PR #3037, opened 2023-07-07
Skip flaky error log check until there is a new snapshot of Beats available.

cmacknz:
We need https://github.com/elastic/beats/pull/36006 available in a snapshot build of Beats to fix the flakiness, until that happens in the next few days just skip the flaky part of the test.


cmacknz:
Requires https://github.com/elastic/elastic-agent/pull/3036 first

cmacknz:
We don't need this on main anymore, there was a successful snapshot today.

PR #3052, opened 2023-07-10
add exception for unit keyword

jlind23:
Fixing sonarcloud coverage issue with `-run Unit` option

### DIFF
--- a/dev-tools/mage/gotest.go
+++ b/dev-tools/mage/gotest.go
@@ -233,7 +233,7 @@ func GoTest(ctx context.Context, params GoTestArgs) error {
 		)
 	}
 
-	if params.TestName != "" {
+	if params.TestName != "" && params.TestName != "Unit" {
 		testArgs = append(testArgs, "-run", params.TestName)
 	}
 


cmacknz:
I see what happened here, Tiago's PR re-used the `TestName` parameter as the argument to the `-run` arg of `go test`. The problem is that the original code intended this to only be use for logging purposes, see the comment in:

https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/dev-tools/mage/gotest.go#L31

The `mage unitTest` target used by sonarcloud and CI is implemented by:

https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/magefile.go#L330-L335

The default Go test args unconditionally set the test name to `Unit`:
https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/dev-tools/mage/gotest.go#L90

The `makeGoTestArgs` function called here takes the input as the name of the test for use in logging and sets `TestName`:
https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/dev-tools/mage/gotest.go#L49

The net effect is that `mage unitTest` is now equivalent to `go test -run Unit ./...` which causes us to skip every unit test. We have no test coverage because we aren't running any unit tests.

The tests in the PRs "pass" in the sense that they aren't running. `mage unitTest` now takes 7s to complete and logs it isn't running any tests, while `go test ./...` to run all tests takes almost 3 minutes on my machine and is obviously running tests.

cmacknz:
Closing, actual fix is in https://github.com/elastic/elastic-agent/pull/3053

PR #3053, opened 2023-07-11
Fix mage unitTest to actually run tests

cmacknz:
Follow up from https://github.com/elastic/elastic-agent/pull/3052

A previous PR re-used the `TestName` parameter as the argument to the `-run` arg of `go test`. The problem is that the original code intended this to only be use for logging purposes, see the comment in:

https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/dev-tools/mage/gotest.go#L31

The `mage unitTest` target used by sonarcloud and CI is implemented by:

https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/magefile.go#L330-L335

The default Go test args unconditionally set the test name to `Unit`:
https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/dev-tools/mage/gotest.go#L90

The `makeGoTestArgs` function called here takes the input as the name of the test for use in logging and sets `TestName`:
https://github.com/elastic/elastic-agent/blob/d20b1085a3c8ba92ff3f7401e363b34f21d97f3e/dev-tools/mage/gotest.go#L49

The net effect is that `mage unitTest` is now equivalent to `go test -run Unit ./...` which causes us to skip every unit test. We have no test coverage because we aren't running any unit tests.

The tests in the PRs "pass" in the sense that they aren't running. `mage unitTest` now takes 7s to complete and logs it isn't running any tests, while `go test ./...` to run all tests takes almost 3 minutes on my machine and is obviously running tests.

I have fixed this by adding a separate `RunExpr` variable to do what the original change intended and also renamed `TestName` to run `LogName` to more clearly indicate its purpose.

### DIFF
--- a/dev-tools/mage/gotest.go
+++ b/dev-tools/mage/gotest.go
@@ -29,6 +29,7 @@ import (
 // how "go test" is invoked. "go test" is always invoked with -v for verbose.
 type GoTestArgs struct {
 	TestName            string            // Test name used in logging.
+	RunExpr             string            // Expression to pass to the -run argument of go test.
 	Race                bool              // Enable race detector.
 	Tags                []string          // Build tags to enable.
 	ExtraFlags          []string          // Extra flags to pass to 'go test'.
@@ -233,8 +234,8 @@ func GoTest(ctx context.Context, params GoTestArgs) error {
 		)
 	}
 
-	if params.TestName != "" {
-		testArgs = append(testArgs, "-run", params.TestName)
+	if params.RunExpr != "" {
+		testArgs = append(testArgs, "-run", params.RunExpr)
 	}
 
 	testArgs = append(testArgs, params.ExtraFlags...)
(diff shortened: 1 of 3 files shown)


cmacknz:
The test failures here are real and don't reproduce with Go 1.19.10, but do reproduce with 1.20.5. We didn't catch these in the Go upgrade PR because the unit tests didn't run because of the bug being fixed here.

jlind23:
buildkite test it

amitkanfer:
how do we prevent this from happening again?

pchila (approved):
Aside from the linter problem (not part of your change but since the file has been touched it appears in this PR) the rest of the change LGTM

jlind23:
> how do we prevent this from happening again?

I believe that setting Sonarcloud as being a mandatory check would have at least forced us to look at the coverage results.

cmacknz:
Yes, sonarcloud found this so making that mandatory would have prevented this.

Issue #3069, opened 2023-07-12
TestRemovePath is flaky


faec:
## Flaky Test

* **Test Name:** `TestRemovePath`
* **Link:** https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/install/uninstall_windows_test.go
* **Branch:** `main`
* **Artifact Link:** https://fleet-ci.elastic.co/blue/organizations/jenkins/elastic-agent%2Felastic-agent-mbp/detail/PR-2781/30/pipeline https://fleet-ci.elastic.co/blue/organizations/jenkins/elastic-agent%2Felastic-agent-mbp/detail/PR-3067/2/pipeline
* **Notes:** There are at least two different types of failures, one seemingly from trying to remove the path while it is in use, and one that fails to execute the test program at all because it is identified as malicious.

### Stack Trace

Failure mode 1:

```
[2023-07-12T17:31:47.445Z] === FAIL: internal/pkg/agent/install TestRemovePath (2.96s)
[2023-07-12T17:31:47.445Z]     uninstall_windows_test.go:54: 
[2023-07-12T17:31:47.445Z]         	Error Trace:	C:/Users/jenkins/workspace/_agent_elastic-agent-mbp_PR-3067/src/github.com/elastic/elastic-agent/internal/pkg/agent/install/uninstall_windows_test.go:54
[2023-07-12T17:31:47.445Z]         	Error:      	Received unexpected error:
[2023-07-12T17:31:47.445Z]         	            	failed to open handle for "C:\\Users\\jenkins\\AppData\\Local\\Temp\\TestRemovePath3514517768\\001\\subdir\\main.exe": The process cannot access the file because it is being used by another process.
[2023-07-12T17:31:47.445Z]         	Test:       	TestRemovePath
```

Failure mode 2:

```
[2023-07-12T13:23:48.438Z] === FAIL: internal/pkg/agent/install TestRemovePath (2.58s)
[2023-07-12T13:23:48.438Z]     uninstall_windows_test.go:47: 
[2023-07-12T13:23:48.438Z]         	Error Trace:	C:/Users/jenkins/workspace/_agent_elastic-agent-mbp_PR-2781/src/github.com/elastic/elastic-agent/internal/pkg/agent/install/uninstall_windows_test.go:47
[2023-07-12T13:23:48.438Z]         	Error:      	Received unexpected error:
[...]

PR #3070, opened 2023-07-12
Disable flaky test `TestRemovePath`

faec:
Skips `TestRemovePath` which is causing CI failures, see https://github.com/elastic/elastic-agent/issues/3069


cmacknz:
Instead of disabling the test lets just revert the change that is causing it, which is recent and is https://github.com/elastic/elastic-agent/commit/b5b805a93dc6a93e9036cab25122a3c5504ef670. There is a chance this is a real bug and not just a flaky test.

I'll forward port https://github.com/elastic/elastic-agent/pull/3056 to main.

Closing this one.

PR #3073, opened 2023-07-13
Integration test - Fetcher must collect SHAs

pchila:
## What does this PR do?

This PR adds support for fetchers to handle `sha512` hashes so that the elastic agent artifacts can be used as version to upgrade to using a `--sourceURI` flag (that will be in a follow-up PR) 

## Why is it important?
This change open up some upgrade test scenarios where we can use the local package on disk instead of waiting for it to be available through artifact API

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...


pchila:
@blakerouse could you have a look at this?
It slightly changes the runner signature for the `Prepare()` function and it may impact the work on the windows runner slightly...

https://github.com/elastic/elastic-agent/blob/7dfff4e81a286b0e005cc99f225567de5cefea69/pkg/testing/runner/runner.go#L55

pchila:
buildkite test this

pkg/testing/fetcher_artifact.go
```diff
@@ -78,16 +78,18 @@ func (f *artifactFetcher) Fetch(ctx context.Context, operatingSystem string, arc
...
 }
 
 type artifactResult struct {
-	doer httpDoer
-	src  string
-	path string
+	doer      httpDoer
+	src       string
+	path      string
+	fetchHash bool
```
ycombinator, line 92: AFAICT, `fetchHash` will always be `true`.  So do we even need this field?  In other words, should we always just fetch the hash? I don't think it adds too much additional cost or risk?

pkg/testing/fetcher_local.go
```diff
@@ -69,12 +69,13 @@ func (f *localFetcher) Fetch(_ context.Context, operatingSystem string, architec
...
-	return &localFetcherResult{src: f.dir, path: build}, nil
+	return &localFetcherResult{src: f.dir, path: build, fetchHash: true}, nil
 }
 
 type localFetcherResult struct {
-	src  string
-	path string
+	src       string
+	path      string
+	fetchHash bool
```
ycombinator, line 78: Same question/comment as https://github.com/elastic/elastic-agent/pull/3073/files#r1262835426.

pkg/testing/fetcher_artifact.go
pchila: I wasn't sure if we want to use fetchers to get other artifacts beyond elastic-agent and so I didn't know if all kind of artifacts always have the hash in the same format. It could be premature optimization here, as you said right now this is always true

pkg/testing/fetcher_local.go
pchila: https://github.com/elastic/elastic-agent/pull/3073#discussion_r1263425066

ycombinator (approved):
LGTM.

pchila:
buildkite test this

Issue #2852, opened 2023-06-13
Agent discards unhealthy state because of unrelated changes


faec:
Agent's `Coordinator` is the controller that supervises the various managers and aggregates their state for Fleet and internal/test listeners. Currently, there are four events that can cause `Coordinator`'s internal state to become unhealthy [there are other events that can cause Coordinator to _report_ an unhealthy state to external listeners, but those aren't relevant here]:

- Receiving a new configuration from the config manager that is invalid, or that couldn't be applied or acknowledged
- Receiving a variable update from the vars manager that is invalid or couldn't be applied
- Receiving a log level update that couldn't be applied (which can only happen if there is already a problem with the config and/or vars)
- One of the managers (runtime, config, or vars) terminating before the shutdown signal is sent

Once the internal state becomes unhealthy, the only thing that can make it healthy again is successfully applying a policy change to the runtime manager, which requires successful generation of the component model from the config, vars, and log level, followed by a successful `Update` call to the runtime manager.

The scenario that causes a problem is this: suppose the Coordinator is healthy, with a successfully generated component model and an updated runtime manager. Now suppose a new policy comes in that is invalid, and can't be processed into a component model. `Coordinator` will transition to an error state but will leave its existing internal config and components unchanged. Now suppose an unrelated change comes in on the vars manager (there are various vars listeners that will trigger updates based on ambient context, e.g. applying environment changes from Kubernetes or Docker). The new vars are combined with the original (non-error) config to produce a successful component model, which the runtime manager receives in an `Update` call, which may succeed (it may not even have to change anything). Because of this call, `Coordinator` sets itself as healthy, and any state related to the failed policy change is lost -- if this happens in a short time span, it may not even be obvious that the policy change has been discarded.

We should figure out a more stable error state model, that can separate the immediate running state of the `Coordinator` (i.e. "I have a working component model and it's running ok") from global error state ("the last policy change was discarded because of an error").

cmacknz:
I am tentatively putting this into the next sprint, because this can lead to the agent appearing healthy when it hasn't actually applied the latest policy successfully.

I would expect we are most likely to observe this in Kubernetes 1) because the agent is more likely to be standalone and thus more likely to encounter manually written invalid policies and 2) more likely to be receiving updates from the vars manager from the k8s providers.

> We should figure out a more stable error state model, that can separate the immediate running state of the Coordinator (i.e. "I have a working component model and it's running ok") from global error state ("the last policy change was discarded because of an error").

[...]

PR #3076, opened 2023-07-13
Add / report persistent policy error metadata in Coordinator

faec:
Update Coordinator's error tracking and aggregate state reporting:
- Add standalone `CoordinatorState` / `CoordinatorMessage` fields to `coordinator.State` ~~and `proto.StateResponse`~~, so Coordinator has its own state instead of overloading the global Agent state.
- Add new error fields to `Coordinator` tracking the three types of errors a policy update can encounter (initial parsing, component model generation, and runtime update).
- Update `(*Coordinator).generateReportableState` to aggregate the new error/state fields into the global `State`/`Message` reported by Fleet.

Fixes https://github.com/elastic/elastic-agent/issues/2852.

## Checklist


control_v2.proto
```diff
@@ -164,19 +164,28 @@ message StateAgentInfo {
...
-  // Overall state message of Elastic Agent.
   string message = 3;
-  // State of each component in Elastic Agent.
-  repeated ComponentState components = 4;
-  // Fleet connectivity state of Elastic Agent.
+
+  // State + message for the Agent Coordinator, reporting manager-level errors
+  // like failed policy updates.
+  State coordinatorState = 7;
+  string coordinatorMessage = 8;
```
blakerouse, line 180: I do not like the idea of exposing an internal detail like the coordinator to the status output.

faec: Well, I expect it wouldn't be user-facing unless we change the UI, and it might be nice to have a way to check it when debugging... but I wouldn't mind being conservative about what we add to the proto in this pass. The most important part of Coordinator state was reporting manager errors and similar, which this PR replaces with explicit aggregation.

What if we break the implicit sync between coordinator state and the proto? Leave the proto unchanged, but keep the coordinator state/message in the internal struct? That way it can still be persisted / aggregated into the reported value, but without any explicit new fields over the wire?

faec:
buildkite test this

michalpristas (approved):
small comments here and there
tested locally scenario
- install agent with system integration logs/metrics enabled
- unplug cable 
- check fleet status is FAILED
- reconnect
- check status is back to green
- removed log collection from system integration
- rename filebeat executable
- add logs back and check status is failed
- rename filebeat back to filebeat
- check it's green

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -214,11 +214,26 @@ type Coordinator struct {
 	// into the reported state before broadcasting -- State() will report
 	// agentclient.Failed if one of these is set, even if the underlying
 	// coordinator state is agentclient.Healthy.
-	runtimeMgrErr error
+	runtimeMgrErr error // Currently unused
```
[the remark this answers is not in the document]
faec: Fair question -- I asked @blakerouse a while ago and he said it was so the runtime manager satisfies the `Runner` interface. A more immediate answer is that it will probably be used soon: I have a PR in progress to centralize more of the runtime manager logic in its own run loop. For that, it's useful to decouple runtime manager `Update` calls from the error result they produce, to avoid accessing runtime internals from the Coordinator goroutine. In that PR I've started using the previously-idle error channel to report component model update failures, and those end up being stored in this field.

faec:
jenkins test this

faec:
jenkins test this

Issue #82, opened 2022-02-07
Update Elastic Agent docker image for azure containers runtime


jlind23:
**Describe the enhancement:**
Elastic Agent docker image is currently not working out of the box for azure container runtime

Docker-entrypoint is executed but it cannot launch elastic-agent because it doesn't find the file.
And it does not find the file because it does not have enough rights to do so.

This is due to the fact that Elastic Agent user is part of the root user group but somehow Azure container runtime has some security restriction.

This is the solution provided so far by @eedugon :
```
FROM docker.elastic.co/beats/elastic-agent:7.16.2
USER root
RUN chown -R :elastic-agent /usr/share/elastic-agent
USER elastic-agent
```

We should ensure that this fix wouldn't break anything related to openshift.

More information can be found here: elastic/elastic-agent#147

**Describe a specific use case for the enhancement or feature:**

narph:
Spent too much time on this one, just to circle back, when running:
```
az container create --resource-group group1  --name elastic-agent --image docker.elastic.co/beats/elastic-agent:7.16.2  --restart-policy Never --ip-address Public  --environment-variables FLEET_ENROLLMENT_TOKEN='...' FLEET_ENROLL='1' FLEET_URL='...'
```
it fails creating the container with:

```
Start streaming logs:
[WARN  tini (19)] Tini is not running as PID 1 and isn't registered as a child subreaper.
Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
/usr/local/bin/docker-entrypoint: line 11: exec: elastic-agent: not found

```

I tried to create a small reproduction to validate `useradd` refuses to add the `elastic-agent` user to `root` group as claimed here: https://github.com/elastic/sdh-beats/issues/1603#issuecomment-1012408050. 

```DockerFile
# Prepare home in a different stage to avoid creating additional layers on
# the final image because of permission changes.
FROM ubuntu:20.04 AS home

RUN mkdir -p /usr/share/elastic-agent/data /usr/share/elastic-agent/data/elastic-agent-4a0299/logs
RUN touch /usr/share/elastic-agent/test-file.txt

FROM ubuntu:20.04

COPY --chown=elastic-agent:root --from=home /usr/share/elastic-agent /usr/share/elastic-agent

# Elastic Agent needs group permissions in the home itself to be able to
[...]

PR #3084, opened 2023-07-14
Agent owns its directories

michalpristas:
## What does this PR do?

This changes owner of agent directories. Tested locally and on azure container instances, seems to work. 

## Why is it important?

Containers running on Azure Container Instances refuse to read config.

## Checklist

### How to test

- build docker image
- push to observability-ci container repo
- create Azure (portal.azure.com login using SSO) Container Instance using published repo (you need to specify Other-Private docker registry with login and pass capture from docker-auth.elastic.co)
- see it runs

Closes: https://github.com/elastic/elastic-agent/issues/82

### DIFF
--- a/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
+++ b/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
@@ -9,7 +9,6 @@ FROM {{ .buildFrom }} AS home
 COPY beat {{ $beatHome }}
 
 RUN mkdir -p {{ $beatHome }}/data {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/logs && \
-    chown -R root:root {{ $beatHome }} && \
     find {{ $beatHome }} -type d -exec chmod 0755 {} \; && \
     find {{ $beatHome }} -type f -exec chmod 0644 {} \; && \
     find {{ $beatHome }}/data -type d -exec chmod 0770 {} \; && \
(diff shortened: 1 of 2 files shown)


dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -144,8 +143,10 @@ true
 RUN mkdir /app
 {{- end }}
 {{- else }}
-RUN groupadd --gid 1000 {{ .BeatName }}
-RUN useradd -M --uid 1000 --gid 1000 --groups 0 --home {{ $beatHome }} {{ .user }}
+RUN groupadd --gid 1000 {{ .BeatName }} && \
+    useradd -M --uid 1000 --gid 1000 --groups 0 --home {{ $beatHome }} {{ .user }} && \
+    chown -R {{ .user }}:{{ .user }} {{ $beatHome }} && \
+    true
```
pchila, line 149: Why do you need a `&& true` at the end ? :thinking:

pchila (approved):
LGTM

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
michalpristas: convention & readability, easier to add other commands later without messing with `&& \` at the end

jlind23:
buildkite test it

111andre111:
@michalpristas 
I tested now with the changes und Docker Desktop as well as Azure and it was working.
Some household questions:
1. Is this one still needed?
https://github.com/elastic/elastic-agent/blob/bbede6e6cbe20932052e9a350d35567c77d13652/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl#L29
Given we later do nevertheless a chown I assume that particular step is not needed.
2. Is there the `--groups 0` still needed here:
https://github.com/elastic/elastic-agent/blob/bbede6e6cbe20932052e9a350d35567c77d13652/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl#L147
Given we don't have root group nevertheless inherited from that command like mentioned in the original issue, is there any need to add this user to supplementary group root?

michalpristas:
first one is needed because beats have a check to see that only root can change the config, some security measure.

second one is needed because without it npm install and check for synthetics fails. i tried to work around it but i was not successful. there's an npm install command with || option to check `/root/something` and without this npm install fails and this check fails as well because `/root`

111andre111:
@michalpristas 
Thanks for these backgrounds. These male semse.
Now the `/root` this is probably happening because of `npm i -g`
https://stackoverflow.com/questions/18088372/how-to-npm-install-global-not-as-root
The location could be probably changed by `npm config set prefix` but I don't want to mess around too much. Thanks for the explanation.
Just to raise attention I still found this issue once agent is working:
https://github.com/elastic/elastic-agent/issues/82#issuecomment-1638233164

Issue #3077, opened 2023-07-13
Agent fails to uninstall endpoint successfully when handling the Unenroll action


cmacknz:
The agent fails to uninstall endpoint when handling the unenroll action, which is sent when a user unenrolls the agent from Fleet without selecting the force unenroll option (which just revokes API keys). 

The uninstall path here is starts from processing the unenroll action which generates an empty agent policy such that all inputs are uninstalled:

https://github.com/elastic/elastic-agent/blob/123579115ad7a330779c2f9b5b9e719f98282dc4/internal/pkg/agent/application/actions/handlers/handler_action_unenroll.go#L55-L57

When this happens the agent keeps running and the Endpoint service directory is left on disk (at /Library/Elastic/Endpoint on a Mac), and the agent logs show the `endpoint-security` command exiting and immediately failing with a context cancelled error:

```json
{"log.level":"info","@timestamp":"2023-07-13T19:09:26.085Z","log.logger":"component.runtime.endpoint-default.service_runtime","log.origin":{"file.name":"runtime/service.go","file.line":216},"message":"uninstall endpoint service","ecs.version":"1.6.0"}
{"log.level":"warn","@timestamp":"2023-07-13T19:09:26.085Z","log.logger":"component.runtime.endpoint-default.service_runtime","log.origin":{"file.name":"runtime/service_command.go","file.line":191},"message":"service command execution failed with error [failed starting the command: failed to start \"/opt/Elastic/Agent/data/elastic-agent-123579/components/endpoint-security\": context canceled], retrying (will be retry [1]) after [10.798795076s]","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-07-13T19:09:26.085Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":1187},"message":"Component state changed endpoint-default (HEALTHY->STOPPED): Stopped: endpoint service runtime","log":{"source":"elastic-agent"},"component":{"id":"endpoint-default","state":"STOPPED","old_state":"HEALTHY"},"ecs.version":"1.6.0"}
[...]

Issue #3060, opened 2023-07-12
Uninstall fails if Elastic Defend integration is installed


cmacknz:
On v8.9-0-SNAPSHOT and 8.10.0-SNAPSHOT if the Elastic Defend integration is installed, the `elastic-agent uninstall` command will fail every time.

This bug was originally found in https://github.com/elastic/elastic-agent/pull/2963 on Linux GCP VMs but also reproduces locally on my M1 Mac.

```sh
sudo elastic-agent uninstall
Elastic Agent will be uninstalled from your system at /Library/Elastic/Agent. Do you want to continue? [Y/n]:y
Error: failed to remove installation directory (/Library/Elastic/Agent): unlinkat /Library/Elastic/Agent: directory not empty
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.10/fleet-troubleshooting.html
```

Looking at what is left in the Elastic Agent directory I see the following log files:

```
sudo ls /Library/Elastic/Agent
elastic-agent-20230711-1.ndjson elastic-agent-20230711-2.ndjson
```

This does not reproduce with the 8.8.2 release.

cmacknz:
The contents of the logs files are:

```sh
sudo cat /Library/Elastic/Agent/elastic-agent-20230711-1.ndjson
```

```json
{"log.level":"error","@timestamp":"2023-07-11T22:25:34.512-0400","log.origin":{"file.name":"runtime/service_command.go","file.line":76},"message":"2023-07-12 02:25:34: info: MainPosix.cpp:235 Executing uninstall","context":"command output","ecs.version":"1.6.0"}
```

```sh
sudo cat /Library/Elastic/Agent/elastic-agent-20230711-2.ndjson
```

```json
{"log.level":"error","@timestamp":"2023-07-11T22:25:34.512-0400","log.origin":{"file.name":"runtime/service_command.go","file.line":76},"message":"2023-07-12 02:25:34: info: Internal.cpp:51 Found config path [/Library/Elastic/Endpoint/elastic-endpoint.yaml]","context":"command output","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2023-07-11T22:25:34.513-0400","log.origin":{"file.name":"runtime/service_command.go","file.line":76},"message":"2023-07-12 02:25:34: info: InstallLib.cpp:719 Skipping uninstall token validation as tamper protection is not enabled.","context":"command output","ecs.version":"1.6.0"}
[...]

PR #2889, opened 2023-06-16
[Service Runtime] Add ability for service operations to be retried

ycombinator:
## What does this PR do?

This PR adds the ability for service operations to be retried.  Service components can optionally specify retries for their operations in their spec files like so:

```yml
version: 2
inputs:
  - name: endpoint
    service:
      operations:
        install:
          retry:
            init_interval: 60s
```

With the above spec definition, the `install` operation of the `endpoint` service will be retried indefinitely, with the duration between retries will start at 60 seconds and grow exponentially (with some random jitter). 

If `retry_sleep_init_duration` is omitted, it will default to **20 seconds**.

The duration between retries will grow exponentially (with some random jitter) but will be capped at approximately **15 minutes**.

## Why is it important?

In cases where a service operation may be flaky due to transient issues, it makes sense to retry the operation  indefinitely, giving it an opportunity to succeed.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

PR #3086, opened 2023-07-14
Fix endpoint unenroll test

cmacknz:
Adds an integration test for the case where endpoint-security is uninstalled because it was removed from the policy. Fixes a bug where the uninstall operation in this case would silently fail, which is detectable because the Endpoint directory would still exist next to the Agent directory in the install path.

This is basically the same as https://github.com/elastic/elastic-agent/issues/3060 but triggered on a different path. I also solved it slightly differently, in that the agent will retry the uninstall but only for a maximum 1 hour period. This is because I don't think there's a way to cancel an uninstall that is retrying forever since endpoint is already removed from the policy. I didn't test this explicitly, but I wanted to guard against it regardless (and we've never seen uninstall fail in practice so the retries probably aren't even necessary). We also have no way to deal with a failed unenroll attempt but that's a separate problem.

- Closes https://github.com/elastic/elastic-agent/issues/3077

### DIFF
--- a/pkg/component/runtime/service.go
+++ b/pkg/component/runtime/service.go
@@ -219,7 +219,7 @@ func (s *serviceRuntime) stop(ctx context.Context, comm Communicator, lastChecki
 	}
 
 	// Force component stopped state
-	s.log.Debug("set %s service runtime to stopped state", name)
+	s.log.Debugf("set %s service runtime to stopped state", name)
 	s.forceCompState(client.UnitStateStopped, fmt.Sprintf("Stopped: %s service runtime", name))
 }
 
--- a/pkg/component/runtime/service.go
+++ b/pkg/component/runtime/service.go
@@ -219,7 +219,7 @@ func (s *serviceRuntime) stop(ctx context.Context, comm Communicator, lastChecki
 	}
 
 	// Force component stopped state
-	s.log.Debugf("set %s service runtime to stopped state", name)
+	s.log.Debug("set %s service runtime to stopped state", name)
 	s.forceCompState(client.UnitStateStopped, fmt.Sprintf("Stopped: %s service runtime", name))
 }
 
@@ -448,11 +448,6 @@ func (s *serviceRuntime) install(ctx context.Context) error {
 
 // uninstall executes the service uninstall command
 func (s *serviceRuntime) uninstall(ctx context.Context) error {
-	if s.comp.InputSpec.Spec.Service.Operations.Uninstall == nil {
-		s.log.Errorf("missing uninstall spec for %s service", s.comp.InputSpec.BinaryName)
-		return ErrOperationSpecUndefined
-	}
-
 	// Always retry for internal attempts to uninstall, because they are an attempt to converge the agent's current state
 	// with its desired state based on the agent policy.
 	return uninstallService(ctx, s.log, s.comp, s.executeServiceCommandImpl)
@@ -468,10 +463,6 @@ func uninstallService(ctx context.Context, log *logger.Logger, comp component.Co
 		log.Errorf("missing uninstall spec for %s service", comp.InputSpec.BinaryName)
 		return ErrOperationSpecUndefined
 	}
-
-	// Do not use infinite retries when uninstalling from the command line. If the uninstall needs to be
-	// retried the entire uninstall command can be retried. Retries may complete asynchronously with the
-	// execution of the uninstall command, leading to bugs like https://github.com/elastic/elastic-agent/issues/3060.
 	log.Debugf("uninstall %s service", comp.InputSpec.BinaryName)
 	return executeServiceCommandImpl(ctx, log, comp.InputSpec.BinaryPath, comp.InputSpec.Spec.Service.Operations.Uninstall)
 }
--- a/pkg/component/runtime/service.go
+++ b/pkg/component/runtime/service.go
@@ -448,6 +448,11 @@ func (s *serviceRuntime) install(ctx context.Context) error {
 
 // uninstall executes the service uninstall command
 func (s *serviceRuntime) uninstall(ctx context.Context) error {
+	if s.comp.InputSpec.Spec.Service.Operations.Uninstall == nil {
+		s.log.Errorf("missing uninstall spec for %s service", s.comp.InputSpec.BinaryName)
+		return ErrOperationSpecUndefined
+	}
+
 	// Always retry for internal attempts to uninstall, because they are an attempt to converge the agent's current state
 	// with its desired state based on the agent policy.
 	return uninstallService(ctx, s.log, s.comp, s.executeServiceCommandImpl)
@@ -463,6 +468,10 @@ func uninstallService(ctx context.Context, log *logger.Logger, comp component.Co
 		log.Errorf("missing uninstall spec for %s service", comp.InputSpec.BinaryName)
 		return ErrOperationSpecUndefined
 	}
+
+	// Do not use infinite retries when uninstalling from the command line. If the uninstall needs to be
+	// retried the entire uninstall command can be retried. Retries may complete asynchronously with the
+	// execution of the uninstall command, leading to bugs like https://github.com/elastic/elastic-agent/issues/3060.
 	log.Debugf("uninstall %s service", comp.InputSpec.BinaryName)
 	return executeServiceCommandImpl(ctx, log, comp.InputSpec.BinaryPath, comp.InputSpec.Spec.Service.Operations.Uninstall)
 }
--- a/pkg/component/runtime/service_command.go
+++ b/pkg/component/runtime/service_command.go
@@ -49,7 +49,7 @@ func executeCommand(ctx context.Context, log *logger.Logger, binaryPath string, 
 
 	proc, err := process.Start(binaryPath, opts...)
 	if err != nil {
-		return fmt.Errorf("failed starting the command with args %s: %w", args, err)
+		return fmt.Errorf("failed starting the command: %w", err)
 	}
 
 	// channel for the last error message from the stderr output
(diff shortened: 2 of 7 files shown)


ycombinator:
buildkite test this

changelog/fragments/1689375911-fix-endpoint-unenroll-test.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Fix bug where Endpoint Security is not being uninstalled when Agent is unenrolled from Fleet
```
cmacknz, line 14: Don't need a changelog entry for this because it was never release, it only exists in the 8.9.0 branch. 8.8.2 is unaffected.

ycombinator: Ah yes, good point — the bug hasn't been released yet! Will remove this file. Thanks.

pkg/component/runtime/service_command.go
```diff
@@ -168,18 +175,14 @@ func (cr *cmdRetrier) Start(
 
 	backoffCtx := backoff.WithContext(expBackoff, retryCtx)
 
-	// Since we will be executing the command with infinite retries, we don't
-	// want to block.  So we execute the command with retries in its own
-	// goroutine.
-	go func() {
+	retryFunc := func() error {
```
cmacknz, line 178: If you end up having to change the code anyway, Aleks pointed out that the contexts created here are only cancelled when `Stop` is called but not on the success path.

I tried a quick fix but one of the tests failed, and then I ran out of time to investigate.

ycombinator:
The `TestFQDN` test (unrelated to this PR) is failing in CI. Re-running...

ycombinator:
buildkite test this

ycombinator:
CI failed due to buildkite pipeline upload failure. Re-running...

ycombinator:
buildkite test this

ycombinator:
buildkite test this

cmacknz:
Disable or skip the failing parts of the test, the number of re-runs is getting silly.

ycombinator:
> Disable or skip the failing parts of the test, the number of re-runs is getting silly.

Good idea.  Are you suggesting disabling these parts on `main` permanently (and then rebasing this PR on `main`) or just as part of this PR, temporarily, until CI is green on it?

cmacknz:
Solve this for every PR, skip on main and then rebase. We can work out how to fix the tests without everyone needing to play the CI test lottery to merge.

aleksmaus:
> Fixes a bug where the uninstall operation in this case would silently fail, which is detectable because the Endpoint directory would still exist next to the Agent directory in the install path.

Could you explain the details how this uninstall bug is fixed with this change?

pkg/component/runtime/service.go
```diff
@@ -435,36 +443,64 @@ func (s *serviceRuntime) check(ctx context.Context) error {
...
 		s.log.Errorf("missing install spec for %s service", s.comp.InputSpec.BinaryName)
 		return ErrOperationSpecUndefined
 	}
 	s.log.Debugf("install %s service", s.comp.InputSpec.BinaryName)
-	return s.executeServiceCommandImpl(ctx, s.log, s.comp.InputSpec.BinaryPath, s.comp.InputSpec.Spec.Service.Operations.Install, true)
+
+	// Install with infinite retries, don't block and execute asynchronously.
+	return s.executeServiceCommandWithRetriesImpl(
+		ctx, s.log, s.comp.InputSpec.BinaryPath, s.comp.InputSpec.Spec.Service.Operations.Install,
+		context.Background(), defaultRetryInitInterval, defaultRetryMaxInterval, false)
```
aleksmaus, line 463: would there be a race condition if the state changes, and there is the go routine running for an hour retrying install?

aleksmaus:
> Therefore, if the uninstall operation fails, it no longer can fail silently.

the issue that we saw was that the uninstall command executor code was failing right away because it was receiving the  cancelled context for uninstall. how does this change address that?

pkg/component/runtime/service.go
aleksmaus: I mean, is it possible that while the retrier install/retrier go routine runs (waits for the next retry( there will be a legitimate state change on the agent that will get Endpoint uninstalled, and then this retrier loop kicks in and installs the Endpoint point? The kind of the race condition, when you have potentially long running (1 hour) go routine that can only be stopped with another install command or 1 hour timeout.

ycombinator: Yeah, I see what you mean.  In the scenario you outlined, there will be nothing to cancel the ongoing install command retries so, yeah, I think what you said will happen. 😞 Thanks for bringing it up!

Thinking about how best to fix this.  Perhaps the retrier should cancel _any_ service command for the same service binary before running the new command?

ycombinator:
> the issue that we saw was that the uninstall command executor code was failing right away because it was receiving the cancelled context for uninstall. how does this change address that?

The key is this block of code (its functionality is unchanged before/after this PR): https://github.com/elastic/elastic-agent/blob/50de0fd03c01d4aef2197fb02f6710285f44fb0f/pkg/component/runtime/service.go#L216-L225

Before this PR, the call to `uninstall(ctx)` lead to a non-blocking call to the retrier/executor for the uninstall command.  This means that the soon after this call was made, a few lines later in the above code block, the endpoint component was put into a STOPPED state.  This, then cancelled the context (`ctx`) that was passed to `uninstall` and further down all the way into the command executor.  

After this PR, the call to `uninstall(ctx)` leads to a blocking call to the retrier, which means until the retrier succeeds, the endpoint component will not be put into a STOPPED state, and the context (`ctx`) will not be cancelled.  But since we don't want the uninstall command to execute forever, we pass in a context with a 1 hour timeout (`safetyCtx`) now: https://github.com/cmacknz/elastic-agent/blob/fc00847fe0f3fc929278f9f11176099a94d12b3f/pkg/component/runtime/service.go#L478-L488

One thing that I'm not sure of is whether the `safetyCtx` should apply to the command itself, or to the retrier, or both.  As implemented in this PR right now, it applies to the command itself.  For the retrier, we are passing a `context.Background()`. My feeling is that it should apply to both — the command and the retrier — as we'd probably want either to stop within an hour.  @cmacknz @aleksmaus WDYT?

ycombinator:
Chatted with @aleksmaus off-PR.  His suggestion is to close this PR unmerged and revert the original non-blocking, indefinite retries implementation PR (https://github.com/elastic/elastic-agent/pull/2889).  I'm inclined to agree, given that that change has introduced the bug that this PR here tries to fix and has the potential to introduce other bugs along the critical install and uninstall paths for Endpoint. 

We should probably step back and consider having blocking retries for all the service operations (as opposed to non-blocking retries, as implemented in #2889).  It would greatly simplify the implementation but also probably reduce the likelihood of introducing bugs due to the unforeseen effects of non-blocking/async code paths.

So I'm going to put up a revert PR for #2889 and, once that's merged and backported to `8.9`, close out this PR here as unmerged.  I'll also reopen the issue associated with #2889.

ycombinator:
> So I'm going to put up a revert PR for #2889 and, once that's merged and backported to `8.9`, close out this PR here as unmerged. I'll also reopen the issue that #2889 tried to fix.

Revert PR: https://github.com/elastic/elastic-agent/pull/3092

pierrehilbert:
@ycombinator the revert has been merged. Should we close this one then?

ycombinator:
> @ycombinator the revert has been merged. Should we close this one then?

As discussed in https://github.com/elastic/elastic-agent/pull/3092#issuecomment-1638794565, the thinking was that we don't close this PR but reduce it to just the integration test and then merge that.  So I will try to do that next.

testing/integration/endpoint_security_test.go
```diff
@@ -87,54 +89,129 @@ func TestEndpointSecurity(t *testing.T) {
...
+		endpointHealthPollingTimeout,
+		time.Second,
+		"All components not removed.",
+	)
+	t.Logf("Verified endpoint component and units are removed")
+
+	// Verify that the Endpoint directory was correctly removed.
+	// Regression test for https://github.com/elastic/elastic-agent/issues/3077
+	agentInstallPath := fixture.WorkDir()
+	files, err := os.ReadDir(filepath.Join(agentInstallPath, ".."))
```
pchila, line 203: Nitpick: since we are going 1 level up from the agent install maybe we want to [clean the path](https://pkg.go.dev/path/filepath#Clean) using `filepath.Clean(filepath.Join(...))` 

We could also use [`filepath.SplitList()`](https://pkg.go.dev/path/filepath#SplitList) + `filepath.Join` removing the last element
I think also `filepath.Dir()` should work here (to be tested)

testing/integration/endpoint_security_test.go
```diff
@@ -169,3 +246,52 @@ func installElasticDefendPackage(t *testing.T, info *define.Info, policyID strin
...
+				if unit.UnitType == client.UnitTypeInput {
+					foundEndpointInputUnit = true
+				}
+				if unit.UnitType == client.UnitTypeOutput {
+					foundEndpointOutputUnit = true
+				}
+			}
+
+			if unit.State != client.Healthy {
+				t.Logf("Unit is not Healthy\n%+v", unit)
```
pchila, line 284: Nitpick: since we are testing the health of all components/units maybe we want to change the function name to something different? something like `agentAndEndpointHealthy` ?

pchila (approved):
Couple of nitpicks, LGTM otherwise

cmacknz:
Thanks for sorting this out! After some thinking I also believe there is a better way to solve this problem, see https://github.com/elastic/ingest-dev/issues/2117#issuecomment-1648365247.

PR #3090, opened 2023-07-17
Support log level setting from policy

### LINKED ISSUE #2851: Configuring agent.logging.level through agent policy do not work for managed agents
## Issue

We now have a feature that allow to add arbitrary key to the agent policy sent to the agent,  https://github.com/elastic/kibana/pull/159414#issuecomment-1587160242

I tried to set agent log level this way to Fleet managed agents, and it seems the agent is still logging at the default log level, when I run `elastic-agent inspect` I saw that my `agent.logging.level: debug` config is here, but it seems not used by the agent

## Definition of done

- [ ] Logging level is taking into account the one in the policy
- [ ] Test confirming the logging level change is working as expected
- [ ] Test covering the logging level change from settings action too
- [ ] Test covering an unexisting logging level target

pchila:
## What does this PR do?

Support setting elastic-agent log level from fleet policy.
The agent will apply (in decreasing order of priority):
1. log level set specifically to the agent via `settings` action, if any
1. log level specified in fleet policy, if any
1. default hard-coded log level for elastic-agent

Whenever a `policy_change` or `settings` action is received, the settings action handler will reevaluate the loglevels specified and set the log level according to the priority above.

## Why is it important?

It allows users to manage elastic-agent verbosity easily through the fleet policy, while at the same time allowing to set a different log level to specific agents for troubleshooting issues.

## Checklist

Create a simple policy in fleet and enroll an elastic-agent with that policy.
The start setting log levels in policy and for the specific agent using dev tools and the requests below

Set policy log level
```http
PUT kbn:/api/fleet/agent_policies/<policy id>
{
   "name": "<policy name>",
   "namespace": "default",
   "overrides": {
       "agent":{
         "logging": {
           "level": "error"
         }
       }
   }
}
```
Set log level for a specific agent
```http
POST kbn:/api/fleet/agents/<elastic agent id>/actions
{
  "action": {
    "type": "SETTINGS",
    "data": {
     "log_level": "debug"   
    }
  }
}
```

A good way to check the effect of the changes of log levels is to keep a terminal with `elastic-agent logs -f` command running: that way we can see the changes in elastic-agent logging in real-time.

Another way to check the current log level currently in use by agent is to use the `inspect` subcommand (grepping or using `yq` is recommended as the output is very verbose, for example:

`sudo elastic-agent inspect | yq .agent`

```yaml
download:
  sourceURI: https://artifacts.elastic.co/downloads/
features: null
headers: null
id: 25bd1b94-9b76-4a2d-bdd8-09d778e8cf44
logging:
  level: warning
monitoring:
  enabled: true
  http:
    buffer: null
    enabled: false
    host: localhost
    port: 6791
  logs: true
  metrics: true
  namespace: default
  use_output: default
protection:
  enabled: false
  signing_key: <redacted>
```

## Related issues

- Closes #2851

## Questions to ask yourself

- ...


internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -104,10 +109,56 @@ func (h *PolicyChangeHandler) Handle(ctx context.Context, a fleetapi.Action, ack
 		return errors.New(err, "could not parse the configuration from the policy", errors.TypeConfig)
 	}
 
+	var configNeedToBeSaved bool
+
 	h.log.Debugf("handlerPolicyChange: emit configuration for action %+v", a)
-	err = h.handleFleetServerHosts(ctx, c)
+	fleetClientCfg, newFleetClient, err := h.handleFleetServerHosts(ctx, c)
```
pchila, line 115: Note to self: Instead of returning the new client, return a simple `func()` callback that can be called after the configuration has been correctly saved. Same thing can be done for handleLogLevel

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -104,10 +109,56 @@ func (h *PolicyChangeHandler) Handle(ctx context.Context, a fleetapi.Action, ack
...
+		configNeedToBeSaved = true
+	}
+
+	newLogLevel, err := h.handleLogLevel(ctx, c)
+	if err != nil {
+		return fmt.Errorf("error handling log level setting: %w", err)
+	}
+
+	if newLogLevel != nil {
+		h.config.Settings.LoggingConfig.Level = *newLogLevel
```
pchila, line 131:
```suggestion
		h.config.Settings.LoggingConfig.Level = *newLogLevel
		configNeedToBeSaved = true
```

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -570,13 +570,16 @@ func (c *Coordinator) PerformComponentDiagnostics(ctx context.Context, additiona
 
 // SetLogLevel changes the entire log level for the running Elastic Agent.
 // Called from external goroutines.
-func (c *Coordinator) SetLogLevel(ctx context.Context, lvl logp.Level) error {
+func (c *Coordinator) SetLogLevel(ctx context.Context, lvl *logp.Level) error {
```
blakerouse, line 573: Why change this to allow a nil value, to then error if a nil value is passed?

You also convert the pointer to value in the function. Seems to me all the changes in this function could be removed and it would have the same result.

blakerouse (commented):
Overall this looks really good. Easy to follow and very well tested.

Would like to see an integration test, testing all the different modes of operations. Control protocol should expose the currently set log level, so should be easy to validate the log information.

internal/pkg/agent/application/coordinator/coordinator.go
pchila: Mostly it is to have a single interface for log level setter and then leaving the check to the implementations...
If I didn't change this signature I would have to define 2 interfaces: one with a pointer (for setting the log level coming from policy which could be cleared) and one with the value implemented by coordinator (which needs to receive a real value)

The signal to noise ration of having 2 interfaces with similar names differing only by a pointer seemed a bit much to me but I still wanted to abstract away the coordinator package so this is the tradeoff I came to.

There's no technical reason why we could not have 2 separate interfaces though...

pchila:
> Would like to see an integration test, testing all the different modes of operations. Control protocol should expose the currently set log level, so should be easy to validate the log information.

@blakerouse I am currently implementing integration tests using `inspect` to check that the log value is correctly set.
Hopefully I will be done by the end of the day :)

internal/pkg/agent/application/coordinator/coordinator.go
blakerouse: No that makes complete sense. I was just confused on the change, but if that unifies it I prefer that.

blakerouse:
> > Would like to see an integration test, testing all the different modes of operations. Control protocol should expose the currently set log level, so should be easy to validate the log information.
> 
> @blakerouse I am currently implementing integration tests using `inspect` to check that the log value is correctly set. Hopefully I will be done by the end of the day :)

You might not need to use `inspect`. Would be easier to use the status output of the control protocol that should include the `agent_info` that has the current log level.

pchila:
> You might not need to use `inspect`. Would be easier to use the status output of the control protocol that should include the `agent_info` that has the current log level.

Depending on how the AgentInfo object is created it may not contain the actual log level as it could be empty in the config/state file but it could be set via policy at runtime when we process the policy_change action... I implemented the integration test with `inspect` and it seems to work pretty well

blakerouse (changes requested):
> > You might not need to use `inspect`. Would be easier to use the status output of the control protocol that should include the `agent_info` that has the current log level.
> 
> Depending on how the AgentInfo object is created it may not contain the actual log level as it could be empty in the config/state file but it could be set via policy at runtime when we process the policy_change action... I implemented the integration test with `inspect` and it seems to work pretty well

That is not good. We should ensure that the output from `elastic-agent status --output=yaml` includes the current active log level. It should not be obscure to the user, it should be very clear. Even if the log level is not set in the agent info context, the output from the `status` command should show the active log level.

pchila:
> > the integration test with `inspect` and it seems to work pretty well
> 
> That is not good. We should ensure that the output from `elastic-agent status --output=yaml` includes the current active log level. It should not be obscure to the user, it should be very clear. Even if the log level is not set in the agent info context, the output from the `status` command should show the active log level.

@blakerouse 

Had a quick look at the `elastic-agent status` command output and I didn't see any information about log level, so I dug deeper into the State() function in our GRPC server
```golang
// State returns the overall state of the agent.
func (s *Server) State(_ context.Context, _ *cproto.Empty) (*cproto.StateResponse, error) {
	state := s.coord.State()
	return stateToProto(&state, s.agentInfo)
}
```
and then in the stateToProto() function
```golang
func stateToProto(state *coordinator.State, agentInfo info.Agent) (*cproto.StateResponse, error) {

// some code omitted here...


	return &cproto.StateResponse{
		Info: &cproto.StateAgentInfo{
			Id:           agentInfo.AgentID(),
			Version:      release.Version(),
			Commit:       release.Commit(),
			BuildTime:    release.BuildTime().Format(control.TimeFormat()),
			Snapshot:     release.Snapshot(),
			Pid:          int32(os.Getpid()),
			Unprivileged: agentInfo.Unprivileged(),
		},
		State:          state.State,
		Message:        state.Message,
		FleetState:     state.FleetState,
		FleetMessage:   state.FleetMessage,
		Components:     components,
		UpgradeDetails: upgradeDetails,
	}, nil
}
```

It turns out that the log level is not part of the `cproto.StateAgentInfo` so it's not output to the user at this point.

To have the correct representation if we decide to add the log level to the `elastic-agent status` output, we need to take the log level from the `coordinator.State` rather than `info.AgentInfo` struct (we pass both structs in ), as the coordinator always have the applied log level wherever it is coming from (policy or agent settings).

Issue #3068, opened 2023-07-12
Integration tests failed to start in Buildike


pazone:
## Flaky Test

Build: https://buildkite.com/elastic/elastic-agent/builds/1696

### Stack Trace
Error:
```
+ mage integration:test
--
  | [Jul 12 15:53:05.992] >>> Pulling latest ogc image
  | [Jul 12 15:53:05.992] >>> Create SSH keys to use for SSH
  | [Jul 12 15:53:05.992] >>> Creating zip archive of repo to send to remote hosts
  | Error: error running test: failed to create zip archive of repo: error adding files: failed to stat file /opt/buildkite-agent/builds/bk-agent-elastic-gcp-1689176529591075389/elastic/elastic-agent/.buildkite/ho: stat /opt/buildkite-agent/builds/bk-agent-elastic-gcp-1689176529591075389/elastic/elastic-agent/.buildkite/ho: no such file or directory
  | + TESTS_EXIT_STATUS=1
```

pazone:
@cmacknz @blakerouse @fearful-symmetry 
I see this path in the error message : `...elastic/elastic-agent/.buildkite/ho`. It's definitely a truncated `elastic/elastic-agent/.buildkite/hooks/...`. 
The error was raised in [archiver.go](https://github.com/elastic/elastic-agent/blob/main/pkg/testing/runner/archiver.go#L80)
It: 
- executes `git ls-files -z` and gets a `\0` separated list of files in the working directory. 
- reads it with `s.Scan()` and `s.Text()` and then somehow gets a truncated line (147 bytes)

We can try to replace the data stream with something simple as `filepath.Walk`. But I'm probably wrong and there is another reason.

cmacknz:
Similar failure here:

```
Error: error running test: failed to create zip archive of repo: error adding files: failed to stat file /opt/buildkite-agent/builds/bk-agent-prod-gcp-1689376416624238497/elastic/elastic-agent/internal/pkg/cli/confirapiKey: stat /opt/buildkite-agent/builds/bk-agent-prod-gcp-1689376416624238497/elastic/elastic-agent/internal/pkg/cli/confirapiKey: no such file or director
```

From https://buildkite.com/elastic/elastic-agent/builds/1787#018956ae-4988-4388-958c-ef7cc15768b5

pazone:
[...]

PR #3091, opened 2023-07-17
Integration tests: fixed couldn't stat file error

pazone:
Figured out with @blakerouse the race condition while reading the output of `git ls-files`

## Related issues
 Relates: https://github.com/elastic/elastic-agent/issues/3068


pazone:
buildklite build this

PR #3096, opened 2023-07-18
Enable and fix race detector for CI

blakerouse:
## What does this PR do?

Enables the race detector for the CI unit tests including fixing the issues detected by enabling the race detector.

## Why is it important?

Ensures that we do not have any data race conditions in the code.

## Checklist


blakerouse:
buildkite test this

internal/pkg/agent/application/dispatcher/dispatcher_test.go
```diff
@@ -248,10 +251,11 @@ func TestActionDispatcher(t *testing.T) {
 		select {
 		case err := <-d.Errors():
 			t.Fatalf("Unexpected error: %v", err)
-		case <-time.After(200 * time.Microsecond):
-			// we're not expecting any reset,
+		case <-calledCh:
+			// Handle was called, expected
```
ycombinator, line 255: Nice replacement of `assert.Eventually`.

internal/pkg/agent/application/dispatcher/dispatcher_test.go
```diff
@@ -248,10 +251,11 @@ func TestActionDispatcher(t *testing.T) {
...
+		case <-time.After(1 * time.Second):
```
ycombinator, line 256: You could probably make this even shorter, given that the `assert.Eventually` before had a timeout of `100*time.Millisecond`?

ycombinator (approved):
LGTM.

internal/pkg/agent/application/dispatcher/dispatcher_test.go
blakerouse: I just like to keep it longer because it returns as soon as it happens. This is just to be sure that in the case that runner is slow that we don't hit that timeout just because of that.

blakerouse: You okay with me leaving it as is?

ycombinator: Yes, OK to leave as-is.

Issue #3001, opened 2023-07-05
Add integration test for an upgrade that falls back to fetching the GPG key from a URL


cmacknz:
https://github.com/elastic/elastic-agent/pull/2980 added a fallback to downloading the GPG key from https://artifacts.elastic.co/GPG-KEY-elastic-agent when verification with the embedded GPG key fails. This is critical functionality in the event of a GPG key rotation. We need to write an integration test that ensures that when verification when the embedded GPG key fails we can successfully verify the upgrade artifact with the key downloaded from the URL.

The primary complication to writing this test will be getting the initial GPG key verification to fail, considering that the default key is embedded into the binary and is still valid. We will need to produce a special build of the agent with an invalid embedded GPG key, or add configuration to alter the GPG precedence order. For example we could have the agent prefer an invalid GPG key from a file to the one that is embedded into it, and then rely on downloading the correct key to have the upgrade succeed.

We used to have configuration to this effect, we could revive that and use it as the basis for this test:

https://github.com/elastic/elastic-agent/blob/f8776d5e9b781dbd334c7a8a2356d57d4ce5e196/elastic-agent.reference.yml#L74-L76

cmacknz:
I should note that we do not need to write an entirely new test for this, we can modify one of the existing upgrade tests to start with an invalid GPG key and it will still test what it did before along with this new functionality.

PR #2955, opened 2023-06-28
Introduce upgrade tests from released versions

pchila:
## What does this PR do?

This PR adds upgrade tests for both standalone and managed agent from some versions of the current and previous major release to the latest snapshot.

The tests will select:
- up to 2 versions with the same major version as the current agent
- up to 1 version from a previous major version
These versions are retrieved using the Artifact API and working backwards through the versions, excluding snapshots.
The _up to_ part is because under certain circumstances (new major release, for example) we may not find enough versions fitting the criteria.

The tests `TestStandaloneUpgrade` and `TestFleetManagedUpgrade` will start a subtest (using `t.Run()`) for each version to be used as a starting point for the upgrade and then upgrade it to the latest version of agent.

Current limitations:
- We are currently using the ArtifactFetcher so the starting elastic agent version all come from the artifact API. For all agent versions that allow the `--skip-verify` we use the packaged elastic agent as artifact for the upgrade via `--sourceURI` parameter
 

## Why is it important?
We need to have automated tests for the common upgrade scenarios to detect upgrade issues as soon as possible in the development cycle. 

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- [x] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## How to test this PR locally

We can run the integration tests using 
`AGENT_VERSION="8.10.0-SNAPSHOT" GOTEST_FLAGS="-test.run ^(TestFleetManagedUpgrade|TestStandaloneUpgrade)$"  mage integration:test`
or if you want to run just a specific version upgrade scenario we can use (for example only standalone upgrade from 8.9.0 to 8.10.0-SNAPSHOT):
`AGENT_VERSION="8.10.0-SNAPSHOT" GOTEST_FLAGS="-test.run ^TestStandaloneUpgrade/Upgrade_8.9.0_to_8.10.0-SNAPSHOT$"`

## Related issues

- Requires #3073 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #3101, opened 2023-07-19
[Integration Test] Upgrade failed default verification test

michalpristas:
## What does this PR do?

Close https://github.com/elastic/elastic-agent/issues/3001

This PR adds an integration test for use case when default option of package verification fails and we need to try remote pgp published at https://artifacts.elastic.co/GPG-KEY-elastic-agent

We test this by introducing a flag `--skip-default-pgp` which disables checks with embedded key and providing corrupted key using `--pgp`.


internal/pkg/agent/application/coordinator/coordinator_test.go
```diff
@@ -447,7 +447,7 @@ func TestCoordinator_Upgrade(t *testing.T) {
 	require.NoError(t, err)
 	cfgMgr.Config(ctx, cfg)
 
-	err = coord.Upgrade(ctx, "9.0.0", "", nil, true)
```
pchila, line 450: Nice little test, it will be fun to see what happens when we will release 9.0 xD

internal/pkg/agent/application/upgrade/artifact/download/http/verifier.go
```diff
@@ -194,7 +196,14 @@ func (v *Verifier) composeURI(filename, artifactName string) (string, error) {
 }
 
 func (v *Verifier) getPublicAsc(sourceURI string) ([]byte, error) {
-	resp, err := v.client.Get(sourceURI)
+	ctx, cancelFn := context.WithTimeout(context.Background(), 30*time.Second)
```
pchila, line 199: I appreciate the usage of a client with a timeout, I am just not sure what triggered this specific change.... :thinking:

testing/integration/upgrade_test.go
```diff
@@ -273,7 +274,165 @@ func TestStandaloneUpgrade(t *testing.T) {
...
+	}, 5*time.Minute, 1*time.Second, "agent never upgraded to expected version")
+
+	checkUpgradeWatcherRan(t, agentFixture)
+
+	version, err := c.Version(ctx)
+	require.NoError(t, err, "error checking version after upgrade")
+	require.Equal(t, expectedAgentHashAfterUpgrade, version.Commit, "agent commit hash changed after upgrade")
+}
+
+func TestStandaloneUpgrade_FailInitialCheck(t *testing.T) {
```
pchila, line 297: I kinda wish that  this new test would be added after PR https://github.com/elastic/elastic-agent/pull/2955 because this will generate more conflicts :sweat:

testing/integration/upgrade_test.go
```diff
@@ -273,7 +274,165 @@ func TestStandaloneUpgrade(t *testing.T) {
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	err = agentFixture.Prepare(ctx)
+	require.NoError(t, err, "error preparing agent fixture")
+
+	err = agentFixture.Configure(ctx, []byte(fastWatcherCfg))
+	require.NoError(t, err, "error configuring agent fixture")
+
+	const minVersionString = "8.9.0-SNAPSHOT"
```
pchila, line 317: shouldn't this version constraint be 8.10.0-SNAPSHOT ?

testing/integration/upgrade_test.go
```diff
@@ -273,7 +274,165 @@ func TestStandaloneUpgrade(t *testing.T) {
...
+	require.NoErrorf(t, err, "error accessing build details for version %q and buildID %q", latestSnapshotVersion.Original(), upgradeVersionString)
+	require.NotNil(t, buildDetails)
+	agentProject, ok := buildDetails.Build.Projects["elastic-agent"]
+	require.Truef(t, ok, "elastic agent project not found in version %q build %q", latestSnapshotVersion.Original(), upgradeVersionString)
+	t.Logf("agent build details: %+v", agentProject)
+	t.Logf("expected agent commit hash: %q", agentProject.CommitHash)
+	expectedAgentHashAfterUpgrade := agentProject.CommitHash
+
+	// Workaround until issue with Artifact API build commit hash are resolved
+	actualAgentHashAfterUpgrade := extractCommitHashFromArtifact(t, ctx, latestSnapshotVersion, agentProject)
```
pchila, line 399: Is there a specific reason why you want to install the penultimate SNAPSHOT build? This is from the upgrade to specific snapshot integration test but you could simply use the latest snapshot or the latest released version, right ?

pchila (commented):
The implementation looks good, I am just not sure that you picked the simplest integration test to take inspiration from... Maybe that can be changed with a simpler Standalone upgrade test ?

pchila:
As an aside, IMHO `skip-embedded-pgp-key` is clearer than `skip-default-pgp` even if it's longer, but I am not a product owner :sweat_smile:

internal/pkg/agent/application/upgrade/artifact/download/http/verifier.go
michalpristas: making linter happy

testing/integration/upgrade_test.go
michalpristas: i just took normal test as a base and added custom flags for my usecase, i havent changed a thing so it's better to ask probably @ycombinator about why this was chosen

testing/integration/upgrade_test.go
michalpristas: true

testing/integration/upgrade_test.go
```diff
@@ -273,7 +274,165 @@ func TestStandaloneUpgrade(t *testing.T) {
...
+	version, err := c.Version(ctx)
+	require.NoError(t, err, "error checking version after upgrade")
+	require.Equal(t, expectedAgentHashAfterUpgrade, version.Commit, "agent commit hash changed after upgrade")
+}
+
+func TestStandaloneUpgrade_FailInitialCheck(t *testing.T) {
+	define.Require(t, define.Requirements{
+		// Stack:   &define.Stack{},
+		Local:   false, // requires Agent installation
+		Isolate: true,
```
blakerouse, line 301: Why is this isolated?

testing/integration/upgrade_test.go
```diff
@@ -273,7 +274,165 @@ func TestStandaloneUpgrade(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	err = agentFixture.Prepare(ctx)
+	require.NoError(t, err, "error preparing agent fixture")
+
+	err = agentFixture.Configure(ctx, []byte(fastWatcherCfg))
+	require.NoError(t, err, "error configuring agent fixture")
+
+	const minVersionString = "8.10.0-SNAPSHOT"
+	minVersion, _ := version.ParseVersion(minVersionString)
```
blakerouse, line 318: I know its probably not going to happen, but just to be sure that in the case that `version.ParseVersion` throws an error it is checked. Online an extra line.

`require.NoError(t, err)`

pchila:
@blakerouse there is already some refactoring pending for the specific test and the upgrade tests in general in PR #2955 
Moreover the test that has been chosen as a base is the upgrade (downgrade) to specific snapshot which has additional complexity (looking at all the snapshot builds and retrieving the actual commit SHA etc.)
It would be probably easier if a different test was used as a base

michalpristas:
vulncheck not related

testing/integration/upgrade_test.go
```diff
@@ -66,6 +66,9 @@ var version_8_7_0 = version.NewParsedSemVer(8, 7, 0, "", "")
 // minimum version for upgrade to specific snapshot + minimum version for setting shorter watch period after upgrade
 var version_8_9_0_SNAPSHOT = version.NewParsedSemVer(8, 9, 0, "SNAPSHOT", "")
 
+// minimum version for upgrade to specific snapshot + minimum version for setting shorter watch period after upgrade
```
pchila, line 69: Nit: maybe we want to update the comment :smile:

testing/integration/upgrade_test.go
```diff
@@ -208,18 +211,18 @@ func TestStandaloneUpgrade(t *testing.T) {
 
 			parsedUpgradeVersion, err := version.ParseVersion(define.Version())
 			require.NoErrorf(t, err, "define.Version() %q cannot be parsed as agent version", define.Version())
-			testStandaloneUpgrade(ctx, t, agentFixture, parsedVersion, parsedUpgradeVersion, "", true)
+			testStandaloneUpgrade(ctx, t, agentFixture, parsedVersion, parsedUpgradeVersion, "", true, true, false, "")
```
pchila, line 214: The values of skip-verify depends on the versions being tested. Since this test will run also for older sets of versions we should keep the check

```suggestion
			skipVerify:=version_8_7_0.Less(parsedVersion)
			testStandaloneUpgrade(ctx, t, agentFixture, parsedVersion, parsedUpgradeVersion, "", skipVerify, true, false, "")
```

testing/integration/upgrade_test.go
```diff
@@ -424,7 +489,18 @@ func testStandaloneUpgrade(
 		require.NoError(t, err)
 		sourceURI := "file://" + filepath.Dir(srcPkg)
 		t.Logf("setting sourceURI to : %q", sourceURI)
-		upgradeCmdArgs = append(upgradeCmdArgs, "--source-uri", sourceURI, "--skip-verify")
```
pchila, line 427: the skipVerify and source-uri have to go hand in hand when we are testing with the locally built package...
How about setting the conditions to something like:
```
useLocalPackage :=  allowLocalPackage && version_8_7_0.Less(*parsedFromVersion)
skipVerifyFlag := useLocalPackage || skipVerify
```

Or is the skipVerify an override?

If this get too complicated we can always extract all of those as options to be set in the caller function...

testing/integration/upgrade_test.go
```diff
@@ -208,18 +211,18 @@ func TestStandaloneUpgrade(t *testing.T) {
...
 			parsedUpgradeVersion, err := version.ParseVersion(define.Version())
 			require.NoErrorf(t, err, "define.Version() %q cannot be parsed as agent version", define.Version())
-			testStandaloneUpgrade(ctx, t, agentFixture, parsedVersion, parsedUpgradeVersion, "", true)
+			testStandaloneUpgrade(ctx, t, agentFixture, parsedVersion, parsedUpgradeVersion, "", true, true, false, "")
 		})
 	}
 }
 
-func TestStandaloneUpgradeToSpecificSnapshotBuild(t *testing.T) {
+func TestStandaloneUpgradeToSpecificSnapshotBuildWithGPGFallback(t *testing.T) {
```
pchila, line 219: I still don't understand why this test has to use a different build of the same snapshot version for the upgrade...
Can't we test the same functionality by downgrading from the locally built package to a previous (distinct) version ? Would that change the nature of the test?

If we can choose a different version you can simplify the test to a `testStandaloneUpgrade` call + a few statements to manipulate the PGP key as it's already there... You have no need to the commit sha workaround or checking the commit sha after upgrade...

pchila (changes requested):
There are some changes that are needed (the --skip-verify value when not using source-uri for example) and I still think that the integration test can be simplified and reduced to a normal up(down)grade test

testing/integration/upgrade_test.go
michalpristas: updated to simpler version

testing/integration/upgrade_test.go
```diff
@@ -208,11 +211,56 @@ func TestStandaloneUpgrade(t *testing.T) {
...
+
+	if parsedVersion.Less(*minVersion) {
+		t.Skipf("Version %s is lower than min version %s", define.Version(), minVersion)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// previous
+	toVersion, err := parsedVersion.GetPreviousMinor()
```
pchila, line 238: Not the biggest fan of trying to guess the previous minor by decreasing the counters...
Since we already have access to artifact api you could retrieve the previous version string from there...
Not a blocker in itself until we release 9.x it's just that we can rely on the actual agent versions...

pchila:
buildkite test this

PR #3109, opened 2023-07-20
Pass go test flags to integration tests

pchila:
## Why is it important?

It allows to propagate go test flags to the actual go test command allowing for finer control over the test execution.
The `-test.short` is also important for allowing a quicker execution for PRs, for example while maintaining a full integration test run in other contexts.

## Checklist

## Author's Checklist

- [ ]

Try to run integration tests (both locally and remotely) with some custom go test flags. There are some examples listed in the [dev-guide](https://github.com/elastic/elastic-agent/blob/b2a0f2dc837a524763b893a134d552e77b3d17ff/docs/test-framework-dev-guide.md#passing-additional-go-test-flags), for example:

We pass a `-test.run` flag along with the names of the tests we want to run in OR
```
GOTEST_FLAGS="-test.run ^(TestStandaloneUpgrade|TestFleetManagedUpgrade)$" mage integration:test
```

## Related issues

- Relates #2955 

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/gotest.go
+++ b/dev-tools/mage/gotest.go
@@ -232,11 +232,10 @@ func GoTest(ctx context.Context, params GoTestArgs) error {
 		)
 	}
 
+	testArgs = append(testArgs, params.ExtraFlags...)
 	if params.RunExpr != "" {
 		testArgs = append(testArgs, "-run", params.RunExpr)
 	}
-
-	testArgs = append(testArgs, params.ExtraFlags...)
 	testArgs = append(testArgs, params.Packages...)
 
 	args := append(gotestsumArgs, append([]string{"--"}, testArgs...)...)
--- a/dev-tools/mage/gotest.go
+++ b/dev-tools/mage/gotest.go
@@ -232,6 +232,22 @@ func GoTest(ctx context.Context, params GoTestArgs) error {
 		)
 	}
 
+	// Pass the go test extra flags BEFORE the RunExpr.
+	// TL;DR: This is needed to make sure that a -test.run flag specified in the GOTEST_FLAGS environment variable does
+	// not interfere with the batching done by the framework.
+	//
+	// Full explanation:
+	// The integration test framework runs the tests twice:
+	// - the first time we pass a special tag that make all the define statements in the tests skip the test and dump the requirements.
+	//   This output is processed by the integration test framework to discover the tests and the set of environments/machines
+	//   we will need to spawn and allocate the tests to the various machines. (see batch.go for details)
+	// - the second time we run the tests (here) the integration test framework adds a -test.run flag when launching go test
+	//   on the remote machine to make sure that only the tests corresponding to that batch are executed.
+	//
+	// By specifying the extra flags before the -test.run for the batch we make sure that the last flag definition "wins"
+	// (have a look at the unit test in batch_test.go), so that whatever run constraint is specified in GOTEST_FLAGS
+	// participates in the discovery and batching (1st go test execution) but doesn't override the actual execution on
+	// the remote machine (2nd go test execution).
 	testArgs = append(testArgs, params.ExtraFlags...)
 	if params.RunExpr != "" {
 		testArgs = append(testArgs, "-run", params.RunExpr)
--- a/docs/test-framework-dev-guide.md
+++ b/docs/test-framework-dev-guide.md
@@ -68,6 +68,7 @@ We want to run only the test named "TestStandaloneUpgrade"
 ##### Run a tests matching a partial expression
 We want to run any test with "Upgrade" in the name
 `GOTEST_FLAGS="-test.run Upgrade" mage integration:test`
+
 ##### Run a single test and signal that we want the short version
 We pass a `-test.short` flag along with the name match
 `GOTEST_FLAGS="-test.run ^TestStandaloneUpgrade$ -test.short" mage integration:test`
--- a/docs/test-framework-dev-guide.md
+++ b/docs/test-framework-dev-guide.md
@@ -49,6 +49,41 @@ pass `[testName]` to `go test` as `--run=[testName]`.
 
 - `mage integration:matrix` to run all tests on the complete matrix of supported operating systems and architectures of the Elastic Agent.
 
+#### Passing additional go test flags
+
+When running the tests we can pass additional go test flag using the env variable `GOTEST_FLAGS`.
+
+These flags are passed also when calculating batches for remote execution of integration tests.
+This allows for selecting a subset of test in a convenient way (see examples below)
+
+This feature is intended mostly for integration tests debugging/development without the need for
+new mage targets corresponding to a new set of test flags.
+
+A few examples:
+
+##### Run a single test with an exact match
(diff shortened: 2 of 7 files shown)


magefile.go
```diff
@@ -1425,17 +1432,17 @@ func (Integration) TestOnRemote(ctx context.Context) error {
 		testPrefix := fmt.Sprintf("%s.%s", prefix, filepath.Base(packageName))
 		testName := fmt.Sprintf("remote-%s", testPrefix)
 		fileName := fmt.Sprintf("build/TEST-go-%s", testName)
+		extraFlags := make([]string, 0, len(goTestFlags)+3)
+		extraFlags = append(extraFlags, goTestFlags...)
+		extraFlags = append(extraFlags, "-test.shuffle", "on",
+			"-test.timeout", "0", "-test.run", "'^("+strings.Join(packageTests, "|")+")$'")
```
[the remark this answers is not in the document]
blakerouse: +1 should be `+6`

blakerouse (approved):
Change looks good to me.

magefile.go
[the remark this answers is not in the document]
pchila: good catch! This is a leftover from when I was thinking about changing the flag form to `-test.foo=bar`... Will change to 6 :smile:

pchila:
ok,  now integration tests are running correctly after a small change in 5c3227092f2e6b53bdd19a0d026a598558fc3fbb

/cc @ycombinator

dev-tools/mage/gotest.go
```diff
@@ -232,11 +232,10 @@ func GoTest(ctx context.Context, params GoTestArgs) error {
 		)
 	}
 
+	testArgs = append(testArgs, params.ExtraFlags...)
```
ycombinator, line 235: Just curious: why was this change needed?

docs/test-framework-dev-guide.md
```diff
@@ -49,6 +49,41 @@ pass `[testName]` to `go test` as `--run=[testName]`.
...
+A few examples:
+
+##### Run a single test with an exact match
+We want to run only the test named "TestStandaloneUpgrade"
+`GOTEST_FLAGS="-test.run ^TestStandaloneUpgrade$" mage integration:test`
+
+##### Run a tests matching a partial expression
+We want to run any test with "Upgrade" in the name
+`GOTEST_FLAGS="-test.run Upgrade" mage integration:test`
+##### Run a single test and signal that we want the short version
```
ycombinator, line 71:
```suggestion

##### Run a single test and signal that we want the short version
```

docs/test-framework-dev-guide.md
```diff
@@ -49,6 +49,41 @@ pass `[testName]` to `go test` as `--run=[testName]`.
...
+We pass a `-test.count` flag along with the name match
+`GOTEST_FLAGS="-test.run ^TestStandaloneUpgrade$ -test.count 10" mage integration:test`
+
+##### Run specific tests
+We pass a `-test.run` flag along with the names of the tests we want to run in OR
+`GOTEST_FLAGS="-test.run ^(TestStandaloneUpgrade|TestFleetManagedUpgrade)$" mage integration:test`
+
+##### Limitations
+Due to the way the parameters are passed to `devtools.GoTest` the value of the environment variable
+is split on space, so not all combination of flags and their values may be correctly split.
```
ycombinator, line 85: Thanks for adding this section.

Does this limitation have something to do with some go test flags taking arguments, e.g. `-test.run <test name regex>`, vs. others not taking arguments, e.g. `-short`?

magefile.go
```diff
@@ -1491,7 +1506,7 @@ func integRunner(ctx context.Context, matrix bool, singleTest string) error {
 	return nil
 }
 
-func createTestRunner(matrix bool, singleTest string, batches ...define.Batch) (*runner.Runner, error) {
+func createTestRunner(matrix bool, singleTest string, goTestFlags string, batches ...define.Batch) (*runner.Runner, error) {
```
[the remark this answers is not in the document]
pchila: We could... I am leaving it in, in case the new flags do not behave like we want but we can deprecate them and create a separate issue to clean up

docs/test-framework-dev-guide.md
pchila: It does not have any problems with flags that take arguments or not... 
The problem is that when one argument contains a space, it will be split incorrectly.
The correct way would be to:
- either reimplement the flag parser of `go test` (those are not exposed sadly and reimplementing it and keeping it in sync with the future versions of go it will be probably too much work)
- or define a completely different set of flags and translate them into go test flags (kinda misses the point of the PR)

Another alternative is to build the `go test` command differently so that we can plonk the flags as a single string being completely agnostic about the contents but that is also a bit of work...

For this PR I went with the simple and stupid solution and I couldn't think of a scenario where I needed a smarter parsing of flags: we'll cross that bridge when we have to.

magefile.go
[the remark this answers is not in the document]
blakerouse: I like `mage integration:single [testName]` it is much clean CLI command to run a test than all the ENV flags being set. I wish we could get mage to take command line arguments for targets. I would prefer to see something like:

`mage integration:test -short -name SingleTestName -platform linux/amd64`

dev-tools/mage/gotest.go
pchila: Sorry for the incoming wall of text :sweat_smile: 

TL;DR: This was needed to make sure that a `-test.run` flag specified in the `GOTEST_FLAGS` environment variable does not interfere with the batching done by the framework.

Full explanation:
The integration test framework runs the tests twice:
- the first time we pass a special tag that make all the `define` statements in the tests skip the test and dump the requirements. This output is processed by the integration test framework to discover the tests and the set of environments/machines we will need to spawn and allocate the tests to the various machines.
- the second time we run the tests (here) the integration test framework adds a `-test.run` flag when launching go test on the remote machine to make sure that only the tests corresponding to that batch are executed.

By specifying the extra flags before the `-test.run` for the batch we make sure that the last flag definition "wins" (have a look at the unit test in `batch_test.go`) so that whatever `run` constraint is specified in GOTEST_FLAGS participated in the discovery and batching (1st `go test` execution) but doesn't override the actual execution on the remote machine.

magefile.go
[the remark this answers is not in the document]
ycombinator: Okay, then could we eliminate _just_ the `singleTest` argument to this function and have `mage integration:single [testName]` internally setup `goTestFlags` appropriately?

docs/test-framework-dev-guide.md
ycombinator: Understood, thanks for clarifying.  I'm good with the simple solution for now.

dev-tools/mage/gotest.go
cmacknz: This should definitely be a comment in the code.

pchila: Added the comment just above the change.

pchila:
buildkite test this

PR #3111, opened 2023-07-20
Fix data race in `TestActionUnenrollHandler` test

ycombinator:
## What does this PR do?

This PR fixes the data race found by `go test -race` in the `TestActionUnenrollHandler` unit test.

## Why is it important?

So the test passes consistently.


blakerouse (approved):
Looks good.

Issue #2141, opened 2023-01-18
Profiles collected by diagnostics should be per component (or process) not per unit


cmacknz:
We are duplicating the profiles per unit, when the profiles apply at the process level and they should likely be at the component level. For example looking at a log component implemented by Filebeat will show the following where the per unit profiles don't add any value.

```sh
tree diagnostics/components/log-default/
├── log-default
│   ├── allocs.txt
│   ├── beat-rendered-config.yml
│   ├── block.txt
│   ├── goroutine.txt
│   ├── heap.txt
│   ├── mutex.txt
│   └── threadcreate.txt
└── logfile-system-b856919c-ade6-4a47-9e0c-28d4822b7ada
    ├── allocs.txt
    ├── beat-rendered-config.yml
    ├── block.txt
    ├── goroutine.txt
    ├── heap.txt
    ├── mutex.txt
    └── threadcreate.txt
```

fearful-symmetry:
Running into this while working on https://github.com/elastic/elastic-agent/issues/2140, and there's a series of implementation and API issues that kind of lead to this:

1) In implementation, the elastic-agent-client will run any diagnostic callback (including the pprof diagnostics mentioned above) that are registered at the client level, and run them per-unit
2) in the protobuf API between the agent CLI and the agent server, the actual API call is `DiagnosticUnits` and it operates at the level of individual units; there's no component-level `DiagnosticComponents` protobuf call or anything.
3) In the protobuf API between the agent and the components, the diagnostic gets compressed down to a single `ActionRequest_DIAGNOSTICS` type.

So, I feel like the correct way to fix this would be to have a `DiagnosticComponents` API call, and then leverage the `ActionRequest` struct in some way so the client can differentiate between unit and component-level diagnostics. At that point we can properly fix the first issue of client-level callback registration.

cmacknz:
Yeah I think the diagnostics API should have accounted for component diagnostics initially, but since it didn't the next best option is a `DiagnosticsComponents` API call.

We need a way to support process global diagnostics like CPU and heap profiles and this is the only way that makes sense long term.

Issue #2140, opened 2023-01-18
The elastic-agent diagnostics command should be able to collect CPU profiles


cmacknz:
The new diagnostics implementation in 8.6 captures memory and goroutine profiles (and several other profile types) by default. We should add a command line option to allow capturing CPU profiles, and allow varying the sample period. CPU profiles take some time to collect (defaulting to 30s per process) so it makes some sense for them to be disabled by default.

cmacknz:
When we implement this, we may need to make collecting CPU profiles optional with a new option for the diagnostics command. If we do, it could also be nice to allow targeting specific components instead of trying to profile all of them at once.

fearful-symmetry:
So, going over this, and I'm wondering how we want to integrate this (specifically the UI element for waiting to collect a CPU diagnostic) into the diagnostic collection UI in Kibana. The CLI interface is simple enough, but the web GUI is backed up by the action API and it's a whole different beast. Do we need to collaborate with someone on the fleet/Kibana team for a new UI element? Should we just do the CLI now?
<img width="905" alt="Screenshot 2023-07-12 at 12 55 21 PM" src="https://github.com/elastic/elastic-agent/assets/8418476/2ad44833-3473-412a-a74e-80b85ac275f0">

fearful-symmetry:
For now just going to work on the CLI until we figure out how the Kibana/actions API changes should work.

cmacknz:
Yes start with the CLI, we can create a separate issue to expose this in Fleet. That will involve changing the diangostics action payload to indicate whether or not CPU profiles should be supported. 

https://github.com/elastic/elastic-agent/blob/8b4ddebda5364e35cf20fd82812595426db180b5/internal/pkg/fleetapi/action.go#L419-L425

cmacknz:
I'm reopening this because we missed updating the documentation explaining the format changes and the new option.

https://github.com/elastic/ingest-docs/blob/main/docs/en/ingest-management/commands.asciidoc#elastic-agent-diagnostics-command

Close this again once there is a PR updating that page in the docs.

PR #3118, opened 2023-07-24
Add CPU profiling and component-level diagnostics 

fearful-symmetry:
## What does this PR do?

Follow-up to https://github.com/elastic/elastic-agent-client/pull/80

Part of https://github.com/elastic/elastic-agent/issues/2140 and https://github.com/elastic/elastic-agent/issues/2141

NOTE: Before the above two issues are actually closed, after we merge this we also need to update `elastic-agent-client` in beats itself.

This PR accomplishes a few things:
- makes changes to the agent control protocol to add a `DiagnosticComponentsRequest` call and an `AdditionalDiagnosticRequest` request field. This allows us to request an optional CPU profile to the diagnostic bundle.
- A number of changes to the coordinator, diagnostics command, and manager to support component-level diagnostics.
- An additional integration test
- Increase the timeout for diagnostic requests
- Restructures the .zip files that elastic-agent produces, so component-level diagnostics are properly placed in the `components/` directory, and not copied per-unit
- Add a CPU profile diag hook for the agent itself

It should be noted that codepath for component diagnostics is different from the normal unit-diagnostics in a number of ways:

- The manager `PerformComponentDiagnostics` method returns an error value, the corresponding unit-level diagnostic does not
- The component level diagnostic aggregator works in parallel, spawning a go routine for every component that requires diagnostics. As CPU profiles run for a set period of time, this is kind of needed so diagnostic request doesn't take 3 minutes.

## Why is it important?

We want CPU profiling in diagnostics 

## Checklist

- Pull down and build in whatever way you normally would
- run `elastic-agent diagnostics -p`


internal/pkg/diagnostics/diagnostics.go
```diff
@@ -139,13 +140,31 @@ func pprofDiag(name string, debug int) func(context.Context) []byte {
...
+func CreateCPUProfile(ctx context.Context, period time.Duration) ([]byte, error) {
+	var writeBuf bytes.Buffer
+	err := pprof.StartCPUProfile(&writeBuf)
+	if err != nil {
+		return nil, fmt.Errorf("error starting CPU profile: %w", err)
+	}
+	tc := time.After(period)
+	select {
+	case <-ctx.Done():
+		return nil, fmt.Errorf("got context done")
```
blakerouse, line 153:
```suggestion
		return nil, ctx.Err()
```

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -139,13 +140,31 @@ func pprofDiag(name string, debug int) func(context.Context) []byte {
...
+	case <-tc:
+		break
+	}
+
+	pprof.StopCPUProfile()
```
blakerouse, line 158: Shouldn't this be called when `ctx.Done()` occurs?

pkg/control/v2/server/server.go
```diff
@@ -199,12 +199,71 @@ func (s *Server) DiagnosticAgent(ctx context.Context, _ *cproto.DiagnosticAgentR
 			Generated:   timestamppb.New(time.Now().UTC()),
 		})
 	}
+
+	for _, metric := range req.AdditionalMetrics {
+		switch metric {
+		case cproto.AdditionalDiagnosticRequest_CPU:
+			duration := time.Second * 30
```
blakerouse, line 206: Best to place this as a const at the top of this file.

pkg/control/v2/server/server.go
```diff
@@ -199,12 +199,71 @@ func (s *Server) DiagnosticAgent(ctx context.Context, _ *cproto.DiagnosticAgentR
...
+			s.logger.Infof("cpu metrics collected, waiting for %s", duration)
```
blakerouse, line 207:
```suggestion
			s.logger.Infof("cpu metrics being collected, waiting for %s", duration)
```

blakerouse (approved):
Change looks good.

A couple comments, I believe the one for stoping the CPU profile on context cancellation needs to be fixed before merge.

blakerouse (approved):
Updates look great!

PR #3014, opened 2023-07-06
Improve uninstall on Windows to remove directory even when uninstall is being executed from same directory

blakerouse:
## What does this PR do?

Previously the Elastic Agent would spawn a `cmd.exe` that would wait 2 seconds and then delete the `C:\Program Files\Elastic\Agent` directory. This triggers a Cloud IOC alert as that is something that nefarious applications do.

This provides a better way of uninstalling of Elastic Agent on Windows, by using the Windows API to allow removal of a directory even if there is an executable running inside of it.

## Why is it important?

Two reasons this is important. Uninstall no longer has a 2 second delay for it to be fully removed, something that caused a problem with integration testing on Windows, and a Cloud IOC alert will no longer trigger when Elastic Agent is uninstalled.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test (enabled with Windows runner in other PR)

## How to test this PR locally

```
PS> .\elastic-agent install -f
PS> "C:\Program Files\Elastic\Agent\elastic-agent.exe" uninstall -f
```

## Related issues

- Closes #2970 

## Logs

```
PS C:\Program Files\Elastic> .\Agent\elastic-agent.exe uninstall
Elastic Agent will be uninstalled from your system at C:\Program Files\Elastic\Agent. Do you want to continue? [Y/n]:y
Elastic Agent has been uninstalled.
```

PR #3120, opened 2023-07-25
Improve uninstall on Windows to remove directory even when uninstall is being executed from same directory

### LINKED ISSUE #2970: Uninstall on Windows triggers a Cloud IOC alert
The `uninstall` command for the Elastic Agent performs an action at the end of the installation process that spawns a `cmd.exe` that uses ping to wait 2 seconds and then delete the `C:\Program Files\Elastic\Agent` directory. This is done because `uninstall` is being executed from that directory and it cannot delete the directory that it is executing from (Windows does not allow it).

See the code here: https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/install/uninstall.go#L143

This type of process execution on Windows triggers a Cloud IOC alert. We should determine if its possible for this removal to be performed differently that doesn't trigger this type of alert.

- Version: All
- Operating System: Windows

blakerouse:
## What does this PR do?

Previously the Elastic Agent would spawn a `cmd.exe` that would wait 2 seconds and then delete the `C:\Program Files\Elastic\Agent` directory. This triggers a Cloud IOC alert as that is something that nefarious applications do.

This provides a better way of uninstalling of Elastic Agent on Windows, by using the Windows API to allow removal of a directory even if there is an executable running inside of it.

This is the same as #3014 that was reverted, except it has retry logic added that keeps trying to remove the directory in the case that `ERROR_SHARING_VIOLATION` occurs. There is no way to delete a file that has that error, but some times Windows reports that error. This takes the same path that golang uses internally to remove a directory. See: https://github.com/golang/go/blob/master/src/testing/testing.go#L1257

## Why is it important?

Two reasons this is important. Uninstall no longer has a 2 second delay for it to be fully removed, something that caused a problem with integration testing on Windows, and a Cloud IOC alert will no longer trigger when Elastic Agent is uninstalled.

## Checklist

- [x] I have added an integration test or an E2E test (enabled with Windows runner in other PR)

```
PS> .\elastic-agent install -f
PS> "C:\Program Files\Elastic\Agent\elastic-agent.exe" uninstall -f
```

## Related issues

- Closes #2970 

## Logs

```
PS C:\Program Files\Elastic> .\Agent\elastic-agent.exe uninstall
Elastic Agent will be uninstalled from your system at C:\Program Files\Elastic\Agent. Do you want to continue? [Y/n]:y
```

### DIFF
--- a/changelog/fragments/1688657261-Don't-trigger-IOC-alert-on-Windows-uninstall.yaml
+++ b/changelog/fragments/1688657261-Don't-trigger-IOC-alert-on-Windows-uninstall.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Don't trigger IOC alert on Windows uninstall
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description:
+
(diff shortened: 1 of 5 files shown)


blakerouse:
Going to run this a few times to ensure that this solves any of the random failures that we where seeing before.

blakerouse:
Seems we got a failure of `TestActionDispatcher/Cancel_queued_action` unrelated to this change. Running again.

michalpristas (approved):
Looks like it works, i havent tested this on win10 thought. just win11

PR #3136, opened 2023-07-27
Improve test runner to re-use instances and make provisioners pluggable

### LINKED ISSUE #2856: Integration testing: suggestions for various performance improvements
Right now, `mage integration:*` involves some fairly heavy operations, even `mage integration:Single` with the lowest-ping GCP datacenter can take around 16-20 minutes to run a single test, which is a tad long when you're actually trying to develop tests. 

After poking around the code, it seems like there's a few different things we can do to speed up the process:

- Create a pre-built machine image for the tests that's updated via automation. Right now each test run also steps through a setup process:
```
>>> (linux/amd64/ubuntu/22.04[p3sl]) Running apt-get update
>>> (linux/arm64/ubuntu/22.04[p3sm]) Install build-essential and unzip
>>> (linux/amd64/ubuntu/22.04[p3sl]) Install build-essential and unzip
>>> (linux/arm64/ubuntu/22.04[p3sm]) Install golang 1.19.9 (arm64)
>>> (linux/arm64/ubuntu/22.04[p3sm]) Copying repo
>>> (linux/arm64/ubuntu/22.04[p3sm]) Running make mage and prepareOnRemote
```

Ideally we should have some automation or something to create a premade machine image with all the needed dependencies like mage, `unzip`, etc, and the test infrastructure can use/boot to that image.

- Constrain the image types we're running on. Right now by default we seem to run on both x86 and ARM:
```
>>> (linux/arm64/ubuntu/22.04[p3sm]) Copying agent build elastic-agent-8.9.0-linux-arm64.tar.gz
>>> (linux/amd64/ubuntu/22.04[p3sl]) Copying agent build elastic-agent-8.9.0-linux-x86_64.tar.gz
```

Although these steps are run in parallel, it still ends up adding time to the test development process, as you have to rebuild two separate binaries, and one of the two steps will end up taking longer, and blocking (for example, the x86 builds are slightly larger, and thus take longer to upload).

### LINKED ISSUE #2933: Allow reusing previously provisioned resources for integration testing
This is a follow up from https://github.com/elastic/elastic-agent/issues/2856. The integration framework works well, but running individual tests can take 10+ minutes. This costs significant time in two common use cases:

1. Developing a test for the first time where unexpected problems are common and the test needs to be re-run many times.
2. Repeatedly running a test to reproduce an intermittent issue.

To speed up these two use cases we should eliminate the need to re-provision the VMs the tests run on, and in the case of E2E tests the Elastic Stack deployment they are tested against.

There are two ways to accomplish this:
1. Split the current `mage integration:test` target into two dependent pieces, one to provision resources and prepare them for test execution and another to actually run the tests. This would allow developers to only run the provisioning step a single time. Developers would need to be given the credentials to SSH or RDP into the created VMs to allow manually executing tests and uploading new elastic agent packages.
2. A simpler option proposed in a discussion with @ycombinator, which is to stop immediately destroying resources after a test executes. We would be default leave VMs and clusters after a test executes and allow targeting them for reuse. We would rely on the QA region's automated cleanup of unused resources to remove them, or as an optimization we could write our own batch job to clean up resources tagged as created by the test framework on a regular scheduled (triggered by Buildkite for example).

blakerouse:
## What does this PR do?

Refactors the way integration test runner works to provide the following benefits:

* Instance re-use - Call to any `mage integration:*` target that requires the usage of a deployed instance will re-use an already deployed instance. A specific call to `mage integration:clean` is required to bring down created instances and stacks.
* Copy builds only if required - With instance re-use support the code has been changed to only copy the Elastic Agent builds to an instance in the case that the build has changed.
* Add `InstanceProvisioner` interface - This interface provides a defined interface between the runner and the instance provisioner. This will allow other instance provisioners to be added to the test runner.
* Add `StackProvisioner` interface - This interface provides a defined interface between the runner and the stack provisioner. This will allow other stack provisioners to be add to the test runner.

## Why is it important?

* Provides the ability to inspect an instance in the case of test failures and test success as the instance remains running when the tests is complete.
* Iteration of integration tests becomes faster for a developer because an instance that has already been prepared can keep being re-used.

## Checklist

```
$ mage integration:test
$ mage integration:test # quicker second run
```

## Related issues

- Closes #2933
- Closes #2856


blakerouse:
Hmm didn't change how the runner uses OGC that much. Getting a strange error from OGC:

```


Status: Downloaded newer image for gorambo/ogc:blake
--
  | Traceback (most recent call last):
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 3246, in execute_sql
  | cursor.execute(sql, params or ())
  | sqlite3.OperationalError: no such table: layouts
  |  
  | During handling of the above exception, another exception occurred:
  |  
  | Traceback (most recent call last):
  | File "/venv/bin/ogc", line 8, in <module>
  | sys.exit(start())
  | ^^^^^^^
  | File "/venv/lib/python3.11/site-packages/ogc/commands/base.py", line 27, in start
  | cli()
  | File "/venv/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
  | return self.main(*args, **kwargs)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/click/core.py", line 1055, in main
  | rv = self.invoke(ctx)
  | ^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/click/core.py", line 1657, in invoke
  | return _process_result(sub_ctx.command.invoke(sub_ctx))
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
  | return ctx.invoke(self.callback, **ctx.params)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/click/core.py", line 760, in invoke
  | return __callback(*args, **kwargs)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/ogc/commands/down.py", line 34, in down
  | _layouts: list[layout.LayoutModel] = [
  | ^
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 7107, in __iter__
  | self.execute()
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 1966, in inner
  | return method(self, database, *args, **kwargs)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 2037, in execute
  | return self._execute(database)
  | ^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 2210, in _execute
  | cursor = database.execute(self)
  | ^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 3254, in execute
  | return self.execute_sql(sql, params)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 3244, in execute_sql
  | with __exception_wrapper__:
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 3014, in __exit__
  | reraise(new_type, new_type(exc_value, *exc_args), traceback)
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 192, in reraise
  | raise value.with_traceback(tb)
  | File "/venv/lib/python3.11/site-packages/peewee.py", line 3246, in execute_sql
  | cursor.execute(sql, params or ())
  | peewee.OperationalError: no such table: layouts
  |  
  | Error: error running clean: failed to run ogc down: docker run exited with code: 1


``

**Comment by fearful-symmetry** [comment]
Getting some odd errors:
```
mage integration:single TestFleetManagedUpgrade 
>>> Creating zip archive of repo to send to remote hosts
>>> Create SSH keys to use for SSH
>>> Pulling latest ogc image
>>> Creating stack 8.10.0-SNAPSHOT (8100-SNAPSHOT)
Error: error running test: failed to run ogc import: docker run exited with code: 1
```

Added a bit of extra debugging, got this:
```
Error: error running test: failed to run ogcPull: docker run exited with code: 1: docker: 'ogcPull' is not a docker command.
```

Looks like the issue is here:
```
	args := []string{
		"ogcPull",
		"docker.io/gorambo/ogc:blake", // switch back to :latest when ready
	}
	p.logger.Logf("Pulling latest ogc image")
	proc, err := process.Start("docker", process.WithContext(ctx), process.WithArgs(args))
```

On the chance `"ogcPull"` was a typo, I just changed it to `pull`, but then OGC just threw a different error:
```
  File "/venv/lib/python3.11/site-packages/ogc/deployer.py", line 215, in _up_async
    provisioner.create()
  File "/venv/lib/python3.11/site-packages/ogc/provision.py", line 381, in create
    size = self.sizes(self.layout.instance_size)[0]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
2023-07-27T18:55:34Z <Greenlet at 0x7f4807a2d080: _up_async(<LayoutModel: 2>)> failed with IndexError
```

**Comment by blakerouse** [comment]
@fearful-symmetry Yeah I added a typo trying to fix the lint issue.

Just pushed latest fix. Try running `mage integration:clean` and then run it again. I see it working locally again with latest commit.

**Comment by fearful-symmetry** [comment]
Yup, still getting ogc errors after `integration:clean`
```
{23-07-27 12:01}shoebill:~/go/src/github.com/elastic/elastic-agent@refactor-testing-runner✗✗✗✗✗✗ alexk% mage -v integration:single TestFleetManagedUpgrade
Running target: Integration:Single
exec: go "list" "-m"
Found Elastic Beats dir at /home/alexk/go/src/github.com/elastic/elastic-agent
>>> Creating zip archive of repo to send to remote hosts
>>> Create SSH keys to use for SSH
>>> Pulling latest ogc image
>>> Creating stack 8.10.0-SNAPSHOT (8100-SNAPSHOT)
>>> Import layouts into ogc
>>> Bring up instances through ogc
[19:01:33] INFO     Bringing up machines                         deployer.py:210
[19:01:33] DEBUG    Establing provider connection...<class      provision.py:303
                    'libcloud.compute.drivers.gce.GCENodeDriver                 
                    '> - {'user_id':                                            
                    'alex-kristiansen-agent-testing@elastic-pla                 
                    tform-ingest.iam.gserviceaccount.com',                      
                    'key':                                                      
                    '/root/agent-testing-service-token.json',                   
                    'project': 'elastic-platform-ingest',                       
                    'datacenter': 'us-west1-a'}                                 
[19:01:33] DEBUG    Establing provider connection...<class      provision.py:303
                    'libcloud.compute.drivers.gce.GCENodeDriver                 
                    '> - {'user_id':                                            
                    'alex-kristiansen-agent-testing@elastic-pla                 
                    tform-ingest.iam.gserviceaccount.com',                      
                    'key':                                                      
                    '/root/agent-testing-service-token.json',                   
                    'project': 'elastic-platform-ingest',                       
                    'datacenter': 'us-west1-a'}                                 
[19:01:35] WARNING  No firewall found, will create one to       provision.py:361
                    attach nodes to.                                            
[19:01:36] WARNING  No firewall found, will create one to       provision.py:361
                    attach nodes to.                                            
{'name': 'ogc-linux-amd64-ubuntu-2204-aadd', 'size': <NodeSize: id=335002, name=e2-standard-2, ram=8192 disk=0 bandwidth=0 price=None driver=Google Compute Engine ...>, 'image': <NodeImage: id=5502244062996746584, name=ubuntu-2204-jammy-v20230727, driver=Google Compute Engine  ...>, 'ex_metadata': {'items': [{'key': 'ssh-keys', 'value': 'ubuntu: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAgDLv5EJUTbA+eCLEL68rz+cidkjxeRSUFGXa8+mVNPkfRXXRgDjK/N1LBC/GVS++hVL4lqcXsLJIoKDHdSGHHEqQAo2jgH9RVZ8kEaOfivzEmxyMDYgxvVuI9pkYVMvLRPorGvxcG4QXhXAgG76aWN82t8aR6COG5HlvOp5R67a5d/my/K9lywjNHj7F8Hlodz81bEVn1ur2WRSrieGM+b9nhiLtnh7hDa2Iu0VOQm8tZyWQUokHK1hDNTcfVgX1w7vm8HCqEtw8NDn5SrqFkbVRs6AN+xin84iJ91yeD77Jb/d/ZZFr94J89d9VtJbu61dNpLbV2E8V/ToTXcXKSM/0Q=='}, {'key': 'startup-script', 'value': '#!/usr/bin/env bash\nif ! test -f "/usr/local/bin/pacapt"; then\n    sudo wget -O /usr/local/bin/pacapt https://github.com/icy/pacapt/raw/ng/pacapt\n    sudo chmod 755 /usr/local/bin/pacapt\n    sudo ln -sv /usr/local/bin/pacapt /usr/local/bin/pacman || true\nfi\npacapt update || true\npacapt install rsync || true\n'}]}, 'ex_tags': ['agent-integration', 'linux', 'amd64', 'ubuntu-22-04', 'created-2023-07-27', 'user-ogc', 'environment-ogc', 'repo-ogc'], 'ex_labels': {'division': 'engineering', 'org': 'platform', 'project': 'elastic-agent', 'team': 'ingest'}, 'ex_disk_type': 'pd-ssd', 'ex_disk_size': 100, 'ex_preemptible': False}
Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/venv/lib/python3.11/site-packages/ogc/deployer.py", line 215, in _up_async
    provisioner.create()
  File "/venv/lib/python3.11/site-packages/ogc/provision.py", line 381, in create
    size = self.sizes(self.layout.instance_size)[0]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
2023-07-27T19:01:45Z <Greenlet at 0x7f3b9fa3d080: _up_async(<LayoutModel: 2>)> failed with IndexError


Error: error running test: failed to find machine for layout ID: linux-arm64-ubuntu-2204
```

**Comment by blakerouse** [comment]
@fearful-symmetry Do you not get that error on main? I have never seen that error before and I am not getting that error locally.

**Comment by fearful-symmetry** [comment]
@blakerouse yah, it appears I'm not hitting the issue on main, just here.

**Comment by fearful-symmetry** [comment]
Yup, seems to be _just_ on this branch. Tried swapping between `main` and here and different combinations of `mage integration:clean` and `rm -r .ogc-cache`, but couldn't get the failure on main.

**Comment by blakerouse** [comment]
@fearful-symmetry I am not seeing that error on my side. I have ran it many times sucessfully.

@adam-stokes ^ any ideas on why this change would be causing this problem?

**Comment by fearful-symmetry** [comment]
Aright, it magically stopped happening. Not sure what combination of `mage integration:clean` and `rm -r .ogc-cache` made this work.

**Comment by fearful-symmetry** [comment]
Alright, so, if you don't have an `.ogc-cache` directory, when you run the test, it'll put things in _some_ kind of weird state that's hard to get out of. Still tinkering with it.

File: pkg/testing/ogc/provisioner.go [hunk 1]
```diff
@@ -0,0 +1,338 @@
...
+		return fmt.Errorf("failed to run ogc import: docker run exited with code: %d", ps.ExitCode())
+	}
+	return nil
+}
+
+// ogcUp brings up all the instances.
+func (p *provisioner) ogcUp(ctx context.Context) ([]byte, error) {
+	p.logger.Logf("Bring up instances through ogc")
+	var output bytes.Buffer
+	proc, err := p.ogcRun(ctx, []string{"up", LayoutIntegrationTag}, false, process.WithCmdOptions(runner.AttachOut(&output), runner.AttachErr(&output)))
```
adam-stokes, line 188: With latest ogc in `:bleed` the layoutintegrationtag should not be needed any longer

pkg/testing/ogc/provisioner.go
```diff
@@ -0,0 +1,338 @@
...
+		return nil, fmt.Errorf("failed to run ogc up: docker run exited with code: %d", ps.ExitCode())
+	}
+	return output.Bytes(), nil
+}
+
+// ogcDown brings down all the instances.
+func (p *provisioner) ogcDown(ctx context.Context) error {
+	p.logger.Logf("Bring down instances through ogc")
+	var output bytes.Buffer
+	proc, err := p.ogcRun(ctx, []string{"down", LayoutIntegrationTag}, false, process.WithCmdOptions(runner.AttachOut(&output), runner.AttachErr(&output)))
```
adam-stokes, line 205: Rather than LayoutIntegrationTag, you can use `ogc down` or `ogc down -q 'instance_name=blah'` for individual ones

pkg/testing/ogc/provisioner.go
```diff
@@ -0,0 +1,338 @@
...
+		SSHPublicKey:  ".integration-cache/id_rsa.pub",
+		Ports:         []string{"22:22"},
+		Tags:          tags,
+		Labels: map[string]string{
+			"division": "engineering",
+			"org":      "platform",
+			"team":     "ingest",
+			"project":  "elastic-agent",
+		},
+		Scripts: "path", // not used; but required by OGC
```
adam-stokes, line 318: No longer required

adam-stokes:
> Alright, so, if you don't have an `.ogc-cache` directory, when you run the test, it'll put things in _some_ kind of weird state that's hard to get out of. Still tinkering with it.

We typically ensure that directory exists on the main entrypoints (up, layout import) etc. Are you saying that you were able to run ogc without it creating the cache dir?

fearful-symmetry:
@adam-stokes no, it appears that running most commands creates that directory, but removing the directory, then running another command, seems to what puts it in that initial error state. Still having trouble with it.

adam-stokes:
> @adam-stokes no, it appears that running most commands creates that directory, but removing the directory, then running another command, seems to what puts it in that initial error state. Still having trouble with it.
> 
> Either way, just running `mage integration:clean` doesn't fix the state once we get into it.

Ok, we want to make sure `.ogc-cache` isn't getting removed, the previous issue with the table missing is resolved.

pierrehilbert:
buildkite test this

pkg/testing/runner/runner.go
```diff
@@ -237,64 +308,77 @@ func (r *Runner) runMachines(ctx context.Context, sshAuth ssh.AuthMethod, repoAr
...
 	// ensure that we have all the requirements for the stack if required
 	env := map[string]string{}
 	if batch.Batch.Stack != nil {
-		ch, err := r.getCloudForBatchID(batch.ID)
-		if err != nil {
-			return OSRunnerResult{}, err
+		// wait for the stack to be ready before continuing
+		r.stacksReady.Wait()
+		if r.stacksErr != nil {
+			return OSRunnerResult{}, fmt.Errorf("failed to save instance state %s: %w", instance.Name, err)
```
aleksmaus, line 359: typo, should be:
```
return OSRunnerResult{}, fmt.Errorf("failed to save instance state %s: %w", instance.Name, r.stacksErr)
```

pkg/testing/ogc/provisioner.go
blakerouse: Going to switch to the newer version in another PR.

pkg/testing/ogc/provisioner.go
blakerouse: When I test out the `:bleed` tag I will make this change.

pkg/testing/ogc/provisioner.go
blakerouse: Will change with `:bleed`.

pkg/testing/runner/runner.go
blakerouse: The error message itself is also wrong. I have updated, thanks for catching that!

blakerouse:
@aleksmaus I will see if we can ensure that `info.Namespace` is unique for each run.

aleksmaus:
> "Defend-"

Or could just derive the the name from the policyID, this seems to sufficient. Don't know if is there is any significance of using info.Namespace in this case.
```
Name:     "Defend-" + packagePolicyID, //info.Namespace,
```

blakerouse:
@aleksmaus Good idea. I have updated that test to do just that.

pkg/testing/runner/runner.go
```diff
@@ -490,288 +562,170 @@ func (r *Runner) setupCloud(ctx context.Context) error {
...
-func (r *Runner) getCloudForBatchID(id string) (chan *ess.CreateDeploymentResponse, error) {
-	r.batchToCloudMx.RLock()
-	essResp, ok := r.batchToCloud[id]
-	if !ok {
-		r.batchToCloudMx.RUnlock()
-		return nil, fmt.Errorf("no batch with ID %s", id)
-	}
-	r.batchToCloudMx.RUnlock()
+func (r *Runner) addOrUpdateInstance(instance StateInstance) error {
+	r.stateMx.Lock()
```
pchila, line 636: are we calling `find*` and `add*` functions from different goroutines? I see a lot of mutex lock/unlock for state...

pkg/testing/runner/utils.go
```diff
@@ -0,0 +1,42 @@
...
+		return nil
+	}
+}
+
+func AttachErr(w io.Writer) process.CmdOption {
+	return func(c *exec.Cmd) error {
+		c.Stderr = w
+		return nil
+	}
+}
```
pchila, line 42: What's the advantage in using this instead of [StdErrPipe](https://pkg.go.dev/os/exec#Cmd.StderrPipe) or [StdOutPipe](https://pkg.go.dev/os/exec#Cmd.StdoutPipe) optionally with an [io.Pipe](https://pkg.go.dev/io#Pipe) if we don't like the ReadCloser types ?

pchila (approved):
Left a few comments, none of which is blocking but I would like to see them answered.
For ease of reviewing these PR, please add some unit testing showing how the various pieces are intended to be used or try to develop iteratively in order to keep a smaller size for the PR.

Since there are already people using this, I am gonna approve anyways in order not to block this till tomorrow

pkg/testing/ess/provisioner.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type provisioner struct {
+	logger runner.Logger
+	cfg    ProvisionerConfig
+	client *Client
+}
+
+// NewProvisoner creates the ESS provisioner
+func NewProvisoner(cfg ProvisionerConfig) (runner.StackProvisioner, error) {
```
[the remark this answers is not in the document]
blakerouse: `runner.StackProvisioner` is an interface, so it is a pointer. I will fix the misspelling.

pkg/testing/ogc/provisioner.go
```diff
@@ -0,0 +1,338 @@
...
+	LayoutIntegrationTag = "agent-integration"
+)
+
+type provisioner struct {
+	logger runner.Logger
+	cfg    Config
+}
+
+// NewProvisoner creates the OGC provisioner
+func NewProvisoner(cfg Config) (runner.InstanceProvisioner, error) {
```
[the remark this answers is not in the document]
blakerouse: Same as above. Will fix misspelling, but reason for return type stays the same.

pkg/testing/runner/provisioner.go
```diff
@@ -0,0 +1,105 @@
...
+	//
+	// The provision should re-use already prepared instances when possible.
+	Provision(ctx context.Context, batches []OSBatch) ([]Instance, error)
+
+	// Clean cleans up all provisioned resources.
+	Clean(ctx context.Context, instances []Instance) error
+}
+
+// InstanceProvisionerCreator creates a new provisioner.
+type InstanceProvisionerCreator func(l Logger, cfg Config) InstanceProvisioner
```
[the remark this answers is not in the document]
blakerouse: Absolutely! That is the whole reason for the refactor. I am already working on a LXD provisioner that will allow me to spin up full ubuntu machines locally on an ubuntu host.

pkg/testing/runner/runner.go
```diff
@@ -192,43 +246,60 @@ func (r *Runner) Run(ctx context.Context) (Result, error) {
 
 // Clean performs a cleanup to ensure anything that could have been left running is removed.
 func (r *Runner) Clean() error {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
-	defer cancel()
-	return r.ogcDown(ctx)
+	r.stateMx.Lock()
+	defer r.stateMx.Unlock()
```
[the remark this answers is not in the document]
blakerouse: Clean is now only called separately from the actual running of the tests. It will always be called from a different process.

pkg/testing/runner/runner.go
```diff
@@ -192,43 +246,60 @@ func (r *Runner) Run(ctx context.Context) (Result, error) {
...
+				logger := &batchLogger{wrapped: r.logger, prefix: i.ID}
+				result, err := r.runInstance(ctx, sshAuth, logger, repoArchive, batch, i)
 				if err != nil {
-					logger.Logf("Failed for instance %s: %s\n", m.PublicIP, err)
+					logger.Logf("Failed for instance %s (@ %s): %s\n", i.ID, i.IP, err)
 					return err
 				}
 				resultsMx.Lock()
 				results[batch.ID] = result
 				resultsMx.Unlock()
```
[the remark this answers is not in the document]
blakerouse: You would have to give me an example of what you mean. It is not clear from your comment what your expecting.

pkg/testing/runner/runner.go
blakerouse: Yes, controlling instances is completely parallelized.

pkg/testing/runner/utils.go
blakerouse: The command execution is the same command execution that is used for the core pieces of Elastic Agent. I prefer to use those command execution as it provides OS level specifics about Pkill and Windows Job so if the parent process dies the child process also dies.

blakerouse:
Seems the stack failed to become ready. Weird we might need to add more retries to the stack provisioner.

blakerouse:
buildkite test this

pkg/testing/ess/provisioner.go
```diff
@@ -0,0 +1,163 @@
...
+		return errors.New("field APIKey must be set")
+	}
+	if c.Region == "" {
+		return errors.New("field Region must be set")
+	}
+	return nil
+}
+
+type provisioner struct {
+	logger runner.Logger
```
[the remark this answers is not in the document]
pchila: Because then each package is not coupled by a common interface definition and any changes to the interface in `runner` package do not affect the `ess` package ? Usually interfaces are always defined at the receiver's end by specifying only the actual functions used... if it's really always the same for all the packages everywhere then we can have a common interface for the whole project with the big disclaimer that any changes would affect the whole codebase

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
pchila: What is really important is that the object returned by the New function fits the interface as defined by the `runner` package.
The compiler will ensure that the object implements such interface regardless of the type used as return value.

Setting the interface as a return value here introduces a dependency from `ess` to `runner` where there's no need for one

pkg/testing/runner/provisioner.go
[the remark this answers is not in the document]
pchila: So we are gonna have tests running both ogc and lxd provisioners at the same time, in the same run, switching between the 2 ? Is that why we need a Factory pattern as defined here ?

pkg/testing/runner/utils.go
pchila: I fail to see how what  you are describing is related to how we attach to stdout and stderr... 
What I meant was that these two functions are a very thin layer around setting up an `exec.Cmd` and in a lot of places we are using them just to dump output and error in the same buffer. So I was wondering: what was wrong with using directly the functions exposed by `exec.Cmd` ?

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
blakerouse: I prefer to know from the compiler that it doesn't implement the interface for the `runner.StackProvisioner` inside of the module that implements that provisioner. I do not want to wait until its used before I know if it implements that interface. I want the compiler to say it doesn't implement the interface at the new method, so its clear that this creator function does not implement that interface.

pkg/testing/runner/provisioner.go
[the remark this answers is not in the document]
blakerouse: The original idea was that it would be defined in a registry, but I removed the registry and changed it to just be defined by the creator of the runner. So both `InstanceProvisionerCreator` and `StackProvisionerCreator` can be removed. Follow up commit will remove them.

The idea is that you can use different provisioners, not at the same time in the same run but when running the tests.

There is already work done to switch from OGC to terraform that will be done first with a flag `STACK_PROVISIONER="terraform" mage integration:test`. I am also working on the LXD provider that will allow the same `STACK_PROVISIONER="lxd" mage integration:test`.

pkg/testing/runner/runner.go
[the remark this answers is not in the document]
blakerouse: Your not wrong the mutex could be removed here. I keep it hear so its clear that the `r.state` is a shared state and in the case that it was used from 2 goroutines it would be used correctly. It currently is not used from 2 goroutines, but when reading the code its clear that the state is protected by a lock.

So in short it could be removed, but I don't because of above.

pkg/testing/runner/utils.go
blakerouse: The difference is because we merge the stdout and stderr into a single buffer where the output is combined. `StdErrPipe` and `StdOutPipe` provide unique buffers.

blakerouse:
Stack failure yet again! Possible that 8.10-SNAPSHOT in cloud is very very flaky. Next commit for retry.

Issue #2579, opened 2023-05-01
Add the full version number to the installation directory name


joshdover:
## Dependencies ##
- Blocked by https://github.com/elastic/elastic-agent/issues/2873
- Blocked by https://github.com/elastic/elastic-agent/issues/2582
- Blocked by ttps://github.com/elastic/elastic-agent/issues/3639

## Issue ##

We need the full version number of agent to be present in the directory we install into. This is necessary for changing how we release agent.

## Suggested solution ##

Add the full version number to the installation directory name.
Example:
```
data/
└── elastic-agent-7ac68h-8.8.1
└── elastic-agent-7ac68h-8.8.1.er1
```

## Definition of done ##

- [ ] Installation directory now includes the full version number.
- [ ] Add a new integration test [here](https://github.com/elastic/elastic-agent/blob/main/testing/integration/install_test.go) to validate that installation directory is the correct one.

cmacknz:
Right now the upgrade process temporarily has multiple versions of the agent extracted, so it can rollback. The agent versions are installed under the /data directory with the short Git SHA as the version identifier:

data/
└── elastic-agent-383752
└── elastic-agent-7ac68h

This works fine besides being impossible to tell which version is which by manual inspection. This approach fails if we want the commit hash of the agent itself to be the same for two emergency releases. What you would want instead is something like:

data/
└── elastic-agent-7ac68h-8.8.1
└── elastic-agent-7ac68h-8.8.1.er1

This allows us to keep using the agent Git SHA in the path and also makes it obvious which version is which, which I want for convenience anyway. These `elastic-agent-xxxxx` paths now show up in our diagnostics and make debugging one step less convenient.

Another approach to this would be to make the unique identifier something other than the agent Git SHA, like computing the hash of the agent package contents at build time. This would also be unique, but the directory name wouldn't be something you could look up in Git. We would need to rely on a manifest file listing the detailed version information of each component of the agent package, but we may want that anyway.

pchila:
Since we have to change the format of the directories and stamp out new versions, I have 1 question and (maybe) 1 proposal.
The question first: is the precedence relationship below correct? :
`8.8.0 < 8.8.1-er.1 < 8.8.1`

If the precedence is correct, can we just have the directories to follow the semver version strings like
```
data/
└── elastic-agent-8.8.1+7ac68h
└── elastic-agent-8.8.1-er.1+7ac68h

```
@cmacknz , @joshdover  what do you think ?

[...]

PR #3140, opened 2023-07-27
Spawn the watcher of the new binary vs the current running version.

### LINKED ISSUE #2873: When upgrading, the elastic-agent watch command should be invoked using the new agent binary
When the agent upgrades it launches an instance of the `elastic-agent watch` command to ensure that the new version of the elastic agent remains healthy and does not crash within 10 minutes of being started.

The `elastic-agent` version invoked is the old version we are upgrading *from*, not the new version we are upgrading *to*:

https://github.com/elastic/elastic-agent/blob/9c062312d1952e68cd928b12d49dde8ec16865cb/internal/pkg/agent/application/upgrade/rollback.go#L109-L118

The `elastic-agent watch` command should instead be invoked using the new version of the elastic-agent binary we are upgrading to. For example, in an upgrade from 8.7.1 to 8.8.2 we should invoke the `elastic-agent watch` command using the 8.8.2 `elastic-agent` binary.

1. This means the upgrade watcher can be aware of any changes in the Elastic Agent that are introduced in the new version of the upgrade, for example the changes in https://github.com/elastic/elastic-agent/issues/2579
2. This protects from the situation where there is a bug in the upgrade watcher by allowing us to solve it by upgrading to a different version. Today a bug in the upgrade watcher that would prevent a successful upgrade cannot be fixed without a reinstall of the existing agent.

blakerouse:
## What does this PR do?

Spawns the new version of the Elastic Agent watcher when upgrading versus using the current running Elastic Agent version to perform the watch.

## Why is it important?

The upgrade watcher can be aware of any changes in the Elastic Agent that are introduced in the new version of the upgrade. Protects from the situation where there is a bug in the upgrade watcher by allowing us to solve it by upgrading to a different version. 

## Checklist

## Related issues

- Closes #2873

### DIFF
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -21,7 +21,6 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/agent/errors"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/install"
 	"github.com/elastic/elastic-agent/internal/pkg/core/backoff"
-	"github.com/elastic/elastic-agent/internal/pkg/release"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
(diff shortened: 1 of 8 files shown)


internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -108,14 +109,18 @@ func Cleanup(log *logger.Logger, currentHash string, removeMarker bool, keepLogs
...
 	}
 
-	versionedHome := paths.VersionedHome(paths.Top())
-	cmd := invokeCmd(versionedHome)
+	if targetHash == "" {
+		// when targetHash is not provided then we use the current running hash
+		targetHash = release.ShortCommit()
+	}
+	hashedDir := fmt.Sprintf("%s-%s", agentName, targetHash)
+	cmd := invokeCmd(filepath.Join(paths.Top(), "data", hashedDir))
```
pchila, line 123: This clashes with https://github.com/elastic/elastic-agent/issues/2579
Since by the time we invoke the watcher we already switched the symlink, can't we just call the watcher from home ?

pchila (approved):
LGTM

pchila:
I just noticed that some integration tests failed but buildkite is green... THere is something wrong in the integration test script...

blakerouse:
@pchila That is interesting, we need to fix that.

blakerouse:
buildkite test this

PR #3145, opened 2023-07-27
Fix E2E tests skipping if requirement OS set to ARM64

aleksmaus:
This fixes the issue with E2E tests being skipped if they target only ARM64 arch.

For example, the test with the following spec
1. was not be included for ARM64 arch tests
2. was attempted to be run on AMD64 arch and skipped
```
		Stack:   &define.Stack{},
		Local:   false, // requires Agent installation
		Isolate: false,
		Sudo:    true, // requires Agent installation
		OS: []define.OS{
			define.OS{Type: define.Linux, Arch: define.ARM64},
		},
	})
```

executed as single test, example:
```
mage integration:single TestInstallWithEndpointSecurityAndRemoveEndpointIntegration
```

After the change the test is only executed on ARM64.

## What does this PR do?

Fixes what seems to be the typo.

## Why is it important?

E2E correctness.

## Checklist


aleksmaus:
> Please remove the changelog entry, we only create changelog entries for end user facing changes and fixes. Being the integration suite is only used by developers and CI it doesn't get changelog entries.

PR was failing the missing changelog step, didn't know it's optional. Removed.

blakerouse:
When you know a changelog is not needed you can add the skip-changelog label and it will bypass the check.

PR #3149, opened 2023-07-28
Fix mage and go buildkit script exit codes

### LINKED ISSUE #3148: CI `mage integration:test` no longer reports as failed when a test does fail
Seems that there was a change to how mage is executed in BuildKit. The `common.sh` script that is sourced in the `integration_test.sh` has a `mage()` function. This change means that the `integration_test.sh` runs the `mage()` script instead of the actual mage binary directory.

Because of this the following happens:

```
>>> JUnit XML written here: build/TEST-go-integration.xml
--
  | + popd
  | ~/builds/bk-agent-prod-gcp-1690558472259851007/elastic/elastic-agent
  | + TESTS_EXIT_STATUS=0
```

That means that the exit code for `mage integration:test` is no longer the last command, `popd` is. Which means that `TESTS_EXIT_STATUS=$?` is no longer getting the exit code from `mage integration:test` it is getting it from `popd` which always returns 0.

blakerouse:
## What does this PR do?

Fixes the buildkit job to fail when `mage` or `go` functions actual fail.

## Why is it important?

Without this change CI integration failures will not mark the tests as failed.

## Related issues

- Closes #3148

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -58,7 +58,9 @@ mage() {
     fi
     pushd "$WORKSPACE"
     command "mage" "$@"
+    ACTUAL_EXIT_CODE=$?
     popd
+    return $ACTUAL_EXIT_CODE
 }
 
 # Wrapper function for executing go
@@ -78,7 +80,9 @@ go(){
     fi
     pushd "$WORKSPACE"
     command go "$@"
+    ACTUAL_EXIT_CODE=$?
     popd
+    return $ACTUAL_EXIT_CODE
 }
 
 google_cloud_auth() {


pchila (approved):
LGTM

blakerouse:
The failing test shows this clearly fixes the issue. What is interesting is the failure, possible that the test was merged thinking that it worked, but really it didn't.

blakerouse:
buildkite test this

blakerouse:
The failures in the integration tests from this PR is from the fact that the integration testing is actually failing. The previous changed hide that fact and I believe some changes or new tests have broken the integration tests.

This should be force merged, so we are forced to fix the CI.

Issue #3155, opened 2023-07-31
TestStandaloneUpgradeRetryDownload is failing regularly on main


cmacknz:
## Flaky Test

* **Test Name:** TestStandaloneUpgradeRetryDownload
* **Link:** https://github.com/cmacknz/elastic-agent/blob/0158f7755cd411443587211b3afa89215b7fd201/testing/integration/upgrade_test.go#L315-L318
* **Branch:** main
* **Artifact Link:** https://buildkite.com/elastic/elastic-agent/builds/2160#0189acba-5002-471d-99e3-b759b8733356
* **Notes:** First noticed in https://github.com/elastic/elastic-agent/pull/3153

### Stack Trace

```
=== RUN   TestStandaloneUpgradeRetryDownload
    upgrade_test.go:384: Testing Elastic Agent upgrade from 8.10.0-SNAPSHOT to 8.8.0...
    fetcher.go:90: Using existing artifact elastic-agent-8.10.0-SNAPSHOT-linux-arm64.tar.gz
    fixture.go:200: Extracting artifact elastic-agent-8.10.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneUpgradeRetryDownload2438491427/001
    fixture.go:213: Completed extraction of artifact elastic-agent-8.10.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneUpgradeRetryDownload2438491427/001
    fixture.go:511: Components were not modified from the fetched artifact
    upgrade_test.go:398: Install the built Agent
    fixture.go:344: >> running agent with: [/tmp/TestStandaloneUpgradeRetryDownload2438491427/001/elastic-agent-8.10.0-SNAPSHOT-linux-arm64/elastic-agent install --force --non-interactive]
    upgrade_test.go:400: Installing in non-interactive mode.Elastic Agent has been successfully installed.
        
    upgrade_test.go:403: Ensure the correct version is running
    fixture.go:344: >> running agent with: [/opt/Elastic/Agent/elastic-agent version --yaml]
    upgrade_test.go:502: binary:
          version: 8.10.0
          commit: 0158f7755cd411443587211b3afa89215b7fd201
          build_time: 2023-07-31T13:54:35Z
          snapshot: true
        could not get version. failed to communicate with running daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/elastic-agent.sock: connect: no such file or directory"
[...]

PR #3156, opened 2023-07-31
Skip flaky retry upgrade test.

cmacknz:
This will fix the build on main and avoid wasting time while we investigate the root cause.

See https://github.com/elastic/elastic-agent/issues/3155

This test fails often enough for me that we just need to disable it while we investigate.


blakerouse (approved):
Yes it needs to be disabled.

PR #3163, opened 2023-08-01
Add auto diagnostic collection on test failure and ability to skip uninstall for inspection of installed agent

### LINKED ISSUE #2953: Testing framework makes it difficult to debug issues during setup and install phase
## Issue

Right now, if something in a remote test fails such that the resulting agent install isn't communicating with the stack, or isn't sending events in the way we expect, there's no easy way to fetch diagnostics or tell what went wrong.

## Definition of done

- [ ] Integration test collects a diagnostics bundle if it detects a test failure
- [ ] A flag/env variable is in place to stop the cleanup process, so we don't remove the VMs or elastic cluster in case of failure

blakerouse:
## What does this PR do?

This provides the automatic collection of diagnostics when a tests fails in the integration testing suite. Including providing them as artifacts to buildkite so they can be used.

This also includes 2 environment variable options to help debug using the integration testing framework.

First one is the `AGENT_COLLECT_DIAG=true` option that instructs the testing fixture to collect the diagnostics of the Elastic Agent that is installed before performing the uninstall. (Doesn't have to be a failure for the collection)

Second one is the `AGENT_KEEP_INSTALLED=true` option that prevents the uninstallation of the Elastic Agent at the end of the test. Should only be used with the `mage integration:single [testName]` but helps with the ability to debug the still running Elastic Agent. With the latest change of not bring down the instances at the end, its a nice helper for inspection.

## Why is it important?

To help with debugging the Elastic Agent and tests.

## Checklist

```
$ AGENT_COLLECT_DIAG=true mage integration:test
$ ls build/diagnostics/
```

```
$ AGENT_KEEP_INSTALLED=true mage integration:single TestInstallWithoutBasePath
$ # ssh into the host see that the Elastic Agent is still installed
$ mage integration:clean # needs to be cleaned as its now installed
```

## Related issues

- Closes #2953


blakerouse:
Build that failed on purpose shows the artifacts include the diagnostics - https://buildkite.com/elastic/elastic-agent/builds/2201#0189b244-6706-4a51-8e14-78c586971027

pkg/testing/runner/debian.go
```diff
@@ -207,6 +207,41 @@ func (DebianRunner) Run(ctx context.Context, verbose bool, sshClient *ssh.Client
...
+		return nil
+	}
+	eachDiagnostic := strings.Split(string(stdOut), "\n")
+	for _, filename := range eachDiagnostic {
+		filename = strings.TrimSpace(filename)
+		if filename == "" {
+			continue
+		}
+
+		// don't use filepath.Join as we need this to work in Windows as well
```
cmacknz, line 227: Is the problem here that filepath.Join will format this for the host system instead of the remote system?

On first read this confused me since filepath.Join is supposed be cross-platform.

cmacknz: Can you improve the comments to they state the problem you are avoiding directly?

cmacknz (commented):
Looks great, was just thinking we needed something like this.

You need to upgrade the testing development guide with the information from the PR description though, https://github.com/elastic/elastic-agent/blob/main/docs/test-framework-dev-guide.md

pkg/testing/runner/debian.go
blakerouse: I improved the comment.

blakerouse:
@cmacknz I have added information to the development guide.

docs/test-framework-dev-guide.md
```diff
@@ -96,6 +85,37 @@ We pass a `-test.run` flag along with the names of the tests we want to run in O
...
+- `mage integration:clean` will de-provision the allocated resources and cleanup any local state.
+
+Tests with external dependencies might need more environment variables to be set
+when running them manually, such as `ELASTICSEARCH_HOST`, `ELASTICSEARCH_USERNAME`,
+`ELASTICSEARCH_PASSWORD`, `KIBANA_HOST`, `KIBANA_USERNAME`, and `KIBANA_PASSWORD`.
+
+### Debugging tests
+
+#### Auto diagnostics retrieval
+When an integration test fails and the testing fixture will try its best to automatically collect the diagnostic
```
AndersonQ, line 102: typo?
```suggestion
When an integration test fails the testing fixture will try its best to automatically collect the diagnostic
```

pkg/testing/fixture_install.go
```diff
@@ -174,3 +200,34 @@ func (f *Fixture) Uninstall(ctx context.Context, uninstallOpts *UninstallOpts, o
...
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+
+	dir, err := findProjectRoot(f.caller)
+	if err != nil {
+		f.t.Logf("failed to collect diagnostics; failed to find project root: %s", err)
+		return
+	}
+	diagPath := filepath.Join(dir, "build", "diagnostics")
+	_ = os.MkdirAll(diagPath, 0755)
```
AndersonQ, line 214: [Blocker]
Why are you ignoring the error? If it fails there will be a cascading effect, better to handle it and stop in case the directory cannot be created.

AndersonQ (changes requested):
Looks good, but there is an error I believe needs to be handled

docs/test-framework-dev-guide.md
blakerouse: Fixed.

pkg/testing/fixture_install.go
blakerouse: Fixed.

AndersonQ (approved):

blakerouse:
buildkite test this

Issue #3154, opened 2023-07-31
[Flaky Test] Integration/e2e tests installing the agent randomly fail


AndersonQ:
## Flaky Test

* **Test Name:** Any tests installing the agent.
* **Link:** N/A.
* **Branch:** main.
* **Artifact Link:** A few examples
   - https://buildkite.com/elastic/elastic-agent/builds/2151#0189abf3-c343-4eb5-b99c-c90655f37481
   - https://buildkite.com/elastic/elastic-agent/builds/2135#01899de3-d93e-4a98-9268-80fd98db9d18
   - https://buildkite.com/elastic/elastic-agent/builds/2126#01899d2d-112f-4b40-b5a5-a9748087a09e

### Notes

 - For some reason tests installing the agent sometimes get stuck with the agent never finishing fetching the 1st config from fleet-server. 
 - It also happens for tests with the mock fleet-server, which leads to believe the issue is on the agent, its installation process or somehow on the host. 
 - Something is preventing the agent finishing its enroll process or making the agent, during the install/enroll process, to start in standalone mode.
 - It also does not seem to be related to the usage of `testify/suite` with SetUp/TearDown as the ProxyURL test do not use it, and other failing tests use `testify/suite`.

Once I spotted this issue while running the tests on a Vagrant box, and I managed to gather the following:

The agent reported not enrolled into fleet, however it did call /enroll and the request succeeded.

in the test logs I can see the call to fleet through the proxy:
```
    fixture.go:344: >> running agent with: [/tmp/TestProxyURLTestEnrollProxyAndNoProxyInThePolicy3423221791/001/elastic-agent-8.10.0-SNAPSHOT-linux-arm64/elastic-agent install --force --insecure --non-interactive --proxy-url=http://localhost:36529 --url http://fleet.elastic.co --enrollment-token anythingWillDO]
    proxytest.go:73: [proxy-proxy-1] [935eef36-5695-4922-8031-4860b131bca9] STARTING - POST http://fleet.elastic.co/api/fleet/agents/enroll? HTTP/1.1 127.0.0.1:36150
    proxytest.go:73: [proxy-proxy-1] original URL: http://fleet.elastic.co/api/fleet/agents/enroll?, new URL: http://localhost:42557/api/fleet/agents/enroll?
[...]

PR #3164, opened 2023-08-01
Disable proxy tests while flakiness is addressed.

cmacknz:
All of the proxy tests are currently flaky and fail regularly, see https://github.com/elastic/elastic-agent/issues/3154

Disabling them while we investigate a fix to unblock CI.


rdner:
Unskipped by https://github.com/elastic/elastic-agent/pull/4321

PR #3165, opened 2023-08-01
Fix diagnostic scrubbing errors

fearful-symmetry:
Closes https://github.com/elastic/ingest-dev/issues/1791

## What does this PR do?

This fixes the redaction component of the diagnostic bundle, so we should actually be redacting passwords, key names, etc. There's a few things going on:

- add a custom YAML marshaler to the `ComponentUnitKey` type
- fix up `redactMap` to make it a little more hardy
- change the diagnostic bundler to only error out if all data collection fails
- Add a bunch of tests

## Why is it important?

This has been an ongoing bug for a while

## Checklist


fearful-symmetry:
Bah, don't have access to buildkite, can't see the failures

cmacknz:
Does this correctly redact SSL private keys as well?

This is one of the problems I was originally observing when the original bug was filed.

Screen Shot 2023-01-11 at 10 00 30 AM

fearful-symmetry:
@cmacknz So, as long as its in a hashmap with key `"certificate"`, it should.

cmacknz:
Ah, the path that isn't getting redacted is `ssl.key` https://www.elastic.co/guide/en/fleet/current/elastic-agent-ssl-configuration.html

fearful-symmetry:
Ah, yah, `key` should be redacted too:
```
strings.Contains(k, "key")
```

fearful-symmetry:
I think the buildkite error is unrelated?

pierrehilbert:
I think so we don't have anything in the artifact so can't tell.
@blakerouse / @pazone why the integration output is missing here? Is it relative to the ` File not found: build/diagnostics/*`

pazone:
The buildkite job was killed.

internal/pkg/diagnostics/diagnostics_test.go
```diff
@@ -20,13 +20,98 @@ import (
...
+	t.Logf("%s", formatted)
+	errOut := strings.Builder{}
+	outWriter := strings.Builder{}
+	res := client.DiagnosticFileResult{Content: formatted, ContentType: "application/yaml"}
+
+	err = writeRedacted(&errOut, &outWriter, "test/path", res)
+	require.NoError(t, err)
+
+	require.Empty(t, errOut.String())
+	require.NotContains(t, outWriter.String(), "unredacted")
```
AndersonQ, line 112: Am I missing something, or there is no `unredacted` on `exampleState`? This `require` seems to have no effect

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -324,23 +324,46 @@ func writeRedacted(errOut, w io.Writer, fullFilePath string, fr client.Diagnosti
...
-			v = redactMap(v.(map[interface{}]interface{}))
-		}
-		if s, ok := k.(string); ok {
-			if redactKey(s) {
-				v = REDACTED
+// Redact sensitive values from the underlying map
+// the whole generic function here is out of paranoia. Although extremely unlikely,
+// we have no way of guaranteeing we'll get a "normal" map[string]interface{},
+// since the diagnostic interface is a bit of a free-for-all
+func redactMap[K comparable](errOut io.Writer, inputMap map[K]interface{}) map[K]interface{} {
```
AndersonQ, line 335: [Suggestion|Go convention-ish]

```suggestion
// redactMap redacts sensitive values from the underlying map
// the whole generic function here is out of paranoia. Although extremely unlikely,
// we have no way of guaranteeing we'll get a "normal" map[string]interface{},
// since the diagnostic interface is a bit of a free-for-all
func redactMap[K comparable](errOut io.Writer, inputMap map[K]interface{}) map[K]interface{} {
```

AndersonQ (approved):

internal/pkg/diagnostics/diagnostics_test.go
```diff
@@ -20,13 +20,98 @@ import (
...
+		"root": mapstr.M{
+			"passphrase": "unredacted",
+			"nested1": mapstr.M{
+				"certificate": "unredacted",
+				"nested2": mapstr.M{
+					"passphrase": "unredacted",
+					"password":   "unredacted",
+					"nested3": mapstr.M{
+						"token": "unredacted",
+						"key":   "unredacted",
```
cmacknz, line 46: Can you use an inline PEM formatted key here (obviously don't use a real one)? We should test that this properly redacts multiline strings.

internal/pkg/diagnostics/diagnostics_test.go
fearful-symmetry: Ahh, yup.

internal/pkg/diagnostics/diagnostics_test.go
fearful-symmetry: Ah, good idea...

internal/pkg/diagnostics/diagnostics_test.go
```diff
@@ -61,6 +112,7 @@ func TestRedactResults(t *testing.T) {
 
 	require.Empty(t, errOut.String())
 	require.NotContains(t, outWriter.String(), "unredacted")
+	require.NotContains(t, outWriter.String(), "END OPENSSH PRIVATE KEY")
```
cmacknz, line 115: nit: you could just put the key in a variable and check for all of it. This leaves us open to a bug where we only remove the last line or something.

cmacknz (approved):

PR #3189, opened 2023-08-04
[Developer Documentation] Add details of Upgrade Watcher interactions with Agent and upgrade marker file

ycombinator:
## What does this PR do?

This PR enhances the sequence diagram in https://github.com/elastic/elastic-agent/blob/main/docs/upgrades.md to include more details about the interplay between the Upgrade Watcher, the upgrade marker file, and the Agents (old or new).

## Why is it important?

To understand exactly the steps taken by the Upgrade Watcher during an ongoing upgrade, either to decide that the new (post-upgrade) Agent is running OK or that it needs to be rolled back to the old (pre-upgrade) Agent.

### DIFF
--- a/docs/upgrades.md
+++ b/docs/upgrades.md
@@ -38,19 +38,26 @@ sequenceDiagram
        A->>A: Extract new Agent artifact
        A->>A: Replace current Agent artifact with new one
        A->>UM: Create
+       A->>A: Update active commit file
        A->>UW: Start
        A->>A: Rexec to start new Agent artifact
        A->>FS: Ack successful upgrade
-       UW->>UM: Remove
        FS->>ES: Write successful ack in `.fleet-actions-results`
        FS->>ES: Update Agent doc in `.fleet-agents`<br />set `upgrade_status` = null<br />`upgraded_at` = <now><br />`upgrade_started_at` = null
        UI->>UI: Show Agent status as "healthy"
-   end
-   opt Rollback
-       UW->>A: Start
-       A->>FS: Ack failed upgrade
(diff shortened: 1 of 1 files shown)


pierrehilbert:
Thanks @ycombinator for this new part of the doc!

pierrehilbert (approved):

docs/upgrades.md
```diff
@@ -38,19 +38,26 @@ sequenceDiagram
...
-       FS->>ES: Update Agent doc in `.fleet-agents`<br />set `upgrade_status` = null<br />`upgraded_at = <now>
-       UI->>UI: Show Agent status as "healthy"
-       UW->>UM: Remove
-   end
+       UW->>UW: Start watching new Agent
+       alt New Agent is OK
+         UW->>UM: Remove
+         UW->>UW: Cleanup old Agent files
+       else Rollback
+         UW->>UW: Replace current Agent artifact with old one
```
cmacknz, line 53: This is just a change to the symlink correct? Is it worth clarifying this to specifically mention the word symlink? 

This could read as if it is doing something involved by manipulating the extracted agent, but it just leaves that alone.

The same applies to the same wording on Line 39 above.

ycombinator: Yes, you're correct.  I don't know why I decided to use the less concrete wording here instead 🤷.  Will fix in both places.  Thanks.

ycombinator: Fixed in b2084030f37c0f161e645a4792153f2deb28e647.

cmacknz (approved):

PR #3202, opened 2023-08-07
Adds an instance provisioner for multipass for integration testing locally

blakerouse:
## What does this PR do?

Adds a new `multipass` integration instance provisioner. `multipass` is a very simple CLI that makes it easy to get an Ubuntu VM.

## Why is it important?

When working on integration tests is much quicker if the instances is local to the developer. Using multipass provides a quick way of getting an Ubuntu machine.

## Checklist

```
$ brew install multipass
$ mage integration:clean  # clean first before changing instance type
$ INSTANCE="multipass" mage integration:test
```

## Logs

```
>>> Creating zip archive of repo to send to remote hosts
>>> Create SSH keys to use for SSH
>>> Creating stack 8.10.0-SNAPSHOT (8100-SNAPSHOT)
>>> Launching multipass image linux-arm64-ubuntu-2204
>>> (linux-arm64-ubuntu-2204) Starting SSH; connect with `ssh -i /Users/blake/Go/src/github.com/elastic/elastic-agent/.integration-cache/id_rsa ubuntu@192.168.64.14`
>>> (linux-arm64-ubuntu-2204) Connected over SSH
>>> (linux-arm64-ubuntu-2204) Preparing instance
```


pierrehilbert:
Shouldn't we add a test for this multipass thing?

pkg/testing/multipass/provisioner.go
```diff
@@ -0,0 +1,229 @@
...
+		return fmt.Errorf("failed to marshal cloud-init configuration: %w", err)
+	}
+
+	var output bytes.Buffer
+	p.logger.Logf("Launching multipass image %s", batch.ID)
+	proc, err := process.Start("multipass", process.WithContext(ctx), process.WithArgs(args), process.WithCmdOptions(runner.AttachOut(&output), runner.AttachErr(&output)))
+	if err != nil {
+		return fmt.Errorf("failed to run multipass launch: %w", err)
+	}
+	_, err = proc.Stdin.Write([]byte(fmt.Sprintf("#cloud-config\n%s", cfgData)))
```
AndersonQ, line 154: So, is the multipass config using `cloud-config` for local VMs? Is so, the ignore the comment above.

pkg/testing/multipass/provisioner.go
```diff
@@ -0,0 +1,229 @@
...
+// list all the instances.
+func (p *provisioner) list(ctx context.Context) (map[string]instance, error) {
+	cmd := exec.CommandContext(ctx, "multipass", "list", "--format", "yaml")
+	result, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("failed to run multipass list: %w", err)
+	}
+
+	// yaml output from multipass gives a list of instances for each instance name,
+	// even though there is only every 1 entry in the list
```
AndersonQ, line 207: typo? ever?
```suggestion
	// yaml output from multipass gives a list of instances for each instance name,
	// even though there is only ever 1 entry in the list
```

AndersonQ (changes requested):
There is one real blocker, everything else are just suggestions, take or ignore them, it's up to you

pkg/testing/multipass/provisioner.go
```diff
@@ -0,0 +1,229 @@
...
+	for _, instance := range instances {
+		func(instance runner.Instance) {
+			deleteCtx, deleteCancel := context.WithTimeout(ctx, 5*time.Minute)
+			defer deleteCancel()
+			err := p.delete(deleteCtx, instance)
+			if err != nil {
+				// prevent a failure from stopping the other instances and clean
+				p.logger.Logf("Delete instance %s failed: %s", instance.Name, err)
+			}
+		}(instance)
```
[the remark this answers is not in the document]
blakerouse: The anonymous function is used because of the defer deleteCancel() to ensure that its always cleaned up.

pkg/testing/multipass/provisioner.go
blakerouse: Yes, you will find this being used by all distros to provide a way of configuring a new instance.

pkg/testing/multipass/provisioner.go
```diff
@@ -0,0 +1,229 @@
...
+
+// Clean cleans up all provisioned resources.
+func (p *provisioner) Clean(ctx context.Context, instances []runner.Instance) error {
+	// doesn't execute in parallel for the same reasons in Provision
+	// multipass just cannot handle it
+	for _, instance := range instances {
+		func(instance runner.Instance) {
+			deleteCtx, deleteCancel := context.WithTimeout(ctx, 5*time.Minute)
+			defer deleteCancel()
+			err := p.delete(deleteCtx, instance)
```
aleksmaus, line 113: do you want to ```prune``` as well here?

docs/test-framework-dev-guide.md
```diff
@@ -184,6 +184,17 @@ Some examples of where namespaces should be used:
 :warning: Not using namespaces when accessing data in a shared persistent store can cause tests to
 be flaky.
 
+## Alternative Providers
+
+### Multipass Instance Provisioner
+By default the integration testing suite uses OGC with GKE to provision instances. In the case that you
+want to use a local VM instead of a remote VM, you can use the [Multipass](https://multipass.run/) provisioner.
+
+- `INSTANCE="multipass" mage integration:test`
```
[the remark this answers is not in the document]
AndersonQ: I get it, and I believe changing both is even better if you don't mind :)

pkg/testing/multipass/provisioner.go
blakerouse: Do you mean `purge`? If so, I do that in the delete call with `multipass delete -p [instanceName]`

docs/test-framework-dev-guide.md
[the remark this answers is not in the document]
blakerouse: I have updated it to `INSTANCE_PROVIDER` and `STACK_PROVIDER`.

cmacknz:
This is awesome, I tried this out and I was able to run one of the endpoint security tests in 30s!

`SNAPSHOT=true INSTANCE="multipass" mage integration:single   2.63s user 4.30s system 22% cpu 31.108 total`

blakerouse:
@AndersonQ can you give this another look?

cmacknz:
This is so much faster we should consider having it be the default execution mode for Linux VMs, and having CI specifically opt in to cloud VMs.

@blakerouse is there any reason not to do this?

blakerouse:
@cmacknz `multipass` is very specific to Ubuntu only. We need to enable Windows, CentOS, SLES, and macOS (don't think that one can be on GKE any way). I prefer to leave the default to using GKE because of that reason. I would prefer that developers are testing there changes on all 3 supports types.

I did have another idea that maybe could provide a middle ground. I was thinking about adding a `integration.cfg.yaml` that developers could configure that would allow them to set the defaults that they prefer versus always having to change using ENVIRONMENT variables.

cmacknz:
A local config file would be helpful, doesn't need to be done as part of this change though.

fearful-symmetry:
So, took me a while to get `multipass` working on Fedora (blame qemu) but otherwise this is pretty reasonable?

nimarezainia:
@blakerouse @cmacknz is this something that can be shared with the rest of the integrations developers (security and obs). Seems very beneficial, perhaps a snippet in the weekly email.

AndersonQ (approved):

PR #3203, opened 2023-08-08
Fix Elastic Defend removal for tamper protected Agent. Add E2E test coverage for tamper protection.

aleksmaus:
## What does this PR do?

* Fixes intermittent issue that leads to Agent being unhealthy and stopping to check-in for up to 10 mins after Elastic Defend removal from the policy with Endpoint Tamper Protection Enabled. This issue was discovered during E2E tests development and was often reproducible with the test.
* Adds E2E test coverage for tamper protection use cases.

There are lots of Kibana API wrappers that were already moved to https://github.com/elastic/elastic-agent-libs and there is one PR https://github.com/elastic/elastic-agent-libs/pull/143 that would allows to move some more cruft around CreatePolicy API wrapper (@cmacknz). Once that PR is merged I will update this one and expect the size of this one to be trimmed quite a bit.

## Why is it important?

Fixes the issue and expands E2E coverage.

## Checklist


pierrehilbert:
buildkite test this

aleksmaus:
Pulled the fleet API wrappers that are moved now to elastic-agent-libs. Going to let CI go through the run. Then do the release of elastic-agent-libs 0.3.11 and use that tag. Expecting one more commit to this PR.

aleksmaus:
@pierrehilbert it looks like the buildkite junit2html step is failing as far as I understand 
```
+ junit2html
--
  | + exit 1
  | 🚨 Error: The command exited with status 1
```

pkg/component/runtime/service.go
```diff
@@ -147,23 +147,22 @@ func (s *serviceRuntime) Run(ctx context.Context, comm Communicator) (err error)
...
 	}
 
-	processTeardown := func(am actionMode, signed *component.Signed) error {
+	processTeardown := func(am actionMode, signed *component.Signed) {
 		s.log.Debugf("start teardown for %s service", s.name())
 		// Inject new signed
 		newComp, err := injectSigned(s.comp, signed)
 		if err != nil {
 			s.log.Errorf("failed to inject signed configuration for %s service, err: %v", s.name(), err)
-			return err
```
cmacknz, line 156: Why don't we return on error here? If there's a reason it should be a comment as I could see this being added back later by someone without context.

pkg/component/runtime/service.go
```diff
@@ -226,7 +225,7 @@ func (s *serviceRuntime) Run(ctx context.Context, comm Communicator) (err error)
 				s.forceCompState(client.UnitStateFailed, err.Error())
 			}
 		case newComp := <-s.compCh:
-			s.processNewComp(newComp, comm)
+			s.processNewComp(newComp, comm, true)
```
cmacknz, line 228: Why do you sometimes need to send the observed state and sometimes not? Can this be clarified with comments?

pkg/testing/tools/agents.go
```diff
@@ -21,7 +21,7 @@ import (
 
 // GetAgentByHostnameFromList get an agent by the local_metadata.host.name property, reading from the agents list
 func GetAgentByHostnameFromList(client *kibana.Client, hostname string) (*kibana.AgentExisting, error) {
-	listAgentsResp, err := client.ListAgents(kibana.ListAgentsRequest{})
+	listAgentsResp, err := client.ListAgents(context.Background(), kibana.ListAgentsRequest{})
```
cmacknz, line 24: Would it make more sense for the context to be passed in as the first argument to `GetAgentByHostnameFromList`?

pkg/testing/tools/fleet_package_policy_api.go
```diff
@@ -1,106 +0,0 @@
-// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
cmacknz, line 1: Thanks!

cmacknz (commented):
Looks good, couple of questions about the changes to the service runtime because I can't tell exactly how the bug was fixed from reading the code.

pkg/testing/tools/agents.go
aleksmaus: That would change another dozen on places, wanted to limit the change.

pkg/component/runtime/service.go
aleksmaus: The implementation before was incorrect it was it was assigning the error returned from processTeardown to the outer scope error returned from Run
https://github.com/elastic/elastic-agent/blob/e8f79865c5c1747d9be5bf1581bc765c89cf9d13/pkg/component/runtime/service.go#L181

The ```processTeardown``` is only called from ```onTeardown``` which doesn't return any errors.

The error from ```injectSigned``` is highly unlikely, cause it remarshalls the same config map
https://github.com/elastic/elastic-agent/pull/3203/files/b7ebf651fdec16a858feb2272f9dc6e4724f7256#diff-4f68d92e6db0175cc6917df7f966feed47cfd66753553f9a86c98ee7ed184b1fR269
that was decoded few lines above
https://github.com/elastic/elastic-agent/pull/3203/files/b7ebf651fdec16a858feb2272f9dc6e4724f7256#diff-4f68d92e6db0175cc6917df7f966feed47cfd66753553f9a86c98ee7ed184b1fR269

The intent here is to log and let the data flow through the rest of the states leading to Endpoint check-in in or the Teardown timeout, which will be both leading to the final ```onStop```.

The implementation without tamper protection immediately go to ```onStop``` that leads to the Endpoint uninstall. The addition of the tamper protection requires an additional step (roundtrip) with Endpoint to let it to unprotect itself before it can be stopped and uninstalled.

pkg/component/runtime/service.go
aleksmaus: With the introduction of the tamper protection there was a need for another internal step for the service runtime in order to handle the Teardown sequence correctly for the protected Endpoint.
For the case of unprotected Endpoint the Teardown immediately led to onStop and only one state change was sent back.

For the case of protected Endpoint the ```Teardown``` sequence sends another configuration update to Endpoint in order to unprotect the Endpoint before it can be stopped/uninstalled.
If service runtime sends two configuration updates ```onTeardown``` from the service runtime the second one gets blocked on attempting to write to the channel:
https://github.com/elastic/elastic-agent/blob/main/pkg/component/runtime/service.go#L524

The solution here was to make the state change consistent/same for both protected and unprotected cases:
* Agent receives the consistent state update onTeardown in both cases (it doesn't really care about additional step for the protected Endpoint teardown before it can be stopped).
* The internals of the "teardown" sequence are encapsulated within the service runtime implementation.

pierrehilbert:
From what I can observe, your test are failing:         	
                ```
Error Trace:	/home/ubuntu/agent/pkg/testing/tools/tools.go:125
        	            				/home/ubuntu/agent/pkg/testing/tools/tools.go:84
        	            				/home/ubuntu/agent/testing/integration/endpoint_security_test.go:351
        	            				/home/ubuntu/agent/testing/integration/endpoint_security_test.go:318
        	Test:       	TestInstallWithEndpointSecurityAndRemoveEndpointIntegration/protected
        	Messages:   	Elastic Agent status is not online
```

It's something @AndersonQ is working to fix as this is not happening only for you.
To check the error, use the artifact tab in Buildkite

**Comment by aleksmaus** [comment]
buildkite test this

**Comment by aleksmaus** [comment]
> From what I can observe, your test are failing:

Yeah, looks different the last run, the same test passed on amd64 and failed arm64 in buildkite, and there is no OS Arch specific code in that PR. 
Some flakiness with setup?

**Comment by pierrehilbert** [comment]
Yes we have some flakiness, @AndersonQ is working on a fix.

**Comment by aleksmaus** [comment]
> Yes we have some flakiness, @AndersonQ is working on a fix.

Awesome! Thanks! Looks like the test passed now.

File: pkg/component/runtime/service.go [hunk 5]
```diff
@@ -364,14 +363,14 @@ func (s *serviceRuntime) awaitCheckin(ctx context.Context, comm Communicator, ti
 	}
 }
 
-func (s *serviceRuntime) processNewComp(newComp component.Component, comm Communicator) {
+func (s *serviceRuntime) processNewComp(newComp component.Component, comm Communicator, sendObs bool) {
```
ycombinator, line 366: Could you add a comment here explaining when it makes sense to pass `sendObs` as `true` and when to pass it as `false`?

aleksmaus: The user of the additional flag here is a bit ugly. I'll split this function in two specific for each use and will add the comment.

cmacknz (approved):
Thanks for the explanations!

aleksmaus:
buildkite test this

aleksmaus:
buildkite test this

aleksmaus:
buildkite test this

PR #3206, opened 2023-08-08
invalidate fetcher cache if hash files are different

AndersonQ:
SNAPSHOT builds, either local or from the artifacts API, are expected to be different even though their names are the same. 

Therefore this change makes the fetcher cache to fetch the hash file and compare with the cached hash file only for SNAPSHOT versions. Non-SNAPSHOT versions are considered to be stable.

## What does this PR do?

This change makes the fetcher cache to fetch the hash file and compare with the cached hash file only for SNAPSHOT versions. Non-SNAPSHOT versions are considered to be stable.

## Why is it important?

SNAPSHOT builds, either local or from the artifacts API, are expected to be different even though their names are the same. 

## Checklist

## Related issues

- N/A

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/fetcher.go
+++ b/pkg/testing/fetcher.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -51,8 +52,12 @@ func GetPackageSuffix(operatingSystem string, architecture string) (string, erro
 type FetcherResult interface {
 	// Name is the name of the fetched result.
 	Name() string
+
 	// Fetch performs the actual fetch into the provided directory.
 	Fetch(ctx context.Context, l Logger, dir string) error
+
+	// FetchHash fetches the hash file for the artefact.
+	FetchHash(ctx context.Context, l Logger, dir string) error
 }
 
 // Fetcher provides a path for fetching the Elastic Agent compressed archive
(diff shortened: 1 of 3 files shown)


pkg/testing/fetcher.go
```diff
@@ -85,18 +90,59 @@ type fetcherCache struct {
...
-	err = res.Fetch(ctx, l, c.dir)
+
+	err := res.Fetch(ctx, l, c.dir)
 	if err != nil {
 		return "", err
 	}
 	return src, nil
 }
 
+func (c *fetcherCache) isSameArtefact(ctx context.Context, path string, res FetcherResult, l Logger) bool {
```
aleksmaus, line 109: nit: Artefact vs Artifact :-)
There are 500+ "Artifact" and only 1 "Artefact" in the agent repo. Maybe keep it consistent.

pkg/testing/fetcher_artifact.go
```diff
@@ -99,11 +99,21 @@ func (r *artifactResult) Name() string {
...
+	err = r.FetchHash(ctx, l, dir)
+	if err != nil {
+		return fmt.Errorf("failed to fetch hash file %s: %w", r.src, err)
+	}
+
+	return nil
+}
+
+// FetchHash fetches the hash file for the artefact.
+func (r *artifactResult) FetchHash(ctx context.Context, l Logger, dir string) error {
```
aleksmaus, line 115: nit: same as above but here it's inconsistent in the same function and it's comment

AndersonQ:
Closing it. The agent's repo folder should be deleted every time, therefore no need to invalidate the fetcher cache. I'll be addressed on another PR yet to come

AndersonQ:
fixed by https://github.com/elastic/elastic-agent/pull/3225

PR #3207, opened 2023-08-08
install fails if enroll fails and surface errors

AndersonQ:
For the agent to be actually enrolled it needs to restart after the enroll process is completed, so it'll pickup the new config and "connect" to fleet-server.

## What does this PR do?

## Why is it important?

For the agent to be actually enrolled it needs to restart after the enroll process is completed, so it'll pickup the new config and "connect" to fleet-server.

## Checklist

## Related issues

- N/A

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1693403216-Surface-errors-during-Agent's-enroll.yaml
+++ b/changelog/fragments/1693403216-Surface-errors-during-Agent's-enroll.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Surface errors during Agent's enroll process, failing if any happens.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: install/enroll
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/3207
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
--- a/internal/pkg/agent/cmd/enroll.go
+++ b/internal/pkg/agent/cmd/enroll.go
@@ -351,7 +351,7 @@ func enroll(streams *cli.IOStreams, cmd *cobra.Command) error {
 	//  Error: failed to fix permissions: chown /Library/Elastic/Agent/data/elastic-agent-c13f91/elastic-agent.app: operation not permitted
 	// This is because we are fixing permissions twice, once during installation and again during the enrollment step.
 	// When we are enrolling as part of installation on MacOS, skip the second attempt to fix permissions.
-	var fixPermissions bool = fromInstall
+	fixPermissions := fromInstall
 	if runtime.GOOS == "darwin" {
 		fixPermissions = false
 	}
--- a/internal/pkg/agent/cmd/enroll_cmd.go
+++ b/internal/pkg/agent/cmd/enroll_cmd.go
@@ -172,7 +172,7 @@ func newEnrollCmd(
 	)
 }
 
-// newEnrollCmdWithStore creates an new enrollment and accept a custom store.
+// newEnrollCmdWithStore creates a new enrollment and accept a custom store.
 func newEnrollCmdWithStore(
 	log *logger.Logger,
 	options *enrollCmdOption,
@@ -187,10 +187,11 @@ func newEnrollCmdWithStore(
 	}, nil
 }
 
-// Execute tries to enroll the agent into Fleet.
+// Execute enrolls the agent into Fleet.
 func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
 	var err error
 	defer c.stopAgent() // ensure its stopped no matter what
+
 	span, ctx := apm.StartSpan(ctx, "enroll", "app.internal")
 	defer func() {
 		apm.CaptureError(ctx, err).Send()
@@ -235,7 +236,7 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
 		// Ensure that the agent does not use a proxy configuration
 		// when connecting to the local fleet server.
 		// Note that when running fleet-server the enroll request will be sent to :8220,
-		// however when the agent is running afterwards requests will be sent to :8221
+		// however when the agent is running afterward requests will be sent to :8221
 		c.remoteConfig.Transport.Proxy.Disable = true
 	}
 
@@ -256,7 +257,7 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
 
 	err = c.enrollWithBackoff(ctx, persistentConfig)
 	if err != nil {
-		return errors.New(err, "fail to enroll")
+		return fmt.Errorf("fail to enroll: %w", err)
 	}
 
 	if c.options.FixPermissions {
@@ -267,17 +268,23 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
 	}
 
 	defer func() {
-		fmt.Fprintln(streams.Out, "Successfully enrolled the Elastic Agent.")
+		if err != nil {
+			fmt.Fprintf(streams.Err, "Something went wrong while enrolling the Elastic Agent: %v\n", err)
+		} else {
+			fmt.Fprintln(streams.Out, "Successfully enrolled the Elastic Agent.")
+		}
 	}()
 
 	if c.agentProc == nil {
-		if err := c.daemonReload(ctx); err != nil {
-			c.log.Infow("Elastic Agent might not be running; unable to trigger restart", "error", err)
-		} else {
-			c.log.Info("Successfully triggered restart on running Elastic Agent.")
+		if err = c.daemonReloadWithBackoff(ctx); err != nil {
+			c.log.Errorf("Elastic Agent might not be running; unable to trigger restart: %v", err)
+			return fmt.Errorf("could not reload agent daemon, unable to trigger restart: %w", err)
 		}
+
+		c.log.Info("Successfully triggered restart on running Elastic Agent.")
 		return nil
 	}
+
 	c.log.Info("Elastic Agent has been enrolled; start Elastic Agent")
 	return nil
 }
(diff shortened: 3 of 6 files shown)


changelog/fragments/1693403216-Surface-errors-during-Agent's-enroll.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
ycombinator, line 11: I think this should probably be a `bug-fix`? Seems like it's a bug for `elastic-agent install --url=... --enrollment-token-...` to succeed if enrollment fails. 
```suggestion
kind: bug-fix
```

internal/pkg/agent/cmd/enroll.go
```diff
@@ -351,7 +351,7 @@ func enroll(streams *cli.IOStreams, cmd *cobra.Command) error {
 	//  Error: failed to fix permissions: chown /Library/Elastic/Agent/data/elastic-agent-c13f91/elastic-agent.app: operation not permitted
 	// This is because we are fixing permissions twice, once during installation and again during the enrollment step.
 	// When we are enrolling as part of installation on MacOS, skip the second attempt to fix permissions.
-	var fixPermissions bool = fromInstall
+	var fixPermissions = fromInstall
```
ycombinator, line 354: Might as well go all the way here?
```suggestion
	fixPermissions := fromInstall
```

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -267,17 +268,23 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
...
 	}()
 
 	if c.agentProc == nil {
-		if err := c.daemonReload(ctx); err != nil {
-			c.log.Infow("Elastic Agent might not be running; unable to trigger restart", "error", err)
-		} else {
-			c.log.Info("Successfully triggered restart on running Elastic Agent.")
+		if err = c.daemonReloadWithBackoff(ctx); err != nil {
+			c.log.Errorf("Elastic Agent might not be running; unable to trigger restart: %v", err)
+			return fmt.Errorf("could not reload agent deamon, unable to trigger restart: %w", err)
```
ycombinator, line 281: Given the error printing that's now happening in the `defer func()` above, would it make sense to set `err` to the wrapped error here and return that instead?
```suggestion
			err = fmt.Errorf("could not reload agent deamon, unable to trigger restart: %w", err)
			return err
```

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -267,17 +268,23 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
 	}
 
 	defer func() {
-		fmt.Fprintln(streams.Out, "Successfully enrolled the Elastic Agent.")
+		if err != nil {
+			fmt.Fprintf(streams.Out, "Something went wrong while enrolling the Elastic Agent: %v\n", err)
```
ycombinator, line 272: Should this error be printed to `streams.Err`?
```suggestion
			fmt.Fprintf(streams.Err, "Something went wrong while enrolling the Elastic Agent: %v\n", err)
```

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -478,8 +485,20 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 	c.log.Infof("Starting enrollment to URL: %s", c.client.URI())
 	err := c.enroll(ctx, persistentConfig)
+	if err == nil {
+		return nil
+	}
+
+	const deadline = 10 * time.Minute
+	const frequency = 60 * time.Second
+
+	c.log.Infof("1st enrollment attempt failed, retrying for %s each %s enrolling to URL: %s",
```
ycombinator, line 495:
```suggestion
	c.log.Infof("1st enrollment attempt failed, retrying for %s, every %s, enrolling to URL: %s",
```

internal/pkg/agent/cmd/install.go
```diff
@@ -83,7 +83,7 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 
 	nonInteractive, _ := cmd.Flags().GetBool("non-interactive")
 	if nonInteractive {
-		fmt.Fprintf(streams.Out, "Installing in non-interactive mode.")
+		fmt.Fprintf(streams.Out, "Installing in non-interactive mode.\n")
```
ycombinator, line 86:
```suggestion
		fmt.Fprintln(streams.Out, "Installing in non-interactive mode.")
```

internal/pkg/agent/cmd/install.go
```diff
@@ -205,6 +205,8 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 				}
 			}()
 		}
+
+		fmt.Fprint(streams.Out, "Elastic Agent successfully installed, starting enrollment.\n")
```
ycombinator, line 209:
```suggestion
		fmt.Fprintln(streams.Out, "Elastic Agent successfully installed, starting enrollment.")
```

ycombinator (commented):
Nice improvement! Left a few minor comments.

internal/pkg/agent/cmd/enroll_cmd.go
AndersonQ: I did not see `streams.Err` being used here, so I just followed the pattern. But I agree with you

internal/pkg/agent/cmd/enroll_cmd.go
AndersonQ: they have the same effect: https://go.dev/play/p/9JnbEQWsxCG, err "will be set by returning the error"

ycombinator (approved):

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -478,8 +493,20 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
 
 	c.log.Infof("Starting enrollment to URL: %s", c.client.URI())
 	err := c.enroll(ctx, persistentConfig)
+	if err == nil {
+		return nil
+	}
+
+	const deadline = 10 * time.Minute
+	const frequency = 60 * time.Second
```
cmacknz, line 501: This feels too slow for me. We can retry faster than once every minute.

I also think waiting for 10 minutes is too long, we should fail earlier than this.

Right now the enroll command is going to appear like it is hung since there won't be any user feedback faster than every 1 minute.

cmacknz: Ah I see this is what it was before for the Fleet URL, I care more about the new backoff  for the restart command on the same host. Commented in the wrong place.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -443,24 +450,32 @@ func (c *enrollCmd) prepareFleetTLS() error {
 
 func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 	err := c.daemonReload(ctx)
+	if err != nil &&
```
cmacknz, line 453: Why do we need to special case this? Why can't we just enter the backoff loop below and exit it once it is finished?

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -443,24 +450,32 @@ func (c *enrollCmd) prepareFleetTLS() error {
...
-	for i := 5; i >= 0; i-- {
+	var i int
+	for ; i < 5; i++ {
 		backExp.Wait()
 		c.log.Info("Retrying to restart...")
 		err = c.daemonReload(ctx)
-		if err == nil {
+		if err == nil ||
+			errors.Is(err, context.DeadlineExceeded) ||
+			errors.Is(err, context.Canceled) {
```
cmacknz, line 472: You could replace this with https://pkg.go.dev/github.com/cenkalti/backoff/v4 which has a `WithContext` option to natively understands contexts. We already use this package for upgrades. Example:

https://github.com/cmacknz/elastic-agent/blob/261b94e5ec8372dda376ac80cdd212902b3a5aba/internal/pkg/agent/application/upgrade/step_download.go#L155-L185

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -267,17 +268,23 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
...
+			fmt.Fprintln(streams.Out, "Successfully enrolled the Elastic Agent.")
+		}
 	}()
 
 	if c.agentProc == nil {
-		if err := c.daemonReload(ctx); err != nil {
-			c.log.Infow("Elastic Agent might not be running; unable to trigger restart", "error", err)
-		} else {
-			c.log.Info("Successfully triggered restart on running Elastic Agent.")
+		if err = c.daemonReloadWithBackoff(ctx); err != nil {
```
cmacknz, line 279: I think the 1 minute retry with a 10 minute wait might be too slow for attempts to restart the agent on the same host. We can likely retry every second to start with a maximum wait of up to 2 minutes. If we are still retrying after 2 minutes changes are we aren't going to get the agent to restart.

We seem to need this logic in integration tests to get them to pass reliably and using a 1 minute minimum interval is going to make every test that hits this case take an additional minute to run.

The same is probably true for the Fleet retries but that one is tied to system scalability during mass enrollments so I am more hesitant to change it.

AndersonQ:
buildkite test this

cmacknz:
You will need to merge in the latest changes in main to get the tests to run properly if you haven't done that already. We had to change the region we provisioned the stack in.

cmacknz:
This is failing with the error it could be fixing:

```
    endpoint_security_test.go:189: Installing in non-interactive mode.
        Copying files................................................... DONE
        Installing service.... DONE
        Starting service... DONE
        Elastic Agent successfully installed, starting enrollment.
        Enrolling Elastic Agent with Fleet....{"log.level":"info","@timestamp":"2023-10-05T17:49:49.581Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":494},"message":"Starting enrollment to URL: https://2bb73e932bae48fba9bd4ac7037849c5.fleet.us-central1.gcp.foundit.no:443/","ecs.version":"1.6.0"}
        .............................................................................{"log.level":"info","@timestamp":"2023-10-05T17:50:10.808Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":468},"message":"Retrying to restart...","ecs.version":"1.6.0"}
        {"log.level":"error","@timestamp":"2023-10-05T17:50:10.826Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":280},"message":"Elastic Agent might not be running; unable to trigger restart: could not reload deamon after 1 retries: %!w()","ecs.version":"1.6.0"}
        Something went wrong while enrolling the Elastic Agent: could not reload deamon after 1 retries: %!w()
        Error: could not reload agent daemon, unable to trigger restart: could not reload deamon after 1 retries: %!w()
        For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.10/fleet-troubleshooting.html
         FAILED
        Stopping service... DONE
        Uninstalling...
           Stopping service... DONE
           Stopping upgrade watcher; none found... DONE
        .   Removing service.... DONE
           Removing install directory... DONE
           DONE
        Error: enroll command failed for unknown reason: exit status 1
        For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.10/fleet-troubleshooting.html
        
    endpoint_security_test.go:191: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/endpoint_security_test.go:191
        	            				/home/ubuntu/agent/testing/integration/endpoint_security_test.go:88
        	            	unable to enroll Elastic Agent: error running agent install command: exit status 1
        	Test:       	TestInstallAndCLIUninstallWithEndpointSecurity/protected
        	Messages:   	failed to install agent with policy
--- FAIL: TestInstallAndCLIUninstallWithEndpointSecurity/protected (55.99s)
```

We need more than 1 retry when we can't restart the agent to fix this.

cmacknz:
> Error: enroll command failed for unknown reason: exit status 1

Also that error message can be improved since we know exactly why enroll failed.

PR #3217, opened 2023-08-09
Fix character on changelog filename

fearful-symmetry:
## What does this PR do?
The filename `1688657261-Don't-trigger-IOC-alert-on-Windows-uninstall.yaml` has a single-quote in it, which breaks `go get`:
```
go get -u  github.com/elastic/elastic-agent@5973faf9d3af82785a4095a7ff6b6ecf83fbf6bc   
go: downloading github.com/elastic/elastic-agent v0.0.0-20230809182735-5973faf9d3af
go: github.com/elastic/elastic-agent@5973faf9d3af82785a4095a7ff6b6ecf83fbf6bc: create zip: changelog/fragments/1688657261-Don't-trigger-IOC-alert-on-Windows-uninstall.yaml: malformed file path "changelog/fragments/1688657261-Don't-trigger-IOC-alert-on-Windows-uninstall.yaml": invalid char '\''
```

As the `'` isn't allowed in zip path names, I think. Which is what go get is using under the hood.

## Why is it important?

This breaks `go get`.


cmacknz (approved):
We should probably upgrade the changelog tool to prevent creating files with single quotes in the name 🤦

PR #3085, opened 2023-07-14
[Integration Test] New Agent after upgrade fails to start

ycombinator:
## What does this PR do?

Relates https://github.com/elastic/elastic-agent/issues/2176

This PR adds an integration test that ensures an Agent upgrade is rolled back because the new Agent (after upgrade) fails to start.

The test creates a fake Agent package containing a fake Agent binary that immediately exits with a non-zero status code.  The test installs the locally-built Agent and then attempts to upgrade to the fake Agent package. The upgrade is expected to fail due to the new (fake) Agent binary failing to start successfully, and therefore the upgrade should be rolled back to the previous Agent version.

## Why is it important?

To ensure the upgrade/rollback behavior is working as expected.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] ~I have made corresponding changes to the documentation~
- [ ] ~I have made corresponding change to the default configuration files~
- [ ] ~I have added tests that prove my fix is effective or that my feature works~
- [ ] ~I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~
- [x] I have added an integration test or an E2E test

## How to test this PR locally

1. Build Agent locally.
   ```
   DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64 PACKAGES=tar.gz mage package
   ```

2. Run the test in this PR.
   ```
   SNAPSHOT=true mage integration:single TestStandaloneUpgradeFailsRestart
   ```

## Related issues
- Related to https://github.com/elastic/elastic-agent/issues/2176

PR #3220, opened 2023-08-10
Prevent `systemd` from killing the Upgrade Watcher process after the main Agent process has crashed

### LINKED ISSUE #3123: `systemd` prematurely kills Upgrade Watcher when upgraded Agent fails to start
This bug was noticed when writing an integration test where the upgraded Agent failed to start up ([PR](https://github.com/elastic/elastic-agent/pull/3085#issuecomment-1636276513)).

- Version: `main` / `8.10.0-SNAPSHOT`
- Operating System: Linux
- Steps to Reproduce:

1. To reproduce this bug, we need to upgrade to an Agent whose binary fails to start.  The integration test in this [PR](https://github.com/elastic/elastic-agent/pull/3085) builds such a failing, fake Agent binary, and packages it up. 

2. Try to upgrade to the Agent on Linux using `elastic-agent upgrade <version> --source-uri file:///path/to/fake/failing/agent/package.tgz`.

3. While the upgrade is in progress, monitor the status of the Elastic Agent service:
   ```
   $ watch -n1 systemctl status elastic-agent.service
   ```
   
4. For about a minute and a half, the above command will show that the fake Agent process is failing with a non-zero status code. It will also show the Upgrade Watcher process running.
   ```
   Every 1.0s: systemctl status elastic-agent.service                                                                  tdwijr: Fri Jul 14 18:51:38 2023

   ● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
        Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; vendor preset: enabled)
        Active: deactivating (stop-sigterm) (Result: exit-code) since Fri 2023-07-14 18:50:50 UTC; 48s ago
       Process: 97938 ExecStart=/usr/bin/elastic-agent (code=exited, status=101)
      Main PID: 97938 (code=exited, status=101)
         Tasks: 6 (limit: 4637)
        Memory: 105.8M
           CPU: 3.386s
        CGroup: /system.slice/elastic-agent.service
                └─98040 /opt/Elastic/Agent/data/elastic-agent-903287/elastic-agent watch --path.config /opt/Elastic/Agent --path.home /opt/Elastic/Agent
   ```

   Then, after about a minute and a half,  systemd will attempt to restart the fake Agent process. You can tell this by seeing the `Main PID` change in the status command's output.  However, as a result of this attempted restart, the Upgrade Watcher process also gets killed.  You can see it disappearing from the output.

   ```
   Every 1.0s: systemctl status elastic-agent.service                                                                  tdwijr: Fri Jul 14 18:52:32 2023

   ● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
        Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; vendor preset: enabled)
        Active: activating (auto-restart) (Result: exit-code) since Fri 2023-07-14 18:52:20 UTC; 12s ago
       Process: 97938 ExecStart=/usr/bin/elastic-agent (code=exited, status=101)
      Main PID: 97938 (code=exited, status=101)
           CPU: 3.396s

   Jul 14 18:52:20 tdwijr systemd[1]: elastic-agent.service: Killing process 98043 (elastic-agent) with signal SIGKILL.
   Jul 14 18:52:20 tdwijr systemd[1]: elastic-agent.service: Killing process 98044 (elastic-agent) with signal SIGKILL.
   Jul 14 18:52:20 tdwijr systemd[1]: elastic-agent.service: Killing process 98045 (elastic-agent) with signal SIGKILL.
   Jul 14 18:52:20 tdwijr systemd[1]: elastic-agent.service: Failed with result 'exit-code'.
   Jul 14 18:52:20 tdwijr systemd[1]: elastic-agent.service: Consumed 3.396s CPU time.
   ```

5. As a consequence of this premature killing of the Upgrade Watcher process, the upgraded Elastic Agent will keep failing to start but there will be no Upgrade Watcher around to monitor these failures and perform a rollback to the Agent that was running prior to the upgrade.

ycombinator:
## What does this PR do?

This PR adds the `KillMode=process` setting to Elastic Agent's `systemd` unit configuration file, installed at `/etc/systemd/system/elastic-agent.service`. 

## Why is it important?

The [`KillMode`](https://www.freedesktop.org/software/systemd/man/systemd.kill.html#KillMode=)  setting governs which processes under the service will be killed by `systemd`. The default value of the setting is `control-group`.  With this value, all processes under the service are killed. However, with the `process` value, only the main process under the service is killed while the others are left running.

In the context of Elastic Agent, the main process it the Agent process.  During an ongoing upgrade, there is a second process in the same cgroup as the main process: the Upgrade Watcher process. 

In the scenario when the main process crashes or is not reporting a healthy status repeatedly, we need the Upgrade Watcher process to keep running so it can detect this repeated crashing or unhealthy status, and then initiate a rollback. 

By setting `KillMode=process` in Elastic Agent's `systemd` unit configuration file, we ensure that when the main process is repeatedly crashing, `systemd` keeps the Upgrade Watcher process running so it can fulfill it's destiny.

## Checklist

### Test A

For this test, we will upgrade from a pre-built version of Agent, say `8.9.0` to the build of Agent with this PR (`8.10.0`) and then to another build of Agent that deliberately crashes (say, `8.11.0`).

The upgrade from the first to the second build of Agent will test that the `/etc/systemd/system/elastic-agent.service` file starts off without the `KillMode=process` setting in it and that the setting is added after the new (post-upgrade) Agent starts running.

The upgrade from the second to the third build of Agent will test that the `KillMode=process` setting is behaving as expected, by ensuring that the Upgrade Watcher process keeps running even after the new (post-upgrade) Agent process crashes.

1. Start with a pre-built version of Elastic Agent < `8.10.0`
2. Unpack and install this Elastic Agent on Linux.
3. Check the contents of the `/etc/systemd/system/elastic-agent.service` file and ensure that the `KillMode=process` setting is absent.
   ```
   grep KillMode /etc/systemd/system/elastic-agent.service
   ```
4. Check out this PR's branch and build Elastic Agent for Linux.
   ```
5. Upgrade to the newly-built Agent.
   ```
   sudo elastic-agent upgrade 8.10.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
   ```
6. Check the contents of the `/etc/systemd/system/elastic-agent.service` file and ensure that the `KillMode=process` setting is now present.  Don't worry about any whitespace around the `=` sign (`systemd` [ignores](https://www.freedesktop.org/software/systemd/man/systemd.syntax.html#) it).
   ```
   grep KillMode /etc/systemd/system/elastic-agent.service
   ```
7. Ensure that the `systemd` unit configuration files have been reloaded.  For this, run `systemctl status elastic-agent.service >/dev/null` and ensure that the following warning message is NOT shown:
   ```
   Warning: The unit file, source configuration file or drop-ins of elastic-agent.service changed on disk. Run 'systemctl daemon-reload' to reload units.
   ```

9. Ensure that the upgrade succeeds before moving on.  For this, wait until the Upgrade Watcher process has finished running and ensure that the running Elastic Agent's version is `8.10.0`.
   ```
   systemctl status elastic-agent.service | grep 'elastic-agent watch'
   sudo elastic-agent version
   ```
1. Bump up the minor version in `version/version.go` (to, say, `8.11.0`), and also return an error at the very start of https://github.com/elastic/elastic-agent/blob/af31709ed6d1d249235e2745eb3cc314d3a7f464/internal/pkg/agent/cmd/run.go#L65 to ensure that the built Agent will crash immediately upon start up.
2. Make sure to make a new commit with these changes so the build ID (SHA) in the Agent build is different. Otherwise the upgrade won't work.
4. Build Elastic Agent for Linux.
   ```
   ```
5. Upgrade to the newly-built Agent.
   ```
   sudo elastic-agent upgrade 8.11.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
   ```
7. Monitor the output of `systemctl status elastic-agent.service`.  Watch the Upgrade Watcher process being spawned, then the new (post-upgrade) Agent process crashing, and ensure that the Upgrade Watcher process keeps running for at least 2 minutes.
   ```
   watch -n1 systemctl status elastic-agent.service
   ```

### Test B

For this test, we will upgrade from a pre-built version of Agent, say `8.9.0` to a build of Agent that has the changes in this PR but also crashes upon start up (say, `8.10.0`).

This will test that that the `/etc/systemd/system/elastic-agent.service` file starts off without the `KillMode=process` setting in it, that the setting is added after the new (post-upgrade) Agent starts running, and that the setting takes effect immediately so as to keep the Upgrade Watcher process running after the Agent process has crashed.

1. Start with a pre-built version of Elastic Agent < `8.10.0`
2. Unpack and install this Elastic Agent on Linux.
3. Check the contents of the `/etc/systemd/system/elastic-agent.service` file and ensure that the `KillMode=process` setting is absent.
   ```
   grep KillMode /etc/systemd/system/elastic-agent.service
   ```
4. Check out this PR's branch.
5. Modify the code to return an error right after the changes in this PR have a chance to take effect, to ensure that the built Agent will crash: https://github.com/elastic/elastic-agent/blob/af31709ed6d1d249235e2745eb3cc314d3a7f464/internal/pkg/agent/cmd/run.go#L108-L111

2. Make sure to make a new commit with these changes so the build ID (SHA) in the Agent build is different. Otherwise the upgrade won't work.
4. Build Elastic Agent for Linux.
   ```
7. Upgrade to the newly-built Agent.
   ```
   sudo elastic-agent upgrade 8.10.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
   ```
8. Check the contents of the `/etc/systemd/system/elastic-agent.service` file and ensure that the `KillMode=process` setting is now present.  Don't worry about any whitespace around the `=` sign (`systemd` [ignores](https://www.freedesktop.org/software/systemd/man/systemd.syntax.html#) it).
   ```
   grep KillMode /etc/systemd/system/elastic-agent.service
   ```
9. Ensure that the `systemd` unit configuration files have been reloaded.  For this, run `systemctl status elastic-agent.service >/dev/null` and ensure that the following warning message is NOT shown:
   ```
   Warning: The unit file, source configuration file or drop-ins of elastic-agent.service changed on disk. Run 'systemctl daemon-reload' to reload units.
   ```
7. Monitor the output of `systemctl status elastic-agent.service`.  Watch the Upgrade Watcher process being spawned, then the new (post-upgrade) Agent process crashing, and ensure that the Upgrade Watcher process keeps running for at least 2 minutes.
   ```
   watch -n1 systemctl status elastic-agent.service
   ```

## Related issues

- Fixes #3123

### DIFF
--- a/internal/pkg/agent/application/upgrade/service_update_linux.go
+++ b/internal/pkg/agent/application/upgrade/service_update_linux.go
@@ -17,6 +17,9 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 )
 
+// EnsureServiceConfigUpToDate modifies, if necessary, Elastic Agent's service
+// configuration file to its latest definition. This change is NOT rolled back
+// if the Agent is rolled back to a previous version.
 func EnsureServiceConfigUpToDate() error {
 	switch service.ChosenSystem().String() {
 	case "linux-systemd":
--- a/internal/pkg/agent/application/upgrade/service_update_linux.go
+++ b/internal/pkg/agent/application/upgrade/service_update_linux.go
@@ -24,8 +24,14 @@ func EnsureServiceConfigUpToDate() error {
 	switch service.ChosenSystem().String() {
 	case "linux-systemd":
 		unitFilePath := "/etc/systemd/system/" + paths.ServiceName + ".service"
-		if err := ensureSystemdServiceConfigUpToDate(unitFilePath); err != nil {
+		updated, err := ensureSystemdServiceConfigUpToDate(unitFilePath)
+		if err != nil {
 			return err
+		}
+
+		if !updated {
+			// Nothing more to do!
+			return nil
 		}
 
 		// Reload systemd unit configuration files
@@ -39,26 +45,29 @@ func EnsureServiceConfigUpToDate() error {
 	return nil
 }
 
-func ensureSystemdServiceConfigUpToDate(unitFilePath string) error {
+// ensureSystemdServiceConfigUpToDate modifies, if necessary, Elastic Agent's systemd
+// unit configuration file to its latest definition. If the the file is modified, this
+// function returns true; otherwise, it returns false.
+func ensureSystemdServiceConfigUpToDate(unitFilePath string) (bool, error) {
 	// It is safe to use an INI file parser/writer for systemd unit files.
 	// See https://www.freedesktop.org/software/systemd/man/systemd.syntax.html
 	cfg, err := ini.Load(unitFilePath)
 	if err != nil {
-		return fmt.Errorf("error opening systemd unit file [%s]: %w", unitFilePath, err)
+		return false, fmt.Errorf("error opening systemd unit file [%s]: %w", unitFilePath, err)
 	}
 
 	// Check if KillMode= is already present
 	if cfg.Section("Service").HasKey("KillMode") {
 		// Nothing more to do
-		return nil
+		return false, nil
 	}
 
 	// If KillMode= is not present, add it and set it to "process"
 	// See https://github.com/elastic/elastic-agent/pull/3220
 	cfg.Section("Service").Key("KillMode").SetValue("process")
 	if err := cfg.SaveTo(unitFilePath); err != nil {
-		return fmt.Errorf("error writing updated systemd unit file [%s]: %w", unitFilePath, err)
+		return false, fmt.Errorf("error writing updated systemd unit file [%s]: %w", unitFilePath, err)
 	}
 
-	return nil
+	return true, nil
 }
--- a/internal/pkg/agent/application/upgrade/service_update_linux.go
+++ b/internal/pkg/agent/application/upgrade/service_update_linux.go
@@ -4,7 +4,7 @@
 
 //go:build linux
 
-package cmd
+package upgrade
 
 import (
 	"fmt"
@@ -17,7 +17,7 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 )
 
-func ensureServiceConfigUpToDate() error {
+func EnsureServiceConfigUpToDate() error {
 	switch service.ChosenSystem().String() {
 	case "linux-systemd":
 		unitFilePath := "/etc/systemd/system/" + paths.ServiceName + ".service"
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -63,9 +63,6 @@ func newRunCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Command {
 		Short: "Start the Elastic Agent",
 		Long:  "This command starts the Elastic Agent.",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			// FIXME: for testing only!
-			<-time.After(30 * time.Minute)
-
 			// done very early so the encrypted store is never used
 			disableEncryptedStore, _ := cmd.Flags().GetBool("disable-encrypted-store")
 			if disableEncryptedStore {
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -109,9 +109,6 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
 		return fmt.Errorf("error checking for and handling upgrade: %w", err)
 	}
 
-	// FIXME: Remove the following line; it is just for testing!
-	return errors.New("deliberately crashing agent")
-
 	locker := filelock.NewAppLocker(paths.Data(), paths.AgentLockFileName)
 	if err := locker.TryLock(); err != nil {
 		return err
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -19,6 +19,8 @@ import (
 	apmtransport "go.elastic.co/apm/transport"
 	"gopkg.in/yaml.v2"
 
+	"github.com/spf13/cobra"
+
 	"github.com/elastic/elastic-agent-libs/api"
 	"github.com/elastic/elastic-agent-libs/logp"
 	monitoringLib "github.com/elastic/elastic-agent-libs/monitoring"
@@ -46,7 +48,6 @@ import (
 	"github.com/elastic/elastic-agent/pkg/control/v2/server"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 	"github.com/elastic/elastic-agent/version"
-	"github.com/spf13/cobra"
 )
 
 const (
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -109,6 +109,9 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
 		return fmt.Errorf("error checking for and handling upgrade: %w", err)
 	}
 
+	// FIXME: Remove the following line; it is just for testing!
+	return errors.New("deliberately crashing agent")
+
 	locker := filelock.NewAppLocker(paths.Data(), paths.AgentLockFileName)
 	if err := locker.TryLock(); err != nil {
 		return err
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -63,6 +63,9 @@ func newRunCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Command {
 		Short: "Start the Elastic Agent",
 		Long:  "This command starts the Elastic Agent.",
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			// FIXME: for testing only!
+			<-time.After(30 * time.Minute)
+
 			// done very early so the encrypted store is never used
 			disableEncryptedStore, _ := cmd.Flags().GetBool("disable-encrypted-store")
 			if disableEncryptedStore {
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -581,6 +581,12 @@ func handleUpgrade() error {
 	if err := ensureInstallMarkerPresent(); err != nil {
 		return err
 	}
+
+	if err := ensureServiceConfigUpToDate(); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func ensureInstallMarkerPresent() error {
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -583,7 +583,7 @@ func handleUpgrade() error {
 		return err
 	}
 
-	if err := ensureServiceConfigUpToDate(); err != nil {
+	if err := upgrade.EnsureServiceConfigUpToDate(); err != nil {
 		return err
 	}
 
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -15,7 +15,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/spf13/cobra"
 	"go.elastic.co/apm"
 	apmtransport "go.elastic.co/apm/transport"
 	"gopkg.in/yaml.v2"
@@ -24,7 +23,6 @@ import (
 	"github.com/elastic/elastic-agent-libs/logp"
 	monitoringLib "github.com/elastic/elastic-agent-libs/monitoring"
 	"github.com/elastic/elastic-agent-libs/service"
-
 	"github.com/elastic/elastic-agent-system-metrics/report"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator"
@@ -48,6 +46,7 @@ import (
 	"github.com/elastic/elastic-agent/pkg/control/v2/server"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 	"github.com/elastic/elastic-agent/version"
+	"github.com/spf13/cobra"
 )
 
 const (
@@ -579,6 +578,12 @@ func handleUpgrade() error {
 		return nil
 	}
 
+	if err := ensureInstallMarkerPresent(); err != nil {
+		return err
+	}
+}
+
+func ensureInstallMarkerPresent() error {
 	// In v8.8.0, we introduced a new installation marker file to indicate that
 	// an Agent was running as installed. When an installed Agent that's older
 	// than v8.8.0 is upgraded, this installation marker file is not present.
--- a/internal/pkg/agent/cmd/svc_update.go
+++ b/internal/pkg/agent/cmd/svc_update.go
@@ -1,0 +1,12 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build !linux
+
+package cmd
+
+func ensureServiceConfigUpToDate() error {
+	// Noop on non-Linux systems
+	return nil
+}
--- a/internal/pkg/agent/cmd/svc_update_linux.go
+++ b/internal/pkg/agent/cmd/svc_update_linux.go
@@ -1,0 +1,48 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build linux
+
+package cmd
+
+import (
+	"fmt"
+
+	"gopkg.in/ini.v1"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/kardianos/service"
+)
+
+func ensureServiceConfigUpToDate() error {
+	switch service.ChosenSystem().String() {
+	case "linux-systemd":
+		unitFilePath := "/etc/systemd/system/" + paths.ServiceName + ".service"
+		return ensureSystemdServiceConfigUpToDate(unitFilePath)
+	}
+
+	return nil
+}
+
+func ensureSystemdServiceConfigUpToDate(unitFilePath string) error {
+	cfg, err := ini.Load(unitFilePath)
+	if err != nil {
+		return fmt.Errorf("error opening systemd unit file [%s]: %w", unitFilePath, err)
+	}
+
+	// Check if KillMode= is already present
+	if cfg.Section("Service").HasKey("KillMode") {
+		// Nothing more to do
+		return nil
+	}
+
+	// If KillMode= is not present, add it and set it to "process"
+	// See https://github.com/elastic/elastic-agent/pull/3220
+	cfg.Section("Service").Key("KillMode").SetValue("process")
+	if err := cfg.SaveTo(unitFilePath); err != nil {
+		return fmt.Errorf("error writing updated systemd unit file [%s]: %w", unitFilePath, err)
+	}
+
+	return nil
+}
--- a/internal/pkg/agent/cmd/svc_update_linux.go
+++ b/internal/pkg/agent/cmd/svc_update_linux.go
@@ -8,6 +8,7 @@ package cmd
 
 import (
 	"fmt"
+	"os/exec"
 
 	"gopkg.in/ini.v1"
 
@@ -45,6 +46,12 @@ func ensureSystemdServiceConfigUpToDate(unitFilePath string) error {
 	cfg.Section("Service").Key("KillMode").SetValue("process")
 	if err := cfg.SaveTo(unitFilePath); err != nil {
 		return fmt.Errorf("error writing updated systemd unit file [%s]: %w", unitFilePath, err)
+	}
+
+	// Reload systemd unit configuration files
+	cmd := exec.Command("systemctl", "daemon-reload")
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error reloading systemd unit configuration files: %w", err)
 	}
 
 	return nil
--- a/internal/pkg/agent/cmd/svc_update_linux.go
+++ b/internal/pkg/agent/cmd/svc_update_linux.go
@@ -11,8 +11,9 @@ import (
 
 	"gopkg.in/ini.v1"
 
-	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 	"github.com/kardianos/service"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 )
 
 func ensureServiceConfigUpToDate() error {
--- a/internal/pkg/agent/cmd/svc_update_linux.go
+++ b/internal/pkg/agent/cmd/svc_update_linux.go
@@ -27,6 +27,8 @@ func ensureServiceConfigUpToDate() error {
 }
 
 func ensureSystemdServiceConfigUpToDate(unitFilePath string) error {
+	// It is safe to use an INI file parser/writer for systemd unit files.
+	// See https://www.freedesktop.org/software/systemd/man/systemd.syntax.html
 	cfg, err := ini.Load(unitFilePath)
 	if err != nil {
 		return fmt.Errorf("error opening systemd unit file [%s]: %w", unitFilePath, err)
--- a/internal/pkg/agent/cmd/svc_update_linux.go
+++ b/internal/pkg/agent/cmd/svc_update_linux.go
@@ -21,7 +21,16 @@ func ensureServiceConfigUpToDate() error {
 	switch service.ChosenSystem().String() {
 	case "linux-systemd":
 		unitFilePath := "/etc/systemd/system/" + paths.ServiceName + ".service"
-		return ensureSystemdServiceConfigUpToDate(unitFilePath)
+		if err := ensureSystemdServiceConfigUpToDate(unitFilePath); err != nil {
+			return err
+		}
+
+		// Reload systemd unit configuration files
+		cmd := exec.Command("systemctl", "daemon-reload")
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("error reloading systemd unit configuration files: %w", err)
+		}
+
 	}
 
 	return nil
@@ -46,12 +55,6 @@ func ensureSystemdServiceConfigUpToDate(unitFilePath string) error {
 	cfg.Section("Service").Key("KillMode").SetValue("process")
 	if err := cfg.SaveTo(unitFilePath); err != nil {
 		return fmt.Errorf("error writing updated systemd unit file [%s]: %w", unitFilePath, err)
-	}
-
-	// Reload systemd unit configuration files
-	cmd := exec.Command("systemctl", "daemon-reload")
-	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("error reloading systemd unit configuration files: %w", err)
 	}
 
 	return nil
--- a/internal/pkg/agent/install/svc.go
+++ b/internal/pkg/agent/install/svc.go
@@ -139,7 +139,7 @@ StandardError=file:/var/log/{{.Name}}.err
 {{if gt .LimitNOFILE -1 }}LimitNOFILE={{.LimitNOFILE}}{{end}}
 {{if .Restart}}Restart={{.Restart}}{{end}}
 {{if .SuccessExitStatus}}SuccessExitStatus={{.SuccessExitStatus}}{{end}}
-{{if .TimeoutStopSec}}TimeoutStopSec={{.TimeoutStopSec}}{{end}}
+{{if .Config.Option.TimeoutStopSec}}TimeoutStopSec={{.Config.Option.TimeoutStopSec}}{{end}}
 RestartSec=120
 EnvironmentFile=-/etc/sysconfig/{{.Name}}
 
--- a/internal/pkg/agent/install/svc.go
+++ b/internal/pkg/agent/install/svc.go
@@ -55,9 +55,9 @@ func newService(topPath string) (service.Service, error) {
 	}
 
 	if runtime.GOOS == "linux" {
-		// The github.com/kardianos/service library doesn't support TimeoutStopSec in their prebuilt template.
+		// The github.com/kardianos/service library doesn't support KillMode in their prebuilt template.
 		// This option allows to pass our own template for the systemd unit configuration, which is a copy
-		// of the prebuilt template with added TimeoutStopSec option
+		// of the prebuilt template with added KillMode option
 		cfg.Option["SystemdScript"] = linuxSystemdScript
 		cfg.Option["KillMode"] = "process"
 	}
@@ -116,7 +116,7 @@ const darwinLaunchdConfig = `<?xml version='1.0' encoding='UTF-8'?>
 `
 
 // A copy of the systemd config template from github.com/kardianos/service
-// with added .Config.Option.TimeoutStopSec option
+// with added .Config.Option.KillMode option
 const linuxSystemdScript = `[Unit]
 Description={{.Description}}
 ConditionFileIsExecutable={{.Path|cmdEscape}}
--- a/internal/pkg/agent/install/svc.go
+++ b/internal/pkg/agent/install/svc.go
@@ -59,7 +59,7 @@ func newService(topPath string) (service.Service, error) {
 		// This option allows to pass our own template for the systemd unit configuration, which is a copy
 		// of the prebuilt template with added TimeoutStopSec option
 		cfg.Option["SystemdScript"] = linuxSystemdScript
-		cfg.Option["TimeoutStopSec"] = 610
+		cfg.Option["KillMode"] = "process"
 	}
 
 	if runtime.GOOS == "darwin" {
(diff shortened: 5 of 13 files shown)


pchila (approved):
LGTM

ycombinator:
buildkite test this

blakerouse (changes requested):
I would like to talk about this change a little bit more.

I read all of the details from the issue that you provided on changing this value. I do worry that this change of the timeout does have a side-effect of `systemctl stop elastic-agent`. I believe that if stopping the Elastic Agent using the above command hangs that it would block for the full 10 minutes. I don't know if that is the behavior we want to provide in that situation.

I also worry about this having an effect on uninstall. We stop the service, if the watcher is running and uninstall is performed will it block for 10mins waiting for the watcher to die before being able to continue.

I actually thinking changing the `KillMode=process` is the proper solution. Being that Elastic Agent is a supervisor and it spawns all of it subprocess with the proper `Pkill` in the situations that require it. It would also equal the same behavior that we have on macOS and WIndows.

ycombinator:
Thanks for taking the time to read through all the investigation details in the issue and bringing up your concerns here. Let me do some more testing for the scenarios you outlined as well as with the alternative idea of specifying `KillMode=process` instead of the `TimeoutStopSec` setting and post my findings here.

ycombinator:
@blakerouse Thanks for taking the time to read through all the investigation details in the issue and bringing up your concerns here. Let me do some more testing for the scenarios you outlined as well as with the alternative idea of specifying `KillMode=process` instead of the `TimeoutStopSec` setting and post my findings here.

ycombinator:
I tested the three scenarios @blakerouse brought up in https://github.com/elastic/elastic-agent/pull/3220#pullrequestreview-1572182008.  Observations and conclusions from each of the three tests are below but tl;dr is that it seems better to go with the `KillMode=process` setting than the `TimeoutStopSec` setting.  I'll update this PR accordingly.

> I do worry that this change of the timeout does have a side-effect of `systemctl stop elastic-agent`. I believe that if stopping the Elastic Agent using the above command hangs that it would block for the full 10 minutes. I don't know if that is the behavior we want to provide in that situation.

I built a version of Agent (on top of the changes in this PR) that will block for 30 minutes (`<-time.After(30 * time.Minute)`) very early in the `elastic-agent run` code path.  

I installed this Agent made sure it started running under the service.

Then I immediately called `systemctl stop elastic-agent` and then looked at the output of `systemctl status elastic-agent`. 

It appears the service is stopped and the Agent process is terminated.

```
$ sudo systemctl status elastic-agent.service
Unit elastic-agent.service could not be found.
Elastic Agent will be installed at /opt/Elastic/Agent and will run as a service. Do you want to continue? [Y/n]:y
$ sudo systemctl status elastic-agent.service
● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2023-08-10 19:50:35 UTC; 26s ago
   Main PID: 478576 (elastic-agent)
      Tasks: 5 (limit: 4637)
     Memory: 10.9M
        CPU: 60ms
     CGroup: /system.slice/elastic-agent.service
             └─478576 /opt/Elastic/Agent/elastic-agent

Aug 10 19:50:35 shaunak-ubuntu-22-arm systemd[1]: Started Elastic Agent is a unified agent to observe, monitor and protect your system..
$  sudo systemctl stop elastic-agent.service
$ sudo systemctl status elastic-agent.service
○ elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Thu 2023-08-10 19:51:11 UTC; 5s ago
    Process: 478576 ExecStart=/usr/bin/elastic-agent (code=killed, signal=TERM)
   Main PID: 478576 (code=killed, signal=TERM)
        CPU: 65ms

Aug 10 19:50:35 shaunak-ubuntu-22-arm systemd[1]: Started Elastic Agent is a unified agent to observe, monitor and protect your system..
Aug 10 19:51:11 shaunak-ubuntu-22-arm systemd[1]: Stopping Elastic Agent is a unified agent to observe, monitor and protect your system....
Aug 10 19:51:11 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Deactivated successfully.
Aug 10 19:51:11 shaunak-ubuntu-22-arm systemd[1]: Stopped Elastic Agent is a unified agent to observe, monitor and protect your system..
```

**Conclusion**: it would appear that setting `TimeoutStopSec=610` does not have an adverse impact on the behavior of `systemctl stop elastic-agent`.

---

> I also worry about this having an effect on uninstall. We stop the service, if the watcher is running and uninstall is performed will it block for 10mins waiting for the watcher to die before being able to continue.

I tested this scenario by starting with a build of Agent from this PR (so it includes the `TimeoutStopSec=610` setting) and upgrading to a build of Agent that crashes immediately upon starting up.

```
$ grep TimeoutStopSec /etc/systemd/system/elastic-agent.service
TimeoutStopSec=610
$ sudo elastic-agent upgrade 8.13.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
Upgrade triggered to version 8.13.0, Elastic Agent is currently restarting
$ 
```

As expected, due to the timeout change in this PR, the Upgrade Watcher process kept running after the upgraded Agent process crashed.

```
$ sudo systemctl status elastic-agent.service
● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; vendor preset: enabled)
     Active: deactivating (stop-sigterm) (Result: exit-code) since Thu 2023-08-10 20:29:06 UTC; 3min 10s ago
    Process: 492383 ExecStart=/usr/bin/elastic-agent (code=exited, status=1/FAILURE)
   Main PID: 492383 (code=exited, status=1/FAILURE)
      Tasks: 6 (limit: 4637)
     Memory: 103.8M
        CPU: 6.234s
     CGroup: /system.slice/elastic-agent.service
             └─493106 /opt/Elastic/Agent/elastic-agent watch --path.config /opt/Elastic/Agent --path.home /opt/Elastic/Agent

Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --environment environmentVar   set environment being ran in (default default)
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:   -h, --help                         help for elastic-agent
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.config string           Config path is the directory Agent looks for its >
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.downloads string        Downloads path contains binaries Agent downloads >
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.home string             Agent root path (default "/opt/Elastic/Agent")
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.install string          Install path contains binaries Agent extracts
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.logs string             Logs path contains Agent log output (default "/op>
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:   -v, --v                            Log at INFO level
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]: Use "elastic-agent [command] --help" for more information about a command.
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]: prematurely crashing agent
```

At this point, I uninstalled the Agent. The command returned almost immediately, as usual, without blocking for too long.

```
$ sudo elastic-agent uninstall
Elastic Agent is installed but currently broken: service is not installed
Continuing will uninstall the broken Elastic Agent at /opt/Elastic/Agent. Do you want to continue? [Y/n]:y
```

Then I checked the service status.

```
$ sudo systemctl status elastic-agent.service
Warning: The unit file, source configuration file or drop-ins of elastic-agent.service changed on disk. Run 'systemctl daemon-reload' to reload uni>
● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; bad; vendor preset: enabled)
     Active: deactivating (stop-sigterm) (Result: exit-code) since Thu 2023-08-10 20:29:06 UTC; 3min 31s ago
   Main PID: 492383 (code=exited, status=1/FAILURE)
      Tasks: 6 (limit: 4637)
     Memory: 103.4M
        CPU: 6.239s
     CGroup: /system.slice/elastic-agent.service
             └─493106 /opt/Elastic/Agent/elastic-agent watch --path.config /opt/Elastic/Agent --path.home /opt/Elastic/Agent

Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --environment environmentVar   set environment being ran in (default default)
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:   -h, --help                         help for elastic-agent
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.config string           Config path is the directory Agent looks for its >
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.downloads string        Downloads path contains binaries Agent downloads >
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.home string             Agent root path (default "/opt/Elastic/Agent")
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.install string          Install path contains binaries Agent extracts
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:       --path.logs string             Logs path contains Agent log output (default "/op>
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]:   -v, --v                            Log at INFO level
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]: Use "elastic-agent [command] --help" for more information about a command.
Aug 10 20:29:06 shaunak-ubuntu-22-arm elastic-agent[492383]: prematurely crashing agent
```

As the output above shows, the service was still running and, notably, the Upgrade Watcher process was still running. 

Interestingly, however, about 2m16s later, the Upgrade Watcher process stopped running, much sooner than the 660s timeout.  Given the odd timing (2m16s), most likely the Upgrade Watcher itself failed somehow.  Unfortunately since the `elastic-agent uninstall` had wiped out everything under `/opt/Elastic/Agent`, I couldn't inspect the Upgrade Watcher logs.  However, this does indicate to me, that had the Upgrade Watcher process not naturally exited it would likely have been kept running by `systemd` for 660s.  

Here we can see that the service is still running but there are no processes under it.

```
$ sudo systemctl status elastic-agent.service
Warning: The unit file, source configuration file or drop-ins of elastic-agent.service changed on disk. Run 'systemctl daemon-reload' to reload uni>
● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; bad; vendor preset: enabled)
     Active: activating (auto-restart) (Result: exit-code) since Thu 2023-08-10 20:35:21 UTC; 1s ago
   Main PID: 492383 (code=exited, status=1/FAILURE)
        CPU: 6.266s

Aug 10 20:35:21 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Failed with result 'exit-code'.
Aug 10 20:35:21 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Consumed 6.266s CPU time.
```

Exactly 2 minutes after the Upgrade Watcher process stopped running, the service itself got uninstalled.

```
$ sudo systemctl status elastic-agent.service
Unit elastic-agent.service could not be found.
```

**Conclusion**: it would appear that setting `TimeoutStopSec=610` does have some impact on the uninstall process but it does not block the uninstall process.

---

> I actually thinking changing the `KillMode=process` is the proper solution. Being that Elastic Agent is a supervisor and it spawns all of it subprocess with the proper `Pkill` in the situations that require it. It would also equal the same behavior that we have on macOS and WIndows.

I tested this scenario by starting with a build of Agent that set `KillMode=process` instead of `TimeoutStopSec=610` in the unit configuration file,  and upgrading to a build of Agent that crashes immediately upon starting up.

```
$ grep TimeoutStopSec /etc/systemd/system/elastic-agent.service
$ grep KillMode /etc/systemd/system/elastic-agent.service
KillMode=process
$ sudo elastic-agent upgrade 8.13.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
Upgrade triggered to version 8.13.0, Elastic Agent is currently restarting
```

As expected, due to `KillMode=process`, the Upgrade Watcher process kept running after the upgraded Agent process crashed.

```
$ sudo systemctl status elastic-agent.service
● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; vendor preset: enabled)
     Active: activating (auto-restart) (Result: exit-code) since Thu 2023-08-10 20:59:03 UTC; 1min 57s ago
    Process: 503266 ExecStart=/usr/bin/elastic-agent (code=exited, status=1/FAILURE)
   Main PID: 503266 (code=exited, status=1/FAILURE)
      Tasks: 6 (limit: 4637)
     Memory: 104.3M
        CPU: 6.068s
     CGroup: /system.slice/elastic-agent.service
             └─504544 /opt/Elastic/Agent/elastic-agent watch --path.config /opt/Elastic/Agent --path.home /opt/Elastic/Agent

Aug 10 20:59:03 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Main process exited, code=exited, status=1/FAILURE
Aug 10 20:59:03 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Failed with result 'exit-code'.
Aug 10 20:59:03 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Unit process 504544 (elastic-agent) remains running after unit stopped.
Aug 10 20:59:03 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Consumed 6.047s CPU time.
```

At this point, I uninstalled the Agent. The command returned almost immediately, as usual, without blocking for too long.

```
$ sudo elastic-agent uninstall
Elastic Agent is installed but currently broken: service is not installed
Continuing will uninstall the broken Elastic Agent at /opt/Elastic/Agent. Do you want to continue? [Y/n]:y
```

Then I checked the service status.

```
$ sudo systemctl status elastic-agent.service
Warning: The unit file, source configuration file or drop-ins of elastic-agent.service changed on disk. Run 'systemctl daemon-reload' to reload uni>
○ elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; bad; vendor preset: enabled)
     Active: inactive (dead) (Result: exit-code) since Thu 2023-08-10 21:05:33 UTC; 15s ago
   Main PID: 506788 (code=exited, status=1/FAILURE)
      Tasks: 6 (limit: 4637)
     Memory: 102.9M
        CPU: 5.377s
     CGroup: /system.slice/elastic-agent.service
             └─507047 /opt/Elastic/Agent/elastic-agent watch --path.config /opt/Elastic/Agent --path.home /opt/Elastic/Agent

Aug 10 21:05:20 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Main process exited, code=exited, status=1/FAILURE
Aug 10 21:05:20 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Failed with result 'exit-code'.
Aug 10 21:05:20 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Unit process 507047 (elastic-agent) remains running after unit stopped.
Aug 10 21:05:20 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Consumed 5.370s CPU time.
Aug 10 21:05:33 shaunak-ubuntu-22-arm systemd[1]: Stopped Elastic Agent is a unified agent to observe, monitor and protect your system..
Aug 10 21:05:33 shaunak-ubuntu-22-arm systemd[1]: elastic-agent.service: Consumed 5.370s CPU time.
```
As the output above shows, the service was still running and, notably, the Upgrade Watcher process was still running. 

I kept watching the output of `sudo systemctl status elastic-agent.service`.  Exactly 2 minutes after the Agent crash happened, the Upgrade Watcher process was killed AND the service was also uninstalled.  This is definitely a better experience than with using the `TimeoutStopSec` setting.  This behavior also alleviates a concern I had with using the `KillMode=process` setting: what if one of the Agent's component sub-processes, e.g. Metricbeat, hangs? Will it prevent the service from being uninstalled?  That doesn't appear to be the case, given what happened to the Upgrade Watcher process.

I also wanted to test that, with the `KillMode=process` setting a rollback would've still been executed had I not uninstalled the Agent while the Upgrade Watcher was running. I essentially repeated all the steps above, skipping over the Agent uninstall step.  Within a couple of minutes after the upgraded Agent crashed, the Upgrade Watcher initiated the rollback.  I confirmed this by checking the Upgrade Watcher logs.

*Conclusions*:
* Using `KillMode=process` does prevent `systemd` from prematurely killing off the Upgrade Watcher process, thus allowing it enough time to do its job.
* Using `KillMode=process`, when `elastic-agent uninstall` is executed while the Upgrade Watcher is running, the service (and any lingering processes, e.g. the Upgrade Watcher) are removed/killed faster than when using the  `TimeoutStopSec` setting.

internal/pkg/agent/install/svc.go
```diff
@@ -106,3 +114,35 @@ const darwinLaunchdConfig = `<?xml version='1.0' encoding='UTF-8'?>
   </dict>
 </plist>
 `
+
+// A copy of the systemd config template from github.com/kardianos/service
+// with added .Config.Option.TimeoutStopSec option
```
pchila, line 119: Nit: this comment is still referring to the first fix...

```suggestion
// A copy of the systemd config template from github.com/kardianos/service
// with added .Config.Option.KillMode option
```

internal/pkg/agent/install/svc.go
```diff
@@ -54,6 +54,14 @@ func newService(topPath string) (service.Service, error) {
 		},
 	}
 
+	if runtime.GOOS == "linux" {
+		// The github.com/kardianos/service library doesn't support TimeoutStopSec in their prebuilt template.
+		// This option allows to pass our own template for the systemd unit configuration, which is a copy
+		// of the prebuilt template with added TimeoutStopSec option
```
pchila, line 60:
```suggestion
		// The github.com/kardianos/service library doesn't support KillMode in their prebuilt template.
		// This option allows to pass our own template for the systemd unit configuration, which is a copy
		// of the prebuilt template with added KillMode option
```

pchila (approved):
Just 2 small updates in the comments, looks good otherwise

internal/pkg/agent/install/svc.go
ycombinator: Thanks for catching.  Fixed in ad152a41ef3ca16763d2f65cfea67b27e54864bd.

pchila (approved):
LGTM

ycombinator:
buildkite test this

ycombinator:
> Should we update the systemd service file that is used for the DEB/RPM? Do we want it to act the same as when the `elastic-agent install` is used? Being that DEB/RPM cannot be self-upgraded it doesn't have to be the same, just didn't know if we want to ensure uniformity.

Yes, we should! @cmacknz had mentioned it earlier too, I think.  Thanks for the reminder.  I'll add those changes to this PR.

> How are we going to get this change in an upgrade? This will only be a fix for new installations.

Hmm, this is a tricky one.  In general I'm thinking we need some way to make this change during the upgrade process but I can already think of a few issues with that idea.  Let me experiment a bit and see what might be possible.  Do you think this issue is is a blocker for this PR?  Or could we address it in a follow up PR?

internal/pkg/agent/install/svc.go
```diff
@@ -54,6 +54,14 @@ func newService(topPath string) (service.Service, error) {
 		},
 	}
 
+	if runtime.GOOS == "linux" {
+		// The github.com/kardianos/service library doesn't support KillMode in their prebuilt template.
+		// This option allows to pass our own template for the systemd unit configuration, which is a copy
+		// of the prebuilt template with added KillMode option
```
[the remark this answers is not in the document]
ycombinator: Added explanation in 4e731ddd27c53092b728404bcd03b88e88f77898.

ycombinator:
> Should we update the systemd service file that is used for the DEB/RPM? Do we want it to act the same as when the `elastic-agent install` is used? Being that DEB/RPM cannot be self-upgraded it doesn't have to be the same, just didn't know if we want to ensure uniformity.

Done in c3f934be8cd35c0ce732b066e55efd9ee7da0c6c.

cmacknz (approved):

blakerouse (approved):
@ycombinator Don't think we should block this PR on upgrades not updating this file. I do think it is something that we need to do though.

ycombinator:
> I can already think of a few issues with that idea [of updating the unit file during an upgrade]. 

The two issues I was/am thinking of are:

1. The `root` requirement, which @cmacknz also mentioned in https://github.com/elastic/elastic-agent/pull/3220#issuecomment-1675272578, especially as we're now actively working on removing this requirement for running Agent.  

   My preference here is to go ahead with this change now and deal with the `root` requirement when we deal with other aspects of Agent that require `root`.

2. Whether to update the unit configuration file (and issue the `systemd daemon-reload`) from the old (pre-upgrade) Agent as one of the upgrade steps OR to do this update as part of a "migration" in the new (post-upgrade) Agent.  There are pros/cons to each approach:
   * Making the change from the new Agent as part of starting up
      * Pro: users upgrading _to_ Agent v8.10.0 (assuming that's when the code is released) or later version will benefit.
      * Con: if the new Agent crashes immediately upon start, the unit configuration change will not get deployed.
   * Making the change from the old Agent as one of the upgrade steps
      * Pro: if the new Agent crashes immediately upon start, the unit configuration change will have been deployed.
      * Con: only users who upgrade _from_ Agent v8.10.0 or later  will benefit.

   My preference is to make the change from the new Agent as:
   - it fits with the philosophy of controlling more of the upgrade process from the new Agent, e.g. the change we recently made to launch the new Agent's watcher, as it gives us the ability to fix bugs with new releases.
   - it should be rare for an Agent to crash immediate upon start; hopefully this is the sort of thing we'd either be able to catch well before a release or, if it happens anyway, hopefully it would affect a small segment of users due to some corner case deployment configuration.

 WDYT?

cmacknz:
> My preference here is to go ahead with this change now and deal with the root requirement when we deal with other aspects of Agent that require root.

👍 Agreed.

> My preference is to make the change from the new Agent as:

Agreed. I would suggest thinking of this as yet another state convergence problem. The agent should eventually be able to edit its unit configuration file, and should periodically retry until the file matches what is expected. A single failure shouldn't cause the change not to take effect. Remember that the power from the system can be removed at essentially any moment, so anything we do that doesn't retry will fail in that scenario.

The only other problem to consider is if users have manually edited parts of the configuration file, which we can likely handle by only changing the keys we care about and not replacing the entire file.

ycombinator:
> Don't think we should block this PR on upgrades not updating this file. I do think it is something that we need to do though.

Based on recent comments on the implementation, I believe I can get this addition into this PR in time for the 8.10 FF.  So I'm going to do that instead of a follow up PR.

blakerouse:
With us updating the unit file on upgrade, I believe we are going to need to do one more test.

The open question I have in my head is what does systemd do when the KillMode is changed will the service is already running?

If systemd creates the cgroup at start time of the process it could be possible that systemd will still exhibit the same behavior until the service gets completely restarted. We need to determine if we need a completely true restart of the Elastic Agent after the reload for the KillMode change to fully take effect.

ycombinator:
> With us updating the unit file on upgrade, I believe we are going to need to do one more test.
> 
> The open question I have in my head is what does systemd do when the KillMode is changed will the service is already running?
> 
> If systemd creates the cgroup at start time of the process it could be possible that systemd will still exhibit the same behavior until the service gets completely restarted. We need to determine if we need a completely true restart of the Elastic Agent after the reload for the KillMode change to fully take effect.

Okay, I will test this out, thanks.

If it turns out we need to restart the `elastic-agent` service after the unit config change and `systemd daemon-reload`, that should be possible too with `systemctl restart elastic-agent` (or the equivalent Golang API call to the https://github.com/kardianos/service/ library).

internal/pkg/agent/cmd/svc_update.go
```diff
@@ -0,0 +1,12 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
ycombinator, line 1: Should this file be moved to `internal/pkg/agent/application/upgrade/service_update.go` so it lives along side `internal/pkg/agent/application/upgrade/service.go`?

internal/pkg/agent/cmd/svc_update_linux.go
```diff
@@ -0,0 +1,48 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
ycombinator, line 1: Should this file be moved to `internal/pkg/agent/application/upgrade/service_update_linux.go` so it lives along side `internal/pkg/agent/application/upgrade/service.go`?

ycombinator:
@cmacknz @blakerouse I've addressed both your recent feedback and updated this PR accordingly.  I've also updated the "How to test this PR locally" section with two tests, the second of which answers @blakerouse's open question from https://github.com/elastic/elastic-agent/pull/3220#issuecomment-1675393562 (the answer is that the `KillMode=process` setting takes effect immediately and has an impact on the _already running_ Agent service).

Please re-review when you get a chance.  Thanks!

internal/pkg/agent/cmd/svc_update.go
blakerouse: Yes I think it should.

internal/pkg/agent/cmd/svc_update_linux.go
blakerouse: Yes.

internal/pkg/agent/cmd/svc_update_linux.go
```diff
@@ -0,0 +1,61 @@
...
+		// Nothing more to do
+		return nil
+	}
+
+	// If KillMode= is not present, add it and set it to "process"
+	// See https://github.com/elastic/elastic-agent/pull/3220
+	cfg.Section("Service").Key("KillMode").SetValue("process")
+	if err := cfg.SaveTo(unitFilePath); err != nil {
+		return fmt.Errorf("error writing updated systemd unit file [%s]: %w", unitFilePath, err)
+	}
```
blakerouse, line 58: I think we should NOTE that on a rollback this will not be changed back. The older Elastic Agent doesn't have a way to ensure that the service file is in its expected content.

I wonder how hard it would be to use the 3rd party `service` module that we use to render the file in memory and then just compare the contents of the file? That way we are not doing something specific here. It could just compare the 2 file contents and if different replace it with the new one and reload.

Is it not possible with the 3rd part `service` module to just have it overwrite the file instead of doing it here directly?

internal/pkg/agent/cmd/run.go
```diff
@@ -579,6 +579,18 @@ func handleUpgrade() error {
 		return nil
 	}
 
+	if err := ensureInstallMarkerPresent(); err != nil {
+		return err
+	}
+
+	if err := ensureServiceConfigUpToDate(); err != nil {
```
blakerouse, line 586: I think it would be good to have this as `upgrade.EnsureServiceConfig()`. Related to your ask below.

blakerouse (commented):
@ycombinator Awesome work! Glad to hear that the `KillMode=process` take effect without a restart!

internal/pkg/agent/cmd/svc_update_linux.go
ycombinator: > I wonder how hard it would be to use the 3rd party `service` module that we use to render the file in memory and then just compare the contents of the file? That way we are not doing something specific here. It could just compare the 2 file contents and if different replace it with the new one and reload.

I looked through https://pkg.go.dev/github.com/kardianos/service#section-documentation and I'm not seeing any APIs exposed by the `service` module that would help load the existing unit configuration file into memory (essentially a `service.Config` object) and compare it against our `service.Config` object.

> Is it not possible with the 3rd part `service` module to just have it overwrite the file instead of doing it here directly?

The reason I didn't go with the overwrite approach is because of a consideration that @cmacknz brought up in https://github.com/elastic/elastic-agent/pull/3220#issuecomment-1675311109: 

> The only other problem to consider is if users have manually edited parts of the configuration file, which we can likely handle by only changing the keys we care about and not replacing the entire file.

That's why I went with the approach of loading the current unit configuration file into memory, making a surgical edit if needed, and writing it back to disk.

internal/pkg/agent/cmd/run.go
ycombinator: Done in 09ec4a120a90bc3e01d92db93c5887a9fc56401f.

cmacknz (approved):
LGTM, thanks! I like how small the code to edit the systemd configuration turned out.

Could we modify one of the upgrade integration tests to check that KillMode process change takes effect? The best case would be to always do this on a test from a pre-8.10 version to 8.10+.

ycombinator:
> Could we modify one of the upgrade integration tests to check that KillMode process change takes effect? The best case would be to always do this on a test from a pre-8.10 version to 8.10+.

Sure, that will be done as part of https://github.com/elastic/elastic-agent/pull/3085. It was while writing the test in that PR that I found this bug.

internal/pkg/agent/cmd/svc_update_linux.go
blakerouse: True, if they manually edited the file.

blakerouse (changes requested):
See inline comment.

internal/pkg/agent/application/upgrade/service_update_linux.go
```diff
@@ -0,0 +1,64 @@
...
+func EnsureServiceConfigUpToDate() error {
+	switch service.ChosenSystem().String() {
+	case "linux-systemd":
+		unitFilePath := "/etc/systemd/system/" + paths.ServiceName + ".service"
+		if err := ensureSystemdServiceConfigUpToDate(unitFilePath); err != nil {
+			return err
+		}
+
+		// Reload systemd unit configuration files
+		cmd := exec.Command("systemctl", "daemon-reload")
```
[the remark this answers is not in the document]
ycombinator: Yeah, good point.  Will change!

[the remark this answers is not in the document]
ycombinator: Done in beb7ce9273f67efce005130fa63075d2ade392b8.

blakerouse (approved):
Looks good! Thanks for addressing all the feedback.

PR #3223, opened 2023-08-10
Propagate apm config

### LINKED ISSUE #2612: Propagate APM tracing configuration to sub-processes via the control protocol
The primary goal of this issue is to enabling trace collection on Integrations Server on Cloud by default to send to our internal APM backends. This tracing configuration will be specified in the `elastic-agent.yml` file on the server side and will not be part of the agent policy retrieved by Fleet.

See the initial discussion in https://github.com/elastic/elastic-agent/issues/1969

### Requirements

- When tracing is configured in Agent's local configuration, but absent in the Fleet agent policy, the local tracing configuration should take precedence
- APM tracing configuration must be propagated to sub-processes via the control protocol
	- We should not use environment variables since this will not always be possible in some runtimes
- elastic-agent-client should be updated to support this configuration
- We need to prevent any sensitive information from leaking via APM tracing data
	- We should audit existing traces
	- We should propose / consider any additional protections that need to be enabled to avoid leaking sensitive data
	- Areas to focus on: requests to Fleet Server, handling of the agent policy (local and remote from Fleet), gRPC control protocol messages to sub-processes

pchila:
## What does this PR do?
This change propagates the APM configuration set up for agent to the components that are managed.

We only support Elastic APM at the moment and at the time of this change this can only be configured in `elastic-agent.yml` configuration file.

For fleet-managed agents we include a workaround that will inject this configuration from the config file; this workaround inject the configuration from the config file in all the config changes received from Fleet, however it does not support hot reloading in this configuration: any changes to the apm configuration will take effect after a restart. 

Here's what an example of APM configuration looks like:
```yaml
agent.monitoring:
  ...
  traces: true
  apm:
    hosts:
      - https://12519f377f2449228f5095a5b549fe49.apm.us-central1.gcp.qa.cld.elstc.co:443
    environment: test-apm-changed
    secret_token: secret
    tls:
      skip_verify: true
```

## Why is it important?

We want to manage the apm configuration from a single place for all components run via Elastic Agent.

## Checklist

## Author's Checklist

- [x] Add a config injector to restore apm config for fleet-managed agents (fleet policy overwrites whatever we have in the config file)

## Related issues

- Requires elastic/elastic-agent-client#82
- Closes #2612 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/apm_config_modifier.go
+++ b/internal/pkg/agent/application/apm_config_modifier.go
@@ -3,9 +3,12 @@ package application
 import (
 	"fmt"
 
+	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
+
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator"
 	"github.com/elastic/elastic-agent/internal/pkg/config"
 	moncfg "github.com/elastic/elastic-agent/internal/pkg/core/monitoring/config"
+	"github.com/elastic/elastic-agent/pkg/component/runtime"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/errors"
@@ -15,9 +18,19 @@ import (
 
 func InjectAPMConfig(comps []component.Component, cfg map[string]interface{}) ([]component.Component, error) {
 
+	tracesEnabled, err := getAPMTracesEnabled(cfg)
+	if err != nil {
+		return comps, fmt.Errorf("error retrieving traces flag: %w", err)
+	}
+
+	if !tracesEnabled {
+		// nothing to do
+		return comps, nil
+	}
+
 	apmConfig, err := getAPMConfigFromMap(cfg)
 	if err != nil {
-		return comps, fmt.Errorf("error injecting apm config: %w", err)
+		return comps, fmt.Errorf("error retrieving apm config: %w", err)
 	}
 
 	if apmConfig == nil {
@@ -26,11 +39,35 @@ func InjectAPMConfig(comps []component.Component, cfg map[string]interface{}) ([
 	}
 
 	for i, _ := range comps {
-		comps[i].APM = new(component.APMConfig)
-		comps[i].APM.Elastic = (*component.ElasticAPM)(apmConfig)
+		// We shouldn't really go straight from config datamodel to protobuf datamodel (a core datamodel would be nice to
+		// abstract from protocol details)
+		if comps[i].Component == nil {
+			comps[i].Component = new(proto.Component)
+		}
+		comps[i].Component.ApmConfig = runtime.MapAPMConfig(apmConfig)
 	}
 
 	return comps, nil
+}
+
+func getAPMTracesEnabled(cfg map[string]any) (bool, error) {
+
+	rawTracesEnabled, err := utils.GetNestedMap(cfg, "agent", "monitoring", "traces")
+	if errors.Is(err, utils.ErrKeyNotFound) {
+		// We didn't find the key, return false without any error
+		return false, nil
+	}
+
+	if err != nil {
+		return false, fmt.Errorf("error accessing trace flag: %w", err)
+	}
+
+	traceEnabled, ok := rawTracesEnabled.(bool)
+	if !ok {
+		return false, fmt.Errorf("trace flag has unexpected type %T", rawTracesEnabled)
+	}
+
+	return traceEnabled, nil
 }
 
 func getAPMConfigFromMap(cfg map[string]any) (*moncfg.APMConfig, error) {
@@ -73,6 +110,12 @@ func PatchAPMConfig(log *logger.Logger, rawConfig *config.Config) func(change co
 		return noop
 	}
 
+	tracesEnabled, err := getAPMTracesEnabled(configMap)
+	if err != nil {
+		log.Errorf("error retrieving trace flag, patching disabled: %v", err)
+		return noop
+	}
+
 	apmConfig, err := getAPMConfigFromMap(configMap)
 
 	if err != nil {
@@ -80,13 +123,18 @@ func PatchAPMConfig(log *logger.Logger, rawConfig *config.Config) func(change co
 		return noop
 	}
 
-	if apmConfig == nil {
-		log.Debug("no apm config set, patching disabled")
+	if tracesEnabled == false && apmConfig == nil {
+		// traces disabled and no apm config -> no patching happening
+		log.Debugf("traces disabled and no apm config: no patching necessary")
 		return noop
+	}
+	monitoringPatch := map[string]any{"traces": tracesEnabled}
+	if apmConfig != nil {
+		monitoringPatch["apm"] = apmConfig
 	}
 
 	return func(change coordinator.ConfigChange) coordinator.ConfigChange {
-		err := change.Config().Merge(map[string]any{"agent": map[string]any{"monitoring": map[string]any{"apm": apmConfig}}})
+		err := change.Config().Merge(map[string]any{"agent": map[string]any{"monitoring": monitoringPatch}})
 		if err != nil {
 			log.Errorf("error patching apm config into configchange: %v", err)
 		}
--- a/internal/pkg/agent/application/apm_config_modifier.go
+++ b/internal/pkg/agent/application/apm_config_modifier.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
 package application
 
 import (
--- a/internal/pkg/agent/application/apm_config_modifier.go
+++ b/internal/pkg/agent/application/apm_config_modifier.go
@@ -42,7 +42,7 @@ func InjectAPMConfig(comps []component.Component, cfg map[string]interface{}) ([
 		return comps, nil
 	}
 
-	for i, _ := range comps {
+	for i := range comps {
 		// We shouldn't really go straight from config datamodel to protobuf datamodel (a core datamodel would be nice to
 		// abstract from protocol details)
 		if comps[i].Component == nil {
@@ -127,7 +127,7 @@ func PatchAPMConfig(log *logger.Logger, rawConfig *config.Config) func(change co
 		return noop
 	}
 
-	if tracesEnabled == false && apmConfig == nil {
+	if !tracesEnabled && apmConfig == nil {
 		// traces disabled and no apm config -> no patching happening
 		log.Debugf("traces disabled and no apm config: no patching necessary")
 		return noop
--- a/internal/pkg/agent/application/apm_config_modifier.go
+++ b/internal/pkg/agent/application/apm_config_modifier.go
@@ -20,6 +20,8 @@ import (
 	"github.com/elastic/elastic-agent/pkg/utils"
 )
 
+// InjectAPMConfig is a modifier passed to coordinator in order to set the global APM configuration used for the agent
+// into each Component coming from input/output configuration
 func InjectAPMConfig(comps []component.Component, cfg map[string]interface{}) ([]component.Component, error) {
 
 	tracesEnabled, err := getAPMTracesEnabled(cfg)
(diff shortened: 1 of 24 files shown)


internal/pkg/agent/application/application.go
```diff
@@ -155,6 +158,7 @@ func New(
 			compModifiers = append(compModifiers, FleetServerComponentModifier(cfg.Fleet.Server),
 				InjectFleetConfigComponentModifier(cfg.Fleet, agentInfo),
 				EndpointSignedComponentModifier(),
+				InjectAPMConfig,
```
blakerouse, line 161: Is there ever a time that `InjectAPMConfig` is not used? If its always used we should probably just place it directly in the component model generator. That would ensure that it matches the same as `elastic-agent inspect`.

blakerouse (commented):
I know this is in draft, but want to provide some early feedback. I like how this is looking.

internal/pkg/agent/application/application.go
pchila: The only case is the testing mode on top of the if/else block... Since that's something coming from an integration test I didn't want the agent to touch anything...

blakerouse: I think even in testing you would want it to be the same as well. If a test provides the configuration for enabling APM, then it should be enabled in the test as well.

AndersonQ (changes requested):
Nothing serious, but I believe you still need to update to a released version of the `elastic-agent-client` and there was an `append` that was doing nothing. So either you forgot to add the element to append, or the line should be removed

pkg/component/config.go
```diff
@@ -15,6 +15,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
+
 	"github.com/elastic/elastic-agent/pkg/limits"
```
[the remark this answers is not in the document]
pchila: I think that is goimports as it separates the `elastic-agent-client` (3rd party lib) from the `elastic-agent` (own module)...

I will check that again...

pkg/component/fake/component/main.go
```diff
@@ -68,29 +68,41 @@ func run() error {
...
-				logger.Info().
-					Str("state", state.String()).
-					Str("expectedState", u.Expected().State.String()).
-					Msg("unit change received")
-			} else {
-				logger.Info().Msg("unit change received, but no unit on it")
+				logger.Info().Msgf("unit: %+v", change.Unit)
+				if change.Unit.Expected().APMConfig != nil {
+					logger.Info().Msgf("apmConfig: %+v", change.Unit.Expected().APMConfig)
+				}
```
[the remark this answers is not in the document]
pchila: That is for me to debug stuff, will remove

pkg/component/fake/component/README.md
```diff
@@ -1,3 +1,53 @@
 # Fake Component
 
-Controllable through GRPC control protocol with actions. Allows unit tests to simulate control and communication with a running sub-process.
+Controllable through GRPC control protocol with actions. Allows tests to simulate control and communication with a running sub-process.
+
+
+## How to use the fake component
+If we need to use the fake component for a manual test, we need to build it suing
```
dmathieu, line 7:
```suggestion
If we need to use the fake component for a manual test, we need to build it using
```

internal/pkg/agent/application/application.go
```diff
@@ -136,19 +138,21 @@ func New(
 			log.Debugf("Reloading of configuration is on, frequency is set to %s", cfg.Settings.Reload.Period)
 			configMgr = newPeriodic(log, cfg.Settings.Reload.Period, discover, loader)
 		}
+
+		compModifiers = append(compModifiers)
```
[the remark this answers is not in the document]
pchila: Nope just the result of iterative changes xD
Will clean that up

internal/pkg/agent/application/application.go
```diff
@@ -136,19 +138,21 @@ func New(
...
 	} else {
 		isManaged = true
 		var store storage.Store
 		store, cfg, err = mergeFleetConfig(rawConfig)
 		if err != nil {
 			return nil, nil, nil, err
 		}
-
+		log.Info("configuration loaded: %v", cfg)
```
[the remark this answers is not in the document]
pchila: it's part of the debug commit  480228cbadc27a790ac9eb0d8757c066824de36f , will get rid of the log as a fixup just in case

internal/pkg/agent/application/coordinator/config_patcher.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (c ConfigPatchManager) Watch() <-chan ConfigChange {
+	return c.outch
+}
+
+func (c ConfigPatchManager) patch(src <-chan ConfigChange, dst chan ConfigChange) {
+	for ccc := range src {
+		dst <- c.patchf(ccc)
+	}
+}
```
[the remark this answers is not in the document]
pchila: This struct just applies patches to config returned by a ConfigManager.
Generating such a patch is task of the patchFn, so this can be used with any number of patching functions and the functions are easier to test

If the compiler inlines it, so be it :rofl:

pkg/component/fake/component/main.go
```diff
@@ -68,29 +68,41 @@ func run() error {
...
+	if change.Unit != nil {
+		u := change.Unit
+		state, _, _ := u.State()
+		logger.Info().
+			Str("state", state.String()).
+			Str("expectedState", u.Expected().State.String()).
+			Msg("unit change received")
+	} else {
+		logger.Info().Msg("unit change received, but no unit on it")
+	}
```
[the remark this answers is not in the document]
pchila: I just moved code from the main loop in `run()`

My assumption is that we still wanted to handle changes without a unit assigned in the fake component but I agree it looks weird... I wouldn't change this behavior in this PR thoughm just in case some test breaks because of it

pkg/component/runtime/apm_config_mapper.go
```diff
@@ -0,0 +1,39 @@
...
+	if conf == nil {
+		// component apm config is nil, so the pb config is nil as well
+		return nil
+	}
+
+	elasticAPMConf := &proto.ElasticAPM{
+		Environment:  conf.Environment,
+		ApiKey:       conf.APIKey,
+		SecretToken:  conf.SecretToken,
+		Hosts:        append([]string{}, conf.Hosts...),
```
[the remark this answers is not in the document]
pchila: It is indeed reduntant, removing

pkg/component/runtime/manager_test.go
```diff
@@ -482,6 +486,334 @@ func TestManager_FakeInput_Features(t *testing.T) {
...
+
+							retrievedApmConfig, err := extractAPMConfigFromActionResult(t, res)
+							if err != nil {
+								subscriptionErrCh <- fmt.Errorf("[case %d]: failed to retrieve APM Config from ActionResult %s: %w",
+									healthIteration, fakecmp.ActionRetrieveAPMConfig, err)
+								return false
+							}
+
+							return gproto.Equal(modifiedAPMConfig, retrievedApmConfig)
+						}, 1*time.Second, 100*time.Millisecond, "APM config was not received by component. expected: %s", modifiedAPMConfig)
```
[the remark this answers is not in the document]
pchila: same as above

pkg/component/runtime/manager_test.go
```diff
@@ -482,6 +486,334 @@ func TestManager_FakeInput_Features(t *testing.T) {
...
+							}
+
+							retrievedApmConfig, err := extractAPMConfigFromActionResult(t, res)
+							if err != nil {
+								subscriptionErrCh <- fmt.Errorf("[case %d]: failed to retrieve APM Config from ActionResult %s: %w",
+									healthIteration, fakecmp.ActionRetrieveAPMConfig, err)
+								return false
+							}
+							return retrievedApmConfig == nil
+						}, 1*time.Second, 100*time.Millisecond, "APM config was not received by component. expected: nil")
```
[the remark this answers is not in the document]
pchila: same as above

internal/pkg/agent/application/coordinator/config_patcher.go
```diff
@@ -0,0 +1,49 @@
...
+package coordinator
+
+import (
+	"context"
+)
+
+type ConfigPatch func(change ConfigChange) ConfigChange
+
+// ConfigPatchManager is a decorator to restore some agent settings from the elastic agent configuration file
+type ConfigPatchManager struct {
```
dmathieu, line 14: It seems this struct is lacking tests?

internal/pkg/agent/application/apm_config_modifier.go
```diff
@@ -0,0 +1,146 @@
...
+	monitoringcfg "github.com/elastic/elastic-agent/internal/pkg/core/monitoring/config"
+	"github.com/elastic/elastic-agent/pkg/component/runtime"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/errors"
+	"github.com/elastic/elastic-agent/pkg/component"
+	"github.com/elastic/elastic-agent/pkg/utils"
+)
+
+func InjectAPMConfig(comps []component.Component, cfg map[string]interface{}) ([]component.Component, error) {
```
dmathieu, line 23: As someone who doesn't know much/anything about elastic-agent, I find it hard to understand why we have to inject and then patch. How about documenting those two public methods?

internal/pkg/agent/application/coordinator/config_patcher.go
```diff
@@ -0,0 +1,49 @@
...
+	inner   ConfigManager
+	outCh   chan ConfigChange
+	patchFn ConfigPatch
+}
+
+func (c ConfigPatchManager) Run(ctx context.Context) error {
+	go c.patch(c.inner.Watch(), c.outCh)
```
dmathieu, line 21: An unchecked goroutine here feels a bit brittle/unsafe. Could we end up with an explosion of running goroutines? Should there be a timeout?

dmathieu (commented):
I don't know if that kind of doc already exists for other configs, but how about documenting how services should handle the added config?

internal/pkg/agent/application/apm_config_modifier.go
pchila: I will add some documentation, the short version is:
- `PatchAPMConfig` will patch agent configuration coming from fleet (which does not support APM configuration at the moment) by adding the APM config values retrieved from the elastic-agent.yaml file
- `InjectAPMConfig` will add APM configuration to the internal datamodel agent uses to determine which components and units to run

internal/pkg/agent/application/coordinator/config_patcher.go
dmathieu: Yes. If somehow, `inner.Run()` returns right away, but `inner.Watch()` doesn't at all, then you could end up with a goroutines leak.
I also don't have much context on this code at all, so maybe that's not really possible.

pchila: `inner.Watch()` returns the channel on which config changes are published. We use that channel to intercept configuration changes that we want to patch.
The basic pattern for a ConfigManager is to start it and then consume all the configuration changes passed on the channel. 
The ConfigPatchManager does the same thing with the addition of a goroutine that will read from the inner channel, patch and publish on its own channel

blakerouse (approved):
I really looked through the code. It is very well documented and has a good amount of tests. I really like the addition in the fake component to really test this change from Elastic Agent down to the actual running component.

pierrehilbert:
buildkite test this

AndersonQ (approved):

pchila:
@dmathieu

> I don't know if that kind of doc already exists for other configs, but how about documenting how services should handle the added config?

This feature is for internal use at the moment, so no official docs are created at this stage.

Regarding how services should handle the added config: it depends :) 
The basic idea is that the agent will propagate the APM configuration it receives to the components and units it manages and it expects that such components take appropriate action to reload/reinstantiate their APM instrumentation.
How that happens in detail depends on what sort of APM objects the component uses, for example:
- if the component uses a decorated http server it may be needed to stop (gracefully) the current server, recreate it with the new configuration and start the new one. 
- if it uses a custom Tracer object, it will need to create the new one, close the old one and swap them safely (this is for example what happens in the e2e test in this PR)

As of now the agent and the agent client **will not** set the APM env variables for the component or restart it, the processing is delegated to the component itself (if then the component decides that setting the env and reexecuting is the easier way to apply the configuration, that is obviously an option)

pchila:
buildkite test this

dmathieu:
> This feature is for internal use at the moment, so no official docs are created at this stage.

I don't mean necessarily official docs, but pointers for folks who (internally) will handle this?

pchila:
buildkite test this

pchila:
buildkite test this

pchila:
buildkite test this

pchila:
buildkite test this

pchila:
buildkite test this

pchila:
buildkite test this

PR #3225, opened 2023-08-10
testing/runner: use sudo rm on host machine and handle errors

AndersonQ:
When tests are executed with sudo, the fetched artefact and other created files are owned by root. Therefore to successfully remove the agent folder it must be done with sudo.

## What does this PR do?

Remove files on the integration tests host machine with sudo and handle the possible errors

## Why is it important?

When the test is run with sudo, some files will belong to root, which caused the removal of the `agent` folder to fail. Also handling the errors will surface any other problem that might still happen.

## Checklist

run any integration test using the local fetcher and the SNAPSHOT version. Rum it at least once . Rebuild the agent in a was the .sha512 file will change (or just change the file). Run the test again. 

Without this fix you should find in the host machine the `~/agent/.agent-cache/` files were not updated whereas the artefacts were copied by the test runner. On the home folder there will be an agent and it's sha512 file, this one should be the new one, whereas the `~/agent/.agent-cache/` is still the previous one.

## Related issues

N/A

## Questions to ask yourself

- ...


pkg/testing/runner/debian.go
```diff
@@ -97,8 +97,16 @@ func (DebianRunner) Copy(ctx context.Context, sshClient *ssh.Client, logger Logg
 	if err != nil {
 		return fmt.Errorf("failed to SCP repo archive %s: %w", repoArchive, err)
 	}
+
 	// ensure that agent directory is removed (possible it already exists if instance already used)
-	_, _, _ = sshRunCommand(ctx, sshClient, "rm", []string{"-rf", "agent"}, nil)
+	stdout, stderr, err := sshRunCommand(ctx,
+		sshClient, "sudo", []string{"rm", "-rf", "agent"}, nil)
+	if err != nil {
+		_ = fmt.Errorf(
```
blakerouse, line 105:
```suggestion
		return fmt.Errorf(
```

blakerouse (approved):
Looks good, I believe you need to add the missing return before merging this.

Issue #2273, opened 2023-02-15
[Agent-Upgrade 8.6.1>8.6.2]: Agent gets unhealthy on triggering upgrade and error `Unable to clean downloads before update`, however gets successfully upgraded.


amolnater-qasource:
**Kibana version:** 8.6.2 BC2 Kibana cloud environment

**Host OS:** All

**Build details:**
VERSION: 8.6.2 BC2 
BUILD: 59020
COMMIT: aabd37e8a9069bed4f63b8c457a6829f4581ab97
Artifact Link: https://www.elastic.co/downloads/past-releases/elastic-agent-8-6-1

**Preconditions:**
1. 8.6.2BC2 Kibana cloud environment should be available.
2. Agent Policy should be available with System and Endpoint defend integrations.

**Steps to reproduce:**
1. Install lower version 8.6.1 agents using agent policy.
2. Add BC artifact link under Fleet>Settings>Agent binary: `https://staging.elastic.co/8.6.2-ca679c34/downloads/`
3. Trigger agent upgrade from Fleet-UI and observe agent gets into Updating>Unhealthy state.
4. Observe error log `Unable to clean downloads before update`.
5. However, agent gets upgraded to 8.6.2 and also gets Healthy.

**Note:**
- This issue is reproducible on Mac, Linux and Windows VMs multiple times on different builds.

**Screenshots:**
![1](https://user-images.githubusercontent.com/77374876/218972701-5914e386-cd79-4e6e-bd38-13e4cbea7694.JPG)
![2](https://user-images.githubusercontent.com/77374876/218971814-075993ef-2990-45e9-8cf4-d19512ff3c8d.JPG)
![3](https://user-images.githubusercontent.com/77374876/218971816-47f893b4-8286-410e-8c55-109780f17afb.JPG)
![4](https://user-images.githubusercontent.com/77374876/218971818-9c1e95d2-a22c-4938-bd9f-499123a2e5c5.JPG)

**Logs:**
[[Upgrade]elastic-agent-diagnostics-2023-02-15T07-52-45Z-00.zip](https://github.com/elastic/elastic-agent/files/10740868/Upgrade.elastic-agent-diagnostics-2023-02-15T07-52-45Z-00.zip)

**Expected Result:**
On triggering upgrade agent should get in Updating state and then should directly get into Healthy state with latest version.

**Related issue with Unhealthy agents:**
https://github.com/elastic/elastic-agent/issues/2272

[...]

PR #3227, opened 2023-08-10
Fix mapping the upgrading state to degraded.

cmacknz:
Degraded is interpreted by users as the agent being in an error state. We should not report degraded for situations where the agent is working as expected.

Long term we should probably just send these states directly to Fleet rather than translating them into a less fine grained set of states.

- Closes https://github.com/elastic/elastic-agent/issues/2273


blakerouse (approved):
Looks good.

I agree it would be better to have Fleet know this states.

pierrehilbert (approved):

PR #3238, opened 2023-08-14
Daily BK job deleting stale GCE instances

pazone:
- Enhancement

## What does this PR do?

Daily buildkite job that cleans up stale GCE instances left after integration tests
Removes stale GCE instances having matching labels, name prefixes and older than 24 hours

## Why is it important?
Integration tests are being interrupted without a proper cleanup when we add new commits to pull requests.
It's a temporary solution. Normally we should make the CI system manage build agents. 
## Checklist

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -14,7 +14,7 @@ if [ -n "$TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE" ]; then
   fi
 fi
 
-sh .buildkite/hooks/unset-secrets.sh
+source .buildkite/scripts/unset-secrets.sh
 
 if command -v docker &>/dev/null; then
   DOCKER_REGISTRY="docker.elastic.co"
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -14,7 +14,7 @@ if [ -n "$TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE" ]; then
   fi
 fi
 
-unset GOOGLE_APPLICATION_GCP_SECRET API_KEY_TOKEN
+sh .buildkite/hooks/unset-secrets.sh
 
 if command -v docker &>/dev/null; then
   DOCKER_REGISTRY="docker.elastic.co"
--- a/.buildkite/misc/gce-cleanup.yml
+++ b/.buildkite/misc/gce-cleanup.yml
@@ -2,7 +2,7 @@
 version: "1.0"
 
 accounts:
-  - name: "elastic-platform-ingest"
+  - name: "${ACCOUNT_PROJECT}"
     driver: "gce"
     options:
       key: "${ACCOUNT_KEY}"
--- a/.buildkite/misc/gce-cleanup.yml
+++ b/.buildkite/misc/gce-cleanup.yml
@@ -1,0 +1,36 @@
+---
+version: "1.0"
+
+accounts:
+  - name: "elastic-platform-ingest"
+    driver: "gce"
+    options:
+      key: "${ACCOUNT_KEY}"
+      secret: "${ACCOUNT_SECRET}"
+      project: "${ACCOUNT_PROJECT}"
+
+scanners:
+  - account_name: "${ACCOUNT_PROJECT}"
+    resources:
+      - type: "node"
+        regions:
+          - "us-central1"
+        filters:
+          - type: "<"
+            pointer: "/extra/creationTimestamp"
+            param: "${CREATION_DATE}"
+            converters:
+              param: "date"
+              value: "date"
+          - type: "="
+            pointer: "/extra/labels/project"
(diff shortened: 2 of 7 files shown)


.buildkite/scripts/steps/gce-cleanup.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -exuo pipefail
+
+export ACCOUNT_KEY_SECRET=$(vault kv get -field=client_email $VAULT_PATH)
```
alexsapran, line 5: For reading information from the vault, we kind of followed a pattern to put those into the pre-command hook.

alexsapran (commented):
I am not against this, but why not have a proper cleanup from within the pipeline, maybe utilizing some hook that would be executed when the step fails?

.buildkite/scripts/steps/gce-cleanup.sh
pazone: Thanks, good point

pazone: Now the variables are being unset regardless of command exit code. What do you think if I keep it within the step script? For me it looks quite straightforward and doesn't require any additional if conditions

.buildkite/misc/gce-cleanup.yml
```diff
@@ -0,0 +1,36 @@
...
+accounts:
+  - name: "elastic-platform-ingest"
+    driver: "gce"
+    options:
+      key: "${ACCOUNT_KEY}"
+      secret: "${ACCOUNT_SECRET}"
+      project: "${ACCOUNT_PROJECT}"
+
+scanners:
+  - account_name: "${ACCOUNT_PROJECT}"
```
amannocci, line 13:
```suggestion
  - account_name: "elastic-platform-ingest"
```
The account name should match the account name defined at L5.

.buildkite/scripts/steps/gce-cleanup.sh
alexsapran: The intention is to have the script executable from ci as well as locally for debug. This is why it's better to set up the creds as part of the hook in ci.

.buildkite/scripts/unset-secrets.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+# Unset all variables ending with _SECRET
+for var in $(printenv | sed 's;=.*;;' | sort); do
+  if [[ "$var" == *_SECRET || "$var" == *_TOKEN]]; then
```
mrodm, line 6: Missing space ?
```suggestion
  if [[ "$var" == *_SECRET || "$var" == *_TOKEN ]]; then
```
In my laptop, this fails:
```shell
 $ if [[ "$var" == *_SECRET || "$var" == *_TOKEN]]; then echo "deleted"; else echo "don't"; fi
bash: syntax error in conditional expression: unexpected token `;'
bash: syntax error near `;'
```

.buildkite/scripts/unset-secrets.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+# Unset all variables ending with _SECRET
```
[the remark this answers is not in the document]
mrodm: What about adding the 5 possible suffixes here ? The default suffixes mentioned in the docs: https://buildkite.com/docs/pipelines/managing-log-output#redacted-environment-variables

For instance, in elastic-package it is used also SECRET_KEY and ACCESS_KEY
https://github.com/elastic/elastic-package/blob/f112509da6b8a8b9a4e54dfa2dec4a20fed74b23/.buildkite/hooks/pre-command#L50-L51

.buildkite/hooks/pre-exit
```diff
@@ -14,7 +14,7 @@ if [ -n "$TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE" ]; then
   fi
 fi
 
-unset GOOGLE_APPLICATION_GCP_SECRET API_KEY_TOKEN
+sh .buildkite/hooks/unset-secrets.sh
```
amannocci, line 17:
```suggestion
source .buildkite/scripts/unset-secrets.sh
```
The script isn't present in `.buildkite/hooks`.

amannocci (changes requested):

.buildkite/scripts/unset-secrets.sh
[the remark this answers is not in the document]
alexsapran: I think making it so that it unsets all of them would be better.
This script would make more sense to be part of our CI image, so we don't actually need to add it. But this is something we can do afterward.

.buildkite/hooks/pre-exit
pazone: Oh!

amannocci (approved):

mrodm (approved):
LGTM

Issue #682, opened 2022-07-06
Elastic Agent non-fleet upgrade between 8.3.x releases is broken 


aleksmaus:
Elastic Agent non-fleet upgrade between 8.3.x released is broken.
The upgrade via fleet works. 

The original discussion thread is here:
[Update to 8.3.1 from 8.3.0 has broken Fleet - please help!](https://discuss.elastic.co/t/update-to-8-3-1-from-8-3-0-has-broken-fleet-please-help/308654)

This is effectively not an upgrade, but install of the different version of the agent without upgrade handling code invoked.

The root cause of the issue is the agent secret is not properly migrated if the new version of the agent is installed on top of the existing agent:
1. the fleet.enc file is in the top directory of the agent
2. the vault directory (the encrypted agent secret on linux/windows) is under the hashed data path data/elastic-agent-7a475d
3. when the new version of the agent is installed with apt it’s installed side by side into it’s own directory something like `data/elastic-agent-3bf26a
4. the vault is only copied over if upgrade is done via fleet (upgrade handler in the agent) (!!!)
5. when the new version of the agent starts from the new location it doesn’t find the vault and creates the new agent key.
6. the new version of the agent can’t decrypt existing fleet.enc

For confirmed bugs, please report:
- Version: 8.3.x
- Operating System: Linux/Windows (Non-fleet initiated upgrades)
- Discuss Forum URL: [Update to 8.3.1 from 8.3.0 has broken Fleet - please help!](https://discuss.elastic.co/t/update-to-8-3-1-from-8-3-0-has-broken-fleet-please-help/308654)

Possible solution:
In the next release 8.3.3 or 8.4.0 invoke migration code scanning for the existing agent data directories and perform the agent secret migration only if the secret is not found. The migration should probably move up the vault dir to the top agent directory, near fleet.enc file so it can be shared between future installations and would not have to be migrated if the newer version of the agent is installed on top of the existing agent directory from .deb/.rpm for example.

christophercutajar:
@aleksmaus we're experiencing this behaviour on one of our Linux boxes after running `yum upgrade elastic-agent`.  We're documenting this issue in https://github.com/elastic/seceng/issues/4248

As a solution, would we need to re-enroll the agent again to Fleet? How can we get the agent in a running state? Thanks

aleksmaus:
The easiest way to make the new install of the agent work again in this situation is to copy over the "vault" directory from the previous agent installation. 
[...]

PR #3240, opened 2023-08-15
Harden linux/windows vault implementation for concurrent access.

aleksmaus:
## What does this PR do?

Harden linux/windows vault implementation for concurrent access.

#### Problem details

During repetitive E2E tests runs (took over 5 hours to reproduce) the file based vault implementation (that is used for linux and Windows) got into the state where the key value was corrupted, the resulted in the error:
```
Enrollment failed after failed to store agent config: could not save enrollment information: failed to ensure key: 
could not get agent key: could not decrypt key: cipher: message authentication failed
```
It's difficult to reproduce in the normal agent usage scenario, but looking at the implementation there is no mitigation against multiple processes accessing vault. The lower level API allows to reproduce the situation artificially see ```TestVaultConcurrent``` test https://github.com/elastic/elastic-agent/compare/main...aleksmaus:feature/harden_fsvault?expand=1#diff-8a5388ab5e868a37ee07711c7978dc51561a302be8fa560946038e71375fe70bR183

#### Solution

Harden linux/windows vault implementation for concurrent access.

* Use "atomic" write/rename for the data files https://github.com/elastic/elastic-agent/compare/main...aleksmaus:feature/harden_fsvault?expand=1#diff-5176984040d0726ff96ca6e27e372355b57829a8f657c706fd6960a6e4f96a45R17
* Use file locking https://github.com/elastic/elastic-agent/compare/main...aleksmaus:feature/harden_fsvault?expand=1#diff-6c9bf532380b28932cd23bc8d308c07263fa77a517d41fc0e2fd390d5e361a9eR71. The ```.lock``` file is created inside of the ```vault``` directory now.
* For the file locking API needed to pass context.Context down to the storage, that lead to the storage API change and the code that uses the storage is modified in order to pass the context down.
* Removed ```migrate_secret.go``` that doesn't seem to be used anywhere anymore.

## Why is it important?

Hardens the vault data agains corruption when it's being accessed concurrently. 

## Checklist


internal/pkg/agent/migration/migrate_secret.go
```diff
@@ -1,163 +0,0 @@
...
-	"github.com/elastic/elastic-agent/internal/pkg/agent/application/secret"
-	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
-	"github.com/elastic/elastic-agent/internal/pkg/fileutil"
-)
-
-const (
-	darwin = "darwin"
-)
-
-// MigrateAgentSecret migrates agent secret if the secret doesn't exists agent upgrade from 8.3.0 - 8.3.2 to 8.x and above on Linux and Windows platforms.
```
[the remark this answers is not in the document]
aleksmaus: I have no idea, but it's not called from anywhere.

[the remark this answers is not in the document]
cmacknz: Looks like we lost this in the big 8.6 merge https://github.com/elastic/elastic-agent/blob/006784462a006cc31fbba70ec43522500c1816bb/internal/pkg/agent/cmd/run.go#L132-L140

Here was the original bug: https://github.com/elastic/elastic-agent/issues/682

Would need to confirm if that situation is still possible with the change in directory structure in 8.6 and above.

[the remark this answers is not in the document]
aleksmaus: Whooops. Been doing ok without it for few releases. 🤷

internal/pkg/agent/vault/vault_linux.go
```diff
@@ -157,3 +224,40 @@ func deriveKey(pw []byte, salt []byte) ([]byte, []byte, error) {
 func (v *Vault) filepathFromKey(key string) string {
 	return filepath.Join(v.path, fileNameFromKey(v.key, key))
 }
+
+// writeFile "atomic" file write, utilizes temp file and replace
+// os.CreateTemp creates the file with 0600 mask, which is what we need
+func writeFile(fp string, data []byte) (err error) {
```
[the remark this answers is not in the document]
aleksmaus: the code is fairly small, will it keep the vault package separate from other agent libs. makes it easier to extract, put into different smaller project to test cross-platform etc.

cmacknz (approved):

internal/pkg/agent/vault/vault_linux.go
```diff
@@ -59,42 +65,90 @@ func New(path string, opts ...OptionFunc) (v *Vault, err error) {
...
+	return r, nil
+}
+
+// try to acquire exclusive lock
+func (v *Vault) tryLock(ctx context.Context) error {
+	_, err := v.lock.TryLockContext(ctx, v.retryDelay)
+	if err != nil {
+		return fmt.Errorf("failed to acquire exclusive lock: %v, err: %w", v.lock.Path(), err)
+	}
+	return err
```
AndersonQ, line 96: [Blocker]
Please, `err` is always `nil` so return `nil`.
```suggestion
	return nil
```

internal/pkg/agent/vault/vault_linux.go
```diff
@@ -59,42 +65,90 @@ func New(path string, opts ...OptionFunc) (v *Vault, err error) {
...
+}
+
+// try to acquire shared lock
+func (v *Vault) tryRLock(ctx context.Context) error {
+	_, err := v.lock.TryRLockContext(ctx, v.retryDelay)
+	if err != nil {
+		return fmt.Errorf("failed to acquire shared lock: %v, err: %w", v.lock.Path(), err)
+	}
+	return err
```
AndersonQ, line 105: [Blocker]
Please, `err` is always `nil` so return `nil`.
```suggestion
	return nil
```

internal/pkg/agent/vault/vault_linux.go
```diff
@@ -59,42 +65,90 @@ func New(path string, opts ...OptionFunc) (v *Vault, err error) {
...
 
-// Close closes the valut store
+// unlock unlocks the file lock and preserves the original error if there was a error
+func (v *Vault) unlock(err error) error {
+	unerr := v.lock.Unlock()
+	if err != nil {
+		return err
+	}
+	return unerr
+}
```
AndersonQ, line 115: [Blocker]
Both errors are relevant. Better to use `errors.Join` and return both of them.
You can even just pass the errors to `errors.Join` and it'll take care of omitting the `nil` errors and returning `nil` if all errors are `nil`

```suggestion
func (v *Vault) unlock(err error) error {
	unerr := v.lock.Unlock()
	if err != nil {
		return err
	}
	return errors.Join(unerr, err)
}
```

internal/pkg/agent/vault/vault_linux.go
```diff
@@ -157,3 +224,40 @@ func deriveKey(pw []byte, salt []byte) ([]byte, []byte, error) {
...
+	dir, fn := filepath.Split(fp)
+	if dir == "" {
+		dir = "."
+	}
+
+	f, err := os.CreateTemp(dir, fn)
+	if err != nil {
+		return fmt.Errorf("failed creating temp file: %w", err)
+	}
+	defer func() {
```
AndersonQ, line 240: [Suggestion]
1 - move the comment about the `os.CreateTem` to where it's used.
2 - I think the sentence could be improved

```suggestion
// writeFile performs an "atomic" file write operation, using a temporary file
// and then replacing the original.
func writeFile(fp string, data []byte) (err error) {
	dir, fn := filepath.Split(fp)
	if dir == "" {
		dir = "."
	}

	// os.CreateTemp creates the file with a 0600 permission mask,
	// which is the required permission setting.
	f, err := os.CreateTemp(dir, fn)
	if err != nil {
		return fmt.Errorf("failed creating temp file: %w", err)
	}
	defer func() {
```

internal/pkg/agent/vault/vault_linux.go
```diff
@@ -157,3 +224,40 @@ func deriveKey(pw []byte, salt []byte) ([]byte, []byte, error) {
...
+		if err != nil {
+			_ = os.Remove(f.Name())
```
AndersonQ, line 242: [Blocker]
Please add this to the error as well.
You can join this error with the other and an extra message to clearly states this error is from the cleanup.

```suggestion
			rerr := os.Remove(f.Name())
			if rerr !=nil{
				err= errors.Join(err, fmt.Errorf("cleanup failed, could not remove temp file: %w", rerr))
			}
```

internal/pkg/agent/vault/vault_options_nondarwin.go
```diff
@@ -0,0 +1,23 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build linux || windows
```
AndersonQ, line 5: If it's _nondarwin_, why not use:
```suggestion
//go:build !darwin
```

internal/pkg/agent/vault/vault_test.go
```diff
@@ -164,3 +174,73 @@ func TestVault(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("failed to set secret: %w", err)
+	}
+
+	sec, err := v.Get(ctx, key)
+	if err != nil {
+		return fmt.Errorf("failed to get secret: %w", err)
+	}
+
+	_ = sec
```
AndersonQ, line 243: [Suggestion]
```suggestion
	_, err = v.Get(ctx, key)
	if err != nil {
		return fmt.Errorf("failed to get secret: %w", err)
	}
```

internal/pkg/agent/vault/vault_test.go
```diff
@@ -164,3 +174,73 @@ func TestVault(t *testing.T) {
...
+			})
+		}
+		err := g.Wait()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func doCrud(t *testing.T, ctx context.Context, vaultPath, key string) error {
```
AndersonQ, line 210: [Suggestion]
I don't see `t` being used, so you could remove it.
```suggestion
func doCrud(ctx context.Context, vaultPath, key string) error {
```

AndersonQ (changes requested):

internal/pkg/agent/vault/vault_windows.go
```diff
@@ -59,42 +65,90 @@ func New(path string, opts ...OptionFunc) (v *Vault, err error) {
 		}
 	}
 
-	entropy, err := getOrCreateSeed(path, options.readonly)
+	r := &Vault{
```
AndersonQ, line 68: [NIT]
Why are you using `r` instead of `v`?

internal/pkg/agent/vault/vault_windows.go
```diff
@@ -59,42 +65,90 @@ func New(path string, opts ...OptionFunc) (v *Vault, err error) {
...
+	return err
+}
+
+// unlock unlocks the file lock and preserves the original error if there was a error
+func (v *Vault) unlock(err error) error {
+	unerr := v.lock.Unlock()
+	if err != nil {
+		return err
+	}
+	return unerr
```
AndersonQ, line 114: [Blocker]
Same as in the other one. Please return both errors, you can use `errors.Join` for that.

internal/pkg/agent/vault/vault_windows.go
AndersonQ, line 1: I'm wondering if it'd be worth having a common file for those vault methods that seems to be the same for windows and linux and a _darwin file with the darwin specific implementation

internal/pkg/agent/vault/vault_windows.go
aleksmaus: want to return v as nil if err happens after this line

internal/pkg/agent/vault/vault_linux.go
aleksmaus: done

internal/pkg/agent/vault/vault_windows.go
aleksmaus: done

pierrehilbert:
Buildkite is failing on main because we are unable to deploy 8.11.0 on Elastic Cloud for now. @AndersonQ is working on a workaround to unlock us in the meantime.

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -0,0 +1,214 @@
...
+		lockRetryDelay: options.lockRetryDelay,
+		lock:           flock.New(filepath.Join(path, lockFile)),
+	}
+
+	err = r.tryLock(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err = r.unlock(err)
```
AndersonQ, line 84: [Blocker]
It'll override the error of `getOrCreateSeed`. Better to `errors.Join` the errors.

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// Get retrieves the key from the vault store
+func (v *Vault) Get(ctx context.Context, key string) (dec []byte, err error) {
+	err = v.tryRLock(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err = v.unlock(err)
```
AndersonQ, line 120: [Blocker]
Same here, it'll override any other error

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to acquire shared lock: %v, err: %w", v.lock.Path(), err)
+	}
+	return nil
+}
+
+// unlock unlocks the file lock and preserves the original error if there was a error
+func (v *Vault) unlock(err error) error {
+	return errors.Join(err, v.lock.Unlock())
+}
```
AndersonQ, line 214: [Blocker]
I don't understand why an unlock function needs to receive an error. The caller is responsible to handle any previous error and then decide what to do if `unlock` fails as well.

Besides it harms the single responsibility principle, unlock is also responsible to aggregate an error for no obvious reason.

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -0,0 +1,214 @@
...
+	defer func() {
+		err = v.unlock(err)
+	}()
+
+	if _, err = os.Stat(v.filepathFromKey(key)); err == nil {
+		ok = true
+	} else if errors.Is(err, fs.ErrNotExist) {
+		err = nil
+	}
+	return ok, err
```
AndersonQ, line 146: [Suggestion]
Avoid `if else` for a clearer, more explicit code
 
```suggestion
	if _, err = os.Stat(v.filepathFromKey(key)); err != nil {
		if errors.Is(err, fs.ErrNotExist) {
			return false, nil
		}
		return false, err
	}
	
	return true, nil
```

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// Remove removes the key
+func (v *Vault) Remove(ctx context.Context, key string) (err error) {
+	err = v.tryLock(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = v.unlock(err)
```
AndersonQ, line 156: [Blocker]
same as the others

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, opt := range opts {
+		opt(&options)
+	}
+
+	return options
+}
+
+// fileNameFromKey returns the filename as a hash of the vault seed combined with the key
+// this ties the key with the vault seed eliminating the change of attempting
+// to decrypt the key for the wrong vault seed value.
```
AndersonQ, line 183:
```suggestion
// fileNameFromKey returns the filename as a hash of the vault seed combined with the key.
// This ties the key with the vault seed eliminating the chance of attempting
// to decrypt the key for the wrong vault seed value.
```

internal/pkg/agent/vault/vault_windows.go
```diff
@@ -152,3 +43,7 @@ func systemAdministratorsOnly(path string, inherit bool) error {
 		acl.GrantSid(0xF10F0000, systemSID), // full control of all acl's
 		acl.GrantSid(0xF10F0000, administratorsSID))
 }
+
+func writeFile(fp string, data []byte) (err error) {
+	return os.WriteFile(fp, data, 0600)
+}
```
AndersonQ, line 49: [Suggestion]
```suggestion
func writeFile(fp string, data []byte) error {
	return os.WriteFile(fp, data, 0600)
}
```

internal/pkg/agent/vault/vault_notdarwin.go
aleksmaus: do not agree with making it a blocker, it's a helper function to make the deferred call code above more compact and consistent, since it's done in multiple places

internal/pkg/agent/vault/vault_notdarwin.go
aleksmaus: explained above why this is done the way it's done

aleksmaus:
> just one small blocker, I marked it as `[Blocker]`

addressed the comments. do not agree with the blocker, added explanation why, it's a private helper function to avoid repeating the same code in 5 places. could you review again?

internal/pkg/agent/vault/vault_notdarwin.go
aleksmaus: it will not, it's a helper func using errors.Join

AndersonQ (approved):

cmacknz:
This is going to go into 8.10 first and then get forward ported to main. This will get us around the issue with the 8.11 snapshot.

AndersonQ:
Merged with failing integration tests because this PR fixes another integration tests issue affecting all tests installing the agent. Therefore we need to have it merged to reduce the number of failing tests and focus on one issue at a time.

PR #3241, opened 2023-08-15
Design discussion on operator PoC

michalpristas:
Draft POC operator

For agent to act as operator we need to set ENV Feature Flag `IS_OPERATOR`. 
Once this flag is set agent will register hooks for 2 CRDs and will setup custom manager, config manager and container runtime.

CRDs are for: policy see `pkg/operator/config/samples/observability_v1alpha1_elasticagentpolicy.yaml`
and for Component which are generated by agent and then processed in a separate reconciliation loop. 

Basic flow is like this
- On policy change reconciliation on Policy type is kicked off. 
- Configuration is extracted and sent to coordinator as with normal config runtime. 
- Coordinator will do its thing and call `Update` to update component model on a runtime manager.
- This runtime manager is our custom one. This one will create custom channel where new components are pushed and insert this channel into a watch channel (controller is watching this channel in a separate async loop)
- Manager groups unit based on k8s config. (Daemonset toghether, deployments together) by hashing k8s config
- manager then creates a runtime for each component, which is sent to a channel watched by controller loop.
- controller in a separate loop gets components and prepare service,account, role, binding, secret with config and pod vehicle (daemonset/deployment) for component to run.

To test this
- have cluster running (i'm using docker desktop one) make sure to have enough resources
- build docker images using normal `EXTERNAL=true SNAPSHOT=true DEV=true PLATFORMS=linux/amd64 PACKAGES=docker mage package`
- run `make install` this will install CRDs
- run `make deploy` this will deploy operator
- run `kubectl get pods --all-namespaces`, you should see operator there
- run `kubectl apply -f pkg/operator/config/samples/observability_v1alpha1_elasticagentpolicy.yaml` this will apply policy
- run `kubectl get pods --all-namespaces`, you should 2 more things deployed. 
- when you're done with it `make undeploy` will clean it

you can check logs, or if you modify `pkg/operator/config/samples/observability_v1alpha1_elasticagentpolicy.yaml` to point to proper ES you can check ES for consumed content (container logs are defined in a policy).
You can modify policy and see what happens.

 **What's missing**
This is still a POC with a lot of TODO comments in implementation so edge cases are missing. 
Detection of scheduled but not deployed resources (due to resource limits or crashes) is entirely missing.
In this implementation k8s provider is resolved on operator side, another idea is to let child components resolve this on their own. Currently it's not disabled on either side, just config coming to component does not contain any placeholders. I think perf wise it would make sense to keep it on operator level and disable on all components.

other grouping strategies can be implemented, e.g grouping by k8s config and output, some sharding logic etc


blakerouse:
I gave this a good look over, and I think the path this is taking is the correct one. I would have like to have seen the runtime manager being re-used, but I can see from the implementation that just replacing it with a different one does provide a clear and easier to understand path for being an operator of kuberenetes.

Being that the coordinator, vars manager, and config manager are able to be re-used really shows the strength of the V2 architecture.

A couple things I would like to see in this PR specifically:
- The `env.*` changes are a great cleanup, I think its best to just pull that out into a different PR and get that merged now.
- `Kubernetes map[string]interface{}` that was added to the component model to be a defined structure that way its self-describing and type checked.

I would also be interested in helping you get kubernetes testing integrated into the integration testing framework. Getting good coverage early would be great!

PR #3245, opened 2023-08-15
[Upgrade Watcher] Add logging to rollback restart step

ycombinator:
## What does this PR do?

This PR adds logging to the restart step of the Agent upgrade rollback process.

## Why is it important?

To better understand the progress of the upgrade rollback process.

## Checklist

1. Build an install Agent from `main` (currently at version `8.11.0`).
   ```
   EXTERNAL=true SNAPSHOT=true  DEV=true PLATFORMS=linux/arm64 PACKAGES=tar.gz mage package
   ```
3. Build Agent package from this PR (currently at version `8.12.0`).  Make sure NOT to use `SNAPSHOT=true` or `EXTERNAL=true`.
   ```
   ```
4. Upgrade the installed `8.11.0` Agent to the `8.12.0` Agent built with this PR.
   ```
   sudo elastic-agent upgrade 8.12.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
   ```
6. Tail the Upgrade Watcher logs and ensure that the log messages in this PR show up.
   ```
   tail -F /opt/Elastic/Agent/data/elastic-agent-b8da42/logs/elastic-agent-watcher-$(date +%Y%m%d).ndjson
   ```

### DIFF
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -45,7 +45,7 @@ func Rollback(ctx context.Context, log *logger.Logger, prevHash string, currentH
 
 	// Restart
 	log.Info("Restarting the agent after rollback")
-	if err := restartAgent(ctx); err != nil {
+	if err := restartAgent(ctx, log); err != nil {
 		return err
 	}
 
@@ -126,7 +126,7 @@ func InvokeWatcher(log *logger.Logger) error {
 	return cmd.Start()
 }
 
-func restartAgent(ctx context.Context) error {
+func restartAgent(ctx context.Context, log *logger.Logger) error {
 	restartFn := func(ctx context.Context) error {
 		c := client.New()
 		err := c.Connect(ctx)
@@ -148,12 +148,18 @@ func restartAgent(ctx context.Context) error {
 
 	for i := maxRestartCount; i >= 1; i-- {
 		backExp.Wait()
+
+		attempt := maxRestartCount - i + 1
+		log.Infof("Restarting Agent; attempt %d of %d", attempt, maxRestartCount)
+
 		err := restartFn(ctx)
 		if err == nil {
+			log.Warnf("Failed to restart agent: %s", err.Error())
 			break
 		}
 
 		if i == 1 {
+			log.Error("Failed to restart agent after final attempt")
 			return err
 		}
 	}
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -148,20 +148,20 @@ func restartAgent(ctx context.Context, log *logger.Logger) error {
 
 	for i := maxRestartCount; i >= 1; i-- {
 		backExp.Wait()
-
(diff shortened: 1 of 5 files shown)


version/version.go
```diff
@@ -4,5 +4,6 @@
 
 package version
 
-const defaultBeatVersion = "8.10.0"
+// FIXME: remove this — testing only!
+const defaultBeatVersion = "8.11.0"
```
ycombinator, line 8: Will remove this after PR is tested and approved.

cmacknz:
While you are in here, can you add calls to the `Sync` method of our logger before the watcher exits? I think the agent might be missing it as well since I don't see calls to it in the agent code directly.

version/version.go
```diff
@@ -4,5 +4,6 @@
 
 package version
 
-const defaultBeatVersion = "8.11.0"
+// FIXME: remove this — testing only!
+const defaultBeatVersion = "8.12.0"
```
ycombinator, line 8: Will remove this after PR is tested and approved.

internal/pkg/agent/cmd/run.go
```diff
@@ -148,6 +152,9 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
 		"source": agentName,
 	})
 
+	// Make sure to flush any buffered logs before we're done.
+	defer l.Sync() //nolint:errcheck // flushing buffered logs is best effort.
```
cmacknz, line 156: Wouldn't it be safer to do this on `baseLogger` in case that ends up being passed somewhere else?

cmacknz (approved):
LGTM assuming you remember to remove the test code you added :p

One nit about which logger gets Sync'd in the agent.

internal/pkg/agent/cmd/run.go
ycombinator: Done in 367dfbf41.

ycombinator:
buildkite test this

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -148,14 +148,20 @@ func restartAgent(ctx context.Context) error {
 
 	for i := maxRestartCount; i >= 1; i-- {
 		backExp.Wait()
+		attempt := maxRestartCount - i + 1
```
michalpristas, line 151: wouldnt it be easier and more optimal to switch for logic and condition on line 159?

michalpristas (approved):

internal/pkg/agent/application/upgrade/rollback.go
ycombinator: You mean to make the loop go from 0 to `maxRestartCount` instead of how it is now, going in the opposite direction?

ycombinator: @michalpristas I made a change to the logic in 5ec79d4ca. Is this what you were suggesting?

michalpristas: yeah, much nicer and no arithmetic operations on top of the loop

pchila (approved):

PR #3252, opened 2023-08-16
fix the GCP credentials variable name

sharbuz:
## What does this PR do?

Fix the GCP credentials variable name according to the doc: https://cloud.google.com/docs/authentication/application-default-credentials#GAC

## Why is it important?

The credentials aren't working properly in some cases, for example, when the pipeline has multiple GCP credentials. 

## Checklist

It was tested in the other elastic repo.

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -92,7 +92,7 @@ google_cloud_auth() {
 
     gcloud auth activate-service-account --key-file ${keyFile} 2> /dev/null
 
-    export GOOGLE_APPLICATIONS_CREDENTIALS=${secretFileLocation}
+    export GOOGLE_APPLICATION_CREDENTIALS=${secretFileLocation}
 }
 
 retry() {


jlind23 (approved):

.buildkite/scripts/common.sh
```diff
@@ -92,7 +92,7 @@ google_cloud_auth() {
 
     gcloud auth activate-service-account --key-file ${keyFile} 2> /dev/null
 
-    export GOOGLE_APPLICATIONS_CREDENTIALS=${secretFileLocation}
```
mrodm, line 95: is there any `unset` command that needs to be updated along with this ?

sharbuz: I'm sure yes, and something is already implemented here with the correct variable name.
Also, I'm not sure I have to change the logic of the pipeline. 
I just found the issue in my pipeline and suggested changing everywhere where it was copied because I spent some time to find and fix the issue.

sharbuz:
buildkite test this

.buildkite/scripts/common.sh
mrodm: Ok, it was just to double check it was not left behind some `unset` of this variable. In some repos, like elastic-package or fleet-server, those unset's are also in some trap functions or pre-exit hooks.

I just checked that there is just one usage of this env. var in this repo :+1: 
```shell
 $ git grep GOOGLE_APPLICATIONS
.buildkite/scripts/common.sh:    export GOOGLE_APPLICATIONS_CREDENTIALS=${secretFileLocation}
```

mrodm (approved):

sharbuz:
buildkite test this

pierrehilbert:
We have a problem @pazone is working on to fix for now. You won't be able to have buildkite green.

sharbuz:
buildkite test this

pazone (approved):

sharbuz:
buildkite test this

sharbuz:
buildkite test this

blakerouse:
buildkite test this

PR #3253, opened 2023-08-16
skip flaky upgrade_test

AndersonQ:
## What does this PR do?

Skip the flaky upgrade tests

## Why is it important?

The test is flaky

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/3154

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/upgrade_test.go
+++ b/testing/integration/upgrade_test.go
@@ -76,6 +76,7 @@ func TestFleetManagedUpgrade(t *testing.T) {
 		Isolate: false,
 		Sudo:    true, // requires Agent installation
 	})
+	t.Skip("Flaky test: https://github.com/elastic/elastic-agent/issues/3154")
 
 	ctx, cancel := context.WithCancel(context.TODO())
 	defer cancel()


testing/integration/upgrade_test.go
```diff
@@ -76,6 +76,7 @@ func TestFleetManagedUpgrade(t *testing.T) {
 		Isolate: false,
 		Sudo:    true, // requires Agent installation
 	})
+	t.Skip("Flaky test: https://github.com/elastic/elastic-agent/issues/3154")
```
cmacknz, line 79: We lose coverage of Fleet upgrades entirely by disabling this, which I don't like. I don't think we should be merging changes that could break this.

We can either leave it as flaky and hope changes get past it eventually, or we are getting close to having to prevent merges while we work through the test flakiness.

AndersonQ: ok, then let's merge the PRs fixing other tlaky tests that are blocked because of this flaky test?

AndersonQ:
The test is too important to skip, we'll deal with the flakiness without skipping the test

PR #3265, opened 2023-08-17
Small improvements on integration tests

AndersonQ:
- remove missing verb on printf-like call
 - remove folder as it might be left over from another test

## What does this PR do?

Fixes a possible cause of flakiness on tests and the missing verb on a printf-like call.

## Why is it important?

To avoid test flakiness

## Checklist

## Related issues

- N/A

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/tools/tools.go
+++ b/pkg/testing/tools/tools.go
@@ -81,14 +81,21 @@ func InstallAgentWithPolicy(t *testing.T, ctx context.Context, installOpts atest
 		agentFixture.SetUninstallToken(uninstallToken)
 	}
 
-	err = InstallAgentForPolicy(t, installOpts, agentFixture, kibClient, policy.ID)
+	err = InstallAgentForPolicy(t, ctx, installOpts, agentFixture, kibClient, policy.ID)
 	return policy, err
 }
 
-// InstallAgentForPolicy enrolls the given agent
-// fixture in Fleet using the default Fleet Server, waits for the agent to be
-// online, and returns error or nil.
-func InstallAgentForPolicy(t *testing.T, installOpts atesting.InstallOpts, agentFixture *atesting.Fixture, kibClient *kibana.Client, policyID string) error {
+// InstallAgentForPolicy enrolls the provided agent fixture in Fleet using the
+// default Fleet Server, waits for the agent to come online, and returns either
+// an error or nil.
+// If the context (ctx) has a deadline, it will wait for the agent to become
+// online until the deadline of the context, or if not, a default 5-minute
+// deadline will be applied.
+func InstallAgentForPolicy(t *testing.T, ctx context.Context,
+	installOpts atesting.InstallOpts,
+	agentFixture *atesting.Fixture,
+	kibClient *kibana.Client,
+	policyID string) error {
 	t.Helper()
 
 	// Create enrollment API key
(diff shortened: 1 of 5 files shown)


version/version.go
```diff
@@ -4,5 +4,5 @@
 
 package version
 
-const defaultBeatVersion = "8.11.0"
+const defaultBeatVersion = "8.10.0"
```
pierrehilbert, line 7: 8.11.0 is the version to have on main

pierrehilbert (changes requested):
Please let 8.11.0 as the branch version

AndersonQ:
> Please let 8.11.0 as the branch version

Right now the tartefacts aren't in the API, so the tests fail. I'm using the old version while I investigate/ensure the fix works. Before merge I'll restore that

pkg/testing/tools/tools.go
```diff
@@ -81,14 +81,19 @@ func InstallAgentWithPolicy(t *testing.T, ctx context.Context, installOpts atest
...
 	return policy, err
 }
 
-// InstallAgentForPolicy enrolls the given agent
-// fixture in Fleet using the default Fleet Server, waits for the agent to be
-// online, and returns error or nil.
-func InstallAgentForPolicy(t *testing.T, installOpts atesting.InstallOpts, agentFixture *atesting.Fixture, kibClient *kibana.Client, policyID string) error {
+// InstallAgentForPolicy enrolls the given agent fixture in Fleet using the
+// default Fleet Server, waits for the agent to be online, and returns error or nil.
+// If ctx has a deadline, it'll wait the agent to become online until 500ms
```
cmacknz, line 90: Why 500ms? What problem is this solving?

AndersonQ: just giving some slack so there is still some time after this function gives up waiting and the the context to be done.
I can remove if you want

version/version.go
AndersonQ: I know, but while the 8.11 artefacts aren't all published, the buildkite build will fail because it cannot download the agent's components. Therefore I changed it to 8.10 so I can have the tests running.

pkg/testing/tools/tools.go
blakerouse: Yeah I am not sure the point of it, should the default timeout of 5 minutes be enough, is 500ms really helping?

pkg/testing/tools/tools.go
```diff
@@ -121,11 +126,15 @@ func InstallAgentForPolicy(t *testing.T, installOpts atesting.InstallOpts, agent
 	}
 	t.Logf(">>> Ran Enroll. Output: %s", output)
 
+	timeout := 5 * time.Minute
+	if deadline, ok := ctx.Deadline(); ok {
+		timeout = deadline.Sub(time.Now().Add(-500 * time.Millisecond))
+	}
```
blakerouse, line 132: I like this idea, allowing the context to control the timeout. But why the adjustment?

testing/integration/install_test.go
```diff
@@ -53,8 +53,8 @@ func TestInstallWithoutBasePath(t *testing.T) {
 	}
 
 	topPath := filepath.Join(defaultBasePath, "Elastic", "Agent")
-	_, err = os.Stat(topPath)
-	require.True(t, os.IsNotExist(err))
+	err = os.RemoveAll(topPath)
+	require.NoError(t, err, "failed to remove %q. The test requires this path not to exist.")
```
blakerouse, line 57: Doesn't this change the tests? The test is checking that the uninstall did the correct thing. This basically ignores if it does and just removes itself. That seems to be masking a possible error.

blakerouse (changes requested):

testing/integration/install_test.go
AndersonQ: No, it's a pre-condition for the test as it's testing the agent successfully installs without the base path (`TestInstallWithoutBasePath`).
Being it a precondition, it'd not be fulfilled only if another test would let the base path there, therefore I don't see why the test preparation couldn't ensure it's absent.

pkg/testing/tools/tools.go
AndersonQ: i removed it

AndersonQ:
@Mergifyio update

pierrehilbert (approved):
LGTM, thx for having addressed every comments

PR #3166, opened 2023-08-01
[Upgrade Watcher][Crash Checker] Consider Agent process as crashed if its PID remains 0

ycombinator:
## What does this PR do?

This PR fixes a bug in the Upgrade Watcher's Crash Checker where it was consider the Agent process as healthy (not crashed) despite its PID remaining 0 every time the Crash Checker retrieved the PID from the service.

## Why is it important?

To detect when Agent has crashed so the Upgrade Watcher can initiate a rollback.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] ~I have made corresponding changes to the documentation~
- [ ] ~I have made corresponding change to the default configuration files~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] ~I have added an integration test or an E2E test~

## How to test this PR locally

_Manually testing this PR is currently blocked on https://github.com/elastic/elastic-agent/issues/3377._

Testing this PR locally is not trivial, but it's possible.  It involves upgrading to an Agent build where the Agent binary deliberately crashes, and making sure that the Agent is then rolled back to the previous (pre-upgrade) version.

1. Install and run Elastic Agent >= `8.10.0`. This is necessary because we need the pre-upgrade Agent to kick off the Upgrade Watcher using the post-upgrade Agent's binary, a change that was [implemented in `8.10.0`](https://github.com/elastic/elastic-agent/pull/3140).

2. Checkout this PR and make changes to the `elastic-agent run` code path such that the Agent process will exit with an error.  The easiest change would probably be to add a small sleep, say 5 seconds, followed by returning an error, right here: https://github.com/elastic/elastic-agent/blob/539a5f2956e423558b2d48fdb8bcdcf257348d51/internal/pkg/agent/cmd/run.go#L65-L66

3. Also bump up the Agent version to `8.12.0` over here so the upgrade is possible: https://github.com/elastic/elastic-agent/blob/243c76b3ebec56130bd6bd172f5d6725d289fbca/version/version.go#L7

4. Build the Agent package with the changes. Since the Agent version has been bumped up and the corresponding component binaries won't be available, make sure to set `AGENT_DROP_PATH` to nothing. Make sure NOT to use `SNAPSHOT=true` otherwise the snapshot artifact downloader will kick in during the upgrade process.
   ```
[...]

PR #3268, opened 2023-08-18
[Upgrade Watcher] Try restarting Agent in multiple ways during rollback

ycombinator:
## What does this PR do?

This PR fixes a potential bug where, during an upgrade, if the upgraded Agent process crashes, the Upgrade Watcher initiates a rollback but isn't able to restart the rolled back Agent.

Before this PR, the Upgrade Watcher would attempt to restart the rolled back Agent via a GRPC call to the Agent's `Restart()` function.  However, if the Agent process has crashed, there is no GRPC server available to receive such a call, thus preventing the Agent from being restarted. 

With this PR, the Upgrade Watcher continues to use the GRPC `Restart()` call to restart the Agent.  However, if that call fails, the Upgrade Watcher uses a fallback method of asking the Elastic Agent service to restart the Agent.

## Checklist

- [ ] ~I have added an integration test or an E2E test~ This bug was found while writing the integration test in https://github.com/elastic/elastic-agent/pull/3085.  As such, this bug fix will be tested by that test.

1. Build an Agent package from `main` (currently at version `8.11.0`), unpack it, and install it. This will serve as the version of Agent we will upgrade _from_.
   ```
   SNAPSHOT=true EXTERNAL=true DEV=true PLATFORMS=linux/arm64 PACKAGES=tar.gz mage package
   ```
3. Build an Agent package from the branch for this PR (say at version `8.12.0`). This will serve as the version of Agent we will upgrade _to_.  This Agent will deliberately crash upon start up but it also has the bugfix in the Upgrade Watcher code path.
   ```
   ```
5. Upgrade to the `8.12.0` Agent.
   ```
   sudo elastic-agent upgrade 8.11.0 --source-uri file://$(pwd)/build/distributions/ --skip-verify
   ```
6. Ensure that the Agent version is now `8.12.0`.  Because the Agent will be crashing, you may only see the binary version, not the daemon version.
   ```
   sudo elastic-agent version
   ```
7. Tail the Upgrade Watcher logs. Ensure that the Upgrade Watcher eventually initiates rollback.
   ```
   sudo tail -F /opt/Elastic/Agent/data/elastic-agent-06be89/logs/elastic-agent-watcher-$(date +%Y%m%d).ndjson
   ```

8. Ensure that the rolled back Agent gets restarted.
9. Ensure that the Agent version is now back to `8.11.0`. 
   ```
   sudo elastic-agent version
   ```

## Related issues

- Fixes bug explained in https://github.com/elastic/elastic-agent/pull/3166#issuecomment-1661275131.

### DIFF
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -145,10 +145,9 @@ func restartAgent(ctx context.Context, log *logger.Logger) error {
 
 	restartViaServiceFn := func(ctx context.Context) error {
 		topPath := paths.Top()
-		install.StopService(topPath)
-		err := install.StartService(topPath)
+		err := install.RestartService(topPath)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to restart agent via service: %w", err)
 		}
 
 		return nil
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -127,7 +127,7 @@ func InvokeWatcher(log *logger.Logger) error {
 }
 
 func restartAgent(ctx context.Context, log *logger.Logger) error {
-	restartFn := func(ctx context.Context) error {
+	restartViaDaemonFn := func(ctx context.Context) error {
 		c := client.New()
 		err := c.Connect(ctx)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/pkg/agent/cmd/run.go
```diff
@@ -211,6 +211,10 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
 		l.Error(errors.New(err, "failed to invoke rollback watcher"))
 	}
 
+	// FIXME: for testing only!
+	time.Sleep(10 * time.Second)
+	return errors.New("deliberately crashing agent (after invoking watcher)")
+
```
ycombinator, line 217: Will be removed once this PR is tested and approved.

version/version.go
```diff
@@ -4,5 +4,5 @@
 
 package version
 
-const defaultBeatVersion = "8.11.0"
+const defaultBeatVersion = "8.12.0"
```
ycombinator, line 7: Will be removed once this PR is tested and approved.

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -143,12 +143,32 @@ func restartAgent(ctx context.Context) error {
 		return nil
 	}
 
+	restartViaServiceFn := func(ctx context.Context) error {
+		topPath := paths.Top()
+		install.StopService(topPath)
+		err := install.StartService(topPath)
```
ycombinator, line 149: Would it better to create a new `install.RestartServive(topPath)` wrapper around these two calls?

michalpristas (approved):
lgtm, restarting by service should work in rollback cases without a need for check if service is there as for now we don't support upgrade scenario without agent being installed

blakerouse (approved):
Change looks good to me.

internal/pkg/agent/application/upgrade/rollback.go
ycombinator: Added `install.RestartService` in 6cb1746c7.  I'm taking care of adding logging to the restart step in #3245.

internal/pkg/agent/cmd/run.go
ycombinator: Removed in 2af8c30f1.

version/version.go
ycombinator: Removed in 2af8c30f1.

internal/pkg/agent/install/install.go
```diff
@@ -172,6 +172,16 @@ func StopService(topPath string) error {
...
 
+// RestartService stops and starts the installed service.
+func RestartService(topPath string) error {
+	err := StopService(topPath)
+	if err != nil {
+		return err
+	}
+
+	return StartService(topPath)
+}
```
cmacknz, line 183: Since we are dealing with unlikely problems already, if the agent were to stop for whatever reason between the `Stop` and `Start` calls it would be left stopped.

The most likely reason this could happen is a laptop system being powered off.

It would be much safer to use `systemctl restart` or equivalent if it exists because it should prevent this situation. The service manager knows you want the services end state to be running in this case, which is not true with an isolated call to stop.

ycombinator: Good point.  Fixed in d81fe7067.

cmacknz (approved):
Thanks!

pchila (approved):

pierrehilbert:
Unrelated CI failures. Force merge this one as it will bring some additional stability to our upgrade process.

Issue #1273, opened 2022-09-22
Elastic Agent relies on artifacts-api for fetching the dependent artifacts when running the packaging


v1v:
### What

As you have probably found, since yesterday the main branch is failing when running the packaging, since artifacts are not yet available for metricbeat, filebeat and so on.

<img width="1296" alt="image" src="https://user-images.githubusercontent.com/2871786/191785461-7976c63d-dec7-4d04-8a9f-9a373002f4db.png">

```
[2022-09-22T10:04:24.060Z] time="2022-09-22T10:04:24Z" level=warning msg="The Elastic artifacts API is not available yet" artifact=filebeat artifactName=filebeat-8.6.0-SNAPSHOT-linux-arm64.tar.gz elapsedTime=268.234246ms error="GET request failed with 404" retry=1 statusEndpoint="https://artifacts-api.elastic.co/v1/search/8.6.0-SNAPSHOT/filebeat?x-elastic-no-kpi=true" version=8.6.0-SNAPSHOT
```

This happens if no artifacts from the upstream projects have been pushed or the branch is not available in the artifacts-api.

This can be problematic since the merge commits are not moving forward in the CI but failing dramatically, hence the other stages in the CI workflow are not executed.

@mgreau pointed to the artifacts are available in https://artifacts-snapshot.elastic.co/beats/8.6.0-SNAPSHOT/downloads/beats/filebeat/filebeat-8.6.0-SNAPSHOT-linux-x86_64.tar.gz, hence there is no need to use the `artifacts-api`

### Questions

- Should the changes be done to the packaging to use `artifacts-snapshot.elastic.co` instead of `artifacts-api`?

cc @narph as you can probably add ore insights

cmacknz:
I am in favour of fixing CI by switching to `artifacts-snapshot.elastic.co` unless someone has a strong reason to fix and keep using `artifacts-api` for this use case. It's not clear what the tradeoffs are, or why we use `artifact-api` now.

v1v:
Just to be clear the error happens locally when running `mage package`, it is unrelated to the CI itself but the build system, the CI is another actor consuming the mage build system

cmacknz:
Ah I had thought this was a CI issue.

jlind23:
@v1v I looked at this issue and didn't see any mention of this artifacts api in the elastic-agent repository.
[...]

PR #3273, opened 2023-08-21
Update the e2e-testing dependency and unblock 8.11 buiilds

cmacknz:
- Depends on https://github.com/elastic/e2e-testing/pull/3622
- Closes https://github.com/elastic/elastic-agent/issues/1273

Update the e2e-testing dependency and allow the agent to build before an official snapshot is ready. For the build to succeed we only need the packaging job of each artifact we depend on to succeed once with the new version number.

The only major change here is that each artifact we download now needs to be linked back to its project name in the unified release manager. This is done with a simple map.

I also changed the download logic to attempt the downloads concurrently, which speeds them up slightly.


ycombinator (approved):
Tested with `SNAPSHOT=true EXTERNAL=true DEV=true PLATFORMS=linux/arm64 PACKAGES=tar.gz mage package` and it works with this PR!  The same command has been failing on `main` for the past few days.  

Also noticed the speedup with the concurrent downloading of artifacts, which is nice 👍.

cmacknz:
```
>> (linux-arm64-ubuntu-2204) Failed for instance linux-arm64-ubuntu-2204 (@ 34.135.199.225): ogc-linux-arm64-ubuntu-2204-89f5 unable to continue because stack never became ready: failed to create ESS cloud for version 8.11.0-SNAPSHOT: failed to create: (clusters.cluster_invalid_plan) Version [8.11.0-SNAPSHOT] is not available. Please use one of the available versions
```

I am going to force merge this to get us past the build errors. Will keep iterating on actually fixing the tests.

PR #3275, opened 2023-08-22
Use artifacts-snapshot instead of e2e-testing dependency

pazone:
- Enhancement

## What does this PR do?
Use artifacts-snapshot instead of the dependency from e2e-testing
## Why is it important?
Artifacts-api has some delay before we can get a new version of a recently published artifact. Artifacts-snapshot can be used to immediate artefacts retrievial especially when we bump the current version.

## Checklist

## Author's Checklist

- [ ]

`DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64 PACKAGES=tar.gz mage package`
## Related issues

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -38,12 +38,12 @@ import (
 	"github.com/elastic/elastic-agent/dev-tools/mage"
 	// mage:import
 	"github.com/elastic/elastic-agent/dev-tools/mage/target/common"
-
 	"github.com/elastic/elastic-agent/pkg/testing/define"
 	"github.com/elastic/elastic-agent/pkg/testing/ess"
 	"github.com/elastic/elastic-agent/pkg/testing/multipass"
 	"github.com/elastic/elastic-agent/pkg/testing/ogc"
 	"github.com/elastic/elastic-agent/pkg/testing/runner"
+	tools "github.com/elastic/elastic-agent/pkg/testing/tools"
 	bversion "github.com/elastic/elastic-agent/version"
 
 	// mage:import
@@ -51,7 +51,7 @@ import (
 	// mage:import
 	"github.com/elastic/elastic-agent/dev-tools/mage/target/test"
 
-	"github.com/sirupsen/logrus"
+	// "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 )
 
@@ -494,9 +494,6 @@ func FixDRADockerArtifacts() error {
 }
 
 func getPackageName(beat, version, pkg string) (string, string) {
-	if _, ok := os.LookupEnv(snapshotEnv); ok {
-		version += "-SNAPSHOT"
-	}
 	return version, fmt.Sprintf("%s-%s-%s", beat, version, pkg)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/testing/tools/artifacts_snapshot.go
```diff
@@ -0,0 +1,204 @@
...
+	manifestPackages, err := c.getManifestPackages(ctx, manResponse.ManifestURL)
+	if err != nil {
+		return fmt.Errorf("getting manifest packages: %w", err)
+	}
+	for _, pkg := range packageRequests {
+		pkgData, ok := manifestPackages[pkg.Name]
+		if !ok {
+			return fmt.Errorf("package %s not found in manifest", pkg)
+		}
+		err := c.downloadPackage(ctx, pkgData, pkg.TargetPath)
```
cmacknz, line 81: All the packages can be downloaded concurrently. There is now an example using errgroup in the magefile on main.

cmacknz:
👍 looking forward to removing the e2e-testing repository. There is a lot of code in there we don't actually need (even in the parts we do use).

pkg/testing/tools/artifacts_snapshot.go
```diff
@@ -0,0 +1,204 @@
...
+
+func downloadFile(url string, path string) error {
+	var filePath string
+	if path == "" {
+		tempParentDir := filepath.Join(os.TempDir(), uuid.NewString())
+		filePath = filepath.Join(tempParentDir, uuid.NewString())
+		path = filePath
+	} else {
+		filePath = filepath.Join(path, uuid.NewString())
+	}
```
AndersonQ, line 134: [Suggestion]
Simplify it by handling an empty path by itself.

```suggestion
	if path == "" {
		tempParentDir := filepath.Join(os.TempDir(), uuid.NewString())
		path = filepath.Join(tempParentDir, uuid.NewString())
	}
	
	filePath = filepath.Join(path, uuid.NewString())
```

pkg/testing/tools/artifacts_snapshot.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return fmt.Errorf("creating file: %w", err)
+	}
+	defer tempFile.Close()
+	exp := getExponentialBackOff(3)
+
+	retryCount := 1
+	var fileReader io.ReadCloser
+	download := func() error {
+		resp, err := http.Get(url)
```
AndersonQ, line 146: Why isn't it a method of `ArtifactSnapshotClient`? If it were, you could use the http client `ArtifactSnapshotClient` has as you did on `(aac *ArtifactSnapshotClient) createAndPerformRequest` or even use `createAndPerformRequest` directly.
It'd be better to use that HTTP client as well as having 1 single way of making HTTP requests, all using the same HTTP client

pkg/testing/tools/artifacts_snapshot.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+	defer fileReader.Close()
+
+	_, err = io.Copy(tempFile, fileReader)
+	if err != nil {
+		return fmt.Errorf("copying file: %w", err)
+	}
+
+	_ = os.Chmod(tempFile.Name(), 0666)
```
AndersonQ, line 167: [Blocker]
Handle the error. If this fails, it might lead to unexpected behaviour which will be hard to understand as we'd missing this error

pkg/testing/tools/artifacts_snapshot.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return nil
+}
+
+func getExponentialBackOff(elapsedTime time.Duration) *backoff.ExponentialBackOff {
```
AndersonQ, line 172: [Suggestion]
Rename `elapsedTime` to something like `timeout`, `timeLimit`, `maxDuration`

magefile.go
```diff
@@ -933,23 +930,29 @@ func packageAgent(platforms []string, packagingFn func()) {
 				"pf-host-agent",
 			}
 
-			ctx := context.Background()
+			packageRequests := []tools.PackageRequest{}
+			downloadVersion := packageVersion
+			if _, ok := os.LookupEnv(snapshotEnv); ok {
+				downloadVersion += "-SNAPSHOT"
+			}
```
AndersonQ, line 937: [Blocker]
Just having `SNAPSHOT` present does not mean it's true, it might as well be set to `false`. You can use `os.Getenv(snapshotEnv) == "true"` or even better, use `[strconv.ParseBool](https://pkg.go.dev/strconv#ParseBool):
```suggestion
			if ok, err := strconv.ParseBool(os.Getenv(snapshotEnv)); err==nil && ok {
				downloadVersion += "-SNAPSHOT"
			}
```

perhaps too much in one line, but you get the idea

magefile.go
```diff
@@ -933,23 +930,29 @@ func packageAgent(platforms []string, packagingFn func()) {
...
-					}
+					_, packageName := getPackageName(binary, downloadVersion, reqPackage)
+					packageRequests = append(packageRequests, tools.PackageRequest{Name: packageName, TargetPath: targetPath})
 				}
 			}
+
+			ctx := context.Background()
+
+			artifactSnapshotClient := tools.NewArtifactSnapshotClient()
+			err = artifactSnapshotClient.DownloadPackages(ctx, packageRequests, downloadVersion)
```
AndersonQ, line 951: [NIT]
as the context is being created just to satisfy `artifactSnapshotClient.DownloadPackages` it seems to be a good place to use `context.TODO()`
```suggestion
			artifactSnapshotClient := tools.NewArtifactSnapshotClient()
			err = artifactSnapshotClient.DownloadPackages(context.TODO(), packageRequests, downloadVersion)
```

AndersonQ (changes requested):

PR #3276, opened 2023-08-22
use 8.10 instead of 8.11 for stack version and .package-version

AndersonQ:
## What does this PR do?

use 8.10 instead of 8.11 for stack version and .package-version

## Why is it important?

8.11 stack isn't available yet on QA/Staging

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -19,7 +20,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/pkg/version"
+	agentVersion "github.com/elastic/elastic-agent/version"
 
 	"github.com/otiai10/copy"
 	"gopkg.in/yaml.v2"
@@ -127,6 +132,8 @@ func NewFixture(t *testing.T, version string, opts ...FixtureOpt) (*Fixture, err
 	for _, o := range opts {
 		o(f)
 	}
+
+	decreasePackageVersionFor8_11(t, f)
 	return f, nil
 }
 
@@ -915,4 +922,44 @@ type AgentInspectOutput struct {
 	Signed struct {
 		Data string `yaml:"data"`
 	} `yaml:"signed"`
+}
+
+func decreasePackageVersionFor8_11(t *testing.T, f *Fixture) {
+	installFS := os.DirFS(f.WorkDir())
+	var matches []string
+
+	err := fs.WalkDir(installFS, ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.Name() == agentVersion.PackageVersionFileName {
+			matches = append(matches, path)
+		}
+		return nil
+	})
+	require.NoError(t, err)
+
+	t.Logf("package version files found: %v", matches)
+
+	for _, m := range matches {
+		versionFile := filepath.Join(f.WorkDir(), m)
+		fmt.Printf("removing package version file %q\n", versionFile)
+
+		pv, err := version.ParseVersion(f.version)
+		if err != nil {
+			require.NoError(t, err, "could not parse fixture version")
+		}
+		if pv.Major() == 8 && pv.Minor() > 11 {
+			prev, err := pv.GetPreviousMinor()
+			if err != nil {
+				require.NoError(t, err, "8.11 cannot be used right now, "+
+					"failed getting previous minor")
+			}
+
+			err = os.WriteFile(versionFile, []byte(prev.String()), 0666)
+			require.NoError(t, err, "could not write package-version file %q",
+				versionFile)
+		}
+	}
 }
--- a/magefile.go
+++ b/magefile.go
@@ -52,6 +52,8 @@ import (
 )
 
 const (
+	VersionAgentStack = "AGENT_STACK_VERSION"
+
 	goLintRepo        = "golang.org/x/lint/golint"
 	goLicenserRepo    = "github.com/elastic/go-licenser"
 	buildDir          = "build"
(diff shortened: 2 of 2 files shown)


pkg/testing/fixture.go
```diff
@@ -916,3 +923,43 @@ type AgentInspectOutput struct {
...
+
+	for _, m := range matches {
+		versionFile := filepath.Join(f.WorkDir(), m)
+		fmt.Printf("removing package version file %q\n", versionFile)
+
+		pv, err := version.ParseVersion(f.version)
+		if err != nil {
+			require.NoError(t, err, "could not parse fixture version")
+		}
+		if pv.Major() == 8 && pv.Minor() > 11 {
```
cmacknz, line 953:
```suggestion
		if pv.Major() == 8 && pv.Minor() == 11 {
```

Shouldn't this be an equals?

cmacknz (commented):
We are going to need to make these changes again in the next feature freeze period. Can the code be refactored out into a dedicated package so both pieces live together? Maybe call the package featurefreeze or something to make the purpose obvious.

pkg/testing/fixture.go
AndersonQ: yes, I fixed here already

pierrehilbert:
If you can just fix the linter things, it will be ready to be merged.

cmacknz:
I ran this locally to try a single test: `TEST_PLATFORMS=linux/amd64 SNAPSHOT=true INSTANCE="multipass" mage integration:single TestInstallAndCLIUninstallWithEndpointSecurity` 

It started an 8.10 stack but the .package-version edit doesn't seem to be taking effect because Fleet is rejecting the version of the agent:
```
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): Error: fail to enroll: fail to execute request to fleet-server: status code: 400, fleet-server returned an error: UnsupportedVersion, message: version is not supported
```

cmacknz:
I branched from this PR and created one with the fix to get the tests to run https://github.com/elastic/elastic-agent/pull/3277

cmacknz:
You need https://github.com/elastic/elastic-agent/pull/3281 to get the build to succeed reliably.

cmacknz:
@Mergifyio update

.buildkite/scripts/steps/integration_tests.sh
```diff
@@ -4,11 +4,21 @@ set -euxo pipefail
 source .buildkite/scripts/common.sh
 
 # PACKAGE
-DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64 PACKAGES=tar.gz mage package
+AGENT_PACKAGE_VERSION=8.10.0 \
```
cmacknz, line 7:
```suggestion
AGENT_PACKAGE_VERSION=8.10.0-SNAPSHOT \
```

This will fix the failing upgrade test at least.

cmacknz:
FYI you had the agent package version set to a non-snapshot version causing some of these test failures. Not sure if we needed anything else for it to pass:

```diff
-AGENT_PACKAGE_VERSION=8.10.0 \
+AGENT_PACKAGE_VERSION=8.10.0-SNAPSHOT \
```

cmacknz:
I don't know if we need https://github.com/elastic/elastic-agent/pull/3276/commits/49c077021465a90bc69c3043a73a244a7e79e2ac

cmacknz:
The actual bug with the package version test is that we are getting 8.10.0 as the agent version from this block of code instead of 8.10.0-SNAPSHOT

https://github.com/elastic/elastic-agent/blob/e2a3a254604bfb4902e21a8ee2495af7ce1e4ce9/testing/integration/pkgversion_common_test.go#L60-L71

cmacknz:
This is because we need both `AGENT_PACKAGE_VERSION=8.10.0-SNAPSHOT` and `SNAPSHOT=true` when running `mage package` but this gives us the final `.tar.gz` with the `-SNAPSHOT-SNAPSHOT` prefix from both places 🤦

cmacknz:
Just using BEAT_VERSION=8.10.0 and SNAPSHOT=true will fix this, with the consequence that we use the 8.10.0 Beats.

cmacknz:
Apparently there is even more to this problem, the `elastic-agent version --yaml` output indicates whether a version is a snapshot or not with a separate boolean flag that the test logic doesn't parse:

```yaml
❯ sudo elastic-agent version --yaml
binary:
  version: 8.10.0
  commit: e2a3a254604bfb4902e21a8ee2495af7ce1e4ce9
  build_time: 2023-08-24T00:32:56Z
  snapshot: true
daemon:
  version: 8.10.0
  commit: e2a3a254604bfb4902e21a8ee2495af7ce1e4ce9
  build_time: 2023-08-24T00:32:56Z
  snapshot: true
```

We ignore the `snapshot` parameter here causing us to try to download a version that doesn't exist:

https://github.com/elastic/elastic-agent/blob/1071b42739e023c989e8b05d1b35d5582f02dc12/testing/integration/pkgversion_common_test.go#L100-L107

I am confused about how this ever passed on the 8.10 branch because the 8.10 version doesn't exist either:

```
failed download of agent binary: unable to download package: 3 errors occurred:
        	            		* package '/opt/Elastic/Agent/data/elastic-agent-1071b4/downloads/elastic-agent-8.10.0-linux-x86_64.tar.gz' not found: open /opt/Elastic/Agent/data/elastic-agent-1071b4/downloads/elastic-agent-8.10.0-linux-x86_64.tar.gz: no such file or directory
        	            		* call to 'https://snapshots.elastic.co/8.10.0-ffe0de5c/downloads/beats/elastic-agent/elastic-agent-8.10.0-linux-x86_64.tar.gz' returned unsuccessful status code: 404
        	            		* call to 'https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.10.0-linux-x86_64.tar.gz' returned unsuccessful status code: 404
        	            		*
```

cmacknz:
I am going to create a separate PR with https://github.com/elastic/elastic-agent/pull/3276/commits/ad73102bdfb64538ba472ab3527f971a263a9ccb to fix the bug in the GetPrevMinor function, this will also let me test out the 8.11 snapshot on main.

We still have work to do to make this work in the next minor, but now that there is an 8.11 snapshot we can leave this PR alone and focus on fixing the tests on main.

cmacknz:
https://github.com/elastic/elastic-agent/pull/3289

AndersonQ:
closing as it isn't needed right now, of we need it again, we can reopen it.

PR #3277, opened 2023-08-22
Test against 8.10 on main with the 8.11 agent by editing the package version

cmacknz:
The same as https://github.com/elastic/elastic-agent/pull/3276/files but with the package version file edit moved to happen in the right place.

Also fixes a bug in the GetPrevMinor() method of our version parser package. We need to keep this change.

I hope we get an 8.11 snapshot soon and we don't need this, but if we don't we'll merge it and then revert once the 8.11 snapshot is ready and work on less ugly solution for the next time this happens.

### DIFF
--- a/pkg/version/version_parser_test.go
+++ b/pkg/version/version_parser_test.go
@@ -360,3 +360,54 @@ func TestLess(t *testing.T) {
 		})
 	}
 }
+
+func TestPreviousMinor(t *testing.T) {
+	testcases := []struct {
+		name             string
+		version          string
+		prevMinorVersion string
+	}{
+		{
+			name:             "basic release version",
+			version:          "8.7.0",
+			prevMinorVersion: "8.6.0",
+		},
+		{
+			name:             "snapshot release version",
+			version:          "8.9.3-SNAPSHOT",
+			prevMinorVersion: "8.8.0-SNAPSHOT",
+		},
+		{
+			name:             "emergency release version",
+			version:          "8.9.0-er1",
+			prevMinorVersion: "8.8.0-er1",
+		},
+		{
+			name:             "previous major version",
+			version:          "8.0.0",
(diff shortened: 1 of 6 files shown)


pkg/version/version_parser_test.go
```diff
@@ -360,3 +360,54 @@ func TestLess(t *testing.T) {
 		})
 	}
 }
+
+func TestPreviousMinor(t *testing.T) {
```
cmacknz, line 364: The `GetPrevMinor` method was discarding the prerelease and build metadata information.

cmacknz:
Looks like we need to fix the upgrade tests to respect the test stack version:

```
    fixture.go:433: >> running agent with: [/opt/Elastic/Agent/elastic-agent upgrade 8.11.0-SNAPSHOT --source-uri file:///home/ubuntu/agent/.agent-testing/local --skip-verify]
    upgrade_test.go:469: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_test.go:469
        	            				/home/ubuntu/agent/testing/integration/upgrade_test.go:215
        	            	exit status 1
        	Test:       	TestStandaloneUpgrade/Upgrade_8.10.0_to_8.11.0-SNAPSHOT
        	Messages:   	error triggering agent upgrade to version "8.11.0-SNAPSHOT", output:
```

This just keeps getting uglier...

AndersonQ:
@cmacknz with @pchila's help I took another approach and just needed to set `AGENT_PACKAGE_VERSION`, `AGENT_STACK_VERSION` and `AGENT_VERSION` when building the agent and running the tests on buildkite to have it working. I've also merged you fix on the `version` package on my PR https://github.com/elastic/elastic-agent/pull/3276/. 

I think we can close this one and finish the review and any last change there.

cmacknz:
Yes that solution is much nicer, closing this one.

PR #3289, opened 2023-08-24
Preserve prerelease and metadata in GetPrevMinor.

cmacknz:
The GetPrevMinor function would discard build metadata, in particular it will drop the SNAPSHOT prerelease signifier which can cause us to upgrade to non-existent or incorrect versions.

Found when attempting to get main to run against the previous stack release: https://github.com/elastic/elastic-agent/pull/3276

### DIFF
--- a/pkg/version/version_parser.go
+++ b/pkg/version/version_parser.go
@@ -113,8 +113,9 @@ func (psv ParsedSemVer) GetPreviousMinor() (*ParsedSemVer, error) {
 
 	if minor > 0 {
 		// We have at least one previous minor version in the current
-		// major version series
-		return NewParsedSemVer(major, minor-1, 0, "", ""), nil
+		// major version series. Set the patch to zero to guarnatee the
+		// version exists, the number of patch releases varies.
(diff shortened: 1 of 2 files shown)


pkg/component/runtime/command_test.go
```diff
@@ -31,6 +32,9 @@ func TestAddToBucket(t *testing.T) {
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
+			if runtime.GOOS == "windows" {
```
cmacknz, line 35: https://github.com/elastic/elastic-agent/issues/3290

We've tried to fix these on Windows a few times already, and since there is nothing Windows specific to this test let's just skip it completely on that OS rather than waste more timing trying to account for the variation between CI machines.

AndersonQ (approved):

cmacknz:
@Mergifyio update

cmacknz:
@Merigfyio update

cmacknz:
@Mergifyio update

cmacknz:
No new test failures, merging.

PR #3293, opened 2023-08-25
skip downgrade test when only one snapshot version is available

AndersonQ:
Commit message:
>**Tile:** 
skip downgrade test when only one snapshot version is available (#3293)
**Message:**
TestStandaloneUpgradeToSpecificSnapshotBuild actually test a downgrade to a previous SNAPSHOT. This fix renames it to TestStandaloneDowngradeToPreviousSnapshotBuild. It also skips the test if there is only one SNAPSHOT available as, in that case, there is no 'previous' SNAPSHOT to downgrade to.

## What does this PR do?

It fixes TestStandaloneUpgradeToSpecificSnapshotBuild to work is there is only one snapshot build available.

## Why is it important?

When there is only one snapshot version for the current version the test fails.

## Checklist

```
INSTANCE_PROVISIONER="multipass" \
SNAPSHOT=true \
TEST_PLATFORMS="linux/amd64" \
mage integration:single TestStandaloneUpgradeToSpecificSnapshotBuild
```

## Related issues

- N/A

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/tools/artifacts_api.go
+++ b/pkg/testing/tools/artifacts_api.go
@@ -23,7 +23,7 @@ const (
 	artifactsAPIV1VersionsEndpoint      = "v1/versions/"
 	artifactsAPIV1VersionBuildsEndpoint = "v1/versions/%s/builds/"
 	artifactAPIV1BuildDetailsEndpoint   = "v1/versions/%s/builds/%s"
-	//artifactAPIV1SearchVersionPackage = "v1/search/%s/%s"
+	// artifactAPIV1SearchVersionPackage = "v1/search/%s/%s"
 )
 
 var (
@@ -275,9 +275,10 @@ func GetLatestSnapshotVersion(ctx context.Context, log logger, aac *ArtifactAPIC
 		return nil, ErrSnapshotVersionsEmpty
 	}
 
-	// normally the output of the versions returned by artifact API is already sorted in ascending order,
-	// if we want to sort in descending order we could use
-	sort.Sort(sort.Reverse(sortedParsedVersions))
+	// normally the output of the versions returned by artifact API is already
+	// sorted in ascending order.If we want to sort in descending order we can
+	// use sort.Reverse.
+	sort.Sort(sortedParsedVersions)
 
 	var latestSnapshotVersion *version.ParsedSemVer
 	// fetch the latest SNAPSHOT build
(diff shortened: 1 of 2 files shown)


belimawr (approved):

pkg/testing/tools/artifacts_api.go
```diff
@@ -275,8 +275,9 @@ func GetLatestSnapshotVersion(ctx context.Context, log logger, aac *ArtifactAPIC
 		return nil, ErrSnapshotVersionsEmpty
 	}
 
-	// normally the output of the versions returned by artifact API is already sorted in ascending order,
-	// if we want to sort in descending order we could use
+	// normally the output of the versions returned by artifact API is already
+	// sorted in ascending order.If we want to sort in descending order we need
+	// to pass a sort.Reverse to sort.Sort.
```
pchila, line 280: After discussing the meaning of this comment maybe we can clarify the wording a bit like
```suggestion
	// normally the output of the versions returned by artifact API is already
	// sorted in ascending order. Since we want to extract the most recent version,
	// we sort in descending order by passing a sort.Reverse operator to sort.Sort.
```

pierrehilbert:
I force merge this one as it helps us to move in a direction with less tests failing.

testing/integration/upgrade_test.go
```diff
@@ -261,7 +261,7 @@ func TestStandaloneUpgradeWithGPGFallback(t *testing.T) {
 	testStandaloneUpgrade(ctx, t, agentFixture, parsedVersion, toVersion, "", false, false, true, customPGP)
 }
 
-func TestStandaloneUpgradeToSpecificSnapshotBuild(t *testing.T) {
+func TestStandaloneDowngradeToPreviousSnapshotBuild(t *testing.T) {
```
pchila, line 264: I would keep the previous test name as it's the feature name. we may add a comment for clarity

```suggestion
// Here we test the ability of elastic agent to upgrade to a given snapshot build (not necessarily the latest)
// by passing a version `x.y.z-SNAPSHOT+<buildid>`. In order to test this, we look for a build for the current
// snapshot version other than the latest, downgrade to such version and verify that we get the correct
// git commit hash for the specified build (to detect that we didn't install the latest snapshot build by mistake)
func TestStandaloneDowngradeToSpecificSnapshotBuild(t *testing.T) {
```

pchila (approved):
A couple of suggestion but overall it looks good

Issue #3294, opened 2023-08-25
Incomplete diagnostic when ran from Kibana


pierrehilbert:
From 8.10.0, when diagnostic collection is initialized from Kibana, lot of files within `components/` directory are missing.
When the diagnostic is ran locally, everything is correctly collected in the package.

pierrehilbert:
I created a followup issue to get E2E for this: https://github.com/elastic/elastic-agent/issues/3301

PR #3295, opened 2023-08-25
Fix components in diag collection from Kibana

pierrehilbert:
## What does this PR do?

This PR is adding components in the diagnostic collection when initialized from Kibana.

## Why is it important?

To get components when diag are collected from Kibana.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3294

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
@@ -20,6 +20,7 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/diagnostics"
 	"github.com/elastic/elastic-agent/internal/pkg/fleetapi"
 	"github.com/elastic/elastic-agent/internal/pkg/fleetapi/acker"
+	"github.com/elastic/elastic-agent/pkg/component"
 
 	"golang.org/x/time/rate"
 )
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
```diff
@@ -241,8 +243,43 @@ func (h *Diagnostics) diagUnits(ctx context.Context) []client.DiagnosticUnitResu
...
 
+// diagUnits gathers diagnostics from components.
+func (h *Diagnostics) diagComponents(ctx context.Context) []client.DiagnosticComponentResult {
+	cDiag := make([]client.DiagnosticComponentResult, 0)
+	h.log.Debug("Performing component diagnostics")
+	startTime := time.Now()
+	defer func() {
+		h.log.Debugf("Component diagnostics complete. Took: %s", time.Since(startTime))
+	}()
+	rr := h.diagProvider.PerformDiagnostics(ctx)
```
pchila, line 254: With this addition we are performing the diagnostics on the components/units twice (one here and one in the diagUnits).

For both performance and consistency reason it would be better to call `PerformDiagnostics` outside of the diagComponents/diagUnits and pass the result as a parameter to both functions

pierrehilbert: Oh you're right, will fix that!

PR #3310, opened 2023-08-29
[crossbuild] Fix path separators

ycombinator:
## What does this PR do?

This PR fixes the path separators in the `crossBuild` target, so if this target is invoked on a Windows host, it creates the correct build command to be run inside a Linux-based Docker container.

## Why is it important?

So we can build Elastic Agent on Windows hosts.

Without this PR, e.g. from `main`, we get the following error while building an Elastic Agent package on a Windows host:

```
$ EXTERNAL=true SNAPSHOT=true DEV=true PLATFORMS=windows/amd64 mage package 
...
sh: 1: buildmage-linux-amd64: not found
...
```

### DIFF
--- a/dev-tools/mage/crossbuild.go
+++ b/dev-tools/mage/crossbuild.go
@@ -271,6 +271,13 @@ func (b GolangCrossBuilder) Build() error {
 		return errors.Wrap(err, "failed to determine mage-linux-"+builderArch+" relative path")
 	}
 
+	// If we're running this mage target on Windows, the path separator used in
+	// buildCmd will be `\`. But the buildCmd is executed inside a Linux Docker
+	// container. So we replace all `\` path separators with `/`.
+	if GOOS == "windows" {
+		buildCmd = strings.ReplaceAll(buildCmd, "\\", "/")
+	}
+
 	dockerRun := sh.RunCmd("docker", "run")
 	image, err := b.ImageSelector(b.Platform)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


AndersonQ (approved):

cmacknz (approved):

ycombinator:
CI failures are unrelated to the changes in this PR.  @pierrehilbert @cmacknz would you mind force merging this PR?

dev-tools/mage/crossbuild.go
```diff
@@ -271,6 +271,13 @@ func (b GolangCrossBuilder) Build() error {
 		return errors.Wrap(err, "failed to determine mage-linux-"+builderArch+" relative path")
 	}
 
+	// If we're running this mage target on Windows, the path separator used in
+	// buildCmd will be `\`. But the buildCmd is executed inside a Linux Docker
+	// container. So we replace all `\` path separators with `/`.
+	if GOOS == "windows" {
+		buildCmd = strings.ReplaceAll(buildCmd, "\\", "/")
```
leehinman, line 278: would you consider using  [filepath.ToSlash](https://pkg.go.dev/path/filepath#ToSlash) ?   I think if we do that we don't have to have the `GOOS == "windows"` check.  If the `os.PathSeparator` is `/` it won't do anything, otherwise it will replace all instances of `os.PathSeparator` with `/`.

pierrehilbert:
I let you reply to @leehinman's question and then I will force merge things.

dev-tools/mage/crossbuild.go
ycombinator: I like it, thanks for the suggestion @leehinman. Changed in 39daec4db.

ycombinator: I like it, thanks for the suggestion @leehinman. Changed in 39daec4db.

pierrehilbert:
@leehinman could you review this again please?

leehinman (approved):
LGTM

pierrehilbert:
I'm force merging as the failures are not related

PR #3317, opened 2023-08-29
Properly check if and error matches a windows/syscall error

### LINKED ISSUE #3316: Flaky test: TestRemovePath, error is not considered retriable
## Flaky Test

* **Test Name:** internal/pkg/agent/install TestRemovePath

* **Link:** [uninstall_windows_test.go](https://github.com/elastic/elastic-agent/blob/0f3229248d82145a49005204ebff2d8a1cb864fc/internal/pkg/agent/install/uninstall_windows_test.go#L31)

* **Branch:** 
 - main
 - others

* **Artifact Link:** 
  - [Jenkins](https://fleet-ci.elastic.co/blue/organizations/jenkins/elastic-agent%2Felastic-agent-mbp/detail/main/1298/pipeline/2133/)

* **Notes:** 
 - the issue seems to be on how the error is considered retriable 
 - this failure happened on Windows 8

### Stack Trace

```
=== RUN   TestRemovePath
    uninstall_windows_test.go:54: 
        	Error Trace:	C:/Users/jenkins/workspace/tic-agent_elastic-agent-mbp_main/src/github.com/elastic/elastic-agent/internal/pkg/agent/install/uninstall_windows_test.go:54
        	            	could not remove "C:\\Users\\jenkins\\AppData\\Local\\Temp\\TestRemovePath2573796789\\001\\subdir", unretriable error: failed to open handle for "C:\\Users\\jenkins\\AppData\\Local\\Temp\\TestRemovePath2573796789\\001\\subdir\\main.exe": The process cannot access the file because it is being used by another process.
        	Test:       	TestRemovePath
--- FAIL: TestRemovePath (0.62s)
```

AndersonQ:
## What does this PR do?

Properly check if and error matches a windows/syscall error

## Why is it important?

RemovePath is failing as it considers the error not retriable when it actually is.

## Checklist

## Related issues

- Closes #3316 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/install/uninstall_windows.go
+++ b/internal/pkg/agent/install/uninstall_windows.go
@@ -16,26 +16,31 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-func isBlockingOnExe(err error) bool {
+func isAccessDeniedError(err error) bool {
 	if err == nil {
 		return false
 	}
-	path, errno := getPathFromError(err)
-	if path == "" {
-		return false
+
+	var errno syscall.Errno
+	if errors.As(err, &errno) {
+		return errno == syscall.ERROR_ACCESS_DENIED
 	}
-	return errno == syscall.ERROR_ACCESS_DENIED
+
+	return false
 }
 
 func isRetryableError(err error) bool {
 	if err == nil {
 		return false
 	}
-	path, errno := getPathFromError(err)
-	if path == "" {
-		return false
+
+	var errno syscall.Errno
+	if errors.As(err, &errno) {
+		return errno == syscall.ERROR_ACCESS_DENIED ||
+			errno == windows.ERROR_SHARING_VIOLATION
 	}
-	return errno == syscall.ERROR_ACCESS_DENIED || errno == windows.ERROR_SHARING_VIOLATION
+
+	return false
 }
 
 func removeBlockingExe(blockingErr error) error {
@@ -72,14 +77,20 @@ func removeBlockingExe(blockingErr error) error {
 	return nil
 }
 
-func getPathFromError(blockingErr error) (string, syscall.Errno) {
+func getPathFromError(err error) (string, syscall.Errno) {
 	var perr *fs.PathError
-	if errors.As(blockingErr, &perr) {
+	if errors.As(err, &perr) {
 		var errno syscall.Errno
 		if errors.As(perr.Err, &errno) {
 			return perr.Path, errno
 		}
 	}
+
+	var errno syscall.Errno
+	if errors.As(err, &errno) {
+		return "", errno
+	}
+
 	return "", 0
 }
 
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/install/uninstall_windows.go
```diff
@@ -20,22 +20,27 @@ func isBlockingOnExe(err error) bool {
 	if err == nil {
 		return false
 	}
-	path, errno := getPathFromError(err)
```
cmacknz, line 23: Where did this call go?

AndersonQ: away - uahuahuahua (sorry I could not resist the joke)

Why it went away:
I don't see any evidence the fact the error does not have a path on it prevents it from being one of the retriable error. Besides the path isn't used at all here, so I don't see the relevance of this check. `err == syscall.ERROR_ACCESS_DENIED` seems to be the only thing relevant to determine if the error `isBlockingOnExe`.

cmacknz: Looking at the way this gets used, the error needs to have a path for the special case for removing a blocking exe to work. The call is unnecessary in the overall logic chain since it gets called again later, but if the error doesn't have a path the work around for it can't work.

cmacknz: I wonder if SHARING_VIOLATION was just forgotten here.

AndersonQ: > I wonder if SHARING_VIOLATION was just forgotten here.
me too.

However the error in the test doe not come from `isBlockingOnExe`, but from `isRetryableError`. So that might be only checking for `syscall.ERROR_ACCESS_DENIED` intentionally.

>Looking at the way this gets used, the error needs to have a path for the special case for removing a blocking exe to work. The call is unnecessary in the overall logic chain since it gets called again later, but if the error doesn't have a path the work around for it can't work.

yes, but it's another case and there it still checks for the path not being empty.

cmacknz: Reading the original PR https://github.com/elastic/elastic-agent/pull/3120 the implemetation was adapted from https://github.com/golang/go/blob/master/src/testing/testing.go#L1257 which does the same.

> yes, but it's another case and there it still checks for the path not being empty.

If we get rid of the path check here, this function is no longer answering the question "is this an error from blocking waiting on an exe" it is just "is this an access denied error" which is mislead. Just because the code happens to be written today so that this doesn't matter, doesn't mean that will stay true in the future.

You could rename the function to something like `possiblyBlockingOnExe` to indicate that the error here might not have anything to do with an executable file. There are a lots of other reasons to get an access is denied error.

AndersonQ: Well either way it does not answer _"is this an error from blocking waiting on an exe"_ as it did not do any check on the path, not even if it was from an exe file.
Also the Go implementation does not check the path. I'm happy to rename it, but unless there is a proper check on the path, I don't believe to be worth checking it

cmacknz (approved):
Thanks, at least the function name matches what it actually checks now.

cmacknz:
No new test failures, merging.

PR #3338, opened 2023-08-30
Show installation progress

ycombinator:
## What does this PR do?

This PR enhances `elastic-agent install` to show users the progress of the installation.

### Before this PR

https://github.com/elastic/elastic-agent/assets/51061/6f47764d-971f-4222-be20-c3bf8f3325ca

### After this PR

https://github.com/elastic/elastic-agent/assets/51061/8eb08be5-d664-4d7e-b1fd-2857af5c6151

## Why is it important?

So users know installation is proceeding and not stalled.

## Checklist

### DIFF
--- a/internal/pkg/agent/install/progress.go
+++ b/internal/pkg/agent/install/progress.go
@@ -10,42 +10,54 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"golang.org/x/exp/rand"
 )
 
 type ProgressTracker struct {
-	writer               io.Writer
-	progressTickDuration time.Duration
-	stepInProgress       bool
-	mu                   sync.RWMutex
-	stop                 chan struct{}
+	writer io.Writer
+
+	tickInterval          time.Duration
+	randomizeTickInterval bool
+
+	stepInProgress bool
+	mu             sync.RWMutex
+	stop           chan struct{}
 }
 
 func NewProgressTracker(writer io.Writer) *ProgressTracker {
 	return &ProgressTracker{
-		writer:               writer,
-		progressTickDuration: 100 * time.Millisecond,
-		stop:                 make(chan struct{}),
+		writer:                writer,
+		tickInterval:          100 * time.Millisecond,
+		randomizeTickInterval: true,
+		stop:                  make(chan struct{}),
 	}
 }
 
-func (pt *ProgressTracker) SetProgressTickDuration(d time.Duration) {
-	pt.progressTickDuration = d
+func (pt *ProgressTracker) SetTickInterval(d time.Duration) {
+	pt.tickInterval = d
+}
+
+func (pt *ProgressTracker) DisableRandomizedTickIntervals() {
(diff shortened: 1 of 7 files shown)


pierrehilbert:
We discussed this on Monday, thanks I'm sure it will help users to see that something is happening there!

internal/pkg/agent/install/progress.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (pt *ProgressTracker) calculateTickInterval() time.Duration {
+	if !pt.randomizeTickInterval {
+		return pt.tickInterval
+	}
+
+	// Randomize interval between 65% and 250% of configured interval
+	// to make it look like the progress is non-linear. :)
+	floor := int64(math.Floor(float64(pt.tickInterval.Milliseconds()) * 0.65))
+	ceiling := int64(math.Floor(float64(pt.tickInterval.Milliseconds()) * 2.5))
```
blakerouse, line 103: Haha, I don't know if this is really needed but okay! Why not.

blakerouse (approved):
Looks good. Simple design and easy to follow.

internal/pkg/agent/install/progress.go
ycombinator: This whole deal is just to show some random progress to make things appear more realistic ;)

ycombinator:
buildkite test this

internal/pkg/agent/install/progress_test.go
```diff
@@ -0,0 +1,115 @@
...
+		pt := NewProgressTracker(w)
+
+		pt.Start()
+
+		pt.StepStart("step 1 starting")
+		pt.StepFailed()
+
+		pt.Stop()
+
+		require.Equal(t, "step 1 starting... FAILED\n", string(w.buf))
```
AndersonQ, line 41: [Blocker]
Isn't there a chance there will be more than 3 dots(`...`) when the assertion happens?
You didn't disable the randomised ticker intervals.

AndersonQ (commented):
A wee blocker, isn't the 1st test susceptible to flakiness? 
Another thing, it won't work well with nested steps, right?

AndersonQ (changes requested):
Please check the previous comment

internal/pkg/agent/install/progress_test.go
ycombinator: We shouldn't see more than three dots because the step starts (line 36) and immediately fails (line 37). Unlike some of the other tests in this PR, there is no delay (sleep) in between those two lines, so more dots will not be shown.

ycombinator:
> Another thing, it won't work well with nested steps, right?

Correct, if/when we have such steps or want to represent such steps in the progress, we'll need to enhance the `ProgressTracker` with a stack or something.

AndersonQ (approved):

pierrehilbert:
Unrelated failures, I'm merging

PR #3359, opened 2023-09-06
Structure the BK pipeline output for failures

### LINKED ISSUE #3347: [Buildkite] Improve Buildkite output logs in case of failure
As of today, if an integration test fails the overall step with output `+exit 1` without letting you know what the problem is.
You then have to expand all groups and start looking for failure.
We have two options:
- We should split this https://github.com/elastic/elastic-agent/blob/main/.buildkite/scripts/steps/integration_tests.sh into smaller scripts
- @alexsapran said that BK has the ability to automatically structure your logging output https://buildkite.com/docs/pipelines/managing-log-output, this should probably be explored too.

@rdner is the one who highlighted that issue.

alexsapran:
## What does this PR do?

This PR is going to make it easier to read the output of the Buildkite output on the integration tests. 

Example output with failed unit and integration tests
Screenshot 2023-09-12 at 5 21 13 PM

## Why is it important?

This is important because it would allow engineers to spot the failure quicker when integration tests are failing.

## Checklist

This is a change intended to be used for Buildkite.

## Related issues

fixes #3347


belimawr (approved):

AndersonQ (approved):

alexsapran:
This PR is failing on the windows, not sure how to proceed with this.
I would like to avoid merging when there are CI-failures, especially when the change is not related to the product.

PR #3361, opened 2023-09-06
Don't create a stack on a single run that doesn't require a stack

blakerouse:
## What does this PR do?

When running `mage integration:single TestFakeComponent` the integration testing runner would create a stack even though the test did not require a stack. 

This is because the `pkg/testing/define` merged batches of tests that didn't require a stack and those that didn't as they can be ran in the same batch. This is fine until then only a single test is selected out of that batch and the test doesn't need a stack.

## Why is it important?

Makes it faster to run the integration tests for a single test in the case that it doesn't need a stack.

## Checklist

Run `mage integration:single TestFakeComponent` and notice that a stack is no longer provisioned.

### DIFF
--- a/pkg/testing/define/batch.go
+++ b/pkg/testing/define/batch.go
@@ -64,7 +64,15 @@ type BatchPackageTests struct {
 	// Name is the package name.
 	Name string `json:"name"`
 	// Tests is the set of tests in the package.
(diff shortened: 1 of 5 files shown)


ycombinator (approved):
Nice improvement.

blakerouse:
buildkite test this

PR #3370, opened 2023-09-07
Fix integration tests by waiting for the watcher to finish during upgrade tests

cmacknz:
This is a work around while we implement a fix for https://github.com/elastic/elastic-agent/issues/3371.

If the watcher is still running at the end of a test, it will observe the normal uninstall+reinstall as a crash and roll back the agent at the start of the next generally causing chaos and failing tests.

### DIFF
--- a/testing/integration/upgrade_test.go
+++ b/testing/integration/upgrade_test.go
@@ -9,7 +9,6 @@ package integration
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/fs"
 	"net/http"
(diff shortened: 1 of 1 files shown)


testing/integration/upgrade_test.go
```diff
@@ -556,28 +574,17 @@ func checkLegacyAgentHealthAndVersion(t *testing.T, ctx context.Context, f *ates
...
-		t.Logf("Version %q is too old for a quick update marker check, skipping...", fromVersion)
-		return
+		t.Logf("Version %q is too old for a quick update marker check", fromVersion)
+		timeout = defaultWatcherDuration
 	}
 
-	t.Log("Waiting for upgrade watcher to finish running...")
-
-	updateMarkerFile := filepath.Join(agentFixture.WorkDir(), "data", ".update-marker")
-	require.FileExists(t, updateMarkerFile)
```
cmacknz, line 573: I tried polling for this file to stop existing, but it never exists at the end of the test. I am actually still a bit confused about why the watcher is still running without this (maybe the path here was wrong?), but waiting for exactly the configured watcher timeout fixes all the tests.

My priority now is to get the tests back into a good state.

ycombinator (approved):

blakerouse (approved):
Good to see this get the tests back to working state. Sadly makes them much longer, especially being that they are just sitting and waiting. Hopefully killing the watcher on uninstall can improve this some.

AndersonQ (approved):

Issue #3372, opened 2023-09-07
occasional uninstall hang on Windows 8.8.1


leehinman:
Windows elastic-agent version 8.8.1

occasionally running `elastic-agent uninstall` will hang indefinitely or report error "service did not respond in a timely fashion".  Running `elastic-agent restart` will break the hang and allow the uninstall to complete.  Also stopping the service prior to running uninstall will allow the uninstall to complete without hanging.

likely related to https://github.com/elastic/elastic-agent/issues/3061

cmacknz:
I bet we aren't handling stop events properly looking at some of the related issues, particularly https://github.com/elastic/elastic-agent/issues/3307

cmacknz:
The root cause has been identified and fixed: https://github.com/elastic/elastic-agent-libs/pull/155

PR #3373, opened 2023-09-07
improve reporting of errors during uninstall

leehinman:
## What does this PR do?

Report errors during uninstall process, rather than ignoring and continuing.

## Why is it important?

Current code will ignore and not log errors that occur during uninstall.  This makes it very hard to debug.

## Checklist

~~- [ ] I have added an integration test or an E2E test~~ current uninstall tests should be sufficient

1. install elastic-agent
2. run `elastic-agent uninstall`

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/3372

### DIFF
--- a/internal/pkg/agent/install/uninstall.go
+++ b/internal/pkg/agent/install/uninstall.go
@@ -30,8 +30,20 @@ import (
 )
 
 // Uninstall uninstalls persistently Elastic Agent on the system.
-func Uninstall(cfgFile, topPath, uninstallToken string) error {
-	// uninstall the current service
+func Uninstall(cfgFile, topPath, uninstallToken string, force bool) error {
+	var log *logp.Logger
+	var err error
+
+	if force {
+		log, err = logger.NewWithLogpLevel("uninstall", logp.DebugLevel, false)
+
+	} else {
+		log, err = logger.NewWithLogpLevel("uninstall", logp.ErrorLevel, false)
+	}
+	if err != nil {
+		return fmt.Errorf("error setting logger: %w", err)
+	}
+
 	svc, err := newService(topPath)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/install/uninstall.go
```diff
@@ -36,7 +36,14 @@ func Uninstall(cfgFile, topPath, uninstallToken string) error {
 	if err != nil {
 		return err
 	}
-	status, _ := svc.Status()
+
+	status, err := svc.Status()
+	if err != nil {
```
cmacknz, line 41: This converts this into a fatal error for the install process? Is this actually necessary?

We have a few open issues where the agent doesn't respond to service manager events properly, so I'm not sure this is the right choice.

Can we log this without aborting the uninstall process?

cmacknz: Ah we won't attempt to stop the service if we can't detect that it is running, so not stopping an unresponsive service probably isn't correct either.

leehinman: what if we make the force flag an argument to `Uninstall`?  That way we can optionally ignore the error messages (but print to stderr) and try to continue.  But it most cases, if Status() returns an error, than Stop() is going to error as well.

As a side note.  Stop() can actually return without an error but the service isn't in the Stopped state.
https://github.com/kardianos/service/blob/9832e01049dd49bb66dd8fc447c72eefba7fc2cd/service_windows.go#L499-L525

If the service doesn't report status.Stopped within the `WaitToKillServiceTimeout` time (default 5 sec), then Stop just returns with a nil error.

cmacknz: Force flag works for me, would like to keep a way to do a best effort uninstall given the amount of bugs in Windows service for the agent right now.

cmacknz: We are probably also mixing the kardianos handling with the service handlers from Beats https://github.com/elastic/elastic-agent-libs/blob/main/service/service_windows.go and I don't think that is helping anything.

internal/pkg/agent/install/uninstall.go
```diff
@@ -30,13 +30,35 @@ import (
...
 	svc, err := newService(topPath)
 	if err != nil {
 		return err
 	}
-	status, _ := svc.Status()
+
+	status, err := svc.Status()
+	if err != nil {
+		log.Debugf("failed to get status for service (%s): %s", paths.ServiceName, err)
+		if !force {
```
cmacknz, line 55: Now that I've seen this, I don't love it. Seems like it makes the code more convoluted without necessarily fixing a real problem, not sure the extra complexity is worth the "just in case" insurance.

PR #3376, opened 2023-09-07
Code cleanup for command checkins

faec:
No behavior changes, just cleanup and adding some documentation comments related to checkin handling.

## Checklist


cmacknz (approved):

pkg/component/runtime/state.go
```diff
@@ -131,15 +134,6 @@ func (s *ComponentState) Copy() (c ComponentState) {
 	return c
 }
 
-func (s *ComponentState) syncComponent(comp *component.Component) bool {
```
faec, line 134: Moving this inline since it only had one caller and its return value was ignored

pkg/component/runtime/state.go
```diff
@@ -151,34 +145,34 @@ func (s *ComponentState) syncExpected(comp *component.Component) bool {
 		}
 
 		touched[key] = true
-		existing, ok := s.expectedUnits[key]
+		expected, ok := s.expectedUnits[key]
```
faec, line 148: Renaming because "existing" is also used in this file for the current units (rather than the expected units) and this makes it clearer that we're comparing to the expected config.

michalpristas (approved):

pierrehilbert:
buildkite test this

cmacknz:
Force merging, ignoring the maintainability warning from SonarQube.

Issue #3362, opened 2023-09-06
Set version in headers requests to Fleet endpoints


criamico:
With the merge of https://github.com/elastic/kibana/pull/163570, Kibana is now enforcing version headers in all requests. 

When a fleet server gets enrolled in Fleet, enrollment fails with this error message:

```
  Requesting service_token from Kibana.
  Error: request to get security token from Kibana failed: Please specify a version via elastic-api-version header. Available versions: [2023-10-31]
```

This can be resolved by adding 
```
  elastic-api-version: '2023-10-31'
```

to the request headers. It should be added to any requests to public Fleet endpoints.

jlind23:
@pierrehilbert @pierrehilbert @fearful-symmetry transferring this to you. Worth fixing it as soon as possible otherwise it will probably block part of your serverless testing.

pchila:
Isn't this a duplicate of elastic/ingest-dev#2253 ?

jlind23:
@pchila this is somehow similar but here for Fleet API (inside kibana) and not fleet-server

fearful-symmetry:
@jlind23 is someone else doing serverless support for agent? Right now all of my testing and such has been for beats.

jlind23:
@fearful-symmetry We have this issue that is left unassigned for now and this [one](https://github.com/elastic/ingest-dev/issues/2253) on which @pchila is working on.
cc @pierrehilbert

joshdover:
Do we need to do this at all? We're not supporting enrolling Fleet Server agents against Serverless deployments and this API versioning on Kibana will only be enforced on Serverless, I believe.

@criamico For Fleet development, we will need to start using standalone Fleet Server when working with Serverless ES and Kibana. We may need some new dev docs for this.

PR #3383, opened 2023-09-08
 Implement basic version negotiation with Fleet Server

pchila:
## What does this PR do?

We add an `Elastic-Api-Version` header to all requests to Fleet. 
We already have support for detecting and logging warnings and downgrade requests (those are not yet implemented though)

The `Warning` header coming from Fleet is logged at fleet client level and propagated to coordinator via the specific error type `WarningError`. When the coordinator receives a `WarningError` will set fleet status to `Degraded` (as opposed to `Failed` that was used for any error up to this point) and set fleet status message to the warning it received.

## Why is it important?

We aim for Elastic Agent to support connecting to multiple versions of Fleet server supporting multiple api versions.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/remote/client.go
+++ b/internal/pkg/remote/client.go
@@ -220,11 +220,28 @@ func (c *Client) Send(
 			c.log.With("error", err).Debugf(msg)
 			continue
 		}
+		c.checkApiVersionHeaders(reqID, resp)
 
 		return resp, nil
 	}
 
 	return nil, fmt.Errorf("all hosts failed: %w", multiErr)
+}
+
+func (c *Client) checkApiVersionHeaders(reqID string, resp *http.Response) {
+	const elasticApiVersionHeaderKey = "Elastic-Api-Version"
+	const warningHeaderKey = "Warning"
+
+	warning := resp.Header.Get(warningHeaderKey)
+	if warning != "" {
+		c.log.With("http.request.id", reqID).Warnf("warning in fleet response: %q", warning)
+	}
+
+	if downgradeVersion := resp.Header.Get(elasticApiVersionHeaderKey); resp.StatusCode == http.StatusBadRequest && downgradeVersion != "" {
+		// fleet server requested a downgrade to a different api version, we should bubble up an error until some kind
+		// of fallback mechanism can instantiate the requested version. This is not yet implemented so we log an error
+		c.log.With("http.request.id", reqID).Errorf("fleet requested a different api version %q but this is currently not implemented: %q", downgradeVersion)
+	}
 }
 
 // URI returns the remote URI.
--- a/.mockery.yaml
+++ b/.mockery.yaml
@@ -3,7 +3,3 @@ testonly: False
 with-expecter: True
 keeptree: True
 case: underscore
-note: |
-  Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
-  // or more contributor license agreements. Licensed under the Elastic License;
-  // you may not use this file except in compliance with the Elastic License.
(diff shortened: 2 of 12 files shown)


joshdover:
@pchila Ideally we could backport this to the next 8.10.x patch. Any concerns with that?

pchila:
@joshdover It shouldn't be too much of a risk since we only add a header, a couple of logs and an additional state to the fleet status in case of warning: nothing that should impact the core functionality of elastic-agent.
Just for completeness: should we backport elastic/elastic-agent#3394 as well ?

joshdover:
> Just for completeness: should we backport #3394 as well ?

See my comment here: https://github.com/elastic/elastic-agent/issues/3362#issuecomment-1711810672

I actually think we might want to revert this as this scenario is not supported (users deploying their own Fleet Server against Serverless projects)

internal/pkg/remote/client.go
```diff
@@ -220,13 +220,30 @@ func (c *Client) Send(
...
 		}
+		c.checkApiVersionHeaders(reqID, resp)
 
 		return resp, nil
 	}
 
 	return nil, fmt.Errorf("all hosts failed: %w", multiErr)
 }
 
+func (c *Client) checkApiVersionHeaders(reqID string, resp *http.Response) {
```
ycombinator, line 231: Would it be possible to write a unit test for this method with cases for when the warning header is set and not set?

internal/pkg/remote/client.go
```diff
@@ -220,13 +220,30 @@ func (c *Client) Send(
...
+	const elasticApiVersionHeaderKey = "Elastic-Api-Version"
```
ycombinator, line 232: Could you use the package-level `elasticApiVersionHeaderKey` const instead of redefining it in local scope here?

internal/pkg/remote/client.go
```diff
@@ -220,13 +220,30 @@ func (c *Client) Send(
...
+	const warningHeaderKey = "Warning"
+
+	warning := resp.Header.Get(warningHeaderKey)
+	if warning != "" {
+		c.log.With("http.request.id", reqID).Warnf("warning in fleet response: %q", warning)
+	}
+
+	if downgradeVersion := resp.Header.Get(elasticApiVersionHeaderKey); resp.StatusCode == http.StatusBadRequest && downgradeVersion != "" {
```
ycombinator, line 240: Is there any relationship between the warning header and the API version header? Is it possible to have one set but not the other? In the context of version negotiation, is there some specific value of the warning header that should make Agent look at the value of the API version header?

internal/pkg/remote/client.go
pchila: This is an internal implementation detail that is subject to change due to refactoring. I added unit tests for the client public function `Send` about the api version and the warning headers above in the `TestElasticApiVersion` where we assert the behavior of the client with a test http server :thinking:

internal/pkg/remote/client.go
pchila: According to the related issue:
- we have `Elastic-Api-Version` ina  response with an http status code `400` when we are using an api version that is too new
- we get a warning if the Api version is too old but fleet server will attempt to serve the request anyways (it won't immediately answer with a `Bad Request` code

I am not sure that we have overlap (like if we get a warning we could also get a significant value of `Elastic-Api-Version` header) but maybe @joshdover  can answer that ;)

ycombinator: Reading through the related issue, it doesn't look like there's any strong connection between the content of the warning header and the API version header. So I think the logic here is good as-is.

internal/pkg/remote/client.go
ycombinator: Ah yes, my bad.  I thought this file here was in the `client` package. Never mind :)

ycombinator (approved):
LGTM.

Issue #2706, opened 2023-05-17
Quick multiple upgrades while agent is still in grace period break agent installation


pchila:
Please post all questions and issues concerning the Elastic Agent on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For security vulnerabilities please only send reports to security@elastic.co.
See https://www.elastic.co/community/security for more information.

Please include configurations and logs if available.

For confirmed bugs, please report:
- Version: 8.9.0-SNAPSHOT
- Operating System: Linux
- Discuss Forum URL:
- Steps to Reproduce: 
  - Upgrade/downgrade agent using CLI
  - After the upgrade initiate immediately (or before 10 minutes of grace period have elapsed) another upgrade using the CLI
  - in my specific case I performed 8.9.0-SNAPSHOT -> 8.8.0-SNAPSHOT -> 8.9.0-SNAPSHOT and I ended up with the following (the agent `0af676` that was running had been cleaned up by the upgrade process)
  ```
  root@pchila-elastic:/opt/Elastic/Agent# ll
  total 1016
  drwxr-x--- 4 root root   4096 May 17 12:25 ./
  drwxr-xr-x 3 root root   4096 May 17 12:13 ../
  -rw-r----- 1 root root     41 May 17 12:13 .build_hash.txt
  drwxr-x--- 4 root root   4096 May 17 12:26 data/
  lrwxrwxrwx 1 root root     58 May 17 12:25 elastic-agent -> /opt/Elastic/Agent/data/elastic-agent-0af676/elastic-agent
  -rw-r----- 1 root root      6 May 17 12:25 .elastic-agent.active.commit
  -rw-r----- 1 root root  10026 May 17 12:13 elastic-agent.reference.yml
  -rw------- 1 root root  10486 May 17 12:13 elastic-agent.yml
  -rw------- 1 root root    274 May 17 12:13 fleet.enc
  -rw------- 1 root root      0 May 17 12:13 fleet.enc.lock
  -rw-r--r-- 1 root root      0 May 17 12:13 .installed
  -rw-r----- 1 root root  13675 May 17 12:13 LICENSE.txt
  -rw-r----- 1 root root 964376 May 17 12:13 NOTICE.txt
  -rw-r----- 1 root root    309 May 17 12:13 README.md
  drwxr-x--- 2 root root   4096 May 17 12:13 vault/
[...]

PR #3384, opened 2023-09-08
Uninstall finds and kills any running `elastic-agent watch` process

### LINKED ISSUE #3371: Uninstall does not stop a running watcher after upgrade
In the case that the Elastic Agent was recently upgraded (aka. inside the 10 minute watcher grace period) running `elastic-agent uninstall` does not stop the running watcher.

This causes a host of problems:

- Uninstall on Windows completely fails
  - This is because the watcher is writing to its log file and Windows prevents removal of the log directory because the watcher is still running and writing to that log file.
- Sequence uninstall/re-install within the grace period causes corrupted Elastic Agent
  - The Elastic Agent will uninstall correctly, but if Elastic Agent is re-installed within the 10 minute grace period the watcher will then notice that the Elastic Agent is not health and result in a rollback to be performed. This rollback results in a corrupted Elastic Agent installation.

- Version: All
- Operating System: All

blakerouse:
## What does this PR do?

Finds any running `elastic-agent watch` process on the host and kills the running watcher.

Adds console output for uninstall so its clear to see what uninstall is performing.

## Why is it important?

Fixes the following problems:

- A orphan watcher process is not remaining when the Elastic Agent has been uninstalled.
- Uninstall on Windows will no longer fail if an upgrade has occurred within the past 10 minutes.
- A sequence of uninstall/re-install within the 10 minute will not cause a corrupted Elastic Agent.
- Adds ability to track what uninstall is performing.

## Checklist

- [x] I have added an integration test or an E2E test (tested by integration tests already)

1. Install Elastic Agent 8.10.0-SNAPSHOT
2. Upgrade to a build with the PR, this will get the watcher running.
3. Run `sudo elastic-agent uninstall` to get it to fully uninstall.

## Related issues

- Closes #3371 

## Logs

After Upgrade:

```
% ps aux | grep elastic-agent
root             76886   0.1  0.1 409760672  47904   ??  Ss   11:23AM   0:01.33 elastic-agent
root             76927   0.1  0.3 409548672 100480   ??  S    11:23AM   0:00.28 /Library/Elastic/Agent/data/elastic-agent-693ca9/components/metricbeat -E setup.ilm.enabled=false -E setup.template.enabled=false -E management.enabled=true -E management.restart_on_output_change=true -E logging.level=info -E logging.to_stderr=true -E gc_percent=${METRICBEAT_GOGC:100} -E metricbeat.config.modules.enabled=false -E http.enabled=true -E http.host=unix:///Library/Elastic/Agent/data/tmp/iThI_df0cBKC6YUNGGlKscMkOfz3FBH3.sock -E path.data=/Library/Elastic/Agent/data/elastic-agent-693ca9/run/system/metrics-default
blake            76955   0.0  0.0 408626944   1232 s003  R+   11:24AM   0:00.00 grep elastic-agent
root             76930   0.0  0.3 409531104  98992   ??  S    11:23AM   0:00.21 /Library/Elastic/Agent/data/elastic-agent-693ca9/components/metricbeat -E setup.ilm.enabled=false -E setup.template.enabled=false -E management.enabled=true -E management.restart_on_output_change=true -E logging.level=info -E logging.to_stderr=true -E gc_percent=${METRICBEAT_GOGC:100} -E metricbeat.config.modules.enabled=false -E http.enabled=true -E http.host=unix:///Library/Elastic/Agent/data/tmp/akSPbdqgaHaTY0_J01-dsfYK6JpMz2zn.sock -E path.data=/Library/Elastic/Agent/data/elastic-agent-693ca9/run/http/metrics-monitoring
root             76929   0.0  0.3 409525744  96864   ??  S    11:23AM   0:00.21 /Library/Elastic/Agent/data/elastic-agent-693ca9/components/metricbeat -E setup.ilm.enabled=false -E setup.template.enabled=false -E management.enabled=true -E management.restart_on_output_change=true -E logging.level=info -E logging.to_stderr=true -E gc_percent=${METRICBEAT_GOGC:100} -E metricbeat.config.modules.enabled=false -E http.enabled=true -E http.host=unix:///Library/Elastic/Agent/data/tmp/Hk6rvk9TDibMPcDvpl0jkLE-qDsHWVYL.sock -E path.data=/Library/Elastic/Agent/data/elastic-agent-693ca9/run/beat/metrics-monitoring
root             76928   0.0  0.3 409520032  95904   ??  S    11:23AM   0:00.26 /Library/Elastic/Agent/data/elastic-agent-693ca9/components/filebeat -E setup.ilm.enabled=false -E setup.template.enabled=false -E management.enabled=true -E management.restart_on_output_change=true -E logging.level=info -E logging.to_stderr=true -E gc_percent=${FILEBEAT_GOGC:100} -E filebeat.config.modules.enabled=false -E http.enabled=true -E http.host=unix:///Library/Elastic/Agent/data/tmp/xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock -E path.data=/Library/Elastic/Agent/data/elastic-agent-693ca9/run/filestream-monitoring
root             76924   0.0  0.1 409355680  32224   ??  Ss   11:23AM   0:00.06 /Library/Elastic/Agent/elastic-agent watch --path.config /Library/Elastic/Agent --path.home /Library/Elastic/Agent
```

After Uninstall:

```
% ps aux | grep elastic-agent
blake            77071   0.0  0.0 408626944   1280 s003  S+   11:24AM   0:00.00 grep elastic-agent
```

Uninstall output:

```
Elastic Agent will be uninstalled from your system at /Library/Elastic/Agent. Do you want to continue? [Y/n]:Y
Stopping service... DONE
Stopping upgrade watcher; none found... DONE
Removing service... DONE
Removing install directory.... DONE
```


cmacknz:
You need to revert https://github.com/elastic/elastic-agent/commit/94764be90ce4a207f2e380390cfb67558095eb9b as part of this to see whether it actually fixes the tests.

internal/pkg/agent/install/uninstall.go
```diff
@@ -41,23 +44,28 @@ func Uninstall(cfgFile, topPath, uninstallToken string) error {
...
 				err,
 				fmt.Sprintf("failed to stop service (%s)", paths.ServiceName),
-				errors.M("service", paths.ServiceName))
+				aerrors.M("service", paths.ServiceName))
 		}
 	}
 
+	// kill any running watcher
+	if err := killWatcher(); err != nil {
+		return fmt.Errorf("failed trying to kill any running watcher: %w", err)
```
cmacknz, line 56: Does this need to be a fatal error? Maybe this makes sense behind a `--force` flag as discussed in https://github.com/elastic/elastic-agent/pull/3373#discussion_r1319119781

internal/pkg/agent/install/uninstall.go
```diff
@@ -310,3 +318,41 @@ func applyDynamics(ctx context.Context, log *logger.Logger, cfg *config.Config)
...
+			continue
+		}
+		if filepath.Base(state.Args[0]) == paths.BinaryName && state.Args[1] == "watch" {
+			// it is the watch subprocess
+			proc, err := os.FindProcess(pid)
+			if err != nil {
+				errs = errors.Join(errs, fmt.Errorf("failed to load watcher process with pid %d: %w", pid, err))
+				continue
+			}
+			err = proc.Kill()
```
cmacknz, line 350: > // Kill causes the Process to exit immediately. Kill does not wait until
> // the Process has actually exited. This only kills the Process itself,
> // not any other processes it may have started.
> func (p *Process) Kill() error {

The docs on this say it doesn't actually wait for the process to exit, which suggests there is a small race condition here with any following attempt to install depending on how long this actually takes. On Unix it is probably almost instant, not sure how long it takes on Windows.

blakerouse: Yeah I am testing this on Windows with my integration runner.

I am still seeing `Error: failed to remove installation directory (C:\\Program Files\\Elastic\\Agent): could not remove path, timeout exeeded after 69 tries during 7s. Last error: remove C:\\Program Files\\Elastic\\Agent\\data\\elastic-agent-162ea9\\logs\\elastic-agent-watcher-20230908-1.ndjson: The process cannot access the file because it is being used by another process.\nFor help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.11/fleet-troubleshooting.html\n`.

Could be that the `Kill` is not instant on Windows, but I also read that waiting for a process that is not your child doesn't always work. Maybe I should add a loop in this function to keep calling `Kill` on all found processes until they are all gone.

internal/pkg/agent/install/uninstall.go
blakerouse: This step is right after the service is stopped, but not removed. Being that it is early I prefer that we ensure that killing the watcher works, before going any futher. This is not leaving the Elastic Agent in a broken state, the service is just off.

We could maybe add skip the error in the case that `--force` is added, but don't want to conflate this PR with an addition of `--force`.

internal/pkg/agent/install/uninstall.go
cmacknz: Yeah I think you will have to poll waiting for the process to exit.

pchila (commented):
Code looks good, maybe needs some testing.
For the testing part I guess we need to create an e2e test that performs an upgrade so that we have the watcher in play, correct ?

ycombinator:
Should we also cleanup the Upgrade Marker file when we kill the Upgrade Watcher process? I think it would be good to try and maintain the invariant that the Upgrade Watcher is expected to be running if there's an Upgrade Marker file present.

blakerouse:
> Should we also cleanup the Upgrade Marker file when we kill the Upgrade Watcher process? I think it would be good to try and maintain the invariant that the Upgrade Watcher is expected to be running if there's an Upgrade Marker file present.

This is only called for Uninstall that is removing everything anyway, do we need to specifically remove that file when killing the watcher?

blakerouse:
I am seeing this branch basically get all tests passing on Windows. I am seeing one issue with uninstall where it seems that the Elastic Agent is still running. I think its possible that on Windows that stopping the service doesn't mean that it fully gets stopped (which I believe we have another reported bug about). But I am not getting any issues with the watcher writing to the log file any more.

ycombinator:
> This is only called for Uninstall that is removing everything anyway, do we need to specifically remove that file when killing the watcher?

True, we don't need to worry about cleaning up the Upgrade Marker file specifically because everything will get cleaned up during Uninstall anyway. 

I'm working on https://github.com/elastic/elastic-agent/issues/2706 and I plan to reuse the work you're doing here over there so I was thinking ahead to that.  Don't worry about changing anything in this PR here.  I'll add the cleanup if it makes sense when I work on https://github.com/elastic/elastic-agent/issues/2706.

cmacknz:
You should be able to revert https://github.com/elastic/elastic-agent/commit/bf467e32de6c87c73d3f411c35fac40c9f3b8c75 on this branch as well to confirm that test passes with this fix.

blakerouse:
> You should be able to revert [bf467e3](https://github.com/elastic/elastic-agent/commit/bf467e32de6c87c73d3f411c35fac40c9f3b8c75) on this branch as well to confirm that test passes with this fix.

Locally I have been testing before that PR was merged, so it is passing without that change. I will revert it in this PR still.

blakerouse:
I have reverted 3b0f040460cba561d9503d716b33c241893c01db and 51ad472440b2136cf483bc49b72d4b09724618ca. I am seeing a failure still with those reverted on test "TestStandaloneUpgradeFailsStatus".

blakerouse:
I un-reverted https://github.com/elastic/elastic-agent/commit/bf467e32de6c87c73d3f411c35fac40c9f3b8c75 so the "TestStandaloneUpgradeFailsStatus" will be skipped.

I kept the https://github.com/elastic/elastic-agent/commit/94764be90ce4a207f2e380390cfb67558095eb9b reverted in this PR so the tests run much faster as the watcher is now killed.

internal/pkg/agent/install/uninstall.go
blakerouse: I added the loop to check for the watcher, this ensures that the watcher is fully killed before continuing.

blakerouse:
Testing failed with:

```
=== RUN   TestFleetManagedUpgrade/Upgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT
    fetcher_artifact.go:176: Downloading artifact from https://staging.elastic.co/7.17.13-d6f555d2/downloads/beats/elastic-agent/elastic-agent-7.17.13-linux-x86_64.tar.gz
    fetcher_artifact.go:255: Downloading artifact progress 28.97%
    fetcher_artifact.go:255: Downloading artifact progress 48.79%
    fetcher_artifact.go:255: Downloading artifact progress 65.56%
    fetcher_artifact.go:255: Downloading artifact progress 87.66%
    fetcher_artifact.go:255: Downloading artifact progress 100.00%
    fetcher_artifact.go:222: Completed downloading artifact from https://staging.elastic.co/7.17.13-d6f555d2/downloads/beats/elastic-agent/elastic-agent-7.17.13-linux-x86_64.tar.gz
    fetcher_artifact.go:176: Downloading artifact from https://staging.elastic.co/7.17.13-d6f555d2/downloads/beats/elastic-agent/elastic-agent-7.17.13-linux-x86_64.tar.gz.sha512
    fetcher_artifact.go:222: Completed downloading artifact from https://staging.elastic.co/7.17.13-d6f555d2/downloads/beats/elastic-agent/elastic-agent-7.17.13-linux-x86_64.tar.gz.sha512
    fixture.go:221: Extracting artifact elastic-agent-7.17.13-linux-x86_64.tar.gz to /tmp/TestFleetManagedUpgradeUpgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT682026941/001
    fixture.go:234: Completed extraction of artifact elastic-agent-7.17.13-linux-x86_64.tar.gz to /tmp/TestFleetManagedUpgradeUpgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT682026941/001
    fixture.go:540: Components were not modified from the fetched artifact
    upgrade_test.go:108: Creating Agent policy...
    upgrade_test.go:121: Creating Agent enrollment API key...
    upgrade_test.go:128: Getting default Fleet Server URL...
    upgrade_test.go:132: Enrolling Elastic Agent...
    fixture_install.go:88: [test TestFleetManagedUpgrade/Upgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT] Inside fixture install function
    fixture.go:365: >> running agent with: [/tmp/TestFleetManagedUpgradeUpgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT682026941/001/elastic-agent-7.17.13-linux-x86_64/elastic-agent install --force --url https://80354456e6344e1ab7362baf3a3ae194.fleet.us-central1.gcp.qa.cld.elstc.co:443 --enrollment-token ZGczZmpvb0JXUkdseGpjSnpaLV86ZEFnV0drMVFRRU9sSVVGZ3ZvNjFfZw==]
    upgrade_test.go:147: 2023-09-13T14:09:11.196Z	INFO	cmd/enroll_cmd.go:454	Starting enrollment to URL: https://80354456e6344e1ab7362baf3a3ae194.fleet.us-central1.gcp.qa.cld.elstc.co:443/
        Error: failed to fix permissions: chown /opt/Elastic/Agent/elastic-agent: no such file or directory
        For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/7.17/fleet-troubleshooting.html
        Error: enroll command failed with exit code: 1
        For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/7.17/fleet-troubleshooting.html
        
    upgrade_test.go:149: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_test.go:149
        	            				/home/ubuntu/agent/testing/integration/upgrade_test.go:99
        	            	error running agent install command: exit status 1
        	Test:       	TestFleetManagedUpgrade/Upgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT
--- FAIL: TestFleetManagedUpgrade/Upgrade_managed_agent_from_7.17.13_to_8.11.0-SNAPSHOT (110.41s)
```

I have not seen this running the tests myself. Seems there is still an issue somewhere.

blakerouse:
buildkite test this

internal/pkg/agent/install/uninstall.go
```diff
@@ -290,3 +297,36 @@ func applyDynamics(ctx context.Context, log *logger.Logger, cfg *config.Config)
...
+	for {
+		// finding and killing watchers is performed in a loop until no
+		// more watchers are existing, this ensures that during uninstall
+		// that no matter what the watchers are dead before going any further
+		pids, errs := utils.GetWatcherPIDs()
+		if errs != nil {
+			return errs
+		}
+		if len(pids) == 0 {
+			return nil
```
cmacknz, line 312: Let's log when we hit this case so we can still what this function is doing.

internal/pkg/agent/install/uninstall.go
```diff
@@ -290,3 +297,36 @@ func applyDynamics(ctx context.Context, log *logger.Logger, cfg *config.Config)
...
+		}
+		for _, pid := range pids {
```
cmacknz, line 314: We should also log the PIDs we are going to kill

cmacknz (commented):
Looks good, but I think we should add some logging to the `killWatcher` function.

blakerouse:
If we want to put logging in it then it would just be to stdout, this is a command that the users runs to uninstall. It's not the agent running this code as a service.

Don't want to pollute the output from uninstall with log lines, so I will add them to the stdout output for the command.

pkg/utils/watcher.go
```diff
@@ -0,0 +1,49 @@
...
+		// that even if the watcher is spawned without the '.exe' suffix (which Windows will allow and supports)
+		// it always results in the watch process being killed
+		if strings.TrimSuffix(filepath.Base(state.Args[0]), ".exe") == "elastic-agent" && state.Args[1] == "watch" {
+			// it is a watch subprocess
+			pids = append(pids, pid)
+		}
+	}
+	if errs != nil {
+		return nil, errs
+	}
```
AndersonQ, line 47: It seems it isn't being used
```suggestion
```

pkg/utils/watcher.go
```diff
@@ -0,0 +1,49 @@
...
+	err := procStats.Init()
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize process.Stats: %w", err)
+	}
+	pidMap, _, err := procStats.FetchPids()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch pids: %w", err)
+	}
+	var pids []int
+	var errs error
```
AndersonQ, line 31:
```suggestion
```

internal/pkg/agent/install/uninstall.go
```diff
@@ -290,3 +297,36 @@ func applyDynamics(ctx context.Context, log *logger.Logger, cfg *config.Config)
...
+
+// killWatcher finds and kills any running Elastic Agent watcher.
+func killWatcher() error {
+	for {
+		// finding and killing watchers is performed in a loop until no
+		// more watchers are existing, this ensures that during uninstall
+		// that no matter what the watchers are dead before going any further
+		pids, errs := utils.GetWatcherPIDs()
+		if errs != nil {
+			return errs
```
AndersonQ, line 309: Even if it'd return several errors, as they aren't individually handled here, I'd suggest to stick to the convention
```suggestion
		pids, err := utils.GetWatcherPIDs()
		if errs != nil {
			return err
```

internal/pkg/agent/install/uninstall.go
```diff
@@ -290,3 +297,36 @@ func applyDynamics(ctx context.Context, log *logger.Logger, cfg *config.Config)
 
 	return config.NewConfigFrom(finalConfig)
 }
+
+// killWatcher finds and kills any running Elastic Agent watcher.
+func killWatcher() error {
+	for {
```
AndersonQ, line 303: [Blocker]
I don't see a stop condition, when will the function return?
Shouldn't you remove the PIDs which `proc.Kill()` return `os.ErrProcessDone` from the list?

AndersonQ (changes requested):

internal/pkg/agent/install/uninstall.go
blakerouse: `utils.GetWatcherPIDs()` returns a list of PID's. It returns when there are no more found watcher processes.

blakerouse:
@cmacknz I have added the same progress tracking that install has to uninstall so its clear on what uninstall is doing and that includes killing the watcher.

@ycombinator This changes the progress tracker a lot allow steps to have other nested steps. It didn't break any of the output formatting you had. Let me know what you think.

internal/pkg/agent/install/uninstall.go
blakerouse: Done!

internal/pkg/agent/install/uninstall.go
blakerouse: Done!

internal/pkg/agent/install/uninstall.go
cmacknz: I had this thought too, the only other option is to add a timeout but we have no idea how long to set it to. We could pick some arbitrarily long value beyond which we think users might consider uninstall to be hanging? That's probably 30s to 1 minute. Part of why I wanted logging was so we could see how long this takes in practice.

I suppose this could be tied into the root context of the application so that SIGTERM causes it to stop, if it doesn't already.

internal/pkg/agent/install/progress.go
```diff
@@ -15,15 +15,98 @@ import (
...
+func (pts *progressTrackerStep) Failed() {
+	prefix := " "
+	if pts.substeps {
+		prefix = pts.prefix + "   "
+	}
+	if !pts.rootstep {
+		pts.tracker.printf("%sFAILED\n", prefix)
+	}
+	pts.finalizeFunc()
+}
```
AndersonQ, line 71: [Suggestion]
you could make a function that takes a string (`DONE` or `FAILED`) and then `Succeeded` and `Failed` call it with the right string.
but the chance someone will need to change either `Succeeded` or `Failed` and forget the other is small, so it's realy up to you

blakerouse:
stack never became ready, I will try again

blakerouse:
buildkite test this

ycombinator:
Tested the progress checker changes in this PR.  I like how this looks!

```
Uninstalling current Elastic Agent...
   Stopping service... DONE
   Stopping upgrade watcher; none found... DONE
   Removing service... FAILED
   Removing install directory... DONE
   DONE
Copying files.................................................................... DONE
Installing service... DONE
Starting service... DONE
```

I have two concerns:
* I don't think we should show the step about the upgrade watcher to the user.  I think of the upgrade watcher as an implementation detail.  So my preference would be to take out that step from the progress output. 
* I don't know how the nested steps would generalize for steps that take long.  Specifically, the `......` could wrap around and "break" the indentation for a nested step that takes long.  But I'm okay deferring this concern for another day; the only nested steps right now should all complete pretty quickly.

blakerouse:
> 
> I have two concerns:
> 
> * I don't think we should show the step about the upgrade watcher to the user.  I think of the upgrade watcher as an implementation detail — something the Agent runs internally to monitor it's upgrade.  So my preference would be to take out that step from the progress output.

@cmacknz specifically asked for this information to be added, so its clear that a watcher was running and what PID was killed. I think overall its good to see, even if its implementation detail.

> * I don't know how the nested steps would generalize for steps that take long.  Specifically, the `......` could wrap around and "break" the indentation for a nested step that takes long.  But I'm okay deferring this concern for another day; the only nested steps right now should all complete pretty quickly.

Yeah I think we would need switch from `....` to like a spinner to ensure that it doesn't wrap. Otherwise we track the terminal width which to me would be more extra code then needed.

ycombinator (approved):
Progress tracker related changes LGTM!

pchila (approved):

blakerouse:
seems the stack was either removed mid test or was never really ready, trying again...

ycombinator:
@blakerouse Could we backport this PR to `8.10`, since it's fixing a bug (https://github.com/elastic/elastic-agent/issues/3371)? Also, this PR introduces a `utils.GetWatcherPIDs()` function which is needed in another `8.10` backport PR (https://github.com/elastic/elastic-agent/pull/3488).

blakerouse:
@ycombinator Yes, backporting it now.

blakerouse:
@ycombinator Actually this cannot be backported without backporting the work you did to add the installation output in 8.11. This work is based off of that. We can either backport that one as well, or not backport this one. Only backport the `utils.GetWatcherPIDs()` into the upgrade tests refactor backport. Thoughts?

@cmacknz your opinion as well?

ycombinator:
> Only backport the `utils.GetWatcherPIDs()` into the upgrade tests refactor backport. 

I'm in favor of this as the installation output work is an enhancement (not a bugfix).

PR #3411, opened 2023-09-13
Fix TestRemovePath being flagged as a possible virus

cmacknz:
My suspicion is this test gets flagged as malware because it writes code to disk and then compiles it by executing `go build` itself. This does seem like something malware could do, so I changed this to use the same build strategy the fake component uses. The `mage build:testBinary` target must be used to build the binaries before the test can run, which is a dependency of `mage unitTest`.

### DIFF
--- a/internal/pkg/agent/install/uninstall_windows_test.go
+++ b/internal/pkg/agent/install/uninstall_windows_test.go
@@ -7,7 +7,6 @@
 package install
 
 import (
-	"os"
 	"os/exec"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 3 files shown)


cmacknz:
Taking out of draft, got one successful run. Will run the tests a few times before merging to confirm. I expect test may still be flaky for other reasons, I just want to eliminate the virus scanning false positive with these changes.

ycombinator (approved):
LGTM.

internal/pkg/agent/install/uninstall_windows_test.go
```diff
@@ -7,49 +7,23 @@
...
+	cmd := exec.Command(binaryPath)
 	err = cmd.Start()
 	require.NoError(t, err)
 	defer func() {
 		_ = cmd.Process.Kill()
 		_ = cmd.Wait()
 	}()
-
-	err = RemovePath(dir)
-	require.NoError(t, err)
```
ycombinator, line 54: Shouldn't these lines still be there? Isn't `RemovePath` the function being tested?

cmacknz: Yes they should whoops

cmacknz: I did this far too quickly the first time, thanks for catching this.

ycombinator (approved):

cmacknz:
```
=== RUN   TestRemovePath
    uninstall_windows_test.go:43: 
        	Error Trace:	C:/Users/jenkins/workspace/_agent_elastic-agent-mbp_PR-3411/src/github.com/elastic/elastic-agent/internal/pkg/agent/install/uninstall_windows_test.go:43
        	            	fork/exec C:\Users\jenkins\AppData\Local\Temp\TestRemovePath532456694\001\subdir\testblocking.exe: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
        	Test:       	TestRemovePath
--- FAIL: TestRemovePath (2.95s)
```

Ugh

AndersonQ (approved):

pchila (approved):

cmacknz:
Merging, let's see if this problem comes back after some more test runs on other branches and adjust from there.

Issue #3368, opened 2023-09-07
When the GPG URL is unreachable, the upgrade should continue if the embedded GPG key still works


cmacknz:
We have a report in https://github.com/elastic/fleet-server/issues/2887#issuecomment-1710016052 that the upgrade process fails in air gapped environments if the fallback GPG key URL is unreachable.

If the embedded GPG key for verification still works, then this is not a fatal error and the upgrade process should still succeed. 

Acceptance Criteria:

* A test exists proving that a GPG verification method that fails does not affect the upgrade process if an alternative verification method suceeds.

jlind23:
@pierrehilbert - I just assigned this to @michalpristas as he is the one who worked on this code base.

pierrehilbert:
Perfect, I would have done the same :-)

cmacknz:
I wrote https://github.com/elastic/elastic-agent/issues/3403 describing how to implement an integration test for this case. While this specific fix can likely be verified purely with unit testing, we should consider the integration test part of the scope of this issue since it is necessary to prevent this problem from happening again.

PR #3426, opened 2023-09-17
If one gpg check is working, upgrade should continue

pierrehilbert:
## What does this PR do?

This PR removes the break in the verify loop to continue in case one verify is successful.

## Why is it important?

This change will allow upgrade to continue if remote gpg key is unreachable.

## Checklist

## Author's Checklist

- [ ]

## Related issues

Closes https://github.com/elastic/elastic-agent/issues/3368

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
@@ -33,8 +33,6 @@ func NewVerifier(verifiers ...download.Verifier) *Verifier {
 // Verify checks the package from configured source.
 func (e *Verifier) Verify(a artifact.Artifact, version string, skipDefaultPgp bool, pgpBytes ...string) error {
 	var err error
-	var checksumMismatchErr *download.ChecksumMismatchError
-	var invalidSignatureErr *download.InvalidSignatureError
 
 	for _, v := range e.vv {
 		e := v.Verify(a, version, skipDefaultPgp, pgpBytes...)
@@ -44,11 +42,6 @@ func (e *Verifier) Verify(a artifact.Artifact, version string, skipDefaultPgp bo
 		}
 
 		err = multierror.Append(err, e)
-
-		if errors.As(e, &checksumMismatchErr) || errors.As(err, &invalidSignatureErr) {
-			// Stop verification chain on checksum/signature errors.
-			break
-		}
 	}
 
 	return err
--- a/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
@@ -22,7 +22,7 @@ type Verifier struct {
 	log *logger.Logger
 }
 
-func (e *Verifier) Name() string {
+func (v *Verifier) Name() string {
 	return "composed.verifier"
 }
 
--- a/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
@@ -31,20 +31,20 @@ func NewVerifier(verifiers ...download.Verifier) *Verifier {
 }
 
 // Verify checks the package from configured source.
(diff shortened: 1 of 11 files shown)


michalpristas (approved):
it's not fixing the issue but we can push this in and have at least one strategy with verifiers to reduce number of failures

internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
```diff
@@ -44,11 +42,6 @@ func (e *Verifier) Verify(a artifact.Artifact, version string, skipDefaultPgp bo
 		}
 
 		err = multierror.Append(err, e)
-
-		if errors.As(e, &checksumMismatchErr) || errors.As(err, &invalidSignatureErr) {
```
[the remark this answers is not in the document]
pierrehilbert: What's the best option here according to you?
Log directly in each Verifier or only those specific errors here?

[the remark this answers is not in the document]
cmacknz: I don't have a strong preference, if you do it here the log message will be consistent for every verifier and nobody can forget. If you do it in each verifier we might get a better log message with more details. As long as we can tell what is failing and why it is fine with me.

internal/pkg/agent/application/upgrade/step_download.go
```diff
@@ -81,10 +81,13 @@ func (u *Upgrader) downloadArtifact(ctx context.Context, version, sourceURI stri
 		return "", errors.New(err, "initiating verifier")
 	}
 
-	if err := verifier.Verify(agentArtifact, parsedVersion.VersionWithPrerelease(), skipDefaultPgp, pgpBytes...); err != nil {
+	success, err := verifier.Verify(agentArtifact, parsedVersion.VersionWithPrerelease(), skipDefaultPgp, pgpBytes...)
+	if !success {
 		return "", errors.New(err, "failed verification of agent binary")
 	}
-
+	if err != nil {
```
cmacknz, line 88: This isn't really the standard way to do this, if the operation succeeded `err` should be `nil`. 

Now that I've seen this approach, I think a better solution would be to have the composed verifier log each verifier that failed but return only a `nil` error if at least one of them succeeded.

If you wanted to keep doing it this way, you would want to return the details of what happened as a separate structure so it can be logged at this level but I think it will be much simpler to handle the logging directly in this block of the composed verifier

Something like the following should handle most of this, it already builds up a joined error of each failure and only returns it if all of the verifiers fail. We just need to add a log statement and a `Name` method to the verifier interface so you can log which one failed.

```diff
--- a/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/composed/verifier.go
@@ -5,7 +5,7 @@
 package composed

 import (
-       "github.com/hashicorp/go-multierror"
+       "errors"

        "github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact"
        "github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact/download"
@@ -43,11 +43,10 @@ func (e *Verifier) Verify(a artifact.Artifact, version string, skipDefaultPgp bo
                        return nil
                }

-               err = multierror.Append(err, e)
+               err = errors.Join(err, e)

                if errors.As(e, &checksumMismatchErr) || errors.As(err, &invalidSignatureErr) {
-                       // Stop verification chain on checksum/signature errors.
-                       break
+                       log.Warningw("Verifier failed!", "verifier", v.Name(), "error", err)
                }
        }
```

cmacknz: You would still want the logger here to log the case where the all fail, which should be logged at the `Error` level.

You should be able to pass the logger you have here into the function call.

cmacknz: Yes, we need success = "no error".

The trouble you are running into is that you need to log twice, once inside the `composed/verifier.go` file for non-fatal errors that won't be returned (since final success = no error) and once when the verifier is called from the upgrade process for the fatal error where all verifiers failed.

cmacknz (approved):

PR #3427, opened 2023-09-18
[Bug] Do not fail when remote PGPs are not available

### LINKED ISSUE #3368: When the GPG URL is unreachable, the upgrade should continue if the embedded GPG key still works

michalpristas:
This PR adds filtering capabilities for incorrect location or download failure when trying to fetch PGP keys before verification. 
When this error is encountered warning is logged and key not considered. This will not prevent from verifying in air gapped environments where default elastic artifact api is not available

Fixes #3368

### DIFF
--- a/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
@@ -122,13 +122,14 @@ func (v *Verifier) verifyAsc(fullPath string, skipDefaultPgp bool, pgpSources ..
 			continue
 		}
 		raw, err := download.PgpBytesFromSource(check, v.client)
+		if errors.Is(err, download.ErrRemotePGPDownloadFailed) {
+			v.log.Warnf("Skipped remote PGP located at %q because it's unavailable: %v", strings.TrimPrefix(check, download.PgpSourceURIPrefix), err)
+			continue
+		}
 		if err != nil {
-			if errors.Is(err, download.ErrRemotePGPDownloadFailed) {
-				v.log.Warnf("Skipped remote PGP located at %q because it's unavailable: %v", strings.TrimPrefix(check, download.PgpSourceURIPrefix), err)
-				continue
-			}
 			return err
 		}
+
 		if len(raw) == 0 {
 			continue
 		}
--- a/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/elastic/elastic-agent-libs/transport/httpcommon"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact"
@@ -122,6 +123,10 @@ func (v *Verifier) verifyAsc(fullPath string, skipDefaultPgp bool, pgpSources ..
 		}
 		raw, err := download.PgpBytesFromSource(check, v.client)
 		if err != nil {
+			if errors.Is(err, download.ErrRemotePGPDownloadFailed) {
+				v.log.Warnf("Skipped remote PGP located at %q because it's unavailable: %v", strings.TrimPrefix(check, download.PgpSourceURIPrefix), err)
+				continue
+			}
 			return err
 		}
 		if len(raw) == 0 {
--- a/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/elastic/elastic-agent-libs/transport/httpcommon"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact"
@@ -125,11 +124,7 @@ func (v *Verifier) verifyAsc(fullPath string, skipDefaultPgp bool, pgpSources ..
 		if len(check) == 0 {
 			continue
 		}
-		raw, err := download.PgpBytesFromSource(check, v.client)
-		if errors.Is(err, download.ErrRemotePGPDownloadFailed) {
-			v.log.Warnf("Skipped remote PGP located at %q because it's unavailable: %v", strings.TrimPrefix(check, download.PgpSourceURIPrefix), err)
-			continue
-		}
+		raw, err := download.PgpBytesFromSource(v.log, check, v.client)
 		if err != nil {
 			return err
 		}
--- a/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier_test.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/fs/verifier_test.go
@@ -173,94 +173,60 @@ func prepareFetchVerifyTests(dropPath, targetDir, targetFilePath, hashTargetFile
 }
 
 func TestVerify(t *testing.T) {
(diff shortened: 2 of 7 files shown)


testing/integration/upgrade_test.go
```diff
@@ -266,13 +266,71 @@ func TestStandaloneUpgradeWithGPGFallback(t *testing.T) {
...
 	)
 
+	customPGP := CustomPGP{
+		PGP: newPgp,
+	}
+
+	testStandaloneUpgrade(ctx, t, agentFixture, fromVersion, toVersion, "", false, false, true, customPGP)
+}
+
+func TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing(t *testing.T) {
```
ycombinator, line 283: This test is good for testing the standalone upgrade path. But it would be good to have a unit test that calls `verifier, err := newVerifier(...)` followed by calling `verifier.Verify(...)` with no `pgpBytes` being passed to `verifier.Verify(...)`, because this is what's happening in the Fleet-managed upgrade path - details below. The unit test could use the [testing logger](https://github.com/elastic/elastic-agent/blob/735a8fbc52c94b78f9b893b761e4a373df46c2e3/pkg/core/logger/testing.go#L15-L17) to assert that the new warning you added in this PR is actually emitted.

Looking at the Fleet-managed upgrade path, no `pgpBytes` are passed to the `coordinator.Upgrade(...)` call here: https://github.com/elastic/elastic-agent/blob/b19e95069e28d0389e2e66a8281e5df823566ecc/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L78

Which means, no `pgpBytes` get passed to the `upgrader.Upgrade(...)` call here: https://github.com/elastic/elastic-agent/blob/b19e95069e28d0389e2e66a8281e5df823566ecc/internal/pkg/agent/application/coordinator/coordinator.go#L448

Which means, no `pgpBytes` are passed to the `u.downloadArtifact(...)` call here: https://github.com/elastic/elastic-agent/blob/b19e95069e28d0389e2e66a8281e5df823566ecc/internal/pkg/agent/application/upgrade/upgrade.go#L122

Which means, no `pgpBytes` are passed to the `verifier.Verify(...)` call here: https://github.com/elastic/elastic-agent/blob/b19e95069e28d0389e2e66a8281e5df823566ecc/internal/pkg/agent/application/upgrade/step_download.go#L84

However, right before that call, when `verifier` is [constructed](https://github.com/elastic/elastic-agent/blob/b19e95069e28d0389e2e66a8281e5df823566ecc/internal/pkg/agent/application/upgrade/step_download.go#L79), it ends up getting the embedded PGP bytes and passing them to each of three types of verifiers (fs, snapshot, and http): https://github.com/elastic/elastic-agent/blob/b19e95069e28d0389e2e66a8281e5df823566ecc/internal/pkg/agent/application/upgrade/step_download.go#L122-L145

So that's why everything works out, but it would be good to have a unit test that actually exercises this code path by calling `verifier, err := newVerifier(...)` followed by calling `verifier.Verify(...)` with no `pgpBytes` being passed to it. The unit test could use the [testing logger](https://github.com/elastic/elastic-agent/blob/735a8fbc52c94b78f9b893b761e4a373df46c2e3/pkg/core/logger/testing.go#L15-L17) to assert that the new warning you added in this PR is actually emitted.

ycombinator (changes requested):
Implementation looks good.  Just added one comment requesting a unit test that exercises the verification code in the same way that a Fleet-managed Agent being upgraded would.

testing/integration/upgrade_test.go
cmacknz: +1 to having a unit test, for an integration test I am more interested in a test that blocks access to the artifacts API entirely because that test will catch other attempts to reach out to the artifacts API unnecessarily, where as this one only catches attempts to get the PGP key.

michalpristas: @ycombinator i see where you're getting but i think without passing pgpBytes we cannot verify much. we append remote pgp in step_download and we want to verify that we won't fail if this is unavailable. 
i'm passing fake uri as a pgpBytes here to simulate the situation with unreachable remote pgp, so i would need to pass something what definitely fails i order to check logger output.

i will add UT with this scenario still. 

@cmacknz i've seen the issue with integration test you're describing but as it was unassigned and unplanned i was treating these separately. that one will most likely require a bit more effort in a sense of client injections.

internal/pkg/agent/application/upgrade/artifact/download/http/verifier.go
```diff
@@ -122,9 +122,14 @@ func (v *Verifier) verifyAsc(a artifact.Artifact, version string, skipDefaultPgp
 			continue
 		}
 		raw, err := download.PgpBytesFromSource(check, v.client)
+		if errors.Is(err, download.ErrRemotePGPDownloadFailed) {
```
cmacknz, line 125: This can't be unit tested like the FS verifier can?

internal/pkg/agent/application/upgrade/artifact/download/fs/verifier.go
```diff
@@ -121,9 +122,14 @@ func (v *Verifier) verifyAsc(fullPath string, skipDefaultPgp bool, pgpSources ..
 			continue
 		}
 		raw, err := download.PgpBytesFromSource(check, v.client)
+		if errors.Is(err, download.ErrRemotePGPDownloadFailed) {
```
cmacknz, line 125: Is there a way to avoid needing to duplicate this check into each individual verifier? 

Or at least is there a higher level test we can write that is guaranteed to try each possible download source so if we change this again in the future this can't be forgotten?

The integration test will catch the most important HTTP case at least.

michalpristas: i moved check to `PgpBytesFromSource` used in both verifiers. this function filters these errors, logs warning and returns empty pgp. empty pgp is ignored as it was before.

internal/pkg/agent/application/upgrade/artifact/download/fs/verifier_test.go
```diff
@@ -173,50 +173,60 @@ func prepareFetchVerifyTests(dropPath, targetDir, targetFilePath, hashTargetFile
...
-	targetDir, err := ioutil.TempDir(os.TempDir(), "")
-	if err != nil {
-		t.Fatal(err)
+	tt := []struct {
+		Name             string
+		RemotePGPUris    []string
+		UnreachableCount int
+	}{
+		{"default", nil, 0},
+		{"unreachable local path", []string{download.PgpSourceURIPrefix + "https://127.0.0.1:2874/path/does/not/exist"}, 1},
```
cmacknz, line 182: It feels like this test could move to the top level download package now, instead of testing it specifically in the fs implementation.

cmacknz (commented):
Latest change looks much nicer, I still think the way it is tested could be improved. I don't think you'd have tested it this way if you wrote the current iteration from the start, with part of the test coverage in one of the specialized implementations.

internal/pkg/agent/application/upgrade/artifact/download/fs/verifier_test.go
michalpristas: will add more unit tests for `PgpBytesFromSource` in a followup PR.

cmacknz: Sure, the test coverage exists today it just isn't in the optimal location so I'm fine with that.

cmacknz (approved):

Issue #3264, opened 2023-08-17
GPG key - Add a second fallback URL


jlind23:
If https://artifacts.elastic.co/GPG-KEY-elastic-agent is unavailable or failed to be reached, Elastic Agent should have a second fallback URL that will download the key from a path hosted by Fleet Server itself.

Related PR - https://github.com/elastic/elastic-agent/pull/2980/files

cmacknz:
The path here doesn't need to be configurable, although we could do that if it is easy. The default value should be the Fleet URL with the path as described in the Fleet server issue.

We will need to add adequate logging so we can tell this is happening, and add a new test or an extension to the GPG fallback upgrade test that already exists at https://github.com/elastic/elastic-agent/blob/65ac6d5793018b10b3489f18da21e3201f910404/testing/integration/upgrade_test.go#L220

nimarezainia:
I concur that the path need not be configurable. Let's reduce the scope. a fixed path relative to the installation layout of the fleet-server would suffice. [the user can modify the install location of the fleet server anyway when deployed on prem]

pchila:
Reopening as we rolled back the merge of this in main and 8.10

michalpristas:
@pchila  can you describe repro steps?

PR #3453, opened 2023-09-21
[Feature] Secondary fallback for package signature verification

michalpristas:
## What does this PR do?

This PR adds a capability to check for fleet server hosted PGP as the last resort in case elastic artifact API is unavailable.
Fleet server will host this PGP per version. any snapshot flags or other version qualifiers are ignored. 

## Why is it important?

Add ability to upgrade securiely in Air gapped environment where fleet server is only reachable URI.

## Checklist

Fixes: #3264

### DIFF
--- a/internal/pkg/agent/application/application.go
+++ b/internal/pkg/agent/application/application.go
@@ -160,7 +160,7 @@ func New(
 				EndpointSignedComponentModifier(),
 			)
 
-			managed, err = newManagedConfigManager(ctx, log, agentInfo, cfg, store, runtime, fleetInitTimeout)
+			managed, err = newManagedConfigManager(ctx, log, agentInfo, cfg, store, runtime, fleetInitTimeout, upgrader)
 			if err != nil {
 				return nil, nil, nil, err
 			}
--- a/internal/pkg/agent/application/managed_mode.go
+++ b/internal/pkg/agent/application/managed_mode.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/elastic/elastic-agent-client/v7/pkg/client"
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/actions"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/actions/handlers"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/dispatcher"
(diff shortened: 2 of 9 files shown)


internal/pkg/remote/client.go
```diff
@@ -247,6 +247,10 @@ func (c *Client) checkApiVersionHeaders(reqID string, resp *http.Response) {
 // URI returns the remote URI.
 func (c *Client) URI() string {
 	host := c.config.GetHosts()[0]
+	if strings.HasPrefix(host, string(ProtocolHTTPS)+"://") ||
+		strings.HasPrefix(host, string(ProtocolHTTP)+"://") {
+		return host + "/" + c.config.Path
+	}
```
pchila, line 253: Nit: wouldn't at this point be better to use a [url.Parse](https://pkg.go.dev/net/url#Parse) to extract the various parts and then [url.JoinPath](https://pkg.go.dev/net/url#URL.JoinPath) the configured path ?

pchila (commented):
Aside from one specific comment in the code, why don't we have an abstraction over the available PGP keys we can use for validating ?

Something like a `PGPKeyProvider` that would return all the available PGP keys (the builtin one, the one retrieved from internet, the one from fleet, the one from command line...) and return the available ones so that the verifier can try them in order to validate the software?

Wouldn't that make this also easier to test ?

changelog/fragments/1695289867-Secondary-fallback-for-package-signature-verification.yaml
```diff
@@ -0,0 +1,31 @@
...
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Secondary fallback for package signature verification
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+description: Ability to upgrade securely in Air gapped environment where fleet server is only reachable URI.
```
AndersonQ, line 18: [Suggestion]
I believe a few small changes here will make it easier to understand. 

```suggestion
description: Ability to upgrade securely in Air gapped environment where fleet server is the only reachable URI.
```

or
```suggestion
description: Ability to securely upgrade in Air gapped environment, where the fleet server is the only reachable URI.
```

internal/pkg/remote/client.go
AndersonQ: I'm wondering why have you added this extra case. Is it a valid case the protocol on `host` to be optional?

Normally I'd agree with Paolo, but as this function does not return an error and [url.Parse](https://pkg.go.dev/net/url#Parse) returns one, it'd not be possible to handle the error here. Therefore, I like how you did it.

AndersonQ (approved):
I don't see any blocker, just a few suggestions

pchila (approved):

PR #2908, opened 2023-06-20
Feature/report version from file

pchila:
## What does this PR do?

Package a new file containing the Elastic Agent "package version" passed using `AGENT_PACKAGE_VERSION` env var and make sure that that's the version reported by the agent. If no version is specified during packaging, the resulting file will be created with the current BEAT_VERSION and the agent will keep reporting version as it did before this change.

The file is read once for the daemon/service process and for every CLI command execution and the content is stored in a package private variable. If there are issues reading the file the `InitVersionInformation()` function will fallback on BEATS_VERSION and return an error. This initialization error is treated as non-blocking and logged by the running agent or printed to stderr by the CLI

## Why is it important?

For our agent early release process we want the ability of repackaging the agent while updating only the components that need to be patched quickly: in other words, we want to take a previously packaged agent, swap in updated versions of some component (maybe beats, maybe endpoint, maybe something else...) without modifying anything else.
This PR gives the possibility to specify a *package version* while performing `mage package` by passing the version using `AGENT_PACKAGE_VERSION` environment variable.
For example a "normal" package command to create a new 8.9.0-SNAPSHOT elastic agent package like:
```BEATS_VERSION=8.9.0 SNAPSHOT=true mage package```
would become something like this if we want to give it a different package version
```AGENT_PACKAGE_VERSION=8.9.0-super_special_sauce_remix BEATS_VERSION=8.9.0 SNAPSHOT=true mage package```
and using the second package form from the same agent source code we can have the agent report itself as `8.9.0-super_special_sauce_remix`

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## Related issues

- Closes #2582

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

Issue #2780, opened 2023-06-05
Test that the currently installed Elastic Agent version can upgrade to itself


cmacknz:
As we add test coverage for the Elastic Agent upgrade process one frequent pain point is that there is no easy way to upgrade to an agent that was just built locally or in the current CI run. Consider a CI job that wants to test that the agent version it just built can be upgraded to and from. 

Rather than require these scenarios start with a previously released version of the agent, or manually editing the version during the build process, we should make it possible for a version of the agent to upgrade to itself. The workflow we want is:

1. Build the latest version of the agent from the HEAD of main or a release branch.
2. Install this latest version of the agent.
3. Instruct the agent to upgrade to itself using the upgrade from local file path. This ensures the latest version can perform the upgrade process and successfully act as the target version for a new upgrade. 

We will still need to test other version combinations but this serves as the most basic sanity check with no dependencies on a previously released version of the agent to do the test.

The primary complication with this approach is that the upgrade artifact in this scenario will be unpacked to the `data/elastic-agent-$hash` directory where the value of $hash will conflict. We will need a way to resolve this collision. 

We can likely assume that this collision can only ever happen once with the current version so when we detect a collision we could add a special suffix and just alternate between adding and removing it. This would allow an agent to upgrade to itself indefinitely.

The solution here likely overlaps with the work in https://github.com/elastic/elastic-agent/issues/2579 and can possibly be done as part of that work.

cmacknz:
This could also be one way to address the request in https://github.com/elastic/elastic-agent/issues/697, where the "repair" just instructs the agent to upgrade to the currently install version.

cmacknz:
With the implementation of https://github.com/elastic/elastic-agent/issues/2579, this is now possible.

We should add a new integration test that explicitly confirms that an upgrade succeeds when the agent version is different but the commit hash is the same.

We could possibly eliminate some of the standalone upgrade scenarios that must "downgrade" to the previous snapshot as part of this, but that is also something we can evaluate as a follow up change.

cmacknz:
I think this test should have the version we upgrade to be the same as the current agent version, but with +buildYYYYMMDD added to it to mimic the early release process.

PR #3458, opened 2023-09-21
Change the selection of downgrade version in TestUpgradeBrokenPackageVersion

### LINKED ISSUE #3454: Flaky test: TestUpgradeBrokenPackageVersion
## Flaky Test

* **Test Name:** TestUpgradeBrokenPackageVersion

* **Link:** https://github.com/elastic/elastic-agent/blob/b1d2e6b04062b8572718e583590afe678579bf9d/testing/integration/upgrade_test.go#L868

* **Branch:** https://github.com/elastic/elastic-agent/tree/main
* **Artifact Link:** https://buildkite.com/elastic/elastic-agent/builds/3422#018ab6de-5d5c-4de6-949e-0f962b039a59
* **Notes:** 
The test tries an upgrade to itself depending on the output of the artifact API (it just takes the penultimate version)

### Stack Trace

```
paste stack trace here
```

pchila:
## What does this PR do?
Fixes selection of the elastic agent version to downgrade to in the test so that we don't try to upgrade the agent on its own version

## Why is it important?
The test outcome today depends on the order of versions returned by Artifact API making it flaky when there are some changes.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #3454 

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/upgrade_test.go
+++ b/testing/integration/upgrade_test.go
@@ -9,7 +9,6 @@ package integration
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/fs"
 	"net/http"
@@ -45,7 +44,14 @@ import (
 	agtversion "github.com/elastic/elastic-agent/version"
 )
 
-const fastWatcherCfg = `
+// The watcher will need the default 10 minutes to complete for a Fleet managed agent, see https://github.com/elastic/elastic-agent/issues/2977.
+const defaultWatcherDuration = 10 * time.Minute
+
+// Configure standalone agents to complete faster to speed up tests.
+const standaloneWatcherDuration = time.Minute
+
+// Note: this configuration can't apply to Fleet managed upgrades until https://github.com/elastic/elastic-agent/issues/2977 is resolved
+var fastWatcherCfg = `
 agent.upgrade.watcher:
   grace_period: 1m
   error_check.interval: 15s
@@ -94,8 +100,6 @@ func TestFleetManagedUpgrade(t *testing.T) {
 			err = agentFixture.Prepare(ctx)
 			require.NoError(t, err, "error preparing agent fixture")
 
-			err = agentFixture.Configure(ctx, []byte(fastWatcherCfg))
-			require.NoError(t, err, "error configuring agent fixture")
 			testUpgradeFleetManagedElasticAgent(t, ctx, info, agentFixture, parsedVersion, define.Version())
 		})
 	}
@@ -158,6 +162,11 @@ func testUpgradeFleetManagedElasticAgent(t *testing.T, ctx context.Context, info
 	t.Logf("Upgrade Elastic Agent to version %s...", toVersion)
 	err = tools.UpgradeAgent(kibClient, policy.ID, toVersion)
 	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		// The watcher needs to finish before the agent is uninstalled: https://github.com/elastic/elastic-agent/issues/3371
+		waitForUpgradeWatcherToComplete(t, agentFixture, parsedFromVersion, defaultWatcherDuration)
+	})
 
 	t.Log(`Waiting for enrolled Agent status to be "online"...`)
 	require.Eventually(t, tools.WaitForAgentStatus(t, kibClient, policy.ID, "online"), 10*time.Minute, 15*time.Second, "Agent status is not online")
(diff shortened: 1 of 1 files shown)


testing/integration/upgrade_test.go
```diff
@@ -936,18 +940,22 @@ func TestUpgradeBrokenPackageVersion(t *testing.T) {
 	versionList, err := aac.GetVersions(ctx)
 	require.NoError(t, err)
 	require.NotEmpty(t, versionList.Versions, "Artifact API returned no versions")
-	latestVersion := versionList.Versions[len(versionList.Versions)-1]
 
-	t.Logf("Upgrading to version %q", latestVersion)
+	// transform and reverse the version list and find the most recent version that is different from the broken version returned
+
+	// look for 1 version with the current major that is lower than the current one
```
AndersonQ, line 946: I'm guessing it's 1 thing? AM I right?

```suggestion
	// transform and reverse the version list and find the most recent version that is different from the broken version returned
	// look for 1 version with the current major that is lower than the current one
```

AndersonQ (approved):

blakerouse (changes requested):
This test overall has a problem. Being that this test is downgrading the installed build to a snapshot build, means that when 'elastic-agent uninstall' is ran at the end of the test will be the snapshots uninstall. 

The snapshot uninstall might not have the required code to kill the watcher, which then means the watcher is still running and will break a following test.

pchila:
@blakerouse since we start from the latest snapshot for test, we don't have that many choices about versions to move to.
The purpose of the test is to ensure that an agent install that has a broken/missing package version file can still be upgraded as [requested when the feature was implemented](https://github.com/elastic/elastic-agent/pull/2908#discussion_r1237153021)
If we cannot downgrade here, what do you suggest? Remove the test entirely? Skip it for a few release cycles?
If we don't merge this PR we still have the test just in the flaky version which is probably a worse outcome

blakerouse:
> @blakerouse since we start from the latest snapshot for test, we don't have that many choices about versions to move to. The purpose of the test is to ensure that an agent install that has a broken/missing package version file can still be upgraded as [requested when the feature was implemented](https://github.com/elastic/elastic-agent/pull/2908#discussion_r1237153021) If we cannot downgrade here, what do you suggest? Remove the test entirely? Skip it for a few release cycles? If we don't merge this PR we still have the test just in the flaky version which is probably a worse outcome

@pchila We have an overall problem with all the upgrade tests in the integration suite. They are either not testing the actual agent under test (https://github.com/elastic/elastic-agent/issues/3461) or the ending Elastic Agent is an older version of Elastic Agent that might not have required fixes for the testing suite to pass (in this case missing watcher kill).

> @blakerouse would running it in isolation address your concerns? for the time being. once we have previous major with proper watcher capabilities we would go back to normal runs

@michalpristas I wish that would work, but uninstall is a core part of the testing framework. Running this test in isolation will still result in it failing on Windows, because it will not be able to kill the watcher.

Only possible solution is to wait until the watcher is completely done running, before performing the uninstall. I think that is the only valid solution. This is the only case that works, that is still assuming that we don't hit an issue with `RemovePath` on the older version of the Elastic Agent.

cmacknz:
> @pchila We have an overall problem with all the upgrade tests in the integration suite. They are either not testing the actual agent under test (https://github.com/elastic/elastic-agent/issues/3461) or the ending Elastic Agent is an older version of Elastic Agent that might not have required fixes for the testing suite to pass (in this case missing watcher kill).

Fundamentally to solve this we need the ability to upgrade to the version of the agent that was built in the PR. We need https://github.com/elastic/elastic-agent/issues/2780.

We need to move tests between previously released versions and official snapshots onto a schedule because the code in PRs doesn't affect them at all.


> Only possible solution is to wait until the watcher is completely done running, before performing the uninstall. I think that is the only valid solution. This is the only case that works, that is still assuming that we don't hit an issue with RemovePath on the older version of the Elastic Agent.

Agreed I think this is the only way to deal with this right now, besides skipping the test or figuring out how to unit test this instead of integration testing it.

pchila:
> Fundamentally to solve this we need the ability to upgrade to the version of the agent that was built in the PR. We need https://github.com/elastic/elastic-agent/issues/2780.

For standalone we mostly do, except for versions of agent where we cannot specify a `file://` source-uri and a `skip-verify` flag. For Fleet managed tests I would expect the same but I didn't actually verify that we do use the packaged version of agent something can be improved there.

For the scope of this PR, do we want to include all of those changes? Or do we create a separate issues/PRs for those ?

pchila:
@blakerouse @cmacknz Adding the wait means partially revert commit 7e86d24cd336145b9641f621a3de10ed3af0c9fe where the cleanup functions have been removed.
Would that be acceptable for the moment in this PR ?

pchila:
@blakerouse could you give another review pass now that we have reintroduced the wait for watcher on the upgrade tests ?

pchila:
@cmacknz 

> We need to move tests between previously released versions and official snapshots onto a schedule because the code in PRs doesn't affect them at all.

https://github.com/elastic/elastic-agent/issues/3475

blakerouse:
@pchila See my PR #3477

I believe I have upgrade testing working correctly, even this one.

pchila:
This is replaced by https://github.com/elastic/elastic-agent/pull/3477

PR #3460, opened 2023-09-21
[BUG] Treat elastic-security container platforms as separate inputs

norrietaylor:
## What does this PR do?

This PR  modifies the elastic security spec file by treating the container platforms as separate inputs without the restriction of requiring the default install location. 

## Why is it important?
This fixes an issue where the beta deployment of Endpoint in K8s would throw an error on deployment. 

## Checklist

### DIFF
--- a/specs/endpoint-security.spec.yml
+++ b/specs/endpoint-security.spec.yml
@@ -5,13 +5,11 @@ inputs:
     platforms:
       - linux/amd64
       - linux/arm64
-      - container/amd64
-      - container/arm64
-    outputs:
+    outputs: &outputs
       - elasticsearch
       - logstash
       - kafka
-    proxied_actions:
+    proxied_actions: &proxied_actions
       - UNENROLL
       - UPGRADE
     runtime:
@@ -22,7 +20,7 @@ inputs:
           message: "Elastic Agent must be running as root"
         - condition: ${install.in_default} == false
           message: "Elastic Defend requires Elastic Agent be installed at the default installation path"
-    service:
+    service: &service
       cport: 6788
       log:
         path: "/opt/Elastic/Endpoint/state/log/endpoint-*.log"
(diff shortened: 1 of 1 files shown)


specs/endpoint-security.spec.yml
```diff
@@ -90,3 +88,26 @@ inputs:
       log:
         path: "C:\\Program Files\\Elastic\\Endpoint\\state\\log\\endpoint-*.log"
       operations: *operations
+  - name: endpoint
+    description: "Endpoint Security"
+    platforms:
+      - container/amd64
+      - container/arm64
+    outputs:
```
blakerouse, line 96: best to use a yaml reference for outputs as they look to be the same

specs/endpoint-security.spec.yml
```diff
@@ -90,3 +88,26 @@ inputs:
...
+      - elasticsearch
+      - logstash
+      - kafka
+    proxied_actions:
```
blakerouse, line 100: should these still be proxied in a container?

if so, use a reference as well

specs/endpoint-security.spec.yml
```diff
@@ -90,3 +88,26 @@ inputs:
...
+      - UNENROLL
+      - UPGRADE
+    runtime:
+      preventions:
+        - condition: ${runtime.arch} == 'arm64' and ${runtime.family} == 'redhat' and ${runtime.major} == '7'
+          message: "No support for RHEL7 on arm64"
+        - condition: ${user.root} == false
+          message: "Elastic Agent must be running as root"
+    service:
```
blakerouse, line 109: can this whole section use a reference? seems only operations are at the moment

specs/endpoint-security.spec.yml
norrietaylor: Will do! Thanks for the suggestion.

specs/endpoint-security.spec.yml
norrietaylor: Roger that!

blakerouse (approved):
Looks good, thanks for the changes.

PR #3462, opened 2023-09-22
add pre exit hook for integration test cleanup

### LINKED ISSUE #3456: Mage: Create cleanup target for integration tests resources
### Prerequisites
Currently integration tests leave stale deployments and cloud resources when a CI command is interrupted or forcibly restarted.

### The solution
We need a cleanup target to be executed in `pre-exit` buildkite hook. We already have `.integration-cache/state.yml` produced by integration target containing all created resources during the tests execution.
The state file should be consumed my the `mage integration:cleanup` target to delete the stack and could resources.
In this case the cleanup will be always executed by buildkite and no stale resources will be left behind

pchila:
## What does this PR do?
Add a buildkite pre exit hook for cleaning up of integration tests stacks and vms

## Why is it important?
Guarantees that we run `mage integration:clean` at the end of `integration test` job avoid leaking VMs and stack deployments.
We can see an example here where the build was cancelled on purpose ---> https://buildkite.com/elastic/elastic-agent/builds/3464#018abc62-f7c8-4d23-b3a5-471039638950

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #3456 

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -2,6 +2,16 @@
 
 set -eo pipefail
 
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent" && "$BUILDKITE_STEP_KEY" == "integration-tests" ]]; then
+  if [[ -z "${WORKSPACE-""}" ]]; then
+      WORKSPACE=$(git rev-parse --show-toplevel)
+  fi
+  source "${WORKSPACE}/.buildkite/scripts/common.sh"
+
+  # Perform cleanup of integration tests resources
+  TEST_INTEG_AUTH_ESS_REGION=azure-eastus2 SNAPSHOT=true mage integration:clean
+fi
+
 if [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
   if test -f "$GOOGLE_APPLICATION_CREDENTIALS"; then
     rm $GOOGLE_APPLICATION_CREDENTIALS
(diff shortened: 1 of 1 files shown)


pchila:
buildkite test this

.buildkite/hooks/pre-exit
```diff
@@ -2,6 +2,16 @@
...
 set -eo pipefail
 
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent" && "$BUILDKITE_STEP_KEY" == "integration-tests" ]]; then
+  if [[ -z "${WORKSPACE-""}" ]]; then
+      WORKSPACE=$(git rev-parse --show-toplevel)
+  fi
+  source "${WORKSPACE}/.buildkite/scripts/common.sh"
+
+  # Perform cleanup of integration tests resources
+  TEST_INTEG_AUTH_ESS_REGION=azure-eastus2 SNAPSHOT=true mage integration:clean
```
alexsapran, line 12: Can we add an extra log line with the `--- ` prefix so that in Buildkite it logs nicely and we know when this is executed?
Maybe something along the lines of 
```suggestion
  echo "--- Cleaning up integration test resources"
  TEST_INTEG_AUTH_ESS_REGION=azure-eastus2 SNAPSHOT=true mage integration:clean
```

pchila: Good idea!
Most of the time the output will be empty if the cleanup already completed in the step, so having the log printed is convenient :D

blakerouse: Could we place the `TEST_INTEG_AUTH_ESS_REGION=azure-eastus2` in the `common.sh` for both the testing step and this pre-exit? That way it only needs to be adjusted in one place.

belimawr (approved):

michalpristas (approved):

PR #3464, opened 2023-09-22
Kubernetes Secrets: Use a cached client

### LINKED ISSUE #3442: Performance issue with K8S Secret Provider
Hi 👋 ,

We have an Agent configuration which is referencing 2 `Secrets` using the [K8S Secret Provider](https://www.elastic.co/guide/en/fleet/current/kubernetes_secrets-provider.html), for example:

```yaml
          streams:
            - id: elasticsearch/metrics-elasticsearch.stack_monitoring.ccr
              data_stream:
                dataset: elasticsearch.stack_monitoring.ccr
                type: metrics
              metricsets:
                - ccr
              hosts:
                - 'https://${kubernetes.pod.ip}:9200'
              headers:
                Authorization: Bearer ${kubernetes_secrets.elastic-agent.somesecret1.value}
                AnotherHeader: SharedSecret ${kubernetes_secrets.elastic-agent.somesecret2.value}
              scope: node
              period: 10s
```

While investigating a global performance issue in our K8S cluster we observed a fair amount of requests to our K8S API server, with a **constant** rate of around 4 to 8 `GET` requests on these 2 Secrets per second. The audit trace suggests that that those requests are made by Agents managed by a `DaemonSet` on our clusters:

<details>
<summary>Audit Event Sample</summary>

```json
{
    "kind": "Event",
    "apiVersion": "audit.k8s.io/v1",
    "level": "Metadata",
    "auditID": "4a1822f1-6509-45ad-a767-4174f0beded",
    "stage": "ResponseComplete",
    "requestURI": "/api/v1/namespaces/elastic-agent/secrets/somesecret1",
    "verb": "get",
    "user": {
        "username": "system:serviceaccount:elastic-agent:elastic-agent",
        "uid": "7322c0d3-f498-4dae-a7e5-699d899fdfdf",
        "groups": [
            "system:serviceaccounts",
            "system:serviceaccounts:elastic-agent",
            "system:authenticated"
        ],
        "extra": {
            "authentication.kubernetes.io/pod-name": [
                "elastic-agent-b5hff"
            ],
            "authentication.kubernetes.io/pod-uid": [
                "64cb7be8-3d16-470f-afe6-c77fd092ffff"
            ]
        }
    },
    "sourceIPs": [
        "192.168.92.12"
    ],
    "userAgent": "Go-http-client/2.0",
    "objectRef": {
        "resource": "secrets",
        "namespace": "elastic-agent",
        "name": "somesecret1",
        "apiVersion": "v1"
    },
    "responseStatus": {
        "metadata": {},
        "code": 200
    },
    "requestReceivedTimestamp": "2023-09-20T14:55:44.652043Z",
    "stageTimestamp": "2023-09-20T14:55:44.655909Z",
    "annotations": {
        "authorization.k8s.io/decision": "allow",
        "authorization.k8s.io/reason": "RBAC: allowed by ClusterRoleBinding \"elastic-agent\" of ClusterRole \"elastic-agent\" to ServiceAccount \"elastic-agent/elastic-agent\""
    }
}
```

</details>

I think these `Secrets` should be cached, as it is usually the case when using the K8S client from the controller runtime for example. It would definitely help in our case.

We tried to have a look at the code and we were wondering if the client calls are coming from [here](https://github.com/elastic/elastic-agent/blob/d2a375831475668fc90f85fc076c6a741547e9c2/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L80).

Thanks!

- Version: `docker.elastic.co/beats/elastic-agent:8.8.2`
- Operating System: K8S 1.25 on EKS

barkbay:
## What does this PR do?

Use a namespace scoped cached client to read Kubernetes Secrets

## Why is it important?

To avoid hammering the K8s API, refer to #3442

## Checklist

## Author's Checklist

- [ ] Not sure about the log levels...

I don't know yet tbh 🤷 

## Related issues

Fixes #3442

## Use cases

n/a

## Screenshots

n/a

## Logs

_I'll update once I have a better understanding of the log levels in this project_

## Questions to ask yourself

Logs should be enough, if not using the k8s audit API can be an alternative.

Reader created, non cached requests...


barkbay:
Unscientific comparison with an Docker image I built myself on a small dev cluster:

image

The remaining spikes come from an external secret sync. operator.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -72,13 +121,20 @@ func (p *contextProviderK8sSecrets) Fetch(key string) (string, bool) {
...
-	ctx := context.TODO()
-	secret, err := secretIntefrace.Get(ctx, secretName, metav1.GetOptions{})
-	if err != nil {
+	secret := corev1.Secret{}
+	if err := reader.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: secretName}, &secret); err != nil {
+		if k8serrors.IsNotFound(err) {
+			p.logger.Errorf("Secret %s/%s not found: %v", ns, secretName, err)
+			return "", false
+		}
 		p.logger.Errorf("Could not retrieve secret from k8s API: %v", err)
```
gizas, line 138: Could be Debug level as this might create too much noise and false positives

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -72,13 +121,20 @@ func (p *contextProviderK8sSecrets) Fetch(key string) (string, bool) {
 		return "", false
 	}
 	ns := tokens[1]
+	reader, err := p.k8sReaderProvider.getReader(ns)
+	if err != nil {
+		p.logger.Errorf("Could not create K8S client: %v", err)
```
gizas, line 126:
```suggestion
		p.logger.Errorf("Could not create K8s client reader: %v", err)
```

gizas:
@barkbay are you gonna provide doc updates and samples on how to configure Elastic Agent with the secrets client?

I would expect some simple commands in the test section here to help me configure the reader for a given namespace. Can you please provide those?

barkbay:
> @barkbay are you gonna provide doc updates and samples on how to configure Elastic Agent with the secrets client?

This PR doesn't change anything from a user point of view, we are just replacing a K8S client implementation by an other one. Do you have something specific in mind?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -32,33 +37,77 @@ type contextProviderK8sSecrets struct {
...
+	// We restrict each cache to watch only 1 namespace. Watching all the namespaces would consume a lot of memory
+	// when we probably want to watch a very limited set of Secrets.
+	readers map[string]client.Reader
+
+	// ctx is the context used to start the informers. Informers will be stopped once the context is done.
+	ctx context.Context
+}
+
+// getReader returns a cached client associated to a given Kubernetes namespace.
+func (p *contextProviderK8sSecrets) getReader(namespace string) (client.Reader, error) {
```
barkbay, line 53: > Failed condition 34.1% 34.1% Coverage on New Code (is less than 40%)

I think Sonar is no happy because that function is not tested. I'll try to fix that.

gizas:
> > @barkbay are you gonna provide doc updates and samples on how to configure Elastic Agent with the secrets client?
> 
> This PR doesn't change anything from a user point of view, we are just replacing a K8S client implementation by an other one. Do you have something specific in mind?

Was referring to this page https://www.elastic.co/guide/en/fleet/current/kubernetes_secrets-provider.html as for now it is the only relevant that I can find. If this can be enhanced with an example with the manifest and refernece of a secret or some logs to troubleshoot references of https://github.com/elastic/elastic-agent/pull/3464/files#diff-8ce13c7c88aad2289f0a95358cafa95589e40293ba2dddeb14832bd2d604737fR141

changelog/fragments/1695390956-kubernetes_secrets-use-cached-client.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Use a cached client to retrieve Kubernetes Secrets
```
blakerouse, line 14: This doesn't really help our users when reading the changelog. Can this be updated more to describe the benefits this provides? Less API calls, better performance, etc.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -31,34 +32,91 @@ type contextProviderK8sSecrets struct {
...
+		}()
+		// Wait for the cache to be initialized.
+		newReader.WaitForCacheSync(p.ctx)
+	}
+	return reader, nil
 }
 
 // ContextProviderBuilder builds the context provider.
-func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed bool) (corecomp.ContextProvider, error) {
+func ContextProviderBuilder(logger *logger.Logger, c *config.Config, _ bool) (corecomp.ContextProvider, error) {
```
blakerouse, line 100: I assume that this change has the Elastic Agent now watching for changes in secrets, it would be great to see this also now trigger a re-render of the policy when a secret changes. Seems that the code is already listening to the API for those events, why not take it to the next level and improve it some more?

blakerouse (changes requested):

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -31,34 +32,91 @@ type contextProviderK8sSecrets struct {
...
+			namespace,
+		)
+		newReader, err := p.k8sCacheProvider.new(p.config, namespace)
+		if err != nil {
+			return nil, err
+		}
+		p.readers[namespace] = newReader
+		reader = newReader
+		go func() {
+			if err := newReader.Start(p.ctx); err != nil {
```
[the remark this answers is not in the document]
blakerouse: This is the blocker, by the way. I didn't make that clear in the review. Sorry about that.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
barkbay: I applied your suggestion. I'm not really familiar with how we want to surface such situations, this was to be consistent with the current behaviour:

```diff
-       secretIntefrace := client.CoreV1().Secrets(ns)
-       ctx := context.TODO()
-       secret, err := secretIntefrace.Get(ctx, secretName, metav1.GetOptions{})
-       if err != nil {
-               p.logger.Errorf("Could not retrieve secret from k8s API: %v", err)
+       secret := corev1.Secret{}
+       if err := reader.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: secretName}, &secret); err != nil {
+               if k8serrors.IsNotFound(err) {
+                       p.logger.Debugf("Secret %s/%s not found: %v", ns, secretName, err)
+                       return "", false
+               }
+               p.logger.Debugf("Could not retrieve secret from k8s API: %v", err)
```

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
barkbay: > it would be great to see this also now trigger a re-render of the policy when a secret changes.

I’m afraid I don’t know what "trigger a re-render of the policy" means.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
blakerouse: The context for a provider is only cancelled in the case that the Elastic Agent is being stopped.

The kubernetes context provider was the only "fetch` provider that was added to Elastic Agent and it was added after the fact. We didn't want the Elastic Agent to read/watch every secret, which we don't. The original design of all context and dynamic providers was for it to keep a local cache of all keys/values and then signal the composable manager when a key/value changes.

Based on that information it was never required to alert a provider to a re-render (which happens anytime the policy is updated, or when a key/value is changed from any provider in the Elastic Agent composable manager). I believe for this to work correctly you will need to handle this correctly to add this type of caching.

See me comment here https://github.com/elastic/elastic-agent/issues/3442#issuecomment-1731352600. Everything in that comment still applies even with the usage of https://github.com/kubernetes-sigs/controller-runtime/blob/d5bc8734caccabddac6a1bea250b0b9d771d318d/pkg/cache/cache.go#L61-L70

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
blakerouse: See my comment above for more context. It means that when a secret value is updated that you would call to the `comm Communicator` that a value has changed. That will signal the Elastic Agent composable manager to re-render the component model because something changed.

Think of it like this:

1. Elastic Agent is running with a policy and inside of that policy it references a password from a kubernetes secret.
2. The kubernetes secret is updated to change the password.
3. Elastic Agent gets the event that the password has changed.
4. It updates its cache for the secret.
5. It alerts the composable manager that a value has changed.
6. The Elastic Agent re-renders the policy with that new password substituted.
7. The Elastic Agent sends the new configuration to the running component.
8. The running component starts using the new password.

Elastic Agent already works that way today, just not for kubernetes secrets (everything else does). The kubernetes secrets provider just needs to "5." in that list and Elastic Agent would do the rest for you.

barkbay: > I assume that this change has the Elastic Agent now watching for changes in secrets,

The benefit of using the controller runtime client is to get an abstraction over the watch mechanism, and therefore to get a eventually up-to-date cache without managing watches. Given your explanations, I feel less and less confident in continuing down this path.

Issue #3191, opened 2023-08-04
Support flattened `data_stream` fields


belimawr:
The Elastic-Agent supports flattened YAML keys in its configuration files (e.g: `agent.logging.level`) however the `data_stream` object in input configurations (both the input root and the `streams` array) do not recognise flattened `data_stream.*` keys.

The following does not work:
```yaml
inputs:
  - type: filestream
    id: unique-id-per-input
    data_stream.dataset: mystuff
    paths:
      - /var/log/my-file/my.log*
```

And this works:
```yaml
inputs:
  - type: filestream
    id: unique-id-per-input
    data_stream:
      dataset: mystuff
    paths:
      - /var/log/my-file/my.log*
```

Both notations should work.

ruflin:
@mdbirnstiehl @yngrdyn This should allow us to simplify some of our configs in the docs and potentially onboarding flow.

cmacknz:
Re-opening as part of this work had to be reverted due to a bug starting Elastic Agent inputs: https://github.com/elastic/beats/pull/36610

PR #3465, opened 2023-09-22
Support flattened data_stream.* fields

belimawr:
## What does this PR do?

An input configuration supports flattened fields, however the 'data_stream' field was not being correctly decoded when flattened. This commit fixes this issue.

## Why is it important?

It increases consistency within the Elastic-Agent configuration and our other projects, like Beats, because they all support flattened configuration keys.

## Checklist

~~## Author's Checklist~~

1. Deploy a standalone Elastic-Agent with a configuration like the following.
```yaml
outputs:
  default:
    type: elasticsearch
    hosts:
      - "${ELASTICSEARCH_HOST}"
    username: "${ELASTICSEARCH_USERNAME}"
    password: "${ELASTICSEARCH_PASSWORD}"

inputs:
  - type: filestream
    id: elastic-agent-input-id
    streams:
      - id: filestream-input-id-1
        data_stream:
          dataset: generic
        data_stream.namespace: "flattened-namespace"
        data_stream.type: "logs"
        paths:
          - /tmp/log.log
```
2. Make sure some fields from `data_stream.*` are flattened. It is even better if there is a mix
3. Look at the ingested data and make sure the fields correctly set.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3191
- Docs PR: https://github.com/elastic/ingest-docs/pull/489

~~## Use cases~~
~~## Screenshots~~
~~## Logs~~

## Questions to ask yourself

- ...


testing/integration/datastreams_test.go
```diff
@@ -0,0 +1,361 @@
...
+	numEvents := uint64(60)
+
+	tempDir := t.TempDir()
+	logFilePath := filepath.Join(tempDir, "log.log")
+	generateLogFile(t, logFilePath, 2*time.Millisecond, numEvents)
+
+	info := define.Require(t, define.Requirements{
+		Local: false,
+		Stack: &define.Stack{
+			Version: version.Agent + "-SNAPSHOT",
```
belimawr, line 43: CI was failing because the stack was too old, so I "hardcoded" the version at least to make sure the tests are passing

pkg/component/config.go
```diff
@@ -100,9 +101,91 @@ func ExpectedConfig(cfg map[string]interface{}) (*proto.UnitExpectedConfig, erro
...
+		Type:      merge(tmp.DataStream.Type, ds.Type),
+		Namespace: merge(tmp.DataStream.Namespace, ds.Namespace),
+		Source:    raw.GetDataStream().GetSource(),
+	}
+
+	return ret, nil
+}
+
+// merge returns b if a is an empty string
+func merge(a, b string) string {
```
faec, line 164: Could this be named something like `valueOrDefault` instead? "Merge" sounds like something that incorporates both values.

pkg/component/config.go
```diff
@@ -100,9 +101,91 @@ func ExpectedConfig(cfg map[string]interface{}) (*proto.UnitExpectedConfig, erro
...
 }
 
+// dataStreamAndSource is a generic way to represent proto mesages
+// that contain a source field and a datastream field.
+type dataStreamAndSource interface {
+	GetDataStream() *proto.DataStream
+	GetSource() *structpb.Struct
+}
+
+func deDotDataStream(raw dataStreamAndSource) (*proto.DataStream, error) {
```
faec, line 118: Non-blocking suggestion: what if this helper took two parameters, the datastream and the source? Then those fields could be unwrapped at the call site without a one-off interface wrapper.

pkg/component/config.go
```diff
@@ -100,9 +101,91 @@ func ExpectedConfig(cfg map[string]interface{}) (*proto.UnitExpectedConfig, erro
...
+	ds := raw.GetDataStream()
+	if ds == nil {
+		ds = &proto.DataStream{}
+	}
+
+	tmp := struct {
```
faec, line 124: Suggestions for clarity: (1) move the declaration of `tmp` to immediately before the `Unpack` call so their association is more obvious, (2) add a comment explaining that `Unpack` handles dotted fields so `tmp` will get the correct values from the raw config tree.

pkg/component/config.go
```diff
@@ -100,9 +101,91 @@ func ExpectedConfig(cfg map[string]interface{}) (*proto.UnitExpectedConfig, erro
...
+	if (ds.Type != tmp.DataStream.Type) && (ds.Type != "" && tmp.DataStream.Type != "") {
+		return nil, errors.New("duplicated key 'datastream.type'")
+	}
+
+	if (ds.Namespace != tmp.DataStream.Namespace) && (ds.Namespace != "" && tmp.DataStream.Namespace != "") {
+		return nil, errors.New("duplicated key 'datastream.namespace'")
+	}
+
+	ret := &proto.DataStream{
+		Dataset:   merge(tmp.DataStream.Dataset, ds.Dataset),
```
faec, line 154: Actually is the `merge` helper really needed? I don't see a way for the `tmp.DataStream` fields to be empty if the `ds` ones are not... if we can always overwrite with the unpacked values, that simplifies these assignments (and some of the duplication logic above). (On the other hand, if there is a way that `DataStream.*` could be empty when `ds.*` is not, please add a comment noting how that can happen.)

faec (approved):
A few nits for code clarity but looks good to me!

pkg/component/config.go
belimawr: > Actually is the merge helper really needed?

I'm not completely sure, but I believe there is a case where it might be needed. The custom logs integration creates the following YAML:
```yaml
inputs:
  - id: logfile-logs-17010054-5623-4437-8079-69fada4de0b1
    name: 'Log-Input-TestFlattenedDatastreamFleetPolicy-2023-10-09T15:25:04Z'
    revision: 1
    type: logfile
    use_output: default
    meta:
      package:
        name: log
        version: 2.3.0
    data_stream:
      namespace: testflatteneddatastreamfleetpolicynamespace17719163731420661242
    package_policy_id: 17010054-5623-4437-8079-69fada4de0b1
    streams:
      - id: logfile-log.logs-17010054-5623-4437-8079-69fada4de0b1
        data_stream:
          dataset: testflatteneddatastreamfleetpolicy-dataset
        paths:
          - /tmp/TestFlattenedDatastreamFleetPolicy3982960466/001/log.log
        ignore_older: 72h
```
The `data_stream.*` fields are defined both in the `inputs` section as well as in the `streams` section. My understanding is that somewhere before `deDotDataStream` is called, some values will be copied over and they will be present only in the `DataStream` struct. Hence this need to 'merge' the contents of `DataStream` and `source`.

If there is only one value or they're equal we just use it, if they differ then the function returns an error.

testing/integration/datastreams_test.go
```diff
@@ -0,0 +1,400 @@
...
+
+	require.Eventually(
+		t,
+		ensureDocumentsInES(t, context.TODO(), info.ESClient, dsType, dsDataset, dsNamespace, int(numEvents)),
+		120*time.Second,
+		time.Second,
+		"could not get all expected documents form ES")
+}
+
+func TestFlattenedDatastreamStandalone(t *testing.T) {
```
cmacknz, line 146: 1. Do we need an entire integration test to test this? Why do we need an Elastic stack? Could this be tested with the fake component in a unit test?
2. Why do we need to test this for both fleet and standalone? What actually differs between these cases that we need to cover them separately? Does how the policy arrives at the agent change how it gets parsed?

belimawr: > Why do we need to test this for both fleet and standalone? What actually differs between these cases that we need to cover them separately?

I'm being extra cautious here. When I first implemented it in Beats, all my test cases were passing, but when running Elastic-Agent managed by Fleet, some fields were set slightly different that my handcrafted standalone policy, which broke all other cases.

Learning from that mistake I thought it would be good to have:
1. A test case managed by Fleet because the change affects how the policy is parsed/interpreted
2. Something that ensures we're not breaking the normal behaviour the Elastic-Agent is run managed by Fleet and a simple integration is added/used.

> Does how the policy arrives at the agent change how it gets parsed?

I'm not quite sure I fully understood this question. Depending on how the policy is sent to Elastic-Agent it can arrive in the `DataStream` field from the input/streams or in the `Source` field as YAML that needs to be decoded/parsed. The test cases should cover a mix of both and the "default behaviour" of just installing Elastic-Agent and adding integrations without changing any of the defaults.

> Do we need an entire integration test to test this? Why do we need an Elastic stack? Could this be tested with the fake component in a unit test?

I believe so, I wanted to test it end-to-end, so ensuring that the Elastic-Agent can properly interpret the `data_stream` configuration no matter how it's sent but also that Beats will work as expected when receiving the information from Elastic-Agent. That is also why I check for the documents on ES with the expected index name rather than asserting it any other way.

Probably the Fleet managed test could be renamed to reflect something like "simple happy path always works".

cmacknz: We are learning that maintaining integration tests is very expensive, so I only want us to write and maintain integration tests for features that can only be tested this way. I don't want us to have integration tests because it was possible, or "just in case" :) Our test suite will take hours and be difficult to maintain if we do this everywhere.

I would not expect the agent policy to be parsed differentlu when sent by Fleet. The part of the agent that converts the policy to components and parses it is centralized into the coorindator, not the configuration providers. See https://github.com/elastic/elastic-agent/blob/main/docs/architecture.md#coordinator

What I would expect is that the Fleet will perhaps use different variants of the syntax than you may have expected when writing standalone configurations by hand.

I could be convinced we need a Fleet integration test for this functionality, but I'm not convinced that we need an entirely new test for it. Could we just extend the monitoring and log test in https://github.com/elastic/elastic-agent/blob/main/testing/integration/monitoring_logs_test.go to also install the custom logs integration? That test is already exactly the same as the one you added minus that one integration. This would avoid the cost of creating an maintaining an entirely new test for a few small additions to an existing tests. Integration tests are not unit tests, they don't need to test only one piece of functionality.

I'm not convinced we need a standalone test case since there should not be a difference between Fleet and standalone agents in this case. If we need to cover all possible permutations of the YAML syntax we should be doing that in unit tests.

belimawr: I see your point and agree in not creating integrations test "just in case". In this specific case both, the Feet managed and standalone, tests also test the most basic use case for Elastic-Agent and Beats: log ingestion. I believe it is valuable to have at least one test like this for each case. They might be more like a smoke test but I still believe they're valuable.

I agree with not creating an entire new test for it. The way I coded both tests they're very isolated in the sense that all documents ingested go to indexes different than the default ones. I can refactor them to use the same stack deployment as `TestMonitoringLogsShipped`.

cmacknz: >  I can refactor them to use the same stack deployment as TestMonitoringLogsShipped.

👍 

I still don't think we need both a standalone and a fleet managed test for this, only one of them is enough. There is no functional difference besides how the policy is delivered to the agent. If you were to copy the agent policy generated by Fleet into a standalone agent the behaviour would be exactly the same.

belimawr: > If you were to copy the agent policy generated by Fleet into a standalone agent the behaviour would be exactly the same.

While that is true, the opposite is not. At least regarding the characters allowed for the `data_stream.*` fields. Fleet server/Kibina is more picky about it than a standalone Elastic-Agent.

If you really want I can remove the standalone test. I tried to re-use the same stack+VM, however the Elastic-Agent wasn't being properly uninstalled and I didn't want to spend time debugging it. I'd be happy to go back to it and make sure the same stack+VM can run all tests.

From my machine, with an existing stack and VM, the standalone Go test takes about 20s and the whole execution of `mage integration:single <test name>` takes about 50s.
```
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): --- PASS: TestFlattenedDatastreamStandalone (19.61s)
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): PASS
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): ok   github.com/elastic/elastic-agent/testing/integration    19.624s
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): DONE 1 tests in 24.121s
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): >> go test: remote-linux-amd64-ubuntu-2204-sudo.integration Test Passed
>>> Testing completed (1 successful)
>>> Console output written here: build/TEST-go-integration.out
>>> Console JSON output written here: build/TEST-go-integration.out.json
>>> JUnit XML written here: build/TEST-go-integration.xml
>>> Diagnostic output (if present) here: build/diagnostics
AGENT_KEEP_INSTALLED=false TEST_PLATFORMS=linux/amd64 mage -v    3.58s user 1.51s system 10% cpu 48.286 total
```

cmacknz: > Fleet server/Kibina is more picky about it than a standalone Elastic-Agent.

It's not clear to me what this means. Do you have an example?

If Fleet and Standalone agents have different views on what syntax is valid that is a bug. The two should be aligned. The allowed syntax must be the same. The actual syntax variant used may differ, but any form should be valid in either place.

pierrehilbert:
buildkite test this

testing/integration/datastreams_test.go
cmacknz: What error do you get from Fleet when you include the `-`?

It should be valid for an index name to contain a `-` going off of https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params.

This smells like Fleet is unnecessarily strict, that doesn't justify an entirely separate agent integration test to me, but we'll confirm first.

belimawr: If I make a request like this:
```
POST kbn:/api/fleet/package_policies
{
  "policy_id": "cfbf25e0-68ee-11ee-8f57-2948faf178f3",
  "package": {
    "name": "log",
    "version": "2.3.0"
  },
  "name": "foo-logs",
  "namespace": "my-namespace",
  "inputs": {
    "logs-logfile": {
      "enabled": true,
      "streams": {
        "log.logs": {
          "enabled": true,
          "vars": {
            "paths": [
              "/tmp/foo.ndjson"
            ],
            "data_stream.dataset": "my-dataset"
          }
        }
      }
    }
  }
}
```

Kibana/Fleet returns:
```
{
  "statusCode": 500,
  "error": "Internal Server Error",
  "message": "[namespace]: Namespace contains invalid characters"
}
```

If I try to put it in the input or stream level I get a Bad Request:
```
POST kbn:/api/fleet/package_policies
{
  "policy_id": "cfbf25e0-68ee-11ee-8f57-2948faf178f3",
  "package": {
    "name": "log",
    "version": "2.3.0"
  },
  "name": "foo-logs",
  "inputs": {
    "logs-logfile": {
      "enabled": true,
      "data_stream": {
        "namespace": "my-namespace"
      },
      "streams": {
        "log.logs": {
          "enabled": true,
          "vars": {
            "paths": [
              "/tmp/foo.ndjson"
            ],
            "data_stream.dataset": "my-dataset"
          }
        }
      }
    }
  }
}
```
Response
```
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": """[request body]: types that failed validation:
- [request body.0.inputs]: expected value of type [array] but got [Object]
- [request body.1.inputs.logs-logfile.data_stream]: definition for this key is missing"""
}
```

belimawr: If you really don't want both tests, I can remove one of them. Do you have any preferences?

I tried to have both, the standalone and Fleet managed using the same Stack/VM, but the Elastic-Agent managed by fleet is not being correctly stopped/uninstalled despite no error from the framework, so the standalone agent won't start :/

belimawr:
I rebased onto `main`, then force push. Let's see if CI goes green.

testing/integration/datastreams_test.go
cmacknz: Found the reason Fleet is not allowing the dash, it is "forbidden" in the data stream naming scheme: https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme

> In all three parts, the “-” character is not allowed. This means all data streams are named in the following way:

Also enforcing this in standalone agents would be ideal, but that is a breaking change for anyone that has done this already. I would just note this as a comment.

> If you really don't want both tests, I can remove one of them. Do you have any preferences?

The Fleet test is the one most likely to catch actual problems with the index names, as we have discovered so let's keep that one.

changelog/fragments/1695389490-Support-flattened-data_stream.-fields.yaml
```diff
@@ -0,0 +1,35 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Support flattened data_stream.* fields
```
pchila, line 14:
```suggestion
summary: Support flattened data_stream.* fields in input configuration
```

pkg/component/component_test.go
```diff
@@ -2361,3 +2361,74 @@ func gatherDurationFieldPaths(s interface{}, pathSoFar string) []string {
...
+		t.Fatalf("cannot convert policy to component: %s", err)
+	}
+
+	if len(result) != 1 {
+		t.Fatalf("expecting result to have one element, got %d", len(result))
+	}
+
+	if len(result[0].Units) != 2 {
+		t.Fatalf("expecting result[0].Units to have two elements, got %d", len(result))
+	}
```
pchila, line 2406: We can be more compact here

```suggestion
        require.NoErrorf(t, err, "cannot convert policy to component: %s", err)
        require.Len(t, result, 1, "unexpected result length")
        require.Len(t, result[0].Units, 2, "unexpected number of units for result[0]")
```

pkg/component/component_test.go
```diff
@@ -2361,3 +2361,74 @@ func gatherDurationFieldPaths(s interface{}, pathSoFar string) []string {
...
+
+	if dataStream.Dataset != expectedDataset {
+		t.Errorf("expecting DataStream.Dataset: %q, got: %q", expectedDataset, dataStream.Dataset)
+	}
+	if dataStream.Type != expectedType {
+		t.Errorf("expecting DataStream.Type: %q, got: %q", expectedType, dataStream.Type)
+	}
+	if dataStream.Namespace != expectedNamespace {
+		t.Errorf("expecting DataStream.Namespace: %q, got: %q", expectedNamespace, dataStream.Namespace)
+	}
```
pchila, line 2433: This can probably be substituted with an [assert.ElementsMatch()](https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch) if we are able to  predictably build DataStream objects (since we are building them from protobuf, that's not always a given)

testing/integration/logs_ingestion_test.go
```diff
@@ -0,0 +1,439 @@
...
+		3*time.Minute,
+		15*time.Second,
+	)
+
+	// TODO: remove after debugging
+	t.Log("--- debugging: results from ES --- START ---")
+	for _, doc := range docs.Hits.Hits {
+		t.Logf("%#v", doc.Source)
+	}
+	t.Log("--- debugging: results from ES --- END ---")
```
pchila, line 216: Is the TODO still relevant ?

testing/integration/logs_ingestion_test.go
```diff
@@ -0,0 +1,439 @@
...
+            ],
+            "data_stream.dataset": "{{.Dataset}}"
+          }
+        }
+      }
+    }
+  }
+}`
+
+type plolicyVars struct {
```
pchila, line 415:
```suggestion
type policyVars struct {
```

pchila (approved):
Overall it looks good, left a couple of (non-blocking) comments on some test code

pkg/component/component_test.go
belimawr: Protobuf types do not play well with testify, as far as I know they don't play well with any comparison package that is not built specifically for protobuf.

I rather keep the code like that and not have to worry about protobuf internal fields.

testing/integration/logs_ingestion_test.go
belimawr: It's not my code, but I also believe it's not relevant, I'll remove it. Thanks for catching it!

pkg/component/component_test.go
pchila: That should probably be a reason not to propagate protobuf objects across layers if they make writing tests more tedious than it needs to be

pchila (approved):
LGTM

pkg/component/component_test.go
belimawr: I get that, but it's not too bad either, `github.com/protocolbuffers/protobuf-go` and `github.com/google/go-cmp` comparing and getting a diff is not too bad. E.g:

https://github.com/elastic/elastic-agent/blob/910d17b42be5a7e04688573c5cfd2d7cbe9e00fa/pkg/component/component_test.go#L2110

belimawr:
`fleet-ci/pr-merge` is stuck :man_facepalming:

belimawr:
I rebased onto `main` to re-trigger the tests

cmacknz (commented):
You should update the reference YAML to include an example of this style of configuration, I would recommend modifying the sample log input:

https://github.com/elastic/elastic-agent/blob/dba216671d99e391ca66381f251d42dd04a6afc9/elastic-agent.yml#L47-L58

If the documentation isn't updated nobody is going to use this configuration style :)

belimawr:
> You should update the reference YAML to include an example of this style of configuration, I would recommend modifying the sample log input:

> If the documentation isn't updated nobody is going to use this configuration style :)

We already have examples since about 3 years ago :scream: 
https://github.com/elastic/elastic-agent/blob/e9be0709e45da4b504aa7a678d4736c329541fb0/elastic-agent.reference.yml#L24-L39

Anyways, I'll edit the `filestream` example as well.

cmacknz (approved):

belimawr:
@cmacknz `mage check` did not like my last changes in the configuration :/

Aside from the flattened examples I pointed out before we also have it shown in our documentation: https://github.com/elastic/ingest-docs/blob/bab384ba275f4ae34696cedbdc1221383b64e1c9/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-configuration.asciidoc?plain=1#L37-L50

I believe we have a good mix of syntax example.

I confess I'm puzzled about `mage check` only editing the commented out section of the reference configuration :thinking:

PR #3473, opened 2023-09-25
[Standalone Agent] Disallow upgrade if upgrade is already in progress

ycombinator:
## What does this PR do?

This PR prevents users of standalone Agents from being able to upgrade the Agent if an upgrade is already in progress.

## Why is it important?

Attempting to upgrade an Agent while an upgrade is already in progress could corrupt the Agent installation.

## Checklist

## Related issues

- Relates #2706


pkg/control/v2/client/mocks/client.go
pchila, line 1: :heart:

pkg/control/v2/client/client.go
```diff
@@ -156,6 +156,7 @@ type DiagnosticComponentResult struct {
 }
 
 // Client communicates to Elastic Agent through the control protocol.
+// go:generate mockery --name Client
```
pchila, line 159: :+1:

pchila (commented):
Code looks good. 
There is a clash between your changes for the upgrade integration test and the modifications proposed in https://github.com/elastic/elastic-agent/pull/3477 so I guess we have to sync a bit to make sure we don't break/lose testcases

For the code coverage, I guess we have to exclude `mocks/*.go` from analysis since most of the uncovered lines are actually the generated code.

ycombinator:
> For the code coverage, I guess we have to exclude `mocks/*.go` from analysis since most of the uncovered lines are actually the generated code.

Done in 98764d81a735503681812e44487ad1043a5cb4c0.

pchila (approved):

PR #3482, opened 2023-09-28
Fix double stop components

pchila:
## What does this PR do?

This PR prevents agent from stopping already stopped components.

## Why is it important?
Calling stop on an already stopped component may have unforeseen effects especially if uninstall is involved

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1695920792-Prevent-multiple-stops-of-services.yaml
+++ b/changelog/fragments/1695920792-Prevent-multiple-stops-of-services.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Prevent multiple attempts to stop an already stopped service
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: runtime
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
--- a/pkg/component/runtime/manager.go
+++ b/pkg/component/runtime/manager.go
@@ -705,6 +705,12 @@ func (m *Manager) update(model component.Model, teardown bool) error {
 		var stoppedWg sync.WaitGroup
 		stoppedWg.Add(len(stop))
 		for _, existing := range stop {
+			if existing.getLatest().State == client.UnitStateStopped {
+				// it's already stopped, move on
+				stoppedWg.Done()
+				continue
+			}
+			m.logger.Debugf("Stopping component %q", existing.id)
 			_ = existing.stop(teardown, model.Signed)
 			// stop is async, wait for operation to finish,
 			// otherwise new instance may be started and components
@@ -755,6 +761,7 @@ func (m *Manager) waitForStopped(comp *componentRuntimeState) {
 	for {
 		latestState := comp.getLatest()
 		if latestState.State == client.UnitStateStopped {
+			m.logger.Debugf("component %q stopped.", compID)
 			return
 		}
 
@@ -767,6 +774,7 @@ func (m *Manager) waitForStopped(comp *componentRuntimeState) {
 
 		select {
 		case <-timeoutCh:
+			m.logger.Errorf("timeout exceeded waiting for component %q to stop", compID)
 			return
 		case <-time.After(stopCheckRetryPeriod):
 		}
--- a/pkg/component/runtime/manager.go
+++ b/pkg/component/runtime/manager.go
@@ -705,8 +705,8 @@ func (m *Manager) update(model component.Model, teardown bool) error {
 		var stoppedWg sync.WaitGroup
 		stoppedWg.Add(len(stop))
 		for _, existing := range stop {
-			if existing.getLatest().State == client.UnitStateStopped {
-				// it's already stopped, move on
+			if existing.shuttingDown.Load() {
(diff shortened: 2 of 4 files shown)


pkg/component/runtime/service.go
```diff
@@ -136,6 +137,11 @@ func (s *serviceRuntime) Run(ctx context.Context, comm Communicator) (err error)
 	defer cisStop()
 
 	onStop := func(am actionMode) {
+		if stopping {
+			s.log.Debugf("service %s is already stopping: skipping...")
+			return
+		}
+		stopping = true
```
cmacknz, line 144: That this is never set to false doesn't seem right. What would happen if we stopped the service and then tried to start it again? This would be uninstalling the endpoint integration and then reinstalling it quickly, or perhaps reassigning the agent to a different policy.

The `ignoreCheckins` on this path is set to false again in the `case actionStart:` block.

pchila: When discussing the fix with @blakerouse , I asked about recycling a service runtime object and we came to the conclusion that a stopped component is ultimately discarded (removed from the maps of running components in the runtime manager) after it eventually stops.

If this assumption does not hold, then we need to keep track of the installation status and if the object gets recycled we need to wait for the ongoing uninstall to finish, then reinstall it (it is done as part of the start()) and set up the component back properly.

@blakerouse any thoughts ?

pchila:
buildkite test this

pkg/component/runtime/service.go
blakerouse: It is correct. Once a manager is stopped it is never restarted again. This follows the same pattern -  https://github.com/elastic/elastic-agent/blob/main/pkg/component/runtime/runtime.go#L190; once set it is not unset.

pkg/component/runtime/manager.go
```diff
@@ -705,6 +705,12 @@ func (m *Manager) update(model component.Model, teardown bool) error {
 		var stoppedWg sync.WaitGroup
 		stoppedWg.Add(len(stop))
 		for _, existing := range stop {
+			if existing.getLatest().State == client.UnitStateStopped {
```
blakerouse, line 708: I think instead of making this check here this could be changed to use the atomic that the `componentRuntimeState` is already setting.

https://github.com/elastic/elastic-agent/blob/d316ee0353e08eba4e45cf584efabb1833c358fe/pkg/component/runtime/runtime.go#L190

I think change line 190 in `runtime.go` to check if already set and do nothing would result in the same behavior. Removing the need to hold the lock that is being held when `getLatest()` is being called.

pkg/component/runtime/manager.go
```diff
@@ -705,6 +705,12 @@ func (m *Manager) update(model component.Model, teardown bool) error {
 		var stoppedWg sync.WaitGroup
 		stoppedWg.Add(len(stop))
 		for _, existing := range stop {
+			if existing.shuttingDown.Load() {
+				// it's already shutting down, move on
+				stoppedWg.Done()
+				continue
+			}
+			m.logger.Debugf("Stopping component %q", existing.id)
```
blakerouse, line 713: Why do it here? Why not do it inside of `existing.stop(`? Seems safer, that way if the code is updated some where else to call `existing.stop(` the same logic would apply. Seems weird that it would be checked outside of the function that sets it.

pchila: If you are referring to the log, it's there because that's where I landed when investigating the bug initially.
If you are referring to the stop() itself, the initial fix was only here in the manager without any modification to the service runtime object.
Would you prefer to have only for a service runtime object ?

blakerouse: @pchila No I am referring to this line function https://github.com/elastic/elastic-agent/blob/main/pkg/component/runtime/runtime.go#L189 that is being called from https://github.com/elastic/elastic-agent/blob/main/pkg/component/runtime/manager.go#L708 (which is what your changing here).

This runtime wraps the actual managers. So this change will apply to all managers.

I thinking the function could easily be changed to:

```
func (s *componentRuntimeState) stop(teardown bool, signed *component.Signed) error {
	if s.shuttingDown.Load() {
                // already stopping
		return nil
        }
	s.shuttingDown.Store(true)
	if teardown {
		return s.runtime.Teardown(signed)
	}
	return s.runtime.Stop()
}
```

testing/integration/endpoint_security_test.go
```diff
@@ -394,17 +395,37 @@ func testInstallWithEndpointSecurityAndRemoveEndpointIntegration(t *testing.T, i
...
+		for _, f := range files {
+			if !f.IsDir() {
+				continue
+			}
 
-		t.Log("Found directory", f.Name())
-		require.False(t, strings.Contains(f.Name(), "Endpoint"), "Endpoint directory was not removed")
+			t.Log("Found directory", f.Name())
+			if strings.Contains(f.Name(), "Endpoint") {
+				t.Logf("Check failed: found existing endpoint directory %q", f.Name())
```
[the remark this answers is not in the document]
pchila: This is what allowed us to detect what was left in the endpoint directory at the end of the test. Without it, we wouldn't know if we were left with an empty cache directory or the full endpoint install. It can be removed of course...

pkg/component/runtime/manager.go
pchila: done in 2606c30428f6e43099f7e262351d1cb2bd95570a

changelog/fragments/1695920792-Prevent-multiple-stops-of-services.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Prevent multiple attempts to stop an already stopped service
```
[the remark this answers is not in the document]
pchila: The issue here is that this applies to service runtimes and not to command runtimes...

AndersonQ (commented):
Any chance you could add a test?

blakerouse (approved):
Looks good. Thanks for all the changes.

pkg/component/runtime/runtime.go
```diff
@@ -187,6 +187,10 @@ func (s *componentRuntimeState) start() error {
 }
 
 func (s *componentRuntimeState) stop(teardown bool, signed *component.Signed) error {
+	if s.shuttingDown.Load() {
+		// already stopping
+		return nil
+	}
```
cmacknz, line 193: Can `shuttingDown` be set to false between this `Load` and the following `Store`?

pkg/component/runtime/service.go
cmacknz: If this is the case we should add a comment explaining this.

michalpristas (approved):

pkg/component/runtime/runtime.go
michalpristas: from a component point of view it should be guarded, but from agent architecture standpoint it should not be concurrent and should be safe to have it as it is

pchila:
> Any chance you could add a test?

@AndersonQ  I thought of that but for this precise timing I need a unit or a "narrow" integration test.
If the fix was in the runtime manager as it was initially, that could have been maybe easier. For service runtime I will have to check if I can figure something out without big refactor just for testing.

AndersonQ (approved):

PR #3486, opened 2023-09-28
Update serverless support in integration tests, expand test tooling, add support for beats testing

fearful-symmetry:
## What does this PR do?

This PR "breaks apart" https://github.com/elastic/elastic-agent/pull/3258 into only changes needed to support serverless instances in integration tests, as well as testing with beats, and various expansions to the tooling available to integration tests. Essentially, this is everything except the actual beats serverless tests.

## Why is it important?

We want serverless testing support, and need to support https://github.com/elastic/elastic-agent/pull/3258 

## Checklist


cmacknz:
Let's add a test to ensure that Elastic Agent can ship data to a Serverless project to ensure this code and the Serverless support in general keeps working.

I suggest parameterizing the [monitoring logs integration test](https://github.com/elastic/elastic-agent/blob/main/testing/integration/monitoring_logs_test.go) to run against both stateful and serverless. This test shouldn't require the DSL support since Fleet takes care of that for us.

You will need to provision either an Observability or Security project to include Fleet server, I'm not sure it matters which one you pick.

fearful-symmetry:
@cmacknz You mean a buildkite target or something? The way the tests are set up, we just need to launch the integration tests with the correct environment variables, and they should just run against a serverless instance, unless I'm misunderstanding.

fearful-symmetry:
hmm, ogc seems to be failing with the usual:
```


Traceback (most recent call last):
--


sqlite3.OperationalError: no such table: layouts
```

... I don't think that was something I did?

fearful-symmetry:
Alright, assuming I can get the (existing) CI to pass, I'll add some serverless tests to buildkite and see what happens

fearful-symmetry:
Alright, hacked in some serverless tests. Let's see what that does..

.buildkite/scripts/steps/integration_tests.sh
```diff
@@ -20,6 +20,8 @@ AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true
 set +e
 AGENT_VERSION="${OVERRIDE_TEST_AGENT_VERSION}" TEST_INTEG_CLEAN_ON_EXIT=true SNAPSHOT=true mage integration:test
 TESTS_EXIT_STATUS=$?
+AGENT_VERSION="${OVERRIDE_TEST_AGENT_VERSION}" STACK_PROVISIONER="ess" TEST_INTEG_CLEAN_ON_EXIT=true SNAPSHOT=true mage integration:test
```
cmacknz, line 23: Hmm, I'm not sure I want to run every integration test twice. That is going to cause the tests to take something like 2.5 hours since it will double the run time.

I would only run a basic sanity check, probably only https://github.com/elastic/elastic-agent/blob/main/testing/integration/monitoring_logs_test.go is enough. 

There aren't any actual changes in agent itself for serverless, running the other tests is interesting but it isn't telling us anything about agent, only Fleet.

cmacknz: We could consider moving this to a separate parallel build step. That will speed it up but I still wouldn't want to run all the tests twice.

https://github.com/elastic/elastic-agent/blob/123ba9ce80c9865f72fa3659b5cafe9b51954f49/.buildkite/pipeline.yml#L34-L42

Just duplicating that pipeline and parameterizing this script to take the provisioner type and the mage target to run would work.

Also if you used `ess` to abbreviate `Elasticsearch Serverless` then that is confusing because that abbreviation already means `Elasticsearch service` which is a common name for the stateful cloud offering. We should rename it to something more obvious.

fearful-symmetry: The script also builds the agent, so just running it twice might just slow it down? Unless we refactor the pipeline to do building as a separate step.

I fear we're gonna get an appreciable amount of slowdown no matter what, though. In my experience bringing up the serverless cluster, and waiting for it to return healthy (specifically, waiting for DNS to resolve) takes a non-zero amount of time.

Perhaps we can run some of the pipelines in parallel? At that point we're kind of the outside the scope of this PR, though.

fearful-symmetry:
Alright, recent testing changes broke some stuff in this PR now..

fearful-symmetry:
So, the tests still add around 20 minutes, taken by the 110-ish minutes for everything else.

.buildkite/pipeline.yml
```diff
@@ -41,6 +41,16 @@ steps:
       provider: "gcp"
       machineType: "n1-standard-8"
 
+  - label: "Serverless integration test"
+    key: "serverless-integration-tests"
+    command: ".buildkite/scripts/steps/integration_tests.sh serverless integration:single TestMonitoringLogsShipped"
```
belimawr, line 46: Are we hardcoding a single test here?

magefile.go
```diff
@@ -1749,10 +1752,10 @@ func createTestRunner(matrix bool, singleTest string, goTestFlags string, batche
 	fmt.Printf(">>>> Using %s instance provisioner\n", instanceProvisionerMode)
 	stackProvisionerMode := os.Getenv("STACK_PROVISIONER")
 	if stackProvisionerMode == "" {
-		stackProvisionerMode = "ess"
+		stackProvisionerMode = "stateful"
 	}
-	if stackProvisionerMode != "ess" && stackProvisionerMode != "serverless" {
-		return nil, errors.New("STACK_PROVISIONER environment variable must be one of 'serverless' or 'ess'")
+	if stackProvisionerMode != "stateful" && stackProvisionerMode != "serverless" {
+		return nil, errors.New("STACK_PROVISIONER environment variable must be one of 'serverless' or 'stateful'")
```
belimawr, line 1758: It would be nice to also return the value we got, something like:
```suggestion
		return nil, errors.New("STACK_PROVISIONER environment variable must be one of 'serverless' or 'stateful', not %q", stackProvisionerMode)
```

We can probably find a better wording, but you get the idea.

pkg/testing/ess/serverless.go
```diff
@@ -243,6 +248,38 @@ func (srv *ServerlessClient) WaitForKibana(ctx context.Context) error {
...
+
+	resp, err := http.DefaultClient.Do(resetHandler)
+	if err != nil {
+		return CredResetResponse{}, fmt.Errorf("error performing HTTP request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		p, _ := io.ReadAll(resp.Body)
+		return CredResetResponse{}, fmt.Errorf("Non-201 status code returned by server: %d, body: %s", resp.StatusCode, string(p))
```
belimawr, line 271: Typo? You're checking for a 200 response code, but the message says 201.

```suggestion
		return CredResetResponse{}, fmt.Errorf("Non-200 status code returned by server: %d, body: %s", resp.StatusCode, string(p))
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+
+	apiReq := esapi.SecurityCreateAPIKeyRequest{Body: &buf}
+	resp, err := apiReq.Do(ctx, client)
+	if err != nil {
+		return APIKeyResponse{}, fmt.Errorf("error creating API key: %w", err)
+	}
+	resultBuf, err := handleResponseRaw(resp)
+	if err != nil {
+		return APIKeyResponse{}, fmt.Errorf("error handling HTTP response: %w", err)
+	}
```
belimawr, line 192: `handleResponseRaw` does not close the response body
```suggestion
	defer resp.Body.Close()
	resultBuf, err := handleResponseRaw(resp)
	if err != nil {
		return APIKeyResponse{}, fmt.Errorf("error handling HTTP response: %w", err)
	}
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+
+// GetIndexTemplatesForPattern lists all index templates on the system
+func GetIndexTemplatesForPattern(ctx context.Context, client elastictransport.Interface, name string) (TemplateResponse, error) {
+	req := esapi.IndicesGetIndexTemplateRequest{Name: name}
+	resp, err := req.Do(ctx, client)
+	if err != nil {
+		return TemplateResponse{}, fmt.Errorf("error fetching index templates: %w", err)
+	}
+
+	resultBuf, err := handleResponseRaw(resp)
```
belimawr, line 235: Same thing here
```suggestion
	defer resp.Body.Close()
	resultBuf, err := handleResponseRaw(resp)
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+	req := esapi.IndicesGetDataStreamRequest{Name: []string{namePattern}, ExpandWildcards: "all,hidden"}
+	resp, err := req.Do(ctx, client)
+	if err != nil {
+		return DataStreams{}, fmt.Errorf("error fetching data streams")
+	}
+
+	raw, err := handleResponseRaw(resp)
+	if err != nil {
+		return DataStreams{}, fmt.Errorf("error handling HTTP response for data stream get: %w", err)
+	}
```
belimawr, line 259:
```suggestion
	defer resp.Body.Close()
	raw, err := handleResponseRaw(resp)
	if err != nil {
		return DataStreams{}, fmt.Errorf("error handling HTTP response for data stream get: %w", err)
	}
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+func DeleteIndexTemplatesDataStreams(ctx context.Context, client elastictransport.Interface, name string) error {
+	req := esapi.IndicesDeleteDataStreamRequest{Name: []string{name}, ExpandWildcards: "all,hidden"}
+	resp, err := req.Do(ctx, client)
+	if err != nil {
+		return fmt.Errorf("error deleting data streams: %w", err)
+	}
+	_, err = handleResponseRaw(resp)
+	if err != nil {
+		return fmt.Errorf("error handling HTTP response for data stream delete: %w", err)
+	}
```
belimawr, line 280:
```suggestion
	defer resp.Body.Close()
	_, err = handleResponseRaw(resp)
	if err != nil {
		return fmt.Errorf("error handling HTTP response for data stream delete: %w", err)
	}
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+	patternReq := esapi.IndicesDeleteIndexTemplateRequest{Name: name}
+	resp, err = patternReq.Do(ctx, client)
+	if err != nil {
+		return fmt.Errorf("error deleting index templates: %w", err)
+	}
+	_, err = handleResponseRaw(resp)
+	if err != nil {
+		return fmt.Errorf("error handling HTTP response for index template delete: %w", err)
+	}
+	return nil
```
belimawr, line 291:
```suggestion
	defer resp.Body.Close()
	_, err = handleResponseRaw(resp)
	if err != nil {
		return fmt.Errorf("error handling HTTP response for index template delete: %w", err)
	}
	return nil
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+func GetPipelines(ctx context.Context, client elastictransport.Interface, name string) (map[string]Pipeline, error) {
+	req := esapi.IngestGetPipelineRequest{PipelineID: name}
+	resp, err := req.Do(ctx, client)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching index templates: %w", err)
+	}
+	resultBuf, err := handleResponseRaw(resp)
+	if err != nil {
+		return nil, fmt.Errorf("error handling HTTP response: %w", err)
+	}
```
belimawr, line 304:
```suggestion
	defer resp.Body.Close()
	resultBuf, err := handleResponseRaw(resp)
	if err != nil {
		return nil, fmt.Errorf("error handling HTTP response: %w", err)
	}
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -103,11 +173,158 @@ func GetIndicesWithContext(ctx context.Context, client elastictransport.Interfac
...
+func DeletePipelines(ctx context.Context, client elastictransport.Interface, name string) error {
+	req := esapi.IngestDeletePipelineRequest{PipelineID: name}
+	resp, err := req.Do(ctx, client)
+	if err != nil {
+		return fmt.Errorf("error deleting index template")
+	}
+	_, err = handleResponseRaw(resp)
+	if err != nil {
+		return fmt.Errorf("error handling HTTP response: %w", err)
+	}
```
belimawr, line 324:
```suggestion
	defer resp.Body.Close()
	_, err = handleResponseRaw(resp)
	if err != nil {
		return fmt.Errorf("error handling HTTP response: %w", err)
	}
```

pkg/testing/tools/elasticsearch.go
```diff
@@ -280,15 +473,40 @@ func GetLogsForDatastreamWithContext(ctx context.Context, client elastictranspor
...
+	req := esapi.InfoRequest{}
+	resp, err := req.Do(ctx, client)
+	if err != nil {
+		return Ping{}, fmt.Errorf("error in ping request")
+	}
+
+	respData, err := handleResponseRaw(resp)
+	if err != nil {
+		return Ping{}, fmt.Errorf("error in HTTP response: %w", err)
+	}
```
belimawr, line 490:
```suggestion
	defer resp.Body.Close()
	respData, err := handleResponseRaw(resp)
	if err != nil {
		return Ping{}, fmt.Errorf("error in HTTP response: %w", err)
	}
```

pkg/testing/tools/kibana.go
```diff
@@ -0,0 +1,91 @@
...
+func DeleteDashboard(ctx context.Context, client *kibana.Client, id string) error {
+	// In the future there should be logic to check if we need this header, waiting for https://github.com/elastic/kibana/pull/164850
+	headers := http.Header{}
+	headers.Add("x-elastic-internal-origin", "integration-tests")
+	status, resp, err := client.Connection.Request("DELETE", fmt.Sprintf("/api/saved_objects/dashboard/%s", id), nil, headers, nil)
+	if err != nil {
+		return fmt.Errorf("error making API request: %w, response: '%s'", err, string(resp))
+	}
+
+	if status != 200 {
```
belimawr, line 44: A little suggestion:
```suggestion
	if status != http.StatusOK {
```

pkg/testing/tools/kibana.go
```diff
@@ -0,0 +1,91 @@
...
+	page := 1
+	for {
+		headers := http.Header{}
+		headers.Add("x-elastic-internal-origin", "integration-tests")
+		status, resp, err := client.Connection.Request("GET", "/api/saved_objects/_find", params, headers, nil)
+		if err != nil {
+			return nil, fmt.Errorf("error making api request: %w", err)
+		}
+
+		if status != 200 {
```
belimawr, line 67: here as well
```suggestion
	if status != http.StatusOK {
```

belimawr (changes requested):

.buildkite/pipeline.yml
fearful-symmetry: Yah. @cmacknz 's idea was to just run a single high-level test for serverless, so we can have some coverage without running the entire test suite twice.

cmacknz: Correct, this is intended to be a basic sanity check since nothing in agent itself has to change to support serverless (besides a single additional HTTP header).

We don't need to run the entire 1+ hour test suite a second time for this.

I think a clarifying comment in this file explaining this should be added.

belimawr: Thanks for the clarification folks, it totally makes sense. A comment clarifying it would be great.

belimawr (approved):

.buildkite/pipeline.yml
```diff
@@ -31,9 +31,19 @@ steps:
       manual:
         allowed: true
 
+  - label: "Serverless integration test"
```
cmacknz, line 34: You should add a comment explaining why there is only a single test here. Otherwise nobody who reads this will know.

pkg/testing/fixture.go
```diff
@@ -235,15 +275,20 @@ func ExtractArtifact(l Logger, artifactFile, outputDir string) error {
...
-// If no `states` are provided then the Elastic Agent runs until the context is cancelled.
+// If no `states` are provided then the Elastic Agent runs until the context is or the timeout specified with WithRunLength is reached.
 func (f *Fixture) Run(ctx context.Context, states ...State) error {
-	if f.installed {
+	isBeat := false
+	if f.binaryName != "elastic-agent" {
+		isBeat = true
+	}
+
+	if !isBeat && f.installed {
```
cmacknz, line 291: I think it might be better to implement a `RunBeat` version of this function. Even in places where you haven't special cased on `isBeat` there is code that will never be hit when running a Beat.

For example the `stateErrCh` and `stateCh` cases in the switch statement below on line 364 don't do anything.

I think the duplication is worth not having this become confusing to read and understand. This function is written to launch a subprocess and then control what it does through the control protocol and the second part does not apply to Beats.

I also suspect that as we work more with Beats we may end up with even more specialized logic for Beats since we have no way to easily interact with it, and I don't think we should encourage that to just be bolted onto this implementation as even more special cases.

cmacknz:
I think this was fixed on main with https://github.com/elastic/elastic-agent/pull/3616

This test should also run on Windows and not just Linux, you can remove the OS restriction.

fearful-symmetry:
Alright, so it looks like those windows tests might have been disabled for a reason, because it's failing. Want to see if it's just timing out, or if there's a deeper issue...

fearful-symmetry:
how ironic

 `Failed to connect to backoff(elasticsearch(https://c00f724935b3496cba74c717249987eb.es.us-east-1.aws.staging.elastic.cloud:443)): Connection marked as failed because the onConnect callback failed: Elasticsearch is too old. Please upgrade the instance. If you would like to connect to older instances set output.elasticsearch.allow_older_versions to true. ES=8.11.0, Beat=8.12.0`

pkg/testing/ess/provisioner.go
```diff
@@ -82,8 +82,10 @@ func (p *provisioner) Provision(ctx context.Context, requests []runner.StackRequ
 		results[r] = resp
 	}
 
-	// wait 15 minutes for all stacks to be ready
-	readyCtx, readyCancel := context.WithTimeout(ctx, 15*time.Minute)
+	// set a long timeout
+	// this context travels up to the magefile, clients that want a shorter timeout can set
+	// it via mage's -t flag
+	readyCtx, readyCancel := context.WithTimeout(ctx, 25*time.Minute)
```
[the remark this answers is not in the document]
fearful-symmetry: This is actually for stateful, I ran into a couple timeouts while testing.

testing/integration/monitoring_logs_test.go
```diff
@@ -181,7 +180,7 @@ func findESDocs(t *testing.T, findFn func() (estools.Documents, error)) estools.
 			docs, err = findFn()
 			return err == nil
 		},
-		3*time.Minute,
+		10*time.Minute,
```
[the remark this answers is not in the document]
fearful-symmetry: Did that while trying to diagnose a flaky test, can probably revert it.

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
cmacknz: Did it fix the problem, or does it just wait longer before failing? Sometimes the stack can fail to come up and waiting doesn't fix it, it just makes you wait longer until finding out.

cmacknz (approved):
Changes look good to me, good to merge once the tests pass again. Looks like the failure was in the serverless tests this time.

fearful-symmetry:
Never seen that before.
```
=== FAIL: testing/integration TestMonitoringLogsShipped (2.87s)
--
 (windows-amd64-2022) Test output (sudo) (stdout): unable to create 
policy: missing authentication credentials for REST request [/_mget]: 
security_exception
```

fearful-symmetry:
Sigh, something seems to be breaking the resulting junit files...
```


Parsing /build/TEST-go-integration.xml
--


/usr/local/lib/ruby/gems/3.1.0/gems/rexml-3.2.5/lib/rexml/parsers/treeparser.rb:96:in
 `rescue in parse': #&lt;RuntimeError: Illegal character "\\u0000" in 
raw string "
```

fearful-symmetry:
Downloaded the XML files, they all appear to parse correctly.

fearful-symmetry:
In typical fashion, can't reproduce the failure, which was a setup issue. I'm guessing there was some intermittent timeout or issue setting up auth against the serverless cluster.

PR #3505, opened 2023-10-03
Add changelog for default ES compression

faec:
Changelog-only PR. Adds an entry for enabling compression by default in Elasticsearch outputs, which is an update that is propagating to Agent now that https://github.com/elastic/beats/pull/36681 changed the default for all beats.


cmacknz (approved):

jlind23:
Tests failures are unrelated force merging this then.

jlind23:
@lucabelluccini FYI from 8.11 on we will change the default compression for ES output to 1.

PR #3518, opened 2023-10-04
Fix: Secondary fallback is not used for agent with local agent

michalpristas:
## What does this PR do?

This PR adds a capability to check for fleet server hosted PGP as the last resort in case elastic artifact API is unavailable.
Fleet server will host this PGP per version. any snapshot flags or other version qualifiers are ignored. 

## Why is it important?

Add ability to upgrade securiely in Air gapped environment where fleet server is only reachable URI.

## Checklist

Fixes: #3264


pchila (approved):

cmacknz:
Just to double check, we tested that this version of the change can successfully deploy to cloud with a working Fleet server?

PR #3267, opened 2023-08-17
Updating variable %BRANCH% to repalce hardcoded value main

gizas:
- Bug

## What does this PR do?

Fixes the variable that produces the BRANCH value of daemonsets. The value was hardcoded. 

## Why is it important?

Relates to https://github.com/elastic/elastic-agent/pull/3246/files

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## How to test this PR locally

```bash
cd elastic-agent/deploy/kubernetes

make generate-k8s
```
Test file `elastic-agent/deploy/kubernetes/elastic-agent-stanadlone/elastic-agent-standalone.yaml`
that contains
```yaml
      #        wget -O - https://github.com/elastic/elastic-agent/archive/main.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-main/deploy/kubernetes/elastic-agent-standalone/templates.d"

```

```bash
cd elastic-agent/deploy/kubernetes

GENERATEKUSTOMIZE=tue make ci-create-kustomize
```
Test file `elastic-agent/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset.yaml`

that contains
```yaml
      #        wget -O - https://github.com/elastic/elastic-agent/archive/main.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-main/deploy/kubernetes/elastic-agent-standalone/templates.d"

```

## Related issues

- 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #3523, opened 2023-10-04
update kubernetes deploy yaml to match `make check` output

leehinman:
## What does this PR do?

Changes url to match what `make check` will output.  Will need to reconcile in future.

## Why is it important?

get the build to run

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...

### DIFF
--- a/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml
+++ b/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml
@@ -683,7 +683,7 @@ spec:
       #      - -c
       #      - >-
       #        mkdir -p /etc/elastic-agent/inputs.d &&
-      #        wget -O - https://github.com/elastic/elastic-agent/archive/8.11.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-8.11/deploy/kubernetes/elastic-agent-standalone/templates.d"
+      #        wget -O - https://github.com/elastic/elastic-agent/archive/main.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-main/deploy/kubernetes/elastic-agent-standalone/templates.d"
       #    volumeMounts:
       #      - name: external-inputs
       #        mountPath: /etc/elastic-agent/inputs.d


cmacknz (approved):

deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml
```diff
@@ -683,7 +683,7 @@ spec:
       #      - -c
       #      - >-
       #        mkdir -p /etc/elastic-agent/inputs.d &&
-      #        wget -O - https://github.com/elastic/elastic-agent/archive/8.11.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-8.11/deploy/kubernetes/elastic-agent-standalone/templates.d"
+      #        wget -O - https://github.com/elastic/elastic-agent/archive/main.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-main/deploy/kubernetes/elastic-agent-standalone/templates.d"
```
cmacknz, line 686: Something seems to be wrong with how these YAML files are templated, everything on the 8.11 branch references main which doesn't seem right.

cmacknz:
Merging the buildkite failure is unrelated, will be fixed by https://github.com/elastic/elastic-agent/pull/3522

deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml
gizas: @cmacknz I had a small fix here: https://github.com/elastic/elastic-agent/pull/3267

So according to this,:
```
cd elastic-agent/deploy/kubernetes
make generate-k8s
```
And should have done it.

gizas: @cmacknz found the reason, the [makefile](https://github.com/leehinman/elastic-agent/blob/894965abd22547f8d399e3721a76f36289fec1a5/deploy/kubernetes/Makefile#L4) reads the https://github.com/leehinman/elastic-agent/blob/894965abd22547f8d399e3721a76f36289fec1a5/version/docs/version.asciidoc#L2

So is this expected is not it?

cmacknz: Nothing ever updates `doc-branch` in the agent repository since the docs are not published from this repository like they are in Beats. I suspect this file was originally copied from Beats without having the unnecessary parts removed.

The `stack-version` field is updated by our release automation, so you could use that instead. The only modification will be to strip the patch version if the intent is to download the latest version of the 8.11 branch. Using a tagged version would also work and probably be more stable, but I don't think that's what is happening today.

PR #3542, opened 2023-10-05
Refactoring HTTP downloader progress reporter to accept multiple observers

ycombinator:
## What does this PR do?

Prior to this PR, the HTTP downloader's `progressReporter` would report download progress in Agent logs. This functionality was implemented entirely within the `progressReporter` code.

This PR refactors the `progressReporter` code to accept multiple observers and introduces a new observer, `loggingProgressObserver`, that logs the observed progress in Agent logs, preserving the behavior prior to this PR. 

## Why is it important?

This PR is purely a refactoring PR; it does not change any functionality.  Functionally, the HTTP downloader's progress will continue to be reported in Agent logs. 

However, in https://github.com/elastic/elastic-agent/pull/3527, we will need the `progressReporter` to also report progress to another observer.  This PR makes it easy to accomplish that in a clean manner.  More context here: https://github.com/elastic/elastic-agent/pull/3527#discussion_r1347016802

## Checklist

## Related issues

- Relates to https://github.com/elastic/elastic-agent/pull/3527#discussion_r1347016802

### DIFF
--- a/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
@@ -206,9 +206,9 @@ func (e *Downloader) downloadFile(ctx context.Context, artifactName, filename, f
 		}
 	}
 
-	plo := newLoggingProgressObserver(e.log, e.config.HTTPTransportSettings.Timeout)
+	lpObs := newLoggingProgressObserver(e.log, e.config.HTTPTransportSettings.Timeout)
 	reportCtx, reportCancel := context.WithCancel(ctx)
-	dp := newDownloadProgressReporter(sourceURI, e.config.HTTPTransportSettings.Timeout, fileSize, plo)
+	dp := newDownloadProgressReporter(sourceURI, e.config.HTTPTransportSettings.Timeout, fileSize, lpObs)
 	dp.Report(reportCtx)
 	_, err = io.Copy(destinationFile, io.TeeReader(resp.Body, dp))
 	if err != nil {
--- a/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
@@ -206,7 +206,7 @@ func (e *Downloader) downloadFile(ctx context.Context, artifactName, filename, f
 		}
 	}
 
-	plo := newLoggerProgressObserver(e.log, e.config.HTTPTransportSettings.Timeout)
+	plo := newLoggingProgressObserver(e.log, e.config.HTTPTransportSettings.Timeout)
 	reportCtx, reportCancel := context.WithCancel(ctx)
 	dp := newDownloadProgressReporter(sourceURI, e.config.HTTPTransportSettings.Timeout, fileSize, plo)
 	dp.Report(reportCtx)
--- a/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
@@ -17,14 +17,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/docker/go-units"
-
-	"github.com/elastic/elastic-agent-libs/atomic"
 	"github.com/elastic/elastic-agent-libs/transport/httpcommon"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact/download"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/errors"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
 const (
@@ -46,13 +44,13 @@ const (
 
 // Downloader is a downloader able to fetch artifacts from elastic.co web page.
 type Downloader struct {
-	log    progressLogger
+	log    *logger.Logger
 	config *artifact.Config
 	client http.Client
 }
 
 // NewDownloader creates and configures Elastic Downloader
-func NewDownloader(log progressLogger, config *artifact.Config) (*Downloader, error) {
+func NewDownloader(log *logger.Logger, config *artifact.Config) (*Downloader, error) {
 	client, err := config.HTTPTransportSettings.Client(
 		httpcommon.WithAPMHTTPInstrumentation(),
 		httpcommon.WithKeepaliveSettings{Disable: false, IdleConnTimeout: 30 * time.Second},
@@ -66,7 +64,7 @@ func NewDownloader(log progressLogger, config *artifact.Config) (*Downloader, er
 }
 
 // NewDownloaderWithClient creates Elastic Downloader with specific client used
-func NewDownloaderWithClient(log progressLogger, config *artifact.Config, client http.Client) *Downloader {
+func NewDownloaderWithClient(log *logger.Logger, config *artifact.Config, client http.Client) *Downloader {
 	return &Downloader{
 		log:    log,
 		config: config,
@@ -208,8 +206,9 @@ func (e *Downloader) downloadFile(ctx context.Context, artifactName, filename, f
 		}
 	}
 
+	plo := newLoggerProgressObserver(e.log, e.config.HTTPTransportSettings.Timeout)
 	reportCtx, reportCancel := context.WithCancel(ctx)
-	dp := newDownloadProgressReporter(e.log, sourceURI, e.config.HTTPTransportSettings.Timeout, fileSize)
+	dp := newDownloadProgressReporter(sourceURI, e.config.HTTPTransportSettings.Timeout, fileSize, plo)
 	dp.Report(reportCtx)
 	_, err = io.Copy(destinationFile, io.TeeReader(resp.Body, dp))
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ycombinator:
While working on this PR, I found a bug with the progress reporter.  It was reporting progress just once (after a delay) instead of reporting it periodically.  As such, this PR here is now blocked on the bugfix PR: https://github.com/elastic/elastic-agent/pull/3548.

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
```diff
@@ -208,8 +206,9 @@ func (e *Downloader) downloadFile(ctx context.Context, artifactName, filename, f
 		}
 	}
 
+	lpObs := newLoggingProgressObserver(e.log, e.config.HTTPTransportSettings.Timeout)
```
michalpristas, line 209: wild name, i had to go up here to figure it what it means

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (dp *downloadProgressReporter) Report(ctx context.Context) {
+	started := time.Now()
+	dp.started = started
+	sourceURI := dp.sourceURI
+	length := dp.length
+	interval := dp.interval
+
+	go func() {
```
michalpristas, line 54: very unnecessary optimization from my side: no need to spin up goroutine or initiate tickers if progressObservers is []. as we're having at least log now, i consider this just a small optimization from component side of view.

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
```diff
@@ -0,0 +1,103 @@
...
+		t := time.NewTicker(interval)
+		defer t.Stop()
+		for {
+			select {
+			case <-ctx.Done():
```
michalpristas, line 59: we're counting on external elements to cancel this. we could introduce Done chan internally which would be closed on ReportComplete or ReportFailed

michalpristas (approved):
just a very tiny comments, it looks good otherwise

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func (dp *downloadProgressReporter) Write(b []byte) (int, error) {
+	n := len(b)
+	dp.downloaded.Add(n)
+	return n, nil
+}
+
+func (dp *downloadProgressReporter) Report(ctx context.Context) {
```
michalpristas, line 47: make a comment that called is responsible for cancelling context

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
```diff
@@ -46,13 +44,13 @@ const (
 
 // Downloader is a downloader able to fetch artifacts from elastic.co web page.
 type Downloader struct {
-	log    progressLogger
```
pchila, line 49: Nit: I always prefer the logger as interfaces for mocking and loose coupling, instead of relying on a more concrete type

internal/pkg/agent/application/upgrade/artifact/download/http/downloader_test.go
```diff
@@ -253,21 +237,21 @@ func assertLogs(t *testing.T, logs []logMessage, minExpectedProgressLogs int, ex
 	// Verify that the first minExpectedProgressLogs messages are about the download progress (for the first file).
 	i := 0
 	for ; i < minExpectedProgressLogs; i++ {
-		assert.Equal(t, logs[i].record, expectedProgressMsg)
+		assert.Regexp(t, expectedProgressRegexp, logs[i].Message)
```
pchila, line 240: If there are other logs mixed in with the progress logs, this assert will fail the test. Is this intentional?

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: Good point, will add, thanks!

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: Done in 89ca86b27d49e1cfa981f9156886689ffc9d1218.

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
ycombinator: Fixed in e9fc0970bac5927bd135c4f0a78b36b74d1eae5a.

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: Added in da06f8c9b.

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
```diff
@@ -0,0 +1,110 @@
...
+		length:            float64(length),
+		progressObservers: progressObservers,
+	}
+}
+
+func (dp *downloadProgressReporter) Write(b []byte) (int, error) {
+	n := len(b)
+	dp.downloaded.Add(n)
+	return n, nil
+}
```
pchila, line 46: Why the downloadProgressReporter needs to implement a `Writer` interface that only increments an integer ?

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: Implemented your suggestion in 67f80f63f.  Let me know what you think.

pchila: what about creating a subcontext that can be closed when the request is complete or cancelled (ofc we need to hold a ref to its cancellation function) ?

pchila (commented):
A few nits, nothing major.
Left a couple of questions...

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
ycombinator: In general, I agree but I'm also a fan of YAGNI.  We can easily change this back to an interface if/when we need it, i.e. have more than one implementation.

internal/pkg/agent/application/upgrade/artifact/download/http/downloader_test.go
ycombinator: Good point.  As things stand, the only other log that could get mixed up is this one: https://github.com/elastic/elastic-agent/blob/359be4f18ef1104f60e9199aca268d5dcbfd77f1/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go#L102-L104

This log _should_ never get triggered but I suppose in some strange circumstances the `os.Remove` could fail. 

Any suggestions on how to make this more robust to only consider progress logs? Maybe the logs could first be filtered to only keep ones that match `expectedProgressRegexp` or `expectedCompletedRegexp` before running any assertions?

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: This being a refactoring PR, I didn't want to change the implementation too much.  So I didn't touch this.  However, the answer lies in how the progress reporter is informed of a download's progress, over here: https://github.com/elastic/elastic-agent/blob/359be4f18ef1104f60e9199aca268d5dcbfd77f1/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go#L214

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: > what about creating a subcontext that can be closed when the request is complete or cancelled (ofc we need to hold a ref to its cancellation function) ?

I like this pattern better as it doesn't require the consumer to remember to necessarily call either `ReportComplete` or `ReportFailed`.  Basically, it gives the consumer two options on how to cancel: either cancel the `ctx` passed to `Report` or call `ReportComplete`/`ReportFailed`. 

As far as implementation goes, we could do a subcontext or we use an internal `done` channel as implemented in https://github.com/elastic/elastic-agent/commit/67f80f63fe8c8359e381bc3233234c99ce4f1c41. Is one implementation necessarily better than the other?

pchila: The context cancellation is more "abandon processing" and the done channel is the simple solution to unblock a go routine.

The context has the advantage of propagation through the subcontexts if the parent context is cancelled/closed/expires, the done channel has to be managed/closed explicitly.

Since you have already implemented the `done` channel you can keep that, I am just quite surprised to see that we only send an empty struct instead of closing the channel and removing the reference

internal/pkg/agent/application/upgrade/artifact/download/http/downloader_test.go
pchila: Filtering and then asserting length is surely one option. The exact semantics depend on what we are trying to assert here: from the code you are asserting that at least the first `minExpectedProgressLogs` log entries must be the ones you expect: this will break as soon as an extra log entry is in the mix (it may even be added in the future).

Did you mean to assert that among the collected log entries you have at least `minExpectedProgressLogs`  expected logs ? In this case filter/count and then assert

ycombinator: Let me put up a commit showing what I meant by filtering and see what you think.

blakerouse (approved):
This looks good to me. I tried to find something to complain about, I was unsuccessful. ;-)

internal/pkg/agent/application/upgrade/artifact/download/http/progress_reporter.go
ycombinator: @pchila @michalpristas I updated the implementation such that there are now two ways to return from the goroutine inside `Report`:
* `Report` is passed a context.  When the context is done, the goroutine will return.
* When the consumer calls `ReportComplete` or `ReportFailed`, we close an internal `done` channel, which will also cause the goroutine in `Report` to return.

internal/pkg/agent/application/upgrade/artifact/download/http/downloader_test.go
ycombinator: @pchila I implemented the filtering in a464077ad.  Let me know what you think.

pchila (approved):

PR #3554, opened 2023-10-06
install fails if enroll fails and surface errors

AndersonQ:
For the agent to be actually enrolled it needs to restart after the enroll process is completed, so it'll pickup the new config and "connect" to fleet-server.

## What does this PR do?

## Why is it important?

For the agent to be actually enrolled it needs to restart after the enroll process is completed, so it'll pickup the new config and "connect" to fleet-server.

## Checklist

## Related issues

- N/A

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/cmd/enroll_cmd.go
+++ b/internal/pkg/agent/cmd/enroll_cmd.go
@@ -460,6 +460,7 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 	}
 
 	signal := make(chan struct{})
+	defer close(signal)
 	backExp := backoff.NewExpBackoff(signal, 10*time.Second, 1*time.Minute)
 
 	var i int
@@ -467,15 +468,18 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 		backExp.Wait()
 		c.log.Info("Retrying to restart...")
 		err = c.daemonReload(ctx)
-		if err == nil ||
-			errors.Is(err, context.DeadlineExceeded) ||
-			errors.Is(err, context.Canceled) {
+		if err != nil &&
+			(errors.Is(err, context.DeadlineExceeded) ||
+				errors.Is(err, context.Canceled)) {
+			return fmt.Errorf("could not reload deamon after %d retries: %w",
+				i+1, err)
+		}
+		if err == nil {
 			break
 		}
 	}
 
-	close(signal)
-	return fmt.Errorf("could not reload deamon after %d retries: %w", i+1, err)
+	return nil
 }
 
 func (c *enrollCmd) daemonReload(ctx context.Context) error {
--- a/internal/pkg/agent/cmd/enroll_cmd.go
+++ b/internal/pkg/agent/cmd/enroll_cmd.go
@@ -172,7 +172,7 @@ func newEnrollCmd(
 	)
 }
 
-// newEnrollCmdWithStore creates an new enrollment and accept a custom store.
+// newEnrollCmdWithStore creates a new enrollment and accept a custom store.
 func newEnrollCmdWithStore(
 	log *logger.Logger,
 	options *enrollCmdOption,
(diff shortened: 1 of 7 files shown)


internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -159,14 +162,24 @@ func TestEnroll(t *testing.T) {
...
+			defer cancel()
+			err = cmd.Execute(ctx, streams)
+
+			if err != nil &&
+				// There is no agent running, therefore nothing to be restarted.
+				// However, this will cause the Enroll command to return an error
+				// which we'll ignore here.
+				!strings.Contains(err.Error(),
+					"could not reload agent daemon, unable to trigger restart") {
+				t.Fatalf("enrrol coms returned and unexpected error: %v", err)
```
ycombinator, line 175: Typos?

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -443,24 +450,35 @@ func (c *enrollCmd) prepareFleetTLS() error {
 
 func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
```
ycombinator, line 451: I wonder if it would be possible to simplify the implementation of this function a bit.

Right now, we call `c.daemonReload(ctx)`, do some error checking, and then enter the backoff loop.  Inside the backoff loop we again call `c.daemonReload(ctx)` and do some error checking.  Would it be possible to consolidate and directly call the backoff loop, with one extra iteration and moving the `backExp.Wait()` to the end of each iteration instead of the start?

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -267,17 +268,23 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
...
 	}()
 
 	if c.agentProc == nil {
-		if err := c.daemonReload(ctx); err != nil {
-			c.log.Infow("Elastic Agent might not be running; unable to trigger restart", "error", err)
-		} else {
-			c.log.Info("Successfully triggered restart on running Elastic Agent.")
+		if err = c.daemonReloadWithBackoff(ctx); err != nil {
+			c.log.Errorf("Elastic Agent might not be running; unable to trigger restart: %v", err)
+			return fmt.Errorf("could not reload agent daemon, unable to trigger restart: %w", err)
```
ycombinator, line 281: There are a number of small fixes in this PR so just want to double check: this line here is the actual fix to the bug, correct?

internal/pkg/agent/install/perms_unix.go
```diff
@@ -18,19 +19,26 @@ func fixPermissions(topPath string) error {
...
-		if err == nil {
-			// all files should be owned by root:root
-			err = os.Chown(name, 0, 0)
-			if err != nil {
-				return err
-			}
-			// remove any world permissions from the file
-			err = os.Chmod(name, info.Mode().Perm()&0770)
-		} else if errors.Is(err, fs.ErrNotExist) {
+func recursiveRootPermissions(root string) error {
```
ycombinator, line 22: Correct me if I'm wrong but the changes to this function (and some other minor changes in this PR) are unrelated to the actual bug being fixed?  In the future, could you please pull such refactoring/cleanup changes into their own PR please? It makes things easier to review and, should a revert be needed, it makes it easier to revert just the relevant changes.

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -159,14 +162,24 @@ func TestEnroll(t *testing.T) {
 			require.NoError(t, err)
 
 			streams, _, _, _ := cli.NewTestingIOStreams()
-			err = cmd.Execute(context.Background(), streams)
-			require.NoError(t, err)
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+			defer cancel()
+			err = cmd.Execute(ctx, streams)
+
+			if err != nil &&
```
ycombinator, line 169: Given the 1 minute timeout context being passed to `cmd.Execute` above, are we _guaranteed_ for `cmd.Execute` to fail with a `context.DeadlineExceeded` error in this test?

internal/pkg/agent/cmd/enroll_cmd.go
pchila: It would be even better if the retry implementation took care of the loop and called a `func() error` on its own till success or deadline/timeout/max attempts. We could separate the concerns of retry from the actual action being performed

pchila (approved):
Aside from the usual backoff awkwardness (not the point of the PR) there are just a couple of small nitpicks.
I agree with @ycombinator  that, since we squash commits it would be better to separate small typo corrections or refactors in a dedicated PR

ycombinator (approved):

PR #3556, opened 2023-10-06
Fixing script to read ascii code

gizas:
-- Bug

## What does this PR do?

WHAT: Updates makefile to read from different value of version.doc file
WHY: The `doc-branch` is a field that no longer is udpated. Thus we replace it with `stack-version` field

## Why is it important?

To avoid wrong references to main and alwasy reference to latest release of templates

## Checklist

1. Checkout PR
2. `cd elastic-agent/deploy/kubernetes`
3. Run make `make generate-k8s`

See in `elastic-agent-standalone-kubernetes.yaml`
```yaml
      #        wget -O - https://github.com/elastic/elastic-agent/archive/8.8.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-8.8/deploy/kubernetes/elastic-agent-standalone/templates.d"
```

- Relates https://github.com/elastic/elastic-agent/pull/3523


gizas:
@cmacknz   I move our discussion here: https://github.com/elastic/elastic-agent/pull/3523#discussion_r1347650550

Only thing is that `stack-version` still seems not updated in main branch. See here: https://github.com/elastic/elastic-agent/blob/main/version/docs/version.asciidoc#L1

Is this expected?

deploy/kubernetes/Makefile
```diff
@@ -1,7 +1,7 @@
 ALL=elastic-agent-standalone elastic-agent-managed
 
 BEAT_VERSION=$(shell grep '^:stack-version:' ../../version/docs/version.asciidoc | cut -c 17- )
-BRANCH_VERSION=$(shell grep '^:doc-branch:' ../../version/docs/version.asciidoc | cut -c 14- )
```
cmacknz, line 4: It appears the `:stack-version` key here isn't being updated properly.

The one you want and is guarnateed to be right is https://github.com/elastic/elastic-agent/blob/99b14c8acc63e65c710dca19e2115200209ff314/version/version.go#L7

There are a couple of examples for parsing this you could reuse. For one example see https://github.com/elastic/elastic-agent/blob/99b14c8acc63e65c710dca19e2115200209ff314/.buildkite/scripts/common.sh#L16

You'd still have to drop the patch version.

cmacknz: Ah I also now see why we were trying to use doc-branch because this needs to be `main` and not `8.12` for main.

I would suggest `git branch --show-current` but that will break when this is run on developer branches.

deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml
```diff
@@ -683,7 +683,7 @@ spec:
       #      - -c
       #      - >-
       #        mkdir -p /etc/elastic-agent/inputs.d &&
-      #        wget -O - https://github.com/elastic/elastic-agent/archive/main.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-main/deploy/kubernetes/elastic-agent-standalone/templates.d"
+      #        wget -O - https://github.com/elastic/elastic-agent/archive/8.8.tar.gz | tar xz -C /etc/elastic-agent/inputs.d --strip=5 "elastic-agent-8.8/deploy/kubernetes/elastic-agent-standalone/templates.d"
```
cmacknz, line 686: All of the files probably should have updated, not just one.

deploy/kubernetes/Makefile
gizas: I changed it to read from version.go

Both managed and standalone updated

cmacknz (approved):

MichaelKatsoulis (approved):

PR #3563, opened 2023-10-06
Remove the agent version override in tests.

cmacknz:
We now have working 8.12.0, 8.11.0, and 8.10.3 snapshots.

### DIFF
--- a/.buildkite/scripts/steps/integration_tests.sh
+++ b/.buildkite/scripts/steps/integration_tests.sh
@@ -5,8 +5,8 @@ source .buildkite/scripts/common.sh
 
 # Override the agent package version using a string with format <major>.<minor>.<patch>
 # NOTE: use only after version bump when the new version is not yet available, for example:
-# OVERRIDE_AGENT_PACKAGE_VERSION="8.10.3"
-OVERRIDE_AGENT_PACKAGE_VERSION="8.10.2"
+# OVERRIDE_AGENT_PACKAGE_VERSION="8.10.3" otherwise OVERRIDE_AGENT_PACKAGE_VERSION="".
+OVERRIDE_AGENT_PACKAGE_VERSION=""
 
 if [[ -n "$OVERRIDE_AGENT_PACKAGE_VERSION" ]]; then
   OVERRIDE_TEST_AGENT_VERSION=${OVERRIDE_AGENT_PACKAGE_VERSION}"-SNAPSHOT"
--- a/internal/pkg/agent/application/upgrade/artifact/download/snapshot/downloader.go
+++ b/internal/pkg/agent/application/upgrade/artifact/download/snapshot/downloader.go
@@ -88,6 +88,11 @@ func snapshotConfig(config *artifact.Config, versionOverride *agtversion.ParsedS
 }
 
 func snapshotURI(versionOverride *agtversion.ParsedSemVer, config *artifact.Config) (string, error) {
+	// Respect a non-default source URI even if the version is a snapshot.
+	if config.SourceURI != artifact.DefaultSourceURI {
+		return "", fmt.Errorf("skip snapshot download due to non-default sourceURI %s", config.SourceURI)
+	}
+
 	// snapshot downloader is used also by the 'localremote' impl in case of agent currently running off a snapshot build:
 	// the 'localremote' downloader does not pass a specific version, implying that we should update to the latest snapshot
 	// build of the same <major>.<minor>.<patch>-SNAPSHOT version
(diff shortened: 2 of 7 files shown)


pchila:
buildkite test this

pchila (changes requested):
The change is ok but the latest buildkite run is worrying: https://buildkite.com/elastic/elastic-agent/builds/3864

The `TestAPM` and `TestStandaloneUpgradeWithGPGFallback` fail again (symptoms look very similar to the fleet bootstrap issue we already experienced during 8.11.0 FF)

I relaunched an integration test run but this should not be merged until we get a clean run

faec:
It looks like the explicit test failure is trying to grab a snapshot URL that doesn't exist (maybe it's getting the hash wrong?) but I also noticed this line which looks weird to me:

```
>>> (linux-amd64-ubuntu-2204) Test output (sudo) (stdout): upgrader.go:190: Upgrading from version "8.12.0-SNAPSHOT" to version "8.11.0"
```

Is this an intentional "upgrade" to a lower version, or is the test misbehaving?

cmacknz:
> the GPG ones look likethey are coming from 8.11.0 not being released but being selected as a version for the downgrade

When we call https://artifacts-api.elastic.co/v1/versions/ it reports that 8.11 is a known version which isn't actually true. Possibly this is because an 8.11 BC exists and we would need to download that.

```json
{"versions":["7.17.10","7.17.11","7.17.12","7.17.13","7.17.14-SNAPSHOT","7.17.14","8.7.1","8.8.0","8.8.1","8.8.2","8.9.0","8.9.1","8.9.2","8.10.0-SNAPSHOT","8.10.0","8.10.1-SNAPSHOT","8.10.1","8.10.2-SNAPSHOT","8.10.2","8.10.3-SNAPSHOT","8.10.3","8.11.0-SNAPSHOT","8.11.0","8.12.0-SNAPSHOT"],"aliases":["7.17-SNAPSHOT","7.17","8.7","8.8","8.9","8.10-SNAPSHOT","8.10","8.11-SNAPSHOT","8.11","8.12-SNAPSHOT"],"manifests":{"last-update-time":"Tue, 10 Oct 2023 18:24:10 UTC","seconds-since-last-update":14}}
```

The problem is it comes from https://staging.elastic.co/8.11.0-8f52d632/summary-8.11.0.html#elastic-agent which we have no special handling for in the agent. This is because there is no obvious qualifier in the version itself to tell us to look in staging like there is for snapshots:

https://github.com/elastic/elastic-agent/blob/7e653ebb09692e4dcf226d8699fd733593eb29dd/internal/pkg/agent/application/upgrade/artifact/download/snapshot/downloader.go#L23

We can either unconditionally use the snapshot version or look at the https://artifacts-api.elastic.co/v1/versions/8.11 to confirm the URL to use. I don't see a reason not to use snapshots since BCs have a larger time window between fixes if we need to fix a problem.

cmacknz:
I think what we can do is check if the highest version number only has a snapshot available (8.12.0-SNAPSHOT). If it does it is unreleased, and the previous version must also be a snapshot to guarantee it exists.

The only caveat to this is we would prefer 8.11.1-SNAPSHOT to 8.11.0 once 8.11.0 is actually released, but I think that is actually preferable if we are still actively iterating on the code for the 8.11 branch.

cmacknz:
And now 8.12.0-SNAPSHOT is broken.

cmacknz:
buildkite test it

pchila (approved):
The only thing missing is a successful run of integration tests...

cmacknz:
The RetryDownload upgrade test appears to be failing because of a bug where the the updated source URI is ignored in favour of the snapshot path. Probably this has to do with the fact that a dedicated snapshot downloader exists and gets tried before the http one so this will always fail for snapshots.

```json
{"log.level":"info","@timestamp":"2023-10-12T21:13:48.976Z","log.origin":{"file.name":"upgrade/step_download.go","file.line":59},"message":"Downloading upgrade artifact","log":{"source":"elastic-agent"},"version":"8.11.0-SNAPSHOT","source_uri":"http://localhost:35383","drop_path":"","target_path":"/opt/Elastic/Agent/data/elastic-agent-f0c25c/downloads","install_path":"/opt/Elastic/Agent/data/elastic-agent-f0c25c/install","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-10-12T21:13:48.976Z","log.origin":{"file.name":"upgrade/step_download.go","file.line":188},"message":"download attempt 1","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-10-12T21:14:26.665Z","log.origin":{"file.name":"http/downloader.go","file.line":319},"message":"download from https://snapshots.elastic.co/8.11.0-60b804d4/downloads/beats/elastic-agent/elastic-agent-8.11.0-SNAPSHOT-linux-arm64.tar.gz completed in 37 seconds @ 17.52MBps","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-10-12T21:14:26.727Z","log.origin":{"file.name":"http/downloader.go","file.line":319},"message":"download from https://snapshots.elastic.co/8.11.0-60b804d4/downloads/beats/elastic-agent/elastic-agent-8.11.0-SNAPSHOT-linux-arm64.tar.gz.sha512 completed in Less than a second @ +InfYBps","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
```

PR #3139, opened 2023-07-27
Revert "Agent owns its directories"

michalpristas:
Reverts elastic/elastic-agent#3084

Relates:
- https://github.com/elastic/elastic-agent/issues/82

PR #3576, opened 2023-10-10
[Fix] Agent incapable of running on Azure Container Instances

michalpristas:
Bringing back #3139 

What this change is introducing on top of bringing back work introduced in #3084 is change of ordrer for some operations.

Changing owner of a file, discards capabilities set. This becomes a problem with heartbeat as it needs `setuid` and `netraw` capabilities to perform properly. 
So setting capabilities was moved after chown.

### How to test
- build docker image
- deploy to Azure Container Instances to see it running (it cannot fail)
- check capabilities `docker run --rm -it --entrypoint /bin/bash {image_id} -c "getcap data/elastic-agent-*/components/heartbeat"` -> this needs to have output (capabilities mentioned earlier)

I also checked with @justinkambic to see if we can test this with their failing E2E but it seems we need to merge it to see it fail. But it won't fail. OK?

Fixes: #82


dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -154,6 +144,16 @@ RUN chown {{ .user }} /app
 {{- end }}
 {{- end }}
 
+RUN setcap cap_net_raw,cap_setuid+p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/components/heartbeat && \
+{{- if .linux_capabilities }}
+# Since the beat is stored at the other end of a symlink we must follow the symlink first
+# For security reasons setcap does not support symlinks. This is smart in the general case
+# but in our specific case since we're building a trusted image from trusted binaries this is
+# fine. Thus, we use readlink to follow the link and setcap on the actual binary
+    readlink -f {{ $beatBinary }} | xargs setcap {{ .linux_capabilities }} && \
```
pchila, line 153: if we are reading a single link, can't we avoid xargs?
```suggestion
   setcap {{ .linux_capabilities }}  $(readlink -f {{ $beatBinary }}) && \
```

michalpristas: will test this to be sure but we don't use this for agent

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -127,25 +126,16 @@ COPY --from=home {{ $beatHome }}/NOTICE.txt /licenses
...
-
 {{- if eq .user "root" }}
 {{- if contains .image_name "-cloud" }}
 # Generate folder for a stub command that will be overwritten at runtime
 RUN mkdir /app
 {{- end }}
 {{- else }}
-RUN groupadd --gid 1000 {{ .BeatName }}
-RUN useradd -M --uid 1000 --gid 1000 --groups 0 --home {{ $beatHome }} {{ .user }}
+RUN groupadd --gid 1000 {{ .BeatName }} && \
```
cmacknz, line 135: Add a comment documenting that the chown resets capabilities and that the order of commands is important.

cmacknz (approved):

cmacknz:
Re-opening this as the fix needed to be reverted due to problems with ECE deployments. See https://github.com/elastic/elastic-agent/issues/3711.

PR #3580, opened 2023-10-10
Add warning messages to magefile for version mismatches in beats

fearful-symmetry:
## What does this PR do?

This adds a few warning lines to the magefile to handle cases where there's a version mismatch between the agent repo and the beat repo. Shot myself in the foot a few times with this, so it will just quit if it detects a version mismatch.

## Checklist


belimawr (commented):
LGTM, however I would make that an error and interrupt the packaging process.

I've shoot myself in the foot because of this as well. If we need mismatched versions for testing/development, we know what we're doing and we can just modify the version files.

As a general behaviour I rather error on the safe side and not generate an inconsistent package.

magefile.go
```diff
@@ -1008,6 +1008,15 @@ func packageAgent(platforms []string, packagingFn func()) {
 					targetPath := filepath.Join(archivePath, rp)
 					os.MkdirAll(targetPath, 0755)
 					for _, f := range files {
+						// safety check; if the user has an older version of the beats repo,
+						// for example right after a relase where you've `git pulled` from on repo and not the other,
```
ycombinator, line 1012: Typo.

```suggestion
						// for example right after a release where you've `git pulled` from on repo and not the other,
```

magefile.go
```diff
@@ -1008,6 +1008,15 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+						// they might end up with a mishmash of packages from different versions.
+						// check to see if we have mismatched versions.
+						if !strings.Contains(f, packageVersion) {
+							// if this panic hits weird edge cases where we don't want actual failures, revert to a printf statement.
+							panic(fmt.Sprintf("the file %s doesn't match agent version %s, beats might be out of date", f, packageVersion))
```
ycombinator, line 1017:
```suggestion
							panic(fmt.Sprintf("the file %s doesn't match agent version %s, beats repo might be out of date", f, packageVersion))
```

magefile.go
```diff
@@ -1017,8 +1026,8 @@ func packageAgent(platforms []string, packagingFn func()) {
 				}
 				// a very basic footcannon protector; if packages are missing and we need to rebuild them, check to see if those files were copied
 				// if we needed to repackage beats but still somehow copied nothing, could indicate an issue. Usually due to beats and agent being at different versions.
-				if packagesMissing && packagesCopied == 0 {
-					fmt.Printf(">>> WARNING: no packages were copied, but we repackaged beats anyway. Check binary to see if intended beats are there.")
+				if packagesCopied == 0 {
+					fmt.Printf(">>> WARNING: no packages were copied, but we repackaged beats anyway. Check binary to see if intended beats are there.\n")
```
ycombinator, line 1030:
```suggestion
					fmt.Println(">>> WARNING: no packages were copied, but we repackaged beats anyway. Check binary to see if intended beats are there.")
```

ycombinator (approved):
A few minor comments but overall LGTM!

PR #3598, opened 2023-10-12
[Non-Root] Add support for running without root permissions on Linux

blakerouse:
## What does this PR do?

Adds the ability to install the elastic-agent where it runs as its own elastic-agent user on Linux. Limited to Linux at the moment just to keep the PR smaller.

## Why is it important?

Allows the ability to install an Elastic Agent that doesn't require root access.

**Note:** Root is still required to install the Elastic Agent, just adds support for `--unprivileged` option.

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (not possible to unit test, would affect the actual system)
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~ (will add changelog when its enabled on all platforms)
- [x] I have added an integration test or an E2E test (TODO)

## Author's Checklist

- [x] Installs and runs
- [x] All files owned by `elastic-agent:elastic-agent`
- [x] Control socket can be used when user is added to `elastic-agent` group

```
$ sudo ./elastic-agent install --unprivileged
$ sudo usermod -a -G elastic-agent $USER
$ newgrp elastic-agent
$ elastic-agent status
```


blakerouse:
This PR enables installation and running as non-root. It does not enable upgrading as non-root, there is an issue with the watcher where it tries to make calls to the service runtime to watch the PID. A non-root user doesn't have that ability to perform those actions. A solution is needed for this problem.

blakerouse:
Moving in draft as I work on a watcher change required for non-root upgrading to work.

blakerouse:
Currently hitting issue with Elastic Agent not showing or using the correct build when testing self-upgrade with local build using `file://` as the `--source-uri`. Working on this PR to fix the issue and provide a better experience using a local file for upgrades.

https://github.com/elastic/elastic-agent/pull/3682

blakerouse:
**Note**: I cannot add an upgrade test, until latest SNAPSHOT exists with this change in it. This change requires the version to upgrade to to support non-root.

internal/pkg/agent/install/perms.go
```diff
@@ -0,0 +1,10 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package install
+
+// FixPermissions fixes the permissions on the installed system.
+func FixPermissions(topPath string, uid string, gid string) error {
+	return fixPermissions(topPath, uid, gid)
+}
```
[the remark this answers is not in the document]
blakerouse: Because each OS has a different `fixPermissions`. I preferred for the public function to be for all OS's and the implementation unique per OS.

internal/pkg/agent/cmd/install.go
```diff
@@ -21,7 +22,10 @@ import (
 	"github.com/elastic/elastic-agent/pkg/utils"
 )
 
-const flagInstallBasePath = "base-path"
+const (
+	flagInstallBasePath = "base-path"
+	flagInstallNonRoot  = "non-root"
```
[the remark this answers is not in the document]
blakerouse: We want the default to be installation as root as that gives the best experience and allows the ability for Endpoint Security to be used. By changing to `--root` that would mean the default becomes non-root which is not what we want.

We will also not allow the ability user a specific user. It will always be the `elastic-agent` user and the `elastic-agent` group. They can always add a user to `elastic-agent` group to allow them to interact with the Elastic Agent.

internal/pkg/agent/application/paths/paths_windows.go
```diff
@@ -22,6 +22,9 @@ const (
 	// ControlSocketPath is the control socket path used when installed.
 	ControlSocketPath = `\\.\pipe\elastic-agent-system`
 
+	// ControlSocketNonRootPath is the control socket path used when installed as non-root.
+	ControlSocketNonRootPath = `\\.\pipe\elastic-agent-system`
```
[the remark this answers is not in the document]
blakerouse: Done.

internal/pkg/agent/cmd/run.go
```diff
@@ -612,7 +612,7 @@ func ensureInstallMarkerPresent() error {
 	// Otherwise, we're being upgraded from a version of an installed Agent
 	// that didn't use an installation marker file (that is, before v8.8.0).
 	// So create the file now.
-	if err := info.CreateInstallMarker(paths.Top()); err != nil {
+	if err := info.CreateInstallMarker(paths.Top(), "", ""); err != nil {
```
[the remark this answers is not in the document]
blakerouse: Done.

testing/integration/install_non_root_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/internal/pkg/agent/install"
+	atesting "github.com/elastic/elastic-agent/pkg/testing"
+	"github.com/elastic/elastic-agent/pkg/testing/define"
+)
+
+func TestInstallNonRootWithoutBasePath(t *testing.T) {
```
cmacknz, line 29: I like these tests but I'm not sure they are covering enough to ensure we never break the non-root mode. I am worried anytime somebody does anything involving a file they will break the non-root mode by failing to set the permissions properly, or placing it in a directory the elastic-agent user can't access by default.

Key things I can think of that don't look like they are covered here:

1. Upgrades.
2. Enrolling with Fleet
3. Shipping monitoring logs to Fleet.
4. Collecting diagnostics.
5. Collecting system logs and metrics as the system integration would do.

cmacknz: I wonder if it would be better to run every test in non-root mode by default since that is the hardest to get right, and only run the sanity check tests for the agent as root. So the inverse of this approach since if it works as non-root it is highly likely to work as root.

The endpoint security tests would have to be root at least, that plus an upgrade test as root might be enough as a sanity check.

cmacknz: The main difference seems to be the socket path being different and the install operation has to create the user and group, otherwise it seems we are just auditing for incorrect file permissions.

internal/pkg/agent/install/install.go
```diff
@@ -44,18 +48,62 @@ func Install(cfgFile, topPath string, pt *progressbar.ProgressBar, streams *cli.
...
 		}
 		pt.Describe("Successfully uninstalled current Elastic Agent")
 	}
 
+	uid := "0"
+	gid := "0"
+	username := ""
+	groupName := ""
+	if nonRoot {
+		username = elasticUsername
```
cmacknz, line 64: nit: This block is big enough to be its own function.

internal/pkg/agent/install/install.go
```diff
@@ -301,3 +369,22 @@ func hasAllSSDs(block ghw.BlockInfo) bool {
 
 	return true
 }
+
+func createSocketDir(uid int, gid int) error {
```
cmacknz, line 373: UID and GID are passed around so much it might be worth encapsulating them into a structure. At minimum that would let us initialize them by name to minimize the chance that someone accidentally reverses them.

internal/pkg/agent/cmd/install.go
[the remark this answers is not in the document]
cmacknz: Agree, we can't change the default which means we need a negative in the flag like this.

My only other suggestion would be to come up with a name for this option that isn't Linux specific, since root doesn't exist on Windows. It's Administrator there. Perhaps `--unprivileged`? Not sure if that is as clear or if there is a better general term for this.

Or maybe just aliasing this to `--non-admin` works for people who only speak Windows.

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -175,19 +179,21 @@ func restartAgent(ctx context.Context, log *logger.Logger) error {
...
 		log.Warnf("Failed to restart agent via control protocol: %s", err.Error())
 
-		// Next, try to restart Agent via the service.
-		log.Infof("Restarting Agent via service; attempt %d of %d", restartAttempt, maxRestartCount)
-		err = restartViaServiceFn(ctx)
-		if err == nil {
-			break
+		// Next, try to restart Agent via the service. (only if root)
+		if root {
+			log.Infof("Restarting Agent via service; attempt %d of %d", restartAttempt, maxRestartCount)
```
cmacknz, line 184: Do we need to keep this in the root case, if it isn't possible in the root case?

I am worried about upgrade bugs becoming specific to the type of installation, although maybe that is unavoidable.

internal/pkg/agent/application/paths/paths.go
```diff
@@ -17,6 +17,9 @@ const (
 	// ControlSocketPath is the control socket path used when installed.
 	ControlSocketPath = "unix:///run/elastic-agent.sock"
 
+	// ControlSocketNonRootPath is the control socket path used when installed as non-root.
+	ControlSocketNonRootPath = "unix:///run/elastic-agent/elastic-agent.sock"
```
cmacknz, line 21: Why does this need a containing directory when not running as root? Because of the permissions on /run?

cmacknz (commented):
My primary concern is about keeping this working long term, we may need to adjust our testing strategy. Otherwise the changes look pretty reasonable.

We should additionally add something to our diagnostics to clearly indicate if the agent was installed as root or not so we can isolate problems specific to one mode or the other quickly. We may want to log the user we are running as at startup as well or something.

testing/integration/install_non_root_test.go
```diff
@@ -0,0 +1,205 @@
...
+		defaultBasePath = `/Library`
+	case "linux":
+		defaultBasePath = `/opt`
+	case "windows":
+		defaultBasePath = `C:\Program Files`
+	}
+
+	topPath := filepath.Join(defaultBasePath, "Elastic", "Agent")
+	err = os.RemoveAll(topPath)
+	require.NoError(t, err, "failed to remove %q. The test requires this path not to exist.")
```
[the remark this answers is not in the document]
blakerouse: It is actually from the other install test that already exists. They do cleanup and it is checked at the end of the test, but if your running with `KEEP_INSTALLED=true` then it can result in it not being removed. This check is just a pre-check to ensure that its clean before performing the test. I think its best to have it just for sanity check before the test runs.

testing/integration/install_non_root_test.go
blakerouse: I could do that in a follow up PR. Especially because I have to wait for this to be in the SNAPSHOT build before I can write the upgrade test, which has to be in a follow up PR.

internal/pkg/agent/application/upgrade/rollback.go
blakerouse: This was added by @ycombinator. It think it was added to ensure that in the worse case that it would be recoverable. Sadly its just not possible to perform in non-root.

internal/pkg/agent/application/paths/paths.go
blakerouse: Correct. The `/run/elastic-agent` directory must be created during install when executing as root, and then the `elastic-agent` user can write to that directory.

internal/pkg/agent/cmd/install.go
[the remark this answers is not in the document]
blakerouse: I like `--unprivileged`, it works for both. Instead of having multiple flags depending on the OS. If there is an agreement on `--unprivileged` I will change it to that.

internal/pkg/agent/install/install.go
blakerouse: Okay will refactor it.

internal/pkg/agent/install/install.go
blakerouse: I like that sugession. Will make that change.

testing/integration/install_non_root_test.go
cmacknz: > I could do that in a follow up PR

Just to confirm, making the non-root agent the default agent in tests? Would be tidier as a follow up PR anyway most likely.

internal/pkg/agent/cmd/install.go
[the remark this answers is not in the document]
nimarezainia: i also agree with `--unprivileged` we need to ensure this doesn't cause a breaking change.

internal/pkg/agent/application/info/state.go
```diff
@@ -28,11 +31,33 @@ func RunningInstalled() bool {
...
 
-func CreateInstallMarker(topPath string) error {
+func CreateInstallMarker(topPath string, uidStr string, gidStr string) error {
 	markerFilePath := filepath.Join(topPath, MarkerFileName)
 	if _, err := os.Create(markerFilePath); err != nil {
 		return err
 	}
 
+	var err error
+	if runtime.GOOS != "windows" {
```
ycombinator, line 41: Since `os.Getuid()` and `os.Getgid()` do actually work on Windows, it's not immediately obvious (without the context of this PR) why we need this `if` condition here.  Consider adding an explanatory comment.

internal/pkg/agent/install/install.go
```diff
@@ -23,13 +24,16 @@ import (
...
 const (
 	darwin = "darwin"
+
+	elasticUsername  = "elastic-agent"
+	elasticGroupName = "elastic-agent"
 )
 
 // Install installs Elastic Agent persistently on the system including creating and starting its service.
-func Install(cfgFile, topPath string, pt *progressbar.ProgressBar, streams *cli.IOStreams) error {
+func Install(cfgFile, topPath string, nonRoot bool, pt *progressbar.ProgressBar, streams *cli.IOStreams) (string, string, error) {
```
ycombinator, line 33: I know you mentioned it in the PR description that you're returning UID and GID as strings for future Windows support.  Consider mentioning that in this function's godoc as well for developers reading this code in the future without the context of this PR.

internal/pkg/agent/install/install.go
ycombinator: If you're going this struct route (which I'm also ++ for), I'd consider encapsulating the string counterparts in the struct and adding accessor methods for returning the corresponding int values.  You could probably just do the conversion once, as part of the struct's constructor, not every time the accessors are called.  Currently, this string->int conversion is being done in a couple of places so moving this into a single place would be nice.

ycombinator:
> Note: Root is still required to install the Elastic Agent, just adds support for --non-root option.

UX suggestion: I think it might be a nice UX touch to clarify this in the help text that's displayed when a user tries to install elastic-agent without sudo.  Today, we display the following help text:

```
$ ./elastic-agent install -n
Error: unable to perform install command, not executed with root permissions
```

I think we could add something here that says if you want to _run_ Elastic Agent as an unprivileged user, you still need to execute the install command as a privileged user but should pass the `--unprivileged` option.

ycombinator:
I built Agent from this PR's branch and tried installing it with `--non-root`.  I think there are some issues with the socket creation.

```
$ sudo ./elastic-agent install -n --non-root
[    ] Done  [16s]
sudo elastic-agent version
Binary: 8.12.0-SNAPSHOT (build: 28faa54f1f3c0cdd50c280d8f3a5bad6a348046a at 2023-11-08 06:54:37 +0000 UTC)
Daemon: <failed to communicate>
could not get version. failed to communicate with running daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/elastic-agent.sock: connect: no such file or directory"
Use --binary-only flag to skip trying to retrieve version from running daemon
```

If I try installing the same build of Agent without the `--non-root` option, socket creation works:

```
$ sudo ./elastic-agent install -n
[=== ] Done  [16s]
$ sudo elastic-agent version
Binary: 8.12.0-SNAPSHOT (build: 28faa54f1f3c0cdd50c280d8f3a5bad6a348046a at 2023-11-08 06:54:37 +0000 UTC)
Daemon: 8.12.0-SNAPSHOT (build: 28faa54f1f3c0cdd50c280d8f3a5bad6a348046a at 2023-11-08 06:54:37 +0000 UTC)
```

internal/pkg/agent/application/paths/paths_darwin.go
```diff
@@ -17,6 +17,9 @@ const (
 	// ControlSocketPath is the control socket path used when installed.
 	ControlSocketPath = "unix:///var/run/elastic-agent.sock"
 
+	// ControlSocketNonRootPath is the control socket path used when installed as non-root.
+	ControlSocketNonRootPath = "unix:///var/run/elastic-agent/elastic-agent.sock"
```
michalpristas, line 21: could you add a comment why we need to place it in subdir, i'm sure we'll forget

internal/pkg/agent/cmd/install.go
```diff
@@ -67,6 +72,12 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 		return fmt.Errorf("unable to perform install command, not executed with %s permissions", utils.PermissionUser)
 	}
 
+	// only support Linux at the moment
+	nonRoot, _ := cmd.Flags().GetBool(flagInstallNonRoot)
+	if runtime.GOOS != "linux" {
```
michalpristas, line 77: shouldn't this be `nonRoot && runtime.GOOS != "linux"`

blakerouse:
> I built Agent from this PR's branch and tried installing it with `--non-root`. I think there are some issues with the socket creation.
> 
> ```
> $ sudo ./elastic-agent install -n --non-root
> Installing in non-interactive mode.
> [    ] Done  [16s]
> Elastic Agent has been successfully installed.
> sudo elastic-agent version
> Binary: 8.12.0-SNAPSHOT (build: 28faa54f1f3c0cdd50c280d8f3a5bad6a348046a at 2023-11-08 06:54:37 +0000 UTC)
> Daemon: <failed to communicate>
> could not get version. failed to communicate with running daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/elastic-agent.sock: connect: no such file or directory"
> Use --binary-only flag to skip trying to retrieve version from running daemon
> ```

You didn't read the description of the PR ;-). You cannot call `sudo elastic-agent version` that is running it as root, you installed it as non-root. See "How to test this PR locally" of the PR description.

ycombinator:
> > I built Agent from this PR's branch and tried installing it with `--non-root`. I think there are some issues with the socket creation.
> > ```
> > $ sudo ./elastic-agent install -n --non-root
> > Installing in non-interactive mode.
> > [    ] Done  [16s]
> > Elastic Agent has been successfully installed.
> > sudo elastic-agent version
> > Binary: 8.12.0-SNAPSHOT (build: 28faa54f1f3c0cdd50c280d8f3a5bad6a348046a at 2023-11-08 06:54:37 +0000 UTC)
> > Daemon: <failed to communicate>
> > could not get version. failed to communicate with running daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/elastic-agent.sock: connect: no such file or directory"
> > Use --binary-only flag to skip trying to retrieve version from running daemon
> > ```
> 
> You didn't read the description of the PR ;-). You cannot call `sudo elastic-agent version` that is running it as root, you installed it as non-root. See "How to test this PR locally" of the PR description.

Ah, I see that now 🤦 - thanks!

internal/pkg/agent/install/perms.go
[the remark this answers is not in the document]
blakerouse: Changed to remove the extra `FixPermissions` wrapper.

internal/pkg/agent/cmd/install.go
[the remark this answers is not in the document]
blakerouse: Changed to `--unprivileged`.

testing/integration/install_non_root_test.go
blakerouse: I will do a follow up once this change is in the latest snapshot to run as many tests as we can both as root and non-root installations.

internal/pkg/agent/install/install.go
blakerouse: Changing the the struct route allows me to define different structs between unix and Windows, which made the code a lot cleaner and much easier to understand.

internal/pkg/agent/application/paths/paths.go
blakerouse: Added comment on why it must be in a directory.

internal/pkg/agent/install/install.go
blakerouse: I was able to change this behavior removing the need for these type of comments.

internal/pkg/agent/application/paths/paths_darwin.go
blakerouse: Added.

internal/pkg/agent/cmd/install.go
blakerouse: Yes it should. Fixed.

blakerouse:
I have updated the PR to switch to `--unprivileged` and to switch to a structure for passing around the UID/GID between the functions. The structure switched allowed it to be much simpler being that I could define different structures between unix and Windows. I fixed and cleaned up all review comments, this is ready for more reviews, thanks!

pkg/utils/perm_windows.go
```diff
@@ -6,6 +6,21 @@
...
 
+// FileOwner is the ownership a file should have.
+type FileOwner struct {
+	UID string
+	GID string
+}
+
+// CurrentFileOwner returns the executing UID and GID of the current process.
+func CurrentFileOwner() FileOwner {
+	// TODO(blakerouse): Make this return the current user and group on Windows.
```
pchila, line 17: Is this TODO to be implemented in this PR ?

pkg/utils/perm_unix.go
```diff
@@ -11,6 +11,20 @@ import (
 	"os"
 )
 
+// FileOwner is the ownership a file should have.
```
pchila, line 14: This describes a user along with its default group, so maybe the name should be `UserGroupIDs` ?

pchila (approved):

internal/pkg/agent/cmd/install.go
```diff
@@ -43,6 +47,7 @@ would like the Agent to operate.
 	cmd.Flags().BoolP("force", "f", false, "Force overwrite the current installation and do not prompt for confirmation")
 	cmd.Flags().BoolP("non-interactive", "n", false, "Install Elastic Agent in non-interactive mode which will not prompt on missing parameters but fails instead.")
 	cmd.Flags().String(flagInstallBasePath, paths.DefaultBasePath, "The path where the Elastic Agent will be installed. It must be an absolute path.")
+	cmd.Flags().Bool(flagInstallUnprivileged, false, "Installed Elastic Agent will create an 'elastic-agent' user and run as that user, instead of running as root.")
```
ycombinator, line 50: Since we've replaced `--non-root` with `--unprivileged`, might as well remove the `root` reference from this help text as well.

```suggestion
	cmd.Flags().Bool(flagInstallUnprivileged, false, "Installed Elastic Agent will create an 'elastic-agent' user and run as that user, instead of running as the privileged user.")
```

ycombinator (approved):
LGTM — nice work!

internal/pkg/agent/application/info/state_windows.go
```diff
@@ -0,0 +1,16 @@
...
+//go:build windows
+
+package info
+
+import (
+	"github.com/elastic/elastic-agent/pkg/utils"
+)
+
+func fixInstallMarkerPermissions(markerFilePath string, ownership utils.FileOwner) error {
+	// TODO(blakerouse): Fix the market permissions on Windows.
```
michalpristas, line 14: missed or future work? can we link an issue if so?

internal/pkg/agent/cmd/install.go
michalpristas: agree, root is OS specific

internal/pkg/agent/install/uninstall.go
```diff
@@ -35,7 +35,8 @@ import (
 // Uninstall uninstalls persistently Elastic Agent on the system.
 func Uninstall(cfgFile, topPath, uninstallToken string, pt *progressbar.ProgressBar) error {
 	// uninstall the current service
-	svc, err := newService(topPath)
+	// not creating the service, so no need to set the username and group to any value
+	svc, err := newService(topPath, "", "")
```
michalpristas, line 39: a lot of these with `""` as args while uid and gid are actually used in not that many use cases. maybe we should have `newService(path)` and `newServiceWithOwner(path, uid,gid)` or fluent alternative `newService(path).WithOwner(uid, gid)`

pkg/utils/perm_windows.go
```diff
@@ -6,6 +6,21 @@
...
+	return FileOwner{
+		UID: "S-1-5-32-544", // Administrator
+		GID: "S-1-5-18",     // SYSTEM
```
michalpristas, line 20: may be worth extracting to consts and using these consts in `internal/pkg/agent/install/perms_windows.go` as well. helps tracking usage

michalpristas (approved):
in general it looks good, still few comments, mostly minor. I;d like to see craigs suggested refactor in though
I havent tested this on windows

pkg/utils/perm_windows.go
blakerouse: Unified this in the PR to use a constant here and anywhere else that used those same strings. It was in a few places before this change.

internal/pkg/agent/install/uninstall.go
blakerouse: I cleaned this up to use the options pattern so it doesn't need to be changed everywhere.

internal/pkg/agent/application/info/state_windows.go
blakerouse: don't really have a public issue, so I didn't link it

I am working on Windows next, this will not live long.

pkg/utils/perm_unix.go
blakerouse: A file is always owned by a user and a group. To me the name is correct, and you cannot have a file owned by multiple groups at the same time.

pkg/utils/perm_windows.go
blakerouse: It is for a follow up that I am working on, which enabled non Administrator on Windows.

internal/pkg/agent/cmd/install.go
blakerouse: Updated.

cmacknz (approved):
Looks good, I created https://github.com/elastic/elastic-agent/issues/3736 to follow up on ensuring we test this thoroughly.

IMO we should prioritize getting solid tests in place before supporting the other OSes.

My biggest problem so far is reliably typing unprivileged correctly. For whatever reason I always type it as unpriveledged or something.

blakerouse:
@cmacknz I cannot merge it because of Sonar and the failed coverage merge in buildkite.

cmacknz:
Yeah looks like https://buildkite.com/elastic/elastic-agent/builds/4755#018bb508-189b-47b2-b5b1-edeb428e28e8 actually passed but isn't posting results.

cmacknz:
Let's run it one more time and to give Anderson and Tiago time to respond to some of the new comments, then I can force merge it.

cmacknz:
buildkite test it

amitkanfer:
related: https://github.com/elastic/ingest-dev/issues/1766

PR #3622, opened 2023-10-17
Change upgrade watcher to use StateWatch

blakerouse:
## What does this PR do?

Refactors how the upgrade watcher works removing the need to parse the PID from any system service manager.

The watcher now tracks the inability to connect to the daemon, lost connections to the daemon, agent failures, component failures, and flip-flopping of healthy/error states. The tracking of the lost connections to the daemon allows the removal of the PID tracking that used to be present. The error tracking is more robust as it requires that an error be an error for at least 30 seconds, allowing the Elastic Agent to possibly have an issue and then recover (before if the State call happened when an error was present it would just become failed).

This also includes unit testing for the `AgentWatcher` service that previously had no unit tests.

## Why is it important?

Completely removes the requirement of executing and reading output of service managers (removing the need for root access). Improves the overall logic and provides better testing.

## Checklist

- [x] I have added an integration test or an E2E test **(covered by existing tests)**


internal/pkg/agent/application/upgrade/watcher.go
```diff
@@ -0,0 +1,217 @@
...
+						}
+					}
+					if compErr != nil {
+						failedCh <- fmt.Errorf("%w: %w", ErrAgentComponentFailed, compErr)
+						continue
+					}
+				}
+
+				// nothing is failed
+				failedCh <- nil
```
ycombinator, line 201: If nothing is failed, it looks like we continue with this inner loop, immediately calling `watch.Recv()` again, followed by the status checks.  Is this what we want?  Or do we want to continue with `LOOP` here so the next `watch.Recv()` happens when the `connectTimer` fires again?

ycombinator:
https://github.com/elastic/elastic-agent/blob/defde80db7ce43db2e319fd2cfb458f16101a69f/testing/upgradetest/watcher.go#L22

Should we be removing this line? Or do we need to leave it in there for older Agent versions (that don't have the watcher implementation in this PR)?

internal/pkg/agent/application/upgrade/watcher.go
```diff
@@ -0,0 +1,217 @@
...
+				return
+			case reset := <-failedReset:
+				if reset {
+					failedCount = 0
+					failedTimer.Stop()
+				}
+			case err := <-failedCh:
+				if err != nil {
+					if failedErr == nil {
+						failedCount++
```
ycombinator, line 93: It looks like we only increment `failedCount` when we received a non-nil error from `failedCh` but previously there was no error (i.e. `failedErr == nil`).  So that means `failedCount` actually represents the number of times we flip-flopped from a non-error state to an error state, right?  If my understanding is correct, could we rename this to something like `flipFlopCount` or something else indicating that it's counting flip-flops?

internal/pkg/agent/application/upgrade/watcher.go
blakerouse: This is a streaming protocol so its a permanent connection that keeps sending state changes. This is correct we want to keep receiving the next state.

ycombinator: I see.  So `watch.Recv()` will block until it has something to return?  In that case, we're good here!

blakerouse: Correct, it is blocking.

internal/pkg/agent/application/upgrade/watcher.go
blakerouse: Done.

ycombinator (approved):
LGTM.

blakerouse:
buildkite test this

blakerouse:
another unrelated failure... this time the endpoint integration

blakerouse:
buildkite test this

Issue #3607, opened 2023-10-16
Flaky Test:  TestProgress/nested_step_delayed_success


pchila:
## Flaky Test

* **Test Name:** TestProgress/nested_step_delayed_success
* **Link:** [Link to file/line number in github.](https://github.com/elastic/elastic-agent/blob/d5b58ef457970e4834cdacb2ee2819b4b0dafbc3/internal/pkg/agent/install/progress_test.go#L136)
* **Branch:** [Git branch the test was seen in. If a PR, the branch the PR was based off.](https://github.com/elastic/elastic-agent/tree/mergify/bp/8.11/pr-3547)
* **Artifact Link:** 
* **Notes:** There's an extra <dot> appearing between substep1 and substep2 which probably makes the regex fail.

https://github.com/elastic/elastic-agent/actions/runs/6530034372/job/17728615531?pr=3605

### Stack Trace

```
=== Failed
=== FAIL: internal/pkg/agent/install TestProgress/nested_step_delayed_success (0.08s)
    progress_test.go:136: 
        	Error Trace:	/Users/runner/work/elastic-agent/elastic-agent/internal/pkg/agent/install/progress_test.go:136
        	Error:      	Expect "step starting...
        	            	   substep 1 starting....... DONE
        	            	.   substep 2 starting..... DONE
        	            	   DONE
        	            	" to match "step starting\.{3,}\n   substep 1 starting\.{3,}\.+ DONE\n   substep 2 starting\.{3,}\.+ DONE\n   DONE\n"
        	Test:       	TestProgress/nested_step_delayed_success

=== FAIL: internal/pkg/agent/install TestProgress (0.28s)
```

fearful-symmetry:
So, just appears to be a race where we can call other operations in the progress helper while we're in the middle of a `tick()` operation. Going to put together a PR...

PR #3623, opened 2023-10-17
Use third-party library for reporting install/uninstall progress

fearful-symmetry:
## What does this PR do?
fixes https://github.com/elastic/elastic-agent/issues/3607

This uses a third-party library for creating a spinner that reports progress during install/uninstall operations.

## Why is it important?

The old code was causing some flaky tests, and for such a small bit of code, it was a bit easier to just use a third-party library. 

## Checklist


ycombinator:
> At that point we're getting kind of complicated for what this is, and we might just want to find a third-party library like https://github.com/schollz/progressbar to just do this for us, which is also an option here. Honestly open to the third-party library idea for a follow-up to this; not sure why we're DIY'ing this.

I'm ++ to using a third-party library here.  I've been thinking along the same lines myself.  The initial implementation was done as one of three OnWeek projects and, at the time, I thought it would be simpler to DIY this especially since we weren't doing anything too fancy like repainting the screen (like some third party progress trackers can do).  I honestly hadn't thought it would introduce so much flakiness and headache.  But after seeing that over the past few weeks, I'm very much in favor of just replacing this with a third-party library at this point.

fearful-symmetry:
Alright, so, it looks like SonarQube is counting all the one-line log statements I added towards the code coverage percentage, which is...not great.

ycombinator:
I like the new progress bar and the time counter too!

There are two things I was wondering if we could change:

1. Would it be possible to display all the steps?  That is, when a step completes, would it be possible to leave it on the screen without it being replaced by the next step?


2. The install ends like so:
   ```
   [=   ] Service Started  [11s] Elastic Agent has been successfully installed.
   ```

   The looks a bit odd.  I think we're missing a newline?

ycombinator:
Right after `sudo ./elastic-agent install` is called, there's a spinner but no text is shown next to it.  I managed to capture a screenshot of the moment:

Screenshot 2023-10-19 at 09 14 19

Seems like there should be some text there about the work being done?

I'm not seeing the same behavior during uninstall, though, which is good.

ycombinator:
Thanks @fearful-symmetry, I see the initial message during `sudo elastic-agent install` now.  

One nit:

During uninstallation (`sudo elastic-agent uninstall`), when uninstallation succeeds, the last progress message that's printed is "Done".

```
$ sudo elastic-agent uninstall -f
[ ===] Done  [1s]
```

However, during installation (`sudo ./elastic-agent install`), when installation succeeds, the last progress message that's printed is the last actual installation step.

```
$ sudo ./elastic-agent install -n
[==  ] Service Started  [11s]
```

Could we make these behaviors consistent? I'm not sure which direction makes more sense but I'm leaning slightly towards the way uninstallation behaves.

ycombinator (approved):
LGTM. Thanks for fixing this, Alex — much needed!

pchila (approved):
LGTM

michalpristas (approved):

cmacknz:
I am going to force merge this. Looking at the code that is missing coverage, some of it is covered by integration tests and I don't see value in simulating the 10+ error conditions we can hit just to verify a log line gets printed as expected.

orestisfl:
Just noticed this on a snapshot and it looks amazing :star_struck: :+1:

Issue #3119, opened 2023-07-25
Track and report upgrade details


ycombinator:
**Describe the enhancement:**

Elastic Agent should track details about an ongoing upgrade and report them to Fleet Server.

**Describe a specific use case for the enhancement or feature:**

To understand where upgrades fail or get stuck without progressing.

**What is the definition of done?**

* [x] Elastic Agent is able to track the various upgrade states (defined below) representing details about an ongoing upgrade and correctly transition between these states.
* [x] Elastic Agent is able to report a change in an upgrade state to Fleet Server via a new `upgrade_details` field in the check-in API.
* [ ] For backwards compatibility, Elastic Agent continues to acknowledge the `UPGRADE` action with Fleet Server, via the acknowledgements API, as before.
 
**Details**

In the future, Elastic Agents will send details about an ongoing upgrade to Fleet Server.  These details will be communicated via a new `upgrade_details` field in check-in API requests.  

The proposed structure of the `upgrade_details` field is:

```json
{
  "upgrade_details": { // new field; present when upgrade is in progress
    "target_version": "8.12.0", // version being upgraded to; always present
    "action_id": "xxxxxxxx", // ID of the UPGRADE action
    "state": "UPG_*",
    "metadata": {
      "scheduled_at": "2023-08-09T10:11:12Z", // when state == "UPG_SCHEDULED"
      "download_percent": 16.4, // when state == "UPG_DOWNLOADING"
      "failed_state": "UPG_*" // when state == "UPG_FAILED"
      "error_msg": "" // when state == "UPG_FAILED"
    }
  }
}
```

Where `upgrade_details.state` is expected to hold one of the following values:

| State | Meaning |
| -- | -- |
| `UPG_REQUESTED` | Upgrade requested by user |
| `UPG_SCHEDULED` | Upgrade scheduled for <date/time> |
| `UPG_DOWNLOADING` | Downloading new Agent artifact version <version> |
| `UPG_EXTRACTING` | Extracting new Agent artifact version <version> |
| `UPG_REPLACING` | Replacing old Agent artifact version <version> with new one version <version> |
| `UPG_RESTARTING` | Starting new Agent version <version> |
| `UPG_WATCHING` | Monitoring new Agent version <version> |
| `UPG_ROLLBACK` | Upgrade unsuccessful; rolling back to Agent version <version> |
| `UPG_FAILED` | Upgrade failed due to error <error> from state <state> |

The possible transitions between these states are:

[...]

PR #3624, opened 2023-10-18
Include upgrade details in diagnostics bundle during ongoing upgrade

ycombinator:
## What does this PR do?

This PR enhances the diagnostics bundle to include upgrade details when Agent is being upgraded.

## Why is it important?

To gain more visibility into the Agent upgrade process.

## Checklist

1. Build `elastic-agent` from this PR's branch and install it.
2. Run `elastic-agent diagnostics`, expand the resulting archive, and check the bundled `state.yaml` file. Verify that NO upgrade details are included in it by checking that there is no `upgrade_details` field.
3. In a new window, upgrade Elastic Agent to a version that hasn't been released yet.  This will ensure that the upgrade starts but remains stalled.
   ```
   sudo elastic-agent upgrade 8.13.0
   ```
4. In the original window, run `elastic-agent diagnostics` again, expand the resulting archive, and check the bundled `state.yaml` file. Verify that upgrade details are included in it by checking that there is an `upgrade_details` field.

## Related issues

- Relates to https://github.com/elastic/elastic-agent/issues/3119#issuecomment-1743638526


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -799,12 +800,13 @@ func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
...
-					FleetMessage: s.FleetMessage,
-					LogLevel:     s.LogLevel,
-					Components:   compStates,
+					State:          s.State,
+					Message:        s.Message,
+					FleetState:     s.FleetState,
+					FleetMessage:   s.FleetMessage,
+					LogLevel:       s.LogLevel,
+					Components:     compStates,
+					UpgradeDetails: s.UpgradeDetails,
```
pchila, line 809: When is this field present? Do we get details about the upgrade after a failure ? Or those get cleared as soon as the grace period is over ?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -781,12 +781,13 @@ func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
...
-					FleetMessage string                 `yaml:"fleet_message"`
-					LogLevel     logp.Level             `yaml:"log_level"`
-					Components   []StateComponentOutput `yaml:"components"`
+					State          agentclient.State      `yaml:"state"`
+					Message        string                 `yaml:"message"`
+					FleetState     agentclient.State      `yaml:"fleet_state"`
+					FleetMessage   string                 `yaml:"fleet_message"`
+					LogLevel       logp.Level             `yaml:"log_level"`
+					Components     []StateComponentOutput `yaml:"components"`
+					UpgradeDetails *details.Details       `yaml:"upgrade_details,omitempty"`
```
pchila, line 790: I don't see any changes in the tests related to this change... could you please add a unit test to document what we can expect in the hook output when UpgradeDetails is populated ?

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: > When is this field present? 

This field is populated when the upgrade process starts: https://github.com/elastic/elastic-agent/blob/2a93086d5c7695f9a08e118667dac0da677157c6/internal/pkg/agent/application/coordinator/coordinator.go#L475-L476

> Do we get details about the upgrade after a failure ?

Yes, if the upgrade process fails at any point, this field will continue to remain populated. 

> Or those get cleared as soon as the grace period is over ?

I assume you are referring to the Upgrade Watcher's grace period.

Exactly when we clear it out after that is not yet implemented. We need to decide whether we want to clear it out based on some timeout (maybe the grace period or maybe something even beyond that) or due to some user intervention (which could be retrying the upgrade or something else just to clear the upgrade details, not sure yet) or either (whichever happens first).

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: Thanks.  Added in 02625e831564e5371bdaf6ca81771216f90df926.

ycombinator: Thanks.  Added in 02625e831564e5371bdaf6ca81771216f90df926.

blakerouse (approved):
Nice! Great this is added to the diagnostics.

pchila (approved):
LGTM

Issue #3628, opened 2023-10-18
Elastic Agent enroll fails to restart daemon on docker 


AndersonQ:
- Relates https://github.com/elastic/elastic-agent/issues/3664

The Elastic Agent fails to restart its daemon during enroll when running from the docker image. Subsequent starts of the docker container succeed.

The culprit is https://github.com/elastic/elastic-agent/commit/f7e558f736d5c17b5488a66e9051df814b95c050

- Version: main (8.12.0-SNAPSHOT), 8.11.0-SNAPSHOT
- Operating System: all
- Discuss Forum URL: N/A
- Steps to Reproduce:

```
docker run \
  --env FLEET_ENROLL=1 \
  --env FLEET_URL=https://fleet-url:8220/ \
  --env FLEET_ENROLLMENT_TOKEN=SOME_TOKEN \
  --env FLEET_INSERUCE=true \
  docker.elastic.co/beats/elastic-agent:8.12.0-SNAPSHOT
```

Some of our tests are failing:
 - https://fleet-ci.elastic.co/job/Ingest-manager/job/integrations/job/main/2928/#showFailuresLink

logs:

```
root@elastic-agent:~# docker run \
  --env FLEET_ENROLL=1 \
  --env FLEET_URL=https://some.fleet.url:port \
  --env FLEET_ENROLLMENT_TOKEN=SOME_TOKE  \
  --env FLEET_INSERUCE=true \
  docker.elastic.co/beats/elastic-agent:8.12.0-SNAPSHOT

{"log.level":"info","@timestamp":"2023-10-18T16:09:56.069Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":497},"message":"Starting enrollment to URL: https://fc2e07ab4001499380ce57a763e698fd.fleet.us-east-1.aws.staging.elastic.cloud:443/","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-10-18T16:10:19.256Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":468},"message":"Retrying to restart...","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-10-18T16:10:59.261Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":468},"message":"Retrying to restart...","ecs.version":"1.6.0"}
[...]

PR #3629, opened 2023-10-18
Revert "install fails if enroll fails and surface errors"

AndersonQ:
Reverts elastic/elastic-agent#3554

Relates to
 - https://github.com/elastic/elastic-agent/issues/3628


cmacknz (approved):

cmacknz:
Force merging, ignoring the SonarQube failure on the revert commit which we shouldn't be modifying from the original.

Issue #3307, opened 2023-08-28
Elastic Agent is not notifying the Windows service manager that it is about to shut down reliably


cmacknz:
Follow up from https://github.com/elastic/elastic-agent/issues/1525. This should have been fixed with https://github.com/elastic/elastic-agent-libs/pull/113 and https://github.com/elastic/elastic-agent/pull/2434 but we have reports that the `The Elastic Agent service terminated unexpectedly.  It has done this 1 time(s).  The following corrective action will be taken in 1000 milliseconds: Restart the service.` message can still appear in the Windows service manager when the host running the Elastic Agent shuts down.

This can cause problems for the endpoint security service, quoting @bjmcnic:

> Basically, when the host goes to shutdown, Agent terminates without notifying Windows that it's service is intentionally stopping. Windows notes Agent isn't running and starts another instance of, which as it's starting up queries Endpoint/Defend as to whether it's in a healthy state (which fails as we encounter errors querying our state with the service control manager because the host is in the process of shutting down)...when Endpoint/Defend reports a failure in validating it's healthy...Agent attempts to take the corrective action of uninstalling/reinstalling Endpoint/Defend. The system then shuts down in the middle of the Endpoint/Defend uninstall. At the point the host shutdowns, we have succeeded deleting the ELAM service, so the certificate that would let our primary service ElasticEndpoint start as PPL will not be loaded on reboot. And that's why they're seeing the error.

The agent logs at the time of the first shutdown are:

```json
{"log.level":"info","@timestamp":"2023-08-21T15:37:45.031-0400","log.origin":{"file.name":"cmd/run.go","file.line":300},"message":"signal \"terminated\" received","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-08-21T15:37:45.031-0400","log.origin":{"file.name":"cmd/run.go","file.line":312},"message":"Shutting down Elastic Agent and sending last events...","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
[...]

PR #3657, opened 2023-10-25
Increase the Windows OnFailureDelayDuration delay to 15s

cmacknz:
Increase the Windows `OnFailureDelayDuration` delay to 15s. This is the delay before the service is restarted when it exits unexpectedly. This is the same value used by endpoint-security by default.

Note that this change only applies to new agent installations. We would need to add code to migrate existing agent installations to the new value.

It was originally suggested that we increase this to 30s+ in https://github.com/elastic/elastic-agent/issues/3307#issuecomment-1696423487 but I am confident the root cause for that problem was addressed by https://github.com/elastic/elastic-agent-libs/pull/155.

Regardless I don't think our current default for this value was chosen for any particular reason and we can at least have agent behave consistently with endpoint. Speaking with @bjmcnic a 15s delay would also mitigate the original problem in all but the most extreme cases even if it were to still go unfixed.

### DIFF
--- a/changelog/fragments/1698259940-Increase-wait-period-between-service-restarts-on-failure-to-15s-on-Windows.yaml
+++ b/changelog/fragments/1698259940-Increase-wait-period-between-service-restarts-on-failure-to-15s-on-Windows.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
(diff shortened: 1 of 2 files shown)


leehinman (approved):

cmacknz:
> >> (linux-arm64-ubuntu-2204) Failed for instance linux-arm64-ubuntu-2204 (@ 34.31.33.167): ogc-linux-arm64-ubuntu-2204-9441 unable to continue because stack never became ready: failed to check for cloud 8.12.0-SNAPSHOT to be ready: context deadline exceeded

Arg

cmacknz:
buildkite test it

pchila (approved):
LGTM

cmacknz:
Force merging, test failure is unrelated. https://github.com/elastic/elastic-agent/pull/3657

Issue #3470, opened 2023-09-25
Allow running the standalone Beats serverless integration tests on a daily schedule


cmacknz:
https://github.com/elastic/elastic-agent/pull/3258 adds various tests for standalone Beats interacting with Serverless projects. These tests were added to the Elastic Agent repository because the Beats CI system is not currently capable of starting serverless projects for testing. While we expect this to be possibly eventually, in the meantime we should regularly run these tests to ensure we can catch regressions.

The scope of this issue is to add a Buildkite job that runs only the standalone Beats serverless integration tests as a nightly test suite. This should gate the merge of https://github.com/elastic/elastic-agent/pull/3258 since these tests should not be triggered for each PR in this repository since the code here cannot cause regressions in these tests.

cmacknz:
We do not need this work once https://github.com/elastic/beats/issues/36675 is completed, this issue only exists as a stop gap solution while we migrate the tests to Beats. We can avoid doing this at all if we can complete https://github.com/elastic/beats/issues/36675 in the near future instead.

PR #3658, opened 2023-10-25
Add serverless beats tests

fearful-symmetry:
## What does this PR do?

This is (hopefully) the final stage of getting serverless testing working. This Adds 17 serverless tests, as well as a mage role to run the tests. Assuming you have a beats binary built, you can run `mage integration:testBeatServerless metricbeat`

This does not include any buildkite steps to run the tests; I'm not totally sure how we want to schedule them, and if that should be part of this PR.

## Why is it important?

We need serverless tests.

## Checklist


testing/integration/beats_serverless_test.go
```diff
@@ -0,0 +1,589 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build integration
```
pchila, line 5: Having just the `integration` build tag will include this file while trying to launch integration tests for elastic agent. Maybe we want to specify an additional tag like:

```suggestion
//go:build integration
//go:build serverless
```
Ofc we need to add the tag in magefile target

testing/integration/beats_serverless_test.go
```diff
@@ -0,0 +1,589 @@
...
+	require.NoError(runner.T(), err)
+
+	// beats likes to add standard ports to URLs that don't have them, and ESS will sometimes return a URL without a port, assuming :443
+	// so try to fix that here
+	fixedKibanaHost := runner.kibHost
+	parsedKibana, err := url.Parse(runner.kibHost)
+	require.NoError(runner.T(), err)
+	if parsedKibana.Port() == "" {
+		fixedKibanaHost = fmt.Sprintf("%s:443", fixedKibanaHost)
+	}
```
pchila, line 134: Shouldn't this conditioned to the url scheme being https (which is 99.9% of the time is)? It's just that setting fixed `:443` to a URL where we don't even check the scheme feels wrong...
Furthermore, since we spent the time parsing the URL couldn't we set `parsedKibana.Host` field and then call `parsedKibanaHost.String()` ?

testing/integration/beats_serverless_test.go
```diff
@@ -0,0 +1,589 @@
...
+	if parsedES.Port() == "" {
+		fixedESHost = fmt.Sprintf("%s:443", fixedESHost)
+	}
+
+	runner.T().Logf("configuring beats with %s / %s", fixedESHost, fixedKibanaHost)
+
+	testUuid, err := uuid.NewV4()
+	require.NoError(runner.T(), err)
+	runner.testUuid = testUuid.String()
+	parsedCfg := fmt.Sprintf(beatOutConfig, fixedESHost, apiResp.Id, apiResp.APIKey, fixedKibanaHost, testUuid.String())
```
pchila, line 148: maybe we can use a `template` instead of `fmt.Sprintf` ? Having positional arguments substituted in the config kinda makes me nervous

testing/integration/beats_serverless_test.go
```diff
@@ -0,0 +1,589 @@
...
+	_ = estools.DeleteIndexTemplatesDataStreams(ctx, runner.requirementsInfo.ESClient, fmt.Sprintf("*%s*", runner.testbeatName))
+
+	err := runner.agentFixture.RunBeat(ctx)
+	require.NoError(runner.T(), err)
+
+	docs, err := estools.GetLatestDocumentMatchingQuery(ctx, runner.requirementsInfo.ESClient, map[string]interface{}{
+		"match": map[string]interface{}{
+			"host.test-id": runner.testUuid,
+		},
+	}, fmt.Sprintf("*%s*", runner.testbeatName))
```
pchila, line 168: We are only doing it once in the whole test, let's keep it simple and allocate one extra string ;)
```suggestion
	docs, err := estools.GetLatestDocumentMatchingQuery(ctx, runner.requirementsInfo.ESClient, map[string]interface{}{
		"match": map[string]interface{}{
			"host.test-id": runner.testUuid,
		},
	}, "*"+runner.testbeatName+"*")
```

testing/integration/beats_serverless_test.go
fearful-symmetry: You would think, but the beats defaults to `5601` if you don't specify the port, hence why we need to add it manually.

testing/integration/beats_serverless_test.go
fearful-symmetry: Yah, considering that this PR is technically temporary until we get buildkite setup in the beats repo, I didn't want to get to clever with anything that would require modifications to the agent code, since eventually this'll become a proper integration test in the agent repo.

cmacknz:
Given that this is finding actual bugs in serverless Beats, my preference to unblock this is to  run these tests on a scheduled https://github.com/elastic/elastic-agent/issues/3470

Otherwise we will be blocked waiting for the Beats repository's migration to Buildkite which is probably 1+ month away.

fearful-symmetry:
@cmacknz agreed. Is adding some scheduled run in buildkite something that can be done in this PR, or is there some other global config?

cmacknz:
You should be able to do it in this PR

.buildkite/pipeline.yml
```diff
@@ -160,6 +160,18 @@ steps:
       provider: "gcp"
       machineType: "n1-standard-8"
 
+  - label: "Serverless Beats Tests"
+    key: "serverless-beats-integration-tests"
+    command: ".buildkite/scripts/steps/beats_tests.sh"
+    if: "build.env('CRON') == 'yes'"
```
cmacknz, line 166: Where is the cron schedule controlled? Can you link to it here?

cmacknz:
Do you have a link to the Beats tests passing on Buildkite, since they aren't running as part of the PR automation?

fearful-symmetry:
@cmacknz here: https://buildkite.com/elastic/elastic-agent/builds/4490#018b87a1-0854-4559-b675-20e3033f15e4

.buildkite/pipeline.yml
fearful-symmetry: So, I was going to add it once the PR was merged, but the control panel for it is here: https://buildkite.com/elastic/elastic-agent/settings/schedules

testing/integration/beats_serverless_test.go
cmacknz: Changing the build tag wouldn't just require you to make a dedicated mage target for running these I think. That is probably a good thing.

How are these getting excluded from `mage integration:test` runs today?

.buildkite/pipeline.yml
cmacknz: Add a link to that as a comment right above this so future us can find it easily

.buildkite/scripts/steps/beats_tests.sh
```diff
@@ -0,0 +1,73 @@
...
+    exit 1
+fi
+
+echo "testing filebeat..."
+if ! run_test_for_beat filebeat; then
+    exit 1
+fi
+
+
+echo "testing auditbeat..."
```
cmacknz, line 70: We can skip testing auditbeat since our team doesn't maintain it. None of the code we are testing is really Beat specific regardless.

.buildkite/scripts/steps/beats_tests.sh
```diff
@@ -0,0 +1,73 @@
...
+    return $TESTS_EXIT_STATUS
+}
+#run mage before setup, since this will install go and mage
+#the setup scripts will do a few things that assume we're running out of elastic-agent and will break things for beats, so run before we do actual setup
+mage -l
+
+mkdir -p build
+cd build
+
+git clone git@github.com:elastic/beats.git
```
cmacknz, line 39: Why do we need to clone and build Beats?

Why don't we just download the latest SNAPSHOT image using the artifacts API?

cmacknz:
Your sample test run doesn't contain any Beat logs. If these tests fail how do we debug them?

There are no test artifacts generated either https://buildkite.com/elastic/elastic-agent/builds/4490#018b87a1-0854-4559-b675-20e3033f15e4

So there is no test report generated to make the failures easier to see.

It also looks like it is running every other test. Is that the intent for the scheduled job? I would have it only run the serverless Beats test so it isn't going to fail for any other random reason.

pchila (commented):
A couple of comments left about the pipeline step script and the mage target.
Overall code is ok, I am not too fond of the mixing of beats serverless and elastic agent integration tests being guarded only by a condition on the TestSuite and a mage target that would allow to run those tests against a stateful stack (not sure that we have a usecase for that).

.buildkite/scripts/common.sh
```diff
@@ -53,7 +53,7 @@ getOSOptions() {
 # Wrapper function for executing mage
 mage() {
     go version
-    if ! [ -x "$(type -p mage | sed 's/mage is //g')" ];
+    if ! [ -x "$(type -P mage | sed 's/mage is //g')" ];
```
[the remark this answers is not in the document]
fearful-symmetry: so, because we define a `mage` bash function that encases the actual mage executable and we only want to check if the actual binary is installed, `type -p` seems to be the only command I found that behaves in the way the script initially expected. The previous version would always return true.

.buildkite/scripts/steps/beats_tests.sh
```diff
@@ -0,0 +1,73 @@
...
+# TEST_INTEG_CLEAN_ON_EXIT=true TEST_PLATFORMS="linux/amd64" STACK_PROVISIONER="$STACK_PROVISIONER" SNAPSHOT=true mage integration:testBeatServerless metricbeat
+# TESTS_EXIT_STATUS=$?
+# set -e
+
+# exit $TESTS_EXIT_STATUS
+
+echo "testing metricbeat..."
+if ! run_test_for_beat metricbeat; then
+    exit 1
+fi
```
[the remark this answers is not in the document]
fearful-symmetry: ah, good point!

magefile.go
```diff
@@ -1567,6 +1567,33 @@ func (Integration) PrepareOnRemote() {
...
+	}
+
+	// a bit of bypass logic; run as serverless by default
+	if os.Getenv("STACK_PROVISIONER") == "" {
+		err := os.Setenv("STACK_PROVISIONER", "serverless")
+		if err != nil {
+			return fmt.Errorf("error setting serverless stack var: %w", err)
+		}
+	} else if os.Getenv("STACK_PROVISIONER") == "stateful" {
+		fmt.Printf(">>> Warning: running TestBeatServerless as stateful\n")
```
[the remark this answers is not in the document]
fearful-symmetry: It's occasionally useful for testing, usually for cases if we aren't sure if something is just broken, or only broken under serverless. Has happened a lot in the past.

fearful-symmetry:
@pchila not sure what you mean? The `if runner.testbeatName == "elastic-agent" ` line was just the simplest way to constrain the tests, considering this entire PR is meant to be reverted once buildkite is available in the beats repo.

pchila (approved):

Issue #2509, opened 2023-04-18
Agent monitoring server is binding to all interfaces instead of localhost


andrewkroh:
For confirmed bugs, please report:
- Version: 8.6.2 (build: 913c02bea9b13dec4d5c5f3057b5b397344e3298 at 2023-02-13 16:52:36 +0000 UTC)
- Operating System: macOS 13.3.1 (22E261)
- Steps to Reproduce:
1. Deploy a Fleet managed Elastic Agent on macOS.
2. Observe the listening ports. (use netstat, Little Snitch, etc)

```json
[
  {
    "connections": [
      {
        "bytesDown": "0",
        "bytesUp": "0",
        "interface": "",
        "localAddress": "::",
        "localPort": "6791",
        "protocol": "TCP",
        "remoteAddress": "::",
        "remoteHostName": "::",
        "remotePort": "0",
        "state": "Listen"
      }
    ],
    "process": {
      "path": "/Library/Elastic/Agent/data/elastic-agent-913c02/elastic-agent.app/Contents/MacOS/elastic-agent",
      "pid": "837",
      "signature(s)": {
        "signatureStatus": 100013
      }
    }
  }
]
```

```
netstat -an | grep 6791
tcp46      0      0  *.6791                 *.*                    LISTEN
```

I expected to see something like this where it binding only to the loopback interface.

```
# MOCK DATA
tcp4       0      0  127.0.0.1.6791         *.*                    LISTEN
```

This may be a bug because the Agent code already appears to bind to localhost by default.

https://github.com/elastic/elastic-agent/blob/d7adf66bc4e462abed24b95ee48b0671b8a952ab/internal/pkg/core/monitoring/config/config.go#L56-L60

WiegerElastic:
Wanted to share the output on my machine (MacOS 13.3.1) running Agent version 8.7.0.

<img width="640" alt="image" src="https://user-images.githubusercontent.com/97174155/232885118-18de52a1-37f8-47b6-9e65-b8748a48386f.png">

<img width="779" alt="image" src="https://user-images.githubusercontent.com/97174155/232887402-74e260af-8e3a-4a9c-ac4b-59bebb9beb0f.png">

SpencerLN:
Are there any plans to prioritize fixing this? 

I noticed that `elastic-agent` seems to listen on all interfaces even when `agent.monitoring.http.enabled: false` is set:
```
agent:
  download:
    sourceURI: https://artifacts.elastic.co/downloads/
[...]

PR #3662, opened 2023-10-26
Disable exposing metrics server on all interfaces

michalpristas:
## What does this PR do?

This PR disables option to configure agent in a way that it will expose monitoring agent on all interfaces. 
Empty value for `agent.monitoring.http.host` is now disabled and is replaced by localhost

Agents upgraded from <8.5 do have empty host due to change in defaults in 8.5. These agents are exposing metrics endpoint on all interfaces not on purpose.

Guard is implemented at the time of configuration unpacking and at the time of server creation to minimize future misbehavior with possible code changes 

Fixes: https://github.com/elastic/elastic-agent/issues/2509

### DIFF
--- a/internal/pkg/core/monitoring/config/config_test.go
+++ b/internal/pkg/core/monitoring/config/config_test.go
@@ -13,6 +13,51 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/config"
 )
 
+func TestHost(t *testing.T) {
+	testCases := []struct {
+		name         string
+		config       string
+		expectedHost string
+	}{
+		{"no host", `enabled: true
+logs: true
+metrics: true
+http:
+  enabled: true`, defaultHost},
+		{"empty host", `enabled: true
+logs: true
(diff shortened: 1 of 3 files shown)


michalpristas:
ci failing on processing failed results 🤦 
integ.tests ok

internal/pkg/core/monitoring/config/config_test.go
```diff
@@ -13,6 +13,57 @@ import (
...
+http:
+  enabled: true
+  host: ""`, DefaultHost},
+		{"whitespace host", `enabled: true
+logs: true
+metrics: true
+http:
+  enabled: true
+  host: "   "`, DefaultHost},
+		{"default", `enabled: true
```
AndersonQ, line 39: [Suggestion]
```suggestion
		{"default host", `enabled: true
```

AndersonQ (approved):

pchila (approved):
LGTM

PR #3680, opened 2023-11-01
[CI]Buildkite k8s tests migration

pazone:
## What does this PR do?
Migrates k8s tests to buildkite.
It now runs tests on push to `main`, `7.x` and `8.x` against next k8s versions:

- 1.28.0
- 1.27.3
- 1.26.6

Here is the [example execution of k8s tests](https://buildkite.com/elastic/elastic-agent/builds/4555#018b92d9-4f8d-476b-968f-23f992693441) on buildkite

Currently, it runs k8s tests on every pull request. Anyway, it takes a few minutes. 
I have a plan to create a conditional step using [monorepo-diff-buildkite-plugin](https://github.com/monebag/monorepo-diff-buildkite-plugin). But it must be thoroughly tested.


.ci/Jenkinsfile
```diff
@@ -248,7 +194,7 @@ pipeline {
         beforeAgent true
         anyOf {
           not { changeRequest() }
-          expression { return isExtendedWindowsEnabled() && env.ONLY_DOCS == "false"}
+          // expression { return isExtendedWindowsEnabled() && env.ONLY_DOCS == "false"}
```
pchila, line 197: Why has this been commented ?

.buildkite/pipeline.yml
```diff
@@ -168,3 +168,30 @@ steps:
     plugins:
       - junit-annotate#v2.4.1:
           artifacts: build/TEST-go-integration*.xml
+
+  # Branch steps
+  - group: "K8s tests"
+    key: "k8s-tests"
+    steps:
+      - label: "K8s tests: {{matrix.k8s_version}}"
+        branches: "main 7.* 8.* v7.* v8.* buildkite_k8s_tests"
```
pchila, line 177: the `buildkite_k8s_tests` should make it run on the source branch but I don't see the k8s tests running anywhere... Is this on purpose or am I misreading the intention here ?

pchila (commented):
Code looks reasonable, however, to have a more informed opinion we should probably involve @gizas and his team to take a look.

As a future improvement I would like to see the docker image being packaged in the pipeline, loaded into kind and tested right away instead of relying on the docker registry.

Finally, is there a run of k8s tests in buildkite anywhere ?

.ci/Jenkinsfile
pazone: It will be removed soon

.buildkite/pipeline.yml
```diff
@@ -168,3 +168,30 @@ steps:
...
+          - "build/TEST-**"
+          - "build/diagnostics/*"
+          - "coverage.out"
+        agents:
+          provider: "gcp"
+          image: "family/core-ubuntu-2204"
+        matrix:
+          setup:
+            k8s_version:
+              - "1.28.0"
```
gizas, line 190: I see you have aligned  with this https://github.com/elastic/elastic-agent/pull/3676?
cc @constanca-m to see any diff

pazone: That's right. I picked the versions according to this PR

.buildkite/pipeline.yml
```diff
@@ -168,3 +168,28 @@ steps:
...
+          K8S_VERSION: "v{{matrix.k8s_version}}"
+          KIND_VERSION: "v0.20.0"
+        command: ".buildkite/scripts/steps/k8s-tests.sh"
+        artifact_paths:
+          - "build/TEST-**"
+          - "build/diagnostics/*"
+          - "coverage.out"
+        agents:
+          provider: "gcp"
+          image: "family/core-ubuntu-2204"
```
gizas, line 186: Does this equal to `ubuntu-22 && immutable` that we specify in Jenkinsfile?

.buildkite/scripts/install-kind.sh
```diff
@@ -0,0 +1,46 @@
...
+
+echo "--- Install Kind"
+
+MSG="environment variable missing."
+DEFAULT_HOME="/usr/local"
+KIND_VERSION=${KIND_VERSION:?$MSG}
+HOME=${HOME:?$DEFAULT_HOME}
+KIND_CMD="${HOME}/bin/kind"
+
+if command -v kind
```
alexsapran, line 12: What that condition here do?

If I understand we check that there is no pre-installed version of Kind that we don't want. How this can be possible?

We update the `PATH` on line `38` so at this stage even if Kind existed in the path, it would not be resolved.

.buildkite/scripts/install-kind.sh
```diff
@@ -0,0 +1,46 @@
...
+    ARCH_SUFFIX=arm64
+else
+    ARCH_SUFFIX=amd64
+fi
+
+mkdir -p "${HOME}/bin"
+
+if curl -sSLo "${KIND_CMD}" "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-${OS}-${ARCH_SUFFIX}" ; then
+    chmod +x "${KIND_CMD}"
+    export PATH=$HOME/bin:${PATH}
```
alexsapran, line 39: we might want to update this and move it into the pipeline env, WDYT?

.buildkite/scripts/install-kubectl.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "--- Install kubectl"
+
+MSG="parameter missing."
+DEFAULT_HOME="/usr/local"
+K8S_VERSION=${K8S_VERSION:?$MSG}
+HOME=${HOME:?$DEFAULT_HOME}
+KUBECTL_CMD="${HOME}/bin/kubectl"
+
+if command -v kubectl
```
alexsapran, line 12: similar to my previous comment, this works because we happened to export PATH in the previous script. I think this is some implicit behavior which can lead to trouble in the future, that's why I think the update of the PATH should be moved somewhere more visible ex, the pipeline

.buildkite/scripts/steps/k8s-tests.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+set -euo pipefail
+
+source .buildkite/scripts/install-kubectl.sh
+source .buildkite/scripts/install-kind.sh
+
+echo $PATH
```
alexsapran, line 7: This is for debugging reasons right?

.buildkite/pipeline.yml
```diff
@@ -168,3 +168,28 @@ steps:
...
+  - group: "K8s tests"
+    key: "k8s-tests"
+    steps:
+      - label: "K8s tests: {{matrix.k8s_version}}"
+        env:
+          K8S_VERSION: "v{{matrix.k8s_version}}"
+          KIND_VERSION: "v0.20.0"
+        command: ".buildkite/scripts/steps/k8s-tests.sh"
+        artifact_paths:
+          - "build/TEST-**"
```
alexsapran, line 181: I wonder what type of reports this generates, do we get JUnit reports?
If so it would be nice to incorporate the following plugin as well.
```
plugins:
      - junit-annotate#v2.4.1:
          artifacts: build/TEST-*.xml
```

So that in case of failure, we get the nice BK annotation and it would save some time to figure out what failed.

.buildkite/pipeline.yml
pazone: now we run it on every PR.

.buildkite/scripts/install-kind.sh
pazone: Do you mean the line 37? If yes then we also need to move `$ARCH` and `$OS` variables which are dynamically initialised

.buildkite/scripts/steps/k8s-tests.sh
pazone: yeah :)

.buildkite/scripts/install-kind.sh
pazone: We check if `kind` with the corresponding version is already installed. If not, we install it. I plan to pre-install `kind` by building custom VM images for elastic-agent CI. Now we install it on the fly

.buildkite/pipeline.yml
pazone: In general, these are both Ubuntu 22.04 images. The main difference is in preinstalled agents and tools because we migrate to buildkite

pazone: I think these are more or less equal ubuntu 22.04 VMs, except for the differences in preinstalled tools because buildkite requires another tool set.

.buildkite/scripts/install-kind.sh
alexsapran: Yes, but my comment I think still stands. 

When this script runs, and let's assume `kind` is already downloaded in the `bin` folder, because the way we define the `PATH` will not locate it under the bin folder, so this `command -v` will always return false. 
Moving `PATH` on the top level, then if its already available under the `bin` folder then it will be resolved and compared.

pazone: Moved the `PATH` declaration to the top-level script. Thanks @alexsapran for the nice catch!

alexsapran (approved):

pazone:
@gizas @AndersonQ @pchila Any comments, suggestions, approvals? :)

pchila (approved):
LGTM but I advise you wait also for the other reviewers feedback after the modifications ;)

PR #3681, opened 2023-11-01
[Buildkite][Integration tests] Change ESS region to `aws-eu-central-1`

ycombinator:
## What does this PR do?

This PR changes the ESS region used by integration tests, when run in Buildkite, to `aws-eu-central-1`. 

## Why is it important?

This is the only region that's working at the moment.  In other regions, we're experiencing a Kibana failure, which is not only causing integration tests to fail because the deployment cannot come up but it's also leaving these broken deployments orphaned which costs money!

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -10,7 +10,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent" && "$BUILDKITE_STEP_KEY" == 
 
   # Perform cleanup of integration tests resources
   echo "--- Cleaning up integration test resources"
-  TEST_INTEG_AUTH_ESS_REGION=aws-eu-central-1 SNAPSHOT=true mage integration:clean
+  TEST_INTEG_AUTH_ESS_REGION=us-east-1 SNAPSHOT=true mage integration:clean
 fi
 
 if [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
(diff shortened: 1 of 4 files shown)


.buildkite/hooks/pre-exit
```diff
@@ -10,7 +10,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent" && "$BUILDKITE_STEP_KEY" ==
 
   # Perform cleanup of integration tests resources
   echo "--- Cleaning up integration test resources"
-  TEST_INTEG_AUTH_ESS_REGION=azure-eastus2 SNAPSHOT=true mage integration:clean
+  TEST_INTEG_AUTH_ESS_REGION=aws-eu-central-1 SNAPSHOT=true mage integration:clean
```
cmacknz, line 13: I originally tested with us-east-1 but it seems this works as well (I just created a deployment manually to check).

Only thing is if our buildkite instances are in NA then this will be cross-continental traffic in the tests.

belimawr (approved):

.buildkite/hooks/pre-exit
ycombinator: > Only thing is if our buildkite instances are in NA then this will be cross-continental traffic in the tests. 

Ooof, yeah, good point. Let me switch it to `us-east-1`.

ycombinator: Changed in 37e5933f4197dc90a8690a8c207366d2ddb26bbe.

cmacknz:
We seem to be hitting a totally unrelated error now:

```
Parsing /build/TEST-go-integration.xml
--
  | 2023-11-01 17:17:31 UTC | /usr/local/lib/ruby/gems/3.1.0/gems/rexml-3.2.5/lib/rexml/parsers/treeparser.rb:96:in `rescue in parse': #<RuntimeError: Illegal character "\\u0000" in raw string
```

cmacknz:
buildkite test it

cmacknz:
The integration tests did pass in that run though

cmacknz:
OK the integration tests passed again, I am going to force merge this since this puts us in a better place than we were before. We hit another known flaky test and failed to get the coverage report (likely because the test failed).

Issue #2789, opened 2023-06-06
Coordinator's Run function never retries on error


faec:
`Coordinator.Run` has a `for` loop that dispatches to the internal helper `Coordinator.runner` and checks the result:

```go
func (c *Coordinator) Run(ctx context.Context) error {
  ...
  for {
    c.state.UpdateState(state.WithState(agentclient.Starting, "Waiting for initial configuration and composable variables"))
    err := c.runner(ctx)
    if err != nil {
      ...
```

The checks here return from `Coordinator.Run` if the context is cancelled or the return value is `ErrFatalCoordinator`, otherwise the loop starts again with another call to `c.runner`.

The problem is, `Coordinator.runner` never returns anything except `ErrFatalCoordinator` (unless the parent context in `Coordinator.Run` has already expired, indicating an external shutdown) -- in either case, the `for` loop will never run a second time, so no real retrying is happening here.

This happens because `c.runner's` return value is generated by `Coordinator.handleCoordinatorDone`, a helper function that collects the return values from the three managers (config, runtime, vars) and, if any of them are errors, collects them all into an `ErrFatalCoordinator`. Since `handleCoordinatorDone` is never called in the first place until there has been at least one error, this is in practice the only possible return value until final Agent shutdown.

Discussing this issue @blakerouse confirmed that we want to retry here when possible, so we need a way to decide which errors are considered fatal (currently the possible errors from the various managers are undocumented so this may take some investigation), and when only non-fatal errors are returned from the managers we should do a rate-limited retry in `Coordinator.Run`.

Note: The current form of `handleCoordinatorDone` came out of a [fix for several deadlocks](https://github.com/elastic/elastic-agent/pull/2352), so this change is potentially sensitive -- it is better to be overcautious about which errors are considered fatal than to risk permanently deadlocking the `Coordinator`.

faec:
Reminder since it looks like this is back in the sprint planning: right now we don't have a defined list of what errors the various managers can return, or a way to decide which ones should be fatal. (There was some work on clarifying these earlier this summer but that was deferred for more urgent tasks.) Mistakenly retrying here for a misclassified error could block Agent shutdown.

cmacknz:
If it is dangerous to mistakenly retry, and the only error we know about today is fatal, then is it simpler to remove this misleading retry logic entirely and always exit on error? 

The agent is almost always running in an environment where it is running as a service or k8s pod where it will restart automatically if the process exits unexpectedly.

[...]

PR #3688, opened 2023-11-01
Clean up Agent Coordinator and manager shutdown logic

faec:
Clean up `Coordinator` shutdown handling and fix the no-op retry loop described in https://github.com/elastic/elastic-agent/issues/2789.

The previous retry logic in `(*Coordinator).Run` never took effect: the only possible error return was considered fatal, and so there was never an actual retry. This returned error (`ErrFatalCoordinator`) was a placeholder returned in part out of caution, since accidentally retrying a fatal error could deadlock the Agent.

Following discussion in https://github.com/elastic/elastic-agent/issues/2789, the new model is:
- All errors that cause a Manager's Run routine to terminate are considered fatal
- Managers should send non-fatal errors to the manager's existing error channel instead of returning

(This is already how managers work in practice.)

This PR:
- Removes the retry loop in `(*Coordinator).Run` and instead explicitly passes on error results returned by `runner`
- Removes the `ErrFatalCoordinator` placeholder error and instead reports any Manager failures directly
- Renames `(*Coordinator).handleCoordinatorDone` to a standalone function `collectManagerErrors` to better reflect its behavior, and makes its timeout a parameter so tests don't need to mess with a shared global variable
- Reports manager shutdown timeouts as explicit errors

This is cleanup only. There should be no functional change from this PR except that some shutdown errors give more precise messages.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/2789

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -1218,7 +1218,7 @@ func (c *Coordinator) filterByCapabilities(comps []component.Component) []compon
 
 // collectManagerErrors listens on the shutdown channels for the
 // runtime, config, and vars managers, and waits for up to
-// CoordinationShutdownTimeout for them to report their final status.
+// the specified timeout for them to report their final status.
 // It returns any resulting errors as a multierror, or nil if no errors
 // were reported.
 // Called on the main Coordinator goroutine.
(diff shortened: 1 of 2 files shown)


michalpristas (approved):
lgtm, i was a bit worried about for loop removal, but this should be fine as we're looping in runner until ctx is cancelled.

blakerouse (approved):
Thanks for the cleanup. Much easier to follow with this change and makes the behavior consistent.

PR #3615, opened 2023-10-17
Include upgrade details in output of `elastic-agent status` during ongoing upgrade

ycombinator:
## What does this PR do?

This PR enhances the output of `elastic-agent status` to include upgrade details when Agent is being upgraded.

## Why is it important?

To gain more visibility into the Agent upgrade process.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] ~I have made corresponding changes to the documentation~
- [ ] ~I have made corresponding change to the default configuration files~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] ~I have added an integration test or an E2E test~

## How to test this PR locally

1. Build `elastic-agent` from this PR's branch and install it.
2. Run `elastic-agent status --output json`. Verify that NO upgrade details are shown.
3. In a new window, upgrade Elastic Agent to a version that hasn't been released yet.  This will ensure that the upgrade starts but remains stalled.
   ```
   sudo elastic-agent upgrade 8.13.0
   ```
4. In the original window, run `elastic-agent status --output json` again.  Verify that upgrade details are included.

## Related issues

- Relates to https://github.com/elastic/elastic-agent/issues/3119#issuecomment-1743638526

PR #3694, opened 2023-11-02
[Upgrade Details] Implement tracking of `UPG_ROLLBACK` state

ycombinator:
## What does this PR do?

This PR enhances the Elastic Agent to track the `UPG_ROLLBACK` state.

## Why is it important?

When an Agent upgrade is rolled back, we want to know about it.

## Checklist

- [x] ~I have added tests that prove my fix is effective or that my feature works~
- [x] ~I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~

## Related issues

- Closes #123
- Relates #123
- Requires #123
- Superseds #123
-->
- Relates to https://github.com/elastic/elastic-agent/issues/3119#issuecomment-1743638526

## Questions to ask yourself

- ...


testing/integration/upgrade_rollback_test.go
```diff
@@ -119,6 +120,12 @@ inputs:
 	}
 	require.NoError(t, err)
 
+	// ensure that upgrade details now show the state as UPG_ROLLBACK
+	state, err := startFixture.Client().State(ctx)
+	require.NoError(t, err)
+	require.NotNil(t, state.UpgradeDetails) // FIXME: there is no UpgradeDetails on State (yet!)
```
ycombinator, line 126: The `UpgradeDetails` field on `state` will become available once https://github.com/elastic/elastic-agent/pull/3615 is merged.

ycombinator:
Unit test failure in CI is relevant to this PR:

```
=== Failed
--
  | === FAIL: internal/pkg/agent/application/upgrade TestWatchMarker (0.05s)
  | step_mark_test.go:79:
  | Error Trace:	/Users/admin/builds/bk-agent-prod-orka-1699028124327725357/elastic/elastic-agent/internal/pkg/agent/application/upgrade/step_mark_test.go:79
  | /Users/admin/.gvm/versions/go1.20.10.darwin.amd64/src/runtime/asm_amd64.s:1598
  | Error:      	Expected value not to be nil.
  | Test:       	TestWatchMarker
```

Will investigate and fix.

ycombinator:
Rollback-related integration tests are failing.  Most likely related to the changes in this PR.  I'm investigating...

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+	// is already present before the watching starts.
+	marker, err := loadMarker(mfw.markerFilePath)
+	if err != nil {
+		return fmt.Errorf("unable to load upgrade marker from watch: %w", err)
+	}
+	if marker != nil && marker.Details != nil {
+		mfw.updateCh <- *marker
+	}
+
+	<-ctx.Done()
```
pchila, line 118: we launch a goroutine but then we block here... 
We have the goroutine just to start watching the file as soon as possible while we try and do the first read? 
Couldn't we refactor this to have a single go routine handle the watch and the first read? This would avoid possible races if a write happens between the start of the goroutine and the first read and both try to send the same marker on the update channel

internal/pkg/agent/application/upgrade/marker_watcher_test.go
```diff
@@ -0,0 +1,84 @@
...
+				testDetailsMu.Lock()
+				testDetails = marker.Details
+				testDetailsMu.Unlock()
+			}
+		}
+	}()
+
+	go func() {
+		err := markerWatcher.Run(testCtx)
+		require.NoError(t, err)
```
pchila, line 56: Because of `t.FailNow()` called by `require`, shouldn't the `require.*` be called from the goroutine running the test function?

internal/pkg/agent/application/upgrade/marker_watcher_test_util.go
```diff
@@ -0,0 +1,30 @@
...
+
+// CreateMarkerTestDir creates a temporary directory that's needed
+// for the MarkerFileWatcher to function correctly. This allows tests that
+// directly or indirectly use the MarkerFileWatcher to succeed. The
+// temporary directory is cleaned up when the test completes.
+func CreateMarkerTestDir(t *testing.T) {
+	execPath, err := os.Executable()
+	require.NoError(t, err)
+
+	testDataDir := filepath.Join(filepath.Dir(execPath), "data")
```
pchila, line 23: Since this is only to be used in tests could you rename the file to `marker_watcher_util_test.go` or make it available only to test code in some way ?

pchila (commented):
A couple of comments mostly about concurrency and encapsulation, the change looks ok for the rest

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1224,14 +1254,15 @@ func (c *Coordinator) filterByCapabilities(comps []component.Component) []compon
...
+func collectManagerErrors(timeout time.Duration, varsErrCh, runtimeErrCh, configErrCh, upgradeMarkerWatchErrCh chan error) error {
 	var runtimeErr error
 	var configErr error
 	var varsErr error
+	var upgradeMarkerErr error
 	// in case other components are locked up, let us time out
 	timeoutWait := time.NewTimer(timeout)
 	defer timeoutWait.Stop()
-	var returnedRuntime, returnedConfig, returnedVars bool
+	var returnedRuntime, returnedConfig, returnedVars, returnedUpgradeMarker bool
```
AndersonQ, line 1265: [NIT]
This could be more consistent

```go
	var runtimeErr, configErr, varsErr, upgradeMarkerErr error
	var returnedRuntime, returnedConfig, returnedVars, returnedUpgradeMarker bool

	// in case other components are locked up, let us time out
	timeoutWait := time.NewTimer(timeout)
	defer timeoutWait.Stop()
```

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -876,14 +890,25 @@ func (c *Coordinator) runner(ctx context.Context) error {
...
+	if c.upgradeMgr != nil && c.upgradeMgr.MarkerWatcher() != nil {
+		go func() {
+			err := c.upgradeMgr.MarkerWatcher().Run(ctx)
+			cancel()
+			upgradeMarkerWatcherErrCh <- err
+		}()
+	} else {
+		upgradeMarkerWatcherErrCh <- nil
+	}
+
```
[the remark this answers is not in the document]
AndersonQ: The [collectManagerErrors](https://github.com/elastic/elastic-agent/pull/3694/files#diff-efef2d2536ab052feff95a0b0565325806b28e733bb9f2abc319b646194d4de5R1290-R1291) uses the error to know the upgrade marker run (`returnedUpgradeMarker`)

internal/pkg/agent/application/upgrade/details/details.go
```diff
@@ -68,6 +68,7 @@ func NewDetails(targetVersion string, initialState State, actionID string) *Deta
 
 // SetState is a convenience method to set the state of the upgrade and
 // notify all observers.
+// Do NOT call SetState with StateFailed; call the Fail method instead.
```
AndersonQ, line 71: Why? So I'm guessing the Fail method is much more than just sugar syntax, am I right?

internal/pkg/agent/application/upgrade/details/details.go
```diff
@@ -118,6 +119,14 @@ func (d *Details) RegisterObserver(observer Observer) {
 	d.notifyObserver(observer)
 }
 
+// Equals compares the non-lock fields of two Details structs.
+func (d *Details) Equals(otherD *Details) bool {
```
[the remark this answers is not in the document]
AndersonQ: +1

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+	errCh    chan error
+}
+
+func newMarkerFileWatcher(upgradeMarkerFilePath string, logger *logger.Logger) (MarkerWatcher, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create upgrade marker watcher: %w", err)
+	} // Watch the upgrade marker file's directory, not the file itself, so we
+
+	// notice the file even if it's deleted and recreated.
```
AndersonQ, line 40: [Suggestion]
I guess those are the same comment, right?
```suggestion
	} 
	
	// Watch the upgrade marker file's directory, not the file itself, so we
	// notice the file even if it's deleted and recreated.
```

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (mfw *MarkerFileWatcher) Run(ctx context.Context) error {
+	// Handle watching
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case err, ok := <-mfw.watcher.Errors:
+				mfw.logger.Debug("after there are watch errors")
```
AndersonQ, line 72: [Suggestion]
Could you add a wee more context to be more clear after _what_ there are watch errors?

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+				if !ok { // Channel was closed (i.e. Watcher.Close() was called).
+					mfw.logger.Debug("upgrade marker watch's error channel was closed")
+					return
+				}
+				mfw.errCh <- fmt.Errorf("upgrade marker watch returned error: %w", err)
+				continue
+			case e, ok := <-mfw.watcher.Events:
+				mfw.logger.Debugf("after there is a watch event: [%s]", e.String())
```
AndersonQ, line 80: Again, after _what_?

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+				if !ok { // Channel was closed (i.e. Watcher.Close() was called).
+					mfw.logger.Debug("upgrade marker watch's events channel was closed")
+					return
+				}
+
+				if e.Name != mfw.markerFilePath {
+					// Event is for a file other than the upgrade marker; ignore it.
+					continue
+				}
```
AndersonQ, line 89: [Suggestion]
if it isn't expected to happen, I'd say there should be a warning log, if it's expected to happen, could you say so in the comment?

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+	}()
+
+	// Do an initial read from the upgrade marker file, in case the file
+	// is already present before the watching starts.
+	marker, err := loadMarker(mfw.markerFilePath)
+	if err != nil {
+		return fmt.Errorf("unable to load upgrade marker from watch: %w", err)
```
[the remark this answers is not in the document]
AndersonQ: +1

internal/pkg/agent/application/upgrade/marker_watcher.go
AndersonQ: I agree with Paolo, if it's gonna block anyway, I don't see why the goroutine is needed.

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+
+					mfw.updateCh <- *marker
+				}
+			}
+		}
+	}()
+
+	// Do an initial read from the upgrade marker file, in case the file
+	// is already present before the watching starts.
+	marker, err := loadMarker(mfw.markerFilePath)
```
AndersonQ, line 110: if it's the first read, shouldn't it happen before the goroutine be started?

internal/pkg/agent/application/upgrade/marker_watcher_test.go
AndersonQ: yeah, `require` does not work on a goroutine

AndersonQ: if you need to abort the test, you'd need to signal the test goroutine to exit

internal/pkg/agent/application/upgrade/marker_watcher_test_util.go
AndersonQ: +1

AndersonQ: or just move it to the file where it's used

internal/pkg/agent/cmd/watch.go
```diff
@@ -107,10 +108,35 @@ func watchCmd(log *logp.Logger, cfg *configuration.Configuration) error {
 	errorCheckInterval := cfg.Settings.Upgrade.Watcher.ErrorCheck.Interval
 	ctx := context.Background()
 	if err := watch(ctx, tilGrace, errorCheckInterval, log); err != nil {
-		log.Error("Error detected proceeding to rollback: %v", err)
+		log.Error("Error detected, proceeding to rollback: %v", err)
```
AndersonQ, line 111: +1

internal/pkg/agent/application/coordinator/coordinator.go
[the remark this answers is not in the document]
ycombinator: Yeah, it's what @AndersonQ said.  It's the same pattern followed in the blocks of code directly above this one for the runtime manager, the config manager, and the vars manager.

internal/pkg/agent/application/upgrade/details/details.go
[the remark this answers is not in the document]
ycombinator: Yeah, good catch.  This method was originally comparing the struct values so I didn't have such checks but it makes sense to add them now.  Thanks!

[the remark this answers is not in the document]
ycombinator: Fixed in 9171b1ada787ef3eb92b4bd5952d1e216e6c5a26.

internal/pkg/agent/application/upgrade/details/details.go
ycombinator: That's right. The [`Fail()` method](https://github.com/elastic/elastic-agent/blob/72653ac2b0b57deedd276c3abc8d19f9085b3154/internal/pkg/agent/application/upgrade/details/details.go#L99-L117) sets some other fields besides `state` as well.  It also takes an `error` as an input parameter, otherwise I would've simply called it directly from within `SetState(s State)` if `s == StateFailed`. 

With this godoc line at least there is some warning to the user not to call `SetState(StateFailed)` but call `Fail(err)` instead.  I'm open to better ideas, though.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Yeah, thanks, not sure what happened here 😄.  Maybe something during a rebase + conflict resolution 🤷.

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: Fixed in 1cd2e07938.

internal/pkg/agent/application/upgrade/marker_watcher.go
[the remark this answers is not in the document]
ycombinator: When I wrote this code here I did inspect `loadMarker` to ensure that [it returned `nil, nil` if the file was not present](https://github.com/elastic/elastic-agent/blob/72653ac2b0b57deedd276c3abc8d19f9085b3154/internal/pkg/agent/application/upgrade/step_mark.go#L156-L158).  So if the file does not exist yet or is still being written to, `loadMarker` here will not return an error (which is what we want).

So the only errors that could be returned from `loadMarker` are:
- if there's some error reading the file into memory, or
- if there's some error parsing the contents of the file as YAML into memory

I think the question is whether either of these errors could be transient.  That is, if a call to `loadMarker` were to be attempted again (as part of the goroutine code), could these errors go away or will they always be returned. 

I can't come up with conditions that would make the first error transient but perhaps there are some.

As for the second error, I suppose there's some scenario in which the file is still being written to so only part of the contents have been written. So that would make the second error transient.

To address the (potentially) transient nature of these errors, I'm good with removing the error check here and just checking if the returned `marker` is `nil` (which we are doing already in the code).

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Agreed, I did not implement this well.  Will refactor.  Thanks for calling it out.

internal/pkg/agent/application/upgrade/marker_watcher.go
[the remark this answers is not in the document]
ycombinator: I will address this as part of the refactoring for https://github.com/elastic/elastic-agent/pull/3694#discussion_r1386317575.

internal/pkg/agent/application/coordinator/coordinator.go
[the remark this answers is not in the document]
ycombinator: With the refactoring for https://github.com/elastic/elastic-agent/pull/3694#discussion_r1386317575, the call to `c.upgradeMgr.MarkerWatcher().Run(ctx)` will no longer return an error.  As such, this whole business with the  `upgradeMarkerWatcherErrCh` will go away.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Actually this is a holdover from some debugging I was doing.  This line should just be deleted.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: The line is also holdover from some debugging I was doing. This line should just be deleted.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: It's expected to happen.  I will expand the comment.  Thanks.

internal/pkg/agent/application/coordinator/coordinator.go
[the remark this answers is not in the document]
ycombinator: > With the refactoring for [#3694 (comment)](https://github.com/elastic/elastic-agent/pull/3694#discussion_r1386317575), the call to `c.upgradeMgr.MarkerWatcher().Run(ctx)` will no longer return an error. As such, this whole business with the `upgradeMarkerWatcherErrCh` will go away.

Addressed in 83577da728.

internal/pkg/agent/application/upgrade/marker_watcher.go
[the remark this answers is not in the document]
ycombinator: > I will address this as part of the refactoring for [#3694 (comment)](https://github.com/elastic/elastic-agent/pull/3694#discussion_r1386317575).

Addressed as part of the refactoring done in 83577da728.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Refactored in 83577da728.

internal/pkg/agent/application/upgrade/marker_watcher_test.go
ycombinator: Addressed as part of the refactoring done in 83577da728.

internal/pkg/agent/application/upgrade/marker_watcher_test_util.go
ycombinator: > Since this is only to be used in tests could you rename the file to `marker_watcher_util_test.go` or make it available only to test code in some way ?

If I rename the file `marker_watcher_util_test.go` it only becomes available to test code in the same package.  However, I need this function in a test in another package too.  That's why my solution was to name it `marker_watcher_test_util.go`.  This is the same approach taken by [`logger.NewTesting`](https://github.com/elastic/elastic-agent/blob/72653ac2b0b57deedd276c3abc8d19f9085b3154/pkg/core/logger/testing.go#L15-L26).

> or just move it to the file where it's used

It's used from two different packages, so unfortunately it can't be moved to where it's used (without duplication).

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Deleted as part of the refactoring done in https://github.com/elastic/elastic-agent/commit/83577da72817b8c9817f1cd3f7b467825a9654f8.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Deleted as part of the refactoring done in https://github.com/elastic/elastic-agent/commit/83577da72817b8c9817f1cd3f7b467825a9654f8.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Done in d874790f7c.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Ideally it should happen _after_ the goroutine has started or as part of the goroutine's `select`.  If the first read were done before the goroutine started we could encounter the following undesirable scenario:
1. Upgrade marker file is read (first read).
2. Upgrade marker file gets updated.
3. Goroutine starts, but update event is now "lost".

The refactoring done in 83577da72817b8c9817f1cd3f7b467825a9654f8 ensures that the first read happens in the same goroutine `select` as the waiting on update events from the `fsnotify.Watcher`.

internal/pkg/agent/cmd/watch.go
ycombinator: @AndersonQ What is the "+1" here for?

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Fixed in ea5177522c58e011b59c354ff55954ef5d69e1eb.

ycombinator:
@pchila @AndersonQ Thank you for the great review.  I believe I've addressed all your feedback now.  Please re-review when you get a chance.  Thanks again!

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1225,13 +1246,13 @@ func (c *Coordinator) filterByCapabilities(comps []component.Component) []compon
...
-	var configErr error
-	var varsErr error
+	var runtimeErr, configErr, varsErr error
+	var returnedRuntime, returnedConfig, returnedVars bool
+
 	// in case other components are locked up, let us time out
 	timeoutWait := time.NewTimer(timeout)
 	defer timeoutWait.Stop()
-	var returnedRuntime, returnedConfig, returnedVars bool
+
```
ycombinator, line 1255: Cleanup unrelated to the functional changes introduced in this PR.

internal/pkg/agent/application/upgrade/details/details.go
```diff
@@ -118,6 +119,22 @@ func (d *Details) RegisterObserver(observer Observer) {
...
+	if d == nil && otherD == nil {
+		return true
+	}
+
+	if d == nil || otherD == nil {
+		return false
+	}
```
michalpristas, line 130: up to you: comparing addresses first to avoid comparing metadata

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+				if e.Name != mfw.markerFilePath {
+					// Since we are watching the directory that will contain the upgrade
+					// marker file, we could receive events here for changes to files other
+					// than the upgrade marker. We ignore such events as we're only concerned
+					// with changes to the upgrade marker.
+					continue
+				}
+
+				switch {
+				case e.Op&fsnotify.Create == fsnotify.Create ||
```
michalpristas, line 94: alt. compare `!=0`

internal/pkg/agent/application/upgrade/details/details.go
ycombinator: Sorry, I didn't follow what you're suggesting.

internal/pkg/agent/cmd/watch.go
```diff
@@ -107,10 +108,35 @@ func watchCmd(log *logp.Logger, cfg *configuration.Configuration) error {
...
+		if marker.Details == nil {
+			actionID := ""
+			if marker.Action != nil {
+				actionID = marker.Action.ActionID
+			}
+			marker.Details = details.NewDetails(version.GetAgentPackageVersion(), details.StateRollback, actionID)
+		}
+
+		marker.Details.SetState(details.StateRollback)
+		err = upgrade.SaveMarker(marker)
```
michalpristas, line 126: i wonder if we want to guard this with some retry. we're accessing marker file from 2 processes now and when one process opens marker even for reading the other may fail on windows

ycombinator: Thanks, I was not aware this was an issue on Windows.  So it seems it would make sense to add retry in `upgrade.SaveMarker` and `upgrade.LoadMarker` both.  I'll make the change.

internal/pkg/agent/application/upgrade/details/details.go
ycombinator: I think you're suggesting that if the addresses are equal then we can short circuit and return true.  I'll make that change, thanks.

ycombinator: Done in 96b824b74e.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Done in 42b8a41805.

internal/pkg/agent/cmd/watch.go
ycombinator: Implemented in c174487f9a.  Let me know what you think.

internal/pkg/agent/application/upgrade/details/details_test.go
```diff
@@ -92,3 +92,26 @@ func TestDetailsDownloadRateJSON(t *testing.T) {
...
+	details1.SetDownloadProgress(0.1234, 34.56)
+	details1.Fail(errors.New("download failed"))
+
+	details2 := NewDetails("8.12.0", StateDownloading, "foobar")
+	details2.SetDownloadProgress(0.1234, 34.56)
+	details2.Fail(errors.New("download failed"))
+
+	details3 := NewDetails("8.12.0", StateDownloading, "foobar")
+
+	require.True(t, details1.Equals(details2))
```
michalpristas, line 107:
```suggestion
	require.True(t, details1.Equals(details1))
	require.True(t, details1.Equals(details2))
```

internal/pkg/agent/application/upgrade/marker_access_windows.go
```diff
@@ -0,0 +1,66 @@
...
+// randomized exponential backoff up to markerAccessTimeout duration. This retry
+// mechanism is necessary since the marker file could be accessed by multiple
+// processes (the Upgrade Watcher and the main Agent process) at the same time,
+// which could fail on Windows.
+func writeMarkerFile(markerFile string, markerBytes []byte) error {
+	writeFn := func() (err error) {
+		return os.WriteFile(markerFile, markerBytes, 0600)
+	}
+
+	if err := accessMarkerFileWithRetries(writeFn); err != nil {
```
michalpristas, line 49: let's see how it behaves and if there will be any risk of out of order writes. e.g retry writes after another call succeeds. for that we would need to implement some queue mechanism but that sounds like too much and it's possible this will just work

internal/pkg/agent/application/upgrade/step_download_test.go
```diff
@@ -83,13 +83,17 @@ func TestDownloadWithRetries(t *testing.T) {
 		},
 	}
 
+	CreateMarkerTestDir(t)
```
michalpristas, line 86: want to do cleanup after?

michalpristas (approved):
I think it looks OK, i let others comment on whether they see their comments addressed

internal/pkg/agent/application/upgrade/step_download_test.go
ycombinator: The `CreateMarkerTestDir` does the cleanup.  That's why it's passed the `t`.

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,124 @@
...
+func newMarkerFileWatcher(upgradeMarkerFilePath string, logger *logger.Logger) (MarkerWatcher, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create upgrade marker watcher: %w", err)
+	}
+
+	// Watch the upgrade marker file's directory, not the file itself, so we
+	// notice the file even if it's deleted and recreated.
+	upgradeMarkerDirPath := filepath.Dir(upgradeMarkerFilePath)
+	err = watcher.Add(upgradeMarkerDirPath)
```
pchila, line 41: Why do we add the files to watch here but we instantiate the goroutine to consume events only in Run() ?

Can't we instantiate the watcher directly in `Run()` so that we can also add a 
```
defer watcher.Close()
```
on the goroutine, making sure we close the watcher ?
If that is the case maybe we should document it on the interface stating that,  whatever happens,  a created watcher must always be closed, even if we didn't call Run() or other functions on it

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+	// Handle watching
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case err, ok := <-mfw.watcher.Errors:
+				if !ok { // Channel was closed (i.e. Watcher.Close() was called).
+					mfw.logger.Debug("fsnotify.Watcher's error channel was closed")
+					return
```
pchila, line 75: shouldn't we close the watcher when returning? Or we are relying on the caller to always do it ?

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+					continue
+				}
+
+				switch {
+				case e.Op&(fsnotify.Create|fsnotify.Write) != 0:
+					// Upgrade marker file was created or updated; read its contents
+					// and send them over the update channel.
+					mfw.processMarker()
+				}
+			case <-doInitialRead:
```
pchila, line 99: This now makes Run() asynchronous (before we were blocking on `<- ctx.Done()`)
What I mean with my comments on the previous version of this was that since Run() was blocking we didn't need the extra goroutine :sweat_smile: .
Did the wait move somewhere else ?

internal/pkg/agent/application/upgrade/marker_watcher_test_util.go
pchila: I would rather have the duplication and break the unit test dependency between the 2 packages...
Why would this detail be needed in 2 packages? Is there more than 1 package that needs to access the marker file directly and that dependency cannot be mocked away ?

pchila (commented):
Code looks ok as far as the feature is concerned.
There are a couple of clarifications needed on the MarkerWatcher usage and lifecycle and some nitpicks about package dependency (non-blocking)

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: While `Run()` was blocking on `<-ctx.Done()` in the previous version, `Run()` was being called from inside a goroutine, effectively making it asynchronous.  

Now, `Run()` itself is asynchronous and I've removed the goroutine wrapper around the call to `Run()`.

internal/pkg/agent/application/upgrade/marker_watcher_test_util.go
ycombinator: >  Is there more than 1 package that needs to access the marker file directly and that dependency cannot be mocked away ?

The unit test outside this package that calls his helper function is [`application.TestLimitsLog`](https://github.com/elastic/elastic-agent/blob/64563943437a4703e56498ca215a2a0881e4984b/internal/pkg/agent/application/application_test.go#L50).  If I remove the call to `upgrade.CreateMarkerTestDir(t)` from that test, it fails like so:

```
=== RUN   TestLimitsLog
    application_test.go:69: 
        	Error Trace:	/Users/shaunak/development/github/elastic-agent/internal/pkg/agent/application/application_test.go:69
        	            	failed to create upgrader: failed to set watch on upgrade marker's directory [/private/var/folders/mc/qr61t2n91z3g63ffq5sy4cm00000gn/T/GoLand/data]: lstat /private/var/folders/mc/qr61t2n91z3g63ffq5sy4cm00000gn/T/GoLand/data: no such file or directory
        	Test:       	TestLimitsLog
--- FAIL: TestLimitsLog (0.00s)
```

That's because that unit test calls `Application.New`, which ends up eventually calling `upgrade.newMarkerFileWatcher()`, which is where this error is coming from because the upgrade marker's directory doesn't exist at that point.  In normal (non-test) execution, the directory exists because it's created at installation time, before Agent is run and `Application.New` is called.

Anyway, I think we can remove this dependency if I move the instantiation of the watch inside Run() as you're alluding to in https://github.com/elastic/elastic-agent/pull/3694#discussion_r1389062050.  So let me give that a shot and see if it works.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: Yes, it should be possible to do all this work inside `Run()`.  Let me give it a shot.

ycombinator: The reason I added the files to watch in the constructor instead of `Run()` was because I only wanted the files to be added to the watch once per `MarkerFileWatcher` instance.  In theory, the calling code could call `Run()` repeatedly on the same `MarkerFileWatcher` instance, and I didn't want to add the same files to the watch each time.  I don't know how `fsnotify` behaves if the same path is watched multiple times — will it emit duplicate events? Or is it smart enough to de-dupe paths each time they're added? 

Let me play with `fsnotify` some more to see if watching the same paths is safe.  If it is, I think we can greatly simplify the implementation here.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: As of the current implementation, we're relying on the caller to always do it but this could all change depending on what I find after https://github.com/elastic/elastic-agent/pull/3694#discussion_r1389177469.

AndersonQ (approved):

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: I experimented with `fsnotify` and confirmed that the `Add()` method is safe.  With that, I was able to simplify the implementation of the `UpgradeMarkerWatcher` quite a bit in 8b9d67c797.

internal/pkg/agent/application/upgrade/marker_watcher.go
ycombinator: After simplifying the `UpgradeMarkerWatcher` implementation in 8b9d67c797848ffc33c2c3fe39bf9751b2aded64, I was able to move the `fsnotify` watcher's initialization and closing into the `Run()` method in 49d16e94ef.  As such, the caller doesn't need to do anything any more.

internal/pkg/agent/application/upgrade/marker_watcher_test_util.go
ycombinator: As expected, after simplifying the implementation of `UpgradeMarkerWatcher` in 8b9d67c797848ffc33c2c3fe39bf9751b2aded64, I was able to remove this dependency as the `UpgradeMarkerWatcher` constructor no longer adds the watch on the upgrade marker's directory; that is now being done in the `Run()` method instead, which is not in the execution path of this test.

pchila (approved):
LGTM

internal/pkg/agent/cmd/watch.go
```diff
@@ -107,10 +108,35 @@ func watchCmd(log *logp.Logger, cfg *configuration.Configuration) error {
...
 	ctx := context.Background()
 	if err := watch(ctx, tilGrace, errorCheckInterval, log); err != nil {
-		log.Error("Error detected proceeding to rollback: %v", err)
+		log.Error("Error detected, proceeding to rollback: %v", err)
+
+		// If we are upgrading from version >= 8.12.0, marker.Details should be non-nil
+		// because the Agent we upgraded FROM would've written upgrade details in the upgrade
+		// marker. However, if we're upgrading from version < 8.12.0, the marker won't
+		// contain upgrade details, so we populate them now.
+		if marker.Details == nil {
```
cmacknz, line 117: Might be worth adding a log line when this happens, in case it happens unexpectedly

internal/pkg/agent/cmd/watch.go
cmacknz: That this can fail makes detecting the upgrade rollback state best effort. We might be able to improve this, and make it so that only getting the reason for a rollback is best effort.

When the agent first starts up it reads the upgrade marker to determine if it needs to acknowledge a pending upgrade action. This eventually ends up here:

https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/internal/pkg/agent/application/upgrade/upgrade.go#L210-L236

If the agent at this point sees that there is an upgrade marker file, and the target version does not match the agent's current version, that should be enough to conclude the upgrade was rolled back. I don't think there's another situation that leads to this.

This should guarantee we can report rollbacks if this write fails for whatever reason, but we just won't know the reason why.

This may be complicated by the fact that it looks like the upgrade watcher sometimes removes the marker file itself? I'm not sure why or if this is correct because per the code above the presence of the marker is how the agent knows to ack the upgrade. This could be causing some of our "stuck in updating" problems if I am following this correctly.

https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/internal/pkg/agent/cmd/watch.go#L118-L128

internal/pkg/agent/cmd/watch.go
ycombinator: Added in fd29a3297e.

internal/pkg/agent/cmd/watch.go
ycombinator: > If the agent at this point sees that there is an upgrade marker file, and the target version does not match the agent's current version, that should be enough to conclude the upgrade was rolled back.

I think the logic here needs to be revised slightly. 

Older versions of Agent that don't contain upgrade details in the Upgrade Marker will only contain the version that Agent was being upgraded _from_ in the Upgrade Marker (the `prevVersion` field), not the target version of the upgrade.

As such, I think the condition indicating a rollback is if the `prevVersion` value from the Upgrade Marker matches the agent's current version.

---

Also, while I totally agree with this idea of doing a best effort under the said circumstances of recording in upgrade details that we've rolled back, I don't think this `Ack()` method is the right place to do it as it's specific to Fleet and also it's concern is acknowledging the upgrade.  I think we can make this logic the responsibility of the Upgrade Marker Watcher introduced in this PR as that's not a Fleet-specific component.

ycombinator: Implemented in 210af5c9a5.  Please let me know what you think.

internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -0,0 +1,131 @@
...
+	// If we are able to load (read) the upgrade marker here, but there are no
+	// upgrade details in it, it's possible (particularly on Windows, due to
+	// multiple processes reading and writing from the same file), that
+	// the Upgrade Watcher failed to update the upgrade marker file with
+	// upgrade details in it. In such a scenario, if we detect that the
+	// version of Agent we're running right now is the same as the prevVersion
+	// recorded in the upgrade marker we just read here, it would mean the
+	// upgrade was rolled back. So we explicitly create upgrade details in the
+	// upgrade marker with the UPG_ROLLBACK state.
+	if marker.Details == nil && marker.PrevVersion == version.GetAgentPackageVersion() {
```
cmacknz, line 126: Does this handle the case where the watcher failed to write the rolled back state? Isn't it possible for details to not be nil, but be stuck in the watching/monitoring state in this case?

ycombinator: You're right.  This sufficient condition for detecting rollback here should be if the marker is present but `marker.PrevVersion == version.GetAgentPackageVersion()`.    Fixing.

ycombinator: Fixed in f735acd21d.

ycombinator: With added/updated unit tests: 99883dd3dc

cmacknz (approved):

ycombinator:
All integration tests involving Agent installation are failing on Windows. Investigating...

ycombinator:
> All integration tests involving Agent installation are failing on Windows. Investigating...

Turns out all the tests fail trying to install the Agent with the error:

```
error starting service: failed to start service (Elastic Agent): The service did not respond to the start or control request in a timely fashion.
```

I'm able to reproduce this error if I build Agent for Windows from this PR's branch and try to install it.

The only Windows-specific code in this PR is in `marker_access_windows.go`.  For debugging, I temporarily replaced this code with the same code as in `marker_access_other.go` and the install worked.  So it's definitely something about that code that's wonky.  Investigating...

cmacknz:
That error means the Windows service control event handler is either blocked or more likely not being started:

https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/internal/pkg/agent/cmd/run.go#L134

My guess its the problem is in the following function getting blocked or taking too long https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/internal/pkg/agent/cmd/run.go#L111-L113

https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/internal/pkg/agent/cmd/run.go#L574-L582

ycombinator:
I've made two changes to fix the installation problems on Windows (see previous two comments for details):
- On Windows, where we implement a retry mechanism for reading the upgrade marker file, overrode the default timeout of 15 minutes to 10 seconds.
- Short circuit the retry mechanism in case the file is not found.  This is the same logic as before this PR, where we were only reading the upgrade marker once without retries in `upgrade.LoadMarker`.

All integration tests involving Agent installation are now passing on Windows.

Issue #3659, opened 2023-10-25
Flaky test: TestManager_FakeInput_GoodUnitToBad


fearful-symmetry:
## Flaky Test

* **Test Name:**  `TestManager_FakeInput_GoodUnitToBad`
* **Link:** `pkg/component/runtime/manager_test.go`
* **Branch:** main
* **Notes:** large select statement involved, probably a race. So far, haven't been able to reproduce locally.

### Stack Trace

```
[2023-10-25T19:46:28.506Z] === FAIL: pkg/component/runtime TestManager_FakeInput_GoodUnitToBad (2.70s)

[2023-10-25T19:46:28.506Z]     manager_test.go:1371: component state changed: {State:STARTING Message:Starting: spawned pid '2160' Units:map[{UnitType:input UnitID:fake-input}:{State:STARTING Message:Starting: spawned pid '2160' Payload:map[] unitState:0 unitMessage:Starting unitPayload:map[] configStateIdx:0 err:<nil>} {UnitType:input UnitID:good-input}:{State:STARTING Message:Starting: spawned pid '2160' Payload:map[] unitState:0 unitMessage:Starting unitPayload:map[] configStateIdx:0 err:<nil>}] Features:<nil> FeaturesIdx:0 Component:<nil> ComponentIdx:0 VersionInfo:{Name: Version: Meta:map[]} expectedUnits:map[{UnitType:0 UnitID:fake-input}:{state:2 configStateIdx:1 config:0xc000284990 err:<nil> logLevel:4} {UnitType:0 UnitID:good-input}:{state:2 configStateIdx:1 config:0xc000284a20 err:<nil> logLevel:4}] expectedFeatures:<nil> expectedFeaturesIdx:1 expectedComponent:<nil> expectedComponentIdx:1}

[...]

PR #3696, opened 2023-11-02
Fix TestManager_FakeInput_GoodUnitToBad

faec:
Fix the test flakiness reported in https://github.com/elastic/elastic-agent/issues/3659 by rewriting the test to have more straightforward sequencing and be more tolerant of the occasional redundant states that can happen during configuration transitions. (The reported issue happened because a component with two units sometimes reports two state changes as each unit finishes startup, but the test failed when this happened because it assumed startup could only ever produce one state change.)

## Checklist

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/3659


cmacknz (approved):
The logic of the test is much easier to follow now with the expected next state separated out now, thanks!

PR #3700, opened 2023-11-03
[CI] Buildkite migration for Windows 10/11

pazone:
## What does this PR do?
Migrates `Extended windows` pipeline step to buildkite. 
Now it runs build and unit tests on Windows 10 and 11 on a PR basis.
Windows 8 was deprecated and hence removed from the build matrix.

The average Desktop windows job time is 18 minutes. Agent wait time is ~7 min.
Important note: Desktop Windows instances are slow. Agent wait time is up to 7 minutes. 

Aggregates unit tests to a single group for convenience:
 
Screenshot 2023-11-06 at 13 30 39

Windows 10 and 11 are placed in a group `Desktop Windows tests`: 
Screenshot 2023-11-09 at 23 31 38


.buildkite/pipeline.yml
```diff
@@ -4,97 +4,100 @@ env:
...
+        command: ".buildkite/scripts/steps/unit-tests.sh"
+        artifact_paths:
+          - "build/TEST-**"
+          - "build/diagnostics/*"
+          - "coverage.out"
+        agents:
+          provider: "aws"
+          imagePrefix: "core-ubuntu-2204-aarch64"
+          diskSizeGb: 200
+          instanceType: "m6g.4xlarge"
```
pchila, line 35: I know it's the same type of machine as it was before the change but isn't that a bit too much to simply run unit tests? ---> https://aws.amazon.com/ec2/instance-types/m6g/

.buildkite/pipeline.yml
```diff
@@ -4,97 +4,100 @@ env:
...
+        command: ".\\.buildkite\\scripts\\steps\\unit-tests.ps1"
+        artifact_paths:
+          - "build/TEST-**"
+          - "build/diagnostics/*"
+          - "coverage.out"
+        agents:
+          provider: "gcp"
+          image: "family/core-windows-2022"
+          machine_type: "n2-standard-8"
+          disk_size: 200
```
pchila, line 51: By comparison the windows 2022 machine is smaller than the linux ARM one

.buildkite/pipeline.yml
```diff
@@ -168,3 +166,41 @@ steps:
...
+        key: "unit-tests-win10"
+        branches: "main v7.* 7.* 8.* v8.*"
+        command: ".\\.buildkite\\scripts\\steps\\unit-tests.ps1"
+        artifact_paths:
+          - "build/TEST-**"
+          - "build/diagnostics/*"
+          - "coverage.out"
+        agents:
+          provider: "gcp"
+          # TODO create own image
```
pchila, line 183: Is this TODO relevant? Why do we need to create our own image ?

.ci/Jenkinsfile
```diff
@@ -240,67 +240,7 @@ pipeline {
...
       }
-    }
-    stage('extended windows') {
-      when {
-        // Always when running builds on branches/tags
-        // Enable if extended windows support related changes.
-        beforeAgent true
-        anyOf {
-          not { changeRequest() }
-          expression { return isExtendedWindowsEnabled() && env.ONLY_DOCS == "false"}
```
pchila, line 251: In this removed code the `extended windows` stage ran on (conditions in inclusive OR):
- any non-pull request build 
- when isExtendedWindowsEnabled() is true (let's set aside `ONLY_DOCS` for a moment)

here is the code of the function
```groovy
/**
* Wrapper to know if the build should enable the windows extended support
*/
def isExtendedWindowsEnabled() {
  return env.EXT_WINDOWS_CHANGES == "true" || params.extended_windows_ci || env.GITHUB_COMMENT?.contains('extended windows') || matchesPrLabel(label: 'ci:extended-windows')
}
```

It seems we tested:
- `EXT_WINDOWS_CHANGES` env variable
- `extended_windows_ci` build parameter
- `extended windows` in the GH comment
- `ci:extended-windows` PR label

while in the new version all these are not there: I personally never used any of those but I have to wonder if we want to keep such triggers in the buildkite version...

Maybe someone more familiar with our builds can clarify that ?

.buildkite/pipeline.yml
dliappis: Same concern from my side, I would think that unless those tests are very CPU (and memory) intensive a `m6g.xlarge` (4vCPUs, 16GiB RAM) should be enough?

dliappis (approved):
I have limited context on this work, but apart from Paolo's comments I don't have any other substantial comment to make, so, once they get resolved, LGTM from me.

.buildkite/pipeline.yml
pazone: You're right. Set to `m6g.xlarge`. Should be enough

.buildkite/pipeline.yml
pazone: Yes. We plan to create our own VM images to reinstall more tools and reduce the build time

.ci/Jenkinsfile
pazone: To make it triggered by a comment or a label we need to move the extended windows workflow to another pipeline. Unfortunately there's no possibility to trigger a separate step in bk.

pazone: After a thorough discussion with @cmacknz we decided to run extended Windows tests on a PR basis. Later we can extract the extended Windows to a separate pipeline and make it triggered by a GH comment

pazone:
@pchila @blakerouse please take a look at this shiny green pull request.

PR #3701, opened 2023-11-03
Switch to CFT region and add more robust tracking and cleanup of stacks.

### LINKED ISSUE #3459: Switch integration tests to provision stacks in the CFT region in production
We have recently had some stability challenges with the QA GCP region and this is because it isn't granted the same reliability guarantees as the staging or production regions.

Luckily there is a dedicated cloud first testing region for internal use with the same stability guarantees as production in gcp-us-west2 Los Angeles.

Let's switch our integration tests to use that region. We could also decide to use staging cloud, which will give us more cloud providers and regions to test against but with slightly less stability as it will still have pre-production cloud code. Generally the CFT region is the best place for internal testing like we are doing with our integration tests.

### LINKED ISSUE #3463: Integration test framework should save created deployments in state even if provisioning fails
While working on creating a pre-exit hook for cleaning up integration tests resources I came across a [build](https://buildkite.com/elastic/elastic-agent/builds/3457#018abbb6-9cf8-4fef-9fa5-09ac878339b3) where the deployment did not get ready in time.
The deployment was leaked anyways (deployment is still visible in qa admin console)  and clean did not reap it because [the framework doesn't write the deployments in the state in case provisioning fails](https://github.com/elastic/elastic-agent/blob/0ad7b4e00637838821b05bc661c8792eeab8bbb5/pkg/testing/runner/runner.go#L617-L634) (the ess provisioner [does not even return them](https://github.com/elastic/elastic-agent/blob/0ad7b4e00637838821b05bc661c8792eeab8bbb5/pkg/testing/ess/provisioner.go#L65-L117))

If we rely on `mage integration:clean` to clean up resources the state.yml must keep track of all the created deployments, regardless of the usability of those or if the deplyments became healthy in time

blakerouse:
## What does this PR do?

Switches the integration testing framework to use the CFT cloud region for spawning of stacks. Adds more robust tracking of cloud deployments that ensure that they are cleaned up even if they don't become ready.

## Why is it important?

Improve the reliability of the testing framework and less resource usage.

## Checklist

## Related issues

- Closes #3459
- Closes #3463


pierrehilbert:
It seems to be failing on permissions for serverless

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,82 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+	createCtx, createCancel := context.WithTimeout(ctx, 4*time.Minute)
+	defer createCancel()
+
+	client := NewServerlessClient(prov.cfg.Region, "observability", prov.cfg.APIKey, prov.log)
+	srvReq := ServerlessRequest{Name: request.ID, RegionID: prov.cfg.Region}
+
+	prov.log.Logf("Creating serverless stack %s (%s)", request.Version, request.ID)
+	proj, err := client.DeployStack(createCtx, srvReq)
+	if err != nil {
+		return runner.Stack{}, fmt.Errorf("error deploying stack for request %s: %w", request.ID, err)
```
pchila, line 77: I am not sure if it's possible, at least from what I saw of the serverless API but it would be nice if we could tag the serverless projects like we do the ess deployments...

pchila (commented):
Just a few comments, mostly typos and suggestion about injecting timeout values rather than hardcoding slightly different  in various functions.
Happy to see that now we write stacks to the state files as soon as they are created, that should reduce the number of deployments we leave behind.

I am a bit confused about the fact that now we can choose the ESS environment  via `TEST_INTEG_AUTH_ESS_URL` env var: does it mean that we may want to run the integration tests in various environments/regions ?

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
-		}(req, resp))
+// WaitForReady should block until the stack is ready or the context is cancelled.
+func (p *provisioner) WaitForReady(ctx context.Context, stack runner.Stack) (runner.Stack, error) {
+	deploymentID, err := p.getDeploymentID(stack)
+	if err != nil {
+		return stack, fmt.Errorf("failed to get deployment ID from the stack: %w", err)
 	}
-	err := g.Wait()
+	// allow up to 10 minutes for it to become ready
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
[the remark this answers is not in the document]
blakerouse: I made it this way allowing the provisioner code to make the decision on what is the best timeout to be used based on what is being used. That way a timeout from a setting or default value doesn't effect the needed time. I believe its the provisioners that should be determine the appropriate amount of time.

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
-		createCancel()
-		if err != nil {
-			return nil, err
-		}
-		results[r] = resp
-	}
+// Create creates a stack.
+func (p *provisioner) Create(ctx context.Context, request runner.StackRequest) (runner.Stack, error) {
+	// allow up to 2 minutes for request
+	createCtx, createCancel := context.WithTimeout(ctx, 2*time.Minute)
```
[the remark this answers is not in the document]
blakerouse: See above on why I didn't do that.

pkg/testing/ess/serverless_test.go
```diff
@@ -48,27 +47,23 @@ func TestStackProvisioner(t *testing.T) {
 	cfg := ProvisionerConfig{Region: "aws-eu-west-1", APIKey: key}
 	provClient, err := NewServerlessProvisioner(cfg)
 	require.NoError(t, err)
-	stacks := []runner.StackRequest{
-		{ID: "stack-test-one", Version: "8.9.0"},
-		{ID: "stack-test-two", Version: "8.9.0"},
-	}
+	request := runner.StackRequest{ID: "stack-test-one", Version: "8.9.0"}
```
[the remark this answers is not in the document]
blakerouse: The second one was not redundant when the provisioner was used to deploy multiple at a time. Now that the code is changed it only does one at a time, so no need for the second stack. It would not be adding anything to the test.

pkg/testing/ess/serverless_provision.go
blakerouse: I have no idea either.

blakerouse:
> I am a bit confused about the fact that now we can choose the ESS environment via `TEST_INTEG_AUTH_ESS_URL` env var: does it mean that we may want to run the integration tests in various environments/regions ?

I added that to allow a developer the option to switch to staging or qa cloud if they need to perform a test in those environments.

blakerouse:
> It seems to be failing on permissions for serverless

Checking with @pazone to get the credential fixed.

pkg/testing/ess/serverless_provision.go
fearful-symmetry: I don't recall seeing a tagging API when I was first writing for the serverless API, but that might have changed. In general the API is much more restrictive than the stateful API.

fearful-symmetry:
As far as the permission error, it looks like the serverless test is attempting to use the stateful shutdown API. This is probably where the permission error is coming from---we send the details for a serverless deployment, and the stateful API doesn't see it, so it probably just reverts to a permissions error.

.buildkite/hooks/pre-exit
```diff
@@ -10,7 +10,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent" && "$BUILDKITE_STEP_KEY" ==
 
   # Perform cleanup of integration tests resources
   echo "--- Cleaning up integration test resources"
-  TEST_INTEG_AUTH_ESS_REGION=us-east-1 SNAPSHOT=true mage integration:clean
+  SNAPSHOT=true mage integration:clean
```
fearful-symmetry, line 13: I assume the serverless errors are originating here. The `.integration-cache` needs to have some kind of data to tell us if the current deployment is serverless or stateful, and then we can create the appropriate provisioner.

blakerouse: I don't believe so. It is originating here - https://github.com/blakerouse/elastic-agent/blob/35dbbdea9b7bf9693864564e58b9e590e68104f3/pkg/testing/ess/serverless.go#L209

```
>>> (linux-amd64-ubuntu-2204) Failed for instance linux-amd64-ubuntu-2204 (@ 34.27.26.94): failed to check for serverless 8.12.0-SNAPSHOT to be ready: error waiting for ES to become available: error waiting for ES to become healthy: got context done; Last HTTP Error: unexpected status code 401 in request to https://snapshot-cae896.es.us-east-1.aws.elastic.cloud, body: {"error":{"root_cause":[{"type":"security_exception","reason":"unable to authenticate user [] for REST request [/]","header":{"WWW-Authenticate":["Bearer realm=\"security\"","ApiKey","Basic realm=\"security\" charset=\"UTF-8\""]}}],"type":"security_exception","reason":"unable to authenticate user [] for REST request [/]","header":{"WWW-Authenticate":["Bearer realm=\"security\"","ApiKey","Basic realm=\"security\" charset=\"UTF-8\""]}},"status":401}
```

I think I have identified the problem though, when that is called in the code now the username/password is not provided so it cannot actually authenticate.

I was thinking it was an issue with the API key and it is not.

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
pchila: Well it seems to me that we have at least one occurrence where these timeouts are related 
https://github.com/elastic/elastic-agent/pull/3701/files#diff-488eb64eb90cc8cd9d898511bc73a09b146c6704f3ce57d9e88d0189cd007498R693-R695

or is it just coincidence?

The timeout should be how long the caller is willing to wait, so why can't we pass that in ?

In general I don't see the downside of putting the timeouts as parameters (even if at first we just pass constants in it) rather than hardcoding different timeouts in different functions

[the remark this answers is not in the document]
blakerouse: That timeout places a hard limit just for caution. I could remove it, but if the provisioner doesn't place a timeout on its own execution then it would result in it blocking forever. If you prefer I can remove it here.

blakerouse:
Everything passes now, except for the Mac ARM runner not working. That is unrelated to this PR, it is currently affecting all PR's.

blakerouse:
Nothing changed in this PR, latest merge with main has the serverless test failing. Is there a reason that the test used for serverless was switched here https://github.com/elastic/elastic-agent/pull/3465

@belimawr

blakerouse:
Retry of the serverless test now has it passing.

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
+func (p *provisioner) WaitForReady(ctx context.Context, stack runner.Stack) (runner.Stack, error) {
+	deploymentID, err := p.getDeploymentID(stack)
+	if err != nil {
+		return stack, fmt.Errorf("failed to get deployment ID from the stack: %w", err)
 	}
-	err := g.Wait()
+	// allow up to 10 minutes for it to become ready
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer cancel()
+	p.logger.Logf("Waiting for cloud %s to be ready (%s) [id: %s]", stack.Version, stack.ID, deploymentID)
```
AndersonQ, line 102: [Blocker]
It might not be obvious at first what is what if someone does not know before hand what each value is.
```suggestion
	p.logger.Logf("Waiting for cloud stack %s to be ready. StackID: %s, deploymentID: %s]", stack.Version, stack.ID, deploymentID)
```

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
+	ready, err := p.client.DeploymentIsReady(ctx, deploymentID, 30*time.Second)
 	if err != nil {
-		return nil, err
+		return stack, fmt.Errorf("failed to check for cloud %s to be ready: %w", stack.Version, err)
```
AndersonQ, line 105: [Blocker]
Wouldn't be valuable to have the deploymetID as well? In case someone wants to investigate what cause the stack to not be ready.
```suggestion
		return stack, fmt.Errorf("failed to check for cloud stack %s to be ready: %w", stack.Version, err)
```

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
-			Kibana:        resp.KibanaEndpoint,
-			Username:      resp.Username,
-			Password:      resp.Password,
-			Internal: map[string]interface{}{
-				"deployment_id": resp.ID,
-			},
-		})
+	if !ready {
+		return stack, fmt.Errorf("cloud %s never became ready: %w", stack.Version, err)
 	}
```
AndersonQ, line 109: [Blocker]
same here.

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
+	if err != nil {
+		return err
 	}
-	return nil
+
+	// allow up to 1 minute for request
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	p.logger.Logf("Destroying cloud %s (%s) [id: %s]", stack.Version, stack.ID, deploymentID)
```
AndersonQ, line 125: [SUggestion]
Same here

```suggestion
	p.logger.Logf("Destroying cloud stack %s. StackID: %s, deploymentID: %s", stack.Version, stack.ID, deploymentID)
```

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
+	return p.client.ShutdownDeployment(ctx, deploymentID)
 }
 
 func (p *provisioner) createDeployment(ctx context.Context, r runner.StackRequest, tags map[string]string) (*CreateDeploymentResponse, error) {
 	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
 	defer cancel()
 
-	p.logger.Logf("Creating stack %s (%s)", r.Version, r.ID)
+	p.logger.Logf("Creating cloud %s (%s)", r.Version, r.ID)
```
AndersonQ, line 133: May I ask why are you changing from `stack` to `cloud`? I find `stack` much more explicit than `cloud`

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
-			isUp, err := client.DeploymentIsReady(ctx)
-			if err != nil {
-				depErrs <- err
+	client := NewServerlessClient(prov.cfg.Region, "observability", prov.cfg.APIKey, prov.log)
+	srvReq := ServerlessRequest{Name: request.ID, RegionID: prov.cfg.Region}
 
-			}
-			depUp <- isUp
-		}()
+	prov.log.Logf("Creating serverless stack %s (%s)", request.Version, request.ID)
```
AndersonQ, line 74: [Blocker]
```suggestion
	prov.log.Logf("Creating serverless stack %s. RequestID: %s", request.Version, request.ID)
```

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+	proj, err := client.DeployStack(createCtx, srvReq)
+	if err != nil {
+		return runner.Stack{}, fmt.Errorf("error deploying stack for request %s: %w", request.ID, err)
+	}
+	err = client.WaitForEndpoints(createCtx)
+	if err != nil {
+		return runner.Stack{}, fmt.Errorf("error waiting for endpoints to become available for request: %w", err)
```
AndersonQ, line 81: Did you miss the request id?
```suggestion
		return runner.Stack{}, fmt.Errorf("error waiting for endpoints to become available for request %s: %w", request.ID, err)
```

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+		Kibana:        client.proj.Endpoints.Kibana,
+		Username:      client.proj.Credentials.Username,
+		Password:      client.proj.Credentials.Password,
+		Internal: map[string]interface{}{
+			"deployment_id":   proj.ID,
+			"deployment_type": proj.Type,
+		},
+		Ready: false,
 	}
+	prov.log.Logf("Created serverless stack %s (%s) [id: %s]", request.Version, request.ID, proj.ID)
```
AndersonQ, line 96: [Blocker]
```suggestion
	prov.log.Logf("Created serverless stack %s. RequestID: %s, projectID: %s", request.Version, request.ID, proj.ID)
```

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+	return stack, nil
+}
+
+// WaitForReady should block until the stack is ready or the context is cancelled.
+func (prov *ServerlessProvision) WaitForReady(ctx context.Context, stack runner.Stack) (runner.Stack, error) {
+	deploymentID, deploymentType, err := prov.getDeploymentInfo(stack)
```
AndersonQ, line 102: [Suggestion]
For serverless it's a project id, right? So:
```suggestion
	projectID, projectType, err := prov.getProjectInfo(stack)
```

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+	client := NewServerlessClient(prov.cfg.Region, "observability", prov.cfg.APIKey, prov.log)
+	client.proj.ID = deploymentID
+	client.proj.Type = deploymentType
+	client.proj.Region = prov.cfg.Region
+	client.proj.Endpoints.Elasticsearch = stack.Elasticsearch
+	client.proj.Endpoints.Kibana = stack.Kibana
+	client.proj.Credentials.Username = stack.Username
+	client.proj.Credentials.Password = stack.Password
+
+	prov.log.Logf("Waiting for serverless stack %s to be ready (%s) [id: %s]", stack.Version, stack.ID, deploymentID)
```
AndersonQ, line 119:
```suggestion
	prov.log.Logf("Waiting for serverless stack %s to be ready. StackID: %s, projectID: %s", stack.Version, stack.ID, deploymentID)
```

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
-			err := client.DeleteDeployment()
-			if err != nil {
-				return fmt.Errorf("error removing deployment after re-creating client: %w", err)
-			}
+// Delete deletes a stack.
+func (prov *ServerlessProvision) Delete(ctx context.Context, stack runner.Stack) error {
+	deploymentID, deploymentType, err := prov.getDeploymentInfo(stack)
+	if err != nil {
+		return fmt.Errorf("failed to get deployment info from the stack: %w", err)
+	}
```
AndersonQ, line 162: [Suggestion]
```suggestion
	projectID, projectType, err := prov.getprojectInfo(stack)
	if err != nil {
		return fmt.Errorf("failed to get project info from the stack: %w", err)
	}
```

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+	client := NewServerlessClient(prov.cfg.Region, "observability", prov.cfg.APIKey, prov.log)
+	client.proj.ID = deploymentID
+	client.proj.Type = deploymentType
+	client.proj.Region = prov.cfg.Region
+	client.proj.Endpoints.Elasticsearch = stack.Elasticsearch
+	client.proj.Endpoints.Kibana = stack.Kibana
+	client.proj.Credentials.Username = stack.Username
+	client.proj.Credentials.Password = stack.Password
+
+	prov.log.Logf("Destroying serverless stack %s (%s) [id: %s]", stack.Version, stack.ID, deploymentID)
```
AndersonQ, line 173: [Blocker]
```suggestion
	prov.log.Logf("Destroying serverless stack %s. StackID: %s, projectID: %s", stack.Version, stack.ID, deploymentID)
```

pkg/testing/runner/runner.go
```diff
@@ -592,59 +591,115 @@ func (r *Runner) startStacks(ctx context.Context) error {
...
+					ch <- res
+				}
+			}
+		}
 	}
 
-	// start go routine to provision the needed stacks
-	go func(ctx context.Context) {
-		defer r.stacksReady.Done()
+	// start go routine's to provision the needed stacks
```
AndersonQ, line 634: typo?
```suggestion
	// start goroutines to provision the needed stacks
```

pkg/testing/runner/runner_test.go
```diff
@@ -123,31 +124,38 @@ func (f *fakeInstanceProvisioner) Clean(_ context.Context, _ Config, instances [
...
+	stack.Ready = true
+	return stack, nil
 }
 
-func (f *fakeStackProvisioner) Clean(_ context.Context, stacks []Stack) error {
-	f.stacks = stacks
+func (f *fakeStackProvisioner) Delete(_ context.Context, stack Stack) error {
+	f.mx.Lock()
+	defer f.mx.Unlock()
+	f.stacks = append(f.stacks, stack)
```
AndersonQ, line 159: I'm confused here. Why `Delete` is saving a stack? if `f.stacks` is an slide of deleted stacks, then could you rename it?

AndersonQ (changes requested):

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/ess/serverless_provision.go
belimawr: +1

pkg/testing/runner/runner.go
belimawr: +1

pkg/testing/runner/runner_test.go
belimawr: +1

belimawr (changes requested):

pkg/testing/ess/provisioner.go
```diff
@@ -62,89 +60,77 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
+	resp, err := p.createDeployment(createCtx, request,
+		map[string]string{
+			"division":          "engineering",
+			"org":               "ingest",
+			"team":              "elastic-agent",
+			"project":           "elastic-agent",
+			"integration-tests": "true",
+		})
+	if err != nil {
+		return runner.Stack{}, err
```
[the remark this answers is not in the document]
blakerouse: Already includes all the context that is required as the call to `createDeployment` adds all that information. Wrapping just for the sake of wrapping does not help.

pkg/testing/ess/serverless_provision.go
blakerouse: Why would I change to project? We call the other one deployments, its stored as `deployment_id` and `deployment_type` in the internal context. So confused on this suggestion.

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
blakerouse: I like how it is.

pkg/testing/ess/serverless_provision.go
blakerouse: Updated.

pkg/testing/ess/provisioner.go
blakerouse: Updated.

pkg/testing/ess/provisioner.go
blakerouse: Updated.

pkg/testing/ess/provisioner.go
blakerouse: Updated.

pkg/testing/ess/provisioner.go
blakerouse: Updated.

pkg/testing/ess/provisioner.go
blakerouse: `r` is request that is clear in the function definition. `r.ID` is the request ID, and is used to reference a created stack, the created stack struct must return the same ID.

pkg/testing/ess/serverless_provision.go
blakerouse: Updated.

pkg/testing/ess/serverless_provision.go
blakerouse: Updated.

pkg/testing/ess/serverless_provision.go
blakerouse: Updated.

pkg/testing/ess/serverless_provision.go
blakerouse: Updated.

pkg/testing/runner/runner.go
blakerouse: Updated.

.buildkite/hooks/pre-command
```diff
@@ -17,8 +17,7 @@ DOCKER_REGISTRY="docker.elastic.co"
 DOCKER_REGISTRY_SECRET_PATH="kv/ci-shared/platform-ingest/docker_registry_prod"
 CI_DRA_ROLE_PATH="kv/ci-shared/release/dra-role"
 CI_GCP_OBS_PATH="kv/ci-shared/observability-ingest/cloud/gcp"
-# CI_AGENT_QA_OBS_PATH="kv/ci-shared/observability-ingest/elastic-agent-ess-qa"
-CI_ESS_STAGING_PATH="kv/ci-shared/platform-ingest/platform-ingest-ec-staging"
+CI_ESS_STAGING_PATH="kv/ci-shared/platform-ingest/platform-ingest-ec-prod"
```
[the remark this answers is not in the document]
blakerouse: Updated.

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
cmacknz: The caller has the flexibility to set a shorter timeout through the context. If the parent context's timeout is shorter than the child context's the child context expires when the parents does.

So as is you can change the timeout with the context, and both the maximum and default timeouts are 10 minutes. 

https://go.dev/play/p/sQQpd4QWt42

```go
package main

import (
	"context"
	"fmt"
	"time"
)

func main() {
	ctx1, cancel1 := context.WithTimeout(context.Background(), 1*time.Millisecond)
	defer cancel1()
	ctx5, cancel5 := context.WithTimeout(ctx1, 5*time.Second)
	defer cancel5()

	select {
	case <-ctx1.Done():
		fmt.Println("Waited 1 millisecond")
	case <-ctx5.Done():
		fmt.Println("Waited 5 seconds")
	}
}
```

This prints `Waited 1 millisecond`.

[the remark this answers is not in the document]
cmacknz: > So as is you can change the timeout with the context, and both the maximum and default timeouts are 10 minutes.

To be clear I think this is a totally reasonable way for this to work, unless we think the timeout should be able to be longer than 10 minutes.

Passing in a timeout or deadline via the context is a fine way to do this.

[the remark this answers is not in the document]
cmacknz: Updated comment with a correct example:

https://go.dev/play/p/KBEtIjKFho_b

```go
package main

import (
	"context"
	"fmt"
	"time"
)

func main() {
	ctx1, cancel1 := context.WithTimeout(context.Background(), 1*time.Millisecond)
	defer cancel1()
	ctx5, cancel5 := context.WithTimeout(ctx1, 5*time.Second)
	defer cancel5()

	start := time.Now()
	select {
	case <-ctx5.Done():
		fmt.Printf("Child context is expired after %s\n", time.Since(start))
	}
}
```

This prints `Child context is expired after 1ms` since the child context respects the parent context's deadline.

So with the code today the maximum and default timeout is 10 minutes, and callers can set a shorter timeout via the context. I think this is a perfectly reasonable way for this to work.

The only debate should be if 10 minutes is a long enough maximum timeout.

pkg/testing/ess/provisioner.go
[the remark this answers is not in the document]
cmacknz: I think he meant the comment chain in https://github.com/elastic/elastic-agent/pull/3701/files#r1383088811 which might have been above at one point but isn't necessarily now.

pkg/testing/ess/serverless_provision.go
```diff
@@ -71,114 +62,118 @@ func (prov *ServerlessProvision) SetLogger(l runner.Logger) {
...
+				lastErr = ctx.Err()
 			}
-			if gotUp >= len(requests) {
-				return stacks, nil
+			return stack, fmt.Errorf("serverless stack %s [stack_id: %s, deployment_id: %s] never became ready: %w", stack.Version, stack.ID, deploymentID, lastErr)
+		case <-ticker.C:
+			go func() {
+				statusCtx, statusCancel := context.WithTimeout(ctx, 30*time.Second)
+				defer statusCancel()
+				ready, err := client.DeploymentIsReady(statusCtx)
```
cmacknz, line 138: It seems like the stateful `DeploymentIsReady` method does this waiting work, and `WaitForReady` just wraps it, where as this `DeploymentIsReady` just gives you the current status.

If you call `WaitForReady` this gives you the same result for both, but calling `DeploymentIsReady` doesn't. This might bite someone in the future since it isn't consistent.

pkg/testing/ess/serverless_provision.go
cmacknz: Anderson is right, the "official" name is a Serverless project, separate from a Stateful deployment. I don't have a good idea for a term that accurately describes both besides something like "instance".

Screenshot 2023-11-09 at 3 23 58 PM

cmacknz (commented):
A couple nits but looks good generally.

Since the only explicit call to `Delete()` for a stack is in the `Clean()` method, if I am following correctly the fix here is that when the integration test script calls `mage integration:clean` it now has the deployment IDs of every stack including those that weren't ready in time?

pkg/testing/ess/serverless_provision.go
blakerouse: I agree it is not consistent. This PR is about changing to use CFT not about making cloud and serverless consistency between the implementations, which I agree should be fixed (but not in this PR). This PR is about switching to the CFT region and making it keep state when a stack doesn't become ready. This has now snowballed into a review of serverless implementation, which would have been better to bring up when it was originally added.

pkg/testing/ess/serverless_provision.go
blakerouse: Same as my other comment. I also don't see why "deployment" cannot be used...

pkg/testing/ess/serverless_provision.go
cmacknz: It wasn't clear that it was already like this before, thanks.

pkg/testing/ess/serverless_provision.go
cmacknz: > Why would I change to project? We call the other one deployments, its stored as deployment_id and deployment_type in the internal context. So confused on this suggestion.

Project is technically the correct term, so the confusion from the above comment is removed. This rename is a completely valid suggestion.

The common term for both could be deployment in contexts that are not serverless specific and probably not cause too much confusion. I don't think we need to block the entire PR on this suggestion, but project is definitely the right word here.

pkg/testing/ess/serverless_provision.go
AndersonQ: are you sure you updated the right line? it's still the same, but it isn't a blocker so it's fine

pkg/testing/ess/serverless_provision.go
AndersonQ: As I said in my 1st comment, it's a suggestion, I prefer to have it consistent with how we talk and what we see in the UI. But it isn't a blocker either

AndersonQ (approved):

pierrehilbert:
`unable to create policy: Unauthorized`

belimawr (approved):

PR #3734, opened 2023-11-09
Fix runner/provisioners to account for the provision r when comparing instances and stacks

AndersonQ:
## What does this PR do?

It adds the provisioner name to Stack and Instance so the test framework can differentiate stack and instances from different providers

## Why is it important?

The runner was not taking into account the provisioner when checking is there was already an instance or stack to be reused

## Checklist

see https://github.com/elastic/elastic-agent/issues/3756

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3756
- Closes https://github.com/elastic/elastic-agent/issues/3770

## Questions to ask yourself

- ...

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -1772,10 +1772,14 @@ func createTestRunner(matrix bool, singleTest string, goTestFlags string, batche
 	fmt.Printf(">>>> Using %s instance provisioner\n", instanceProvisionerMode)
 	stackProvisionerMode := os.Getenv("STACK_PROVISIONER")
 	if stackProvisionerMode == "" {
-		stackProvisionerMode = "stateful"
+		stackProvisionerMode = ess.ProvisionerStateful
 	}
-	if stackProvisionerMode != "stateful" && stackProvisionerMode != "serverless" {
-		return nil, fmt.Errorf("STACK_PROVISIONER environment variable must be one of 'serverless' or 'stateful', not %s", stackProvisionerMode)
+	if stackProvisionerMode != ess.ProvisionerStateful &&
+		stackProvisionerMode != ess.ProvisionerServerless {
+		return nil, fmt.Errorf("STACK_PROVISIONER environment variable must be one of %q or %q, not %s",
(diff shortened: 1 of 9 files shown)


pchila (approved):
LGTM

fearful-symmetry:
So, I'm not sure if this is a problem this PR is intended to fix, but it feels adjacent to the problem. If you spin up a stack with something like `STACK_PROVISIONER=serverless TEST_PLATFORMS="linux/amd64" INSTANCE_PROVISIONER=multipass mage integration:single` and you run `mage integration:clean` with no additional env vars after, we'll still fail due to the provisioner trying to spin up the wrong kind of stack for the shutdown command, even though `provisioner: serverless` appears in the `state.yml` file. Presumably during initialization we're just not using the `provisioner` tag from the state file.

Might be outside the scope of this, but it makes me wonder if there's other cases where we're not properly using the `provisioner` tag.

AndersonQ:
> So, I'm not sure if this is a problem this PR is intended to fix, but it feels adjacent to the problem. If you spin up a stack with something like `STACK_PROVISIONER=serverless TEST_PLATFORMS="linux/amd64" INSTANCE_PROVISIONER=multipass mage integration:single` and you run `mage integration:clean` with no additional env vars after, we'll still fail due to the provisioner trying to spin up the wrong kind of stack for the shutdown command, even though `provisioner: serverless` appears in the `state.yml` file. Presumably during initialization we're just not using the `provisioner` tag from the state file.
> 
> Might be outside the scope of this, but it makes me wonder if there's other cases where we're not properly using the `provisioner` tag.

good point. but not really, the PR was to only address the reuse of the saved state to avoid the frameworks using the wrong instances or stacks.

In reality, as you said, I think the framework isn't quite ready to deal with different providers in a completely automated way. IIRC the idea is we manually clean up one stack before changing providers, which is error prone :/

fearful-symmetry (approved):

PR #3735, opened 2023-11-09
[CI] macos arm tests trigger by comment

pazone:
## What does this PR do?

Makes MacOS 13 ARM tests run only on demand by a comment `buildkite test macosarm`

## Why is it important?
We lack MacOS ARM machines and testing on ARM for each pull request exhausts the cluster


pazone:
buildkite test macosarm

Issue #3738, opened 2023-11-10
runtime.update can block all checkins and state processing while waiting for shutdown


leehinman:
[update](https://github.com/elastic/elastic-agent/blob/5f891b0664cd453ed40a48321f4c97f80b62a784/pkg/component/runtime/manager.go#L663) can block because it waits for [waitForStopped](https://github.com/elastic/elastic-agent/blob/5f891b0664cd453ed40a48321f4c97f80b62a784/pkg/component/runtime/manager.go#L740) to finish.  `waitForStopped` can take up to [15 seconds](https://github.com/elastic/elastic-agent/blob/5f891b0664cd453ed40a48321f4c97f80b62a784/pkg/component/runtime/conn_info_server.go#L19). Because this function is called by the Coordinator, the entire Coordinator can block for that time.

This block causes problems because the runtime Manager update channel needs to write to [coordinator.watchRuntimeComponets](https://github.com/elastic/elastic-agent/blob/5f891b0664cd453ed40a48321f4c97f80b62a784/internal/pkg/agent/application/coordinator/coordinator.go#L606). If it cannot do that it will block.  This in turn blocks [runtime.stateChange](https://github.com/elastic/elastic-agent/blob/5f891b0664cd453ed40a48321f4c97f80b62a784/pkg/component/runtime/manager.go#L803) which prevents [runtime.runLoop](https://github.com/elastic/elastic-agent/blob/5f891b0664cd453ed40a48321f4c97f80b62a784/pkg/component/runtime/runtime.go#L150) from processing checkins and state changes, leading to stalled communication and "missed checkin" errors for components that are still sending checkins.

This can lead to the bugs seen in #3617 and #3654

Issue #3721, opened 2023-11-07
Agent with Elastic Defend remains unhealthy for long when invalid configuration is added for sometime and then updated with correct.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.11.0 BC9
Build : 68160
Commit : f2ea0c43ec0d854259d63d926b97e5c556b5f6b2
```

**Host OS:**  Windows

**Preconditions:**
1. 8.11.0 BC9 Kibana cloud environment should be available.
2. Windows agent should be installed using System and Defend integration.
3. Create invalid logstash outputs or invalid fleet server.
_[Update]_
4. Agent tamper protection settings are enabled.

**Steps to reproduce:**
1. Set invalid output/invalid fleet server under agent policy.
2. Observe agent gets unhealthy and error under Defend integration.
3. Wait 5-10 minutes and update the changes to the correct configuration.
6. Wait for 10-15 minutes and observe agent still remains unhealthy.

**NOTE:**
- Issue is consistently reproducible.

**Screenshot:**

![image](https://github.com/elastic/elastic-agent/assets/77374876/f751b6f1-00d2-43ee-95e9-c2dd1a90a0a6)

**Expected Result:**
Agent with Endpoint remains should get back Healthy when invalid configuration is added for sometime and then updated with correct.

**Agent.json:**
[ec2amaz-8ag00sp-agent-details.zip](https://github.com/elastic/elastic-agent/files/13279350/ec2amaz-8ag00sp-agent-details.zip)

**Logs:**
[elastic-agent-diagnostics-2023-11-07T11-28-06Z-00.zip](https://github.com/elastic/elastic-agent/files/13279351/elastic-agent-diagnostics-2023-11-07T11-28-06Z-00.zip)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

cmacknz:
The degraded input I see is:

```
    {
      "id": "beat/metrics-monitoring",
      "type": "beat/metrics",
      "status": "DEGRADED",
      "message": "Degraded: pid '4652' missed 1 check-in",
      "units": [
        {
          "id": "beat/metrics-monitoring-metrics-monitoring-beats",
          "type": "input",
          "status": "STARTING",
          "message": "Starting: spawned pid '4652'"
        },
        {
          "id": "beat/metrics-monitoring",
          "type": "output",
[...]

Issue #3654, opened 2023-10-25
Linux agents gets unhealthy on enabling/disabling modules for System/Linux integration.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.11.0 BC3
BUILD: 67923
COMMIT: 714189fa2b0f0a4d9f3865a8fce08261211570c8
```

**Host OS:**  Linux

**Preconditions:**
1. 8.11.0 BC3 Kibana cloud environment should be available.
2. Few Linux agents- .tar and .deb agents should be installed.

**Steps to reproduce:**
1. Add System and linux integrations to the agent policy.
2. Navigate to integration settings and disable the integration modules.
3. Wait for policy to get updated.
4. Observe few linux agents gets unhealthy inconsistently.
5. Restart the disabled modules and wait for policy to get updated.
6. Observe few linux agents again gets unhealthy inconsistently.

**NOTE:**
- We are consistently able to reproduce this issue for linux agents.
- Issue is reproducible for enabling/disabling system and linux integrations.

**Screen Recording:**

https://github.com/elastic/elastic-agent/assets/77374876/93041db3-488e-4071-b46d-a9475bbe75f7

**Expected Result:**
Linux agents should remain Healthy on enabling/disabling modules for System/Linux integration.

**Agent.json:**
[ip-172-31-66-47-agent-details.zip](https://github.com/elastic/elastic-agent/files/13162600/ip-172-31-66-47-agent-details.zip)
[ip-172-31-79-43-agent-details.zip](https://github.com/elastic/elastic-agent/files/13162601/ip-172-31-79-43-agent-details.zip)

**Logs:**
[elastic-agent-diagnostics-2023-10-25T07-20-34Z-00.zip](https://github.com/elastic/elastic-agent/files/13162607/elastic-agent-diagnostics-2023-10-25T07-20-34Z-00.zip)
[elastic-agent-diagnostics-2023-10-25T07-27-25Z-00.zip](https://github.com/elastic/elastic-agent/files/13162608/elastic-agent-diagnostics-2023-10-25T07-27-25Z-00.zip)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

cmacknz:
I can reproduce this on my Mac by adding and removing the system integration.

cmacknz:
[...]

Issue #3617, opened 2023-10-17
Agent gets unhealthy on assigning from policy with Elastic Defend integration to without Defend integration.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.11.0 BC2
BUILD: 67841
COMMIT: 636a8339cfad92998a5a5adb7be81e3546525ebf
```

**Host OS:**  All

**Preconditions:**
1. 8.11.0 BC2 Kibana cloud environment should be available.
2. Windows agent should be installed using System and Elastic Defend integration.

**Steps to reproduce:**
1. Navigate to Fleet>Agents tab.
2. Assign the agent to the policy without Elastic Defend integration.
3. Observe agent gets unhealthy.

**Screen Recording:**

https://github.com/elastic/elastic-agent/assets/77374876/3442901e-4afa-4a35-8199-e6193c52912e

https://github.com/elastic/elastic-agent/assets/77374876/ca9125ed-6ec8-45cf-b4a0-2bec588a234e

**Agent Logs:**
[elastic-agent-diagnostics-2023-10-17T09-17-18Z-00.zip](https://github.com/elastic/elastic-agent/files/12926784/elastic-agent-diagnostics-2023-10-17T09-17-18Z-00.zip)

**Agent JSON:**
[ec2amaz-auvpg7n-agent-details.zip](https://github.com/elastic/elastic-agent/files/12926790/ec2amaz-auvpg7n-agent-details.zip)

**Expected Result:**
Agent shouldn't get unhealthy on assigning from policy with Elastic Defend integration to without Defend integration.

**Impacted Testcase:**
https://elastic.testrail.io/index.php?/tests/view/2166951

**Similar issue:**
https://github.com/elastic/elastic-agent/issues/3507

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

cmacknz:
```json
"components": [
    {
      "id": "log-default",
      "type": "log",
      "status": "HEALTHY",
      "message": "Healthy: communicating with pid '128'",
      "units": [
        {
          "id": "log-default",
          "type": "output",
          "status": "HEALTHY",
          "message": "Healthy"
        },
        {
          "id": "log-default-logfile-system-0a5c1691-2ce5-4a4f-8ead-f5b1e767f61a",
          "type": "input",
          "status": "HEALTHY",
          "message": "Healthy"
        }
[...]

PR #3747, opened 2023-11-10
Rework runtime manager updates to block the coordinator less

faec:
## What does this PR do?

Fix https://github.com/elastic/elastic-agent/issues/3738, which happens when `Coordinator` and `runtime.Manager` block on a policy update and state reporting gets backed up.

Summary of changes:
- Give the runtime manager an explicit run loop in its own function instead of managing all changes via mutex
- `Manager.Update` now sends the component model update to the runtime manager loop via a channel
- Component model updates are triggered asynchronously from the runtime manager loop, so neither `Coordinator` nor the runtime manager block their communication while an update is in progress
- Results from the update process are sent back to `Coordinator` on the runtime manager's (previously unused) error channel instead of returning directly from `Update`
  * on the `Coordinator` side, the internal variables for the runtime manager error and the update error have been merged
- The runtime manager now uses a done channel to signal that it is shutting down and will no longer accept incoming requests
- (adjacent cleanup) some unneeded / no longer needed synchronization handling has been removed from the runtime manager (`updateMx` which is no longer needed since `(*Manager).runLoop` ensures `update` calls do not conflict, `netMx` which was only used to get the tcp listener port, and some startup / shutdown atomics)

## Checklist

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/3738
- Fixes https://github.com/elastic/elastic-agent/issues/3654
- Fixes https://github.com/elastic/elastic-agent/issues/3617
- Fixes https://github.com/elastic/ingest-dev/issues/2560
- Fixes https://github.com/elastic/elastic-agent/issues/3721
- Fixes https://github.com/elastic/elastic-agent/issues/3507


pkg/component/runtime/manager.go
```diff
@@ -93,22 +93,18 @@ type Manager struct {
...
-	netMx    sync.RWMutex
-	listener net.Listener
-	server   *grpc.Server
-
 	// Set when the RPC server is ready to receive requests, for use by tests.
 	serverReady *atomic.Bool
 
-	// updateMx protects the call to update to ensure that
-	// only one call to update occurs at a time
-	updateMx sync.Mutex
```
pchila, line 111: :heart:

pkg/component/runtime/manager.go
```diff
@@ -93,22 +93,18 @@ type Manager struct {
...
 	ca         *authority.CertificateAuthority
 	listenAddr string
+	listenPort int
 	agentInfo  *info.AgentInfo
 	tracer     *apm.Tracer
 	monitor    MonitoringManager
 	grpcConfig *configuration.GRPCConfig
 
-	// netMx synchronizes the access to listener and server only
-	netMx    sync.RWMutex
```
pchila, line 102: :heart:

pkg/component/runtime/manager_test.go
```diff
@@ -1564,7 +1588,8 @@ LOOP:
 		case <-endTimer.C:
 			// no deadlock after timeout (all good stop the component)
 			updatedCancel()
-			_ = m.Update(component.Model{Components: []component.Component{}})
+			m.Update(component.Model{Components: []component.Component{}})
+			<-errCh // Don't care about the result of Update, just that it runs
```
pchila, line 1592: This is where we have this test hanging in CI... 
Shouldn't this be 
```suggestion
			m.Update(component.Model{Components: []component.Component{}})
			<-m.errCh // Don't care about the result of Update, just that it runs
```
since `errCh` is a channel used internally in the test ?

pchila (commented):
Code looks good, I think that the suggestion I left should take carel of the deadlock we have in unit test ;)

blakerouse (approved):
Looks really good.

I like the removal of the locks and the fact that the error now comes through the errCh. Unit testing has good coverage with the change as well.

faec:
There's still an issue with this PR -- in manual testing it gets stuck forever during the `install` command, enrollment reaches Kibana but is never properly reported Agent-side. That might also be why the integration tests seem stalled without a result on CI. Investigating.

faec:
Found the problem: Coordinator blocks to tell the runtime manager about the new update, and the runtime manager blocks to tell Coordinator about the result of the update. Usually these two aren't in conflict, but during an install two Update calls happen right after each other, so both objects end up waiting on each other.

I still ran the update logic synchronously on the runtime manager goroutine with the idea that it was a "less drastic change" that could be done in a smaller first-pass PR, but I think I just need to fix the blocking entirely.

faec:
New version is ready that fixes the full issue instead of just mitigating it. I had to add some more state handling variables and code but I've tried to comment thoroughly, and this should prevent any number of policy changes from ever blocking state update reporting.

pkg/component/runtime/manager.go
```diff
@@ -205,87 +216,123 @@ func (m *Manager) Run(ctx context.Context) error {
...
-			State:     crs.getLatest(),
-			LegacyPID: legacyPID,
-		})
+// If calling from a test, you should read from errCh afterwards to avoid
+// blocking Manager's main loop.
+func (m *Manager) Update(model component.Model) {
+	select {
+	case m.updateChan <- model:
+	case <-m.doneChan:
+		// Manager is shutting down, ignore the udpate
```
ycombinator, line 334: Typo.
```suggestion
		// Manager is shutting down, ignore the update
```

pkg/component/runtime/manager.go
```diff
@@ -205,87 +216,123 @@ func (m *Manager) Run(ctx context.Context) error {
...
+			// There is a component model update available, apply it.
+			go func(model component.Model) {
+				// Run the update with tearDown set to true since this is coming
+				// from a user-initiated policy update
+				result := m.update(model, true)
+
+				// When update is done, send its result back to the coordinator,
+				// unless we're shutting down.
+				select {
+				case m.errCh <- result:
```
ycombinator, line 278: Nit: the result here is an error.
```suggestion
				err := m.update(model, true)

				// When update is done, send its result back to the coordinator,
				// unless we're shutting down.
				select {
				case m.errCh <- err:
```

ycombinator (approved):
LGTM.  Very easy to follow the new implementation and thanks for the adjacent cleanup as well.

pchila (approved):
LGTM

PR #3743, opened 2023-11-10
Revert "Support flattened data_stream.* fields"

belimawr:
Reverts elastic/elastic-agent#3465 because it seems to be breaking some integration tests: https://github.com/elastic/elastic-agent/issues/3741

Issue #3741, opened 2023-11-10
TestLogIngestionFleetManaged/Monitoring_logs_are_shipped is flaky


cmacknz:
## Flaky Test

* **Test Name:** TestLogIngestionFleetManaged/Monitoring_logs_are_shipped
* **Link:** https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/testing/integration/logs_ingestion_test.go#L39
* **Branch:** main
* **Artifact Link:**
  * https://buildkite.com/elastic/elastic-agent/builds/4753 (note multiple failures here but one of the is this test)
  * https://buildkite.com/elastic/elastic-agent/builds/4763
* **Notes:** This seems to mostly fail against serverless, which has a longer delay from ingestion to data being queryable (there is a longer refresh interval). Possibly the timeout needs to be longer.

### Stack Trace

```
=== RUN   TestLogIngestionFleetManaged/Monitoring_logs_are_shipped
    logs_ingestion_test.go:106: Making sure metricbeat logs are populated
    logs_ingestion_test.go:110: metricbeat: Got 0 documents
    logs_ingestion_test.go:111: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/logs_ingestion_test.go:111
        	            				/home/ubuntu/agent/testing/integration/logs_ingestion_test.go:90
        	Error:      	Should not be zero, but was 0
        	Test:       	TestLogIngestionFleetManaged/Monitoring_logs_are_shipped
--- FAIL: TestLogIngestionFleetManaged/Monitoring_logs_are_shipped (15.19s)
```

cmacknz:
Believe this new flakiness was introduced in https://github.com/elastic/elastic-agent/pull/3465

cmacknz:
Reverting the change while we investigate what the failure is https://github.com/elastic/elastic-agent/pull/3743

cmacknz:
Looking at the logs it looks like the `filestream-monitoring` instance never connected to Elasticsearch, or legitimately didn't publish events. There are no acked events or output stats and the number of published events equals the number of events in the pipeline.

```json
{
  "log.level": "info",
  "@timestamp": "2024-03-27T14:50:08.758Z",
  "message": "Non-zero metrics in the last 30s",
  "component": {
    "binary": "filebeat",
    "dataset": "elastic_agent.filebeat",
    "id": "filestream-monitoring",
    "type": "filestream"
  },
  "log": {
[...]

Issue #3756, opened 2023-11-13
[Integration Test Framework] does not keep track of the stack type (statefull/serverless)


belimawr:
When the framework creates a stack deployment it does not save it's type (statefull/serverless), so when re-using existing stacks, it will use whatever stack from the required version exists, regardless of the stack provisioned selected.

This makes statefull tests run against serverless and vice versa depending on which one was run first.

**How to reproduce:**
1. Make sure there is not existing stack, run: `mage integration:clean`
2. Run a test against serverless: `STACK_PROVISIONER=serverless TEST_PLATFORMS="linux/amd64" SNAPSHOT=true mage -v integration:single TestLogIngestionFleetManaged`
3. Everything works, a serverless project is created.
4. Run the same test against statefull: `TEST_PLATFORMS="linux/amd64" SNAPSHOT=true mage -v integration:single TestLogIngestionFleetManaged`
5. The serverless project is used, see the logs:
```
>>>> Using ogc instance provisioner
>>>> Using stateful stack provisioner
>>> Creating zip archive of repo to send to remote hosts
>>> Reusing stack 8.12.0-SNAPSHOT (8120-SNAPSHOT)
```

AndersonQ:
it's seems to be the same I'm fixing here https://github.com/elastic/elastic-agent/pull/3734

AndersonQ:
actually, I was looking better at it, this issue isn't the same as the one I'm fixing on https://github.com/elastic/elastic-agent/pull/3734. However they're similar, so I think they can be the same PR

Issue #3664, opened 2023-10-26
Enrolling into Fleet can fail because the control socket doesn't exist


cmacknz:
- Relates https://github.com/elastic/elastic-agent/issues/3628

This is frequently affecting our integration tests because we are frequently enrolling as part of the agent installation. A typical failure looks like:

```
fixture.go:365: >> running agent with: [/tmp/TestInstallWithEndpointSecurityAndRemoveEndpointIntegrationunprotected1325593366/001/elastic-agent-8.11.0-SNAPSHOT-linux-arm64/elastic-agent install --force --non-interactive --url https://xxxxxx.fleet.us-central1.gcp.foundit.no:443 --enrollment-token <REDACTED>]
    endpoint_security_test.go:358: >>> Ran Enroll. Output: Installing in non-interactive mode.
        Copying files................................................................. DONE
        Installing service...... DONE
        Starting service... DONE
        Enrolling Elastic Agent with Fleet...{"log.level":"info","@timestamp":"2023-10-26T17:31:05.010Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":479},"message":"Starting enrollment to URL: https://da362739b9424443a6e94773953da8d7.fleet.us-central1.gcp.foundit.no:443/","ecs.version":"1.6.0"}
        ......{"log.level":"info","@timestamp":"2023-10-26T17:31:06.610Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":275},"message":"Elastic Agent might not be running; unable to trigger restart","error":{"message":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /opt/Elastic/Agent/data/tmp/elastic-agent-control.sock: connect: no such file or directory\""},"ecs.version":"1.6.0"}
        Successfully enrolled the Elastic Agent.
```

1. The error tells us exactly what is wrong, the `elastic-agent-control.sock` file doesn't exist yet. Possibly we just need to retry accessing it, but there could have been an unlogged problem creating it.
2. The command reports `Successfully enrolled the Elastic Agent.` which I don't think is expected. The agent is enrolled, but it won't connect to Fleet until it is manually restarted.

cmacknz:
A solution to this was started in https://github.com/elastic/elastic-agent/pull/3631, but changing the agent to wait longer for this error broke enrollment in a Docker container.

belimawr:
[...]

PR #3765, opened 2023-11-14
Fix index pattern when querying ES and condition when searching logs

belimawr:
## What does this PR do?

Two issues were causing flakiness on integration tests and are fixed by this PR:
1. The pattern used to query ES was not working on serverless, this commit updates it to a pattern that works on both stateful and serverless as well as make it more specific to the indexes/data streams we want to query
2. `findESDocs` did not wait for the data to be indexed, only to a successful query on ES. In some cases the documents the test wanted were not indexed yet, leading to 0 documents being returned and the test failing with no error in the logs/diagnostics. This is fixed by waiting for a document count > 0 and no error.

## Why is it important?

It fixes flaky integration tests.

## Checklist

~~## Author's Checklist~~

Run the integration test that was failing with both serverless and stateful
```
STACK_PROVISIONER=serverless TEST_PLATFORMS="linux/amd64" SNAPSHOT=true mage -v integration:single TestLogIngestionFleetManaged
mage integration:clean
TEST_PLATFORMS="linux/amd64" SNAPSHOT=true mage -v integration:single TestLogIngestionFleetManaged
```

It is needed to clean the deployments because at the moment the integration test framework does not keep track of the type of stack it deployed, more information: https://github.com/elastic/elastic-agent/issues/3756

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3741
- Closes https://github.com/elastic/elastic-agent/pull/3743

~~## Use cases~~
~~## Screenshots~~
~~## Logs~~

## Questions to ask yourself

- ...


pkg/testing/tools/estools/elasticsearch.go
```diff
@@ -488,7 +488,7 @@ func GetLogsForDatasetWithContext(ctx context.Context, client elastictransport.I
 		},
 	}
 
-	return performQueryForRawQuery(ctx, indexQuery, "*ds-logs*", client)
```
cmacknz, line 491: This must have been querying the underlying data stream backing indexes directly then? That wasn't correct if so.

Everything follows the data stream naming scheme so `log*` is correct if you are querying for all logs data streams.

In our own tests for elastic agent logs you should be able to do `logs-elastic_agent*` but this ES instance contains such few documents I'm not sure this make much of a difference.

testing/integration/logs_ingestion_test.go
```diff
@@ -104,15 +104,11 @@ func testMonitoringLogsAreShipped(
 ) {
 	// Stage 1: Make sure metricbeat logs are populated
 	t.Log("Making sure metricbeat logs are populated")
-	require.Eventually(t,
```
cmacknz, line 107: Will removing the eventually here make this flaky again?

testing/integration/logs_ingestion_test.go
```diff
@@ -199,14 +195,21 @@ func testMonitoringLogsAreShipped(
 
 func findESDocs(t *testing.T, findFn func() (estools.Documents, error)) estools.Documents {
 	var docs estools.Documents
+	start := time.Now()
+	defer func() {
+		fmt.Println(">>>>>>>>>>>>>>>>>>>> Took", time.Now().Sub(start))
```
cmacknz, line 200:
```suggestion
		t.Log(">>>>>>>>>>>>>>>>>>>> Took", time.Now().Sub(start))
```

Same applies for the one below, assuming you plan to keep these and they aren't just for debugging

testing/integration/logs_ingestion_test.go
```diff
@@ -199,14 +195,21 @@ func testMonitoringLogsAreShipped(
...
+	}()
+	count := 0
 	require.Eventually(
 		t,
 		func() bool {
+			count++
+			fmt.Println(">>>>>>>>>>>>>>>>>>>> Iteration ", count)
```
cmacknz, line 207:
```suggestion
			t.Logf(">>>>>>>>>>>>>>>>>>>> Iteration %d", count)
```

testing/integration/logs_ingestion_test.go
cmacknz: Ah I see there is a require eventually inside findESDocs.

pkg/testing/tools/estools/elasticsearch.go
cmacknz: We do this in other places in this file 

https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/pkg/testing/tools/estools/elasticsearch.go#L437

https://github.com/elastic/elastic-agent/blob/f7dcbd74ce0e27d3b46b478755168c66d78dc111/pkg/testing/tools/estools/elasticsearch.go#L365

cmacknz: I also see this filters by dataset already which should catch bugs where the data exists but in the wrong index, but limiting the query only to the index things are supposed to be in would be a stricter test.

testing/integration/logs_ingestion_test.go
belimawr: Exactly.

testing/integration/logs_ingestion_test.go
belimawr: That's just for debugging, I'm still trying to understand why it only fails to find the docs in CI....

belimawr:
The failure seems to be caused by: https://github.com/elastic/elastic-agent/issues/3664

belimawr:
buildkite test this

testing/integration/endpoint_security_test.go
```diff
@@ -493,7 +493,7 @@ func TestEndpointSecurityNonDefaultBasePath(t *testing.T) {
 	info := define.Require(t, define.Requirements{
 		Stack:   &define.Stack{},
 		Local:   false, // requires Agent installation
-		Isolate: false,
+		Isolate: true,
```
blakerouse, line 496: Why all the changes to isolate? That means this will add 4 more machines to perform tests, why?

This seems unrelated to the change implemented in this PR. I didn't see anything in the PR description describing this change and its reason.

blakerouse (changes requested):

testing/integration/endpoint_security_test.go
belimawr: That was me trying to get the integration tests to pass. Some exploratory testing because I just can't reproduce any of the CI failures when running the tests from my machine. I undid this change.

But I still have the changes from https://github.com/elastic/elastic-agent/pull/3786 to try getting CI happy. 

I'll request a re-review once it's the final code that can pass CI.

pkg/testing/tools/estools/elasticsearch.go
belimawr: Thanks Craig! I'll update it to be more strict then.

belimawr:
Folks, this PR is finally (well, hopefully) ready for a final review. I found another source of flakiness and fixed it in my last commit ff4edf2824.

cmacknz (approved):

blakerouse (approved):
Looks good!

Issue #3711, opened 2023-11-06
8.11 Docker image size increased to 4.13 GB from 2.3 GB


cmacknz:
Our Docker image size in 8.11 has increased to 8.11, which appears to be because of an additional Docker layer touching all files in the file system.

```
docker images | grep elastic-agent-cloud
docker.elastic.co/cloud-release/elastic-agent-cloud 8.11.0-e75ef729 fe8243cd3cad 45 hours ago 4.13GB
docker.elastic.co/cloud-release/elastic-agent-cloud 8.10.0          d500f2e33781  8 weeks ago 2.3GB
```

Using https://github.com/GoogleContainerTools/container-diff shows the actual container size has barely changed:

```
container-diff diff docker.elastic.co/cloud-release/elastic-agent-cloud:8.11.0-e75ef729 docker.elastic.co/cloud-release/elastic-agent-cloud:8.10.0

-----Size-----

Image size difference between docker.elastic.co/cloud-release/elastic-agent-cloud:8.11.0-e75ef729 and docker.elastic.co/cloud-release/elastic-agent-cloud:8.10.0:
SIZE1        SIZE2
2.2G         2.1G
```

The reason for the increase is revealed looking at the layer history:

```
8.11.0-e75ef729
IMAGE          CREATED        CREATED BY                                      SIZE      COMMENT
fe8243cd3cad   45 hours ago   RUN /bin/sh -c echo -e '#!/bin/sh\nexec /usr…   51B       buildkit.dockerfile.v0
<missing>      45 hours ago   CMD ["/app/apm.sh"]                             0B        buildkit.dockerfile.v0
<missing>      45 hours ago   ENTRYPOINT ["/usr/bin/tini" "--"]               0B        buildkit.dockerfile.v0
<missing>      45 hours ago   WORKDIR /usr/share/elastic-agent                0B        buildkit.dockerfile.v0
<missing>      45 hours ago   ENV LIBBEAT_MONITORING_CGROUPS_HIERARCHY_OVE…   0B        buildkit.dockerfile.v0
<missing>      45 hours ago   USER elastic-agent                              0B        buildkit.dockerfile.v0
<missing>      45 hours ago   RUN /bin/sh -c setcap cap_net_raw,cap_setuid…   105MB     buildkit.dockerfile.v0
<missing>      45 hours ago   RUN /bin/sh -c chown elastic-agent /app # bu…   0B        buildkit.dockerfile.v0
[...]

PR #3778, opened 2023-11-16
Elastic Agent ACI compliant image

michalpristas:
## What does this PR do?

Bringing back work which allows 
- image to run on Azure Container Instances (no root group)
- run cloud defend (heartbeat capabilities set)
- smaller size (got rid of double chown and reduced size of npm modules setup for synthethics)
- allow container to run as different user (get rid of world permission stripping)

Size comparison
```
branch
docker images | grep cloud
docker.elastic.co/beats-ci/elastic-agent-cloud. 8.12.0-SNAPSHOT  331c23d10995  5 minutes ago. 3.08GB

main
docker images | grep cloud
docker.elastic.co/beats-ci/elastic-agent-cloud  8.12.0-SNAPSHOT  a6e4e5b9ff81  6 minutes ago. 3.08GB
```

- Closes https://github.com/elastic/elastic-agent/issues/3711
- Closes https://github.com/elastic/elastic-agent/issues/82


michalpristas:
still working on -complete image size

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -6,14 +6,25 @@
...
+{{- end}}
+    # ECE needs to create config here under non-1000 user
+    chmod 0777 {{ $beatHome}} && \
+    mkdir -p {{ $beatHome }}/data {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/logs && \
     find {{ $beatHome }} -type d -exec chmod 0755 {} \; && \
     find {{ $beatHome }} -type f -exec chmod 0644 {} \; && \
-    find {{ $beatHome }}/data -type d -exec chmod 0770 {} \; && \
-    find {{ $beatHome }}/data -type f -exec chmod 0660 {} \; && \
+    find {{ $beatHome }}/data -type d -exec chmod 0777 {} \; && \
+    find {{ $beatHome }}/data -type f -exec chmod 0666 {} \; && \
```
pchila, line 27: Why do we need to create the user and the correct ownership in the intermediate stage ? Wouldn't it be enough to copy with the correct ownership and user in the last stage (line ~127) ?

pchila (approved):
Aside from some unnecessary change of ownership in the earlier stage, the change looks good.

Just out of curiosity: how hard would it be to always create the `elastic-agent` user and group, associate all the files to that and as last instruction add a USER instruction ? (ideally with the value being an argument instead of a template)

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
michalpristas: hmm that;s a nice idea, will check

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -27,7 +38,7 @@ RUN mkdir -p {{ $beatHome }}/data {{ $beatHome }}/data/elastic-agent-{{ commit_s
     (chmod 0755 {{ $beatHome }}/data/elastic-agent-*/components/pf-elastic-collector || true) && \
     (chmod 0755 {{ $beatHome }}/data/elastic-agent-*/components/pf-elastic-symbolizer || true) && \
     (chmod 0755 {{ $beatHome }}/data/elastic-agent-*/components/pf-host-agent || true) && \
-    find {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/components -name "*.yml*" -type f -exec chown root:root {} \; && \
+    find {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/components -name "*.yml*" -type f -exec chown {{ .user }}:{{ .user }} {} \; && \
```
michalpristas, line 41: this is a no, modules permissions are checked to be owned by root

cmacknz:
Did we confirm this image can be deployed to Elastic Cloud successfully and that ECE stack can come up.

I would also like @blakerouse to review this before merge to see if there is anything that will cause problems for the non-root/unprivileged agent work.

michalpristas:
i created cloud instance using our regular testing/cloud env. this was fine. 
for the use case lastly reported i haven't found a way to deploy custom image (QA labs does not seem to work)

michalpristas:
i'm seeing some weird behavior but i see it on main as well. 
i need to test this on another machine to rule out my env. i don't trust my mac

michalpristas:
seems like some env issues on my side, rebuilt on my linux machine and it seems to produce system metrics normally.

blakerouse (approved):
Looks good to me

PR #3779, opened 2023-11-16
Fix comparison of upgrade metadata

pchila:
## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/dispatcher/dispatcher.go
+++ b/internal/pkg/agent/application/dispatcher/dispatcher.go
@@ -321,7 +321,7 @@ func reportNextScheduledUpgrade(input []fleetapi.Action, detailsSetter details.O
 
 	upgradeDetails := details.NewDetails(nextUpgrade.Version, details.StateScheduled, nextUpgrade.ID())
 	startTime, _ := nextUpgrade.StartTime()
-	upgradeDetails.Metadata.ScheduledAt = startTime
+	upgradeDetails.Metadata.ScheduledAt = &startTime
 
 	detailsSetter(upgradeDetails)
 }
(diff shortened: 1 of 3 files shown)


michalpristas (approved):
let's get this in, this is blocking other PRs

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -321,7 +321,7 @@ func reportNextScheduledUpgrade(input []fleetapi.Action, detailsSetter details.O
 
 	upgradeDetails := details.NewDetails(nextUpgrade.Version, details.StateScheduled, nextUpgrade.ID())
 	startTime, _ := nextUpgrade.StartTime()
-	upgradeDetails.Metadata.ScheduledAt = startTime
+	upgradeDetails.Metadata.ScheduledAt = &startTime
```
pchila, line 324: I am not 100% sure this is correct, we probably need to test startTime for zero value (checked the `StartTime()` function)

pchila: see 73a9031524ab62212f12eef2428c29ea620aa458

rdner (approved):

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -320,8 +320,10 @@ func reportNextScheduledUpgrade(input []fleetapi.Action, detailsSetter details.O
 	}
 
 	upgradeDetails := details.NewDetails(nextUpgrade.Version, details.StateScheduled, nextUpgrade.ID())
-	startTime, _ := nextUpgrade.StartTime()
-	upgradeDetails.Metadata.ScheduledAt = startTime
+	startTime, err := nextUpgrade.StartTime()
```
michalpristas, line 323: we should not have an error here as we're guarding against this in a loop. 
in case we end up with an error here following lines are questionable
cc @ycombinator

internal/pkg/agent/application/dispatcher/dispatcher.go
ycombinator: It built before because two PRs independently succeeded in CI.  So they got merged into `main`.  But after that, the changes in them collided and we got this failure.

To prevent this in the future, we need to enable the *Require branches to be up to date before merging branch* protection.

Issue #3197, opened 2023-08-04
Collecting diagnostics in 8.10 takes 3 minutes if Elastic Defend is installed


cmacknz:
Collecting diagnostics now takes almost exactly 3 minutes every time when using the default configuration with the system integration on main (8.10).

```sh
❯ time sudo elastic-agent diagnostics
[WARNING] Could not redact state.yaml due to unmarshalling error: yaml: invalid map key: map[interface {}]interface {}{"unitid":"log-default-logfile-system-ee627672-4cb2-49fd-8a56-6406b056490a", "unittype":0}
Created diagnostics archive "elastic-agent-diagnostics-2023-08-04T18-09-42Z-00.zip"
***** WARNING *****
Created archive may contain plain text credentials.
Ensure that files in archive are redacted before sharing.
*******************
sudo elastic-agent diagnostics  0.12s user 0.06s system 0% cpu 3:00.21 total
```

Using the official 8.9 release this takes 40s, which is still long but much faster than 3 minutes:

```sh
 time sudo elastic-agent diagnostics
[WARNING] Could not redact state.yaml due to unmarshalling error: yaml: invalid map key: map[interface {}]interface {}{"unitid":"log-default", "unittype":1}
Created diagnostics archive "elastic-agent-diagnostics-2023-08-04T18-17-29Z-00.zip"
***** WARNING *****
Created archive may contain plain text credentials.
Ensure that files in archive are redacted before sharing.
*******************
sudo elastic-agent diagnostics  0.13s user 0.03s system 0% cpu 40.364 total
```

Reverting https://github.com/elastic/elastic-agent/pull/3118 brings the time back down to 40s, so it looks like something in that change caused this regression.

cmacknz:
@fearful-symmetry FYI

fearful-symmetry:
looking into this. @cmacknz what was the config? What were the enabled components/units?

fearful-symmetry:
Perhaps unsurprisingly, my normal agent set-up-and-test isn't reproducing this. Trying to figure out what I missed.

cmacknz:
I installed the agent using the default configuration, then enrolled it into Fleet where the agent policy had monitoring enabled and the system integration installed.

cmacknz:
```sh
sudo elastic-agent status --output=full
┌─ fleet
│  └─ status: (HEALTHY) Connected
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
[...]

PR #3793, opened 2023-11-20
[Upgrade Details] Use `details.DownloadRate` type from elastic-agent-libs

ycombinator:
## What does this PR do?

This is a refactoring PR. It moves the `details.DownloadRate` type from the `github.com/elastic/elastic-agent` module to the `github.com/elastic/elastic-agent-libs` module.

## Why is it important?

So packages within the `github.com/elastic/elastic-agent-libs` module can reuse the `details.DownloadRate` type; see https://github.com/elastic/elastic-agent-libs/pull/166.

## Related issues
* Closes https://github.com/elastic/elastic-agent/issues/3792

PR #3794, opened 2023-11-20
Use different timeout values for CPU and non-CPU diagnostic requests

fearful-symmetry:
## What does this PR do?

Part of https://github.com/elastic/elastic-agent/issues/3197, this is a fairly simple change that has the diagnostic request use two different timeout values depending on if we've requested additional CPU metrics. 

## Why is it important?

This is meant to prevent us from waiting extended periods of time when `endpoint` isn't responding to diagnostic requests.

## Checklist

~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have made corresponding change to the default configuration files~

### DIFF
--- a/changelog/fragments/1700520158-dynamic-timeout-for-diags.yaml
+++ b/changelog/fragments/1700520158-dynamic-timeout-for-diags.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: dynamic-timeout-for-diags
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: Use different timeout values for diagnostics requests, depending on if CPU diagnostics are requested
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: diagnostics
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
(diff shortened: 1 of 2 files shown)


pierrehilbert:
Test failing because of https://github.com/elastic/elastic-agent/pull/3793

changelog/fragments/1700520158-dynamic-timeout-for-diags.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: dynamic-timeout-for-diags
```
pchila, line 14: Since this ends up in the release notes maybe we can put something more friendly, like
```suggestion
summary: Use longer diagnostics timeout when CPU diagnostics are requested
```

pkg/component/runtime/manager.go
```diff
@@ -943,7 +946,14 @@ func (m *Manager) getListenAddr() string {
 // performDiagAction creates a diagnostic ActionRequest and executes it against the runtime that's mapped to the specified component.
 // if the specified actionLevel is ActionRequest_COMPONENT, the unit field is ignored.
 func (m *Manager) performDiagAction(ctx context.Context, comp component.Component, unit component.Unit, actionLevel proto.ActionRequest_Level, params client.DiagnosticParams) ([]*proto.ActionDiagnosticUnitResult, error) {
-	ctx, cancel := context.WithTimeout(ctx, diagnosticTimeout)
+	// if we're gathering CPU diagnostics, request a longer timeout; CPU diag collection requires the diagnostic hook to sit and gather a CPU profile.
+	finalDiagnosticTime := diagnosticTimeout
+	for _, tag := range params.AdditionalMetrics {
+		if tag == "CPU" {
+			finalDiagnosticTime = diagnosticTimeoutCPU
```
pchila, line 953: maybe we can break here? Not sure how many additional metrics there can be but we don't need to iterate more after we found the "CPU" tag...

```suggestion
			finalDiagnosticTime = diagnosticTimeoutCPU
			break
```

pchila (commented):
A couple of minor suggestions, code is ok otherwise.

Maybe it would be worth adding a test (doesn't have to be a full-fledged integration test, probably a simple combination of runtime manager and fake input would work) so that we can put a hard limit on how long it takes for diagnostics in case we have a component that does not support the diagnostics action ?

fearful-symmetry:
@pchila yah, was kinda on the fence about tests, since I wasn't sure how effective it would be to test essentially a single `if` statement, unless you're thinking of something else.

pchila:
Well I was thinking of adding a test here to detect if code changes (not just in this PR) make elastic-agent diagnostics wait a lot longer in case some component does not implement the diagnostic action. 
The idea is to reproduce the bug/usecase conditions and test for expected behavior, not the code that patches the bug. 
We don't have guidelines about test coverage beyond a simple measurement so it's up to you to choose if you want to add a test here

pchila (approved):

AndersonQ (approved):

PR #3800, opened 2023-11-22
Embedded Otel Collector into Agent

michalpristas:
## What does this PR do?

This PR adds an ability to start agent in collector mode.
So far logic is simple. When specifying config using `-c` flag it checks the name of the config.
If config is `otel`, `otlp`, `otlpcol` then it starts in otel mode.

When name does not match but config contains `service` and `exporters`, `importers` then we assume collector config and start in otel mode.

Otherwise we start in a standard way.

On top of that, we provide custom config provider which serves a default value for logging output == `stdout` instead of otel default `stderr`.

Tested on darwin and windows. when changing `elastic-agent.yml` to contain otel config even service is started in otel mode

## DoD from issue

- [x] A log message clearly states we're starting in OTel mode and its experimental status
- [x] We should not open any ports on the system for the coordinator
- [x] Log messages from OTel collector and agent should go to `stdout`

## Checklist


michalpristas:
i'm seeing some weird behavior but i see it on main as well. 
i need to test this on another machine to rule out my env. i don't trust my mac

cmacknz:
Duplicating the test requirements here so we know when this is done, at least the first test should exist as part of this PR. The second could be a follow up.

> ### Test requirements
> 
> We should start with some very basic e2e tests that ensure that we don't break OTel mode:
> - One test that uses an `otlpreceiver` and the file exporter to verify we can receive OTLP logs
> - One test that uses the `filelogreceiver` to ingest a very simple log file and exports it via OTLP to APM Server on ESS. Query the `logs-apm*` indices to verify the logs were delivered.

internal/pkg/agent/cmd/run.go
```diff
@@ -133,6 +134,22 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
...
+	}
+
+	if runAsOtel {
+		return otel.Run(ctx, cancel, stop, testingMode)
+	}
+
+	// not otel continue as usual
+	return runElasticAgent(ctx, cancel, override, stop, testingMode, fleetInitTimeout, modifiers...)
+
+}
```
blakerouse, line 150: Would really like to see otel integrated into the coordinator and runtime system. That way we get all the status reporting, and the ability to run both at the same time.

michalpristas: @joshdover what was your idea of the scope for the initial integration?

internal/pkg/agent/cmd/run.go
```diff
@@ -133,6 +134,22 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
...
 
+	// detect otel
+	runAsOtel, err := otel.IsOtelConfig(ctx, paths.ConfigFile())
+	if err != nil {
+		return err
+	}
+
+	if runAsOtel {
+		return otel.Run(ctx, cancel, stop, testingMode)
+	}
```
pchila, line 145: Shouldn't we check and test the config before parsing args and in case we have to run in otel mode let the collector parse its own flags ? I had a quick look at otelcol CLI args:

```
Flags:
      --config --config=file:/path/to/first --config=file:path/to/second   Locations to the config file(s), note that only a single location can be set per flag entry e.g. --config=file:/path/to/first --config=file:path/to/second. (default [])
      --feature-gates flag                                                 Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature. (default confmap.expandEnabled,coreinternal.attraction.hash.sha256,-filter.filterlog.useOTTLBridge,-filter.filtermetric.useOTTLBridge,-filter.filterspan.useOTTLBridge,pkg.translator.prometheus.NormalizeName,-pkg.translator.prometheus.PermissiveLabelSanitization,receiver.jaeger.DisableRemoteSampling,receiver.jaegerreceiver.replaceThriftWithProto,-receiver.prometheusreceiver.UseCreatedMetric,-telemetry.disableHighCardinalityMetrics,-telemetry.useOtelForInternalMetrics,-telemetry.useOtelWithSDKConfigurationForInternalTelemetry)
  -h, --help                                                               help for otelcol
      --set func                                                           Set arbitrary component config property. The component has to be defined in the config file and the flag has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s
  -v, --version                                                            version for otelcol
```
I imagine that we won't support feature gates or arbitrary component property overrides in this PR... Do we have plan supporting other collector CLI args other than config ?

michalpristas: it is validated in a run function of otel collector why do it twice?

michalpristas: config is validated in a run function of otel collector, no need to do it twice. i don;t think we need to match collector subcommands or flags as well

ycombinator: One idea might be to introduce an `otel` or `otelcol` subcommand so we can hang as many otel-specific options off it without risk of collisions with existing options.  So, for example:

```
elastic-agent otel --config
elastic-agent otel --feature-gates
...
```

or 

```
elastic-agent otelcol --config
elastic-agent otelcol --feature-gates
...
```

This way, users of the Elastic OTel Collector distribution need only change the command and the sub-command to use it over any other distribition, e.g. `otelcol ...` -> `elastic-agent otel ...` or `elastic-agent otelcol ...`.

michalpristas: Pushed support for `--config` into this PR and create issues for followup work for things discussed

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+      receivers: [filelog]
+      exporters:
+        - file`)
+
+var apmProcessingContent = `2023-06-19 05:20:50 ERROR This is a test error message
+2023-06-20 12:50:00 DEBUG This is a test debug message 2
+2023-06-20 12:51:00 DEBUG This is a test debug message 3
+2023-06-20 12:52:00 DEBUG This is a test debug message 4`
+
+var apmOtelConfig = `receivers:
```
pchila, line 52: Nit: these can be const
```suggestion
const apmProcessingContent = `2023-06-19 05:20:50 ERROR This is a test error message
2023-06-20 12:50:00 DEBUG This is a test debug message 2
2023-06-20 12:51:00 DEBUG This is a test debug message 3
2023-06-20 12:52:00 DEBUG This is a test debug message 4`

const apmOtelConfig = `receivers:
```

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+	// otel mode should be detected automatically
+	err = fixture.Configure(ctx, fileProcessingConfig)
+	require.NoError(t, err)
+
+	var fixtureWg sync.WaitGroup
+	fixtureWg.Add(1)
+	go func() {
+		err = fixture.RunWithClient(ctx, false)
+		fixtureWg.Done()
+	}()
```
pchila, line 127: Nit: safety from panics ?
```suggestion
	go func() {
		defer fixtureWg.Done()
		err = fixture.RunWithClient(ctx, false)
	}()
```

internal/pkg/otel/config_file_provider_test.go
```diff
@@ -0,0 +1,41 @@
...
+		expectedOutputs []string
+	}{
+		{"default", "otel.yml", []string{"stdout"}},
+		{"stderr", "otlp.yml", []string{"stderr"}},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			cp := NewFileProviderWithDefaults()
+			confMap, err := cp.Retrieve(context.TODO(), "file:"+filepath.Join(".", "testdata", "otel", tc.configFile), nil)
```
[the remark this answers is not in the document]
michalpristas: this is what otel accepts. they strip the `file:` prefix and use a path

internal/pkg/agent/cmd/run.go
ycombinator: @michalpristas I don't see a follow up issue for implementing the `elastic-agent otel` or `elastic-agent otelcol` subcommand as discussed above.  I checked the elastic-agent repo and our private repos as well.  

Could you please make one link to it from here?  Thanks.

pchila (approved):

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+const buildDescription = "Elastic opentelemetry-collector distribution"
+
+func IsOtelConfig(ctx context.Context, pathConfigFile string) (bool, error) {
+	fileName := filepath.Base(pathConfigFile)
+	cleanFileName := strings.TrimSpace(strings.ToLower(strings.TrimSuffix(fileName, filepath.Ext(fileName))))
+	if cleanFileName == "otel" || cleanFileName == "otlp" || cleanFileName == "otelcol" {
+		return true, nil
+	}
```
ycombinator, line 32: IIUC, the only expected extensions are `.yml` or `.yaml`.  Could we be a bit stricter here and only accept those extensions?

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/internal/pkg/agent/errors"
+	"github.com/elastic/elastic-agent/internal/pkg/config"
+	"github.com/elastic/elastic-agent/internal/pkg/release"
+)
+
+const buildDescription = "Elastic opentelemetry-collector distribution"
+
+func IsOtelConfig(ctx context.Context, pathConfigFile string) (bool, error) {
```
ycombinator, line 27: Could we add a godoc for this function documenting exactly the heuristics to determine if a configuration is an OTel configuration?  That way one doesn't need to read the function's implementation to figure that out.

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if hasAgent || hasInputs || hasOutputs {
+		return false, nil
+	}
+
+	// contains otel service definition
+	_, hasService := mapConfig["service"]
+	_, hasExporters := mapConfig["exporters"]
+	_, hasReceivers := mapConfig["receivers"]
+	_, hasProcessors := mapConfig["processors"]
```
ycombinator, line 62: Should [`extensions`](https://opentelemetry.io/docs/collector/configuration/#extensions) be added to this list as well?

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,121 @@
...
+	go func() {
+		<-stop
+		cancel()
+	}()
+	defer cancel()
+
+	return svc.Run(cancelCtx)
+}
+
+func newSettings(configPaths []string, version string) (*otelcol.CollectorSettings, error) {
```
ycombinator, line 95: Any reason `configPaths` is a `[]string` and not just `string`? AFAICT, the only caller of `newSettings` is just passing a single path for this argument.

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,121 @@
...
+	provider, err := otelcol.NewConfigProvider(configProviderSettings)
+	if err != nil {
+		return nil, err
+	}
+
+	return &otelcol.CollectorSettings{
+		Factories:               components,
+		BuildInfo:               buildInfo,
+		ConfigProvider:          provider,
+		DisableGracefulShutdown: true,
```
ycombinator, line 119: Probably worth commenting here that we're disabling graceful shutdown as we're handling that via the `cancelCtx` being passed to the collector's `Run` method in the `Run` function above?

pkg/testing/fixture.go
```diff
@@ -369,7 +440,9 @@ func (f *Fixture) RunBeat(ctx context.Context) error {
 //
 // The `elastic-agent.yml` generated by `Fixture.Configure` is ignored
 // when `Run` is called.
-func (f *Fixture) Run(ctx context.Context, states ...State) error {
+//
+// if isClientEnabled is set to bool, communicating state does not happen.
```
ycombinator, line 444: Typo?
```suggestion
// if isClientEnabled is set to false, communicating state does not happen.
```

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+	var fixtureWg sync.WaitGroup
+	fixtureWg.Add(1)
+	go func() {
+		fixture.RunWithClient(ctx, false)
+		fixtureWg.Done()
+	}()
+
+	go func() {
+		// delayed write
+		<-time.After(30 * time.Second)
```
ycombinator, line 233: Why do we need this delay? Does it have something to do with Agent starting up in collector mode first or is there some other reason?

Thirty seconds is a sizeable delay IMO. If we could replace this with blocking on a receive from a channel or something like a wait-for pattern that polls for something every second up to 30 seconds, that would be better.

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func TestAPMIngestion(t *testing.T) {
+	info := define.Require(t, define.Requirements{
+		Group: Default,
+		Stack: &define.Stack{},
+		Local: true,
+		OS: []define.OS{
+			// apm server not supported on darwin
+			{Type: define.Linux},
```
ycombinator, line 150: Should we also run this on Windows?

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+	require.True(t, len(esHost) > 0)
+
+	esUsername := os.Getenv("ELASTICSEARCH_USERNAME")
+	require.True(t, len(esUsername) > 0)
+
+	esPass := os.Getenv("ELASTICSEARCH_PASSWORD")
+	require.True(t, len(esPass) > 0)
+
+	esClient := info.ESClient
+	if esClient == nil {
```
ycombinator, line 192: Why do we need this check?  Under what circumstances would `info.ESClient` not be set in this test?

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return "", err
+	}
+	if parsedES.Port() == "" {
+		fixedESHost = fmt.Sprintf("%s:443", fixedESHost)
+	}
+	return fixedESHost, nil
+}
+
+func getESApiKey(esClient *elasticsearch.Client, esHost, esUser, esPass string) (string, error) {
```
ycombinator, line 272: Nit: this is a side-effecting function so could we name it `createESApiKey`?

testing/integration/otel_test.go
```diff
@@ -0,0 +1,295 @@
...
+	require.NoError(t, err)
+
+	// prepare input
+	agentWorkDir := fixture.WorkDir()
+	fileName := "content.log"
+	fixture.WriteFileToWorkDir(ctx, "", fileName)
+
+	testUuid, err := uuid.NewV4()
+	require.NoError(t, err, "failed to create test id")
+	testId := testUuid.String()
```
ycombinator, line 170: Could you use `info.Namespace` here instead?

internal/pkg/otel/run.go
michalpristas: yes it was just to conform with underlying interface. i can change that

testing/integration/otel_test.go
michalpristas: i had this for use cases when commented out using Stack and used my own

testing/integration/otel_test.go
michalpristas: yes this is just an arbitrary number of seconds we let otel to work, detect file .... we don't have a way of being informed about processing being done atm.

michalpristas:
hittin issue with apm version skew. need to figure that out

michalpristas:
@ycombinator updated, i skipped widows in e2e apm test for now due to this apm version mismatch at this time. test is skipped. 
once we have override gone i will revisit for windows

testing/integration/otel_test.go
ycombinator: Since we are using `Stack: &define.Stack{},` now, can we remove this section and the `getESClient` function?  None of the other integration tests have something like this so I don't think we should introduce this usage pattern here. 

If you want to use your own stack while debugging this test, you should be able to keep `Stack: &define.Stack{},` uncommented and run the test like so (this is what I do):

```
sudo ELASTICSEARCH_USERNAME=xxxx ELASTICSEARCH_PASSWORD=xxxxxxxxxxxx ELASTICSEARCH_HOST=xxxxxxxxxxx TEST_DEFINE_PREFIX=foo  go test -tags integration -v ./testing/integration/... -test.run TestAPMIngestion -count 1
```

testing/integration/otel_test.go
```diff
@@ -0,0 +1,306 @@
...
+		},
+		5*time.Minute, 500*time.Millisecond,
+		"there should be exported logs by now")
+
+	cancel()
+	fixtureWg.Wait()
+	require.True(t, err == nil || err == context.Canceled || err == context.DeadlineExceeded, "Retrieved unexpected error: %s", err.Error())
+}
+
+func TestAPMIngestion(t *testing.T) {
```
ycombinator, line 145: Let's put OTel somewhere in the name of the test as all integration tests are in a single namespace.

testing/integration/otel_test.go
```diff
@@ -0,0 +1,306 @@
...
+service:
+  pipelines:
+    logs:
+      receivers: [filelog]
+      processors: [resource]
+      exporters:
+        - debug
+        - otlp/elastic`
+
+func TestFileProcessing(t *testing.T) {
```
ycombinator, line 96: Let's put OTel somewhere in the name of the test as all integration tests are in a single namespace.

testing/integration/otel_test.go
```diff
@@ -0,0 +1,306 @@
...
+	var fixtureWg sync.WaitGroup
+	fixtureWg.Add(1)
+	go func() {
+		fixture.RunWithClient(ctx, false)
+		fixtureWg.Done()
+	}()
+
+	go func() {
+		// delayed write
+		<-time.After(10 * time.Second)
```
ycombinator, line 244: Ten seconds delay is better than thirty seconds (as was the case previously), but is there no way to poll for elastic agent logs and look for a specific string or something to know that it's safe to write the file?  

Agent supports a `logs` subcommand.  We could define a corresponding `ExecLogs` method on the Agent test fixture in https://github.com/elastic/elastic-agent/blob/main/pkg/testing/fixture.go and also a `WaitForLogs` function (similar to [`upgradetest.WaitForWatcher`](https://github.com/elastic/elastic-agent/blob/e3c0695757b3799335214ae20af2e847c0d8cc00/testing/upgradetest/watcher.go#L36)) that periodically calls `ExecLogs` under the hood until some timeout or the expected log entry is found.

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,138 @@
...
+	go func() {
+		<-stop
+		cancel()
+	}()
+	defer cancel()
+
+	return svc.Run(cancelCtx)
+}
+
+func newSettings(configPaths string, version string) (*otelcol.CollectorSettings, error) {
```
ycombinator, line 110: Nit: rename to `configPath` (singular) since we're only expecting one path here now.

internal/pkg/agent/cmd/run.go
ycombinator: Created the issue myself: https://github.com/elastic/elastic-agent/issues/3903

internal/pkg/otel/run.go
ycombinator: I don't think this has been implemented yet?  If this code sees `otel.json` it will return `(true, nil)`, right?

ycombinator:
@michalpristas Left a few more comments.  Couple of those might be hidden by GitHub so please check all unresolved conversations, including any hidden ones.  Thanks!

testing/integration/otel_test.go
```diff
@@ -0,0 +1,306 @@
...
+	stackVersion := info.KibanaClient.GetVersion()
+	agentVersion, err := version.New(aVersion.Agent)
+	require.NoError(t, err, "failed to get agent version")
+
+	if stackVersion.LessThan(agentVersion) {
+		// at the transition period when new version of stack is not yet released
+		// and we use override with previous version this test would fail with
+		// events being produced by newer APM than stack version
+		t.Skip("agent version needs to be equal to stack version")
+	}
```
ycombinator, line 165: @cmacknz @blakerouse I thought we ran into this sort of issue with other integration tests and solved it in a more global fashion, that is, in a way that would work for all integration tests and every test wouldn't need skip logic like what's implemented here?

testing/integration/otel_test.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return "", err
+	}
+	if parsedES.Port() == "" {
+		fixedESHost = fmt.Sprintf("%s:443", fixedESHost)
+	}
+	return fixedESHost, nil
+}
+
+func createESApiKey(esClient *elasticsearch.Client, esHost, esUser, esPass string) (string, error) {
```
ycombinator, line 288: Doesn't look like `esHost`, `esUser`, `esPass` are used in this function?

pkg/testing/log_watcher.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+)
+
+var _ Logger = &LogWatcher{}
```
ycombinator, line 15: What is this used for?

testing/integration/otel_test.go
ycombinator: Nice idea with the `LogWatcher`!

ycombinator:
@michalpristas Left a couple of very minor comments/questions, mostly looks like some missing cleanup.

pkg/testing/log_watcher.go
michalpristas: just to make sure it implements Logger interface, even if you're not using it anywhere (in case it's removed from test) check is performed

michalpristas:
test this

testing/integration/otel_test.go
```diff
@@ -0,0 +1,294 @@
...
+	// start apm default config just configure ES output
+	esHost, err := getESHost()
+	require.NoError(t, err, "failed to get ES host")
+	require.True(t, len(esHost) > 0)
+
+	esUsername := os.Getenv("ELASTICSEARCH_USERNAME")
+	require.True(t, len(esUsername) > 0)
+
+	esPass := os.Getenv("ELASTICSEARCH_PASSWORD")
+	require.True(t, len(esPass) > 0)
```
ycombinator, line 192: After the cleanup you did in 60dcc2c110019f738a8323c30508a3d46cc6fd40, I don't think you need these variables any more.

ycombinator (approved):
LGTM — we now have a foundation to build on — thanks!

michalpristas:
ci is failing on typo no longer present

michalpristas:
e2e failing on endpoint not being able to connect to ES 
@cmacknz @pierrehilbert could we force push? sonar is complaining about coverage but this PR is mostly tests (code lines wise)not sure why it's so pessimistic. 
e2e failures not related

internal/pkg/otel/run.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return false, errors.New(err,
+			errors.TypeConfig,
+			errors.M(errors.MetaKeyPath, pathConfigFile))
+	}
+
+	// contains agent definition
+	_, hasAgent := mapConfig["agent"]
+	_, hasInputs := mapConfig["inputs"]
+	_, hasOutputs := mapConfig["outputs"]
```
cmacknz, line 66: These terms are so generic I worry we may regret this one day. Especially since we aren't in complete control of what keys end up in an otel collector configuration.

I can't see an Elastic Agent configuration ever using the terms `exporters` or `receivers` since we already have names for those, can you simplify into just checking for the presence of those?

cmacknz: Although we may need to consider the possibility that we allow otel configuration to co-exist inside of the elastic-agent.yml file one day.

Could we just support configuration detection based on the file name initially?

cmacknz: For example this would break if we decided to start using `---` to separate the two configurations in one file.

testing/integration/otel_test.go
```diff
@@ -0,0 +1,288 @@
...
+    logs:
+      receivers: [filelog]
+      processors: [resource]
+      exporters:
+        - debug
+        - otlp/elastic`
+
+func TestOtelFileProcessing(t *testing.T) {
+	define.Require(t, define.Requirements{
+		Group: Default,
```
cmacknz, line 94: You could create an Otel specific test group to help speed these tests up depending on how long they take. Each group runs on its own VM.

cmacknz:
For reference this PR increases the size of the elastic-agent binary by 22 MB on my Mac.

```
du -sh data/elastic-agent-1af429/elastic-agent.app/Contents/MacOS/elastic-agent
 74M    data/elastic-agent-1af429/elastic-agent.app/Contents/MacOS/elastic-agent

du -sh ../elastic-agent-8.11.3-darwin-aarch64/data/elastic-agent-f4f6fb/elastic-agent.app/Contents/MacOS/elastic-agent
 52M    ../elastic-agent-8.11.3-darwin-aarch64/data/elastic-agent-f4f6fb/elastic-agent.app/Contents/MacOS/elastic-agent
```

testing/integration/otel_test.go
```diff
@@ -0,0 +1,288 @@
...
+	}()
+
+	require.Eventually(t,
+		func() bool {
+			// verify file exists
+			content, err := os.ReadFile(fileProcessingFilename)
+			return err == nil && len(content) > 0
+		},
+		5*time.Minute, 500*time.Millisecond,
+		"there should be exported logs by now")
```
cmacknz, line 134: Isn't this just testing that a file gets created? What if it were full of garbage? What if the collector creates the file and immediately dies?

Would it not be a better test to ingest from a file with static contents and verify the log lines are what we expect?

testing/integration/otel_test.go
cmacknz: Since these tests don't require a stack they will probably execute pretty quickly

testing/integration/otel_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// apm mismatch or proper docs in ES
+	err = logWatcher.WaitForKeys(context.Background(),
+		20*time.Second,
+		500*time.Millisecond,
+		apmVersionMismatch)
+	if err == nil {
+		// err nil means there's a match within time limit
+		// mark skipped to make it explicit it was not successfully evaluated
+		t.Skip("agent version needs to be equal to stack version")
```
cmacknz, line 246: Why can we only make this check at the end? Don't we know the versions we are starting at the start of the test?

cmacknz: Also why is this check timing dependent? How would you catch cases where this test is being skipped when it shouldn't be?

testing/integration/otel_test.go
```diff
@@ -0,0 +1,288 @@
...
+	} else {
+		// failed to get APM version mismatch in time
+		// processing should be running
+		require.Eventually(t,
+			func() bool {
+				docs := findESDocs(t, func() (estools.Documents, error) {
+					return estools.GetLogsForIndexWithContext(context.Background(), esClient, "logs-apm*", match)
```
cmacknz, line 253: You should probably pass in a real context, if the underlying HTTP client doesn't have a timeout on it this request could hang forever.

testing/integration/otel_test.go
```diff
@@ -0,0 +1,288 @@
...
+				})
+				hits = len(docs.Hits.Hits)
+				return hits > 0
```
cmacknz, line 256: Similar to the test above, you should be verifying that the index contains what you expect it to contain, not just that it contains anything at all.

internal/pkg/otel/run.go
michalpristas: i'm ok with file name

testing/integration/otel_test.go
michalpristas: we cannot check the version because we know version of agent and version of stack. what's causing issues is version of APM integration. This bubbles up only during runtime as an APM log.

testing/integration/otel_test.go
michalpristas: maybe we should think about updating beats test as well as this is a copy of it.

michalpristas:
failing on unrelated e2e again

testing/integration/otel_test.go
cmacknz: You could probably use the Fleet API to get the version of the APM integration instead. Probably the package policies API to find the APM package details https://www.elastic.co/guide/en/fleet/8.11/fleet-apis.html#PackagePolicies

cmacknz (approved):

cmacknz:
Both test failures are known issues. I'm going to force merge this one.

PR #3815, opened 2023-11-24
Fix install/enroll command not  failing when the daemon restart fails

belimawr:
For the agent to be enrolled it needs to restart after the enrol process is completed, in order to pickup the new config and "connect" to fleet-server.

This change makes the enrol command to fail if it cannot restart the agent after enrolling on fleet. A `--skip-daemon-reload` CLI flag is also added to the `enroll` command because the `container` command calls `enroll`, however there is no running daemon to restart. Skipping the daemon reload on this case allows the `container` command to succeed.

This PR brings back the changes introduced by https://github.com/elastic/elastic-agent/pull/3554 that were reverted due to the `container` command not working and fixes the issues with the `container` command.

## What does this PR do?

## Why is it important?

For the agent to be enrolled it needs to restart after the enrol process is completed, in order to pickup the new config and "connect" to fleet-server.

## Checklist

~~## Author's Checklist~~

Export the follwing environment variables
```
export FLEET_ENROLL=1
export FLEET_URL=https://testing-this-PR.elastic-cloud.com:443
export FLEET_ENROLLMENT_TOKEN=my-secret-token
```
Then execute the `container` command: 
```
./elastic-agent container -e -v -d "*"
```

The Elastic-Agent should enrol in Fleet and start working.

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/3664
- Brings back: https://github.com/elastic/elastic-agent/pull/3554

~~## Use cases~~
~~## Screenshots~~
~~## Logs~~

## Questions to ask yourself

- ...


internal/pkg/agent/cmd/enroll.go
```diff
@@ -75,6 +75,7 @@ func addEnrollFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolP("delay-enroll", "", false, "Delays enrollment to occur on first start of the Elastic Agent service")
 	cmd.Flags().DurationP("daemon-timeout", "", 0, "Timeout waiting for Elastic Agent daemon")
 	cmd.Flags().DurationP("fleet-server-timeout", "", 0, "Timeout waiting for Fleet Server to be ready to start enrollment")
+	cmd.Flags().Bool("skip-daemon-reload", false, "Skip daemon reload after enrolling")
```
pchila, line 78: If this flag is not intended for generic use (that is if a regular user is not supposed to specify it) maybe we want to hide it ?

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -445,24 +453,35 @@ func (c *enrollCmd) prepareFleetTLS() error {
...
 func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 	err := c.daemonReload(ctx)
+	if err != nil &&
+		(errors.Is(err, context.DeadlineExceeded) ||
+			errors.Is(err, context.Canceled)) {
+		return fmt.Errorf("could not reload daemon: %w", err)
+	}
 	if err == nil {
 		return nil
 	}
```
pchila, line 463: Maybe  we can simplify a bit
```suggestion
	if err == nil {
		return nil
	}
	if errors.Is(err, context.DeadlineExceeded) ||
			errors.Is(err, context.Canceled) {
		return fmt.Errorf("could not reload daemon: %w", err)
	}
```

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -445,24 +453,35 @@ func (c *enrollCmd) prepareFleetTLS() error {
...
+		if err != nil &&
+			(errors.Is(err, context.DeadlineExceeded) ||
+				errors.Is(err, context.Canceled)) {
+			return fmt.Errorf("could not reload daemon after %d retries: %w",
+				i+1, err)
+		}
 		if err == nil {
-			break
+			return nil
 		}
```
pchila, line 481: Same here (cannot suggest because of the deleted line -.- )

```
		if err == nil {
			return nil
		}
		if errors.Is(err, context.DeadlineExceeded) ||
				errors.Is(err, context.Canceled) {
			return fmt.Errorf("could not reload daemon after %d retries: %w",
				i+1, err)
		}

```

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -445,24 +453,35 @@ func (c *enrollCmd) prepareFleetTLS() error {
 
 func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 	err := c.daemonReload(ctx)
+	if err != nil &&
```
pchila, line 456: Maybe we can add a debug log about the errors we receive and retry while trying to restart the agent ?

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -216,16 +228,24 @@ func TestEnroll(t *testing.T) {
...
 			streams, _, _, _ := cli.NewTestingIOStreams()
-			err = cmd.Execute(context.Background(), streams)
-			require.NoError(t, err)
-
-			require.True(t, store.Called)
-
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+			defer cancel()
+			err = cmd.Execute(ctx, streams)
+			if err != nil &&
```
pchila, line 234: is the error always returned here ?

pchila (approved):
A couple of nitpicks here and there, overall the change looks ok

internal/pkg/agent/cmd/enroll.go
blakerouse: I agree with @pchila it is only being used by the container it should be hidden. I don't think we want users to use it and then complain that things didn't work as expected.

internal/pkg/agent/cmd/enroll_cmd_test.go
belimawr: I'm not quite sure what you mean.

If there is an error it should be returned.

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -216,16 +228,24 @@ func TestEnroll(t *testing.T) {
...
+				// There is no agent running, therefore nothing to be restarted.
+				// However, this will cause the Enroll command to return an error
+				// which we'll ignore here.
+				!strings.Contains(err.Error(),
+					"could not reload agent daemon, unable to trigger restart") {
+				t.Fatalf("enrrol coms returned and unexpected error: %v", err)
```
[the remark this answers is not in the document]
belimawr: Here the opposite is being done: the test fails if the error **does not** contain `could not reload agent daemon, unable to trigger restart`.

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -275,16 +295,24 @@ func TestEnroll(t *testing.T) {
...
+			defer cancel()
+			err = cmd.Execute(ctx, streams)
+
+			if err != nil &&
+				// There is no agent running, therefore nothing to be restarted.
+				// However, this will cause the Enroll command to return an error
+				// which we'll ignore here.
+				!strings.Contains(err.Error(),
+					"could not reload agent daemon, unable to trigger restart") {
+				t.Fatalf("enrrol coms returned and unexpected error: %v", err)
```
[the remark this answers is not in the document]
belimawr: Same as above, we want to fail if the error does not contain the given string.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -448,21 +456,31 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 	if err == nil {
 		return nil
 	}
+	if errors.Is(err, context.DeadlineExceeded) ||
```
cmacknz, line 459: Why are we only considering the context errors here?

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -448,21 +456,31 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
...
+			errors.Is(err, context.Canceled) {
+		return fmt.Errorf("could not reload daemon: %w", err)
+	}
 
 	signal := make(chan struct{})
+	defer close(signal)
 	backExp := backoff.NewExpBackoff(signal, 10*time.Second, 1*time.Minute)
 
-	for i := 5; i >= 0; i-- {
+	for i := 0; i < 5; i++ {
```
cmacknz, line 468: Where did the 5 here come from? 60s / 10s = 6 so I'd expect 6 iterations.

I think this would be simpler and clearer if we used https://github.com/cenkalti/backoff since it has native support for contexts, see example in  https://github.com/elastic/elastic-agent/blob/97e82175953af75fbe29fccc3cfcc5688a96dc8f/internal/pkg/agent/application/upgrade/step_download.go#L241-L265

internal/pkg/agent/cmd/enroll_cmd.go
belimawr: That's the retry with backoff, so if the context is cancelled or the deadline is exceeded, then we should stop retrying and just return. Any other error we can keep on retrying.

Deadline exceeded and context cancelled are errors that are not related to the daemon reload and because the context is cancelled retrying will always result in the same error.

cmacknz: Why is this first call separated out? I guess it was like that before but the way this function is implemented is harder to follow than usual.

internal/pkg/agent/cmd/enroll_cmd.go
cmacknz: I still don't obviously see where the number 5 came from. The advantage of the package I suggested is you can just do `context.WithoutTimeout(ctx, timeout)` and limit the maximum retry time easily and also automatically handle the context errors instead of doing that separately.

We could improve our backoff implementation, but that seems like a waste when there's an off the shelf implementation that already does it (which I have used widely even prior to Elastic).

cmacknz: Anyway I'm not blocking the review on changing the backoff implementation, I just think it would make this easier to follow.

belimawr:
buildkite test this

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -448,21 +456,31 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
...
 		backExp.Wait()
 		c.log.Info("Retrying to restart...")
+
```
cmacknz, line 471: I can't comment on the line `c.log.Info("Retrying to restart...")` above since it didn't change, but we should log which retry this is and what the last error was.

cmacknz: We aren't logging the error message from failed retries right now.

cmacknz:
I tried this out by modifying agent so that it unconditionally retried restarting the daemon with the following patch applied:

```diff
diff --git a/internal/pkg/agent/cmd/enroll_cmd.go b/internal/pkg/agent/cmd/enroll_cmd.go
index e704990426..a1facdf185 100644
--- a/internal/pkg/agent/cmd/enroll_cmd.go
+++ b/internal/pkg/agent/cmd/enroll_cmd.go
@@ -452,10 +452,10 @@ func (c *enrollCmd) prepareFleetTLS() error {
 }

 func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
-       err := c.daemonReload(ctx)
-       if err == nil {
-               return nil
-       }
+       err := errors.New("this is a test error")
+       // if err == nil {
+       //      return nil
+       // }
        if errors.Is(err, context.DeadlineExceeded) ||
                errors.Is(err, context.Canceled) {
                return fmt.Errorf("could not reload daemon: %w", err)
@@ -470,9 +470,7 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
                c.log.Info("Retrying to restart...")

                err = c.daemonReload(ctx)
-               if err == nil {
-                       return nil
-               }
+
                if errors.Is(err, context.DeadlineExceeded) ||
                        errors.Is(err, context.Canceled) {
                        return fmt.Errorf("could not reload daemon after %d retries: %w",
(END)
```

I got the following output. Note that the existing implementation of the progress bar seems to buffer all logs until the step completes, so I didn't see any output until it failed completely and then everything was dumped out.

```
❯ sudo ./elastic-agent install --url=https://574feabc9fff4d4fbc4d6994e7f0d556.fleet.eastus2.staging.azure.foundit.no:443 --enrollment-token=d0tZZUFvd0J2M1UxSW91VlRDX3c6SFVTRlpJWXBTY200VXM3RlRlT1VZQQ==
[  ==] Service Started  [13s] Elastic Agent successfully installed, starting enrollment.
[==  ] Waiting For Enroll...  [14s] {"log.level":"info","@timestamp":"2023-11-28T15:27:18.441-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":497},"message":"Starting enrollment to URL: https://574feabc9fff4d4fbc4d6994e7f0d556.fleet.eastus2.staging.azure.foundit.no:443/","ecs.version":"1.6.0"}
[==  ] Waiting For Enroll...  [36s] {"log.level":"info","@timestamp":"2023-11-28T15:27:40.866-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":470},"message":"Retrying to restart...","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [1m16s] {"log.level":"info","@timestamp":"2023-11-28T15:28:20.870-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":470},"message":"Retrying to restart...","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [2m16s] {"log.level":"info","@timestamp":"2023-11-28T15:29:20.873-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":470},"message":"Retrying to restart...","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [3m16s] {"log.level":"info","@timestamp":"2023-11-28T15:30:20.886-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":470},"message":"Retrying to restart...","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [4m16s] {"log.level":"info","@timestamp":"2023-11-28T15:31:20.887-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":470},"message":"Retrying to restart...","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [4m16s] {"log.level":"error","@timestamp":"2023-11-28T15:31:20.895-0500","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":283},"message":"Elastic Agent might not be running; unable to trigger restart: could not reload agent's daemon, all retries failed. Last error: %!w(<nil>)","ecs.version":"1.6.0"}
Something went wrong while enrolling the Elastic Agent: could not reload agent's daemon, all retries failed. Last error: %!w(<nil>)
Error: could not reload agent daemon, unable to trigger restart: could not reload agent's daemon, all retries failed. Last error: %!w(<nil>)
[==  ] Uninstalled  [5m17s] Error: enroll command failed for unknown reason: exit status 1
```

There are several problems here:

1. As already noted, we don't actually log the error we get on retry attempts: `"message":"Retrying to restart..."`.
2. The time between retries is too long. The exponential backoff is almost pointless because it maxes out after two iterations. The retry timestamps are at 14s, 36s, 1m16s, 2m16s, 3m16s, 4m16s. So we spend 4 minutes retrying but only retry 5 times. I think 4 minutes is too long to wait before giving up here, we could cut this in half to 2 minutes. I'm also not sure the exponential backoff is doing anything for us when trying to restart the agent, we could just use a constant backoff retrying every 1s with better results I think.
3. The error formatter in `could not reload agent's daemon, all retries failed. Last error: %!w(<nil>` is wrong.
4. `[==  ] Uninstalled  [5m17s] Error: enroll command failed for unknown reason: exit status 1` is wrong, we know the reason but say it is unknown.

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -216,16 +228,24 @@ func TestEnroll(t *testing.T) {
...
-
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+			defer cancel()
+			err = cmd.Execute(ctx, streams)
+			if err != nil &&
+				// There is no agent running, therefore nothing to be restarted.
+				// However, this will cause the Enroll command to return an error
+				// which we'll ignore here.
+				!strings.Contains(err.Error(),
+					"could not reload agent daemon, unable to trigger restart") {
```
AndersonQ, line 239: [Blocker]
As there is the `--skip-daemon-reload` it'd be better to use it here instead of excluding some errors.
I added this exclusion because on my PR there was no way to skip the daemon reload

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -275,16 +295,24 @@ func TestEnroll(t *testing.T) {
...
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+			defer cancel()
+			err = cmd.Execute(ctx, streams)
+
+			if err != nil &&
+				// There is no agent running, therefore nothing to be restarted.
+				// However, this will cause the Enroll command to return an error
+				// which we'll ignore here.
+				!strings.Contains(err.Error(),
+					"could not reload agent daemon, unable to trigger restart") {
```
AndersonQ, line 307: [Blocker]
As there is the `--skip-daemon-reload` it'd be better to use it here instead of excluding some errors.
I added this exclusion because on my PR there was no way to skip the daemon reload

testing/integration/container_cmd_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Log("Container command output:")
+		t.Log(string(output))
+		t.FailNow()
+	}
+
+	require.Eventually(t, func() bool {
+		healthy, err := agentFixture.IsHealthy(ctx)
+		if err != nil {
+			t.Logf("error checking agent health, retrying soon. Err: %s", err)
+		}
```
AndersonQ, line 97: [Suggestion]
Use `Eventuallyf` and add the error to the message instead of logging every single error

AndersonQ (changes requested):

testing/integration/container_cmd_test.go
belimawr: I see two reasons not to move this message to the `Eventually` call:
1. It's a message that will be logged on every interaction, so we can track what is happening in the test.
2. The `msgAndArgs` `Eventuallyf` takes get evaluated when the function is called, if any variable changes when the `condition` function is called, it won't be reflected in the final error message.

internal/pkg/agent/cmd/enroll_cmd_test.go
```diff
@@ -209,23 +216,28 @@ func TestEnroll(t *testing.T) {
...
+				t.Fatalf("enrrol coms returned and unexpected error: %v", err)
+			}
 
-			config, err := readConfig(store.Content)
+			assert.True(t, store.Called)
 
-			require.NoError(t, err)
-			require.Equal(t, "my-access-api-key", config.AccessAPIKey)
-			require.Equal(t, host, config.Client.Host)
+			config, err := readConfig(store.Content)
```
AndersonQ, line 236: [Blocker]
you missed the `require.NoErrorf(t, err, "some relevant message")

testing/integration/container_cmd_test.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		t.Fatalf("could not get Fleet URL: %s", err)
+	}
+
+	cmd, err := agentFixture.PrepareAgentCommand(ctx, []string{"container"})
+	cmd.Env = append(os.Environ(), []string{
+		"FLEET_ENROLL=1",
+		"FLEET_URL=" + fleetURL,
+		"FLEET_ENROLLMENT_TOKEN=" + enrollmentToken.APIKey,
+	}...)
```
AndersonQ, line 82: why did you create an array here?
```suggestion
	cmd.Env = append(os.Environ(),
		"FLEET_ENROLL=1",
		"FLEET_URL=" + fleetURL,
		"FLEET_ENROLLMENT_TOKEN=" + enrollmentToken.APIKey)
```

AndersonQ (changes requested):
you missed a check on one of the tests, but apart from that all good

belimawr:
@cmacknz 

> The time between retries is too long. The exponential backoff is almost pointless because it maxes out after two iterations. The retry timestamps are at 14s, 36s, 1m16s, 2m16s, 3m16s, 4m16s. So we spend 4 minutes retrying but only retry 5 times. I think 4 minutes is too long to wait before giving up here, we could cut this in half to 2 minutes. I'm also not sure the exponential backoff is doing anything for us when trying to restart the agent, we could just use a constant backoff retrying every 1s with better results I think.

I see your point there, I reduced the time so iterations are faster now. How many times do you think we should retry?

I don't know the details of the shutdown process from the Elastic-Agent so I'm really not sure what makes sense here. If there is any sort of communication with Fleet-Server or even waiting Beats finishing to flush some data, it might make sense to have longer wait times in case ES is overloaded.

> `[==  ] Uninstalled  [5m17s] Error: enroll command failed for unknown reason: exit status 1` is wrong, we know the reason but say it is unknown.

That comes from a totally different bit of the codebase, the bit that logs it is waiting for an external process to finish, hence it does not know the reason of the failure.
https://github.com/elastic/elastic-agent/blob/e3c0695757b3799335214ae20af2e847c0d8cc00/internal/pkg/agent/cmd/install.go#L256-L262

testing/integration/logs_ingestion_test.go
```diff
@@ -126,40 +126,15 @@ func testMonitoringLogsAreShipped(
...
-		})
-	})
-	t.Logf("error logs: Got %d documents", len(docs.Hits.Hits))
-	for _, doc := range docs.Hits.Hits {
-		t.Logf("%#v", doc.Source)
-	}
-	require.Empty(t, docs.Hits.Hits)
-
-	// Stage 4: Make sure we have message confirming central management is running
+	// Stage 3: Make sure we have message confirming central management is running
```
belimawr, line 129: I'll remove this change once https://github.com/elastic/elastic-agent/pull/3917 is merged. I kept it here so the integration tests will, hopefully, pass.

belimawr:
Folks, I've resolved all comments, some I implemented what was asked, others I explained why it is implemented the way it is.

Once all tests are passing I'll ask for a re-review.

cmacknz:
> How many times do you think we should retry?

In a situation where the control socket isn't available yet, most likely it will become available fairly quickly once the agent creates it. The problem is we previously only waited once, when really we probably needed to wait a few seconds.

I would say retrying every 1s is reasonable for 120 total retries over 2 minutes.

> I don't know the details of the shutdown process from the Elastic-Agent so I'm really not sure what makes sense here. If there is any sort of communication with Fleet-Server or even waiting Beats finishing to flush some data, it might make sense to have longer wait times in case ES is overloaded.

Does this retry logic actually impact the shutdown process? This wasn't something I was expecting to have to consider.

AndersonQ:
buildkite test this

AndersonQ (approved):

jlind23:
Sonarqube is failing due to code coverage below 40%, force merging it.

Issue #3594, opened 2023-10-12
Kubernetes secret provider: Caching it to avoid performance issues


constanca-m:
This issue was first mentioned in https://github.com/elastic/elastic-agent/issues/3442.

When we use a kubernetes secret provider, we need to get its result every time:

https://github.com/elastic/elastic-agent/blob/69cc86060ca4b50f701f037e169b7bc5f48f4a81/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L80

This is not ideal, because we could be just fetching the result one time and using it for the future, instead of repeating the same steps every time. 

We need to implement a way to reduce the amount of requests to the API server for this by caching the result.

constanca-m:
Caching the result by default is not as straightforward as a better solution as it seemed.

To cache the results we need to find a way to update them. There are several ways to do that, but most ways (TTL, for example) might lead to incorrect results if the cache is not updated in time. One way that seems efficient is to use watchers, but that will lead to an impact in memory. We have seen this in other cases, like `add_resource_metadata` (issue for that can be found [here](https://github.com/elastic/beats/issues/36858)).

So even though there is a way to cache the results, we have two solutions that don't seem ideal:

1. Make a request to the API Server every time we need a secret (the current one).
2. Use a cache and have watchers so we know that the cache always had the correct value for the secret.

Now we need to make sure if it is less bad to use option 1 or 2.

cmacknz:
> One way that seems efficient is to use watchers, but that will lead to an impact in memory. We have seen this in other cases, like add_resource_metadata (issue for that can be found https://github.com/elastic/beats/issues/36858).

Do we know why the watchers in that case increased memory, and can confirm for sure it will have the same effect here? If we do know why, can we fix it? Watching the secret keys seems like it is the ideal solution here.

There are an increasing number of support cases where we are making problematic amounts of calls to the k8s API so trying to reduce them seems better in the long term.

constanca-m:
> Do we know why the watchers in that case increased memory, and can confirm for sure it will have the same effect here? If we do know why, can we fix it? Watching the secret keys seems like it is the ideal solution here.

We some SDHs related to that, and an [issue](https://github.com/elastic/beats/issues/36858) is well. Since we would be taking the same approach for this, it seems sure that memory would be affected.

@axw Could you please share your input? Should we try to go for this solution and see the results? If they are badly received (many SDHs about memory), maybe we undo this decision?

axw:
[...]

PR #3822, opened 2023-11-27
[Kubernetes secret provider] Add cache for the secrets

constanca-m:
## What does this PR do?

Currently, we make a request to the API Server every time we need to get the value of a secret. This issue is better explained [here](https://github.com/elastic/elastic-agent/issues/3594). Discussion also present in the issue.

With this PR, we use a map to store the secrets like this:

1. Once the kubernetes secret provider starts running, we launch a go routine that will update the cache every TTL minutes. [TTL](https://github.com/constanca-m/elastic-agent/blob/dadc79b3d338bd2f4b37e3488c453d145450e436/internal/pkg/composable/providers/kubernetessecrets/config.go#L14) is a new option in the configuration of the provider. By default it is set to 10 minutes.
2. To update the cache, we range over every secret stored there, and make a new request to the API Server to obtain the current value.

The secrets are accessed from the outside through the function [Fetch](https://github.com/constanca-m/elastic-agent/blob/dadc79b3d338bd2f4b37e3488c453d145450e436/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L59). If we try to retrieve a secret that is not present in our map, then we make a request to the API Server to obtain its value and store it in the map. This function never causes an update on the secret values.

The only difference from this approach to the current one is that we know store the values of the secrets and secret values.

**Warning**:  Sometimes the secret values are not updated in time and we can be using incorrect ones until the update happens. This is also happening now.

## Why is it important?

To stop overwhelming the API Server with secret requests.

## Checklist

1. Clone this branch.
2. Follow [these steps](https://github.com/constanca-m/elastic-agent/blob/kubernetes-cache-secrets/README.md#testing-elastic-agent-on-kubernetes) from the README file.

## Related issues

- Relates to https://github.com/elastic/elastic-agent/issues/3594

## Screenshots

There should not be a change in behavior apart from a decrease in calls to the API Server. Everything else works as expected/before:

image

### DIFF
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -15,11 +15,11 @@ type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
 
-	TTLUpdate time.Duration `config:"ttl_update"`
-	TTLDelete time.Duration `config:"ttl_delete"`
+	RefreshInterval time.Duration `config:"refresh_interval"`
+	TTLDelete       time.Duration `config:"ttl_delete"`
 }
 
 func (c *Config) InitDefaults() {
-	c.TTLUpdate = 60 * time.Second
+	c.RefreshInterval = 60 * time.Second
 	c.TTLDelete = 1 * time.Hour
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -15,5 +15,5 @@ type Config struct {
 }
 
 func (c *Config) InitDefaults() {
-	c.TTL = 10
+	c.TTL = 1
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -10,4 +10,10 @@ import "github.com/elastic/elastic-agent-autodiscover/kubernetes"
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	TTL int `config:"ttl"`
+}
+
+func (c *Config) InitDefaults() {
+	c.TTL = 10
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -11,9 +11,11 @@ type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
 
-	TTL int `config:"ttl"`
+	TTL string `config:"ttl"`
 }
 
+var defaultTTL = "60s"
+
 func (c *Config) InitDefaults() {
-	c.TTL = 1
+	c.TTL = defaultTTL
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -17,9 +17,11 @@ type Config struct {
 
 	RefreshInterval time.Duration `config:"refresh_interval"`
 	TTLDelete       time.Duration `config:"ttl_delete"`
+	Timeout         time.Duration `config:"timeout"`
 }
 
 func (c *Config) InitDefaults() {
 	c.RefreshInterval = 60 * time.Second
 	c.TTLDelete = 1 * time.Hour
+	c.Timeout = 5 * time.Second
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -4,18 +4,22 @@
 
 package kubernetessecrets
 
-import "github.com/elastic/elastic-agent-autodiscover/kubernetes"
+import (
+	"time"
+
+	"github.com/elastic/elastic-agent-autodiscover/kubernetes"
+)
 
 // Config for kubernetes provider
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
 
-	TTL string `config:"ttl"`
+	TTLUpdate time.Duration `config:"ttl_update"`
+	TTLDelete time.Duration `config:"ttl_delete"`
 }
 
-var defaultTTL = "60s"
-
 func (c *Config) InitDefaults() {
-	c.TTL = defaultTTL
+	c.TTLUpdate = 60 * time.Second
+	c.TTLDelete = 1 * time.Hour
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -17,11 +17,13 @@ type Config struct {
 
 	RefreshInterval time.Duration `config:"refresh_interval"`
 	TTLDelete       time.Duration `config:"ttl_delete"`
-	Timeout         time.Duration `config:"timeout"`
+	RequestTimeout  time.Duration `config:"request_timeout"`
+	DisableCache    bool          `config:"disable_cache"`
 }
 
 func (c *Config) InitDefaults() {
 	c.RefreshInterval = 60 * time.Second
 	c.TTLDelete = 1 * time.Hour
-	c.Timeout = 5 * time.Second
+	c.RequestTimeout = 5 * time.Second
+	c.DisableCache = false
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -16,9 +16,9 @@ type Config struct {
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
 
 	RefreshInterval time.Duration `config:"cache_refresh_interval"`
-	TTLDelete       time.Duration `config:"cache_ttl_delete"`
+	TTLDelete       time.Duration `config:"cache_ttl"`
 	RequestTimeout  time.Duration `config:"cache_request_timeout"`
-	DisableCache    bool          `config:"cache_disable_cache"`
+	DisableCache    bool          `config:"cache_disable"`
 }
 
 func (c *Config) InitDefaults() {
--- a/internal/pkg/composable/providers/kubernetessecrets/config.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/config.go
@@ -15,10 +15,10 @@ type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
 
-	RefreshInterval time.Duration `config:"refresh_interval"`
-	TTLDelete       time.Duration `config:"ttl_delete"`
-	RequestTimeout  time.Duration `config:"request_timeout"`
-	DisableCache    bool          `config:"disable_cache"`
+	RefreshInterval time.Duration `config:"cache_refresh_interval"`
+	TTLDelete       time.Duration `config:"cache_ttl_delete"`
+	RequestTimeout  time.Duration `config:"cache_request_timeout"`
+	DisableCache    bool          `config:"cache_disable_cache"`
 }
 
 func (c *Config) InitDefaults() {
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -91,25 +91,14 @@ func getK8sClient(kubeconfig string, opt kubernetes.KubeClientOptions) (k8sclien
 
 // Update the secrets in the cache every TTL minutes
 func (p *contextProviderK8sSecrets) updateSecrets(ctx context.Context) {
-	d, err := time.ParseDuration(p.config.TTL)
-	if err != nil {
-		p.logger.Debugf(
-			"TTL %v is not valid: %v. Default of %v will be used.",
-			p.config.TTL,
-			err,
-			defaultTTL,
-		)
-		d, _ = time.ParseDuration(defaultTTL)
-	}
-	timer := time.NewTimer(d)
-
+	timer := time.NewTimer(p.config.TTLUpdate)
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-timer.C:
 			p.updateCache()
-			timer.Reset(d)
+			timer.Reset(p.config.TTLUpdate)
 		}
 	}
 }
@@ -125,8 +114,8 @@ func (p *contextProviderK8sSecrets) updateCache() {
 		} else {
 			// if the secret has not been accessed in over 1h, delete it
 			// to avoid keeping secrets in cache that are no longer in use
-			diff := time.Since(data.lastAccess).Hours()
-			if diff > 1 {
+			diff := time.Since(data.lastAccess)
+			if diff > p.config.TTLDelete {
 				delete(p.secretsCache, name)
 			} else {
 				data.value = newValue
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -140,12 +140,13 @@ func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
 		}
 	}
 
+	var pass string
 	p.secretsCacheMx.Lock()
 	data, ok := p.secretsCache[key]
 	data.lastAccess = time.Now()
+	pass = data.value
 	p.secretsCacheMx.Unlock()
-
-	return data.value, ok
+	return pass, ok
 }
 
 func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -6,7 +6,6 @@ package kubernetessecrets
 
 import (
 	"context"
-	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -129,10 +128,10 @@ func (p *contextProviderK8sSecrets) updateCache() {
 }
 
 func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
-	p.secretsCacheMx.RLock()
-	_, ok := p.secretsCache[key]
-	p.secretsCacheMx.RUnlock()
+	p.secretsCacheMx.Lock()
+	defer p.secretsCacheMx.Unlock()
 
+	_, ok := p.secretsCache[key]
 	// if value is still not present in cache, it is possible we haven't tried to fetch it yet
 	if !ok {
 		data, ok := p.addToCache(key)
@@ -142,16 +141,14 @@ func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
 		}
 	}
 
-	var pass string
-	p.secretsCacheMx.Lock()
 	data, ok := p.secretsCache[key]
 	data.lastAccess = time.Now()
-	fmt.Println(data.lastAccess)
-	pass = data.value
-	p.secretsCacheMx.Unlock()
+	pass := data.value
+
 	return pass, ok
 }
 
+// This is only called by getFromCache function, which is already locking the cache.
 func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
 	// Make sure the key has the expected format "kubernetes_secrets.somenamespace.somesecret.value"
 	tokens := strings.Split(key, ".")
@@ -176,9 +173,7 @@ func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
 		value: value,
 	}
 	if ok {
-		p.secretsCacheMx.Lock()
 		p.secretsCache[key] = &data
-		p.secretsCacheMx.Unlock()
 	}
 	return data, ok
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -104,24 +104,28 @@ func (p *contextProviderK8sSecrets) updateSecrets(ctx context.Context) {
 }
 
 func (p *contextProviderK8sSecrets) updateCache() {
-	p.secretsCacheMx.Lock()
 	// deleting entries does not free the memory, so we need to create a new map
 	// to place the secrets we want to keep
 	cacheTmp := make(map[string]*secretsData)
+
+	p.secretsCacheMx.RLock()
 	for name, data := range p.secretsCache {
 		diff := time.Since(data.lastAccess)
-		if diff > p.config.TTLDelete {
-			delete(p.secretsCache, name)
-		} else {
-			newValue, ok := p.fetchSecret(name)
-			if !ok {
-				delete(p.secretsCache, name)
-			} else {
-				data.value = newValue
-				cacheTmp[name] = data
+		if diff < p.config.TTLDelete {
+			value, ok := p.fetchSecretWithTimeout(name)
+			if ok {
+				newData := &secretsData{
+					value:      value,
+					lastAccess: data.lastAccess,
+				}
+				cacheTmp[name] = newData
 			}
+
 		}
 	}
+	p.secretsCacheMx.RUnlock()
+
+	p.secretsCacheMx.Lock()
 	p.secretsCache = cacheTmp
 	p.secretsCacheMx.Unlock()
 }
@@ -168,7 +172,7 @@ func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
 		}, false
 	}
 
-	value, ok := p.fetchSecret(key)
+	value, ok := p.fetchSecretWithTimeout(key)
 	data := secretsData{
 		value: value,
 	}
@@ -180,12 +184,34 @@ func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
 	return data, ok
 }
 
-func (p *contextProviderK8sSecrets) fetchSecret(key string) (string, bool) {
+type Result struct {
+	value string
+	ok    bool
+}
+
+func (p *contextProviderK8sSecrets) fetchSecretWithTimeout(key string) (string, bool) {
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), p.config.Timeout)
+	defer cancel()
+
+	resultCh := make(chan Result, 1)
+	go p.fetchSecret(ctxTimeout, key, resultCh)
+
+	select {
+	case <-ctxTimeout.Done():
+		p.logger.Errorf("Could not retrieve value for key %v: %v", key, ctxTimeout.Err())
+		return "", false
+	case result := <-resultCh:
+		return result.value, result.ok
+	}
+}
+
+func (p *contextProviderK8sSecrets) fetchSecret(context context.Context, key string, resultCh chan Result) {
 	p.clientMx.Lock()
 	client := p.client
 	p.clientMx.Unlock()
 	if client == nil {
-		return "", false
+		resultCh <- Result{value: "", ok: false}
+		return
 	}
 
 	tokens := strings.Split(key, ".")
@@ -199,17 +225,19 @@ func (p *contextProviderK8sSecrets) fetchSecret(key string) (string, bool) {
 	secretVar := tokens[3]
 
 	secretInterface := client.CoreV1().Secrets(ns)
-	ctx := context.TODO()
-	secret, err := secretInterface.Get(ctx, secretName, metav1.GetOptions{})
+	secret, err := secretInterface.Get(context, secretName, metav1.GetOptions{})
+
 	if err != nil {
 		p.logger.Errorf("Could not retrieve secret from k8s API: %v", err)
-		return "", false
+		resultCh <- Result{value: "", ok: false}
+		return
 	}
 	if _, ok := secret.Data[secretVar]; !ok {
 		p.logger.Errorf("Could not retrieve value %v for secret %v", secretVar, secretName)
-		return "", false
+		resultCh <- Result{value: "", ok: false}
+		return
 	}
-	secretString := secret.Data[secretVar]
 
-	return string(secretString), true
+	secretString := secret.Data[secretVar]
+	resultCh <- Result{value: string(secretString), ok: true}
 }
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -106,16 +106,12 @@ func (p *contextProviderK8sSecrets) updateSecrets(ctx context.Context) {
 func (p *contextProviderK8sSecrets) updateCache() {
 	p.secretsCacheMx.Lock()
 	for name, data := range p.secretsCache {
-		newValue, ok := p.fetchSecret(name)
-
-		// remove the secret from the cache
-		if !ok {
+		diff := time.Since(data.lastAccess)
+		if diff > p.config.TTLDelete {
 			delete(p.secretsCache, name)
 		} else {
-			// if the secret has not been accessed in over 1h, delete it
-			// to avoid keeping secrets in cache that are no longer in use
-			diff := time.Since(data.lastAccess)
-			if diff > p.config.TTLDelete {
+			newValue, ok := p.fetchSecret(name)
+			if !ok {
 				delete(p.secretsCache, name)
 			} else {
 				data.value = newValue
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -125,7 +125,7 @@ func (p *contextProviderK8sSecrets) updateCache() {
 		} else {
 			// if the secret has not been accessed in over 1h, delete it
 			// to avoid keeping secrets in cache that are no longer in use
-			diff := time.Now().Sub(data.lastAccess).Hours()
+			diff := time.Since(data.lastAccess).Hours()
 			if diff > 1 {
 				delete(p.secretsCache, name)
 			} else {
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -102,7 +102,7 @@ func (p *contextProviderK8sSecrets) updateSecrets(ctx context.Context) {
 
 func (p *contextProviderK8sSecrets) updateCache() {
 	p.secretsCacheMx.Lock()
-	for name, _ := range p.secretsCache {
+	for name := range p.secretsCache {
 		newValue, ok := p.fetchSecret(name)
 
 		// remove the secret from the cache
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -6,6 +6,7 @@ package kubernetessecrets
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -108,7 +109,7 @@ func (p *contextProviderK8sSecrets) updateCache() {
 	// to place the secrets we want to keep
 	cacheTmp := make(map[string]*secretsData)
 
-	p.secretsCacheMx.RLock()
+	p.secretsCacheMx.Lock()
 	for name, data := range p.secretsCache {
 		diff := time.Since(data.lastAccess)
 		if diff < p.config.TTLDelete {
@@ -123,9 +124,6 @@ func (p *contextProviderK8sSecrets) updateCache() {
 
 		}
 	}
-	p.secretsCacheMx.RUnlock()
-
-	p.secretsCacheMx.Lock()
 	p.secretsCache = cacheTmp
 	p.secretsCacheMx.Unlock()
 }
@@ -148,6 +146,7 @@ func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
 	p.secretsCacheMx.Lock()
 	data, ok := p.secretsCache[key]
 	data.lastAccess = time.Now()
+	fmt.Println(data.lastAccess)
 	pass = data.value
 	p.secretsCacheMx.Unlock()
 	return pass, ok
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -149,6 +149,24 @@ func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
 }
 
 func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
+	// Make sure the key has the expected format "kubernetes_secrets.somenamespace.somesecret.value"
+	tokens := strings.Split(key, ".")
+	if len(tokens) > 0 && tokens[0] != "kubernetes_secrets" {
+		return secretsData{
+			value: "",
+		}, false
+	}
+	if len(tokens) != 4 {
+		p.logger.Debugf(
+			"not valid secret key: %v. Secrets should be of the following format %v",
+			key,
+			"kubernetes_secrets.somenamespace.somesecret.value",
+		)
+		return secretsData{
+			value: "",
+		}, false
+	}
+
 	value, ok := p.fetchSecret(key)
 	data := secretsData{
 		value: value,
@@ -169,19 +187,12 @@ func (p *contextProviderK8sSecrets) fetchSecret(key string) (string, bool) {
 		return "", false
 	}
 
-	// key = "kubernetes_secrets.somenamespace.somesecret.value"
 	tokens := strings.Split(key, ".")
-	if len(tokens) > 0 && tokens[0] != "kubernetes_secrets" {
-		return "", false
-	}
-	if len(tokens) != 4 {
-		p.logger.Debugf(
-			"not valid secret key: %v. Secrets should be of the following format %v",
-			key,
-			"kubernetes_secrets.somenamespace.somesecret.value",
-		)
-		return "", false
-	}
+	// key has the format "kubernetes_secrets.somenamespace.somesecret.value"
+	// This function is only called from:
+	// - addToCache, where we already validated that the key has the right format.
+	// - updateCache, where the results are only added to the cache through addToCache
+	// Because of this we no longer need to validate the key
 	ns := tokens[1]
 	secretName := tokens[2]
 	secretVar := tokens[3]
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"strings"
 	"sync"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "k8s.io/client-go/kubernetes"
@@ -33,6 +34,9 @@ type contextProviderK8sSecrets struct {
 
 	clientMx sync.Mutex
 	client   k8sclient.Interface
+
+	secretsCacheMx sync.Mutex
+	secretsCache   map[string]string
 }
 
 // ContextProviderBuilder builds the context provider.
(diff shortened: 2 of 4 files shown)


internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -10,4 +10,10 @@ import "github.com/elastic/elastic-agent-autodiscover/kubernetes"
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	TTL int `config:"ttl"`
+}
+
+func (c *Config) InitDefaults() {
+	c.TTL = 10
```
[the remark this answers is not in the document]
constanca-m: I honestly don't know what should be best. But we were updating on bigger intervals than 1. I can change to 1 as default, and then the user can update it as they wish.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +50,102 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+			p.updateCache()
+			timer.Reset(d)
+		}
+	}
+}
+
+func (p *contextProviderK8sSecrets) updateCache() {
+	p.secretsCacheMx.Lock()
+	for name := range p.secretsCache {
+		newValue, ok := p.fetchSecret(name)
```
gizas, line 106: Shall we first check the value of each secret and if it is not the same then update it?

internal/pkg/composable/providers/kubernetessecrets/config.go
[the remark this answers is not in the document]
gizas: Can you please check with 1min to see if we have less api calls?  (now rememberd saying 3-5-7 minutes)
And we can decide then

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: I believe that getting or putting a new value into a map has the same complexity, so it does not make a difference. Would it make it more clear for readability? If so, I can add an `if` statement to check it before updating.

gizas: Ok leave it. Ignore that comment

gizas (approved):

internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -10,4 +10,10 @@ import "github.com/elastic/elastic-agent-autodiscover/kubernetes"
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	TTL int `config:"ttl"`
```
blakerouse, line 14: You wrap this value with `time.Duration` when it is being used. Why not just use `time.Duration` in the configuration? That would allow the TTL to be set using `10s` instead of it being set to a specific unit.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +50,102 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+	return comm.Err()
+}
+
+func getK8sClient(kubeconfig string, opt kubernetes.KubeClientOptions) (k8sclient.Interface, error) {
+	return kubernetes.GetKubernetesClient(kubeconfig, opt)
+}
+
+// Update the secrets in the cache every TTL minutes
+func (p *contextProviderK8sSecrets) updateSecrets(ctx context.Context) {
+	d := time.Duration(p.config.TTL) * time.Minute
```
blakerouse, line 89: See my comment about about changing `p.config.TTL` to a `time.Duration` in the structure.

blakerouse (changes requested):
The biggest issue I see in this PR is how does a key get removed from the cache if it is no longer referenced from the policy? Let say the policy references `kubernetes_secrets.somenamespace.somesecret.value1` and then changes to `kubernetes_secrets.somenamespace.somesecret.value2`. This change will keep caching `kubernetes_secrets.somenamespace.somesecret.value1` forever.

To solve the problem you should add a last accessed time per cached key, and during each update cycle and based on an interval you should remove keys that have not been accessed. That will solve the issue I described above.

constanca-m:
> > To solve the problem you should add a last accessed time per cached key
> 
> @blakerouse thanks for this . Makes sense.

And should we give that option also to the user or should we enforce that time ourselves @gizas ?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: Thanks, I updated the configuration according to this

internal/pkg/composable/providers/kubernetessecrets/config.go
constanca-m: > Why not just use time.Duration in the configuration? 

Like it is now as a string? Or use time duration exactly in this place?

internal/pkg/composable/providers/kubernetessecrets/config.go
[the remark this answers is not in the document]
constanca-m: When I checked, we were having around 24 calls at the same time because we were getting calls from every resource. Now to obtain the same number of calls, even with TTL set to 1, we would need 24 minutes.

internal/pkg/composable/providers/kubernetessecrets/config.go
constanca-m: I left it as a string, because I don't see the place to set it to the default in case the user sets it wrong :thinking:

internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -10,4 +10,12 @@ import "github.com/elastic/elastic-agent-autodiscover/kubernetes"
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	TTL string `config:"ttl"`
```
blakerouse, line 14: Why are you making this a string? Make it a `time.Duration`. Elastic Agent uses https://github.com/elastic/go-ucfg which supports `time.Duration` as a native configuration type.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +55,131 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+}
+
+func getK8sClient(kubeconfig string, opt kubernetes.KubeClientOptions) (k8sclient.Interface, error) {
+	return kubernetes.GetKubernetesClient(kubeconfig, opt)
+}
+
+// Update the secrets in the cache every TTL minutes
+func (p *contextProviderK8sSecrets) updateSecrets(ctx context.Context) {
+	d, err := time.ParseDuration(p.config.TTL)
+	if err != nil {
```
blakerouse, line 95: You won't need to do this if you use a `time.Duration` in the structure. `go-ucfg` with Elastic Agent will handle this for you and report the errors when it parses the whole configuration.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +55,131 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+		newValue, ok := p.fetchSecret(name)
+
+		// remove the secret from the cache
+		if !ok {
+			delete(p.secretsCache, name)
+		} else {
+			// if the secret has not been accessed in over 1h, delete it
+			// to avoid keeping secrets in cache that are no longer in use
+			diff := time.Since(data.lastAccess).Hours()
+			if diff > 1 {
```
blakerouse, line 129: If you make the TTL configurable, why not make this configuration as a `time.Duration`?

blakerouse (changes requested):

blakerouse:
This is also missing unit tests coverage for any of these additions, that needs to be added as well.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: And if it fails, does it go back to the defaults?

constanca-m:
> This is also missing unit tests coverage for any of these additions, that needs to be added as well.

I just added the unit test to check the results inside the cache @blakerouse

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
```diff
@@ -138,3 +138,110 @@ func Test_K8sSecretsProvider_FetchWrongSecret(t *testing.T) {
...
+		},
+		Data: map[string][]byte{
+			"secret_value": []byte(newPass),
+		},
+	}
+	_, err = client.CoreV1().Secrets(ns).Update(context.Background(), secret, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	// wait for ttl update
+	<-time.After(ttlUpdate)
```
blakerouse, line 236: Relying on time here leads to random test failures because its possible that its executed out of order when the goroutines are scheduled. If your going to wait then you should at least double the wait time to ensure that it happens.

Another more full proof solution would be to use `assert.Eventually` after the wait to allow some chance that it has not be updated yet.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
```diff
@@ -138,3 +138,110 @@ func Test_K8sSecretsProvider_FetchWrongSecret(t *testing.T) {
...
+	val, found = fp.Fetch("kubernetes_secrets.test_namespace.testing_secret.secret_value")
+	assert.True(t, found)
+	assert.Equal(t, val, newPass)
+
+	// After TTL delete, secret should no longer be found in cache since it was never
+	// fetched during that time
+	<-time.After(ttlDelete)
```
blakerouse, line 243: Same as above on this wait.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +55,116 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+}
+
+func (p *contextProviderK8sSecrets) updateCache() {
+	p.secretsCacheMx.Lock()
+	for name, data := range p.secretsCache {
+		diff := time.Since(data.lastAccess)
+		if diff > p.config.TTLDelete {
+			delete(p.secretsCache, name)
+		} else {
+			newValue, ok := p.fetchSecret(name)
```
blakerouse, line 113: This is another issue that I am seeing. The lock is being held when making API calls to fetch the secret. What is the worry that this blocks for a long time holding the lock and then causing the rendering of the policy to be blocked?

Is it possible to be multiple key values in a single API call? It might be better to change this to first get all the keys that new values are needed. Then make a single API call to get all the values (if possible), and then update the cache with the values.

blakerouse (changes requested):

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +55,116 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+	data.lastAccess = time.Now()
+	p.secretsCacheMx.Unlock()
+
+	return data.value, ok
+}
+
+func (p *contextProviderK8sSecrets) addToCache(key string) (secretsData, bool) {
+	value, ok := p.fetchSecret(key)
+	data := secretsData{
+		value: value,
```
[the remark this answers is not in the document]
constanca-m: It's on purpose. `addToCache` is only used inside `getFromCache`.

https://github.com/elastic/elastic-agent/blob/9d948597bbd3448e3ff66693cf2431a89d05bb5f/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L124

`getFromCache` is the only function that updates `lastAcess`.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -33,6 +34,14 @@ type contextProviderK8sSecrets struct {
 
 	clientMx sync.Mutex
 	client   k8sclient.Interface
+
+	secretsCacheMx sync.Mutex
```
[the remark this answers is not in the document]
constanca-m: Yes, fixed.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,19 +55,116 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+		}
+	}
+}
+
+func (p *contextProviderK8sSecrets) updateCache() {
+	p.secretsCacheMx.Lock()
+	for name, data := range p.secretsCache {
+		diff := time.Since(data.lastAccess)
+		if diff > p.config.TTLDelete {
+			delete(p.secretsCache, name)
```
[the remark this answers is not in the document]
constanca-m: Fixed.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
constanca-m: I updated the code to use `assert.Eventually`.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: > Is it possible to be multiple key values in a single API call? 

It is not possible. The only way to get multiple secrets at once is to use the list function instead of get [here](https://github.com/elastic/elastic-agent/blob/b3ba2192a841d4e353ac625620a88643b87144ef/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L202). However, the options to filter the secrets we return are [these](https://github.com/kubernetes/kubernetes/blob/ad9b60e2c9ddb21e8b00cabbe27e639638a0ea88/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L322), and as we can see, there are none to filter by the name. So in the end we would be retrieving all the secrets from a namespace when there is no need.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
blakerouse: I see, thanks.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
blakerouse: Dang. Okay, I am still worried about it holding the lock the entire time it is refreshing all the secrets.

blakerouse (changes requested):

constanca-m:
Are there any blockers you still want to address @blakerouse , since you did not remove the requested changes?

blakerouse:
> Are there any blockers you still want to address @blakerouse , since you did not remove the requested changes?

Yes I am still waiting on my comment about holding the lock the entire time the code is refreshing the cache. I think that is still a big issue that needs to be looked at.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,38 +55,152 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+			p.updateCache()
+			timer.Reset(p.config.TTLUpdate)
+		}
 	}
+}
```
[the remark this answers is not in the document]
constanca-m: > General question: why do we update all secrets periodically instead of adding/refreshing entries based on access?

We cannot do that, because otherwise we would need to make a request to the API Server every time we want to access the value. That is the [issue](https://github.com/elastic/elastic-agent/issues/3594) that led to this PR.

> If we set a TTL for deletion (basically the time we consider a cache value "not stale") and in case of cache miss we fetch the specific secret value woudn't that be more efficient ?

You mean to just update the secret after TTL delete has passed? TTL delete is quite a long duration.

> it's just to understand the rationale behind having a cache where (almost) all entries get refreshed at once: do we have an idea of what the access pattern of secrets would be ?

I am not sure, this a feature request. In the issue I [linked](https://github.com/elastic/elastic-agent/issues/3594) we discuss the possible workarounds. In the future, this cache will likely be replaced by a watcher, but we need to solve [this](https://github.com/elastic/beats/issues/34717) first.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,38 +55,152 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+			return data.value, ok
+		}
+	}
+
+	var pass string
+	p.secretsCacheMx.Lock()
+	data, ok := p.secretsCache[key]
+	data.lastAccess = time.Now()
+	pass = data.value
+	p.secretsCacheMx.Unlock()
```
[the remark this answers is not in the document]
constanca-m: > Are we using a write lock here because we are modifying the lastAccess attribute of a cache entry in place ?

Yes, we are updating the value there.

internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -4,10 +4,22 @@
...
+
+	"github.com/elastic/elastic-agent-autodiscover/kubernetes"
+)
 
 // Config for kubernetes provider
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	TTLUpdate time.Duration `config:"ttl_update"`
```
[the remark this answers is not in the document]
constanca-m: Updated.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
cmacknz: Does `fetchSecrets` have a timeout? What if you never get a response? What if you do get a response, but it takes one hour?

Now that `cacheTmp` is being rebuilt is holding the lock for each `fetchSecret` call still necessary? Can `cacheTmp` be built and then assigned to `p.secretsCache` with only that one line holding the lock?

Since we are iterating over secretsCache that would likely require duplicating the keys first.

> However, the options to filter the secrets we return are these, and as we can see, there are none to filter by the name. So in the end we would be retrieving all the secrets from a namespace when there is no need.

What places more load on the k8s API? Returning a single response that is unnecessarily large or multiple small requests made rapidly one after another?

We should bias towards making sure the k8s api-server remains stable over optimizing the request size if possible.

constanca-m: > Does fetchSecrets have a timeout? What if you never get a response? What if you do get a response, but it takes one hour?

I added a timeout just now, so we will never be waiting for 1h.

> Now that cacheTmp is being rebuilt is holding the lock for each fetchSecret call still necessary?

I changed the implementation to this way:
1. Created the temporary cache.
2. Apply read lock and iterate over the entries of the secrets cache. If entry remains, then write on temporary cache.
3. Now that the iteration is over, apply writing lock to secrets cache and make it the same as the temporary one.

Is this ok this way @cmacknz ?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,189 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+		// if it was not possible to fetch the secret, return
+		if !ok {
+			return data.value, ok
+		}
 	}
+
+	var pass string
+	p.secretsCacheMx.Lock()
+	data, ok := p.secretsCache[key]
+	data.lastAccess = time.Now()
```
cmacknz, line 150: `data` is not a pointer so you are getting a copy of it, so the update to `lastAccess` does not change the `lastAccess` that was originally stored in the map.

You are missing test coverage for two consecutive `Fetch` calls of the same value to ensure that `lastAccess` updates in the expected way.

Here's a sample modification to your test showing the problem:

```diff
diff --git a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
index 3fea2dc455..a7348c40a9 100644
--- a/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
+++ b/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
@@ -212,12 +212,33 @@ func Test_K8sSecretsProvider_Check_TTL(t *testing.T) {
        val, found := fp.Fetch(key)
        assert.True(t, found)
        assert.Equal(t, val, pass)
+
        fp.secretsCacheMx.RLock()
        assert.Equal(t, len(fp.secretsCache), 1)
-       assert.NotNil(t, fp.secretsCache[key])
-       assert.NotZero(t, fp.secretsCache[key].lastAccess)
+       cachedVal1 := fp.secretsCache[key]
        fp.secretsCacheMx.RUnlock()

+       assert.NotNil(t, cachedVal1)
+       assert.NotZero(t, cachedVal1.lastAccess)
+
+       val, found = fp.Fetch(key)
+       assert.True(t, found)
+       assert.Equal(t, val, pass)
+
+       // Arbitrary delay to ensure that the value returned by time.Now() used when setting TTL is not
+       // the same between Fetch() calls.
+       time.Sleep(time.Millisecond)
+
+       fp.secretsCacheMx.RLock()
+       assert.Equal(t, len(fp.secretsCache), 1)
+       cachedVal2 := fp.secretsCache[key]
+       fp.secretsCacheMx.RUnlock()
+
+       assert.NotNil(t, cachedVal2)
+       t.Logf("Fetch 1 LastAccess: %s", cachedVal1.lastAccess)
+       t.Logf("Fetch 2 LastAccess: %s", cachedVal2.lastAccess)
+       assert.NotEqual(t, cachedVal1.lastAccess, cachedVal2.lastAccess)
+
        // Update the secret and check after TTL time, the secret value is correct
        newPass := "new-pass"
        secret = &v1.Secret{

```

```sh
❯ go test ./internal/pkg/composable/providers/kubernetessecrets/... -run Test_K8sSecretsProvider_Check_TTL

# github.com/elastic/elastic-agent/internal/pkg/composable/providers/kubernetessecrets.test
ld: warning: ignoring duplicate libraries: '-lproc'
--- FAIL: Test_K8sSecretsProvider_Check_TTL (2.22s)
    kubernetes_secrets_test.go:238: Fetch 1 LastAccess: 2023-12-05 16:35:13.749574 -0500 EST m=+0.016378918
    kubernetes_secrets_test.go:239: Fetch 2 LastAccess: 2023-12-05 16:35:13.749574 -0500 EST m=+0.016378918
    kubernetes_secrets_test.go:240:
                Error Trace:    /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go:240
                Error:          Should not be: time.Date(2023, time.December, 5, 16, 35, 13, 749574000, time.Local)
                Test:           Test_K8sSecretsProvider_Check_TTL
FAIL
FAIL    github.com/elastic/elastic-agent/internal/pkg/composable/providers/kubernetessecrets    2.754s
FAIL
```

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,189 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+}
+
+func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
+	p.secretsCacheMx.RLock()
+	_, ok := p.secretsCache[key]
+	p.secretsCacheMx.RUnlock()
+
+	// if value is still not present in cache, it is possible we haven't tried to fetch it yet
+	if !ok {
+		data, ok := p.addToCache(key)
```
cmacknz, line 140: It is possible that another goroutine acquires could call addToCache before this goroutine. That is this goroutine is preempted immediately after `RUnlock`, then another goroutine calls `addToCache`, and `addToCache` is then called again unnecessarily.

I think this is more of an optimization than a bug, since all it results in is an unnecessary call to the secrets API.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,189 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+				newData := &secretsData{
+					value:      value,
+					lastAccess: data.lastAccess,
+				}
+				cacheTmp[name] = newData
+			}
+
+		}
+	}
+	p.secretsCacheMx.RUnlock()
```
cmacknz, line 126: If this goroutine is preempted immediately after the `RUnlock` call here, and then another goroutine calls `addToCache` via a call to `Fetch` for a secret that wasn't previously in the map, then the write is lost and `p.secretsCache = cacheTmp` discards that secret.

You would need to iterate over `p.secretsCache` again while holding the write lock to check for values that were added when you released the `RLock` above to include those as well.

In theory the next call to `Fetch` would populate the missing value but you'd need to make a duplicate call to the k8s API for this.

cmacknz (commented):
Took another look, there is still at least one bug and a few cases I could spot where splitting the cache changes across separate "read lock" and "write lock" sections introduces concurrency problems. 

This is because the manipulation you are doing to the cache is no longer atomic across a single function body allowing the cache to change in the middle of the function execution when you unlock the read lock and acquire the write lock.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: Thanks @cmacknz for such a complete example.

I checked this:

> data is not a pointer so you are getting a copy of it, so the update to lastAccess does not change the lastAccess that was originally stored in the map.

And I ran the test. I added a println on the code right where the last access changes:

```
=== RUN   Test_K8sSecretsProvider_Check_TTL
2023-12-06 08:26:10.267182544 +0100 CET m=+0.016935938
2023-12-06 08:26:10.267266524 +0100 CET m=+0.017019919
    kubernetes_secrets_test.go:239: Fetch 1 LastAccess: 2023-12-06 08:26:10.267266524 +0100 CET m=+0.017019919
    kubernetes_secrets_test.go:240: Fetch 2 LastAccess: 2023-12-06 08:26:10.267266524 +0100 CET m=+0.017019919
    kubernetes_secrets_test.go:241: 
        	Error Trace:	/home/c/Documents/Git-Code/Elastic-Forks/elastic-agent/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go:241
        	Error:      	Should not be: time.Date(2023, time.December, 6, 8, 26, 10, 267266524, time.Local)
        	Test:       	Test_K8sSecretsProvider_Check_TTL
--- FAIL: Test_K8sSecretsProvider_Check_TTL (0.01s)
```

So, as you can see, the last value got updated. But since `cachedVal1` and `cachedVal2` are the same, this assertion will always fail `assert.NotEqual(t, cachedVal1.lastAccess, cachedVal2.lastAccess)`. Notice that the map contains the exact values, not copies:

```
type contextProviderK8sSecrets struct {
        ...
	secretsCache   map[string]*secretsData <---------------------- pointer here
}
```

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: I switched the reading lock to a writing lock, so this doesn't happen anymore.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
cmacknz: Yes you are correct, my example proves myself wrong I just needed to read it at the start of my day instead of the end :)

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,183 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+	value string
+	ok    bool
+}
+
+func (p *contextProviderK8sSecrets) fetchSecretWithTimeout(key string) (string, bool) {
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), p.config.Timeout)
+	defer cancel()
+
+	resultCh := make(chan Result, 1)
+	go p.fetchSecret(ctxTimeout, key, resultCh)
```
cmacknz, line 191: Why does this need to be in a goroutine if we are just going to synchronously wait for it to complete anyway?

cmacknz (commented):
One last thought, it seems like it would be straightforward to add a configuration entry that disables the cache.

That way if there is some unexpected surprise we can revert to the previous behavior without needing a release.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: You are right. I will remove the `go`

constanca-m:
>  it seems like it would be straightforward to add a configuration entry that disables the cache.

I added this option in my last commit. I also added the proper unit test for it.

blakerouse (changes requested):

internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -4,10 +4,26 @@
...
 
 // Config for kubernetes provider
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	RefreshInterval time.Duration `config:"refresh_interval"`
+	TTLDelete       time.Duration `config:"ttl_delete"`
+	RequestTimeout  time.Duration `config:"request_timeout"`
+	DisableCache    bool          `config:"disable_cache"`
```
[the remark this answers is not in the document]
constanca-m: I will add the prefix `cache_` to all of those then, and this way we will be following the same logic as we followed for kubernetes leader:
https://github.com/elastic/elastic-agent/blob/e3c0695757b3799335214ae20af2e847c0d8cc00/internal/pkg/composable/providers/kubernetesleaderelection/config.go#L14-L20

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,201 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+					value:      value,
+					lastAccess: data.lastAccess,
+				}
+				cacheTmp[name] = newData
+			}
+
+		}
+	}
+	p.secretsCache = cacheTmp
+	p.secretsCacheMx.Unlock()
```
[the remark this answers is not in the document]
gizas: @blakerouse I can understand your concern. For now the option to disable the cache has been added [0f8d86a](https://github.com/elastic/elastic-agent/pull/3822/commits/0f8d86a4955f6737e74b504e83f11a6188fcb3fb)

So in case of any problem like you describe we can revert to ption where we just get the secrret [here](https://github.com/elastic/elastic-agent/blob/0f8d86a4955f6737e74b504e83f11a6188fcb3fb/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L240) with a small lock to get the [client](https://github.com/elastic/elastic-agent/blob/0f8d86a4955f6737e74b504e83f11a6188fcb3fb/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L223)

Can we agree that Syncronisation of cache can be  done in a new story? We also investigate this on other cases for kubernetes watchers and if we have any update we can link this issue as well. 
For now I will approve this story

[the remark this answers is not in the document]
MichaelKatsoulis: I agree that holding the lock during the whole duration of update is not a good solution due to the blocking it can create to Fetch. I understand from the code that the reason to lock, is because we read all the keys and the lastaccess from the cache in the iteration. We actually write to the cache only after all secrets are updated. Why don't we keep the keys of the cache map in a slice and only lock the cache when we actually read the last access value for each key. Something like this 

```
func (p *contextProviderK8sSecrets) updateCache() {
	cacheTmp := make(map[string]*secretsData)

	p.secretsCacheMx.Lock()
	secretKeys := make([]string, len(p.secretsCache))
	i := 0
	for k := range p.secretsCache {
		secretKeys[i] = k
		i++
	}
	p.secretsCacheMx.Unlock()
	for _, name := range secretKeys {
		p.secretsCacheMx.Lock()
		lastAccess := p.secretsCache[name].lastAccess
		p.secretsCacheMx.Unlock()
		diff := time.Since(lastAccess)

		if diff < p.config.TTLDelete {
			value, ok := p.fetchSecretWithTimeout(name)
			if ok {
				newData := &secretsData{
					value:      value,
					lastAccess: lastAccess,
				}
				cacheTmp[name] = newData
			}

		}
	}
	p.secretsCacheMx.Lock()
	p.secretsCache = cacheTmp
	p.secretsCacheMx.Unlock()
}
```

internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -4,10 +4,26 @@
...
 
 // Config for kubernetes provider
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	RefreshInterval time.Duration `config:"cache_refresh_interval"`
+	TTLDelete       time.Duration `config:"cache_ttl_delete"`
+	RequestTimeout  time.Duration `config:"cache_request_timeout"`
+	DisableCache    bool          `config:"cache_disable_cache"`
```
MichaelKatsoulis, line 21:
```suggestion
	DisableCache    bool          `config:"cache_disable"`
```

internal/pkg/composable/providers/kubernetessecrets/config.go
```diff
@@ -4,10 +4,26 @@
...
+	"github.com/elastic/elastic-agent-autodiscover/kubernetes"
+)
 
 // Config for kubernetes provider
 type Config struct {
 	KubeConfig        string                       `config:"kube_config"`
 	KubeClientOptions kubernetes.KubeClientOptions `config:"kube_client_options"`
+
+	RefreshInterval time.Duration `config:"cache_refresh_interval"`
+	TTLDelete       time.Duration `config:"cache_ttl_delete"`
```
MichaelKatsoulis, line 19:
```suggestion
	TTLDelete       time.Duration `config:"cache_ttl"`
```

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
constanca-m: That has already been suggested, and @cmacknz explains in this https://github.com/elastic/elastic-agent/pull/3822#discussion_r1416314549 why that does not work. @MichaelKatsoulis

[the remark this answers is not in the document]
MichaelKatsoulis: You mean that they keys can change by fetch method which can create a new entry in the cache, and then in update due to the temp cache we will override and lose the new key? 
Because if this is the problem, it could be mitigated if we then merge the `p.secretsCache` and `cacheTmp` while holding the cache.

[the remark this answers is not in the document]
constanca-m: > You mean that they keys can change by fetch method which can create a new entry in the cache, and then in update due to the temp cache we will override and lose the new key?

Yes.

> Because if this is the problem, it could be mitigated if we then merge the p.secretsCache and cacheTmp while holding the cache.

We cannot do that. It is explained in this https://github.com/elastic/elastic-agent/pull/3822#discussion_r1406557666 why not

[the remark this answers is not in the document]
MichaelKatsoulis: In that comment it is explained that deleting a key from a map does not delete it from memory. I am saying that after the update is done and new keys are added in the temp cache we can do a merge of the temp cache and p.secretsCache and keep the keys that may have been added while we were updating. See this example:

```
p.secretsCacheMx.Lock()
newCache := mergeMaps(p.secretsCache, cacheTmp, p.config.TTLDelete)
p.secretsCache = newCache
p.secretsCacheMx.Unlock()
```

```
func mergeMaps[K comparable, V any](m1 map[K]V, m2 map[K]V, ttl time.Duration) map[K]V {
	merged := make(map[K]V)
	for key, value := range m1 {
		lastAccess := any(value).(secretsData).lastAccess
		diff := time.Since(lastAccess)
		if diff < ttl {
			merged[key] = value
		}
	}
	for key, value := range m2 {
		merged[key] = value
	}
	return merged
}
```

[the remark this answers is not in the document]
blakerouse: I agree with @MichaelKatsoulis suggestion but I believe you need to make an approvement to the mergeMaps. I believe that it is updating the lastAccess time from the `m2` map. It should most likely only overwrite the value and not the `lastAccess` timestamp.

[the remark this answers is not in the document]
MichaelKatsoulis: Yeah, in case there was a key accessed by fetch, while the update took place, then the last access of that key will have been updated. And by 
```
for key, value := range m2 {
		merged[key] = value
	}
```
the new last access will be overridden by lets say the old last access of the cache. 
Maybe using generics in this use case does not add anything and makes it more complicated as we already know the data types of the maps.
Something like this could work
```
func mergeMaps(m1, m2 map[string]*secretsData, ttl time.Duration) map[string]*secretsData {
	merged := make(map[string]*secretsData)
	for key, value := range m1 {
		lastAccess := value.lastAccess
		diff := time.Since(lastAccess)
		if diff < ttl {
			merged[key] = value
		}
	}
	for key, value := range m2 {
		if val, ok := merged[key]; ok {
			merged[key] = &secretsData{
				value:      value.value,
				lastAccess: val.lastAccess,
			}
		} else {
			merged[key] = value
		}
	}
	return merged
}
```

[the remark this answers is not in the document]
blakerouse: I don't believe you want that last else statement. If the key was removed due to ttl in the previous loop you don't want to add it back in the case that it was fetched in the cache refresh, otherwise I believe the function will work and prevent the holding of the lock for the full refresh.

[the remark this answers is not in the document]
constanca-m: I made a new commit, so now:

1. We hold only a read lock to update the secrets.
2. After that, we merge the maps.

Is it ok this way @blakerouse @MichaelKatsoulis ?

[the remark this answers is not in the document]
MichaelKatsoulis: @blakerouse you are right. I somehow thought that update adds extra keys as well but that is not the case.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,223 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+func (p *contextProviderK8sSecrets) mergeWithCurrent(updatedMap map[string]*secretsData) map[string]*secretsData {
+	merged := make(map[string]*secretsData)
+	for name, data := range p.secretsCache {
+		diff := time.Since(data.lastAccess)
+		if diff < p.config.TTLDelete {
+			merged[name] = data
+		}
+	}
+	for name, data := range updatedMap {
+		merged[name] = data
```
MichaelKatsoulis, line 129: In https://github.com/elastic/elastic-agent/pull/3822/files#r1425592648 I explain why this will override the last access of the secrets that were accessed by fetch , while the update was taking place. 
The right thing to do here would be 
```
for name, data := range updatedMap {
		if val, ok := merged[name]; ok {
			merged[name] = &secretsData{value: data.value, lastAccess: val.lastAccess}
		}
	}
```

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,223 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+	// While the cache was updated, it is possible that some secret was added through another go routine.
+	// We need to merge the updated map with the current cache map to catch the new entries and avoid
+	// loss of data.
+	p.secretsCacheMx.Lock()
+	p.secretsCache = p.mergeWithCurrent(cacheTmp)
+	p.secretsCacheMx.Unlock()
+}
+
+func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
+	p.secretsCacheMx.Lock()
```
MichaelKatsoulis, line 165: I believe now RLock and RUnlock should be used.

constanca-m: Not possible, because we are calling `addToCache` in this function, and we need that lock

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: Thanks, I updated the code to fix the mistake.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
MichaelKatsoulis: TBH I don't understand why we need to hold the lock while we are waiting to fetchSecretWithTimeout

constanca-m: We are adding a value to the cache, and then we are returning the value that is inside the cache. If we don't hold the lock there, we get a racing condition from the tests.

MichaelKatsoulis: Ok can you explain to me what is wrong with this code? I am not holding the lock while waiting for 
fetchSecretWithTimeout. 

```
func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
	p.secretsCacheMx.RLock()
	defer p.secretsCacheMx.RUnlock()

	data, ok := p.secretsCache[key]
	if ok {
		data.lastAccess = time.Now()
		pass := data.value
		return pass, ok
	}
	data, ok = p.addToCache(key)
	// if it was not possible to fetch the secret, return
	return data.value, ok
}

func (p *contextProviderK8sSecrets) addToCache(key string) (*secretsData, bool) {
	valid := p.validateKey(key)
	if !valid {
		return &secretsData{
			value: "",
		}, false
	}

	value, ok := p.fetchSecretWithTimeout(key)
	data := &secretsData{
		value: value,
	}
	if ok {
		p.secretsCacheMx.Lock()
		p.secretsCache[key] = data
		p.secretsCacheMx.Unlock()
	}
	return data, ok
}
```

MichaelKatsoulis: Yes because of the defer wich will keep the read lock while adToCache runs. Actually what it needs is to just remove the read lock after reading

```
func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
	p.secretsCacheMx.RLock()
	data, ok := p.secretsCache[key]
	p.secretsCacheMx.RUnlock()
	if ok {
		data.lastAccess = time.Now()
		pass := data.value
		return pass, ok
	}
	data, ok = p.addToCache(key)
	return data.value, ok
}
```
One of the tests is not passing though ...

MichaelKatsoulis: From my side it's good to go! Thanks for patience!

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -46,67 +55,229 @@ func ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed boo
...
+}
+
+func (p *contextProviderK8sSecrets) getFromCache(key string) (string, bool) {
+	p.secretsCacheMx.RLock()
+	_, ok := p.secretsCache[key]
+	p.secretsCacheMx.RUnlock()
+
+	// if value is still not present in cache, it is possible we haven't tried to fetch it yet
+	if !ok {
+		value, ok := p.addToCache(key)
```
blakerouse, line 179: Why grab the lock inside of the function to set the new value. Then only 4 lines below grab it again to set the lastAccess time?

Why not just set `lastAccess` inside of `addToCache` and then return inside of the if statement? That would prevent the need to grab the lock twice for the same key.

blakerouse (changes requested):

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
constanca-m: > Why not just set lastAccess inside of addToCache and then return inside of the if statement?

Because then we would be changing `lastAccess` in two functions, when we don't need it. The logic is that `lastAccess` is only changed when it is required (by calling the get from cache).

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
constanca-m: It is a reading lock now @blakerouse (this file preview is not updated to the last commit)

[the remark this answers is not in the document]
constanca-m: I updated the code to do that in the latest commit @blakerouse

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
blakerouse: To me its worse to grab a lock, release the lock, grab the lock again, then release it again all in the same path. It should grab the lock do what it needs to do and release it.

blakerouse (approved):
Thanks for all the changes and fixes. I am glad a solution was able to be created for not holding the lock the entire time a refresh loop occurs.

My last point of hold/releasing followed by another hold/release is not a blocker, just seems like it would be less CPU cycles to grab the lock once. I understand your trying to be my DRY with the code, but for something as simple as setting a single attribute timestamp I think less lock contention would be worth it.

PR #3824, opened 2023-11-27
Pass the parsed version to all downloaders

### LINKED ISSUE #3813: Upgrades from 8.9 to present strip build metadata from valid semantic versions
https://github.com/elastic/elastic-agent/pull/2752 was merged in 8.9 to introduce the ability to upgrade to a specific snapshot version. It also adds the ability to parse a semantic version into its component parts. Unfortunately it only considers the major.minor.path and prerelease components when constructing non-snapshot upgrade URLs stripping out build metadata. This is unfortunate because we are considering a new versioning scheme that depends on build metadata.

Let's use `8.12.0+buildYYYYMMDD` as the example version we are attempting to upgrade to. The upgrade will fail as this version doesn't exist, but we should expect the download URL to contain `8.12.0+buildYYYYMMDD`.

### 8.8.2
When executing `elastic-agent upgrade 8.12.0+buildYYYYMMDD` in 8.8.2 and earlier we get the expected result where the `elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz` artifact is attempted.

```json
    {"log.level":"warn","@timestamp":"2023-11-23T19:54:34.850Z","log.origin":{"file.name":"upgrade/step_download.go","file.line":156},"message":"unable to download package: 2 errors occurred:\n\t* package '/Library/Elastic/Agent/data/elastic-agent-cdc5ba/downloads/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz' not found: open /Library/Elastic/Agent/data/elastic-agent-cdc5ba/downloads/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz: no such file or directory\n\t* call to 'https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz' returned 
```

### 8.11.1

When executing `elastic-agent upgrade 8.12.0+buildYYYYMMDD` from 8.9.0 to 8.11.1 the build metadata is stripped and we attempt to download `elastic-agent-8.12.0-darwin-aarch64.tar.gz`.

```json
    {"log.level":"warn","@timestamp":"2023-11-23T20:12:33.460Z","log.origin":{"file.name":"upgrade/step_download.go","file.line":207},"message":"unable to download package: 2 errors occurred:\n\t* package '/Library/Elastic/Agent/data/elastic-agent-03ef9d/downloads/elastic-agent-8.12.0-darwin-aarch64.tar.gz' not found: open /Library/Elastic/Agent/data/elastic-agent-03ef9d/downloads/elastic-agent-8.12.0-darwin-aarch64.tar.gz: no such file or directory\n\t* call to 'https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.12.0-darwin-aarch64.tar.gz' returned unsuccessful status code: 404\n\n; retrying (will be retry 1) in 42.682256265s.","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
```

### Root Cause

This happens because we explicitly strip the build metadata in the download step:

https://github.com/elastic/elastic-agent/blob/a14c51f043070c1fbdbd9c5c83307c5f34216b97/internal/pkg/agent/application/upgrade/step_download.go#L219-L225

This is because we allowed upgrading to specific snapshot versions by encoding the snapshot build ID in build metadata for the agent as `major.minor.patch-SNAPSHOT+buildID` for example `8.11.2-SNAPSHOT+ae92f19f`. The only downloader that understands build metadata is the snapshot downloader which explicitly handles it, so this solution hides the new build metadata for snapshot build IDs by not passing it through to the other downloaders. It is handled explicitly in the snapshot downloader as the build metadata needs to be included with a `-` instead of a `+` in the artifacts API.

https://github.com/elastic/elastic-agent/blob/a14c51f043070c1fbdbd9c5c83307c5f34216b97/internal/pkg/agent/application/upgrade/artifact/download/snapshot/downloader.go#L97-L109

### Solution

When we use the non-snapshot HTTP downloader we should pass through the entire version string. We should exclusively use the snapshot downloader when `SNAPSHOT` appears in the version prerelease information and use the generic HTTP downloader otherwise. There is a TODO in the existing code about this that can be addressed.

https://github.com/elastic/elastic-agent/blob/a14c51f043070c1fbdbd9c5c83307c5f34216b97/internal/pkg/agent/application/upgrade/step_download.go#L167-L179

### Acceptance Criteria
Assuming the host system is a Mac (substitute the architecture below as appropriate):

* `elastic-agent upgrade 8.12.0+buildYYYYMMDD` must be downloaded from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz
* `elastic-agent upgrade 8.11.2-SNAPSHOT+ae92f19f` must be downloaded from https://snapshots.elastic.co/8.11.2-ae92f19f/downloads/beats/elastic-agent/elastic-agent-8.11.2-SNAPSHOT-darwin-aarch64.tar.gz as it is today.
* Tests are in place to confirm the above behavior. Note that the +buildYYYYMMDD artifacts do not exist yet so it cannot be integration tested.

pchila:
## What does this PR do?

Avoid stripping build metadata off the version string for agent upgrade versions

## Why is it important?

It should be possible to use build metadata to identify specific builds of a given version for any downloader (fs, http and snapshot)

## Checklist

## Author's Checklist

- [ ]

Let's create 2 packages with versions containing build metadata from the elastic-agent code:

Note: the way we create the version containing `+buildYYYYMMDD` is a hack and it's not compatible with `SNAPSHOT` versions
- Create the first package camouflaging the agent as an `8.11.1` build by running:
  ```shell
  AGENT_PACKAGE_VERSION=8.11.1+build20231130 BEAT_VERSION=8.11.1  DEV=true EXTERNAL=true PLATFORMS="linux/amd64" PACKAGES=tar.gz mage package
  ```
- Create an empty commit before creating the second package (to avoid collisions in directory names during upgrade):
  ```shell
  git commit --allow-empty -m "Empty commit to change agent hash"
  ```
- Create a second package:
   ```shell
  AGENT_PACKAGE_VERSION=8.11.1+build20231201 BEAT_VERSION=8.11.1  DEV=true EXTERNAL=true PLATFORMS="linux/amd64" PACKAGES=tar.gz mage package
  ```
- Install the first package (no need to enroll in fleet)
- Test that upgrading using the fs downloader works (adapt the `file://` url to the location where the second package is located on disk):
  ```shell
  elastic-agent upgrade 8.11.1+build20231201 --skip-verify --source-uri file:///vagrant/build/distributions/
  ```
- Verify that the upgrade proceeds and that at the end we get an elastic-agent version similar to this:
  ```shell
  elastic-agent version --yaml
  ```
  ```shell
  root@elastic-agent-dev:~# elastic-agent version --yaml
  binary:
    version: 8.11.1+build20231201
    commit: 29def7ee77ea9f10de7cb0b24e85153f07372ec2
    build_time: 2023-12-01T09:59:54Z
    snapshot: false
  daemon:
    version: 8.11.1+build20231201
    commit: 29def7ee77ea9f10de7cb0b24e85153f07372ec2
    build_time: 2023-12-01T09:59:54Z
    snapshot: false
  ```
- Create a simple directory structure, link the last part of the path to where the packages are located and start a simple http server (in the example I am using `python`)
  ```shell
  mkdir beats
  ln -s /vagrant/build/distributions beats/elastic-agent
   python3 -m http.server 
  ```
  when the http server starts, it should look like this
  ```shell
  vagrant@elastic-agent-dev:~$  python3 -m http.server
  Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
  ```
- Now we can test the http downloader by specifying a `source-uri` by going back to the previous version
  ```shell
    root@elastic-agent-dev:~# elastic-agent upgrade 8.11.1+build20231130 --skip-verify --source-uri http://localhost:8000
  Upgrade triggered to version 8.11.1+build20231130, Elastic Agent is currently restarting
  ```
  ```shell
  root@elastic-agent-dev:~# elastic-agent version --yaml                                                    
  binary:
    version: 8.11.1+build20231130
    commit: 8439fe6a814319e3ad9575abd5dcc01012678d35
    build_time: 2023-12-01T13:50:11Z
    snapshot: false
  daemon:
    version: 8.11.1+build20231130
    commit: 8439fe6a814319e3ad9575abd5dcc01012678d35
    build_time: 2023-12-01T13:50:11Z
    snapshot: false

  ```

## Related issues

- Closes #3813

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/paths/common.go
+++ b/internal/pkg/agent/application/paths/common.go
@@ -59,7 +59,7 @@ func init() {
 	fs.StringVar(&configPath, "path.config", configPath, "Config path is the directory Agent looks for its config file")
 	fs.StringVar(&configFilePath, "c", DefaultConfigName, "Configuration file, relative to path.config")
 	fs.StringVar(&logsPath, "path.logs", logsPath, "Logs path contains Agent log output")
-	fs.StringVar(&installPath, "path.install", installPath, "Install path contains binaries Agent extracts")
+	fs.StringVar(&installPath, "path.install", installPath, "DEPRECATED, setting this flag has no effect since v8.6.0")
 
 	// enable user to download update artifacts to alternative place
 	// TODO: remove path.downloads support on next major (this can be configured using `agent.download.targetDirectory`)


internal/pkg/agent/application/upgrade/artifact/download/snapshot/downloader.go
```diff
@@ -66,8 +66,10 @@ func (e *Downloader) Reload(c *artifact.Config) error {
 
 // Download fetches the package from configured source.
 // Returns absolute path to downloaded package and an error.
-func (e *Downloader) Download(ctx context.Context, a artifact.Artifact, version string) (string, error) {
-	return e.downloader.Download(ctx, a, version)
+func (e *Downloader) Download(ctx context.Context, a artifact.Artifact, version *agtversion.ParsedSemVer) (string, error) {
+	// remove build metadata to match filename of the package for the specific snapshot build
+	strippedVersion := agtversion.NewParsedSemVer(version.Major(), version.Minor(), version.Patch(), version.Prerelease(), "")
```
cmacknz, line 71: What happens if we don't do this? Why can't we have a version like `8.13.0-SNAPSHOT+buildYYYYMMDD` where -SNAPSHOT is the prerelease component?

internal/pkg/agent/application/upgrade/artifact/download/http/verifier.go
```diff
@@ -115,13 +116,13 @@ func (v *Verifier) Verify(a artifact.Artifact, version string, skipDefaultPgp bo
 	return nil
 }
 
-func (v *Verifier) verifyAsc(a artifact.Artifact, version string, skipDefaultKey bool, pgpSources ...string) error {
-	filename, err := artifact.GetArtifactName(a, version, v.config.OS(), v.config.Arch())
+func (v *Verifier) verifyAsc(a artifact.Artifact, aVersion version.ParsedSemVer, skipDefaultKey bool, pgpSources ...string) error {
+	filename, err := artifact.GetArtifactName(a, aVersion, v.config.OS(), v.config.Arch())
```
cmacknz, line 120: nit: all these renames from `version` to `aVersion` are making it harder to pick out the meaningful changes.

internal/pkg/agent/application/upgrade/artifact/download/snapshot/verifier.go
```diff
@@ -41,8 +41,9 @@ func NewVerifier(log *logger.Logger, config *artifact.Config, pgp []byte, versio
 }
 
 // Verify checks the package from configured source.
-func (v *Verifier) Verify(a artifact.Artifact, version string, skipDefaultPgp bool, pgpBytes ...string) error {
-	return v.verifier.Verify(a, version, skipDefaultPgp, pgpBytes...)
+func (v *Verifier) Verify(a artifact.Artifact, version agtversion.ParsedSemVer, skipDefaultPgp bool, pgpBytes ...string) error {
+	strippedVersion := agtversion.NewParsedSemVer(version.Major(), version.Minor(), version.Patch(), version.Prerelease(), "")
```
cmacknz, line 45: Same as in the download case, what happens if we don't strip the build metadata?

(I also worry about having to duplicate this across two files but that is probably too much refactoring to pull into this change)

internal/pkg/agent/application/upgrade/step_download.go
```diff
@@ -219,7 +219,7 @@ func (u *Upgrader) downloadOnce(
 	// All download artifacts expect a name that includes <major>.<minor.<patch>[-SNAPSHOT] so we have to
 	// make sure not to include build metadata we might have in the parsed version (for snapshots we already
 	// used that to configure the URL we download the files from)
-	path, err := downloader.Download(ctx, agentArtifact, version.VersionWithPrerelease())
+	path, err := downloader.Download(ctx, agentArtifact, version)
```
cmacknz, line 222: This is the actual fix, are we able to unit or integration test this?

internal/pkg/agent/application/upgrade/artifact/download/http/verifier.go
pchila: Well there was a name conflict with the `version` package which I tried to disambiguate. Would you prefer a rename of the package import perhaps ?

internal/pkg/agent/application/upgrade/artifact/download/snapshot/verifier.go
pchila: If we leave the build metadata in, no existing snapshot package file name will match, even the ones that currently work for the upgrade to a specific snapshot build, for example looking at https://artifacts-api.elastic.co/v1/versions/8.12.0-SNAPSHOT/builds/8.12.0-7d5211d6

```json
 "elastic-agent-8.12.0-SNAPSHOT-linux-x86_64.tar.gz": {
            "url": "https://snapshots.elastic.co/8.12.0-7d5211d6/downloads/beats/elastic-agent/elastic-agent-8.12.0-SNAPSHOT-linux-x86_64.tar.gz",
            "sha_url": "https://snapshots.elastic.co/8.12.0-7d5211d6/downloads/beats/elastic-agent/elastic-agent-8.12.0-SNAPSHOT-linux-x86_64.tar.gz.sha512",
            "asc_url": "https://snapshots.elastic.co/8.12.0-7d5211d6/downloads/beats/elastic-agent/elastic-agent-8.12.0-SNAPSHOT-linux-x86_64.tar.gz.asc",
            "type": "tar",
            "architecture": "x86_64",
            "os": [
              "linux"
            ]
          },
```
We can avoid stripping the metadata only once the downloaders become a lot smarter at looking up versions but that is a big change

The duplication is due to the way the code is structured and I wouln't want

internal/pkg/agent/application/upgrade/artifact/download/http/verifier.go
pchila: Renamed the imports and restored `version` variable names

internal/pkg/agent/application/upgrade/artifact/download/snapshot/downloader.go
cmacknz: Right the problem is that we have decided that snapshot IDs are specified as build metadata, even though that is not actually the version in the artifacts API.  ability to handle the case where we actually want build metadata in the version.

We could solve this conflict in the artifacts API by doing something like:

https://snapshots.elastic.co/8.11.2-buildYYYYMMDD/downloads/beats/elastic-agent/elastic-agent-8.11.2-SNAPSHOT-darwin-aarch64.tar.gz

That said this doesn't exist yet, and because this is only for snapshots we can fix it without worrying about the stack release cadence. So let's just leave it as is.

internal/pkg/agent/application/upgrade/artifact/download/snapshot/verifier.go
cmacknz: As noted above, let's leave this alone.

internal/pkg/agent/application/upgrade/step_download.go
blakerouse: I believe a mock HTTP server which is something that is really easily to create with golang is the way to test this. I agree that we need to test this to ensure that the behavior is working as expected.

I would prefer to see the test at this level as well, so its testing the whole step_download section.

blakerouse (changes requested):
Overall this looks good.

I am worried that it doesn't have an associated unit test with it. I think we need to get that added before this gets merged.

internal/pkg/agent/application/upgrade/step_download.go
AndersonQ: >For unit tests, given the current code structure we would have to mock an artifact api on a test HTTP server, not exactly a "unit" test

It won't be the first _not exactly a "unit" test_, besides as Blake pointed out, it's easy in Go and if you want, you could make it an integration test. 

I agree a test is important, for me it can be in any form you prefer as long as there is a test

AndersonQ (approved):

cmacknz (approved):
The new tests are great! Thank you!

Did a quick sanity check running `sudo elastic-agent upgrade 8.12.0+buildYYYYMMDD` and the build metadata is no longer stripped as expected.

```
❯ sudo elastic-agent status
┌─ fleet
├─ elastic-agent
│  └─ status: (UPGRADING) Upgrading to version 8.12.0+buildYYYYMMDD
└─ upgrade_details
   ├─ target_version: 8.12.0+buildYYYYMMDD
   ├─ state: UPG_DOWNLOADING
   └─ metadata
      ├─ download_percent: 0.00%
      ├─ retry_until: 1h59m44.87304s
      └─ retry_error_msg: unable to download package: 3 errors occurred:
             * package '/Library/Elastic/Agent/data/elastic-agent-f29cd5/downloads/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz' not found: open /Library/Elastic/Agent/data/elastic-agent-f29cd5/downloads/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz: no such file or directory
             * call to 'https://snapshots.elastic.co/8.12.0-51bw2wko/downloads/beats/elastic-agent/elastic-agent-8.12.0-darwin-aarch64.tar.gz' returned unsuccessful status code: 404
             * call to 'https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.12.0+buildYYYYMMDD-darwin-aarch64.tar.gz' returned unsuccessful status code: 404

```

blakerouse (approved):
Looks good.

I really appreciate you adding the tests. I know that part is not fun and can be really boring, but it really does provide confidence in this PR and more confidence in the Elastic Agent overall. Thank you!

Issue #3821, opened 2023-11-25
Watcher does not consider build hash when determining if upgrade was rolled back


AndersonQ:
With the introduction of the tracking of `UPG_ROLLBACK` state the watcher compares the new version of the agent with the previous one. However when comparing it only uses the major.minor.path[-SNAPSHOT], ignoring the build hash.

If the agent is upgraded from an SNAPSHOT version to another the watcher considers they as being the same version and therefore it marks the upgrade as `UPG_ROLLBACK` and neither logs why it changed the upgrade status nor adds any information to the new state.
This leads to an successful upgrade, a healthy agent but marked as `UPG_ROLLBACK` as shown below. Without any log or information explaining why the agent is on an `UPG_ROLLBACK` state.

![Screenshot from 2023-11-24 15-35-48](https://github.com/elastic/elastic-agent/assets/203669/4eb900c7-5724-4d6b-892d-e49fb0160d84)

The culprit seems to be here:

https://github.com/elastic/elastic-agent/blob/d58ac9cb9183317ea0bc58bc6d1e38244000b5a3/internal/pkg/agent/application/upgrade/marker_watcher.go#L117-L123

The `UpgradeMarker` has both hashes, it just need to be used when comparing the versions

https://github.com/elastic/elastic-agent/blob/d58ac9cb9183317ea0bc58bc6d1e38244000b5a3/internal/pkg/agent/application/upgrade/step_mark.go#L27-L43

For confirmed bugs, please report:
- Version: main
- Operating System: all
- Discuss Forum URL: N/A
- Steps to Reproduce: Upgrade from a snapshot build to another from the same version.

pchila:
Reopening as it needs fixes from PRs #3827 #3837

PR #3827, opened 2023-11-27
[Upgrade Details] Ensure details report `UPG_WATCHING` for the entire time that the upgrade is being watched

### LINKED ISSUE #3816: The UPG_WATCHING state is not reported while the new agent version is being watched
The upgrade watching state is only reported briefly before the upgrade restarting state and then is never reported again. This is incorrect because the watcher runs for ~10 minutes after the new version of the agent is started. The agent should remain in the upgrade watching state as long as the watcher is running.

Here are the upgrade details logs capture up to the point where the agent restarts:

```json
{"log.level":"info","@timestamp":"2023-11-24T16:18:08.241Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_REQUESTED","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:08.241Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_DOWNLOADING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:08.967Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_DOWNLOADING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:26.438Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_DOWNLOADING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"16.87223912MBps"}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:26.540Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_DOWNLOADING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"16.87223912MBps"}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:26.541Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_DOWNLOADING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"+Inf bps"}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:28.397Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_EXTRACTING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"+Inf bps"}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:35.133Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_REPLACING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"+Inf bps"}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:35.138Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_WATCHING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"+Inf bps"}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2023-11-24T16:18:35.154Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":500},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.11.1","state":"UPG_RESTARTING","action_id":"aa72eec8-90b0-4889-b329-8472f5ee9e1d","metadata":{"download_percent":1,"download_rate":"+Inf bps"}},"ecs.version":"1.6.0"}
```

Here is the output of the status command through this same transition, notice that the new agent restarts without upgrade details.

```
~/Downloads/elastic-agent-8.3.0-SNAPSHOT-darwin-aarch64 ····································· 03:15:58 PM
❯ sudo elastic-agent status
┌─ fleet
├─ elastic-agent
│  └─ status: (UPGRADING) Upgrading to version 8.11.1
└─ upgrade_details
   ├─ target_version: 8.11.1
   ├─ state: UPG_EXTRACTING
   ├─ action_id: 00dd5998-c0ce-49d5-8536-d64b12a964ff
   └─ metadata
~/Downloads/elastic-agent-8.3.0-SNAPSHOT-darwin-aarch64 ····································· 03:15:59 PM
❯ sudo elastic-agent status
┌─ fleet
│  └─ status: (STOPPED) Requested to be stopped
└─ elastic-agent
   └─ status: (STOPPING) Re-executing
~/Downloads/elastic-agent-8.3.0-SNAPSHOT-darwin-aarch64 ····································· 03:16:25 PM
❯ sudo elastic-agent status
┌─ fleet
│  └─ status: (STARTING)
└─ elastic-agent
   └─ status: (HEALTHY) Running
```

The expectation is that the agent should be in the UPG_WATCHING state and report itself as such to Fleet. The watcher is still running at this point:

```
~/Downloads/elastic-agent-8.3.0-SNAPSHOT-darwin-aarch64 ······························· ✘ INT 03:19:46 PM
❯ ps aux | grep elastic-agent
root             92594   0.0  0.2 409891920  54624   ??  Ss    3:15PM   0:01.17 elastic-agent
root             93003   0.0  0.1 409369760  35888   ??  Ss    3:15PM   0:00.19 /Library/Elastic/Agent/elastic-agent watch --path.config /Library/Elastic/Agent --path.home /Library/Elastic/Agent
```

It may be worth reporting the UPG_WATCHING state when we report that the `elastic-agent watch` command is executing since we can now detect that. That would eliminate the need to report this through the upgrade marker file.

ycombinator:
## What does this PR do?

This PR fixes the upgrade details such that they are in the `UPG_WATCHING` state the entire time the Agent upgrade is being watched by the Upgrade Watcher.

## Why is it important?

So the state of the upgrade is accurately reported.

## Checklist

- [ ] ~I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~ Bug was never released.

1. Build Elastic Agent from this PR but give it a lower-than-current version number.  This will be the starting (pre-upgrade) version of the Agent.
   ```
   AGENT_PACKAGE_VERSION=8.11.0 EXTERNAL=true SNAPSHOT=true PLATFORMS=darwin/arm64 PACKAGES=targz mage package
   ```
2. Make a no-op commit.  Without this the Agent upgrade will not succeed.
3. Build Elastic Agent again. This will be the target (post-upgrade) version of the Agent.
   ```
   EXTERNAL=true SNAPSHOT=true PLATFORMS=darwin/arm64 PACKAGES=targz mage package
   ```
4. Install the starting version of the Agent.
5. Upgrade to the target version of the Agent.
   ```
   sudo elastic-agent upgrade 8.12.0-SNAPSHOT --source-uri file:///Users/shaunak/development/github/elastic-agent/build/distributions --skip-verify
   ```
6. Check that the Agent status reports upgrade details with a state of `UPG_WATCHING`.
   ```
   sudo elastic-agent status --output json | jq '.upgrade_details'
   {
     "target_version": "8.12.0",
     "state": "UPG_WATCHING",
     "metadata": {}
   }
   ```
8. Check the Agent logs and verify the upgrade details states are in order.  In particular, make sure that **chronologically**, we see `UPG_WATCHING` after `UPG_RESTARTING`.
   ```
   sudo grep -R -h --include=\*.ndjson UPG_ /Library/Elastic/Agent | jq -c -s 'sort_by(.["@timestamp"]) | .[]'
   {"log.level":"info","@timestamp":"2023-11-28T11:13:45.530Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0-SNAPSHOT","state":"UPG_REQUESTED","metadata":{}},"ecs.version":"1.6.0"}
   {"log.level":"info","@timestamp":"2023-11-28T11:13:45.531Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0-SNAPSHOT","state":"UPG_DOWNLOADING","metadata":{}},"ecs.version":"1.6.0"}
   {"log.level":"info","@timestamp":"2023-11-28T11:13:45.729Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0-SNAPSHOT","state":"UPG_EXTRACTING","metadata":{}},"ecs.version":"1.6.0"}
   {"log.level":"info","@timestamp":"2023-11-28T11:13:52.308Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0-SNAPSHOT","state":"UPG_REPLACING","metadata":{}},"ecs.version":"1.6.0"}
   {"log.level":"info","@timestamp":"2023-11-28T11:13:52.313Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0-SNAPSHOT","state":"UPG_REPLACING","metadata":{}},"ecs.version":"1.6.0"}
   {"log.level":"info","@timestamp":"2023-11-28T11:13:52.320Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0-SNAPSHOT","state":"UPG_RESTARTING","metadata":{}},"ecs.version":"1.6.0"}
   {"log.level":"info","@timestamp":"2023-11-28T11:14:00.374Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":499},"message":"updated upgrade details","log":{"source":"elastic-agent"},"upgrade_details":{"target_version":"8.12.0","state":"UPG_WATCHING","metadata":{}},"ecs.version":"1.6.0"}
   ```
6. Wait until the Upgrade Watcher has finished running.
   ```
   pgrep -f 'elastic-agent watch' | wc -l    # should report 0 eventually
   ```
9. Check the Agent version and verify that it's the target version.
   ```
   sudo elastic-agent version

   Binary: 8.12.0-SNAPSHOT (build: 80bb6a61369c20e054478a73c6c866aadfcc52b1 at 2023-11-27 23:38:37 +0000 UTC)
   Daemon: 8.12.0-SNAPSHOT (build: 80bb6a61369c20e054478a73c6c866aadfcc52b1 at 2023-11-27 23:38:37 +0000 UTC)
   ```
11. Cleanup: revert/remove the no-op commit from step 2.

## Related issues
- Fixes #3816

### DIFF
--- a/internal/pkg/agent/application/upgrade/marker_watcher.go
+++ b/internal/pkg/agent/application/upgrade/marker_watcher.go
@@ -139,5 +139,7 @@ func (mfw *MarkerFileWatcher) processMarker(currentVersion string, commit string
 		}
 	}
 
+	// FIXME: Special case for UPG_COMPLETED -> nil details
+
 	mfw.updateCh <- *marker
 }
--- a/internal/pkg/agent/application/upgrade/marker_watcher.go
+++ b/internal/pkg/agent/application/upgrade/marker_watcher.go
@@ -139,7 +139,5 @@ func (mfw *MarkerFileWatcher) processMarker(currentVersion string, commit string
 		}
 	}
 
-	// FIXME: Special case for UPG_COMPLETED -> nil details
-
 	mfw.updateCh <- *marker
 }
--- a/internal/pkg/agent/application/upgrade/marker_watcher.go
+++ b/internal/pkg/agent/application/upgrade/marker_watcher.go
@@ -129,9 +129,12 @@ func (mfw *MarkerFileWatcher) processMarker(currentVersion string, commit string
 	// isn't for some reason, we fallback to explicitly setting that state as
 	// part of the upgrade details in the marker.
 	if marker.PrevVersion == currentVersion && marker.PrevHash == commit && !mfw.upgradeStarted.Load() {
+		// If there are no upgrade details in the marker or the state in the
+		// details is not set for some reason, we assume the worst and
+		// explicitly set the state to UPG_ROLLBACK
 		if marker.Details == nil {
 			marker.Details = details.NewDetails("unknown", details.StateRollback, marker.GetActionID())
-		} else {
+		} else if marker.Details.State == "" {
 			marker.Details.SetState(details.StateRollback)
 		}
 	}
--- a/internal/pkg/agent/application/upgrade/marker_watcher.go
+++ b/internal/pkg/agent/application/upgrade/marker_watcher.go
@@ -29,6 +29,7 @@ type MarkerFileWatcher struct {
 	updateCh       chan UpdateMarker
 
 	upgradeStarted atomic.Bool
+	lastMarker     *UpdateMarker
 }
 
 func newMarkerFileWatcher(upgradeMarkerFilePath string, logger *logger.Logger) MarkerWatcher {
@@ -100,6 +101,16 @@ func (mfw *MarkerFileWatcher) Run(ctx context.Context) error {
 					// Upgrade marker file was created or updated; read its contents
 					// and send them over the update channel.
 					mfw.processMarker(version.GetAgentPackageVersion(), version.Commit())
+				case e.Op&(fsnotify.Remove) != 0:
+					// Upgrade marker file was removed.
+					// - Upgrade could've been rolled back
+					// - Upgrade could've been successful
+					// If last known Upgrade Details state is not `UPG_ROLLBACK`, assume
+					// upgrade was successful
+					if mfw.lastMarker != nil && mfw.lastMarker.Details != nil && mfw.lastMarker.Details.State != details.StateRollback {
+						mfw.lastMarker.Details = nil
+						mfw.updateCh <- *mfw.lastMarker
+					}
 				}
 			case <-doInitialRead:
 				mfw.processMarker(version.GetAgentPackageVersion(), version.Commit())
@@ -139,5 +150,6 @@ func (mfw *MarkerFileWatcher) processMarker(currentVersion string, commit string
 		}
 	}
 
+	mfw.lastMarker = marker
 	mfw.updateCh <- *marker
 }
--- a/testing/upgradetest/upgrader.go
+++ b/testing/upgradetest/upgrader.go
@@ -34,6 +34,10 @@ type upgradeOpts struct {
 	customPgp        *CustomPGP
 	customWatcherCfg string
 
+	// TODO: should be removed along with all references once 8.13.0 has been released.
+	// See also WithDisableUpgradeWatcherUpgradeDetailsCheck.
+	disableUpgradeWatcherUpgradeDetailsCheck bool
+
 	preInstallHook  func() error
 	postInstallHook func() error
 	preUpgradeHook  func() error
@@ -108,6 +112,17 @@ func WithCustomWatcherConfig(cfg string) upgradeOpt {
 	}
 }
 
+// WithDisableUpgradeWatcherUpgradeDetailsCheck disables any assertions for
+// upgrade details that are being set by the Upgrade Watcher. This option is
+// useful in upgrade tests where the end Agent version does not contain changes
+// in the Upgrade Watcher whose effects are being asserted upon in PerformUpgrade.
+// TODO: should be removed along with all references once 8.13.0 has been released.
+func WithDisableUpgradeWatcherUpgradeDetailsCheck() upgradeOpt {
+	return func(opts *upgradeOpts) {
+		opts.disableUpgradeWatcherUpgradeDetailsCheck = true
+	}
+}
+
 // PerformUpgrade performs the upgrading of the Elastic Agent.
 func PerformUpgrade(
 	ctx context.Context,
@@ -159,6 +174,16 @@ func PerformUpgrade(
 	if err != nil {
 		return fmt.Errorf("failed to get end agent build version info: %w", err)
 	}
+
+	// For asserting on the effects of any Upgrade Watcher changes made in 8.13.0, we need
+	// the endVersion to be >= 8.13.0.  Otherwise, these assertions will fail as those changes
+	// won't be present in the Upgrade Watcher.
+	endVersion, err := version.ParseVersion(endVersionInfo.Binary.Version)
+	if err != nil {
+		return fmt.Errorf("failed to parse version of upgraded Agent binary: %w", err)
+	}
+	upgradeOpts.disableUpgradeWatcherUpgradeDetailsCheck = upgradeOpts.disableUpgradeWatcherUpgradeDetailsCheck ||
+		version.NewParsedSemVer(8, 13, 0, "", "").Less(*endVersion)
 
 	if upgradeOpts.preInstallHook != nil {
 		if err := upgradeOpts.preInstallHook(); err != nil {
@@ -253,16 +278,8 @@ func PerformUpgrade(
 	logger.Logf("upgrade watcher started")
 
 	// Check that, while the Upgrade Watcher is running, the upgrade details in Agent status
-	// show the state as UPG_WATCHING. However, this UPG_WATCHING state is set by the Upgrade
-	// Watcher, which will be of the target version of the upgrade. And the logic to set this
-	// state is only present in Upgrade Watcher code >= v8.13.0. So we must only run this check
-	// if the endFixture's version, that is the target version of the upgrade, is >= 8.13.0.
-	endVersion, err := version.ParseVersion(endVersionInfo.Binary.Version)
-	if err != nil {
-		return fmt.Errorf("failed to parse version of upgraded Agent binary: %w", err)
-	}
-
-	if !endVersion.Less(*version.NewParsedSemVer(8, 13, 0, "", "")) {
+	// show the state as UPG_WATCHING.
+	if !upgradeOpts.disableUpgradeWatcherUpgradeDetailsCheck {
 		logger.Logf("Checking upgrade details state while Upgrade Watcher is running")
 		if err := waitUpgradeDetailsState(ctx, startFixture, details.StateWatching, 2*time.Minute, 10*time.Second, logger); err != nil {
 			// error context added by waitUpgradeDetailsState
@@ -302,11 +319,8 @@ func PerformUpgrade(
 	logger.Logf("upgrade watcher finished")
 
 	// Check that, upon successful upgrade, the upgrade details have been cleared out
-	// from Agent status.  However, the logic to clear out upgrade details is implemented in
-	// the Upgrade Watcher, which will be of the target version of the upgrade. And this logic
-	// is only present in Upgrade Watcher code >= v8.13.0. So we must only run this check
-	// if the endFixture's version, that is the target version of the upgrade, is >= 8.13.0.
-	if !endVersion.Less(*version.NewParsedSemVer(8, 13, 0, "", "")) {
+	// from Agent status.
+	if !upgradeOpts.disableUpgradeWatcherUpgradeDetailsCheck {
 		logger.Logf("Checking upgrade details state after successful upgrade")
 		if err := waitUpgradeDetailsState(ctx, startFixture, "", 2*time.Minute, 10*time.Second, logger); err != nil {
 			// error context added by checkUpgradeDetailsState
--- a/testing/upgradetest/upgrader.go
+++ b/testing/upgradetest/upgrader.go
@@ -244,6 +244,8 @@ func PerformUpgrade(
 		return fmt.Errorf("failed to start agent upgrade to version %q: %w\n%s", endVersionInfo.Binary.Version, err, upgradeOutput)
 	}
 
+	// Check that, right before the Upgrade Watcher is started, the upgrade details in Agent status
+	// show the state as UPG_REPLACING.
 	if err := waitUpgradeDetailsState(ctx, startFixture, details.StateReplacing, 2*time.Minute, 10*time.Second); err != nil {
 		// error context added by checkUpgradeDetailsState
 		return err
@@ -281,6 +283,13 @@ func PerformUpgrade(
 		}
 
 		// error context added by WaitHealthyAndVersion
+		return err
+	}
+
+	// Check that, upon successful upgrade, the upgrade details have been cleared out
+	// from Agent status.
+	if err := waitUpgradeDetailsState(ctx, startFixture, "", 2*time.Minute, 10*time.Second); err != nil {
+		// error context added by checkUpgradeDetailsState
 		return err
 	}
 
--- a/testing/upgradetest/upgrader.go
+++ b/testing/upgradetest/upgrader.go
@@ -434,7 +434,7 @@ func waitUpgradeDetailsState(ctx context.Context, f *atesting.Fixture, expectedS
 		select {
 		case <-ctx.Done():
 			if lastErr != nil {
-				return fmt.Errorf("failed waiting for healthy agent and version (%w): %w", ctx.Err(), lastErr)
+				return fmt.Errorf("failed waiting for status (%w): %s", ctx.Err(), lastErr.Error())
 			}
 			return ctx.Err()
 		case <-t.C:
--- a/testing/upgradetest/upgrader.go
+++ b/testing/upgradetest/upgrader.go
@@ -434,7 +434,7 @@ func waitUpgradeDetailsState(ctx context.Context, f *atesting.Fixture, expectedS
 		select {
 		case <-ctx.Done():
 			if lastErr != nil {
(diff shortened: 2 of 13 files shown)


internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -115,15 +115,18 @@ func (mfw *MarkerFileWatcher) processMarker(currentVersion string) {
...
-	// now is the same as the prevVersion recorded in the marker, it means
-	// the upgrade was rolled back. Ideally, this UPG_ROLLBACK state would've
-	// been recorded in the marker's upgrade details field but, in case it
-	// isn't for some reason, we fallback to explicitly setting that state as
-	// part of the upgrade details in the marker.
+	// now is the same as the prevVersion recorded in the marker, it could mean
+	// that we are in one of two states:
+	// - the marker was just created and the upgrade is about to start
+	//   (marker.details.state should not be empty), or
+	// - the upgrade was rolled back (marker.details.state should be UPG_ROLLBACK)
```
cmacknz, line 122: Is there a third option we haven't considered? The case where the agent is restarted in the middle of an upgrade, for example when it is in the UPGRADE_EXTRACTING state. The obvious example would be the host system powering off.

What happens in this case? What is reported in the upgrade details?

cmacknz: If we just rolled back, is the previous version of the agent guaranteed to be next to the currently running agent in the data path? That is is the state of the filesystem always something like:

```
data/
  elastic-agent-current/
  elastic-agent-next/
```

If this is true, can we look to see if there's another agent in the data directory next to us to detect if there was a roll back?

cmacknz: We will likely have both versions next to each other if the upgrade is interrupted immediately after the artifact is extracted, so that may not be 100% either.

One more thing, in the case that the host system powers off we should start calling https://pkg.go.dev/os#File.Sync on the marker file. It doesn't look like we do this today.

https://github.com/elastic/elastic-agent/blob/97e82175953af75fbe29fccc3cfcc5688a96dc8f/internal/pkg/agent/application/upgrade/marker_access_other.go#L26-L30

This might only be worth doing arounds critical transitions, like when the agent process re-execs or the watcher first starts up.

ycombinator: I've implemented the fsync change in its own PR, since it's not strictly related to this PR here: https://github.com/elastic/elastic-agent/pull/3836

ycombinator: > Is there a third option we haven't considered? The case where the agent is restarted in the middle of an upgrade, for example when it is in the UPGRADE_EXTRACTING state. The obvious example would be the host system powering off.

> What happens in this case? What is reported in the upgrade details?

It depends on whether the state transition happens before or after the upgrade marker file comes into existence.  

The following states occur before the upgrade marker file comes into existence and, as such, are never persisted in it: `UPG_REQUESTED`, `UPG_SCHEDULED`, `UPG_DOWNLOADING`, `UPG_EXTRACTING`.   Additionally, the `UPG_RESTARTING` state is also currently not being persisted to the upgrade marker file, mostly because of where this state transition happens in the code vs. where the upgrade marker file is being created. With some refactoring, we could start persisting this state to the upgrade marker file as well. So if the Agent were to restart during one of these states, the upgrade details that are stored in the Coordinator state (and from there sent to Fleet) would get reset to nothing and the upgrade state would be lost.

The following states occur right before or after the upgrade marker file comes into existence and, as such, do get persisted to it: `UPG_REPLACING`, `UPG_WATCHING`, `UPG_ROLLBACK`.  So if the Agent were to restart during one of these states, the upgrade details from the upgrade marker would be restored to the Coordinator state (and from there sent to Fleet).

We may want to consider either persisting upgrade details in their own file throughout the upgrade process OR creating the upgrade marker file at the start of the upgrade process instead of where it's being created now (right before the Upgrade Watcher is invoked from the old Agent).

ycombinator: > If we just rolled back, is the previous version of the agent guaranteed to be next to the currently running agent in the data path? That is is the state of the filesystem always something like:

```
data/
  elastic-agent-current/
  elastic-agent-next/
```

> If this is true, can we look to see if there's another agent in the data directory next to us to detect if there was a roll back?

Two folders will exist the moment we go past the `UPG_EXTRACTING` state and, yes, two folders will also exist when we are in the `UPG_ROLLBACK` state.  So I'm not sure if checking if the number of folders > 1 is sufficient to determine if we're about to upgrade or if we've just rolled back.

But I think there might be another solution to detecting if we're about to upgrade: the code that creates the Upgrade Marker file runs in the same process as the code that's watching the Upgrade Marker file for changes.  As such, the former code can communicate to the latter in memory that we're about to upgrade.  In the case of a rollback, this communication will not happen.  

Let me explore this solution in a separate PR as it's not strictly related to this PR here.

ycombinator: > But I think there might be another solution to detecting if we're about to upgrade: the code that creates the Upgrade Marker file runs in the same process as the code that's watching the Upgrade Marker file for changes. As such, the former code can communicate to the latter in memory that we're about to upgrade. In the case of a rollback, this communication will not happen.

> Let me explore this solution in a separate PR as it's not strictly related to this PR here.

https://github.com/elastic/elastic-agent/pull/3837

AndersonQ (approved):
I tested it using the air-gapped test, it works. There is still what I believe is something similar to https://github.com/elastic/elastic-agent/issues/3821, but it's because it's upgrading from an snapshot to anoter. Thus all good for this PR

testing/upgradetest/upgrader.go
```diff
@@ -250,6 +259,13 @@ func PerformUpgrade(
 	}
 	logger.Logf("upgrade watcher started")
 
+	// Check that, while the Upgrade Watcher is running, the upgrade details in Agent status
+	// show the state as UPG_WATCHING.
+	if err := checkUpgradeDetailsState(ctx, startFixture, details.StateWatching); err != nil {
+		// error context added by checkUpgradeDetailsState
+		return err
+	}
+
```
pchila, line 268: Maybe we want to check for a final state of the upgrade like COMPLETED or FAILED ?

internal/pkg/agent/application/upgrade/marker_watcher.go
cmacknz: > The following states occur right before or after the upgrade marker file comes into existence and, as such, do get persisted to it: UPG_REPLACING, UPG_WATCHING, UPG_ROLLBACK. So if the Agent were to restart during one of these states, the upgrade details from the upgrade marker would be restored to the Coordinator state (and from there sent to Fleet).

What happens to the upgrade action if an upgrade is interrupted after it is started but before it is completed? Does the agent start over from the beginning? Does it acknowledge it as if the upgrade had happened if though it didn't? Does it never get acknowledged?

What does the watcher do if it starts up, sees an upgrade marker, but the version of the agent that is currently running isn't the version that should be running?

It would be surprising for a user to see an upgrade stuck in `UPG_REPLACING` for example. If the upgrade is essentially aborted by a host reboot then `UPG_ROLLBACK` could be considered the correct state. The `UPG_WATCHING` always clears itself when the watcher stops but I'm not sure what happens in this state in this situation today.

I think the ideal thing to happen in this situation is that the upgrade restarts from the beginning when the agent host system comes back online.

ycombinator: > What happens to the upgrade action if an upgrade is interrupted after it is started but before it is completed? Does the agent start over from the beginning? Does it acknowledge it as if the upgrade had happened if though it didn't? Does it never get acknowledged?

Looking at the code, once the Upgrade Marker has been created, if the Agent restarts, it will acknowledge the upgrade with Fleet _even if the upgrade may not have completed_.  In fact, it's entirely possible that Agent acknowledges the upgrade with Fleet while the Upgrade Watcher is still running and then the Upgrade Watcher decides to roll back the Agent.  As far as Fleet is concerned, the upgrade would've been reported as successful, and then within 10 minutes, the previous version of Agent would start showing again without any explanation as to why.

With Upgrade Details, in the above rollback scenario, the Upgrade Watcher will record the state as `UPG_ROLLBACK` in the Upgrade Marker file, which will get picked up by the main Agent process and sent to Fleet.

> What does the watcher do if it starts up, sees an upgrade marker, but the version of the agent that is currently running isn't the version that should be running?

Again, looking at the code...

First, if the Upgrade Watcher was running before an interruption stopped/killed it, and then the Upgrade Watcher was restarted, it will exit immediately because the watcher lock file, `watcher.lock`, will still exist.  In this case, there are two possibilities as to which state will be reported in Upgrade Details to Fleet:
- if the Upgrade Marker contains Upgrade Details, the state recorded in it will be reported to Fleet. This should be the `UPG_REPLACING` state as it's the last state that's persisted to the Upgrade Marker before the old Agent's upgrade code restarts the new Agent.
- if, for some reason, the Upgrade Marker does not contain Upgrade Details and the version of the running Agent is the same as the previous version recorded in the Upgrade Marker, `UPG_ROLLBACK` will be reported to Fleet.

If the Upgrade Watcher wasn't running yet when the upgrade process was interrupted, the Upgrade Watcher will start monitoring the Agent regardless of what version of Agent is currently running or what version is recorded in the Upgrade Marker file.  In this case, `UPG_WATCHING` will be reported to Fleet.  If the watch succeeds, the Upgrade Details will stop being reported to Fleet; note that the version of Agent being reported to Fleet will already be the new one in this case.  If the watch fails and Agent has to be rolled back, `UPG_ROLLBACK` will be reported to Fleet.

> I think the ideal thing to happen in this situation is that the upgrade restarts from the beginning when the agent host system comes back online.

Agreed but I think restarting the upgrade from the beginning after a crash is beyond the scope of this PR so I've created https://github.com/elastic/elastic-agent/issues/3860 to track this improvement.

cmacknz: Agree fixing this is outside the scope of this PR given your explanation.

I also think we acknowledge the upgrade too early since it isn't synchronized with the watcher but that is also out of scope.

cmacknz (approved):
Tested manually and confirmed it works, thanks!

ycombinator:
All upgrade-related integration tests are passing now.  But all Endpoint-related integration tests are failing now with the same symptom:

```
Endpoint component or units are not healthy
```

Failures don't seem related to the changes in this PR.

jlind23:
buildkite test this

PR #3837, opened 2023-11-29
[Upgrade Details] Inform the Upgrade Marker Watcher when an upgrade starts

ycombinator:
## What does this PR do?

This PR allows the part of the upgrade process performed by the old (pre-upgrade) Agent to inform the Upgrade Marker Watcher when an upgrade starts.

## Why is it important?

The Upgrade Marker Watcher's job is to watch for changes to the Upgrade Marker file and process said changes.  As part of this processing, it implements fallback logic to determine if the upgrade has been rolled back.  For this, it checks whether the running Agent's version is the same as the `prev_version` noted in the Upgrade Marker. 

However, this version check alone is not sufficient to say that the Agent has rolled back, as the check will also succeed in the period right after the old (pre-upgrade) Agent creates the Upgrade Marker but before it re-execs and starts up the new (post-upgrade) Agent.  If the old (pre-upgrade) Agent can communicate to the Upgrade Marker Watcher when it starts an upgrade, the Upgrade Marker Watcher can then know that it's not being rolled back even though the versions are the same.

## Checklist

## Related issues

- Relates to https://github.com/elastic/elastic-agent/pull/3827#discussion_r1408553007

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/marker_watcher.go
+++ b/internal/pkg/agent/application/upgrade/marker_watcher.go
@@ -19,12 +19,15 @@ import (
 type MarkerWatcher interface {
 	Watch() <-chan UpdateMarker
 	Run(ctx context.Context) error
+	SetUpgradeStarted()
 }
 
 type MarkerFileWatcher struct {
 	markerFilePath string
 	logger         *logger.Logger
 	updateCh       chan UpdateMarker
+
+	upgradeStarted bool
 }
 
 func newMarkerFileWatcher(upgradeMarkerFilePath string, logger *logger.Logger) MarkerWatcher {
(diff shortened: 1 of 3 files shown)


internal/pkg/agent/application/upgrade/marker_watcher.go
```diff
@@ -34,13 +37,18 @@ func newMarkerFileWatcher(upgradeMarkerFilePath string, logger *logger.Logger) M
...
+		upgradeStarted: false,
 	}
 }
 
 func (mfw *MarkerFileWatcher) Watch() <-chan UpdateMarker {
 	return mfw.updateCh
 }
 
+func (mfw *MarkerFileWatcher) SetUpgradeStarted() {
+	mfw.upgradeStarted = true
```
pchila, line 49: even if setting a boolean is probably atomic according to Go's memory model we may want to use a mutex or an atomic boolean here for correctness ?

pchila (approved):

blakerouse (approved):
Lgtm

PR #3839, opened 2023-11-29
Remove failing processors from the query for errors

belimawr:
## What does this PR do?

The `add_cloud_metadata` and some other `add_*_metadata` processors are expected to log some errors if they cannot fetch the necessary information. It is normal to find their error logs in pretty much any deployment, some examples:
 - When Docker is not installed/running `add_docker_metadata` will log some errors
 - When the Elastic-Agent is deployed in a non-cloud VM `add_cloud_metadata` will log some errors.

This commit removes all those processors from the queries for log errors, as they're expected.

## Why is it important?

It fixes some flaky test

## Checklist

~~## Author's Checklist~~

```
DEV=true SNAPSHOT=true EXTERNAL=true PACKAGES=tar.gz PLATFORMS=linux/amd64 mage -v package
SNAPSHOT=true TEST_PLATFORMS="linux/amd64" mage -v integration:single  TestLogIngestionFleetManaged 
```

~~## Related issues~~
~~## Use cases~~
~~## Screenshots~~
~~## Logs~~

## Questions to ask yourself

- ...


pkg/testing/tools/estools/elasticsearch.go
```diff
@@ -392,42 +428,11 @@ func CheckForErrorsInLogsWithContext(ctx context.Context, client elastictranspor
...
-					"must": []map[string]interface{}{
-						{
-							"match": map[string]interface{}{
-								"log.level": "error",
-							},
-						},
-						{
-							"term": map[string]interface{}{
-								"data_stream.namespace": map[string]interface{}{
-									"value": namespace,
```
AndersonQ, line 420: [Blocker]
Have you forgotten to add this part of the query, the one that selects the data stream and the error logs?

AndersonQ (changes requested):

pkg/testing/tools/estools/elasticsearch.go
AndersonQ: ah, I missed the other part because it was not changed.

AndersonQ (approved):

AndersonQ:
buildkite test this

cmacknz:
I am going to force merge this, the test failure is unrelated and this fixes several other frequent test failures.

pkg/testing/tools/estools/elasticsearch.go
```diff
@@ -375,7 +375,43 @@ func CheckForErrorsInLogs(client elastictransport.Interface, namespace string, e
...
+
+	// Some if not all of those processors are likely to always log some error
+	// just because they cannot fetch the information, like if docker is not running
+	// or the Beat is not running on Kubernetes, so we already exclude all errors from them.
+	processors := []string{
+		"add_host_metadata",
+		"add_cloud_metadata",
+		"add_docker_metadata",
+		"add_kubernetes_metadata",
+	}
```
cmacknz, line 388: Hmm I'm not sure I agree with unconditionally ignoring every error from these processors.

This would miss legitimate regressions, for example if the add_host_metadata processor were to start reporting errors when we added the FQDN feature.

I would rather us exclude specific errors.

cmacknz:
Actually the way we are fixing this might be too broad.

pkg/testing/tools/estools/elasticsearch.go
cmacknz: It looks like we are filtering the specific errors on main now

https://github.com/elastic/elastic-agent/blob/a0e49afa6788917e713dd6127799ae5199bcdbaa/testing/integration/logs_ingestion_test.go#L140-L144

Is this change still necessary?

belimawr: I wasn't sure if that was the best approach. I believe it is a bit hard to extensively exclude all possible expected errors from those loggers (or any other).

I agree it might be too broad. I'm fine with closing this PR if it has already been fixed in a different way.

belimawr:
Closing this PR as it has already been fixed by: https://github.com/elastic/elastic-agent/commit/8a8abd046a1f549557d006031223919d71314f96

Regardless of closing this PR, other tests that use `estools.CheckForErrorsInLogs` might face the same issue. We should think about a more robust way to exclude know error messages when looking from unexpected errors.

PR #3849, opened 2023-11-30
Add check for deadline set in contexts passed to Fixture.Run()

pchila:
This change adds a check on the contexts used by integration test fixture when calling `fixture.Run()`, failing the test if no deadline has been specified.

This is to avoid having the test hanging on t.Run() when the state specified in the FSM are never reached.

## What does this PR do?
Adds a check on context passed to agent fixture.Run() to ensure that we have set a deadline

## Why is it important?
Calling f.Run() with a context that does not expire could result in tests hanging if they don't reach any of the states specified in the Run() FSM

## Checklist

- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-~~agent#changelog)~~

## Author's Checklist

- [ ]

## Questions to ask yourself

- ...


cmacknz (approved):
Thanks! The tests will tell us if we missed any other instances of this

blakerouse (approved):
Looks good.

belimawr (approved):

cmacknz (approved):

PR #3884, opened 2023-12-07
Fix component control protocol to allow checkin to be chunked across multiple messages

### LINKED ISSUE #2460: Control protocol checkin payloads can exceed the gRPC maximum message size when using autodiscovery
# Intro
[Kubernetes Autodiscovery](https://www.elastic.co/guide/en/fleet/current/elastic-agent-kubernetes-autodiscovery.html) allows the agent to automatically add and update inputs to its policy as containers, pods, and nodes in a Kubernetes cluster come and go.

This works well, but when autodiscovery is configured such that it will generate inputs for each pod on a node or each node in a cluster the resulting agent policies can be so large they cause problems. For example, the officially recommended limit for the number of pods on a node is 110 but individual Kubernetes runtimes can allow more than this. [Amazon EKS allows](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt) 737 pods per node for the largest node types, and we have seen individual nodes with 400+ pods in support cases.

# Problem
We have a recent example where the agent failed in this case when a user was attempting to monitor 700+ pods. The agent logs were flooded with errors like:

```json
{"log.level":"error","@timestamp":"2023-02-13T09:00:01.944+0100","log.logger":"centralmgmt.fleet","log.origin":{"file.name":"management/manager.go","file.line":251},"message":"elastic-agent-client got error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4269853 vs. 4194304)","service.name":"filebeat","ecs.version":"1.6.0"}
```

The biggest problem caused by these large agent policies is that they begin to exceed the default size limits for the messages in the control protocol. This prevents the system from functioning by default. 

This problem also affects diagnostics, see https://github.com/elastic/elastic-agent/issues/1808. All communication between the agent and its subprocesses is affected by this problem, and it may also affect communication with Fleet server since the Fleet checkin payload contains an entry for each input.

## Solution 1: Configurable Message Size Limits
There is a way to configure the maximum message size for the Elastic Agent itself, but there is a similar limit that must be configured for each client.

https://github.com/elastic/elastic-agent/blob/b60b8b04c41f67a07958b59d18977ba0229bf233/elastic-agent.reference.yml#L111-L113

Even if we could allow changing the size limits everywhere, this solution to the problem is not automatic and requires users to experience a failure and diagnose policy size as the problem.

## Solution 2: Chunked gRPC Transfers
For a more transparent solution, we can introduce configuration chunking into the agent control protocol as described in https://github.com/elastic/elastic-agent/issues/2460#issuecomment-1525938065. This would solve the problem for the agent as it exists today.

However we are also working to implement a Kubernetes operator for the Elastic Agent, and in this case the agent policy would not be transported using the control protocol but rather stored in a Kubernetes primitive like a ConfigMap or a Secret, both of which have fixed 1 MB size limits. Changing the control protocol would not solve the problem in the case of an agent Kubernetes operator or other Kubernetes native technology.

## Solution 3: Require Components to Render Input Templates
Yet another alternative approach is to change the way we generate inputs when using autodiscovery entirely. Today each discovered input is rendered completely in the agent policy as the need for them is discovered. This is convenient as it requires no logic in each input, but problematic because the generated agent policies can be enormous.

For a single example, it is common to generate a filestream input to collect logs from each container on a node using an input like (see the [Dynamic Logs Path documentation](https://www.elastic.co/guide/en/fleet/current/conditions-based-autodiscover.html#_example_dynamic_logs_path)):

```yaml
- data_stream:
    namespace: production
  id: filestream-container-logs-b203b554-f873-4d4a-8b1e-3884282db838
  meta:
    package:
      name: kubernetes
      version: 1.31.0
  name: kubernetes-container-logs
  package_policy_id: b203b554-f873-4d4a-8b1e-3884282db838
  revision: 3
  streams:
  - data_stream:
      dataset: kubernetes.container_logs
      type: logs
    id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
    parsers:
    - container:
        format: auto
        stream: all
    paths:
    - /var/log/pods/*${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
    prospector.scanner.symlinks: true
  type: filestream
  use_output: default
```

The autodiscovery logic will further expand this configuration to add processors resulting in an even larger configuration, which is then repeated in the policy 100s of times:

<details>

```yaml
- data_stream:
    namespace: production
  id: filestream-container-logs-b203b554-f873-4d4a-8b1e-3884282db838-kubernetes-008b0185-59e6-4d41-b9ac-450d7c19f974.deployment
  meta:
    package:
      name: kubernetes
      version: 1.31.0
  name: kubernetes
  original_id: filestream-container-logs-b203b554-f873-4d4a-8b1e-3884282db838
  package_policy_id: b203b554-f873-4d4a-8b1e-3884282db838
  policy:
    revision: 5
  processors:
  - add_fields:
      fields:
        id: 9ecb5ce8b4370ac1025f158b782375f0b8efd36a5dd4b1b3ee3073c58998b37e
        image:
          name: docker/some-docker-image
        runtime: cri-o
      target: container
  - add_fields:
      fields:
        container:
          name: deployment
        labels:
          some-label: my-label
        namespace: pod-namespace
        namespace_labels:
          some-label: my-namespace-label
        namespace_uid: dcf37b8c-bd61-4d7f-97b5-19fd8cfde8e7
        node:
          hostname: node-hostname
          labels:
            beta_kubernetes_io/arch: amd64
            beta_kubernetes_io/instance-type: Standard_D16s_v3
            beta_kubernetes_io/os: linux
            failure-domain_beta_kubernetes_io/region: westeurope
            failure-domain_beta_kubernetes_io/zone: westeurope-1
            kubernetes_io/arch: amd64
            kubernetes_io/hostname: node-hostname
            kubernetes_io/os: linux
            node-role_kubernetes_io/worker: ""
            node_kubernetes_io/instance-type: Standard_D16s_v3
            topology_disk_csi_azure_com/zone: westeurope-1
            topology_kubernetes_io/region: westeurope
            topology_kubernetes_io/zone: westeurope-1
          name: node-name
          uid: 3158c8c6-ea34-4321-9755-090e4a426079
        pod:
          ip: 10.128.8.146
          name: pod-name
          uid: 008b0185-59e6-4d41-b9ac-450d7c19f974
      target: kubernetes
  revision: 3
  streams:
  - data_stream:
      dataset: kubernetes.container_logs
      type: logs
    id: kubernetes-container-logs-pod-name-pattern-9ecb5ce8b4370ac1025f158b782375f0b8efd36a5dd4b1b3ee3073c58998b37e
    parsers:
    - container:
        format: auto
        stream: all
    paths:
    - /var/log/pods/*pod-name-pattern/deployment/*.log
    prospector.scanner.symlinks: true
  type: filestream
```

</details>

Rather than having the agent render these inputs from templates itself, we could introduce the concept of a templated input directly into the control protocol.

I am imaging that we introduce a message type that contains a base template, which could similar to the unrendered agent input in the agent policy (repeated below for clarity) but in the same message type we also include the list of variables to substitute.

<details>

```yaml
- data_stream:
    namespace: production
  id: filestream-container-logs-b203b554-f873-4d4a-8b1e-3884282db838
  meta:
    package:
      name: kubernetes
      version: 1.31.0
  name: kubernetes-container-logs
  package_policy_id: b203b554-f873-4d4a-8b1e-3884282db838
  revision: 3
  streams:
  - data_stream:
      dataset: kubernetes.container_logs
      type: logs
    id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
    parsers:
    - container:
        format: auto
        stream: all
    paths:
    - /var/log/pods/*${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
    prospector.scanner.symlinks: true
  type: filestream
  use_output: default
```

</details>

This configuration require us to provide `${kubernetes.pod.name}`, `${kubernetes.pod.uid}`, and `${kubernetes.container.name}` for each pod on the node. A representative set of messages could look like:

```protobuf
message TemplatedInput {
   string InputTemplate = 1;
   repeated TemplateVariables = 2;
}

// Repeated maps aren't supported directly
message TemplateVariables {
   map<string, string> vars = 1;
}
```

This would compress the configuration down to the minimum set of information necessary to transport to each sub-process, however it would require each component support templated input rendering in their implementation language. This solution is not as easily generalizable as introducing a chunked control protocol, but it also solves the problem in the case where the agent policy needs to be stored in 

## Solution 4: Use YAML Anchors to Avoid Repetition

Similar to the approach above to avoid duplicating information in the control protocol, we could leave the control protocol as is and attempt to eliminate the duplication using YAML anchors or custom YAML syntax in the agent policy itself.

For an example, see Gitlab's documentation on simplifying configurations: https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#anchors

Those examples use anchors in combination with [map merging](https://yaml.org/type/merge.html) to attempt to perform the templating described above using the more advanced parts of the YAML syntax.

This will make the computed agent policy harder to read, and also requires that the YAML features we use be well supported in multiple implementation languages (at least Go and C++).

# Scope

Provide a recommended solution to this problem to ensure that the Elastic Agent can scale on to monitor any size of Kubernetes cluster without arbitrary internal limits. Consider solutions beyond those proposed here, each of which as their own set of pros and cons.

Keep in mind that this problem so far only affects Kubernetes deployments, and solutions that add a fixed resource cost to all use cases (enabling compression in the control protocol for example) should be avoided if possible.

blakerouse:
## What does this PR do?

Fixes component control protocol to allow checkin to be chunked across multiple messages. It is backwards compatible as well, so only clients that inform the Elastic Agent that they support chunking will chunking be used when the message is larger than that max message size.

## Why is it important?

Fixes a core issue where when running large amounts of inputs in a single component that the Elastic Agent was unable to communicate the information because the amount of data being sent between the Elastic Agent and the component was larger than the max message size.

## Checklist

- ~~[ ] I have added an integration test or an E2E test~~ (tested in a unit test)

## Author's Checklist

- [x] Enables chunking support when communicating with the Elastic Agent.

```
$ cd ../beats
$ git checkout blakerouse/agent-client-chunking
$ cd ../elastic-agent
$ mage package  # not using `EXTERNAL=true`
```

## Related issues

- Closes #2460


blakerouse:
go.mod replace will be removed once https://github.com/elastic/elastic-agent-client/pull/89 is merged

pkg/component/runtime/manager.go
```diff
@@ -658,6 +658,15 @@ func (m *Manager) CheckinV2(server proto.ElasticAgent_CheckinV2Server) error {
 		return status.Error(codes.PermissionDenied, "invalid token")
 	}
 
+	// enable chunking with the communicator if the initial checkin
+	// states that it supports chunking
+	runtime.comm.chunkingAllowed = false
+	for _, support := range initCheckin.Supports {
+		if support == proto.ConnectionSupports_CheckinChunking {
+			runtime.comm.chunkingAllowed = true
```
michalpristas, line 666: very minor: no need to continue with loop right, although it will be very fast

pkg/component/runtime/runtime_comm.go
```diff
@@ -401,3 +411,20 @@ func genServerName() (string, error) {
 	}
 	return strings.Replace(u.String(), "-", "", -1), nil
 }
+
+func sendExpectedChunked(server proto.ElasticAgent_CheckinV2Server, msg *proto.CheckinExpected, chunk bool, maxSize int) error {
```
michalpristas, line 415: s/chunk/chunkingAllowed

michalpristas (approved):

changelog/fragments/1701975889-Fix-component-control-protocol-to-allow-checkin-to-be-chunked-across-multiple-messages.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Fix component control protocol to allow checkin to be chunked across multiple messages
```
AndersonQ, line 14: Is it a fix or a feature? If we're considering a checkin too big to be sent a bug, this is a bug fix

pkg/component/runtime/manager_test.go
```diff
@@ -3481,6 +3481,164 @@ LOOP:
...
+	defer endTimer.Stop()
+LOOP:
+	for {
+		select {
+		case <-endTimer.C:
+			t.Fatalf("timed out after 30 seconds")
+		case err := <-errCh:
+			require.NoError(t, err)
+		case err := <-subErrCh:
+			require.NoError(t, err)
```
AndersonQ, line 3626: [Suggestion]
Use `Errorf` and add a message to better differentiate both error cases

pkg/component/runtime/manager_test.go
```diff
@@ -3481,6 +3481,164 @@ LOOP:
...
+						} else if unit.State == client.UnitStateHealthy {
+							healthyCount += 1
+						} else if unit.State == client.UnitStateStopped {
+							stoppedCount += 1
+						} else if unit.State == client.UnitStateStarting {
+							// acceptable
+						} else {
+							// unknown state that should not have occurred
+							subErrCh <- fmt.Errorf("unit reported unexpected state: %v", unit.State)
+						}
```
AndersonQ, line 3592: [Suggestion]
Use a `switch` statement instead of so mane `else if`

blakerouse: This is copied from the other tests, so I am going to keep it just so its consistent.

changelog/fragments/1701975889-Fix-component-control-protocol-to-allow-checkin-to-be-chunked-across-multiple-messages.yaml
blakerouse: Updated to bug-fix.

pkg/component/runtime/manager.go
blakerouse: Added `break`

pkg/component/runtime/manager_test.go
blakerouse: Each error that comes over those channels are specific and its clear. Adding more context would be redundant.

pkg/component/runtime/runtime_comm.go
blakerouse: Updated.

blakerouse:
go.mod replace has been removed now that v7.6.0 of the elastic-agent-client has been released.

pkg/component/runtime/manager.go
```diff
@@ -658,6 +658,16 @@ func (m *Manager) CheckinV2(server proto.ElasticAgent_CheckinV2Server) error {
...
```
cmacknz, line 666: Can we get this to be included in the `state.yaml` or some other part of the diagnostics so we can see if it is enabled or not when debugging?

pkg/component/runtime/manager_test.go
```diff
@@ -3481,6 +3481,164 @@ LOOP:
...
 }
 
+func TestManager_FakeInput_Chunk(t *testing.T) {
+	testPaths(t)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	grpcConfig := configuration.DefaultGRPCConfig()
+	grpcConfig.MaxMsgSize = 120 // very small size to test chunking
```
cmacknz, line 3491: This affects (and therefore tests) the expected direction correct?

Is there a way to force chunking on the checkin observed path as well?

pkg/component/runtime/manager_test.go
```diff
@@ -3481,6 +3481,164 @@ LOOP:
...
+						} else if unit.State == client.UnitStateStopped {
+							stoppedCount += 1
+						} else if unit.State == client.UnitStateStarting {
+							// acceptable
+						} else {
+							// unknown state that should not have occurred
+							subErrCh <- fmt.Errorf("unit reported unexpected state: %v", unit.State)
+						}
+					}
+					if healthyCount == 4 {
```
cmacknz, line 3594: Is there a way that you can assert that these units were received via the chunking protocol?

If there were a bug that caused the grpcConfig above not to apply, then this test would still pass and we wouldn't know if it were testing the chunking protocol or not.

pkg/component/runtime/manager.go
blakerouse: I will look at how difficult that is to propagate to get that information exposed.

pkg/component/runtime/manager_test.go
blakerouse: This affects both directions as the `MaxMsgSize` is sent to the client over stdin, so it will also use that max message size for the observed path.

pkg/component/runtime/manager_test.go
blakerouse: I understand the reasoning for wanting to confirm this, but I believe we would really be adding lots of extras in the code path to test that. Its rather deep into the communication had basically handled all transparently by the elastic-agent-client library.

I think to really have something to verify that it would require modification to the elastic-agent-client and they would have to be public hooks, or callbacks to track that it was sent/receive chunked.

pkg/component/runtime/manager_test.go
cmacknz: Ah I missed that part of this.

pkg/component/runtime/manager_test.go
cmacknz: I get that it's more work, but I don't like that this test could still pass without the feature actually doing anything.

If we added atomic stats counters to the client, and added some for the number of chunked messages sent/received, then you'd probably be able to confirm it was working pretty easily.

This would generally be useful to expose via the agent diagnostics and HTTP interface one day as well.

blakerouse: Thats exploding this bug fix into adding metrics to the elastic-agent-client, seems very unrelated and unreasonable.

Another more isolated option is make the check-in messages always larger than the default GRPC configuration message size, and remove the setting of that option in the test. That way if that value is changed the test will always create a message size that is larger than that amount to ensure that it always still works. The test can also ensure the minimum size is 4MB which is GRPC default, not Agent default to ensure that if the default is ever broken it still works.

cmacknz: That works, it's the least disruptive to the rest of the PR.

I still think it would be nice to have a way to tell that chunking is happening or has happened so we can isolate possible bugs to it, but I agree that is more work.

cmacknz:
One other thought to minimize risk when we release this, and give us an option to isolate whether chunking is causing problems: can we add a configuration option in the policy to disable chunking? Perhaps by having the agent declare that it doesn't support chunking?

pkg/component/runtime/manager.go
blakerouse: How much do we really want to propagate this information? At the moment this runtime communication is separate from the component state and no lock is required to ensure its consistent. I have looked into this and its rather difficult to propagate this information in a manner that it can be exposed in the diagnostics output.

Seems like a large amount of work for some information of a true/false where it should basically always become the default as the beats/agent versions are fixed to the same release cycle.

cmacknz: I agree it is not worth introducing complexity or new locks for.

I still think we want a way to tell when this functionality is taking effect, my primary concern is not having a way to know it is causing us problems. Perhaps some well placed log lines can do this job effectively.

If we get a configuration flag to disable the chunking that gives us one way to isolate it as the cause of a problem, but without some observability into if it's actually happening we will still be guessing as to when to try that.

blakerouse: I added log line for when it is enabled as well as a specific log line when it could be enabled, but it is explicitly disabled.

pkg/component/runtime/manager_test.go
blakerouse: Added to the PR. It found an issue with the chunking function, as it was not taking into the account of overhead for each element in an repeated structure in. https://github.com/elastic/elastic-agent-client/issues/94 will need to be fixed before this PR will pass the unit test.

blakerouse:
> One other thought to minimize risk when we release this, and give us an option to isolate whether chunking is causing problems: can we add a configuration option in the policy to disable chunking? Perhaps by having the agent declare that it doesn't support chunking?

I added this.

cmacknz:
Latest changes LGTM, will approve once we get passing tests.

pkg/component/runtime/manager_test.go
blakerouse: The fix in elastic-agent-client has been released and I have updated the PR to use it.

pierrehilbert:
The CI failure seems unrelated (https://github.com/elastic/elastic-agent/issues/4076)

AndersonQ (approved):

PR #3892, opened 2023-12-08
Add missing contexts (and therefore timeouts) to integration test code

cmacknz:
We are still seeing integration tests time out after 2+ hours. This appears to be because we are making network requests without timeouts in several instances, with the root cause being that we do not propagate contexts through the calls even though we are using the `WithContext` variant of the network call.

For example https://buildkite.com/elastic/elastic-agent/builds/5302#018c2582-6e37-488c-a13f-4964c8fc51ea got stuck in a new place. It timed out in https://github.com/elastic/elastic-agent/blob/33ad31eae14026fa1151c7b42ee3cf44ead00b9b/pkg/testing/tools/tools.go#L126-L132
That eventually ends up in https://github.com/elastic/elastic-agent/blob/33ad31eae14026fa1151c7b42ee3cf44ead00b9b/pkg/testing/tools/fleettools/fleet.go#L18-L22

Which hard codes context.Background. 

I searched for `context.Background` in pkg/testing and removed every instance of it that was not already using it as an input to `context.WithTimeout`.

This will not fix the actual test timeouts, but it should tell us which specific parts of the tests are timing out instead of having to root cause it through go test execution panics.

### DIFF
--- a/pkg/testing/ess/serverless_test.go
+++ b/pkg/testing/ess/serverless_test.go
@@ -16,6 +16,9 @@ import (
 )
 
 func TestProvisionGetRegions(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*240)
+	defer cancel()
+
 	_ = logp.DevelopmentSetup()
 	key, found, err := GetESSAPIKey()
 	if !found {
@@ -29,13 +32,16 @@ func TestProvisionGetRegions(t *testing.T) {
 		cfg: cfg,
 		log: &defaultLogger{wrapped: logp.L()},
 	}
-	err = prov.CheckCloudRegion()
+	err = prov.CheckCloudRegion(ctx)
 	require.NoError(t, err)
 	require.NotEqual(t, "bad-region-ID", prov.cfg.Region)
 
 }
 
 func TestStackProvisioner(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	defer cancel()
+
 	_ = logp.DevelopmentSetup()
 	key, found, err := GetESSAPIKey()
 	if !found {
@@ -45,12 +51,10 @@ func TestStackProvisioner(t *testing.T) {
 	require.True(t, found)
 
 	cfg := ProvisionerConfig{Region: "aws-eu-west-1", APIKey: key}
-	provClient, err := NewServerlessProvisioner(cfg)
+	provClient, err := NewServerlessProvisioner(ctx, cfg)
 	require.NoError(t, err)
 	request := runner.StackRequest{ID: "stack-test-one", Version: "8.9.0"}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
-	defer cancel()
 	stack, err := provClient.Create(ctx, request)
 	require.NoError(t, err)
 	t.Logf("got results:")
--- a/pkg/testing/ess/serverless_test.go
+++ b/pkg/testing/ess/serverless_test.go
@@ -16,7 +16,7 @@ import (
 )
 
 func TestProvisionGetRegions(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*240)
+	ctx, cancel := context.WithTimeout(context.Background(), 4*time.Minute)
 	defer cancel()
 
 	_ = logp.DevelopmentSetup()
@@ -39,7 +39,7 @@ func TestProvisionGetRegions(t *testing.T) {
 }
 
 func TestStackProvisioner(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
 	_ = logp.DevelopmentSetup()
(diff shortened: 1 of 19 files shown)


cmacknz:
This is a follow up to https://github.com/elastic/elastic-agent/pull/3849 which added contexts with timeouts to every integration test.

blakerouse (approved):
Looks good to me.

Very mechanical change.

pkg/testing/ess/serverless_test.go
```diff
@@ -16,6 +16,9 @@ import (
 )
 
 func TestProvisionGetRegions(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*240)
```
belimawr, line 19: [Question]

Why 240s instead of 4min?
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*4)
```

belimawr (approved):

pkg/testing/ess/serverless_test.go
cmacknz: No idea, copied from another part of this file. I switched everything to minutes for consistency in here.

cmacknz:
>     upgrader.go:202: Upgrading from version "8.12.0-SNAPSHOT" to version "8.11.3"
>     fixture.go:520: >> running binary with: [/opt/Elastic/Agent/elastic-agent upgrade 8.11.3 --pgp abcDEFgGIN PGP PUBLIC KEY BLOCK-----
>         Version: GnuPG v2.0.14 (GNU/Linux)

>     upgrade_gpg_test.go:78: 
>         	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_gpg_test.go:78
>         	Error:      	Received unexpected error:
>         	            	failed to start agent upgrade to version "8.11.3": exit status 1
>         	            	Error: Failed trigger upgrade of daemon: failed download of agent binary: unable to download package: 3 errors occurred:
>         	            		* package '/opt/Elastic/Agent/data/elastic-agent-769b15/downloads/elastic-agent-8.11.3-linux-x86_64.tar.gz' not found: open /opt/Elastic/Agent/data/elastic-agent-769b15/downloads/elastic-agent-8.11.3-linux-x86_64.tar.gz: no such file or directory
>         	            		* call to 'https://snapshots.elastic.co/8.12.0-nobcbsl4/downloads/beats/elastic-agent/elastic-agent-8.11.3-linux-x86_64.tar.gz' returned unsuccessful status code: 404
>         	            		* call to 'https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.11.3-linux-x86_64.tar.gz' returned unsuccessful status code: 404

Now hitting this, which is expected because 8.11.3 does indeed not exist. Hmm.

cmacknz:
This problem with 8.11.3 is going to fix itself shortly. Long term we need to enable automatically upgrading to builds hosted on staging.

cmacknz:
The latest test failure here still has the TestAPMConfig panicing after hitting the 2 hour timeout, but the stack trace now more clearly shows we are stuck calling `<-proc.Wait()`. The goroutine trace is similarly full of goroutines also stuck waiting on Windows processes.

Something deeper is going on here related to our ability to kill and wait on processes on Windows.

I specifically see us blocked on `syscall.WaitForSingleObject` in a few places.

```

T | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): goroutine 351 [syscall, 117 minutes, locked to thread]:
-- | --
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): syscall.SyscallN(0x7ffedf703ef0?, {0xc0002ede90?, 0x3?, 0x0?})
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): C:/Program Files/Go/src/runtime/syscall_windows.go:557 +0x109
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): syscall.Syscall(0x1?, 0x0?, 0x0?, 0x0?, 0x0?)
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): C:/Program Files/Go/src/runtime/syscall_windows.go:495 +0x3b
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): syscall.WaitForSingleObject(0xc0000a47d0?, 0xffffffff)
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): C:/Program Files/Go/src/syscall/zsyscall_windows.go:1145 +0x65
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): os.(*Process).wait(0xc0006a2000)
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): C:/Program Files/Go/src/os/exec_windows.go:18 +0x65
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): os.(*Process).Wait(...)
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): C:/Program Files/Go/src/os/exec.go:132
  | 2023-12-14 19:12:30 EST | >>> (windows-amd64-2022-default) Test output (non-sudo) (stdout): github.com/elastic/elastic-agent/pkg/core/process.(*Info).Wait.func1()
```

cmacknz:
I think the most sensible thing to do right now is merge this since it makes the failure more obvious, and keep investigating the root cause separately.

I'm also considering disabling the affected test while we investigate this since we now know exactly what is causing the problem.

PR #3895, opened 2023-12-11
Add PR guidelines to the CONTRIBUTING.md guide.

cmacknz:
Add guidelines for how to review PRs as discussed in one of our recent team meetings. These are copied almost verbatim from the ones @ycombinator wrote in https://github.com/elastic/logstash/blob/main/CONTRIBUTING.md#as-a-reviewee-ie-author-of-a-pr. 

Note that I omitted a specific PR size limit. The guiding principle should be "can I understand this?" regardless of size, not necessary a strict size limit. If we find this does not work we can add and enforce a line limit.

I also updated the contributing guide with relevant links to Slack and Discuss.


cmacknz:
Link to rendered CONTRIBUTING.md: https://github.com/cmacknz/elastic-agent/blob/update-contributing-guide/CONTRIBUTING.md

CONTRIBUTING.md
```diff
@@ -1,17 +1,61 @@
...
+* You can ask a question in the [forum](https://discuss.elastic.co/c/elastic-stack/elastic-agent).
+* You are welcome to join Elastic Stack Community slack (https://elasticstack.slack.com) and ask for help on the #elastic-agent channel.
+
+## Have an Idea or Feature Request?
+
+* File an issue on [GitHub](https://github.com/elastic/elastic-agent/issues). Please remember that GitHub is used only for bugs and feature requests. If you have a general question, the [forum](https://discuss.elastic.co/c/elastic-stack/elastic-agent) or Elastic Stack Community slack (https://elasticstack.slack.com) is the best place to ask.
+
+## Something Not Working? Found a Bug?
+
+If you think you found a bug, it probably is a bug. File it in [GitHub](https://github.com/elastic/elastic-agent/issues)
```
pierrehilbert, line 16:
```suggestion
If you think you found a bug, it probably is a bug. Fill it in [GitHub](https://github.com/elastic/elastic-agent/issues)
```

pierrehilbert (approved):
Thanks @cmacknz for putting this by written, it will be really useful!

pchila (approved):

CONTRIBUTING.md
belimawr: Question: Do we want to direct people to [Discuss](https://discuss.elastic.co/) like we do for Beats instead of directly opening an issue here?

Regarding the wording, I believe `File` reads better in this case.

CONTRIBUTING.md
```diff
@@ -1,17 +1,61 @@
...
+* You can ask a question in the [forum](https://discuss.elastic.co/c/elastic-stack/elastic-agent).
+* You are welcome to join Elastic Stack Community slack (https://elasticstack.slack.com) and ask for help on the #elastic-agent channel.
+
+## Have an Idea or Feature Request?
+
+* File an issue on [GitHub](https://github.com/elastic/elastic-agent/issues). Please remember that GitHub is used only for bugs and feature requests. If you have a general question, the [forum](https://discuss.elastic.co/c/elastic-stack/elastic-agent) or Elastic Stack Community slack (https://elasticstack.slack.com) is the best place to ask.
+
+## Something Not Working? Found a Bug?
+
+If you think you found a bug, it probably is a bug. Fill it in [GitHub](https://github.com/elastic/elastic-agent/issues)
```
belimawr, line 16: [Question]

Do we want to direct users to Discuss like we do in Beats instead of directly opening an issue on GitHub?

belimawr (changes requested):

michalpristas (approved):
💙

AndersonQ (changes requested):
:heart: but please fix the `its -> it's`

AndersonQ:
buildkite test this

CONTRIBUTING.md
cmacknz: We don't generally have a problem with users filing issues for things that aren't bugs or feature requests in this repository, so I don't think there's a need to introduce an extra step until there is a problem to solve.

blakerouse (approved):
Easy to read and very clear. Like the improvement!

cmacknz:
@AndersonQ @belimawr ping since you've both requested changes that have been addressed.

cmacknz:
Just going to merge this one as it is documentation only (test failures are irrelevant).

PR #3896, opened 2023-12-12
Always use the latest snapshot for each minor.

cmacknz:


cmacknz:
Closing a better solution is to allow us to download BCs from staging rather than trying to infer if an artifact is on staging and avoiding it.

PR #3909, opened 2023-12-13
[Unprivileged] Adjust the control socket path add `--path.socket` option

### LINKED ISSUE #3840: [Unprivileged] `/run/elastic-agent` removed on reboot when `/run` is mounted as a tmpfs
With the latest addition of `--unprivileged` installation mode the Elastic Agent creates a `/run/elastic-agent` directory that is owned with 'elastic-agent:elastic-agent'. In the case that the host does not have `/run` mounted as a tmpfs this works fine, in the case that `/run` is mount as a tmpfs the Elastic Agent will work until the next restart of the host. Upon next restart of the host that directory will be removed and the spawned Elastic Agent will not have the correct permissions to re-create that directory.

The only thing placed into this directory is the `/run/elastic-agent/elastic-agent.sock`.

To fix this issue the following need to occur:

1. Move the `elastic-agent.sock` to the top-level `/opt/Elastic/Agent` or the `--base-path` path.
2. On installation without `--unprivileged` create a symlink from `/run/elastic-agent.sock` to the new top-level path.

blakerouse:
## What does this PR do?

Adjusts the control socket path to always be inside of the top level of the Elastic Agent installation directory. Backwards compatibility creates a symlink from original `/run/elastic-agent.sock` to the new path, as long as the Elastic Agent is running as root.

A new `--path.socket` is also added to allow users to adjust this path for running and communicating with an Elastic Agent.

## Why is it important?

In unprivileged mode the control socket was moved to `/run/elastic-agent/elastic-agent.sock`, this doesn't work when the run directory is mounted with a tmpfs (which is highly possible) because on reboot the elastic-agent directory inside of run will be deleted and root will no longer be available to create the directory. Moving the control socket into the topPath ensures that the Elastic Agent will always have the permissions to create the socket.

## Checklist

## Related issues

- Closes #3840


internal/pkg/agent/application/paths/paths_unix.go
```diff
@@ -0,0 +1,27 @@
...
+
+func initialControlSocketPath(topPath string) string {
+	socketPath := filepath.Join(topPath, ControlSocketName)
+	unixSocket := fmt.Sprintf("unix://%s", socketPath)
+	if len(unixSocket) < 104 {
+		return unixSocket
+	}
+	// place in global /tmp to ensure that its small enough to fit; current path is way to long
+	// for it to be used, but needs to be unique per Agent (in the case that multiple are running)
+	return fmt.Sprintf(`unix:///tmp/elastic-agent/%x.sock`, sha256.Sum256([]byte(socketPath)))
```
cmacknz, line 23: Can you use the code that already exists for this?

https://github.com/elastic/elastic-agent/blob/1410d2a171015456174663ac64b06d26b8ad4509/pkg/utils/socket_unix.go#L21-L44

Same goes for Windows.

changelog/fragments/1702501610-Adjust-control-socket-path-into-the-installation-directory.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Adjust control socket path into the installation directory
```
AndersonQ, line 14: Perhaps "move" is better
```suggestion
summary: Move control socket path to the installation directory
```

internal/pkg/agent/cmd/run.go
```diff
@@ -275,6 +283,24 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
 	}
 	defer control.Stop()
 
+	// create symlink from /run/elastic-agent.sock to `paths.ControlSocket()` when running as root
+	// this provides backwards compatibility as the control socket was moved with the addition of --unprivileged
+	// option during installation
+	if isRoot && paths.RunningInstalled() && paths.ControlSocketRunSymlink != "" {
+		socketPath := strings.TrimPrefix(paths.ControlSocket(), "unix://")
+		socketLog := controlLog.With("path", socketPath).With("link", paths.ControlSocketRunSymlink)
+		_ = os.Remove(paths.ControlSocketRunSymlink) // ensure it doesn't exist before creating the symlink
```
AndersonQ, line 292: [Blocker]
If it fails for some other reason that it does not exist?

pkg/control/addr.go
```diff
@@ -2,37 +2,47 @@
...
+	}
+	unixSocket := fmt.Sprintf("unix://%s", socketPath)
+	if len(unixSocket) < 104 {
+		// small enough to fit
+		return unixSocket, nil
 	}
 	// place in global /tmp to ensure that its small enough to fit; current path is way to long
 	// for it to be used, but needs to be unique per Agent (in the case that multiple are running)
-	return fmt.Sprintf(`unix:///tmp/elastic-agent/%x.sock`, sha256.Sum256([]byte(path)))
+	return fmt.Sprintf(`unix:///tmp/elastic-agent/%x.sock`, sha256.Sum256([]byte(socketPath))), nil
```
AndersonQ, line 47: [Blocker]
It's replicating the logic from `initialControlSocketPath`. Shouldn't it be extracted to a function and both use this single function? With the logic duplicated it's at least risking a bug if someone changes one but not the other

AndersonQ (changes requested):

pkg/testing/fixture_install.go
```diff
@@ -116,9 +116,15 @@ func (f *Fixture) Install(ctx context.Context, installOpts *InstallOpts, opts ..
...
-		socketPath = paths.ControlSocketUnprivilegedPath
+	socketPath := fmt.Sprintf("unix://%s", paths.ControlSocketRunSymlink) // use symlink as that works for all versions
+	if runtime.GOOS == "windows" {
+		// Windows uses a fixed named pipe, that is always the same.
+		// It is the same even running in unprivileged mode.
+		socketPath = `\\.\pipe\elastic-agent-system`
+	} else if installOpts.Unprivileged {
+		// Unprivileged versions move the socket to inside the installed directory
+		// of the Elastic Agent.
+		socketPath = fmt.Sprintf("unix://%s", filepath.Join(f.workDir, paths.ControlSocketName))
```
cmacknz, line 127: Can we import the paths from the agent code instead of duplicating them here?

blakerouse: I was able to update this to use the paths module.

pkg/control/addr.go
blakerouse: I extracted the function to a single location.

internal/pkg/agent/cmd/run.go
blakerouse: I updated this to log if it gets another error other than not existing.

blakerouse:
I was able to finally get this cleaned up. The code for the control socket paths was really all over the place and there was issues with the upgrade path and issues when the `.installed` marker being created to late during installation before the Elastic Agent was restarted.

Anyway this is ready for another review @AndersonQ and @cmacknz

internal/pkg/agent/application/paths/paths_unix.go
blakerouse: Updated.

AndersonQ (changes requested):

pkg/control/v2/server/server.go
```diff
@@ -85,6 +85,7 @@ func (s *Server) Start() error {
 		s.logger.Errorf("unable to create listener: %s", err)
 		return err
 	}
+	s.logger.With("address", control.Address()).Infof("GRPC control socket listening at %s", control.Address())
```
cmacknz, line 88: Doesn't this log the control address twice? Once in the message and once with the `address` key?

internal/pkg/agent/cmd/run.go
```diff
@@ -287,6 +295,27 @@ func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override cf
 	}
 	defer control.Stop()
 
+	// create symlink from /run/elastic-agent.sock to `paths.ControlSocket()` when running as root
+	// this provides backwards compatibility as the control socket was moved with the addition of --unprivileged
+	// option during installation
+	if isRoot && paths.RunningInstalled() && paths.ControlSocketRunSymlink != "" {
```
cmacknz, line 301: Shouldn't there be an explicit platform check here, or is it implicit in `ControlSocketRunSymlink` being the empty string? If that is the case I would add a comment explaining that this isn't necessary on Windows since that isn't obvious without the context of reading this entire at once.

internal/pkg/agent/cmd/run.go
```diff
@@ -287,6 +295,27 @@ func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override cf
...
+			socketLog.Errorf("Failed to remove existing control socket symlink %s: %s", paths.ControlSocketRunSymlink, err)
+		}
+		if err := os.Symlink(socketPath, paths.ControlSocketRunSymlink); err != nil {
+			socketLog.Errorf("Failed to create control socket symlink %s -> %s: %s", paths.ControlSocketRunSymlink, socketPath, err)
+		} else {
+			socketLog.Infof("Created control socket symlink %s -> %s; allowing unix://%s connection", paths.ControlSocketRunSymlink, socketPath, paths.ControlSocketRunSymlink)
+		}
+		defer func() {
+			// delete the symlink on exit; ignore the error
+			_ = os.Remove(paths.ControlSocketRunSymlink)
```
cmacknz, line 315: Would it be valuable to log the error even if it is still ignored?

internal/pkg/agent/install/install_windows.go
```diff
@@ -46,8 +46,7 @@ func postInstall(topPath string) error {
 	return nil
 }
 
-// createSocketDir creates the socket directory.
-func createSocketDir(ownership utils.FileOwner) error {
-	// doesn't do anything on windows, no directory is needed.
+func fixInstallMarkerPermissions(markerFilePath string, ownership utils.FileOwner) error {
+	// TODO(blakerouse): Fix the market permissions on Windows.
```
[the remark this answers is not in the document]
blakerouse: That is just because the function was moved, it already existed and is in main just as you see it in this PR. The typo is also in main, I do not want to inflate this PR with un-related changes because a function had to be moved.

I will not be fixing this in this PR. Follow up PR will finish this TODO when Windows --unprivileged is implemented, which it is not.

internal/pkg/agent/application/paths/common.go
```diff
@@ -14,15 +14,30 @@ import (
...
 	// AgentLockFileName is the name of the overall Elastic Agent file lock.
 	AgentLockFileName = "agent.lock"
-	tempSubdir        = "tmp"
-	tempSubdirPerms   = 0o770
+
+	// ControlSocketName is the control socket name.
+	ControlSocketName = "elastic-agent.sock"
+
+	// WindowsControlSocketInstalledPath is the control socket path used when installed on Windows.
+	WindowsControlSocketInstalledPath = `npipe:///elastic-agent-system`
```
cmacknz, line 31: Is this equivalent to the old name `\\.\pipe\elastic-agent-system`? My Google searches haven't given me an obvious answer yet.

cmacknz (commented):
I believe the standalone upgrade test from 7.17.x to the current version covers this code path, correct?

https://github.com/elastic/elastic-agent/blob/3fe6d3224b5870269c278cd87a9fa18fb1cb7fda/testing/integration/upgrade_standalone_test.go#L34-L36

Just want to double check we don't need a new test here.

internal/pkg/agent/install/install_windows.go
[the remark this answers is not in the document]
AndersonQ: ok

internal/pkg/agent/cmd/run.go
AndersonQ: I agree it'd be good to log it

AndersonQ (approved):

pkg/control/v2/server/server.go
blakerouse: Yes it does. I really want it to have the address key in the log line, but also having it in the message is nice. I don't think it hurts to have it twice.

internal/pkg/agent/cmd/run.go
blakerouse: Yes on Windows the `paths.ControlSocketRunSymlink` is `""`. I will add a comment.

internal/pkg/agent/application/paths/common.go
blakerouse: Yes it is. Look at our `npipe.TransformString`, we use `npipe://` inside of `SocketURLWithFallback` which you asked me to use. So this makes it consistent.

internal/pkg/agent/cmd/run.go
blakerouse: Done.

internal/pkg/agent/cmd/run.go
blakerouse: Added log.

blakerouse:
> I believe the standalone upgrade test from 7.17.x to the current version covers this code path, correct?
> 
> https://github.com/elastic/elastic-agent/blob/3fe6d3224b5870269c278cd87a9fa18fb1cb7fda/testing/integration/upgrade_standalone_test.go#L34-L36
> 
> Just want to double check we don't need a new test here.

Correct this is well tested by the integration testing suite.

cmacknz (approved):
Thanks. I sanity checked this on my Mac and it does what I expect it to. Upgraded from a version without this change to this version.

```
ls -la /var/run
...
lrwxr-xr-x   1 root         daemon         41 Jan  5 14:31 elastic-agent.sock -> /Library/Elastic/Agent/elastic-agent.sock
...
```

PR #3916, opened 2023-12-14
download current agent version to upgrade to

AndersonQ:
## What does this PR do?

Fixes the `TestFleetAirGappedUpgrade` which was using the latest snapshot, but during minor releases the stack isn't available on the same version. Therefore it should use the "current" version which is adjusted so the tests wont fail during minor releases.

## Why is it important?

The failing test prevent PRs from being merged

## Checklist

## Related issues

 - N/A

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/upgrade_fleet_test.go
+++ b/testing/integration/upgrade_fleet_test.go
@@ -28,7 +28,6 @@ import (
 	"github.com/elastic/elastic-agent-libs/kibana"
 	atesting "github.com/elastic/elastic-agent/pkg/testing"
 	"github.com/elastic/elastic-agent/pkg/testing/define"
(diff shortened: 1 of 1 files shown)


ycombinator (approved):
LGTM.  Thanks for fixing this!

pchila (commented):
QQ: we are upgrading from `define.Version()` to `define.Version()`: even if the starting version is the one we packaged from the commit and the end version is coming from the artifact API substitute (this is air gapped after all) does the upgrade work for both SNAPSHOT and release versions ?

pchila (approved):

michalpristas (approved):

AndersonQ:
buildkite test this

pierrehilbert:
Thanks for fixing this!

PR #3924, opened 2023-12-18
Fix VCS errors in daily serverless tests 

fearful-symmetry:
This fixes an issue where the tests were failing with an `error obtaining VCS status: exit status 128` message. Turns out that something in a newer go version doesn't like our `--filter=tree:0` flag we were passing to git.

Technically that's the only part of this required to fix things, the shuffling around of directory setup is to deal with the _other_ circumstances where we might get the same VCS error.

### DIFF
--- a/.buildkite/scripts/steps/beats_tests.sh
+++ b/.buildkite/scripts/steps/beats_tests.sh
@@ -17,7 +17,9 @@ run_test_for_beat(){
     
     #build
     export WORKSPACE="/tmp/beats-build/beats/x-pack/${beat_name}"
+    pushd $WORKSPACE
     SNAPSHOT=true PLATFORMS=linux/amd64 PACKAGES=tar.gz,zip mage package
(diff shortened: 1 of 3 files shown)


jlind23 (approved):

.buildkite/scripts/steps/beats_tests.sh
```diff
@@ -33,11 +38,13 @@ run_test_for_beat(){
...
-mkdir -p build
-cd build
+# mkdir -p build
+# cd build
+mkdir -p /tmp/beats-build
+pushd /tmp/beats-build
 
-git clone --filter=tree:0 git@github.com:elastic/beats.git
-cd ..
+git clone git@github.com:elastic/beats.git
```
alexsapran, line 46: Might help if we do shallow clone

```suggestion 
git clone --depth=1 git@github.com:elastic/beats.git
```

PR #3929, opened 2023-12-19
Move FQDN tests to its own group

blakerouse:
## What does this PR do?

Move FQDN test into their own group as the tests manipulate the instances /etc/hosts file.

## Why is it important?

The integration test modifies the `/etc/hosts` file, and should be in its own group. This ensures that if the tests fails or makes a mistake that it doesn't affect other tests.


pierrehilbert (approved):
Thx

blakerouse:
buildkite test this

pierrehilbert:
Unrelated failure (that @AndersonQ is investigating), I'm merging.

PR #3932, opened 2023-12-19
Skip APM propagation test on Windows.

cmacknz:
This test causes the Windows tests to hang for 2 hours until they time out. While we investigate the root cause skip the test. The test will still run on Linux.


ycombinator (approved):

blakerouse (approved):

cmacknz:
buildkite test it

PR #3935, opened 2023-12-20
Allow opentelemetry dependency updates

michalpristas:
This PR updates dependabot rules and allows updating dependencies from
- `go.opentelemetry.io/*`
- `github.com/open-telemetry/*`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,8 @@ updates:
     allow:
       # Only update internal dependencies for now while we evaluate this workflow.
       - dependency-name: "github.com/elastic/*"
+      - dependency-name: "go.opentelemetry.io/*"
+      - dependency-name: "github.com/open-telemetry/*"
     reviewers:
       - "elastic/elastic-agent-control-plane"
     open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -12,6 +12,8 @@ updates:
     allow:
       # Only update internal dependencies for now while we evaluate this workflow.
```
ycombinator, line 13: Remove this comment since we'll be tracking non-internal dependencies now.

.github/dependabot.yml
```diff
@@ -12,6 +12,8 @@ updates:
...
       - dependency-name: "github.com/elastic/*"
+      - dependency-name: "go.opentelemetry.io/*"
+      - dependency-name: "github.com/open-telemetry/*"
```
ycombinator, line 16: Should we bit a bit more restrictive, at least for now?
```suggestion
      - dependency-name: "github.com/open-telemetry/opentelemetry-collector-contrib/*"
```

michalpristas: i have no issues with being more strict, updated first one as well

joshdover:
Will we have any issues if dependencies aren't updated in sync? For instance if `go.opentelemetry.io/collector/otelcol` gets bumped in one PR and `github.com/open-telemetry/opentelemetry-collector-contrib/receiver/filelogreceiver` in a separate one, could things be broken in hard-to-detect ways if there's an API change to the components API?

I think we can use the `groups` feature to avoid this scenario: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

michalpristas:
i added grouping this should help as long as versions across repos are synced. if one is released upfront grouping won't guard against that

joshdover (approved):

cmacknz (approved):

PR #3939, opened 2023-12-20
Fix issue on Windows for creating a copy of the install directory

blakerouse:
## What does this PR do?

Fixes issue where the integration tests would fail to zip the installation directory because the filename contained a `:` which is reserved on Windows.

## Why is it important?

Without this change when the diagnostics cannot be collected due to an error, then the zip of the installation directory cannot also be collected.

## Checklist


cmacknz (approved):

pierrehilbert:
Unrelated failure (that @AndersonQ is investigating), I'm merging.

Issue #3936, opened 2023-12-20
Kafka output error is displayed for Elastic Defend on 8.12.0 BC2.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.12.0 BC2
BUILD: 69899
COMMIT: 15a6cc8236b4828b97da733746ec36bd33f03bba
Artifact Link: https://staging.elastic.co/8.12.0-f6277726/downloads/beats/elastic-agent/elastic-agent-8.12.0-windows-x86_64.zip
```

**Host OS:**  Windows

**Preconditions:**
1. 8.12.0 BC2 Kibana cloud environment should be available.
2. 8.12.0 Agent should be installed using agent policy having Elastic Defend integration.
3. Kafka broker should be setup.

**Steps to reproduce:**
1. Select output for integrations as Kafka.
2. Observe agent gets unhealthy and observe output not supported error for Defend integration.

**Validated as working fine under:**
https://github.com/elastic/endpoint-dev/issues/13192#issuecomment-1833616128

**Screenshot:**
![image](https://github.com/elastic/elastic-agent/assets/77374876/b66b0c47-75e5-47ca-a048-78ca9ed4737c)

**Expected Result:**
As per https://github.com/elastic/endpoint-dev/issues/13192 Kafka output should be supported for Elastic Defend on 8.12.0 BC2.

**Logs:**
[elastic-agent-diagnostics-2023-12-20T11-32-54Z-00.zip](https://github.com/elastic/elastic-agent/files/13726758/elastic-agent-diagnostics-2023-12-20T11-32-54Z-00.zip)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

pierrehilbert:
@nfritts Keep me honest here but it should be supported in 8.12 right?

nfritts:
Yeah, Kafka should be supported in 8.12

intxgo:
Hey Agent team, something werid happened with the diagnostics.zip. There's so many empty directories inside agent's log dir, but no `services` dir with Endpoint logs. Also the `components/endpoint*` diagnostics didn't succeed.

cmacknz:
`output not supported` is coming from the agent and it is refusing to start endpoint because the specification file says it doesn't support Kafka on Windows or MacOS.

Fix is in https://github.com/elastic/elastic-agent/pull/3941

[...]

PR #3941, opened 2023-12-20
Endpoint outputs should be the same on Windows, Mac, and Linux

cmacknz:
- Closes https://github.com/elastic/elastic-agent/issues/3936

Endpoint wasn't using the anchored outputs section from its Linux definition for Windows and Mac, so it wasn't allowing use of Kafka when it should be.


blakerouse (approved):

cmacknz:
Test failures are unrelated, force merging.

nfritts:
This is already merged but it looks good from the endpoint side.

Issue #3947, opened 2023-12-21
[8.12] Upgrades being incorrectly rolled back because the agent cannot parse the upgrade marker


cmacknz:
Primarily observed on Linux. This is the error we are seeing:

```
Error: error checking for and handling upgrade: unable to load upgrade marker to check if Agent is being upgraded: yaml: line 17: mapping values are not allowed in this context
```

# How to reproduce
Run the Fleet managed upgrade test:
```
AGENT_KEEP_INSTALLED=true SNAPSHOT=true TEST_PLATFORMS="linux/amd64" mage -v integration:single TestFleetManagedUpgrade
```

It consistently fails on for me.

# Some extra information
The process that fails is the new Elastic-Agent, looking at the systemd logs we can see:
```
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e systemd[1]: Started Elastic Agent is a unified agent to observe, monitor and protect your system..
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]: Error: error checking for and handling upgrade: unable to load upgrade marker to check if Agent is being upgraded: yaml: line 17: mapping values are not allowed in this context
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]: For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.13/fleet-troubleshooting.html
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]: Error: error checking for and handling upgrade: unable to load upgrade marker to check if Agent is being upgraded: yaml: line 17: mapping values are not allowed in this context
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]: Usage:
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]:   elastic-agent [subcommand] [flags]
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]:   elastic-agent [command]
Dec 21 15:13:49 ogc-linux-amd64-ubuntu-2204-fleet-331e elastic-agent[65235]: Available Commands:
[...]

PR #3948, opened 2023-12-21
[Upgrade Details] Fix corruption when writing upgrade marker file

ycombinator:
## What does this PR do?

This PR fixes the corruption that could take place when writing the Upgrade Marker file.

## Why is it important?

Agent tries to reads the Upgrade Marker file at startup to determine if the Agent is in the midst of an upgrade.  If it cannot read the Upgrade Marker file because it's corrupted, an error is returned, and Agent fails to start up.

## Note about integration/E2E tests for this PR

The bug that's being fixed in this PR was caught via an E2E test that was failing consistently: `TestFleetManagedUpgrade `.  Unfortunately, though this PR fixes the bug, this test will not pass on the PR.  That's to be expected because the non-truncating write that was causing the bug is happening from the Upgrade Watcher, which is run as part of the new (i.e. upgraded) Agent. And the Agent that's upgraded to in the test is one pulled down from the artifacts API, not the one that's locally built.

I did add a unit test to catch the truncation bug to this PR.

## Checklist

- [ ] ~I have added an integration test or an E2E test~ Bug was already caught by integration tests

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/3947

### DIFF
--- a/internal/pkg/agent/application/upgrade/marker_access_common.go
+++ b/internal/pkg/agent/application/upgrade/marker_access_common.go
@@ -10,7 +10,7 @@ import (
 )
 
 func writeMarkerFileCommon(markerFile string, markerBytes []byte, shouldFsync bool) error {
-	f, err := os.OpenFile(markerFile, os.O_WRONLY|os.O_CREATE, 0600)
+	f, err := os.OpenFile(markerFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return fmt.Errorf("failed to open upgrade marker file for writing: %w", err)
 	}
--- a/internal/pkg/agent/application/upgrade/marker_access_common_test.go
+++ b/internal/pkg/agent/application/upgrade/marker_access_common_test.go
@@ -1,0 +1,56 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package upgrade
+
+import (
+	"math/rand"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWriteMarkerFileWithTruncation(t *testing.T) {
+	tmpDir := t.TempDir()
+	testMarkerFile := filepath.Join(tmpDir, markerFilename)
+
+	// Write a long marker file
+	err := writeMarkerFileCommon(testMarkerFile, randomBytes(40), true)
+	require.NoError(t, err)
+
+	// Get length of file
+	fileInfo, err := os.Stat(testMarkerFile)
+	require.NoError(t, err)
+	originalSize := fileInfo.Size()
+
+	// Write a shorter marker file
+	err = writeMarkerFileCommon(testMarkerFile, randomBytes(25), true)
+	require.NoError(t, err)
+
+	// Get length of file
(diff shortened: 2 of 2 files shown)


.buildkite/scripts/steps/integration_tests.sh
```diff
@@ -4,8 +4,8 @@ set -euo pipefail
 source .buildkite/scripts/common.sh
 
 STACK_PROVISIONER="${1:-"stateful"}"
-MAGE_TARGET="${2:-"integration:test"}"
-MAGE_SUBTARGET="${3:-""}"
```
ycombinator, line 8: For debugging only.  Should be removed before merging PR.

internal/pkg/agent/application/upgrade/marker_access_common_test.go
```diff
@@ -0,0 +1,56 @@
...
+	// Make sure shorter file has is smaller in length than
+	// the original long marker file
+	require.Less(t, newSize, originalSize)
+}
+
+func randomBytes(length int) []byte {
+	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZÅÄÖ" +
+		"abcdefghijklmnopqrstuvwxyzåäö" +
+		"0123456789" +
+		"~=+%^*/()[]{}/!@#$?|")
```
belimawr, line 47: Nit: Because it's just a test for the file size, I'd avoid special characters:
```suggestion
	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
		"abcdefghijklmnopqrstuvwxyz" +
		"0123456789" +
		"~=+%^*/()[]{}/!@#$?|")
```

belimawr (approved):

michalpristas (approved):

pierrehilbert:
Looks like the issue is not fixed unfortunately
`    upgrader.go:434: waiting for healthy agent and proper version: could not unmarshal agent status output: error: exit status 1, output: Error: failed to communicate with Elastic Agent daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/elastic-agent.sock: connect: connection refused"
        
        invalid character 'E' looking for beginning of value`

pierrehilbert:
As mentioned [here](https://github.com/elastic/elastic-agent/issues/3947#issuecomment-1867008703), failures on integration test are expected so I'm force merging this one as it will solve the original integration test failure.

internal/pkg/agent/application/upgrade/marker_access_common.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func writeMarkerFileCommon(markerFile string, markerBytes []byte, shouldFsync bool) error {
-	f, err := os.OpenFile(markerFile, os.O_WRONLY|os.O_CREATE, 0600)
+	f, err := os.OpenFile(markerFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
cmacknz, line 13: One thing to note with this approach is that the truncate and write do not necessarily happen atomically in the same IO operation. This could create a brief period where a read would see an empty upgrade marker file.

I am unsure how likely this is, but it seems possible. Ways to avoid this are to introduce retries when reading a 0 byte marker file or writing to a tmp file and then replacing the original file.

ycombinator: > or writing to a tmp file and then replacing the original file

Implemented this in https://github.com/elastic/elastic-agent/pull/3953.

PR #3910, opened 2023-12-13
Add manifest to package

pchila:
## What does this PR do?
This PR adds a manifest to .zip, tar.gz, .deb and .rpm packages according to the outcome of #3639
In this PR the manifest is only generated and included in the packages but not yet used by install or upgrade processes.

## Why is it important?
This is a prerequisite for #3950 where the manifest included in the package will be used for remapping package directories during install and upgrade

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Author's Checklist

- [ ]

## How to test this PR locally

It can be interesting having a look at .tar.gz or .zip packages and see how the manifest content (especially path mappings) changes when specifying different versions from the same commit for example:
```shell
SNAPSHOT=true PLATFORMS="linux/amd64" PACKAGES="tar.gz"  mage package
```
and 

```shell
 AGENT_PACKAGE_VERSION=8.13.1-exp SNAPSHOT=true PLATFORMS="linux/amd64" PACKAGES="tar.gz"  mage package
```

## Related issues

- Relates #3950 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #3950, opened 2023-12-21
Use version in path

### LINKED ISSUE #2579: Add the full version number to the installation directory name

pchila:
## What does this PR do?
This PR introduces support for the elastic-agent package manifest during install and upgrade operations.
This implements #2579 as it will remap the legacy `elastic-agent-<hash>` directory in `elastic-agent-<version>-<hash>`

The version check during upgrade has been changed from comparing the hashes of the old and new agent version to comparing the triplets `(version, snapshot flag, hash)`: as long as at least one of these elements differs between 2 elastic agent versions, those version can be installed side-by-side without collision in their paths.

## Why is it important?
This allows for faster releases of agent by allowing repackaging (creation of new elastic agent package versions based on the same elastic agent commit) by updating only components impacted by bugs/vulnerabilities.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Requires #3910
- Closes #2579 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/paths/common.go
+++ b/internal/pkg/agent/application/paths/common.go
@@ -167,11 +167,16 @@ func ExternalInputs() string {
 
 // Data returns the data directory for Agent
 func Data() string {
+	return DataFrom(Top())
+}
+
+// DataFrom returns the data directory for Agent using the passed directory as top path
+func DataFrom(topDirPath string) string {
 	if unversionedHome {
 		// unversioned means the topPath is the data path
-		return topPath
+		return topDirPath
 	}
-	return filepath.Join(Top(), "data")
+	return filepath.Join(topDirPath, "data")
 }
 
 // Run returns the run directory for Agent
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -70,18 +70,27 @@ func Cleanup(log *logger.Logger, topDirPath, currentVersionedHome, currentHash s
 	log.Infow("Cleaning up upgrade", "hash", currentHash, "remove_marker", removeMarker)
 	<-time.After(afterRestartDelay)
 
+	// data directory path
+	dataDirPath := paths.DataFrom(topDirPath)
+
 	// remove upgrade marker
 	if removeMarker {
-		if err := CleanMarker(log, paths.DataFrom(topDirPath)); err != nil {
+		if err := CleanMarker(log, dataDirPath); err != nil {
 			return err
 		}
 	}
 
 	// remove data/elastic-agent-{hash}
-	dataDir, err := os.Open(paths.DataFrom(topDirPath))
+	dataDir, err := os.Open(dataDirPath)
 	if err != nil {
 		return err
 	}
+	defer func(dataDir *os.File) {
+		err := dataDir.Close()
+		if err != nil {
+			log.Errorw("Error closing data directory", "file.directory", dataDirPath)
+		}
+	}(dataDir)
 
 	subdirs, err := dataDir.Readdirnames(0)
 	if err != nil {
@@ -113,7 +122,7 @@ func Cleanup(log *logger.Logger, topDirPath, currentVersionedHome, currentHash s
 			continue
 		}
 
-		hashedDir := filepath.Join(paths.DataFrom(topDirPath), dir)
+		hashedDir := filepath.Join(dataDirPath, dir)
 		log.Infow("Removing hashed data directory", "file.path", hashedDir)
 		var ignoredDirs []string
 		if keepLogs {
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -138,13 +138,13 @@ func Cleanup(log *logger.Logger, topDirPath, currentVersionedHome, currentHash s
 
 // InvokeWatcher invokes an agent instance using watcher argument for watching behavior of
 // agent during upgrade period.
-func InvokeWatcher(log *logger.Logger) error {
+func InvokeWatcher(log *logger.Logger, agentExecutable string) error {
 	if !IsUpgradeable() {
 		log.Info("agent is not upgradable, not starting watcher")
 		return nil
 	}
 
-	cmd := invokeCmd()
+	cmd := invokeCmd(agentExecutable)
 	defer func() {
 		if cmd.Process != nil {
 			log.Infof("releasing watcher %v", cmd.Process.Pid)
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -72,7 +72,7 @@ func Cleanup(log *logger.Logger, currentVersionedHome, currentHash string, remov
 
 	// remove upgrade marker
 	if removeMarker {
-		if err := CleanMarker(log); err != nil {
+		if err := CleanMarker(log, paths.Data()); err != nil {
 			return err
 		}
 	}
@@ -89,8 +89,8 @@ func Cleanup(log *logger.Logger, currentVersionedHome, currentHash string, remov
 	}
 
 	// remove symlink to avoid upgrade failures, ignore error
-	prevSymlink := prevSymlinkPath()
-	log.Infow("Removing previous symlink path", "file.path", prevSymlinkPath())
+	prevSymlink := prevSymlinkPath(paths.Top())
+	log.Infow("Removing previous symlink path", "file.path", prevSymlinkPath(paths.Top()))
 	_ = os.Remove(prevSymlink)
 
 	dirPrefix := fmt.Sprintf("%s-", agentName)
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -46,7 +46,7 @@ func Rollback(ctx context.Context, log *logger.Logger, c client.Client, topDirPa
 		symlinkTarget = paths.BinaryPath(filepath.Join(paths.DataFrom(topDirPath), hashedDir), agentName)
 	}
 	// change symlink
-	if err := changeSymlinkInternal(log, topDirPath, symlinkPath, symlinkTarget); err != nil {
+	if err := ChangeSymlink(log, topDirPath, symlinkPath, symlinkTarget); err != nil {
 		return err
 	}
 
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -34,24 +34,25 @@ const (
 
 // Rollback rollbacks to previous version which was functioning before upgrade.
 func Rollback(ctx context.Context, log *logger.Logger, prevVersionedHome, prevHash string, currentHash string) error {
-	symlinkPath := filepath.Join(paths.Top(), agentName)
+	topDirPath := paths.Top()
+	symlinkPath := filepath.Join(topDirPath, agentName)
 
 	var symlinkTarget string
 	if prevVersionedHome != "" {
-		symlinkTarget = paths.BinaryPath(filepath.Join(paths.Top(), prevVersionedHome), agentName)
+		symlinkTarget = paths.BinaryPath(filepath.Join(topDirPath, prevVersionedHome), agentName)
 	} else {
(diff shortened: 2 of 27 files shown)


internal/pkg/agent/application/paths/common.go
```diff
@@ -59,6 +61,11 @@ var (
 )
 
 func init() {
+	// this is needed to correctly initialize paths containing the agent package version
+	// TODO find a better solution and a way to handle errors gracefully
+	version.InitVersionInformation()
```
blakerouse, line 66: Make lint happy.

```suggestion
	_ = version.InitVersionInformation()
```

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -199,3 +219,7 @@ func restartAgent(ctx context.Context, log *logger.Logger) error {
 	close(signal)
 	return nil
 }
+
+type mockAgentServer struct {
+	cproto.UnimplementedElasticAgentControlServer
+}
```
blakerouse, line 225: Should be deleted?

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -7,52 +7,103 @@ package upgrade
...
 	"github.com/hashicorp/go-multierror"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/errors"
+	v1 "github.com/elastic/elastic-agent/pkg/api/v1"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
+type UnpackResult struct {
+	Hash string `json:"hash" yaml:"hash"`
+	// TODO add mapped path of executable
```
blakerouse, line 30: This TODO still needed?

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -7,52 +7,103 @@ package upgrade
...
+
+	// Load manifest, the use of path.Join is intentional since in .zip file paths use slash ('/') as separator
+	manifestFile, err := r.Open(path.Join(fileNamePrefix, "manifest.yaml"))
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		// we got a real error looking up for the manifest
+		return UnpackResult{}, fmt.Errorf("looking up manifest in package: %w", err)
+	}
+	if err == nil {
+		// load manifest
+		defer manifestFile.Close()
```
blakerouse, line 77: should this maybe be called directly instead of with a defer? defer will wait to the end of the function, the contents seems to be read here and done. I would prefer to see this closed as soon as possible

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -7,52 +7,103 @@ package upgrade
...
+		versionedHome = filepath.Clean(pm.Map(manifest.Package.VersionedHome))
+	}
+
+	// Load hash, the use of path.Join is intentional since in .zip file paths use slash ('/') as separator
+	hashFile, err := r.Open(path.Join(fileNamePrefix, agentCommitFile))
+	if err != nil {
+		// we got a real error looking up for the manifest
+		return UnpackResult{}, fmt.Errorf("looking up %q in package: %w", agentCommitFile, err)
+	}
+	defer hashFile.Close()
```
blakerouse, line 92: same as above

version/helper.go
```diff
@@ -31,6 +31,17 @@ var (
 
 const PackageVersionFileName = "package.version"
 
+var versionInitError error
+
+func init() {
+	versionInitError = InitVersionInformation()
```
blakerouse, line 37: This is already called in another init that I commented on.

You should make the internal of this function `InitVersionInformation` use a `sync.Once` and store the error inside of the `sync.once` instead of on the outside. That will sure that this is initialized only once and the error is always set correctly.

blakerouse (commented):
This is a preliminary review as the PR is still in draft.

 I did notice that there are a few TODO/FIX around the PR. I would clean those up and ensure all those points have been solved before placing into full review.

Overall I really like the testing in this PR it has lots of tests which is great to see. It doesn't have any integration tests. I wonder if those could possibly be added that would maybe modify the zip/tarball to tests some upgrade scenarios.

version/helper.go
pchila: The init() function is guaranteed to be initialized only once already by the go runtime, so no extra synchronization mechanisms are necessary. The only concurrent access we might have is reading versionInitError after initialization but since the value of the variable never changes after initialization it is not a problem.

pchila:
>  I wonder if those could possibly be added that would maybe modify the zip/tarball to tests some upgrade scenarios.

In order to test the feature (allowing 2 different versions of agent built from the same commit to be installed side by side) in an integration test we need to package agent twice from the same commit and assign a different version to the extra package. I don't know if the current integration fwk is capable of handling 2 packages for the same commit with different versions in the same test (will try and see what happens).

As for the general upgrade flow there from previous releases or current snapshots there already plenty of tests, adding another would not add value.

internal/pkg/agent/application/upgrade/rollback.go
pchila: Ah yes, that's a leftover from before I extracted the client and replaced that with a mock. Will delete it

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: It was a nice-to-have that I abandoned in this first implementation. Will remove the TODO and add some documentation for the struct

internal/pkg/agent/application/paths/common.go
pchila: Removed in favor of an anonymous import

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -209,36 +345,190 @@ func untar(log *logger.Logger, version string, archivePath, dataDir string) (str
...
+
+type tarCloser struct {
+	tarFile    *os.File
+	gzipReader *gzip.Reader
+}
+
+func (tc *tarCloser) Close() error {
+	var err error
+	if tc.gzipReader != nil {
+		err = multierror.Append(err, tc.gzipReader.Close())
```
cmacknz, line 452: Shouldn't we be using `errors.Join` now?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -269,15 +370,19 @@ func (u *Upgrader) sourceURI(retrievedURI string) string {
...
 }
 
-func rollbackInstall(ctx context.Context, log *logger.Logger, hash string) {
-	os.RemoveAll(filepath.Join(paths.Data(), fmt.Sprintf("%s-%s", agentName, hash)))
-	_ = ChangeSymlink(ctx, log, release.ShortCommit())
+func rollbackInstall(ctx context.Context, log *logger.Logger, topDirPath, versionedHome string) {
+	err := os.RemoveAll(filepath.Join(topDirPath, versionedHome))
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		// TODO should this be a warning or an error ?
+		log.Warnw("error rolling back install", "error.message", err)
```
cmacknz, line 377: It looks like this is only failing to remove the version we are upgrading to.

If another upgrade attempt can succeed with this situation, it's a warning. If not, it's an error.

Also I think the log message text could be clearer to state it is an error removing a path along with which path that is.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -170,49 +194,126 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
+	newHash := unpackRes.Hash
 	if newHash == "" {
 		return nil, errors.New("unknown hash")
 	}
 
-	if strings.HasPrefix(release.Commit(), newHash) {
-		u.log.Warn("Upgrade action skipped: upgrade did not occur because its the same version")
-		return nil, nil
+	if unpackRes.VersionedHome == "" {
+		// FIXME this should be treated as an error
```
cmacknz, line 241: It looks like it is an error already?

internal/pkg/agent/install/install.go
```diff
@@ -229,6 +226,108 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 	return ownership, nil
 }
 
+func copyFiles(streams *cli.IOStreams, pathMappings []map[string]string, dir string, topPath string) error {
```
cmacknz, line 229: This function could use some comments explaining why it is doing what it is doing, considering how much code it contains.

cmacknz (commented):
This is looking good, especially the new upgrade tests. Left a few comments from my first pass, will probably have to read this more than once.

As I mentioned earlier, please add a summary of what the manifest file is, how it works, and why it is necessary to https://github.com/elastic/elastic-agent/tree/main/docs.

I think this will also help others who haven't been following this as closely review the PR.

internal/pkg/agent/application/upgrade/upgrade.go
pchila: Yeah, decided for error and then forgot to remove the fixme, will remove that

internal/pkg/agent/application/upgrade/upgrade.go
pchila: I changed it a bit in the latest commit, I think it should be ok now

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -170,49 +194,126 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
-	if err := InvokeWatcher(u.log); err != nil {
+	minParsedVersionForNewUpdateMarker := agtversion.NewParsedSemVer(8, 13, 0, "", "")
+	var watcherExecutable string
+	if parsedVersion.Less(*minParsedVersionForNewUpdateMarker) {
+		// use the current agent executable for watch
+		watcherExecutable = paths.BinaryPath(paths.VersionedHome(paths.Top()), agentName)
+	} else {
+		// use the new agent executable as it should be able to parse the new update marker
+		watcherExecutable = paths.BinaryPath(filepath.Join(paths.Top(), unpackRes.VersionedHome), agentName)
+	}
```
cmacknz, line 309: What exactly is it in the upgrade marker that breaks backwards compatibility?

We AFAIK didn't have this problem when we introduced upgrade details.

Or is the problem not the upgrade marker, is it that the older agent doesn't understand the path structure of the current 8.13.0 agent? If this is the case I think the comment should clarify this, and I'm unsure if there's anything else we could do about this.

cmacknz: P.S. now that I've actually looked at this snippet I see how it'd be hard to pull out on its own, because it's pointless by itself.

pchila:
buildkite test this

internal/pkg/agent/application/upgrade/upgrade.go
pchila: The upgrade marker in itself is backward compatible (I only added the fields with version and paths of new and old versions of agent).
The issue is that a watcher that does not read those fields will break the agent installation that uses the new paths on Cleanup/Rollback ( it will remove any agent directory that uses a path different from `elastic-agent-<hash>`).

The reason why we didn't have this problem for upgrade details is: those are just informational states that we send along with our status, if those are not there we simply don't set the status and the upgrade completes normally. In this case the watcher will actively delete files and directories and if it does it the legacy way on installations that use the new paths the agent is broken.

pchila:
This pr is substituted by the smaller PRs:

#4173 
#4174 
#4176 
#4177 
#4178

PR #3971, opened 2023-12-29
Avoid nested Eventually calls in APM Ingestion E2E test

michalpristas:
`findESDocs` uses `require.Eventually` internally. so if this is used inside another eventually call like in APM ingestion test. it may lead to running just one iteration of the main loop.


michalpristas:
looks like we're getting stuck on this
```
docs := findESDocs(t, func() (estools.Documents, error) {
				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
				defer cancel()
				return estools.GetLogsForIndexWithContext(ctx, esClient, "logs-apm*", match)
			})
```

and only one run of eventually condition check is performed and not finished

michalpristas:
Failing on unrelated tests now

AndersonQ (approved):

PR #3982, opened 2024-01-02
Return the current agent state on Fixture.IsHealthy method

AndersonQ:
## What does this PR do?

Modifies `Fixture.IsHealthy` to return the current agent status if the agent isn't healthy.

## Why is it important?

When debugging a failing test it's valuable to know the current status of the agent.

## Checklist

## Related issues

- N/A

## Questions to ask yourself

- ...


pkg/testing/fixture.go
```diff
@@ -724,16 +724,23 @@ func (f *Fixture) ExecVersion(ctx context.Context, opts ...process.CmdOption) (A
...
 func (f *Fixture) IsHealthy(ctx context.Context, opts ...process.CmdOption) (bool, error) {
 	status, err := f.ExecStatus(ctx, opts...)
 	if err != nil {
 		return false, fmt.Errorf("agent status returned and error: %w", err)
 	}
 
-	return status.State == int(cproto.State_HEALTHY), nil
+	if status.State != int(cproto.State_HEALTHY) {
+		return false, fmt.Errorf("agent isn't health, current status: %s",
+			client.State(status.State))
```
blakerouse, line 740: I understand the reason for adding this, but this kinda changes the meaning of the error in this function. The error was more that it failed to get the actual health. Now this changes it to return an error when its not healthy, that is probably not was the caller is expecting.

AndersonQ: This is true, but it's also why I changed it docs. Besides there is just one usage of it, that is the test which needed this change. I could change the function signature, but I believe the doc change clarifies what is returned in the error.

blakerouse: Based on the way your changing it, would you be willing to change it to `EnsureHealthy` and remove the boolean? Doesn't seem like true/false is relevant to what this function is even doing at more, so just have it return error or nil.

pkg/testing/fixture.go
```diff
@@ -724,16 +724,23 @@ func (f *Fixture) ExecVersion(ctx context.Context, opts ...process.CmdOption) (A
...
+// This function is compatible with any Elastic Agent version 8.6 or later.
 func (f *Fixture) IsHealthy(ctx context.Context, opts ...process.CmdOption) (bool, error) {
 	status, err := f.ExecStatus(ctx, opts...)
 	if err != nil {
 		return false, fmt.Errorf("agent status returned and error: %w", err)
 	}
 
-	return status.State == int(cproto.State_HEALTHY), nil
+	if status.State != int(cproto.State_HEALTHY) {
+		return false, fmt.Errorf("agent isn't health, current status: %s",
```
ycombinator, line 739: Nit: typo.
```suggestion
		return false, fmt.Errorf("agent isn't healthy, current status: %s",
```

pkg/testing/fixture.go
AndersonQ: It does make sense, I just kept the name because it isn't ensuring anything, just checking the status

blakerouse (approved):
+1

Issue #3981, opened 2024-01-02
Flaky Test TestMarkerWatcher


blakerouse:
## Flaky Test

* **Test Name:** TestMarkerWatcher
* **Link:** https://github.com/elastic/elastic-agent/blob/1ee9cc7ca9d88d676ab3124f1f57b2013ed17a9c/internal/pkg/agent/application/upgrade/marker_watcher_test.go#L25
* **Branch:** https://github.com/elastic/elastic-agent/pull/3929
* **Artifact Link:** https://buildkite.com/elastic/elastic-agent/builds/5878

### Stack Trace

```

=== Failed
--
  | === FAIL: internal/pkg/agent/application/upgrade TestMarkerWatcher (unknown)
  | panic: send on closed channel
  |  
  | goroutine 61 [running]:
  | github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade.watchFileNotEmpty.func1()
  | C:/buildkite-agent/builds/bk-agent-prod-gcp-1704207293125079587/elastic/elastic-agent/internal/pkg/agent/application/upgrade/marker_access_test.go:124 +0x3ae
  | created by github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade.watchFileNotEmpty
  | C:/buildkite-agent/builds/bk-agent-prod-gcp-1704207293125079587/elastic/elastic-agent/internal/pkg/agent/application/upgrade/marker_access_test.go:91 +0x20f


```

ycombinator:
> * **Artifact Link:** https://buildkite.com/elastic/elastic-agent/builds/5878

Looking at this build, I'm not seeing `TestMarkerWatcher` failing.  

I also ran it locally on my Mac with 5000 iterations, a couple of times, and I wasn't able to get it to fail.

```
$ go test ./internal/pkg/agent/application/upgrade/... -test.run TestMarkerWatcher -test.count 5000
?   	github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact/download/localremote	[no test files]
ok  	github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade	75.885s
ok  	github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact	1.031s [no tests to run]
ok  	github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact/download	1.089s [no tests to run]
[...]

PR #3987, opened 2024-01-02
Fix TestWriteMarkerFileWithTruncation flaky test

aleksmaus:
## What does this PR do?

Fixes TestWriteMarkerFileWithTruncation flaky test
Addresses [Flaky Test TestMarkerWatcher](https://github.com/elastic/elastic-agent/issues/3981) 

This is one way to fix this without rewriting the test too much.

The actual problem is in the TestWriteMarkerFileWithTruncation test, which also corresponds with the stack trace in the original issue.

There are few problems with that tests:
1. Race condition between Go routines that this tests starts upon the cancelling the shared context. 
Need to wait for the watcher to exit first, and only then exit from the Go routine that listens on  errCh (otherwise the watcher Go routine that writes into errCh can get blocked and leak).
https://github.com/aleksmaus/elastic-agent/blob/main/internal/pkg/agent/application/upgrade/marker_access_test.go#L44
Only after that close the errCh.

2. Another issue was not breaking out on the error case here https://github.com/aleksmaus/elastic-agent/blob/main/internal/pkg/agent/application/upgrade/marker_access_test.go#L124, that would cause double write into unbuffered errCh.

## Why is it important?

Test stability

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3981

### DIFF
--- a/internal/pkg/agent/application/upgrade/marker_access_test.go
+++ b/internal/pkg/agent/application/upgrade/marker_access_test.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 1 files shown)


ycombinator (approved):
LGTM - thanks for adding the missing synchronization and the missed `return` as well.

aleksmaus:
Unrelated test failure:
```
>>> (linux-arm64-ubuntu-2204-default) Test output (non-sudo) (stdout): === FAIL: testing/integration TestAPMConfig (22.82s)
```

blakerouse (approved):

blakerouse:
How does this solve #3981?

aleksmaus:
> How does this solve #3981?

I mentioned that in this comment
https://github.com/elastic/elastic-agent/issues/3981#issuecomment-1874386321

blakerouse:
@aleksmaus Thanks, missed that comment on the issue.

Issue #3960, opened 2023-12-27
[French-German]: Agent service is not restarted itself on installation for French and German Windows 10.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.12.0 BC3
BUILD: 69985
COMMIT: 2a8afed8572a4c709aa1c64216748197eeb9b18f
Artifact Link: https://staging.elastic.co/8.12.0-61156bc6/summary-8.12.0.html
```

**Host OS:**  Windows French and German

**Preconditions:**
1. 8.12.0 BC3 Kibana cloud environment should be available.

**Steps to reproduce:**
1. Run agent install command on Windows 10 French/German VM.
2. Observe agent service is not restarted itself on installation.
3. Navigate to services and manually restart the Elastic Agent service.
4. Observe agent gets Healthy under Agents tab.

**NOTE:**
- Issue is consistently reproducible on French and German VMs.
- Issue is not reproducible on Korean and Spanish VMs.

**Screenshot:**
![image](https://github.com/elastic/elastic-agent/assets/77374876/d1c44fba-07c3-467d-9cf9-d7e889130c45)
![image](https://github.com/elastic/elastic-agent/assets/77374876/662ce37a-73d7-41f1-bd5b-61171cc7f574)

**Expected Result:**
Agent service should be restarted itself on installation for French and German Windows 10.

**Logs:**
Unable to collect logs till manually restarted:
![image](https://github.com/elastic/elastic-agent/assets/77374876/d3b0ef1e-8f4d-495b-b00b-24c67d1181e3)

**Logs after restart:**
[elastic-agent-diagnostics-2023-12-27T08-05-55Z-00.zip](https://github.com/elastic/elastic-agent/files/13776780/elastic-agent-diagnostics-2023-12-27T08-05-55Z-00.zip)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

intxgo:
Hello, interestingly this is not a new bug, straight from an elevated console of my Windows 11 Pro English OS:

```
PS C:\Users\[redacted]> & 'C:\Program Files\Elastic\Agent\elastic-agent.exe' diagnostics
[...]

PR #3988, opened 2024-01-03
switch to consistent HasRoot method on windows and fix memory leak

leehinman:
## What does this PR do?

replaces custom `isWindowsAdmin` in listener_windows.go with call to `HasRoot` implementation in `utils`.  Also fix memory leak in `HasRoot` function.

## Why is it important?

In some localized Windows installs calls to `GroupIds` for the user of the process fail, this was causing the `isWindowsAdmin` check to fail, which causes problems with the control socket for `elastic-agent`.

The `HasRoot` implementation doesn't depend on `GroupIds`

## Checklist

## Author's Checklist

- [ ]

### Prior to this PR.

1. Run agent install command on Windows 10 French/German VM.
2. Observe agent service is not restarted itself on installation.
3. Navigate to services and manually restart the Elastic Agent service.
4. Observe agent gets Healthy under Agents tab.

### With this PR

1. Run agent install command on Windows 10 French/German VM.
2. Observe agent service is restarted on installation.
3. `elastic-agent restart` should work

## Related issues

- #3960

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/control/v2/server/listener_windows.go
+++ b/pkg/control/v2/server/listener_windows.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"net"
 	"os/user"
-	"strings"
 
 	"github.com/elastic/elastic-agent-libs/api/npipe"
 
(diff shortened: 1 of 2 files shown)


pierrehilbert:
Failure seems to be unrelated (your change is impacting windows only and the failure is on Ubuntu).
I created an issue for this failure: https://github.com/elastic/elastic-agent/issues/3989

blakerouse (commented):
Does this have the same semantics? Will this allow a user that is part of the Administrators group to still be able to communicate with the Elastic Agent control socket?

leehinman:
> Does this have the same semantics? Will this allow a user that is part of the Administrators group to still be able to communicate with the Elastic Agent control socket?

yes.  take a look at https://github.com/golang/go/issues/28804 for more in depth discussion.  But basically the check is to see if the current process is part of the "well known" admin SID, and if so then the Administrators group gets added to the control socket.  None of the logic changed about adding to the security descriptor on the socket, just in how we detect if the process is "Admin" or not.

blakerouse (approved):
Thanks for the details. Awesome!

michalpristas (approved):

PR #4005, opened 2024-01-04
[Internal Documentation] Data flow of CPU metrics presented in Fleet UI

ycombinator:
## What does this PR do?

This PR documents the end-to-end data flow of CPU metrics presented in the Fleet UI, including any transformations/calculations/aggregations that are done at the various stages along the way.

## Why is it important?

To understand how the CPU utilization % metric presented in the Fleet UI for every Agent in the Agent Listing view is ultimately calculated, as this has resulted in confusion for a lot of users.


nimarezainia:
fyi @kpollich

cmacknz:
Relates: https://github.com/elastic/kibana/issues/174458

docs/cpu-metrics-in-fleet.md
```diff
@@ -0,0 +1,320 @@
...
+2024-01-05T23:21:59.164Z	34810
+```
+
+We can see that the metrics in the Agent + Beats `/stats` API outputs match up with those in the Elasticsearch
+`metrics-elastic_agent*` indices.
+
+## Unanswered Questions
+
+1. Why do we divide by 10000 (`params._interval`) in the ES query?
+2. Why do we use a `10s` interval for the derivative aggregation in the ES query? Is this related to the division above?
```
cmacknz, line 320: I think we used to use a 10s metrics reporting interval but it was changed to 60s, perhaps the 10s was related to the rate at which we reported metrics originally.

cmacknz: @nchaulet might be able to answer this correctly.

nchaulet: If I remember correctly how this was calculated we use the derivative aggregation to get the number of ms of cpu used during an interval (the 10s) than we divided this by 10000 to get the % of cpu used during that interval. 

If the reporting interval changed this probably need to be changed too. We probably should use `metricset.period` in our query here to avoid to change the query each time the period change.

ycombinator: Thank you. I understand now that the two are indeed related - if we use `units: "10s"` in the `derivative` aggregation, we're asking for CPU time consumed (in milliseconds) per 10-second period.  To get a percentage of the elapsed time (10s), we divide that value by 10*1000 milliseconds.  Makes sense.

ycombinator:
@cmacknz @fearful-symmetry @nchaulet I've completed documenting the data path taken by CPU utilization metrics all the way from `/proc/{pid}/stat` files (for Agent and Beats processes running on Linux) to the single value shown in the Fleet UI.  

I've also included, at the end of this documentation, sections on my observations and suggested improvements.  Please review the entire document but especially the observations and suggested improvements sections and provide any feedback.  I'd like to file issues (or reuse existing ones) for, and link to them from, each of the improvements in this documentation. 

The short/medium-term goal is that next time there's an SDH on Agent CPU utilization metrics in the Fleet UI being confusing, we can link to this documentation. The medium/long-term goal would be to implement the improvements and then delete them from this documentation, while also updating the rest of the documentation as needed.

Thank you!

docs/cpu-metrics-in-fleet.md
```diff
@@ -0,0 +1,374 @@
...
+  icon in CPU column in the Agent Listing page in the Fleet UI.
+
+* Rather than showing a single value for every Agent in the Agent Listing page in the Fleet UI, we should consider showing
+  the constituent values for Agent and each of the Beats processes managed by the Agent. Concretely, this would mean
+  getting rid of the `sum_cpu` aggregation in the Elasticsearch query. The challenge, of course, would be in presenting
+  these values in a user-friendly manner, especially in cases where an Agent is managing several Beats. Also, by showing
+  the values for the Beat processes (along with the Agent process), we would be exposing internals of Agent to users, which
+  may be confusing for some users.
+
+* We should enhance collection and aggregation to include CPU utilization for Beats, which are Agent components managed by
```
fearful-symmetry, line 372: Can you elaborate on this? You mean provide the CPU metrics in the `stats` metrics, instead of requiring fleet to perform its own aggregation query?

docs/cpu-metrics-in-fleet.md
```diff
@@ -0,0 +1,374 @@
...
+  time, the CPU utilization % shown varies for each process over time.
+
+* To relate the output seen in `top` or `htop` for Agent and Beat processes with the single value shown in the Fleet UI,
+  one must observe the values in the `top` / `htop` output over five minutes, arrive at an average CPU utilization (in %)
+  value for each process, and sum it up.  The resulting value will roughly be equal to the single value shown for that Agent
+  in the Agent Listing page in the Fleet UI.
+
+## Suggested improvements
+
+* We should document the observations above in an appropriate location and perhaps link to this documentation from the "i"
```
fearful-symmetry, line 362: Ideally, the tooltip text should also be improved, as right now it's pretty vague.

docs/cpu-metrics-in-fleet.md
```diff
@@ -0,0 +1,374 @@
...
+## Observations
+
+* The CPU utilization (in %) of Agent and each of the Beat processes can vary wildly. Generally speaking, the Agent process
+  itself does not utilize much CPU. And each Beat process may utilize CPU depending on the type of computations it is performing
+  on the data.
+
+* Also, CPU utilization is rarely constant. If the output of `top` or `htop` for Agent and Beat processes is observed over
+  time, the CPU utilization % shown varies for each process over time.
+
+* To relate the output seen in `top` or `htop` for Agent and Beat processes with the single value shown in the Fleet UI,
```
fearful-symmetry, line 355: I wonder if users also find this confusing? Do we _need_ to average over 5 minutes? It smooths out spikes, but also might confuse existing users to used to peeping at metrics.

docs/cpu-metrics-in-fleet.md
ycombinator: Sorry, I don't think I wrote this in the clearest way.  I've changed the language now but what I wanted to convey was that we should also include and present CPU metrics about service-runtime components (e.g. Endpoint), not just command-runtime components (e.g. Beats).  And I deliberately didn't want to get into exactly _how_ we should do that.

docs/cpu-metrics-in-fleet.md
ycombinator: ++ added an entry about this in the "Suggested improvements" section below.

docs/cpu-metrics-in-fleet.md
ycombinator: Yeah, if we can **briefly** explain what the metric represents and a bit about how it's calculated, we should put all that in the tooltip.  Otherwise, I think we should just replace the tooltip with a  link to a documentation page.

cmacknz: IMO if we have to link to a documentation page we are probably presenting the wrong value. Most likely the CPU should just document itself as the 5 minute average CPU utilization, or switch to just presenting an instantaneous value from the last checkin with a link to see a graph of the values over time.

docs/cpu-metrics-in-fleet.md
cmacknz: I think we account for endpoint CPU now as well, which will even further complicate users attempt to correlate what is shown in Fleet to what is in top.

Top will never account for endpoint resource usage because it is in a different process and cgroup.

docs/cpu-metrics-in-fleet.md
```diff
@@ -0,0 +1,374 @@
...
+  one must observe the values in the `top` / `htop` output over five minutes, arrive at an average CPU utilization (in %)
+  value for each process, and sum it up.  The resulting value will roughly be equal to the single value shown for that Agent
+  in the Agent Listing page in the Fleet UI.
+
+## Suggested improvements
+
+* We should document the observations above in an appropriate location and perhaps link to this documentation from the "i"
+  icon in CPU column in the Agent Listing page in the Fleet UI.
+
+* Rather than showing a single value for every Agent in the Agent Listing page in the Fleet UI, we should consider showing
```
cmacknz, line 365: We also have to account for there being multiple instances of each individual Beat process when we display it them. We have to break it down by unique input-output combination as that is how we create the processes.

docs/cpu-metrics-in-fleet.md
```diff
@@ -0,0 +1,374 @@
...
+curl -s -u $ES_USER:$ES_PASS -H 'Content-Type: application/json' 'https://test-cpu.es.us-central1.gcp.cloud.es.io:9243/metrics-elastic_agent*/_search' -d '{
+  "sort": [
+    {
+      "@timestamp": {
+        "order": "desc"
+      }
+    }
+  ],
+  "collapse": {
+    "field": "elastic_agent.process"
```
cmacknz, line 258: Does this account for there being multiple instances of each Beat process? It could depending on if we use the component ID (filestream-default) or the just the process name (e.g. filebeat) as the value.

For reference the default Fleet policy with the system integration and monitoring runs 3 metricbeat instances (system/metrics-default, http/metrics-monitoring, beat/metrics-monitoring) and 2 filebeat instances (filestream-monitoring, log-default).

If this is averaging the 3 metricbeats and 2 filebeats together as if they are one process instead of summing the usage of each distinct process I don't think it is correct.

The memory usage query has this problem as described in https://github.com/elastic/kibana/issues/174458

docs/cpu-metrics-in-fleet.md
ycombinator: Agreed, I guess this suggestion should be more of a fallback if we can't come up with a good UX in the Fleet UI itself.

> ... just presenting an instantaneous value from the last checkin 

I'm partial to idea of showing a single value that's the sum the of the latest CPU utilization %s for Agent + all it's component processes with a tooltip explaining that the value can range from 0 to the number of cores * 100.  This value should be easily correlatable to what's seen in the output of `top` / `htop`.

> ... with a link to see a graph of the values over time.

++,  single value could link to a chart showing the CPU utilization %, per process, over time, in case the user wants to drill down further.

docs/cpu-metrics-in-fleet.md
ycombinator: It doesn't.  The value of this field is just `filebeat`.  The query should be using the `component.id` field itself, which has values such as `log-default`.  I will add this fix to the list of suggestions.

cmacknz: Looks like this is already tracked as a TODO in https://github.com/elastic/kibana/issues/174458 now that I look closer at it

docs/cpu-metrics-in-fleet.md
ycombinator: Added suggestion in 2ded3c4782.

docs/cpu-metrics-in-fleet.md
ycombinator: Added suggestion in e490363791.

docs/cpu-metrics-in-fleet.md
ycombinator: Added suggestion in 2ded3c4782.

ycombinator:
@fearful-symmetry @cmacknz Thanks for all the great feedback!  Based on it, I've reworked the suggestions section in the doc now.  Please re-review and I will iterate further as needed.  Once you're both happy with the suggestions, I'll file issues for them and link them from this doc before merging this PR.

docs/cpu-metrics-in-fleet.md
ycombinator: If you're referring to https://github.com/elastic/kibana/issues/174458#issuecomment-1885770207, I added that based on this conversation here :).

docs/cpu-metrics-in-fleet.md
ycombinator: Hmmm, I'm not seeing Endpoint CPU being recorded today.

I just added the Elastic Defend integration to my policy and then, a few minutes later, queried the `metrics-elastic_agent*` indices with the following query:

```
POST metrics-elastic_agent*/_search?filter_path=hits.hits._source
{
  "sort": [
    {
      "@timestamp": {
        "order": "desc"
      }
    }
  ],
  "_source": [
    "@timestamp", 
    "elastic_agent.process",
    "component.id"
  ],
  "collapse": {
    "field": "component.id"
  },
  "query": {
    "bool": {
      "must": [
        {
          "terms": {
            "elastic_agent.id": [
              "c406741d-7673-447f-b7e0-43417eeb10d6"
            ]
          }
        }
      ]
    }
  }
}
```

Results:

```json
{
  "hits": {
    "hits": [
      {
        "_source": {
          "@timestamp": "2024-01-10T22:03:45.069Z",
          "component": {
            "id": "log-default"
          },
          "elastic_agent": {
            "process": "filebeat"
          }
        }
      },
      {
        "_source": {
          "@timestamp": "2024-01-10T22:03:45.069Z",
          "component": {
            "id": "elastic-agent"
          },
          "elastic_agent": {
            "process": "elastic-agent"
          }
        }
      }
    ]
  }
}```

File: docs/cpu-metrics-in-fleet.md (hunk 6 above)
Comment (cmacknz) [reply docs/cpu-metrics-in-fleet.md:258]: One more question: did you observe us reporting CPU for the monitoring components?

https://github.com/elastic/kibana/issues/174458#issuecomment-1887717174

There is some evidence we are not reporting them, causing us to undercount CPU and memory usage by omitting the monitoring component contributions.

**Review by cmacknz — APPROVED** [review]
I only have questions about why we are doing what we are doing at this point, the content here LGTM.

File: docs/cpu-metrics-in-fleet.md (hunk 6 above)
Comment (ycombinator) [reply docs/cpu-metrics-in-fleet.md:258]: I did not observe us reporting CPU for monitoring components themselves.

File: docs/cpu-metrics-in-fleet.md [hunk 7]
```diff
@@ -0,0 +1,380 @@
...
+* The `processes` aggregation in the Elasticsearch query should use the field `component.id` instead of `elastic_agent.process`.
+  This is to correctly account for multiple instances of the same type of Beat (e.g. Filebeat).  This can happen if there
+  are multiple outputs defined in the Agent policy and some inputs of a type (e.g. log) use one output while other inputs
+  of the same type use another output.
+
+* We should reconsider taking a 5-minute average in the Elasticsearch query made by the Fleet UI and instead take a
+  30-second or 1-minute average (making corresponding adjustments to the `calendar_interval` value in the `cpu_time_series`
+  aggregation). This would result in a value closer to what's observed in `top` / `htop` output.
+
+* We should link the value shown in the Fleet UI to a chart that breaks it down for that Agent by `component.id` over time,
```
nchaulet, line 371: should this be the `elastic_agent` dashboard?

ycombinator: Yes, specifically the Agent Metrics one, e.g.:

Screenshot 2024-01-11 at 11 51 54

fearful-symmetry (approved):
I would say this is all pretty reasonable. Do we plan to break down the `Suggested Improvements` down to a meta-issue or something?

ycombinator:
> I would say this is all pretty reasonable. Do we plan to break down the `Suggested Improvements` down to a meta-issue or something?

Yes, I plan to file separate issues for them.

docs/cpu-metrics-in-fleet.md
ycombinator: Added a suggestion for collecting CPU usage metrics for monitoring components.

ycombinator:
Filed issues for each of the suggested improvements and linked to them from this doc.

PR #4023, opened 2024-01-05
Support for validate subcommand for otel configuration

michalpristas:
performs validation of config passed to the command using flags supported also in `otel` subcommand (`--config` and `--set`)


michalpristas:
UT for logic not for boilerplate results in low soundqube rating

internal/pkg/agent/cmd/testdata/otel/elastic-agent.yml
```diff
@@ -0,0 +1,301 @@
+###################### Agent Configuration Example #########################
```
ycombinator, line 1: WDYT about removing all the comments from this file, so this file also becomes more like the `otel.yml` file?  Just thinking that the comments are not adding much value to test fixture files.

ycombinator:
With this PR, the syntax for running OTel configuration validation becomes `elastic-agent validate`.  However, the syntax for validating OTel configuration with the vanilla OTel collector is `otelcol validate --config /path/to/first.yml --config /path/to/first.yml  ...`. 

To ease migration, I think we'd want Agent (with embedded OTel Collector) to be as much of a drop-in replacement for the vanilla OTel collector.  So I think we need to support the same syntax, that is, something like `elastic-agent validate --config /path/to/first.yml --config /path/to/first.yml  ...`, or [better yet](https://github.com/elastic/elastic-agent/pull/4023#issuecomment-1884869351), `elastic-agent otel validate --config /path/to/first.yml --config /path/to/first.yml  ...`.

michalpristas:
once i introduce otel subcommand i will move this under otel subcommand and add multiple files and other flags. for now i prefer to stay with flags we have and not support multiple files using --config as normal agent does not understand that. 

the order in my mind is 
https://github.com/elastic/elastic-agent/pull/4047 to go in
introduce otel
bring this even with normal collector CLI

internal/pkg/agent/cmd/validate.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/internal/pkg/cli"
+	"github.com/elastic/elastic-agent/internal/pkg/otel"
+)
+
+func newValidateCommandWithArgs(_ []string, _ *cli.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "validate",
+		Short:         "Validates the config without running the collector",
```
ycombinator, line 21:
```suggestion
		Short:         "Validates the OpenTelemetry collector configuration without running the collector",
```

michalpristas:
elastic-agent.yml is a default config value if not provided by `--config` flag. may be it

michalpristas:
i changed default for validate subcommand to `otel.yml`

michalpristas:
fixed

internal/pkg/agent/cmd/validate.go
```diff
@@ -0,0 +1,40 @@
...
+		},
+	}
+
+	return cmd
+}
+
+func validateOtelConfig(ctx context.Context, cfgFile string) error {
+	if runAsOtel := otel.IsOtelConfig(ctx, cfgFile); !runAsOtel {
+		return fmt.Errorf("%q is not an otel config. file should be named 'otel.yml', 'otlp.yml' or 'otelcol.yml'", cfgFile)
+	}
```
ycombinator, line 37: After this PR is merged but before we change `elastic-agent validate` to `elastic-agent otel validate`, the user experience would be that `elastic-agent validate` only works for OTel configurations.  To me, this is a big awkward as `validate` is a top-level sub-command and, therefore, suggests that it can validate both OTel and Agent configurations.

So my suggestion for this PR would be to return early from this function with `nil` if `otel.IsOtelConfig(ctx, cfgFile)` returns `false` and also make `DefaultConfigName` the default value for the `-c/--config` global CLI options.

Once we have the follow up PR to move `elastic-agent validate` to `elastic-agent otel validate`, it would make sense to error out if `otel.IsOtelConfig(ctx, cfgFile)` returns false.

internal/pkg/agent/cmd/validate.go
```diff
@@ -0,0 +1,40 @@
...
+			cfgFile := paths.ConfigFileWithDefault("otel.yml")
+			return validateOtelConfig(ctx, cfgFile)
+		},
+	}
+
+	return cmd
+}
+
+func validateOtelConfig(ctx context.Context, cfgFile string) error {
+	if runAsOtel := otel.IsOtelConfig(ctx, cfgFile); !runAsOtel {
```
ycombinator, line 35: Any reason this shouldn't be shortened?

```suggestion
	if !otel.IsOtelConfig(ctx, cfgFile) {
```

pierrehilbert:
@michalpristas should we put this back on draft for now?

michalpristas:
@pierrehilbert after `otel` subcommand, this is up to date and ready for review

ycombinator:
If I run `./elastic-agent otel validate` (without any `--config` options), the command succeeds with a 0 exit code.

As comparison, if I run the OTel Collector binary's `validate` sub-command without any `--config` options, that command fails like so:

```sh
$ ./bin/otelcorecol_darwin_arm64 validate
Error: at least one config flag must be provided
2024/02/02 04:18:29 collector server run finished with error: at least one config flag must be provided
```

Would it be possible to mimic the OTel Collector binary's behavior in `./elastic-agent otel validate`?  Just trying to make this command as much as a drop-in replacement for the native OTel Collector binary as much as possible.

ycombinator (approved):
Apart from my request/suggestion in https://github.com/elastic/elastic-agent/pull/4023#issuecomment-1923695648, which I would say is a nice to have, this PR LGTM.

testing/integration/otel_test.go
```diff
@@ -133,6 +133,11 @@ func TestOtelFileProcessing(t *testing.T) {
 		`"stringValue":"system.log"`, // system.log is being processed
 	})
 
+	// check `elastic-agent otel validate` command works for otel config
+	out, err := fixture.Exec(ctx, []string{"otel", "validate", "--config", cfgFilePath})
```
cmacknz, line 137: I would like to see a test (here or elsewhere) that the validate command actually detects errors in otel configuration files.

That is the real value of the validate command, lets make sure it works. This test would pass even if we never called the otel validate function and just parsed the command line arguments.

michalpristas: this is now addressed at https://github.com/elastic/elastic-agent/pull/4023/commits/e432f98e14709f96e2ca849df0a00f285ce3f4db

michalpristas: i pulled it outside into separate function but kept within same test to spare some resources

cmacknz (approved):

cmacknz:
LGTM, fine to force merge this. Will wait until Michal is back online before doing so.

PR #4024, opened 2024-01-05
Send package.version to components

AndersonQ:
## What does this PR do?

The agent now includes AgentInfo alongside the connection information during component startup. It also stores the BuildHash sent by the components as part of the ComponentVersionInfo during check-in.

To facilitate testing, an info.Agent interface has been created to abstract the implementation of info.AgentInfo. This abstraction allows for the creation of mocks, enabling testing scenarios where using `info.NewAgentInfo` is impractical. This is because `info.NewAgentInfo` relies on the agent vault, which, on Mac systems, is the system's keychain. Accessing the keychain requires root permissions, which are not available during testing.

## Why is it important?

The components should report the package version instead of their own.

## Checklist

With a beats build containing https://github.com/elastic/beats/pull/37553:
 - build the agent
 - change the package.version file to contain a version with extra build information, e.g.: `8.13.0+build20131123`
 - enroll it to fleet
 - the events sent by the beats should have the `agent.version`, `elastic_agent.version` and `metadata.version` set as the version defined on the `package.version` file.
 - in the `state.yaml` from the agent diagnostics, for each component, there should be:
   - no `version` field in the `version_info`
   - a new `version_info.build_hash` with the beats build hash

## Related issues

- N/A

## Related PRs
 - https://github.com/elastic/elastic-agent-client/pull/93
 - https://github.com/elastic/beats/pull/37553

## Questions to ask yourself

- ...


dev-tools/mage/gotool/go.go
AndersonQ, line 1: Once https://github.com/elastic/elastic-agent/pull/4039 is merged, this changes won't me o this PR anymore

dev-tools/notice/overrides.json
```diff
@@ -1,4 +1,5 @@
 {"name": "github.com/elastic/beats/v7", "licenceType": "Elastic"}
+{"name": "github.com/AndersonQ/elastic-agent-client/v7", "licenceType": "Elastic"}
```
AndersonQ, line 2: It'll be removed once https://github.com/elastic/elastic-agent-client/pull/93 is merged

cmacknz:
The integration test we are going to want here will be to change the agent package version file at install time to be something different from what the agent was built with. You can edit it to something the build will never produce, like set it to 8.0.5000+build19700101 for example.

Then install the agent, ensure it can enroll in Fleet, and verify the documents ingested into Elasticsearch actually have that version in the `agent.version` field (and the `elastic_agent.version` field if it exists in the document).

You could probably just update the existing log ingestion tests to operate with a modified package version to cover this. https://github.com/elastic/elastic-agent/blob/main/testing/integration/logs_ingestion_test.go

You'd have to inspect the version fields in the returned documents in addition to what it does now, but this will be easier and execute faster than writing a completely new test.

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -96,18 +96,25 @@ func New(ctx context.Context, path string, opts ...OptionFunc) (v *Vault, err er
...
 	enc, err := v.encrypt(data)
 	if err != nil {
-		return err
+		return fmt.Errorf("vault Set: could not encrypt key: %w", err)
 	}
 
 	err = v.tryLock(ctx)
 	if err != nil {
-		return err
+		return fmt.Errorf("vault Set: could aqquire lock: %w", err)
```
pchila, line 104:
```suggestion
		return fmt.Errorf("vault Set: could acquire lock: %w", err)
```

internal/pkg/agent/vault/vault_notdarwin.go
```diff
@@ -96,18 +96,25 @@ func New(ctx context.Context, path string, opts ...OptionFunc) (v *Vault, err er
...
 	defer func() {
 		err = v.unlockAndJoinErrors(err)
+		if err != nil {
+			err = fmt.Errorf("vault Set: unlockAndJoinErrors failed: %w", err)
+		}
 	}()
 
-	return writeFile(v.filepathFromKey(key), enc)
+	err = writeFile(v.filepathFromKey(key), enc)
+	if err != nil {
```
pchila, line 114: Error handling is missing ?

pkg/component/runtime/runtime_comm_test.go
```diff
@@ -0,0 +1,80 @@
...
+	require.NoError(t, err, "failed to write ConnInfo")
+
+	clientv2, _, err := client.NewV2FromReader(&buff, client.VersionInfo{
+		Name: "TestRuntimeComm_WriteConnInfo",
+		Meta: nil,
+	})
+	require.NoError(t, err, "failed creating V2 client")
+
+	assert.Equal(t, &want, clientv2.AgentInfo(),
+		"package version does not match")
```
pchila, line 79: Nit: We should try to make the text explain the reason for the assertion, something like:
```suggestion
		"agent info returned by client must match what has been written on command input")
```

pkg/component/runtime/state.go
```diff
@@ -345,6 +347,10 @@ func (s *ComponentState) syncCheckin(checkin *proto.CheckinObserved) bool {
 			s.VersionInfo.Name = checkin.VersionInfo.Name
 			changed = true
 		}
+		if checkin.VersionInfo.BuildHash != "" && s.VersionInfo.BuildHash != checkin.VersionInfo.BuildHash {
+			s.VersionInfo.BuildHash = checkin.VersionInfo.BuildHash
+			changed = true
+		}
```
pchila, line 353: If we are trying to detect a change in the version reported by the component we should check the triplet (version, snapshot flag, hash) since repackaged agent versions will have the same hash as other versions.

If that's not what we are doing here, then I don't understand what purpose this snippet serves :sweat_smile:

pkg/testing/tools/fleettools/fleet.go
```diff
@@ -144,3 +152,43 @@ func DefaultURL(ctx context.Context, client *kibana.Client) (string, error) {
 
 	return "", errors.New("unable to determine default fleet server URL")
 }
+
+// NewEnrollParams creates a new policy with monitoring logs and metrics,
+// an enrollment token and returns an EnrollParams with the information to enroll
+// an agent. If an error happens, it returns nil and a non-nil error.
+func NewEnrollParams(ctx context.Context, client *kibana.Client) (*EnrollParams, error) {
```
pchila, line 159: How is this new test related to sending the agent version to components?
I don't mind more coverage but I fail to see the relation with the purpose of the PR

testing/integration/diagnostics_test.go
```diff
@@ -139,28 +139,44 @@ func TestDiagnosticsCommand(t *testing.T) {
...
 		require.NoError(t, err)
 
-		verifyDiagnosticArchive(t, files[0], diagFiles, diagCompFiles, avi)
+		verifyDiagnosticArchive(t, diagZip, diagFiles, diagCompFiles, avi)
 
 		return nil
 	}
 }
 
+// collectDiagnostics collects the agent diagnostics and returns the path to the
```
pchila, line 154: Nit: this something shared among different integration tests in different files, maybe we want to move it to a common utilities file

testing/integration/package_version_test.go
```diff
@@ -50,6 +55,107 @@ func TestPackageVersion(t *testing.T) {
...
+	err = f.Prepare(ctx)
+	require.NoError(t, err, "could not prepare fixture")
+
+	// Set package version
+	pkgVersionFiles := findPkgVersionFiles(t, f.WorkDir())
+	pkgVersion := "8.13.0+build20131123"
+	for _, pkgVerFile := range pkgVersionFiles {
+		err := os.WriteFile(pkgVerFile, []byte(pkgVersion), 0o644)
+		require.NoError(t, err)
+	}
```
pchila, line 85: I would prefer we prepare a dedicated package with a specific version instead of hacking through the version files changing the values. I am not sure this will work once the version is part of the agent path...

testing/integration/package_version_test.go
```diff
@@ -50,6 +55,107 @@ func TestPackageVersion(t *testing.T) {
...
+			} `yaml:"state"`
+		} `yaml:"components"`
+	}{}
+	err = yaml.NewDecoder(stateYAML).Decode(&state)
+	require.NoError(t, err, "could not parse state.yaml (%s)", stateYAML.Name())
+
+	for _, c := range state.Components {
+		assert.Equal(t, wantBuildHash, c.State.VersionInfo.BuildHash)
+		assert.Equal(t, wantBuildHash, c.State.VersionInfo.Meta.Commit)
+	}
```
pchila, line 156: Same problem here: we are checking just the hash but that is not uniquely identifying an agent version

pchila (changes requested):
There are a couple of nitpicks but the thing that worries me most is that this PR is using the agent hash as the sole identifier of an agent version. With the elastic agent early releases on the horizon, where we will potentially repackage the same agent binary with different patched components I am not sure that this holds...

pkg/component/runtime/state.go
AndersonQ: it's the components build hash. the agent does not have it bu default. the component sends it on its 1st checkin

cmacknz:
Your PR description still just says WIP, you at least need to document how to manually test this.

pkg/component/runtime/state.go
cmacknz: Yes this is the build hash of the sub-processes, which are not going to match the build hash of the agent. 

We stopped receiving the version string because it is now just going to be equal to the package version of the agent. 

Looking at only the hash is intentionally, because it is the only thing that tells us which commit and branch the component version is from, since the version string no longer uniquely identifies a stack release branch.

testing/integration/package_version_test.go
AndersonQ: ok, I agree, but other tests are doing the same. Therefore it could be a followup PR changing all the test. Also to do so it'd require to build another agent for the test, changing the step on buildkite as well. What ins't quite the scope of this PR

changelog/fragments/1707143167-send-package-version-to-components.yaml
```diff
@@ -0,0 +1,34 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Agent information is sent to components together with the connection information.
```
blakerouse, line 14: Is this really informational for our users? I don't think they care, this is more an internal change and not something that affects the usage of the Elastic Agent.

blakerouse (approved):
Overall this looks good. I question the need for the changelog entry though.

changelog/fragments/1707143167-send-package-version-to-components.yaml
pierrehilbert: I tend to agree, this change is more an internal change than something the users should pay attention at

cmacknz: +1 we can remove the changelog entry here.

cmacknz:
Tested this manually, looks like it works. The `agent.name` field in each event matches the agent package version.

I did note that the build hash field is not included in the version info output from `elastic-agent status --output=yaml` (and json, and full)

```yaml
- id: system/metrics-default
  name: system/metrics
  state: 2
  message: 'Healthy: communicating with pid ''86008'''
  units:
  - unit_id: system/metrics-default
    unit_type: 1
    state: 2
    message: Healthy
  - unit_id: system/metrics-default-system/metrics-system-56875859-e15e-48ab-ad98-29698fd331ad
    unit_type: 0
    state: 2
    message: Healthy
  version_info:
    name: beat-v2-client
    meta:
      build_time: 2024-02-06 11:59:31 +0000 UTC
      commit: ac6e2230bcfe766f1bd5695a59aa7d55116098f7
```

It is included in the `state.yaml` field in the diagostics though:

```yaml
    - id: system/metrics-default
      state:
        component:
            apmconfig: null
            limits:
                gomaxprocs: 0
                source:
                    fields:
                        go_max_procs:
                            kind:
                                numbervalue: 0
        component_idx: 2
        features_idx: 2
        message: 'Healthy: communicating with pid ''86008'''
        state: 2
        units:
            input-system/metrics-default-system/metrics-system-56875859-e15e-48ab-ad98-29698fd331ad:
                message: Healthy
                state: 2
            output-system/metrics-default:
                message: Healthy
                state: 2
        version_info:
            build_hash: ac6e2230bcfe766f1bd5695a59aa7d55116098f7
            meta:
                build_time: 2024-02-06 11:59:31 +0000 UTC
                commit: ac6e2230bcfe766f1bd5695a59aa7d55116098f7
            name: beat-v2-client
```

AndersonQ:
> Tested this manually, looks like it works. The `agent.name` field in each event matches the agent package version.
> 
> I did note that the build hash field is not included in the version info output from `elastic-agent status --output=yaml` (and json, and full)
> 
> ```yaml
> - id: system/metrics-default
>   name: system/metrics
>   state: 2
>   message: 'Healthy: communicating with pid ''86008'''
>   units:
>   - unit_id: system/metrics-default
>     unit_type: 1
>     state: 2
>     message: Healthy
>   - unit_id: system/metrics-default-system/metrics-system-56875859-e15e-48ab-ad98-29698fd331ad
>     unit_type: 0
>     state: 2
>     message: Healthy
>   version_info:
>     name: beat-v2-client
>     meta:
>       build_time: 2024-02-06 11:59:31 +0000 UTC
>       commit: ac6e2230bcfe766f1bd5695a59aa7d55116098f7
> ```

Interesting, the `--output full` already did not show it:

```
❯ sudo /opt/Elastic/Agent/elastic-agent status --output full
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: 7c2067ff-30ff-4628-b8cf-af36fa56a524
   │  ├─ version: 8.12.0
   │  └─ commit: 5cbf2e403c761f91d11eca6b9cb5385e0f07f2ce
   ├─ beat/metrics-monitoring
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '2343'
   │  ├─ beat/metrics-monitoring
   │  │  └─ type: OUTPUT
   │     └─ type: INPUT

❯ sudo /opt/Elastic/Agent/elastic-agent status --output yaml
info:
  id: 7c2067ff-30ff-4628-b8cf-af36fa56a524
  version: 8.12.0
  commit: 5cbf2e403c761f91d11eca6b9cb5385e0f07f2ce
  build_time: 2024-01-11 13:25:50 +0000 UTC
  snapshot: false
  pid: 1274
state: 2
message: Running
components:
- id: beat/metrics-monitoring
  name: beat/metrics
  state: 2
  message: 'Healthy: communicating with pid ''2343'''
  units:
  - unit_id: beat/metrics-monitoring
    unit_type: 1
    state: 2
    message: Healthy
  - unit_id: beat/metrics-monitoring-metrics-monitoring-beats
    unit_type: 0
    state: 2
    message: Healthy
  version_info:
    name: beat-v2-client
    version: 8.12.0
    meta:
      build_time: 2024-01-10 21:09:16 +0000 UTC
      commit: 27c592782c25906c968a41f0a6d8b1955790c8c5
```

I'll open an issue to fix the status command as a whole then

AndersonQ:
done: 
 - https://github.com/elastic/elastic-agent/issues/4214

PR #3901, opened 2023-12-12
Remove the OVERRIDE_AGENT_PACKAGE_VERSION value as 8.13 is now built

pierrehilbert:
## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4036, opened 2024-01-06
[Upgrade Integration/E2E tests] Update version in guard for running upgrade details assertions

ycombinator:
## What does this PR do?

This PR makes it so the upgrade integration/E2E tests run assertions about upgrade details being present or absent (as expected) from the upgrade marker file IF the target Agent version is >= 8.12.0.

## Why is it important?

When the version guard was originally introduced, the code being exercised by the test was not yet present in `8.12.0` builds. That's because the code being exercised was in the same PR as the version guards, that PR was targetting `main`, and hadn't yet been backported to `8.12`.  To get tests passing on that PR the guard only ran the relevant assertions if the version of the upgraded Agent was >= 8.13.0.  

Now that PR has been merged AND backported to `8.12` as well, the code being exercised is in `8.12` and `main` both.  So now we can update the guard to run the relevant assertions if the version of the upgraded Agent is >= 8.12.0.

## Related issues

Related to https://github.com/elastic/elastic-agent/pull/3901

### DIFF
--- a/testing/integration/upgrade_downgrade_test.go
+++ b/testing/integration/upgrade_downgrade_test.go
@@ -88,7 +88,7 @@ func TestStandaloneDowngradeToSpecificSnapshotBuild(t *testing.T) {
 	// We pass the upgradetest.WithDisableUpgradeWatcherUpgradeDetailsCheck option here because the endFixture
 	// is fetched from the artifacts API and it may not contain changes in the Upgrade Watcher whose effects are
 	// being asserted upon in upgradetest.PerformUpgrade.
-	// TODO: Stop passing this option and remove these comments once 8.13.0 has been released.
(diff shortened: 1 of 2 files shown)


michalpristas (approved):

cmacknz:
https://github.com/elastic/elastic-agent/pull/4038 is the same as this one but with the additional line that I think fixes the problem. We need to consider the start version as well as the end version when deciding whether to check upgrade details. Was easier to make a new PR to test that this works as expected.

ycombinator:
> #4038 is the same as this one but with the additional line that I think fixes the problem. We need to consider the start version as well as the end version when deciding whether to check upgrade details. Was easier to make a new PR to test that this works as expected.

Thanks.  Closing this PR in favor of #4038.

Issue #4041, opened 2024-01-08
Cannot send `+Inf` download rate to Fleet API in `upgrade_details.metadata.download_rate` double JSON field in check-in API request


ycombinator:
When upgrading Agent via Fleet, I noticed the following warning in the Elastic Agent logs:

```
{"log.level":"warn","@timestamp":"2024-01-08T17:30:45.639Z","log.origin":{"file.name":"fleet/fleet_gateway.go","file.line":193},"message":"Possible transient error during checkin with fleet-server, retrying","log":{"source":"elastic-agent"},"error":{"message":"status code: 500, fleet-server returned an error: BadRequest, message: failed to update upgrade_details: elastic fail 400: document_parsing_exception: [1:3872] failed to parse field [upgrade_details.metadata.download_rate] of type [double] in document with id '92df91f9-aa51-4279-9d9c-cef6dabb3f04'. Preview of field's value: '+Inf bps': number_format_exception: For input string: \"+Inf bps\""},"request_duration_ns":1072107250,"failed_checkins":2,"retry_after_ns":217255338807,"ecs.version":"1.6.0"}
```

cmacknz:
That looks like an ES mapping error, I don't think the ES mapping for double accepts +Inf as implied by the "restricted to finite values" in: https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html

> A double-precision 64-bit IEEE 754 floating point number, restricted to finite values.

Is `download_rate` optional? If so when we hit `+Inf` the easiest fix might be to just omit it. Once we hit `+Inf` I don't think we are actually downloading anymore, correct? `+Inf` doesn't mean anything useful to a user here anyway.

ycombinator:
Yes, `download_rate` is optional.  We hit `+Inf` when the download completes very fast. So yes, the download is complete at that point and reporting the download rate to the user as a progress indicator is not useful. ++ to detecting `+Inf` and just omitting it.

cmacknz:
Will put a PR with that change shortly. Looks like this causes the checkin to fail so want to avoid releasing this problem.

PR #4044, opened 2024-01-08
Omit download rate when it is calculated as +Inf to prevent checkin errors

cmacknz:
- Closes https://github.com/elastic/elastic-agent/issues/4041

We can calculate download rates of +Inf if the download happens quickly enough, which when sent as part of the checkin payload causes the checkin to fail because only finite float values can be indexed in ES.

Since +Inf isn't informative to the user, just omit it since `download_rate` is optional. Do the same for -Inf and NaN in case those values ever happen.


cmacknz:
buildkite test it

ycombinator:
With the changes in this PR, I wonder if we even need the [`details.DownloadRate` type](https://github.com/elastic/elastic-agent-libs/blob/1cf0afdc6c934ab507ef043e7d920435419af61d/upgrade/details/details.go#L31) any more?

cmacknz:
buildkite test it

cmacknz:
> With the changes in this PR, I wonder if we even need the [details.DownloadRate type](https://github.com/elastic/elastic-agent-libs/blob/1cf0afdc6c934ab507ef043e7d920435419af61d/upgrade/details/details.go#L31) any more?

Hmm, didn't notice we had that. We don't need the +Inf handling, we want to keep the edit to the precision in the string.

ycombinator (approved):
LGTM.

cmacknz:
```
=== RUN   TestFQDN
    fqdn_test.go:49: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/fqdn_test.go:49
        	            	Get "https://api.ipify.org": net/http: TLS handshake timeout
        	Test:       	TestFQDN
--- FAIL: TestFQDN (23.09s)
```

Failure unrelated, will open an issue for this. Force merging.

Issue #3406, opened 2023-09-12
Log a summary of each policy change received from Fleet


cmacknz:
One thing I have noticed looking through diagnostics is that I can often tell what the agent is doing, but I can't tell why unless I also have access to Fleet or logs of the requests made to Fleet. This also applies to standalone agents, but the configuration is more likely to be static in that case or subject to changes from autodiscovery events.

Each time we get a policy change from Fleet, we should log a summary of what it contained. We cannot log the complete policy because it will contain credentials (both the output and for individual integrations), but we could log a summary indicating which sections of the policy has changed. For example we could log:

1. That a policy change has occurred.
2. Which outputs were added/removed/updated.
3. Which inputs were added/removed/updated.

This would more easily allow correlating the actions of the coordinator to the actions from Fleet. For example today, if an input stops in the logs it is difficult to tell whether it was supposed to stop unless the final policy in the agent diagnostics does not contain the input (in which case we can assume it was added and then removed).

PR #4050, opened 2024-01-09
Add fleet config change logger

fearful-symmetry:
closes https://github.com/elastic/elastic-agent/issues/3406

## What does this PR do?

This is a basic implementation of https://github.com/elastic/elastic-agent/issues/3406. When triggered, it prints a change detailing added and removed units:
```
{"log.level":"info","@timestamp":"2024-01-08T23:52:56.107Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":1297},"message":"The following components have been added: [udp-default journald-default]","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
```

Right now this PR is fairly basic and does what we want it to do, so I figured I'd ask if there's anything else we think we should do here. We could drill down even farther, and diff individual streams in an input, etc. The downside here is that diffing an array is kind of complex, so we might be incurring a large cost for complex configs.

### Example log lines:
```
{"log.level":"info","@timestamp":"2024-01-17T22:08:16.612Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":1382},"message":"component model updated","log":{"source":"elastic-agent"},"changes":{"components":{"added":["udp-default","journald-default"],"updated":["http/metrics-monitoring: [(http/metrics-monitoring-metrics-monitoring-agent: updated)]","log-default: [(log-default-logfile-system-9bf446fc-58d4-4767-b42d-3450815d5d3d: updated)]","system/metrics-default: [(system/metrics-default-system/metrics-system-9bf446fc-58d4-4767-b42d-3450815d5d3d: updated)]","beat/metrics-monitoring: [(beat/metrics-monitoring-metrics-monitoring-beats: updated)]"],"count":7},"outputs":{}},"ecs.version":"1.6.0"}

{"log.level":"info","@timestamp":"2024-01-17T22:08:18.558Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":1382},"message":"component model updated","log":{"source":"elastic-agent"},"changes":{"components":{"updated":["beat/metrics-monitoring: [(beat/metrics-monitoring-metrics-monitoring-beats: updated)]","http/metrics-monitoring: [(http/metrics-monitoring-metrics-monitoring-agent: updated)]"],"count":7},"outputs":{}},"ecs.version":"1.6.0"}
```

## Why is it important?

We need debug data showing us what config has changed before a config is applied.

## Checklist


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1219,103 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
+	if len(removedList) > 0 {
+		c.logger.Debugf("The following components have been removed: %v", removedList)
+	}
+
+	if len(addedOutputs) > 0 {
+		c.logger.Debugf("The following outputs have been added: %v", addedOutputs)
+	}
+	if len(removedOutputs) > 0 {
+		c.logger.Debugf("The following outputs have been removed: %v", removedOutputs)
+	}
```
cmacknz, line 1314: This doesn't log components that have been updated, so that's the biggest thing to add. It is probably the most complicated to compute though.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1219,103 @@ func (c *Coordinator) generateComponentModel() (err error) {
 	// If we made it this far, update our internal derived values and
 	// return with no error
 	c.derivedConfig = cfg
+	c.lastComponentModel = c.componentModel
 	c.componentModel = comps
+
+	// don't run a bunch of relatively costly code unless we're in debug mode
+	if c.logger.IsDebug() {
+		c.checkAndLogUpdate()
+	}
```
cmacknz, line 1228: IMO this is valuable even when not in debug mode, policy changes aren't that frequent so the actual amortized cost of this should be low.

I'd vote we turn this on by default.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1219,103 @@ func (c *Coordinator) generateComponentModel() (err error) {
 	// If we made it this far, update our internal derived values and
 	// return with no error
 	c.derivedConfig = cfg
+	c.lastComponentModel = c.componentModel
```
cmacknz, line 1222: Do you actually need to save this? How big is it? Remember we can have agents with 100s of units.

Can you not just move the `c.componentModel = comps` to after this function call and pass in both the new and updated model as arguments? Like `checkAndLogUpdate(c.componentModel, comps)`?

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: Yah, that's a fairly complex operation, since it involves diffing two relatively complex objects, though.

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: Yah, seems useful, but I was kinda worried about overhead, particularly if we start adding info on component updates. Should probably run some benchmarks and see what it looks like.

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: Ah, good idea.

fearful-symmetry:
Alright, completely refactored this to properly print updated component/units, and also clean it up.

There's still a few more tests to add, but other than that, should be good.

AndersonQ (changes requested):
just a small blocker

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1224,157 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
 	// return with no error
 	c.derivedConfig = cfg
+
+	lastComponentModel := c.componentModel
 	c.componentModel = comps
+
+	// don't run a bunch of relatively costly code unless we're in debug mode
+	if c.logger.IsDebug() {
+		c.checkAndLogUpdate(lastComponentModel)
+	}
```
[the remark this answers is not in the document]
cmacknz: Agreed this should be at the info level. Policy changes producing differences are rare but important. Also the resource cost of this will be dominated by starting and stopping subprocesses.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1224,157 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
+		if comp.inLast && !comp.inNew {
+			removedOutputs = append(removedOutputs, output)
+		}
+		if !comp.inLast && comp.inNew {
+			addedOutputs = append(addedOutputs, output)
+		}
+	}
+
+	if len(addedList) > 0 {
+		c.logger.Debugf("The following components have been added: %v", addedList)
```
cmacknz, line 1358: Put all of these in the same log statement, I suggest using the `Infow` variant with each category being its own key value: added, removed, updated, etc.

This will associate them all with a single timestamp and event which will make them much easier to analyze and query for when they are shipped to Fleet.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1224,157 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
+		c.logger.Debugf("Received initial component update; total of %d components", len(c.componentModel))
+		return
+	}
+
+	type compCheck struct {
+		inNew  bool
+		inLast bool
+
+		diffUnits map[string]diffCheck
+	}
```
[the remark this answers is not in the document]
fearful-symmetry: Yah, that was done entirely because `compCheck` isn't used outside of one method, and `diffCheck` is. I didn't want to clutter up `coordinator.go`.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1224,157 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
+	compDiffMap := map[string]compCheck{}
+	outDiffMap := map[string]diffCheck{}
+	// kinda-callbacks for dealing with output logic
+	foundInLast := func(outName string) {
+		if outDiff, ok := outDiffMap[outName]; ok {
+			outDiff.inLast = true
+			outDiffMap[outName] = outDiff
+		} else {
+			outDiffMap[outName] = diffCheck{inLast: true}
+		}
```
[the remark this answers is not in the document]
fearful-symmetry: yah, the logic is to update if it exists, add otherwise.

fearful-symmetry:
test failed on windows, but not for the reasons expected?
```
 testing.go:1225: TempDir RemoveAll cleanup: remove C:\Users\buildkite\AppData\Local\Temp\TestComponentUpdateDiff2228549831\001\testlog-20240116.ndjson: The process cannot access the file because it is being used by another process.
```

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1239,149 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
+				if state.inLast && !state.inNew {
+					action = "removed"
+				}
+				if state.inNew && !state.inLast {
+					action = "added"
+				}
+				if state.updated {
+					action = "updated"
+				}
+				if action != "" {
```
cmacknz, line 1348: Should have a case for "unmodified"?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1216,10 +1239,149 @@ func (c *Coordinator) generateComponentModel() (err error) {
...
+			Count:   len(c.componentModel),
+			Updated: formattedUpdated,
+		},
+		Outputs: UpdateComponentChange{
+			Added:   addedOutputs,
+			Removed: removedOutputs,
+		},
+	}
+
+	c.logger.Infow("state update", "changes", logStruct)
```
cmacknz, line 1382:
```suggestion
	c.logger.Infow("component model updated", "changes", logStruct)
```

Let's be specific about what we are logging.

cmacknz (commented):
Can you post a few sample logs that this outputs?

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: My assumption was that if a unit was unmodified, we don't want to log it?

internal/pkg/agent/application/coordinator/coordinator_test.go
```diff
@@ -97,6 +99,180 @@ func waitForState(
...
+
+	logp.ObserverLogs()
+
+	err = logp.Sync()
+	require.NoError(t, err)
+	err = logp.Configure(logp.Config{
+		ToStderr: true,
+	})
+	require.NoError(t, err)
+
```
AndersonQ, line 267: Why are you configuring the logger after the tests?

fearful-symmetry: Ah, should add a comment there. There's a bug in `logp` for windows where if refuses to close any open file handles, causing the test to fail when go cleans up the logging directory.

fearful-symmetry: Oh wait, that's from before I removed the file logging in the test, I can just get rid of that.

AndersonQ:
@cmacknz @fearful-symmetry I'll be on PTO until next Tue, so feel free to proceed without my review. I've already removed my "request changes"

blakerouse (approved):
Looks good, has nice test coverage!

PR #4054, opened 2024-01-10
[Kubernetes] Update K8s version to 1.29

constanca-m:
## What does this PR do?

Add 1.29 K8s version.

## Related issues

Relates to https://github.com/elastic/observability-dev/issues/2926.


pierrehilbert (approved):

gizas (approved):

.buildkite/pipeline.yml
```diff
@@ -191,6 +191,7 @@ steps:
         matrix:
           setup:
             k8s_version:
+              - "1.29.0"
```
gizas, line 194: Although I approved, seeing this: https://kubernetes.io/releases/

So I would advise also to update the latest releases to version 1.28.5, 1.27.9, 1.26.12 
(all relative prs might need same update)

constanca-m: The versions used are from the image kindest/node. The latest versions are [here](https://hub.docker.com/r/kindest/node/tags?page=1), which are the same as these

gizas: You can ignore above as we align with latest available versions of kind https://hub.docker.com/r/kindest/node/tags?page=1

So LGTM

Issue #3683, opened 2023-11-01
Fix data race in tests that modify global agent paths in component runtime tests


cmacknz:
```
[2023-10-27T11:37:31.791Z] === FAIL: pkg/component/runtime TestManager_FakeInput_Configure (1.93s)
[2023-10-27T11:37:31.791Z]     manager_test.go:1656: component state changed: {State:STARTING Message:Starting: spawned pid '44999' Units:map[{UnitType:input UnitID:fake-input}:{State:STARTING Message:Starting: spawned pid '44999' Payload:map[] unitState:0 unitMessage:Starting unitPayload:map[] configStateIdx:0 err:<nil>}] Features:<nil> FeaturesIdx:0 Component:<nil> ComponentIdx:0 VersionInfo:{Name: Version: Meta:map[]} expectedUnits:map[{UnitType:0 UnitID:fake-input}:{state:2 configStateIdx:1 config:0xc000366750 err:<nil> logLevel:4}] expectedFeatures:<nil> expectedFeaturesIdx:1 expectedComponent:<nil> expectedComponentIdx:1}
[2023-10-27T11:37:31.791Z]     manager_test.go:1656: component state changed: {State:HEALTHY Message:Healthy: communicating with pid '44999' Units:map[{UnitType:input UnitID:fake-input}:{State:STARTING Message:Starting: spawned pid '44999' Payload:map[] unitState:0 unitMessage: unitPayload:map[] configStateIdx:0 err:<nil>}] Features:<nil> FeaturesIdx:0 Component:<nil> ComponentIdx:0 VersionInfo:{Name:fake Version:1.0 Meta:map[input:fake]} expectedUnits:map[{UnitType:0 UnitID:fake-input}:{state:2 configStateIdx:1 config:0xc000366750 err:<nil> logLevel:4}] expectedFeatures:<nil> expectedFeaturesIdx:1 expectedComponent:<nil> expectedComponentIdx:1}
[2023-10-27T11:37:31.791Z] {"log.level":"info","@timestamp":"0000-01-01T00:00:00.000Z","message":"","component":{"binary":"","dataset":"elastic_agent.","id":"fake-default","type":"fake"},"log":{"source":"fake-default"},"level":"info","state":"STARTING","expectedState":"HEALTHY","time":"2023-10-27T11:36:55Z","message":"unit change received","ecs.version":"1.6.0"}
[...]

Issue #4076, opened 2024-01-11
Windows tests failing with: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: open \\\\.\\pipe\\08aEXWqjpu7u7318BQm1ZyV6ip9MmEzd.sock: The system cannot find the file specified."


cmacknz:
## Flaky Test

* **Test Name:** TestFakeComponent, TestDiagnosticsCommand, etc.
* **Branch:** main
* **Artifact Link:** 
  * https://buildkite.com/elastic/elastic-agent/builds/6215
  * https://buildkite.com/elastic/elastic-agent/builds/6233

### Stack Trace

Stack trace with logs from https://github.com/elastic/elastic-agent/pull/4088:
```
=== RUN   TestFakeComponent
    fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip to C:\Users\windows\AppData\Local\Temp\TestFakeComponent363186115\001
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip to C:\Users\windows\AppData\Local\Temp\TestFakeComponent363186115\001
    fixture.go:880: All component specifications where removed
    fixture.go:921: Placed component specifications: fake, fake-shipper
    log.go:96: {"log.level":"info","@timestamp":"2024-01-16T20:53:03.207Z","log.origin":{"file.name":"cmd/run.go","file.line":170},"message":"Elastic Agent started","log":{"source":"elastic-agent"},"process.pid":6140,"agent.version":"8.13.0","ecs.version":"1.6.0"}
[...]

PR #4079, opened 2024-01-12
Move FakeInput tests into their own test suite

faec:
Tests using the FakeInput and FakeShipper components need to run Agent test code in a temporary test directory. To do this, they need to modify the global flags `unversionedHome` and `topPath`. However, they can't safely restore these flags at the end of the test, because it is detected as a data race with access of these flags during the test execution (logically this is not a real race but the race detector doesn't know that).

This PR resolves that by moving these tests into their own test suite, which sets the correct temporary paths at the beginning of each test instead of through a manual helper function call. This way the globals don't need to be explicitly cleaned up after the tests, and there's less risk of omitting the temporary directory setup in a test where it's needed.

The diff is less than it looks like, since (other than moving to a different file) most tests only have changes in their first couple lines.

## Checklist

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/3683


michalpristas (approved):

pierrehilbert:
Unrelated failure https://github.com/elastic/elastic-agent/issues/4076

pierrehilbert:
@faec do we need another round of review or should I force merge this one as the failure is unrelated to this change?

pchila (approved):

Issue #4091, opened 2024-01-17
TestFleetAirGappedUpgrade fails with "context deadline exceeded. upgrade watcher did not start"


belimawr:
## Flaky Test

* **Test Name:** TestFleetAirGappedUpgrade
* **Link:** https://github.com/elastic/elastic-agent/blob/17f048008fcea105d0b86cc90743c7bfb8ce2a41/testing/integration/upgrade_fleet_test.go#L250
* **Branch:** main
* **Artifact Link:** https://buildkite.com/elastic/elastic-agent/builds/6287#018d1146-595d-4ba5-b82e-c5b493b703d3/399-1234

belimawr:
The test is failing because the Elastic-Agent is trying to upgrade to the same version, with the same hash. That causes extracting the archive to fail.

We can see the following error messages in the logs:

```
Failed to unpack upgrade artifact

upgrade to version 8.13.0-SNAPSHOT failed: TarInstaller: creating file /opt/Elastic/Agent/data/elastic-agent-17f048/elastic-agent: open /opt/Elastic/Agent/data/elastic-agent-17f048/elastic-agent: text file busy
```

cmacknz:
Duplicate of https://github.com/elastic/elastic-agent/issues/4089, thanks!

Issue #4089, opened 2024-01-17
TestStandaloneDowngradeToSpecificSnapshotBuild  in daily builds can fail with busy file `elastic-agent` error due to upgrade hash collision


rdner:
## Flaky Test

* **Test Name:** TestStandaloneDowngradeToSpecificSnapshotBuild
* **Link:** https://github.com/elastic/elastic-agent/blob/17f048008fcea105d0b86cc90743c7bfb8ce2a41/testing/integration/upgrade_downgrade_test.go#L26
* **Branch:** main
* **Artifact Link:** https://buildkite.com/elastic/elastic-agent/builds/6287#018d1146-595d-4ba5-b82e-c5b493b703d3 on Windows and Linux
* **Notes:** The file which is used by another process is actually the `elastic-agent` binary itself. Looks like something is modifying this binary while the test is running.

It started failing quite consistently on 12th of January 2024. The commit has no code changes https://github.com/elastic/elastic-agent/commit/158fd47bf44cff0a1e56a4189c9bb411968058a1

### Stack 

Linux

```
  fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneDowngradeToSpecificSnapshotBuild3993626959/002
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneDowngradeToSpecificSnapshotBuild3993626959/002
    fixture.go:820: Components were not modified from the fetched artifact
    fixture.go:615: >> running binary with: [/tmp/TestStandaloneDowngradeToSpecificSnapshotBuild3993626959/001/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/elastic-agent version --binary-only --yaml]
    fixture.go:615: >> running binary with: [/tmp/TestStandaloneDowngradeToSpecificSnapshotBuild3993626959/002/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/elastic-agent version --binary-only --yaml]
    upgrader.go:201: Installing version "8.13.0-SNAPSHOT"
    fixture_install.go:103: [test TestStandaloneDowngradeToSpecificSnapshotBuild] Inside fixture install function
[...]

PR #4095, opened 2024-01-18
Look for the latest build with a non-matching commit hash

rdner:
Upgrade tests were failing because of the matching commit hashes in picked snapshot builds. Multiple builds can have the same commit hash for the agent which leads to file system collisions.

Fixes https://github.com/elastic/elastic-agent/issues/4089
Fixes https://github.com/elastic/elastic-agent/issues/4091


rdner:
The linter is failing in the unchanged code.

testing/integration/upgrade_downgrade_test.go
```diff
@@ -41,25 +42,29 @@ func TestStandaloneDowngradeToSpecificSnapshotBuild(t *testing.T) {
...
+	require.NoError(t, err)
+	startVersion, err := startFixture.ExecVersion(ctx)
 	require.NoError(t, err)
-	if len(builds.Builds) < 2 {
-		t.Skipf("not enough SNAPSHOT builds exist for version %s. Found %d", latestSnapshotVersion.VersionWithPrerelease(), len(builds.Builds))
+
+	// when picking the build we need to find one that has a different commit hash to avoid file system
+	// collisions during the upgrade process.
+	// multiple builds can have different IDs but the same commit hash.
+	build, err := aac.GetLastBuildForVersion(ctx, latestSnapshotVersion.VersionWithPrerelease(), startVersion.Binary.Commit)
```
pchila, line 59: To properly test the specific feature we should choose a build that is not the latest SNAPSHOT build.
If we pick the latest snapshot the result after the upgrade in the form `elastic-agent upgrade x.y.z-SNAPSHOT-<buildid>` would be indistinguishable from a regular `elastic-agent upgrade x.y.z-SNAPSHOT`

pchila (changes requested):
Please restore the test behavior for `TestStandaloneDowngradeToSpecificSnapshotBuild` to NOT use the latest SNAPSHOT build

testing/integration/upgrade_downgrade_test.go
rdner: @pchila can you elaborate more? 

* We get a list of snapshot builds of the same version as we're running
* Then we look for the latest snapshot that **does not match** the current hash of `main`
* This tests upgrading to the previous SNAPSHOT

Under what circumstances there will be a top/last snapshot with a different commit hash? My idea was that the commit hash filter will always find the previous snapshot that was not built with the same commit hash, is not what we want here?

rdner: Changed in https://github.com/elastic/elastic-agent/pull/4095/commits/dd512b44169c539789719942ed60aa9fb3924c9f

testing/upgradetest/upgrader.go
```diff
@@ -233,7 +237,7 @@ func PerformUpgrade(
 		}
 	}
 
-	logger.Logf("Upgrading from version %q (%s) to version %q (%s)", startParsedVersion, startVersionInfo.Binary.Commit, endVersionInfo.Binary.String(), endVersionInfo.Binary.Commit)
+	logger.Logf("Upgrading from version \"%s-%s\" to version \"%s-%s\"", startParsedVersion, startVersionInfo.Binary.Commit, endVersionInfo.Binary.String(), endVersionInfo.Binary.Commit)
```
rdner, line 240: To make it consistent with https://github.com/elastic/elastic-agent/blob/17f048008fcea105d0b86cc90743c7bfb8ce2a41/testing/integration/upgrade_fleet_test.go#L229

testing/integration/upgrade_downgrade_test.go
```diff
@@ -53,18 +56,34 @@ func TestStandaloneDowngradeToSpecificSnapshotBuild(t *testing.T) {
...
 
-	// when picking the build we need to find one that has a different commit hash to avoid file system
-	// collisions during the upgrade process.
-	// multiple builds can have different IDs but the same commit hash.
-	build, err := aac.GetLastBuildForVersion(ctx, latestSnapshotVersion.VersionWithPrerelease(), startVersion.Binary.Commit)
-	if errors.Is(err, tools.ErrNoMatchingBuild) {
+	// We need to find a build which is not the latest (so, we can make sure we address it by a build ID
+	// in the version prefix, e.g. x.y.z-SNAPSHOT-<buildid>) and does not have the same commit hash
+	// as the currently running binary (so, we don't have a file system collision).
+	// Multiple builds can have different IDs but the same commit hash.
```
pchila, line 62: :+1: :heart:

pchila (approved):

blakerouse (approved):
Looks good.

PR #4100, opened 2024-01-18
Fix creation of containing directories for files in tar packages

### LINKED ISSUE #4093: [Unprivileged] Upgrade broken due to permission issues
Upgrading an Elastic Agent that is installed with unprivileged fails to perform upgrade unless the tarball is built with the correct permissions.

`
Failed trigger upgrade of daemon: TarInstaller: creating file /opt/Elastic/Agent/data/elastic-agent-5b3535/package.version: open /opt/Elastic/Agent/data/elastic-agent-5b3535/package.version: permission denied
`

```
root@ogc-linux-arm64-ubuntu-2204-upgrade-cb2f:/opt/Elastic/Agent/data# ls -lh
total 8.0K
-rw------- 1 elastic-agent elastic-agent    0 Jan 17 18:37 agent.lock
drw-r----- 2 elastic-agent elastic-agent 4.0K Jan 17 18:37 elastic-agent-5b3535
drwxr-x--- 5 elastic-agent elastic-agent 4.0K Jan 17 18:37 elastic-agent-666fd7
```

The `elastic-agent-5b3535` is the extracted directory for the upgrade and it doesn't have the correct permissions. It should have the same permissions as `elastic-agent-666fd7`.

https://github.com/elastic/elastic-agent/blob/71c7cb3e51841f555f305dd373f86fe44cd5576a/internal/pkg/agent/application/upgrade/step_unpack.go#L217

Only applies a bitmask for `0770` meaning that the contents of the tarball have an a effect on the extracted contents.

pchila:
## What does this PR do?
This change fixes the creation of destination directories for file contained in a tar package.
If a file needs to be extracted to a destination directory that does not yet exist we create the necessary directories on the fly with permission `0750`.
If we encounter the same directories after the file, we fix the permissions according to what is specified in the tar archive

## Why is it important?
When running without superuser privileges we need to create a directory structure that has permissions that do not block access to the current user

## Checklist

- ~~[x] I have added tests that prove my fix is effective or that my feature works~~

## Author's Checklist

- [ ]

## Related issues

- Closes #4093

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/step_unpack_test.go
+++ b/internal/pkg/agent/application/upgrade/step_unpack_test.go
@@ -1,3 +1,7 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
 package upgrade
 
 import (
--- a/internal/pkg/agent/application/upgrade/step_unpack_test.go
+++ b/internal/pkg/agent/application/upgrade/step_unpack_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
@@ -119,10 +118,8 @@ func TestUpgrader_unpack(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		prevTop := paths.Top()
 		t.Run(tt.name, func(t *testing.T) {
 			testTop := t.TempDir()
-			paths.SetTop(testTop)
 			testDataDir := filepath.Join(testTop, "data")
 			err := os.MkdirAll(testDataDir, 0o777)
 			assert.NoErrorf(t, err, "error creating initial structure %q", testDataDir)
@@ -134,13 +131,12 @@ func TestUpgrader_unpack(t *testing.T) {
 			archiveFile, err := tt.args.archiveGenerator(t, tt.args.archiveFiles)
 			require.NoError(t, err, "creation of test archive file failed")
 
-			got, err := u.unpack(tt.args.version, archiveFile)
+			got, err := u.unpack(tt.args.version, archiveFile, testDataDir)
 			if !tt.wantErr(t, err, fmt.Sprintf("unpack(%v, %v)", tt.args.version, tt.args.archiveGenerator)) {
 				return
 			}
 			assert.Equalf(t, tt.want, got, "unpack(%v, %v)", tt.args.version, tt.args.archiveGenerator)
 		})
-		paths.SetTop(prevTop)
 	}
 }
 
--- a/internal/pkg/agent/application/upgrade/step_unpack_test.go
+++ b/internal/pkg/agent/application/upgrade/step_unpack_test.go
@@ -1,0 +1,192 @@
+package upgrade
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+)
+
+const foo_component_spec = `
+version: 2
+inputs:
+  - name: foobar
+    description: "Foo input"
+    platforms:
+      - linux/amd64
+      - linux/arm64
+      - darwin/amd64
+      - darwin/arm64
+    outputs:
+      - elasticsearch
+      - kafka
+      - logstash
+    command:
+      args:
+        - foo
+        - bar
+        - baz
+`
+
+type fileType uint
+
+const (
+	REGULAR fileType = iota
+	DIRECTORY
+	SYMLINK
+)
+
+type files struct {
+	fType   fileType
+	path    string
+	content string
+	mode    fs.FileMode
+}
+
+func (f files) Name() string {
+	return path.Base(f.path)
+}
+
+func (f files) Size() int64 {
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/application/upgrade/step_unpack_test.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+			want:    "abcdef",
+			wantErr: assert.NoError,
+		},
+	}
+	for _, tt := range tests {
+		prevTop := paths.Top()
+		t.Run(tt.name, func(t *testing.T) {
+			testTop := t.TempDir()
+			paths.SetTop(testTop)
```
blakerouse, line 125: There was a race condition that @faec just fixed related to adjusting paths in tests. Would it be possible to just pass in the top directory into the unpack function. That way the test can use that directory versus changing a global.

blakerouse (changes requested):

internal/pkg/agent/application/upgrade/step_unpack_test.go
pchila: Removed the references to `paths.Data()` that required us to set `paths.Top()` in the unit test.

blakerouse (approved):
Looks good. Thanks for the change to remove the need for path adjustments in the unit tests.

cmacknz:
This should have a changelog entry shouldn't it? It's a bug fix. It's not one anyone is hitting today, that doesn't mean we shouldn't document the change.

internal/pkg/agent/application/upgrade/step_unpack_test.go
```diff
@@ -0,0 +1,197 @@
...
+	return f.fType == DIRECTORY
+}
+
+func (f files) Sys() any {
+	return nil
+}
+
+type createArchiveFunc func(*testing.T, []files) (string, error)
+
+func TestUpgrader_unpack(t *testing.T) {
```
cmacknz, line 89: 1. Thank you for this, this is great.
2. This test still passes if you take out the `os.Stat` call and associated logic


```diff
diff --git a/internal/pkg/agent/application/upgrade/step_unpack.go b/internal/pkg/agent/application/upgrade/step_unpack.go
index 4f5b0bd744..c7d8100335 100644
--- a/internal/pkg/agent/application/upgrade/step_unpack.go
+++ b/internal/pkg/agent/application/upgrade/step_unpack.go
@@ -10,7 +10,6 @@ import (
        "compress/gzip"
        "fmt"
        "io"
-       "io/fs"
        "os"
        "path/filepath"
        "runtime"
@@ -214,20 +213,20 @@ func untar(log *logger.Logger, version string, archivePath, dataDir string) (str
                case mode.IsDir():
                        log.Debugw("Unpacking directory", "archive", "tar", "file.path", abs)
                        // remove any world permissions from the directory
-                       _, err = os.Stat(abs)
-                       if errors.Is(err, fs.ErrNotExist) {
-                               if err := os.MkdirAll(abs, mode.Perm()&0770); err != nil {
-                                       return "", errors.New(err, "TarInstaller: creating directory for file "+abs, errors.TypeFilesystem, errors.M(errors.MetaKeyPath, abs))
-                               }
-                       } else if err != nil {
-                               return "", errors.New(err, "TarInstaller: stat() directory for file "+abs, errors.TypeFilesystem, errors.M(errors.MetaKeyPath, abs))
-                       } else {
-                               // set the appropriate permissions
-                               err = os.Chmod(abs, mode.Perm()&0o770)
-                               if err != nil {
-                                       return "", errors.New(err, fmt.Sprintf("TarInstaller: setting permissions %O for directory %q", mode.Perm()&0o770, abs), errors.TypeFilesystem, errors.M(errors.MetaKeyPath, abs))
-                               }
+                       // _, err = os.Stat(abs)
+                       // if errors.Is(err, fs.ErrNotExist) {
+                       if err := os.MkdirAll(abs, mode.Perm()&0770); err != nil {
+                               return "", errors.New(err, "TarInstaller: creating directory for file "+abs, errors.TypeFilesystem, errors.M(errors.MetaKeyPath, abs))
                        }
+                       // } else if err != nil {
+                       //      return "", errors.New(err, "TarInstaller: stat() directory for file "+abs, errors.TypeFilesystem, errors.M(errors.MetaKeyPath, abs))
+                       // } else {
+                       //      // set the appropriate permissions
+                       //      err = os.Chmod(abs, mode.Perm()&0o770)
+                       //      if err != nil {
+                       //              return "", errors.New(err, fmt.Sprintf("TarInstaller: setting permissions %O for directory %q", mode.Perm()&0o770, abs), errors.TypeFilesystem, errors.M(errors.MetaKeyPath, abs))
+                       //      }
+                       // }
                default:
                        return "", errors.New(fmt.Sprintf("tar file entry %s contained unsupported file type %v", fileName, mode), errors.TypeFilesystem, errors.M(errors.MetaKeyPath, fileName))
                }
(END)
```

```sh
go test ./internal/pkg/agent/application/upgrade/... -run TestUpgrader_unpack -count 1 -v
?       github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact/download/localremote     [no test files]
=== RUN   TestUpgrader_unpack
=== RUN   TestUpgrader_unpack/targz_with_file_before_containing_folder
--- PASS: TestUpgrader_unpack (0.00s)
    --- PASS: TestUpgrader_unpack/targz_with_file_before_containing_folder (0.00s)
PASS
ok      github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade 0.591s
```

pchila: The test passes because of the default 0750 permission when we create the directory on the fly (so we are able to write the file anyways). The code you removed is needed to restore permissions once we come across the directory in the tar archive, but that is not in the assertions. I will modify the test

pchila:
> This should have a changelog entry shouldn't it? It's a bug fix. It's not one anyone is hitting today, that doesn't mean we shouldn't document the change.
@cmacknz  not sure about the changelog as the change that introduced the issue is related to the unprivileged elastic-agent feature. Before such change, directories created on the fly would always have `0755` permissions and that would not be corrected if we encountered the directory with different permission later in the tar.gz archive. So a customer cannot have encountered the exact same bug in the released versions.
I added a changelog regardless as we improve on the previous behavior by fixing the permissions as specified in the tar.gz  (minus the world ones, which I think was the original intention of the unprivileged change)

PR #4106, opened 2024-01-18
Remove all io/ioutil imports.

cmacknz:
What the title says. io/ioutil is deprecated and I kept seeing the linter fail because of it, so remove all imports of io/ioutil. The linter will warn us if we bring it back.

For the most part this was just simple renames, except the conversion to os.ReadDir which changes the return type.


dev-tools/mage/copy.go
```diff
@@ -104,20 +104,20 @@ func (t *CopyTask) fileCopy(src, dest string, info os.FileInfo) error {
...
 	if mode == 0 {
-		mode = info.Mode()
+		mode = entry.Type()
 	}
 	if err := os.MkdirAll(dest, mode&os.ModePerm); err != nil {
 		return errors.Wrap(err, "failed creating dirs")
 	}
 
-	contents, err := ioutil.ReadDir(src)
+	contents, err := os.ReadDir(src)
```
cmacknz, line 120: Conversion to os.ReadDir, the return type changed to fs.DirEntry. Double check this code.

dev-tools/mage/gotest.go
```diff
@@ -126,14 +125,14 @@ func DefaultTestBinaryArgs() TestBinaryArgs {
 // Use MODULE=module to run only tests for `module`.
 func GoTestIntegrationForModule(ctx context.Context) error {
 	module := EnvOr("MODULE", "")
-	modulesFileInfo, err := ioutil.ReadDir("./module")
+	modulesDirEntry, err := os.ReadDir("./module")
```
cmacknz, line 128: Conversion to os.ReadDir, the return type changed to fs.DirEntry. Double check this code.

cmacknz:
There was only one place that resulted in more code changes than just a simple rename, I've commented it on it to point out where more careful review is required.

blakerouse (approved):
Thanks for this change.

I reviewed the mage copy code, nothing stood out to me that maybe me think the logic didn't match and work the same.

cmacknz:
Yeah that Copy change broke mage package but I don't see how yet.

michalpristas (approved):

cmacknz:
Only failure is the Windows 10 unit tests failing to get a buildkite agent to run on. https://buildkite.com/elastic/elastic-agent/builds/6439

The windows 11 unit tests and integration tests passed, force merging.

Issue #4143, opened 2024-01-25
[Flaky Test]: TestStandaloneUpgradeUninstallKillWatcher – TarInstaller: creating file open package.version: permission denied


cmacknz:
### Failing test case

TestStandaloneUpgradeUninstallKillWatcher

### Error message

Failed trigger upgrade of daemon: TarInstaller: creating file /opt/Elastic/Agent/data/elastic-agent-cb1c24/package.version: open /opt/Elastic/Agent/data/elastic-agent-cb1c24/package.version: permission denied

### Build

Main: https://buildkite.com/elastic/elastic-agent/builds/6552#018d412d-b1b7-4958-8294-b8f761011c38
Branch: https://buildkite.com/elastic/elastic-agent/builds/6555#018d4136-3471-4cbb-b66e-3cab14032f1a

### OS

Linux

### Stacktrace and notes

```markdown
=== RUN   TestStandaloneUpgradeUninstallKillWatcher
    fetcher.go:91: Using existing artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz
    fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneUpgradeUninstallKillWatcher3431983411/001
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneUpgradeUninstallKillWatcher3431983411/001
    fixture.go:829: Components were not modified from the fetched artifact
    fixture.go:624: >> running binary with: [/tmp/TestStandaloneUpgradeUninstallKillWatcher3431983411/001/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/elastic-agent version --binary-only --yaml]
    fetcher.go:91: Using existing artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz
    fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneUpgradeUninstallKillWatcher3431983411/002
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestStandaloneUpgradeUninstallKillWatcher3431983411/002
    fixture.go:829: Components were not modified from the fetched artifact
[...]

PR #4146, opened 2024-01-25
flaky TestStandaloneUpgradeUninstallKillWatcher

leehinman:
## What does this PR do?

In TestStandaloneUpgradeUninstallKillWatcher,  we need to wait for watcher process to start before starting the uninstall.  If watcher hasn't started before the uninstall begins, the uninstall won't find it's pid, so won't kill it, but the watcher can start before the directory gets deleted.

## Why is it important?

Fixes flaky test.

## Checklist

## Author's Checklist

- [ ]

```
mage integration:single TestStandaloneUpgradeUninstallKillWatcher
```

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/upgrade_uninstall_test.go
+++ b/testing/integration/upgrade_uninstall_test.go
@@ -9,7 +9,6 @@ package integration
 import (
 	"context"
 	"errors"
-	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


testing/integration/upgrade_uninstall_test.go
```diff
@@ -80,6 +81,18 @@ func TestStandaloneUpgradeUninstallKillWatcher(t *testing.T) {
 	}
 	require.NoError(t, err)
 
+	// on windows watcher needs to start before uninstall,
+	// otherwise you can get a race condition where watcher hasn't
+	// started yet, so it doesn't show up in PID list, but will
+	// start before deletion of directory that will lead to
+	// sharing violation on Windows
```
cmacknz, line 88: Why is this only true on Linux? Won't we also fail to find the watcher PID there?

It won't cause a sharing violation, but it will let the watcher keep running and potentially interfere with the next test, which is the reason why we started killing it.

cmacknz: I think we should get rid of the GOOS restriction here.

leehinman: switched to all OSes

cmacknz (approved):

cmacknz:
We got https://github.com/elastic/elastic-agent/issues/4143 and a new one in https://github.com/elastic/elastic-agent/issues/4148 that I haven't observed before.

The only code change here is isolated to a specific test so these are unrelated. @leehinman want me to force merge this?

leehinman:
> The only code change here is isolated to a specific test so these are unrelated. @leehinman want me to force merge this?

yeah, lets try to get at least one flaky test closed :-)

PR #4147, opened 2024-01-25
Pass context with timeout to FQDN lookup

ycombinator:
This PR improves the FQDN lookup function calls by passing a context with a 1 minute timeout, thus preventing the calls from blocking indefinitely due to DNS issues and such.

Depends on: https://github.com/elastic/go-sysinfo/pull/199 and a new release of that library.


internal/pkg/agent/application/info/agent_metadata.go
```diff
@@ -207,7 +211,10 @@ func (i *AgentInfo) ECSMetadataFlatMap(l *logger.Logger) (map[string]interface{}
 	info := sysInfo.Info()
 	hostname := info.Hostname
 	if features.FQDN() {
-		fqdn, err := sysInfo.FQDN()
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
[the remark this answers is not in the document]
ycombinator: Done in 92151dabf0.

internal/pkg/composable/providers/host/host.go
```diff
@@ -157,3 +150,20 @@ func getHostInfo(log *logger.Logger) func() (map[string]interface{}, error) {
...
+		return hostInfo.Hostname
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+
+	fqdn, err := host.FQDNWithContext(ctx)
+	if err != nil {
+		log.Debugf("unable to lookup FQDN: %s, using hostname = %s", err.Error(), hostInfo.Hostname)
+		return ""
```
cmacknz, line 165: The parent function calling this can return an error? Why do we drop the error?

cmacknz:
Any ideas on how to test for this?

Perhaps blocking DNS with iptables as is done in the air gapped fleet test? https://github.com/elastic/elastic-agent/blob/ece91fe559940316376e3c61ce33903881e3dd9c/testing/integration/upgrade_fleet_test.go#L349

There is a mock DNS package that looks like it might help https://github.com/foxcpp/go-mockdns

Possibly we could use that in beats as well.

cmacknz:
Unlike Beats (https://github.com/elastic/beats/pull/37756#discussion_r1475201892) I *think* both calls here happen periodically to allow us to recover from failure.

ycombinator:
> Any ideas on how to test for this?

I added a test for this in `go-sysinfo`: https://github.com/elastic/go-sysinfo/blob/489579dd56bb8a6f8e32a0112f6119887000a58c/providers/shared/fqdn_test.go#L62-L67.  Let me know if you think that's not sufficient.

internal/pkg/composable/providers/host/host.go
ycombinator: I _think_ the idea here was to just log the error and fallback to the OS-provided hostname because this code will be called again on a timer: https://github.com/elastic/elastic-agent/blob/ece91fe559940316376e3c61ce33903881e3dd9c/internal/pkg/composable/providers/host/host.go#L63-L72

However, I see now that the timer loop itself just logs the error and continues to the next iteration.  So this function here should return the error.  Fixing.

ycombinator: I _think_ the idea here was to just log the error and fallback to the OS-provided hostname because this code will be called again on a timer: https://github.com/elastic/elastic-agent/blob/ece91fe559940316376e3c61ce33903881e3dd9c/internal/pkg/composable/providers/host/host.go#L63-L72

If we return an error, there will be no way to fallback to the OS-provided hostname.

internal/pkg/agent/application/info/agent_metadata.go
```diff
@@ -151,7 +152,10 @@ func (i *AgentInfo) ECSMetadata(l *logger.Logger) (*ECSMeta, error) {
...
```
michalpristas, line 155: isn't all this (L152-L164) what you have in `getHostName`?

ycombinator: Yes, good spot.  I need to move that function to a common location so I can share it between two two packages.

ycombinator: Refactored in f4878696dc. Thanks for catching.

ycombinator: Refactored in f4878696dc. Thanks for catching.

ycombinator: Added unit tests as well: f2f3355cc2

internal/pkg/composable/providers/host/host.go
cmacknz: Can you add that comment to the code so nobody misses that in the future?

It seems little bit weird that the function being called is making assumptions about how its errors are getting used, but I'm not going to make you try to fix that.

cmacknz:
> I added a test for this in go-sysinfo: https://github.com/elastic/go-sysinfo/blob/489579dd56bb8a6f8e32a0112f6119887000a58c/providers/shared/fqdn_test.go#L62-L67. Let me know if you think that's not sufficient.

If its tested in go-sysinfo we don't need to retest the same thing here, thanks.

cmacknz (approved):

PR #4155, opened 2024-01-27
Elastic Defend arm64 is not supported on Windows yet

intxgo:
## What does this PR do?

This PR enhances elastic-endpoint spec to gracefully refrain from ever installing elastic-endpoint on Windows ARM, because Elastic Defend is not supported on it.

The win32 API used `IsWow64Process2` is available only on Windows >= Win10, however Agent is currently compiled using Go 1.21 which requires at least Windows 10 or Server 2016.

The native architecture constants can be found [here on MSDN web](https://learn.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants), however I believe it's unimportant to port them all since Elastic Defend is compiled only for x86_64 for Windows.

Reading environment variables could be an alternative solution, as Windows sets them by default, examples:
``` 
PROCESSOR_ARCHITECTURE=ARM64
PROCESSOR_IDENTIFIER=ARMv8 (64-bit) Family 8 Model 0 Revision   0, QEMU
```

```
PROCESSOR_ARCHITECTURE=AMD64
PROCESSOR_IDENTIFIER=Intel64 Family 6 Model 151 Stepping 2, GenuineIntel
```

```
PROCESSOR_ARCHITECTURE=AMD64
PROCESSOR_IDENTIFIER=AMD64 Family 25 Model 97 Stepping 2, AuthenticAMD
```
However environment variables could be easily modified, therefore calling native API is better solution.

## Why is it important?

Elastic Defend is not provided yet for Windows ARM (Microsoft Surface X or VM on Apple M1/M2/M3) but since Windows Elastic Agent amd64 works fine on Windows ARM in emulation people install it.

Handling gracefully the incompatibility enables installing Agent across a fleet of Windows AMD64 and ARM64 devices with the same policy.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...


nfritts (approved):
This looks good to me from Endpoints perspective

internal/pkg/agent/application/info/inject_config_windows.go
```diff
@@ -0,0 +1,87 @@
...
+	var kernel32 = syscall.NewLazyDLL("kernel32.dll")
+	var isWow64Process2 = kernel32.NewProc("IsWow64Process2")
+
+	var currentProcessHandle, _ = syscall.GetCurrentProcess()
+	_, _, _ = isWow64Process2.Call(uintptr(currentProcessHandle), uintptr(unsafe.Pointer(&processMachine)), uintptr(unsafe.Pointer(&nativeMachine)))
+
+	var nativeMachineStr string
+
+	switch nativeMachine {
+	case 0x8664:
```
pchila, line 77: Nitpick: it would be nice to define constants for values corresponding to "amd64" and "arm64".
We could use the same names as https://learn.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants (even better if we put a comment with a link to the docs)

pchila (approved):
Aside from the small nitpick, the change looks good to me

internal/pkg/agent/application/info/inject_config_windows.go
intxgo: fixed

intxgo:
I'm going to look at the unit test failure

internal/pkg/agent/application/info/inject_config_windows.go
```diff
@@ -0,0 +1,92 @@
...
+			"config": paths.Config(),
+			"home":   paths.Home(),
+			"logs":   paths.Logs(),
+		},
+		"host": map[string]interface{}{
+			"id": hostInfo.Info().UniqueID,
+		},
+		"runtime.os":             runtime.GOOS,
+		"runtime.arch":           runtime.GOARCH,
+		"runtime.native_arch":    nativeArch(),
```
cmacknz, line 55: Rather than duplicating the entire map here, you should be able to call the non-windows (or lets say the common) version and then just add the single additional `runtime.native_arch` key.

For example:

```go
globalCfg, err := agentGlobalConfig()
if err != nil {
    return fmt.Errorf("getting common global config: %w", err)
}

globalCfg["runtime.native_arch"] = nativeArch()

return globalCfg
```

Essentially just augmenting the common configuration with the windows key, rather than duplicating it for both cases.

internal/pkg/agent/application/info/inject_config_windows.go
```diff
@@ -0,0 +1,92 @@
...
+	}, nil
+}
+
+func nativeArch() string {
+	var processMachine, nativeMachine uint16
+
+	var kernel32 = syscall.NewLazyDLL("kernel32.dll")
+	var isWow64Process2 = kernel32.NewProc("IsWow64Process2")
+
+	var currentProcessHandle, _ = syscall.GetCurrentProcess()
```
cmacknz, line 71: Do you need to close this? We recently had a severe memory leak from forgetting to close syscall handles on Windows so this is a sensitive subject :)

internal/pkg/agent/application/info/inject_config_windows.go
intxgo: thanks, indeed. I'm also now wondering if this should rather be added to the go-sysinfo package

cmacknz: I can't think of a reason why you couldn't put it in go-sysinfo, that will allow us to reuse this outside the agent.

Easiest thing to do is open a PR there.

intxgo: I've found out that go-sysinfo is only extending golang.org/x/sys/windows, where the latter already includes `IsWow64Process2` so I use it directly instead of loading the dll.

internal/pkg/agent/application/info/svc_windows.go
```diff
@@ -21,9 +23,10 @@ func RunningUnderSupervisor() bool {
 	if err != nil {
 		return false
 	}
-	defer windows.FreeSid(serviceSid)
+	defer func() { _ = windows.FreeSid(serviceSid) }()
```
intxgo, line 26: unrelated change, fixing lint error: Error return value of `windows.FreeSid` is not checked (errcheck)

internal/pkg/agent/application/info/svc_windows.go
```diff
@@ -21,9 +23,10 @@ func RunningUnderSupervisor() bool {
...
 
-	t, err := windows.OpenCurrentProcessToken()
+	var t windows.Token
+	err = windows.OpenProcessToken(windows.CurrentProcess(), windows.TOKEN_QUERY|windows.TOKEN_QUERY_SOURCE, &t)
```
intxgo, line 29: unrelated change, fixing lint error: SA1019: windows.OpenCurrentProcessToken is deprecated: Explicitly call OpenProcessToken(CurrentProcess(), ...) with the desired access instead, or use GetCurrentProcessToken for a TOKEN_QUERY token.  (staticcheck)

internal/pkg/agent/application/info/svc_windows.go
```diff
@@ -21,9 +23,10 @@ func RunningUnderSupervisor() bool {
 	if err != nil {
 		return false
 	}
-	defer windows.FreeSid(serviceSid)
+	defer func() { _ = windows.FreeSid(serviceSid) }()
 
-	t, err := windows.OpenCurrentProcessToken()
+	var t windows.Token
+	err = windows.OpenProcessToken(windows.CurrentProcess(), windows.TOKEN_QUERY, &t)
```
intxgo, line 29: unrelated change, fixing lint error: SA1019: windows.OpenCurrentProcessToken is deprecated: Explicitly call OpenProcessToken(CurrentProcess(), ...) with the desired access instead, or use GetCurrentProcessToken for a TOKEN_QUERY token. (staticcheck)

intxgo:
could you review this again since the implementation has changed significantly

@michalpristas @pchila 

The unit test error on Windows Server 2016 (`Failed to find IsWow64Process2 procedure in kernel32.dll: The specified procedure could not be found.`) will be resolved by `go-sysinfo` update, as soon as [the improvement](https://github.com/elastic/go-sysinfo/pull/201) is merged

michalpristas (approved):
i think it looks good, fix from sysinfo is included for older windows versions as well.
haven't tested this on arm machine

cmacknz:
The one test failure here is unrelated:

```
	Received unexpected error:
        	            	failed to download https://snapshots.elastic.co/8.13.0-e2cda7bd/downloads/beats/elastic-agent/elastic-agent-8.13.0-SNAPSHOT-linux-x86_64.tar.gz: failed to write file /home/ubuntu/agent/.agent-testing/artifact/elastic-agent-8.13.0-SNAPSHOT-linux-x86_64.tar.gz: stream error: stream ID 1; INTERNAL_ERROR; received from peer
```

cmacknz:
Going to force merge this one.

Issue #4080, opened 2024-01-12
[OTel Collector] Agent should accept all schemes allowed by the OTel `--config` flag


ycombinator:
**Describe the enhancement:**

According to https://github.com/open-telemetry/opentelemetry-collector/tree/main/service#how-to-provide-configuration, OTel configuration can be supplied to the collector via the `--config` flag using a few different schemes. 

When Agent is being run in OTel mode, it should accept all the different schemes mentioned in the OTel documentation. 

**Describe a specific use case for the enhancement or feature:**

To allow existing OTel users to run Agent in OTel mode with minimal changes.

**What is the definition of done?**

Agent running in OTel mode should be able accept OTel configuration via the `--config` flag using all the schemes described in the OTel documentation.

At the moment, Agent decides if it's running in OTel mode if he value of the `--config` flag conforms to a certain filename pattern.  We may need to change how this decision is made.  After implementing https://github.com/elastic/elastic-agent/issues/3903, we may want to say that Agent is running in OTel mode simply if the `elastic-agent otel ...` subcommand was run, instead of basing it on the filename pattern of the `--config` flag value.

Issue #3903, opened 2023-12-12
[OTel Collector] Run embedded OTel Collector via `otel` sub-command


ycombinator:
Users should be able to run the OTel Collector embedded in Elastic Agent via the `elastic-agent otel` subcommand.  This will allow OTel users to hang as many otel-specific options off the command without risk of collisions with existing Agent options. So, for example:

```
elastic-agent otel --config
elastic-agent otel --feature-gates
```

Also, this way, OTel users need only change the command and the sub-command to use it over any other distribition, e.g. `otelcol ...` -> `elastic-agent otel ...`.

This is similar to the approach taken by [`minikube kubectl`](https://minikube.sigs.k8s.io/docs/handbook/kubectl/).

Based on this discussion: https://github.com/elastic/elastic-agent/pull/3800#discussion_r1410588093

Issue #3870, opened 2023-12-05
[OTel Collector] Process references to other files from OTel Collector configuration file


ycombinator:
As per https://opentelemetry.io/docs/collector/configuration/:

> [OTel Collector configuration] can also include other files, causing the Collector to merge them in a single in-memory representation of the YAML configuration:

> ```yaml
> receivers:
>   otlp:
>     protocols:
>       grpc:
> 
> exporters: ${file:exporters.yaml}
> 
> service:
>   extensions: []
>   pipelines:
>     traces:
>       receivers: [otlp]
>       processors: []
>       exporters: [otlp]
> ```

Elastic Agent running in OTel mode should be able to process such references, ensuring that the referenced file's contents are loaded into memory, parsed, and merged with the referencing file's contents.

PR #4159, opened 2024-01-29
[OTel] Introduce otel subcommand

michalpristas:
This PR introduces `otel` subcommand to elastic agent.

Pre PR agent detected mode based on a name passed using a config flag. If it was otel/otelcol/otlp it switched to otel mode otherwise it used normal agent mode. This functionality is no longer present.

When running using `run` command, pre-otel behavior is used regardless of config file name.

When using `otel` subcommand agent starts in otel mode keeping `status` subcommand  working. 
In this PR no other subcommand are added such as validate. 
multiple `--config` flags are allowed with values supported with normal collector as described here: [how-to-provide-configuration](https://github.com/open-telemetry/opentelemetry-collector/tree/main/service#how-to-provide-configuration)

`--set` is supported for overriding config options as described here [--set](https://github.com/open-telemetry/opentelemetry-collector/tree/main/service#the-format-and-limitations-of---set)

Fixes: https://github.com/elastic/elastic-agent/issues/3903
Fixes: https://github.com/elastic/elastic-agent/issues/4080
Fixes: https://github.com/elastic/elastic-agent/issues/3870


michalpristas:
`status` and `run` are covered by existing integ tests

ycombinator:
I tested this PR with a few different ways of specifying OTel configuration and they all worked as expected.  Adding a few "Fixes: #xxxx" to the issue description to reflect this.

internal/pkg/agent/cmd/otel_flags.go
```diff
@@ -0,0 +1,51 @@
...
+	var sets []string
+	setVals, err := cmd.Flags().GetStringArray(setFlagName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve set flags: %w", err)
+	}
+
+	for _, s := range setVals {
+		idx := strings.Index(s, "=")
+		if idx == -1 {
+			// No need for more context, see TestSetFlag/invalid_set.
```
ycombinator, line 45: > see TestSetFlag/invalid_set

Where is this test case?

internal/pkg/agent/cmd/otel_flags.go
```diff
@@ -0,0 +1,51 @@
...
+			return nil, fmt.Errorf("missing equal sign for set value %q", s)
+		}
+		sets = append(sets, "yaml:"+strings.TrimSpace(strings.ReplaceAll(s[:idx], ".", "::"))+": "+strings.TrimSpace(s[idx+1:]))
+	}
```
ycombinator, line 49: Nice to have: extract into own function(s) and add unit tests.

internal/pkg/agent/cmd/otel.go
```diff
@@ -0,0 +1,123 @@
...
+			})
+		},
+		SilenceUsage:  true,
+		SilenceErrors: true,
+	}
+
+	cmd.SetHelpFunc(func(c *cobra.Command, s []string) {
+		c.InheritedFlags().VisitAll(func(f *pflag.Flag) {
+			f.Hidden = true
+		})
```
ycombinator, line 50: Nit: repeated in `PreRun` above; consider extracting into common function, assuming the reason for this hiding behavior is the same in both places.

ycombinator: Also, would be good to add a comment as to why we're hiding inherited flags.

internal/pkg/agent/cmd/run.go
```diff
@@ -208,7 +177,7 @@ func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override cf
 	}
 
 	// agent ID needs to stay empty in bootstrap mode
-	createAgentID := true
+	createAgentID := !runAsOtel
```
ycombinator, line 180: Just double checking: this change is not strictly related to this PR but a fix for a small bug you found while working on this PR?

internal/pkg/agent/cmd/run.go
```diff
@@ -405,7 +374,11 @@ LOOP:
 
 func loadConfig(ctx context.Context, override cfgOverrider, runAsOtel bool) (*configuration.Configuration, error) {
 	if runAsOtel {
-		return configuration.DefaultConfiguration(), nil
+		defaultCfg := configuration.DefaultConfiguration()
+		// disable monitoring to avoid injection
```
ycombinator, line 378: Appreciate the comment here but I didn't follow what you mean by "avoid injection".  Could you expand on this a bit in the comment, please?

internal/pkg/otel/run.go
```diff
@@ -95,3 +74,11 @@ func newSettings(configPath string, version string) (*otelcol.CollectorSettings,
 		DisableGracefulShutdown: true,
 	}, nil
 }
+
+func makeMapProvidersMap(providers ...confmap.Provider) map[string]confmap.Provider {
```
ycombinator, line 78: Easy candidate for a unit test?

internal/pkg/agent/application/paths/common.go
```diff
@@ -20,6 +20,8 @@ import (
 const (
 	// DefaultConfigName is the default name of the configuration file.
 	DefaultConfigName = "elastic-agent.yml"
+	// DefaultConfigName is the default name of the configuration file.
```
ycombinator, line 23: Comment needs to be updated.

internal/pkg/agent/application/paths/common.go
```diff
@@ -146,8 +148,19 @@ func SetConfig(path string) {
 
 // ConfigFile returns the path to the configuration file.
 func ConfigFile() string {
+
+	return configFileWithDefaultOverride(DefaultConfigName)
+}
+
+// ConfigFile returns the path to the configuration file.
```
ycombinator, line 155: Comment needs to be updated.

internal/pkg/agent/application/paths/common.go
```diff
@@ -146,8 +148,19 @@ func SetConfig(path string) {
...
+func OtelConfigFile() string {
+	return configFileWithDefaultOverride(DefaultOtelConfigName)
+}
+
+// ConfigFile returns the path to the configuration file.
```
ycombinator, line 160: Comment needs to be updated.

ycombinator:
This is looking good.  Left a few minor comments/questions.

internal/pkg/agent/cmd/run.go
michalpristas: yes, on mac it fails to run otel because agent tries to create agent id and do not have permissions if not running as root

internal/pkg/agent/cmd/run.go
michalpristas: updated a comment. it's not really an issue now as we're passing `inputs: []`
added this just to be sure and have default reflecting use case a bit more

ycombinator (approved):
LGTM.

michalpristas:
@cmacknz this is similar case to PR force-merged recently.
do you see some requirements missing preventing proceeding with the same path?

cmacknz (approved):
The integration tests use the `otel` sub-command so I'm fine to bypass the unit test coverage check.

Issue #4152, opened 2024-01-26
[Flaky Test]: TestStandaloneUpgradeUninstallKillWatcher / TestStandaloneDowngradeToSpecificSnapshotBuild / TestStandaloneUpgradeRetryDownload – target version has the same commit hash


pierrehilbert:
### Failing test case

TestStandaloneUpgradeUninstallKillWatcher / TestStandaloneDowngradeToSpecificSnapshotBuild / TestStandaloneUpgradeRetryDownload

### Error message

target version has the same commit hash

### Build

https://buildkite.com/elastic/elastic-agent/builds/6583

### OS

Linux, Windows

### Stacktrace and notes

```markdown
Stacktrace: 
=== RUN   TestStandaloneUpgradeUninstallKillWatcher
    fetcher.go:91: Using existing artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip
    fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip to C:\Users\windows\AppData\Local\Temp\TestStandaloneUpgradeUninstallKillWatcher2473922599\001
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip to C:\Users\windows\AppData\Local\Temp\TestStandaloneUpgradeUninstallKillWatcher2473922599\001
    fixture.go:829: Components were not modified from the fetched artifact
    fixture.go:624: >> running binary with: [C:\Users\windows\AppData\Local\Temp\TestStandaloneUpgradeUninstallKillWatcher2473922599\001\elastic-agent-8.13.0-SNAPSHOT-windows-x86_64\elastic-agent.exe version --binary-only --yaml]
    fetcher.go:91: Using existing artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip
    fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip to C:\Users\windows\AppData\Local\Temp\TestStandaloneUpgradeUninstallKillWatcher2473922599\002
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-windows-x86_64.zip to C:\Users\windows\AppData\Local\Temp\TestStandaloneUpgradeUninstallKillWatcher2473922599\002
    fixture.go:829: Components were not modified from the fetched artifact
[...]

PR #4163, opened 2024-01-31
Add support fetching artefacts for a build ID

rdner:
Previously it was not implemented but some of the tests requested a specific build by its ID. The fetcher was incorrectly returning just the latest snapshot in this case.

Also, the artifact caching system was not aware of build IDs and didn't distinguish snapshots of different builds. This led to caching the latest snapshot and returning it for any requested build ID.

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/4152

### DIFF
--- a/pkg/testing/fetcher_artifact.go
+++ b/pkg/testing/fetcher_artifact.go
@@ -64,28 +64,29 @@ func (f *artifactFetcher) Fetch(ctx context.Context, operatingSystem string, arc
 	}
 
 	var uri string
-	var prevErr error
-	if !f.snapshotOnly {
-		uri, prevErr = findURI(ctx, f.doer, version)
-	}
-	preVersion := version
-	version, _ = splitBuildID(version)
-	if uri == "" {
-		if !strings.HasSuffix(version, "-SNAPSHOT") {
-			version += "-SNAPSHOT"
-		}
-		uri, err = findURI(ctx, f.doer, version)
-		if err != nil {
-			return nil, fmt.Errorf("failed to find snapshot URI for version %s: %w (previous error: %w)", preVersion, err, prevErr)
-		}
+
+	noBuildVersion, _ := splitBuildID(version)
+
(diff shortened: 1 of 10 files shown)


pkg/testing/fetcher_artifact.go
```diff
@@ -63,29 +65,34 @@ func (f *artifactFetcher) Fetch(ctx context.Context, operatingSystem string, arc
...
-		}
-		uri, err = findURI(ctx, f.doer, version)
-		if err != nil {
-			return nil, fmt.Errorf("failed to find snapshot URI for version %s: %w (previous error: %w)", preVersion, err, prevErr)
+
+	if f.snapshotOnly && !ver.IsSnapshot() {
+		if ver.Prerelease() == "" {
+			ver = semver.NewParsedSemVer(ver.Major(), ver.Minor(), ver.Patch(), "SNAPSHOT", ver.BuildMetadata())
+		} else {
+			ver = semver.NewParsedSemVer(ver.Major(), ver.Minor(), ver.Patch(), ver.Prerelease()+"-SNAPSHOT", ver.BuildMetadata())
```
rdner, line 77: This is now consistent with https://github.com/elastic/elastic-agent/blob/d8b446f5a5686c99415e823edf321699fbea4b26/pkg/testing/fetcher_local.go#L57-L76

pkg/testing/fetcher_artifact.go
```diff
@@ -102,56 +109,117 @@ func (r *artifactResult) Name() string {
...
 
-	if len(body.Packages) == 0 {
+	return &body, nil
+}
+
+func findURI(ctx context.Context, doer httpDoer, version *semver.ParsedSemVer) (string, error) {
+	var (
+		project *projectResponse
+		err     error
+	)
```
AndersonQ, line 206: [Suggestion]
This form isn't really used in our code, I'd suggest keeping it consistent with the rest of the code by.

cmacknz:
This can go into 8.12 to fix the tests there as well.

Issue #4145, opened 2024-01-25
Properly support disabling providers and changing init time settings in containers


cmacknz:
Recently there have been several situations where users have needed to disable the leader election provider. Providers in agent cannot be configured in Fleet today, but even if they were they require a restart of the agent to take effect, which also isn't supported through Fleet. The providers are initialized when the composable controller is created at startup:

https://github.com/elastic/elastic-agent/blob/9d25f79b159744801daefca36ae227674e28d920/internal/pkg/composable/controller.go#L52-L71

It is possible to disable providers by editing the elastic-agent.yml file read by the agent container when it starts, which on Kubernetes is most easily accomplished by mounting the file as a ConfigMap. Essentially the process is:
1. Mount the elastic-agent.yml as a ConfigMap in the container.
2. Mount the ConfigMap as a volume
3. Ensure the agent is configured to read the configuration file mounted from that volume.
4. Mounting the ConfigMap as a volume creates a bind mount inside the container, which has the effect that the agent's attempt to overwrite the initial configuration on enrollment fails (because the bind mount is essentially a different file system and we can't move the file) and the flag disabling the providers is preserved. 
5. The enrollment API key and agent ID are stored in the state path which is usually stored on a host path volume, so subsequent restarts of the pod do not attempt to re-enroll and skip over trying to replace the contents of the ConfigMap. This work around does not work if you do not preserve the state path between restarts of the container.

A simplified example follows:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: agent-yml
  namespace: kube-system
  labels:
    app: elastic-agent
data:
  agent.yml: |-
    providers.kubernetes_leaderelection.enabled: false
    fleet.enabled: true
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: elastic-agent
  # ...
spec:
   # ...
    spec:
      containers:
        - name: elastic-agent
          image: docker.elastic.co/beats/elastic-agent:8.12.0
          args: ["-c", "/etc/elastic-agent/agent.yml", "-e"]
          volumeMounts:
            - name: agent-yml
              mountPath: /etc/elastic-agent/agent.yml
              readOnly: true
              subPath: agent.yml
      volumes:
        - name: agent-yml
          configMap:
            defaultMode: 416
            name: agent-yml
```

[...]

PR #4166, opened 2024-01-31
Add disable providers by default option

eyalkraft:
## What does this PR do?

Two main things (let me know if you prefer two separate PRs)

- Adding the `providers_default_disable` configuration flag, to allow changing the default providers behavior.
- Skipping replacing the current agent configuration with default fleet configuration upon enrollment, in case the current configuration already contains the configuration from the default fleet configuration.

## Why is it important?

- Being able to disable all providers by default is required in cases such as agentless.
- Skipping the configuration replacement is required to allow persisting pre-enrollment configuration for a containerized agent that enrolls to fleet.

## Checklist

- [ ] I have made corresponding changes to the documentation https://github.com/elastic/ingest-docs/pull/884

## Author's Checklist

Post-merge:
- [ ] https://github.com/elastic/agentless-controller/pull/241
- [ ] https://github.com/elastic/serverless-gitops/pull/1445

1. Deploy elastic-agent on k8s configured with a configmap
2. configure with fleet enabled and providers disabled by default
3. no provider logs!

## Related issues

- https://github.com/elastic/elastic-agent/issues/4145

## Screenshots

Tested on my serverless dev cluster.

Old agent image, and New image with old configmap: docker provider tries starting.

image

image

New image with new configmap: no provider is starting

image

image

New image with docker provider explicitly enabled: provider is starting

image

image

## Logs

```
elastic-agent@agentless-58995d99d6-hm8n2:~$ elastic-agent inspect | grep provider
providers_default_disable: true
```

## Questions to ask yourself

- ...


internal/pkg/composable/controller.go
```diff
@@ -65,7 +65,7 @@ func New(log *logger.Logger, c *config.Config, managed bool) (Controller, error)
 	contextProviders := map[string]*contextProviderState{}
 	for name, builder := range Providers.contextProviders {
 		pCfg, ok := providersCfg.Providers[name]
-		if ok && !pCfg.Enabled() {
+		if (ok && !pCfg.Enabled()) || (!ok && providersCfg.ProvidersDefaultDisable) {
```
olegsu, line 68: In case a provider is configured and the `ProvidersDefaultDisable` is set, should we continue or not?

elastic-agent.yml
```diff
@@ -261,6 +261,11 @@ agent.logging.to_stderr: true
 # and conditionals. Each provider's keys are automatically prefixed with the name
 # of the provider.
 
+# All registered providers are enabled by default.
+
+# Disable all providers by default and only enable explicitly configured providers.
+# providers_default_disable: true
```
blakerouse, line 267: I don't really like this as top-level field. The agent has a couple sections in its configuration `outputs`, `inputs`, `management`, `agent`. This as a top-level option is kinda floating out in the wrong area. Being that you are changing the default option of the Elastic Agent for all provides I would think this should be under `agent.*`.

What about: `agent.providers.initial_default: false`? The `initial_default` can be a `*bool` to allow the default to be true when unset, true (explicity set), or false.

blakerouse: That does work but I always worry about issues with defaults actually being initialized and not accidentally cleared to false before the unpack. I prefer to be safer with the `*bool`.

internal/pkg/composable/controller.go
olegsu: So in that test-case, what is the expected output?
```go
		{
			cfg: map[string]interface{}{
				"agent.providers.initial_default": "true",
				"providers": map[string]interface{}{
					"env": map[string]interface{}{
						"enabled": "false",
					},
                                 },
			},
		},

```

cmacknz:
`buildkite test it` is the command to re-trigger that build step.

Issue #4051, opened 2024-01-09
Missing `.installed` file prevents uninstall or upgrade


fearful-symmetry:
While trying to fix a broken install on a test machine, an attempt at running `elastic-agent install` failed, and a following `uninstall` command failed:
```
sudo /usr/bin/elastic-agent uninstall                                                                                                                                                                       1 ↵
[sudo] password for alexk: 
Error: can only be uninstalled by executing the installed Elastic Agent at: /usr/bin/elastic-agent
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.13/fleet-troubleshooting.html
```

There's two problems here:

1. ) we can't deal with partial/damaged installs. In this case, the error happened because the `.installed` file in the root directory is missing.
2.  The `can only be uninstalled by executing the installed Elastic Agent at: /usr/bin/elastic-agent` message is a bit deceptive. It doesn't actually care about what path you're running the binary from, but instead it's checking the validity of the install:
```
	if status == install.Installed && !info.RunningInstalled() {
		return fmt.Errorf("can only be uninstalled by executing the installed Elastic Agent at: %s", install.ExecutablePath(paths.Top()))
	}
```
We should print a message that more accurately describes why we bailed out of the uninstall process.

pierrehilbert:
It seems that it has been introduced by https://github.com/elastic/elastic-agent/pull/3909 then.
@blakerouse this is not backported to 8.12 so less urgent than https://github.com/elastic/elastic-agent/issues/3342 but we will need to fix this.

cmacknz:
Missing the `.installed` file also marks the agent as unupgradable which makes this a more severe problem. We have seen internal InfoSec agents hit this problem but we haven't been able to explain why.

https://github.com/elastic/elastic-agent/blob/bf513a5ee635c33fbca11b176d350d17a35b2863/internal/pkg/agent/application/upgrade/upgrade.go#L62-L68

https://github.com/elastic/elastic-agent/blob/bf513a5ee635c33fbca11b176d350d17a35b2863/internal/pkg/agent/application/paths/common.go#L301-L310

We should consider relaxing the `Upgradable` check not to depend on being installed unless we can find the root cause. I'd rather have people accidentally upgrade uninstalled agents than not be able to upgrade installed agents that should have been upgradable.

cmacknz:
For the problem Alex is describing we may be able to improve this by just creating the `.installed` marker as early as we can.

[...]

Issue #3890, opened 2023-12-08
The elastic-agent upgrade command can fail even though the upgrade succeeds


cmacknz:
First observed in https://buildkite.com/elastic/elastic-agent/builds/5460#018c4990-62ec-4dc1-a9c0-81953d923d80. One of the package version tests failed because the `elastic-agent ugprade` command exited with an EOF error.

```
=== RUN   TestUpgradeBrokenPackageVersion
    upgrade_broken_package_test.go:59: Testing Elastic Agent upgrade from 8.12.0-SNAPSHOT to 8.11.2-SNAPSHOT...
    fetcher.go:91: Using existing artifact elastic-agent-8.12.0-SNAPSHOT-linux-arm64.tar.gz
    fixture.go:261: Extracting artifact elastic-agent-8.12.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestUpgradeBrokenPackageVersion2415228739/001
    fixture.go:274: Completed extraction of artifact elastic-agent-8.12.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestUpgradeBrokenPackageVersion2415228739/001
    fixture.go:717: Components were not modified from the fetched artifact
    fetcher.go:91: Using existing artifact elastic-agent-8.11.2-SNAPSHOT-linux-arm64.tar.gz
    fixture.go:261: Extracting artifact elastic-agent-8.11.2-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestUpgradeBrokenPackageVersion2415228739/002
    fixture.go:274: Completed extraction of artifact elastic-agent-8.11.2-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestUpgradeBrokenPackageVersion2415228739/002
    fixture.go:717: Components were not modified from the fetched artifact
    fixture.go:520: >> running binary with: [/tmp/TestUpgradeBrokenPackageVersion2415228739/001/elastic-agent-8.12.0-SNAPSHOT-linux-arm64/elastic-agent version --binary-only --yaml]
    fixture.go:520: >> running binary with: [/tmp/TestUpgradeBrokenPackageVersion2415228739/002/elastic-agent-8.11.2-SNAPSHOT-linux-arm64/elastic-agent version --binary-only --yaml]
    upgrader.go:167: Installing version "8.12.0-SNAPSHOT"
    fixture_install.go:99: [test TestUpgradeBrokenPackageVersion] Inside fixture install function
[...]

PR #4172, opened 2024-01-31
Fix filemarker creation process, update error message on uninstall

fearful-symmetry:
## What does this PR do?

Closes https://github.com/elastic/elastic-agent/issues/4051

## Why is it important?

This does a few things:
- moves the command to create the `.installed` file to earlier on in the install process, so a partial install won't break the `uninstall` command.
- Rename `RunningInstalled()` to make it more clear what the function does.
- Change the error message that's printed when the uninstall command detects an issue with the install path

## Checklist

~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have made corresponding change to the default configuration files~


internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -63,7 +63,7 @@ type Upgrader struct {
 func IsUpgradeable() bool {
 	// only upgradeable if running from Agent installer and running under the
 	// control of the system supervisor (or built specifically with upgrading enabled)
-	return release.Upgradeable() || (paths.RunningInstalled() && info.RunningUnderSupervisor())
+	return release.Upgradeable() || (paths.InstallMarkerExists() && info.RunningUnderSupervisor())
```
cmacknz, line 66: Same as https://github.com/elastic/elastic-agent/pull/4172/files#r1473537703

I would be in favour of removing the `InstallMarkerExists()` check here.

internal/pkg/agent/application/info/agent_metadata.go
```diff
@@ -169,7 +169,7 @@ func (i *AgentInfo) ECSMetadata(l *logger.Logger) (*ECSMeta, error) {
 				BuildOriginal: release.Info().String(),
 				// only upgradeable if running from Agent installer and running under the
 				// control of the system supervisor (or built specifically with upgrading enabled)
-				Upgradeable: release.Upgradeable() || (paths.RunningInstalled() && RunningUnderSupervisor()),
+				Upgradeable: release.Upgradeable() || (paths.InstallMarkerExists() && RunningUnderSupervisor()),
```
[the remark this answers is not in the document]
fearful-symmetry: Yah, it seems like `RunningUnderSupervisor()` should be sufficient?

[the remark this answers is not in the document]
cmacknz: I think so, on Unix that check makes sure the parent PID is 1 which can't happen unless we are running as an installed service.

cmacknz:
Coming up with an automated test for this is a bit tricky. To reproduce the original problem, we'd need to SIGINT the install command at the correct moment which would be hard to do deterministically. 

@pchila added a good test of the "Copy files" portion of the installer in https://github.com/elastic/elastic-agent/pull/4173/files, see the changes in internal/pkg/agent/install/install_test.go.

Could we split the install up into individual function steps and write a unit test that ensures we can call uninstall if you only complete the first one?

You could also add a hidden test option to the install command to get the install to exit right after the files were copied or something. I like breaking the install code up into explicit steps a bit better since a test that says "ensure we can uninstall after the first step" is better than "ensure we can uninstall after we hit the part of the code I arbitrarily exited from".

fearful-symmetry:
@cmacknz when you say "ensure we can uninstall after the first step" are you thinking of a specific part of the process? After we create the `.installed` file? After the files are copied?

cmacknz:
Yes, should have been clearer, I am thinking of the install process as a sequence of steps like:

1. Create directory and install marker
2. Copy the agent package.
3. Setup the shell wrapper.
5. Fix permissions.
6. Setup the OS service.
7. Etc

internal/pkg/agent/install/install.go
```diff
@@ -101,14 +101,18 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
...
-			fmt.Sprintf("failed to create installation parent directory (%s)", filepath.Dir(topPath)),
-			errors.M("directory", filepath.Dir(topPath)))
+			fmt.Sprintf("failed to create installation parent directory (%s)", topPath),
+			errors.M("directory", topPath))
 	}
 
+	// create the install marker
+	if err := CreateInstallMarker(topPath, ownership); err != nil {
+		return utils.FileOwner{}, fmt.Errorf("failed to create install marker: %w", err)
+	}
```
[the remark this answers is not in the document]
fearful-symmetry: Yah, in the previous version, the `Copy` call would create the `Agent` directory. I tested it like this, but wasn't considering permissions.Lemme check...

internal/pkg/agent/application/info/agent_metadata.go
```diff
@@ -169,7 +168,7 @@ func (i *AgentInfo) ECSMetadata(l *logger.Logger) (*ECSMeta, error) {
 				BuildOriginal: release.Info().String(),
 				// only upgradeable if running from Agent installer and running under the
 				// control of the system supervisor (or built specifically with upgrading enabled)
-				Upgradeable: release.Upgradeable() || (paths.RunningInstalled() && RunningUnderSupervisor()),
+				Upgradeable: release.Upgradeable() || RunningUnderSupervisor(),
```
blakerouse, line 171: This is changing behavior no?

internal/pkg/agent/install/install.go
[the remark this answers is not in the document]
cmacknz: Good catch, we need to preserve the previous permissions. This is something else we should consider adding tests for here.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -63,7 +63,7 @@ type Upgrader struct {
 func IsUpgradeable() bool {
 	// only upgradeable if running from Agent installer and running under the
 	// control of the system supervisor (or built specifically with upgrading enabled)
-	return release.Upgradeable() || (paths.RunningInstalled() && info.RunningUnderSupervisor())
+	return release.Upgradeable() || info.RunningUnderSupervisor()
```
blakerouse, line 66: We don't allow upgradable unless installed from the `elastic-agent install`. This will break this check. This means that even RPM and DEB's can be self-upgraded, that is not true.

blakerouse (changes requested):

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz: Ah I wasn't sure which case this was covering and you are right it will allow RPM and DEB to appear upgradable through Fleet.

We need to put something back back to prevent support cases coming in about this when the upgrade process doesn't work for these installs.

I still don't really like the idea of the entire upgrade process being disabled by deleting a single file. I wonder if we shouldn't put the special case in the RPM and DEB instead?

Is there a way a running agent can detect that it was installed via a DEB or RPM package directly, instead of us inferring it via the presence of a magic file?

fearful-symmetry: Yah, removing that was @cmacknz 's idea, since the idea (i think) is that `RunningUnderSupervisor` will be functionally the same as `RunningInstalled`. Is there a case where `RunningUnderSupervisor` will be true but `RunningInstalled` won't be? 

Does the RPM/DEB install not place the `.installed` file?

cmacknz: > Does the RPM/DEB install not place the .installed file?

It doesn't since `elastic-agent install` is never invoked. The installation is done by the `dpkg` and `rpm` commands.

I ran `AGENT_PACKAGE_VERSION=8.12.0 EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/arm64 PACKAGES=deb mage package`, installed it in a Linux VM, enrolled it with an 8.13.0-SNAPSHOT Fleet instance, and confirmed Fleet shows it as upgradable.

cmacknz: Easiest thing to do now is to put the `Installed` check along with a comment about what its for. We can then evaluate if just creating the installed marker file earlier solves all problems related to upgrade.

Reworking how we detect we are installed on RPM and DEB separately from `elastic-agent install` is out of scope of this PR I think.

fearful-symmetry: @cmacknz to be clear, are you saying we should revert the placement of the `.installed` file, and just have this PR fix the error string and add some docs?

internal/pkg/agent/install/install.go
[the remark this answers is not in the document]
fearful-symmetry: So, I ended up moving the `CreateInstallMarker` to after the initial copy operation, which avoids the permission issue here, and also seems more useful? If we create the install marker and then the copy operation fails than we might not be able to run or uninstall the agent at all, as the executable or base config, etc, might be missing.

internal/pkg/agent/cmd/uninstall.go
```diff
@@ -54,8 +54,9 @@ func uninstallCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 	if status == install.NotInstalled {
 		return fmt.Errorf("not installed")
 	}
-	if status == install.Installed && !paths.RunningInstalled() {
-		return fmt.Errorf("can only be uninstalled by executing the installed Elastic Agent at: %s", install.ExecutablePath(paths.Top()))
+
+	if status == install.Installed && !paths.InstallMarkerExists() {
+		return fmt.Errorf("cannot find %s file relative to running elastic agent; agent should be run from %s, or the install may be corrupt", paths.MarkerFileName, install.ExecutablePath(paths.Top()))
```
AndersonQ, line 59: When you say _agent should be run from %s_, you mean in this case the uninstall command right? At first it seems the agent itself should be run from that does not seem to be the information you want to convey.

I'd suggest
```suggestion
		return fmt.Errorf("cannot find %s file relative to running elastic agent; agent uninstall should be run from %s, or the install may be corrupt", paths.MarkerFileName, install.ExecutablePath(paths.Top()))
```

internal/pkg/agent/application/info/agent_metadata.go
cmacknz, line 172: Let's add a comment that `paths.InstallMarkerExists()` exists to mark RPM and DEB packages as not upgradable from the UI.

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz, line 66: Same as https://github.com/elastic/elastic-agent/pull/4172/files#r1480469016. Let's add a comment that paths.InstallMarkerExists() exists to mark RPM and DEB packages as not upgradable from the UI.

Also if we could stop duplicating this logic that would be nice.

fearful-symmetry:
Alright, moved around `IsUpgradeable()` to remove some of the duplication. Hopefully CI and everything will be happy, we've had a bit of back-and-forth feature creep here.

fearful-symmetry:
HOW DID THE TEST COVERAGE GO DOWN

fearful-symmetry:
Alright, despite the fact that I added a bunch of tests, sonarqube is probably gonna be unhappy....

fearful-symmetry:
Current CI errors are (i think?) unrelated?

fearful-symmetry:
Not sure if the lone CI failure is related:
```
Error: error uninstalling agent: failed to remove installation directory (C:\\Program Files\\Elastic\\Agent): timed out while removing \"C:\\\\Program Files\\\\Elastic\\\\Agent\". Last error: remove C:\\Program Files\\Elastic\\Agent\\data\\elastic-agent-d8f513\\elastic-agent.exe: Access is denied
```
I assume it is?

internal/pkg/agent/install/install.go
```diff
@@ -123,110 +176,69 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
...
-				return utils.FileOwner{}, errors.New(
-					err,
-					fmt.Sprintf("failed to write shell wrapper (%s)", paths.ShellWrapperPath),
-					errors.M("destination", paths.ShellWrapperPath))
-			}
-		}
-	}
-
 	// create the install marker
 	if err := CreateInstallMarker(topPath, ownership); err != nil {
```
pchila, line 193: I was expecting this call to be moved before the copy of other files, aren't we going to have the same issue if the copy of files above fails or gets interrupted for whatever reason ?

internal/pkg/agent/install/install.go
```diff
@@ -32,81 +32,134 @@ const (
...
+		if err != nil {
+			return "", errors.New(err, fmt.Sprintf("failed to uninstall Agent at (%s)", filepath.Dir(topPath)), errors.M("directory", filepath.Dir(topPath)))
+		}
+		pt.Describe("Successfully uninstalled current Elastic Agent")
 	}
+	return dir, nil
+}
 
+// perform the file copy operations
+func copyFiles(pt *progressbar.ProgressBar, topPath string, streams *cli.IOStreams, sourceDir string, ownership utils.FileOwner) error {
```
pchila, line 158: This function does too much: I understand that this is the result of a refactor that extracted a big chunk of the install code into its own function but we don't need a ProgressBar to perform a copy of the files.

As an example have a look at https://github.com/elastic/elastic-agent/issues/3890 (and even then the cli.IOStreams can be pulled out of the function)

fearful-symmetry:
Yeah @pchila the entire refactor is entirely in service of adding 1-2 unit tests and making SonarQube happy. Between the added complexity of the PR and the coming conflict of https://github.com/elastic/elastic-agent/pull/4193 I'm not sure it makes sense, and I wonder if we should revert this PR to a simpler version without the refactor.

internal/pkg/agent/install/install.go
fearful-symmetry: So, the install marker is just there so `elastic-agent` doesn't complain during an uninstall operation. However, if that initial copy doesn't complete, we might not even have an `elastic-agent` binary or any files, and anything that is copied can either be removed via `rm` or just overwritten on  the next install. This is a less troublesome failure mode than creating the install marker at the end, where we'd have to remove service files, systemd config, etc, by hand.

fearful-symmetry:
So, talked with @pierrehilbert , we're going to merge https://github.com/elastic/elastic-agent/pull/4193, then I'll rebase this without all the refactoring, and either rely on the tests in https://github.com/elastic/elastic-agent/pull/4193 or expand on them for the sake of testing this.

internal/pkg/agent/install/install_test.go
```diff
@@ -185,3 +187,10 @@ func TestCopyFiles(t *testing.T) {
 	}
 
 }
+
+func TestInstallMarker(t *testing.T) {
```
pchila, line 191: Maybe add some assertion about ownership and permissions?
I know it's not vital as we install/uninstall as root/administrator but it's part of the install marker creation and it would be nice to see this part of `CreateInstallMarker` behavior documented in the test as well

pchila (approved):
Small nit on some additional assertions, looks good otherwise

internal/pkg/agent/install/install.go
```diff
@@ -125,6 +125,11 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 		return utils.FileOwner{}, err
 	}
 
+	// create the install marker
```
michalpristas, line 128: can we just make a small comment to make sure we don't move it after

internal/pkg/agent/install/install.go
```diff
@@ -475,8 +475,10 @@ func hasAllSSDs(block ghw.BlockInfo) bool {
 
 func CreateInstallMarker(topPath string, ownership utils.FileOwner) error {
 	markerFilePath := filepath.Join(topPath, paths.MarkerFileName)
-	if _, err := os.Create(markerFilePath); err != nil {
+	handle, err := os.Create(markerFilePath)
```
michalpristas, line 478: 🤦 nice catch. we should have a linter for this

michalpristas (approved):

internal/pkg/agent/install/install.go
fearful-symmetry: So, the windows CI tests will fail on this if you leave the file open in a temp directory created by the golang `testing` library.

internal/pkg/agent/install/install.go
```diff
@@ -125,6 +125,14 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 		return utils.FileOwner{}, err
 	}
 
+	// create the install marker
+	// do this after the file copy but before installing service files, etc
+	// That way if the service install process fails, a user can still run `elastic-agent uninstall`
+	// to clean up.
+	if err := CreateInstallMarker(topPath, ownership); err != nil {
```
cmacknz, line 132: This is still happening after the `copyFiles` call, it should be immediately after the `err = os.MkdirAll(filepath.Dir(topPath), 0755)` earlier in this function.

fearful-symmetry: So, I chose to not put it there, since putting the `.installed` file before anything has been installed seemed counter-intuitive, and if the copy fails we might not be able to run `elastic-agent` at all. Unless there's something I'm not factoring in?

cmacknz: You can't run `elastic-agent uninstall` unless the `.installed` marker exists, so if you put it after the copy you can't remove the copied files from disk with the `elastic-agent uninstall` command.

internal/pkg/agent/install/install.go
```diff
@@ -216,6 +207,27 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
...
+// setup the basic topPath, and the .installed file
+func setupInstallPath(topPath string, ownership utils.FileOwner) error {
+	// ensure parent directory exists
+	err := os.MkdirAll(filepath.Dir(topPath), 0755)
+	if err != nil {
+		return errors.New(err, fmt.Sprintf("failed to create installation parent directory (%s)", filepath.Dir(topPath)), errors.M("directory", filepath.Dir(topPath)))
+	}
+
+	// create Agent/ directory with more locked-down permissions
+	err = os.MkdirAll(topPath, 0750)
```
cmacknz, line 219: Why do we need to do this now? It might have been discussed already but I missed it.

cmacknz:
Needs a changelog entry.

internal/pkg/agent/install/install.go
fearful-symmetry: So, this is normally performed by the later `Copy` operation that does the actual file copying. However, we're now creating the directory before we call `Copy`, so we need to create the directory manually.

changelog/fragments/1707951532-change-install-marker-creation.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: change-install-marker-creation
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: Create the .installed marker earlier on in the install process, allowing the user of `elastic-agent uninstall` in cases where the install fails.
```
cmacknz, line 19:
```suggestion
description: Create the .installed marker earlier on in the install process, allowing the use of `elastic-agent uninstall` to cleanup if the install fails.
```

cmacknz (approved):

pchila (approved):

cmacknz:
I am going to force merge this since Blake has requested changes but is on PTO. The change is now small enough that I don't expect he'd object.

Issue #4138, opened 2024-01-25
Remove deprecated  "golang.org/x/tools/go/vcs":  replace it with go list command with -json flag instead


pchila:
In PR #3910 linter complains about the usage of vcs package in dev-tools

```
SA1019: "golang.org/x/tools/go/vcs" is deprecated: Use the go list command with -json flag instead, which implements up-to-date import path resolution behavior, module support, and includes the latest security fixes. (staticcheck)
```

Since this does not look like a quick fix, I am opening this issue to track the removal of deprecated vcs package

PR #4173, opened 2024-02-01
Use package manifest for install

pchila:
## What does this PR do?

This is a subset of a bigger PR #3950, so CI is not expected to be green, especially regarding the integration tests

This PR uses the package manifest included in elastic-agent packages to map the package contents onto disk when installing. In the context of #2579 the path mapping is used to install agent in a directory like `data/elastic-agent-<package version>-<hash>`.

The install process now looks for the package manifest file and start copying mapped paths first, then copy all the unmapped files and finally creates the symlinks after remapping their targets.

## Why is it important?

This PR is needed to ensure that new versions of agents get installed to a new path containing the version from a package `data/elastic-agent-<package version>-<hash>` that still uses the legacy structure `data/elastic-agent-<hash>` for backward compatibility

## Checklist

## Author's Checklist

- [ ]

To test this PR locally, package an elastic-agent (with or without specifying a custom AGENT_PACKAGE_VERSION) like:
```
```
or (this is an example, AGENT_PACKAGE_VERSION can be something else as long as it respects semver convention)
```
AGENT_PACKAGE_VERSION=8.13.0-version-in-path EXTERNAL=true SNAPSHOT=true PLATFORMS="linux/amd64" PACKAGES="tar.gz"  mage package
```

Install the agent and verify that it's healthy and it runs from `data/elastic-agent-<package version>-<hash>` directory.
Upgrade support will come in a follow-up PR and at this point this has been tested only on Linux. MacOS and Windows fixes will come in a follow-up PR (we need small modifications to make this work mostly related to OS-dependendent agent paths).

## Related issues

- Relates #2579 
- Relates #3950 

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/packaging/packages.yml
+++ b/dev-tools/packaging/packages.yml
@@ -62,19 +62,19 @@ shared:
       /etc/init.d/{{.BeatServiceName}}:
         template: '{{ elastic_beats_dir }}/dev-tools/packaging/templates/{{.PackageType}}/elastic-agent.init.sh.tmpl'
         mode: 0755
-      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{ commit_short }}/{{.BeatName}}{{.BinaryExt}}:
+      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{agent_package_version}}{{snapshot_suffix}}-{{ commit_short }}/{{.BeatName}}{{.BinaryExt}}:
         source: build/golang-crossbuild/{{.BeatName}}-{{.GOOS}}-{{.Platform.Arch}}{{.BinaryExt}}
         mode: 0755
-      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{ commit_short }}/package.version:
+      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{agent_package_version}}{{snapshot_suffix}}-{{ commit_short }}/package.version:
         content: >
           {{ agent_package_version }}
         mode: 0644
-      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{ commit_short }}/components:
+      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{agent_package_version}}{{snapshot_suffix}}-{{ commit_short }}/components:
         source: '{{.AgentDropPath}}/{{.GOOS}}-{{.AgentArchName}}.tar.gz/'
         mode: 0755
         config_mode: 0644
         skip_on_missing: true
-      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{ commit_short }}/manifest.yaml:
+      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{agent_package_version}}{{snapshot_suffix}}-{{ commit_short }}/manifest.yaml:
         mode: 0644
         content: >
           {{ manifest }}
--- a/dev-tools/packaging/templates/linux/postinstall.sh.tmpl
+++ b/dev-tools/packaging/templates/linux/postinstall.sh.tmpl
@@ -16,8 +16,9 @@ if test -L "$symlink"; then
 fi
 
 commit_hash="{{ commit_short }}"
+version_dir="{{agent_package_version}}{{snapshot_suffix}}"
 
-new_agent_dir="/var/lib/elastic-agent/data/elastic-agent-$commit_hash"
+new_agent_dir="/var/lib/elastic-agent/data/elastic-agent-$version_dir-$commit_hash"
 
 # copy the state files if there was a previous agent install
 if ! [ -z "$old_agent_dir" ] && ! [ "$old_agent_dir" -ef "$new_agent_dir" ]; then
--- a/internal/pkg/agent/install/install.go
+++ b/internal/pkg/agent/install/install.go
@@ -110,21 +110,11 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 			errors.M("directory", filepath.Dir(topPath)))
 	}
 
-	manifestFile, err := os.Open("manifest.yaml")
+	manifest, err := readPackageManifest(dir)
 	if err != nil {
-		return utils.FileOwner{}, errors.New(
-			err,
-			fmt.Sprintf("failed to open package manifest file (%s)", "manifest.yaml"),
-			errors.M("file", "manifest.yaml"))
+		return utils.FileOwner{}, fmt.Errorf("reading package manifest: %w", err)
 	}
-	defer manifestFile.Close()
-	manifest, err := v1.ParseManifest(manifestFile)
-	if err != nil {
-		return utils.FileOwner{}, errors.New(
-			err,
-			fmt.Sprintf("failed to parse package manifest file contents (%s)", "manifest.yaml"),
-			errors.M("file", "manifest.yaml"))
-	}
+
 	pathMappings := manifest.Package.PathMappings
 
 	pt.Describe("Copying install files")
@@ -223,6 +213,21 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 	pt.Describe("Installed service")
 
 	return ownership, nil
+}
+
+func readPackageManifest(extractedPackageDir string) (*v1.PackageManifest, error) {
+	manifestFilePath := filepath.Join(extractedPackageDir, "manifest.yaml")
+	manifestFile, err := os.Open(manifestFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open package manifest file (%s): %w", manifestFilePath, err)
+	}
+	defer manifestFile.Close()
+	manifest, err := v1.ParseManifest(manifestFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse package manifest file %q contents: %w", manifestFilePath, err)
+	}
+
+	return manifest, nil
 }
 
 func copyFiles(streams *cli.IOStreams, pathMappings []map[string]string, dir string, topPath string) error {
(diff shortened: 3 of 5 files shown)


dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -16,8 +16,9 @@ if test -L "$symlink"; then
 fi
 
 commit_hash="{{ commit_short }}"
+version_dir="{{agent_package_version}}{{snapshot_suffix}}"
```
michalpristas, line 19: nit: name suggests it's the name of the whole dir. what's the reason of this separation.can we either include it in new_agent_dir or exclude whole dir into this one?

internal/pkg/agent/install/install.go
```diff
@@ -109,35 +110,30 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
...
 
-	// copy source into install path
-	//
-	// Try to detect if we are running with SSDs. If we are increase the copy concurrency,
-	// otherwise fall back to the default.
-	copyConcurrency := 1
-	hasSSDs, detectHWErr := HasAllSSDs()
-	if detectHWErr != nil {
-		fmt.Fprintf(streams.Out, "Could not determine block hardware type, disabling copy concurrency: %s\n", detectHWErr)
+	manifestFile, err := os.Open("manifest.yaml")
```
michalpristas, line 113: can we extract work with manifest into separate function and release handler a bit sooner?

ycombinator:
To silence the `vcs` linter error, would it be possible to add a nolint directive with a link to https://github.com/elastic/elastic-agent/issues/4138?

internal/pkg/agent/install/install.go
```diff
@@ -229,6 +225,103 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 	return ownership, nil
 }
 
+func copyFiles(streams *cli.IOStreams, pathMappings []map[string]string, dir string, topPath string) error {
```
ycombinator, line 228: I would like to see more comments sprinkled throughout this function, focussing more on why we're doing the various copying and link fixing steps here.

ycombinator, line 228: I think it will be a bit easier to follow the logic if we use `srcDir` instead of the generic `dir`.

```suggestion
func copyFiles(streams *cli.IOStreams, pathMappings []map[string]string, srcDir string, topPath string) error {
```

dev-tools/packaging/packages.yml
```diff
@@ -62,19 +62,19 @@ shared:
...
         content: >
           {{ agent_package_version }}
         mode: 0644
-      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{ commit_short }}/components:
+      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{agent_package_version}}{{snapshot_suffix}}-{{ commit_short }}/components:
         source: '{{.AgentDropPath}}/{{.GOOS}}-{{.AgentArchName}}.tar.gz/'
         mode: 0755
         config_mode: 0644
         skip_on_missing: true
-      /var/lib/{{.BeatName}}/data/{{.BeatName}}-{{ commit_short }}/manifest.yaml:
```
ycombinator, line 77: Why do we make these changes for the deb/rpm spec but not others, e.g. MacOS? Or, to put it differently, if we made the corresponding changes in other specs would we still need the install code changes in this PR?

I'm sure there's a valid explanation for this difference; I'm just not understanding it yet.

cmacknz (commented):
There is an important test missing here and from the test instructions:

> Install the agent and verify that it's healthy and it runs from data/elastic-agent-<package version>-<hash> directory.

This verifies that the new installation code to copy the agent with path mappings works.

It does not verify that we continue to produce elastic-agent-* packages with the old `elastic-agent-$hash` format, which is critical to ensure that the upgrade process continues to work.

See https://github.com/elastic/elastic-agent/issues/2579#issuecomment-1813693225

> Continue to package the agent using the elastic-agent-$hash pattern with $hash being the first 6 characters read from the .elastic-agent.active.commit file. This guarantees all existing agent versions can upgrade with no changes that interact with previously released versions.

We need a test to guarantee that the original format of the agent package *before* installation is correct, in addition to checking that the one *after* installation is remapped.

It would be nice to do this in one of the installation integration tests with assertions on the path structure, ensuring we are testing the direct output of `mage package` and the result of running `elastic-agent install` directly.

Also, maybe a controversial question, but do we need to remap the paths at install time? It is definitely more consistent, it simulates the world we we stop having to deal with agents using the "old" structure, but we have to deal the old structure during upgrades until there are no more 7.17 or pre-8.13 agents anywhere in any deployment. I don't expect that to be anytime soon.

dev-tools/packaging/packages.yml
ycombinator: Never mind, ignore this question.  It looks like the MacOS section is not even being used in packaging.  I tried deleting that section, repackaged Agent, installed it, and it worked fine.  I'll make a separate PR to just remove that section so I don't get confused by it again in the future.

ycombinator: https://github.com/elastic/elastic-agent/pull/4182

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
pchila: We can certainly nest this in the new_agent_dir variable directly, I thought that for debugging the script this would be easier than a big string concatenation done all at once (similar to what is done for the commit_hash

dev-tools/packaging/packages.yml
pchila: Didn't check yet if we are actually using that MacOS package specification, will do that later.
As for why we are only changing this on .deb and .rpm: we install those files without calling the elastic-agent install command but we rely on CLI utilities to so that.

We want to have the version in the installation path when installing the agent so we need to change the directory structure at packaging time.

We use tar.gz and .zip packages for both install and upgrade: as discussed in a recent RFC we want to maintain the package directory structure with `elastic-agent-<hash>` for backward compatibility with previous versions of agent (hence this change is not applied to those package types) and we handle the remapping to `elastic-agent-<version>-<hash>` within the `install` and `upgrade` commands

internal/pkg/agent/install/install.go
```diff
@@ -229,6 +225,103 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
...
+	}
+
+	err := copy.Copy(dir, topPath, copy.Options{
+		OnSymlink: func(source string) copy.SymlinkAction {
+
+			target, err := os.Readlink(source)
+			if err != nil {
+				// error reading the link, not much choice to leave it unchanged and hope for the best
+				return copy.Shallow
+			}
```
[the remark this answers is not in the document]
pchila: addressed in f1d85ee0550007bff3837d05fa7bdd44d91da992

internal/pkg/agent/install/install.go
pchila: done in acacac7efb7732c819ab80fe334706b841775cab

internal/pkg/agent/install/install.go
pchila: done in 3781d6f702bc010d806c1e44e0a865c234206f20

pchila: done in 3781d6f702bc010d806c1e44e0a865c234206f20

pchila:
@cmacknz

> This verifies that the new installation code to copy the agent with path mappings works.
> 
> It does not verify that we continue to produce elastic-agent-* packages with the old elastic-agent-$hash format, which is critical to ensure that the upgrade process continues to work.

 the check for the package to have the directory `elastic-agent-hash` is checked during the package tests we have already on main (we check that the versionedHome declared in manifest exists). I am not sure why you would want an integration test for that.
For backward compatibility we already have upgrade tests from previous minor and previous major: those will go red if we break the backward compatibility

> Also, maybe a controversial question, but do we need to remap the paths at install time? It is definitely more consistent, it simulates the world we we stop having to deal with agents using the "old" structure, but we have to deal the old structure during upgrades until there are no more 7.17 or pre-8.13 agents anywhere in any deployment. I don't expect that to be anytime soon.

As for the controversial we don't *need* to remap at installation time but I would like for the old `elastic-agent-hash` directory to disappear gradually as people update to newer versions of agent:
- It makes it easier on us because we would have less and less cases where the legacy format appears and the new format becomes the norm (I will not have to ask somebody if the installed or upgraded the agent to know where the path is, we will only have the case for agent upgrades from <8.13)
- It is the less surprising option: 
  - if we don't remap and somebody upgrades to some agent >8.13 and gets the new path, reinstall the agent from scratch they got `elastic-agent-hash` instead
  - if somebody is coming from an earlier version of agent and upgrades it just sees that new versions of agent contains the version in the directory name, it's a change but a less surprising one

ycombinator (approved):
LGTM.  Thanks for splitting up the large PR into smaller ones.  I know it's a PITA to manage but it definitely helps as a reviewer to focus on one specific aspect of the larger change at a time, so much appreciated 🙇.

cmacknz:
> the check for the package to have the directory elastic-agent-hash is checked during the package tests we have already on main (we check that the versionedHome declared in manifest exists). I am not sure why you would want an integration test for that.

I may have missed this, will double check.

> As for the controversial we don't need to remap at installation time but I would like for the old elastic-agent-hash directory to disappear gradually as people update to newer versions of agent:

Thanks your reasoning makes sense to me. I do not have a strong opinion about getting rid of it, but I do feel like we don't need to keep code just because we've already written it, so wanted to double check.

cmacknz:
> the check for the package to have the directory elastic-agent-hash is checked during the package tests we have already on main (we check that the versionedHome declared in manifest exists). I am not sure why you would want an integration test for that.

What I'm worried about is someone changing the packaging including the package tests, but not realizing the impact it has on our backwards compatibility. The packaging tests exists in isolation without the ability to see the system impact. I am usually biased towards using integration tests for this reason, you can encode the actual product use case as a user would experience into the test.

> For backward compatibility we already have upgrade tests from previous minor and previous major: those will go red if we break the backward compatibility

This covers the case I care about. LGTM.

cmacknz (approved):

PR #4174, opened 2024-02-01
Use package manifests for upgrade

pchila:
## What does this PR do?
This is a subset of a bigger PR #3950, so CI is not expected to be green, especially regarding the integration tests.

This PR uses the package manifest included in elastic-agent packages to map the package contents onto disk when upgrading . In the context of #2579 the path mapping is used to install the agent we are upgrading to in a directory like `data/elastic-agent-<package version>-<hash>`.

The upgrade process now looks for the package manifest file and copies the files in the archive after mapping the name using the path mappings in the manifest. This means that also the new agent will end up in `data/elastic-agent-<package version>-<hash>`

## Why is it important?

This PR is needed to ensure that new versions of agents get installed to a new path containing the version from a package `data/elastic-agent-<package version>-<hash>` that still uses the legacy structure `data/elastic-agent-<hash>` for backward compatibility. In practice it allows for upgrading to an agent version built from the same commit as the existing agent as long as the package version is different (the goal of #2579)

## Checklist

## Author's Checklist

- [ ]

To test this PR locally, we package elastic-agent twice from the same commit specifying a different AGENT_PACKAGE_VERSION like:
```
```
and
```
AGENT_PACKAGE_VERSION=8.13.0-repackaged EXTERNAL=true SNAPSHOT=true PLATFORMS="linux/amd64" PACKAGES="tar.gz"  mage package
```

Install the first agent package and verify that it's healthy and it runs from `data/elastic-agent-8.13.0-SNAPSHOT-<hash>` directory.
Upgrade this agent using the second package
```
elastic-agent upgrade 8.13.0-repackaged-SNAPSHOT --source-uri file://<directory where the package is located on disk> --skip-verify
```

Upgrade will unpack and store the new agent in `data/elastic-agent-8.13.0-repackaged-<hash>` and rotate the links.
Verify the agent directory structure contains the 2 distinct directories (you have to do it before the grace period ends).

*Known issues*:
- When the grace period expires the watcher will break the agent install since it will clean anything that does not look like `elastic-agent-<hash>`, this is addressed already in a follow-up PR
- An extra `elastic-agent-<hash>` directory is created during upgrade due to using the wrong path during copy of component states. This is addressed in a follow-up PR
- This has been tested only on Linux. MacOS and Windows fixes will come in a follow-up PR (we need small modifications to make this work mostly related to OS-dependent agent paths).
- To clean up the broken agent using uninstall (it can be done manually as well but I find this more convenient):
  1. remove the broken symlink in `/opt/Elastic/Agent`
  2. use the extracted package to force  install the agent again `./elastic-agent install --force`
  3. remove using uninstall `sudo elastic-agent uninstall`

## Related issues

- Relates #2579 
- Relates #3950 
- Requires #4173

- 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -33,9 +33,20 @@ const (
 )
 
 // Rollback rollbacks to previous version which was functioning before upgrade.
-func Rollback(ctx context.Context, log *logger.Logger, prevHash string, currentHash string) error {
+func Rollback(ctx context.Context, log *logger.Logger, prevVersionedHome, prevHash string, currentHash string) error {
+	symlinkPath := filepath.Join(paths.Top(), agentName)
+
+	var symlinkTarget string
+	if prevVersionedHome != "" {
+		symlinkTarget = paths.BinaryPath(filepath.Join(paths.Top(), prevVersionedHome), agentName)
+	} else {
+		// fallback for upgrades that didn't use the manifest and path remapping
+		hashedDir := fmt.Sprintf("%s-%s", agentName, prevHash)
+		// paths.BinaryPath properly derives the binary directory depending on the platform. The path to the binary for macOS is inside of the app bundle.
+		symlinkTarget = paths.BinaryPath(filepath.Join(paths.Top(), "data", hashedDir), agentName)
+	}
 	// change symlink
-	if err := ChangeSymlink(ctx, log, prevHash); err != nil {
+	if err := changeSymlinkInternal(log, symlinkPath, symlinkTarget); err != nil {
 		return err
 	}
 
@@ -51,11 +62,11 @@ func Rollback(ctx context.Context, log *logger.Logger, prevHash string, currentH
 	}
 
 	// cleanup everything except version we're rolling back into
-	return Cleanup(log, prevHash, true, true)
+	return Cleanup(log, prevVersionedHome, prevHash, true, true)
 }
 
 // Cleanup removes all artifacts and files related to a specified version.
-func Cleanup(log *logger.Logger, currentHash string, removeMarker bool, keepLogs bool) error {
+func Cleanup(log *logger.Logger, currentVersionedHome, currentHash string, removeMarker bool, keepLogs bool) error {
 	log.Infow("Cleaning up upgrade", "hash", currentHash, "remove_marker", removeMarker)
 	<-time.After(afterRestartDelay)
 
@@ -83,7 +94,16 @@ func Cleanup(log *logger.Logger, currentHash string, removeMarker bool, keepLogs
 	_ = os.Remove(prevSymlink)
 
 	dirPrefix := fmt.Sprintf("%s-", agentName)
-	currentDir := fmt.Sprintf("%s-%s", agentName, currentHash)
+	var currentDir string
+	if currentVersionedHome != "" {
+		currentDir, err = filepath.Rel("data", currentVersionedHome)
+		if err != nil {
+			return fmt.Errorf("extracting elastic-agent path relative to data directory from %s: %w", currentVersionedHome, err)
+		}
+	} else {
+		currentDir = fmt.Sprintf("%s-%s", agentName, currentHash)
+	}
+
 	for _, dir := range subdirs {
 		if dir == currentDir {
 			continue
--- a/internal/pkg/agent/application/upgrade/step_mark.go
+++ b/internal/pkg/agent/application/upgrade/step_mark.go
@@ -39,7 +39,7 @@ type UpdateMarker struct {
 	PrevVersion string `json:"prev_version" yaml:"prev_version"`
 	// PrevHash is a hash agent is updated from
 	PrevHash string `json:"prev_hash" yaml:"prev_hash"`
-	// PrevVersionedHome represents the path where the new agent is located relative to top path
+	// PrevVersionedHome represents the path where the old agent is located relative to top path
 	PrevVersionedHome string `json:"prev_versioned_home" yaml:"prev_versioned_home"`
 
 	// Acked is a flag marking whether or not action was acked
--- a/internal/pkg/agent/application/upgrade/step_mark.go
+++ b/internal/pkg/agent/application/upgrade/step_mark.go
@@ -6,6 +6,7 @@ package upgrade
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
@@ -24,8 +25,13 @@ const markerFilename = ".update-marker"
 
 // UpdateMarker is a marker holding necessary information about ongoing upgrade.
 type UpdateMarker struct {
+	// Version represents the version the agent is upgraded to
+	Version string `json:"version" yaml:"version"`
 	// Hash agent is updated to
 	Hash string `json:"hash" yaml:"hash"`
+	// VersionedHome represents the path where the new agent is located relative to top path
+	VersionedHome string `json:"versioned_home" yaml:"versioned_home"`
+
 	//UpdatedOn marks a date when update happened
 	UpdatedOn time.Time `json:"updated_on" yaml:"updated_on"`
 
@@ -33,6 +39,8 @@ type UpdateMarker struct {
 	PrevVersion string `json:"prev_version" yaml:"prev_version"`
 	// PrevHash is a hash agent is updated from
 	PrevHash string `json:"prev_hash" yaml:"prev_hash"`
+	// PrevVersionedHome represents the path where the new agent is located relative to top path
+	PrevVersionedHome string `json:"prev_versioned_home" yaml:"prev_versioned_home"`
 
 	// Acked is a flag marking whether or not action was acked
 	Acked  bool                    `json:"acked" yaml:"acked"`
@@ -83,42 +91,55 @@ func convertToActionUpgrade(a *MarkerActionUpgrade) *fleetapi.ActionUpgrade {
 }
 
 type updateMarkerSerializer struct {
-	Hash        string               `yaml:"hash"`
-	UpdatedOn   time.Time            `yaml:"updated_on"`
-	PrevVersion string               `yaml:"prev_version"`
-	PrevHash    string               `yaml:"prev_hash"`
-	Acked       bool                 `yaml:"acked"`
-	Action      *MarkerActionUpgrade `yaml:"action"`
-	Details     *details.Details     `yaml:"details"`
+	Version           string               `yaml:"version"`
+	Hash              string               `yaml:"hash"`
+	VersionedHome     string               `yaml:"versioned_home"`
+	UpdatedOn         time.Time            `yaml:"updated_on"`
+	PrevVersion       string               `yaml:"prev_version"`
+	PrevHash          string               `yaml:"prev_hash"`
+	PrevVersionedHome string               `yaml:"prev_versioned_home"`
+	Acked             bool                 `yaml:"acked"`
+	Action            *MarkerActionUpgrade `yaml:"action"`
+	Details           *details.Details     `yaml:"details"`
 }
 
 func newMarkerSerializer(m *UpdateMarker) *updateMarkerSerializer {
 	return &updateMarkerSerializer{
-		Hash:        m.Hash,
-		UpdatedOn:   m.UpdatedOn,
-		PrevVersion: m.PrevVersion,
-		PrevHash:    m.PrevHash,
-		Acked:       m.Acked,
-		Action:      convertToMarkerAction(m.Action),
-		Details:     m.Details,
+		Version:           m.Version,
+		Hash:              m.Hash,
+		VersionedHome:     m.VersionedHome,
+		UpdatedOn:         m.UpdatedOn,
+		PrevVersion:       m.PrevVersion,
+		PrevHash:          m.PrevHash,
+		PrevVersionedHome: m.PrevVersionedHome,
+		Acked:             m.Acked,
+		Action:            convertToMarkerAction(m.Action),
+		Details:           m.Details,
 	}
 }
 
 // markUpgrade marks update happened so we can handle grace period
-func (u *Upgrader) markUpgrade(_ context.Context, log *logger.Logger, hash string, action *fleetapi.ActionUpgrade, upgradeDetails *details.Details) error {
-	prevVersion := release.Version()
+func (u *Upgrader) markUpgrade(_ context.Context, log *logger.Logger, version, hash, versionedHome string, action *fleetapi.ActionUpgrade, upgradeDetails *details.Details) error {
+	prevVersion := release.VersionWithSnapshot()
 	prevHash := release.Commit()
+	prevVersionedHome, err := filepath.Rel(paths.Top(), paths.Home())
+	if err != nil {
+		return fmt.Errorf("calculating home path relative to top, home: %q top: %q : %w", paths.Home(), paths.Top(), err)
+	}
 	if len(prevHash) > hashLen {
 		prevHash = prevHash[:hashLen]
 	}
 
 	marker := &UpdateMarker{
-		Hash:        hash,
-		UpdatedOn:   time.Now(),
-		PrevVersion: prevVersion,
-		PrevHash:    prevHash,
-		Action:      action,
-		Details:     upgradeDetails,
+		Version:           version,
+		Hash:              hash,
+		VersionedHome:     versionedHome,
+		UpdatedOn:         time.Now(),
+		PrevVersion:       prevVersion,
+		PrevHash:          prevHash,
+		PrevVersionedHome: prevVersionedHome,
+		Action:            action,
+		Details:           upgradeDetails,
 	}
 
 	markerBytes, err := yaml.Marshal(newMarkerSerializer(marker))
@@ -183,13 +204,16 @@ func loadMarker(markerFile string) (*UpdateMarker, error) {
 	}
 
 	return &UpdateMarker{
-		Hash:        marker.Hash,
-		UpdatedOn:   marker.UpdatedOn,
-		PrevVersion: marker.PrevVersion,
-		PrevHash:    marker.PrevHash,
-		Acked:       marker.Acked,
-		Action:      convertToActionUpgrade(marker.Action),
-		Details:     marker.Details,
+		Version:           marker.Version,
+		Hash:              marker.Hash,
+		VersionedHome:     marker.VersionedHome,
+		UpdatedOn:         marker.UpdatedOn,
+		PrevVersion:       marker.PrevVersion,
+		PrevHash:          marker.PrevHash,
+		PrevVersionedHome: marker.PrevVersionedHome,
+		Acked:             marker.Acked,
+		Action:            convertToActionUpgrade(marker.Action),
+		Details:           marker.Details,
 	}, nil
 }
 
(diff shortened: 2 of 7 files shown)


internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -7,52 +7,105 @@ package upgrade
...
+
+	// Load manifest, the use of path.Join is intentional since in .zip file paths use slash ('/') as separator
+	manifestFile, err := r.Open(path.Join(fileNamePrefix, "manifest.yaml"))
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		// we got a real error looking up for the manifest
+		return UnpackResult{}, fmt.Errorf("looking up manifest in package: %w", err)
+	}
+	if err == nil {
+		// load manifest
+		defer manifestFile.Close()
```
michalpristas, line 79: same here would be nice to release manifest handler sooner

michalpristas: consider adding helper which parses it and returns structure as this is recurring or pathmapper constructor with filepath as arg

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -169,46 +170,65 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
-	if strings.HasPrefix(release.Commit(), newHash) {
-		u.log.Warn("Upgrade action skipped: upgrade did not occur because its the same version")
-		return nil, nil
+	if unpackRes.VersionedHome == "" {
+		// FIXME this should be treated as an error
+		return nil, fmt.Errorf("versionedhome is empty: %v", unpackRes)
 	}
 
-	if err := copyActionStore(u.log, newHash); err != nil {
+	//if strings.HasPrefix(release.Commit(), newHash) {
```
michalpristas, line 189: how is this handled

cmacknz (commented):
Would it be possible to have an integration test that intentionally tries to upgrade with a hash collision to test this?

Could you clone the package built for the current commit, edit the package version to be different, and then perform the upgrade?

You may not even need to clone it, you could install, then edit the package version, and tell the installed agent to upgrade to itself.

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: Fixed in #4177 
https://github.com/elastic/elastic-agent/pull/4177/files#diff-7f896b3c21ff10574e4b2bfaff289b3ce21f7f34450b815a8a6e838433caa848R252

pchila:
@cmacknz 

> 
> Would it be possible to have an integration test that intentionally tries to upgrade with a hash collision to test this?
> 
> Could you clone the package built for the current commit, edit the package version to be different, and then perform the upgrade?
> 
> You may not even need to clone it, you could install, then edit the package version, and tell the installed agent to upgrade to itself.

I would need a new version both in package version and in the manifest (both version and mappings), probably I would also need to change the package name and top directory to make sure that it's identified as a different version by both the testing framework and the agent upgrade command, that's a lot of boilerplate code and it's hacky.

At this point an integration test would fail as the cleanup/rollback is still broken (we could try it once we have the whole code in the last PR or in a new one)

cmacknz:
> I would need a new version both in package version and in the manifest (both version and mappings), probably I would also need to change the package name and top directory to make sure that it's identified as a different version by both the testing framework and the agent upgrade command, that's a lot of boilerplate code and it's hacky.

> At this point an integration test would fail as the cleanup/rollback is still broken (we could try it once we have the whole code in the last PR or in a new one)

Agree it will be a bit hacky, at the same time this entire PR series has the ultimate goal of making upgrades work despite of hash collisions, so I think we should have a test to prove that the entire end to end upgrade process works in this situation.

We don't need to do it as part of this PR. We can use https://github.com/elastic/elastic-agent/issues/2780 to track the work since it is essentially that issue, we can reword it if necessary.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -7,52 +7,105 @@ package upgrade
...
+	hashBytes, err := io.ReadAll(hashFile)
+	if err != nil {
+		return UnpackResult{}, fmt.Errorf("reading elastic-agent hash file content: %w", err)
+	}
+	if len(hashBytes) < hashLen {
+		return UnpackResult{}, fmt.Errorf("elastic-agent hash %q is too short (minimum %d)", string(hashBytes), hashLen)
+	}
+	hash = string(hashBytes[:hashLen])
+	if versionedHome == "" {
+		// if at this point we didn't load the manifest et the versioned to the backup value
```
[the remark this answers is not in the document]
pchila: Went latin for a moment :sweat_smile: , will fix it without using a dead language :smile:

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -24,15 +25,22 @@ const markerFilename = ".update-marker"
...
+	VersionedHome string `json:"versioned_home" yaml:"versioned_home"`
+
 	//UpdatedOn marks a date when update happened
 	UpdatedOn time.Time `json:"updated_on" yaml:"updated_on"`
 
 	// PrevVersion is a version agent is updated from
 	PrevVersion string `json:"prev_version" yaml:"prev_version"`
 	// PrevHash is a hash agent is updated from
 	PrevHash string `json:"prev_hash" yaml:"prev_hash"`
+	// PrevVersionedHome represents the path where the new agent is located relative to top path
```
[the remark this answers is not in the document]
pchila: yes, yes it is

ycombinator:
Followed the instructions in "how to test this PR locally" and upgrade was successful.  Output of `elastic-agent status` before and after upgrade looks good, including healthy status and upgrade details.  File/folder naming/structure in Agent installation location looks good too. 👍 

Reviewing code now...

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -83,7 +94,16 @@ func Cleanup(log *logger.Logger, currentHash string, removeMarker bool, keepLogs
...
+	var currentDir string
+	if currentVersionedHome != "" {
+		currentDir, err = filepath.Rel("data", currentVersionedHome)
+		if err != nil {
+			return fmt.Errorf("extracting elastic-agent path relative to data directory from %s: %w", currentVersionedHome, err)
+		}
+	} else {
+		currentDir = fmt.Sprintf("%s-%s", agentName, currentHash)
+	}
+
```
ycombinator, line 106: If you extract this logic into its own function, you could write a unit test with a couple of test cases around it.

internal/pkg/agent/application/upgrade/step_relink.go
```diff
@@ -32,21 +32,26 @@ func ChangeSymlink(ctx context.Context, log *logger.Logger, targetHash string) e
 	// paths.BinaryPath properly derives the binary directory depending on the platform. The path to the binary for macOS is inside of the app bundle.
 	newPath := paths.BinaryPath(filepath.Join(paths.Top(), "data", hashedDir), agentName)
 
+	return changeSymlinkInternal(log, symlinkPath, newPath)
+}
+
+func changeSymlinkInternal(log *logger.Logger, symlinkPath, newTarget string) error {
```
ycombinator, line 38: Nit: why not just `changeSymlink`? The "internal" part is enforced by the fact that the function is unexported.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -269,15 +289,19 @@ func (u *Upgrader) sourceURI(retrievedURI string) string {
 	return u.settings.SourceURI
 }
 
-func rollbackInstall(ctx context.Context, log *logger.Logger, hash string) {
-	os.RemoveAll(filepath.Join(paths.Data(), fmt.Sprintf("%s-%s", agentName, hash)))
+func rollbackInstall(ctx context.Context, log *logger.Logger, versionedHome string) {
+	err := os.RemoveAll(filepath.Join(paths.Top(), versionedHome))
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		// TODO should this be a warning or an error ?
+		log.Warnw("error rolling back install", "error.message", err)
```
ycombinator, line 296: I think warning is fine.  It's better than before this PR when we weren't even logging anything if `os.RemoveAll` returned an error.

Also, from my understanding, this function is being called when one of the upgrade steps fails and we want to clean up the target agent's files. The current (pre-upgrade) Agent is still running fine at this point and the callers of this function will return an error about the upgrade step that failed anyway.  So I think just a warning here is fine.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -64,34 +117,46 @@ func unzip(log *logger.Logger, archivePath, dataDir string) (string, error) {
...
+			_, err = os.Stat(dstPath)
+			if errors.Is(err, fs.ErrNotExist) {
+				if err := os.MkdirAll(dstPath, f.Mode().Perm()&0770); err != nil {
+					return fmt.Errorf("creating directory %q: %w", dstPath, err)
+				}
+			} else if err != nil {
+				return fmt.Errorf("stat() directory %q: %w", dstPath, err)
+			} else {
+				// set the appropriate permissions
+				err = os.Chmod(dstPath, f.Mode().Perm()&0o770)
```
ycombinator, line 148: Typo?

```suggestion
				err = os.Chmod(dstPath, f.Mode().Perm()&0770)
```

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -64,34 +117,46 @@ func unzip(log *logger.Logger, archivePath, dataDir string) (string, error) {
...
+			}
+
+			_ = os.MkdirAll(dstPath, f.Mode()&0770)
 		} else {
-			log.Debugw("Unpacking file", "archive", "zip", "file.path", path)
+			log.Debugw("Unpacking file", "archive", "zip", "file.path", dstPath)
 			// remove any world permissions from the directory/file
-			_ = os.MkdirAll(filepath.Dir(path), f.Mode()&0770)
-			f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode()&0770)
+			_ = os.MkdirAll(filepath.Dir(dstPath), 0770)
```
ycombinator, line 158: This is a subtle change from before this PR, where we are now no longer doing a logical AND between `f.Mode()` and `0770. Just want to double check this change was intentional and not a miss.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -153,26 +263,25 @@ func untar(log *logger.Logger, version string, archivePath, dataDir string) (str
...
 		fileName := strings.TrimPrefix(f.Name, fileNamePrefix)
 
 		if fileName == agentCommitFile {
-			hashBytes, err := io.ReadAll(tr)
-			if err != nil || len(hashBytes) < hashLen {
-				return "", err
-			}
-
-			hash = string(hashBytes[:hashLen])
 			continue
```
ycombinator, line 276: Could you add a comment here that we already loaded the hash, similar to what you did in the zip code path? Thanks.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -209,31 +318,38 @@ func untar(log *logger.Logger, version string, archivePath, dataDir string) (str
...
 
-	return hash, nil
+	return UnpackResult{
+		Hash:          hash,
+		VersionedHome: versionedHome,
+	}, nil
+}
+
+func getFileNamePrefix(archivePath string) string {
+	return strings.TrimSuffix(filepath.Base(archivePath), ".tar.gz") + "/" // omitting `elastic-agent-{version}-{os}-{arch}/` in filename
```
ycombinator, line 352: Should we use `os.Separator` here instead of the hardcoded `"/"`?

```suggestion
	return strings.TrimSuffix(filepath.Base(archivePath), ".tar.gz") + string(os.PathSeparator) // omitting `elastic-agent-{version}-{os}-{arch}/` in filename
```

internal/pkg/agent/application/upgrade/step_unpack.go
cmacknz: Also looks like we could calculate `f.Mode().Perm()&0770)` once and just reuse in each of the following lines.

internal/pkg/agent/application/upgrade/step_unpack.go
cmacknz: I think this logic around directory permissions needs some more comments. 

I think the intent is that when we create the file we potentially create the containing directory as 0770 to make sure it exists, because we aren't guaranteed to have seen the directory yet when iterating over the contents.

In the case where we get to the directory that is above this, the intent is to set the permissions the way they were in the zip file, overwriting the permissions set here.

If we already say the directory in the `IsDir` case above, this call does nothing per the documentation of `MkdirAll`.

There was a bug here and/or in the tar installer where we were create the directory with the permissions of the first file we saw that was in that directory, instead of applying the permissions of the directory as they were in the archive.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -242,3 +358,110 @@ func validFileName(p string) bool {
...
+
+type tarCloser struct {
+	tarFile    *os.File
+	gzipReader *gzip.Reader
+}
+
+func (tc *tarCloser) Close() error {
+	var err error
+	if tc.gzipReader != nil {
+		err = multierror.Append(err, tc.gzipReader.Close())
```
cmacknz, line 385: nit: couldn't this be errors.Join()?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -169,46 +170,65 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	}
 
 	det.SetState(details.StateExtracting)
-
-	newHash, err := u.unpack(version, archivePath, paths.Data())
+	// TODO: add check that no archive files end up in the current versioned home
```
cmacknz, line 173: Is this also addressed later?

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz: I'd vote error so we can also see it happening, but it also isn't fatal, so technically warning is fine.

I think the error message should note that the error was specifically from `os.RemoveAll`. That doesn't mean that changing the symlink failed, that is the fatal part of rollbackInstall. We could change this to "failed to remove %directory contents" instead.

If this does fail though, does the agent directory that was left behind just stick around forever? If it does that is something for us to follow up on separately.

cmacknz (commented):
LGTM beside a few small things. Holding approval pending resolution of Shaunak's comments. If he approves feel free to merge.

internal/pkg/agent/application/upgrade/step_relink.go
pchila: This function replaces `ChangeSymlink` in #4178 https://github.com/elastic/elastic-agent/pull/4178/files/6fd94ced07902a2a9baf79ce3e695251b8f63468..b60b518fcaa18ccac4e8801ff1f07090adc0690d#diff-e3bd2c5b20d7e38241b1e80a7fd53c3e2aba0aa4cc1a8d15ced461a13e588a45R22

At this point in time I still had references to the old ChangeSymlink that I wanted to remove so I thought of giving changeSymlinkInternal a very specific name ;)

internal/pkg/agent/application/upgrade/upgrade.go
pchila: @cmacknz
Any leftover directories  other than the one we want to keep gets removed during watcher Cleanup() or Rollback() calls: this is the reason why at this stage watcher breaks the agent install (it still removes anything that is not `elastic-agent-<hash we want to keep>`).

About the error/warning and the message, it has been changed in PR #4178 https://github.com/elastic/elastic-agent/pull/4178/files/6fd94ced07902a2a9baf79ce3e695251b8f63468..b60b518fcaa18ccac4e8801ff1f07090adc0690d#diff-42a05457d0720cf1e61083ad5d04a0b2956e6cfa0e0d018f2f815327cb0f1d2dR377-R383
where the logs are both at warn level and the message explicitly say what went wrong

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: That is consistency with the old code, I tried to be consistent in using `0o770` format in my changes. Not sure if there's a guideline about one or the other

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: This change is the same fix as the one done for `.tar.gz` extraction in #4100.

As Craig said, we create not exisiting containing directories for the file that we are extracting with 0770, and then we we encounter them again while extracting we set the correct permissions.
In the old code we set the same permissions we have for the file to its containing directories which poses a problem for access, modification and traversal since the file may be read-only or not executable.

Fun fact: the bug became apparent the moment I tried to run some unit tests on this code because we don't run those as root so permissions matter.

I will add more comments to explain why we had to change this

internal/pkg/agent/application/upgrade/upgrade.go
pchila: This was an idea for an additional protection of the current agent install against the issue we currently have on main where we extract first, ~~ask questions~~ check the hash later to see if the version we just extracted is the same as the current one. In the end I didn't implement this because for correctly generated packages the new version check is better than what was in place before

For details about the new version check see https://github.com/elastic/elastic-agent/pull/4174/files/ee9979ca0194eedb5c3d8e04e36389e5754ec287#r1474739456

internal/pkg/agent/application/upgrade/rollback.go
pchila: There are unit tests added in #4176 where the cleanup and rollback were fixed https://github.com/elastic/elastic-agent/pull/4176/files/bf2e849b16f8852a59593e45b8ada9095e190ad1..b17c7132e2b6cd2ff157835379c833435f325fc4#diff-cc4058ffab0db853a72dc782e00a1a7c3c7b7314175bec4538b1f32e69332c66

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: added comments in https://github.com/elastic/elastic-agent/pull/4174/commits/9f2a4cc546d99c58e4e4325a46a97abedbffb874

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: Done in https://github.com/elastic/elastic-agent/pull/4174/commits/9f2a4cc546d99c58e4e4325a46a97abedbffb874

internal/pkg/agent/application/upgrade/step_unpack.go
pchila: it can be, this was taken from the existing code which used `multierror` package. I will change it

pchila: done in 9945a142d817b9ae2c21437e6b5a9cb813d61801

ycombinator (approved):
LGTM.

PR #4176, opened 2024-02-01
Support version in watcher cleanup rollback

pchila:
## What does this PR do?

This is a subset of a bigger PR #3950, so CI is not expected to be green, especially regarding the integration tests.

This PR adds new fields in the update marker so that the upgrade watcher Cleanup and Rollback processes can operate correctly with the new `elastic-agent-<version>-<hash>`.

There is some refactoring to make Cleanup() and Rollback() testable that involves injecting paths rather than obtaining through `paths.*` functions: this is needed to be able to write unit tests without calling `paths.SetTop()` which already created flakiness and introduced race conditions in our tests in the past

## Why is it important?

Now the agent installation is correctly cleaned up leaving only the new or old version of agent installed, since the watcher is reading the correct paths from the new update marker

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Author's Checklist

- [ ]

## How to test this PR locally

Testing this PR is mostly the same as the test procedure for #4174, but now the watcher cleans the installation correctly
After the end of grace period the new agent directory should be the only left inside `/opt/Elastic/Agent/data`.
If we want to trigger a rollback, it's enough to kill the new agent a few times and then only the old agent installation will be left inside `/opt/Elastic/Agent/data`.

*Known issues*:
- This has been tested only on Linux. MacOS and Windows fixes will come in a follow-up PR (we need small modifications to make this work mostly related to OS-dependent agent paths).

## Related issues

- Relates #2579 
- Supersedes #3950 
- Requires #4173
- Requires #4174

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4177, opened 2024-02-01
Reintroduce upgrade version check

pchila:
## What does this PR do?

This is a subset of a bigger PR #3950, so CI is not expected to be green, especially regarding the integration tests.

This PR reintroduces a modified version check when upgrading, to ensure we don't upgrade to the exact same version of the current agent. The code will check if the triplet (package version, snapshot flag and hash) are the same between the existing agent installation and the package we downloaded for the upgrade.

## Why is it important?

This prevents any collisions on the filesystem between the agent install and the files we are going to extract from the package.

## Checklist

## Author's Checklist

- [ ]

We can package the agent using

```
```

Install the resulting agent and then try to reuse the same package to upgrade

```
elastic-agent upgrade 8.13.0-SNAPSHOT --source-uri file://<directory where the package is located on disk> --skip-verify
```
We should get an error saying that we are already at the specified version.

To test with a different package we can play with AGENT_PACKAGE_VERSION or SNAPSHOT environment variables to generate another package from the same commit and verify that we can indeed upgrade.

*Known issues*
- This has been tested only on Linux. MacOS and Windows fixes will come in a follow-up PR (we need small modifications to make this work mostly related to OS-dependent agent paths).

## Related issues

- Relates #2579 
- Relates #3950 
- Requires #4173
- Requires #4174
- Requires #4176 

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/step_unpack.go
+++ b/internal/pkg/agent/application/upgrade/step_unpack.go
@@ -87,7 +87,7 @@ func unzip(log *logger.Logger, archivePath, dataDir string) (UnpackResult, error
 	fileNamePrefix := strings.TrimSuffix(filepath.Base(archivePath), ".zip") + "/" // omitting `elastic-agent-{version}-{os}-{arch}/` in filename
 
 	pm := pathMapper{}
-	versionedHome := ""
+	var versionedHome string
 
 	metadata, err := getPackageMetadataFromZipReader(r, fileNamePrefix)
 	if err != nil {
--- a/internal/pkg/agent/application/upgrade/step_unpack.go
+++ b/internal/pkg/agent/application/upgrade/step_unpack.go
@@ -54,6 +54,28 @@ func (u *Upgrader) unpack(version, archivePath, dataDir string) (UnpackResult, e
 	return unpackRes, nil
 }
 
+type packageMetadata struct {
+	manifest *v1.PackageManifest
+	hash     string
+}
+
+func (u *Upgrader) getPackageMetadata(archivePath string) (packageMetadata, error) {
+	ext := filepath.Ext(archivePath)
+	if ext == ".gz" {
+		// if we got gzip extension we need another extension before last
+		ext = filepath.Ext(strings.TrimSuffix(archivePath, ext)) + ext
+	}
+
+	switch ext {
+	case ".zip":
+		return getPackageMetadataFromZip(archivePath)
+	case ".tar.gz":
+		return getPackageMetadataFromTar(archivePath)
+	default:
+		return packageMetadata{}, fmt.Errorf("unknown package format %q", ext)
+	}
+}
+
 func unzip(log *logger.Logger, archivePath, dataDir string) (UnpackResult, error) {
 	var hash, rootDir string
 	r, err := zip.OpenReader(archivePath)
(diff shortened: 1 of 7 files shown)


internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -170,7 +186,40 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
+		// extract version info from the version string (we can ignore parsing errors as it would have never passed the download step)
+		parsedVersion, _ := agtversion.ParseVersion(version)
+		newVersion.version = strings.TrimSuffix(parsedVersion.VersionWithPrerelease(), "-SNAPSHOT")
+		newVersion.snapshot = parsedVersion.IsSnapshot()
+	}
+	newVersion.hash = metadata.hash
+
+	u.log.Debugw("Comparing current and new agent version", "current_version", currentVersion, "new_version", newVersion)
+
+	if currentVersion == newVersion {
```
cmacknz, line 216: This looks like it works by inspection, but it feels like this section could be unit tested. Unless I am missing that this is already covered somewhere else, or the intent is to rely only on the integration tests.

Cases I think of that could be tested:
1. Ensuring that we allow upgrade when only the hash differs. This should be covered by our integration test that upgrades to the same snapshot version.
2. Ensuring we cover the case where one version is from the manifest and the other is from a legacy agent. Again I think the upgrade from 7.17 unit test covers this.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -170,7 +186,40 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
+
+	metadata, err := u.getPackageMetadata(archivePath)
+	if err != nil {
+		return nil, fmt.Errorf("reading metadata for elastic agent version %s package %q: %w", version, archivePath, err)
+	}
+
+	currentVersion := agentVersion{
+		version:  release.Version(),
+		snapshot: release.Snapshot(),
+		hash:     release.Commit(),
```
cmacknz, line 198: This doesn't use the truncated hash that we would read out of a legacy agent without a manifest file. Does that cause problems?

```go
func readCommitHash(reader io.Reader) (string, error) {
	commitBytes, err := io.ReadAll(reader)
	if err != nil {
		return "", fmt.Errorf("reading agent commit hash file: %w", err)
	}
	hash := strings.TrimSpace(string(commitBytes))
	if len(hash) < hashLen {
		return "", fmt.Errorf("hash %q is shorter than minimum length %d", string(commitBytes), hashLen)
	}
	return hash, nil
}
```

pchila: We already truncate the hash in all the relevant functions for the legacy part to 6. I preferred err on the side of caution and use the full hash in the new checks.

internal/pkg/agent/application/upgrade/upgrade.go
pchila: It can surely be unit tested, I didn't do it because I considered the comparison of 2 structs "too simple to fail".
If we a mock package  and a mock agent install into play then it could become interesting but it is also true that it should be pretty simple to have an integration test for this. 
I do not have a strong preference between the 2 options

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -351,6 +378,46 @@ func untar(log *logger.Logger, version string, archivePath, dataDir string) (Unp
 	}, nil
 }
 
+func getPackageMetadataFromTar(archivePath string) (packageMetadata, error) {
+	// quickly open the archive and look up manifest.yaml file
+	fileContents, err := getFilesContentFromTar(archivePath, "manifest.yaml", agentCommitFile)
```
michalpristas, line 383: better to have `manifest.yaml` in a const

pchila: Done already as part of #4178 https://github.com/elastic/elastic-agent/pull/4178/commits/8bbd55f8820e29440db1b80a6c757fcb45ee5a7b

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -195,7 +194,54 @@ func unzip(log *logger.Logger, archivePath, dataDir string) (UnpackResult, error
...
+		ret.manifest, err = v1.ParseManifest(manifestFile)
+		if err != nil {
+			return packageMetadata{}, fmt.Errorf("parsing package manifest: %w", err)
+		}
+	}
+
+	// Load hash, the use of path.Join is intentional since in .zip file paths use slash ('/') as separator
+	hashFile, err := r.Open(path.Join(fileNamePrefix, agentCommitFile))
+	if err != nil {
+		// we got a real error looking up for the manifest
```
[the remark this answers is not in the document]
pchila: Done in b635e49b7fb804a7bb6bf71dcb1f8c180e403458

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -466,5 +533,5 @@ func getFilesContentFromTar(archivePath string, files ...string) (map[string]io.
 }
 
 func createVersionedHomeFromHash(hash string) string {
-	return filepath.Join("data", fmt.Sprintf("elastic-agent-%s", hash[:hashLen]))
+	return fmt.Sprintf("data/elastic-agent-%s", hash[:hashLen])
```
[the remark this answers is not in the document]
pchila: Reverted in b635e49b7fb804a7bb6bf71dcb1f8c180e403458

internal/pkg/agent/application/upgrade/upgrade.go
blakerouse: I agree with @cmacknz on adding a unit test. I don't think he meant specifically to test this one line, I think the whole function in general should have some coverage. Testing the manifest usage and string parsing before this line to ensure that those are filling in the data as expected and the expected error is returned or not returned.

pchila: added in c9b41aae094713fd1ad7a13615644be63fe0840b

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz: Ah I see what I was missing. `metadata.hash` is not truncated anymore, it is truncated when it is used in places where that matters.

```go
	metadata, err := getPackageMetadataFromZipReader(r, fileNamePrefix)
	if err != nil {
		return UnpackResult{}, fmt.Errorf("retrieving package metadata from %q: %w", archivePath, err)
	}

	hash = metadata.hash[:hashLen]
```

I was thinking that `metadata.hash` was truncated and this comparison might evaluate to false but manual testing showed this was working properly.

cmacknz (approved):

blakerouse (approved):
Looks good. Thanks for the tests. Splitting it out into its own function was nice as well, helped test it easier!

PR #4178, opened 2024-02-01
Use watcher that supports paths in update marker

pchila:
## What does this PR do?

This is the last subset of a bigger PR #3950, so CI should be green.

This change checks if the target version of the upgrade is at least 8.13.0 to decide which upgrade watcher to launch. This is needed to ensure that the watcher performs correct cleanup and rollbacks by reading paths in the update marker.

In this PR it's also included a necessary fix to the shutdownCallback since we do some path manipulation to support the legacy structure of component run directory where we replace versions from old to new.

This PR includes also a set of minor fixes to fix the elastic-agent upgrade on MacOS and Windows.

## Why is it important?

This is the final block of code necessary to implement #2579 on Linux, MacOS and Windows. It will make use of the mapping described in the package manifest to correctly place agent files in directories containing the agent version and the hash in the name.

## Checklist

## Author's Checklist

- [ ]

To test this PR we can create multiple packages from the same agent commit and test if we can upgrade between them. The procedure is the same as described in #4174 but without the OS limitation and the known issues.

This change is supposed to be backward compatible so it can be tested also using older agent versions (like we do in our integration tests)

## Related issues

- Relates #2579 
- Relates #3950 
- Requires #4173
- Requires #4174
- Requires #4176 
- Requires #4177 

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/packaging/templates/darwin/elastic-agent.tmpl
+++ b/dev-tools/packaging/templates/darwin/elastic-agent.tmpl
@@ -6,6 +6,9 @@ set -e
 symlink="/Library/Elastic/Agent/elastic-agent"
 
 if test -L "$symlink"; then
-    ln -sfn "data/elastic-agent-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent" "$symlink"
+    symlinkTarget="data/elastic-agent-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent"
+    if test -f "data/elastic-agent-{{ agent_package_version }}{{ snapshot_suffix }}-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent"; then
+        symlinkTarget="data/elastic-agent-{{ agent_package_version }}{{ snapshot_suffix }}-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent"
+    ln -sfn "$symlinkTarget" "$symlink"
 fi
 
--- a/docs/upgrades.md
+++ b/docs/upgrades.md
@@ -67,6 +67,12 @@ sequenceDiagram
 Starting from version 8.13.0 an additional file `manifest.yaml` is present in elastic-agent packages.
 The purpose of this file is to present some metadata and package information to be used during install/upgrade operations.
 
+The first enhancement that makes use of this package manifest is [#2579](https://github.com/elastic/elastic-agent/issues/2579)
+as we use the manifest to map the package directory structure (based on agent commit hash) into one that takes also the
+agent version into account. This allows releasing versions of the agent package where only the component versions change,
+with the agent commit unchanged.
+
+
 The [structure](../pkg/api/v1/manifest.go) of such manifest is defined in the [api/v1 package](../pkg/api/v1/).
 The manifest data is generated during packaging and the file is added to the package files. This is an example of a
 complete manifest:
--- a/docs/upgrades.md
+++ b/docs/upgrades.md
@@ -61,3 +61,32 @@ sequenceDiagram
        end
     end
 ```
+
+### Introducing package manifest
+
+Starting from version 8.13.0 an additional file `manifest.yaml` is present in elastic-agent packages.
+The purpose of this file is to present some metadata and package information to be used during install/upgrade operations.
+
+The [structure](../pkg/api/v1/manifest.go) of such manifest is defined in the [api/v1 package](../pkg/api/v1/).
+The manifest data is generated during packaging and the file is added to the package files. This is an example of a
+complete manifest:
+
+```yaml
+version: co.elastic.agent/v1
+kind: PackageManifest
+package:
+  version: 8.13.0
+  snapshot: true
+  hash: 15658b38b48ba4487afadc5563b1576b85ce0264
+  versioned-home: data/elastic-agent-15658b
+  path-mappings:
+    - data/elastic-agent-15658b: data/elastic-agent-8.13.0-SNAPSHOT-15658b
+      manifest.yaml: data/elastic-agent-8.13.0-SNAPSHOT-15658b/manifest.yaml
+```
+
+The package information describes the package version, whether it's a snapshot build, the elastic-agent commit hash it
+has been built from and where to find the versioned home of the elastic agent within the package.
+
+Another section lists the path mappings that must be applied by an elastic-agent that is aware of the package manifest
+(version >8.13.0): these path mappings allow the incoming agent version to have some control over where the files in
+package will be stored on disk.
--- a/docs/upgrades.md
+++ b/docs/upgrades.md
@@ -101,10 +101,28 @@ Legacy elastic-agent upgrade is a pretty straightforward affair:
 - After extraction check if the hash we read from the package matches with the one from the current agent:
   - if it's the same hash the upgrade fails because we are trying to upgrade to the same version
   - if we extracted a package with a different hash, the upgrade keeps going
-- Copy the elastic agent action store and components run directory into the new agent directories
-- Rotate the symlink in the top directory to point to the new agent executable
-- Write the update marker containing the information about the new and old agent versions in `data` directory
+- Copy the elastic agent action store and components run directory into the new agent directories `elastic-agent-<hash>`
+- Rotate the symlink in the top directory to point to the new agent executable `data/elastic-agent-<hash>/elastic-agent`
+- Write the update marker containing the information about the new and old agent versions/hashes in `data` directory
 - Invoke the watcher `elastic-agent watch` command to ensure that the new version of agent works correctly after restart
 - Shutdown current agent and its command components, copy components state once again and restart
 
 #### Upgrading using the manifest
+
+Upgrading using the manifest allows for the new version to pass along some information about the package to the upgrading agent.
+The new process looks like this:
+- Download the elastic-agent package to use for upgrade
+- Extract package metadata from the new elastic-agent package (manifest and hash):
+  - if the package has a manifest we extract `version` and `snapshot` flag as declared by the package
+  - if there is no manifest for the package we extract the version and snapshot for the version passed to the upgrade
+  - hash from the agent commit file (always present in the package)
+- compare the tuple of new `(version, snapshot, hash)` to the current `(version, snapshot, hash)`: if they are the same
+  the upgrade fails because we are trying to upgrade to the same version as current
+- Extract any package file (after mapping it using file mappings in manifest if present) that should go under `/data`.
+  Return the new versionedHome (where the new version of agent has its files, returned as path relative to the top directory)
+- Copy the elastic agent action store and components run directory into the new agent in `<versionedHome>/run`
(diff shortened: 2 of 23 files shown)


docs/upgrades.md
```diff
@@ -61,3 +61,68 @@ sequenceDiagram
        end
     end
 ```
+
+### Introducing package manifest
+
+Starting from version 8.13.0 an additional file `manifest.yaml` is present in elastic-agent packages.
```
cmacknz, line 67: Documentation is great, let's add why we did this. To allow releasing versions of the agent package where only the component versions change, with the agent commit unchanged.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -258,30 +265,51 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
-		rollbackInstall(ctx, u.log, paths.Top(), hashedDir)
+		rollbackInstall(ctx, u.log, paths.Top(), hashedDir, currentVersionedHome)
 		return nil, err
 	}
 
-	if err := InvokeWatcher(u.log); err != nil {
+	minParsedVersionForNewUpdateMarker := agtversion.NewParsedSemVer(8, 13, 0, "", "")
+	var watcherExecutable string
+	if parsedVersion.Less(*minParsedVersionForNewUpdateMarker) {
+		// use the current agent executable for watch
```
cmacknz, line 304:
```suggestion
		// use the current agent executable for watch, if downgrading the old agent doesn't understand the current agent's path structure.
```

I think this could be a little bit clearer.

docs/upgrades.md
pchila: Have a look at b87a013af7e3719dde052c5375eba0d92d95f26c
I tried adding a bit more context on that

cmacknz: Perfect, thanks!

dev-tools/packaging/templates/darwin/elastic-agent.tmpl
```diff
@@ -6,6 +6,9 @@ set -e
 symlink="/Library/Elastic/Agent/elastic-agent"
 
 if test -L "$symlink"; then
-    ln -sfn "data/elastic-agent-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent" "$symlink"
+    symlinkTarget="data/elastic-agent-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent"
+    if test -f "data/elastic-agent-{{ agent_package_version }}{{ snapshot_suffix }}-{{ commit_short }}/elastic-agent.app/Contents/MacOS/elastic-agent"; then
```
cmacknz, line 10: Out of curiosity why is this necessary? I though the package structure was static at the point of packaging?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -244,30 +251,51 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
+	current := agentInstall{
+		version:       version,
+		hash:          unpackRes.Hash,
+		versionedHome: unpackRes.VersionedHome,
+	}
+
+	previous := agentInstall{
+		version:       release.VersionWithSnapshot(),
+		hash:          release.Commit(),
+		versionedHome: currentVersionedHome,
```
cmacknz, line 274: It looks like you have swapped the `current` and `previous` names, since the `previous` version is getting the current values of version, hash, and version home.

dev-tools/packaging/templates/darwin/elastic-agent.tmpl
pchila: This is the script that fixes the links to the elastic-agent executable within the `elastic.app` directory. Now the path has potentially changed depending on how the package has been installed so the symlink target needs to be tested and set to the existing path

cmacknz: Ah I missed that this script was executed at runtime, thanks. Thought it was part of the packaging process itself.

internal/pkg/agent/application/upgrade/upgrade.go
pchila: It is correct: we have already rotated the symlink, so in the update we write the `current ` as the version we just upgraded to and the `previous` is the old agent installation. Confusing, I know but the nomenclature of the update marker is based on these definitions

cmacknz: Thanks, this would be worth a comment in the code. The `previous.versionedHome.currentVersionedHome` seems suspicious when you read it.

pchila: 22606d7985ef6190e9ab0f89a60305b3136ffe13

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -46,7 +46,7 @@ func Rollback(ctx context.Context, log *logger.Logger, c client.Client, topDirPa
 		symlinkTarget = paths.BinaryPath(filepath.Join(paths.DataFrom(topDirPath), hashedDir), agentName)
 	}
 	// change symlink
-	if err := changeSymlinkInternal(log, topDirPath, symlinkPath, symlinkTarget); err != nil {
+	if err := ChangeSymlink(log, topDirPath, symlinkPath, symlinkTarget); err != nil {
```
ycombinator, line 49: I think this could be left as-is or even just renamed to `changeSymlink`.  IOW, it doesn't look like this function is called from outside the `upgrade` package so no need to export it.

internal/pkg/agent/application/upgrade/step_download.go
```diff
@@ -127,7 +122,7 @@ func (u *Upgrader) downloadArtifact(ctx context.Context, version, sourceURI stri
 	return path, nil
 }
 
-func (u *Upgrader) appendFallbackPGP(targetVersion string, pgpBytes []string) []string {
+func (u *Upgrader) appendFallbackPGP(tpv *agtversion.ParsedSemVer, pgpBytes []string) []string {
```
ycombinator, line 125: Could we just continue to call it `targetVersion` instead of the less descriptive `tpv` please? 🙂

ycombinator (approved):
Left a couple of minor comments about naming; otherwise, LGTM!

blakerouse:
The description says this is the final block of code required to implement #2579, but doesn't mark the issue as closed when this lands. Is a follow up needed? If so does that follow-up include an integration test? I would really like to see an integration tests that possibly takes the currently extracted Elastic Agent, re-archives it with a different manifest and performs an upgrage.

cmacknz (approved):
Same, a manual test of an upgrade where the agent commit hash is the same but the version is different succeeds. Thanks!

```
❯ sudo elastic-agent version
Binary: 8.16.1-SNAPSHOT (build: f6a3b8203762d074ae57eb8e3b28b15ca68cdea7 at 2024-02-05 21:06:09 +0000 UTC)
Daemon: 8.16.1-SNAPSHOT (build: f6a3b8203762d074ae57eb8e3b28b15ca68cdea7 at 2024-02-05 21:06:09 +0000 UTC)
~/Downloads ················································································· 04:23:55 PM
❯ sudo elastic-agent upgrade 8.16.2-SNAPSHOT --skip-verify --source-uri file:///Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/distributions
Upgrade triggered to version 8.16.2-SNAPSHOT, Elastic Agent is currently restarting
~/Downloads ············································································· 17s 04:24:18 PM
❯ sudo elastic-agent version
Binary: 8.16.2-SNAPSHOT (build: f6a3b8203762d074ae57eb8e3b28b15ca68cdea7 at 2024-02-05 21:11:59 +0000 UTC)
Daemon: 8.16.2-SNAPSHOT (build: f6a3b8203762d074ae57eb8e3b28b15ca68cdea7 at 2024-02-05 21:11:59 +0000 UTC)
```

Grabbing diagnostics the versions appear as expected as well:

```
 tree logs
logs
├── elastic-agent-8.16.1-SNAPSHOT-f6a3b8
│   ├── elastic-agent-20240205.ndjson
│   ├── elastic-agent-watcher-20240205-1.ndjson
│   └── elastic-agent-watcher-20240205.ndjson
└── elastic-agent-8.16.2-SNAPSHOT-f6a3b8
    ├── elastic-agent-20240205.ndjson
    ├── elastic-agent-watcher-20240205-1.ndjson
    └── elastic-agent-watcher-20240205.ndjson

2 directories, 6 files
```

Great job working through all this @pchila!

cmacknz:
> The description says this is the final block of code required to implement https://github.com/elastic/elastic-agent/issues/2579, but doesn't mark the issue as closed when this lands. Is a follow up needed? If so does that follow-up include an integration test? I would really like to see an integration tests that possibly takes the currently extracted Elastic Agent, re-archives it with a different manifest and performs an upgrage.

Agree, I think I made the same comment in one of the earlier PRs. I did this exact test manually several times through these PRs, would be nice to automate it.

Let's track this in https://github.com/elastic/elastic-agent/issues/2780. I'll assign it to @pchila.

pchila:
> The description says this is the final block of code required to implement #2579, but doesn't mark the issue as closed when this lands.

@blakerouse The big PR from the feature branch to main will close the issue. I would not consider the feature implemented if it's not available on main.

For the integration test we need to hack an existing package which is not something I am extremely fond of...

blakerouse:
> > The description says this is the final block of code required to implement #2579, but doesn't mark the issue as closed when this lands.
> 
> @blakerouse The big PR from the feature branch to main will close the issue. I would not consider the feature implemented if it's not available on main.

Sorry was not paying attention, didn't realize this was for a feature branch.

> 
> For the integration test we need to hack an existing package which is not something I am extremely fond of...

If you have a better way then that would be great, otherwise I am not sure we can workaround it. Better to have a hacky test that tests a core feature, then no test at all.

blakerouse (approved):

internal/pkg/agent/application/upgrade/rollback.go
pchila: 280202026ca9bd5b2c78412f70ba5e77cd27f474

internal/pkg/agent/application/upgrade/step_download.go
pchila: b852a8371293bfcd808ddc63e0e1488e6c11bb09

pchila:
>> For the integration test we need to hack an existing package which is not something I am extremely fond of...
> 
> If you have a better way then that would be great, otherwise I am not sure we can workaround it. Better to have a hacky test that tests a core feature, then no test at all.

@blakerouse Ideally we should package the commit twice with different versions and have both packages available on the runner but the current integration test framework does not allow that AFAIK.

I will have a look at what is possible in #2780 without sinking more time in the integration test framework

pchila:
Integration test failed because of known flakiness in (linux-arm64-ubuntu-2204-upgrade)(sudo)  runner. Merging on feature branch an preparing the PR to main...

PR #4179, opened 2024-02-01
Fix daemon reload with backoff SIGINT handling

aleksmaus:
## What does this PR do?

Fixes SIGINT handling while doing daemon reload with backoff.

## Why is it important?

Currently SIGINT doesn't interrupt the daemon reload backoff retries upon enrollment. 
Can be reproduced for example by starting ```./elastic-agent enroll ...........``` command, and issuing Ctrl+C, agent continues running.
This fixes it. 

## Checklist


internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -486,8 +490,10 @@ func (c *enrollCmd) daemonReloadWithBackoff(ctx context.Context) error {
 		lastErr = err
 
 		c.log.Errorf("Restart attempt %d failed: '%s'. Waiting for %s", attempt, err, backExp.NextWait().String())
-		backExp.Wait()
-
+		// backoff Wait returns false if context.Done()
+		if !backExp.Wait() {
```
cmacknz, line 494: We should add a unit test here to catch this, the simplest way I can see would be something like:

1. Setup a context that is either already cancelled or with a timeout much shorter than `daemonReloadInitBackoff`.
2. Use our mock control client to get past the call to daemonReload returning an error that is not a context error.
3. Ensure the function returns a context error faster than `daemonReloadInitBackoff`.

The mock control client:
https://github.com/elastic/elastic-agent/blob/d8b446f5a5686c99415e823edf321699fbea4b26/pkg/control/v2/client/mocks/client.go#L633-L637

Example usage:

https://github.com/elastic/elastic-agent/blob/d8b446f5a5686c99415e823edf321699fbea4b26/internal/pkg/agent/application/upgrade/upgrade_test.go#L192-L196

pchila (commented):
The code looks reasonable,
I agree with @cmacknz : a unit test would be nice.

aleksmaus:
Added unit tests. Injecting ```daemonReloadFunc``` instead. 
From the point of view of testing the  ```daemonReloadInitBackoff``` function we don't need to to inject the full mock client nor define it's behavior (it calls three different functions).

cmacknz (approved):
Thanks for the test!

PR #4183, opened 2024-02-02
Set GOTMPDIR to a child directory

rdner:
So, Windows does not detect temporary Go binaries (e.g. magefile) as a virus.

This solves:

```
failed to run compiled magefile: fork/exec C:\Users\windows\magefile\6e561802bbe21387b05a72ec1f51afbf0b1d8e8d.exe: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
```


pierrehilbert (approved):

cmacknz:
This can also go into 8.12

PR #4184, opened 2024-02-02
WIP: fix empty version when loading upgrade action from disk

### LINKED ISSUE #3912: Scheduled upgrade fails with `error parsing version ""` if agent restarts before the upgrade starts
- Version: 8.12/main
- Operating System: all
- Discuss Forum URL: N/A
- Steps to Reproduce:
  - schedule and upgrade on fleet for 2 min form now
  - check the agent report the upgrade is scheduled
  - restart the agent
  - wait the upgrade to start
  - it'll fail:
```
# elastic-agent status                                                                                                 

┌─ fleet
├─ elastic-agent
└─ upgrade_details
   ├─ target_version:
   ├─ state: UPG_FAILED
   ├─ action_id: 9c06cd82-50e5-41bc-8a68-ed73b55c1912
   └─ metadata
      ├─ failed_state: UPG_DOWNLOADING
      └─ error_msg: could not download artifact: error parsing version "": version string does not match expected format
```

The issue seems to be on how the actions are loaded from disk. Using the patch [patch.txt](https://github.com/elastic/elastic-agent/files/13670257/patch.txt) we can see the test will fail because the upgrade action version isn't loaded.
If you stop the test and check the store file, it contains all the data, but it sin't loaded correctly:
 - state.yml:
```yaml
action_queue:
- action_id: test
  type: UPGRADE
  start_time: "2023-12-14T08:09:59Z"
  version: 1.2.3
  source_uri: https://example.com
  retry_attempt: 1
- action_id: abc123
  type: POLICY_CHANGE
  policy:
    hello: world
``` 

### Bug root cause

TL;DR: we deserialise the action from one schema (the [`fleetapi.FleetAction`](https://github.com/elastic/elastic-agent/blob/v8.12.1/internal/pkg/fleetapi/action.go#L93-L106)) and serialise a different schema (our concrete action types, like the [`fleetapi.ActionUpgrade`](https://github.com/elastic/elastic-agent/blob/v8.12.1/internal/pkg/fleetapi/action.go#L226-L236). When reading from disk, we loose data, such as the `version` from the upgrade actions.

### Proposed fix
 - use only JSON to serialise/deserialise the actions
 - make all `fleetapi.ActionTYPE` models to match the schema we receive from fleet. Add any action specific properties nested under `data`
 - migrate the current state store in YAML to a new state store using JSON
   - properly deserialise the old store to do not lose data during the migration

### Fix implementation phases (using a feature branch):
 - [x] Create a test for the migration from the old action_store to the current state_store using a golden file. The current test serialises the action_store on the fly, which renders the test pointless if there is any changes to the serialisation process. https://github.com/elastic/elastic-agent/pull/4235
   - https://github.com/elastic/elastic-agent/pull/4235
 - [x] https://github.com/elastic/elastic-agent/pull/4253
 - [x] change the `fleetapi.ActionTYPE` to match the fleet-server model by adding the specific action properties nested under a `data` property. https://github.com/elastic/elastic-agent/pull/4240
 - [x] write the migration from the current state_store in YAML to a new one in JSON https://github.com/elastic/elastic-agent/pull/4305
 - [ ] merge the feature branch to main https://github.com/elastic/elastic-agent/pull/4441
 - [ ] integration test: https://github.com/elastic/elastic-agent/issues/4560
 - [ ] QA test: https://github.com/elastic/elastic-agent/issues/4962

AndersonQ:
## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #3912 

## Questions to ask yourself

- ...


AndersonQ:
TL;DR: YAML isn't JSON. Ok, let me explain that:

When using JSON it's possible to have a `[]byte` field and when unmarshaling it have the byte representation of the JSON string and it can be parsed later. That's the strategy used to marshal/unmarshal a 
[`[].fleet.Action`](https://github.com/AndersonQ/elastic-agent/blob/dfd9554406f581660747663d15476c636095fa20/internal/pkg/fleetapi/action.go#L47), an interface, which actually parses the data into a [`[]fleet.FleetAction`](https://github.com/AndersonQ/elastic-agent/blob/dfd9554406f581660747663d15476c636095fa20/internal/pkg/fleetapi/action.go#L93-L100) then into the specific `Action`.

It works for the [fleet defined JSON `action`](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/elastic/fleet-server/main/model/openapi.yml#/), which has a opaque `data` field with a different schema for each cation type. In Go, it's a `json.RawMessage`, a.k.a `[]byte` and it works just fine. However the same isn't true for YAML, it's not even possible to have a raw field (see [this](https://github.com/go-yaml/yaml/issues/13#issuecomment-758118980)), and we use `yaml.v2`. Besides the concrete `Action` structs we use do not have a `data` field, they're flat. What makes their marshaled version different from what they were unmarshaled from.

That's why we're loosing data like the `version` for the upgrade actions, the original data inside the `data` field is marshaled to disk in a different place than the one expected by unmarshal. Again, as JSON isn't YAML, even if the specific fields for each action were saved to disk under a `data` field, the code would break anyway because the `data` field would fail to unmarshal to a `[]byte/json.RawMessage`.

Some possible solutions:

 1) **Refactor Actions to Reflect JSON Schema**:
So far I believe the best solution would be to refactor all the actions to reflect the [JSON schema](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/elastic/fleet-server/main/model/openapi.yml#/) by having a nested `Data` struct with the action's specific fields. This should make the marshal/unmarshal trivial. However this means changing a substantial amount of code and therefore the potential for things to break.

 1) **Do not use YAML for actions**:
We could also just get rid of the yaml for actions. I confess right now I wonder if we really need yaml for them. Using only JSON could be a easy solution. And of course adding some good old panic when marshalling/unmarshalling the actions so we'd be sure no one uses them with YAML. 

 1) **Explore if `yaml.v3` can replicate the JSON raw message behaviour**:
We could try to use `yaml.v3` to see if it'd be possible to replicate the JSON behaviour. However, I haven't explored all that could be done if we'd to use `yaml.v3` because I'm concerned if we don't change everywhere we might end up with some package using the v2 and messing things when marshalling/unmarshalling the actions.

 1) **Have specific structs used only to marshal/unmarshal actions**:
Another option is to have specific structs to represent the YAML version of the actions, those structs would only be used for marshalling/unmarshalling what would hide the complexity of making the actions work in JSON form as well as in YAML form.

**Regardless of the solution**: we also need some good set of tests to ensure this mix and match of types that an action can be do not loose data when persisting them to disk and reading them back. Mainly if we add new fields or action types. It's exactly what our current tests are missing, to assert the specific fields for each action type.

Right now, for me, I'd like to avoid future problems, what requires a solution allowing the model of the actions be altered or new actions be created without requiring to change any marshal/unmarshal code. What makes me side with *1)* and accept the burden of the refactor. However I haven't accessed how far all the actions spread in the agent's code, so I cannot guess the amount of work it'd require.

cmacknz:
Consistently using JSON everywhere seems like the correct approach to eliminate compatibility errors. I am unsure why we use YAML to store actions on disk. Perhaps @michel-laterman knows or has can give us guidance on the best approach since he was involved in the original implementation.

The one complication to changing the format is that actions are persisted on disk, which means if we change the format we have to treat this as a data migration. 

I don't think there are a lot of actions persisted in general, in fact I'm not sure the most important one in upgrades even gets persisted. That said we should avoid throwing away actions that otherwise would have been processed. I can think of a couple of ways to deal with this, but am open to suggestion on which approach we should take.

michel-laterman:
@cmacknz  I think yaml was used to be consistent with the format of the other files the elastic-agent wrote to disk.

AndersonQ:
closed in favour of https://github.com/elastic/elastic-agent/pull/4441

PR #4191, opened 2024-02-05
Fix upgrade tests, so they use correct snapshots

rdner:
TestStandaloneUpgradeRetryDownload,
TestStandaloneUpgradeUninstallKillWatcher were using snapshots with a matching hash instead of filtering builds by a commit hash.

Fixes https://github.com/elastic/elastic-agent/issues/4152

### DIFF
--- a/pkg/testing/tools/artifacts_api.go
+++ b/pkg/testing/tools/artifacts_api.go
@@ -179,10 +179,11 @@ func (aac ArtifactAPIClient) GetBuildsForVersion(ctx context.Context, version st
 	return checkResponseAndUnmarshal[VersionBuilds](resp)
 }
 
-// FindBuild returns a build of the given `version` that does not match the `excludeHash` commit hash
-// and starts with the `offset` index on the current list of builds.
+// FindBuild returns a build of the given `version` that does not match the
+// `excludeHash` commit hash. It searches for a matching build from the latest
+// to the oldest, starting the search at the `offset` index in the list of builds.
+// Setting `offset` to 0 includes all builds, 1 skips the latest, and so forth.
 // If there are no builds matching these conditions, returns `ErrBuildNotFound`.
-// `offset` value is used when the consumer does not wish to request the latest build.
 func (aac ArtifactAPIClient) FindBuild(ctx context.Context, version, excludeHash string, offset int) (buildDetails *BuildDetails, err error) {
 	resp, err := aac.GetBuildsForVersion(ctx, version)
 	if err != nil {


rdner:
The linter complaints about unchanged code.

pierrehilbert:
@rdner even if you didn't touch that part of the code, wouldn't be better to make the linter happy by adding the returned values instead of the naked return and therefore improve the readability?

rdner:
@pierrehilbert 1. I don't think the readability is impaired. I'd like to keep my change focused on a fix, not touching the code that does not need to be touched.

The linter is supposed to report issues only in the current diff, there is some kind of bug that triggers the whole file instead.

pkg/testing/tools/artifacts_api.go
```diff
@@ -178,6 +181,27 @@ func (aac ArtifactAPIClient) GetBuildsForVersion(ctx context.Context, version st
 	return checkResponseAndUnmarshal[VersionBuilds](resp)
 }
 
+func (aac ArtifactAPIClient) FindBuild(ctx context.Context, version, excludeHash string, offset int) (buildDetails *BuildDetails, err error) {
```
AndersonQ, line 184: [Blocker]
Please, could you document what does `offset` do? It's not trivial to understand without reading the code and even reading the code it isn't clear why one would set it or hot to know which offset to use

AndersonQ: I'm trying to understand how one would know which offset to set if a priori the number of builds isn't known.

testing/integration/upgrade_downgrade_test.go
```diff
@@ -61,33 +58,13 @@ func TestStandaloneDowngradeToSpecificSnapshotBuild(t *testing.T) {
...
-		require.NoError(t, err)
-		if details.Build.Projects[artifactElasticAgentProject].CommitHash != startVersion.Binary.Commit {
-			upgradeVersionString = buildID
-			break
-		}
-		t.Logf("build %q matches the current commit hash %q, skipping...", buildID, startVersion.Binary.Commit)
-	}
-
-	if upgradeVersionString == "" {
+	buildInfo, err := aac.FindBuild(ctx, preReleaseVersion, startVersion.Binary.Commit, 1)
```
AndersonQ, line 61: Why setting the offset to 1? Is it just to guarantee there will be at least 2 snapshots available? If so, is it necessary? There is already the build hash to guarantee the same build will not be used.
I'd understand using the offset to guarantee only newer builds are used, but from what understand, it's not the case

michalpristas:
fixing linter will free you from being dependent on force-pushers. quick fix

testing/integration/upgrade_standalone_retry_test.go
```diff
@@ -38,13 +41,37 @@ func TestStandaloneUpgradeRetryDownload(t *testing.T) {
...
+	buildFragments := strings.Split(upgradeVersionString, "-")
+	require.Lenf(t, buildFragments, 2, "version %q returned by artifact api is not in format <version>-<buildID>", upgradeVersionString)
+	endVersion := version.NewParsedSemVer(
+		startVersion.Major(),
+		startVersion.Minor(),
+		startVersion.Patch(),
+		startVersion.Prerelease(),
+		buildFragments[1],
+	)
 	endFixture, err := atesting.NewFixture(
```
AndersonQ, line 72: [Suggestion]
Perhaps it could be moved to a helper function

AndersonQ (changes requested):

testing/integration/upgrade_downgrade_test.go
rdner: This is already documented above, please read the comment.

pkg/testing/tools/artifacts_api.go
rdner: Documented in https://github.com/elastic/elastic-agent/pull/4191/commits/d0dea5d98a4894dd004ebeeebb3e08ab3cb62fc2

rdner:
@michalpristas AFAIK the linter check is not required unless someone changed that recently. I fixed the issues but I still disagree that one must fix all of them in untouched code.

testing/integration/upgrade_downgrade_test.go
rdner: > If so, is it necessary? There is already the build hash to guarantee the same build will not be used.

I gave the same argument to @pchila (who wrote this test) in a call after his comment here https://github.com/elastic/elastic-agent/pull/4095#discussion_r1457430896 but apparently you and me don't understand the test well enough. So, I kept the logic as it was before.

pierrehilbert:
I thought it was part of our guidelines to make the linter happy when we are touching a file with linter issue to use this as an opportunity instead of chasing them on purpose.
Let's discuss this in our next team weekly meeting to define our way to handle this.

pkg/testing/tools/artifacts_api.go
```diff
@@ -165,32 +167,55 @@ func (aac ArtifactAPIClient) GetVersions(ctx context.Context) (list *VersionList
...
 
 	defer resp.Body.Close()
 	return checkResponseAndUnmarshal[VersionBuilds](resp)
 }
 
+// FindBuild returns a build of the given `version` that does not match the `excludeHash` commit hash
+// and starts with the `offset` index on the current list of builds.
+// If there are no builds matching these conditions, returns `ErrBuildNotFound`.
+// `offset` value is used when the consumer does not wish to request the latest build.
+func (aac ArtifactAPIClient) FindBuild(ctx context.Context, version, excludeHash string, offset int) (buildDetails *BuildDetails, err error) {
```
AndersonQ, line 186: [Blocker]
So if I understood this doc, and considering the artifactAPI return a list of builds where the first, index 0, is the most recent build, then I get how to use the offset. However I think either it could be a boolean `skipLatestBuild` if there isn't a reason to set offset to anything grater than 1. Or the doc should explain _the artifactAPI return a list of builds where the first, index 0, is the most recent build ..._. Without further explanation it's still cryptic why sometimes it's set to `1`, other times to `0` and never to another value.
Would you mind adding some more explanation?

Just as an example of what I mean:

  ```suggestion
  // FindBuild returns a build of the given `version` that does not match the
  // `excludeHash` commit hash. It searches for a matching build from the latest
  // to the oldest, starting the search at the `offset` index in the list of builds.
  // Setting `offset` to 0 includes all builds, 1 skips the latest, and so forth.
  // If there are no builds matching these conditions, returns `ErrBuildNotFound`.
  func (aac ArtifactAPIClient) FindBuild(ctx context.Context, version, excludeHash string, offset int) (buildDetails *BuildDetails, err error) {
  ```

michalpristas: would naming it `FindLastBuildBut(offset, ...)` make it more readable for you @AndersonQ 
i think using offset gives us flexibility

testing/integration/upgrade_downgrade_test.go
pchila: Adding this here, if the comment is not clear feel free to update that.

This test verifies that agent is able to upgrade to a *specific snapshot build* using `elastic-agent upgrade x.y.z-SNAPSHOT+<buildid>` . This functionality has been added on top of the existing "upgrade to latest snapshot build" which has the form of `elastic-agent upgrade x.y.z-SNAPSHOT`.

The reason why the test does not select the latest snapshot build is that the result would be indistinguishable between the 2 forms of the upgrade command (those would select the same build) so we could silently break the upgrade to specific snapshot feature while still having a green test.

Hope that helps...

AndersonQ (approved):

Issue #3873, opened 2023-12-06
[OTel Collector] Support for feature gates


ycombinator:
Open telemetry collector now supports feature gate for controlling usage of experimental features. 

Elastic agent should support this flag on a run command when running in otel mode. 
https://github.com/open-telemetry/opentelemetry-collector/blob/cefe1b7d647cd5859f10a7de4e18b46eb3378c5d/featuregate/README.md?plain=1#L41

DoD:
- `elastic-agent otel --feature-gate` can be used and is replacement for `otelcol run --feature-gates` 
- same syntax of feature gate values is accepted
- disabled features are not used

PR #4198, opened 2024-02-06
Enable feature-gates for otel subcommands

michalpristas:
Added support for [feature gates](https://github.com/open-telemetry/opentelemetry-collector/blob/main/featuregate/README.md) for `otel` and `validate` commands

### How to test

- Build/Package agent
- run `./elastic-agent otel --feature-gates "-telemetry.useOtelForInternalMetrics"`
- as this is `stable` feature gate, you should see error returned saying `useOtelForInternalMetrics` is stable and cannot be disabled
- running `./elastic-agent otel --feature-gates telemetry.useOtelForInternalMetrics` (difference is `-` in front of a gate name) should result in a warning message in output saying it's stable and collector running normally 
- running `./elastic-agent otel -h` needs to include flag in help message

Fixes: #3873

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,6 +65,7 @@ require (
 	go.opentelemetry.io/collector/exporter v0.93.0
 	go.opentelemetry.io/collector/exporter/debugexporter v0.93.0
 	go.opentelemetry.io/collector/exporter/otlpexporter v0.93.0
+	go.opentelemetry.io/collector/featuregate v1.0.1
 	go.opentelemetry.io/collector/otelcol v0.93.0
 	go.opentelemetry.io/collector/processor v0.93.0
 	go.opentelemetry.io/collector/processor/batchprocessor v0.93.0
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -65,6 +65,7 @@ require (
 	go.opentelemetry.io/collector/exporter v0.93.0
 	go.opentelemetry.io/collector/exporter/debugexporter v0.93.0
 	go.opentelemetry.io/collector/exporter/otlpexporter v0.93.0
+	go.opentelemetry.io/collector/featuregate v1.0.1
```
ycombinator, line 68: Should this version be kept in sync with the other OTel modules?

ycombinator (dismissed):
Minor comment / question about the `featuregate` module's version; otherwise LGTM.

ycombinator:
Actually, one more suggestion: could you add an integration test that proves that the `--feature-gates` flag works?  Something like `./elastic-agent otel --feature-gates foo.bar` returns an error starting with `invalid argument "foo.bar" for "--feature-gates" flag: no such feature gate "foo.bar". valid gates:`?

go.mod
michalpristas: good point

ycombinator (commented):
Actually, one more suggestion: could you add an integration test that proves that the `--feature-gates` flag works?  Something like `./elastic-agent otel --feature-gates foo.bar` returns an error starting with `invalid argument "foo.bar" for "--feature-gates" flag: no such feature gate "foo.bar". valid gates:`?

internal/pkg/agent/cmd/common.go
```diff
@@ -43,7 +43,7 @@ func NewCommandWithArgs(args []string, streams *cli.IOStreams) *cobra.Command {
 
 	// Init version information contained in package version file
 	err := version.InitVersionError()
-	if err != nil {
+	if isOtel := len(args) > 1 && args[1] == "otel"; !isOtel && err != nil {
```
ycombinator, line 46: Should we even try to initialize the version (previous line) if we are in `otel` mode?

ycombinator (approved):

PR #4202, opened 2024-02-06
Serverless: switch to _reset-internal-credentials.

cmacknz:
The previous _reset-credentials endpoint will soon return a less privileged user. The internal variant returns one with operator privileges to allow inspecting cluster health for testing purposes.


cmacknz:
buildkite test it

AndersonQ (approved):

belimawr (approved):

PR #4222, opened 2024-02-07
[Independent Agent Release] New pipeline upload step for agent package for Independent Agent Release

dwhyrock:
## What does this PR do?
This adds an alternate upload process for the `elastic-agent-package` pipeline for the Independent Agent Release workflow.

### Workflow
The [Independent Agent Staging pipeline](https://buildkite.com/elastic/independent-agent-release-staging) will call into the `elastic-agent-package` pipeline.

Instead of calling into the Release Manager container for uploading/processing artifacts, this will use a more simple approach to use `buildkite-agent artifact upload` to store all the artifacts, and let the calling pipeline handle the rest.

In this case, only use the new workflow when `BUILDKITE_TRIGGERED_FROM_BUILD_PIPELINE_SLUG == "independent-agent-release-staging"`

The new workflow will use the Release Manager `collect` in `dry-run` mode to verify that all of the expected artifacts are present.

### Docker Tag Adjustment
There is also an issue with Docker tags where it doesn't accept the `+` character in the tag name.  This PR also changes the `+` to a `.` if the `+` is in the version string.  We use the `.` since the `-` may be evaluated as a pre-release in semver.

### Ironbank Workaround
Currently, the `ironbank` container image does not build and use the `AGENT_PACKAGE_VERSION` env var to set the version.  This causes the Release Manager container to fail to verify since it can't find the expected `ironbank` file with the new version.

It was decided that we would not include the `ironbank` containers as part of the Independent Agent Release.  Due to that, I added a simple workaround to copy the `ironbank` container file to a new file with the proper version.

### pkg/errors Linting Fixes
There was a linting error complaining about using `github.com/pkg/errors` as an import.

The built-in `errors` import doesn't have the `Wrap` or `Wrapf` functions, so I replaced them with `fmt.Errorf()` function calls.  **This code is probably worth some scrutiny to make sure I did it correctly.**

Issue: https://github.com/elastic/ingest-dev/issues/2898

## Why is it important?
To provide a way to upload Agent Package artifacts for the Independent Agent workflow that doesn't use the Release Manager container.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...


AndersonQ:
@dwhyrock as it's a draft, would you mind changing the PR to draft?

dev-tools/mage/manifest/manifest.go
```diff
@@ -99,6 +101,13 @@ func DownloadComponentsFromManifest(manifest string, platforms []string, platfor
 	}
 	projects := manifestResponse.Projects
 
+	// If using new Independent Agent version, just use the first X.Y.Z part
+	majorMinorPatchVersion := manifestResponse.Version
```
alexsapran, line 105: nit, for parsing the version we can use the https://github.com/elastic/elastic-agent/tree/main/pkg/version package and extract the part of the version we want.

dev-tools/mage/dockerbuilder.go
```diff
@@ -178,6 +178,7 @@ func (b *dockerBuilder) expandDockerfile(templatesDir string, data map[string]in
 
 func (b *dockerBuilder) dockerBuild() (string, error) {
 	tag := fmt.Sprintf("%s:%s", b.imageName, b.Version)
+	tag = strings.Replace(tag, "+", "_", 1)
```
cmacknz, line 181: https://github.com/opencontainers/distribution-spec/issues/154

I see a dash '-' used most frequently to replace the '+', this seems to be what OCI itself was doing before https://github.com/opencontainers/distribution-spec/pull/417/files

So I have a small preference for that just to avoid doing this in a way that isn't consistent with other projects that also have this problem.

dwhyrock:
I confirmed that this code was able to successfully upload the artifacts and set a metadata value that can be used by the triggering pipeline.

https://buildkite.com/elastic/independent-agent-release-staging/builds/52

dwhyrock:
Example of the Independent Agent Staging pipeline calling into the `elastic-agent-package` pipeline on this branch and the new code working successfully:

https://buildkite.com/elastic/independent-agent-release-staging/builds/69

dev-tools/mage/dockerbuilder.go
```diff
@@ -178,6 +178,10 @@ func (b *dockerBuilder) expandDockerfile(templatesDir string, data map[string]in
 
 func (b *dockerBuilder) dockerBuild() (string, error) {
 	tag := fmt.Sprintf("%s:%s", b.imageName, b.Version)
+	// For Independent Agent releases, replace the "+" with a "." since the "+" character
+	// currently isn't allowed in a tag in Docker
+	// E.g., 8.13.0+build202402191057 -> 8.13.0.build202402191057
+	tag = strings.Replace(tag, "+", ".", 1)
```
alexsapran, line 184: I have a question, if `b.Snapshot` is defined then the resulting tag would be 
`8.13.0.build202402191057-SNAPSHOT` is this the expected behavior for a snapshot build, or is this scenario never going to happen?

dev-tools/mage/manifest/manifest.go
```diff
@@ -61,7 +62,7 @@ func DownloadManifest(manifest string) (tools.Build, error) {
 	}
 	if mg.Verbose() {
 		log.Printf(">>>> Downloaded manifest %s", manifest)
-		log.Printf(">>>> Packaing version: %s, build_id: %s, manifest_version:%s", manifestResponse.Version, manifestResponse.BuildID, manifestResponse.ManifestVersion)
+		log.Printf(">>>> Packaging version: %s, build_id: %s, manifest_version:%s", manifestResponse.Version, manifestResponse.BuildID, manifestResponse.ManifestVersion)
```
alexsapran, line 65: 🥇

.buildkite/pipeline.elastic-agent-package.yml
```diff
@@ -126,3 +127,38 @@ steps:
...
+      ls -alR build
+      echo "--- Copy workaround for ironbank container filename"
+      .buildkite/scripts/steps/ironbank-cp-workaround.sh
+      echo "--- File listing after workaround"
+      ls -alR build
+      echo "+++ Checking artifact validity with release-manager collect dry run"
+      DRA_DRY_RUN="--dry-run"
+      export DRA_DRY_RUN
+      .buildkite/scripts/steps/dra-publish.sh
+      echo "+++ Call buildkite-agent upload to save artifacts for job that triggered us"
```
alexsapran, line 155: why we want to explicitly call the upload, we normally define the uploaded artifacts in the respected field in the `step` definition.

.buildkite/pipeline.elastic-agent-package.yml
```diff
@@ -126,3 +127,38 @@ steps:
...
+      buildkite-agent artifact upload "build/distributions/*"
+      echo "+++ Set job metadata if TRIGGER_JOB_ID is properly set"
+      if [[ -z "${TRIGGER_JOB_ID}" ]]; then
+        echo "TRIGGER_JOB_ID is not set, so not setting metadata"
+      else
+        echo "Setting metadata for job that trigger this one"
+        buildkite-agent meta-data set "triggered_build_id" "$BUILDKITE_BUILD_ID" --job $TRIGGER_JOB_ID
```
alexsapran, line 162: How are we going to use this information later, why do we want to save them?

pazone (approved):

alexsapran:
I see that this PR has a failing lint error, let's try and address those before merging. 

It might seem strange, but in this way, we strive towards a more consistent codebase and potentially find and fix issues.

.buildkite/pipeline.elastic-agent-package.yml
dwhyrock: By setting the `triggered_build_id` metadata field in the build that triggered this one, it passes is the Build ID of this `elastic-agent-package` build to the triggering build.  

The triggering build then uses this in a command to pull down the artifacts from this specific build.

Here's the command in the `unified-release` repo where we reference this: https://github.com/elastic/unified-release/blob/main/catalog/step-get-agent-bk-artifacts.go#L15

It's currently one of the easiest ways for a calling build to pull artifacts back from a triggered build in Buildkite.  Other ways involve obtaining a Buildkite API token that then has to be managed in vault, etc.  

Here's the example we took this from: https://forum.buildkite.community/t/how-to-download-artifacts-back-from-triggered-pipeline/3480/2

.buildkite/pipeline.elastic-agent-package.yml
dwhyrock: My thought was that we are verifying that the artifacts are all present with the above `--dry-run` of the `dra-publish.sh` script.  The normal DRA step here relies on the Release Manager container to do the proper uploads, so the expected files aren't currently defined in this repo.

We will have code for the Independent Release added to the `unified-release` repo that will verify that each expected file is present, which will happen after the artifacts are downloaded from this build.

If we also do the expected file check here, and if there's a change in which artifacts are expected, then we will have two places to change it instead of one.

With the Buildkite migration of the main Unified Release workflow, I believe there will be something that eventually replaces the Release Manager container to upload/verify the artifacts.  But that's more on David and Alpar's plate since I'm not on the Release Eng team.

dev-tools/mage/dockerbuilder.go
dwhyrock: We only plan on having a Staging build for the Independent Agent Release.  No Snapshot builds.

dwhyrock:
> I see that this PR has a failing lint error, let's try and address those before merging.

I will work on fixing the linting errors.

.buildkite/pipeline.elastic-agent-package.yml
alexsapran: oh ok, it make sense then. 

Can you put some comment about this reference, because if someone is going to touch this line and break it, the time to debug would be huge, especially if they don't have the full context.

.buildkite/pipeline.elastic-agent-package.yml
alexsapran: Why we need explicitly to call the `buildkite-agent artifact upload` is my point. 
We normally set it like this
```yaml
  - label: "Blah Blah"
    command: run_me.sh
    artifact_paths:
      - "build/distributions/*"
```

This way no matter your code inside the step, at the end, buildkite is going to try and upload those artifacts.

dev-tools/mage/dockerbuilder.go
```diff
@@ -46,15 +46,15 @@ func newDockerBuilder(spec PackageSpec) (*dockerBuilder, error) {
 
 func (b *dockerBuilder) Build() error {
 	if err := os.RemoveAll(b.buildDir); err != nil {
-		return errors.Wrapf(err, "failed to clean existing build directory %s", b.buildDir)
+		return fmt.Errorf("failed to clean existing build directory %s: %w", b.buildDir, err)
```
dwhyrock, line 49: It's probably worth someone to look over these changes from `pkg/errors` (due to linting complaints) to `fmt.Errorf` calls to make sure I did them correctly.

.buildkite/pipeline.elastic-agent-package.yml
dwhyrock: I will change it to use the `artifact_paths` entry instead of `buildkite-agent artifact upload`

dwhyrock:
I believe I've addressed all feedback and this is ready for another review.

pierrehilbert (approved):
Thx @dwhyrock for your work here!

Issue #4204, opened 2024-02-06
[Flaky Test]: TestStandaloneUpgrade/Upgrade_8.12.0-SNAPSHOT_to_8.13.0-SNAPSHOT_(privileged) – error installing package: failed to copy source directory... /opt/Elastic/Agent/elastic-agent: file exists


pchila:
### Failing test case

TestStandaloneUpgrade/Upgrade_8.12.0-SNAPSHOT_to_8.13.0-SNAPSHOT_(privileged)

### Error message

/opt/Elastic/Agent/elastic-agent: file exists during install, looks like a previous install was not cleaned or is still being cleaned

### Build

https://buildkite.com/elastic/elastic-agent/builds/6855#018d7f0a-a775-42a3-aa0b-0fb496da1a97

### OS

Linux

### Stacktrace and notes

```markdown
fixture.go:631: >> running binary with: [/tmp/TestStandaloneUpgradeUpgrade_8.12.0-SNAPSHOT_to_8.13.0-SNAPSHOT_(privileged)332107157/001/elastic-agent-8.12.0-SNAPSHOT-linux-arm64/elastic-agent install --force --non-interactive]
    upgrade_standalone_test.go:77: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_standalone_test.go:77
        	            				/home/ubuntu/agent/testing/integration/upgrade_standalone_test.go:52
        	Error:      	Received unexpected error:
        	            	failed to install start agent (err: error running agent install command: exit status 1) [output: Installing in non-interactive mode.
        	            	
[    ] Copying install files  [0s] 
                                   

[   =] Error copying files  [12s] Error: error installing package: failed to copy source directory (/tmp/TestStandaloneUpgradeUpgrade_8.12.0-SNAPSHOT_to_8.13.0-SNAPSHOT_(privileged)332107157/001/elastic-agent-8.12.0-SNAPSHOT-linux-arm64) to destination (/opt/Elastic/Agent): symlink data/elastic-agent-bf5180/elastic-agent /opt/Elastic/Agent/elastic-agent: file exists
        	            	For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.12/fleet-troubleshooting.html
        	            	]
        	Test:       	TestStandaloneUpgrade/Upgrade_8.12.0-SNAPSHOT_to_8.13.0-SNAPSHOT_(privileged)
[...]

PR #4225, opened 2024-02-08
Check for agent rogue processes in integration tests

pchila:
This commit adds assertions at the beginning and at the end of Fixture.Install() checking that no agent processes are running before installing or after uninstalling.

If any elastic-agent processes are detected, the test is marked as failed but it still runs to completion.

## What does this PR do?

This PR adds assertions for `elastic-agent` processes not running before installing or after uninstalling elastic agent in our integration tests. 

## Why is it important?

This is needed to highlight any test that leaves agent processes running (usually watchers) which may influence other tests runs and cause flakyness.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Relates #4204 

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/fixture_install.go
+++ b/pkg/testing/fixture_install.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/elastic/elastic-agent-libs/mapstr"
 	"github.com/elastic/elastic-agent-libs/opt"
-	eamprocess "github.com/elastic/elastic-agent-system-metrics/metric/system/process"
+	agentsystemprocess "github.com/elastic/elastic-agent-system-metrics/metric/system/process"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 	"github.com/elastic/elastic-agent/pkg/control/v2/client"
 	"github.com/elastic/elastic-agent/pkg/core/process"
@@ -129,7 +129,8 @@ func NewBool(value bool) *bool {
 func (f *Fixture) Install(ctx context.Context, installOpts *InstallOpts, opts ...process.CmdOption) ([]byte, error) {
 	f.t.Logf("[test %s] Inside fixture install function", f.t.Name())
 
-	assertNoRogueAgentProcesses(f.t, "there should be no running agent at beginning of Install()")
+	// check for running agents before installing
+	assert.Empty(f.t, getElasticAgentProcesses(f.t), "there should be no running agent at beginning of Install()")
 
 	installArgs := []string{"install"}
 	if installOpts == nil {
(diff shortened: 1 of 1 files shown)


leehinman (approved):

leehinman (approved):

pkg/testing/fixture_install.go
```diff
@@ -218,6 +231,79 @@ func (f *Fixture) Install(ctx context.Context, installOpts *InstallOpts, opts ..
...
+	for _, p := range pids {
+		processes = append(processes, mapProcess(p))
+	}
+
+	return processes
+}
+
+func assertNoRogueAgentProcesses(t *gotesting.T, message string) {
+	assert.Empty(t, getElasticAgentProcesses(t), message)
+}
```
AndersonQ, line 305: I think it's a bit unnecessary this function. But it's up to you

pkg/testing/fixture_install.go
```diff
@@ -16,10 +16,15 @@ import (
...
 	"strings"
+	gotesting "testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/elastic/elastic-agent-libs/mapstr"
+	"github.com/elastic/elastic-agent-libs/opt"
+	eamprocess "github.com/elastic/elastic-agent-system-metrics/metric/system/process"
```
AndersonQ, line 27: Why the `m` in `eaMprocess`? If it'd to abbreviate _elastic-agent-system-metrics_ I'd suggest to include the _system_ as well
```suggestion
	easmprocess "github.com/elastic/elastic-agent-system-metrics/metric/system/process"
```
However I think something like `systemprocess` or `agentsystemprocess` is better. For me, those abbreviations always look cryptic at first glance.

AndersonQ (approved):
I don't see why not fail the test as soon we detect there it another agent running, it'll lead to unexpected behaviour in the test anyway. Besides failing the test will block the PR anyway, so I'd say fail fast is better.

Also perhaps it could be more optimistic and do not fail the test when it fails getting the process information. I see it limits the information we have, but it isn't a failure in the test

pkg/testing/fixture_install.go
pchila: the funny effects of refactoring :smile: 
This was more complicated before extracting `getElasticAgentProcesses` function out, at this point it can be nested in the 2 places that still reference this...

pchila:
> I don't see why not fail the test as soon we detect there it another agent running, it'll lead to unexpected behaviour in the test anyway. Besides failing the test will block the PR anyway, so I'd say fail fast is better.
> 
> Also perhaps it could be more optimistic and do not fail the test when it fails getting the process information. I see it limits the information we have, but it isn't a failure in the test

Originally this was to try and correlate with logs I saw in test runs that did not have this guard.

I want to fail the test because it attracts attention (nobody ever looks at the green ones) but if a test is still leaking or adding processes between start and end I want to see it, hence I let it run to completion.

There was another part to this on the branch I cherry-picked this from where we try to clean up at the end of every test that installed agent after asserting no agent processes but that may be overkill for the moment

pkg/testing/fixture_install.go
pchila: Went with `agentsystemprocess` 👍

blakerouse (approved):

PR #4235, opened 2024-02-09
Fix action store -> state store migration tests

AndersonQ:
## What does this PR do?

Fixes the state store migration tests so they actually run and skip them for darwin.
It also changes the migration test to use a golden file when testing the migration from the old action store to the new state store.
Finally, it removes the `withFile` function as its miss usage caused the store migration tests to not actually run.

## Why is it important?

 - The tests need to fully run
 - Without using a golden file and creating the old action store on the fly, the test would not catch changes in the action model. If the model would to be changed, and it will as part of #3912, the test would still pass. However the actual migration would not work as expected.

## Checklist

## Related issues

- Relates #3912

## Questions to ask yourself

- ...


AndersonQ:
> I am a bit surprised by the 0% new code coverage... are the tests actually running?

the new code is test code, which isn't covered by tests. I don't think it's calculating the overall coverage.
Let me calculate it

AndersonQ:
> Moreover, I think that there's a bit of overlap with a bugfix I merged last year where I was moving unencrypted config to the new encrypted diskstores (as the agent was "forgetting" fleet config because `fleet.enc` was not found), so maybe we need to clean up a bit. Here's where the bugfix is located
> 
> https://github.com/elastic/elastic-agent/blob/bdefb2cd63ee6f97851fe1af06d2dcd0cda0076a/internal/pkg/agent/cmd/run.go#L194-L206

good point, it might be messing stuff. I added a TODO on my currently WIP PR: https://github.com/elastic/elastic-agent/pull/4253/commits/4926df0d884e07f7f4459bbd295e91dae33c4db6

pchila (approved):

Issue #4248, opened 2024-02-12
Agent/beats grpc comms over domain socket/named pipe


aleksmaus:
**Describe the enhancement:**

Elastic Agent uses localhost port for GRPC comms with beats.
Make it possible to configure the Agent to use domain socket/named pipe for GRPC with beats.

A core of work is already in done in the agent client lib https://github.com/elastic/elastic-agent-client/pull/91 and is tagged as v7.8.0.

Need to adjust the Agent to:
1. Make this configurable
2. Instantiate Agent GRPC server side over domain socket/named pipe
3. Propagate the server domain socket/named pipe address to the beats.

**Describe a specific use case for the enhancement or feature:**

There were few SDH related to the fact that the current GRPC comms are over IP socket and ask for ability to use domain socket/named pipe instead.

**What is the definition of done?**

Agent can communicate with beats over domain socket/named pipe.

Issue #3998, opened 2024-01-03
All communication with components should occur using unix sockets / named pipes


cmacknz:
Today the control protocol uses local TCP on port 6789 to communicate with components.

https://github.com/elastic/elastic-agent/blob/3a3d31c1b3e3af253bd37af4beb3dd979f3020cf/elastic-agent.yml#L165-L169

We regularly see issues where a firewall or iptables rules are configured to drop all traffic except for ports that have explicitly been whitelisted. The symptoms of this problem are not obvious in the agent logs and require uses to take extra steps to have Elastic Agent run at all.

We can solve this problem and a few others by moving away from local TCP to using Unix sockets or Windows named pipes to communicate with subprocesses.

This change should be made for:
1. The control protocol connection to supervised components. This change should be transparent to components as they can simply be passed the new address in the connection information message used to bootstrap communication with agent. The file ownership of the sockets must be set to match the privilege level of the agent.

https://github.com/elastic/elastic-agent-client/blob/f57f63489dbbce98522c174dae00158f895ddc84/elastic-agent-client.proto#L458-L463

2. The connection info server used to bootstrap endpoint security. In this case the connection info socket must be owed by root. This change must be coordinated with the endpoint security team.

https://github.com/elastic/elastic-agent/blob/3a3d31c1b3e3af253bd37af4beb3dd979f3020cf/pkg/component/runtime/conn_info_server.go#L31-L43

lucabelluccini:
This would be useful for users on Openshift who have [Red Hat OpenShift Data Foundation Managed Service](https://access.redhat.com/documentation/en-us/red_hat_openshift_data_foundation_managed_service/2022-q2/html/getting_started_with_red_hat_openshift_data_foundation_managed_service/con_ports-required-for-redhat-openshift-data-foundation-managed-service_deploy-models), as it uses the port 6789 for Ceph Monitor

cmacknz:
In progress in https://github.com/elastic/elastic-agent/pull/4249

rgarcia89:
I would love to see this being configurable as well.

rgarcia89:
@aleksmaus in which version will this be included?

cmacknz:
There is follow up work to do before we can enable this by default and recommend it to users: https://github.com/elastic/elastic-agent/issues/4899

It isn't enabled by default yet, Defend doesn't support it properly, and we haven't turned it on in our testing framework. All the plumbing through agent to enable it is done though. 

[...]

PR #4249, opened 2024-02-12
Agent/beats grpc comms over domain socket/named pipe

aleksmaus:
## What does this PR do?

Implements configurable GRPC allowing the agent to use domain socket/named pipe for comms with beats.

The core of the support for this is already merged in the agent client lib https://github.com/elastic/elastic-agent-client/pull/91 and is tagged as v7.8.0. Beats just work automagically since they already picked up this tag
https://github.com/elastic/beats/blob/main/go.mod#L72

Changes:
* Made the option configurable with the new```Local``` GRPC bool flag. Agent still uses IP socket by default
* Refactored the code that created the domain socket/named pipe socket, so it could be shared for GRPC comms
* Propagate the server domain socket/named pipe address to the beats.

I tested on all 3 platforms: darwin, linux and winderz. 
Will reach out to Endpoint team, since they likely would need to adjust comms on Endpoint side.

This is the first cut, can change based on the feedback.

Open for feedback/opinions on:
1. Agent on the new GRPC configuration option, currently it's a boolean flag.
2. The comms socket is derived from the control domain socket/named pipe path if GRPC is set to Local. Wanted to have it consistent in the same location as control domain socket path, with the same length limits fallback.

## Why is it important?

Addresses https://github.com/elastic/elastic-agent/issues/4248

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/4248
- Closes https://github.com/elastic/elastic-agent/issues/3998


internal/pkg/agent/application/application.go
```diff
@@ -111,10 +116,18 @@ func New(
 	}
 	monitor := monitoring.New(isMonitoringSupported, cfg.Settings.DownloadConfig.OS(), cfg.Settings.MonitoringConfig, agentInfo)
 
+	controlAddress := control.Address()
+	log.Debugf("Control address: %s", controlAddress)
+	address, err := deriveCommsAddress(controlAddress, cfg.Settings.GRPC)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to dervice comms GRPC: %w", err)
```
intxgo, line 123:
```suggestion
		return nil, nil, nil, fmt.Errorf("failed to derive comms GRPC: %w", err)
```

aleksmaus:
Talked to @jrmolin, one issue left is the connection info discovery is still as defined in the endpoint config an IP port
https://github.com/elastic/elastic-agent/blob/main/specs/endpoint-security.spec.yml#L24

We could potentially allow some base path in the endpoint spect for domain socket/named pipe for connection info API.

Thoughts?

cmacknz:
> The question now, that the endpoint doesn’t know what mode the agent is started in, so it would have to check both IP and domain socket/pipe for the connection info server.

It seems simpler to just align this change in both endpoint and agent rather than introduce complexity to deal with this that we would remove as soon as endpoint supports domain sockets. The endpoint version is always aligned with the agent version so we ideally wouldn't have to release a version where endpoint doesn't know what to expect.

> Talked to @jrmolin, one issue left is the connection info discovery is still as defined in the endpoint config an IP port
https://github.com/elastic/elastic-agent/blob/main/specs/endpoint-security.spec.yml#L24

> We could potentially allow some base path in the endpoint spect for domain socket/named pipe for connection info API.

We could use a fixed base path for the unix socket connection instead of the a fixed port. The socket/pipe would be owned by root/admin so it is already more secure than an arbitrary port.

aleksmaus:
> It seems simpler to just align this change in both endpoint and agent rather than introduce complexity to deal with this that we would remove as soon as endpoint supports domain sockets. The endpoint version is always aligned with the agent version so we ideally wouldn't have to release a version where endpoint doesn't know what to expect.

I made this mode configurable, there might be the cases where you might want to use IP socket, I suspect, like possibly running agent and beats in the separate pods. So the Endpoint would know in advance if IP or domain socket is used for RPC.

cmacknz:
Looking at the code we can use the `agent.grpc.local` as a feature flag initially that defaults to false and is only flipped to true once all of the inputs are confirmed to work with unix sockets/named pipes. 

The Go based inputs all need to be updated to https://github.com/elastic/elastic-agent-client/releases/tag/v7.8.0 first which shouldn't be that hard to coordinate, endpoint is slightly more complicated.

Thinking more about this it would actually be best if endpoint could handle either a TCP port or a domain socket/pipe. This way we can keep the configuration and if we hit an unexpected problem we have the ability to easily go back.

So I'd say the path forward is:
1. We define how we are going to specify the domain socket location for endpoint to connect to.
2. Endpoint looks for this location, and if it doesn't exist it falls back to checking the TCP port. Maybe it just looks for both at the same time knowing agent will only support one of them.
3. We confirm all inputs are updated to at least https://github.com/elastic/elastic-agent-client/releases/tag/v7.8.0 and support domain sockets/pipes.
4. We can flip the agent.grpc.local flag to true by default.

cmacknz:
> I made this mode configurable, there might be the cases where you might want to use IP socket, I suspect, like possibly running agent and beats in the separate pods. So the Endpoint would know in advance if IP or domain socket is used for RPC.

Yes it is also a way to run the agent twice on the same machine since you can more easily change the port than the unix socket location.

blakerouse:
@aleksmaus @cmacknz 

Endpoint will not install when the Elastic Agent is not installed in the default location: https://github.com/elastic/elastic-agent/blob/main/specs/endpoint-security.spec.yml#L21 This is because Endpoint protects the Elastic Agent directory and its hard coded to that directory.

In that case the Elastic Agent should just create a domain unix socket inside of its default installation directory for Endpoint to connect to and get the connection information. Windows is even easier it should just use a defined npipe.

I also agree that Endpoint should check the unix socket/npipe first and then fallback checking for the network port for the information.

I also prefer that this just become the default.

intxgo:
Taking into account all the basic hardening we've added to Endpoint recently (policy signing, Tamper Protection), we ended with a flow:
1. there's a clean machine
2. user install Agent
3. Agent installs Endpoint (with no config)
4. Endpoint tries to get it's first config from Agent
5. Endpoint is "hardened" and from that point it's hard to mock with it...

so for this question "how to tell Endpoint how it should talk to Agent" it's a matter of the bootstrap phase 1-4. Maybe the npipe/socket should be a plain install parameter? I assume once Endpoint is configured with a given comms method it won't change.

aleksmaus:
> I assume once Endpoint is configured with a given comms method it won't change.

This is configurable, so it's possible to change, don't know why would anybody do that, but possible.

intxgo:
> This is configurable, so it's possible to change, don't know why would anybody do that, but possible.

ok. Then it's not so straightforward. Maybe we could define more constrains, like it's possible to change the comms method but it will require Endpoint re-install? The latter would ensure that sufficient prerequisites are met to actually be able to uninstall Endpoint.

cmacknz:
> ok. Then it's not so straightforward. Maybe we could define more constrains, like it's possible to change the comms method but it will require Endpoint re-install? 

Changing the comms method will require every component to restart at minimum I think. I don't love the complexity this brings however I also don't like the idea of changing this without giving users who experience unexpected problems with it an escape hatch. We could require an agent restart but there's no way to initiate this from Fleet right now. Perhaps the agent automatically re-execs itself whenever this changes in the configuration.

One thing that isn't touched in the PR yet is that the bootstrap process for endpoint today also involves local TCP. We should change this to also use a unix/named pipe at a fixed location instead of a fixed TCP port so that you have to be root to connect to it.

https://github.com/elastic/elastic-agent/blob/d075097c676355752c8df5ee06a0c4fa79280369/pkg/component/runtime/conn_info_server.go#L43-L49

The process today is:
1. Endpoint connects to this connection information server. Agent responds with the gRPC address and the TLS certificates for it.
2. Endpoint then connects to the gRPC server and is given its initial configuration.

We want both the bootstrapping and the control protocol gRPC address to be unix/named pipes.

As Blake suggests for the bootstrapping we can just used a fixed location for the bootstrapping pipe. That can then communicate the address of the gRPC pipe.

We could also try to eliminate the need to bootstrap endpoint entirely since we can use a fixed pipe address for gRPC, but we'd still need a way to give endpoint the certs for mTLS.

aleksmaus:
> We want both the bootstrapping and the control protocol gRPC address to be unix/named pipes.

Yes, that's one on the remaining things for this feauture to complete and is on my TODO list.

aleksmaus:
Implemented domain socket/named pipe support for connection info endpoint.
The domain socket path is in the Agent install directory, per @blakerouse.
Please review and let me know if there are any cases I didn't cover or this approach might not work.

The socket path with my tests on linux is ```/opt/Elastic/Agent/.eaci.sock```. Let me know if you prefer a different name/path. Tried to come up with something short and something that is less likely to collide with any file in the agent install dir.
Screenshot 2024-03-10 at 8 00 14 PM


Will test more tomorrow (specifically on Windows) and reach out to Endpoint team for integration tests.

internal/pkg/agent/application/application.go
```diff
@@ -111,10 +116,18 @@ func New(
...
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to derive comms GRPC: %w", err)
+	}
+	log.Debugf("Comms GRPS address: %s", address)
+
 	runtime, err := runtime.NewManager(
 		log,
 		baseLogger,
-		cfg.Settings.GRPC.String(),
+		address,
```
blakerouse, line 130: I think it would be better to just pass the whole `cfg.Settings.GRPC` into the `runtime.NewManager` and allow the manager to create the correct address to communicate. Don't like it outside of the manager, because then different callers could come to different conclusions.

internal/pkg/agent/application/application.go
```diff
@@ -111,10 +116,18 @@ func New(
 	}
 	monitor := monitoring.New(isMonitoringSupported, cfg.Settings.DownloadConfig.OS(), cfg.Settings.MonitoringConfig, agentInfo)
 
+	controlAddress := control.Address()
+	log.Debugf("Control address: %s", controlAddress)
```
blakerouse, line 120: We already have a info level message that logs the control address.

https://github.com/elastic/elastic-agent/blob/main/pkg/control/v2/server/server.go#L88

internal/pkg/agent/application/application.go
```diff
@@ -111,10 +116,18 @@ func New(
...
+	address, err := deriveCommsAddress(controlAddress, cfg.Settings.GRPC)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to derive comms GRPC: %w", err)
+	}
+	log.Debugf("Comms GRPS address: %s", address)
```
blakerouse, line 125: I would prefer this to be a nice looking message that is at the info level. 

https://github.com/elastic/elastic-agent/blob/main/pkg/control/v2/server/server.go#L88

internal/pkg/agent/configuration/grpc.go
```diff
@@ -4,14 +4,17 @@
...
+	"fmt"
+)
 
 // GRPCConfig is a configuration of GRPC server.
 type GRPCConfig struct {
 	Address                 string `config:"address"`
 	Port                    uint16 `config:"port"`
 	MaxMsgSize              int    `config:"max_message_size"`
 	CheckinChunkingDisabled bool   `config:"checkin_chunking_disabled"`
+	Local                   bool   `config:"local"` // Enables RPC via domain socker/named pipe
```
blakerouse, line 17: This really is an internal functionality and I would prefer to see it enabled by default. Would it be possible to just change the port default to a `int32` with `-1` as the default. That would allow anyone that has configured a different port to be used how it works today, and if its `-1`. Possibly the same with the address field. If its empty and its -1 for the port, it used the unix/npipe otherwise it fallsback to how it was configured.

If we hit an issue the solution is to set the `address` to `localhost` and `port` to `6789`.

pkg/component/runtime/conn_info_server_nonwin_test.go
```diff
@@ -0,0 +1,21 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
blakerouse, line 1: Can you name this file `conn_info_server_unix_test.go`? We normally use `unix` in the filename when its for non-Windows, and being that this specifically use `net.Dial("unix", ...)` its more correct.

pkg/component/runtime/conn_info_server_win_test.go
```diff
@@ -0,0 +1,19 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
blakerouse, line 1: Please rename to `conn_info_server_windows_test.go`.

blakerouse (changes requested):

internal/pkg/agent/configuration/grpc.go
aleksmaus: Are we absolutely positive that we want to break the default behavior?

I thought that it would be safer to keep default as IP socket: ```address``` to ```localhost``` and ```port``` to ```6789```
and the let the user to switch to domain socket/named pipe.

Changing ```Port``` to ```int16``` will limit the port number that users can use. Could potentially change it to ```int32```.

Introducing the separate/additional attribute seemed cleaner. 

Let me know if you still want to have -1 ```Port``` as default and remove ```Local``` attribute.

aleksmaus:
Addressed code review comments from @blakerouse. Left one reply, in order to confirm the change request regarding default configuration, see above.

internal/pkg/agent/configuration/grpc.go
blakerouse: I would prefer to see this change and become the default due to the SDH we get with using network based communication. I think being that this communication is internal its not a breaking change.

@cmacknz What do you think?

cmacknz: Agree, we need to stop using localhost by default. The communication method is internal to the agent so I do not consider it a breaking change.

I do think it is important that we be able to switch from unix sockets/pipes back to localhost via configuration. This is how we can make the change safely and protect ourselves from unknown unknowns.

aleksmaus: Sounds good. Thanks for confirmation, will update PR then to use -1 for the port default (pipes rpc)  as suggested.

aleksmaus: Addressed this requirement now.

pkg/component/runtime/service.go
```diff
@@ -275,6 +281,26 @@ func (s *serviceRuntime) Run(ctx context.Context, comm Communicator) (err error)
...
+		u := url.URL{}
+		u.Path = "/"
+
+		if os == "windows" {
+			u.Scheme = "npipe"
+			return u.JoinPath("/", elasticAgentConnInfoSocket).String()
+		}
+
+		u.Scheme = "unix"
+		return u.JoinPath(paths.InstallPath(paths.DefaultBasePath), elasticAgentConnInfoSocket).String()
```
blakerouse, line 298: This is not unique from a standpoint of the Elastic Agent in the future supporting another service type. The `.eaci.sock` should be set in the component specification, not something that should be hard-coded into the Elastic Agent.

blakerouse (changes requested):
This looks really close. Only issue is the hard coding of the socket name for service runtimes. That should come from the component specification.

Obviously this cannot land until Endpoint is working as well with sockets.

pkg/component/runtime/service.go
aleksmaus: Makes sense, will update

aleksmaus:
Addressed code review blocker, made the connection info socket name configurable through the service spec, consistent with connection info port ```"cport"```.

pkg/component/runtime/service.go
```diff
@@ -284,18 +284,22 @@ func (s *serviceRuntime) Run(ctx context.Context, comm Communicator) (err error)
 // .eaci.sock == Elastic Agent Connection Info socket
 const elasticAgentConnInfoSocket = ".eaci.sock"
 
-func getConnInfoServerAddress(os string, isLocal bool, port int) string {
+func getConnInfoServerAddress(os string, isLocal bool, port int, socket string) string {
 	if isLocal {
+		if socket == "" {
```
blakerouse, line 289: This should error, and not start. The field should be required for a service. We don't have a default when `CPort` is missing.

aleksmaus: Updated

gabriellandau:
The C++ grpc library that Endpoint uses doesn't support domain sockets or named pipes.  There's a [longstanding FR](https://github.com/grpc/grpc/issues/13447), but it's not implemented.  We don't think we can land full gRPC-over-docket/npipe in 8.14, but we do think we think we can land the updated bootstrap.

Would it be possible for Agent to support the new comms for bootstrap, while maintaining localhost gRPC TCP for Endpoint in 8.14?

aleksmaus:
> Would it be possible for Agent to support the new comms for bootstrap, while maintaining localhost gRPC TCP for Endpoint in 8.14?

This would require some agent code rework, don't know the scope of changes yet, need to dig more. 
I assume it could be possible to have two "GRPC servers" in 8.14: TCP one for Endpoint only, and the local domain sockets/names pipes for the rest of the agent/beats comms and "connection info" endpoint. 
Kind of odd but that's one way to work around the current limitation on Endpoint side.
Thoughts? @blakerouse @cmacknz

aleksmaus:
Added the commit that uses TCP gRPC for comms and forces domain socket/named pipe for connection info server.

@blakerouse I could not rely on the gRPC Port config (-1) in this case, since we were asked to keep gRPC on TCP socket. And only use domain socket for connection info server. With this change the connection info server is always forced to use domain socket. 
In order to make it configurable I would have to introduce a new configuration option for connection info server only. And then it could become confusing when we just start using Port -1 for setting everything to use domain sockets.
@blakerouse @cmacknz 
Let me know if you still need to make connection info server configurable (able to switch between domain socket and TCP socket) for 8.14.

gabriellandau:
> I also think the security benefits of this change are not as strong if we add a configuration flag to undo it. So for now I'd say we don't need a configuration flag for this as long as we can align it with endpoint merging the change at the same time.

I agree.

ycombinator:
👋 @aleksmaus, just wondering what your plans are for this PR, given the last activity on it was just over a month ago?  Do you want to keep iterating on it or should we close it out?

aleksmaus:
This PR is done and was waiting on Endpoint PR merge in order pass the integration tests. The Endpoint was merged today
https://github.com/elastic/endpoint-dev/pull/14386#event-13077664540
but missed the build window. So I think it should be available by next week in theory, then we can kick off the integration tests again, pass the CI validation and merge.

PR #4256, opened 2024-02-14
Add integration test to update an agent with another built from the same commit

### LINKED ISSUE #2780: Test that the currently installed Elastic Agent version can upgrade to itself

pchila:
## What does this PR do?

This PR will add integration tests checking that:
- we cannot upgrade an agent with the exact same version
- we can modify the version of the agent package and the upgrade using this modified version with an agent built from the same commit

## Why is it important?

The new integration tests prove that #4193 makes it possible to have 2 agent installs from the same hash at the same time on disk (we have this situation when performing an upgrade before the end of the grace period).

It also proves that we can use the same package built from the commit to test certain upgrade flows in CI.

## Checklist

## Author's Checklist

- [ ]

The new tests can be run using
```shell
GOTEST_FLAGS="-test.run ^TestStandaloneUpgradeSameCommit$" TEST_PLATFORMS="linux/amd64,windows/amd64"  SNAPSHOT=true TEST_PLATFORMS="linux/amd64/ubuntu" mage integration:test
```

## Related issues

- Closes #2780

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/upgrade_standalone_same_commit_test.go
+++ b/testing/integration/upgrade_standalone_same_commit_test.go
@@ -1,3 +1,9 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build integration
+
 package integration
 
 import (
@@ -22,6 +28,7 @@ import (
 
 	"github.com/elastic/elastic-agent/dev-tools/mage"
 	v1 "github.com/elastic/elastic-agent/pkg/api/v1"
+	atesting "github.com/elastic/elastic-agent/pkg/testing"
 	"github.com/elastic/elastic-agent/pkg/testing/define"
 	"github.com/elastic/elastic-agent/pkg/testing/tools/testcontext"
 	"github.com/elastic/elastic-agent/pkg/version"
@@ -36,7 +43,7 @@ func TestStandaloneUpgradeSameCommit(t *testing.T) {
 		Sudo:  true,  // requires Agent installation
 	})
 
-	// we upgrade unto the same version
+	// parse the version we are testing
 	currentVersion, err := version.ParseVersion(define.Version())
 	require.NoError(t, err)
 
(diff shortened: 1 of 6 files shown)


testing/integration/upgrade_standalone_same_commit_test.go
```diff
@@ -0,0 +1,381 @@
...
+		t.Skipf("Minimum version for running this test is %q, current version: %q", *upgradetest.Version_8_13_0_SNAPSHOT, currentVersion)
+	}
+
+	unprivilegedAvailable := true
+	if runtime.GOOS != define.Linux {
+		// only available on Linux at the moment
+		unprivilegedAvailable = false
+	}
+	// This is probably redundant: see the skip statement above
+	if unprivilegedAvailable && (currentVersion.Less(*upgradetest.Version_8_13_0) || currentVersion.Less(*upgradetest.Version_8_13_0)) {
```
cmacknz, line 61:
```suggestion
	if unprivilegedAvailable && currentVersion.Less(*upgradetest.Version_8_13_0) {
```

Both sides of this `||` are the same expression, since the start version and end version here are the same I suspect you can remove one of them.

testing/integration/upgrade_standalone_same_commit_test.go
```diff
@@ -0,0 +1,381 @@
...
+		// I wish I could just pass the location of the package on disk to the whole upgrade tests/fixture/fetcher code
+		// but I would have to break too much code for that, when in Rome... add more code on top of inflexible code
+		repackagedLocalFetcher := atesting.LocalFetcher(newPackageContainingDir)
+
+		endFixture, err := atesting.NewFixture(t, versionForFixture.String(), atesting.WithFetcher(repackagedLocalFetcher))
+		require.NoErrorf(t, err, "error creating end fixture with LocalArtifactFetcher with dir %q", newPackageContainingDir)
+
+		err = upgradetest.PerformUpgrade(ctx, startFixture, endFixture, t,
+			upgradetest.WithUnprivileged(unprivilegedAvailable),
+			upgradetest.WithDisableHashCheck(true),
```
cmacknz, line 160: Rather than disabling the check, could you invert it? That is require the hash to be the same?

That would enforce that the commit still reads the same after all of the package hacking that goes on as a condition of the test.

pchila: This is to disable the check before starting the upgrade in our integration tests: it's the "fix" we have to avoid hash collisions then upgrading from a snapshot to another snapshot of the same version (it has been introduced by @rdner not too long ago)

The hash after upgrade is checked anyways using the endFixture which in our case has obviously the same hash

testing/integration/upgrade_standalone_same_commit_test.go
pchila: Oups I forgot to remove it after I changed the second version to 8.13.0 (we have a similar check for the privileged/unprivileged switch that you can find in other tests).
Will remove.

cmacknz (approved):

PR #4264, opened 2024-02-14
[Unprivileged] Windows enablement

### LINKED ISSUE #3868: [Unprivileged] Enable `--unprivileged` on Windows
**Describe the enhancement:**

Current support for `--unprivileged` is isolated to Linux, this needs to be expanded to support Windows.

To allow Elastic Agent to run on Windows without requiring Administrator privileges.

When `install --unprivileged` installs and Elastic Agent works running on Windows in this mode.

blakerouse:
## What does this PR do?

Enables the installation mode of `--unprivileged` for Windows. During installation the `elastic-agent-user` user and `elastic-agent` group is created, permissions are set correctly on the copied files, service is created running as the `elastic-agent-user`. While running in unprivileged mode the daemon creates the vault with the correct permissions and gives users in the `elastic-agent` group the ability to communicate with the running Elastic Agent daemon.

## Why is it important?

Allowing users to install the Elastic Agent with less permissions.

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (cannot test with unit tests)
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~ (skipped as we don't want to expose this feature until ready)

## Author's Checklist

- [x] `elastic-agent-user` is created
- [x] `elastic-agent` group is created
- [x] `elastic-agent-user` is added to `elastic-agent` group
- [x] file permission/owner is set for `elastic-agent-user` and `elastic-agent` group
- [x] `elastic-agent` runs correctly

```
$ .\elastic-agent.exe install -f --unprivileged
```

Add your local user to `elastic-agent` group. Be sure to re-open command prompt. Run `C:\Program Files\Elastic\Agent\elastic-agent.exe status`.

## Related issues

- Closes #3868


control_v2.proto
```diff
@@ -167,6 +167,8 @@ message StateAgentInfo {
   bool snapshot = 5;
   // Current running PID.
   int32 pid = 6;
+  // True when running as privileged.
+  bool privileged = 7;
```
pchila, line 171: To keep nomenclature simple maybe we can flip this boolean (it would be coherent with the feature being introduced, the command line flags and the default value corresponds to the default mode of operation for agent. Something like
```suggestion
  // True when running unprivileged.
  bool unprivileged = 7;
```
The same would be for the agent_info down below privileged -> unprivileged
It doesn't change the behavior and using a privileged flag would work just as well, just trying to reduce cognitive load and possible confusion :sweat_smile:

blakerouse: Thats true, I will switch the logic.

blakerouse: I switched the logic in the latest commit.

internal/pkg/agent/install/svc.go
```diff
@@ -57,25 +58,30 @@ func newService(topPath string, opt ...serviceOpt) (service.Service, error) {
...
+		// Linux (systemd) always restart on failure
+		"Restart": "always",
+
+		// Windows setup restart on failure
+		"OnFailure":              "restart",
+		"OnFailureDelayDuration": "15s", // Matches the value used by endpoint-security.
+		"OnFailureResetPeriod":   10,
+	}
+	if opts.Password != "" {
+		option["Password"] = opts.Password
```
leehinman, line 74: I'm pretty sure this password is stored encrypted in the Registry.  We need to document how the user would go about changing it.  It's not uncommon for password policies to require changing static passwords like this on a regular basis.

leehinman (changes requested):
just documentation requests.

pkg/utils/perm_windows.go
```diff
@@ -21,10 +27,14 @@ type FileOwner struct {
 
 // CurrentFileOwner returns the executing UID and GID of the current process.
 func CurrentFileOwner() FileOwner {
-	// TODO(blakerouse): Make this return the current user and group on Windows.
+	u, err := user.Current()
+	if err != nil {
+		// should not fail; if it does then there is a big problem
+		panic(err)
```
pchila, line 33: I understand why you used `panic()` here as such an error would not be easily recoverable but I would still prefer to give the chance to the callers to explicitly handle an error condition by returning `error`.
If we want to keep the `panic()` for windows we should at least add this to the documentation of the function for all the OSes (as some people developing on a platform different from windows should be aware that this function can panic on windows)

testing/integration/install_unix_test.go
```diff
@@ -17,11 +18,12 @@ import (
...
+	atesting "github.com/elastic/elastic-agent/pkg/testing"
 )
 
-func checkPlatformUnprivileged(t *testing.T, topPath string) {
+func checkPlatformUnprivileged(t *testing.T, _ *atesting.Fixture, topPath string) {
 	// Check that the elastic-agent user/group exist.
-	uid, err := install.FindUID("elastic-agent")
+	uid, err := install.FindUID("elastic-agent-user")
 	require.NoError(t, err)
 	gid, err := install.FindGID("elastic-agent")
```
pchila, line 28: Nit: maybe we should use the constants for user and group here, just to be on the safe side

pchila (approved):
Code looks good, no major blockers.
Left a few comments that are mostly for readability/documentation, those don't fundamentally change the implementation

internal/pkg/agent/install/install.go
```diff
@@ -28,7 +29,7 @@ import (
 const (
 	darwin = "darwin"
 
-	elasticUsername  = "elastic-agent"
+	elasticUsername  = "elastic-agent-user"
 	elasticGroupName = "elastic-agent"
```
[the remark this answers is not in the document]
blakerouse: It was discussed and agreed to have it as it is in the PR.

internal/pkg/agent/cmd/reexec_windows.go
```diff
@@ -34,43 +34,94 @@ func newReExecWindowsCommand(_ []string, streams *cli.IOStreams) *cobra.Command
...
+	for {
+		ready, err := ensureAnotherProcess(serviceName, servicePid)
+		if err == nil && ready {
+			// all done
+			return
+		}
+		if err != nil {
+			fmt.Fprintf(writer, "%s", err)
+		}
+		<-time.After(300 * time.Millisecond)
```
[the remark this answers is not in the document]
blakerouse: Updated.

internal/pkg/agent/cmd/reexec_windows.go
```diff
@@ -34,43 +34,94 @@ func newReExecWindowsCommand(_ []string, streams *cli.IOStreams) *cobra.Command
...
+	}
+
+	if status.State == svc.Stopped {
+		// fully stopped
+		err = startService(serviceName)
+		return false, err
+	}
+
+	// not stopped and not running as a different PID, just wait
+	return false, nil
```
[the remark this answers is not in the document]
blakerouse: I prefer the if statement, to mean its easier to follow.

internal/pkg/agent/install/install_windows.go
```diff
@@ -45,6 +49,56 @@ func postInstall(topPath string) error {
...
+		// no user, running with LOCAL SYSTEM (do nothing)
+		return nil
+	}
+
+	securityDescriptor, err := windows.SecurityDescriptorFromString(
+		"D:(A;;GA;;;SY)" + // SDDL_LOCAL_SYSTEM -> SDDL_GENERIC_ALL
+			"(A;;GA;;;BA)" + // SDDL_BUILTIN_ADMINISTRATORS -> SDDL_GENERIC_ALL
+			"(A;;GR;;;WD)" + // SDDL_EVERYONE -> SDDL_GENERIC_READ
+			"(A;;GRGX;;;NS)" + // SDDL_NETWORK_SERVICE -> SDDL_GENERIC_READ|SDDL_GENERIC_EXECUTE
+			fmt.Sprintf("(A;;GA;;;%s)", ownership.UID), // Ownership UID -> SDDL_GENERIC_ALL
```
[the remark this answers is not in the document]
blakerouse: Added

internal/pkg/agent/install/user_windows.go
```diff
@@ -6,29 +6,246 @@
...
+		return "", fmt.Errorf("invalid SID type for user %s; should be user account type, not %d", name, t)
+	}
+	return sid.String(), nil
 }
 
 // CreateUser creates a user on the machine.
-func CreateUser(name string, gid string) (string, error) {
-	return "", errors.New("not implemented")
+//
+// User is created with about interactive rights, no logon rights, and only service rights.
```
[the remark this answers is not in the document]
blakerouse: Updated.

pkg/utils/perm_windows.go
blakerouse: Added comment.

testing/integration/install_unix_test.go
blakerouse: Switched to constants.

internal/pkg/agent/install/svc.go
blakerouse: I don't know if that is true in this case. The user cannot login at all, it can only be used as a service and the password is setup to never expire.

internal/pkg/agent/cmd/reexec_windows.go
```diff
@@ -34,43 +34,94 @@ func newReExecWindowsCommand(_ []string, streams *cli.IOStreams) *cobra.Command
...
-func reExec(serviceName string, servicePid int) error {
+func reExec(serviceName string, servicePid int, writer io.Writer) {
+	for {
+		ready, err := ensureAnotherProcess(serviceName, servicePid)
+		if err == nil && ready {
+			// all done
+			return
+		}
+		if err != nil {
+			fmt.Fprintf(writer, "%s", err)
```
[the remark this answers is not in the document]
blakerouse: More context was added.

internal/pkg/agent/application/info/svc_windows.go
```diff
@@ -7,44 +7,65 @@
...
-		return false
+		return false, fmt.Errorf("allocate sid error: %w", err)
 	}
-	defer windows.FreeSid(serviceSid)
+	defer func() {
+		_ = windows.FreeSid(sid)
+	}()
 
-	t, err := windows.OpenCurrentProcessToken()
+	token := windows.Token(0)
```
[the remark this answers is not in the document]
blakerouse: `windows.Token(0)` is used because its passed into `checkTokenMembership` that when the token is zero that check uses the current process token. It is a simpler way of saying this processes current token, without the need to grab a handle that needs to be closed for the current process.

What do you want documented here? That `windows.Token(0)` means the current process token?

internal/pkg/agent/install/svc.go
leehinman: > I don't know if that is true in this case.

See things like: https://www.stigviewer.com/stig/windows_server_2019/2020-06-15/finding/V-93209

internal/pkg/agent/application/info/svc_windows.go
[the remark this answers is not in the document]
leehinman: > What do you want documented here? That `windows.Token(0)` means the current process token?

Probably a few links to MSDN documentation, and just a comment that `windows.Token(0)` is current process.  I'm just thinking in 3 years if we have a bug around this code in some new Azure environment it is nice to know what we were thinking and hopefully pointers to full documentation which might give clues if it has been deprecated etc.

internal/pkg/agent/install/svc.go
blakerouse: That is interesting finding, but even in this case the installing administrator doesn't even know the password. Being the finding is speaking about an administrator knowing the password, and in this case the password is completely unknown.

We can definitely add documentation about it for unprivileged installation on Windows. If they choose to change the user account password they will also have to update the password for the service.

internal/pkg/agent/application/info/svc_windows.go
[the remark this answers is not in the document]
blakerouse: Added some more comments and a link for the `windows.Token(0)`.

leehinman (changes requested):
need to use `crypto/rand` not `math/rand` for generating random passwords.

Other than that, LGTM.

blakerouse:
@leehinman Okay I have updated to `crypto/rand`.

leehinman (approved):

pierrehilbert:
What can be tested is tested and additional tests will require this to be merged first.
I'm forcing the merge even if the SonarQube requirement is red.

Issue #4275, opened 2024-02-19
[Flaky Test]: TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing tried to download an incorrect artifact


rdner:
### Failing test case

TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing

### Error message

404 when downloading an artifact

### Build

https://buildkite.com/elastic/elastic-agent/builds/7244#018db45f-064c-4ba3-a4eb-eb93b04613f1

### OS

Linux, Mac, Windows

### Stacktrace and notes

```markdown
upgrade_gpg_test.go:143: 
        	Error Trace:	C:/Users/windows/agent/testing/integration/upgrade_gpg_test.go:143
        	Error:      	Received unexpected error:
        	            	failed to start agent upgrade to version "8.12.2": exit status 1
        	            	Error: Failed trigger upgrade of daemon: failed download of agent binary: unable to download package: 3 errors occurred:
        	            		* package 'C:\Program Files\Elastic\Agent\data\elastic-agent-8.13.0-SNAPSHOT-f23c27\downloads\elastic-agent-8.12.2-windows-x86_64.zip' not found: open C:\Program Files\Elastic\Agent\data\elastic-agent-8.13.0-SNAPSHOT-f23c27\downloads\elastic-agent-8.12.2-windows-x86_64.zip: The system cannot find the file specified.
        	            		* call to 'https://snapshots.elastic.co/8.13.0-772867d3/downloads/beats/elastic-agent/elastic-agent-8.12.2-windows-x86_64.zip' returned unsuccessful status code: 404
        	            		* call to 'https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.12.2-windows-x86_64.zip' returned unsuccessful status code: 404
        	            	
        	            	
        	            	For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.13/fleet-troubleshooting.html
        	Test:       	TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing
        	Messages:   	perform upgrade failed
    fixture_install.go:179: [test TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing] Inside fixture cleanup function
    fixture_install.go:195: collecting diagnostics; test failed
```

not the artifact URL is incorrect and contains 2 different versions in it.
```

rdner:
Looks like the fixture was setup properly:

```json
{
[...]

PR #4276, opened 2024-02-19
Check if non-snapshot version is released before using for testing

rdner:
## What does this PR do?

Currently, we don't check whether a version without the `-SNAPSHOT` suffix is actually released and available on our public CDN.

The version is present on the artifact API once the first BC is built, however, this does not mean the version is published on the CDN.

This leads to failing upgrade attempts by the agent in our integration tests.

## Why is it important?

Our integration tests are not passing.

## Checklist

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/4275

### DIFF
--- a/pkg/testing/tools/artifacts_api.go
+++ b/pkg/testing/tools/artifacts_api.go
@@ -12,8 +12,10 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"runtime"
 	"sort"
 
+	"github.com/elastic/elastic-agent/pkg/testing"
 	"github.com/elastic/elastic-agent/pkg/version"
 )
 
@@ -26,6 +28,7 @@ const (
 	// artifactAPIV1SearchVersionPackage = "v1/search/%s/%s"
 
 	artifactElasticAgentProject = "elastic-agent-package"
+	artifactReleaseCDN          = "https://artifacts.elastic.co/downloads/beats/elastic-agent"
 )
 
 var (
@@ -119,6 +122,10 @@ func WithUrl(url string) ArtifactAPIClientOpt {
 	return func(aac *ArtifactAPIClient) { aac.url = url }
 }
 
+func WithCDNUrl(url string) ArtifactAPIClientOpt {
+	return func(aac *ArtifactAPIClient) { aac.cdnURL = url }
+}
+
 func WithHttpClient(client httpDoer) ArtifactAPIClientOpt {
 	return func(aac *ArtifactAPIClient) { aac.c = client }
 }
@@ -127,15 +134,17 @@ func WithHttpClient(client httpDoer) ArtifactAPIClientOpt {
 // More information about the API can be found at https://artifacts-api.elastic.co/v1
 // which will print a list of available operations
 type ArtifactAPIClient struct {
-	c   httpDoer
-	url string
+	c      httpDoer
+	url    string
+	cdnURL string
 }
 
 // NewArtifactAPIClient creates a new Artifact API client
 func NewArtifactAPIClient(opts ...ArtifactAPIClientOpt) *ArtifactAPIClient {
 	c := &ArtifactAPIClient{
-		url: defaultArtifactAPIURL,
-		c:   new(http.Client),
+		url:    defaultArtifactAPIURL,
+		cdnURL: artifactReleaseCDN,
+		c:      new(http.Client),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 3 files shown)


pkg/testing/tools/artifacts_api.go
```diff
@@ -127,15 +134,17 @@ func WithHttpClient(client httpDoer) ArtifactAPIClientOpt {
...
 }
 
 // NewArtifactAPIClient creates a new Artifact API client
 func NewArtifactAPIClient(opts ...ArtifactAPIClientOpt) *ArtifactAPIClient {
 	c := &ArtifactAPIClient{
-		url: defaultArtifactAPIURL,
-		c:   new(http.Client),
+		url:    defaultArtifactAPIURL,
+		cdnURL: artifactReleaseCDN,
+		c:      new(http.Client),
```
michalpristas, line 147: not your PR but there's a potential eternal hang

michalpristas (approved):
seems ok, left a couple comments

pkg/testing/tools/artifacts_api.go
```diff
@@ -161,6 +170,62 @@ func (aac ArtifactAPIClient) GetVersions(ctx context.Context) (list *VersionList
...
+		if err != nil {
+			return fmt.Errorf("failed to create an HTTP request to %q: %w", url, err)
+		}
+
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return fmt.Errorf("failed to request %q: %w", url, err)
+		}
+		// we don't read the response, we just need the status code
+		resp.Body.Close()
```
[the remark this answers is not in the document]
rdner: Good catch. I didn't know that.
Here is the documentation section for the future reference https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/net/http/response.go;l=62-64

> The default HTTP client's Transport may not reuse HTTP/1.x "keep-alive" TCP connections if the Body is not read to completion and closed.

pkg/testing/tools/artifacts_api.go
```diff
@@ -161,6 +170,62 @@ func (aac ArtifactAPIClient) GetVersions(ctx context.Context) (list *VersionList
...
+		parsedVersion, err := version.ParseVersion(versionItem)
+		if err != nil {
+			return fmt.Errorf("failed to parse version %s: %w", versionItem, err)
+		}
+		// we check only release versions without `-SNAPSHOT`
+		if parsedVersion.Prerelease() != "" {
+			results = append(results, versionItem)
+			continue
+		}
+		url := fmt.Sprintf("%s/elastic-agent-%s-%s", aac.cdnURL, versionItem, suffix)
```
jlind23, line 196: Will this work the same with the new agent release that will be available out of band thanks to @pchila's work?

rdner: I can't confirm that, not familiar with the spec/implementation details but I'll talk to @pchila and I've assigned him as a reviewer.

pkg/testing/tools/artifacts_api.go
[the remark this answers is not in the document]
rdner: Addressed in https://github.com/elastic/elastic-agent/pull/4276/commits/6b80387e4bb11a28d865439563569a45ae5a10e6

pkg/testing/tools/artifacts_api.go
rdner: it's not obvious to me, can you elaborate more?

AndersonQ: Go default http client does not have a timeout

pkg/testing/tools/artifacts_api.go
```diff
@@ -161,6 +170,68 @@ func (aac ArtifactAPIClient) GetVersions(ctx context.Context) (list *VersionList
...
+			continue
+		default:
+			return fmt.Errorf("unexpected status code from %s - %d", url, resp.StatusCode)
+		}
+	}
+
+	// nothing changed
+	if len(vList.Versions) == len(results) {
+		return nil
+	}
```
AndersonQ, line 228: [NIT]
You could remove it, if they're equal, there is no harm in swamping one for the other.

pkg/testing/tools/artifacts_api.go
```diff
@@ -119,6 +122,10 @@ func WithUrl(url string) ArtifactAPIClientOpt {
 	return func(aac *ArtifactAPIClient) { aac.url = url }
 }
 
+func WithCDNUrl(url string) ArtifactAPIClientOpt {
```
AndersonQ, line 125: [Suggestion]
I won't put it as a blocker, but so far I haven't seen any reference to the artifacts API as _CDN_. From my perspective using _CDN_ is at least counter intuitive, if not confusing.

I'd suggest to use `WithURL` as there is just one URL to be configured in the client

pkg/testing/tools/artifacts_api_test.go
```diff
@@ -216,3 +219,74 @@ func TestDefaultArtifactAPIClient(t *testing.T) {
...
+			"7.17.9",
+			"7.17.10",
+			"8.6.0",
+			"8.6.2",
+			"8.7.0",
+			"8.7.1",
+			"8.8.1",
+			"8.9.0-SNAPSHOT",
+		}
+		require.Equal(t, exp, versionList.Versions)
```
AndersonQ, line 269: [Suggestion]
No need to use `require` here, it's a normal test assertion, therefore `assert` should be used.

pkg/testing/tools/artifacts_api_test.go
```diff
@@ -216,3 +219,74 @@ func TestDefaultArtifactAPIClient(t *testing.T) {
...
+			"8.6.0",
+			"8.6.1",
+			"8.6.2",
+			"8.7.0",
+			"8.7.1",
+			"8.8.0",
+			"8.8.1",
+			"8.9.0-SNAPSHOT",
+		}
+		require.Equal(t, exp, versionList.Versions)
```
AndersonQ, line 290: [Suggestion]
No need to use `require` here, it's a normal test assertion, therefore `assert` should be used.

AndersonQ (approved):

pkg/testing/tools/artifacts_api.go
rdner: @AndersonQ this code is used only for testing, would not our test timeout stop the test anyway?

pkg/testing/tools/artifacts_api.go
rdner: As a rule, I don't mutate the function argument unless necessary.

pkg/testing/tools/artifacts_api.go
rdner: I'm not sure I understand your comment. There is the artifact API URL and there is the CDN URL where we publish our artifacts for the customers once versions are released. We need both at the same time and we already have `WithURL` option for the artifact API.

pkg/testing/tools/artifacts_api_test.go
rdner: What is the benefit of using `assert` instead? This is the only test check in the entire test. We use `assert` when do multiple assertions in a row, so the test does not stop after the first assertion. In this case there is not difference.

pkg/testing/tools/artifacts_api_test.go
rdner: See https://github.com/elastic/elastic-agent/pull/4276#discussion_r1494768403

pkg/testing/tools/artifacts_api.go
AndersonQ: oh, I mixed things up and I forgot there was already another `WithURL`. Never mind

pkg/testing/tools/artifacts_api.go
AndersonQ: They will, besides using a context when creating the requests, as you did, will respect the context's cancellation and interrupt the request.

AndersonQ: anyway, for me it isn't a blocker and I'd say it's resolved.
@michalpristas, is it solved for you?

michalpristas: sorry for late response, this was not a blocker for me

pkg/testing/tools/artifacts_api.go
pchila: I don't know details yet on how the new `elastic-agent-x.y.z+buildYYYYMMDDHHMMSS ` are gonna be distributed. If Artifact API has those versions correctly returned in the version list and the *same exact version* appear in the CDN URL this could work otherwise this check will always fail for the early releases

PR #4285, opened 2024-02-20
Extract vault as reusable component

### LINKED ISSUE #3866: [Unprivileged] Split the macOS vault out into a reusable component for macOS only running as root
**Describe the enhancement:**

The Elastic Agent on macOS needs to work without being able to use the system level key store for the vault. There is a filesystem vault but is currently only used on Linux and Windows. When running the Elastic Agent on macOS in unprivileged mode it needs to use the filesystem fault.

This will allow the Elastic Agent to work properly on macOS installed with `--unprivileged`.

pchila:
## What does this PR do?

Separate vault file and darwin keychain implementations so they don't implement the same exact type on different OSes but can be used side-by-side.

Darwin keychain vault implementation is not expected to work on non-Darwin OSes and the compile dependency is decoupled by using an empty implementation that will always error on other OSes.

## Why is it important?

This will allow to have a new Factory function [`vault.New()`](https://github.com/elastic/elastic-agent/compare/main...pchila:extract-vault-as-reusable-component?expand=1#diff-7981f8301ac84c3dd5a2873439485aac0a33cc36c04b8b218758bf4e28c5750aR34) that can return multiple underlying types that implement the `Vault` interface

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #3866

## Questions to ask yourself

- ...


internal/pkg/agent/application/paths/paths.go
```diff
@@ -2,32 +2,25 @@
...
-	ShellWrapperPath = "/usr/bin/elastic-agent"
+	// defaultAgentVaultName is keychain item name for keychain based vault (available on MacOS at the moment)
+	defaultAgentVaultName = "co.elastic.elastic-agent"
 
-	// ShellWrapper is the wrapper that is installed.  The %s must
-	// be substituted with the appropriate top path.
-	ShellWrapper = `#!/bin/sh
-exec %s/elastic-agent $@
-`
+	// defaultAgentVaultPath is the directory name where the file-based vault store is located
```
AndersonQ, line 14: [Suggestion]
I'd say the vault isn't a store, besides the agent already have a state store. So to avoid confusion, I'd suggest to remove the `store` here.


```suggestion
	// defaultAgentVaultPath is the directory name where the file-based vault is located
```

internal/pkg/agent/vault/vault.go
```diff
@@ -0,0 +1,38 @@
...
+	Set(ctx context.Context, key string, data []byte) (err error)
+	Remove(ctx context.Context, key string) (err error)
+	Close() error
+}
+
+func New(ctx context.Context, opts ...OptionFunc) (Vault, error) {
+	options := ApplyOptions(opts...)
+
+	if runtime.GOOS == "darwin" && !options.unprivileged {
+		// TODO add checks for unprivileged and proper fallback
```
AndersonQ, line 33: Is it done?

internal/pkg/agent/vault/vault_keychain_notdarwin.go
```diff
@@ -0,0 +1,42 @@
...
+//go:build !darwin
+
+package vault
+
+import (
+	"context"
+	"errors"
+)
+
+var ErrorImplementationNotAvailable = errors.New("vault keychain implementation is not available")
```
AndersonQ, line 14: [Consistency]
We use `Err` instead of `Error` throughout the code, so I'd ask you to keep it consistent.

AndersonQ (approved):

internal/pkg/agent/vault/vault.go
pchila: Yes, yes it is :sweat_smile:

pchila: Addressed in d3af9f9d69b4be829de6bb302445d2a95127106b

internal/pkg/agent/vault/vault_keychain_notdarwin.go
pchila: Addressed in d3af9f9d69b4be829de6bb302445d2a95127106b

pchila:
Failure in build https://buildkite.com/elastic/elastic-agent/builds/7400#018dd144-6f59-4222-ae20-38083ef77cd1 is another occurrence of #3890 but the upgrade happened correctly

Issue #4086, opened 2024-01-15
The integration test framework should test the DEB and RPM installations of the agent


cmacknz:
- Relates https://github.com/elastic/elastic-agent/issues/4084
- https://github.com/elastic/elastic-agent/issues/3784

At minimum we need to ensure that these packages install and can be enrolled in Fleet. The RPM and DEB tests should run in their own test group to ensure they run in parallel with other installation types.

The Fleet log ingestion test could be used for this purpose. https://github.com/elastic/elastic-agent/blob/main/testing/integration/logs_ingestion_test.go#L40

We should also test that they can be upgraded, see https://github.com/elastic/elastic-agent/issues/3784. This will require dedicated test as this is a new upgrade path, this part of the work could be split into a separate issue if needed.

pierrehilbert:
Reopening as only the DEB part has been done for now

Issue #4084, opened 2024-01-15
[Fleet]: On enrolling RPM and Deb agents, `Restarting agent failed` error is displayed in CLI.


harshitgupta-qasource:
**Kibana Build details:**
```
VERSION: 8.12.0 BC6
BUILD: 70088
COMMIT: e9092c0a17923f4ed984456b8a5db619b0a794b3
Artifact Link: https://staging.elastic.co/8.12.0-3eba7f46/summary-8.12.0.html#elastic-agent
```

**Host OS and Browser version:** All, All

**Preconditions:**
1. 8.12.0 Kibana Cloud environment should be available.
2. Policy should be created.
3. Deb/RPM agent should be extracted.

**Steps to reproduce:**

1. Run agent enroll command for RPM/DEB
2. Observe that on enrolling RPM and Deb agents, `Restarting agent failed` error is displayed in CLI.

**What's working fine:**

- We are able to enroll the agents on running enable and start elastic-agent command.

**Expected:**
On enrolling RPM and Deb agents restarting agent error should not display in CLI.

**Screenshot:**
![image (1)](https://github.com/elastic/elastic-agent/assets/101545338/294fafa3-58f3-4eaa-912e-b4f2db559e61)

amolnater-qasource:
Secondary Review for this ticket is Done.

cmacknz:
I can reproduce this, suspect another unintended consequence of https://github.com/elastic/elastic-agent/pull/3815 where we now always consider a failure to restart with the control socket a fatal error.

The agent service isn't automatically started after running `dpkg -i` so the `enroll` commands attempt to restart it cannot succeed. Likely the fix will be similar to https://github.com/elastic/elastic-agent/pull/4042, we need to skip the attempt to restart the agent in this case because it is supposed to be manual.

```sh
ubuntu@valuable-gudgeon:~$ sudo dpkg -i ./elastic-agent-8.12.0-arm64.deb
Selecting previously unselected package elastic-agent.
(Reading database ... 66270 files and directories currently installed.)
Preparing to unpack .../elastic-agent-8.12.0-arm64.deb ...
Unpacking elastic-agent (8.12.0) ...
Setting up elastic-agent (8.12.0) ...
found symlink /usr/share/elastic-agent/bin/elastic-agent, unlink
create symlink /usr/share/elastic-agent/bin/elastic-agent to /var/lib/elastic-agent/data/elastic-agent-5cbf2e/elastic-agent
[...]

PR #4301, opened 2024-02-21
add a DEB integration test

leehinman:
## What does this PR do?

Add an integration test that installs elastic-agent via the DEB package and makes sure that logs are sent to Elasticsearch.

## Why is it important?

Need to test that DEB packages work.

## Checklist

## Author's Checklist

- [ ]

```
mage integration:single TestDebLogIngestFleetManaged
```

## Related issues

- Relates #4086

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/steps/integration_tests.sh
+++ b/.buildkite/scripts/steps/integration_tests.sh
@@ -19,7 +19,7 @@ else
  OVERRIDE_TEST_AGENT_VERSION=""
 fi
 # PACKAGE
-AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64,windows/amd64 PACKAGES=tar.gz,zip mage package
+AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64,windows/amd64 mage package
 
 # Run integration tests
 set +e
--- a/.buildkite/scripts/steps/integration_tests.sh
+++ b/.buildkite/scripts/steps/integration_tests.sh
@@ -19,7 +19,7 @@ else
  OVERRIDE_TEST_AGENT_VERSION=""
 fi
 # PACKAGE
-AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64,windows/amd64 mage package
+AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64,windows/amd64 PACKAGES=tar.gz,zip,rpm,deb mage package
 
 # Run integration tests
 set +e
--- a/pkg/testing/fetch_test.go
+++ b/pkg/testing/fetch_test.go
@@ -1,0 +1,39 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package testing
+
+import (
+	"errors"
+	gtesting "testing"
+)
+
+func TestGetPackageSuffix(t *gtesting.T) {
+	tests := map[string]struct {
+		os       string
+		arch     string
+		format   string
+		expected string
+		err      error
+	}{
+		"windows zip":       {os: "windows", arch: "amd64", format: "zip", expected: "windows-x86_64.zip", err: nil},
+		"windows msi":       {os: "windows", arch: "amd64", format: "msi", expected: "", err: ErrUnsupportedPlatform},
+		"linux deb":         {os: "linux", arch: "amd64", format: "deb", expected: "amd64.deb", err: nil},
+		"linux rpm":         {os: "linux", arch: "amd64", format: "rpm", expected: "x86_64.rpm", err: nil},
+		"linux tar.gz":      {os: "linux", arch: "amd64", format: "targz", expected: "linux-x86_64.tar.gz", err: nil},
+		"linux pkg.tar.zst": {os: "linux", arch: "amd64", format: "pkg.tar.zst", expected: "", err: ErrUnsupportedPlatform},
+		"darwin arm64":      {os: "darwin", arch: "arm64", format: "targz", expected: "darwin-aarch64.tar.gz", err: nil},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *gtesting.T) {
+			got, err := GetPackageSuffix(tc.os, tc.arch, tc.format)
+			if !errors.Is(err, tc.err) {
+				t.Fatalf("wrong error.  expected: %v got: %v", tc.err, err)
+			}
+			if got != tc.expected {
+				t.Fatalf("wrong output. expected: %s got: %s", tc.expected, got)
+			}
+		})
+	}
+}
--- a/pkg/testing/fixture_install.go
+++ b/pkg/testing/fixture_install.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -132,11 +133,32 @@ func (f *Fixture) Install(ctx context.Context, installOpts *InstallOpts, opts ..
 	// check for running agents before installing
 	assert.Empty(f.t, getElasticAgentProcesses(f.t), "there should be no running agent at beginning of Install()")
 
-	installArgs := []string{"install"}
+	switch f.packageFormat {
+	case "targz", "zip":
+		return f.installNoPkgManager(ctx, installOpts, opts)
+	case "deb":
+		return f.installDeb(ctx, installOpts, opts)
+	default:
+		return nil, fmt.Errorf("package format %s isn't supported yet", f.packageFormat)
+	}
+}
+
+// installNoPkgManager installs the prepared Elastic Agent binary from
+// the tgz or zip archive and registers a t.Cleanup function to
+// uninstall the agent if it hasn't been uninstalled. It also takes
+// care of collecting a diagnostics when AGENT_COLLECT_DIAG=true or
+// the test has failed.
+// It returns:
+//   - the combined output of Install command stdout and stderr
+//   - an error if any.
+func (f *Fixture) installNoPkgManager(ctx context.Context, installOpts *InstallOpts, opts []process.CmdOption) ([]byte, error) {
+	f.t.Logf("[test %s] Inside fixture installNoPkgManager function", f.t.Name())
 	if installOpts == nil {
 		// default options when not provided
 		installOpts = &InstallOpts{}
 	}
+
+	installArgs := []string{"install"}
 	installOptsArgs, err := installOpts.toCmdArgs(f.operatingSystem)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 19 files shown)


.buildkite/scripts/steps/integration_tests.sh
```diff
@@ -19,7 +19,7 @@ else
  OVERRIDE_TEST_AGENT_VERSION=""
 fi
 # PACKAGE
-AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64,windows/amd64 PACKAGES=tar.gz,zip mage package
+AGENT_PACKAGE_VERSION="${OVERRIDE_AGENT_PACKAGE_VERSION}" DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64,linux/arm64,windows/amd64 mage package
```
cmacknz, line 22: By removing PACKAGES, I think this also builds all the Docker container variants we don't need. Not wrong, but slower than it could be.

Does adding `deb,rpm` to the packages list or does it cause problems with windows?

cmacknz: Is it faster if we just build Windows with a separate invocation of `mage package`? I'm not sure if the underlying cross-compilation is actually parallelized.

pkg/testing/fixture_install.go
```diff
@@ -302,6 +324,80 @@ func getElasticAgentProcesses(t *gotesting.T) []runningProcess {
...
+	})
+
+	// start elastic-agent
+	out, err = exec.CommandContext(ctx, "sudo", "systemctl", "start", "elastic-agent").CombinedOutput()
+	if err != nil {
+		return out, fmt.Errorf("systemctl start elastic-agent failed: %w", err)
+	}
+
+	// apt install doesn't enroll, so need to do that
+	enrollArgs := []string{"elastic-agent", "enroll"}
```
cmacknz, line 374: Was this affected by https://github.com/elastic/elastic-agent/issues/4084? Or are we just logging errors and not actually changing the exit code of the command?

pkg/testing/fixture_install.go
```diff
@@ -322,6 +418,30 @@ func (i UninstallOpts) toCmdArgs() []string {
...
+	case "deb":
+		return f.uninstallDeb(ctx, uninstallOpts, opts)
+	default:
+		return nil, fmt.Errorf("uninstall of package format '%s' not supported yet", f.packageFormat)
+	}
+}
+
+func (f *Fixture) uninstallDeb(ctx context.Context, uninstallOpts *UninstallOpts, opts []process.CmdOption) ([]byte, error) {
+	// stop elastic-agent, non fatal if error, might have been stopped before this.
+	out, err := exec.CommandContext(ctx, "sudo", "systemctl", "stop", "elastic-agent").CombinedOutput()
```
cmacknz, line 433: Is making the deb smart enough to start/stop the service a possibility? 

Not something for this PR, just wondering if this another point of friction we've arbitrarily introduced to using these packages.

testing/integration/logs_ingestion_test.go
```diff
@@ -143,6 +211,7 @@ func testMonitoringLogsAreShipped(
 			"add_cloud_metadata: received error failed requesting openstack metadata: Get \\\"https://169.254.169.254/2009-04-04/meta-data/instance-type\\\": dial tcp 169.254.169.254:443: connect: connection refused",               // okay for the cloud metadata to not work
 			"add_cloud_metadata: received error failed with http status code 404", // okay for the cloud metadata to not work
 			"add_cloud_metadata: received error failed fetching EC2 Identity Document: operation error ec2imds: GetInstanceIdentityDocument, http response error StatusCode: 404, request to EC2 IMDS failed", // okay for the cloud metadata to not work
+			"failed to invoke rollback watcher: failed to start Upgrade Watcher: fork/exec /var/lib/elastic-agent/elastic-agent: no such file or directory",                                                   //on debian this happens probably need to fix.
```
cmacknz, line 214: I assume this only happens for the DEB? If we are the DEB we probably shouldn't even be launching the watcher, or I suppose we could fix the path.

Either way this is worth a follow up issue.

pkg/testing/fixture_install.go
cmacknz: I am interested to know why this doesn't doesn't catch that problem.

cmacknz (commented):
LGTM, couple questions before approving. 

Looking at the test time using a sample size of 1:

* This PR's integration test step took 2h11 from https://buildkite.com/elastic/elastic-agent/builds/7435#018dd27c-0e2a-47c9-8e70-14b8a46d9677
* The most recent success on main took 1h52m from https://buildkite.com/elastic/elastic-agent/builds/7437#018dd291-acc2-46e4-a2ed-ec33f9c51afb

That's an additional 19 minutes which is a sizeable increase. I am wondering if we can shave some of that off by not building the packages we don't need yet, namely `rpm` and `docker`. At the same time, we need them eventually.

I have a strong suspicion the docker packages cost the most time, because it builds several container variants for each architecture, and even if we were testing it we'd only need one of them.

.buildkite/scripts/steps/integration_tests.sh
cmacknz: We could also make the windows target smart enough to ignore the deb and rpm packages if that was in fact the problem.

AndersonQ: +1

testing/integration/logs_ingestion_test.go
```diff
@@ -99,6 +99,74 @@ func TestLogIngestionFleetManaged(t *testing.T) {
...
+	agentFixture, err := define.NewFixture(t, define.Version(), atesting.WithPackageFormat("deb"))
+	require.NoError(t, err)
+
+	// 1. Create a policy in Fleet with monitoring enabled.
+	// To ensure there are no conflicts with previous test runs against
+	// the same ESS stack, we add the current time at the end of the policy
+	// name. This policy does not contain any integration.
+	t.Log("Enrolling agent in Fleet with a test policy")
+	createPolicyReq := kibana.AgentPolicy{
+		Name:        fmt.Sprintf("test-policy-enroll-%d", time.Now().Unix()),
```
AndersonQ, line 128: [NIT]
IIRC, most tests do so by adding a uuid to the policy name.

AndersonQ (approved):

pkg/testing/fetch_test.go
```diff
@@ -0,0 +1,39 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package testing
+
+import (
+	"errors"
+	gtesting "testing"
+)
+
+func TestGetPackageSuffix(t *gtesting.T) {
```
pchila, line 12: :heart:

pkg/testing/fixture_install.go
pchila: +1

testing/integration/groups_test.go
```diff
@@ -30,4 +30,7 @@ const (
 
 	// Upgrade group of tests. Used for testing upgrades.
 	Upgrade = "upgrade"
+
+	// Deb group of tests. Used for testing .deb packages install & upgrades
+	Deb = "deb"
```
pchila, line 35: Why do we define a separate group here? we could run the deb tests on any linux machine so I would be inclined to add them to default/other existing group (we save spawning one extra runner too)

testing/integration/logs_ingestion_test.go
pchila: +1

pchila (approved):
Code looks good, I have just a small doubt about defining a new Deb group (this entails provisioning an extra runner) as I think we could put the test into the default group and skip it on non-debian OSes

.buildkite/scripts/steps/integration_tests.sh
leehinman: Sorry, forgot to add the comment on this.  I tried with `PACKAGES="tar.gz,zip,deb,rpm"` and it failed to build in CI, "compiled mage file failed to run" if I remember correctly.  So I think we have a bug in the mage file, I'm still looking into it.  If I don't find it today or tomorrow I'll open up an issue.

cmacknz: Thanks, not opposed to optimizing this separate from merging this PR.

pkg/testing/fixture_install.go
leehinman: Not sure I'm following.  If you run `dpkg -i elastic-agent-8.13.0-SNAPSHOT-arm64.deb`, to "install" the deb package, there is no place to add the enroll token.  So that is what is happening here.

cmacknz: Right you need to run `elastic-agent enroll`. The "bug" in https://github.com/elastic/elastic-agent/issues/4084 is that if you do this right after `dpkg -i` it fails with an error because it can't restart the agent, because the service isn't running yet.

I now see you don't hit this problem because you `systemctl start elastic-agent` first. This is not what our default instructions for installing the deb suggest however, although it is a totally reasonable thing to do.

testing/integration/groups_test.go
leehinman: > Why do we define a separate group here? we could run the deb tests on any linux machine so I would be inclined to add them to default/other existing group (we save spawning one extra runner too)

Abundance of caution and me not knowing all the internals of the testing suite :-). I'll drop this.  FYI we will have to add Rocky or Alma linux as target distros for the RPM tests.  We really need to test DEB & RPM in their native environments.

testing/integration/logs_ingestion_test.go
leehinman: https://github.com/elastic/elastic-agent/issues/4332

pkg/testing/fixture_install.go
leehinman: I think we should add `systemctl start elastic-agent` to the DEB postinst script.  Any reason not to do that?

cmacknz: I think it would prevent having `elastic-agent enroll --delay-enroll` work reliably now that I think about it a bit.

The use case would be to install the deb in a VM image and set it to enrol when it next boots up.

testing/integration/groups_test.go
pchila: I mean we don't necessarily need to drop it, putting it into the default group at the moment would save us the overhead. When we add the other runners we may want to re-assess the grouping and see what we can do.
I think ubuntu for the moment should do as a distro, maybe later we want to have proper debian server or other distros that we want to focus our testing on)

blakerouse: Please don't drop this. I think the isolation tests between deb/rpm's and `install` command installations are good to have.

testing/integration/logs_ingestion_test.go
```diff
@@ -99,6 +99,74 @@ func TestLogIngestionFleetManaged(t *testing.T) {
...
+		Stack: &define.Stack{},
+		OS: []define.OS{
+			{
+				Type:   define.Linux,
+				Distro: "ubuntu",
+			},
+		},
+		Local: false,
+		Sudo:  true,
+	})
```
blakerouse, line 114: I am not saying do it in this PR, but another way of handling the different between deb/RPM vs `install` subcommand would be to define it in the test requirements. An example could be to change to:

```
		Group: Deb,
		Stack: &define.Stack{},
                 Packaging: define.Deb,
		Local: false,
		Sudo:  true,
	})
```

The testing framework could then look at the `Packaging: define.Deb` and determine automatically that this needs to be on an OS that supports deb's. Each OS can say what it supports in its definition. Then when running the framework with `MATRIX="true"` it could run on Ubuntu, Debian, etc. By default it would just pick Ubuntu. Doing it this way would also allow only the `deb` to be copied to the host. Currently with this change, it is going to copy both the `targz` and `deb` to the host, when both are not needed.

.buildkite/scripts/steps/integration_tests.sh
leehinman: So I finally found the timing issue in the magefile for building with the multiple package types defined.  We tried build all the package types at the same time.  So if you have "zip" and the timing is right you can have a compiled magefile that isn't for the GOOS you want.  The "fix" will do the GOOS-ARCH in parallel, so all linux-amd64, then all linux-arm, then windows-amd64.

pkg/testing/fixture_install.go
leehinman: I think this requires it's own issue to discuss I added #4381 to do that.  I think we can just leave this as-is for this PR, and adjust when we get #4381 figured out.

testing/integration/groups_test.go
leehinman: @pchila & @blakerouse are you OK if we keep the "Deb" group for now, so we can get some deb testing started.  And we come to a consensus when I work on the RPM test cases?

leehinman:
@blakerouse @pchila @cmacknz are we "OK" to merge or anything else we have to address for this PR?  

We absolutely have changes that need to happen before adding RPM.

testing/integration/groups_test.go
pchila: Good for me :+1:

.buildkite/scripts/steps/integration_tests.sh
cmacknz: Interestingly the timing vs main is still exactly the same as what it was before (caveat of sample size of 1 still applies):

* 1h52m for a recent build on main https://buildkite.com/elastic/elastic-agent/builds/7645
* 2h11m for the most recent build here https://buildkite.com/elastic/elastic-agent/builds/7624

Blake's suggestion for the future in https://github.com/elastic/elastic-agent/pull/4301#discussion_r1516304403 might be a way to reclaim some of this time but not uploading an unnecessary .tar.gz package. Probably worth putting that in an issue so it isn't lost in this PR.

cmacknz (approved):

cmacknz:
I am fine to defer future improvements to other PRs and issues.

The only thing I noticed after approving is that https://github.com/elastic/elastic-agent/pull/4334 was merged before this. We could add a check that the files were actually deleting after running `apt-get purge` as part of this test to ensure we don't regress on cleanup of the DEB package.

PR #4313, opened 2024-02-22
remove 8.12.0 version pinning

pchila:
## What does this PR do?
Remove the agent version pinning to 8.12.0 for the integration tests.
Such pinning is meant to keep the integration tests running when we release a new version but we cannot package the new  version because the dependencies for such version are not available.
This has been set and forgotten causing the agent to keep using 8.12.0 version even after newer versions have been created on the branch

## Why is it important?

This restores the correct run of elastic agent integration tests in CI for 8.12 branch

## Checklist

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- ...


cmacknz:
There is another problem here now, which is that we download the 8.12.1 release from staging where it has a different commit than the build that was promoted to artifacts.elastic.co https://github.com/elastic/elastic-agent/issues/4317

This PR fixes issues with the snapshot download failing, allowing more tests to run. I am going to force merge this, and we can fix https://github.com/elastic/elastic-agent/issues/4317 separately so that we are fixing this problem by peeling back the layers of problems one at a time.

Issue #4082, opened 2024-01-12
Agent should collect and report CPU and memory usage of monitoring components 


ycombinator:
Currently, Agent is only collecting and, therefore, reporting CPU and memory usage for itself and it's non-monitoring Beats components.  It is not collecting and reporting CPU and memory usage for the monitoring Beats, leading to undercounting.

So, for a default Fleet policy with the system integration and monitoring (logs and metrics both) enabled, Agent runs:
* 1 metricbeat instance for the system integration: `system/metrics-default`, 
* 1 filebeat instance for the system integration: `log-default`,
* 2 metricbeat instances for metrics monitoring: `http/metrics-monitoring` and `beat/metrics-monitoring`, and 
* 1 filebeat instance for logs monitoring: `filestream-monitoring`.

In the above scenario, Agent is currently only collecting and reporting CPU and memory usage for the metricbeat and filebeat instances for the system integration: `system/metrics-default` and `log-default`.  It is not collecting and reporting CPU and memory usage for the metricbeat and filebeat instances for monitoring: `http/metrics-monitoring`, `beat/metrics-monitoring`, and `filestream-monitoring`.

ycombinator:
> @ycombinator when you say "collect and report" what are you imagining? We make sure metrics for these end up in the `last 30s` log lines? Something else?

I meant collect and report monitoring metrics of the monitoring beats' resource usage the same way we do today for any non-monitoring beats.  The issue description has an example that should help clarify things further, but let me know if it doesn't.

fearful-symmetry:
@ycombinator I think we're on different pages here. So, the `last 30s` metrics for the monitoring beats are exposed via the agent logs:
```
grep -rh "last 30s" . | jq .component.id
"filestream-monitoring"
"system/metrics-default"
"beat/metrics-monitoring"
"http/metrics-monitoring"
"filestream-monitoring"
"system/metrics-default"
"beat/metrics-monitoring"
"http/metrics-monitoring"
"filestream-monitoring"
"system/metrics-default"
"beat/metrics-monitoring"
"http/metrics-monitoring"
"filestream-monitoring"
"system/metrics-default"
"beat/metrics-monitoring"
"http/metrics-monitoring"
"filestream-monitoring"
```

What specific monitoring interface or UX are you thinking of here? The Kibana/fleet agent resource usage GUI? Something else?

ycombinator:
[...]

PR #4326, opened 2024-02-22
Add metrics-monitoring beats to resource monitoring

fearful-symmetry:
## What does this PR do?
 Closes https://github.com/elastic/elastic-agent/issues/4082

This PR is currently a draft, as the test borrows the integration setup code from https://github.com/elastic/elastic-agent/pull/4150. Once that PR is merged, we can refactor this to use the same code to setup + install integrations

## Why is it important?

This PR changes the behavior of the monitoring beats, so they _also_ monitor and report metrics on themselves. This fixes an issue where the CPU and memory usage that agent reports to fleet can be deceptive, as they don't include _all_ the beats that are running under agent.

I also did a light refactor of the monitoring setup so we use constants for the monitoring beats IDs.

## Checklist


cmacknz:
Do the `beat/metrics` and `http/metrics` inputs report the resource usage and stats of the Beat they themselves are running in? Not just the other monitoring beats, as in `beat/metrics` is reporting the CPU usage of `beat/metrics-monitoring`, `http/metrics-monitoring` and `filestream-monitoring` and not just the latter two?

fearful-symmetry:
> Do the `beat/metrics` and `http/metrics` inputs report the resource usage and stats of the Beat they themselves are running in? Not just the other monitoring beats, as in `beat/metrics` is reporting the CPU usage of `beat/metrics-monitoring`, `http/metrics-monitoring` and `filestream-monitoring` and not just the latter two?

That's the goal of this PR, yeah. If you want, you can verify the negative, and comment-out the lines that map in `v1_monitor.go`, and the resulting integration test will fail.

cmacknz (commented):
How can we verify that the Fleet memory and CPU usage calculations are actually using the fields from the new documents?

If I install an agent built from this branch the CPU and memory in Fleet are exactly the same as one built from main. Possibly this is because they aren't using enough resources for their to be a detectable difference, but I would have expected memory to be higher at a minimum.

Screenshot 2024-03-08 at 3 59 37 PM

internal/pkg/agent/application/monitoring/v1_monitor_test.go
```diff
@@ -186,11 +186,20 @@ func TestMonitoringConfigComponentFields(t *testing.T) {
...
+						t.Errorf("expecting fields['binary'] = 'filebeat', got %q", binary)
+					}
+					if componentID != "filestream-default" {
+						t.Errorf("expecting fields['id'] = 'filestream-default', got %q", componentID)
+					}
+				} else {
+					if binary != "filebeat" && binary != "metricbeat" {
+						t.Errorf("expected monitoring compoent to be metricbeat or filebeat, got %s", binary)
+					}
+					if componentID != "filestream-monitoring" && componentID != "beat/metrics-monitoring" && componentID != "http/metrics-monitoring" {
```
leehinman, line 200: optional.  I'm wondering if using the const `monitoringFilesUnitsID` might be better than the string "filestream-monitoring"

leehinman (approved):

testing/integration/metrics_monitoring_test.go
```diff
@@ -0,0 +1,138 @@
...
+		"filestream-monitoring",
+	}
+
+	require.Eventually(runner.T(), func() bool {
+		for _, cid := range componentIds {
+			query := genESQuery(agentStatus.Info.ID, cid)
+			res, err := estools.PerformQueryForRawQuery(ctx, query, "metrics-elastic_agent*", runner.info.ESClient)
+			require.NoError(runner.T(), err)
+			runner.T().Logf("Fetched metrics for %s, got %d hits", cid, res.Hits.Total.Value)
+			if res.Hits.Total.Value < 5 {
```
cmacknz, line 103: Why `< 5`? Doesn't any amount of hits mean there was at least one document matching the query?

testing/integration/metrics_monitoring_test.go
```diff
@@ -0,0 +1,138 @@
...
+						},
+					},
+					{
+						"match": map[string]interface{}{
+							"component.id": componentID,
+						},
+					},
+					{
+						"exists": map[string]interface{}{
+							"field": "system.process.cpu.total.value", // make sure we fetch documents that have the metric field used by fleet monitoring
```
cmacknz, line 129: You could also check for the memory metric.

testing/integration/metrics_monitoring_test.go
fearful-symmetry: Yah, my concern was that we could end up in some freak accident where a test misconfiguration causes us to re-use an agent install, and thus an agent ID, leading to some overlap of results. Not sure if this is realistic, though.

cmacknz: The agent ID you get on enrollment is unique within Fleet, so shouldn't be possible unless Fleet is completely broken.

This isn't protecting against that properly anyway IMO, it is just a magic number. 

A way that should actually work would be:
1. Install and enroll an agent with monitoring disabled.
2. Wait one full metrics collection cycle.
3. Ensure there are no hits in the metrics-* indices for that agent.
4. Turn monitoring on.
5. Ensure there are metrics now.

I'm not sure if all of that is actually worth it though.

cmacknz (approved):

PR #4329, opened 2024-02-23
Refactor EncryptedDiskStore to return io.Reader instead of io.ReadCloser

AndersonQ:
## What does this PR do?

Refactors EncryptedDiskStore to return io.Reader instead of io.ReadCloser

## Checklist

- ~~[ ] I have made corresponding change to the default configuration files
- ~~[ ] I have added tests that prove my fix is effective or that my feature work~~s~~

## Related issues

- Relates #3912

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/storage/disk_store.go
+++ b/internal/pkg/agent/storage/disk_store.go
@@ -97,7 +97,7 @@ func (d *DiskStore) Save(in io.Reader) error {
 }
 
 // Load return a io.ReadCloser for the target file.
-func (d *DiskStore) Load() (io.ReadCloser, error) {
+func (d *DiskStore) Load() (io.Reader, error) {
 	fd, err := os.OpenFile(d.target, os.O_RDONLY|os.O_CREATE, perms)
 	if err != nil {
 		return nil, errors.New(err,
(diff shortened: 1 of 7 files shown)


internal/pkg/agent/storage/disk_store.go
```diff
@@ -97,7 +97,7 @@ func (d *DiskStore) Save(in io.Reader) error {
 }
 
 // Load return a io.ReadCloser for the target file.
-func (d *DiskStore) Load() (io.ReadCloser, error) {
+func (d *DiskStore) Load() (io.Reader, error) {
```
michalpristas, line 100: you sure this is allright?

pchila (commented):
I can see that using os.ReadFile instead of opening the file with `os.O_RDONLY` removes the need for the Close() on the Reader we return from Load(), however looking at the code it seems that in most places the Reader (or ReadCloser) is passed to `config.NewConfigFrom` which calls `io.ReadAll()` before unmarshaling.

With this modification we are doubling the buffers with the diskstore content (one in the new ByteReader + the one in `config.NewConfigFrom`, I am worried that we are churning more memory as the `EncryptedDiskStore`s are created in many places throughout the code.

Maybe we should have a small benchmark of EncryptedDiskStore.Load() +  `config.NewConfigFrom` to have a better idea of the memory impact

Issue #4296, opened 2024-02-20
The versions used in upgrade integration tests should be driven by a configuration file


cmacknz:
The available set of versions for use in our upgrade integration tests is currently discovered dynamically using the artifacts API. https://github.com/elastic/elastic-agent/blob/b425e4fc8cc1b5f2ef9a4bac74055927ca14b4a4/testing/upgradetest/versions.go#L38-L53

This approach has been a common source of problems for several reasons:

1. The set of available versions changes relatively frequently as new minor and patch releases move from feature freeze to official release.
2. The artifacts API will return release build candidates as upgradable, but they are only available  from staging.elastic.co which is not and should not be a source URI that is supported by default in the agent. See https://github.com/elastic/elastic-agent/issues/3944.
3. Dynamically discovering the version creates the risk that a bug in the version discovery logic or the artifacts API causes us to unintentionally skip or change the set of versions we test against. The worst case scenario would be accidentally skipping the 7.17 upgrade tests leading to a broken upgrade process for older agents. Today our versions are selected dynamically in several places without assertions that the expected set of versions was returned, for example https://github.com/elastic/elastic-agent/blob/b425e4fc8cc1b5f2ef9a4bac74055927ca14b4a4/testing/integration/upgrade_standalone_test.go#L36-L37.

We should instead define the versions available for use in upgrade tests in a static configuration file so that running the tests has no dependency on dynamic version discovery or the artifacts API. 

We should move the version discovery logic into a periodic GitHub action or buildkite job that creates a PR to update the set of upgradable versions as new versions are made available. The file content should at minimum include:
* The release version number e.g. 8.12.0, 8.13.0-SNAPSHOT
* The source URI the artifacts can be downloaded from e.g. artifacts.elastic.co, staging.elastic.co, etc.
* Including the snapshot build ID should be possible but implementation can be left as a future exercise, where we automatically bump the snapshot build ID to use in tests as new snapshots become available.

rdner:
@pchila how would it affect your work on https://github.com/elastic/elastic-agent/issues/2579 ?

Should I start making these changes now or should I wait until the transition is over?

pchila:
While I don't like the idea of one more file we need to bump periodically, this should not affect my work on #2579 since it's already been merged and it will work like all the other upgrade tests

rdner:
I found 14 tests that perform an upgrade/downgrade:

Here is a one-liner that you can run in the `./testing` directory:

```bash
[...]

Issue #4268, opened 2024-02-15
[Flaky Test]: Multiple test cases fail because of the artifact API / CDN unavailability


rdner:
Any test case that downloads artifacts (e.g. `TestStandaloneUpgrade*`, `TestUpgrade*`, `TestFleetManagedUpgrade*`) may fail because the artifact API or CDN server is unavailable.

Here is a list with a few examples (the timestamp is recorded on the buildkite failure, so it's approximate):

### 2024-02-12T20:28:05.483Z
https://buildkite.com/elastic/elastic-agent/builds/7062#018d9ec4-9b4a-4877-b32e-4d9810e99d75

```
    fixture.go:632: >> running binary with: [/tmp/TestStandaloneUpgradeRetryDownload3070379000/001/elastic-agent-8.13.0-SNAPSHOT-linux-x86_64/elastic-agent version --binary-only --yaml]
    upgrade_standalone_retry_test.go:58: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_standalone_retry_test.go:58
        	Error:      	Received unexpected error:
        	            	failed to get a list of builds: 503: bad http status code
        	Test:       	TestStandaloneUpgradeRetryDownload
```

### 2024-02-13T21:24:31.816Z
https://buildkite.com/elastic/elastic-agent/builds/7141#018da41e-2c94-4883-a427-5120735369e5

```
    upgrade_broken_package_test.go:45: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_broken_package_test.go:45
        	Error:      	Received unexpected error:
        	            	error retrieving versions from Artifact API: getting versions: executing http request &{GET https://artifacts-api.elastic.co/v1/versions/ HTTP/1.1 1 1 map[]   0 [] false artifacts-api.elastic.co map[] map[]  map[]      0xc0000fb960}: Get "https://artifacts-api.elastic.co/v1/versions/": dial tcp 35.188.12.98:443: i/o timeout
        	Test:       	TestUpgradeBrokenPackageVersion
```

### 2024-02-13T16:11:14.713Z
https://buildkite.com/elastic/elastic-agent/builds/7119#018da2fc-1a7d-443b-9ed7-89566b637e75

```
fetcher_artifact.go:344: Downloading artifact progress 39.78%
    upgrade_standalone_test.go:77: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_standalone_test.go:77
[...]

PR #4330, opened 2024-02-23
Fix GPG test hash mismatch

pchila:
## What does this PR do?

Add a simple http artifact fetcher with configurable baseURL defaulting to `artifacts.elastic.co/beats/elastic-agent`

## Why is it important?

This allows to create an agent fixture that uses officially released elastic-agent artifacts.

Downloading directly the artifact from artifacts.elastic.co ensures that we use the same artifact for the end fixture as the one downloaded by agent during an upgrade without source-uri, resolving the hash mismatch problem we have due to artifact API presenting additional builds on top of the officially released one.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Author's Checklist

- [ ]

## Related issues

- Closes #4317 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4331, opened 2024-02-23
Use a version file instead of artifact API in integration tests

rdner:
## What does this PR do?

Now there is a new mage target `mage integrations:updateVersions` that requests a list of version from the artifact
API according to the set requirements and creates a `.agent-versions.json` file in the root of the repository.

This file is later used for running integration tests.

## Why is it important?

This will make our integration tests more stable and deterministic.

## Checklist

```
mage integration:updateVersions
cd testing/upgradetest
go test -v
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/4268
- Closes https://github.com/elastic/elastic-agent/issues/4296


.github/workflows/bump-agent-versions.yml
```diff
@@ -0,0 +1,57 @@
...
+            git push --set-upstream origin "update-agent-versions-$GITHUB_RUN_ID"
+            gh pr create \
+              --base "$GITHUB_REF_NAME" \
+              --fill-first \
+              --head "update-agent-versions-$GITHUB_RUN_ID" \
+              --label 'Team:Elastic-Agent' \
+              --label 'update-versions' \
+              --reviewer 'elastic/elastic-agent-control-plane' \
+              --repo $GITHUB_REPOSITORY
+          fi
```
rdner, line 57: I tested this Github workflow using my fork repo as a target. Here is the example of a created PR https://github.com/rdner/elastic-agent/pull/3

rdner:
`TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing` is failing, hopefully would be fixed by https://github.com/elastic/elastic-agent/pull/4330

testing/upgradetest/versions.go
```diff
@@ -36,33 +41,102 @@ var (
...
+var (
+	agentVersions       *AgentVersions
+	ensureAgentVersions = sync.OnceFunc(setAgentVersions)
+)
+
+func setAgentVersions() {
+	v, err := getAgentVersions()
 	if err != nil {
-		return nil, fmt.Errorf("error retrieving versions from Artifact API: %w", err)
+		panic(err)
```
belimawr, line 71: [Question]
Why panic instead of returning an error?

I believe it would be better to return an error with some added context.

testing/upgradetest/versions.go
```diff
@@ -81,68 +155,84 @@ func getUpgradableVersions(ctx context.Context, vList *tools.VersionList, upgrad
...
-			upgradableVersions = append(upgradableVersions, parsedVersion)
-			previousMajorSelected++
-			continue
+		case recentSnapshotsToFind > 0 && version.IsSnapshot():
+			upgradableVersions = append(upgradableVersions, version.String())
+			recentSnapshotsToFind--
+
+		// for the rest of the checks we capture only released versions
+		case version.Prerelease() != "" || version.BuildMetadata() != "":
+
```
belimawr, line 189: If I understood correctly, we don't want pre-release or versions with build metadata, so this `case` entry is here to skip those versions. In that case I believe explicitly adding a `continue` helps to understand the code.
```suggestion
		// for the rest of the checks we capture only released versions
		case version.Prerelease() != "" || version.BuildMetadata() != "":
			continue

```

testing/upgradetest/versions_test.go
```diff
@@ -5,88 +5,128 @@
...
 	t.Logf("act: %s", versions)
-	for i, exp := range expectedUpgradableVersions {
-		require.Equal(t, exp, versions[i].String())
-	}
+	require.Equal(t, expectedUpgradableVersions, versions)
+}
+
+func TestGetUpgradableVersions(t *testing.T) {
+	versions, err := GetUpgradableVersions()
+	t.Log(versions)
```
belimawr, line 114: Is this log intended to be here? I don't think it makes sense to log it on every run.
```suggestion
```

testing/upgradetest/versions_test.go
```diff
@@ -5,88 +5,128 @@
...
+	require.Truef(t, len(versions) > 1, "expected at least one version for testing, got %d", len(versions))
+}
+
+func TestPreviousMinor(t *testing.T) {
+	currentParsed, err := version.ParseVersion(bversion.Agent)
+	require.NoError(t, err)
+
+	v, err := PreviousMinor()
+	require.NoError(t, err)
+	t.Log(v)
```
belimawr, line 125: Same here:
```suggestion
```

testing/upgradetest/versions_test.go
```diff
@@ -5,88 +5,128 @@
...
+		UpgradeToVersion: "8.13.0", // to test that 8.14 is not returned
+		CurrentMajors:    3,        // should return 8.12.2, 8.12.1, 8.12.0
+		PreviousMajors:   3,        // should return 7.17.18
+		PreviousMinors:   2,        // should return 8.12.2, 8.11.4
+		RecentSnapshots:  2,        // should return 8.13.0-SNAPSHOT, 8.12.2-SNAPSHOT
+	}
+	versions, err := FetchUpgradableVersions(ctx, aac, reqs)
+	require.NoError(t, err)
 	t.Logf("exp: %s", expectedUpgradableVersions)
 	t.Logf("act: %s", versions)
```
belimawr, line 108: I know those lines were already there, but why log it if there is no error?

testing/upgradetest/versions.go
rdner: The answer is on the line 65. This function is used in the `OnceFunc` that is assigned on the package import. I don't know of a better way how to handle errors in this case.

magefile.go
```diff
@@ -1571,6 +1573,41 @@ func (Integration) Single(ctx context.Context, testName string) error {
...
+
+	aac := tools.NewArtifactAPIClient()
+	versions, err := upgradetest.FetchUpgradableVersions(ctx, aac, reqs)
+	if err != nil {
+		return fmt.Errorf("failed to fetch upgradable versions: %w", err)
+	}
+	versionFileData := upgradetest.AgentVersions{
+		TestVersions: versions,
+	}
+	file, err := os.OpenFile(upgradetest.AgentVersionsFilename, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0644)
```
AndersonQ, line 1596: [Nit]
You can use [`os.WriteFile`](https://pkg.go.dev/os#WriteFile)

magefile.go
```diff
@@ -1571,6 +1573,41 @@ func (Integration) Single(ctx context.Context, testName string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to open %s for write: %w", upgradetest.AgentVersionsFilename, err)
+	}
+	defer file.Close()
+
+	encoder := json.NewEncoder(file)
+	encoder.SetIndent("", "  ")
```
AndersonQ, line 1603: +1

testing/upgradetest/versions.go
```diff
@@ -81,68 +155,84 @@ func getUpgradableVersions(ctx context.Context, vList *tools.VersionList, upgrad
...
 	return upgradableVersions, nil
 }
 
-// PreviousMinor gets the previous minor version of the provided version.
-//
-// This checks with the artifact API to ensure to only return version that have actual builds.
-func PreviousMinor(ctx context.Context, version string) (string, error) {
-	versions, err := GetUpgradableVersions(ctx, version, 1, 0)
+// PreviousMinor gets the previous minor version available for upgrade.
```
AndersonQ, line 217: [Suggestion]
```suggestion
// PreviousMinor returns the previous minor version available for upgrade.
```

testing/upgradetest/versions_test.go
```diff
@@ -5,88 +5,128 @@
...
-	for i, exp := range expectedUpgradableVersions {
-		require.Equal(t, exp, versions[i].String())
-	}
+	require.Equal(t, expectedUpgradableVersions, versions)
```
AndersonQ, line 109: [Suggestion]
This is not a case for `require`. It's just a test that if it fails and the code would proceed it would not lead to unexpected behaviour or a panic

```suggestion
	assert.Equal(t, expectedUpgradableVersions, versions)
```

AndersonQ (changes requested):

testing/upgradetest/versions_test.go
rdner: @belimawr @AndersonQ Why not log it in the verbose level of the test? `t.Log` prints only in the verbose level and it's just a list of 5-10 items, I don't see a problem here.

rdner: Okay, I can see your point now, it's better to print it only in case of error, I'll do that.

magefile.go
rdner: I prefer to use streaming, instead of buffering. In this particular case it does not matter match, I agree, but I prefer to keep it this way.

testing/upgradetest/versions.go
rdner: @belimawr I reconsidered and added the `continue` because the switch has a `default` block with `break` anyway. If it reads better with `continue` we should add it, you're right.

cmacknz:
I checked out this PR and generated the versions file:

```sh
❯ mage integration:updateVersions
~/go/src/github.com/elastic/elastic-agent versions-file-generator !1 ······················ 01:21:00 PM
❯ cat .agent-versions.json
```
```json
{
  "testVersions": [
    "8.14.0-SNAPSHOT",
    "8.13.0",
    "8.12.3",
    "7.17.18"
  ]
}
```

I have a few questions about the output:
* Why do we list `8.13.0` instead of `8.13.0-SNAPSHOT`? Because the build candidate exists?
* Why do we list 8.12.3? The latest minor release is 8.12.2 and there is no planned 8.12.3 yet.

cmacknz:
If I change the agent version to 8.13.0 in this PR to simulate what it would do on the 8.13 branch I get:

```diff
diff --git a/version/version.go b/version/version.go
index 7b092deae1..eefb327ba2 100644
--- a/version/version.go
+++ b/version/version.go
@@ -4,5 +4,5 @@

 package version

-const defaultBeatVersion = "8.14.0"
+const defaultBeatVersion = "8.13.0"
 const Agent = defaultBeatVersion
```
```
❯ cat .agent-versions.json
{
  "testVersions": [
    "8.13.0-SNAPSHOT",
    "8.12.3",
    "8.12.2",
    "7.17.18"
  ]
}
```

This has the same problem of including 8.12.3 which shouldn't exist. If this is supposed to be 8.12.3-SNAPSHOT then it seems more reasonable.

testing/upgradetest/versions.go
```diff
@@ -36,33 +41,102 @@ var (
...
-	currentMajor := parsedUpgradeToVersion.Major()
-	var currentMajorSelected, previousMajorSelected int
+	return parsedVersions, nil
+}
+
+// FetchUpgradableVersions returns a list of versions that meet the specified requirements.
+//
+// Every version on the resulting list will meet the given requirements (by OR condition).
+// However, it's not guaranteed that the list contains the amount of versions per requirement.
+// For example, if only 2 previous minor versions exist but 5 requested, the list will have only 2.
```
cmacknz, line 131: Would it be better to return an error in this case? Requesting a specific number of versions is not a requirement if it is actually considered optional by the implementation.

testing/upgradetest/versions.go
```diff
@@ -36,33 +41,102 @@ var (
...
+		if strings.HasSuffix(dir, string(filepath.Separator)) {
+			return nil, fmt.Errorf("failed to find %s", AgentVersionsFilename)
+		}
 	}
-	err = aac.RemoveUnreleasedVersions(ctx, vList)
+
+	f, err := os.OpenFile(filePath, os.O_RDONLY, 0)
 	if err != nil {
-		return nil, fmt.Errorf("failed to remove unreleased versions: %w", err)
+		return nil, fmt.Errorf("failed to open %s", filePath)
```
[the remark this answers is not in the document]
AndersonQ: that's a good point

testing/upgradetest/versions.go
rdner: @cmacknz this is going to be a reply to your comments above too: https://github.com/elastic/elastic-agent/pull/4331#issuecomment-1977200360 and https://github.com/elastic/elastic-agent/pull/4331#issuecomment-1977224749

Every time there is a different output from the mage target we're going to have a PR for a review (unless there is an existing one). So, there will be a CI build where we can see if the versions are used properly, it's up to us when to approve and merge those PRs manually.

The purpose of the generator is to create a file closest to our requirements, that we can review and approve.

We currently have all these issues too, we just don't see them, we don't even know what versions we're testing against in each run unless we check the detailed logs. This is the main problem this PR is solving.

.github/workflows/bump-agent-versions.yml
```diff
@@ -0,0 +1,57 @@
...
+          then
+            echo "The versions file didn't change, skipping..."
+          else
+            echo "The versions file changed"
+            open=$(gh pr list --repo "$GITHUB_REPOSITORY" --label="update-versions" --limit 1 --state open --base "$GITHUB_REF_NAME")
+            if [ -n "$open" ]
+            then
+              echo "Another PR for $GITHUB_REF_NAME is in review, skipping..."
+              exit 0
+            fi
```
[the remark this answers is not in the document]
rdner: Updating is much harder, I can follow up on this later. Let's how this works first.

.agent-versions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "testVersions": [
+    "8.14.0-SNAPSHOT",
+    "8.13.0-SNAPSHOT",
+    "8.12.2",
+    "7.17.18"
+  ]
```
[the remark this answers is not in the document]
rdner: I can't see any functional difference of having a separate list. This is how the list was formed before my PR and I'd like to integrate this feature without changing the interface and how our tests work. This basically replaces what `GetUpgradableVersions` function returned.

testing/upgradetest/versions.go
```diff
@@ -36,33 +41,102 @@ var (
...
+}
+
+func getAgentVersions() (*AgentVersions, error) {
+	var filePath string
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the current directory: %w", err)
+	}
+	for {
+		dir = filepath.Dir(dir)
```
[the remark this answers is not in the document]
rdner: How is it better? We're already using the same approach in the integration tests when finding the project root by looking for `go.mod`.

Using `go list` is more complex, since it's an external command. I'd like to keep it as it is.

testing/upgradetest/versions.go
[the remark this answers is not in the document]
rdner: Fixed in https://github.com/elastic/elastic-agent/pull/4331/commits/99c61c01caf0d3cc28cc3fe4f01f8bb0476ab45c

.github/workflows/bump-agent-versions.yml
```diff
@@ -0,0 +1,57 @@
...
+            git diff -p
+            git add ".agent-versions.json"
```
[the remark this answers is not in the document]
rdner: Addressed in https://github.com/elastic/elastic-agent/pull/4331/commits/e5d10edb87ca03f7a2e8135305bdd3d59e958f2f

testing/upgradetest/versions.go
rdner: Addressed in https://github.com/elastic/elastic-agent/pull/4331/commits/e5d10edb87ca03f7a2e8135305bdd3d59e958f2f

testing/upgradetest/versions.go
rdner: We can also fine-tune the version requirements in a follow up later as we see the incoming automation PRs.

rdner: > I have a few questions about the output:
>
> Why do we list 8.13.0 instead of 8.13.0-SNAPSHOT? Because the build candidate exists?
> Why do we list 8.12.3? The latest minor release is 8.12.2 and there is no planned 8.12.3 yet.

In both cases these versions are [returned](https://artifacts-api.elastic.co/v1/versions/) by the artifact API as released and there is no additional check whether the artifact exists on the CDN or not.

<details>
<summary>response</summary>
```json
{
  "versions": [
    "7.17.14",
    "7.17.15",
    "7.17.16",
    "7.17.17",
    "7.17.18-SNAPSHOT",
    "7.17.18",
    "7.17.19-SNAPSHOT",
    "8.10.0",
    "8.10.1",
    "8.10.2",
    "8.10.3",
    "8.10.4",
    "8.11.0",
    "8.11.1",
    "8.11.2",
    "8.11.3",
    "8.11.4",
    "8.12.0",
    "8.12.1-SNAPSHOT",
    "8.12.1",
    "8.12.2-SNAPSHOT",
    "8.12.2",
    "8.12.3-SNAPSHOT",
    "8.12.3",
    "8.13.0-SNAPSHOT",
    "8.13.0",
    "8.14.0-SNAPSHOT"
  ],
  "aliases": [
    "7.17-SNAPSHOT",
    "7.17",
    "8.10",
    "8.11",
    "8.12-SNAPSHOT",
    "8.12",
    "8.13-SNAPSHOT",
    "8.13",
    "8.14-SNAPSHOT"
  ],
  "manifests": {
    "last-update-time": "Tue, 05 Mar 2024 13:33:02 UTC",
    "seconds-since-last-update": 161
  }
}
```
</details>

.github/workflows/bump-agent-versions.yml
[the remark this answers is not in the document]
rdner: https://github.com/elastic/elastic-agent/issues/4359

pchila:
> If I change the agent version to 8.13.0 in this PR to simulate what it would do on the 8.13 branch I get:
> 
> ```diff
> diff --git a/version/version.go b/version/version.go
> index 7b092deae1..eefb327ba2 100644
> --- a/version/version.go
> +++ b/version/version.go
> @@ -4,5 +4,5 @@
> 
>  package version
> 
> -const defaultBeatVersion = "8.14.0"
> +const defaultBeatVersion = "8.13.0"
>  const Agent = defaultBeatVersion
> ```
> 
> ```
> ❯ cat .agent-versions.json
> {
>   "testVersions": [
>     "8.13.0-SNAPSHOT",
>     "8.12.3",
>     "8.12.2",
>     "7.17.18"
>   ]
> }
> ```
> 
> This has the same problem of including 8.12.3 which shouldn't exist. If this is supposed to be 8.12.3-SNAPSHOT then it seems more reasonable.

To solve these problems maybe we can get the released versions from the github releases of [agent](https://github.com/elastic/elastic-agent/releases) and [beats (for 7.X)](https://github.com/elastic/beats/releases) ?

pkg/testing/tools/artifacts_api.go
```diff
@@ -138,20 +136,17 @@ func WithHttpClient(client httpDoer) ArtifactAPIClientOpt {
...
 // which will print a list of available operations
 type ArtifactAPIClient struct {
-	c      httpDoer
-	url    string
-	cdnURL string
-
-	logger logger
+	c       httpDoer
+	url     string
+	logFunc logFunc
```
pchila, line 141: I understand that this is needed because of the logger that has been added to the ArtifactAPIClient but I was hoping for a cleaner solution 😅

testing/upgradetest/versions.go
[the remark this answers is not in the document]
pchila: The difference is in code: we don't need to go up one level at the time till the root of the filesystem (potentially in case the file is not the in current directory tree), we could just compose the expected path starting from the repository root. Not blocking anyways

pkg/testing/tools/artifacts_api.go
rdner: @pchila any better ideas? We use this client with 2 different logger interfaces: `t.Logf` in tests and `log.Printf` in the magefile. So, I came up with a common solution.

pchila (approved):
Looks good for a first iteration, we can always refine it later as needed.
All the outstanding comments on my side are non-blocking

pkg/testing/tools/artifacts_api.go
rdner: I made this change to make this PR work (conflict resolution), refactoring of the artifact client is not the goal of this PR. We can follow up on this while working on https://github.com/elastic/elastic-agent/issues/4354

belimawr (approved):

testing/upgradetest/versions.go
cmacknz: OK so an 8.12.3 artifact does exist. Looking into https://artifacts-staging.elastic.co/elastic-agent-package/8.12.3-511be3e7/manifest-8.12.3.json and tacking down the job that produced it TIL that Buildkite now does a daily staging build for example https://buildkite.com/elastic/unified-release-staging/builds/813.

So a staging build for the next release patch release will always exist, even if we are not planning to actually release it. This is just a slower version of the snapshot build so we would never want to test both this and a snapshot, we don't gain enough additional test coverage. 

Every non-snapshot build return by the artifacts API is potentially an unreleased build candidate. I think we need a way to obviously tell official releases from build candidates when doing version selection. There are a few ways to do this:
1. We could use https://www.elastic.co/api/product_versions which returns every available release
2. We could try to download each version from artifacts.elastic.co to see if we get a 404.
3. We could look at the available releases on the elastic-agent GitHub page as Paolo suggested earlier.

I think for upgrades we want to always test the two previous minor versions, using the official releases when they are available so we simulate what real users would be doing. So for main, 8.13, and 8.12 currently we'd want to test:

* main: 8.14.0-SNAPSHOT, 8.13.0-SNAPSHOT, 8.12.2, 7.17.18
* 8.13: 8.13.0-SNAPSHOT, 8.12.2 8.11.4, and 7.17.18
* 8.12: 8.12.3-SNAPSHOT, 8.11.4, 8.10.4, 7.17.18

I am admittedly a bit torn on whether we should just always test the latest snapshot (e.g 8.12.3-SNAPSHOT) for a given branch or use the most recent public release if it exists (e.g. 8.12.2). I am currently biased towards using the public release since that is what will happen in reality when a branch is released.

AndersonQ: but wasn't the whole point of testing the latest snapshot to ensure upgrade isn't broken on the new version? Or we're covering that with that test which modifies the agent's package to simulate it's another version?

AndersonQ (approved):

testing/upgradetest/versions.go
cmacknz: There are two halves of the upgrade process to test:

1. We need to make sure that previous releases can upgrade to the version we are about to release. Previous releases need to be able to find the agent executable so they can replace themselves with it and ensure the new watcher works properly. For this we want to test the two previous minors and 7.17.
2. We need to make sure the version we are about to release can upgrade itself to another version. In practice it is a downgrade to a snapshot from the same version. This  tests logic to download, extract, and replace the binary in the current version.

PR #4348, opened 2024-02-29
Integration Testing Framework: Add retries for artifacts API call

### LINKED ISSUE #4345: Integration Testing Framework: Add retries to Artifacts API call
Currently in the integration testing framework, the `ArtifactsAPIClient` doesn't retry calls to the Artifacts API. 
 Speaking with @davesys911, he suggested we should add retries to this call, somewhere in this function:

https://github.com/elastic/elastic-agent/blob/f9310c7f664bb25f37d671cca4e2fcf3749732c8/pkg/testing/tools/artifacts_api.go#L306-L319

ycombinator:
## What does this PR do?

This PR adds retries to the Artifacts API call in the integration testing framework, as recommended by @DaveSys911.

## Why is it important?

The Artifacts API can sometimes be flaky. 

## Related issues
- Resolves #4345

### DIFF
--- a/pkg/testing/tools/artifacts_api.go
+++ b/pkg/testing/tools/artifacts_api.go
@@ -159,13 +159,13 @@ func NewArtifactAPIClient(opts ...ArtifactAPIClientOpt) *ArtifactAPIClient {
 }
 
 // GetVersions returns a list of versions as server by the Artifact API along with some aliases and manifest information
-func (aac ArtifactAPIClient) GetVersions(ctx context.Context) (list *VersionList, err error) {
+func (aac ArtifactAPIClient) GetVersions(ctx context.Context, log logger) (list *VersionList, err error) {
 	joinedURL, err := aac.composeURL(artifactsAPIV1VersionsEndpoint)
 	if err != nil {
 		return nil, err
 	}
 
-	resp, err := aac.createAndPerformRequest(ctx, joinedURL)
+	resp, err := aac.createAndPerformRequest(ctx, joinedURL, log)
 	if err != nil {
 		return nil, fmt.Errorf("getting versions: %w", err)
 	}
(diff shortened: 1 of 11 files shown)


ycombinator:
Moving PR into draft to a) update unit test assertions and b) refactor logging logic so logger doesn't need to be passed down in several places.

pkg/testing/tools/artifacts_api.go
```diff
@@ -310,9 +317,27 @@ func (aac ArtifactAPIClient) createAndPerformRequest(ctx context.Context, URL st
 		return nil, err
 	}
 
-	resp, err := aac.c.Do(req)
+	// Make the request with retries as the artifacts API can sometimes be flaky.
+	var resp *http.Response
+	// TODO (once we're on Go 1.22): replace with for numAttempts := range maxAttemptsForArtifactsAPICall {
+	for numAttempts := 0; numAttempts < maxAttemptsForArtifactsAPICall; numAttempts++ {
+		resp, err = aac.c.Do(req)
+		if err == nil {
```
cmacknz, line 325: You should not continue to retry if the error is context.Cancelled or context.DeadlineExceeded.

ycombinator: Fixed in adf118f6cd.

cmacknz (approved):

rdner (commented):
@ycombinator Unfortunately, I didn't look at this PR in time. Sorry about that.

I don't think this is how we should approach it, we should:

1. Implement the `httpDoer` interface with retries and use this implementation as an option
2. Continue using the existing option pattern and remove the logger parameter.
3. The logger interface is not implemented by any logger, only by `testing.T` and this client is used beyond testing.

I addressed 2 and 3 in my PR https://github.com/elastic/elastic-agent/pull/4331 (had to resolve some conflicts)

For 1 @pchila filed an issue here https://github.com/elastic/elastic-agent/issues/4354

PR #4362, opened 2024-03-05
Enable unprivileged on macos

### LINKED ISSUE #3867: [Unprivileged] Enable `--unprivileged` on macOS
**Describe the enhancement:**

Current support for `--unprivileged` is isolated to Linux, this needs to be expanded to support macOS.

To allow Elastic Agent to run on macOS without requiring root privileges.

When `install --unprivileged` installs and Elastic Agent works running on macOS in this mode.

pchila:
## What does this PR do?

This PR allows running elastic-agent (either installed or just extracted from the package) as an unprivileged user, using a file-based vault as fallback instead of the keychain-based one.

When creating an EncryptedDiskStore, we can specify an `unprivileged` flag which will instantiate the file-based vault on Mac OS (the default value of such flag is defined as whether elastic-agent is running with administrative rights).

In most cases, such default value is enough to choose the correct vault implementation: the exception is `elastic-agent uninstall` command where a specific check for a file vault containing the agent secret has been added.
The additional check during uninstall is necessary since the command needs to run with administrative rights for both privileged and unprivileged elastic-agent installs (a simple check of the user permissions would not be useful as we would always come up with privileged permissions)

## Why is it important?

This change enables `elastic-agent` to be installed and run as an unprivileged user on Mac OS

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #3867

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -140,7 +140,13 @@ func run(override cfgOverrider, testingMode bool, fleetInitTimeout time.Duration
 }
 
 func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override cfgOverrider, stop chan bool, testingMode bool, fleetInitTimeout time.Duration, runAsOtel bool, awaiters awaiters, modifiers ...component.PlatformModifier) error {
-	cfg, err := loadConfig(ctx, override, runAsOtel)
+	// try early to check if running as root
+	isRoot, err := utils.HasRoot()
+	if err != nil {
+		return fmt.Errorf("failed to check for root permissions: %w", err)
+	}
+
+	cfg, err := loadConfig(ctx, override, runAsOtel, !isRoot)
 	if err != nil {
 		return err
 	}
@@ -163,19 +169,13 @@ func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override cf
 
 	l.Infow("Elastic Agent started", "process.pid", os.Getpid(), "agent.version", version.GetAgentPackageVersion())
 
-	cfg, err = tryDelayEnroll(ctx, l, cfg, override)
+	cfg, err = tryDelayEnroll(ctx, l, cfg, override, !isRoot)
 	if err != nil {
 		err = errors.New(err, "failed to perform delayed enrollment")
 		l.Error(err)
 		return err
 	}
 	pathConfigFile := paths.AgentConfigFile()
-
-	// try early to check if running as root
-	isRoot, err := utils.HasRoot()
-	if err != nil {
-		return fmt.Errorf("failed to check for root permissions: %w", err)
-	}
 
 	// agent ID needs to stay empty in bootstrap mode
 	createAgentID := !runAsOtel
@@ -373,7 +373,7 @@ LOOP:
 	return err
 }
 
-func loadConfig(ctx context.Context, override cfgOverrider, runAsOtel bool) (*configuration.Configuration, error) {
+func loadConfig(ctx context.Context, override cfgOverrider, runAsOtel, unprivileged bool) (*configuration.Configuration, error) {
 	if runAsOtel {
 		defaultCfg := configuration.DefaultConfiguration()
 		// disable monitoring to avoid injection of monitoring components
@@ -392,7 +392,7 @@ func loadConfig(ctx context.Context, override cfgOverrider, runAsOtel bool) (*co
 			errors.M(errors.MetaKeyPath, pathConfigFile))
 	}
 
-	if err := getOverwrites(ctx, rawConfig); err != nil {
+	if err := getOverwrites(ctx, rawConfig, unprivileged); err != nil {
 		return nil, errors.New(err, "could not read overwrites")
 	}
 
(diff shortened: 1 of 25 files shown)


internal/pkg/agent/cmd/run.go
```diff
@@ -88,7 +89,11 @@ func newRunCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Command {
 	// feature of the Elastic Agent. On Mac OS root privileges are required to perform the disk
 	// store encryption, by setting this flag it disables that feature and allows the Elastic Agent to
 	// run as non-root.
+	//
+	// Deprecated: MacOS can be run/installed without root privileges
 	cmd.Flags().Bool("disable-encrypted-store", false, "Disable the encrypted disk storage (Only useful on Mac OS)")
+	_ = cmd.Flags().MarkHidden("disable-encrypted-store")
+	_ = cmd.Flags().MarkDeprecated("disable-encrypted-store", "agent on Mac OS can be run/installed without root privileges, see elastic-agent install --help")
```
blakerouse, line 96: +1

internal/pkg/agent/cmd/run.go
```diff
@@ -162,6 +167,12 @@ func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override cf
 
 	l.Infow("Elastic Agent started", "process.pid", os.Getpid(), "agent.version", version.GetAgentPackageVersion())
 
+	// try early to check if running as root
+	isRoot, err := utils.HasRoot()
+	if err != nil {
+		return fmt.Errorf("failed to check for root permissions: %w", err)
```
blakerouse, line 173: The others you added used the word `administrator` and this one used `root`. I believe that was just because this was moved. I would like to see this consitent, maybe `root/Administrator` would be better and work for unix and windows.

internal/pkg/agent/storage/encrypted_disk_store.go
```diff
@@ -46,11 +47,21 @@ func NewEncryptedDiskStore(ctx context.Context, target string, opts ...OptionFun
 	if encryptionDisabled {
 		return NewDiskStore(target)
 	}
+
+	unprivileged := false
+
+	hasRoot, err := utils.HasRoot()
+	if err != nil || !hasRoot {
+		// TODO log error
```
blakerouse, line 55: TODO here?

pchila:
> We don't have a MacOS runner for integration tests yet. What is the plan for ensuring this works with the same level of confidence as Windows and Linux? We have a manual QA team that can help verify this, but we'd have to provide a detailed list of test cases first.

Not sure what other testcases may come to mind but here's what I have tested:
1. `elastic-agent run` from an uncompressed elastic-agent package as a regular Mac user
2. `sudo elastic-agent install --unprivileged` for standalone mode and fleet managed

Aside from checking manually that privileged still works I cannot think of specific additional scenarios to test.
Maybe we can add a testcase with endpoint to verify that it doesn't run unless we install agent as privileged?

We have a real problem with automated testing for this kind of changes, unit tests are unfeasible without massive refactoring in the touched parts of code (I would like to avoid a big-bang scenario) and we don't have runners for integration tests on Mac as you pointed out.

internal/pkg/agent/storage/encrypted_disk_store.go
pchila: Propagated the error in https://github.com/elastic/elastic-agent/pull/4362/files/6b2dc5973182afc60dd44aad888f291e57c3adba..78425bf5734c815486b57c4430bb5e6c0daf642a

internal/pkg/agent/cmd/run.go
pchila: cleaned up a bit in b50857049b98e7eb7f2d04fa6e3eeeea3f14ce7b

blakerouse (approved):
Looks great!

Thanks for adding tests where you could, a lot of these areas are really hard to unit tests.

cmacknz:
> Not sure what other testcases may come to mind but here's what I have tested:
> 
> 1. elastic-agent run from an uncompressed elastic-agent package as a regular Mac user
> 2. sudo elastic-agent install --unprivileged for standalone mode and fleet managed

Do upgrades work?

> We have a real problem with automated testing for this kind of changes, unit tests are unfeasible without massive refactoring in the touched parts of code (I would like to avoid a big-bang scenario) and we don't have runners for integration tests on Mac as you pointed out.

Long term the answer is to add a MacOS integration test runner, I don't expect this to be covered purely with unit tests.

pchila:
> Do upgrades work?

@cmacknz I just tested by repackaging the agent (we cannot downgrade as the old agent versions do not support running unprivileged) and everything works as expected when launching the upgrade with `sudo -u elastic-agent elastic-agent upgrade ...` so at least that still works

pierrehilbert:
> 
> Long term the answer is to add a MacOS integration test runner, I don't expect this to be covered purely with unit tests.

cc @rowlandgeoff to ensure this is under your radar.

pierrehilbert:
As discussed in our weekly meeting, this PR contains what is possible to be tested until we have our MacOS runner so I'm force merging even if SonarQube is not happy with the coverage level.

Issue #4168, opened 2024-01-31
The kuberenetes secrets provider should regenerate the agent configuration when a secrets change is detected


cmacknz:
The Kubernetes [secrets provider](https://www.elastic.co/guide/en/fleet/current/kubernetes_secrets-provider.html) today implements the Fetch context provider interface, which provides the secret value whenever the agent configuration is being regenerated for another reason.

https://github.com/elastic/elastic-agent/blob/4b404b18debf0f83d19fa72132a9907e5fcf8641/internal/pkg/core/composable/providers.go#L9-L17

In cases where secrets are short lived and rotated frequently, this can result in the agent policy using stale secrets values. In general there is currently always a risk that a secret remains stale if the agent configuration is not regenerated because of another detected change in the system, for example a policy change.

The Kuberenetes secrets provider should instead implement the Comm (or communicating) context provider interface, which informs the agent when a change is detected so the running configuration can be regenerated.

https://github.com/elastic/elastic-agent/blob/4b404b18debf0f83d19fa72132a9907e5fcf8641/internal/pkg/core/composable/providers.go#L18-L25

For an example, this is the way the Kubernetes leader election provider is implemented:

https://github.com/elastic/elastic-agent/blob/4b404b18debf0f83d19fa72132a9907e5fcf8641/internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go#L48-L49

https://github.com/elastic/elastic-agent/blob/4b404b18debf0f83d19fa72132a9907e5fcf8641/internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go#L112-L121

Since the secrets provider now implements an optional cache, whenever the cache refreshes we could detect if a value has changed and notify the agent.

https://github.com/elastic/elastic-agent/blob/4b404b18debf0f83d19fa72132a9907e5fcf8641/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L104-L117

Ideally we would use a watcher on the secret values, but this has been found to have performance and memory usage concerns.

constanca-m:
Hey @cmacknz , is this issue still ongoing? Is this still a priority?

adamkasztenny:
@constanca-m @cmacknz We're still running into this issue

constanca-m:
The ContextProvider has

> // Set sets the current mapping for this context. 
> 	Set(map[string]interface{}) error 

[...]

PR #4368, opened 2024-03-06
Add Signal to context provider interface

blakerouse:
## What does this PR do?

Adds a `Signal` function to the context provider interface.

## Why is it important?

This allows a fetch based context provider to signal that something has changed and the component model needs to be re-computed.

## Checklist

## Related issues

- Relates #4168


internal/pkg/core/composable/providers.go
```diff
@@ -19,6 +19,12 @@ type FetchContextProvider interface {
 type ContextProviderComm interface {
 	context.Context
 
+	// Signal signals that something has changed in the provider.
+	//
+	// Note: This should only be used by fetch context providers, standard context
+	// providers should use Set to update the overall state.
+	Signal()
```
cmacknz, line 26: I want to make sure I understand why fetch providers shouldn't call set.

Is this because `Set` is meant to add a mapping and its value, where as a `Fetch` provider is supposed to be provided the list of keys to get?

In the context of the k8s secrets provider we wouldn't call `Set` because we aren't adding secrets keys to the policy, the policy is using the provider to set/fetch their value?

blakerouse: I think you are saying it correctly, but the working of set/fetch might be a little off.

`Set` should not be called by the k8s secrets provider because the available keys are not known in advanced like all the other providers. That is the main reason the k8s secrets provider is a fetch provider, because we don't want it to load every possible secret. Because of this reason it needs a way to just signal that something has changed, and then `Fetch()` will be called again as the Elastic Agent processes the variables in the configuration.

cmacknz (approved):

PR #4371, opened 2024-03-07
[Kubernetes secret provider] Send signal to agent if cache is updated

constanca-m:
## What does this PR do?

If the secrets in the cache get updated, then we notify the agent.

## Why is it important?

See issue https://github.com/elastic/elastic-agent/issues/4168 and the comments there.

## Checklist

Check the README file in the root directory to see how to build your own custom image and loaded to kind cluster.

<details>
<summary>Example secret.</summary>

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-secret
data:
  extra: bXktc2VjcmV0LW1lc3NhZ2UK
```
</details>

<details>
<summary>Example input that you can add to EA standalone using the secret.</summary>

```yaml
      - id: kubernetes-node-metrics
        type: kubernetes/metrics
        use_output: default
        meta:
          package:
            name: kubernetes
            version: 1.52.0
        data_stream:
          namespace: default
        streams:
          - data_stream:
              dataset: kubernetes.proxy
              type: metrics
            metricsets:
              - proxy
            hosts:
              - 'localhost:10249'
            period: 10s
            processors:
              - add_fields:
                  target: myfield
                  fields:
                    value: ${kubernetes_secrets.default.my-secret.extra}
```
</details>

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/4168

## Screenshots

Using the setup as in the section on how to test locally, we can check that the new field has the correct value associated with the secret:

image

## Logs

If a secret gets updated, then the cache will be updated and this wall cause the signal to get triggered. Logs:

```
{"log.level":"info","@timestamp":"2024-03-07T12:07:46.095Z","log.logger":"composable","log.origin":{"file.name":"kubernetessecrets/kubernetes_secrets.go","file.line":119},"message":"Secrets cache was updated, the agent will be notified.","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2024-03-07T12:07:46.273Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":1382},"message":"component model updated","log":{"source":"elastic-agent"},"changes":{"components":{"updated":["kubernetes/metrics-default: [(kubernetes/metrics-default-kubernetes-node-metrics: updated)]"],"count":4},"outputs":{}},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2024-03-07T12:07:46.273Z","log.origin":{"file.name":"coordinator/coordinator.go","file.line":1181},"message":"Updating running component model","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
```

### DIFF
--- a/changelog/fragments/1709824109-k8s-secret-provider-trigger-signal.yaml
+++ b/changelog/fragments/1709824109-k8s-secret-provider-trigger-signal.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: When cache gets update, a signal gets triggered to notify the agent
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/4371
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: https://github.com/elastic/elastic-agent/issues/4168
--- a/changelog/fragments/1709824109-k8s-secret-provider-trigger-signal.yaml
+++ b/changelog/fragments/1709824109-k8s-secret-provider-trigger-signal.yaml
@@ -11,7 +11,7 @@
 kind: feature
 
 # Change summary; a 80ish characters long description of the change.
-summary: When cache gets update, a signal gets triggered to notify the agent
+summary: Kubernetes secrets provider has been improved to update a kubernetes secret  when the secret value changes.
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.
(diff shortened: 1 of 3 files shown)


internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -102,44 +102,61 @@ func getK8sClient(kubeconfig string, opt kubernetes.KubeClientOptions) (k8sclien
...
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-timer.C:
-			p.updateCache()
+			updatedCache := p.updateCache()
+			if updatedCache {
+				p.logger.Info("Secrets cache was updated, the agent will be notified.")
+				comm.Signal()
```
constanca-m, line 115: The signal only gets triggered when the cache is updated.

However, the use of the cache is optional. Should we set it as mandatory so we know that in all configurations, we always trigger the signal?

blakerouse: Cache is on by default, I think if they disable the cache then its fine that signal is not performed. Without the cache then it cannot perform this action. Would be good to document that behavior, but is acceptable to me.

I didn't see in the unit tests where the call the `comm.Signal` is tested. There is a testing communicator that you can use with `SetOnSignal(func() { ... })`, that will call the function when `comm.Signal` is called. You should test both that signal is called when something is changed, as well as not-called when something is not (using a timeout, is acceptable).

constanca-m: I added a unit test to check if the signal gets triggered.

blakerouse (changes requested):
Looks good to me.

Please fix the typo's that the linter is highlighting, otherwise this looks good. Please also add a changelog entry.

Only setting to request changes because I would like to review the changelog entry text, as its user facing.

changelog/fragments/1709824109-k8s-secret-provider-trigger-signal.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: When cache gets update, a signal gets triggered to notify the agent
```
blakerouse, line 14: This is very vague and not specific to what cache in the Elastic Agent. I recommend something like:

```suggestion
summary: Kubernetes secrets provider has been improved to update a kubernetes secret  when the secret value changes.
```

blakerouse (changes requested):

blakerouse:
You also have a race condition that is being caught by the race detector during the unit tests of your code.

MichaelKatsoulis (approved):

blakerouse (approved):
Thanks for the fixes, looks good!

cmacknz:
Backporting to 8.13.0, IMO this is a bug fix in the provider

Issue #4207, opened 2024-02-06
Add a test that can detect goroutine leaks


cmacknz:
Part of https://github.com/elastic/elastic-agent/issues/3833. 

The test in https://github.com/elastic/elastic-agent/issues/4206 has shown that we can fairly rapidly detect OS handle leaks by watching the rate of increase of handles over time. 

We can hopefully do something similar for goroutines, which are one of the primary causes of memory leaks in Go programs. Much like OS handles, the count of goroutines may change frequently but should not experience sustained growth.

To do this we will need to expose https://pkg.go.dev/runtime#NumGoroutine for each process running under agent and watch the rate of change over time.

The hope is that we can have this test check for goroutine leaks as part of each of our PRs.

PR #4388, opened 2024-03-08
Expand extended runtime tests to check for gorutine usage

fearful-symmetry:
## What does this PR do?

closes https://github.com/elastic/elastic-agent/issues/4207

## Why is it important?

This expands the extended runtime test to check for gorutine leaks. This more or less borrows the same strategy used by the handle leaks: measure the handle usage at regular intervals, then calculate the rate of change.

This also refactors the runtime test for readability and reusability. 

## Checklist


cmacknz:
How did you test that this detects leaks? I hope we don't have any by default :) 

How slow can the leak be before detection of it fails?

fearful-symmetry:
@cmacknz I ended up inadvertently testing this while I was developing it; I forgot to close the unix socket connection, causing both the tests to fail after every call to the status socket spawned a new connection. In general, "the metric under test increasing by one every measurement period" is enough to fail the test.

Should probably find a more reliable way to test this, though. I'll add a bug to the system integration and see what happens.

fearful-symmetry:
So, adding a few lines of code that spins off a goroutine every time `Fetch` is called in a metricset is enough to trigger a failure:
```
>>> (linux-amd64-ubuntu-2204-fleet) Test output (sudo) (stdout): Error:         "1.003423582991148" is not less than or equal to "0.1"
>>> (linux-amd64-ubuntu-2204-fleet) Test output (sudo) (stdout): Test:          TestLongRunningAgentForLeaks/TestHandleLeak
>>> (linux-amd64-ubuntu-2204-fleet) Test output (sudo) (stdout): Messages:      increase in count of gorutines exceeded threshold for system/metrics: system/metrics
```

michalpristas (commented):
test makes sense to me. i know finding right balance between false positives and negatives may be difficult, does 0.1 has some buffer so we avoid false failures?

fearful-symmetry:
@michalpristas I haven't gotten a false positive yet with 0.1, most beats will come in a few orders of magnitude less, usually 0.00001 to 0.001.

michalpristas (approved):

PR #4391, opened 2024-03-11
Fix GRPC server shutdown sequence

aleksmaus:
## What does this PR do?

While working on the agent I noticed that the gRPC server shutdown sequence is not working as expected, the 
```m.serverLoop``` is executed as a go routine within go routine. Thus the ```defer wgServer.Done()``` is called right away without waiting for the server loop to exit. 
So when the shutdown sequence is called at:
https://github.com/elastic/elastic-agent/blame/main/pkg/component/runtime/manager.go#L244
the ```wgServer.Wait()``` doesn't wait because it is already ```wgServer.Done()```.

This change seems only went in into 8.12. 
https://github.com/elastic/elastic-agent/commit/112f61896951f738c98f0a46c6df4160d31b718f#diff-ca07503cdc90a78b407b2bbf402f63604226ce4b1ec042a58ceb5f79bd325012R226

## Why is it important?

Fixed grpc server shutdown sequence.

## Checklist

Regression testing.

### DIFF
--- a/pkg/component/runtime/manager.go
+++ b/pkg/component/runtime/manager.go
@@ -231,7 +231,7 @@ func (m *Manager) Run(ctx context.Context) error {
 	wgServer.Add(1)
 	go func() {
 		defer wgServer.Done()
-		go m.serverLoop(ctx, listener, server)
+		m.serverLoop(ctx, listener, server)
 	}()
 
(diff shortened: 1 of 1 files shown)


faec (approved):
Good catch, thank you!

PR #2727, opened 2023-05-25
Add pipeline for packaging Elastic-agent

alexsapran:
## What does this PR do?

This pull request introduces a new packaging pipeline integrated with DRA. 

This new pipeline is intended to be triggered via an automation system as part of the release process of elastic-agent.

The idea is that you provide a manifest and it will detect and download the required components for elastic-agent and build, package it, and finally publish it with the release manager to the appropriate bucket.

The following parameters should be explicitly set when calling the pipeline (by env vars):
- DRA_WORKFLOW: what workflow the release manager should use
- ManifestURL: when packaging the elastic-agent, where to find the components
- DRA_VERSION: what version the release-manager should use

## Why is it important?

This is important for the work we are doing to improve the elastic-agent release process.

PR #714, opened 2022-07-13
Bundle elastic-agent.app for MacOS, needed to be able to enable the  …

aleksmaus:
## What does this PR do?

* Bundles elastic-agent as MacOS app bundle. This is need to be able to give Elastic Agent Full Disk Access that can be propagated down to osquery.
* Replaces the startup shell script with symlink
* Updated code in few places that relied on the os.Executable path discovery

Addresses [FDA support for Osquery Manager](https://github.com/elastic/security-team/issues/3983)

ATTN!!!: This was tested with signing the app bundle with my developer certificate. Need to figure out how to sign the app bundle with our production cert, probably requires some work with #infra team 

## Why is it important?

Addresses [FDA support for Osquery Manager](https://github.com/elastic/security-team/issues/3983)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas

## How to test this PR locally

1. Install the agent
2. Verify that it is running and healthy
2. Enable Full disk access through JAMF policy for example
3. Verify the the files in the restricted areas accessible. For example with planted test files the query 
```
SELECT f.path FROM file AS f JOIN mdfind ON mdfind.path = f.path AND mdfind.query = "kMDItemTextContent == 'testest123*'";
```
should be able to find them in the Desktop and Download locations:

Screen Shot 2022-07-13 at 9 54 06 AM

## Related issues

- Closes https://github.com/elastic/security-team/issues/3983

## Screenshots

The symlink replacing the shell script wrapper:
Screen Shot 2022-07-12 at 7 27 15 PM

The agent install with the app bundle (symlink points to the binary in the app bundle)
Screen Shot 2022-07-12 at 7 30 32 PM

Running agent:
Screen Shot 2022-07-12 at 7 30 49 PM

App bundle dir in installed agent directory:
Screen Shot 2022-07-12 at 7 31 19 PM

Dev cert signed app bundle:
Screen Shot 2022-07-12 at 7 37 05 PM

JAMF profile allow-listing the agent app bundle:
Screen Shot 2022-07-13 at 9 54 27 AM

Query results with Full Disk Access enabled:
Screen Shot 2022-07-13 at 9 54 06 AM

PR #4403, opened 2024-03-12
Add a mage target that packages elastic-agent using elastic-agent-core DRA

### LINKED ISSUE #4367: Add a mage target that packages the elastic-agent-core package instead of recompiling the agent
To support the early agent releases project we need to add a mage target that can package the agent using the elastic-agent-core DRA artifact produced by `mage packageAgentCore` instead of recompiling the agent directly.

The intent is for the early agent release project to be able to create releases where the agent binary itself does not change, but rather uses the agent artifact from the most recent stack release.

The early agent releases project produces agent builds by triggering the [elastic-agent-package](https://buildkite.com/elastic/elastic-agent-package) pipeline. Today this pipeline uses the existing `mage package` target which recompiles the agent:

https://github.com/elastic/elastic-agent/blob/d075097c676355752c8df5ee06a0c4fa79280369/.buildkite/pipeline.elastic-agent-package.yml#L76

https://github.com/elastic/elastic-agent/blob/d075097c676355752c8df5ee06a0c4fa79280369/.buildkite/scripts/steps/package.sh#L16-L17

We need to add a new target that downloads the `elastic-agent-core` DRA artifact and update the package pipeline to use that target. Locally we will almost never want to use this workflow, so an initial suggestion for what to call the target would be `mage packageDRA` or similar.

pchila:
## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #4367 

## Questions to ask yourself

- ...


dwhyrock (approved):
LGTM.

Love all the added verbose logging, btw.

pchila:
buildkite package this

pchila:
buildkite package this

changelog/fragments/1710863577-Add-mage-target-to-package-elastic-agent-using-DRAs.yaml
AndersonQ, line 1: I believe we don't need an entry for it as it's not a customer facing thing

pierrehilbert: Agree we shouldn't have a changelog here.

pchila:
test run of `elastic-agent-package` pipeline https://buildkite.com/elastic/elastic-agent-package/builds/1034#018e61a2-3b33-46f6-87a9-4e74d26478c1

changelog/fragments/1710863577-Add-mage-target-to-package-elastic-agent-using-DRAs.yaml
pchila: Changelog commit dropped.

cmacknz:
Something is up with the `elastic-agent` binary in the core package. I can't execute it on a Mac.

Screenshot 2024-03-22 at 4 02 35 PM

cmacknz (commented):
Generally looks good except the issue I found with the agent binary on Mac.

At some point we should probably start adding tests given the growing complexity of the build targets.

It would also be nice to have only one way to package the agent, perhaps one day we can drive the `EXTERNAL=true` case from a simplified version of the DRA manifest file or something.

magefile.go
```diff
@@ -1146,15 +1211,388 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+			}(pkgName, pkgDesc)
+
+			errGrp.Go(downloadFunc)
+		}
+	}
+
+	return downloadedArtifacts, errGrp.Wait()
+}
+
+func validateChecksum(hasher hash.Hash, fileToValidate, checksumFile string) error {
```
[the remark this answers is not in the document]
pchila: The piece of code that we have in the verifier that validates sha512 is https://github.com/elastic/elastic-agent/blob/d55869484332108b060ebe31bbca779706e77fc4/internal/pkg/agent/application/upgrade/artifact/download/verifier.go#L129-L202

While I am not too comfortable sharing implementations between our build/package system and the code being built/packaged I guess with some light refactor (pulling out the `.sha512` file name instead of assuming that it will always be the same as the name of the file we are validating + '.sha512' extension) I can reuse that part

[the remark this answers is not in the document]
pchila: refactor and usage in 19b974aba37e4b45bc9ffb2456ee6690b9398b95

pchila:
> Something is up with the `elastic-agent` binary in the core package. I can't execute it on a Mac.
> Screenshot 2024-03-22 at 4 02 35 PM

Not sure what is wrong with the `elastic-agent-core` artifacts described in manifest `https://staging.elastic.co/independent-agent/8.13.0+build202403190101/agent-artifacts-8.13.0+build202403190101.json` (I think this manifest is a patched one coming from the early release pipeline) but the binaries themselves seem to be invalid (maybe not signed or some other issue? )

I retested the packaging using a recent 13.0 staging build taken from https://staging.elastic.co/latest/8.13.json

```shell
mage clean && PLATFORMS="darwin" AGENT_DROP_PATH="build/elastic-agent-drop" DRA_VERSION="8.13.0+build202403250000" AGENT_PACKAGE_VERSION="8.13.0+build202403250000" DRA_WORKFLOW="staging" ManifestURL="https://staging.elastic.co/8.13.0-382f439d/manifest-8.13.0.json"  mage -v  downloadManifest packageUsingDRA
```

and everything works as expected.

One issue I discovered while testing is that the commit SHA used for packaging was the local git one, instead of forcing the commit hash from the binaries, fixed in 661c97ce54931d6694898bad0bb383b6adb74250

cmacknz:
I ran a slightly modified version of your command above to only build the darwin/arm64 targz package and am still getting the `"elastic-agent" is damaged` error when I try to execute it:

```sh
mage clean && PACKAGES=targz PLATFORMS=darwin/arm64 AGENT_DROP_PATH="build/elastic-agent-drop" DRA_VERSION="8.13.0+build202403250000" AGENT_PACKAGE_VERSION="8.13.0+build202403250000" DRA_WORKFLOW="staging"
ManifestURL="https://staging.elastic.co/8.13.0-382f439d/manifest-8.13.0.json"  mage -v  downloadManifest packageUsingDRA
```

Relevant section of the build logs downloading the elastic agent core artifact:

```sh
>>>> Packaging version: 8.13.0, build_id: 8.13.0-382f439d, manifest_version:2.1.0
url:"https://staging.elastic.co/8.13.0-382f439d/manifest-8.13.0.json", err <nil>
build "8.13.0-382f439d" project elastic-agent-core packages: {Branch:8.13 CommitHash:818d20468f031265ad7415ef5567b934ffff35c0 CommitURL:https://github.com/elastic/elastic-agent-core/commits/818d20468f031265ad7415ef5567b934ffff35c0 ExternalArtifactsManifestURL:https://artifacts-staging.elastic.co/elastic-agent-core/8.13.0-20209877/manifest-8.13.0.json BuildDurationSeconds:0 Packages:map[elastic-agent-core-8.13.0-darwin-aarch64.tar.gz:{URL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz ShaURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512 AscURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.asc Type:tar Architecture:aarch64 Os:[darwin] Classifier: Attributes:{IncludeInRepo: ArtifactNoKpi: Internal: ArtifactID: Oss: Group:}} elastic-agent-core-8.13.0-darwin-x86_64.tar.gz:{URL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz ShaURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz.sha512 AscURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz.asc Type:tar Architecture:x86_64 Os:[darwin] Classifier: Attributes:{IncludeInRepo: ArtifactNoKpi: Internal: ArtifactID: Oss: Group:}} elastic-agent-core-8.13.0-linux-arm64.tar.gz:{URL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz ShaURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz.sha512 AscURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz.asc Type:tar Architecture:arm64 Os:[linux] Classifier: Attributes:{IncludeInRepo: ArtifactNoKpi: Internal: ArtifactID: Oss: Group:}} elastic-agent-core-8.13.0-linux-x86_64.tar.gz:{URL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz ShaURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz.sha512 AscURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz.asc Type:tar Architecture:x86_64 Os:[linux] Classifier: Attributes:{IncludeInRepo: ArtifactNoKpi: Internal: ArtifactID: Oss: Group:}} elastic-agent-core-8.13.0-windows-x86_64.zip:{URL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip ShaURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip.sha512 AscURL:https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip.asc Type:zip Architecture:x86_64 Os:[windows] Classifier: Attributes:{IncludeInRepo: ArtifactNoKpi: Internal: ArtifactID: Oss: Group:}}] Dependencies:[]}
unfiltered packages: map[elastic-agent-core-8.13.0-darwin-aarch64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.asc tar aarch64 [darwin]  {     }} elastic-agent-core-8.13.0-darwin-x86_64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz.asc tar x86_64 [darwin]  {     }} elastic-agent-core-8.13.0-linux-arm64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz.asc tar arm64 [linux]  {     }} elastic-agent-core-8.13.0-linux-x86_64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz.asc tar x86_64 [linux]  {     }} elastic-agent-core-8.13.0-windows-x86_64.zip:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip.asc zip x86_64 [windows]  {     }}]
checking if elastic-agent-core-8.13.0-darwin-aarch64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.asc tar aarch64 [darwin]  {     }} should be included
platforms include darwin/arm64
checking if elastic-agent-core-8.13.0-windows-x86_64.zip:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-windows-x86_64.zip.asc zip x86_64 [windows]  {     }} should be included
checking if elastic-agent-core-8.13.0-linux-x86_64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-x86_64.tar.gz.asc tar x86_64 [linux]  {     }} should be included
checking if elastic-agent-core-8.13.0-linux-arm64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-linux-arm64.tar.gz.asc tar arm64 [linux]  {     }} should be included
checking if elastic-agent-core-8.13.0-darwin-x86_64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-x86_64.tar.gz.asc tar x86_64 [darwin]  {     }} should be included
filtered packages: map[elastic-agent-core-8.13.0-darwin-aarch64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.asc tar aarch64 [darwin]  {     }}]
packages to download: map[elastic-agent-core-8.13.0-darwin-aarch64.tar.gz:{https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512 https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.asc tar aarch64 [darwin]  {     }}]
<<<<<<<<< Downloaded: https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz to /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/dra/8.13.0-382f439d/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz
<<<<<<<<< Downloaded: https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512 to /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/dra/8.13.0-382f439d/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz.sha512
Source binary "/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/dra/8.13.0-382f439d/extracted/elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent" stat: &{name:elastic-agent size:80969936 mode:493 modTime:{wall:82948219 ext:63846986236 loc:0x103918aa0} sys:{Dev:16777232 Mode:33261 Nlink:1 Ino:93985876 Uid:501 Gid:20 Rdev:0 Pad_cgo_0:[0 0 0 0] Atimespec:{Sec:1711389435 Nsec:712631626} Mtimespec:{Sec:1711389436 Nsec:82948219} Ctimespec:{Sec:1711389436 Nsec:82984302} Birthtimespec:{Sec:1711389435 Nsec:712631626} Size:80969936 Blocks:158152 Blksize:4096 Flags:0 Gen:0 Lspare:0 Qspare:[0 0]}}
copying "/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/dra/8.13.0-382f439d/extracted/elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent" to "/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/golang-crossbuild/elastic-agent-darwin-arm64"
Running dependency: github.com/elastic/elastic-agent/dev-tools/mage.Package
```

The binary does appear to be notarized, I just can't execute it:

```sh
❯ file elastic-agent
elastic-agent: Mach-O 64-bit executable arm64

~/Downloads/builds/elastic-agent-8.13.0+build202403250000-darwin-aarch64 ············ ✘ INT 01:58:48 PM
❯ codesign -dvvvv ./elastic-agent
Executable=/Users/cmackenzie/Downloads/builds/elastic-agent-8.13.0+build202403250000-darwin-aarch64/data/elastic-agent-818d20/elastic-agent.app/Contents/MacOS/elastic-agent
Identifier=elastic-agent
Format=app bundle with Mach-O thin (arm64)
CodeDirectory v=20500 size=627737 flags=0x10000(runtime) hashes=19611+2 location=embedded
VersionPlatform=1
VersionMin=720896
VersionSDK=721664
Hash type=sha256 size=32
CandidateCDHash sha256=1bae5b69feccc33542691df9812cad330329c19a
CandidateCDHashFull sha256=1bae5b69feccc33542691df9812cad330329c19af29429fd99cfdfb60d876bb3
Hash choices=sha256
CMSDigest=1bae5b69feccc33542691df9812cad330329c19af29429fd99cfdfb60d876bb3
CMSDigestType=2
Executable Segment base=0
Executable Segment limit=33636352
Executable Segment flags=0x1
Page size=4096
CDHash=1bae5b69feccc33542691df9812cad330329c19a
Signature size=8986
Authority=Developer ID Application: Elasticsearch, Inc (2BT3HPN62Z)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=Mar 25, 2024 at 4:05:37 AM
Info.plist=not bound
TeamIdentifier=2BT3HPN62Z
Runtime Version=11.3.0
Sealed Resources=none
Internal requirements count=1 size=176
```

cmacknz:
Curiously if I download the core binary included in the build directly, it lets me execute it:

```sh
❯ curl -LO https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 19.3M  100 19.3M    0     0  7306k      0  0:00:02  0:00:02 --:--:-- 7304k
~/Downloads/builds ········································································ 02:08:41 PM
❯ tar xvfz ./elastic-agent-core-8.13.0-darwin-aarch64.tar.gz
x elastic-agent-core-8.13.0-darwin-aarch64/
x elastic-agent-core-8.13.0-darwin-aarch64/LICENSE.txt
x elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent
~/Downloads/builds ········································································ 02:08:50 PM
❯ cd elastic-agent-
~/Downloads/builds ·································································· ✘ INT 02:08:53 PM
❯ ./elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent version
Error initializing version information: reading package version from file "/Users/cmackenzie/Downloads/builds/elastic-agent-core-8.13.0-darwin-aarch64/package.version": open /Users/cmackenzie/Downloads/builds/elastic-agent-core-8.13.0-darwin-aarch64/package.version: no such file or directory
Binary: 8.13.0 (build: 818d20468f031265ad7415ef5567b934ffff35c0 at 2024-03-21 19:38:10 +0000 UTC)
Daemon: <failed to communicate>
could not get version. failed to communicate with running daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /Users/cmackenzie/Downloads/builds/elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent.sock: connect: no such file or directory"
Use --binary-only flag to skip trying to retrieve version from running daemon
```

I just can't execute whatever is being included as part of the agent package the new dra targets produce.

pchila:
> Curiously if I download the core binary included in the build directly, it lets me execute it:
> 
> ```shell
> ❯ curl -LO https://staging.elastic.co/8.13.0-382f439d/downloads/elastic-agent-core/elastic-agent-core-8.13.0-darwin-aarch64.tar.gz
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100 19.3M  100 19.3M    0     0  7306k      0  0:00:02  0:00:02 --:--:-- 7304k
> ~/Downloads/builds ········································································ 02:08:41 PM
> ❯ tar xvfz ./elastic-agent-core-8.13.0-darwin-aarch64.tar.gz
> x elastic-agent-core-8.13.0-darwin-aarch64/
> x elastic-agent-core-8.13.0-darwin-aarch64/LICENSE.txt
> x elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent
> ~/Downloads/builds ········································································ 02:08:50 PM
> ❯ cd elastic-agent-
> ~/Downloads/builds ·································································· ✘ INT 02:08:53 PM
> ❯ ./elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent version
> Error initializing version information: reading package version from file "/Users/cmackenzie/Downloads/builds/elastic-agent-core-8.13.0-darwin-aarch64/package.version": open /Users/cmackenzie/Downloads/builds/elastic-agent-core-8.13.0-darwin-aarch64/package.version: no such file or directory
> Binary: 8.13.0 (build: 818d20468f031265ad7415ef5567b934ffff35c0 at 2024-03-21 19:38:10 +0000 UTC)
> Daemon: <failed to communicate>
> could not get version. failed to communicate with running daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /Users/cmackenzie/Downloads/builds/elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent.sock: connect: no such file or directory"
> Use --binary-only flag to skip trying to retrieve version from running daemon
> ```
> 
> I just can't execute whatever is being included as part of the agent package the new dra targets produce.

Did a quick check on the SHAs of elastic agent binary after running the same mage command and compared the SHAs of the elastic-agent-core binaries:

1) elastic agent binary extracted from the generated package
```shell
➜  elastic-agent-8.13.0+build202403250000-darwin-aarch64 git:(add-mage-target-packageDRA) ✗ cd data/elastic-agent-818d20/elastic-agent.app/Contents/MacOS
➜  MacOS git:(add-mage-target-packageDRA) ✗ sha512sum elastic-agent  
672d94b50884b5992b7a9d5fd948a8cb9d8213140a6af3ba3f2702cc60589881d635e5c46857c1e1dc8693f71536d7a8ea7ae6c82bfbaa7062da35683d82dabf  elastic-agent
```
2) elastic agent binary extracted from the `elastic-agent-core` DRA package used in the build before copying/renaming
``` shell
➜  MacOS git:(add-mage-target-packageDRA) ✗ sha512sum ~/dev/elastic-agent/build/golang-crossbuild/elastic-agent-darwin-arm64
672d94b50884b5992b7a9d5fd948a8cb9d8213140a6af3ba3f2702cc60589881d635e5c46857c1e1dc8693f71536d7a8ea7ae6c82bfbaa7062da35683d82dabf  /home/paolo/dev/elastic-agent/build/golang-crossbuild/elastic-agent-darwin-arm64
```
3) elastic-agent binary extracted from the `elastic-agent-core` package downloaded manually using the URL contained in the manifest
```shell
➜  MacOS git:(add-mage-target-packageDRA) ✗ sha512sum ~/Downloads/elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent
672d94b50884b5992b7a9d5fd948a8cb9d8213140a6af3ba3f2702cc60589881d635e5c46857c1e1dc8693f71536d7a8ea7ae6c82bfbaa7062da35683d82dabf  /home/paolo/Downloads/elastic-agent-core-8.13.0-darwin-aarch64/elastic-agent
```

All the hashes match in my case (I packaged on Linux). I am gonna check the same thing when packaging on Mac and see if there's a difference. If all hashes match also there, it means that the problem is not in the binary itself but maybe something to do with permissions/flags

pchila:
> 
> At some point we should probably start adding tests given the growing complexity of the build targets.
> 

Magefiles are not really testable, I agree that we have too much logic in there, maybe it should be moved out in normal packages for testing (devtools package has its own issues)

> It would also be nice to have only one way to package the agent, perhaps one day we can drive the `EXTERNAL=true` case from a simplified version of the DRA manifest file or something.

EXTERNAL=true should behave like fetching all dependencies from the latest snapshot manifest for the "nearest" branch (master, 8.13 or 8.12) but I am not sure if we should support building with older versions of Beats/other dependencies or if the DRA APIs have some restrictions...

cmacknz:
We can change how elastic-agent-core is built if needed, its only purpose is being used by this build target.

pchila:
Quick recap after investigating the MacOS issue with @cmacknz's help:

- `elastic-agent-core` darwin binaries are signed (as single executables)
- when using those binaries inside `elastic-agent.app` directory there's a discrepancy between the binary signature and the bundling as an app we do during packaging (implemented via #714)
- looking at the signature of the repackaged agent binary
  ```shell
  elastic-agent-818d20 git:(add-mage-target-packageDRA) ✗ codesign -dv elastic-agent.app
  Executable=/Users/pchila/dev/elastic-agent/build/distributions/elastic-agent-8.13.0+build202403250000-darwin-aarch64/data/elastic-agent-818d20/elastic-agent.app/Contents/MacOS/elastic-agent
  Identifier=elastic-agent
  Format=app bundle with Mach-O thin (arm64)
  CodeDirectory v=20500 size=627737 flags=0x10000(runtime) hashes=19611+2 location=embedded
  Signature size=8986
  Timestamp=25 Mar 2024 at 09:05:37
  Info.plist=not bound
  TeamIdentifier=2BT3HPN62Z
  Runtime Version=11.3.0
  Sealed Resources=none
  Internal requirements count=1 size=176
  ➜  elastic-agent-818d20 git:(add-mage-target-packageDRA) ✗ codesign --verify elastic-agent.app
  elastic-agent.app: code has no resources but signature indicates they must be present
  ➜  elastic-agent-818d20 git:(add-mage-target-packageDRA) ✗ 
  ```
- here an elastic-agent binary built from source for comparison
  ```shell
  ➜  elastic-agent-8cf031 git:(add-mage-target-packageDRA) ✗ codesign -dv elastic-agent.app
  elastic-agent.app: code object is not signed at all
  ➜  elastic-agent-8cf031 git:(add-mage-target-packageDRA) ✗ codesign --verify elastic-agent.app                                      
  elastic-agent.app: code object is not signed at all
  In architecture: x86_64
  ```
- Having the wrong kind of signature seems to be the cause for the repackaged agent to be quarantined, while no signature at all should work locally (to be tested)

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+	manifestUrl := devtools.ManifestURL
+
+	build, err := manifest.DownloadManifest(manifestUrl)
+	if err != nil {
+		return fmt.Errorf("downloading manifest from %q: %w", manifestUrl, err)
+	}
+
+	parsedVersion, err := version.ParseVersion(build.Version)
+	if err != nil {
+		return fmt.Errorf("parsing version string %q: %w", build.Version, err)
```
AndersonQ, line 1283: [Suggestion]
```suggestion
		return fmt.Errorf("error parsing version string %q: %w", build.Version, err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+		log.Printf("filtered packages: %v", filteredPackages)
+	}
+	return filteredPackages
+}
+
+func downloadDRAArtifacts(ctx context.Context, manifestUrl string, downloadDir string, projects ...string) (map[string]tools.Package, error) {
+
+	build, err := manifest.DownloadManifest(manifestUrl)
+	if err != nil {
+		return nil, fmt.Errorf("downloading manifest from %q: %w", manifestUrl, err)
```
AndersonQ, line 1393:
```suggestion
		return nil, fmt.Errorf("error downloading manifest from %q: %w", manifestUrl, err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+	}
+
+	// Create a dir with the buildID at <downloadDir>/<buildID>
+	draDownloadDir := filepath.Join(downloadDir, build.BuildID)
+	err = os.MkdirAll(draDownloadDir, 0o770)
+
+	if err != nil {
```
AndersonQ, line 1400:
```suggestion
	err = os.MkdirAll(draDownloadDir, 0o770)
	if err != nil {
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+		return nil, fmt.Errorf("creating %q directory: %w", err)
```
AndersonQ, line 1401:
```suggestion
		return nil, fmt.Errorf("error creating %q directory: %w", err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+	}
+
+	if mg.Verbose() {
+		log.Printf("url:%q, err %v", manifestUrl, err)
```
AndersonQ, line 1405: [Blocker]
`err` is `nil`, why are you printing it here?
Also, just printing `url` will likely be confusing, better to write it in full
```suggestion
		log.Printf("manifest url:%q", manifestUrl)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+			log.Printf("build %q project %s packages: %+v", build.BuildID, projectName, project)
+		}
+		// filter down the packages to the platforms we are building/support
+		filteredPackages := filterPackagesByPlatform(project.Packages)
+		if mg.Verbose() {
+			log.Printf("packages to download: %v", filteredPackages)
+		}
+		for pkgName, pkgDesc := range filteredPackages {
+			downloadFunc := func(pkgName string, pkgDesc tools.Package) func() error {
+				return func() error {
```
AndersonQ, line 1429: [Suggestion]
It'd be more straight forward, more clear, to just build a closure instead of a function which returns a function. At 1st glace it looks like you're already calling the function before passing it to `errGrp`

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+					artifactDownloadPath := filepath.Join(draDownloadDir, pkgName)
+					err := manifest.DownloadPackage(errCtx, pkgDesc.URL, artifactDownloadPath)
+
+					if err != nil {
+						return fmt.Errorf("downloading %q: %w", pkgName, err)
```
AndersonQ, line 1434:
```suggestion
					err := manifest.DownloadPackage(errCtx, pkgDesc.URL, artifactDownloadPath)
					if err != nil {
						return fmt.Errorf("downloading %q: %w", pkgName, err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+					}
+
+					// download the SHA to check integrity
+					artifactSHADownloadPath := filepath.Join(draDownloadDir, pkgName+".sha512")
+					err = manifest.DownloadPackage(errCtx, pkgDesc.ShaURL, artifactSHADownloadPath)
+					if err != nil {
+						return fmt.Errorf("downloading SHA for %q: %w", pkgName, err)
```
AndersonQ, line 1441:
```suggestion
					err = manifest.DownloadPackage(errCtx, pkgDesc.ShaURL, artifactSHADownloadPath)
					if err != nil {
						return fmt.Errorf("error downloading SHA for %q: %w", pkgName, err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+					}
+
+					err = download.VerifyChecksum(sha512.New(), artifactDownloadPath, artifactSHADownloadPath)
+					if err != nil {
+						return fmt.Errorf("validating checksum for %q: %w", pkgName, err)
```
AndersonQ, line 1446:
```suggestion
						return fmt.Errorf("error validating checksum for %q: %w", pkgName, err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+					}
+
+					// we should probably validate the signature, it can be done later as we return the package metadata
+					// see https://github.com/elastic/elastic-agent/issues/4445
+
```
AndersonQ, line 1451: [Suggestion]
Then I think it'd be worth adding a warning log so it's clear the signature isn't verified

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+	if err != nil {
+		return fmt.Errorf("looking up for repository root: %w", err)
+	}
+
+	downloadDir := filepath.Join(repositoryRoot, "build", "dra")
+
+	// fetch the agent-core DRA artifacts for the current branch
+	artifacts, err := downloadDRAArtifacts(ctx, manifestUrl, downloadDir, agentCoreProjectName)
+
+	if err != nil {
```
AndersonQ, line 1479:
```suggestion
	artifacts, err := downloadDRAArtifacts(ctx, manifestUrl, downloadDir, agentCoreProjectName)
	if err != nil {
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+		return fmt.Errorf("downloading elastic-agent-core artifacts: %w", err)
```
AndersonQ, line 1480:
```suggestion
		return fmt.Errorf("error downloading elastic-agent-core artifacts: %w", err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+	mg.Deps(EnsureCrossBuildOutputDir)
+
+	// place the artifacts where the package.yml expects them (in build/golang-crossbuild/{{.BeatName}}-{{.GOOS}}-{{.Platform.Arch}}{{.BinaryExt}})
+	for artifactFile, artifactMeta := range artifacts {
+		// uncompress the archive first
+		const extractionSubdir = "extracted"
+		extractDir := filepath.Join(filepath.Dir(artifactFile), extractionSubdir)
+		err = devtools.Extract(artifactFile, extractDir)
+		if err != nil {
+			return fmt.Errorf("extracting %q: %w", artifactFile, err)
```
AndersonQ, line 1492:
```suggestion
			return fmt.Errorf("error extracting %q: %w", artifactFile, err)
```

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+		dstFileName := fmt.Sprintf("elastic-agent-%s", dstPlatform) + binaryExt
+		dstBinaryPath := filepath.Join(repositoryRoot, "build", "golang-crossbuild", dstFileName)
+
+		log.Printf("copying %q to %q", srcBinaryPath, dstBinaryPath)
+
+		err = copy.Copy(srcBinaryPath, dstBinaryPath, copy.Options{
+			PermissionControl: copy.PerservePermission,
+		})
+		if err != nil {
+			return fmt.Errorf("copying %q to %q: %w", srcBinaryPath, dstBinaryPath, err)
```
AndersonQ, line 1528:
```suggestion
			return fmt.Errorf("error copying %q to %q: %w", srcBinaryPath, dstBinaryPath, err)
```

AndersonQ (changes requested):

dev-tools/mage/settings.go
```diff
@@ -44,6 +44,10 @@ const (
 	// Env vars
 	// agent package version
 	agentPackageVersionEnvVar = "AGENT_PACKAGE_VERSION"
+	// Manifest URL to be used for packaging agent
+	ManifestUrlEnvVar = "ManifestURL"
```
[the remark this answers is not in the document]
pchila: This is a pre-existing env var introduced with https://github.com/elastic/elastic-agent/pull/2727 and is used by the early-agent-release pipelines already in place. If we change this, those pipelines will need to change...
@dwhyrock what do you think?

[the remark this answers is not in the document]
dwhyrock: Makes sense to me to change the names to fit the convention.  Once done here, we can put in a PR in `unified-release` to sync up when we pass that env var into the packaging pipeline.

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+
+	if !devtools.PackagingFromManifest {
+		return fmt.Errorf("elastic-agent PackageUsingDRA is expected to build from a manifest. Check that %s is set to a manifest URL", devtools.ManifestUrlEnvVar)
+	}
+
+	manifestUrl := devtools.ManifestURL
+
+	build, err := manifest.DownloadManifest(manifestUrl)
+	if err != nil {
+		return fmt.Errorf("downloading manifest from %q: %w", manifestUrl, err)
```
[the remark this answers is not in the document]
pchila: the failure is implicit in the fact that it's an error object, the string is only for context. I wanted to avoid the pattern:
```
failed thing A: failed nested operation B : <real error here>
```
if we avoid the `failed` we have
```
thing A: nested operation B: <real error here>
```

magefile.go
pchila: same as above

magefile.go
pchila: Side effect of refactoring, will remove

magefile.go
pchila: The signature is never verified anywhere in the packaging process.

This enhancement is already tracked by the issue in the comment, a warning would just appear multiple times in the logs of the early releases pipelines, it's not for any final user just for packagers. I would prefer to add a debug log saying that we verified the signature when the check is implemented

pchila:
> Another thing, why doesn't it create the drop path if it isn't provided as the `package` target does?

The drop path itself is created in the same way (the `collectPackageDependencies` refactored function is common to both target). The drop path is created only if it's not set via env variable and we need to build the dependencies from the `beats` repository: it's an optional action under a condition derived from global or environment variables for a possible way (out of several) of getting components to package within agent.

The main difference with the way the dependencies (components) are handled and this change is that I preferred injecting a target to call to having global variables and ifs. Also the creation of the golang-crossbuild directory has been moved to an independent target so that it will be executed exactly once regardless of the order of execution of the dependencies.

As for the creation of the `build/dra` directory, that is done only when packaging using DRAs, so it's called just before downloading the archives, once the buildID is known.

magefile.go
```diff
@@ -574,13 +589,23 @@ func Update() {
 	mg.SerialDeps(Config, BuildPGP, BuildFleetCfg, Otel.Readme)
 }
 
+func EnsureCrossBuildOutputDir() error {
+	repositoryRoot, err := findRepositoryRoot()
+	if err != nil {
+		return fmt.Errorf("finding repository root: %w", err)
+	}
+	return os.MkdirAll(filepath.Join(repositoryRoot, "build", "golang-crossbuild"), 0o770)
+}
```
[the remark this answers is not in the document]
pchila: Before the crossbuild was created by the `CrossBuild` target directly which wit root privileges (created from inside the docker container).
Now both the Crossbuild target and the PackageUsingDRA targets call this as a dependency so it's created once and with packaging user ownership (we still have a concurrent execution between PackageUsingDRA and CrossbuildGoDaemon but mage ensures that this is executed only once)

magefile.go
pchila: this is a closure that returns a function() (which is what the errGrp expects)
The reason why this is done this way is to create a copy of the iterator variables so that the internal values are not tied to the `for` iterators. When we switch to Go 1.22 we can simplify this.

dev-tools/mage/settings.go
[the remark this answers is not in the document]
pchila: Changed the var name in f8434620bafa91892b5f9743eb28baebfecdda67

cmacknz:
Verified the MacOS artifact works now. Thanks!

When we merge this, it is going to product snapshots and build candidates that depend on the output of the elastic-agent-binary-dra buildkite pipeline that produces the elastic-agent-core artifact.

Looking at the pipeline https://buildkite.com/elastic/elastic-agent-binary-dra I don't think it runs often enough. Main seems to be on a daily schedule, and I'm not sure what triggers 8.13 as it is now several days behind the head of the 8.13 branch.

It looks like this job only takes 5m most of the time, as part of this PR can we change this agent core DRA step to execute for every commit? It is cheap and that will avoid the situation where someone forgets to kick it manually for a build candidate.

Looking at the buildkite pipeline definition it isn't obvious to me why it isn't already working this way for these branches. https://github.com/elastic/elastic-agent/blob/8.13/.buildkite/pipeline.elastic-agent-binary-dra.yml

pchila:
> Verified the MacOS artifact works now. Thanks!
> 
> When we merge this, it is going to make it so that snapshots and build candidates depend on the output of the elastic-agent-binary-dra buildkite pipeline that produces the elastic-agent-core artifact.
> 
> Looking at the pipeline https://buildkite.com/elastic/elastic-agent-binary-dra I don't think it runs often enough. Main seems to be on a daily schedule, and I'm not sure what triggers 8.13 as it is now several days behind the head of the 8.13 branch.
> 
> It looks like this job only takes 5m most of the time, as part of this PR can we change this agent core DRA step to execute for every commit? It is cheap and that will avoid the situation where someone forgets to kick it manually for a build candidate.
> 
> Looking at the buildkite pipeline definition it isn't obvious to me why it isn't already working this way for these branches. https://github.com/elastic/elastic-agent/blob/8.13/.buildkite/pipeline.elastic-agent-binary-dra.yml

There's nothing in the pipeline definition that would limit the `elastic-agent-binary-dra` from running on any release branch or main (we just have a regex identifiying \d+.\d+ branches or exact match with main).
You are right about the schedule as we have a daily on main in the buildkite settings directly.
image
 
I think that the reason why we don't get a build out of every commit is that the option has not been selected on the pipeline Github configuration `Build branches` option in the pic below
image

If we want to build on every commit on main and release branches I think we need to enable that option (not sure if we need to keep the schedule ?)

I would like a second opinion from our eng-prod team before making any changes though...

cmacknz (approved):

magefile.go
```diff
@@ -1154,15 +1216,319 @@ func packageAgent(platforms []string, packagingFn func()) {
...
+
+	branchInfo, err := findLatestBuildForBranch(ctx, baseURLForStagingDRA, branch)
+
+	// Create a dir with the buildID at <root>/build/dra/<buildID>
+	repositoryRoot, err := findRepositoryRoot()
+	if err != nil {
+		return fmt.Errorf("finding repository root: %w", err)
+	}
+	draDownloadDir := filepath.Join(repositoryRoot, "build", "dra")
+	err = os.MkdirAll(draDownloadDir, 0o770)
```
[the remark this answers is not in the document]
AndersonQ: ok, thanks for the clarification

magefile.go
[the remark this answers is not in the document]
AndersonQ: I thought that was your idea. What I intended to avoid with those suggestions is the situation where the errors end up being printed without further explanation and then it could look like an informational log rather than an error.

AndersonQ (approved):

Issue #4393, opened 2024-03-11
[Flaky Test]: TestLogIngestionFleetManaged – unable to create policy: Agent Policy '<redacted>' already exists with name 'test-policy-enroll-1710143889'


rdner:
### Failing test case

TestLogIngestionFleetManaged

### Error message

unable to create policy: Agent Policy '<redacted>' already exists with name 'test-policy-enroll-1710143889'

### Build

https://buildkite.com/elastic/elastic-agent/builds/7667#018e2c73-6325-4b6d-a6e7-5e4c99314aff

### OS

Linux

### Stacktrace and notes

```markdown
=== RUN   TestLogIngestionFleetManaged
    logs_ingestion_test.go:58: Enrolling agent in Fleet with a test policy
    logs_ingestion_test.go:89: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/logs_ingestion_test.go:89
        	Error:      	Received unexpected error:
        	            	unable to create policy: Agent Policy 'd9698871-b94d-41ba-9cda-a6cd1d613f02' already exists with name 'test-policy-enroll-1710143889'
        	Test:       	TestLogIngestionFleetManaged
--- FAIL: TestLogIngestionFleetManaged (4.24s)
```

rdner:
Happens on main too https://buildkite.com/elastic/elastic-agent/builds/7763#018e3960-7db8-47ab-830b-56cd7e6e3e3e

```
=== RUN   TestDebLogIngestFleetManaged
    logs_ingestion_test.go:126: Enrolling agent in Fleet with a test policy
    logs_ingestion_test.go:157: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/logs_ingestion_test.go:157
        	Error:      	Received unexpected error:
        	            	unable to create policy: Agent Policy '5b612cdc-040d-42c2-b3fe-98222d73bf24' already exists with name 'test-policy-enroll-1710361822'
        	Test:       	TestDebLogIngestFleetManaged
--- FAIL: TestDebLogIngestFleetManaged (4.36s)
```

cmacknz:
https://github.com/elastic/elastic-agent/blob/6135d60613964712ae7b574f04b4ad112acecba7/testing/integration/logs_ingestion_test.go#L127-L128

We are using the time since epoch in seconds as the policy name, which is not unique if there are two concurrent requests in the same second, which is not unlikely in our test framework given we run multiple tests concurrently on different VMs.

[...]

PR #4421, opened 2024-03-18
Make policy names unique in integration tests

rdner:
Otherwise we have name collisions and errors from the Fleet server.

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/4393

### DIFF
--- a/testing/integration/container_cmd_test.go
+++ b/testing/integration/container_cmd_test.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 
 	"github.com/elastic/elastic-agent-libs/kibana"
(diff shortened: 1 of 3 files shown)


testing/integration/logs_ingestion_test.go
```diff
@@ -253,8 +253,9 @@ func testFlattenedDatastreamFleetPolicy(
 	policy kibana.PolicyResponse,
 ) {
 	dsType := "logs"
-	dsNamespace := cleanString(fmt.Sprintf("%snamespace%d", t.Name(), rand.Uint64()))
-	dsDataset := cleanString(fmt.Sprintf("%s-dataset", t.Name()))
+	id := uuid.New().String()
+	dsNamespace := cleanString(fmt.Sprintf("namespace-%s", id))
```
cmacknz, line 257: We should be able to just use `info.Namespace` for this, but id is unique so it doesn't matter too much.

https://github.com/elastic/elastic-agent/blob/a7034ef9c4a1eea4b8caaf983c7ed2f2c18468e9/pkg/testing/define/define.go#L53-L57

testing/integration/container_cmd_test.go
```diff
@@ -41,7 +42,7 @@ func TestContainerCMD(t *testing.T) {
 	require.NoError(t, err)
 
 	createPolicyReq := kibana.AgentPolicy{
-		Name:        fmt.Sprintf("test-policy-enroll-%d", time.Now().Unix()),
+		Name:        fmt.Sprintf("test-policy-enroll-%s", uuid.New().String()),
```
cmacknz, line 45: Might be worth creating a helper for this like `UniquePolicyName()` to help people do this properly.

cmacknz (approved):

PR #4436, opened 2024-03-19
Use IgnoreCommas in default config options

fearful-symmetry:
## What does this PR do?

This is the final round of fixes for https://github.com/elastic/go-ucfg/issues/196. This imports the `IgnoreCommas` option added to go-ucfg and places it as a default option.

I'm a little paranoid about this, and open to alternate suggestions as for how to fix this. I'm fairly sure that the only legitimate use case for the "original" behavior is for parsing command line args (probably like `-E list,of,things`), and I don't see anywhere in agent where this is being used. The alternative fix for this would be to make the `IgnoreCommas` option opt-in, but then we have to play whack-a-mole with the dozens of instances of `NewConfigFrom()` and `ToMapStr()` instead.

## Why is it important?

Right now, the config setup process is breaking under certain conditions where a user passes a config that looks like `'a', 'b'`.

## Checklist

### DIFF
--- a/internal/pkg/config/config_test.go
+++ b/internal/pkg/config/config_test.go
@@ -91,7 +91,7 @@ func TestCommaParsing(t *testing.T) {
 	parsedMap, err := c.ToMapStr()
 	require.NoError(t, err)
 	t.Logf("got :%#v", parsedMap)
-	assert.True(t, reflect.DeepEqual(outMap, parsedMap))
+	require.Equal(t, outMap, parsedMap)
 }
 
 func testLoadFiles(t *testing.T) {
--- a/internal/pkg/config/config_test.go
+++ b/internal/pkg/config/config_test.go
@@ -78,6 +78,22 @@ func testToMapStr(t *testing.T) {
 	assert.True(t, reflect.DeepEqual(m, nm))
 }
 
+func TestCommaParsing(t *testing.T) {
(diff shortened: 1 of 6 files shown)


cmacknz (approved):
LGTM, thanks!

internal/pkg/config/config_test.go
```diff
@@ -78,6 +78,22 @@ func testToMapStr(t *testing.T) {
...
+		"test": "startsWith('${testname}','motmot')",
+	}
+	outMap := map[string]interface{}{
+		"test": "startsWith('motmot','motmot')",
+	}
+	c := MustNewConfigFrom(inMap)
+	parsedMap, err := c.ToMapStr()
+	require.NoError(t, err)
+	t.Logf("got :%#v", parsedMap)
+	assert.True(t, reflect.DeepEqual(outMap, parsedMap))
```
AndersonQ, line 94: [Blocker]
Did you use `DeepEqual` because the normal `assert.Equal` won't do the same check?
If so, could you add a message and also the diff? Otherwise if it fails the message will be just something like "false is not true", which isn't helpful at all

AndersonQ (changes requested):
please the the comment with `[Blocker]`

pchila (approved):

AndersonQ (approved):

Issue #3083, opened 2023-07-14
Ubuntu Agent fails to upgrade if shutdown on triggering upgrade and started after 10 minutes.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.9.0 BC4
BUILD: 64661
COMMIT: ddf0c1972e43898b6890ddb38f4c016e96538239
Agent Binary: https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.8.2-linux-x86_64.tar.gz
```

**Host OS and Browser version:** Ubuntu, All

**Preconditions:**
1. 8.9.0 BC4 Kibana cloud environment should be available.
2. Lower version 8.8.2 agent should be installed using Elastic Defend Integration.

**Steps to reproduce:**
1. Navigate to Fleet>Agents tab.
2. Trigger upgrade for lower version agent.
3. Observe the log line- Upgrading agent under agent logs tab.
4. Immediately shutdown the Ubuntu machine before the elastic-agent upgrades.
5. Keep the machine off for 10-12 minutes.
6. Turn the machine ON, and observe agent gets to Updating state.
7. Observe agent gets stuck in updating state.

**Note:**
- The issue is reproducible for multiple Ubuntu agents.
- If we waited for 1 minute before shutdown, it upgrades the elastic-agent to the latest version till then and only Endpoint is left to upgrade.
  So in this scenario, we Stopped the agent to upgrade to latest version.

**Expected:**
Ubuntu Agent should upgrade successfully if shutdown on triggering upgrade and started after 10 minutes.

**Screenshot:**
![image](https://github.com/elastic/elastic-agent/assets/77374876/8cd1fb59-8f38-409e-a5f6-34b86c0ac6f0)

https://github.com/elastic/elastic-agent/assets/77374876/a121a322-d3b7-42a0-8549-457119350388

**Debug Logs:**
[elastic-agent-diagnostics-2023-07-14T08-29-12Z-00.zip](https://github.com/elastic/elastic-agent/files/12048208/elastic-agent-diagnostics-2023-07-14T08-29-12Z-00.zip)

[elastic-agent-diagnostics-2023-07-14T09-01-16Z-00.zip](https://github.com/elastic/elastic-agent/files/12048207/elastic-agent-diagnostics-2023-07-14T09-01-16Z-00.zip)

amolnater-qasource:
@manishgupta-qasource Please review.

amolnater-qasource:
[...]

Issue #4384, opened 2024-03-08
Refactor/Expand elastic-agent diagnostics integration tests


pchila:
As demonstrated by #4376, the current diagnostics test does not cover enough of elastic-agent diagnostics functionality: the current tests run the `elastic-agent diagnostics` command with fake components and without installing agent.

We need add testing for collecting diagnostic bundles from an installed agent running with real components (ideally we should also include endpoint or some other service component) and verify that we collect all the expected files.

Optional: we can think of testing also triggering the diagnostics using Fleet API and then download the archives from there.

PR #4441, opened 2024-03-20
Fix and refactor the state store

AndersonQ:
## Proposed commit message:
```
Title: refactor the state store to use JSON

Body:
Fix action store -> state store migration tests (#4235)
* use golden file for store migration test
* remove withFile function
* migrate take in a storage.Store instead of the storage's path. It's needed so we can set the encrypted store vault's path

refactor state store (#4253)
It modifies the state store API to match the current needs.

update action model to match fleet-server schema (#4240)
* simplify fleetapi.Actions.UnmarshalJSON
* add test to ensure the state store is correctly loaded from disk
* skip state store migration tests, they will be fixes on a follow-up PR as part of #3912

add migrations for action and state stores (#4305)
```

## What does this PR do?

Fixes the agent's state store. This PR is the following PRs combined and a few needed changes to fix the conflicts with `main`.
 - https://github.com/elastic/elastic-agent/pull/4235
 - https://github.com/elastic/elastic-agent/pull/4253
 - https://github.com/elastic/elastic-agent/pull/4240
 - https://github.com/elastic/elastic-agent/pull/4305

As explained [here](https://github.com/elastic/elastic-agent/pull/4441#issuecomment-2049898034), the integration test for this will be done on another issue.

## Why is it important?

Currently, the agent loses data when loading actions from the state store. With the fix, it won't loose data anymore and now the action schema and serialisation are the same for the state store and the fleetapi

## Checklist

reproduce the issue decribed on https://github.com/elastic/elastic-agent/issues/3912. Now it should not fail.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3912
- Closes https://github.com/elastic/elastic-agent/issues/3083

## Questions to ask yourself

- ...


internal/pkg/agent/protection/action.go
```diff
@@ -21,7 +21,7 @@ var (
 )
 
 type fleetActionWithAgents struct {
-	ActionID         string          `json:"action_id"` // Note the action_id here, since the signed action uses action_id for id
```
AndersonQ, line 24: @aleksmaus is it the ES schema or the API schema? The API schema is `id` and now the whole agent also used `id` instead of `action_id`.

aleksmaus: As far as I remember, it is ES field name, and I think the signed JSON payload uses action_id as well.

AndersonQ: So if I understood correctly, for those actions,there is an action inside the action. Is that correct?
[Here](https://github.com/elastic/elastic-agent/pull/4441/files#diff-401f5547d841bfdeff1bf4f32bba334bea2bff7d1147525983e726d601bfe785L64-L65) you're deserialising a whole action from the `data` field.
Is it just a copy of the action or something else?

Another question. Is there any reason for copying the action model to this package instead of using the models from `fleetapi`. Part of this PR consists on unifying action serialisation/deserialisation on the `fleetapi` package.

aleksmaus: Here is some old example I found of the signed action:
```
{
    "@timestamp": "2023-06-12T21:02:08.783Z",
    "expiration": "2023-07-12T21:02:08.783Z",
    "agents": [
        "f19a072f-49c9-4a1d-94f7-9c1c519eab19"
    ],
    "action_id": "164a6819-5c58-40f7-a33c-821c98ab0a8c",
    "data": {
        "version": "8.9.0"
    },
    "type": "UPGRADE",
    "traceparent": "00-baf54d4e5b51487c5c7dd520eef3ba6f-6959cd1c34a1bfd9-01",
    "signed": {
        "data": "eyJAdGltZXN0YW1wIjoiMjAyMy0wNi0xMlQyMTowMjowOC43ODNaIiwiZXhwaXJhdGlvbiI6IjIwMjMtMDctMTJUMjE6MDI6MDguNzgzWiIsImFnZW50cyI6WyJmMTlhMDcyZi00OWM5LTRhMWQtOTRmNy05YzFjNTE5ZWFiMTkiXSwiYWN0aW9uX2lkIjoiMTY0YTY4MTktNWM1OC00MGY3LWEzM2MtODIxYzk4YWIwYThjIiwiZGF0YSI6eyJ2ZXJzaW9uIjoiOC45LjAifSwidHlwZSI6IlVQR1JBREUiLCJ0cmFjZXBhcmVudCI6IjAwLWJhZjU0ZDRlNWI1MTQ4N2M1YzdkZDUyMGVlZjNiYTZmLTY5NTljZDFjMzRhMWJmZDktMDEifQ==",
        "signature": "MEQCIGxsrI742xKL6OSIfT6abkBfH+plFiTSO26RHNV1pHE8AiBxXgt3LlIowCiLZ/ayFqQWk0ge+c/J1L58csqg6R9nyA=="
    }
} 
```

The ```data``` field contains the base64 encoded action JSON that is signed by Kibana. This is the source of truth for the signed actions for the tamper protection for Endpoint for example. The agent initially had signature validation, but it was commented out for the initial release due to risks concerns raised by PM. So the Agent just passes the action through to the beats/Endpoint leaving it up to them to parse and validate the signed payload if it is present.

aleksmaus: Yes it is Kibana only by design for end-to-end signing. Only Kibana has the key(s) to sign the policy and actions. The goal was to prevent any tampering in the middle.

aleksmaus: Would need some good end to end test coverage with Endpoint here in order to make sure the schema changes will not break Endpoint.

AndersonQ: > Would need some good end to end test coverage with Endpoint here in order to make sure the schema changes will not break Endpoint.

You mean future schema changes or the ones in this PR?

The changes here should not have affected the actions schema outside the state/action store. The changes are to match fleet-server schema.

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+
+	actionStoreExists, err := actionDiskStore.Exists()
+	if err != nil {
+		log.Errorf("failed to check if action store %s exists: %v", actionStorePath, err)
+		return err
+	}
+
+	// delete the actions store file upon successful migration
+	defer func() {
+		if err == nil && actionStoreExists {
```
cmacknz, line 52:
```suggestion
		if err == nil {
```

Can you get rid of this check if you put the defer after the `if !actionStoreExists` check below?

internal/pkg/fleetapi/action.go
```diff
@@ -65,45 +68,59 @@ type ScheduledAction interface {
...
+	var action Action
+
+	// keep the case statements alphabetically sorted
+	switch actionType {
+	case ActionTypeCancel:
+		action = &ActionCancel{}
+	case ActionTypeDiagnostics:
+		action = &ActionDiagnostics{}
+	case ActionTypeInputAction:
+		// Only INPUT_ACTION type actions could possibly be signed https://github.com/elastic/elastic-agent/pull/2348
```
cmacknz, line 107: This comment will eventually be out of date if it isn't already. I think. the intent is to sign all of the actions that can result in endpoint be removed from the policy at minimum.

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+			err = actionDiskStore.Delete()
```
cmacknz, line 53: Is the file here being deleted global or stored in the agent data directory? That is, if we roll back during an upgrade can the old version of the agent read this or is it doomed?

cmacknz (commented):
We need to sanity check that tests exist for the various action types the agent supports and that they still work: upgrade, unenroll, settings log level changes, signed actions, scheduled actions, custom actions like osqueries, etc. A lot of code got touched here but not all of it in significant ways.

However it seems like the majority of the impact and change is in the migration from the old YAML format to the new JSON format for the stored version of the policy on disk at startup. It seems to me like if this fails the agent can be left without the policy  it should be running (which would cause data loss and stop all inputs, even endpoint), or under some circumstances can exit early before it ever checks in with Fleet. Assuming I am understanding this correctly, I don't really like this possibility :)

Is there a way to deal with this migration later on in the agent in a more resilient way? What if you made it so the agent could read either the YAML or JSON format at init time, but in check ins only ever wrote out the new JSON format? 

You could potentially create a completely new file name for the JSON version and then prefer that version once it exists. That would leave the old file around previous agent versions in case we need to consider the upgrade roll back case (where the action queue is not in the version data directory).

internal/pkg/agent/protection/action.go
AndersonQ: I tested, the signed actions aren't affected. They still working

aleksmaus: Thank you!

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+	if !actionStoreExists {
+		log.Debugf("action store %s doesn't exists, nothing to migrate", actionStorePath)
+		return nil
+	}
+
+	action, err := migrations.LoadActionStore(actionDiskStore)
+	if err != nil {
+		log.Errorf("failed to create action store %s: %v", actionStorePath, err)
+		return err
+	}
```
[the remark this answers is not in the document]
AndersonQ: >[...] the action store gets deleted when this function returns, meaning the next trip through this function will succeed.
>
>If this is the correct behavior, can we just treat this error as non-fatal and move on?

It only gets deleted on success, so no, it will not succeed on the next run.

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+			"Found %s, but only %s is suported. Ignoring action and proceeding.",
+			action.Type, fleetapi.ActionTypePolicyChange)
+		// If it isn't ignored, the agent will be stuck here and require manual
+		// intervention to fix the store.
+		return nil
+	}
+
+	stateStore, err := NewStateStore(log, stateDiskStore)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
AndersonQ: as above, the action store is deleted on success, `err == nil`:
```
	// delete the actions store file upon successful migration
	defer func() {
		if err == nil {
			err = actionDiskStore.Delete()
			if err != nil {
				log.Errorf("failed to delete action store %s after migration: %v", actionStorePath, err)
			}
		}
	}()
```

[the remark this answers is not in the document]
cmacknz: Yes I read this backwards for some reason.

internal/pkg/agent/storage/store/migrations.go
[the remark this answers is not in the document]
cmacknz: Yes I read the behavior backwards for some reason, but it doesn't work out as nicely for this comment.

This adds a new way for the agent to get stuck at init time, which as mentioned worries me. I'd prefer if we can design this failure mode out of the system.

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// no actions stored nothing to migrate
+	if action == nil {
+		log.Debugf("no action stored in the action store %s, nothing to migrate", actionStorePath)
+		return nil
+	}
+
+	if action.Type != fleetapi.ActionTypePolicyChange {
```
cmacknz, line 78: Why do we ignore everything that isn't a policy change?

cmacknz:
I went to test this manually, and realized that none of our integration tests are actually testing the migration code. We can conclude this because:

1. Actions and the state store require an agent to be enrolled in Fleet to exist.
2. To trigger the migration, you must upgrade from any previously released version to the version in this PR.
3. You can't upgrade to the version in this PR with a Fleet managed agent because it isn't GPG signed.

That actions still work with the changes in this PR can be tested with an agent installed for the first time. I'm not sure how you'd test the migration logic here prior to merging without relying entirely on unit tests.

cmacknz:
Thinking a bit more about how to handle the failure cases, I think the current implementation works with the condition that every error during the migration causes the agent to exit.

The key learning for me is that the migration only happens on upgrade, and the action state store only exists in the versioned home directory.

This means that if we get the agent to exit on failure, it will roll back to the unmodified version of the state store in the original agent, and the failed or corrupted store in the new agent will be removed. We can then try again for deliver a bug fix with a new patch version.

Ideally there would be a test proving it behaves like this, that errors cause a rollback and the original agent is unmodified. This is complicated by the observation above that that we can't test the migration logic in its current form in integration tests at all because until this code is in a snapshot build. 

The other risk to mitigate with this that the mechanical changes to the on disk format have broken or degraded existing actions in some subtle way.

The prior art for this PR is https://github.com/elastic/elastic-agent/pull/398 where we introduced state encryption, which went on to break upgrades https://github.com/elastic/elastic-agent/pull/578. I see similar discussion around `action_id` vs `id` in that PR https://github.com/elastic/elastic-agent/pull/578/files#r901834206, I wonder if there is an explicit test we can build for this interaction with Fleet server.

cmacknz:
Adding @michel-laterman since he has the most context on the Fleet Server action implementation. @michel-laterman the goal here is to make the action store in the agent have the same on disk format as fleet server (JSON with the same schema).

internal/pkg/agent/storage/store/migrations.go
cmacknz: https://github.com/elastic/elastic-agent/pull/4441#issuecomment-2038402088

internal/pkg/agent/storage/store/migrations.go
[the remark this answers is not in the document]
cmacknz: https://github.com/elastic/elastic-agent/pull/4441#issuecomment-2038402088

AndersonQ:
> Ideally there would be a test proving it behaves like this, that errors cause a rollback and the original agent is unmodified. This is complicated by the observation above that that we can't test the migration logic in its current form in integration tests at all because until this code is in a snapshot build. Perhaps a unit test can prove this.

we can do an upgrade to the PR build if we use a custom binary source pointing to a local http server which will serve the PR build.

The issue I see is to force the agent to add actions to the store and then check it after the migration.
The queue is for upgrade actions, which will get cleaned in the upgrade process. But also if the migration breaks, we'll get an error, so perhaps a feasible test is to ensure there are no errors related to the migration.  during the 

> Perhaps a unit test can prove this

For the migration itself, there are already unit test in place for the migration, including tests for the function called by agent: https://github.com/elastic/elastic-agent/pull/4441/files#diff-c4620db6b6efe5946032ce0788dcb60462c88220840108d12e616a7e2f7057abR541-R602

also there are more test for the function which initiates the migration. The only difference between this one and the above is the creation of the encrypted store:
https://github.com/elastic/elastic-agent/pull/4441/files#diff-c4620db6b6efe5946032ce0788dcb60462c88220840108d12e616a7e2f7057abR334-R539

Or do you meant to test more than that? If so, what do you have in mind?

AndersonQ:
I was doing some experimentation, we can even assert the stored action after upgrade. Anyway the agent would not be quite healthy if the state store data is lost. It's just the queue I don't see how to check after the upgrade as the upgrade action is removed from the queue as soon as the upgrade starts.

We do have a `TODO` to address the agent starting without the policy (action policy change) persisted in the state store:
https://github.com/elastic/elastic-agent/blob/4fa6bb67c9f74a22d9145809320714ed4ad165f5/internal/pkg/agent/application/managed_mode.go#L163-L165

cmacknz:
> we can do an upgrade to the PR build if we use a custom binary source pointing to a local http server which will serve the PR build.

You still won't be able to get past the binary signature check. The Fleet upgrade action doesn't support skip verify. This is a common pain point but I haven't thought of a way to get around it that doesn't weaken security.

cmacknz:
We could maybe try to verify this transitively. For example if we:

1. Prove that the agent exiting with an error during an upgrade always leads to a rollback.
2. Prove that the state store migration for any reason always causes the agent to exit with an error.

Then we will have proved that an error during state store migration causes a rollback if it happens during an upgrade, without actually having to test the state store migration in an upgrade by itself.

You will still be left with figuring out how to get the agent to do a state store migration when not upgrading, but I don't think there is any actual dependency on an upgrade. The migration happens if the agent sees the old action store file(s) exist when it starts, it doesn't care how it got there. Can you just artificially set this up? Usually it would happen because the upgrade copied the store files, but you could just do that part manually.

AndersonQ:
> The migration happens if the agent sees the old action store file(s) exist when it starts, it doesn't care how it got there

Exactly, the store migration happens every time the agent starts up. So as long as the new agent reads the old store, the migration will happen. Thus it should not be an issue to set it up. It'd be the following:
 - start an older agent
 - copy the `state.enc` and `fleet.enc`
 - install a new agent (or just move everything to the new agent's folder. That might be easier)
 - stop the new agent
 - override the new agent's `state.enc` and `fleet.enc`
 - start the new agent
 - confirm it's healthy, connected to fleet and all the components are running
 - check `state.enc` to ensure it's in JSON

I'll manually test it now. @cmacknz, do you want it to be an integration test or just the manual test is fine for you?

cmacknz:
> @cmacknz, do you want it to be an integration test or just the manual test is fine for you?

We need an automated test for this. In 8.6 we actually left the unit tests for the encrypted store in place but forget to keep them in the agent code, so we need proof that the agent binary itself can do the migration. https://github.com/elastic/elastic-agent/issues/2249#issuecomment-1425850787 for reference.

cmacknz:
> so we need proof that the agent binary itself can do the migration

Specifically, we need proof that the agent's entrypoint from the run command can do this. If you can do this with a unit test that's fine, it doesn't have to be an integration test, but that would be fine too.

internal/pkg/agent/storage/store/migrations.go
AndersonQ: Because there should not be any other type of action. IIRC the action store would silently reject any action that wasn't a policy change.
Also to avoid silently discarding actions, I [log a warning](https://github.com/elastic/elastic-agent/pull/4441/files#diff-8c3cf214bf832af7042942a6a48cd1a841471b1f99e3d4eb44c2e94b253f0267R80-R82) if it ever happens.

jlind23:
buildkite test it

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,204 @@
...
+
+	stateStoreExists, err := stateDiskStore.Exists()
+	if err != nil {
+		log.Errorf("failed to check if state store exists: %v", err)
+		return err
+	}
+
+	// do not migrate if the state store already exists
+	if stateStoreExists {
+		log.Debugf("not attempting to migrare from action store: state store already exists")
```
cmacknz, line 41:
```suggestion
		log.Debugf("not attempting to migrate from action store: state store already exists")
```

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,204 @@
...
+	log = log.Named("state_migration")
+	actionDiskStore, err := storage.NewDiskStore(actionStorePath)
+	if err != nil {
+		return fmt.Errorf(
+			"could not create disk store when migratins action store: %w", err)
+	}
+
+	stateStoreExists, err := stateDiskStore.Exists()
+	if err != nil {
+		log.Errorf("failed to check if state store exists: %v", err)
```
cmacknz, line 35: The error log statements here should probably be used to wrap the error instead, with the error logged by the caller.

internal/pkg/agent/storage/store/migrations.go
cmacknz: I see it storing Unenroll as well. https://github.com/elastic/elastic-agent/blob/cb95c52546535d00fee6d753ea5c1ebee2c69e0d/internal/pkg/agent/storage/store/state_store.go#L169-L174

I am worried that someone will extend the store with a new action type in the future and forget to update the migrations to handle it. Is there a test we can write to detect this problem? Or should the migrations not be aware of action type restrictions at all?

cmacknz: Or is this only handling actions that were persisted by previous versions of the agent which are known to only write this action type?

If that is the case, a link to the old code that proves this as a comment would help avoid this confusion later.

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,204 @@
...
+	var action fleetapi.Action
+	switch yamlStore.Action.Type {
+	case fleetapi.ActionTypePolicyChange:
+		action = &fleetapi.ActionPolicyChange{
+			ActionID:   yamlStore.Action.ActionID,
+			ActionType: yamlStore.Action.Type,
+			Data: fleetapi.ActionPolicyChangeData{
+				Policy: conv.YAMLMapToJSONMap(yamlStore.Action.Policy)},
+		}
+	case fleetapi.ActionTypeUnenroll:
```
cmacknz, line 159: Similar to above, how can we help people avoid forgetting to add migrations if they extend the store with a new type sometime later?

I do not want this problem to surface only after release.

internal/pkg/agent/storage/store/migrations.go
```diff
@@ -0,0 +1,204 @@
...
+			yamlStore.Action.Type)
+	}
+
+	var queue actionQueue
+	for _, a := range yamlStore.ActionQueue {
+		queue = append(queue,
+			&fleetapi.ActionUpgrade{
+				ActionID:         a.ActionID,
+				ActionType:       a.Type,
+				ActionStartTime:  a.StartTime.Format(time.RFC3339),
```
cmacknz, line 176: Does Format handle the case where the time is empty, in a different format, or completely malformed?

internal/pkg/agent/storage/store/internal/migrations/migrations.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load action store: %w", err)
+	}
+	defer func() {
+		err2 := reader.Close()
+		if err2 != nil {
+			err2 = fmt.Errorf("migration storeLoad failed to close reader: %w", err2)
+		}
+		if err != nil {
+			err = errors.Join(err, err2)
```
cmacknz, line 77: Does failing to close the reader need to be a fatal error?

internal/pkg/fleetapi/action.go
```diff
@@ -641,101 +611,13 @@ func (a *Actions) UnmarshalJSON(data []byte) error {
...
-				ActionID:     n.ActionID,
-				ActionType:   ActionTypeUnknown,
-				originalType: n.ActionType,
-			}
-		}
-		actions = append(actions, action)
-	}
-	*a = actions
-	return nil
+// UnmarshalYAML prevents to decode actions from .
```
cmacknz, line 614:
```suggestion
// UnmarshalYAML prevents to decode actions from .
```

To decode actions from what?

internal/pkg/fleetapi/action.go
```diff
@@ -641,101 +611,13 @@ func (a *Actions) UnmarshalJSON(data []byte) error {
...
+func (a *Actions) UnmarshalYAML(_ func(interface{}) error) error {
+	// TODO(AndersonQ): we need this to migrate the store from YAML to JSON
```
cmacknz, line 616: Is this still relevant? The migration seems to have been written without this.

internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
```diff
@@ -148,10 +148,10 @@ func (h *Diagnostics) collectDiag(ctx context.Context, action *fleetapi.ActionDi
 	cDiag := h.diagComponents(ctx, action)
 
 	var r io.Reader
-	// attempt to create the a temporary diagnostics file on disk in order to avoid loading a
-	// potentially large file in memory.
+	// attempt to create a temporary diagnostics file on disk in order to avoid
+	// loading a potentially large file in memory.
 	// if on-disk creation fails an in-memory buffer is used.
-	f, s, err := h.diagFile(aDiag, uDiag, cDiag, action.ExcludeEventsLog)
+	f, s, err := h.diagFile(aDiag, uDiag, cDiag, action.Data.ExcludeEventsLog)
```
cmacknz, line 154: I don't think we have an integration test for the diagnostics action yet. Now would be a good time to add one, breaking that unintentionally would be Very Bad.

Related: https://github.com/elastic/elastic-agent/issues/4384

I'd suggest implementing it separately on main and then updating this PR to include it, since it should pass in both cases. I would not add it to this already large PR.

cmacknz (commented):
This is a big one, a lot of it is mechanical but there is a lot going on in the state store and it's hard to tell what problems are just inherited from the code that existed before.

I think the state store allowing access to either a single action or a list of scheduled actions is weird. It seems like it should be an action queue that may contain only one action, or only allow storing a single action. It is a mix of both things now.

I also think we should add an integration test for the diagnostics action separately. We have test coverage for policy changes and unenroll actions within the same version of the agent now I believe. 

Do our upgrade tests always implicitly involve migrating the stored policy action from the starting version of the agent? I hope so as that tests the migration for the most common case.

internal/pkg/agent/storage/store/state_store.go
```diff
@@ -104,136 +122,63 @@ func NewStateStore(log *logger.Logger, store storeLoad) (*StateStore, error) {
...
+func (s *StateStore) SetAction(a fleetapi.Action) {
 	s.mx.Lock()
 	defer s.mx.Unlock()
 
 	switch v := a.(type) {
 	case *fleetapi.ActionPolicyChange, *fleetapi.ActionUnenroll:
 		// Only persist the action if the action is different.
-		if s.state.action != nil && s.state.action.ID() == v.ID() {
+		if s.state.ActionSerializer.Action != nil &&
+			s.state.ActionSerializer.Action.ID() == v.ID() {
```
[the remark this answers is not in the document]
michel-laterman: I added the action queue when we wanted to have scheduled upgrades.
IIRC the single-action storage was historically used to store a policy_change action (the latest policy) or unenroll action (so the agent would not contact fleet)

internal/pkg/agent/storage/store/state_store.go
```diff
@@ -269,22 +213,18 @@ func (s *StateStore) Save() error {
...
+		*fleetapi.ActionUnenroll,
+		nil:
+		// ok
+	default:
+		return fmt.Errorf("incompatible type, expected ActionPolicyChange, "+
+			"ActionUnenroll or nil, but received %T", a)
 	}
 
-	reader, err := yamlToReader(&serialize)
+	reader, err := jsonToReader(&s.state)
```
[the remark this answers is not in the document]
AndersonQ: good catch

internal/pkg/agent/storage/store/state_store_test.go
```diff
@@ -6,311 +6,477 @@ package store
...
+		require.NoError(t, err)
+
+		got := store1.Action()
+		require.NotNil(t, got, "store should have an action stored")
+		require.Empty(t, store1.Queue())
+		require.Equal(t, want, got)
+		require.Equal(t, ackToken, store.AckToken())
+	})
+
+	t.Run("errors when saving invalid action type", func(t *testing.T) {
```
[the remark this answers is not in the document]
AndersonQ: done

internal/pkg/fleetapi/action.go
AndersonQ: fixed

internal/pkg/fleetapi/action.go
AndersonQ: nop, fixed

AndersonQ (commented):
There is still one thing I need to check, I put a panic so the tests will fail with a TODO message for myself

internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
cmacknz: It could be if you meet the criteria it proposes. I am mostly interested in ensuring that we can collect diagnostics and send them to fleet here, not verifying the structure of the diagnostics. Re-reading it that issue considers the Fleet diagnostics part optional, I found it via key word search. You can create a new issue with reduced scope if you want.

internal/pkg/agent/storage/store/migrations.go
cmacknz: What I am worried about is we released 8.16.0 with a new action type that hasn't occurred yet so we don't migrate it, then we release 8.16.1 and at the time of upgrade the 8.16.0 agent has the action in the store, but 8.16.1 doesn't know how to handle it.

You are right that this doesn't matter for the first upgrade where old agents can't send the new action yet, but it will matter for every upgrade from that version onward.

cmacknz: I suppose in this situation we will already be the YAML store so we don't need to do a migration at all and everything is fine.

cmacknz:
Are there any implications for being able to downgrade the agent when a migration has to occur? 

This is something Fleet does not allow today, but we want to make it possible: https://github.com/elastic/kibana/issues/172745

internal/pkg/agent/storage/store/state_store.go
```diff
@@ -104,136 +122,63 @@ func NewStateStore(log *logger.Logger, store storeLoad) (*StateStore, error) {
...
+// reuse this.
+func readState(reader io.ReadCloser) (state, error) {
+	st := state{}
 
-	actionStore, err := newActionStore(log, actionDiskStore)
+	data, err := io.ReadAll(reader)
 	if err != nil {
-		log.Errorf("failed to create action store %s: %v", actionStorePath, err)
-		return err
+		return state{}, fmt.Errorf("could not read store state: %w", err)
```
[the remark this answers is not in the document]
cmacknz: Thanks yeah I don't think I followed it correctly the first time. What was tripping me up is [this](https://github.com/elastic/elastic-agent/blob/bb44a11df7a638f84f806357388b65c3c0a6b213/internal/pkg/agent/storage/store/state_store.go#L113-L116):

```go
// NewStateStore creates a new state store.
func NewStateStore(log *logger.Logger, store saveLoader) (*StateStore, error) {
	// If the store exists we will read it, if an error is returned we log it
	// and return an empty store.
	reader, err := store.Load()
```

Specifically that the error handling here is to return an empty store, but further down we don't follow this approach and return an error:

````go
	st, err := readState(reader)
	if err != nil {
		return nil, fmt.Errorf("could not parse store content: %w", err)
	}

	if st.Version != Version {
		return nil, fmt.Errorf(
			"invalid state store version, got %q isntead of %s",
			st.Version, Version)
	}
```

The error handling isn't consistent. Going off the comment, I was thinking the empty `state{}` returned by `readState` would be returned with an unset Version field. 

This makes me wonder if we should be return an empty store on error in all cases, and not just the call to `store.Load()`. Is there a reason for the error handling to be different here?

internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
pchila: As author of #4384 , the main idea behind it is that we don't trigger diagnostic on a real agent with real components. The Fleet part was marked as optional as I didn't see that part of the code exercised anywhere yet in the integration tests.

For reference, all we have for diagnostics in integration tests is triggered with an `elastic-agent diagnostics` command, see:
https://github.com/elastic/elastic-agent/blob/main/testing/integration/diagnostics_test.go
https://github.com/elastic/elastic-agent/blob/8a30900233f4922ff842320bc0da7b2413607536/testing/integration/endpoint_security_test.go#L742

Feel free to keep the issue and make the Fleet part mandatory if you want or create a new one with reduced scope as @cmacknz suggested

AndersonQ:
> Are there any implications for being able to downgrade the agent when a migration has to occur?
> 
> This is something Fleet does not allow today, but we want to make it possible: [elastic/kibana#172745](https://github.com/elastic/kibana/issues/172745)

Yes, the same as for the other migrations, the store isn't designed with a rollback in mind, the old files are deleted and the store does not support a migration rollback.

**However, it isn't a problem in itself.** Considering an agent rollback is intended to replace a broken agent, I'd assume it'll install the older version overriding the current installation. It'd recreate the store from scratch, thus no problem. Also trying to revert the store migration might even be worse, if the issue is on the store migration, it could corrupt the store of the rollback agent. Without extensive thought I see as the best having the agent to recreate its store from scratch.

What we might need is to keep the fleet.enc file/state for the rollback agent, avoiding to perform a new enroll during the rollback.

internal/pkg/agent/storage/store/internal/migrations/migrations.go
```diff
@@ -0,0 +1,97 @@
...
+		}
```
pchila: You can probably join the errors whether err is nil or not
```suggestion
		if err2 != nil {
			err = errors.Join(err, fmt.Errorf("migration storeLoad failed to close reader: %w", err2))
		}
```

internal/pkg/agent/storage/store/internal/migrations/migrations.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, fmt.Errorf("failed to load action store: %w", err)
+	}
+	defer func() {
+		errClose := reader.Close()
+		if errClose != nil {
+			errClose = fmt.Errorf(
+				"migration storeLoad failed to close reader: %w", errClose)
+		}
+		err = errors.Join(err, errClose)
+	}()
```
AndersonQ, line 78: @cmacknz, answering your question:

I tested on windows and as I expected it breaks if it fails to close the reader.
Also the current implementation only returns the error if there is already an error. But I changed it now, better to return it anyway.

here is what happens if the reader isn't closed:
```
--- FAIL: TestStoreMigrations (1.65s)
    --- FAIL: TestStoreMigrations/action_store_to_YAML_state_store (1.65s)
        --- FAIL: TestStoreMigrations/action_store_to_YAML_state_store/unenroll (1.65s)
            migrations_test.go:147:
                        Error Trace:    C:/vagrant/internal/pkg/agent/storage/store/migrations_test.go:147
                        Error:          Received unexpected error:
                                        migration storeLoad failed to close reader: fake reader.Close() error
                        Test:           TestStoreMigrations/action_store_to_YAML_state_store/unenroll
                        Messages:       migration action store -> state store failed
            testing.go:1225: TempDir RemoveAll cleanup: remove C:\Users\vagrant\AppData\Local\Temp\1\TestStoreMigrationsaction_store_to_YAML_state_storeunenroll2169855176\001\action_store.yml: The process cannot access the file because it is being used by another process.
FAIL
FAIL    github.com/elastic/elastic-agent/internal/pkg/a
```

to emulate this error, I just removed the code closing the reader

internal/pkg/agent/storage/store/migrations.go
AndersonQ: > What I am worried about is we released 8.16.0 with a new action type that hasn't occurred yet so we don't migrate it [...]

This won't happen, as soon as this PR is released, there will be no mo migration needed. Unless another refactor is done on the state store. Once it's released, if we add a new action type to be stored, it'll be added on the new store. Besides now there is a version on the store, making it easier to migrate or evolve the store.

AndersonQ: for that to happen, we'd need to release a new action being persisted, then after that have this PR merged without updating it. Which I'd say has a negligible chance of happening, besides most likely changing the store will always result on a merge conflict with this PR, forcing it to be updated.

internal/pkg/agent/storage/store/state_store.go
```diff
@@ -41,51 +40,70 @@ type action = fleetapi.Action
 // Fleet. The store is not thread safe.
 type StateStore struct {
 	log   *logger.Logger
-	store storeLoad
+	store saveLoader
```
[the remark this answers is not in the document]
AndersonQ: I've updated the description

internal/pkg/agent/storage/store/state_store.go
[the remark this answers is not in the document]
AndersonQ: It was like that already, this is the current behaviour so I did not change it. What I remember from my tests before I was on sick leave, the agent would run normally but with an empty policy if the store is empty because something deleted/messed up with the store.

pchila (commented):
A couple of typos and I'm not entirely comfortable with silently making SetAction a no-op if the action type is unexpected...

AndersonQ:
> A couple of typos and I'm not entirely comfortable with silently making SetAction a no-op if the action type is unexpected...

this is the [current behaviour](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/storage/store/state_store.go#L223-L224). I understand your concern and actually agree with you. I just don't think this PR is the right place to change the store's behaviour

pchila (approved):

Issue #4442, opened 2024-03-20
[Flaky Test]: TestStandaloneUpgrade – Failed trigger upgrade of daemon: rpc error: code = Unavailable desc = error reading from server: EOF


rdner:
### Failing test case

TestStandaloneUpgrade

### Error message

Failed trigger upgrade of daemon: rpc error: code = Unavailable desc = error reading from server: EOF

### Build

https://buildkite.com/elastic/elastic-agent/builds/7867#018e5927-80f8-4c1e-bd1a-82ba1b92ce6d

### OS

Linux

### Stacktrace and notes

```markdown
fixture.go:656: >> running binary with: [/tmp/TestStandaloneUpgradeUpgrade_8.14.0-SNAPSHOT_to_8.14.0-SNAPSHOT_(privileged)3269148483/001/elastic-agent-8.14.0-SNAPSHOT-linux-arm64/elastic-agent install --force --non-interactive]
    fixture.go:656: >> running binary with: [/opt/Elastic/Agent/elastic-agent status --output json]
    upgrader.go:289: Upgrading from version "8.14.0-SNAPSHOT-1d84d2091343549ac3b02e2a1b6cb054563ccbce" to version "8.14.0-SNAPSHOT-ddd832b4c3a16296ff6d0b2693629de42a375446"
    fixture.go:656: >> running binary with: [/opt/Elastic/Agent/elastic-agent upgrade 8.14.0-SNAPSHOT --source-uri file:///home/ubuntu/agent/.agent-testing/local --skip-verify]
    upgrade_standalone_test.go:83: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_standalone_test.go:83
        	            				/home/ubuntu/agent/testing/integration/upgrade_standalone_test.go:49
        	Error:      	Received unexpected error:
        	            	failed to start agent upgrade to version "8.14.0": exit status 1
        	            	Error: Failed trigger upgrade of daemon: rpc error: code = Unavailable desc = error reading from server: EOF
        	            	For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.14/fleet-troubleshooting.html
        	Test:       	TestStandaloneUpgrade/Upgrade_8.14.0-SNAPSHOT_to_8.14.0-SNAPSHOT_(privileged)
[...]

PR #4448, opened 2024-03-20
Fix fixture diagnostics unconditionally logging an error.

cmacknz:
Observed in https://github.com/elastic/elastic-agent/issues/4442#issuecomment-2010403202

We are unconditionally logging an error even when diagnostics were successfully collected on a retry.


blakerouse (approved):
Looks good.

pchila (approved):

Issue #4447, opened 2024-03-20
[Flaky Test]: TestLongRunningAgentForLeaks/TestHandleLeak – Condition never satisfied


rdner:
### Failing test case

TestLongRunningAgentForLeaks/TestHandleLeak

### Error message

Condition never satisfied

### Build

https://buildkite.com/elastic/elastic-agent/builds/7876#018e5c3a-00ae-4972-b890-37cad38f2a96

### OS

Windows

### Stacktrace and notes

```markdown
agent_long_running_leak_test.go:232: component state: Healthy: communicating with pid '3216'
    agent_long_running_leak_test.go:223: unit ID: winlog-fleet-default-output
    agent_long_running_leak_test.go:223: unit ID: winlog-fleet-default-output-winlog-system-e3f438c4-571c-44b0-a8d5-25e4ea556eef
    agent_long_running_leak_test.go:232: component state: Healthy: communicating with pid '2344'
    agent_long_running_leak_test.go:210: 
        	Error Trace:	C:/Users/windows/agent/testing/integration/agent_long_running_leak_test.go:210
        	            				C:/Users/windows/agent/testing/integration/agent_long_running_leak_test.go:142
        	Error:      	Condition never satisfied
        	Test:       	TestLongRunningAgentForLeaks/TestHandleLeak
        	Messages:   	install never became healthy
--- FAIL: TestLongRunningAgentForLeaks/TestHandleLeak (180.00s)
```

rdner:
It failed in the daily build too, perhaps it's a real leak then https://buildkite.com/elastic/elastic-agent/builds/7884#018e5e50-9390-41d6-855d-2979484718c7

rdner:
I skipped the tests for now to avoid more failures, needs to be investigated ASAP
https://github.com/elastic/elastic-agent/pull/4452

cmacknz:
Are we sure this is a bug in this test, and it is not that this test is finding a real bug in the underlying inputs it is testing?

rdner:
@cmacknz it can be a real failure, we need to properly investigate of course. The most concerning fact is that it started failing without any connection to a new change in the code.

It first started failing after https://github.com/elastic/elastic-agent/issues/4447 was merged. Initially, I thought this change introduced something that fails the test, despite the PR being green.

However, then I opened a new PR to revert this change https://github.com/elastic/elastic-agent/pull/4451 and in that PR the leak tests also failed which proves that the failure is not related to the change but just came out of nowhere.

[...]

PR #4454, opened 2024-03-21
Re-enable resource leak tests, change timeout, improve message

fearful-symmetry:
## What does this PR do?

Attempt to fix https://github.com/elastic/elastic-agent/issues/4447.

This increases the timeout on the health check, and improves the error message. As to why the test is flaky, my best guess right now is that the 3-minute timeout just isn't long enough on windows.

### DIFF
--- a/testing/integration/agent_long_running_leak_test.go
+++ b/testing/integration/agent_long_running_leak_test.go
@@ -78,7 +78,7 @@ func TestLongRunningAgentForLeaks(t *testing.T) {
 	}
 
 	suite.Run(t, &ExtendedRunner{info: info,
-		healthCheckTime:        time.Minute * 3,
+		healthCheckTime:        time.Minute * 6,
 		healthCheckRefreshTime: time.Second * 20,
 		resourceWatchers: []StatusWatcher{ // select which tests to run
 			&handleMonitor{},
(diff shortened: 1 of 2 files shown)


cmacknz:
In case you are missing the discussion in Slack, there was a change in 8.14.0-SNAPSHOT in Fleet that changed the output name from `default` that is probably what is causing this failure and the one in https://github.com/elastic/elastic-agent/issues/4082

rdner (approved):

testing/integration/agent_long_running_leak_test.go
```diff
@@ -253,11 +254,12 @@ func (runner *ExtendedRunner) CheckHealthAtStartup(ctx context.Context) {
 			}
 			runner.T().Logf("component state: %s", comp.Message)
 			if comp.State != int(cproto.State_HEALTHY) {
+				compDebugName = comp.Name
 				allHealthy = false
 			}
 		}
 		return allHealthy && foundApache && foundSystem
-	}, runner.healthCheckTime, runner.healthCheckRefreshTime, "install never became healthy")
+	}, runner.healthCheckTime, runner.healthCheckRefreshTime, "install never became healthy: components did not return a healthy state: %s", compDebugName)
```
pchila, line 262: Nit: printing out the entire last agent status (in a structured way) may be more useful for debugging

PR #4451, opened 2024-03-21
Revert "Add metrics-monitoring beats to resource monitoring"

rdner:
Reverts elastic/elastic-agent#4326

These changes caused multiple failures on the CI:

* https://buildkite.com/elastic/elastic-agent/builds/7876
* https://buildkite.com/elastic/elastic-agent/builds/7884

In both builds ` TestLongRunningAgentForLeaks/TestHandleLeak` fails with condition never satisfied (see https://github.com/elastic/elastic-agent/issues/4447)

`TestMetricsMonitoringCorrectBinaries/TestBeatsMetrics` fails with `Condition never satisfied`.

PR #4462, opened 2024-03-21
Revert removal of metrics-monitoring change, fix tests to account for fleet naming changes

fearful-symmetry:
## What does this PR do?
reverts https://github.com/elastic/elastic-agent/pull/4451 and changes the tests to account for recent changes in https://github.com/elastic/kibana/pull/178857

We're not 100% sure if the change to name generation will remain, so the test is designed in such a way that we can make a 1-line change if the way kibana creates the output ID changes in the future.


testing/integration/metrics_monitoring_test.go
```diff
@@ -0,0 +1,147 @@
...
+	policyResp, err := tools.InstallAgentWithPolicy(ctx, runner.T(), installOpts, runner.agentFixture, runner.info.KibanaClient, basePolicy)
+	require.NoError(runner.T(), err)
+
+	_, err = tools.InstallPackageFromDefaultFile(ctx, runner.info.KibanaClient, "system", "1.53.1", "system_integration_setup.json", uuid.New().String(), policyResp.ID)
+	require.NoError(runner.T(), err)
+
+}
+
+func (runner *MetricsRunner) TestBeatsMetrics() {
+	UnitOutputName := "fleet-default-output"
```
cmacknz, line 84: Fleet reverted the change, it just isn't in the latest snapshot yet: https://github.com/elastic/kibana/commit/3069efd3875b7b8bd8855caf199f848df595265e

Don't merge until that happens or you'll just be undoing this again.

cmacknz (approved):

PR #2787, opened 2023-06-06
Fix race detector errors in TestFQDNFeatureFlagToggle, TestContextProvider

faec:
Fix race detector errors in `TestFQDNFeatureFlagToggle` and `TestContextProvider` by removing a leaking error variable and reworking the test listeners to use channels.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] ~~I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- [ ] ~~I have added an integration test or an E2E test~~

## How to test this PR locally

Running `go test -race` these tests previously produced consistent race detector errors. With this PR they pass.

## Related issues

- Followup from https://github.com/elastic/elastic-agent/pull/2743

PR #4473, opened 2024-03-22
Insert ES hosts from policy to fleet-server output

### LINKED ISSUE #2784: Fleet Server configuration does not contain all the hosts available in the Elasticsearch cluster
- Version: 8.7.1 (and probably newer versions)
- Operating System: Linux
- Discuss Forum URL: N/A

Following the instructions to enroll a new Fleet Server, if the host used in the install command is down or stopped, fleet-server becomes unhealhty (and all the agents connected to it).

## Steps to Reproduce:

- Setup a Elastic stack cluster with 3 Elasticsearch hosts and 1 Kibana.
    - Ensure that the 3 ES hosts are accessible to port 9200
        - In this example, these ES hosts are accessible through:
            - https://10.13.233.1:9200
            - https://10.13.233.1:9201
            - https://10.13.233.1:9202
    - Ensure that Kibana has configured the 3 ES hosts in its configuration.
- Add in the Fleet Settings, all three hosts available in the Output
  fleet output configuration

- Create a new VM. In this VM, enroll a new fleet-server using the command suggested by the UI in another VM.
    - NOTE: there is just one ES host set in the `--fleet-server-es` parameter
    - Added some parameters to avoid issues with SSL/TLS
      ```bash
      sudo ./elastic-agent install \
        --fleet-server-es=https://10.13.233.1:9200 \
        --fleet-server-es-ca=/home/ubuntu/ca/ca.crt \
        --fleet-server-es-insecure \
        --insecure \
        --fleet-server-service-token=<token> \
        --fleet-server-policy=fleet-server-policy
      ```
- Create another VM. In this VM, enroll a new agent following the instructions too.
    - Adding `--insecure` to avoid issues with SSL/TLS
      ```
      sudo ./elastic-agent install \
        --url=https://fleetserver:8220 \
        --enrollment-token=<token> \
        --insecure
      ```
- Confirm that there are logs in Elasticsearch
  logs from fleet server and agent

At this point, this new fleet-server should be shown in the Fleet view along with the agents:
fleet healthy

In this stage, if any ES host different to the one used in the `install` command is stopped, there are no issues with fleet-server or any agent that has been installed:
- Agents and fleet server are healthy
- Agents and Fleet server are ingesting logs

However, if the ES host (`https://10.13.233.1:9200`) used in the install command (`--fleet-server-es`) is down:
- fleet-server becomes **unhealhty** as well as the agents.
- both fleet-server and agents are still sending data (logs) to Elasticsearch

unhealthy fleet server and agents

logs still are sent

Output of the `elastic-agent status`:
```
$ sudo elastic-agent status
State: HEALTHY
Message: Running
Fleet State: FAILED
Fleet Message: status code: 503, fleet-server returned an error: ServiceUnavailable, message: Fleet server unable to communicate with Elasticsearch
Components:
  * beat/metrics    (HEALTHY)
                    Healthy: communicating with pid '32799'
  * http/metrics    (HEALTHY)
                    Healthy: communicating with pid '32798'
  * system/metrics  (HEALTHY)
                    Healthy: communicating with pid '32761'
  * log             (HEALTHY)
                    Healthy: communicating with pid '32768'
  * fleet-server    (HEALTHY)
                    Healthy: communicating with pid '32714'
  * filestream      (HEALTHY)
                    Healthy: communicating with pid '32773'
```

Thanks to this PR https://github.com/elastic/fleet-server/pull/2632, when the fleet-server was healthy, a diagnostics bundle was generated and in the fleet-server configuration there is just one Elasticsearch host (the one used in the install command).

```yaml
fleet:
    agent:
        id: e1a89fa0-5c2a-41ed-a267-b615ebb34f92
        version: 8.7.1
        logging:
            level: info
    host:
        id: ""
        name: ""
output:
    elasticsearch:
        protocol: https
        hosts:
            - 10.13.233.1:9200
        path: ""
        headers: {}
        apikey: ""
        servicetoken: '[redacted]'
        proxyurl: ""
        proxydisable: false
        proxyheaders: {}
        tls:
            verification_mode: none
            certificate_authorities:
                - /home/ubuntu/ca/ca.crt
            renegotiation: never
        maxretries: 3
        maxconnperhost: 128
        timeout: 1m30s
        maxcontentlength: 104857600
...
```

michel-laterman:
## What does this PR do?

When running the fleet-server component, elastic-agent will insert
all Elasticsearch hosts from the policy to the component's expected
config. These hosts are merged with the local values that are
provided when the agent/server is first enrolled.

## Why is it important?

We are not able to change the ES hosts that the fleet-server component uses without re-enrollment.

## Checklist

1. Enroll agent with a policy that contains the fleet-server integration.
2. Add another URL to the elastisearch output associated with that policy.
3. Gather diagnostics bundle from agent and inspect `fleet-server.yml`, it will now contain the additional host

## Related issues

- Closes #2784

### DIFF
--- a/internal/pkg/agent/application/fleet_server_bootstrap.go
+++ b/internal/pkg/agent/application/fleet_server_bootstrap.go
@@ -155,7 +155,7 @@ func injectESHosts(dst, src map[string]interface{}) error {
 		}
 	}
 	hosts := make([]interface{}, 0, len(m))
-	for host, _ := range m {
+	for host := range m {
 		hosts = append(hosts, host)
 	}
 	dst["hosts"] = hosts
--- a/internal/pkg/agent/application/fleet_server_bootstrap.go
+++ b/internal/pkg/agent/application/fleet_server_bootstrap.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 
 	"gopkg.in/yaml.v2"
 
@@ -26,6 +27,9 @@ const (
 	endpoint      = "endpoint"
 	apmServer     = "apm"
 )
+
+// schemaReg is a regular expression to match the prefix schema for a host
+var schemaReg = regexp.MustCompile(`^https?://`)
 
 // ErrFleetServerNotBootstrapped set on fleet-server component and units when the Elastic Agent has not been
 // bootstrapped with the required command-line arguments for the Elastic Agent to be able to run the Fleet Server.
@@ -50,7 +54,7 @@ var injectFleetServerInput = config.MustNewConfigFrom(map[string]interface{}{
 
 // FleetServerComponentModifier modifies the comps to inject extra information from the policy into
 // the Fleet Server component and units needed to run Fleet Server correctly.
-func FleetServerComponentModifier(serverCfg *configuration.FleetServerConfig) coordinator.ComponentsModifier {
+func FleetServerComponentModifier(serverCfg *configuration.FleetServerConfig, log *logger.Logger) coordinator.ComponentsModifier {
 	return func(comps []component.Component, _ map[string]interface{}) ([]component.Component, error) {
 		for i, comp := range comps {
 			if comp.InputSpec != nil && comp.InputSpec.InputType == fleetServer && comp.Err == nil {
@@ -65,11 +69,15 @@ func FleetServerComponentModifier(serverCfg *configuration.FleetServerConfig) co
 				} else {
 					for j, unit := range comp.Units {
 						if unit.Type == client.UnitTypeOutput && unit.Config.Type == elasticsearch {
-							unitCfgMap, err := toMapStr(unit.Config.Source.AsMap(), &serverCfg.Output.Elasticsearch)
+							uMap := unit.Config.Source.AsMap()
+							unitCfgMap, err := toMapStr(uMap, &serverCfg.Output.Elasticsearch)
 							if err != nil {
 								return nil, err
 							}
 							fixOutputMap(unitCfgMap)
+							if err := injectESHosts(unitCfgMap, uMap); err != nil {
+								log.With("error.message", err).Warn("Unable to inject hosts from unit.")
+							}
 							unitCfg, err := component.ExpectedConfig(unitCfgMap)
 							if err != nil {
 								return nil, err
@@ -98,6 +106,60 @@ func FleetServerComponentModifier(serverCfg *configuration.FleetServerConfig) co
 		}
 		return comps, nil
 	}
+}
+
+// injectESHosts will take the hosts list in src and add them to the hosts list in dst if they are not there.
+// Items in the hosts list will not have the http(s) schema prefix.
+// If src does not have an api_key attribute the function is a nop.
+// api_key attribute is missing when src is not retrieved from fleet.
+func injectESHosts(dst, src map[string]interface{}) error {
+	if _, ok := src["api_key"]; !ok {
+		return nil
+	}
+
+	// get src hosts as a string slice
+	srcAttr, ok := src["hosts"]
+	if !ok {
+		return fmt.Errorf("source does not contain hosts attribute")
+	}
+	srcList, ok := srcAttr.([]interface{})
+	if !ok {
+		return fmt.Errorf("source hosts attribute is not []interface{}: type=%T", srcAttr)
+	}
+
+	//get dst hosts as a string slice
+	dstAttr, ok := dst["hosts"]
+	if !ok {
+		return fmt.Errorf("destination does not contain hosts attribute")
+	}
+	dstList, ok := dstAttr.([]interface{})
+	if !ok {
+		return fmt.Errorf("destination hosts attribute is not []inteface{}: type=%T", dstAttr)
+	}
+
+	m := make(map[string]struct{}, len(srcList)+len(dstList))
+	for i, o := range srcList {
+		if host, ok := o.(string); ok {
+			host = schemaReg.ReplaceAllString(host, "")
+			m[host] = struct{}{}
+		} else {
+			return fmt.Errorf("unable to cast source host item %d as string: type=%T", i, o)
+		}
+	}
+	for i, o := range dstList {
+		if host, ok := o.(string); ok {
+			host = schemaReg.ReplaceAllString(host, "")
+			m[host] = struct{}{}
+		} else {
+			return fmt.Errorf("unable to cast destination host item %d as string: type=%T", i, o)
+		}
+	}
+	hosts := make([]interface{}, 0, len(m))
+	for host, _ := range m {
+		hosts = append(hosts, host)
+	}
+	dst["hosts"] = hosts
+	return nil
 }
 
 // InjectFleetConfigComponentModifier The modifier that injects the fleet configuration for the components
--- a/internal/pkg/agent/application/fleet_server_bootstrap_test.go
+++ b/internal/pkg/agent/application/fleet_server_bootstrap_test.go
@@ -23,11 +23,13 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/remote"
 	"github.com/elastic/elastic-agent/internal/pkg/testutils"
 	"github.com/elastic/elastic-agent/pkg/component"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
 func TestFleetServerComponentModifier_NoServerConfig(t *testing.T) {
 	cfg := map[string]interface{}{}
-	modifier := FleetServerComponentModifier(nil)
+	log, _ := logger.NewTesting("test")
+	modifier := FleetServerComponentModifier(nil, log)
 	fleetServerInputSource, err := structpb.NewStruct(map[string]interface{}{
 		"id":   "fleet-server",
 		"type": "fleet-server",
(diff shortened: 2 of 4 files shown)


internal/pkg/agent/application/fleet_server_bootstrap.go
```diff
@@ -100,6 +108,60 @@ func FleetServerComponentModifier(serverCfg *configuration.FleetServerConfig) co
 	}
 }
 
+// injectESHosts will take the hosts list in src and add them to the hosts list in dst if they are not there.
```
michel-laterman, line 111: Open question: Should we keep the original host that is part of `dst` or just insert the hosts from `src` (if `src` is valid)?

michel-laterman (commented):
Another thing we can do as part of this PR, or as a follow up would be to allow users to specify multiple ES hosts on initial enrolment

internal/pkg/agent/application/fleet_server_bootstrap.go
cmacknz: If the user wants `dst` to be completely removed then we should do it, otherwise you are creating a bug where Fleet server maintains a host that possibly isn't valid anymore with no way to remove it.

Before replacing hosts, we should verify that at least one of them is actually valid to save users from bricking their system with simple typos and such.

This is the approach taken in the policy change handler when changing the agent's Fleet connection.

https://github.com/elastic/elastic-agent/blob/c9bb1643c6817879545c9200e1d3d00c02cd3287/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go#L192-L198

This is probably difficult to do in a component modifier, which leads me to the next question of why we need to do this in a component modifier?

Why can't this be done by updating Fleet server's output unit to include the new hosts?

internal/pkg/agent/application/fleet_server_bootstrap.go
```diff
@@ -100,6 +108,60 @@ func FleetServerComponentModifier(serverCfg *configuration.FleetServerConfig) co
...
+	}
+	dstList, ok := dstAttr.([]interface{})
+	if !ok {
+		return fmt.Errorf("destination hosts attribute is not []inteface{}: type=%T", dstAttr)
+	}
+
+	m := make(map[string]struct{}, len(srcList)+len(dstList))
+	for i, o := range srcList {
+		if host, ok := o.(string); ok {
+			host = schemaReg.ReplaceAllString(host, "")
```
cmacknz, line 143: Why do you need a regex to do this? Could you use https://pkg.go.dev/net/url#Parse and then just override the `Scheme` member?

cmacknz:
The linked issue https://github.com/elastic/elastic-agent/pull/2787 is incorrect

internal/pkg/agent/application/fleet_server_bootstrap.go
michel-laterman: I'm not super familiar with the agent codebase; is there somewhere else that is better suited to insert the output block?

As I understand, the component modifier  (currently) inserts local config (passed through enrollment flags) into the policy; one of these flags is the hosts list. But it also may include tls/proxy information

internal/pkg/agent/application/fleet_server_bootstrap.go
michel-laterman: I did a quick test of this; If you parsed an input that's `https://HOST`, and replaced `u.Schema` with an empty string it results in `//HOST`

cmacknz: I had to remind myself about exactly what the problem was. We are in the right part of the code.

https://github.com/elastic/elastic-agent/issues/2784#issuecomment-1578741803

> Yes I forgot. Elasticsearch output configuration for the Fleet Server only every uses what it was bootstrapped with. It does not use the configuration that comes from the policy. It has been that way since it was implemented.

So this is the root of our problem and we need to decide how to fix this in a non-breaking way. The way it is designed now, the Fleet server output configuration is always overridden by the agent to be the output it was bootstrapped with:

https://github.com/elastic/elastic-agent/blob/c9bb1643c6817879545c9200e1d3d00c02cd3287/internal/pkg/agent/application/fleet_server_bootstrap.go#L67-L77

The simplest possible mitigation for this would be to remove that block completely, and just take the Fleet server outputs directly from the policy.

However I think this could be a breaking change for users who have created outputs that work for everything except Fleet server. That is, there could be somebody out there who is relying on the bootstrap output never changing and Fleet server ignoring the updated output in Fleet. We don't want to break their setups with an unfortunate surprise. For example, the output in Fleet could have a proxy configuration assuming its the output for Beats/Agent while fleet-server is deployed behind the proxy.

In particular if we look at the definition of a Fleet server output (which almost but not quite matches the [configuration in fleet-server](https://github.com/elastic/fleet-server/blob/fbb07ee59549179a3772149bf350ee2722705bb6/internal/pkg/config/output.go#L36) itself):

https://github.com/elastic/elastic-agent/blob/c9bb1643c6817879545c9200e1d3d00c02cd3287/internal/pkg/agent/configuration/fleet_server.go#L36-L48

What we are doing is extending the `hosts` list assuming that the rest of the bootstrap output configuration entries are still right (for example the TLS section). This isn't necessarily true, and at least if we keep doing it that way we've only allowed managing the `hosts` list from Fleet. Even though you can change more parameters than that in the UI they all get silently ignored. I don't think this is completely correct.

We have a similar situation to this in the agent when we introduced managing proxy parameters from Fleet **after* we had let users put them on the command line. The way we dealt with it was by letting the Fleet proxy replace the command line one only if we confirmed it still allowed us to connect to Fleet, otherwise [we fall back](https://github.com/elastic/elastic-agent/blob/c9bb1643c6817879545c9200e1d3d00c02cd3287/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go#L168-L197) to what was set before.

I think we'd want to do something similar here, where fleet-server uses the outputs set in the Fleet UI and ignores the bootstrap output completely only if confirms the outputs from Fleet UI work.

The trouble is to do this you have to send both outputs to Fleet server, and mark one as being the bootstrap output (or move the original bootstrap output under a `bootstrap` key in the single output configuration or something) so it can do the swap correctly.

I think that makes this more complicated, but I worry if we don't do it this way the implementation will be both incomplete (we ignore everything but `hosts`) and possibly break some users that implicitly depend on the outputs in Fleet UI being ignored by Fleet server.

michel-laterman: One thing to note is the output block modification:

https://github.com/elastic/elastic-agent/blob/c9bb1643c6817879545c9200e1d3d00c02cd3287/internal/pkg/agent/application/fleet_server_bootstrap.go#L67-L77
applies to fleet-server outputs only (and not any other component's elasticsearch output). It's required for bootstrapping where the initial "policy" from fleet-server that is used lists default values (which is why my modification method checks to see if an `api_key` is present as that indicates it's from fleet).

For example, if I add logging to the component modifier we can see the initial value of `unit.Config.Source.AsMap()` is:

```json
{
  "hosts":["localhost:9200"],
  "type":"elasticsearch"
}
```

And when it's retrieved from fleet we see:

```json
{
  "api_key":"REDACTED",
  "hosts":["https://HOST1:PORT","https://HOST2:PORT"],
  "preset":"balanced",
  "type":"elasticsearch"
}
```

If tls or proxy settings are available we see them in the object as well.

I don't think we want to send multiple output blocks to fleet-server as then fleet policies will need to be aware of things like mTLS usage and proxies.

I think we can adjust the output modification to assemble the output block using the retrieved policy and local settings.
In order to assemble the output on elastic-agent we can check if the policy from fleet contains tls/proxy settings and use the only if the keys are present; otherwise fallback to using the local settings.
For example if the local settings are:
```json
{
  "hosts":["HOST:9200"],
  "type":"elasticsearch",
  "ssl":{"certificate_authorities": "/path/to/ca.crt"},
  "proxy_url":"http://PROXY:PORT"
}
```
And it retrieves
```json
{
  "hosts":["OTHER:9200"],
  "type":"elasticsearch",
  "ssl":{"verification_mode": "none"}
}
```

It will create and test:
```json
{
  "hosts":["OTHER:9200"],
  "type":"elasticsearch",
  "ssl":{"verification_mode": "none"},
  "proxy_url":"http://PROXY:PORT"
}
```
If it's able to connect to elasticsearch the output block is passed to fleet-server; otherwise the block from local settings are used.
We would need to be able to persist the new block if it succeeds in order to properly handle restarts

internal/pkg/agent/application/fleet_server_bootstrap_test.go
```diff
@@ -141,6 +143,51 @@ func TestInjectFleetConfigComponentModifier(t *testing.T) {
...
+	resComps, err := modifier(comps, nil)
+	require.NoError(t, err)
+	t.Logf("RESCOMPS %v", resComps)
+	require.Len(t, resComps, 1)
+	require.Len(t, resComps[0].Units, 1)
+
+	resCfg := resComps[0].Units[0].Config.GetSource().AsMap()
+	assert.NotContains(t, resCfg, "api_key")
+	assert.Contains(t, resCfg, "hosts")
+	assert.Len(t, resCfg["hosts"], 2)
```
AndersonQ, line 188: [Blocker]
Why didn't you assert the new `hosts` have the right hosts there? Currently if there is any bug which causes the wrong hots to be used, as log as it's the right amount of hosts the test will not catch it.

internal/pkg/agent/application/fleet_server_bootstrap.go
AndersonQ: one question/thing to consider here. In the case the ES hosts in the policy aren't reachable and the agent keeps the cli configuration, what will be the expected behaviour? A unhealthy agent with a healthy fleet-server component?
Or as it's the bootstrap, it just fails all together?

AndersonQ (changes requested):
I believe there is an issue in the test, I commented there

internal/pkg/agent/application/fleet_server_bootstrap.go
cmacknz: Another way to model this (and perhaps the way it was intended) is to treat it like a bootstrap broker in Kafka, where you manually provide a subset of the total available broker hosts which are used to discover the others.

In this situation, we wouldn't change anything for the agent. It would continue to pass Fleet server the bootstrap output. What would change is in Fleet server, where it internally looks up the other outputs configured for itself and merges them with the bootstrap output as long as doing that maintains connectivity to ES. 

With the changes in this PR, what happens is:

1. agent bootstraps fleet server
2. fleet server gets the agent policy from ES
3. fleet server sends that policy to agent
4. agent sees the policy has more outputs for fleet-server
5. agent merges them with the bootstrap output for fleet server
6. agent passes the merged output to fleet server.

We actually only need to get to step 2 for Fleet server to discover it has more possible outputs. I think keeping 3-6 here is unnecessary complexity and is probably putting the change in the wrong place (agent). If we push this change into Fleet server, it always has the policy outputs persisted in ES, and agent is always passing it the bootstrap output, so it always has all the information it needs to merge or switch between them as needed.

The only thing this doesn't do is make it possible for a user to change the bootstrap output from Fleet, but nobody is asking us to do that yet, and that isn't the change the issue here is requesting.

michel-laterman:
We'll consider the (output) config that the elastic-agent passes to fleet-server as bootstrap input.
Once fleet-server is able to read it's policy we will merge the local (bootstrap) output config with the policy output block when fleet-server runs in agent mode

Issue #4475, opened 2024-03-23
[Flaky Test]: Integration tests keep running forever until manually cancelled


rdner:
* https://buildkite.com/elastic/elastic-agent/builds/7938#018e669f-d4f2-423c-84f7-e6df2caf22b5 is cancelled after 15 hours 
* https://buildkite.com/elastic/elastic-agent/builds/7939#018e675d-553b-444c-8521-e0739a0a2bed is cancelled after 12 hours
* https://buildkite.com/elastic/elastic-agent/builds/7942#018e689d-374d-4154-9c5b-91e8729c11ae cancelled after 6 hours

Need to carefully inspect the logs, improve logging, if necessary, and find the root cause.
Could be related to https://github.com/elastic/elastic-agent/issues/4356

rdner:
Another build got stuck (> 10h) https://buildkite.com/elastic/elastic-agent/builds/7946#018e72ea-209d-411e-8881-e83fe768fdc9

cmacknz:
In the first 15 hour log and the most recent log https://github.com/elastic/elastic-agent/issues/4475#issuecomment-2017699066 we never see the SLES runner get past this point:


```
_bk;t=1711326854275>>> (linux-amd64-sles-15-rpm) Starting SSH; connect with `ssh -i /opt/buildkite-agent/builds/bk-agent-prod-gcp-1711324950210306583/elastic/elastic-agent/.integration-cache/id_rsa sles@34.67.132.10`
_bk;t=1711326854906>>> (linux-amd64-sles-15-rpm) Connected over SSH
_bk;t=1711326854906>>> (linux-amd64-sles-15-rpm) Preparing instance
_bk;t=1711326854906>>> (linux-amd64-sles-15-rpm) install devel_basis
```

What is interesting is that this operation has a timeout on the context that should have timed out:

https://github.com/elastic/elastic-agent/blob/d55869484332108b060ebe31bbca779706e77fc4/pkg/testing/runner/sles.go#L20-L26

Following the context here the call stack is:

https://github.com/elastic/elastic-agent/blob/d55869484332108b060ebe31bbca779706e77fc4/pkg/testing/runner/runner.go#L347-L355

https://github.com/elastic/elastic-agent/blob/d55869484332108b060ebe31bbca779706e77fc4/pkg/testing/runner/runner.go#L309

[...]

PR #4478, opened 2024-03-25
Ensure integration test setup has a timeout.

cmacknz:
The top level context for integration tests comes from mage, which doesn't set a default timeout unless you give it one on the command line, which we don't do.

Detect when the timeout wasn't set, and force a reasonable default. Hopefully this prevents the tests from running indefinitely when the failure is in the setup and provisioning stage.

- Relates https://github.com/elastic/elastic-agent/issues/4475

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -79,14 +79,22 @@ const (
 	cloudImageTmpl = "docker.elastic.co/observability-ci/elastic-agent:%s"
 )
 
-// Aliases for commands required by master makefile
-var Aliases = map[string]interface{}{
-	"build": Build.All,
-	"demo":  Demo.Enroll,
-}
-var errNoManifest = errors.New("missing ManifestURL environment variable")
-var errNoAgentDropPath = errors.New("missing AGENT_DROP_PATH environment variable")
-var errAtLeastOnePlatform = errors.New("elastic-agent package is expected to build at least one platform package")
+var (
+	// Aliases for commands required by master makefile
+	Aliases = map[string]interface{}{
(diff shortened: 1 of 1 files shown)


magefile.go
```diff
@@ -79,14 +79,22 @@ const (
...
+	}
+
+	errNoManifest         = errors.New("missing ManifestURL environment variable")
+	errNoAgentDropPath    = errors.New("missing AGENT_DROP_PATH environment variable")
+	errAtLeastOnePlatform = errors.New("elastic-agent package is expected to build at least one platform package")
+
+	// goIntegTestTimeout is the timeout passed to each instance of 'go test' used in integration tests.
+	goIntegTestTimeout = 2 * time.Hour
+	// goProvisionAndTestTimeout is the timeout used for both provisioning and running tests.
+	goProvisionAndTestTimeout = goIntegTestTimeout + 30*time.Minute
```
cmacknz, line 96: Scientifically chosen by looking at the total time of the integration test stage for recent builds. It frequently goes over 2h so I couldn't also use 2h for this.

* 2h9m https://buildkite.com/elastic/elastic-agent/builds/7928#018e6376-e786-4ba4-b2eb-fc521ed64ae7
* 2h7m https://buildkite.com/elastic/elastic-agent/builds/7938#018e6a11-a2db-466f-8b8f-548febe75ddb
* 1h53m https://buildkite.com/elastic/elastic-agent/builds/7916

We could probably shorten this but I want to avoid false positives.

Issue #4377, opened 2024-03-07
[Flaky Test]: TestFleetAirGappedUpgradeUnprivileged – Error while dialing: dial unix /opt/Elastic/Agent/elastic-agent.sock: connect: no such file or directory


rdner:
### Failing test case

TestFleetAirGappedUpgradeUnprivileged

### Error message

Error while dialing: dial unix /opt/Elastic/Agent/elastic-agent.sock: connect: no such file or directory

### Build

https://buildkite.com/elastic/elastic-agent/builds/7621#018e1986-3969-42c4-91d2-071745693aa4

### OS

Linux

### Stacktrace and notes

```markdown
upgrade_fleet_test.go:234: Creating Agent policy...
    upgrade_fleet_test.go:239: Creating Agent enrollment API key...
    upgrade_fleet_test.go:246: Getting default Fleet Server URL...
    upgrade_fleet_test.go:250: Installing Elastic Agent (unprivileged: true)...
    fixture_install.go:130: [test TestFleetAirGappedUpgradeUnprivileged] Inside fixture install function
    fixture.go:632: >> running binary with: [/tmp/TestFleetAirGappedUpgradeUnprivileged1821186027/003/elastic-agent-8.14.0-SNAPSHOT-linux-arm64/elastic-agent install --force --non-interactive --unprivileged --url https://3da74bc2f27846ea9f056981dbca1fbb.fleet.us-west2.gcp.elastic-cloud.com:443 --enrollment-token N0NLWUdZNEIxa0wzTzdlbUl4NTA6QWRVTEEwbF9Sb2VwcmJ3U1IzejJvZw==]
    upgrade_fleet_test.go:265: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_fleet_test.go:265
        	            				/home/ubuntu/agent/testing/integration/upgrade_fleet_test.go:195
        	            				/home/ubuntu/agent/testing/integration/upgrade_fleet_test.go:115
        	Error:      	Received unexpected error:
        	            	error running agent install command: exit status 1
        	Test:       	TestFleetAirGappedUpgradeUnprivileged
        	Messages:   	failed to install start agent [output: Unprivileged installation mode enabled; this is an experimental and currently unsupported feature.
        	            	Installing in non-interactive mode.
        	            	
[    ] Creating group elastic-agent  [0s] 
                                          

[   =] Successfully created group elastic-agent  [0s] 
                                                      

[...]

PR #4480, opened 2024-03-26
Add TEST_PACKAGES environment variable for integration tests

AndersonQ:
## What does this PR do?

add TEST_PACKAGES environment variable for integration tests

## Why is it important?

It allows to select which packages the integration test will copy to the test runner

## Checklist

Run an integration tests with on;y the selected agent package built. The test should run
```
AGENT_KEEP_INSTALLED=true INSTANCE_PROVISIONER="multipass" SNAPSHOT=true TEST_PLATFORMS="linux/amd64" TEST_PACKAGES="tar.gz" mage integration:single TestFleetManagedUpgradePrivileged
```

## Related issues

- Relates #4377 

## Questions to ask yourself

- ...

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -2285,6 +2285,7 @@ func createTestRunner(matrix bool, singleTest string, goTestFlags string, batche
 		DiagnosticsDir: diagDir,
 		StateDir:       ".integration-cache",
 		Platforms:      testPlatforms(),
+		Packages:       testPackages(),
 		Groups:         testGroups(),
 		Matrix:         matrix,
 		SingleTest:     singleTest,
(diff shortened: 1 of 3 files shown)


cmacknz:
Why does setting the test filter to exclude tests that don't require the RPM or DEB not work?  https://github.com/elastic/elastic-agent/blob/main/docs/test-framework-dev-guide.md#passing-additional-go-test-flags

Is it because the `define.Require` statement doesn't declare what packages it needs?

AndersonQ:
> Is it because the `define.Require` statement doesn't declare what packages it needs?

Yes, the `define.Require` isn't used to define the agent packages needed for the tests. Just as there is `TEST_PLATFORMS` to define the platforms for the artefacts the framework will copy, a `TEST_PACKAGES` is also needed to define the packages.

Of course another option would be to use the `define.Require`, however there will still be the case where a test runs everywhere and we want to run it on a single platform using a single package. It's much more for allowing us to run tests as we develop them or to troubleshoot than restricting where the tests should run.

pchila (approved):

PR #4481, opened 2024-03-26
Check if the Fleet API version is different between request and response to detect an api version switch request

### LINKED ISSUE #4477: Elastic agent interprets any `400-Bad request` status code from Fleet as request to switch Fleet API version
- Version: >= 8.11.*
- Operating System: Any

Elastic agent introduced version negotiation in 8.11 with #3383, however it interprets any `400-Bad Request` that includes an `Elastic-Api-Version` Header as a request to switch API version (which at the time of writing is not yet implemented as Agent supports a single Fleet API version `2023-06-01`).
It seems that Fleet will include the `Elastic-Api-Version` in all responses, hence triggering some false positives logs on Agent side.

The behavior should be changed so that:
- If elastic agent receives a `400 - Bad request` status code it should compare the value of `Elastic-Api-Version` to the one used in the HTTP request and report it as a request to switch API versions only if the values are different.
- Elastic agent should try to decode the response body and log the error message if present to help with debugging issues.

pchila:
## What does this PR do?
This PR adds a check before logging a `400 - Bad request` status code in Fleet response as a request to change `Elastic-Api-Version`.
Before this change any response indicating a bad request, containing an Elastic-Api-Version would be considered (and logged) as a switch version request.
With this change we log only if Fleet answered with an api version that is *different* from what was sent in the request.

This part of the code did not influence the response processing even before this change and the error message extraction is left to the command layer (CheckinCmd, EnrollCmd, AckCmd).

## Why is it important?

This should reduce the noise in the logs and avoid misleading during investigation of issues by indicating a request to switch version when in reality there was none.

## Checklist

## Author's Checklist

- [ ]

To test this fix we need either a Fleet server that returns 400 (may be a mock one) or simply run the `TestElasticApiVersion` unit tests in `internal/pkg/fleetapi/client/client_test.go`

## Related issues

- Closes #4477

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1711474205-Reduce-false-positives-in-logging-an-API-switch-request-from-Fleet-server.yaml
+++ b/changelog/fragments/1711474205-Reduce-false-positives-in-logging-an-API-switch-request-from-Fleet-server.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
 summary: Reduce false positives in logging an API switch request from Fleet server
@@ -26,7 +26,9 @@ component: elastic-agent
 # NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
 # Please provide it if you are adding a fragment for a different PR.
 #pr: https://github.com/owner/repo/1234
+pr: https://github.com/elastic/elastic-agent/pull/4481
 
 # Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
 # If not present is automatically filled by the tooling with the issue linked to the PR number.
 #issue: https://github.com/owner/repo/1234
+issue: https://github.com/elastic/elastic-agent/issues/4477
--- a/changelog/fragments/1711474205-Reduce-false-positives-in-logging-an-API-switch-request-from-Fleet-server.yaml
+++ b/changelog/fragments/1711474205-Reduce-false-positives-in-logging-an-API-switch-request-from-Fleet-server.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Reduce false positives in logging an API switch request from Fleet server
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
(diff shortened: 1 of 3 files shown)


internal/pkg/fleetapi/client/client_test.go
```diff
@@ -183,6 +183,37 @@ func TestElasticApiVersion(t *testing.T) {
 		}
 	})
 
+	t.Run("verify that we don't log a generic 400 status as a downgrade request", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		mux := http.NewServeMux()
+		mux.HandleFunc("/someotheroperationreturnsbadrequest", func(writer http.ResponseWriter, request *http.Request) {
+			assert.Equal(t, request.Header.Get(elasticApiVersionHeaderKey), defaultFleetApiVersion)
```
AndersonQ, line 192: [Blocker]
What exactly do you intend to test here?
Because if for some reason the code was changed and the agent client ins't sending `defaultFleetApiVersion` as the version, it'll be caught at assertion at the end of the test. I'm not sure you need this assert here. I don't believe this test is testing the client indeed sends `defaultFleetApiVersion` as the api version.

internal/pkg/fleetapi/client/client_test.go
```diff
@@ -183,6 +183,37 @@ func TestElasticApiVersion(t *testing.T) {
 		}
 	})
 
+	t.Run("verify that we don't log a generic 400 status as a downgrade request", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		mux := http.NewServeMux()
+		mux.HandleFunc("/someotheroperationreturnsbadrequest", func(writer http.ResponseWriter, request *http.Request) {
```
AndersonQ, line 191: [Suggestion]
Please could you at least use camelCase. It's hard to read the path as it's is and ensure the call uses the same path.

AndersonQ (changes requested):
A few questions that needs to be addressed or answered, I marked them as [blockers] the others comments, you might ignore if you want, they're only suggestions.

changelog/fragments/1711474205-Reduce-false-positives-in-logging-an-API-switch-request-from-Fleet-server.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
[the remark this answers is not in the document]
pchila: Yep, forgot to change that

internal/pkg/fleetapi/client/client_test.go
pchila: The agent client now ALWAYS send the defaultFleetApiVersion (it's the only version we support).
This assertion is to verify that we get the correct header value in the request on the server side.

The assertion at the end of the test is to check the behaviour of the client to the header in the response.

internal/pkg/fleetapi/client/client_test.go
pchila: I will put in the same url like the others

changelog/fragments/1711474205-Reduce-false-positives-in-logging-an-API-switch-request-from-Fleet-server.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
```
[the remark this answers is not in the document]
pchila: Filled both of them, however according to comments that should not be necessary...

internal/pkg/fleetapi/client/client_test.go
pchila: done in 279b5275531a1932333e0d523e07d34c3cc38ba7

AndersonQ (approved):

PR #4491, opened 2024-03-27
Always launch more recent watcher

### LINKED ISSUE #4072: Agent should always launch the more recent watcher version when downgrading
- Relates https://github.com/elastic/kibana/issues/172745

**Describe the enhancement:**
Currently elastic-agent starts the watcher using the binary of the version it's upgrading to.
This may cause issues when downgrading since the older elastic-agent binary may not know about new features introduced only by the new version.

If elastic-agent starts the watcher always from the most recent version regardless of the direction of the upgrade/downgrade it should be able to cope better on the cleanup/rollback phase

A use case for this feature is when downgrading an agent that contains #2579 to a version that does not, since the upgrade marker contains extra information needed for the rollback that would be ignored by the older watcher.

Given an installed agent version x.y.z
When downgrading to a previous version a.b.c
Then the watcher is spawned using elastic-agent binary x.y.z

Given an installed agent version a.b.c
When upgrading to a more recent version x.y.z
Then the watcher is spawned using elastic-agent binary x.y.z

pchila:
## What does this PR do?

This PR changes the way elastic-agent launches the upgrade watcher in the final phases of the upgrade process.
Before this change, the elastic-agent binary from the incoming version (the one specified in the `upgrade` command) was always used but this meant that during a downgrade an *older* version of elastic agent would be launched, potentially not supporting new fields in the upgrade marker generated during the upgrade.

Now the agent will check the current and incoming version and always select the newer one to launch the watcher process: during a downgrade this means that the *current* agent binary will function as a watcher to check that the incoming version functions correctly after the restart.

## Why is it important?

We want to avoid issues during watch/cleanup/rollback phases of upgrades even if the incoming version is older than the current install.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #4072 

## Questions to ask yourself

- ...


internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -272,12 +272,15 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	// while the `previous` install is the currently executing elastic-agent that is no longer reachable via the symlink.
 	// After the restart at the end of the function, everything lines up correctly.
 	current := agentInstall{
+		parsedVersion: parsedVersion,
 		version:       version,
 		hash:          unpackRes.Hash,
 		versionedHome: unpackRes.VersionedHome,
 	}
 
+	previousParsedVersion := getParsedVersionWithFallback(u.log, release.VersionWithSnapshot(), agtversion.NewParsedSemVer(8, 13, 0, "", ""))
```
cmacknz, line 281: Needs a comment explaining why 8.13.0 is hard coded here.

cmacknz: Also it's not clear why we need a fallback. Under what circumstances would parsing `release.VersionWithSnapshot()` fail?

pkg/version/version_parser.go
```diff
@@ -99,34 +145,66 @@ func (psv ParsedSemVer) Less(other ParsedSemVer) bool {
...
+		minPrereleaseTokens = len(otherPrereleaseTokens)
 	}
 
-	// We are at the first minor of the current major version series. To
-	// figure out the previous minor, we need to rely on knowledge of
-	// the release versions from the past major series'.
-	switch major {
-	case 8:
-		return NewParsedSemVer(7, 17, 10, psv.Prerelease(), psv.BuildMetadata()), nil
+	for i := 0; i < minPrereleaseTokens; i++ {
```
cmacknz, line 174: What problem is this solving? Do we need this code to handle something in our current or planned release versioning scheme or is it just in case?

internal/pkg/agent/application/upgrade/upgrade.go
pchila: It may be excess of caution on my side but release.VersionWithSnapshot() is a string returned by reading the package.version (which can be edited and made it to contain corrupt data) or referencing the BEATS_VERSION as a fallback. 
I didn't want to deal with a parsing error for the current version string at this stage of the upgrade process so I added the fallback to 8.13.0 which is the last version where the new update marker was introduced.

I could make the upgrade rollback here at the very end if we cannot parse the current version but I tried to give priority to the upgrade which at this point we managed to unpack, rotate links and copy state to...

pkg/version/version_parser.go
pchila: This is the correct implementation for comparing 2 semver version strings. Prerelease strings should be compared token by token as strings or numbers depending on their format.

Right now we only use "SNAPSHOT" prerelease string and the previous impl would just compare 2 versions based on the presence of prerelease string without comparing those, leading to versions like
`1.2.3-SNAPSHOT` , `1.2.3-alpha`, `1.2.3-SNAPHOT.extraprereleasetoken` not to have an definite order.

This new implementation is able to parse the tokens and correctly determine precedence based on the prerelease token: an example when this is gonna come into play is of we decide to use repackaged agents in upgrade tests as `1.2.3-SNAPSHOT.repackaged` now has a defined order when compared with `1.2.3-SNAPSHOT`

This is a part of a general overhaul of agent version handling correct versioning both at packaging and runtime which I then extracted to put in a separate PR as it would make this change too big.

internal/pkg/agent/application/upgrade/upgrade.go
michalpristas: can you make a comment out of it. i bet i won't remember in a month

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -328,6 +324,26 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
+		previousParsedVersion = fallbackVersion
+		log.Warnw(fmt.Sprintf("parsing of current version failed, watcher selection will fall back to comparing to %q", previousParsedVersion), "error.message", err)
+	}
+	return previousParsedVersion
+}
+
+func selectWatcherExecutable(topDir string, previous agentInstall, current agentInstall) string {
+	// check if the upgraded version is less than the previous (currently installed) version
+	if current.parsedVersion.Less(*previous.parsedVersion) {
+		// use the current agent executable for watch, if downgrading the old agent doesn't understand the current agent's path structure.
```
michalpristas, line 339: comment says use the current as currently installed but it's not current as a variable name. it gets confusing

michalpristas (approved):

changelog/fragments/1712737337-always-launch-more-recent-watcher.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11:
```suggestion
kind: bug-fix
```

changelog/fragments/1712737337-always-launch-more-recent-watcher.yaml
```diff
@@ -0,0 +1,32 @@
...
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
```
blakerouse, line 28:
```suggestion
pr: https://github.com/elastic/elastic-agent/pull/4491
```

internal/pkg/agent/application/upgrade/upgrade.go
blakerouse: I don't really like that this is hard-coded. Will this need to be updated when we create a new release? Why not use `version.Agent`?

blakerouse (changes requested):
Really don't like the hard-coding of 8.13.0.

internal/pkg/agent/application/upgrade/upgrade.go
pchila: Agreed but this is the existing names that follow where the symlink points.
So after rotating the symlink the currently running agent is named previous and the new one is named current.
I don't know if this is more confusing or swapping the variable names so when we look at it in different places in the code it gets confusing...

pchila:
Removed `getParsedVersionWithFallback` function in favor of reading the current parsed version at startup from the `package.version` file

blakerouse (approved):
Thanks for the fixes.

Like this implement of the pressed version string much better.

internal/pkg/agent/application/upgrade/upgrade.go
pchila: This has now been removed in favor of using the agent package version we initialize at startup.

internal/pkg/agent/application/upgrade/upgrade.go
michalpristas: can we maybe rephrase
`use the current agent executable for watch, if downgrading the old agent doesn't understand the current agent's path structure.`
to 
`if downgrading the target agent doesn't understand the current agent's path structure.`

pchila:
buildkite test this

PR #4461, opened 2024-03-21
add integration test for install via RPM on SLES

leehinman:
## What does this PR do?

Adds Integration test that uses SLES 15 AMD to test fleet managed elastic-agent installed via RPM.

Only AMD was added because at the moment ogc doesn't support SLES ARM on GCP.  Also looked at implementing https://github.com/elastic/elastic-agent/pull/4301#discussion_r1516304403 This was skipped because adding a slice of strings (for supported package formats) to the OS struct broke comparison of OS structs.  I think this is still worth doing, just in a separate PR because it is likely to change a fair number of things.

## Why is it important?

Need to test our RPM builds

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## How to test this PR locally

```
mage integration:single TestRpmLogIngestFleetManaged
```

## Related issues

- Relates #4086

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4498, opened 2024-03-28
Add RPM integration tests

leehinman:
## What does this PR do?
Second attempt at adding RPM integration tests.

Previous attempt #4461 was reverted because SLES platform was failing due to a race condition where the rpm repositories were removed.  This PR switches the tests to RHEL, and adds additional SSH logging to make it easier to identify events like this.  TCP keepalives were also added to SSH connections to better detect network outages.

## Why is it important?

Need to test RPM builds

## Checklist

## Author's Checklist

- [ ]

```
mage integration:single TestRpmLogIngestFleetManaged
```

## Related issues

- Relates #4086
- Supersedes #4461

## Questions to ask yourself

- ...


pkg/testing/runner/sles.go
```diff
@@ -0,0 +1,112 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
rdner, line 1: Do we still need the SLES runner?

pkg/testing/runner/ssh.go
```diff
@@ -189,12 +222,21 @@ func (s *sshClient) Exec(ctx context.Context, cmd string, args []string, stdin i
...
-		return nil, nil, fmt.Errorf("could not create new SSH session: %w", err)
+		s.logger.Logf("new session failed: %q, trying reconnect", err)
+		lErr := s.Reconnect(ctx)
+		if lErr != nil {
+			return nil, nil, fmt.Errorf("ssh reconnect failed: %w, after new session failed: %w", lErr, err)
+		}
+		session, lErr = s.NewSession()
+		if lErr != nil {
+			return nil, nil, fmt.Errorf("new session failed after reconnect: %w, original new session failure was: %w", lErr, err)
+		}
```
rdner, line 239: If understand this correctly, we have only one retry without a delay here, should we use a configurable backoff/retry here?

pkg/testing/runner/ssh.go
```diff
@@ -225,6 +267,7 @@ func (s *sshClient) ExecWithRetry(ctx context.Context, cmd string, args []string
 		if err == nil {
 			return stdout, stderr, nil
 		}
+		s.logger.Logf("ssh exec error: %q, will try again in %s", err, interval)
```
rdner, line 270: Very nice that we now have logging.

pkg/testing/runner/supported.go
```diff
@@ -70,6 +72,26 @@ var (
...
+	// SlesAMD64_15 - SUSE Linux Enterprise Server (amd64) 15
+	SlesAMD64_15 = SupportedOS{
+		OS: define.OS{
+			Type:    define.Linux,
+			Arch:    define.AMD64,
+			Distro:  Sles,
+			Version: "15",
+		},
+		Runner: SlesRunner{},
+	}
```
rdner, line 94: Same question, if SLES does not work for us, does it make sense to keep it in our tests?

pkg/testing/runner/sles.go
leehinman: No, we don't really need it.  I was using it for my testing because the SSH connections had failures often enough to make it easy to see that the SSH changes were working.  Anyway, I can remove it.

pkg/testing/runner/ssh.go
leehinman: The retry isn't obvious, `Reconnect` calls `Close` then `Connect`, and the `Connect` has the retry with exponential back off.  If `NewSession` fails, that means we failed twice after what should have been valid connections via `Connect`, at that point I think it is good to just print the errors and let a human figure out what is going on.

rdner (approved):

Issue #390, opened 2022-04-28
Liveness Probe HTTP Endpoint


blakerouse:
**Describe the enhancement:**

Currently the Elastic Agent running in a container does not have a `liveness` HTTP endpoint where kubernetes can check the overall health of the Elastic Agent container. This needs to be added so that in the case that the Elastic Agent is not working correctly it can be restarted by Kubernetes.

Some items that would be good for the liveness probe to be alerted to on failure:

 * Not able to connect to Fleet Server (in managed mode)
 * Overall bad state of inputs after a period of time

Liveness probe should have some subpaths defined for inputs that need to monitor there own liveness:

`/liveness/endpoint` - Checks if endpoint should be alive (see https://github.com/elastic/security-team/issues/3449#issuecomment-1112559420) for more details

**Describe a specific use case for the enhancement or feature:**

This needs to be added so that in the case that the Elastic Agent (or an integration that runs in a sidecar) is not working correctly it can be restarted by Kubernetes.

**Additional Requirements**

Enabling the liveness endpoint requires the ability to enable and possibly modify the agent HTTP configuration. This is currently not reloadable and cannot be configured from Fleet. For Fleet managed users to benefit from this we should make sure this can be turned on from Fleet.

https://github.com/elastic/elastic-agent/blob/b3e82754527d10e0ccde5e081d38db3f97de5b35/elastic-agent.yml#L75-L82

ph:
@jlind23 @nimarezainia We should discuss this with our priority with @norrietaylor for the endpoint on k8s project.

maneta:
Hey folks, we are also hitting an issue in the synthetics-service (https://github.com/elastic/synthetics-service/issues/435) where the agent fails in acquire the leader lease: 

```
May 2, 2022 @ 14:56:46.498	elastic-agent-6qrjw	I0502 14:56:46.498250       7 leaderelection.go:278] failed to renew lease kube-system/elastic-agent-cluster-leader: timed out waiting for the condition

May 2, 2022 @ 14:56:46.498	elastic-agent-6qrjw	E0502 14:56:46.498350       7 leaderelection.go:301] Failed to release lock: resource name may not be empty
``` 

We think that this issue might be caused by the Apiserver being too slow to provide the lease. We are digging into metrics to certify that, but for now, the only solution we have to re-establish the agent is to manually restart the daemonsets. Would be great if we could set a liveness/readiness endpoint in the k8s template so k8s takes care of re-starting the pods for us.

[...]

PR #2729, opened 2023-05-25
Fix deadlock in RuntimeManager

blakerouse:
## What does this PR do?

Fixes a deadlock issue that can occur in the communication of updated components with the `RuntimeManager` and each of the `CommandRuntime` and `ServiceRuntime` component managers.

## Why is it important?

It is possible for a component to become disconnected from the Elastic Agent control protocol and unable to complete re-connection because a concurrent update of the component model was occurring.

This change also allows for the runtime manager to not need to settle before calling the initial `m.Update`. You can see that I was able to remove the initial delay in the existing runtime manager tests.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- ~~[ ] I have added an integration test or an E2E test~~

## How to test this PR locally

Only way to really verify is to remove the changes in this PR, except for the `TestManager_FakeInput_NoDeadlock` test. Run the `TestManager_FakeInput_NoDeadlock` test and see that it fails. Then re-add the changes and see that it causes the test to pass.

## Related issues

- Closes #2691

PR #4499, opened 2024-03-28
Add `/liveness` endpoint to elastic-agent

fearful-symmetry:
## What does this PR do?

Closes https://github.com/elastic/elastic-agent/issues/390

## Why is it important?

This adds a `/liveness` endpoint to elastic agent that we enable along with the `processes` endpoint. However, unlike the `/processes` endpoint, `liveness` will return a `500` is any components are in an Unhealthy or Degraded state.

This is a pretty simple change as-is, since I figured it would be easier to discuss other theoretical changes when we can actually see code. In particular:

1) Should this be toggled by something other than the `monitoring.http` flag? Should we enabled it based on some k8s autodiscover config or env variable or something else? If this feature is fundamentally tied to a k8s workflow, it seems like enabling it should also be tied to k8s in some way.
2) Should we fail the health check based on any other conditions?

I'm also holding off on changing any docs until we're sure we have what we want with regards to config.

## How to test this PR

- set `monitoring.http.enabled` to `true`
- Check `/liveness/` and `/liveness/[component-id]` using either curl or a k8s health check.

## Checklist


internal/pkg/agent/application/monitoring/processes_cloud.go
```diff
@@ -18,7 +18,7 @@ func cloudComponentIDToAgentInputType(componentID string) string {
 	return componentID
 }
 
-func expectedCloudProcessID(c *component.Component) string {
+func expectedCloudProcessID(c component.Component) string {
```
pchila, line 21: do we know what the impact (with particular regard to memory churn) is of copying the struct instead of just passing the pointer? Especially considering that the liveness  probe will run every few seconds ?

internal/pkg/agent/application/monitoring/process.go
```diff
@@ -82,14 +81,18 @@ func processHandler(coord *coordinator.Coordinator, statsHandler func(http.Respo
 
 		state := coord.State()
 
-		for _, c := range state.Components {
-			if matchesCloudProcessID(&c.Component, componentID) {
+		unhealthyComponent := false
+		for _, comp := range state.Components {
+			if comp.State.State == client.UnitStateFailed || comp.State.State == client.UnitStateDegraded {
+				unhealthyComponent = true
+			}
```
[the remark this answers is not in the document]
fearful-symmetry: Hmm, good question. @cmacknz thoughts?

internal/pkg/agent/application/monitoring/processes_cloud.go
fearful-symmetry: So, I initially did that because it set off a linter error in `processesHandler()` 's `for` loop, as we're borrowing a reference from the for loop. This behavior is [changing in 1.22](https://go.dev/blog/loopvar-preview). I wasn't thinking about memory usage, though. I wonder if we can just only pass the parts of the component struct that we actually need.

internal/pkg/agent/application/monitoring/process.go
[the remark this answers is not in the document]
pierrehilbert: It's a public holiday today for Craig, he will reply on Tuesday.
I'm not sure restarting the full container when a component is degraded is the best thing to do but maybe I'm wrong.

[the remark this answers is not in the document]
leehinman: From my point of view (which could be totally wrong) we only want to fail the "liveness" check if we believe that restarting the container has a chance of fixing the problem.  For example if something in the configuration sent down from fleet caused a component to be in Degraded state, I don't want to fail the "liveness" check.  Restarting the container won't fix the issue, and we will just be constantly restarting, which makes figuring out what is wrong even harder.

So it might be worth asking the question "What situations do we have where a reboot of the container is likely to fix the problem?"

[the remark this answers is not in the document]
cmacknz: Yes, there actually aren't many situations where a container restart will fix a problem. At the same time, we don't know the exact conditions of every possible error an input will report. Some people may also end up using this as a crude alert on agent pods needing intervention.

The nicest way to deal with this is to make it configurable. We could have several options.
1. Liveness only checks that the container can respond to an HTTP request at all. This should be the default.
2. Liveness fails when any component or unit is in the failed state.
3. Liveness fails when any component or unit is in the degraded state.
4. Liveness fails when the agent determines it is offline from Fleet (no check ins for X minutes).

We could also optionally introduce the ability to filter which components or unit names are included in these conditions, in case someone wants to limit checks to only outputs or log collection or something.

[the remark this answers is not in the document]
fearful-symmetry: Alright, so what we need is some config subsection in`monitoring.http` that lets us define when we actually return a `500`. 

> Liveness only checks that the container can respond to an HTTP request at all. This should be the default.

So the default should just be that the endpoint works at all, and not based on any component/unit state? That seems like it might be confusing?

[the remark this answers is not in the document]
cmacknz: > So the default should just be that the endpoint works at all, and not based on any component/unit state? That seems like it might be confusing?

Yeah thinking a bit more we could have the default be to fail if anything is in the failed state, with the option to restrict it or loosen it to include degraded. I have a suspicion people want to solve the problem where an input can't be created at all and the container is considered "alive" even though it is sitting there not collecting the data they want silently. This makes it more useful that just hitting `/stats` by default but allows users to control what it does.

For a concrete example of how we could configure this:
```yaml
 agent.monitoring:
  http:
      # enables http endpoint
      enabled: true
      # The criteria used to determine with the /liveness endpoint should not return a 200 OK response
      # Must be one of: "failed", "degraded", "none". Defaults to "failed".
      # - "failed" returns an error when any component or unit is in the failed state. 
      # - "degraded" returns an error when any component or unit is in the degraded state.
      # - "none" always returns success as long as the agent can respond over HTTP at all.
      liveness_criteria: "failed"
```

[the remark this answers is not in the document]
cmacknz: If you want to you can make `liveness` its own sub-object if you think it's nicer or we could add more keys in the future. Also open to a different name than `criteria` if someone can think of a better one.

```yaml
 agent.monitoring:
  http:
      # enables http endpoint
      enabled: true
      liveness: 
        # The criteria used to determine with the /liveness endpoint should not return a 200 OK response
        # Must be one of: "failed", "degraded", "none". Defaults to "failed".
        # - "failed" returns an error when any component or unit is in the failed state. 
        # - "degraded" returns an error when any component or unit is in the degraded state.
        # - "none" always returns success as long as the agent can respond over HTTP at all.
        criteria: "failed"
```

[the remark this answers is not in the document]
blakerouse: I think if we want to provide options to adjust the result of the HTTP response it would be better if its in the HTTP request, like `/healthz?criteria=failed`.

I also had some original ideas on this in the linked issue - https://github.com/elastic/elastic-agent/issues/390#issuecomment-1400796372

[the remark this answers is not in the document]
cmacknz: +1, query parameters are better than configuration.

fearful-symmetry:
So, I'm still doing manual testing, but updating this as-is while I figure out how integration tests should work.

internal/pkg/agent/application/monitoring/stats.go
```diff
@@ -12,6 +12,35 @@ import (
...
+	if userConfig == "failed" {
+		// same as default
+		return defaultUserCfg, nil
+	} else if userConfig == "degraded" {
+		return LivenessFailConfig{Failed: true, Degraded: true}, nil
+	} else if userConfig == "" {
+		return defaultUserCfg, nil
+	} else {
+		return defaultUserCfg, fmt.Errorf("got unexpected value for `%s` attribute: %s", formValueKey, userConfig)
+	}
```
leehinman, line 41: a switch statement would probably be easier to read, especially since you have "" and "failed" doing the same thing.

What if the user doesn't want a "failed" component to mean that elastic-agent is dead?  I don't think you can get that behavior with this code.

internal/pkg/agent/application/monitoring/process.go
```diff
@@ -99,6 +107,9 @@ func processHandler(coord *coordinator.Coordinator, statsHandler func(http.Respo
 				} else {
 					content = string(bytes)
 				}
+				if livenessMode && unhealthyComponent {
+					w.WriteHeader(500)
```
leehinman, line 111:
```suggestion
					w.WriteHeader(http.StatusInternalServerError)
```

leehinman (changes requested):
Couple code suggestions.

But my biggest problem is with the whole idea of calling this a liveness check and having things like degraded state failing it.  Just because a component is failed, doesn't mean elastic-agent isn't alive.

Looking at the [K8s documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) I think this "caution" is very relevant.

```
Caution: Liveness probes can be a powerful way to recover from application failures,
 but they should be used with caution. Liveness probes must be configured carefully 
to ensure that they truly indicate unrecoverable application failure, for example a 
deadlock.
```

cmacknz:
The liveness endpoint behavior and configuration needs to be documented. I'd start with the HTTP section of the default and reference configs:
- https://github.com/elastic/elastic-agent/blob/92ae4f421ca0bd93667145d53929d4f8709cd3bc/elastic-agent.yml#L75
- https://github.com/elastic/elastic-agent/blob/92ae4f421ca0bd93667145d53929d4f8709cd3bc/elastic-agent.reference.yml#L154

We can also add a section to https://github.com/elastic/ingest-docs/blob/main/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc once this is merged.

cmacknz:
> But my biggest problem is with the whole idea of calling this a liveness check and having things like degraded state failing it. Just because a component is failed, doesn't mean elastic-agent isn't alive.

This is why that behavior is configurable, but it isn't documented yet. I don't know how else to handle this really but to let the user decide what they want it to do. 

For example, people might configure the liveness probe to restart the agent container on failed and degraded statuses in a staging environment as an obvious way to detect problems, and then only have it be a basic check that the agent responds to HTTP at all in production.

cmacknz:
We should have an opinionated default as well, so we should add something with explanation on the pitfalls of using this liveness probe to our reference configurations:

- https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-managed/elastic-agent-managed-daemonset.yaml
- https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-standalone/elastic-agent-standalone-daemonset.yaml

I would suggest we add a default liveness probe that just makes sure the agent responds to HTTP by default and cautious that opting in to restarts for the failed and degraded states can restart the agent in situations where restarts won't necessarily help.

leehinman:
> This is why that behavior is configurable, but it isn't documented yet. I don't know how else to handle this really but to let the user decide what they want it to do.
> 
> For example, people might configure the liveness probe to restart the agent container on failed and degraded statuses in a staging environment as an obvious way to detect problems, and then only have it be a basic check that the agent responds to HTTP at all in production.

I know it is just a name, but "liveness" might be part of what bothers me about this.  We seem to be returning the status anyway, maybe change the name to "status", and document the http return code, so if you want to use it for basic "liveness" you can just by checking the http return code.  It would also be nice to not calculate or include the payload in that case (make it cheaper).

I think ideally liveness would be that we have an unblocked go routine to handle a request, the internal coordinator is running, and can process communication to it's children and fleet server (if applicable).  That is an elastic-agent that is "alive", it can receive signals, new configuration, upgrades, etc.  Adding the status of the individual components is more like the "health" of the elastic-agent.   I agree that a user may want to a "live" but "unhealthy" agent the same as "dead", but that shouldn't be the default (which it currently is).

nimarezainia:
@kilfoyle just wanted to bring this issue to your attention also.

pchila:
> > This is why that behavior is configurable, but it isn't documented yet. I don't know how else to handle this really but to let the user decide what they want it to do.
> > For example, people might configure the liveness probe to restart the agent container on failed and degraded statuses in a staging environment as an obvious way to detect problems, and then only have it be a basic check that the agent responds to HTTP at all in production.
> 
> I know it is just a name, but "liveness" might be part of what bothers me about this. We seem to be returning the status anyway, maybe change the name to "status", and document the http return code, so if you want to use it for basic "liveness" you can just by checking the http return code. It would also be nice to not calculate or include the payload in that case (make it cheaper).
> 
> I think ideally liveness would be that we have an unblocked go routine to handle a request, the internal coordinator is running, and can process communication to it's children and fleet server (if applicable). That is an elastic-agent that is "alive", it can receive signals, new configuration, upgrades, etc. Adding the status of the individual components is more like the "health" of the elastic-agent. I agree that a user may want to a "live" but "unhealthy" agent the same as "dead", but that shouldn't be the default (which it currently is).

To expand on what has been said, liveness should succeed if the application is in a sane state (i.e. not blocked, in a terminal failure state where only a full restart will help etc.).
If we want to build in an escape hatch via config it's fine but by default I would say that we need to be very careful and strict about signaling that the agent is not alive.

The proper solution will come only when the agent will orchestrate components as separate containers and then a liveness probe can be specified for each of those.

One more thought: is it worth restarting even if a component is in "FAILED" state? When I reviewed this PR initially I thought it would make sense but the more I think about it, the less sure I am.
For example we could have a failure on 1 component out of 20 (completely arbitrary number) due to some misconfiguration: is it worth interrupting the work of every component to try and recover something that can be fixed only by changing config ?

I am leaning more towards @leehinman opinion on the liveness: if the agent is unblocked and able to handle "request" (in our case this has a rather broad meaning) the liveness probe should succeed regardless of components state

cmacknz:
> I think ideally liveness would be that we have an unblocked go routine to handle a request, the internal coordinator is running, and can process communication to it's children and fleet server (if applicable). That is an elastic-agent that is "alive",

We have this check built into the agent, and it reports lack of check ins by marking the components as degraded or failed. The core problem may actually be that the agent health reporting model does not distinguish problems caused by internal errors (can't communicate with sub-components) from those more likely to be caused by user error (can't create component because configuration is wrong). I'm not sure there agent can distinguish these.

If we want to make the default case more useful, perhaps we need to consider the recency of the state update we are looking at. Right now when we get the state is retrieved asynchronously and could be quite old, that is if the coordinator were deadlocked nothing here could tell.

https://github.com/elastic/elastic-agent/blob/b7617cf47f1acd0e77022e7817e2103c347af917/internal/pkg/agent/application/coordinator/coordinator.go#L409-L413

Looking at the the definition of coodinator state, maybe what we need to do is add a `LastUpdated` timestamp then we can fail the liveness check if it is sufficiently old (minutes most likely):

https://github.com/elastic/elastic-agent/blob/b7617cf47f1acd0e77022e7817e2103c347af917/internal/pkg/agent/application/coordinator/coordinator_state.go#L19-L37

That makes this much more useful than just executing an HTTP request and processing it, which has rightly been identified as not a strong enough check.

> One more thought: is it worth restarting even if a component is in "FAILED" state? When I reviewed this PR initially I thought it would make sense but the more I think about it, the less sure I am.

I agree we shouldn't do this by default but I want to make sure it can be configured as an escape hatch for unknown unknowns. Let's imagine an agent is only running fleet server and there is some problem that can be fixed by restarting it (TLS configuration not being picked up, stuck connections to ES, whatever). I do not want such workarounds to be impossible. I don't want to have to come back to this implementation reactively if we discover we needed this in a few months.

cmacknz:
> If we want to make the default case more useful, perhaps we need to consider the recency of the state update we are looking at. 

I have made an assumption here that the coordinator state updates regularly, which may not actually be true if the agent configuration is static. This is something we'll have to check. 

I don't think the underlying component state has an updated timestamp and probably it should have something that updates for every successful checkin.

https://github.com/elastic/elastic-agent/blob/b7617cf47f1acd0e77022e7817e2103c347af917/pkg/component/component.go#L153

fearful-symmetry:
So, as I kind of expected, there's a lot of discussion happening here, which is actually why I held of adding to the docs, since I wasn't sure how much the actual design would change.

> I would suggest we add a default liveness probe that just makes sure the agent responds to HTTP by default and cautious that opting in to restarts for the failed and degraded states can restart the agent in situations where restarts won't necessarily help.

So, just to clarify, the default behavior of the endpoint should just return 200, and let the user assume that if the agent is in a state to return a value at all, it's good?

My concern with that is because the HTTP handlers live in their own thread separate from most of the agent runtime, this could miss a lot of conditions where a failed response would be appropriate. I feel like the best way to do this would be to somehow hook the liveness probe into the actual coordinator main loop. Not sure how practical that would be. However, as @cmacknz mentioned, failures in the agent backend itself will tend to result in components getting marked as unhealthy, so perhaps it's just the same.


> I have made an assumption here that the coordinator state updates regularly, which may not actually be true if the agent configuration is static. This is something we'll have to check.

> I don't think the underlying component state has an updated timestamp and probably it should have something that updates for every successful checkin.

Yeah, if we don't have some "heartbeat" for the checkin, I'm not sure how this should work. Add some component that updates regularly? Some other mechanism that reaches out to the main coordinator loop and makes sure we're working?

cmacknz:
> Yeah, if we don't have some "heartbeat" for the checkin, I'm not sure how this should work. Add some component that updates regularly? Some other mechanism that reaches out to the main coordinator loop and makes sure we're working? Been a while since I've looked at the coordinator so I'll probably have to poke around the code and figure it out.

Every process agent supervises needs to check in every 30s or it gets marked as degraded. The heartbeat is already there.

The main problem with the current implementation is that a call to `coordinator.State` could return a state that is 20 minutes old because the coordinator is actually deadlocked. We have had bugs like this before (one example: https://github.com/elastic/elastic-agent/pull/2729) so this isn't a hypothetical situation.

My suggested way to fix this is to consider the recency of the state update we pulled out of `coordinator.State` in the liveness condition. We need to add an updated timestamp to state, and make sure this timestamp is updated regularly. The agent already has a natural heartbeat of the components checkin in, we just need to make sure the new updated based on that or via some other mechanism that ensures the coordinator is still alive.

fearful-symmetry:
Ah, alright, I sort of took your two comments to mean "we can't rely on an updated timestamp if we're using a static agent  config"

cmacknz:
> "we can't rely on an updated timestamp if we're using a static agent config"

Maybe, I don't know off the top of my head if this is the case but it is a risk we have to investigate and address. The next step is to test it and find out if we need to change anything for this to work.

I think having the liveness condition correctly prove that the coordinator itself is alive and not stuck somewhere is a great idea that we should implement. This is a much stronger default liveness condition than just being able to respond to HTTP requests.

fearful-symmetry:
Agreed that this needs to be coupled to the coordinator in some way. Poking around now to see how this can be implemented.

fearful-symmetry:
@faec I want your input here. So, if we want some kind of "heartbeat" to make sure the main coordinator is actually running, it seems like `stateBroadcaster.Subscribe()` probably isn't a good idea, since its behavior will be tied to `stateNeedsRefresh`. Similarly, adding some kind of timestamp to the `State` object that we can update with each call to `runLoopIteration()` we also call `refreshState()` also kinda seems like a bad idea, since we'll hammer anything that's also subscribing to `stateBroadcaster`. I feel like the best option here is to add some kind of `*broadcaster.Broadcaster[Heartbeat]` object to the coordinator for this express purpose. Does that make sense? Is there something I'm missing?

fearful-symmetry:
Had a chat with Fae over Zoom; just updating the integration tests.

pierrehilbert:
It seems to be green

internal/pkg/agent/application/monitoring/server.go
```diff
@@ -42,47 +39,56 @@ func NewServer(
...
+		r.Handle("/stats", createHandler(statsHandler))
+
+		if isProcessStatsEnabled(cfg) {
+			r.Handle("/processes", createHandler(processesHandler(coord, false)))
+			r.Handle("/processes/{componentID}", createHandler(processHandler(coord, false, statsHandler)))
+			r.Handle("/processes/{componentID}/", createHandler(processHandler(coord, false, statsHandler)))
+			r.Handle("/processes/{componentID}/{metricsPath}", createHandler(processHandler(coord, false, statsHandler)))
+
+			r.Handle("/liveness", createHandler(processesHandler(coord, true)))
+			r.Handle("/liveness/{componentID}", createHandler(processHandler(coord, true, statsHandler)))
```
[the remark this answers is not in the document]
fearful-symmetry: If that's the case, should we try to make the actual returned JSON behave the same as the `processes/` endpoint? Do we even need a `/processes/{componentID}` sub-path?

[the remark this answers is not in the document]
cmacknz: For the liveness probe no. Nobody has specifically asked us for this, so we could omit it.

testing/integration/monitoring_probe_reload_test.go
```diff
@@ -0,0 +1,180 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	policyResp, err := tools.InstallAgentWithPolicy(ctx, runner.T(), installOpts, runner.agentFixture, runner.info.KibanaClient, basePolicy)
+	require.NoError(runner.T(), err)
+
+	runner.policyID = policyResp.ID
+	runner.policyName = basePolicy.Name
+
+	_, err = tools.InstallPackageFromDefaultFile(ctx, runner.info.KibanaClient, "system", "1.53.1", "system_integration_setup.json", uuid.New().String(), policyResp.ID)
```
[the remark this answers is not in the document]
fearful-symmetry: It's probably not _needed_, but it seemed like a good idea to have some kind of user config before we test things.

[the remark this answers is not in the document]
cmacknz: It is one more thing that can fail, which is why I ask. The less our tests need to do to succeed, the fewer things that can fail for unrelated reasons.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -977,6 +999,8 @@ func (c *Coordinator) runLoopIteration(ctx context.Context) {
 	case upgradeDetails := <-c.upgradeDetailsChan:
 		c.setUpgradeDetails(upgradeDetails)
 
+	case c.heartbeatChan <- struct{}{}:
```
leehinman, line 1002: question on this.  This looks like we will put something on the channel as soon as possible, so if the coordinator gets blocked after that, you would read from the heartbeatChan and think that it was up, because in the past it had been able to write to the channel.  On your next read it would fail.  Is that correct?

If it is, I'd rather see something that records a timestamp every time `runLoopIteration` is called, then we can check to see if that timestamp is within our timeout window.

internal/pkg/agent/application/monitoring/stats.go
```diff
@@ -12,6 +12,41 @@ import (
...
+
+	userConfig := req.Form.Get(formValueKey)
+	switch userConfig {
+	case "failed":
+		return defaultUserCfg, nil
+	case "degraded":
+		return LivenessFailConfig{Failed: true, Degraded: true}, nil
+	case "":
+		return defaultUserCfg, nil
```
leehinman, line 44: nit

```suggestion
	case "failed", "":
		return defaultUserCfg, nil
	case "degraded":
		return LivenessFailConfig{Failed: true, Degraded: true}, nil
```

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: > so if the coordinator gets blocked after that, you would read from the heartbeatChan and think that it was up, because in the past it had been able to write to the channel. On your next read it would fail. Is that correct?

So, if I understand you correctly, yes. We could end up in a state where the coordinator blocks right after a heartbeat call. Because the liveness endpoint is meant to be repeated on some kind of regular period, I'm not too worried about that.

I specially didn't go with some kind of timestamp mechanism on @faec 's advice, since she was worried about the added complexity/edge cases of a time comparison, compared to a simple signal like this.

internal/pkg/agent/application/monitoring/stats.go
```diff
@@ -12,6 +12,39 @@ import (
...
 
+const formValueKey = "failon"
+
+type LivenessFailConfig struct {
+	Degraded bool `yaml:"degraded" config:"degraded"`
+	Failed   bool `yaml:"failed" config:"failed"`
+}
+
+// process the form values we get via HTTP
+func handleFormValues(req *http.Request) (LivenessFailConfig, error) {
```
cmacknz, line 23: This would make more sense if it lived in liveness.go since that's the only place it is called.

internal/pkg/agent/application/monitoring/stats.go
```diff
@@ -12,6 +12,39 @@ import (
...
+	err := req.ParseForm()
+	if err != nil {
+		return LivenessFailConfig{}, fmt.Errorf("Error parsing form: %w", err)
+	}
+
+	defaultUserCfg := LivenessFailConfig{Degraded: false, Failed: true}
```
cmacknz, line 29: Can the default be to rely only the on the coordinator check? That is, default `Failed: false`?

With the current implementation, I don't see how you could turn off the failed check if it turned out to be problematic. You'd have to stop using the liveness probe.

There are inputs that will fail because of invalid configurations, which isn't correctable via restarting, and also doesn't indicate the agent itself isn't "alive".

_meta/config/common.p2.yml.tmpl
```diff
@@ -66,6 +66,15 @@ inputs:
...
 #   use_output: monitoring
 #   # exposes agent metrics using http, by default sockets and named pipes are used
+#   #
+#   # `http` Also exposes a /liveness endpoint that will return an HTTP code depending on agent status:
+#   # 200: Agent is healthy
+#   # 500: A component or unit is in a failed state
+#   # 503: The agent coordinator is unresponsive
+#   # You can pass a `failon` parameter to the /liveness endpoint to determine what component state will result in a 500.
+#   # For example: `curl 'localhost:6792/liveness?failon=degraded'` will return 500 if a component is in a degraded state.
+#   # The two possible values for `failon` are `degraded` and `failed`. If no `failon` parameter is provided, the default
```
cmacknz, line 76: As mentioned later the default should just be the coordinator liveness check, with degraded and failed as optional additional extensions to it.

internal/pkg/agent/application/monitoring/liveness.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// process the form values we get via HTTP
+func handleFormValues(req *http.Request) (LivenessFailConfig, error) {
+	err := req.ParseForm()
+	if err != nil {
+		return LivenessFailConfig{}, fmt.Errorf("Error parsing form: %w", err)
+	}
+
+	defaultUserCfg := LivenessFailConfig{Degraded: false, Failed: true, Coordinator: true}
```
cmacknz, line 30: We do not want `Failed` to be on by default. If somebody typos an integration configuration, that can restart the agent rapidly, so rapidly it can't process the configuration change from Fleet to correct the typo.

I also don't think we need `coordinator` as an explicit setting, mostly because no user it going to understand what that means and it is leaking implementation details of the agent.

Degraded and Failed are part of the Fleet API contract. We could rename the coordinator at any time and then this setting name makes no sense.

If you want an explicit name for this mode, call it "heartbeat" or something more generic.

fearful-symmetry: Yeah, good point about how `failed` might behave. So, just to be clear, the default behavior should be like this?

- If no `failon` param is specified, _just_ check the coordinator
- if `failon == failed` just check for a `Failed` state
- if `failon == degraded` check for both `Failed` and `Degraded` state

Does that sound right?

internal/pkg/agent/application/coordinator/coordinator.go
leehinman: I think this just tells us that the `runLoopIteration` function can run and the select can hit the heartbeat case.  It doesn't tell us if the coordinator can/has processed any of the other cases. 

I was thinking to be "alive", we want to know that one of the case statements besides heartbeat statement has run.  We would probably need to bound that comparison with the check-in Interval.  Or to put another way, the `runLoopIteration` function should happen at least every check-in Interval (2x is probably safer) and it might happen more frequently.

internal/pkg/agent/application/monitoring/liveness.go
cmacknz: Yes

leehinman (approved):

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: Yeah, I sort of agree, this only really works as a basic heartbeat. 

My concern is that "has the coordinator done anything else?" would be a bit flaky, as we can't really guarantee what state other sub-components will be in at any given time. @faec can comment more, but for that to work we might need to refactor part of the coordinator to allow for a more sophisticated health check.

internal/pkg/agent/application/monitoring/liveness.go
```diff
@@ -0,0 +1,85 @@
...
+		return defaultUserCfg, fmt.Errorf("got unexpected value for `%s` attribute: %s", formValueKey, userConfig)
+	}
+}
+
+func livenessHandler(coord CoordinatorState) func(http.ResponseWriter, *http.Request) error {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+
+		state := coord.State()
+		isUp := coord.CoordinatorActive(time.Second * 10)
```
cmacknz, line 56: You could just immediately return StatusServiceUnavailable here, since if this is false none of the form values matter.

internal/pkg/agent/application/monitoring/liveness.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/elastic/elastic-agent-client/v7/pkg/client"
+)
+
+const formValueKey = "failon"
+
+type LivenessFailConfig struct {
+	Degraded    bool `yaml:"degraded" config:"degraded"`
+	Failed      bool `yaml:"failed" config:"failed"`
+	Coordinator bool `yaml:"coordinator" config:"coordinator"`
```
cmacknz, line 20:
```suggestion
```

You don't actually need this since this check can't be turned off. If you keep it, it should be renamed to Heartbeat.

internal/pkg/agent/application/monitoring/server_test.go
```diff
@@ -5,11 +5,89 @@
...
+	"github.com/elastic/elastic-agent-client/v7/pkg/client"
+	"github.com/elastic/elastic-agent-libs/api"
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator"
+	"github.com/elastic/elastic-agent/internal/pkg/core/monitoring/config"
+	"github.com/elastic/elastic-agent/pkg/component"
+	"github.com/elastic/elastic-agent/pkg/component/runtime"
 )
 
+func TestConfigUpdateOnReload(t *testing.T) {
```
cmacknz, line 27: Am I blind or is this test just starting the server and not actually reloading anything?

cmacknz (commented):
Mostly LGTM, a couple of minor things in particular the changelog has a typo.

internal/pkg/agent/application/monitoring/server_test.go
fearful-symmetry: Yeah, the name is a tad deceptive, the integration test is the only one doing the reloading.

rgarcia89:
@cmacknz is there also a possibility to enable monitoring via the UI of the agent policies?

cmacknz:
Logs and Metrics monitoring has always been possible to configure in the UI.

This PR makes it possible to configure the agent to expose the liveness endpoint via the UI, but what polls the /liveness endpoint is deployment specific and Fleet itself can't/won't monitor it since it is local to the machine the agent is on.

rgarcia89:
Thanks I am well aware about the functionality of the liveness and readiness probes. However adding the probe definition to the daemonset is not enough. By default the agents are not listening on the port which is being used in your snipped. Therefore, I am trying to figure out where to enable the exposure of it, so that it can be used for the probing. I might have overseen the section in the UI to enable it. Could you maybe provide a screenshot?

PS: I am talking about fleet managed agents.

cmacknz:
The first problem you have is that the code in this PR is not released yet, and won't be until 8.15.0. The steps in https://github.com/elastic/elastic-agent/pull/4499#pullrequestreview-2002153481 won't work until then.

The second thing with a Fleet managed agent prior to 8.15.0 is that the `agent.monitoring.http` settings need to be set in the local configuration when the agent container first starts up. This is because you can't change them from Fleet yet. Following the process in https://www.elastic.co/guide/en/fleet/current/advanced-kubernetes-managed-by-fleet.html to modify the initial contents of the elastic-agent.yml should get you around this problem, but the `/liveness` endpoint wont' exist until 8.15.0.

The third thing is that you likely need to change the default host the monitoring server binds to to `0.0.0.0` from `localhost` for the liveness probe to work.

PR #4519, opened 2024-04-05
Fix failing upgrade command when gRPC server interrupts connection

rdner:
## What does this PR do?

The upgrade CLI command creates a client, connects to the Elastic Agent server and triggers the upgrade process which is supposed to shut down the agent and start the upgrade. Sometimes, this happens before the client gets a successful response and the upgrade command returns an error code although the upgrade succeeded.

This change checks if the connection was interrupted and treats it as a successful result instead of error.

## Why is it important?

This race condition occasionally fails our integration tests.

## Checklist

## Related issues

- Fixes https://github.com/elastic/elastic-agent/issues/3890

### DIFF
--- a/testing/upgradetest/upgrader.go
+++ b/testing/upgradetest/upgrader.go
@@ -12,9 +12,12 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/otiai10/copy"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/details"
 	v1client "github.com/elastic/elastic-agent/pkg/control/v1/client"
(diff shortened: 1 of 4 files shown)


testing/upgradetest/upgrader.go
```diff
@@ -325,7 +328,15 @@ func PerformUpgrade(
 
 	upgradeOutput, err := startFixture.Exec(ctx, upgradeCmdArgs)
 	if err != nil {
-		return fmt.Errorf("failed to start agent upgrade to version %q: %w\n%s", endVersionInfo.Binary.Version, err, upgradeOutput)
+		s, ok := status.FromError(err)
```
blakerouse, line 331: Does this really work here? The output from `startFixture.Exec` is an error from a command output and not directly from the GRPC client as the code above is.

blakerouse: Wow really from string output? That is surprising, but awesome if that works!

rdner: Fixed in https://github.com/elastic/elastic-agent/pull/4519/commits/1b17b6da507e7b170e9686f9be14b012d21f19e1

blakerouse (approved):
Looks good, thanks for the fix!

testing/upgradetest/upgrader.go
rdner: Added a test in https://github.com/elastic/elastic-agent/pull/4519/commits/5cb86cd6016a9e100705f8b493ea9cf796b55b22

rdner:
I still need to fix the new test for Windows.

PR #4520, opened 2024-04-05
Extract the bump-agent-versions workflow into a script

rdner:
When the script is in the multiline YAML it breaks the desired formatting of the commit message.
Also editing this script separately is a better developer experience.


rdner:
I created this PR during the Github outage, it's broken and cannot be checked out. Closing and recreating it now.

Issue #4336, opened 2024-02-24
[Flaky Test]: TestDownloadWithRetries – "3" is not greater than or equal to "4"


rdner:
### Failing test case

TestDownloadWithRetries

### Error message

"3" is not greater than or equal to "4"

### Build

https://buildkite.com/elastic/elastic-agent/builds/7478#018ddb31-9e7e-4a35-b4e6-b451d15b8c65

### OS

Windows

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: internal/pkg/agent/application/upgrade TestDownloadWithRetries/download_timeout_expired (0.26s)
    step_download_test.go:255:
        	Error Trace:	C:/buildkite-agent/builds/bk-agent-prod-gcp-1708779493233339393/elastic/elastic-agent/internal/pkg/agent/application/upgrade/step_download_test.go:255
        	Error:      	"3" is not greater than or equal to "4"
        	Test:       	TestDownloadWithRetries/download_timeout_expired
=== FAIL: internal/pkg/agent/application/upgrade TestDownloadWithRetries (0.31s)
```

rdner:
Failed again in https://buildkite.com/elastic/elastic-agent/builds/8160#018eb408-dcf0-4bac-aad0-d7f69d6e3a89

rdner:
Failed again in https://buildkite.com/elastic/elastic-agent/builds/8484#018f0842-7b47-4198-aa95-994495c5782d

Reopening...

rdner:
Another failure https://buildkite.com/elastic/elastic-agent/builds/8519#018f128f-4994-4035-a734-7a7508974df1

pchila:
New failure in https://buildkite.com/elastic/elastic-agent/builds/8550#018f288b-cb51-4b9d-a4f3-813b41c2ffa8

PR #4541, opened 2024-04-08
Make the logs test assertion in upgrade retries deterministic

rdner:
The number of attempts cannot be calculated as
`testCaseSettings.Timeout / testCaseSettings.RetrySleepInitDuration` since the exponential backoff is using
`testCaseSettings.RetrySleepInitDuration` only as an `InitialInterval` and then increases the next delay using multipliers and randomized values.

However, we can tell approximately how many attempts should fit in the timeout.

I validated this by running the test 1000 times and it never failed.

Closes https://github.com/elastic/elastic-agent/issues/4336

### DIFF
--- a/internal/pkg/agent/application/upgrade/step_download_test.go
+++ b/internal/pkg/agent/application/upgrade/step_download_test.go
@@ -6,6 +6,7 @@ package upgrade
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


blakerouse (approved):
Looks good.

PR #4542, opened 2024-04-08
[Kubernetes leader election] Run leader elector at all times

constanca-m:
## What does this PR do?

This PR covers the issue https://github.com/elastic/beats/issues/38543.

The fix was already merged for beats: https://github.com/elastic/beats/pull/38471.

Now we need to align the agent with the same logic as well.

We do that by making sure the leader elector is always running. Previously, when an instance lost the lease, the leader elector would finish running, and it would not start again.

## Checklist

## Related issues

- Closes https://github.com/elastic/beats/issues/38543
- Closes https://github.com/elastic/elastic-agent/issues/4543

## How to test

1. Build elastic-agent:
```bash
EXTERNAL=true SNAPSHOT=true DEV=true PLATFORMS=linux/amd64 PACKAGES=docker mage package
```

2. Create stack and kind cluster, and connect the cluster to the stack:
```bash
elastic-package stack up --version=8.14.0-SNAPSHOT
kind create cluster
docker network connect elastic-package-stack_default kind-control-plane
```

3. Build docker image:
```bash
cd build/package/elastic-agent/elastic-agent-linux-amd64.docker/docker-build
docker build -t custom-agent-image .
```
4. Load this image in your kind cluster:
```bash
kind load docker-image custom-agent-image:latest
```

5. Deploy [agent](https://raw.githubusercontent.com/elastic/elastic-agent/8.13/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml) with that image:
```yaml
containers:
  - name: elastic-agent
    image: custom-agent-image:latest
    imagePullPolicy: Never
```

## Results

The following test was done to make sure the leader elector run again after losing the lease:
1. Create 2 node cluster, so there are two agent running:

2. Change the lease, so that the first leader loses it.
First leader:
```
c@c:~$ kubectl get leases -n kube-system | grep elastic-*
elastic-agent-cluster-leader           elastic-agent-leader-elastic-agent-standalone-kmx9m                         9m24s
```
<details>
<summary>Change the lease holder. You can do it like this.

```yaml
apiVersion: coordination.k8s.io/v1
kind: Lease
metadata:
  name: elastic-agent-cluster-leader
  namespace: kube-system
spec:
  holderIdentity: change
```
</summary></details>

After a while, check the new leader:
```yaml
c@c:~$ kubectl get leases -n kube-system | grep elastic-*
elastic-agent-cluster-leader           elastic-agent-leader-elastic-agent-standalone-67x8n                         11m
```

3. Change the lease again, and make sure the leader gets repeated. This way we know that the leader elector kept running.
```yaml
c@c:~$ kubectl get leases -n kube-system | grep elastic-*
elastic-agent-cluster-leader           elastic-agent-leader-elastic-agent-standalone-kmx9m                         14m
```
As we can see, the leader repeated.


changelog/fragments/1712583231-leader-election-issue.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Make sure leader elector is running at all times
```
blakerouse, line 14:
```suggestion
summary: Fix issue where kubernetes_leaderelection provider would not try to require the lease once lost
```

changelog/fragments/1712583231-leader-election-issue.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/4542
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: https://github.com/elastic/beats/issues/38543
```
blakerouse, line 32: Another issue should be filed with Elastic Agent and linked here.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go
```diff
@@ -104,7 +104,19 @@ func (p *contextProvider) Run(ctx context.Context, comm corecomp.ContextProvider
...
+		case <-ctx.Done():
+			break runLeaderElector
+		default:
+			// Run returned because the lease was lost. Run the leader elector again, so this instance
+			// is still a candidate to get the lease.
+		}
+	}
+
 	p.logger.Debugf("Stopped Leader Elector")
 	return comm.Err()
```
blakerouse, line 121: You can simplify all this logic, to:

```suggestion
        for {
		le.Run(ctx)
		if ctx.Err() != nil {
			p.logger.Debugf("Stopped Leader Elector")
			return comm.Err()
		}
	}
```

blakerouse (changes requested):

changelog/fragments/1712583231-leader-election-issue.yaml
constanca-m: Done

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go
```diff
@@ -104,9 +104,14 @@ func (p *contextProvider) Run(ctx context.Context, comm corecomp.ContextProvider
 		p.logger.Errorf("error while creating Leader Elector: %v", err)
 	}
 	p.logger.Debugf("Starting Leader Elector")
-	le.Run(comm)
-	p.logger.Debugf("Stopped Leader Elector")
-	return comm.Err()
+
+        for {
+		le.Run(ctx)
```
cmacknz, line 109: If there is a cluster error resulting in the leader continuously losing the lease, will this result in attempts to acquire it as quickly as possible with no rate limit?

The implementation of `Run` I see is:

```go
// Run starts the leader election loop. Run will not return
// before leader election loop is stopped by ctx or it has
// stopped holding the leader lease
func (le *LeaderElector) Run(ctx context.Context) {
	defer runtime.HandleCrash()
	defer func() {
		le.config.Callbacks.OnStoppedLeading()
	}()

	if !le.acquire(ctx) {
		return // ctx signalled done
	}
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	go le.config.Callbacks.OnStartedLeading(ctx)
	le.renew(ctx)
}
```

There have been several escalations showing that failing to appropriately rate limit k8s control plane API calls like leader election can destabilize clusters.

What testing have we done to ensure this change won't cause issues like this?

cmacknz (changes requested):
Requesting changes until we get an answer on if this needs rate limiting and the impacts on the k8s control plane traffic.

The current implementation definitely needs to change, but it's not clear to me that this change won't just make rapid calls to acquire leases as fast as possible yet.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go
constanca-m: We have already been doing this. This change will not affect the number of calls, it just makes sure that at least 1 agent will be reporting metrics. The problem with the implementation now is that run goes like this:
- Keeps trying to acquire the lease
- Acquire the lease
- Loose the lease and stop running
All the while, all the other instances were trying to acquire the lease already.
We do not have many SDHs on this bug, which leads me to believe that it is rare for an agent to lose the lease. But we do have problems when agents stop reporting metrics, and the only way we knew how to restart that was to make the pod run again - that is, force `run()` to run again.

constanca-m: There are some parameters in the config regarding the lease:

https://github.com/elastic/elastic-agent/blob/43cb1487314d64b3f87761d9f07dc50b1a382fbd/internal/pkg/composable/providers/kubernetesleaderelection/config.go#L17-L19

If it becomes necessary to reduce the amount of times an agent tries to acquire it.

cmacknz:
Before I remove my requested changes, do we have a way to test this in this repository?

It looks like you'd need a test that you can get the lease again after you've lost it.

I see it is possible to inject the client used in the lock implementation, which takes an interface:

https://github.com/elastic/elastic-agent/blob/09b7410d65b5b0f459175f395bfb6aba2535047d/internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go#L80

I also see a fake client is used in the secrets tests: https://github.com/elastic/elastic-agent/blob/09b7410d65b5b0f459175f395bfb6aba2535047d/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go#L36

Is is possible to mock out the lease client to the test this?

constanca-m:
> Before I remove my requested changes, do we have a way to test this in this repository?

For the beats PR, I did [add an unit test](https://github.com/elastic/beats/pull/38471/files#diff-c3df47b0b6d6f8c6bd72218906c71d2154d36864ae01ae61984434152dfff1aa) to check that. However, we add a way to change the start leading and stop function there, and this repository I cannot get inside those functions to check if that is working. Because of that, I don't think it is possible to test.

There is a way to force the lease to be updated, and sometimes the leader changes. The problem is that we are working with seconds in the lease duration fields, so if I push a test for that, it might take 2min, is that a good idea? I cannot do less that seconds for those fields unlike in the beats PR, because these fields are integers in seconds, and changing them to time.duration means a breaking change. @cmacknz

cmacknz:
> For the beats PR, I did [add an unit test](https://github.com/elastic/beats/pull/38471/files#diff-c3df47b0b6d6f8c6bd72218906c71d2154d36864ae01ae61984434152dfff1aa) to check that. However, we had a way to change the start leading and stop function there, and this repository I cannot get inside those functions to check if that is working. Because of that, I don't think it is possible to test.

Can't you allow access to the lease callbacks you need through an internal constructor, so that the `ContextProviderBuilder` used externally remains unmodified?

> There is a way to force the lease to be updated, and sometimes the leader changes. The problem is that we are working with seconds in the lease duration fields, so if I push a test for that, it might take 2min, is that a good idea? I cannot do less than seconds for those fields unlike in the beats PR, because these fields are integers in seconds, and changing them to time.duration means a breaking change. @cmacknz 

As above you could bypass the Config object in `ContextProviderBuilder(logger *logger.Logger, c *config.Config, managed bool)` using an internal constructor for tests that gives you access to the underlying `time.Duration` fields.

It seems possible to test this here, although it does look like it'll be more work than it was to write than it was in Beats.

The upstream tests are also an interesting source of inspiration https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection_test.go, but they test at a lower level and use a lot more test machinery from their own packages.

blakerouse (changes requested):
You made the requested changes I asked for and those changes look good.

I agree with Craig on all points for adding unit tests to cover this code

constanca-m:
I added a unit test @blakerouse @cmacknz. The test does the following:

> // TestNewLeaderElectionManager will test the leader elector.
// We will try to check if an instance can acquire the lease more than one time. This way, we will know that
// the leader elector starts running again after it has stopped - which happens once a leader looses the lease.
// To make sure that happens we will do the following:
// 1. We will create the lease to be used by the leader elector.
// 2. We will create two context providers - in the default context, this would mean two nodes, each one with an agent running.
// We will wait for one of the agents, agent1, to acquire the lease, before starting the other.
// 3. We force the lease to be acquired by the other agent, agent2.
// 4. We will force the lease to be acquired by the agent1 again. To avoid the agent2 reacquiring it multiple times,
// we will stop this provider and make sure the agent1 can reacquire it.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,218 @@
...
+			if currentHolder == leaderElectorPrefix+podNames[1] {
+				done <- 1
+				break
+			}
+		}
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(time.Duration(leaseDuration+leaseRetryPeriod) * 20 * time.Second):
```
constanca-m, line 165: Technically, this seems possible to happen...

There are only two candidates two acquire the lease, `agent1` and `agent2`. In this case, we need to wait for `agent2` to acquire it. Since I am forcing a new leader, this should imply `agent1` losing the lease, which also means the leader elector needs to start again. This should be enough time for `agent2` to acquire it.

However, I still added the timeout to make sure this test does not run for very, very long. Should I remove this?

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,218 @@
...
+			if currentHolder == leaderElectorPrefix+podNames[0] {
+				done <- 1
+				break
+			}
+		}
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(time.Duration(leaseDuration+leaseRetryPeriod) * 3 * time.Second):
```
constanca-m, line 212: I also added a timeout here to avoid waiting this test running for a long, long time. I cannot think of reasons why this would ever fail here, since we give a long time for the lease to be acquired.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
constanca-m: Windows unit tests hit the timeout... So I removed it. This unit test might take a while to run.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+// TestNewLeaderElectionManager will test the leader elector.
+// We will try to check if an instance can acquire the lease more than one time. This way, we will know that
+// the leader elector starts running again after it has stopped - which happens once a leader looses the lease.
+// To make sure that happens we will do the following:
+// 1. We will create the lease to be used by the leader elector.
+// 2. We will create two context providers - in the default context, this would mean two nodes, each one with an agent running.
+// We will wait for one of the agents, agent1, to acquire the lease, before starting the other.
+// 3. We force the lease to be acquired by the other agent, agent2.
+// 4. We will force the lease to be acquired by the agent1 again. To avoid the agent2 reacquiring it multiple times,
+// we will stop this provider and make sure the agent1 can reacquire it.
```
blakerouse, line 73: This explanation at the top is fantastic, thank you.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+		"leader_leaseduration": leaseDuration,
+		"leader_renewdeadline": leaseRenewDeadline,
+		"leader_retryperiod":   leaseRetryPeriod,
+	}
+	cfg, err := config.NewConfigFrom(c)
+	require.NoError(t, err)
+
+	getK8sClientFunc = func(kubeconfig string, opt autodiscoverK8s.KubeClientOptions) (kubernetes.Interface, error) {
+		return client, nil
+	}
```
blakerouse, line 100: What would be nice would be setting the `getK8sClientFunc` back to the original after the test. Something like:

```
defer func() {
   getK8sClientFunc = defaultGetK8sClientFunc
}()
```

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+	podNames := [2]string{"agent1", "agent2"}
+	cancelFuncs := [2]context.CancelFunc{}
+
+	// Create two leader election providers representing two agents running
+	for i := 0; i < 2; i++ {
+		p, err := ContextProviderBuilder(logger, cfg, true)
+		require.NoError(t, err)
+
+		ctx, cancel := context.WithCancel(context.Background())
+		cancelFuncs[i] = cancel
```
blakerouse, line 112: You should probably still have `defer cancel()`.

If a below `require.*` fails the test will end immediately and those contexts will not be cancelled. I know its just tests, but being pragmatic is always good.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+			_ = p.Run(ctx, comm)
+		}()
+
+		if i == 1 {
+			break
+		}
+
+		// We need to wait for the first agent to acquire the lease, so we can POD_NAME environment variable again
+		expectedLeader := leaderElectorPrefix + podNames[i]
+		for {
```
blakerouse, line 129: What `holder == expectedLeader` doesn't happens here if that never happens? Seems this will loop forever?

might be better to switch to `require.Eventually` with a timeout just to ensure and a break of the code somewhere doesn't have this block the unit test forever

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+			// In this case, we already have an agent as holder
+			if currentHolder == leaderElectorPrefix+podNames[0] || currentHolder == leaderElectorPrefix+podNames[1] {
+				break
+			}
+		}
+
+		if currentHolder == leaderElectorPrefix+podNames[1] {
+			break
+		}
+	}
```
blakerouse, line 161: This forloop has the same issue as above, will loop forever.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+			// In this case, we already have an agent as holder
+			if currentHolder == leaderElectorPrefix+podNames[0] || currentHolder == leaderElectorPrefix+podNames[1] {
+				break
+			}
+		}
+
+		if currentHolder == leaderElectorPrefix+podNames[0] {
+			break
+		}
+	}
```
blakerouse, line 188: This forloop has the same issue as above, will loop forever.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	cancelFuncs[0]()
```
blakerouse, line 190: If you do the `defer cancel()` above you can remove this line.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+		if currentHolder == leaderElectorPrefix+podNames[1] {
+			break
+		}
+	}
+
+	// Now that the holder is agent2, let's wait for agent1 to be reelected.
+	// To avoid having to wait very long, the context of agent2 will be canceled so the leader elector will not be
+	// running anymore. This way there is only one instance fighting to acquire the lease.
+	cancelFuncs[1]()
```
blakerouse, line 166: You can still have this line here if you add a `defer cancel()` above, the defer will just become a noop.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
constanca-m: I added a timeout. I explained in this comment why I removed it before https://github.com/elastic/elastic-agent/pull/4542#discussion_r1557216856.

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,229 @@
...
+			err = applyLease(client, lease, false)
+			require.NoError(t, err)
+
+			var currentHolder string
+			for {
+				currentHolder, err = getLeaseHolder(client)
+				require.NoError(t, err)
+
+				// In this case, we already have an agent as holder
+				if currentHolder == leaderElectorPrefix+podNames[0] || currentHolder == leaderElectorPrefix+podNames[1] {
```
constanca-m, line 170: I am not adding a timeout to this for cycle because this is a sure thing to happen, since there are only 2 pods running to acquire the lease.

changelog/fragments/1712583231-leader-election-issue.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Fix issue where kubernetes_leaderelection provider would not try to require the lease once lost
```
cmacknz, line 14:
```suggestion
summary: Fix issue where kubernetes_leaderelection provider would not try to reacquire the lease once lost
```

internal/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection_test.go
```diff
@@ -0,0 +1,229 @@
...
+	holder := lease.Spec.HolderIdentity
+	if holder == nil {
+		return "", err
+	}
+	return *holder, nil
+}
+
+// TestNewLeaderElectionManager will test the leader elector.
+// We will try to check if an instance can acquire the lease more than one time. This way, we will know that
+// the leader elector starts running again after it has stopped - which happens once a leader looses the lease.
```
cmacknz, line 67:
```suggestion
// the leader elector starts running again after it has stopped - which happens once a leader loses the lease.
```

cmacknz:
Added the backport-v8.14.0 label as this is a bug fix. Feel free to remove it if you do not want this in 8.14.0.

blakerouse (approved):
Looks good.

PR #4550, opened 2024-04-09
Add build for main and release branches on elastic-agent-core DRA pipeline

pchila:
## What does this PR do?

This PR enables build on branches for publishing elastic-agent-core DRA.

## Why is it important?

We need regular `elastic-agent-core` publishing to use the artifacts for elastic-agent early releases.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## Related issues

- Relates #4403 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4555, opened 2024-04-09
fix branch_configuration location in catalog-info

jonahbull:
The `branch_configuration` property needs to be under the `spec` block, otherwise terrazzo refuses to parse the config because it's invalid.

Related: https://github.com/elastic/elastic-agent/pull/4550


cmacknz (approved):
Is there anyway for this to have been found before we merged it?

cmacknz:
I'm just going to force merge this, we don't need to wait for the rest of CI on this which is totally unaffected by this change.

PR #4581, opened 2024-04-16
Delete resources created by OGC more than 5 hours ago

rdner:
Previously it was set to 24 hours but since our build load hits some resource limits, the configuration should be more aggressive.


pazone (approved):

pazone:
LGTM

PR #4584, opened 2024-04-16
Exclude cloud-defend from linux .tar.gz packages

pchila:
## What does this PR do?

This change filters out `cloud-defend` and `cloud-defend.spec.yml` when packaging .tar.gz archives

## Why is it important?
Cloud defend is needed only on docker images, removing it reduces the size of linux .tar.gz package by about 40 MB. 

## Checklist

## Author's Checklist

- [ ]

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1713289288-Removed-cloud-defend-from-.tar.gz-archives.yaml
+++ b/changelog/fragments/1713289288-Removed-cloud-defend-from-.tar.gz-archives.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: enhancement
 
 # Change summary; a 80ish characters long description of the change.
 summary: Removed cloud-defend from .tar.gz archives, now it only appears in docker images
--- a/changelog/fragments/1713289288-Removed-cloud-defend-from-.tar.gz-archives.yaml
+++ b/changelog/fragments/1713289288-Removed-cloud-defend-from-.tar.gz-archives.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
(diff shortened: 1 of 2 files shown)


dev-tools/mage/pkgtypes.go
```diff
@@ -904,15 +905,24 @@ func addFileToZip(ar *zip.Writer, baseDir string, pkgFile PackageFile) error {
...
 			if pkgFile.SkipOnMissing && os.IsNotExist(err) {
 				return nil
 			}
-
 			return err
 		}
 
+		if slices.Contains(excludedFiles, info.Name()) {
+			// it's a file we have to exclude
+			return nil
```
cmacknz, line 923: We should log what we are ignoring so we don't forget this is happening.

pchila: Done in 5d67bf4e97e7b7f8d32cdfd8fff93dc6294e9b5d

changelog/fragments/1713289288-Removed-cloud-defend-from-.tar.gz-archives.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11:
```suggestion
kind: enhancement
```

dev-tools/mage/pkgtypes.go
blakerouse: This seems like a hack.

I think it would be better to define this higher up so its clean and clearly defined why its included and not. At this level it just seems like a hack.

pierrehilbert: Actually this is a hack: the best solution would be to be able to have a clear way to define package for docker builds compare to other builds but this would require a way bigger change.

cmacknz: This is definitely a hack, but it's a small hack that helps solve a customer problem.

I'm going to approve this but the hack version of this can't live for very long.

cmacknz (approved):

dev-tools/mage/pkgtypes.go
pchila: This is a hack but it is a quick win to reduce the package size for linux... I filed a new issue to replace this change with something more structured https://github.com/elastic/elastic-agent/issues/4588

pchila:
buildkite test this

Issue #4582, opened 2024-04-16
Disable automatic reload of the agent monitoring HTTP server until handling of default values is figured out 


cmacknz:
In https://github.com/elastic/elastic-agent/pull/4499 we added a liveness probe and had the agent properly support reloading the HTTP monitoring server settings so that Fleet managed agents that defaulted to having the server off can turn it on.

This however has caused several unexpected problems, because there are agents (including the agent used to run Fleet Server in ESS) that have enabled HTTP monitoring in the initial agent configuration expecting it to stay enabled. When we enabled reload, what happened is the agent began respecting the default fleet policy which has HTTP monitoring disabled by default, thus turning off the HTTP server with no simple way in Fleet to turn it back on.

As a short term solution while we figure out how to make the default settings work correctly (I believe for ESS we need a way for the default fleet policy used to specify HTTP monitoring server settings), we need to keep the new liveness probe but disable reloading the HTTP monitoring server settings again.

cmacknz:
Another alternative would be to treat an unset HTTP configuration as equivalent to a no-op, falling back to what the agent was installed at. We would only turn it off if we get an explicit set of settings from Fleet. This is similar to what we have to do with proxy settings which have the exact same problem.

That is we interpret configuration from fleet being missing as no-op, the problem is our config unpacking doesn't represent this unset state and likely unpacks into default values of empty strings and false booleans.

cmacknz:
While we work this out let's revert https://github.com/elastic/elastic-agent/pull/4499 and then we can bring it back with a fix in case it takes longer to fix than we realize.

Also all the integration tests will fail until we get this reverted.

cmacknz:
Revert PR https://github.com/elastic/elastic-agent/pull/4583

fearful-symmetry:
Alright, just so I'm clear on the chain of events, it's something like this:

1) An initial `elastic-agent.yml` text config sets the monitoring enabled flag
2) We then get a fleet policy
3) the fleet policy has no `http.monitoring` section, so we default to off
4) Now that reloading works properly, we reload, and disable monitoring

I agree that the best way to handle this is to is to treat an unset config as effectively a no-op, and leave it enabled. That shouldn't be too difficult to implement. 

Also agree that we really need https://github.com/elastic/elastic-agent/issues/3511 ....

cmacknz:
Yes your understanding is correct.

[...]

PR #4583, opened 2024-04-16
Revert "Add `/liveness` endpoint to elastic-agent (#4499)"

cmacknz:
This reverts commit 29ce53e0c6d8124f37774d7cf70321f111fd8241.

We need this to fix 8.14.0-SNAPSHOT ESS deployments.

PR #4586, opened 2024-04-16
Fix liveness reload config handling

fearful-symmetry:
## What does this PR do?

Closes https://github.com/elastic/elastic-agent/issues/4582

This reverts the revert in https://github.com/elastic/elastic-agent/pull/4583 and adds some extra logic so that if the HTTP monitoring server is enabled, and we get a config reload that _does not explicitly set it to disabled_ we keep it enabled.

This also adds a ton of tests, because now I'm paranoid.

## Checklist

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -279,6 +279,11 @@ type Coordinator struct {
 
 	// mx         sync.RWMutex
 	// protection protection.Config
+
+	// a sync channel that can be called by other components to check if the main coordinator
+	// loop in runLoopIteration() is active and listening.
+	// Should only be interacted with via CoordinatorActive() or runLoopIteration()
+	heartbeatChan chan struct{}
 }
 
 // The channels Coordinator reads to receive updates from the various managers.
@@ -372,6 +377,7 @@ func New(logger *logger.Logger, cfg *configuration.Configuration, logLevel logp.
 		logLevelCh:         make(chan logp.Level),
 		overrideStateChan:  make(chan *coordinatorOverrideState),
 		upgradeDetailsChan: make(chan *details.Details),
+		heartbeatChan:      make(chan struct{}),
 	}
 	// Setup communication channels for any non-nil components. This pattern
 	// lets us transparently accept nil managers / simulated events during
@@ -410,6 +416,22 @@ func New(logger *logger.Logger, cfg *configuration.Configuration, logLevel logp.
 // Called by external goroutines.
 func (c *Coordinator) State() State {
 	return c.stateBroadcaster.Get()
+}
+
+// CoordinatorActive is a blocking method that waits for a channel response
+// from the coordinator loop. This can be used to as a basic health check,
+// as we'll timeout and return false if the coordinator run loop doesn't
+// respond to our channel.
+func (c *Coordinator) CoordinatorActive(timeout time.Duration) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	select {
+	case <-c.heartbeatChan:
+		return true
+	case <-ctx.Done():
+		return false
+	}
 }
 
 func (c *Coordinator) RegisterMonitoringServer(s configReloader) {
@@ -976,6 +998,8 @@ func (c *Coordinator) runLoopIteration(ctx context.Context) {
 
 	case upgradeDetails := <-c.upgradeDetailsChan:
 		c.setUpgradeDetails(upgradeDetails)
+
+	case c.heartbeatChan <- struct{}{}:
 
 	case componentState := <-c.managerChans.runtimeManagerUpdate:
 		// New component change reported by the runtime manager via
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -418,11 +418,11 @@ func (c *Coordinator) State() State {
 	return c.stateBroadcaster.Get()
 }
 
-// CoordinatorActive is a blocking method that waits for a channel response
+// IsActive is a blocking method that waits for a channel response
 // from the coordinator loop. This can be used to as a basic health check,
 // as we'll timeout and return false if the coordinator run loop doesn't
 // respond to our channel.
-func (c *Coordinator) CoordinatorActive(timeout time.Duration) bool {
+func (c *Coordinator) IsActive(timeout time.Duration) bool {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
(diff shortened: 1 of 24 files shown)


cmacknz:
The instructions in https://github.com/elastic/elastic-agent?tab=readme-ov-file#testing-on-elastic-cloud should let you create a cloud deployment that uses this version of the agent.

You can use that to confirm integrations server comes up as healthy.

cmacknz:
I tested this locally and it doesn't seem to work. I installed a standalone agent with the HTTP server enabled

In elastic-agent.yml:

```yml
agent.monitoring:
  enabled: true
  http:
      enabled: true
      port: 6791
```

```sh
❯ sudo ./elastic-agent install -f
❯ curl localhost:6791/processes
{"processes":[{"id":"system/metrics-default","binary":"metricbeat","source":{"kind":"configured","outputs":["default"]}},{"id":"filestream-monitoring","binary":"filebeat","source":{"kind":"internal","outputs":["monitoring"]}},{"id":"beat/metrics-monitoring","binary":"metricbeat","source":{"kind":"internal","outputs":["monitoring"]}},{"id":"http/metrics-monitoring","binary":"metricbeat","source":{"kind":"internal","outputs":["monitoring"]}}]}%
```

Then I enrolled it in fleet with the default policy:

```sh
sudo elastic-agent enroll --url=https://1cdab201df9f4dbb936887167f9d4aa2.fleet.eastus2.staging.azure.foundit.no:443 --enrollment-token=XXXX
```

And I can't hit the processes endpoint anymore:

```sh
❯ curl localhost:6791/liveness
404 page not found
```

The pre-config.yaml has:

```yaml
agent:
    download:
        sourceURI: https://artifacts.elastic.co/downloads/
    features: null
    monitoring:
        enabled: true
        logs: true
        metrics: true
        namespace: default
        use_output: default
```

The local-config.yaml has:

```yaml
    monitoring:
        diagnostics:
            limit:
                burst: 1
                interval: 1m0s
            uploader:
                initdur: 1s
                maxdur: 10m0s
                maxretries: 10
        enabled: true
        http:
            buffer: null
            enabled: true
            host: localhost
            port: 6791
```

fearful-symmetry:
I'm still trying to figure out what's going on here. @cmacknz it looks like you're using staging.found.no for the fleet server, but the 8.14 snapshot build seems broken, probably as fallout from this PR. Did you just get lucky with the cloud or am I being dumb?

fearful-symmetry:
Interestingly, I can't reproduce this with the local file config, despite the fact that it should be the same reloading mechanism passing the same config.

fearful-symmetry:
Alright, it's getting late. Hopefully tomorrow the snapshots will be updated and I'll have a few more options for reproducing this.

cmacknz:
> I'm still trying to figure out what's going on here. @cmacknz it looks like you're using staging.found.no for the fleet server, but the 8.14 snapshot build seems broken, probably as fallout from this PR. Did you just get lucky with the cloud or am I being dumb?

I created my cloud deployment before the problem was introduced.

Using an 8.13.3-SNAPSHOT deployment or a stack from elastic-package are ways around this if the 8.14.0-SNAPSHOT isn't fixed tomorrow.

fearful-symmetry:
Alright, can FINALLY reproduce this.

cmacknz:
Now that I remember more on enroll we actually replace the current agent configuration with a blank one before getting it from Fleet.

This happens in a non-obvious place here: https://github.com/elastic/elastic-agent/blob/eca5bc714b9a5569e733644641deabbf4b949e3b/internal/pkg/agent/cmd/enroll.go#L497-L502

The starting configuration is here: https://github.com/elastic/elastic-agent/blob/eca5bc714b9a5569e733644641deabbf4b949e3b/_meta/elastic-agent.fleet.yml#L1-L7

cmacknz:
The thing is this reset of the configuration happened even without your reload changes so I'm now confused how monitoring remained enabled in cloud to begin with.

fearful-symmetry:
So, in the code currently in main, the HTTP config gets set remarkably early, in `loadConfig()`: https://github.com/elastic/elastic-agent/blob/eca5bc714b9a5569e733644641deabbf4b949e3b/internal/pkg/agent/cmd/run.go#L387

This eventually loads a bunch of the state from the encrypted state store. This all happens before fleet starts up, I think. I assume what's happening is that that when we install, the config in the `.enc` files still have our old hard-coded config, and those are acting as overrides that get passed to the HTTP monitoring config. In this PR, we don't care about the state store, we just get our config from the reloader, which gets it from fleet.

pchila:
@fearful-symmetry  Maybe you can test this PR with https://github.com/elastic/elastic-agent/blob/main/testing/environments/cloud/README.md ?

fearful-symmetry:
Alright @cmacknz this should work. 

However, this raises a few more implementation questions, since it means we need to actually care about not _just_ the fleet config, but also the config at init time. Do we try to merge them somehow? Just care about the `enabled` value?

cmacknz:
> However, this raises a few more implementation questions, since it means we need to actually care about not just the fleet config, but also the config at init time. Do we try to merge them somehow? Just care about the enabled value?

The version of this change that is not breaking for anybody is to work exactly the way it did before, except now it respects changed configurations from fleet and standalone configuration reloads.

That is we want the init time behavior to be unchanged, only what happens after that needs to change.

fearful-symmetry:
Alright, so, this should be a relatively reasonable default. If there's no fleet config set, just revert to the original overrides settings.

fearful-symmetry:
Alright, well, I've tested this, but I'm having some permissions issues with the [cloud testing component](https://github.com/elastic/elastic-agent?tab=readme-ov-file#testing-on-elastic-cloud), the `cloud:push` command just results in `denied: requested access to the resource is denied`, despite setting the API key. Either I'm doing something dumb or there's an extra layer of permissions needed somewhere in my cloud account.

fearful-symmetry:
Alright, I added a second integration test that checks for the condition that gave us so much trouble. The test sets a text config, installs agent, makes sure monitoring is still enabled, then uses the overrrides to change the port.

pierrehilbert:
@fearful-symmetry is it ready for review then?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -412,6 +418,22 @@ func (c *Coordinator) State() State {
 	return c.stateBroadcaster.Get()
 }
 
+// CoordinatorActive is a blocking method that waits for a channel response
+// from the coordinator loop. This can be used to as a basic health check,
+// as we'll timeout and return false if the coordinator run loop doesn't
+// respond to our channel.
+func (c *Coordinator) CoordinatorActive(timeout time.Duration) bool {
```
blakerouse, line 425: Why is this prefixed with `Coordinator`? It is the coordinator, why not just `Active()`?

blakerouse (approved):
Looks good. Just a comment on the name of a function.

Overall this is easy to follow and very well tested.

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: yeah, good point

PR #2790, opened 2023-06-06
Remove race detector platform restrictions and enable it in CI

faec:
(copied from @cmacknz at https://github.com/elastic/elastic-agent/pull/2743 so I can work through the remaining failed cases):

We discovered in a recent PR that the RACE_DETECTOR environment variable seemingly was not having any effect (https://github.com/elastic/elastic-agent/pull/2729#issuecomment-1563320866), it turns out it is ignored on arm64. This is now an unnecessary restriction.

I've also enabled the race detector in CI, I'll need to see what this does to the test time on each platform first though. We will keep it at least for Linux.

PR #4621, opened 2024-04-25
Add `unprivileged` and `privileged` subcommand to Elastic Agent

blakerouse:
## What does this PR do?

Adds a new `unprivileged` and `privileged` subcommand that converts an installed Elastic Agent from privileged to unprivileged or back again.

## Why is it important?

To allow users to change the permission execution mode without having to re-install the Elastic Agent.

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ **(will have integration tests)**

## Author's Checklist

- [x] can convert from privileged to unprivileged
- [x] can convert from unprivileged to privileged

```
$ sudo elastic-agent unprivileged
```

```
$ sudo elastic-agent privileged
```

## Related issues

- Closes [#2790](https://github.com/elastic/ingest-dev/issues/2790)


blakerouse:
SonarQube you are correct about unit tests, but you are very wrong about integration tests...

This has integration tests for both privileged and unprivileged going in both directions.

pkg/control/v2/client/wait/agent.go
```diff
@@ -0,0 +1,60 @@
...
+	"errors"
+	"time"
+)
+
+// ForAgent waits for the agent daemon to be able to be communicated with.
+func ForAgent(ctx context.Context, timeout time.Duration) error {
+	if timeout == 0 {
+		timeout = 1 * time.Minute // default of 1 minute
+	}
+	if timeout > 0 {
```
michel-laterman, line 18: I don't think we need this condition

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func privilegedCmd(streams *cli.IOStreams, cmd *cobra.Command) (err error) {
+	isAdmin, err := utils.HasRoot()
+	if err != nil {
+		return fmt.Errorf("unable to perform privileged command while checking for root/Administrator rights: %w", err)
+	}
+	if !isAdmin {
+		return fmt.Errorf("unable to perform privileged command, not executed with %s permissions", utils.PermissionUser)
+	}
```
michel-laterman, line 57: Should we detect if the agent is in privileged mode and nop/error out (and do something similar for the unprivileged command)?

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// TODO(blakerouse): More work to get this working on macOS.
+	// Need to switch the vault from keystore based to file based vault.
+	if runtime.GOOS == "darwin" {
+		return errors.New("unable to perform unprivileged on macOS (not supported)")
+	}
+
+	// cannot switch to unprivileged when Elastic Defend exists in the policy
+	err = ensureNoElasticDefend()
```
michel-laterman, line 70: 👍

internal/pkg/agent/cmd/privileged.go
blakerouse: I didn't do that on purpose, because its possible that the Elastic Agent is not running at the time that this command is ran. The only reliable way to check is to ask a running Elastic Agent.

This command is built in a way to fix a broken switch. Say you where trying to switch from privileged to unprivileged and mid way the power goes out and now the installation is really broken. You can now just call this command again and it will ensure that it fixes the permissions and gets the Elastic Agent started.

I explain in the `--help` information for this command that it doesn't check and that it always restarts the running Elastic Agent even it its already operating in this mode.

pkg/control/v2/client/wait/agent.go
blakerouse: It is doing exactly that. If `-1` timeout is provided then it will run until the `ctx` is cancelled.

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,127 @@
...
+	endpoint = "endpoint"
+)
+
+func newUnprivilegedCommandWithArgs(s []string, streams *cli.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unprivileged",
+		Short: "Switch installed Elastic Agent to run as unprivileged",
+		Long: `This command converts the installed Elastic Agent from running privileged to running as unprivileged.
+
+By default this command will ask or a confirmation before making this change. You can bypass the confirmation request
```
kaanyalti, line 33:
```suggestion
By default this command will ask for a confirmation before making this change. You can bypass the confirmation request
```

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/elastic/elastic-agent/pkg/utils"
+)
+
+func newPrivilegedCommandWithArgs(s []string, streams *cli.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "privileged",
+		Short: "Switch installed Elastic Agent to run as privileged",
+		Long: `This command converts the installed Elastic Agent from running unprivileged to running as privileged.
+
+By default this command will ask or a confirmation before making this change. You can bypass the confirmation request
```
kaanyalti, line 29:
```suggestion
By default this command will ask for a confirmation before making this change. You can bypass the confirmation request
```

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,127 @@
...
+using the -f flag. This will always stop the running Elastic Agent (if running) before performing the switch it is
```
kaanyalti, line 34:
```suggestion
using the -f flag. This will always stop the running Elastic Agent (if running) before performing the switch. It is
```

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+using the -f flag. This will always stop the running Elastic Agent (if running) before performing the switch it is
```
kaanyalti, line 30:
```suggestion
using the -f flag. This will always stop the running Elastic Agent (if running) before performing the switch. It is
```

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+
+	pt := install.CreateAndStartNewSpinner(streams.Out, "Converting Elastic Agent to privileged...")
+	err = install.SwitchExecutingMode(topPath, pt, "", "")
+	if err != nil {
+		// error already adds context
+		return err
+	}
+
+	// wait for the service
+	if daemonTimeout >= 0 {
```
kaanyalti, line 86: Nitpicking here. Do we have to check `daemonTimeout >= 0` since the default is 0?

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,127 @@
...
+
+	pt := install.CreateAndStartNewSpinner(streams.Out, "Converting Elastic Agent to unprivileged...")
+	err = install.SwitchExecutingMode(topPath, pt, install.ElasticUsername, install.ElasticGroupName)
+	if err != nil {
+		// error already adds context
+		return err
+	}
+
+	// wait for the service
+	if daemonTimeout >= 0 {
```
kaanyalti, line 97: Same question as https://github.com/elastic/elastic-agent/pull/4621/files?diff=split&w=1#r1603913060

internal/pkg/agent/cmd/privileged.go
blakerouse: Yes because they can provide a negative number which means to not wait.

internal/pkg/agent/cmd/unprivileged.go
blakerouse: Yes because they can provide a negative number which means to not wait.

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+
+func newPrivilegedCommandWithArgs(s []string, streams *cli.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "privileged",
+		Short: "Switch installed Elastic Agent to run as privileged",
+		Long: `This command converts the installed Elastic Agent from running unprivileged to running as privileged.
+
+By default this command will ask for a confirmation before making this change. You can bypass the confirmation request
+using the -f flag. This will always stop the running Elastic Agent (if running) before performing the switch. It is
+possible that loss of metrics could occur during this small window of time. The command also always starts the
```
cmacknz, line 31: > possible that loss of metrics could occur during this small window of time

Loss of metrics or data in general, if this is intended to be a warning that this is not a zero downtime operation you could just say that directly.

cmacknz: For both commands you may want to caution about what happens if their execution is interrupted.

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,127 @@
...
+	l, err := newErrorLogger()
+	if err != nil {
+		return fmt.Errorf("failed to create error logger: %w", err)
+	}
+	comps, err := getComponentsFromPolicy(ctx, l, paths.ConfigFile(), 0)
+	if err != nil {
+		return fmt.Errorf("failed to create component model from policy: %w", err)
+	}
+	for _, comp := range comps {
+		if comp.InputSpec != nil && comp.InputSpec.InputType == endpoint {
```
cmacknz, line 122: Is there a way to actually evaluate the runtime preventions and see if something has the root one?

That would be a bit more future proof. The constrain isn't really "is defend running" it's "is there something that requires root to function" running, which today is only defend.

internal/pkg/agent/cmd/unprivileged.go
aleksmaus: Normally osquery is running as privileged. Quite a lot of functionality will be unavailable if running unprivileged or returning different results. 

Osqueryd also checks for "safe permissions" when it loads our custom extension.

```
 When osquery launches, each path is evaluated for "safe permissions" 
(extension executable files must be owned by root or Administrator) 
and executed as a monitored child process
```
https://osquery.readthedocs.io/en/stable/deployment/extensions/#autoloading-extensions

```
If you observe a runtime error from osquery, Extension binary has unsafe permissions, 
you have to lock down the filesystem permissions on the extension executable. 
See the steps in "Extensions Binary Permissions," above.
```
https://osquery.readthedocs.io/en/stable/deployment/extensions/#troubleshooting

kaanyalti (approved):

internal/pkg/agent/cmd/unprivileged.go
blakerouse: I think we need to determine if that is the behavior that we actually want. An input specification that requires root, when switched to unprivileged mode will still allow the Elastic Agent to run and the input will just report as failed. This would then be viewable to the user in `status` as well as in Fleet and would allow them to adjust/remove the integration.

In the case of Endpoint it is more detrimental to switch the mode with it enabled. Endpoint being a separate piece of software that is installed, once switched it cannot be uninstalled. That leaves the Elastic Agent in a completely broken state, unless switched back to privileged and then remove the integration.

I believe that we have 4 options for handling this:

1. Prevent switch if any input has a runtime protection for requiring root.
2. Prevent switch if any service input has a runtime protection for requiring root.
3. Add new option to input specification that says it doesn't support unprivileged mode directly. This is a little different then a runtime protection, being that it could be used in switch to prevent the switch. Endpoint specification can set this specifically and then future inputs can as well.
4. Keep it how it is at the moment and hard-code it to endpoint.

aleksmaus: > That would be a bit more future proof. The constrain isn't really "is defend running" it's "is there something that requires root to function" running, which today is only defend.

Osquery as well

internal/pkg/agent/cmd/unprivileged.go
cmacknz: For osquery, you should mark the integration as requiring root privileges like defend does in https://github.com/elastic/endpoint-package/blob/ce921805467ab55953adc3721e7db8581915bd35/package/endpoint/manifest.yml#L23-L25. You can also do this for individual data streams but I don't think that applies to OSquery.

You could also add a prevention to the spec file to prevent us from even starting osquery if you aren't root, but since it doesn't fundamentally not work when not root, this might not be necessary.

cmacknz: If you do update the spec, you would want to update the agentbeat spec.

internal/pkg/agent/cmd/unprivileged.go
cmacknz: > 3. Add new option to input specification that says it doesn't support unprivileged mode directly.

Let's avoid building something new for information we already have, so this option is out.

> This would then be viewable to the user in status as well as in Fleet and would allow them to adjust/remove the integration.

Agree with this, along the lines of not building anything new, we can rely on this to inform the user after the fact. When this is exposed from Fleet, Fleet will have information about the installed packages and will likely warn or prevent the user in the same way. Most users are going to do this en-mass at won't be reading warning text anyway.

> Prevent switch if any service input has a runtime protection for requiring root.
> Keep it how it is at the moment and hard-code it to endpoint.

Given the distinction that a service input may be totally broken if we do this, it makes sense to only add additional functionality for service components to deal with this distinction.

If it isn't a lot of effort to check the spec file so we don't have to hard code this case, then doing this in a general way is preferable.

Otherwise given endpoint is the only service component we expect to have for the foreseeable future, we can leave this as is and add a comment explaining why it specifically is special cased.

blakerouse: > > That would be a bit more future proof. The constrain isn't really "is defend running" it's "is there something that requires root to function" running, which today is only defend.
> 
> Osquery as well

@aleksmaus Root is required to function or it doesn't work well without root? Trying to understand the difference here.

blakerouse: Another option is to allow the switch with Endpoint enabled, but Elastic Agent uninstall Endpoint just like the `uninstall` command does. It can perform uninstall and then switch to non-root. This provides more of a way of not blocking the user from switching, we could even prompt if they want to remove it an switch. [y/N] could be the default with being a force `-f` switch where it always uninstalls it.

I like this option a little more because it doesn't prevent the switch just because a service component that requires root is enabled.

aleksmaus: > @aleksmaus Root is required to function or it doesn't work well without root? Trying to understand the difference here.

We could run osqueryd binary standalone without root permissions and some of the functionality will not be available and some of the queries will return different results (with no errors). 

So there is no indication from the user point of view that something is wrong with permissions, the data just would be missing, which creates "blind spots" from security point of view without user being aware. 

The file system searches would be missing some valid results because of the lack of the permissions, and osquery doesn't return the error in these cases, it just returns a smaller resultset.

Here are just a few examples of running osquery with sudo and without on the same query:

Screenshot 2024-05-20 at 9 52 11 AM


Screenshot 2024-05-20 at 10 02 48 AM


Screenshot 2024-05-20 at 10 04 02 AM


Screenshot 2024-05-20 at 10 13 00 AM


Also the file permissions for our extensions should be set to root or Administrator:
https://osquery.readthedocs.io/en/stable/deployment/extensions/#autoloading-extensions
```
(extension executable files must be owned by root or Administrator)
```

blakerouse: @aleksmaus I think it's best to move this to its own issue and discussion on osquerybeat as non-root. Its could be possible to allow it to run or not run. I believe that discussion is better to be had in a specific issue and then come to a conclusion. We can adjust in a follow up PR based on that.

testing/integration/switch_privileged_test.go
```diff
@@ -0,0 +1,139 @@
...
+	switch runtime.GOOS {
+	case define.Linux:
+		basePath = `/usr`
+	case define.Windows:
+		basePath = `C:\`
+	default:
+		// Set up random temporary directory to serve as base path for Elastic Agent
+		// installation.
+		tmpDir := t.TempDir()
+		basePath = filepath.Join(tmpDir, strings.ToLower(randStr(8)))
```
pchila, line 110: Why do we need to add a randomizedString to `tmpDir` ? `tmpDir` will already contain the testcase name and a progressive number...

On the same topic: why don't we use TmpDir for all OSes ? Does it not work on windows/linux ?

internal/pkg/agent/cmd/unprivileged.go
cmacknz: > Another option is to allow the switch with Endpoint enabled, but Elastic Agent uninstall Endpoint just like the uninstall command does. It can perform uninstall and then switch to non-root. This provides more of a way of not blocking the user from switching, we could even prompt if they want to remove it an switch. [y/N] could be the default with being a force -f switch where it always uninstalls it.

You would have to include and test that tamper protection still applies on this path which will be extra testing we have to keep around. I think it is much simpler if we force users to put unprivileged agents in policies that do not include Defend. There are no edge cases around Defend being conditionally uninstalled or removed that way. The composable agent policies work in the UI will help users deal with any shared content between the privileged and privileged policies.

So I'd say we should continue to refuse the conversion if defend is in the policy.

blakerouse: Okay. I was just worrying that it prevents issues with different personas in an organization. A persona that manages the policies and a persona that manages installation of the endpoints. That is probably rare and not something we need to worry about.

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// wait for the service
+	if daemonTimeout >= 0 {
+		pt.Describe("Waiting for running service")
+		ctx := handleSignal(context.Background()) // allowed to be cancelled
+		err = wait.ForAgent(ctx, daemonTimeout)
+		if err != nil && !errors.Is(err, context.Canceled) {
+			pt.Describe("Failed waiting for running service")
+			return err
+		}
```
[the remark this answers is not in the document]
blakerouse: I updated this. Thanks for pointing that out.

pkg/control/v2/client/wait/agent.go
blakerouse: Added it to `--help` output.

internal/pkg/agent/cmd/install.go
```diff
@@ -230,7 +231,7 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 			defer func() {
 				if err != nil {
 					progBar.Describe("Stopping Service")
-					innerErr := install.StopService(topPath)
+					innerErr := install.StopService(topPath, 30*time.Second, 250*time.Millisecond)
```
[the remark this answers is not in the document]
blakerouse: Done.

internal/pkg/agent/cmd/privileged.go
```diff
@@ -0,0 +1,98 @@
...
+		Run: func(c *cobra.Command, args []string) {
+			if err := privilegedCmd(streams, c); err != nil {
+				fmt.Fprintf(streams.Err, "Error: %v\n%s\n", err, troubleshootMessage())
+				os.Exit(1)
+			}
+		},
+	}
+
+	cmd.Flags().BoolP("force", "f", false, "Do not prompt for confirmation")
+	cmd.Flags().DurationP("daemon-timeout", "", 0, "Timeout waiting for Elastic Agent daemon")
```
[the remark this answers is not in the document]
blakerouse: Updated close to this, with explaining -1 option.

pkg/control/v2/client/wait/common.go
```diff
@@ -0,0 +1,43 @@
...
+const (
+	daemonTimeout = 30 * time.Second // max amount of for communication to running Agent daemon
+)
+
+type waitResult struct {
+	err error
+}
+
+func getDaemonState(ctx context.Context) (*client.AgentState, error) {
+	ctx, cancel := context.WithTimeout(ctx, daemonTimeout)
```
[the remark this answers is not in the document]
blakerouse: Done.

testing/integration/switch_privileged_test.go
blakerouse: This comes from the install tests, there are reasons for using different directories. It has to do with the permissions on each OS. Possibly something that could be changed, but this PR is already large enough and not something I want to change here. If they are changed here then they should be changed in all the install tests as well.

pchila (approved):

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,141 @@
...
+			pt.Describe("Failed waiting for running service")
+			return err
+		}
+		pt.Describe("Service is up and running")
+	}
+
+	return nil
+}
+
+func ensureNoServiceComponentIssues() error {
```
cmacknz, line 105: I don't think this has any test coverage. You could add a check in the endpoint test suite that the unprivileged command fails, or add a unit test.

blakerouse:
I have validated manually that this works on macOS, because we do not have integration testing coverage on macOS. I was able to switch between the modes successfully and the vaults are handled properly.

internal/pkg/agent/cmd/unprivileged.go
blakerouse: True I will add an integration test, as I would prefer to see a true scenario be tested.

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -0,0 +1,141 @@
...
+		},
+	}
+
+	cmd.Flags().BoolP("force", "f", false, "Do not prompt for confirmation")
+	cmd.Flags().DurationP("daemon-timeout", "", 0, "Timeout waiting for Elastic Agent daemon restart after the change is applied (-1 = no wait)")
+
+	return cmd
+}
+
+func unprivilegedCmd(streams *cli.IOStreams, cmd *cobra.Command) (err error) {
```
michalpristas, line 50: nit: seems like there's quite some repetition meaning this and privilegedCmd function

michalpristas (approved):

jlind23:
@ycombinator feel free to force merge this is required as it seems sonar coverage is the only missing condition.
There are tests it is just that sonar calculates coverage using exclusively unit and not integration tests.

PR #4633, opened 2024-04-29
Reduce agent logs by default

### LINKED ISSUE #4252: Reduce the amount the agent logs by default
**Describe the enhancement:**

We'd like an audit to occur on the current logged messages to make sure that the information at the default INFO log level is useful and is valuable when shipped to Fleet.

10-15% of data produced by Agent is internal logs and metrics, we'd like to reduce the volume of Agent internal monitoring as excessive production of this data consumes significant disk space within customer environments.

We have measurably reduced the amount of logs agent generates by default. Ideally we will have reduced agent monitoring log volume by 80% in normal usage and have run out of obvious ideas or superfluous messages to remove.

pchila:
## What does this PR do?
This PR drops the `Non-zero metrics after 30s...` logs before they get sent to Elasticsearch.

## Why is it important?
The goal is to reduce the amount of elastic-agent monitoring events ingested in Elasticsearch in order to reduce index disk size.

## Checklist

## Author's Checklist

- [ ]

Run an elastic agent with and without this change on a deployment making a note of the start times and elastic-agent id of both.
Using the included ES script we then extract the documents ingested during those runs and store them in dedicated indices.

In my tests I ran 2 elastic agents managed by fleet with a default policy including system integration, first without this change and then including this change, then sliced the first 10 minutes of logs and metrics from startup using the included script.

This is a screenshot from the Index management page were we can see the disk sizes and document count for each index:
Screenshot 2024-05-02 at 18-48-16 Index Management - Elastic

The measured impact on `*logs-elastic_agent.filebeat*` and `*logs-elastic_agent.metricbeat*` is:
- reduction of ~27% of document count for `*logs-elastic_agent.filebeat*` with a disk size reduction of ~18%
- reduction of ~21% of document count for `*logs-elastic_agent.metricbeat*` with a disk size reduction of ~16% 

## Related issues

- Relates #4252 

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1714717717-Reduce-agent-logs-by-default.yaml
+++ b/changelog/fragments/1714717717-Reduce-agent-logs-by-default.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Reduce agent logs by default by dropping "Non-zero metrics..." logs before ES ingestion
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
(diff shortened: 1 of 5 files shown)


amitkanfer:
@pchila does this really close #4252 ? or only address [this](https://github.com/elastic/elastic-agent/issues/4252#issuecomment-2057428599) suggestion by @cmacknz ?

pchila:
> @pchila does this really close #4252 ? or only address [this](https://github.com/elastic/elastic-agent/issues/4252#issuecomment-2057428599) suggestion by @cmacknz ?

At this stage it addresses @cmacknz 's suggestion and removes some unnecessary logs, not sure if we want to stop here for this PR or we are going deeper, it depends on the measurement of how much data we are saving as it is and general 8.14 timeframe... Will change the "Closes" to "Relates"

pierrehilbert:
I agree here, we shouldn't try to close https://github.com/elastic/elastic-agent/issues/4252 but try to get as many improvement as possible for 8.14 and then continue for the next versions what we won't have time to do in that timeframe.

cmacknz (approved):
I enrolled an agent built from this branch and observed 30s metrics in the local agent logs but not in Fleet as expected.

pierrehilbert:
@cmacknz any objections with backporting this in 8.14?

cmacknz:
No objections to backporting.

Issue #4617, opened 2024-04-24
[Flaky Test]: TestOtelFileProcessing – there should be exported logs by now


rdner:
### Failing test case

TestOtelFileProcessing

### Error message

there should be exported logs by now

### Build

https://buildkite.com/elastic/elastic-agent/builds/8506#018f0d68-b5e4-46e6-93a8-91612f46c4e8

### OS

Linux

### Stacktrace and notes

```markdown
=== RUN   TestOtelFileProcessing
    fetcher.go:95: Using existing artifact elastic-agent-8.14.0-SNAPSHOT-linux-x86_64.tar.gz
    fixture.go:281: Extracting artifact elastic-agent-8.14.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestOtelFileProcessing3713650306/002
    fixture.go:299: Completed extraction of artifact elastic-agent-8.14.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestOtelFileProcessing3713650306/002
    fixture.go:957: All component specifications where removed
    fixture.go:998: Placed component specifications: fake, fake-shipper
    log.go:85: Starting in otel mode
    log.go:96: {"log.level":"info","@timestamp":"2024-04-24T00:32:50.019Z","log.origin":{"file.name":"cmd/run.go","file.line":180},"message":"Elastic Agent started","log":{"source":"elastic-agent"},"process.pid":18312,"agent.version":"8.14.0","agent.unprivileged":true,"ecs.version":"1.6.0"}
    log.go:85: 2024-04-24T00:32:50.028Z	info	service@v0.99.0/service.go:99	Setting up own telemetry...
    log.go:85: 2024-04-24T00:32:50.028Z	info	service@v0.99.0/telemetry.go:103	Serving metrics	{"address": ":8888", "level": "Normal"}
    log.go:85: 2024-04-24T00:32:50.034Z	info	service@v0.99.0/service.go:166	Starting /tmp/TestOtelFileProcessing3713650306/002/elastic-agent-8.14.0-SNAPSHOT-linux-x86_64/elastic-agent...	{"Version": "8.14.0", "NumCPU": 2}
    log.go:85: 2024-04-24T00:32:50.034Z	info	extensions/extensions.go:34	Starting extensions...
[...]

PR #4638, opened 2024-04-30
Use otel version 0.97.0

michalpristas:
Otel versions >0.97.0 have a bug where they expose metrics on :8888 even for validate subcommands causing our tests to fail.

related to this issue here: https://github.com/open-telemetry/opentelemetry-collector/issues/10031
cause by pull https://github.com/open-telemetry/opentelemetry-collector/pull/9257

Closes: #4617


rdner (approved):

ycombinator (dismissed):

michalpristas:
Pavel and Olga are working on addressing macos failures

go.mod
```diff
@@ -41,11 +41,11 @@ require (
 	github.com/mitchellh/hashstructure v1.1.0
 	github.com/mitchellh/mapstructure v1.5.1-0.20231216201459-8508981c8b6c
 	github.com/oklog/ulid v1.3.1
-	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.98.0
```
ycombinator, line 44: Could we add a couple of newlines around this block of OTel dependencies or add a separate `require` block for them to indicate that they are all related?  

This would then give us an opportunity to add a comment here explaining why we're pinning to `0.97.0` for now and what is the criteria for undoing that pinning.  That way, when dependabot tries to automatically bump the version, whoever reviews those PRs would immediately know whether to approve them or not.

michalpristas: i think we could separate

go.mod
```diff
@@ -104,6 +81,35 @@ require (
 	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448
 )
 
+require (
+	// open telemetry dependencies
+	// pinned to v0.97.0 due to issue with ports binding which causes our tests to fail
+	// ref: https://github.com/open-telemetry/opentelemetry-collector/issues/10031
```
ycombinator, line 87: Nit: use more action-oriented language so we know when its OK to remove the pinning.

```suggestion
	// Do NOT upgrade the version beyond 0.97.0 until https://github.com/open-telemetry/opentelemetry-collector/issues/10031 is fixed!
```

ycombinator:
Backporting to `8.14` as it's needed for https://github.com/elastic/elastic-agent/pull/4711.

Issue #4660, opened 2024-05-02
[OTel mode] `elastic-agent otel` does not work if `elastic-agent.yml` is absent


ycombinator:
Please post all questions and issues concerning the Elastic Agent on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For security vulnerabilities please only send reports to security@elastic.co.
See https://www.elastic.co/community/security for more information.

Please include configurations and logs if available.

For confirmed bugs, please report:
- Version: `8.14.0-SNAPSHOT` but also reproducible with build off `main`.
- Operating System:  Tested on MacOS but likely reproducible on any.
- Steps to Reproduce:
   1. Extract the Elastic Agent package into a folder.
   2. Deliberately remove `elastic-agent.yml` from the folder.
   3. From the folder, run `./elastic-agent otel`.

Observed: Elastic Agent starts up in OTel mode but immediately quits with an error like so:
```
./elastic-agent otel
Starting in otel mode
2024-05-02T10:27:49.591-0700	info	service@v0.96.0/telemetry.go:55	Setting up own telemetry...
2024-05-02T10:27:49.591-0700	info	service@v0.96.0/telemetry.go:97	Serving metrics	{"address": ":8888", "level": "Basic"}
2024-05-02T10:27:49.591-0700	info	exporter@v0.98.0/exporter.go:275	Development component. May change in the future.	{"kind": "exporter", "data_type": "logs", "name": "debug"}
2024-05-02T10:27:49.591-0700	info	service@v0.96.0/service.go:143	Starting ./elastic-agent...	{"Version": "8.14.0", "NumCPU": 12}
2024-05-02T10:27:49.591-0700	info	extensions/extensions.go:34	Starting extensions...
2024-05-02T10:27:49.591-0700	info	adapter/receiver.go:45	Starting stanza receiver	{"kind": "receiver", "name": "filelog", "data_type": "logs"}
2024-05-02T10:27:49.591-0700	info	service@v0.96.0/service.go:169	Everything is ready. Begin running and processing data.
2024-05-02T10:27:49.591-0700	warn	localhostgate/featuregate.go:63	The default endpoints for all servers in components will change to use localhost instead of 0.0.0.0 in a future version. Use the feature gate to preview the new default.	{"feature gate ID": "component.UseLocalHostAsDefaultHost"}
[...]

PR #4667, opened 2024-05-03
Fix dependency on `elastic-agent.yml`

michalpristas:
Fix dependency on `elastic-agent.yml`

Fixes: #4660


ycombinator:
Would it be possible to add a test, probably one that tries to start up Agent in OTel mode without `elastic-agent.yml` present and asserts that the Agent did indeed start up successfully?

pchila (approved):

PR #4579, opened 2024-04-16
Remove state beat state metricset

pchila:
## What does this PR do?
This change disable collection of beats `state` metricset for elastic-agent self-monitoring components 

## Why is it important?
The `state` metricset is not used and no [exported fields](https://www.elastic.co/guide/en/beats/metricbeat/current/exported-fields-beat.html) are mapped in [elastic-agent integration](https://github.com/elastic/integrations/tree/main/packages/elastic_agent)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## Related issues

- Closes #4153

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4671, opened 2024-05-03
Patched changelog for PR 4579

### LINKED ISSUE #4153: Stop collecting the beat state metricset as part of agent monitoring
Our agent monitoring implementation currently uses the [beat Metricbeat module](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-module-beat.html) to monitor Beat subprocesses. We collect both the [stats](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-beat-stats.html) and [state](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-beat-state.html) metricsets.

https://github.com/elastic/elastic-agent/blob/b39b9af521fcbf1fcae6bab14762b0a120febdb7/internal/pkg/agent/application/monitoring/v1_monitor.go#L617-L625

It seems to me that nothing actually uses the data from the state metricset. We don't map the fields in the Elastic Agent integration. I believe we can remove this metricset and stop pointlessly storing this data for every Beat process we start.

We currently store both the state and stats metricset in the same datastream, and as such include the metricset name as a TSDB dimension which could probably be removed after this change. 

https://github.com/elastic/integrations/blob/a2c55c4cbf752e0490f9fe2d3e68698517c7b74d/packages/elastic_agent/data_stream/elastic_agent_metrics/fields/ecs.yml#L21-L23
```yaml
- name: metricset.name
  type: keyword
  dimension: true
```

Acceptance Criteria:
- The beat state documents are no longer queryable from the metrics-elastic_agent.* datastream and any existing mappings the elastic agent package are removed https://github.com/elastic/integrations/tree/main/packages/elastic_agent
- The data storage savings after removing this metricset are calculated and included in the release notes

pchila:
## What does this PR do?
This PR patches changelog fragment for already merged PR #4579 specifying the savings coming from removal of beat state metricset

## Checklist

## Related issues

- Closes #4153

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1713257367-Remove-beat-state-metricset-from-elastic-agent-monitoring.yaml
+++ b/changelog/fragments/1713257367-Remove-beat-state-metricset-from-elastic-agent-monitoring.yaml
@@ -11,12 +11,12 @@
 kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
-summary: Removed beat state metricset from monitoring collect metrics are unused
+summary: Removed beat state metricset from monitoring metrics
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
-#description:
+description: Removing beat state metricsets reduces the document count of indices metrics-elastic_agent.*-* by 50%, saving around 13% disk space
 
 # Affected component; a word indicating the component this changeset affects.
 component: elastic-agent


cmacknz (approved):

changelog/fragments/1713257367-Remove-beat-state-metricset-from-elastic-agent-monitoring.yaml
```diff
@@ -11,12 +11,12 @@
...
 
 # Change summary; a 80ish characters long description of the change.
-summary: Removed beat state metricset from monitoring collect metrics are unused
+summary: Removed beat state metricset from monitoring metrics
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
-#description:
+description: Removing beat state metricsets reduces the document count of indices metrics-elastic_agent.*-* by 50%, saving around 13% disk space
```
cmacknz, line 19: https://github.com/elastic/elastic-agent/issues/4153#issuecomment-2093194405

> Isn't the number of metrics produced dependent on the number of components running under agent? i.e. something like x document per beat per interval? so the % savings depends on the number of deployed integrations/managed beats?

We need to document which integrations we used to obtain this number and that that the savings depend on the number of inputs. More inputs = more monitoring data being written, so the total savings should be greater in terms of absolute numbers (MB) in that case.

pchila: The numbers I got are obtained using only the system integration... Those should be replicated n times for any other component that belongs in the [supported component list](https://github.com/elastic/elastic-agent/blob/e7c4ac8bd7c5e1ae61fce9c97b373c64cb75559e/internal/pkg/agent/application/monitoring/v1_monitor.go#L67)

cmacknz: Makes sense we drop the same amount of documents from every Beat

jlind23:
@pchila Can we merge this? :-)

PR #4731, opened 2024-05-09
Use same version of golangci-lint in makefile as github actions

therealdwright:
## What does this PR do?

The .github/workflows/golangci-lint.yml spec uses v1.55.2 for the
golangci-lint version. This commit updates the make target to use
the same version for consistency as the current version causes a
panic when running on a darwin arm64 system.

## Why is it important?

Provide a consistent local experience to CI.

## Checklist
- [X] I have made corresponding changes to the documentation
- [X] I have made corresponding change to the default configuration files
- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ check:
 ## check-go: download and run the go linter.
 .PHONY: check-go
 check-go: ## - Run golangci-lint
-	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.2
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.55.2
 	@./bin/golangci-lint run -v
 
 ## check-no-changes : Check there is no local changes.


Makefile
```diff
@@ -48,7 +48,7 @@ check:
 ## check-go: download and run the go linter.
 .PHONY: check-go
 check-go: ## - Run golangci-lint
-	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.2
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.55.2
```
cmacknz, line 51: I don't know if anyone is actually using this target. Let me double check that, it might make sense to delete the target instead. It can't get out of sync if it doesn't exist anymore.

therealdwright: I used it myself so personally think it's useful to keep in but LMK if you want me to repurpose this PR to delete it.

cmacknz (approved):

Makefile
cmacknz: Sure, it doesn't cost us anything to have it here if it makes it easier for external contributors to figure out what to do.

cmacknz:
buildkite test it

PR #4746, opened 2024-05-14
Fix wrong permissions when installation with enrollment on Windows in privileged mode

blakerouse:
## What does this PR do?

Fixes an installation issue on Windows in privileged mode where the wrong permissions are given to the Elastic Agent installation directory.

## Why is it important?

Otherwise it would allow any non-Administrator to read/modify/delete a file in the Elastic Agent installation directory.

## Checklist

## Disruptive User Impact

Fixes a disruptive user issue of invalid permissions.

1. Install without `--unprivileged` and with enrollment.
2. Verify the permissions of the installation directory only have access by Administrators.

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/3339

### DIFF
--- a/internal/pkg/agent/perms/windows.go
+++ b/internal/pkg/agent/perms/windows.go
@@ -42,7 +42,7 @@ func FixPermissions(topPath string, opts ...OptFunc) error {
 
 	// user gets grant based on the mask
 	var userSID *windows.SID
-	if o.ownership.UID != "" {
+	if o.mask&0700 != 0 && o.ownership.UID != "" {
 		userSID, err = windows.StringToSid(o.ownership.UID)
 		if err != nil {
 			return fmt.Errorf("failed to get user %s: %w", o.ownership.UID, err)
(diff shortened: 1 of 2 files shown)


internal/pkg/agent/perms/windows.go
```diff
@@ -42,7 +42,7 @@ func FixPermissions(topPath string, opts ...OptFunc) error {
 
 	// user gets grant based on the mask
 	var userSID *windows.SID
-	if o.ownership.UID != "" {
+	if o.mask&0700 != 0 && o.ownership.UID != "" {
```
cmacknz, line 45: What does the 0700 mask mean for Windows? Same thing as for Unix?

cmacknz (commented):
Can we add a test to prevent a regression like this again?

Something like using an unprivileged shell/terminal to try to access/modify/delete a file in the installation directory?

cmacknz:
This needs to be merged before the 16th to make the final build candidate, so we have some time to add the necessary validation. If it turns out we need longer than that to add validation, we can merge this first.

I've also tracked this issue as blocker on the release issue so it can't be forgotten.

blakerouse:
Sounds good. Working on test.

cmacknz (approved):
Approving to avoid blocking the final 8.14 BC. Tests will follow separately.

Issue #4749, opened 2024-05-14
Better log message for resetting log level


jen-huang:
Follow up to https://github.com/elastic/kibana/issues/180778

When resetting agent log level back to default (using `null` value), the log message is:
`Settings action done, setting agent log level to ` 

This should be improved to something like
`Settings action done, setting agent log level to default (info)`

PR #4751, opened 2024-05-14
fix log message when settings log level from SETTINGS action

nchaulet:
## Description 

Resolve https://github.com/elastic/elastic-agent/issues/4749

Fix log message when setting log level from settings action to display the log level we are setting from action, policy or default one.


blakerouse (approved):
Looks good.

PR #4757, opened 2024-05-15
Remove pinning to 8.14.0.

cmacknz:
What the title says. 8.15.0-SNAPSHOT exists, remove version pinning to 8.14.0-SNAPSHOT.


jlind23 (approved):

.package-version
```diff
@@ -1 +1 @@
-8.14.0
\ No newline at end of file
+8.15.0
```
cmacknz, line 1: The automation should do this tomorrow, https://github.com/elastic/elastic-agent/pull/4768

At which point this just becomes removing the skipped test.

cmacknz: Automation update is in https://github.com/elastic/elastic-agent/pull/4773

PR #4759, opened 2024-05-15
Add new cloud metadata processor error to the ignored list in integration tests

cmacknz:
Also remove an upgrade watcher error that appears like it could be a real bug if it is still happening. If this log appears in this PR run I'll restore it.

The new error was first observed in https://buildkite.com/elastic/elastic-agent/builds/8939#018f7d48-21cf-4525-9538-641ceef01269

### DIFF
--- a/testing/integration/logs_ingestion_test.go
+++ b/testing/integration/logs_ingestion_test.go
@@ -281,7 +281,7 @@ func testMonitoringLogsAreShipped(
 			"add_cloud_metadata: received error failed with http status code 404",                                                                                                                             // okay for the cloud metadata to not work
 			"add_cloud_metadata: received error failed fetching EC2 Identity Document: not found, Signing",                                                                                                    // okay for the cloud metadata to not work
 			"add_cloud_metadata: received error failed fetching EC2 Identity Document: operation error ec2imds: GetInstanceIdentityDocument, http response error StatusCode: 404, request to EC2 IMDS failed", // okay for the cloud metadata to not work
-			"failed to invoke rollback watcher: failed to start Upgrade Watcher: fork/exec /var/lib/elastic-agent/elastic-agent: no such file or directory",                                                   // on debian this happens probably need to fix.
+			"SDK 2024/05/15 17:49:30 WARN falling back to IMDSv1: operation error ec2imds: getToken, http response error StatusCode: 405, request to EC2 IMDS failed",                                         // okay for the cloud metadata to not work
 		})
 	})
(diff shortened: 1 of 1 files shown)


blakerouse:
All of these errors are terrible. The processor should do a better job and providing a nice clear error, and in most cases this should not be an error. It should only be a warning, if you not running on a cloud you don't want a bunch of error logs that are not true for the situation.

testing/integration/logs_ingestion_test.go
```diff
@@ -281,7 +281,7 @@ func testMonitoringLogsAreShipped(
...
 			"add_cloud_metadata: received error failed fetching EC2 Identity Document: not found, Signing",                                                                                                    // okay for the cloud metadata to not work
 			"add_cloud_metadata: received error failed fetching EC2 Identity Document: operation error ec2imds: GetInstanceIdentityDocument, http response error StatusCode: 404, request to EC2 IMDS failed", // okay for the cloud metadata to not work
-			"failed to invoke rollback watcher: failed to start Upgrade Watcher: fork/exec /var/lib/elastic-agent/elastic-agent: no such file or directory",                                                   // on debian this happens probably need to fix.
+			"SDK 2024/05/15 17:49:30 WARN falling back to IMDSv1: operation error ec2imds: getToken, http response error StatusCode: 405, request to EC2 IMDS failed",                                         // okay for the cloud metadata to not work
```
blakerouse, line 284: How is that going to work, it has a timestamp in it?

blakerouse (changes requested):

testing/integration/logs_ingestion_test.go
cmacknz: This is a good point, I just blindly copy pasted this without looking too closely. It's going to need to be a pattern, or I need to go remove this log message from Beats.

cmacknz: What's worse is this isn't even a JSON log, it is just getting dumped into stderr and we are picking it up. I need to figure out where this one comes from.

cmacknz: I am going to look at just fixing the logging in add_cloud_metadata in general. Users trying to build a similar alert run into this exact same problem.

rdner:
Closing in favor of https://github.com/elastic/elastic-agent/pull/4767

cmacknz:
Thanks, I am going to try to fix this in the processor itself so we can stop fighting it.

PR #4761, opened 2024-05-15
Add permission checks to all install and upgrade integration tests

blakerouse:
## What does this PR do?

Improves the integration install and upgrade tests for both Linux and Windows to ensure that the permissions set during installation and installation + enrollment are correct.

## Why is it important?

Ensures that any future changes that can cause bad permissions are caught during the PR stage.

## Checklist


blakerouse:
One of the issues here is that the Elastic Agent user was renamed from `elastic-agent` to `elastic-agent-user` in 8.14. This causes the tests that start at 8.13 and upgrade to fail, because the `elastic-agent-user` doesn't exist it is actually `elastic-agent`.

testing/installtest/checks_unix.go
```diff
@@ -0,0 +1,149 @@
...
+		// Ensure entire installation tree has the correct permissions.
+		err = validateFileTree(topPath, uint32(uid), uint32(gid))
+		if err != nil {
+			// context already added
+			return err
+		}
+
+		// Check that the socket is created with the correct permissions.
+		socketPath := filepath.Join(topPath, paths.ControlSocketName)
+		err = waitForNoError(context.Background(), func(_ context.Context) error {
```
cmacknz, line 44: You should accept a context as an argument through `CheckSuccess` so that this cancels or expires when the context in the test calling it does.

blakerouse: Fixed.

cmacknz (approved):

PR #4767, opened 2024-05-16
Make error checks in TestLogIngestionFleetManaged more resilient

rdner:
This check should not use exact matching but rather "contains" for accounting for know errors since details might change.


blakerouse (approved):
Looks good.

An improvement over the current checks, until the actual errors are improved.

Issue #4723, opened 2024-05-09
Make `resourcedetectionprocessor` available in `otel` mode


ycombinator:
**Describe the enhancement:**

Include the `resourcedetectionprocessor` in the OTel Collector [components](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/components.go) we bundle with Elastic Agent.

**Describe a specific use case for the enhancement or feature:**

Users of Elastic Agent in OTel mode (`elastic-agent otel ...`) would like to be able to use the `resourcedetectionprocessor` in their OTel Collector pipeline configurations.  

**What is the definition of done?**

* [ ] A test that runs Elastic Agent in OTel mode (`elastic-agent otel ...`) includes the `resourcedetectionprocessor` in it's test OTel Collector configuration and the test passes.

PR #4787, opened 2024-05-21
Add test for OTel Collector start

andrzej-stencel:
## What does this PR do?

Adds a test that verifies that the OTel Collector embedded in the Elastic Agent can be started successfully with all the components included in it.

## Why is it important?

We want to make sure all the components we include in the OTel distro can be actually run - e.g. there are no runtime crashes, port conflicts, etc.

## Checklist

```shell
cd internal/pkg/otel
go test -v .
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/4723


internal/pkg/otel/run_test.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+		{
+			configFile:           "nonexistent-component.yml",
+			expectedErrorMessage: `error decoding 'extensions': unknown type: "zpages"`,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.configFile, func(t *testing.T) {
+			configFiles := []string{"file:" + filepath.Join(".", "testdata", tc.configFile)}
```
blakerouse, line 36: Surprised that doesn't need to be `file://`

blakerouse (approved):
Looks good.

internal/pkg/otel/run_test.go
andrzej-stencel: Yeah, "file" is the default when there's no scheme provided.

- https://opentelemetry.io/docs/collector/configuration/#location
- https://github.com/open-telemetry/opentelemetry-collector/blob/v0.100.0/confmap/resolver.go#L82-L83
- https://github.com/open-telemetry/opentelemetry-collector/blob/v0.100.0/confmap/resolver.go#L131-L132

andrzej-stencel:
I need to fix the test for Windows, converting to draft PR until I fix this.

internal/pkg/otel/run_test.go
andrzej-stencel: Oh you mean the `//` in `file://`. Yeah that's how it works.

andrzej-stencel:
The test is fixed now.

internal/pkg/otel/run_test.go
blakerouse: What do you mean that is how it works? It works with `file://` or `file:`? I was just stating that I am surprised if doesn't require the `//` before the actual path.

internal/pkg/otel/testdata/all-components.windows.yml
```diff
@@ -0,0 +1,3 @@
+exporters:
+  file:
+    path: ${env:TMP}\file-exporter-output.json
```
blakerouse, line 3: seems that windows supports a lot less? is this from upstream or just how we have enabled it?

internal/pkg/otel/run_test.go
andrzej-stencel: Yeah, by "this is how it works" I meant that this is how it's implemented in the file provider: https://github.com/open-telemetry/opentelemetry-collector/blob/v0.100.0/confmap/provider/fileprovider/provider.go#L42-L73.

Whether it's compliant with the https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax or https://en.wikipedia.org/wiki/File_URI_scheme or https://datatracker.ietf.org/doc/html/rfc3986 is probably another question. The file provider definitely allows relative paths like `file:./local-file.txt`, which I'm not sure the RFC does.

internal/pkg/otel/run_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+// getConfigFiles returns a collection of config file paths for the collector to use.
+// In the simplest scenario, the collection will contains only one path.
+// In case there is an operating system-specific override file found, it will be added to the collection.
+// E.g. if the input file name is `all-components.yml` and a file named `all-components.windows.yml` exists,
+// the config path collection will have two elements on Windows, and only one element on other OSes.
+// Use `darwin` for MacOS, `linux` for Linux and `windows` for Windows.
+func getConfigFiles(configFileName string) []string {
+	// Add base file to the collection.
+	baseFilePath := filepath.Join(".", "testdata", configFileName)
```
pchila, line 69: Is the `.` initial token of the path required by the `file:` location? Aside from the `//` that was already discussed in Blake's comments I wonder if omitting the current directory would still work. something like
```suggestion
	baseFilePath := filepath.Join("testdata", configFileName)
```

pchila (approved):
Looks good.
Just a little curiosity about the exact semantics of `file:` locator but that is obviously non-blocking

internal/pkg/otel/run_test.go
andrzej-stencel: Yes, this should work equally well.

Issue #4083, opened 2024-01-12
Agent should collect and report CPU and memory usage of service runtime components


ycombinator:
- Relates https://github.com/elastic/elastic-agent/issues/1944

Currently, Agent collects and reports CPU and memory usage of command runtime components (and even that is [incomplete](https://github.com/elastic/elastic-agent/issues/4082)).  It is not collecting and reporting CPU and memory usage of service runtime components, e.g. Endpoint, leading to undercounting.

cmacknz:
This was blocked by https://github.com/elastic/beats/issues/17314, which is now resolved.

fearful-symmetry:
So, I'm tinkering around with how to add monitoring config for something like endpoint to the `v1_monitor.go` setup code. It's a little harder than I expected, as most of the monitoring config seems to be generated before components start, which means we can't just pass a PID to the monitoring code that it'll use to run some normal process metrics code. The alternative would be to search for the process based on a regex name match, witch seems...kinda hacky. The other option is to invent some new monitoring scheme that that we can send a PID to.

There's also the question of how we report these metrics to begin with; the existing monitoring code uses the `beat/stats` and `json` metricsets, which aren't available to us if we're monitoring something like endpoint. If we use the `system/process` metricset, which seems like the most obvious solution, we either have to format the metrics so they'll be in a structure that Kibana expects for its monitoring queries, or we change the kibana queries used for generating usage metrics. Not sure which implementation is best.

fearful-symmetry:
So, still poking around the codebase, but I'm not sure I see good way to pass the PID, since the monitoring config seems to happen at the same time as the rest of the config, which looks like it would happen before a given be is (re)started. Assuming I'm right, what we could do is restart the monitoring beat after everything else has started, which seems a bit needless and could also result in missing data.

cmacknz:
@nfritts @brian-mckinney we want to include endpoint's CPU and memory usage metrics in Fleet. To do this we need to know the endpoint process's PID. 

If we added the current PID as a field that can be reported in the `CheckinObserved` message is that something that would be easy for endpoint to implement?

This seemed like the simplest approach to us but wanted to confirm. Otherwise we'd be iterating through every PID on the system looking for the one that maps to endpoint-security regularly. If you have other ideas feel free to propose them.

nfritts:
[...]

PR #4789, opened 2024-05-21
Allow agent to monitor endpoint

fearful-symmetry:
## What does this PR do?

closes https://github.com/elastic/elastic-agent/issues/4083

This is a large and fairly wide-ranging set of changes that allows agent to monitor and report usage metrics for endpoint.
Specifically: 

- adds code to the coordinator and runtime controllers to filter PID changes up to the config system
- Adds logic to the metrics input generator to add a `system/process` metricset to monitor endpoint
- Add an integration test
- Change the behavior of the `local` env var in mage, so `local=agentbeat` will only build agentbeat, then pull other binaries from upstream. Needed this for testing, but kept it because it seemed vaguely useful. However, we don't need to merge it.

## Checklist


fearful-symmetry:
CI may be various kinds of broken until the snapshot beats build are updated, taking out of draft since technically everything needed is merged.

fearful-symmetry:
Blocked by https://github.com/elastic/beats/pull/39730

pierrehilbert:
Looks like this is no more blocked @pchila @faec @blakerouse

fearful-symmetry:
@pierrehilbert CI will probably fail until snapshots are updated, but other than that, should be ready.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1032,6 +1035,15 @@ func (c *Coordinator) runLoopIteration(ctx context.Context) {
 
 	case c.heartbeatChan <- struct{}{}:
 
+	case <-c.servicePidUpdate:
+		c.logger.Infof("got pid service update, refreshing config")
+		err := c.refreshComponentModel(ctx)
```
cmacknz, line 1040: If endpoint crashloops, which I've seen happen, we are going to be doing this continuously.

I suspect this is expensive and we likely want to rate limit it.

cmacknz: This will be particularly bad if multiple sub-processes were to enter crash loops, or just restart frequently.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1032,6 +1035,15 @@ func (c *Coordinator) runLoopIteration(ctx context.Context) {
 
 	case c.heartbeatChan <- struct{}{}:
 
+	case <-c.servicePidUpdate:
```
cmacknz, line 1038: Service PID update is a bit of a misleading title, it implies it has to do with the service runtime specifically, but you are just generically pulling this out of the checkin observed so it isn't limited to that.

Remove service from the name to clarify this.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -908,6 +910,87 @@ func (b *BeatsMonitor) injectMetricsInput(cfg map[string]interface{}, componentI
 		},
 	}
 
+	// add system/process metrics for services that can't be monitored via json/beats metrics
+	// If there's a checkin PID and it contains the "endpoint" string, assume we want to monitor it
```
cmacknz, line 914: Instead of just "endpoint" can you look for inputs using the service runtime https://github.com/elastic/elastic-agent/blob/e401afe9d7b3ac9f5b2aef7999642f8194d9d243/specs/endpoint-security.spec.yml#L23

That will remove the hard coding of the input name here and generalize better.

magefile.go
```diff
@@ -1077,101 +1077,126 @@ func collectPackageDependencies(platforms []string, packageVersion string, requi
...
+				targetFile := filepath.Join(targetPath, filepath.Base(f))
+				packagesCopied += 1
+				if err := sh.Copy(targetFile, f); err != nil {
+					panic(err)
 				}
 			}
 		}
-	} else {
-		archivePath = movePackagesToArchive(dropPath, requiredPackages)
+		// a very basic footcannon protector; if packages are missing and we need to rebuild them, check to see if those files were copied
```
cmacknz, line 1194: Can all of this be a separate PR with a description for what it fixes? Doesn't seem related to the endpoint change.

testing/integration/helpers.go
```diff
@@ -0,0 +1,133 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build integration
+
+package integration
```
cmacknz, line 7: The filename should include endpoint or just be endpoint.go or something. I don't think you want this file to be a generic dumping ground for utility functions.

testing/integration/monitoring_endpoint_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+type EndpointMetricsMonRunner struct {
+	suite.Suite
+	info    *define.Info
+	fixture *atesting.Fixture
+}
+
+func TestEndpointAgentServiceMonitoring(t *testing.T) {
+	info := define.Require(t, define.Requirements{
+		Group: "fleet",
```
cmacknz, line 34: Use one of the predefined constants, like `Fleet`

testing/integration/monitoring_endpoint_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	require.Eventually(runner.T(),
+		func() bool { return agentAndEndpointAreHealthy(runner.T(), ctx, agentClient) },
+		time.Minute*3,
+		time.Second,
+		"Endpoint component or units are not healthy.",
+	)
+}
+
+func (runner *EndpointMetricsMonRunner) TestEndpointMetrics() {
```
cmacknz, line 103: Is there a reason this can't just be another test in endpoint_security_test.go as well? 

Extending on that, could this just be sub-test added to the existing endpoint installation test to avoid having to install agent again just for this?

These are unit tests, they don't have to be completely isolated into their own test cases.

testing/integration/monitoring_endpoint_test.go
```diff
@@ -0,0 +1,154 @@
...
+	require.NoError(runner.T(), err)
+
+	endpointID := "endpoint-default"
+	require.Eventually(runner.T(), func() bool {
+
+		query := genESQueryByBinary(agentStatus.Info.ID, endpointID)
+		res, err := estools.PerformQueryForRawQuery(ctx, query, "metrics-elastic_agent*", runner.info.ESClient)
+		require.NoError(runner.T(), err)
+		runner.T().Logf("Fetched metrics for %s, got %d hits", endpointID, res.Hits.Total.Value)
+		return res.Hits.Total.Value >= 1
```
cmacknz, line 117: You should probably verify the metrics aren't 0. This test would pass if you monitored nothing, I think this same flaw might exist in all of these tests if you copied what was done in the others.

dev-tools/mage/settings.go
```diff
@@ -88,7 +88,7 @@ var (
 
 	Snapshot      bool
 	DevBuild      bool
-	ExternalBuild bool
+	ExternalBuild string
```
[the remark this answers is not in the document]
fearful-symmetry: Yeah, the mage changes were done purely to make testing easier.  I can remove them.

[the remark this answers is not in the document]
cmacknz: This one I think actually changes the way the flag is parsed, bool I think will accept 0, 1, true, and false as input but string only accepts the ones you explicitly check for.

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: yeah, the debug line isn't supposed to be there, oops. Do you think the overall routine should have a rate limit of some kind? We do a PID comparison in `applyComponentState`, so we won't be hammering the update if we get the same value a bunch of times. 

If we want to handle a sort of crashloop scenario, (endpoint restarts frequently, each time coming up with a new PID) that could be interesting, since we'd still want to track the PID, so we can actually update the config once the crash loop stops or slows down?

internal/pkg/agent/application/monitoring/v1_monitor.go
fearful-symmetry: So, just to clarify, it looks like there's a component to the input spec: 
https://github.com/elastic/elastic-agent/blob/6573c38318ced4906c2bab2e43c1f72a00d3f210/pkg/component/input_spec.go#L25

That's only populated by endpoint and other components that we'd need to separately catch the PID for?

cmacknz: Yes that means it is going to be installed as an independent service and not a sub-process.

internal/pkg/agent/application/coordinator/coordinator.go
fearful-symmetry: Hmm, I see, that makes sense. So it sounds like we need some kind of separate thread that operates on a delay or otherwise bundles the PID updates.

fearful-symmetry:
One again, some field definitions in Fleet are missing. Will need to update the integration...

fearful-symmetry:
Still tinkering with the tests...

fearful-symmetry:
So, I ended up adding a decent bit of logic to the integration test, hopefully to better shake out any problems that might exist with the metric update logic.

testing/integration/monitoring_endpoint_test.go
fearful-symmetry: So, I ended up making the test a bit more complex instead, to the point where it feels like it makes more sense to have it as its own test? You might disagree, though.

internal/pkg/agent/application/coordinator/coordinator_state.go
```diff
@@ -135,16 +135,25 @@ func (c *Coordinator) refreshState() {
...
+				pidRequiresUpdate = true
+			}
 			c.state.Components[i] = state
 			found = true
 			break
 		}
 	}
 	if !found {
 		c.state.Components = append(c.state.Components, state)
+		if state.State.CheckinPid != 0 {
```
blakerouse, line 154: Does the component only send the PID once or only when it changes?

The reason I ask is it seems like to me at this point, every checkin would result in the 30 second ticker recomputing the component model.

internal/pkg/agent/application/coordinator/diagnostics_test.go
```diff
@@ -488,6 +488,7 @@ log_level: "warning"
 components:
   - id: "comp-1"
     state:
+      checkinpid: 0
```
blakerouse, line 491: Why is this prefixed with `checkin`? All of this information in the state comes from checkin, so I am confused on the prefix naming.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -908,6 +910,92 @@ func (b *BeatsMonitor) injectMetricsInput(cfg map[string]interface{}, componentI
 		},
 	}
 
+	for _, comp := range componentList {
+		logp.L().Infof("input component %s: %#v", comp.ID, comp.InputSpec.Spec.Service)
```
blakerouse, line 914: Using the global logger is not really recommened, I don't think this should be used in new code.

fearful-symmetry: ooops, debugging statement I kept in there...

internal/pkg/agent/application/coordinator/coordinator_state.go
fearful-symmetry: It should only mark `pidRequiresUpdate` if the component has an updated PID, or it's a new component with a non-zero pid. I don't see any part in the logic where `pidRequiresUpdate` would always be true?

internal/pkg/agent/application/coordinator/diagnostics_test.go
fearful-symmetry: Is everything in the `ComponentState` object the result of a checkin? I didn't think it was, and that's where that struct name is coming from.

blakerouse (approved):
Overall this looks good.

I would really like to see `checkinpid` to just be `pid`, but not a blocker.

faec (approved):
Some suggestions on atomic handling, otherwise looks good

Issue #4794, opened 2024-05-22
Add retries to all network communication when running integration tests


rdner:
Our integration tests sometimes hit errors like so:

### Error calling deployment retrieval API (**quite often**)

We should add a short timeout and retries.

> Error: error running test: failed to check for cloud 8.13.0-SNAPSHOT [stack_id: 8130-SNAPSHOT, deployment_id: REDACTED] to be ready: error calling deployment retrieval API: Get "https://cloud.elastic.co/api/v1/deployments/REDACTED": context deadline exceeded

Examples:

* https://buildkite.com/elastic/elastic-agent/builds/7614#018e1960-28f9-46f5-a197-a01dc2a76bcf
* https://buildkite.com/elastic/elastic-agent/builds/7654#018e1f49-c387-4973-b8a0-dd10dba598f2
* https://buildkite.com/elastic/elastic-agent/builds/7912#018e61fc-5be2-4ab4-98b7-d9bd98e32c8a
* https://buildkite.com/elastic/elastic-agent/builds/7264#018dc323-a3fd-423d-bed4-75fcf4027ffc
* https://buildkite.com/elastic/elastic-agent/builds/8252#018ec342-292a-4a31-9ae4-c360c3c54166

### Network issues (SSH failure, TCP i/o timeout)

Should add retries.

> Error: error running test: failed to connect to instance ogc-windows-amd64-2022-default-ff13: dial tcp REDACTED: i/o timeout

* https://buildkite.com/elastic/elastic-agent/builds/7876#018e5c3a-00b1-455f-b8c0-dab34135d3ea
* https://buildkite.com/elastic/elastic-agent/builds/7916#018e627d-a512-44bb-938e-2e4d62c4f792
* https://buildkite.com/elastic/elastic-agent/builds/7261#018dc1f1-d326-455c-bd41-fcad4fa2e0af
* https://buildkite.com/elastic/elastic-agent/builds/7296#018dc731-ae08-4a86-86f8-0f25e6edde84 

> (linux-arm64-ubuntu-2204-fleet-airgapped) Failed to execute tests on instance: error running sudo tests: failed to fetched test output at $HOME/agent/build/TEST-go-remote-linux-arm64-ubuntu-2204-fleet-airgapped-sudo.integration.out

* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/193#018fa634-b198-4497-908d-2e6dea654e28
[...]

PR #4798, opened 2024-05-22
Improve integration testing cloud client DeploymentIsReady.

blakerouse:
## What does this PR do?

Improves the logic in `DeploymentIsReady` to not return `false` and the error when it gets an error back from the API. Instead it will keep retrying until the timeout.

## Why is it important?

Improve reliability of integration tests, when the cloud API returns an HTTP error or disconnects without a response (just the nature of HTTP protocol).

## Checklist

## Related issues

- Relates #4794

### DIFF
--- a/pkg/testing/ess/deployment.go
+++ b/pkg/testing/ess/deployment.go
@@ -252,19 +252,18 @@ func (c *Client) DeploymentIsReady(ctx context.Context, deploymentID string, tic
 	defer ticker.Stop()
 
 	statusCh := make(chan DeploymentStatus, 1)
-	errCh := make(chan error)
 
 	for {
 		select {
 		case <-ctx.Done():
 			return false, ctx.Err()
 		case <-ticker.C:
-			statusCtx, statusCancel := context.WithTimeout(ctx, tick)
-			defer statusCancel()
 			go func() {
+				statusCtx, statusCancel := context.WithTimeout(ctx, tick)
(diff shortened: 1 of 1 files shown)


pkg/testing/ess/deployment.go
```diff
@@ -252,19 +252,18 @@ func (c *Client) DeploymentIsReady(ctx context.Context, deploymentID string, tic
...
 		case <-ticker.C:
-			statusCtx, statusCancel := context.WithTimeout(ctx, tick)
-			defer statusCancel()
 			go func() {
+				statusCtx, statusCancel := context.WithTimeout(ctx, tick)
+				defer statusCancel()
 				status, err := c.DeploymentStatus(statusCtx, deploymentID)
 				if err != nil {
-					errCh <- err
+					// ignore the error (most likely a transient HTTP error)
```
andrzej-stencel, line 266: The typical question in this case: would it make sense to at least log this error, to have a trace of what is happening?

rdner: I agree, every error should be logged, so it's easier to debug CI failures.

blakerouse: No logger is passed to this level. I changed the implementation to return the errors that occur, so if it failed to get ready and any errors occurred why checking they will be reported and wrote to stdout.

rdner (approved):

PR #4799, opened 2024-05-22
[CI] Integration:clean on exit

pazone:
## What does this PR do?
Forces `integration:clean` on exit when we run integration tests.  

## Why is it important?
In some cases when the tests are failed we leave test lingering stacks behind. 
Forced `integration:clean` may cover the majority cases but there's no 100% guarantee that it completely solves the problem.
Expect a follow-up PR with a cronjob that will delete all dangling stacks
## Checklist

## Related issues
- Relates https://github.com/elastic/ingest-dev/issues/2887

## Questions to ask yourself

- ...


cmacknz:
Why does the pre-exit hook work for this? Any chance the conditional `if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent-extended-testing" && "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]];` isn't evaluating to `true` as expected?

https://github.com/elastic/elastic-agent/blob/26f5288ab88fcfc658c26bbe63139c9d13acd53f/.buildkite/hooks/pre-exit#L5-L15

PR #4812, opened 2024-05-24
Fix how `updatecli` version is pinned

andrzej-stencel:
## What does this PR do?

Changes how the version of `updatecli` tool is pinned in CI to be more explicit and safer. We should allow updates to the `updatecli-action` (to allow e.g. bug fixes or security updates). If we want to pin the version of the `updatecli` tool, let's use the option that the action exposes for it: https://github.com/updatecli/updatecli-action/blob/v2.58.0/README.md#usage.

## Why is it important?

Keeps us up to date on the action versions, preventing security vulnerabilities or bugs.

## Checklist

### DIFF
--- a/.github/workflows/bump-golang.yml
+++ b/.github/workflows/bump-golang.yml
@@ -22,7 +22,9 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install Updatecli in the runner
-        uses: updatecli/updatecli-action@9a37c7e35598d7b37d8e7568b40ed9538112be01 # v0.76.1
+        uses: updatecli/updatecli-action@v2
+        with:
+          version: v0.76.1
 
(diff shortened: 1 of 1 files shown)


v1v (changes requested):
Dependabot has been enabled to automate the bumps. 

In addition, https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions is the policy we use for third-party dependencies

andrzej-stencel:
> Dependabot has been enabled to automate the bumps.
> 
> In addition, https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions is the policy we use for third-party dependencies

Oh I see, that's why the action version is referenced by a hash - thanks for explaining @v1v.

PR #4816, opened 2024-05-24
Add elastic-otel shortcut to `elastic-agent otel`

### LINKED ISSUE #4661: [OTel mode] Provide `otelcol` executable as a shortcut to `elastic-agent otel`
**Describe the enhancement:**

Currently, to run Elastic Agent in OTel mode, users must run `./elastic-agent otel ...` (or the Windows equivalent).  It would be convenient if users could simply run something like `./otelcol ...` instead.

For users who are familiar with OpenTelemetry but not with Elastic Agent, the less we expose Elastic Agent concepts to such users the better.

- [ ] Users should be able to run `./otelcol ...` in addition to `./elastic-agent otel ...`.
- [ ] These two commands should be functionally identical.

nchaulet:
## Description

Resolve #4661 

Provide an `elastic-otel` shortcut to `elastic-agent otel` cmd

Work in progress

### DIFF
--- a/dev-tools/packaging/files/darwin/elastic-otel.sh
+++ b/dev-tools/packaging/files/darwin/elastic-otel.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+BASEDIR=$(dirname "$0")
+
+exec $BASEDIR/elastic-agent otel $@
--- a/dev-tools/packaging/packages.yml
+++ b/dev-tools/packaging/packages.yml
@@ -201,6 +201,12 @@ shared:
       content: >
         {{ agent_package_version }}
       mode: 0644
+    'elastic-otel':
+      source: '{{ repo.RootDir }}/dev-tools/packaging/files/darwin/elastic-otel.sh'
+      mode: 0755
+    'data/{{.BeatName}}-{{ commit_short }}/elastic-otel':
+      source: '{{ repo.RootDir }}/dev-tools/packaging/files/darwin/elastic-otel.sh'
+      mode: 0755
     <<: *agent_darwin_app_bundle_files
     <<: *agent_binary_common_files
 


strawgate:
thoughts on doing otelcol instead of elastic-otel?

ycombinator:
> thoughts on doing otelcol instead of elastic-otel? or elastic-otel-collector?

https://github.com/elastic/elastic-agent/issues/4661#issuecomment-2135727812

dev-tools/packaging/files/darwin/elastic-otel.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+BASEDIR=$(dirname "$0")
+
+exec $BASEDIR/elastic-agent otel $@
```
strawgate, line 5: For Darwin and Linux I believe we actually want
```
exec "$BASEDIR/elastic-agent" otel "$@"
```

If you put elastic-agent into a folder with a space and then cd to root and try to run it, it will fail due to the space without the path being quoted.

In addition I believe $@ needs to be in quotes:
>($@) Expands to the positional parameters, starting from one. When the expansion occurs within double quotes, each parameter expands to a separate word.

```
./a.sh 2 "3     4" 5
2 3     4 5                  # output for "$@"
2 3 4 5                      # output for $@  -> spaces are lost!
```

strawgate (changes requested):

dev-tools/packaging/files/darwin/elastic-otel.sh
cmacknz: Yes, run [shellcheck](https://www.shellcheck.net/) on this and it should find this problem and any other Bash quirks.

kaanyalti (approved):

dev-tools/packaging/packages.yml
```diff
@@ -811,10 +832,7 @@ specs:
           'data/{{.BeatName}}-{{ commit_short }}/elastic-agent':
             template: '{{ elastic_beats_dir }}/dev-tools/packaging/templates/darwin/elastic-agent.tmpl'
             mode: 0755
-          '{{.BeatName}}{{.BinaryExt}}':
```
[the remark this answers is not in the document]
nchaulet: No, I will re-add this

michalpristas (approved):

Issue #2190, opened 2023-01-26
Processes left as zombies when they exit independently


cmacknz:
Before we fixed https://github.com/elastic/beats/issues/34178 we had some reports of an excessive number of zombie processes being left on a machine. In one case a single MacOS machine was left with hundreds of zombie processes. Example attached below.

[ps-aux.txt](https://github.com/elastic/elastic-agent/files/10504868/ps-aux.txt)

It seems like there is a correlation between the Beats restarting themselves and zombie processes being left behind. We should investigate this and see if there is anything we should be doing to ensure these processes are properly cleaned up. 

The path the Beats follow to exit in this case is [here](https://github.com/elastic/beats/blob/50dad3dab4d4cc6c8cfc6b0892f785471cd356d7/x-pack/libbeat/management/managerV2.go#L523-L528).

cmacknz:
@blakerouse does anything come to mind on what could cause this? Clearly the Beats restarting all the time was causing the unusual number of zombies, but the exited processes should not have been hanging around as zombies.

lucabelluccini:
Other users are hitting this issue.

cmacknz:
Do we have logs or diagnostics available when this is happening?

cmacknz:
I reproduced this without trying just installing a standalone agent with the default policy 
on a fresh Ubuntu 22.04 VM.
```
craig_mackenzie@cmackenzie-ubuntu-test:~/elastic-agent-8.10.2-linux-x86_64$ ps aux | grep elastic
root        1860  1.5  1.3 1506628 53864 ?       Ssl  19:24   0:01 /opt/Elastic/Agent/elastic-agent
root        1866  0.0  0.0      0     0 ?        Zs   19:24   0:00 [elastic-agent] <defunct>
[...]

PR #4834, opened 2024-05-31
Wait on the watcher at startup instead of just releasing resources associated with it

cmacknz:
- Relates https://github.com/elastic/elastic-agent/issues/2190

Today every time the agent starts the watcher it results in a zombie process on Linux. This is because we never wait on it. There are two places where this happens:

1. As part of the upgrade process: https://github.com/elastic/elastic-agent/blob/7b81fea37ed79b26a641f9d2c28cfcd913cdb645/internal/pkg/agent/application/upgrade/upgrade.go#L302-L308
2. At startup in case we are starting as a result of an upgrade: https://github.com/elastic/elastic-agent/blob/7b81fea37ed79b26a641f9d2c28cfcd913cdb645/internal/pkg/agent/cmd/run.go#L243-L247

I don't believe the first case is addressed by this change. This is because the agent process re-execs itself before the `Wait` can complete. I still have to do some experimenting to see exactly what happens in this case.

In the second case, 99% of the time the watcher exits quickly because the restart is not after an upgrade, and so simply waiting will prevent the zombie. If we are rolled back we are in a similar situation to case 1, as we may reexec before the wait completes.

This PR made me read the implementation of [process.Release()](https://pkg.go.dev/os#Process.Release) which on Unix makes no system calls at all, it just cleans up resources:

https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/os/exec_unix.go;l=85-91
```go
func (p *Process) release() error {
	// NOOP for unix.
	p.Pid = -1
	// no need for a finalizer anymore
	runtime.SetFinalizer(p, nil)
	return nil
}
```

On Windows it makes a system call to close the handle:

https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/os/exec_windows.go;l=65-77
```go
func (p *Process) release() error {
	handle := atomic.SwapUintptr(&p.handle, uintptr(syscall.InvalidHandle))
	if handle == uintptr(syscall.InvalidHandle) {
		return syscall.EINVAL
	}
	e := syscall.CloseHandle(syscall.Handle(handle))
	if e != nil {
		return NewSyscallError("CloseHandle", e)
	}
	// no need for a finalizer anymore
	runtime.SetFinalizer(p, nil)
	return nil
}
```

The reason `release()` not making a system call is interesting is because this means that the watcher is still child of the agent process, meaning it dies when the agent dies. This cannot be changed without a system call happening on Unix. This means there is an opportunity where the agent is dead and the watcher is also dead, unable to roll back. I think we may need to use something like https://github.com/sevlyar/go-daemon?tab=readme-ov-file#how-it-works to fix this. I suspect our existing test for this case isn't precise enough. The watcher from the agent starting the upgrade would need to be killed, and the agent we upgraded to would have to exit before it launched the watcher. This is not how it works today. I don't think this PR makes anything any worse but it doesn't fix this problem.

https://github.com/elastic/elastic-agent/blob/7b81fea37ed79b26a641f9d2c28cfcd913cdb645/testing/integration/upgrade_rollback_test.go#L201-L231

On Windows child processes do not share the lifetime of their parent process unless they are manually assigned to the same Job group, which we manually do when launching component sub-processes but do not do here. This means this should work as intended under all circumstances on Windows.

https://github.com/elastic/elastic-agent/blob/7b81fea37ed79b26a641f9d2c28cfcd913cdb645/pkg/core/process/process.go#L169-L177

I am going to do some manual testing and open an issue if I can confirm there is a window where the watcher isn't running and the agent process can have exited. Fixing that requires more work.

### DIFF
--- a/changelog/fragments/1717185708-Stop-creating-a-zombie-process-on-each-restart.yaml
+++ b/changelog/fragments/1717185708-Stop-creating-a-zombie-process-on-each-restart.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Stop creating a zombie process on each restart.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: "elastic-agent"
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/4834
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
--- a/changelog/fragments/1717185708-Stop-creating-a-zombie-process-on-each-restart.yaml
+++ b/changelog/fragments/1717185708-Stop-creating-a-zombie-process-on-each-restart.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
 summary: Stop creating a zombie process on each restart.
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -146,13 +146,6 @@ func InvokeWatcher(log *logger.Logger, agentExecutable string) (*exec.Cmd, error
 	}
 
 	cmd := invokeCmd(agentExecutable)
-	defer func() {
-		if cmd.Process != nil {
-			log.Infof("releasing watcher %v", cmd.Process.Pid)
-			_ = cmd.Process.Release()
-		}
-	}()
-
 	log.Infow("Starting upgrade watcher", "path", cmd.Path, "args", cmd.Args, "env", cmd.Env, "dir", cmd.Dir)
 	if err := cmd.Start(); err != nil {
 		return nil, fmt.Errorf("failed to start Upgrade Watcher: %w", err)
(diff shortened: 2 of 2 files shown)


internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -146,23 +146,24 @@ func InvokeWatcher(log *logger.Logger, agentExecutable string) (*exec.Cmd, error
...
-
 	log.Infow("Starting upgrade watcher", "path", cmd.Path, "args", cmd.Args, "env", cmd.Env, "dir", cmd.Dir)
 	if err := cmd.Start(); err != nil {
 		return nil, fmt.Errorf("failed to start Upgrade Watcher: %w", err)
 	}
 
 	upgradeWatcherPID := cmd.Process.Pid
 	agentPID := os.Getpid()
+
+	go func() {
```
cmacknz, line 157: I haven't added a test for this, because there is a test in https://github.com/elastic/elastic-agent/pull/4822 that catches this 100% of the time.

Specifically it is this block of fixture_install.go:

```go
	f.t.Cleanup(func() {
		// check for running agents after uninstall had a chance to run
		processes := getElasticAgentProcesses(f.t)

		// there can be a single agent left when using --develop mode
		if f.installOpts != nil && f.installOpts.Develop {
			assert.LessOrEqual(f.t, len(processes), 1, "More than one agent left running at the end of the test when --develop was used: %v", processes)
			// The agent left running has to be the non-development agent. The development agent should be uninstalled first as a convention.
			if len(processes) > 0 {
				assert.NotContains(f.t, processes[0].Cmdline, paths.DevelopmentInstallDirName,
					"The agent installed with --develop was left running at the end of the test or was not uninstalled first: %v", processes)
			}
			return
		}

		assert.Empty(f.t, processes, "there should be no running agent at the end of the test")
	})
```

The `assert.LessOrEqual(f.t, len(processes), 1, ...` will fail without this change, because if there is an agent process running we alway pick up 2. One is the main process, and the other is the zombie created from the watcher. 

That test can't be extracted from that PR since it depends on the `--develop` option, and I didn't think it made sense to add another separate test, but let me know and I'll come up with something. Probably an integration test that calls `elastic-agent restart` and checks to make sure there's only one process running afterwards.

pchila (approved):
This change would make the current situation not worse for the watcher launched during the upgrade and avoid a zombie at agent startup if the watch ends before the main elastic-agent process

I still believe that to completely solve the zombie process problem we will end up with some C code (either written by ourselves or implemented by a dependency) to properly detach the watcher or we will have some service manager (like systemd) to do it for us...

blakerouse:
> The reason release() not making a system call is interesting is because this means that the watcher is still child of the agent process, meaning it dies when the agent dies. 

Why would the watcher die if the parent dies? https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/application/upgrade/rollback_linux.go#L41 has `Pdeathsig` set to SIGINT. That means that when the starting process dies that SIGINT is sent to the child (I believe that is the meaning). The watcher would then exit? If that is the case we should not set that and it would remain. @ycombinator did a lot of adjustments with the watcher and systemd as well to ensure that the watcher stays running. So understanding the exact scenario you are speaking of and ensuring that we fix it, is important.

changelog/fragments/1717185708-Stop-creating-a-zombie-process-on-each-restart.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11:
```suggestion
kind: bug-fix
```

blakerouse (approved):
Agree this improves the situation but doesn't solve the re-exec problem.

PR #4846, opened 2024-06-04
[windows] if `elastic-agent run` fails, log error to Application EventLog

### LINKED ISSUE #4627: [windows] capture early start-up errors
**Describe the enhancement:**

Capture standard error of `elastic-agent` when run as a Windows service.  This isn't necessary under Linux since systemd will capture the standard error.

There are several steps that happen before internal logging is started.  If `elastic-agent` fails to start before internal logging is started the only place the error can be found is on the standard error of the `elastic-agent` process.  When `elastic-agent` is run as a service under Windows, the standard error is not captured.  This is a problem because `elastic-agent` can fail to start and there is no record of the failure.  Adding this enhancement will allow us to "see" the error.

When `elastic-agent` fails as a service under Windows the standard error of the process can be retrieved.

leehinman:
## What does this PR do?

On Windows if the run command fails, then the error message will be sent to the Application EventLog.  Other OSes do not require this since they capture stderr from processes that run as services, Windows does not.

This is only for the `run` command because the other commands should be run from the CLI, where the error is already published on stderr.

## Why is it important?

If any error happens before the elastic-agent internal logging is setup and the run command exits while running as a Windows service, the error is lost.  Without this error it is very difficult to determine why `elastic-agent run` is failing.

## Checklist

1. Install `elastic-agent` with `-a` flag pointing to CA certificates on the local filesystem
2. Stop `elastic-agent` service
3. Rename the CA certificates
4. Try to start `elastic-agent` service
5. Verify that Application EventLog has event showing that elastic-agent failed and what the error was.

## Related issues

- Closes #4627

## Questions to ask yourself

- ...

## Screenshots
Screenshot 2024-06-04 at 10 19 55

### DIFF
--- a/changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
+++ b/changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Capture early errors on Windows in Application eventlog.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
--- a/changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
+++ b/changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
@@ -19,14 +19,14 @@ summary: Capture early errors on Windows in Application eventlog.
 #description:
 
 # Affected component; a word indicating the component this changeset affects.
-component: agent
+component: elastic-agent
 
 # PR URL; optional; the PR number that added the changeset.
 # If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
 # NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
 # Please provide it if you are adding a fragment for a different PR.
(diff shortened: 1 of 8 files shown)


internal/pkg/agent/install/install_windows.go
```diff
@@ -80,6 +82,12 @@ func withServiceOptions(username string, groupName string) ([]serviceOpt, error)
 // gives user the ability to control the service, needed when installed with --unprivileged or
 // ReExec is not possible on Windows.
 func servicePostInstall(ownership utils.FileOwner) error {
+	// Modify registry to allow logging to eventlog as "Elastic Agent".
+	err := eventlog.InstallAsEventCreate(paths.ServiceName, eventlog.Info|eventlog.Warning|eventlog.Error)
+	if err != nil && !strings.Contains(err.Error(), "registry key already exists") {
```
cmacknz, line 87: Does this work for both `elastic-agent install` and `elastic-agent install --unprivileged`?

cmacknz: I assume CI will tell us this since it happens unconditionally and CI tests unprivileged agents on Windows.

leehinman: `--unprivileged` should have no effect.  This code only runs during install, where you still need Admin privs, a little farther down in this function we set the SecurityDescriptor, also something that needs Admin privs.

changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
```diff
@@ -0,0 +1,32 @@
...
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
```
blakerouse, line 28: Link to this PR.

changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
```
blakerouse, line 32: Link to the issue.

internal/pkg/agent/install/uninstall.go
```diff
@@ -116,6 +116,14 @@ func Uninstall(cfgFile, topPath, uninstallToken string, log *logp.Logger, pt *pr
 		pt.Describe("Successfully uninstalled service")
 	}
 
+	// Removing the external logger isn't fatal to uninstall
+	err = removeExternalLogger()
```
blakerouse, line 120: Should this really be done? Is it harmful to keep it around? Upon performing uninstall does it remove the entries, or does this still keep the entries?

With https://github.com/elastic/elastic-agent/pull/4822 two Elastic Agent's will be installed at once. In this case if only one is uninstalled then this will be removed and the still installed Elastic Agent will not be able to log.

leehinman: This would definitely break in the 2 agent case.  The Remove just removes the registry entries, so not a big deal to leave them around.  I'll remove.

changelog/fragments/1717516439-Capture-early-errors-on-Windows.yaml
blakerouse: Did it ever? I have always had to do it manually.

blakerouse (approved):
Looks good.

cmacknz:
Bypass the unit test coverage requirement, unit test things makes no sense.

Issue #4810, opened 2024-05-24
[Flaky Test]: Failed to prepare an instance – could not run "unzip agent-repo.zip -d agent" though SSH: wait: remote command exited without exit status or exit signal


rdner:
### Failing test case

Multiple upgrade tests

### Error message

could not run "unzip agent-repo.zip -d agent" though SSH: wait: remote command exited without exit status or exit signal

### Build

* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/130
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/153
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/155
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/180
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/219
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/244
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/284
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/292
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/309

### OS

Linux

### Stacktrace and notes

```markdown
>>> (linux-amd64-ubuntu-2204-upgrade) Failed for instance linux-amd64-ubuntu-2204-upgrade (@ 34.16.95.179): failed to execute tests on instance ogc-linux-amd64-ubuntu-2204-upgrade-9d60: error running sudo tests: failed to fetched test output at $HOME/agent/build/TEST-go-remote-linux-amd64-ubuntu-2204-upgrade-sudo.integration.out
Error: error running test: failed to copy files to instance ogc-linux-arm64-ubuntu-2204-container-b24e: failed to unzip agent-repo.zip to agent directory: could not run "unzip agent-repo.zip -d agent" though SSH: wait: remote command exited without exit status or exit signal (stdout: Archive:  agent-repo.zip
  inflating: agent/.agent-versions.json
  inflating: agent/.buildkite/hooks/post-checkout
  inflating: agent/.buildkite/hooks/post-checkout.ps1
  inflating: agent/.buildkite/hooks/pre-command
  inflating: agent/.buildkite/hooks/pre-command.ps1
  inflating: agent/.buildkite/hooks/pre-exit
  inflating: agent/.buildkite/integration.pipeline.yml
  inflating: agent/.buildkite/misc/gce-cleanup.yml
[...]

PR #4847, opened 2024-06-04
Use higher tier ARM VM for integration testing

rdner:
ARM cores are lower performance, to make it equal to our AMD64 VMs we should use higher tier instances.


cmacknz (approved):

cmacknz:
Rationale https://github.com/elastic/elastic-agent/issues/4810#issuecomment-2147870609

PR #4859, opened 2024-06-06
Repoint kardianos/service to fork elastic/kardianos-service

ycombinator:
## What does this PR do?

This PR repoints the dependency on github.com/kardianos/service to our fork of it at github.com/elastic/kardianos-service via a `replace` directive.

## Why is it important?

The upstream fork at https://github.com/kardianos/service does not seem to be actively maintained based on the latest commit timestamp and the lack of response to https://github.com/kardianos/service/issues/396.

More pressingly, we need a fix to https://github.com/kardianos/service/issues/395.

## Disruptive User Impact

No.


go.mod
```diff
@@ -291,6 +291,7 @@ replace (
 	github.com/Shopify/sarama => github.com/elastic/sarama v1.19.1-0.20220310193331-ebc2b0d8eef3
 	github.com/dop251/goja => github.com/andrewkroh/goja v0.0.0-20190128172624-dd2ac4456e20
 	github.com/dop251/goja_nodejs => github.com/dop251/goja_nodejs v0.0.0-20171011081505-adff31b136e6
+	github.com/kardianos/service => github.com/elastic/kardianos-service v0.0.0-20230215200102-9832e01049dd
```
michel-laterman, line 294: Do we have a follow up issue to change the import in our codebase?

michel-laterman (approved):

go.mod
ycombinator: I wasn't planning on changing the imports.  Do you think we need to? 

If we changed the imports, we could get rid of this replace directive and directly require `github.com/elastic/kardianos-service` instead.  Would we then also have to update the module name here: https://github.com/elastic/kardianos-service/blob/master/go.mod#L1?

Not saying I'm opposed to changing the imports, just wondering if we need to and, if so, what other changes might need to go hand in hand with that.

leehinman, line 294: I think go.mod will allow a comment, if so can you please add one saying why we need to the replace directive for kardianos.

blakerouse:
I would rather see this as a hard fork. Where we just maintain it as its own project. We have some other template changes in agent that could be merged as well, making it cleaner as well.

What about naming it 'elastic/go-service'?

ycombinator:
> I would rather see this as a hard fork. Where we just maintain it as its own project. We have some other template changes in agent that could be merged as well, making it cleaner as well.
> 
> What about naming it 'elastic/go-service'?

I'm not opposed to this idea, considering it's probably unlikely that the upstream repo will become active again any time soon.  But if the upstream repo did become active again some day, it would be nice to contribute our fixes back to the community.

Just to understand the full picture, what are the downsides of keeping the fork as-is? Are the template changes you're mentioning not generic enough? Is it the `kardianos` in the name? Something else?

blakerouse:
The template changes are very generic, could also go upstream if they were active.

The reason for request to use name 'elastic/go-service' is that it will only require updating the import and not the actual usage of the module in the code. Golang will remove the 'go-' in the front automatically allowing all the 'service.*' usage to remain the same in the code base. It's also very common naming scheme. I also just generally like it, haha.

ycombinator:
> The reason for request to use name 'elastic/go-service' is that it will only require updating the import and not the actual usage of the module in the code. Golang will remove the 'go-' in the front automatically allowing all the 'service.*' usage to remain the same in the code base.

Ah, clever!  I've renamed the repository to https://github.com/elastic/go-service and updated the reference in `go.mod` in this PR via 4c49871df534bf62044cc54580f89b3218c92435.

Did you also want me to get rid of the `replace`, directly `require` https://github.com/elastic/go-service, and change the imports in the code to match?

blakerouse:
> > The reason for request to use name 'elastic/go-service' is that it will only require updating the import and not the actual usage of the module in the code. Golang will remove the 'go-' in the front automatically allowing all the 'service.*' usage to remain the same in the code base.
> 
> Ah, clever! I've renamed the repository to https://github.com/elastic/go-service and updated the reference in `go.mod` in this PR via [4c49871](https://github.com/elastic/elastic-agent/commit/4c49871df534bf62044cc54580f89b3218c92435).
> 
> Did you also want me to get rid of the `replace`, directly `require` https://github.com/elastic/go-service, and change the imports in the code to match?

Yes, please!

ycombinator:
>> Did you also want me to get rid of the replace, directly require https://github.com/elastic/go-service, and change the imports in the code to match?

> Yes, please!

This change depends on making this change first: https://github.com/elastic/go-service/pull/6

blakerouse:
@ycombinator I updated this to point to the new `github.com/elastic/go-service` and removed and cleaned up the items that where merged into `go-service`.

ycombinator:
Integration tests are failing consistently on this PR.  I'm going to temporarily remove the code changes in `internal/pkg/agent/install/svc.go` made as part of ba18e5e566b8ef59778ade501dcaed898c59b7ad and also point the `github.com/elastic/go-service` dependency to an earlier version of the fork, after we forked it but before we made an code changes, so 32a7f923456f477b48ecef50525d54735fcda6b7.  

If this configuration works, we know something's off with either the code changes I'm removing from this PR or with changes made in https://github.com/elastic/go-service after 32a7f923456f477b48ecef50525d54735fcda6b7 or both.

cc: @blakerouse

ycombinator:
I have zero recollection of merging this PR 😮.  CI is still failing so I'm not sure why / how it got merged.  Reverting...

PR #4862, opened 2024-06-06
Disable controller code for otel mode

michalpristas:
Disables vars manager for otel mode as it's not needed and it's producing a lot of confusing logs


cmacknz:
A configuration option to disable all providers was added in https://github.com/elastic/elastic-agent/pull/4166, although it would be better if we just didn't execute any part of the controller in otel mode.

ycombinator:
@michalpristas This feels like a bug fix.  Should we backport it?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -683,10 +683,16 @@ func (c *Coordinator) Run(ctx context.Context) error {
...
 	defer close(c.stateBroadcaster.InputChan)
 
-	// The usual state refresh happens in the main run loop in Coordinator.runner,
-	// so before/after the runner call we need to trigger state change broadcasts
-	// manually with refreshState.
-	c.setCoordinatorState(agentclient.Starting, "Waiting for initial configuration and composable variables")
+	if c.varsMgr != nil {
+		// The usual state refresh happens in the main run loop in Coordinator.runner,
+		// so before/after the runner call we need to trigger state change broadcasts
+		// manually with refreshState.
```
blakerouse, line 689: This comment is talking about the `c.refreshState()` call on line 696. This should really stay above that line.

blakerouse (approved):
Simple change. Just one comment to address before merging.

PR #4885, opened 2024-06-07
[Ind Agent] Update packaging to properly package from manifest if given

dwhyrock:
## What does this PR do?

This PR updates the packaging code to operate differently when packaging from a local build vs packaging from a manifest.

The manifest packaging relies on the items in the manifest, while the local one uses globbing to find all packages/directories that match the expected version string.

## Why is it important?
For the Independent Agent Release, a manifest is created that includes opted-in projects that have a patch version one higher than the rest of the opted-out/previously-released projects.  The file globbing expecting one single version is what skips the opted-in projects.

Previously, we had been packaging the incorrect opted-in projects, which is why we had verified those projects were properly packaged.  But when that bug was fixed, we did not properly verify that everything was still there since the packaging still succeeded.

## Checklist

 - [ ] packaging dependencies with a manifest for an Independent Agent build
 - [ ] packaging dependencies with a manifest for a Unified Release / Stack build
 - [ ] packaging dependencies pulling the latest unified release version
 - [ ] packaging dependencies pulling from a DROP_PATH
 - [ ] packaging by building beats dependencies located side by side with the agent code

## Related issues

- Relates https://github.com/elastic/security-team/issues/9684

## Questions to ask yourself

- ...


dwhyrock:
I've added in some code here so that the packaging code looks for the expected version (e.g., `8.14.0`) as well as the bumped Independent Agent version for opted-in projects (e.g., `8.14.1`).

However, it's starting to get a little complex.  The latest error I'm seeing [is this](https://buildkite.com/elastic/elastic-agent-package/builds/1628#018ff44b-5df5-47bf-a0e7-08cbfe880cda/439-448):

```
Error: failed copying spec file "specs/endpoint-security-8.14.1-darwin-x86_64.spec.yml" to "build/elastic-agent-drop/darwin-x86_64.tar.gz/endpoint-security-8.14.1-darwin-x86_64.spec.yml": copy failed: cannot stat source file specs/endpoint-security-8.14.1-darwin-x86_64.spec.yml: stat specs/endpoint-security-8.14.1-darwin-x86_64.spec.yml: no such file or directory
```

It appears as if it's not copying the proper `endpoint-security` spec file to the right location with the bumped version number.

I'm thinking that for `elastic-agent-package` pipeline builds kicked off by the Independent Agent workflow, we should maybe pass in an env var that includes a comma-separated list of opted-in projects.  Maybe something like:

```
INDEPENDENT_AGENT_OPTIN_PROJECTS="endpoint-dev,another-project"
```

Then we could use that var as kind of a gate around the logic I have put in here already that includes a bumped version, and for the copying of the `spec` files to the proper versions for the proper projects.

Thoughts?  @gogochan @cmacknz @pchila @pierrehilbert

gogochan:
IMO, having optional `INDEPENDENT_AGENT_OPTIN_PROJECTS` environment variable to control what to package makes sense. If it is not specified, the pipeline can always fall back to the default logic. I do not have in-depth understanding of this pipeline to foresee any issues with this approach though.

dev-tools/mage/manifest/manifest.go
```diff
@@ -110,6 +110,16 @@ func DownloadComponentsFromManifest(manifest string, platforms []string, platfor
 	// This eliminates the "+buildYYYYMMDDHHMM" suffix on Independent Agent Release builds
 	majorMinorPatchVersion := parsedManifestVersion.VersionWithPrerelease()
 
+	// For Independent Agent Releases, any opted-in projects will have a version that is
+	// one Patch version ahead since we are using the new bumped DRA artifacts for those projects.
+	// This is acceptable since it is being released only as bundled with Elastic Agent.
+	patchPlusOneString := fmt.Sprintf("%d.%d.%d", parsedManifestVersion.Major(), parsedManifestVersion.Minor(), parsedManifestVersion.Patch()+1)
```
cmacknz, line 116: Is there a reason why the input manifest can't specify the patch+1 version? That would keep the logic for this in the release jobs, with the agent package target just doing what the manifest tells it to.

cmacknz:
> IMO, having optional INDEPENDENT_AGENT_OPTIN_PROJECTS environment variable to control what to package makes sense. 

Instead of an environment variable, can we have the manifest specify the mixed set of versions? Alternatively, could this variable just be in the manifest? That way the manifest would define the contract for what agent should do.

dev-tools/mage/manifest/manifest.go
dwhyrock: that may be a decent way to go.

I think the way it works is that it uses the `version` value in the header, which is something like `8.13.4+build202406101001`.  And then we strip off the metadata to get the base 3-number version for the previously-released (opted-out) projects.

We could probably somewhat easily add something like an `optin_version` in the header, but that doesn't tell us exactly _which_ projects should use that version.

As it is now, we'd probably have to go through and parse all of the packages in each of the projects and find any that have a filename that includes the `patch+1` version in it.

Perhaps we could also include just a list of `opted_in_projects` at the top level?  

We probably need to ask @DaveSys911 whether adding fields to the Ind Agent manifest would be a big issue or not.

cmacknz:
CC @pchila for opinions since I think he implemented the packaging target that is used here.

dwhyrock:
Thanks for the great response, @pchila !  

I agree that a more robust solution is probably better for the long and should make it more reliable by relying on the manifest.

I'll work on changes based on your recommendation.  When closer to completion, I may need help verifying that it works in all of those cases as I'm not fully fluent in how they are executed.

dwhyrock:
@pchila @cmacknz @gogochan 

I have this PR ready for review and testing now.  I am testing what I know how to test, namely the staging builds for both unified release and independent agent release.

I don't believe I have quite a full/proper setup locally to do the DROP_PATH or locally built beats, etc.  Could someone from the Agent/Ingest side test those out with this branch, please?

Also, let me know what else, if anything, I need to do about docs/changelog/etc.

dev-tools/mage/manifest/manifest.go
```diff
@@ -68,16 +69,75 @@ func DownloadManifest(manifest string) (tools.Build, error) {
 }
 
 func resolveManifestPackage(project tools.Project, pkg string, reqPackage string, version string) []string {
+	var val tools.Package
+	var ok bool
+
+	// Try the normal/easy case first
 	packageName := fmt.Sprintf("%s-%s-%s", pkg, version, reqPackage)
-	val, ok := project.Packages[packageName]
+	val, ok = project.Packages[packageName]
```
cmacknz, line 77: Is there a sample manifest we can check in here and use to write a unit test against this function? Seeing an actual example would also help me understand if this is correctly doing what it is supposed to do.

dev-tools/mage/manifest/manifest.go
```diff
@@ -68,16 +69,75 @@ func DownloadManifest(manifest string) (tools.Build, error) {
...
+					}
+
+					foundIt = true
+				}
+			}
+		}
+
+		if !foundIt {
+			return nil
+		}
```
cmacknz, line 133: Should this actually be an error? I am worried about silently dropping a package from the build.

magefile.go
```diff
@@ -980,8 +1004,28 @@ func packageAgent(platforms []string, dependenciesVersion string, agentPackaging
...
+				// When getting the packageVersion from snapshot we should also update the env of SNAPSHOT=true which is
+				// something that we use as an implicit parameter to various functions
+				if parsedVersion.IsSnapshot() {
+					os.Setenv(snapshotEnv, "true")
+					mage.Snapshot = true
+				}
+				os.Setenv("BEAT_VERSION", parsedVersion.CoreVersion())
+			}
+		}
+	}
```
cmacknz, line 1025:
```suggestion
		if manifestResponse, err = manifest.DownloadManifest(devtools.ManifestURL); err != nil {
			log.Panicf("failed to download remote manifest file %s", err)
		}

		if parsedVersion, err := version.ParseVersion(manifestResponse.Version); err != nil {
			log.Panicf("the manifest version from manifest is not semver, got %s", manifestResponse.Version)
		} else {
			// When getting the packageVersion from snapshot we should also update the env of SNAPSHOT=true which is
			// something that we use as an implicit parameter to various functions
			if parsedVersion.IsSnapshot() {
				os.Setenv(snapshotEnv, "true")
				mage.Snapshot = true
			}
			os.Setenv("BEAT_VERSION", parsedVersion.CoreVersion())
		}
	}
```

The else here makes this harder to follow and you should be able to remove it.

magefile.go
```diff
@@ -1174,9 +1179,248 @@ func collectPackageDependencies(platforms []string, packageVersion string, requi
...
+
+	return checksums
+}
+
+// This function is used when building with a Manifest.  In that manifest, it's possible
+// for projects in an Independent Agent Release to have different versions since the opted-in
+// ones will be one patch version higher than the opted-out/previously released projects.
+// This function tries to find the versions from the package name
+func getComponentVersion(componentName string, requiredPackage string, componentProject tools.Project) string {
```
cmacknz, line 1239: This could probably be moved into manifest.go and unit tested. It seems like it is doing a similar job to `resolveManifestPackage` but I'm not sure if the `parsing` could be refactored out and shared between the two.

magefile.go
```diff
@@ -1174,9 +1179,248 @@ func collectPackageDependencies(platforms []string, packageVersion string, requi
...
+		panic(errMsg)
+	}
+
+	return componentVersion
+}
+
+// This is a helper function for the cloud-defend package.
+// When it is untarred, it does not have the same dirname as the package name.
+// This adjusts for that and returns the actual path on disk for cloud-defend
+func fixCloudDefendDirPath(dirPath string, componentVersion string, expectedArch string, actualArch string) string {
```
cmacknz, line 1297: Could we just fix this in cloud-defend directly at some point? https://github.com/elastic/cloud-defend

cmacknz: Also FYI that we only include defend in the container build: https://github.com/elastic/elastic-agent/pull/4584/files

dev-tools/mage/manifest/manifest.go
dwhyrock: Adding a few manifests (snapshot, staging, and ind. agent staging) to test this function.

dev-tools/mage/manifest/manifest.go
dwhyrock: Good call, I'll make it an error.

dwhyrock: I just tested this out, and I don't think it should be an error.  The logic for whether or not something was found is here, and seems to just log that "Project [%s] does not have [%s]".

https://github.com/elastic/elastic-agent/blob/main/dev-tools/mage/manifest/manifest.go#L125-L139

Going to leave it as returning `nil` for now.

I do think there should be some checking at some point to verify all expected packages for a given os/arch exists.  That probably would have caught when endpoint was no longer being packaged.

That said, I'm not sure I understand all of the intricacies to make that change right now.

magefile.go
dwhyrock: I agree that there is a fair bit of overlap in these two functions, but they do loop over and compare slightly different things.

I'm heading out for 2-week vacation, so @gogochan will be watching this PR.  If he has some spare time, he may be able to come up w/ a decent solution.

dev-tools/mage/manifest/test_payload/manifest-8.14.0+build202406201002.json
```diff
@@ -0,0 +1,6167 @@
...
+    "version": "8.14.0+build202406201002",
+    "build_id": "8.14.0+build202406201002",
+    "start_time": "Thu, 20 Jun 2024 10:02:13 UTC",
+    "end_time": "Thu, 20 Jun 2024 10:49:18 UTC",
+    "build_duration_seconds": 2825,
+    "manifest_version": "2.1.0",
+    "prefix": "",
+    "projects": {
+        "apm-server": {
+            "branch": "main",
```
cmacknz, line 13: Not part of this PR, but seeing this as `main` is a bit weird since it should be `8.14.

dev-tools/mage/manifest/manifest_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"os"
+	"path"
+	"runtime"
+	"testing"
+
+	"github.com/elastic/elastic-agent/pkg/testing/tools"
+	"github.com/stretchr/testify/assert"
+)
+
+// Helper to get the absolute path of the parent directory name of the current file
```
cmacknz, line 15: Instead of this you can just embed the file content directly into a variable.

https://pkg.go.dev/embed

Example: https://github.com/elastic/elastic-agent/blob/92eeced1af2e5e46a8345a88406fb195b6e8cddd/testing/integration/endpoint_test_tools.go#L27-L28

cmacknz: Here's a patch making this change.

```diff
diff --git a/dev-tools/mage/manifest/manifest_test.go b/dev-tools/mage/manifest/manifest_test.go
index 50c2e39f1f..7d3dde46ff 100644
--- a/dev-tools/mage/manifest/manifest_test.go
+++ b/dev-tools/mage/manifest/manifest_test.go
@@ -1,41 +1,40 @@
 package manifest
 
 import (
+	_ "embed"
 	"encoding/json"
 	"log"
-	"os"
-	"path"
-	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/elastic/elastic-agent/pkg/testing/tools"
 	"github.com/stretchr/testify/assert"
 )
 
-// Helper to get the absolute path of the parent directory name of the current file
-func ParentDir() string {
-	_, filename, _, _ := runtime.Caller(1)
-	return path.Dir(filename)
-}
+var (
+	//go:embed test_payload/manifest-8.14.2.json
+	manifest8_14_2 string
 
-func getManifestJsonData(t *testing.T, filePath string) tools.Build {
-	var response tools.Build
+	//go:embed test_payload/manifest-8.14.2-SNAPSHOT.json
+	manifest8_14_2_SNAPSHOT string
 
-	contents, err := os.Open(filePath)
-	assert.NoError(t, err)
+	//go:embed test_payload/manifest-8.14.0+build202406201002.json
+	manifest8_14_0_build202406201002 string
+)
 
-	err = json.NewDecoder(contents).Decode(&response)
+func getManifestJsonData(t *testing.T, contents string) tools.Build {
+	var response tools.Build
+
+	err := json.NewDecoder(strings.NewReader(contents)).Decode(&response)
 	assert.NoError(t, err)
 
 	return response
 }
 
 func TestBlah(t *testing.T) {
-	parentDir := ParentDir()
-
 	tcs := []struct {
 		name            string
-		filePath        string
+		file            string
 		componentName   string
 		packageName     string
 		requiredPackage string
@@ -43,7 +42,7 @@ func TestBlah(t *testing.T) {
 	}{
 		{
 			name:            "Unified Release Staging 8.14 apm-server",
-			filePath:        path.Join(parentDir, "test_payload", "manifest-8.14.2.json"),
+			file:            manifest8_14_2,
 			componentName:   "apm-server",
 			packageName:     "apm-server",
 			requiredPackage: "linux-x86_64.tar.gz",
@@ -55,7 +54,7 @@ func TestBlah(t *testing.T) {
 		},
 		{
 			name:            "Unified Release Snapshot 8.14 apm-server",
-			filePath:        path.Join(parentDir, "test_payload", "manifest-8.14.2-SNAPSHOT.json"),
+			file:            manifest8_14_2_SNAPSHOT,
 			componentName:   "apm-server",
 			packageName:     "apm-server",
 			requiredPackage: "linux-x86_64.tar.gz",
@@ -67,7 +66,7 @@ func TestBlah(t *testing.T) {
 		},
 		{
 			name:            "Independent Agent Staging 8.14 apm-server",
-			filePath:        path.Join(parentDir, "test_payload", "manifest-8.14.0+build202406201002.json"),
+			file:            manifest8_14_0_build202406201002,
 			componentName:   "apm-server",
 			packageName:     "apm-server",
 			requiredPackage: "linux-x86_64.tar.gz",
@@ -79,7 +78,7 @@ func TestBlah(t *testing.T) {
 		},
 		{
 			name:            "Unified Release Staging 8.14 endpoint-dev",
-			filePath:        path.Join(parentDir, "test_payload", "manifest-8.14.2.json"),
+			file:            manifest8_14_2,
 			componentName:   "endpoint-dev",
 			packageName:     "endpoint-security",
 			requiredPackage: "linux-x86_64.tar.gz",
@@ -91,7 +90,7 @@ func TestBlah(t *testing.T) {
 		},
 		{
 			name:            "Unified Release Snapshot 8.14 endpoint-dev",
-			filePath:        path.Join(parentDir, "test_payload", "manifest-8.14.2-SNAPSHOT.json"),
+			file:            manifest8_14_2_SNAPSHOT,
 			componentName:   "endpoint-dev",
 			packageName:     "endpoint-security",
 			requiredPackage: "linux-x86_64.tar.gz",
@@ -103,7 +102,7 @@ func TestBlah(t *testing.T) {
 		},
 		{
 			name:            "Independent Agent Staging 8.14 endpoint-dev",
-			filePath:        path.Join(parentDir, "test_payload", "manifest-8.14.0+build202406201002.json"),
+			file:            manifest8_14_0_build202406201002,
 			componentName:   "endpoint-dev",
 			packageName:     "endpoint-security",
 			requiredPackage: "linux-x86_64.tar.gz",
@@ -118,7 +117,7 @@ func TestBlah(t *testing.T) {
 
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			manifestJson := getManifestJsonData(t, tc.filePath)
+			manifestJson := getManifestJsonData(t, tc.file)
 			log.Printf("Manifest Version: [%s]", manifestJson.Version)
 
 			projects := manifestJson.Projects
```

cmacknz: Was able to just push this change myself.

dev-tools/mage/manifest/manifest.go
cmacknz: This does need to be an error. We have an internal list of expected binaries, and this ensures we can find all of them in the manifest. One of them missing is a fatal error.

I suspect there are some platform dependencies to handle though.

cmacknz: Handled the platform dependencies.

cmacknz:
Tried the following and they both work:

```
MANIFEST_URL=https://snapshots.elastic.co/8.15.0-875a6caa/manifest-8.15.0-SNAPSHOT.json AGENT_DROP_PATH=./build/drop PLATFORMS=darwin/arm64 PACKAGES=targz mage -v downloadManifest packageUsingDRA
MANIFEST_URL=https://staging.elastic.co/independent-agent/8.14.0+build202406261002/manifest-8.14.0+build202406261002.json AGENT_DROP_PATH=./build/drop PLATFORMS=darwin/arm64 PACKAGES=targz mage -v downloadManifest packageUsingDRA
```

The only oddity was the 8.14.0 manifest created a `elastic-agent-8.14.0-darwin-aarch64.tar.gz` but this is probably because I didn't set the agent package version properly. Possibly a nice enhancement someday would be to pick up the agent version based on the manifest version.

magefile.go
```diff
@@ -961,27 +952,41 @@ func runAgent(env map[string]string) error {
...
 	requiredPackages := []string{}
 	for _, p := range platforms {
-		requiredPackages = append(requiredPackages, platformPackages[p])
+		requiredPackages = append(requiredPackages, manifest.PlatformPackages[p])
 	}
 	if mg.Verbose() {
 		log.Printf("--- Packaging dependenciesVersion[%s], %+v \n", dependenciesVersion, requiredPackages)
 	}
 
+	if devtools.PackagingFromManifest {
```
pchila, line 968: I would move this block in `PackageUsingDRA` target (that's what we are gonna use for independent releases anyways and we already have to download the manifest ) and `Package` target  (in case someone want to still build the agent binary but use the dependencies from a manifest) and pass in a pointer to a `tools.Build` struct (possibly nil if no MANIFEST_URL is there)

cmacknz: Done in https://github.com/elastic/elastic-agent/pull/4885/commits/7754ffdc96a0cc14e2148b0ab35e3419fdc770ba

pchila (approved):

blakerouse (approved):
Looks good. Like that it has tests for `dev-tools` code, always nice to have.

PR #4909, opened 2024-06-11
Ensure socket path is < 104 characters when STATE_PATH is set.

belimawr:
## What does this PR do?

It ensures the `statePath` set by the `container` command will always generate an Unix socket path that is smaller than 108 characters. This ensures the Unix socket can be created and the Elastic-Agent always successfully starts.

## Why is it important?

Unix socket paths need to be less than 104 characters long, if they're not Elastic-Agent will fail to create a socket and will not start. When it fails it logs those messages:
```
{"log.level":"error","@timestamp":"2024-06-10T21:34:44.818Z","log.logger":"control","log.origin":{"function":"github.com/elastic/elastic-agent/pkg/control/v2/server.(*Server).Start","file.name":"server/server.go","file.line":85},"message":"unable to create listener: listen unix /tmp/TestContainerCMDEventToStderr1447633574/001/elastic-agent-8.15.0-SNAPSHOT-linux-x86_64/data/elastic-agent.sock: bind: invalid argument","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2024-06-10T21:34:44.818Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.logReturn","file.name":"cmd/run.go","file.line":147},"message":"listen unix /tmp/TestContainerCMDEventToStderr1447633574/001/elastic-agent-8.15.0-SNAPSHOT-linux-x86_64/data/elastic-agent.sock: bind: invalid argument","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
```

## Checklist

## Disruptive User Impact
The Elastic-Agent container command will not always respect the `STATE_PATH` environment variable. In cases where it would lead to a Unix socket path too long, which prevents the Elastic-Agent from starting, the default path `/usr/share/elastic-agent/state` is used.

1. Get a Stack + Fleet deployed
2. Set the following environment variables according to your deployment:
    - `FLEET_ENROLL`
    - `FLEET_URL`
    - `FLEET_ENROLLMENT_TOKEN`
3. Set `STATE_PATH` to be longer than 100 characters
3. Run the Elastic-Agent container command
4. Assert:
    i. The Elastic-Agent runs and can communicate with Fleet
    ii. The state path is the default path :`/usr/share/elastic-agent/state`

Another option is to run the integration test `TestContainerCMDWithAVeryLongStatePath` from `testing/integration/container_cmd_test.go`

~~## Related issues~~

## Questions to ask yourself

- ...


belimawr:
> ## [Quality Gate failed](https://sonar.elastic.dev/dashboard?id=elastic_elastic-agent_AYluowg0xMq8P7b4moiZ&pullRequest=4909) **Quality Gate failed**
> Failed conditions [](https://sonar.elastic.dev/component_measures?id=elastic_elastic-agent_AYluowg0xMq8P7b4moiZ&pullRequest=4909&metric=new_coverage&view=list) [0.0% Coverage on New Code](https://sonar.elastic.dev/component_measures?id=elastic_elastic-agent_AYluowg0xMq8P7b4moiZ&pullRequest=4909&metric=new_coverage&view=list) (required ≥ 40%)
> 
> [See analysis details on SonarQube](https://sonar.elastic.dev/dashboard?id=elastic_elastic-agent_AYluowg0xMq8P7b4moiZ&pullRequest=4909)

The changes are covered by integration tests.

internal/pkg/agent/cmd/container.go
```diff
@@ -771,9 +771,10 @@ func logToStderr(cfg *configuration.Configuration) {
 
 func setPaths(statePath, configPath, logsPath, socketPath string, writePaths bool) error {
 	statePath = envWithDefault(statePath, "STATE_PATH")
-	if statePath == "" {
+	if isStatePathTooLong(statePath) || statePath == "" {
```
cmacknz, line 774: I'm not sure this is correct, people change the state path to mount persistent volumes to the container so that the filebeat registry and fleet state aren't ephemeral. Silently doing something else isn't right.

It appears all you need to do imove where the control socket gets created, why didn't you do that instead?

We already have a helper for that, does it not work in the container file system? 

https://github.com/elastic/elastic-agent/blob/414ae849ebd5d67b9f6e369753607083a6c013f9/pkg/utils/socket_unix.go#L21-L30

internal/pkg/agent/cmd/container.go
```diff
@@ -962,3 +963,13 @@ func isContainer(detail component.PlatformDetail) component.PlatformDetail {
...
 }
+
+// isStatePathTooLong returns true if the final Unix socket path
+// will be too long and needs to be shortened.
+//
+// Source: https://www.man7.org/linux/man-pages/man7/unix.7.html
+func isStatePathTooLong(statePath string) bool {
+	// This replicates the logic from `setPaths`
+	socketPath := filepath.Join(statePath, "data", paths.ControlSocketName)
+	return len(socketPath) > 107
```
cmacknz, line 974: Why 107? We are using 104 everywhere else.

internal/pkg/agent/cmd/container.go
belimawr: > I'm not sure this is correct, people change the state path to mount persistent volumes to the container so that the filebeat registry and fleet state aren't ephemeral. Silently doing something else isn't right.

That did cross my mind when creating this PR. The Agent failing to start didn't seem a good behaviour either.

> It appears all you need to do imove where the control socket gets created, why didn't you do that instead?

It didn't work, the Agent was filing to find the configuration file, the config path and state path can be the same if `CONFIG_PATH` is not set: https://github.com/elastic/elastic-agent/blob/d79b4ef6887959833e8fea0d31dc5c1a5678d1d1/internal/pkg/agent/cmd/container.go#L778-L781

I don't know much about this part of the Elastic-Agent, so I tried to keep the changes to a minimum.

I can go back dig deeper on why I got failures when changing only the socket path.

internal/pkg/agent/cmd/container.go
belimawr: After reading https://www.man7.org/linux/man-pages/man7/unix.7.html, 107 seemed the most appropriate. It was also the longest I could get without the socket failing to be created.

I can chance to 104.

internal/pkg/agent/cmd/container.go
cmacknz: We need to do something else other than silently not use the state path, this will lead to extremely surprising behavior for some people on Kubernetes. Failing to start the agent is probably more correct than doing this as it tells the user they have something to fix.

belimawr: I managed to get the container command to work with the socket path outside of the state path. Now we explicitly create the config path (before it was only the top path).

I hope nothing outside this function makes any assumption about the socket.

internal/pkg/agent/cmd/container.go
belimawr: It got removed with my lats few commits

testing/integration/container_cmd_test.go
```diff
@@ -2,14 +2,16 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-//go:build integration
+//go:build integration && linux
```
michel-laterman, line 5: Do we also want to run these tests on darwin?

michel-laterman (commented):
Just have a couple of nitpicks

Also if the STATE_PATH is too long and the default is used, would that be clear in diagnostics?

testing/integration/container_cmd_test.go
```diff
@@ -112,22 +94,52 @@ func TestContainerCMD(t *testing.T) {
...
+		"STATE_PATH=" + agentFixture.WorkDir(),
+	}
 
+	cmd := prepareContainerCMD(t, ctx, agentFixture, info, env)
 	t.Logf(">> running binary with: %v", cmd.Args)
 	if err := cmd.Start(); err != nil {
 		t.Fatalf("error running container cmd: %s", err)
 	}
 
+	agentOutput := cmd.Stderr.(*strings.Builder)
```
pchila, line 141: I would prefer that the `strings.Builder` pointer is returned as a separate value by `prepareContainerCMD` instead of a direct cast here

pchila (approved):
LGTM, no major blockers, Just a nitpick on a direct cast for the agent cmd in test.

While I understand that this may not work for everyone, I feel that the state path should be mounted on a well known location as a local persistent volume or some other volume stored on the host that should have the same lifetime as the deployment: having the `STATE_PATH` env var does not feel very container/cloud-native..

testing/integration/container_cmd_test.go
belimawr: It's a test, I didn't want to add returned values not every test is gonna use. If I end up making changes, I might add it as well ;)

testing/integration/container_cmd_test.go
belimawr: They're testing the container command, we only release Linux images, so I don't see the point of running them on Darwin.

If you look at the tests in the file, they all define only Linux:
```go
OS: []define.OS{
			{Type: define.Linux},
		},
```

Either way they will never run on Darwin. Having it only running on Linux also allows us to use some Linux specific code, which I had used in previous iterations of this test.

internal/pkg/agent/cmd/container.go
```diff
@@ -774,13 +775,20 @@ func setPaths(statePath, configPath, logsPath, socketPath string, writePaths boo
 	if statePath == "" {
 		statePath = defaultStateDirectory
 	}
+
 	topPath := filepath.Join(statePath, "data")
 	configPath = envWithDefault(configPath, "CONFIG_PATH")
 	if configPath == "" {
 		configPath = statePath
 	}
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
belimawr: Thanks!

internal/pkg/agent/cmd/container.go
cmacknz: If you exec into the container and run the `elastic-agent status` command does it work?

testing/integration/container_cmd_test.go
cmacknz: I don't think we want to run the tests on Darwin but we don't want engineers on MacOS to be able to compile this file without jumping through hoops.

What problem is this solving? This seems like it is likely to waste MacOS users time when CI finds build errors they could have found locally.

testing/integration/container_cmd_test.go
```diff
@@ -140,6 +152,114 @@ func TestContainerCMD(t *testing.T) {
...
+			statePath:          "",
+			expectedStatePath:  "/usr/share/elastic-agent/state",
+			expectedSocketPath: "/usr/share/elastic-agent/state/data/Td8I7R-Zby36_zF_IOd9QVNlFblNEro3.sock",
+		},
+		"107 characters path": { // Longest path that still works for creating a unix socket, but will use /tmp/elastic-agent
+			statePath:          "/tmp/tttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt",
+			expectedStatePath:  "/tmp/tttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt",
+			expectedSocketPath: "/tmp/elastic-agent/oKUUJbxrLlGSh3z6wZWYleLeMuUN4P0_.sock",
+		},
+		"long path": { // Path too long to create a unix socket, it will use /tmp/elastic-agent
```
cmacknz, line 199: Aren't most of these test cases just tests of the `SocketURLWithFallback()` function? How long does this new set of tests take to run?

Do you get the same effective test coverage if you update the single existing container command to use a path longer than 107 characters as the exception path?

cmacknz:
Once this compiles, make sure the `elastic-agent status` command still works inside the container.

testing/integration/container_cmd_test.go
belimawr: > Aren't most of these test cases just tests of the SocketURLWithFallback() function? How long does this new set of tests take to run?

No, they're testing the [`https://github.com/elastic/elastic-agent/blob/a6d91b029ae8c0b8ccf7bb391c8ac304fe1e9368/internal/pkg/agent/cmd/container.go#L773-L837`](https://github.com/elastic/elastic-agent/blob/a6d91b029ae8c0b8ccf7bb391c8ac304fe1e9368/internal/pkg/agent/cmd/container.go#L773-L837) function and how it decides to create the socket path. The fact that it calls `SocketURLWithFallback` is an implementation detail. It also needs to create some folders:
- https://github.com/elastic/elastic-agent/blob/a6d91b029ae8c0b8ccf7bb391c8ac304fe1e9368/internal/pkg/agent/cmd/container.go#L784-L788
- https://github.com/elastic/elastic-agent/blob/a6d91b029ae8c0b8ccf7bb391c8ac304fe1e9368/internal/pkg/agent/cmd/container.go#L794-L796

All of this gets tested by the integration test.

Maybe we can remove one test case: `107 characters path` as it essentially is the same as `small path`.

belimawr:
> Once this compiles, make sure the `elastic-agent status` command still works inside the container.

It works:
```
% k exec -it elastic-agent-vspdt -- /bin/bash
]0;root@kind-control-plane: /usr/share/elastic-agentroot@kind-control-plane:/usr/share/elastic-agent# elastic-agent status
┌─ fleet
└─ elastic-agent
   └─ status: (HEALTHY) Running
]0;root@kind-control-plane: /usr/share/elastic-agentroot@kind-control-plane:/usr/share/elastic-agent# 
```

testing/integration/container_cmd_test.go
belimawr: Indeed at the moment it solves no problem because I hardcoded `/tmp/elastic-agent/`, but there is a constant for it that is on a Unix-only file. I can remove it if you folks prefer.

cmacknz: Yeah let's remove it, the intent isn't that these tests run on the host system, only that they compile there. Let's keep it that way as much as possible.

jlind23:
Everything works except he sonar unit code coverage which is superseded by the integration tests hence force merging this.

PR #4912, opened 2024-06-12
simplify packaging pre-requisites

codefromthecrypt:
## What does this PR do?

Changes the packaging docs to be more concise about the need for beats to exist in the same directory as elastic-agent was cloned into.

## Why is it important?

Before, the discussion of x-pack had me curious if I needed to have that directory as a peer instead. 

## Checklist

~~- [ ] My code follows the style guidelines of this project~~

## Disruptive User Impact

N/A

N/A

## Questions to ask yourself

N/A


README.md
```diff
@@ -47,7 +47,7 @@ pr: https://github.com/elastic/elastic-agent/pull/823
 Prerequisites:
 - installed [mage](https://github.com/magefile/mage)
 - [Docker](https://docs.docker.com/get-docker/)
-- [X-pack](https://github.com/elastic/beats/tree/main/x-pack) to pre-exist in the parent folder of the local Git repository checkout
+- [beats](https://github.com/elastic/beats) to pre-exist at ../beats
```
pierrehilbert, line 50: Any reason to change the wording here and not keeping ` in the parent folder of the local Git repository checkout`?

AndersonQ: it comes from a slack thread. The agent does not need `x-pack` it actually needs `beats` and needs it in the same level as the agent repo itself. I believe it might have been copied from beats and not updated.

This is what we need:

```
your-root-folder
├── agent
└── beats
    └── x-pack
```

AndersonQ, line 50: Actually the `beats` repo is only required if compiling and packaging the beats as well, with `EXTERNAL=false`. Therefore it can also be stated there:

```suggestion
- [beats](https://github.com/elastic/beats) to pre-exist at ../beats if, and only if, packaging with `EXTERNAL=false` to package the beats as well
```

AndersonQ: @codefromthecrypt would you mind adding what I mentioned above?

pierrehilbert: Yes my point was not about changing X-Pack by Beats but keeping the fact that beats repository should be in the parent folder of the Agent git repository.

codefromthecrypt: no worries @AndersonQ I put back the original wording. My concision hammer went too far ;)

codefromthecrypt: good idea, done!

pierrehilbert (approved):

Issue #4087, opened 2024-01-15
Remove the Docker elastic-agent user from the root group


cmacknz:
The non-root elastic-agent user with UID 1000 is a member of the root group 0 in our Docker file.

This is configured here with the `--gid 0` argument to `useradd`:

https://github.com/elastic/elastic-agent/blob/17f048008fcea105d0b86cc90743c7bfb8ce2a41/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl#L115-L118

```sh
elastic-agent@e31774b0cea6:~$ whoami
elastic-agent
elastic-agent@e31774b0cea6:~$ groups elastic-agent
elastic-agent : elastic-agent root
```

The root group on an Ubuntu container does not have any particular privileges besides the ability to read files owned by the root group. Being a member of the root group does occasionally cause security scans to consider our user as "running as root". This is confusing now that we are developing an Elastic Agent that can be installed as a non-root user (including both UID and GID). The container should also follow this convention. Today our container `elastic-agent` user can read files an agent installed with `elastic-agent install --unprivileged` cannot.

We should remove the default addition to the root group in our container. This may cause problems in situations where directories or volumes mounted into the container from the host system for monitoring require membership in the root group to read the files in the volume. Possibly examples of this situation are using the elastic-agent container to monitoring node metrics or system logs on Kubernetes by mounting the file system of the node into the elastic-agent container.

pierrehilbert:
Should it be the new default or should it be something optional in the same way we are doing for this unpriviledged option?

jlind23:
> Should it be the new default or should it be something optional in the same way we are doing for this unpriviledged option?

After chatting with Craig we should probably ship two different docker images, the one we already have today and a new one compliant with this request.

cmacknz:
> After chatting with Craig we should probably ship two different docker images, the one we already have today and a new one compliant with this request.

I think we need to establish why this exists first, that is not entirely clear to me. If we consider removing our user from the root group a breaking change, then yes we'll probably need to publish a separate image without it. 

For Docker ideally the unprivileged one would be the default but I'm not sure if this is possible depending on what it breaks.

blakerouse:
I don't think we should publish two docker containers. A possible solution is to add a container option that drops the root group from the elastic-agent in the docker entrypoint script before it spawns the Elastic Agent. That would allow us to have the Elastic Agent as running as not having root privileges.

blakerouse:
[...]

Issue #1540, opened 2022-10-11
osquery won't install when deployed via Elastic Agent integrations on k8s


ty-elastic:
Hi, I deployed Endpoint and Agent 8.4.x into the daemonset on my self-managed k8s cluster [per this yaml](https://raw.githubusercontent.com/elastic/endpoint/main/releases/8.4.0/kubernetes/deploy/elastic-endpoint-security.yaml), and then deployed osquery via Fleet Integrations.

Upon deployment, I see this error in the agent logs:
```
06:25:26.900
elastic_agent.osquerybeat
[elastic_agent.osquerybeat][error] Failed to run osquery:W1005 11:24:26.929394  7382 extensions.cpp:426] Will not autoload extension with unsafe directory permissions: /usr/share/elastic-agent/data/elastic-agent-d3eb3e/install/osquerybeat-8.4.2-linux-x86_64/osquery-extension.ext
E1005 11:24:26.953727  7382 shutdown.cpp:79] Cannot activate osq_config config plugin: Unknown registry plugin: osq_config: exit status 78
06:25:26.900
elastic_agent.osquerybeat
[elastic_agent.osquerybeat][info] osquerybeat context cancelled, exiting
```

[this](https://osquery.readthedocs.io/en/stable/deployment/extensions/#extensions-binary-permissions) suggests that osquery extensions must not have write permissions for non-privileged accounts.

Yet after install, if I ssh into the daemonset I see this:
```
# pwd
/usr/share/elastic-agent/data/elastic-agent-d3eb3e/install/osquerybeat-8.4.2-linux-x86_64
# ls -l
total 431832
-rw-r--r-- 1 elastic-agent elastic-agent     13675 Sep 13 21:23 LICENSE.txt
-rw-r--r-- 1 elastic-agent elastic-agent   2571228 Sep 13 21:23 NOTICE.txt
-rw-r--r-- 1 elastic-agent elastic-agent       828 Sep 13 21:58 README.md
drwxr-xr-x 2 elastic-agent elastic-agent        23 Sep 14 22:38 certs
-rw-r--r-- 1 root          root             389399 Sep 13 21:50 fields.yml
drwxr-x--- 3 root          root                 88 Oct  5 19:30 osquery
-rwxr-xr-x 1 elastic-agent elastic-agent   6173182 Sep 13 21:58 osquery-extension.ext
-rwxr-xr-x 1 elastic-agent elastic-agent 219834144 Sep 13 21:57 osquerybeat
[...]

PR #4925, opened 2024-06-13
Extend agent container initialisation logic

pkoutsovasilis:
## What does this PR do?

This PR affects only the `DOCKER` package of elastic-agent (code path of `elastic-agent container ...`). Specifically it does the following:
- allow all capabilities as permitted at the elastic-agent binary

- adds init logic to raise all capabilities present in Permitted set when running as non-root
- adds init logic to chown agent-related paths, including state, config, logs, data, etc. if either CAP_CHOWN or DAC_OVERRIDE capabilities is given

Flowchart for visual aid
```mermaid
flowchart TD
    classDef default line-height:1.5,text-align:center;
    entrypoint((("elastic-agent<br>container #quot;$@#quot;")))
    style entrypoint fill:#222,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
    ifIsRoot{Is Root?}
    canChown{Has either CHOWN<br>or DAC_OVERRIDE<br>capability?}
    raiseEffective[raise all capabilities from Permitted<br>set to Effective and Inheritable sets]
    canPcap{Has SETPCAP<br>capability?}
    raiseAmbient[raise all capabilities from<br>Effective set to Ambient set]
    chownPaths["chown all agent-related paths"]
    existingCode[existing code]
    style existingCode fill:#222,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5

    entrypoint --> ifIsRoot
    ifIsRoot -->|True|canChown
    ifIsRoot --> |False|raiseEffective
    raiseEffective --> canPcap
    canPcap --> |True|raiseAmbient
    canPcap ---> |False|canChown
    raiseAmbient -->canChown
    canChown --> |True|chownPaths
    canChown --> |False|existingCode
    chownPaths --> existingCode
```

## Why is it important?

I consider this PR important for the following reasons:
- it can allow detaching elastic-agent user (inside the container) from the root group by enforcing CAP_CHOWN capability which amplifies the security
- allows for non-root invocation of elastic-agent to do pretty much everything, given the correct capabilities e.g. if we want to read k8s container logs just add the capability `DAC_READ_SEARCH` without the need to run as root or being part of the root group.

## Checklist

## Disruptive User Impact

- For any reason that elastic-agent container is invoked for a completely unknown user (e.g. uid=500 and gid=500) the chowing phase of the initialisation can take some seconds to complete before the normal elastic-agent operation starts. Based on some initial experiment the "chowing" phase for this particular case can take ~15 seconds.

```
$ DEV=true SNAPSHOT=true PLATFORMS=linux/amd64 PACKAGES=docker mage -v package
$ docker run --entrypoint bash --user 1000:1000 -v $PWD/tmp_state:/etc/state --env STATE_PATH=/etc/state  --cap-drop ALL --cap-add CAP_BPF --cap-add CAP_CHOWN --cap-add CAP_SETFCAP --cap-add CAP_SETPCAP --rm -it docker.elastic.co/beats/elastic-agent-complete:8.16.0-SNAPSHOT
$ elastic-agent container
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/4087#issuecomment-2158699181
- Relates https://github.com/elastic/ingest-dev/issues/3253


internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,313 @@
...
+				args = append(args, os.Args[1:]...)
+			}
+			// add skipFileCapabilitiesFlag flag to skip reapplying the file capabilities
+			args = append(args, fmt.Sprintf("--%s", skipFileCapabilitiesFlag))
+
+			cmd := exec.Command(executable, args...)
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.Env = os.Environ()
+			return cmd, nil
```
blakerouse, line 63: This isn't really a re-exec, this is running it as a subprocess for the live of the Elastic Agent. Is that what really should be done or should it be a true re-exec?

We have the ability to do a true re-exec, that is how the Elastic Agent handles self-upgrading.

internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,313 @@
...
+// excluding CAP_CHOWN, CAP_SETPCAP, and CAP_SETFCAP.
+func getAmbientCapabilitiesFromEffectiveSet() ([]cap.Value, error) {
+	set := capProc()
+	var caps []cap.Value
+
+	for capVal := cap.Value(0); capVal < cap.MaxBits(); capVal++ {
+
+		switch capVal {
+		case cap.CHOWN, cap.SETPCAP, cap.SETFCAP:
+			// don't set these as they shouldn't be required by any exec'ed child process
```
blakerouse, line 128: does this affect the ability for the Elastic Agent running as non-root to write a file and set its ownership to its current user? I believe there are specific code paths where it does perform chown always to ensure that the ownership is correct, will that need to be skipped?

blakerouse (commented):
Haven't made it through the whole PR. Just some early comments as I was going.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -216,6 +217,8 @@ RUN for iter in {1..10}; do \
     (exit $exit_code)
 
 {{- end }}
+# root group no more for elastic-agent user
+RUN gpasswd --delete {{ .user }} root
```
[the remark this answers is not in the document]
pkoutsovasilis: my initial approach was to remove it there but apparently the npm flow is a little bit entangled with root and non-root and when I do remove it there I get the following our.

```
> [stage-1 15/17] RUN npm i -g --loglevel verbose --engine-strict @elastic/synthetics@stack_release &&     chmod ugo+rwX -R /usr/share/elastic-agent/.node:
0.479 npm verb exit -13
0.480 npm verb unfinished npm timer command:i 1718110923068
0.480 npm verb unfinished npm timer reify 1718110923070
0.480 npm verb unfinished npm timer reify:loadTrees 1718110923094
0.480 npm verb unfinished npm timer idealTree 1718110923094
0.480 npm verb unfinished npm timer idealTree:userRequests 1718110923097
0.480 npm verb code -13
0.480
0.480 npm ERR! Log files were not written due to an error writing to the directory: /usr/share/elastic-agent/.npm/_logs
0.480 npm ERR! You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
------
Dockerfile:164
--------------------
 163 |     # If this fails dump the NPM logs
 164 | >>> RUN npm i -g --loglevel verbose --engine-strict @elastic/synthetics@stack_release && \
 165 | >>>     chmod ugo+rwX -R $NODE_PATH
 166 |     USER root
--------------------
ERROR: failed to solve: process "/bin/sh -c npm i -g --loglevel verbose --engine-strict @elastic/synthetics@stack_release &&     chmod ugo+rwX -R $NODE_PATH" did not complete successfully: exit code: 243
```

Thus I preferred to tackle this by removing it at the end 😄

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: fantastic! true re-exec is the way to go, can you please point me to the appropriate line in the code so I can try to incorporate it? 🙂

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: no these are skipped only for Ambient permissions thus only subprocesses won't have them. the agent binary has cap_chown so it will always be able to chown files

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: oh unix.Exec 🥲 ok I am gonna try to incorporate it

jlind23 (commented):
Does this PR only Relates or it actually resolves https://github.com/elastic/elastic-agent/issues/4087#issuecomment-2158699181 and https://github.com/elastic/ingest-dev/issues/3253

pkoutsovasilis:
> Does this PR only Relates or it actually resolves [#4087 (comment)](https://github.com/elastic/elastic-agent/issues/4087#issuecomment-2158699181) and [elastic/ingest-dev#3253](https://github.com/elastic/ingest-dev/issues/3253)

Hello @jlind23 👋  this PR actually resolves https://github.com/elastic/elastic-agent/issues/4087 as it removes elastic-agent from root group. https://github.com/elastic/ingest-dev/issues/3253 essentially has two requirements rootless agent and helm chart for agent; this PR resolves the rootless requirement

internal/pkg/agent/cmd/container.go
```diff
@@ -145,6 +150,9 @@ occurs on every start of the container set FLEET_FORCE to 1.
 			}
 		},
 	}
+
+	cmd.Flags().BoolVar(&skipFileCapabilities, skipFileCapabilitiesFlag, false, "")
```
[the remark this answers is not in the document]
pkoutsovasilis: I switched to this pattern, feel free to unresolved if you spot anything else

internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,313 @@
...
+)
+
+// initContainer applies the following container initialisation steps:
+// - set any missing capabilities of Effective set based on the Bounding set at elastic-agent binary
+// - raise capabilities of the Ambient set to match the Effective set
+// - chown all agent-related paths if DAC_OVERRIDE capability is not in the Effective set
+// If new binary capabilities are set then the returned cmd will be not nil. Note that it is up to caller to invoke
+// the returned cmd and spawn an agent instance with all the capabilities.
+func initContainer(streams *cli.IOStreams) (*exec.Cmd, error) {
+	isRoot, _ := utils.HasRoot()
```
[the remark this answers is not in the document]
pkoutsovasilis: you are 100% right, error is handled now, feel free to unresolve if you spot anything else

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: thanks for the hint/tip here, I switched to unix.Exec, feel free to unresolve if you spot anything else

cmacknz:
What is the best way to add automated tests for this?

We don't have a great template for testing agent in a container yet.
- Our integration testing framework can give you a Linux VM you can do whatever you want on.
- There is a very basic testing using kind to ensure agent can deploy https://github.com/elastic/elastic-agent/blob/main/.buildkite/scripts/steps/k8s-tests.sh
- The Buildkite agents can start containers but we don't do this with the agent container yet.

cmacknz:
No, the state of our k8s specific testing is poor

blakerouse:
I don't think we should block this PR on having an integration test. We are not in a worse place with this PR, versus how it is today. Would really like to see if validated with an integration test, but we need to spend more time on enabling that. Blocking this PR because of that is not great.

internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,312 @@
...
+		return true, err
+	}
+	if !skipFileCapabilities && !isRoot {
+		executable, err := os.Executable()
+		if err != nil {
+			return true, err
+		}
+
+		logInfo(streams, "agent container initialisation - file capabilities")
+		updated, err := updateFileCapsFromBoundingSet(executable)
```
blakerouse, line 48: Could this be done with out the `--skip-file-capabilities` flag? Like determine which file caps and missing and apply them and then re-exec? Now the new re-exec will do the same thing and see that it has all the capabilities and continue on?

Would that be bettter and cleaner?

internal/pkg/agent/cmd/container_init_linux.go
blakerouse: wonder how that affects some of the subprocess that store state, like filebeat that writes a registry.

I would hope that it doesn't cause an issue.

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: I totally agree with skipping this flag completely, it was just my just-in-case thinking that made me add this flag if somebody wanted to skip this step for any reason 🥲 but I will remove it

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: I have tested manually this agent image with kubernetes integration and the container logs appear just fine but I trust your gut instincts here. What do you say to allow CAP_CHOWN? CAP_SETPCAP and CAP_SETFCAP shouldn't be necessary right?

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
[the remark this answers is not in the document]
pkoutsovasilis: what do you think, should we leave this as is for this PR or should I try to untangle root vs rootless npm?

internal/pkg/agent/cmd/container_init_linux.go
blakerouse: Can you think of a reason they would want to skip it?

They can still choose to run as root and it will still work the same, correct?

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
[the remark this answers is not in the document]
blakerouse: NPM is only really use for the -complete image. It would be nice if it could be figured out. Downside of leaving it like this is it adds another layer to the container. Very small layer but still another layer.

internal/pkg/agent/cmd/container_init_linux.go
blakerouse: I would say it's only the chown that I worry about. Just because it shipping the logs doesn't mean it's having issues with writing the registry. You could also just check the file system and logs to see if the registry is working correctly. 

If it doesn't seem like much of a concern to allow them to keep that ability we might want to just to be safe.

pkoutsovasilis: ok CAP_CHOWN allowed then just for now to be better safe than sorry. But maybe in the future we would like to extend a bit the package spec (agent integrations) and there capture all linux capabilities needed for each integration (and beat) so we know exactly what to allow and what to filter out 🙂

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: >Can you think of a reason they would want to skip it?

Tbh I don't have any reason reason in mind 

>They can still choose to run as root and it will still work the same, correct?

Yep 100% correct, running agent with root equals that we won't deal with agent capabilities at all

cmacknz:
> I don't think we should block this PR on having an integration test. We are not in a worse place with this PR, versus how it is today. Would really like to see if validated with an integration test, but we need to spend more time on enabling that. Blocking this PR because of that is not great.

We don't need to block the PR on this, but we should:
1. Define what the ideal test actually is and what the minimum we need for that is so we can actually build it. 
2. We do not need this to be a complete functional test of agent on Kubernetes, only the permissions related logic added in this PR. Is there a way we can partially test this with what we have today? For example in the very basic kind test today, can we at least make sure the agent will actually start as both root and non-root? Is there anything we can validate just be running the `elastic-agent container` on a Linux host that is not a container (probably not given this is updating the Dockerfile but you get the spirit of the idea).

cmacknz:
For another example may be able to get Buildkite to just directly invoke the agent container as a build step since Buildkite can run containers without modification. See https://github.com/elastic/elastic-agent-system-metrics/pull/140 for an example buildkite pipeline that has a step running a specific container and a define command to run on that container.

pkoutsovasilis:
> We don't need to block the PR on this, but we should:
> 
> 1. Define what the ideal test actually is and what the minimum we need for that is so we can actually build it.
> 2. We do not need this to be a complete functional test of agent on Kubernetes, only the permissions related logic added in this PR. Is there a way we can partially test this with what we have today? For example in the very basic kind test today, can we at least make sure the agent will actually start as both root and non-root? Is there anything we can validate just be running the `elastic-agent container` on a Linux host that is not a container (probably not given this is updating the Dockerfile but you get the spirit of the idea).

An ideal test IMO would be something similar to [this](https://github.com/elastic/cloud-on-k8s/pull/7356/files#diff-5c93c76848058958da5d8cc4c66d0f169ff6f95ef56dd4d6c2dbbe263d40a18dR19). This particular test, does exactly what we need, deploys an Elasticsearch cluster though ECK operator, deploys agent with kubernetes integration enabled through a Helm chart (spoiler alert 🚨), and checks that the expected events are getting delivered to the expected data_streams. So some ways forward that pop up in my mind are:

- utilise ECK operator in the kind clusters part of k8s tests to spawn an Elasticsearch and then do a similar testing here [I think this is what ECK is doing]
- since this is kind based k8s cluster maybe we can utilise elastic-package to spawn a stack that is reachable through k8s workloads and once again do similar testing
- spawn an Elasticsearch on Elastic cloud through the terraform here (I think I've seen some tf here in the repo doing such a provision) and do our testing.

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: I ve removed the flag, and I have manually tested that this doesn't affect the flow, feel free to unresolve if you spot anything

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: ok CAP_CHOWN is part of Ambient Set and thus given to the individual beats spawned by agent, feel free to unresolve if you spot anything

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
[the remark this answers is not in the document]
pkoutsovasilis: ok I sorted out the npm installation under elastic-agent user, the feel free to unresolve if you spot anything

cmacknz:
> Merging this PR will break elastic-package because of the change introduced https://github.com/elastic/elastic-package/pull/1794

This is interesting, and this indicates this could be considered a breaking change for users who have done this, because clearly Elastic Agent can do useful work with no capabilities in some situations. We at least cannot merge this without coordinating with elastic-packge. CC @mrodm

pkoutsovasilis:
> > Merging this PR will break elastic-package because of the change introduced https://github.com/elastic/elastic-package/pull/1794
> 
> This is interesting, and this indicates this could be considered a breaking change for users who have done this, because clearly Elastic Agent can do useful work with no capabilities in some situations. We at least cannot merge this without coordinating with elastic-package. CC @mrodm 

that's true and I was planning to open a PR in elastic package before merging this PR. but the drop all addition there is still kinda incompatible with the agent because of this line in dockerfile

```
RUN setcap cap_net_raw,cap_setuid+p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/components/heartbeat && \
```

right? and now that I mention this out load since heartbeat requires these capabilities doesn't agentbeat require the same?!? 🤔

cmacknz:
It should, `heartbeat` shouldn't even exist anymore in the agent build so that line may failing silently which is great.

blakerouse (commented):
I am +1 on the code and changes that this introduces.

I believe the integration testing is up in the air for this change. We do have this test https://github.com/elastic/elastic-agent/blob/main/testing/integration/container_cmd_test.go#L25 which runs the `elastic-agent container` subcommand directly. We also support running as without root permissions by changing this flag to false (`sudo: false`). Could that possibly be used to test this?

blakerouse (approved):
Not looking to block this PR on the integration tests. I believe we know what we are missing, I would like to get this in before 8.15 FF.

Please merge if you think this is ready, I do.

pkoutsovasilis:
> Not looking to block this PR on the integration tests. I believe we know what we are missing, I would like to get this in before 8.15 FF.
> 
> Please merge if you think this is ready, I do.

Thanks for the review and the reminder of the 8.15 FF @blakerouse. Since at the moment we lack k8s integration testing, should we try to get a review also from fellows that know how to test a container image for [this issue](https://github.com/elastic/elastic-agent/issues/4087)?  and maybe also from the maintainers of elastic-package as they currently drop ALL capabilities?

blakerouse:
@pkoutsovasilis More testers the better!

jsoriano:
> Ofc I can help testing this if you have something specific to openshift.

If I remember correctly, to test if this would work on Openshift, we should check that the image works when running it with `-u <somerandomid>`.

pkoutsovasilis:
> > Ofc I can help testing this if you have something specific to openshift.
> 
> If I remember correctly, to test if this would work on Openshift, we should check that the image works when running it with `-u <somerandomid>`.

you mean as a normal docker container invocation right?

jsoriano:
> you mean as a normal docker container invocation right?

yes, well, or the equivalent if the container is started in some other way

pkoutsovasilis:
@blakerouse @cmacknz @jsoriano I added two video with two different cases in the PR description. When you find some time, please do take a look and tell me your thoughts 🙂

cmacknz:
Will take a look. The most important thing to establish is whether this is a breaking change for some users or use cases. 

I have some concern that removing the ability to drop all capabilities may put us in violation of the internal security policies of some users, but I don't know of any specific examples off the top of my head.

jsoriano:
> @blakerouse @cmacknz @jsoriano I added two video with two different cases in the PR description. When you find some time, please do take a look and tell me your thoughts 🙂

Thanks for the videos, the behaviour looks good to me. Only one detail, in Openshift and other securized kubernetes environments, when the user id is randomized, no group id is assigned. In these cases the user ends up in the root group, and this is the main reason to give ownership to the root group, you can find more details in the point about supporting arbitrary user ids in https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines.
 
In your tests you are using `-u 140:140`, I wonder if there would be any difference if you use `-u 140`, or `-u 140:0`.

pkoutsovasilis:
> Thanks for the videos, the behaviour looks good to me. Only one detail, in Openshift and other securized kubernetes environments, when the user id is randomized, no group id is assigned. In these cases the user ends up in the root group, and this is the main reason to give ownership to the root group, you can find more details in the point about supporting arbitrary user ids in https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines.
> 
> In your tests you are using `-u 140:140`, I wonder if there would be any difference if you use `-u 140`, or `-u 140:0`.

@jameadded added videos with just `-u 140` 🙂

pkoutsovasilis:
> Will take a look. The most important thing to establish is whether this is a breaking change for some users or use cases.
> 
> I have some concern that removing the ability to drop all capabilities may put us in violation of the internal security policies of some users, but I don't know of any specific examples off the top of my head.

@cmacknz if the case you mention does actually exist, then this is a breaking change for such users. That said, every k8s manifest of elastic-agent in the repo has it running as root user without dropping ALL capabilities. So the drop ALL is something that the users have added on their own.

cmacknz:
>  So the drop ALL is something that the users have added on their own.

Yes, we unfortunately have no way to know if anybody has done that. That we have at least one internal example of this already highly suggests there is at least one external user out there who has done this.

I am still thinking on if removing our ability to drop all capabilities will be something we regret long term. There is an alternate distroless base image coming soon, and it may be simpler for us to default that to not being in the root group and then allow adding the capabilities to it to escalate back to root.

pkoutsovasilis:
> Yes, we unfortunately have no way to know if anybody has done that. That we have at least one internal example of this already highly suggests there is at least one external user out there who has done this.
> 
> I am still thinking on if removing our ability to drop all capabilities will be something we regret long term. There is an alternate distroless base image coming soon, and it may be simpler for us to default that to not being in the root group and then allow adding the capabilities to it to escalate back to root.

we can update the existing k8s manifests to capture the change of the added capabilities and thus every user that grabs the latest manifest won't face this issue?! That said, I see your point and based on this I would propose to take a decision and maybe close this PR as this will be handled differently?!

internal/pkg/agent/cmd/container_init_test.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+func Test_chownPaths(t *testing.T) {
+	firstParentDir, err := os.MkdirTemp("", "test_chown")
+	require.NoError(t, err)
+	defer os.RemoveAll(firstParentDir)
+
+	secondParentDir, err := os.MkdirTemp("", "test_chown")
+	require.NoError(t, err)
+	defer os.RemoveAll(secondParentDir)
```
AndersonQ, line 27: [Suggestion]
You could use `t.TempDir()` instead, then Go will take care of the cleanup.

AndersonQ (dismissed):

cmacknz:
> we can update the existing k8s manifests to capture the change of the added capabilities and thus every user that grabs the latest manifest won't face this issue (PS: even if somebody drops ALL, the capabilities defined explicitly in the add section remain)?! 

In theory updating the manifests would work, but I have observed in support cases that users don't always upgrade the manifest when the upgrade the elastic-agent container version. There is nothing forcing these two things to happen together, so the manifest update sometimes gets forgotten, especially since 99% of the time a container update works fine with an out of date manifest.

> That said, I see your point and based on this I would propose to take a decision and maybe close this PR as this will be handled differently?!

The ideal is we always follow the principle of least privilege with our container by default, which means we can work with a non-root user in a non-root group with all capabilities dropped. We can't do this today without a break change even without this PR, because some users depend on agent being in a non-root group. There are users using the agent container to do things like pull data from s3, where they don't need the additional capabilities like DAC_READ_SEARCH that k8s container monitoring requires.

That we are going to support a new container with a distroless base image soon is a new development that I think lets us change the approach here to get to this least privileged default.

We may still need parts of this PR if we default to a completely unprivileged container, it will just start from a different baseline. TBD.

ycombinator:
Heya @pkoutsovasilis, looks like there are some conflicts on this PR now.  Also looks like there is some review feedback from @cmacknz and @AndersonQ.

cmacknz:
I'm still figuring out how much from this PR we need to keep. I don't think we need it to get the agent out of the root group anymore, we can do that with the upcoming Wolfi based container https://github.com/elastic/elastic-agent/pull/5062.

We may still need some of the chown related logic to deal with cases where agent has to run as root in security applications, https://github.com/elastic/elastic-agent/issues/1540 for example.

pkoutsovasilis:
> I'm still figuring out how much from this PR we need to keep. I don't think we need it to get the agent out of the root group anymore, we can do that with the upcoming Wolfi based container #5062.
> 
> We may still need some of the chown related logic to deal with cases where agent has to run as root in security applications, #1540 for example.

thanks for the reminder @ycombinator and thanks for jumping in @cmacknz , I went idle while waiting for the TBD. If chowning files to arbitrary uids is still needed I think the chown logic of this PR is helpful. Also if we need to be able to extract logs with rootless invocation, or do anything else that requires extra capabilities, then the ambient capabilities approach of this PR can also help. The only way to make this less "invasive" to existing k8s manifest that drop ALL capabilities is to implement this PR with "two binaries approach" that the agent container calls the other binary only when it detects that CAP_CHOWN, CAP_SETPCAP and CAP_SETFCAP are part of the bounding set. However such an approach was deemed not ideal through the discussion [here](https://github.com/elastic/elastic-agent/issues/4087#issuecomment-2152901739).

pkoutsovasilis:
@blakerouse @cmacknz @AndersonQ I have revisited the approach of this PR achieving the following:
- no need to add `cap_chown`, `cap_setfcap`, `cap_setpcap` at elastic-agent binary
- no need to re-exec elastic-agent
- no crashes when user drops ALL capabilities just the init steps introduced by this PR are gonna be skipped

so please re-review 🙂

cmacknz:
Will do, thanks for sticking with this!

internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,283 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package cmd
```
cmacknz, line 5: Can you add the explanation and rationale for why this exists from the PR description directly to this file?

internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,283 @@
...
+	pathsToChown.addPath(envWithDefault("", "LOGS_PATH"))
+	pathsToChown.addPath(envWithDefault("", "STATE_PATH"))
+	pathsToChown.addPath(envWithDefault("", "CONFIG_PATH"))
+	pathsToChown.addPath(envWithDefault("", "DATA_PATH"))
+	pathsToChown.addPath(envWithDefault("", "HOME_PATH"))
+	return pathsToChown.chown(uid, gid)
+}
+
+// distinctPaths represents a set of paths that do not overlap.
+type distinctPaths map[string]struct{}
```
cmacknz, line 222: There's no way deal with this in the standard library? https://pkg.go.dev/path/filepath#Rel perhaps?

internal/pkg/agent/cmd/container_init_linux.go
```diff
@@ -0,0 +1,283 @@
...
+		logInfo(streams, "agent container initialisation - ambient capabilities")
+		if err := raiseAmbientCapabilities(); err != nil {
+			logWarning(streams, err)
+		}
+	}
+
+	// we need to chown all paths
+	logInfo(streams, "agent container initialisation - chown paths")
+	if err := chownPaths(agentBaseDirectory); err != nil {
+		logWarning(streams, err)
```
cmacknz, line 77: Is this one actually fatal, or can we detect situations where it is likely to lead to problems? Like mounted files being owned by root and agent not being root?

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -53,6 +53,7 @@ RUN true && \
     true
 
 # Keep this after any chown command, chown resets any applied capabilities
+RUN setcap =p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/elastic-agent
```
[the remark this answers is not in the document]
pkoutsovasilis: it is not about the need of resetting any capabilities, this binary to begin with doesn't have any capabilities so none is reset. To be frank, I went with `=p` (resetting) and not `all+p` (appending) because we don't want any capability to be left in the `effective` set, as this defeats the support of dropping ALL capabilities. PS: when a binary has a capability set in it;s `effective` set but the capability can't be granted, then binary doesn's start and the OS returns permission denied. Hopefully that explanation helps?! 🙂

cmacknz:
This effectively deals with the problem in https://github.com/elastic/elastic-agent/issues/1540#issuecomment-2216866347 correct? The `uid: 0` in the k8s manifest will result in us chown-ing everything to uid 0?

This problem exists regardless of if the container defaults to elastic-agent:root or elastic-agent:elastic-agent as we want to do with the Wolfi image I think.

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: so my end goal here is with `distinctPaths` is to detect and exclude the overlapping paths (child->parent relation) from the recursive chowning. In a simple example 

I have these two paths:
- /path1 (agent-base)
- /path1/path2 (agent-state)

I shouldn't chown `/path1/path2` as this is gonna be chowed during `/path1` either way.
Sure `https://pkg.go.dev/path/filepath#Rel` can be used in a way to identify this parent->children path relationship but it isn't like the most straightforward way to do so 🙂 So, I don't see any great gains from switching to that but please if you tell me if have another idea I might miss something

pkoutsovasilis:
> This effectively deals with the problem in [#1540 (comment)](https://github.com/elastic/elastic-agent/issues/1540#issuecomment-2216866347) correct? The `uid: 0` in the k8s manifest will result in us chown-ing everything to uid 0?
> 
> This problem exists regardless of if the container defaults to elastic-agent:root or elastic-agent:elastic-agent as we want to do with the Wolfi image I think.

yep! given that we can raise to either CAP_CHOWN or CAP_DAC_OVERRIDE all files will change their uid and gid to the same ones as the process

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: I was thinking about the same, but I end up in the conclusion that this doesn't need to become fatal here because it is gonna become fatal later on when agent tries to touch/write something that it can't due to permissions 😄 thus I resorted to a warning like "hey grant this permission if you face any ownership problems"

cmacknz: Yeah my concern is that this error is very specific and includes the correction a user needs to make, and whatever we get later is going to be way more vague and this log may have rotated away or be easy to miss.

cmacknz:
Thanks, we definitely need this then to solve that problem and allow us to change UID regardless of what the default user and group membership is.

Now that we have k8s tests is it worth capturing the problem with osquery and mixed UIDs into a test that will only pass with this change? That will guarantee we don't regress.

internal/pkg/agent/cmd/container_init_linux.go
pkoutsovasilis: the thing that I am not entirely sure how to handle is that to detect imminent access problems I have to know if a given file will be read, written or both from the agent/agentbeat. This is somewhat tricky to capture and if you add also the access needs of individual beats through agentbeat becomes more tricky?! So not plenty of ideas at the moment 🙂

pkoutsovasilis:
> Thanks, we definitely need this then to solve that problem and allow us to change UID regardless of what the default user and group membership is.
> 
> Now that we have k8s tests is it worth capturing the problem with osquery and mixed UIDs into a test that will only pass with this change? That will guarantee we don't regress.

the challenges I can think of with such a test at the moment are:
- need to deploy managed agent and not standalone as the current k8s tests do
- need to code helper functions to install the osquery through fleet
- need to exec in all agent pods and do either an elastic-agent status (assuming this will report the status of osquery beat) or at least a ps aux?!

maybe all the above should be handled in a dedicated PR?!

pkoutsovasilis:
@cmacknz I introduced the support for k8s inner tests, which are essentially go unit-tests that will run inside the agent pod. 

Some background of the design choices to support k8s inner tests:
* tests requiring the tag `kubernetes_inner` as we don't want these to be part of the existing elastic-agent unit-tests
* the above binary is essentially copied by the k8s provisioner to the agent image that is about to get tested for the following reasons:
  - the provisioner knows the architecture of the k8s cluster thus it can compile correctly the tests binary
  - an equivalent of `kubectl cp ...` isn't suitable as it doesn't maintain the file mode and you have to chmod inside the pod to make it an executable. However, chmoding inside the pod isn't always allowed 🙂
* introduced the first k8s inner test that checks that all agent-related paths have the same uid and gid as the elastic-agent process

Now elastic-agent standalone in k8s is tested that runs without issues against the following scenarios:
- rootful - default deployment [no inner tests]
- rootful - drop ALL capabilities [no inner tests]
- rootful - drop ALL add CHOWN, SETPCAP capabilities [inner tests included]
- rootless - drop ALL add CHOWN, SETPCAP capabilities [inner tests included]
- rootless - drop ALL add CHOWN, SETPCAP capabilities with random uid and gid [inner tests included]

blakerouse (commented):
Overall I like the initialization logic.

The whole inner testing logic of executing a separate binary inside of the container seems complex and error prone to debug. Would it be possible to just change this to running commands directly from like `kubectl exec` instead of just running a golang binary inside of the container?

pkoutsovasilis:
> Overall I like the initialization logic.
> 
> The whole inner testing logic of executing a separate binary inside of the container seems complex and error prone to debug. Would it be possible to just change this to running commands directly from like `kubectl exec` instead of just running a golang binary inside of the container?

I agree that doing a separate binary is more error prone than something that runs as 

> Overall I like the initialization logic.
> 
> The whole inner testing logic of executing a separate binary inside of the container seems complex and error prone to debug. Would it be possible to just change this to running commands directly from like `kubectl exec` instead of just running a golang binary inside of the container?

I see your point, but given the limitation of testing inside a pod mentioned [here](https://github.com/elastic/elastic-agent/pull/4925#issuecomment-2245970808), this is the best approach I could come up with. However, I'm more than happy to remove the Kubernetes inner tests from here and isolate them in a separate PR to give it more thought. Does that sound like a good plan? 🙂

pkoutsovasilis:
@elastic/obs-cloudnative-monitoring @MichaelKatsoulis @constanca-m could you please review the files that you own? 🙂

MichaelKatsoulis (approved):

blakerouse (approved):
I am good with this change and I really appreciate all the testing you added. I think its better to get this in and go from there to sit on it any longer.

Issue #4874, opened 2024-06-06
Allow Fleet-Managed agents to send event log to stderr


belimawr:
Once https://github.com/elastic/elastic-agent/pull/4549 gets merged, Elastic-Agent will start logging events to a different file, at the moment this is configurable via `agent.logging.event_data` for a standalone Elastic-Agent, however Fleet-Managed agents do not receive this configuration via Fleet.

This we need to make it configurable via Fleet, and ideally using an environment variable in container environments like Docker and Kubernetes.

PR #4932, opened 2024-06-13
Update event logger configuration via Fleet and environment variable (container command)

belimawr:
## What does this PR do?

If Fleet sends an event logging output configuration different than the one that's running, save it to the encrypted store and re-exec the Elastic-Agent to use the new configuration.

This PR adds the ability to receive event logger configuration via Fleet. Previously only the log level was received via Fleet and persisted.

Fleet can store the logging configuration in the policy via the `overrides` option, whenever the Elastic-Agent receives the policy (including at startup), it correctly parses this configuration. This PR enable those values to be used to configure the event logger.

When a policy is received the policy handler compares `agent.logging.event_data.to_stderr` and `agent.logging.event_data.to_files` with it's current values, if the policy contains different values, they're persisted in the disk store, and the Elastic-Agent re-execs. When it re-starts it reads the new values from the persistent store and applies them.

## Note fore reviewers
~~This PR is built on top of https://github.com/elastic/elastic-agent/pull/4909, hence I'm keeping it in draft until https://github.com/elastic/elastic-agent/pull/4909 is merged and I can rebase onto `main`.~~

Because this PR enables changing the log output (from disk to stderr or vice versa) I believe it's better to just re-start the Elastic-Agent instead of trying to do it at runtime. This will help to keep the logs consistent and avoid any possible race condition or the necessity to lock the logger while stopping/starting new outputs.

## Why is it important?

Once https://github.com/elastic/elastic-agent/pull/4549 gets merged the event logger won't be configurable via Fleet, this PR enables their configuration via Fleet.

## Checklist

### Test the container command
1. Get VM and a Stack deployment (ESS is easier) (the deployment needs to be accessible by the VM) 
2. Get [`mock-es`](https://github.com/elastic/mock-es) and run it to return StatusNotAcceptable on all requests: `mock-es -nonindex 100`
3. Add an output to Fleet pointing to mock-es
4. Create a log file with a few lines, anything will do, the content is not important as long as there are at least a few lines
5. Create a Elastic-Agent policy using the output you just created
6. Add the "Custom Logs Integration" pointing to the file you created
7. Export the following environment variables, adjust the values to your environment:
    - `FLEET_ENROLL=1`
    - `FLEET_URL=https://fleet-server.elastic.co:443`
    - `FLEET_ENROLLMENT_TOKEN=c2VjcmV0LXRva2Vu`
    - `EVENTS_TO_STDERR=true`
8. Run the Elastic-Agent container command: `./elastic-agent container`
9. Assert you can see the dropped event logs on stderr:
    ```json
    {
      "log.level": "warn",
      "@timestamp": "2024-06-13T17:17:39.365Z",
      "message": "Cannot index event publisher.Event{Content:beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, Flags:0x1, Cache:publisher.EventCache{m:mapstr.M(nil)}, EncodedEvent:(*elasticsearch.encodedEvent)(0xc0031fd480)} (status=406): , dropping event!",
      "component": {
        "binary": "filebeat",
        "dataset": "elastic_agent.filebeat",
        "id": "log-mock-es-4ba82ed3_5321_43cc_9ac4_ac2737e02545",
        "type": "log"
      },
      "log": {
        "source": "log-mock-es-4ba82ed3_5321_43cc_9ac4_ac2737e02545"
      },
      "ecs.version": "1.6.0",
      "log.logger": "elasticsearch",
      "log.origin": {
        "file.line": 490,
        "file.name": "elasticsearch/client.go",
        "function": "github.com/elastic/beats/v7/libbeat/outputs/elasticsearch.(*Client).applyItemStatus"
      },
      "service.name": "filebeat",
      "log.type": "event"
    }
    ```

### Fleet-Managed Elastic-Agent
1. Get VM and a Stack deployment (ESS is easier) (the deployment needs to be accessible by the VM) 
2. Get [`mock-es`](https://github.com/elastic/mock-es) and run it to return StatusNotAcceptable on all requests: `mock-es -nonindex 100`
3. Add an output to Fleet pointing to mock-es
4. Create a log file with a few lines, anything will do, the content is not important as long as there are at least a few lines
5. Create a Elastic-Agent policy using the output you just created
6. Install the Elastic-Agent enrolling in the policy you just created
7. Add the new logging settings to the policy using the API/DevTools. You'll need the policy ID and policy name:
    ```
    PUT kbn:/api/fleet/agent_policies/071c91f3-8fce-480e-b625-381d70f98e53
    {
      "name": "aafoobar",
      "namespace": "default",
      "overrides": {
        "agent": {
          "logging": {
            "event_data": {
              "to_stderr": true,
              "to_files": false
            }
          }
        }
      }
    }
    ```
9. Add the "Custom Logs Integration" pointing to the file you created
10. Assert you can see the dropped event logs on stderr. Run `journalctl -lfu elastic-agent.service` and ensure you see logs like this:
    ```json
    {
      "log.level": "warn",
      "@timestamp": "2024-06-13T17:17:39.365Z",
      "message": "Cannot index event publisher.Event{Content:beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, Flags:0x1, Cache:publisher.EventCache{m:mapstr.M(nil)}, EncodedEvent:(*elasticsearch.encodedEvent)(0xc0031fd480)} (status=406): , dropping event!",
      "component": {
        "binary": "filebeat",
        "dataset": "elastic_agent.filebeat",
        "id": "log-mock-es-4ba82ed3_5321_43cc_9ac4_ac2737e02545",
        "type": "log"
      },
      "log": {
        "source": "log-mock-es-4ba82ed3_5321_43cc_9ac4_ac2737e02545"
      },
      "ecs.version": "1.6.0",
      "log.logger": "elasticsearch",
      "log.origin": {
        "file.line": 490,
        "file.name": "elasticsearch/client.go",
        "function": "github.com/elastic/beats/v7/libbeat/outputs/elasticsearch.(*Client).applyItemStatus"
      },
      "service.name": "filebeat",
      "log.type": "event"
    }
    ```

## Related issues
- Blocked by https://github.com/elastic/elastic-agent/pull/4909

## Questions to ask yourself

- ...

Closes https://github.com/elastic/elastic-agent/issues/4874


internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -311,9 +323,27 @@ func (h *PolicyChangeHandler) handlePolicyChange(ctx context.Context, c *config.
...
+	switch {
+	case p.config.Settings.EventLoggingConfig.ToFiles != new.Settings.EventLoggingConfig.ToFiles:
+		return true
+	case p.config.Settings.EventLoggingConfig.ToStderr != new.Settings.EventLoggingConfig.ToStderr:
+		return true
+	default:
+		return false
+	}
+}
+
```
blakerouse, line 346:
```suggestion
func (h *PolicyChangeHandler) eventLoggingHasChanged(new *configuration.Configuration) bool {
	switch {
	case h.config.Settings.EventLoggingConfig.ToFiles != new.Settings.EventLoggingConfig.ToFiles:
		return true
	case h.config.Settings.EventLoggingConfig.ToStderr != new.Settings.EventLoggingConfig.ToStderr:
		return true
	default:
		return false
	}
}

```

to make lint happy, easy change (should be done)

blakerouse (approved):
Looks good. Just get lint happy.

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -317,9 +329,27 @@ func (h *PolicyChangeHandler) handlePolicyChange(ctx context.Context, c *config.
 		return fmt.Errorf("applying FleetClientConfig: %w", err)
 	}
 
+	if loggingHasChanged {
```
cmacknz, line 332: This restart is only for the logging output correct? A log level change does not require this correct?

cmacknz: You should clarify this in a comment to ensure future readers do not interpret this as broader than it needs to be.

Also you should explain exactly why it won't work if you don't restart, so that we know if/when we can remove this later if we for some reason need to.

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -317,9 +329,27 @@ func (h *PolicyChangeHandler) handlePolicyChange(ctx context.Context, c *config.
...
+		// The only way to update the logging configuration
+		// is to re-start the Elastic-Agent
+		h.coordinator.ReExec(nil)
```
cmacknz, line 335: If you exposed EVENT_LOGGING_TO_STDERR and EVENT_LOGGING_TO_FILES as environment variables would this code even be necessary? On k8s this would automatically restart the container on every change.

The major con would be that changing these wouldn't work from Fleet outside of the container use case I suppose.

testing/integration/container_cmd_test.go
```diff
@@ -258,3 +275,175 @@ func TestContainerCMDWithAVeryLongStatePath(t *testing.T) {
...
+		2*time.Minute, time.Second,
+		"Elastic-Agent did not report healthy. Agent status error: \"%v\", Agent logs\n%s",
+		err, agentOutput,
+	)
+
+	assert.Eventually(t, func() bool {
+		agentOutputStr := agentOutput.String()
+		scanner := bufio.NewScanner(strings.NewReader(agentOutputStr))
+		for scanner.Scan() {
+			if strings.Contains(scanner.Text(), "Cannot index event publisher.Event") {
```
cmacknz, line 342: This doesn't actually test the logic that requires the agent re-exec because the settings don't change.

Can you modify it to test that as well? That could potentially be tested without the container command since I don't think the re-exec is container specific.

cmacknz:
> Because this PR enables changing the log output (from disk to stderr or vice versa) I believe it's better to just re-start the Elastic-Agent instead of trying to do it at runtime. This will help to keep the logs consistent and avoid any possible race condition or the necessity to lock the logger while stopping/starting new outputs.

Out of curiosity, did you try changing the log destination at runtime? Does Zap support it?

I'm not opposed to the re-exec, especially because it's on a path that will rarely change, but I think we should document the concrete reasons why it won't work. Having to put a lock inside the logger for every log statement would definitely be a good reason for example, assuming this just doesn't work.

I want to avoid this being like the Beats restart on output change, where we don't quite know exactly what the historical reason for it being necessary is without a bunch of investigation.

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
belimawr: > This restart is only for the logging output correct? A log level change does not require this correct?

Yes.

I'll better document it.

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
belimawr: Yes, this code is for the Fleet case. This PR addresses both cases: Fleet-managed agents and agents running in container, both standalone and fleet managed.

`EVENTS_TO_STDERR` is already exposed in the container command, the advantage of this single one over `EVENT_LOGGING_TO_STDERR` and `EVENT_LOGGING_TO_FILES` it prevents user error of configuring those two in a way that's invalid.


But to clarify, are you referring to the container case or a non-container execution of the agent?

testing/integration/container_cmd_test.go
belimawr: > This doesn't actually test the logic that requires the agent re-exec because the settings don't change.

No it does not. The intent of the test is to test the environment variable `EVENTS_TO_STDERR`

I thought I had added a test in the handler for that, but it seems I didn't :thinking: . I'll add another test to test the re-exec path.

belimawr:
> Out of curiosity, did you try changing the log destination at runtime? Does Zap support it?

It has been a while since I looked that deep into the code, but I recall that there was a way to at least change the zapcore from our logger, so at least for a single logger (`logp.Logger`) held by a single component/goroutine it is safe. Because we keep a global logger and create derivative loggers, I could not assure all logger instances could be modified at once and that doing so would be goroutine safe.

On a quick look at the Zap documentation it does not seem to be possible.

I remember trying it first and I recall having it working, but then I started thinking on the implications with multiple components running, more goroutines and larger scale, and got to the conclusion I wasn't confident enough it was safe to do. The effort to ensure it or make it safe did not seem a good option comparing with the re-exec that is much simpler and quicker to implement.

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
cmacknz: When I made this comment, I was thinking only of the container use case. Then I realized it also handles Fleet which we want to keep. Marking this as resolved.

cmacknz (approved):
Thanks! Let's backport to 8.15.0, this is small, well tested, and the event logging feature isn't as nice as it should be without this.

cmacknz:
Force merging past the unit test coverage failure.

PR #4964, opened 2024-06-19
Remove now unused Beat spec files

cmacknz:
Remove the Beat spec files since they have been obsoleted by the agentbeat spec file to avoid confusion over where to make changes.

The Filebeat spec file was used in several tests for creating components from complex spec files. Rather than rewrite them all, I renamed the filebeat spec file to testbeat and added a comment that it is a frozen in time 8.12.0 filebeat spec file used for testing.


blakerouse (approved):
Looks good.

PR #4995, opened 2024-06-24
Write `container-paths.yml` into the `STATE_PATH`

blakerouse:
## What does this PR do?

Adjusts the `container` subcommand to write the `container-paths.yml` into the `STATE_PATH` on startup.

## Why is it important?

When running the Elastic Agent container with `--read-only` it is not possible for the container to write the `container-paths.yml` files into the root filesystem of the container.

## Checklist

```
$ PLATFORMS="linux/arm64" PACKAGES="docker" SNAPSHOT=true EXTERNAL=true mage package
...
$ docker run --rm --mount type=tmpfs,destination=/state -e STATE_PATH=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0-SNAPSHOT
```

In another shell with the container running verify that the `status` command works.

```
$ docker exec -it ${container_id} bash
elastic-agent@7d7447b8f608:~$ elastic-agent status
┌─ fleet
└─ elastic-agent
   └─ status: (HEALTHY) Running
```


ycombinator:
Looking at https://github.com/elastic/ingest-dev/issues/3267, I see that we'd like to run Agent in a read-only container like so:

```
docker run --read-only docker.elastic.co/beats/elastic-agent:8.13.2
```

However, looking at the How to test this PR locally section of this PR, I see that we'd need to run Agent in a read-only container like so:

```
docker run --rm --mount type=tmpfs,destination=/state -e STATE_PATH=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0-SNAPSHOT
```

Just want to confirm that, even with the changes in this PR, the `-mount type=tmpfs,destination=/state -e STATE_PATH=/state` arguments will _also_ be needed?  I don't think there's any getting around them but just want to confirm that so we can then document this requirement along with the caveat you mentioned in this PR about not being able to use `elastic-agent` sub-commands inside the read-only container.

ycombinator:
@blakerouse Looking at https://github.com/elastic/ingest-dev/issues/3267, I see that we'd like to run Agent in a read-only container like so:

```
docker run --read-only docker.elastic.co/beats/elastic-agent:8.13.2
```

However, looking at the How to test this PR locally section of this PR, I see that we'd need to run Agent in a read-only container like so:

```
docker run --rm --mount type=tmpfs,destination=/state -e STATE_PATH=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0-SNAPSHOT
```

Just want to confirm that, even with the changes in this PR, the `-mount type=tmpfs,destination=/state -e STATE_PATH=/state` arguments will _also_ be needed?  I don't think there's any getting around them but just want to confirm that so we can then document this requirement along with the caveat you mentioned in this PR about not being able to use `elastic-agent` sub-commands inside the read-only container.

cc: @kilfoyle

blakerouse:
Moved back to draft, as I adjust this to write the `container-paths.yml` into the `STATE_PATH`.

blakerouse:
I have updated this PR to write the `container-paths.yml` into the `STATE_PATH`. This provides a much better experience then the implementation this PR previously had, as now all sub-commands once exec into the container still work.

pchila (approved):

ycombinator:
SonarQube, as usual, isn't able to account for coverage from integration tests so that check is falsely failing.  Other checks are passing, so force merging.

Issue #4785, opened 2024-05-21
Add per-platform sample OTel configurations


strawgate:
We should include a folder in the Agent distribution called something like `otel_samples` or similiar that contains per-platform use-case based configurations.

Each config should have a name and there should be a readme which indicates what each config is used for.

During onboarding we will provide installation instructions that copy the chosen use-case-specific config out of the samples folder, replace otel.yml with it, and populate it with keys/endpoints:
```
curl --proto '=https' --tlsv1.2 -fOL https://snapshots.elastic.co/8.14.0-6b2f3648/downloads/beats/elastic-agent/elastic-agent-8.14.0-SNAPSHOT-darwin-x86_64.tar.gz
tar -xvf elastic-agent-8.13.4-darwin-x86_64.tar.gz
rm elastic-agent-8.13.4-darwin-x86_64/otel.yml && cp elastic-agent-8.13.4-darwin-x86_64/otel_templates/logs_hostmetrics.yaml elastic-agent-8.13.4-darwin-x86_64/otel.yml
sed -i '' 's/<<ES_ENDPOINT>>/'$elasticsearch_url'/g' elastic-agent-8.13.4-darwin-x86_64/otel.yml && sed -i '' 's/<<ES_API_KEY>>/'$api_key'/g' elastic-agent-8.13.4-darwin-x86_64/otel.yml

elastic-agent-8.13.4-darwin-x86_64/elastic-agent otel
```

Here are two examples of macOS sample configurations. I have not tested these and we should assume these will need to be updated and that we will need to provide platform specific working configurations.

**Logs and Host Metrics**
```
receivers:

  # Receiver for platform specific log files
  filelog/platformlogs:
    include: [ /var/log/*.log ]
#    start_at: beginning

  # Receiver for CPU, Disk, Memory, and Filesystem metrics
  hostmetrics/system:
    collection_interval: 30s
    scrapers:
      disk:
      filesystem:
      cpu:
      memory:

processors:
  elasticinframetrics:

exporters:

  elasticsearch/bulk:
    endpoints: [<<ES_ENDPOINT>>]
    api_key: <<ES_API_KEY>>
    logs_index: logs-generic-default
    metrics_index: metrics-generic-default

service:
  pipelines:
    metrics/hostmetrics:
      receivers: [hostmetrics/system]
      processors: [elasticinframetrics]
      exporters: [elasticsearch/bulk]

    logs/platformlogs:
      receivers: [filelog/platformlogs]
      processors: []
[...]

PR #5002, opened 2024-06-25
[otel] Add basic configuration samples

michalpristas:
~~Written in a way `validate` passes on them~~ now using placeholders

Closes https://github.com/elastic/elastic-agent/issues/4785


strawgate:
afaik these will need to follow the format provided in the ticket with the specified placeholders present

michalpristas:
will we replace placeholders by any means?

strawgate:
indeed, the onboarding flow will replace the placeholders

strawgate:
I updated the issue https://github.com/elastic/elastic-agent/issues/4785 with updated configs that should work with the infra processor, just waiting on if the index names need to change

michalpristas:
updated configs

strawgate:
we may remove the logs_index and metrics_index parameters, I guess there's some discussion about having the index pulled from the datastream values, will follow-up

flash1293:
@strawgate In the snippet for the onboarding flow, we try to get the `otel_samples/patformlogs_hostmetrics.yml` for both linux and mac - on this PR it looks like it only exists for mac though? What's the right file for linux in the onboarding flow? Is it `logs_hostmetrics.yml`? Not sure why the difference

flash1293:
@michalpristas Here is the PR for the onboarding flow: https://github.com/elastic/kibana/pull/183732

The snippet for Linux looks like this:
```
arch=$(if ([[ $(arch) == "arm" || $(arch) == "aarch64" ]]); then echo "arm64"; else echo $(arch); fi)

curl --output elastic-distro-8.15.0-SNAPSHOT-linux-$arch.tar.gz --url https://snapshots.elastic.co/8.15.0-af44b0a5/downloads/beats/elastic-agent/elastic-agent-8.15.0-SNAPSHOT-linux-$arch.tar.gz --proto '=https' --tlsv1.2 -fOL && mkdir elastic-distro-8.15.0-SNAPSHOT-linux-$arch && tar -xvf elastic-distro-8.15.0-SNAPSHOT-linux-$arch.tar.gz -C "elastic-distro-8.15.0-SNAPSHOT-linux-$arch" --strip-components=1 && cd elastic-distro-8.15.0-SNAPSHOT-linux-$arch 
        
rm ./otel.yml && cp ./otel_samples/platformlogs_hostmetrics.yml ./otel.yml && sed -i 's#<<ES_ENDPOINT>>#https://myinstance.es.eu-west-1.aws.qa.elastic.cloud:443#g' ./otel.yml && sed -i 's/<<ES_API_KEY>>/base64encodedkey==/g' ./otel.yml
```

it should prepare everything to collect and ship host logs and metrics

internal/pkg/otel/samples/darwin/hostmetrics_resources.yml
```diff
@@ -0,0 +1,34 @@
...
+    detectors: ["system"]
+    system:
+      hostname_sources: ["os"]
+
+exporters:
+  # Exporter to print the first 5 logs/metrics and then every 1000th
+  debug:
+    verbosity: detailed
+    sampling_initial: 5
+    sampling_thereafter: 1000
```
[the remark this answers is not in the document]
strawgate: I agree and removed it from my samples in the orig ticket

internal/pkg/otel/samples/darwin/hostmetrics_resources.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  # Exporter to send logs and metrics to Elasticsearch
+  elasticsearch/bulk:
```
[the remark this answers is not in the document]
strawgate: I think we were thinking we'd have two exporters in this release but we can remove it

internal/pkg/otel/samples/darwin/hostmetrics_resources.yml
```diff
@@ -0,0 +1,34 @@
...
+    endpoints: [<<ES_ENDPOINT>>]
+    api_key: <<ES_API_KEY>>
```
andrzej-stencel, line 27: Not sure if this is a good idea, but maybe use environment variables here? :thinking: 

```suggestion
    endpoints: [${env:ELASTIC_ENDPOINT}]
    api_key: ${env:ELASTIC_API_KEY}
```

internal/pkg/otel/samples/darwin/logs_hostmetrics.yml
```diff
@@ -0,0 +1,36 @@
+receivers:
```
andrzej-stencel, line 1: This is the same file as the above `logs_host_metrics.yml`, right? :slightly_smiling_face:

internal/pkg/otel/samples/darwin/logs_metrics_traces.yml
```diff
@@ -0,0 +1,63 @@
+receivers:
```
andrzej-stencel, line 1: This file has everything that the above `logs_host_metrics.yml` has, and then some more. Is it valuable to have both?

internal/pkg/otel/samples/darwin/platformlogs_hostmetrics.yml
```diff
@@ -0,0 +1,37 @@
+receivers:
```
andrzej-stencel, line 1: Another version of the same file as `logs_host_metrics.yml`? :thinking:

andrzej-stencel:
I'm probably missing some context here, the content in the files is a bit repetitive, is there a reason for this?

internal/pkg/otel/samples/darwin/logs_metrics_traces.yml
strawgate: adding traces requires adding a listening port so i think it makes sense for it to be optional

strawgate:
> I'm probably missing some context here, the content in the files is a bit repetitive, is there a reason for this?

There should be 2 configs for each platform:
1. Logs and Infrastructure-only Host Metrics
2. Logs, Host Metrics, and Traces to APM

I need to double check if we need to include the resource detection processor in these

internal/pkg/otel/samples/darwin/logs_hostmetrics.yml
michalpristas: yeah i wanted to do resource enrichment but i will take it out

internal/pkg/otel/samples/darwin/hostmetrics_resources.yml
michalpristas: @flash1293 can you alter your script to replace variable like this so we have at least something usefull in our configs?

internal/pkg/otel/samples/darwin/platformlogs_hostmetrics.yml
michalpristas: removed the first one

AlexanderWert:
> I need to double check if we need to include the resource detection processor in these

Yes, we need the resource detection processor to populate the resource attributes properly.

flash1293:
@michalpristas In the latest state, these samples now use env variables for ES host and API key: https://github.com/elastic/elastic-agent/pull/5002/files#diff-f10f8ed3393ab509de41e1eaea7fc6bbced0a461041b3fd79ad0eb53165ae030R25

However, the [quickstart snippet](https://github.com/elastic/elastic-agent/pull/5002#issuecomment-2194555483) in the Kibana UI is looking for placeholders: `<<ES_ENDPOINT>>` and `<<ES_API_KEY>>` which are replaced with the actual values.

How should we bring them together?
Options:
* Switch it back to `<<ES_ENDPOINT>>` and `<<ES_API_KEY>>`
* Change the snippet to replace `${env:ELASTIC_ENDPOINT}` and `${env:ELASTIC_API_KEY}` with the actual values instead
* Change the snippet to set the env vars via `export` instead of using `sed` to search/replace in the file

I would prefer the first option, the second one seems like a weird middle ground and the third one makes the host and api key a little too ephemeral to me - as they are not persisted anywhere, it would be very easy for the user to lose them by closing the terminal session. Writing them to the config file seems like the right call.

Any thoughts?

michalpristas:
my issue with first option is that it provides terrible experience for user, while env is standard when it comes to otel config, you can validate and use config right away, with weird placeholders we force users to do something that is not standard way of handling configuration. 

providing them with sed may work for linux/mac users, but this won't work on windows

i'd be for 2nd or 3rd option and as this is implementation detail on your side, i don't really have a preference. my goal is to have usable config packed with agent

flash1293:
@michalpristas fair points, let's go with the second option for now. I will adjust the snippet

flash1293:
@michalpristas On testing with `https://snapshots.elastic.co/8.15.0-af44b0a5/downloads/beats/elastic-agent`, I'm getting 
> failed to build pipelines: failed to create "elasticsearch" exporter for data type "metrics": telemetry type is not supported

is this expected at the moment?

AlexanderWert:
> is this expected at the moment?

Yes, because the other PR with metrics support is not included yet

internal/pkg/otel/samples/windows/hostmetrics.yml
```diff
@@ -0,0 +1,28 @@
+receivers:
```
AlexanderWert, line 1: Since we haven't included the `hostmetricsreceiver` for Windows, how about we remove the `windows` directory alltogether?

internal/pkg/otel/samples/darwin/logs_metrics_traces.yml
```diff
@@ -0,0 +1,96 @@
...
+    endpoint: "${env:APM_ENDPOINT}"
+    headers:
+    # Elastic APM Server secret token or API key
+      Authorization: "Bearer ${env:APM_SECRET_KEY}"
+
+  elasticsearch:
+    endpoints: ["${env:ELASTIC_ENDPOINT}"]
+    api_key: ${env:ELASTIC_API_KEY}
+    mapping:
+      mode: ecs
```
carsonip, line 69: To enable data stream routing when https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33794 is merged, `*_dynamic_index` needs to be enabled
```suggestion
  elasticsearch:
    endpoints: ["${env:ELASTIC_ENDPOINT}"]
    api_key: ${env:ELASTIC_API_KEY}
    mapping:
      mode: ecs
    logs_dynamic_index:
      enabled: true
    metrics_dynamic_index:
      enabled: true
    traces_dynamic_index:
      enabled: true
```

carsonip: the same for all other ymls

strawgate:
> providing them with sed may work for linux/mac users, but this won't work on windows

`sed` won't work, but this is a trivial transformation on Windows with powershell

> it provides terrible experience for user, while env is standard when it comes to otel config

Can you help me understand why this is a terrible experience for users? I understand it is a less fun experience for developers, but a user is only going to be impacted by this once, as they only go through onboarding once. A user who wants to use environment variables can easily do so?

I also believe it depends on the use-case. I believe that using environment variables for credentials is standard for application environments, but is not standard (would be considered insecure), for endpoint systems.

michalpristas:
i agree on the second point, but I believe these configs are not meant for production environments.
as I understand them is this scenario:
i'm normal dev asked or want to try collection done by elastic, i set up local ES and Kibana and here's agent tar package. they say they have logs, metrics and traces and they provided me even a config file, great.
what do i need to do to have perfect experience? ideally `./elastic-agent otel --config=logs_metrics_traces.yml` 
if i don't have a default then modifying the file or providing ENV, these are still standard for trying out. 

I havent' seen onboarding flow Joe is working on, but copy/paste script generated by that I find fine. 
what i would like to have still is effortless experience for that basic scenario I described above, conscious of our out of the box deployment defaults, just like we had with any other product.

so my ideal behavior would require [this](https://github.com/open-telemetry/opentelemetry-collector/issues/5228) to be implemented upstream

michalpristas:
I understood that @flash1293 implemented replace of env, which is great. I enable auto merge and as soon as somebody from my team hit approve, this gets in.

ycombinator:
I'm +1 to the environment variables approach as it provides a standard way to pass information to the config for replacement while also not precluding sed-style replacements if someone chooses to do that instead.

ycombinator (approved):

PR #5030, opened 2024-07-02
Updated config files, removed redundant ones

michalpristas:


internal/pkg/otel/samples/darwin/logs_metrics_traces.yml
```diff
@@ -53,6 +53,11 @@ processors:
           enabled: true
         os.type:
           enabled: true
+  attributes/dataset:
+          actions:
```
ChrsMark, line 57: Is this indention correct? It seems to have too many spaces :)

AlexanderWert:
@michalpristas We also need to extend the configuration for the hostmetrics receiver scrapers:

````
          cpu:
            metrics:
              system.cpu.utilization:
                enabled: true
              system.cpu.logical.count:
                enabled: true
          memory:
            metrics:
              system.memory.utilization:
                enabled: true
          process:
            mute_process_exe_error: true
            mute_process_io_error: true
            mute_process_user_error: true
            metrics:
              process.threads:
                enabled: true
              process.open_file_descriptors:
                enabled: true
              process.memory.utilization:
                enabled: true
              process.disk.operations:
                enabled: true
          network:
          processes:
          load:
          disk:
          filesystem:
````

For Mac, we can remove the `cpu` and `disk` scrapers as those are not supported on Mac

**Comment by AlexanderWert** [comment]
@michalpristas Let's also add the file offset tracking config for filelog receiver:

````
    extensions:
      file_storage:
        directory: /var/lib/otelcol
    receivers:
      filelog:
        retry_on_failure:
          enabled: true
        start_at: end
        storage: file_storage
````

With that we should have all the things uncovered in testing regarding the config samples.

**Comment by ChrsMark** [comment]
@AlexanderWert if `/var/lib/otelcol` is accessible by the collector it should be fine. Most probably that'd require `sudo`. On k8s the Collector runs as `user: 0`.

**Comment by michalpristas** [comment]
can you comment a suggestion with changes needed?

**Review by ChrsMark — COMMENTED** [review]
And the `receiver` definition should be:

```yaml
receivers:
  # Receiver for platform specific log files
  filelog/platformlogs:
    include: [ /var/log/*.log ]
    storage: file_storage
```

In addition, the onboarding should run `mkdir -p ./data/otelcol && sed -i 's#\${env:STORAGE_DIR}#'"$PWD"/data/otelcol'#g' ./otel.yml` .

michalpristas:
@flash1293 was `STORAGE_DIR` merged into onboarding snippet, is this safe to merge?

andrzej-stencel (approved):

Issue #5033, opened 2024-07-02
High memory and CPU consumption when fleet-server fails to start during enroll


AndersonQ:
For confirmed bugs, please report:
#### Steps to Reproduce:
 - Follow the steps outlined in https://github.com/elastic/elastic-agent/issues/5032
 - monitor the memory and CPU consumption of the running elastic-agent processes
   - if running in a dedicated VM, jut run `top` and sort by memory usage
 - after a while you will observe:
   - an error from fleet-server
   - an increase on memory and CPU usage
   - after a while it'll print a plethora of logs like the following will be printed.
`{"log.level":"info","@timestamp":"2024-07-02T13:18:54.078Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":812},"message":"Waiting for Elastic Agent to start: rpc error: code = Canceled desc = context canceled","ecs.version":"1.6.0"}
`

PR #5034, opened 2024-07-02
Fix indefinite memory and CPU consumption when waiting fleet to be ready

AndersonQ:
## What does this PR do?

Fixes the wait for Fleet Server to be ready 

## Why is it important?

When waiting for Fleet Server to start, the Elastic Agent does not account for the timeout when waiting for Fleet Server to be ready.

Currently, when the timeout is reached, the operation isn't interrupted and the goroutine waiting for Fleet Server to be ready gets stuck in an infinite loop without any delay between iterations. It continually prints a log like:

`{"log.level":"info","@timestamp":"2024-07-02T13:18:59.354Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":812},"message":"Waiting for Elastic Agent to start: rpc error: code = Canceled desc = context canceled","ecs.version":"1.6.0"}
`.

This causes a spike in memory and CPU consumption until the agent is killed by the OS, potentially jeopardising the normal operation of the host.

## Checklist

Try to reproduce https://github.com/elastic/elastic-agent/issues/5033, the issue should not be reproducible with this fix.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5033

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1719934992-Fix-indefinite-memory-and-CPU-consumption-when-waiting-fleet-to-be-ready.yaml
+++ b/changelog/fragments/1719934992-Fix-indefinite-memory-and-CPU-consumption-when-waiting-fleet-to-be-ready.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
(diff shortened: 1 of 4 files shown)


blakerouse (approved):
Looks good and the test covers the fix. Nice!

PR #5035, opened 2024-07-02
Add ability to escape variable expansion in policy

### LINKED ISSUE #2177: No way to escape variables in transpiler parsing
There is no easy way to provide a raw `${var}` to a computed unit configuration. At the moment the transpiler will determine that to be a variable and try to replace it, which could be something that the unit configuration would not want.

We should support the usage of `$${var}`, where the transpiler just ignores it and removes the initial `$` resulting in the unit configuration get the constant value of `${var}`.

At the moment the workaround is to use the pipe `|` to work around this and provide a default.

`${vars|'$vars'}`

This is not a great user experience and also has an issue where `vars` must also not exist or it will do the wrong thing.

- Version: 8.6.0
- Operating System: All

blakerouse:
## What does this PR do?

Adds the ability to a variable not be expanded and replaced in inputs.

A variable that is defined as `$${...}` will be replaced with `${...}` in the policy.

`$${...}` is not allowed in any `condition` statement, because a condition statement requires that strings always be wrapped in `'...'` and so it doesn't make sense to allow it to be escaped and converted into a real string. That is already supported by just doing `'${...}'`.

## Why is it important?

Allows users to provide values that should not be replaced in the policy.

## Checklist

- ~~[ ] I have added an integration test or an E2E test~~(unit test coverage is really good)

## Disruptive User Impact

Adds the ability of users to escape variables.

## Related issues

- Closes #2177


blakerouse:
Seems SonarQube is mad about the generated code from antlr, not from the code added.

pierrehilbert:
I'm okay with force merging this if we are getting the approval from code reviewer, @ycombinator is probably sharing my thoughts here

internal/pkg/eql/visitor.go
```diff
@@ -250,6 +250,13 @@ func (v *expVisitor) VisitExpNot(ctx *parser.ExpNotContext) interface{} {
 	return !val
 }
 
+func (v *expVisitor) VisitExpEVariable(ctx *parser.ExpEVariableContext) interface{} {
+	// escaped variables are not allowed in Eql conditions
+	// they should be wrapped in a constant to be allowed as a valid string in the expression
+	v.err = errors.New("escaped variable $${ is not allowed")
```
michel-laterman, line 256: (nitpick) Would this error message be clear in the logs that this is not allowed in a condition?

michel-laterman (approved):
lgtm

internal/pkg/eql/visitor.go
blakerouse: Yes the condition checker adds that context.

ycombinator:
Force merging as this PR contains a lot of generated code and SonarQube is not able to exclude it.

PR #5044, opened 2024-07-02
Inject `HTTPS_PROXY/HTTP_PROXY` env var into endpoint elasticsearch output units as proxy_url 

### LINKED ISSUE #2602: Forward `HTTP_PROXY/HTTPS_PROXY/NO_PROXY` to components.
**Describe the enhancement:**

Currently there are 3 different proxies that can be configured for the Elastic Agent:
 - [proxy to communicate with ES](https://www.elastic.co/guide/en/fleet/8.7/fleet-agent-proxy-support.html#_set_the_proxy_for_communicating_with_elasticsearch)
 - [proxy to communicate with Fleet Server](https://www.elastic.co/guide/en/fleet/8.7/fleet-agent-proxy-support.html#cli-proxy-settings)
 - [a default proxy](https://www.elastic.co/guide/en/fleet/8.7/fleet-agent-proxy-support.html#host-proxy-env-vars).

The default proxy is set though the environment variables `HTTP_PROXY/HTTPS_PROXY/NO_PROXY`. This is actually a Go feature, we don't handle them explicitly, but any other proxy will override it.

The proxy setting doe thorough `HTTP_PROXY/HTTPS_PROXY/NO_PROXY` are inherited by the components the Elastic Agent runs as sub process, that means the beats "receive" those settings when they're set for the Elastic Agent, but other components don't. What is at least an inconsistent behaviour.

A user using the environment variables to set up a global proxy and having problems because some of the Elastic Agent components, such as Endpoint Security, aren't receiving the proxy.

michel-laterman:
## What does this PR do?

Inject `proxy_url` value into endpoint's elasticsearch output configuration, and enpoint/apm's fleet configuration if the attribute is missing and `HTTPS_PROXY/HTTP_PROXY` env var is set.

The first host value is used to determine if the `HTTPS_PROXY`, or `HTTP_PROXY` value is injected.
If that can't be used to determine then the `HTTPS_PROXY` is preferred.
No Injection occurs if the `proxy_url` key exists, `proxy_disable:  true` is set, or the env vars are empty.

## Why is it important?

Go HTTP clients automatically use various proxy env vars. This leads to inconsistent behaviour where some components get config from the environment, but others do not.
This effects endpoint specifically as it is not a go process.

## Checklist

## Related issues

- Closes #2602


internal/pkg/agent/application/inject_proxy_component_modifier.go
```diff
@@ -0,0 +1,61 @@
...
+// InjectProxyComponentModifier injects a proxy_url value into a component's config if one is not set.
+//
+// This will inject a proxy_url into each output unit if there is no existing value and the agent has a proxy set through an environment variable.
+// The URL used is the HTTPS_PROXY env var. If that's not set the HTTP_PROXY env var is used.
+// If there are no env vars set, or the unit's config has `proxy_disable: true`, nothing is injected
+func InjectProxyComponentModifier() coordinator.ComponentsModifier {
+	proxyURL, ok := os.LookupEnv("HTTPS_PROXY")
+	if !ok {
+		proxyURL, _ = os.LookupEnv("HTTP_PROXY")
+	}
```
blakerouse, line 24: This means that the `HTTPS_PROXY` will take over the `HTTP_PROXY` even in the case that `HTTPS_PROXY` is set to `""`. That might not be the behavior wanted, might be better to use:

```suggestion
	proxyURL := os.Getenv("HTTPS_PROXY")
	if proxyURL == "" {
		proxyURL = os.LookupEnv("HTTP_PROXY")
	}
```

internal/pkg/agent/application/inject_proxy_component_modifier.go
```diff
@@ -0,0 +1,61 @@
...
+	return func(comps []component.Component, cfg map[string]interface{}) ([]component.Component, error) {
+		if proxyURL == "" {
+			return comps, nil
+		}
+		for i, comp := range comps {
+			if comp.InputSpec != nil {
+				for j, unit := range comp.Units {
+					if unit.Type == client.UnitTypeOutput {
```
blakerouse, line 32: This is going to apply to all output types not just elasticsearch. Based on the logic here I believe this should be elasticsearch specific? Would we have a logstash one?

internal/pkg/agent/application/inject_proxy_component_modifier.go
```diff
@@ -0,0 +1,61 @@
...
+						unitCfgMap := unit.Config.Source.AsMap()
+						// Check if proxy_url is part of config
+						if _, ok := unitCfgMap["proxy_url"]; ok {
+							continue
+						}
```
blakerouse, line 37: I assume this is the case even when `proxy_url: ""` in the configuration? I think that makes sense, but we should make that clear in the code comment.

blakerouse (changes requested):
Reading the issue this seems like this should also be endpoint specific, so you should really be filter this only to the endpoint component. I don't think we want this to be done across the board, as passing the ENV down to the subprocesses already does the correct thing.

I believe there is already a endpoint specific component modifier, it might be better to integrate this there so multiple modifiers are not needed, each looping through all components and units each time.

See comment amount elasticsearch specific, we should ensure it also works for logstash. Format might be different.

blakerouse (approved):
Thanks for scoping this to endpoint and elasticsearch only. Good test coverage as well, nice!

michel-laterman:
@intxgo @ferullo
Any concerns about this approach?

intxgo:
> @intxgo @ferullo Any concerns about this approach?

The PR is incomplete, Endpoint requires fleet proxy as well. I guess that environment proxy also overrides fleet connection on Agent.

However, I'm concerned about the configuration update flow. Endpoint is caching permanently received config so that even if Agent gets compromised the machine remains protected by Endpoint (using last known configuration).

The configuration update protocol is using state index value to determine if the Expected and Actual config are the same or different, so I guess to have the proxy override applied the index would have to be bumped. Is the index solely managed by Agent or is it a Fleet policy property?

https://github.com/elastic/elastic-agent-client/blob/94f110545198e910527a9db7e1a21f8553d70ea2/elastic-agent-client.proto#L185-L201

michel-laterman:
Thanks @intxgo,
I've added a proxy_url injection into `InjectFleetConfigComponentModifier` to handle that area

intxgo (approved):

Issue #4347, opened 2024-02-29
Remove dependency from snapshot api by removing the lookup for a specific build in upgrade to specific snapshot feature  tests


pchila:
**Describe the enhancement:**

In order to reduce flakiness of elastic agent integration tests we need to avoid querying the snapshot API for a list of builds for a given snapshot version during an integration test run.
We need to remove only the testing part to get the list of builds as the URL `https://snapshots.elastic.co/...` should still be reliable (that's independent from the API)

Check if this is already implemented by #4331

**Describe a specific use case for the enhancement or feature:**
TODO
**What is the definition of done?**
TODO

/cc @rdner @cmacknz

PR #5050, opened 2024-07-03
Remove use of artifacts-api.elastic.co and its client

rdner:
This includes removal of
TestStandaloneDowngradeToSpecificSnapshotBuild integration test.

Discussed with @pchila and @cmacknz 

1. Independent agent patch releases will be ready very soon.
2. We are likely to handle downgrades in a different way than “upgrade to lower version” in the actual product.
3. This test fails our integration tests sue to instability of `artifacts-api.elastic.co` and that's the only code that is still using this API.

Closes https://github.com/elastic/elastic-agent/issues/4347


pkg/testing/tools/artifacts_api.go
pchila, line 1: I think that parts of this are used in magefile in relation of parsing build manifests (mostly datamodel)
For example https://github.com/elastic/elastic-agent/blob/db40ac25757dc35cc61f92a4fea3e8864a97d1cb/magefile.go#L486

rdner: Okay, I'll move structures then.

rdner: https://github.com/elastic/elastic-agent/pull/5050/commits/2a2ba2ea59c3da8ca57017af37e8a09cc4bc244c

rdner:
@cmacknz are we still doing this?

cmacknz (approved):
I'm fine with it, this leaves some infrequently and not critical path code without a test, but eliminates the most frequent test failure.

PR #5087, opened 2024-07-09
[otel] docs: Add links to components in OTel README

andrzej-stencel:
## What does this PR do?

Adds links to OTel components in the [OTel README](https://github.com/elastic/elastic-agent/blob/v8.14.2/internal/pkg/otel/README.md).

## Why is it important?

Knowing the name of the component is not enough for users. They need to know where to find more information.

## Checklist


michalpristas (approved):
it looks good, i was wondering if we should do some kind of validation to check remote part exists, but this could be problematic in CI in case if there are traffic filters.

ycombinator:
> it looks good, i was wondering if we should do some kind of validation to check remote part exists, but this could be problematic in CI in case if there are traffic filters.

++ to adding such validation but if it's not trivial to implement, let's do it in a follow up PR (please file an issue for it in that case). If this validation starts to become flaky in CI, we can remove it or make it more resilient.  But IMO the benefit of having it and detecting broken links is greater than the _potential_ risk of flakiness.

PR #5091, opened 2024-07-09
Use new ARM Google Cloud instances for integration testing

rdner:
We're trying this new instance type out as a private preview.


rdner:
`{'message': '[pd-ssd, c4a-standard-4] features are not compatible for creating instance.', 'domain': 'global', 'reason': 'badRequest'}`

So, persistent disks are not supported for the new instance, we need to make more changes.

rdner:
@adam-stokes I think we don't attach `pd-ssd` in our integration framework, looks like it's done in OGC https://github.com/adam-stokes/ogc/blob/f4abea238be539c3c4e94b439c8d54fc22cbbcf7/ogc/provision.py#L450

We need a condition in OGC that does not attach `pd-ssd` for `c4a` instances. Or any other idea is welcome.

blakerouse:
@rdner We need a disk to boot from, what version of disk does the instance type support?

rdner:
@blakerouse I checked and only [Google Cloud Hyperdisk](https://cloud.google.com/compute/docs/disks/hyperdisks) is currently supported. I'm pretty sure `Hyperdisk Balanced` would be enough for this purpose. There are some [limitations](https://cloud.google.com/compute/docs/disks/hyperdisks#limitations) though.

andrzej-stencel (approved):

adam-stokes (approved):

rdner:
Not relevant anymore since we switched to buildkite agents.

PR #5096, opened 2024-07-09
TestUpgradeBrokenPackageVersion: Detect version bump and conditionally skip test

pchila:
## What does this PR do?

This PR checks if there has been a version bump along with an elastic-agent version pinning to the previous version.
If that is the case, skip `TestUpgradeBrokenPackageVersion` since removing the `package.version` file as it's done in that test will mess up the paths elastic-agent uses.

## Why is it important?

To prevent build failures during version bumps

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/upgrade_broken_package_test.go
+++ b/testing/integration/upgrade_broken_package_test.go
@@ -12,6 +12,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -76,7 +77,7 @@ func skipTestIfVersionIsBumped(ctx context.Context, t *testing.T, startFixture *
 
 	packageVersionBytes, err := os.ReadFile(filepath.Join(workDir, packageVersionFiles[0]))
 	require.NoError(t, err, "error reading package.version file from startFixture in %q", workDir)
-	agentPackageVersion := string(packageVersionBytes)
+	agentPackageVersion := strings.TrimSpace(string(packageVersionBytes))
 	if agtversion.GetDefaultVersion() != agentPackageVersion {
 		t.Skipf(
 			"Package version %q and default version %q differ: this means that we are probably running a bumped version of agent pinned as previous, skipping...",
--- a/testing/integration/upgrade_broken_package_test.go
+++ b/testing/integration/upgrade_broken_package_test.go
@@ -32,8 +32,6 @@ func TestUpgradeBrokenPackageVersion(t *testing.T) {
 		Sudo:  true,  // requires Agent installation
 	})
 
-	t.Skip("Skipping test until 8.16.0 bump is complete and agent version pinning is removed")
-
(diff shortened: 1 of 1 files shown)


pierrehilbert (approved):
Thanks @pchila !

blakerouse (approved):

pchila:
Example of skipping the test because package version and default version are different: https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1263#01909ca3-7329-49ff-b71e-009dd07d53fc

image


Remains to be tested on main once we remove the pinning to 8.15.0 but I guess we could (either/or):
- wait for the bump to be finished and rebase (the test should run at that point)
- merge now and go back to check once the agent version pinning is removed

Any preference?

ycombinator:
> merge now and go back to check once the agent version pinning is removed
This is my preference as it would let us test the conditional logic in both directions:
- now, before pinning is removed => test should not run
- after pinning is removed => test should run

cmacknz:
Why does this test fail during the version bump? The agent was built as 8.15.0, and we fallback to being 8.16.0, and this confuses the test or the agent somehow?

I'm not opposed to skipping it if that is all we can do but I want to understand if we can just fix the problem causing the failure so our test framework doesn't temporarily let potential problems onto main.

Are we still changing the agent package version to the previous minor? If that is the root of this, we can probably eliminate that. If I recall we needed this because we failed the Fleet Server version compatibility check, for stateful this was relaxed for fleet-server snapshot versions some time ago. https://github.com/elastic/fleet-server/pull/3039

pchila:
> Why does this test fail during the version bump? The agent was built as 8.15.0, and we fallback to being 8.16.0, and this confuses the test or the agent somehow?
> 

The version of the agent is part of some of the agent paths, removing the package.version file after installing as an 8.15 agent and falling back to BEAT_VERSION screws up where the agent thinks it's installed.

> I'm not opposed to skipping it if that is all we can do but I want to understand if we can just fix the problem causing the failure so our test framework doesn't temporarily let potential problems onto main.
> 
> Are we still changing the agent package version to the previous minor? If that is the root of this, we can probably eliminate that. If I recall we needed this because we failed the Fleet Server version compatibility check, for stateful this was relaxed for fleet-server snapshot versions some time ago. [elastic/fleet-server#3039](https://github.com/elastic/fleet-server/pull/3039)

The initial reason for the pinning was Fleet not accepting an agent newer than itself, if we don't need that we can avoid pinning agent version to the previous version and that in turn would let this test run...

cmacknz:
> The initial reason for the pinning was Fleet not accepting an agent newer than itself, if we don't need that we can avoid pinning agent version to the previous version and that in turn would let this test run...

Yes this restriction is already removed for ESS agents, and we can change Fleet to do the same for serverless. So I vote we remove the version pinning of agent.

pchila:
PR is in standby as we see what happens when removing the agent version pinning directly in https://github.com/elastic/elastic-agent/pull/5119

pchila:
Closed in favor of #5119 5119

PR #5097, opened 2024-07-09
Add unit tests for `$$` appearing within policy inputs

michel-laterman:
## What does this PR do?

Add unit tests for special characters, mainly the escape sequence of `$$` in the transpiler, and input parsing of policies.

## Why is it important?

Secrets are being incorrectly formatted when they contain these characters.
This also applies to any `key: value` pair the user specifies in the fleet-ui.

Testing found that this is caused by the UI escaping `$$` to `$`; but we should have unit tests in the agent to ensure we don't inadvertently introduce a regression.

## Related issues

- Relates https://github.com/elastic/kibana/issues/188377


pchila (approved):

pierrehilbert:
Skipping the changelog check as this PR is only adding Unit Tests

Issue #5007, opened 2024-06-26
Development agent gets unhealthy on adding Elastic Defend. when not added to the primary agent.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.15.0 SNAPSHOT
BUILD: 75674
COMMIT: c47f4833230ac0c14af5c4669d68ab41ceac4dd6
```

**Preconditions:**
1. 8.15.0-SNAPSHOT Kibana cloud environment should be available.

**Steps to reproduce:**
1. Install a primary agent with Agent policy 1.
2. Install development agent with Agent policy 2.
3. Now add Elastic Defend to Agent policy 2.
4. Observe agent gets unhealthy on adding Elastic Defend to development agent.

**Expected Result:**
Development agent  should remain healthy on adding Elastic Defend. when not added to the primary agent.

**Screenshot:**
![image](https://github.com/elastic/elastic-agent/assets/77374876/c4846181-bc15-4d2b-ad70-90d0a280269d)

**Agent Logs:**
[elastic-agent-diagnostics-2024-06-26T09-29-38Z-00.zip](https://github.com/user-attachments/files/15985941/elastic-agent-diagnostics-2024-06-26T09-29-38Z-00.zip)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary Review for this ticket is Done

amolnater-qasource:
JFI @pierrehilbert

pierrehilbert:
As discussed earlier today I think this is an expected behavior (as Endpoint probably requires specific path to be able to work correctly) but @cmacknz will be able to confirm.

cmacknz:
You should not be able to run Defend twice. You should be able to run Defend in the development agent when it is not in the primary agent.

blakerouse:
@cmacknz I don't believe that is true, because the development Elastic Agent will be installed into a slightly different path. Endpoint requires that Elastic Agent be installed at a specific location any deviation from that path will cause it not to work.

jlind23:
Hence I believe it validates https://github.com/elastic/elastic-agent/issues/5007#issuecomment-2191375776 and this is indeed an expected behaviour. @pierrehilbert @cmacknz do you think we should document this somewhere?

cmacknz:
I will document this in the readme section for the development agent. Internal uses will have the primary agent running elastic agent all the time, so it wouldn't work because of that anyway.

PR #5098, opened 2024-07-09
Clarify --develop installations cannot run Defend.

cmacknz:
Clarifies the behavior reported in https://github.com/elastic/elastic-agent/issues/5007. Namespaced installations cannot run Defend because Defend expects the agent to be in the default path, similar to the restriction on `--base-path`.

- Closes https://github.com/elastic/elastic-agent/issues/5007

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,10 @@ sudo elastic-development-agent status
 sudo -u elastic-agent-user elastic-development-agent status
 ```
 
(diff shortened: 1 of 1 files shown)


pierrehilbert (approved):

jlind23:
Force merging this as SonarQube analysis will never be triggered.

PR #5119, opened 2024-07-11
Remove Agent override version in integration tests

pchila:
We used to override the agent version to a previous version between the version bump for a new release and the availability of the first snapshot build (usually a period of a few days) because Fleet had a limitation about not accepting agent versions more recent than itself.

This limitation should have been removed by now so we can get rid of the agent version override and use the default agent version

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


jlind23 (approved):

jlind23:
Buildkite test it

.buildkite/scripts/steps/integration_tests.sh
```diff
@@ -8,22 +8,13 @@ MAGE_TARGET="${2:-"integration:test"}"
 MAGE_SUBTARGET="${3:-""}"
 
 
-# Override the agent package version using a string with format <major>.<minor>.<patch>
-# There is a time when the snapshot is not built yet, so we cannot use the latest version automatically
-# This file is managed by an automation (mage integration:UpdateAgentPackageVersion) that check if the snapshot is ready.
-OVERRIDE_AGENT_PACKAGE_VERSION="$(cat .package-version)"
```
pierrehilbert, line 14: I don't think we should stop doing that!
We need https://github.com/elastic/elastic-agent/pull/5100 to be merged instead.

pierrehilbert (changes requested):
We shouldn't remove all those parts according to me.

pchila:
@pierrehilbert This PR came after a discussion in the elastic-agent control plane weekly meeting about #5096 where we determined that Fleet should be able to handle a version of elastic-agent newer than itself.

If Fleet is able to manage a newer agent we don't need to package elastic-agent with a previous version after the version bump. If we still need to use an older stack version on cloud that should be accomplished by specifying `AGENT_STACK_VERSION` when running the tests... Maybe that is the reason why you say we still need this code ?

/cc @cmacknz @ycombinator

pierrehilbert:
Yes indeed my objective here is to ensure we can have our tests green even if the new stack version is not available on our cloud.
But if you have another idea, I'm ofc opened to it.

pchila:
@pierrehilbert  what you is probably what is in commit 83a3a830df4dcea100c13b33f3fea06003ba5daf
Please have another look

pierrehilbert (approved):
Thanks for switching AGENT_VERSION to AGENT_STACK_VERSION as this is what we will need in the future.

andrzej-stencel (approved):

PR #5130, opened 2024-07-12
refactor: drop golang.org/x/lint

kruskall:
## What does this PR do?

drop golang.org/x/lint dependency

## Why is it important?

the dependency was used to install the golint tool. The golint tool is frozen and deprecatde according to the upstream repository: https://github.com/golang/lint

> NOTE: Golint is https://github.com/golang/go/issues/38968. There's no drop-in replacement for it, but tools such as [Staticcheck](https://staticcheck.io/) and go vet should be used instead.

CI and automation are using golangci-lint and enable staticcheck which is mentioned as a valid replacement.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -771,8 +771,8 @@ github.com/elastic/go-windows v1.0.1 h1:AlYZOldA+UJ0/2nBuqWdo90GFCgG9xuyw9SYzGUt
 github.com/elastic/go-windows v1.0.1/go.mod h1:FoVvqWSun28vaDQPbj2Elfc0JahhPB7WQEGa3c814Ss=
 github.com/elastic/gosigar v0.14.3 h1:xwkKwPia+hSfg9GqrCUKYdId102m9qTJIIr7egmK/uo=
 github.com/elastic/gosigar v0.14.3/go.mod h1:iXRIGg2tLnu7LBdpqzyQfGDEidKCfWcCMS0WKyPWoMs=
-github.com/elastic/mock-es v0.0.0-20240605193845-b5546a703d6f h1:qo0Nn9G+p7HbG+CmiWwSoiKJ/cyRq59TRn9Z2PRHTi8=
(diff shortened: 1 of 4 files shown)


go.sum
```diff
@@ -1739,7 +1739,6 @@ golang.org/x/lint v0.0.0-20191125180803-fdd1cda4f05f/go.mod h1:5qLYkcX4OjUUV8bRu
 golang.org/x/lint v0.0.0-20200130185559-910be7a94367/go.mod h1:3xt1FjdF8hUf6vQPIChWIBhFzV8gjjsPE/fR3IyQdNY=
 golang.org/x/lint v0.0.0-20200302205851-738671d3881b/go.mod h1:3xt1FjdF8hUf6vQPIChWIBhFzV8gjjsPE/fR3IyQdNY=
 golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5/go.mod h1:3xt1FjdF8hUf6vQPIChWIBhFzV8gjjsPE/fR3IyQdNY=
-golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 h1:VLliZ0d+/avPrXXH+OakdXhpJuEoBZuwh1m2j7U6Iug=
 golang.org/x/lint v0.0.0-20210508222113-6edffad5e616/go.mod h1:3xt1FjdF8hUf6vQPIChWIBhFzV8gjjsPE/fR3IyQdNY=
```
pchila, line 1742: I was expecting for all the others `golang.org/x/lint` entries to disappear as well as it doesn't seem like we have  an indirect dependency according to go.mod

pchila (commented):
Just a small comment/question on the leftover `golang/x/lint` entries in go.sum, looks good otherwise

go.sum
ycombinator: ++, to elaborate a bit, that's because there are still indirect dependencies on all these various versions of `golang.org/x/lint`.  You can see them if you run `go mod graph | grep ' golang.org/x/lint'`.  [This article](https://www.ardanlabs.com/blog/2019/12/modules-03-minimal-version-selection.html), particularly Listing 8, explains why older versions of a dependency might still hang around in `go.sum`.

pchila (approved):

PR #5132, opened 2024-07-12
Require a valid version of github.com/openshift/api

mauri870:
## What does this PR do?

Added a replace directive to pin the library to a known commit that points to the now retracted v3.9.0 release.

## Why is it important?

If you run go mod tidy on main it fails with:

	go: downloading github.com/openshift/api v3.9.0+incompatible
	unknown revision v3.9.0

The go.sum points to github.com/openshift/api@v3.9.0 but this version does not exist, it was retracted in https://github.com/openshift/api/pull/1515.


mauri870:
> Thank you for the contribution! 🎉
> 
> What do you think about setting the version in the go.mod entry instead of using replace ?

I tend to avoid using replace as it makes the main binary package [non go-installable](https://go.dev/issue/44840). Unfortunately using require does not work as this is as indirect dependency pulled from [github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8sclusterreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/4d618025f8839a15d1681b692fd082313d13f24d/receiver/k8sclusterreceiver/go.mod). Interesting that they specify a replace as well but that does not seem to take effect for some reason, perhaps they need to use require with the commit from 2018 instead of the v3.9.0 + replace?

mauri870:
Yeah, this whole situation is quite messed up. I still think they should just use the commit equivalent for v3.9.0 and call it a day. It’s not worth messing around with replace to untangle this situation, which is clearly not working well.

cmacknz:
Not opposed to the change, but with Go 1.22.4 I can run `go mod tidy` without issue. Our CI system is doing the same and not failing. What am I missing?

michalpristas (approved):

michalpristas:
@cmacknz  with proxy you're protected of this behavior as well as unintended force pushes to labels.

go.mod
```diff
@@ -335,6 +335,7 @@ replace (
 	github.com/Shopify/sarama => github.com/elastic/sarama v1.19.1-0.20220310193331-ebc2b0d8eef3
 	github.com/dop251/goja => github.com/andrewkroh/goja v0.0.0-20190128172624-dd2ac4456e20
 	github.com/dop251/goja_nodejs => github.com/dop251/goja_nodejs v0.0.0-20171011081505-adff31b136e6
+	github.com/openshift/api v3.9.0+incompatible => github.com/openshift/api v0.0.0-20180801171038-322a19404e37
```
cmacknz, line 338: Let's add a comment for why we've done this replacement. 

This dependency comes in from opentelemetry-collector-contrib, so you could just link back to them and use the same comment since we should likely just follow the same version they use. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/12d41f40b0d408b0167633d8095160d3343d46ac/go.mod#L38

mauri870: Thanks, I added a comment and link for reference.

cmacknz:
buildkite test it

ycombinator:
Looks like CI is failing on the `NOTICE.txt` being up-to-date check:

```
a/NOTICE.txt | 0s
-- | --
  | b/NOTICE.txt | 0s
  | @@ -38403,11 +38403,11 @@ Contents of probable licence file $GOMODCACHE/github.com/opencontainers/image-sp
  |  
  | --------------------------------------------------------------------------------
  | Dependency : github.com/openshift/api
  | -Version: v3.9.0+incompatible
  | +Version: v0.0.0-20180801171038-322a19404e37
  | Licence type (autodetected): Apache-2.0
  | --------------------------------------------------------------------------------
  |  
  | -Contents of probable licence file $GOMODCACHE/github.com/openshift/api@v3.9.0+incompatible/LICENSE:
  | +Contents of probable licence file $GOMODCACHE/github.com/openshift/api@v0.0.0-20180801171038-322a19404e37/LICENSE:
  |  
  | Apache License
  | Version 2.0, January 2004
  | NOTICE.txt: needs update
  | make[1]: *** [Makefile:60: check-no-changes] Error 1
  | make[1]: Leaving directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1721767272486308266/elastic/elastic-agent'
  | make: *** [Makefile:41: check-ci] Error 2
  | 🚨 Error: The command exited with status 2
```

@mauri870 Please run `make notice` in the `elastic-agent` root folder and commit the changes.  Thanks!

mauri870:
> Please run `make notice` in the `elastic-agent` root folder and commit the changes. Thanks!

Done, thank you!

cmacknz:
buildkite test it

PR #5136, opened 2024-07-16
Fixing the hints autodiscovery configuration path in Elastic Agent Manifests

gizas:
- Bug

## What does this PR do?

WHAT: Fixes the path that is being used in hints autodiscovery of elastic-agent
WHY: The path for hint configuration was not correct when users where extracting the manifest from Kibana page

Note: We have added `hints.default_container_logs: true` also in the standalone configmap

## Why is it important?

Improves the UX for elastic agent users that want to use kibana to create their manifests

## Checklist

1. Download this nranch
2. Navigate to `deploy/kubernetes`
3. Run: 
```bash
WITHOUTCONFIG=true make generate-k8s
 ./creator_k8s_manifest.sh .
``
4. Check the new `elastic_agent_manifest.ts` that is being created should include `elastic-agent-standalone/templates.d` line
```yaml
      #        curl -sL https://github.com/elastic/elastic-agent/archive/8.16.tar.gz | tar xz -C /usr/share/elastic-agent/state/inputs.d --strip=5 "elastic-agent-8.16/deploy/kubernetes/elastic-agent-standalone/templates.d"

```

### DIFF
--- a/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
+++ b/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
@@ -28,7 +28,7 @@ data:
     providers.kubernetes:
       node: ${NODE_NAME}
       scope: node
-      #Uncomment to enable hints' support https://www.elastic.co/guide/en/fleet/current/hints-annotations-autodiscovery.html
+      #Uncomment to enable hints' support - https://www.elastic.co/guide/en/fleet/current/hints-annotations-autodiscovery.html
       #hints.enabled: true
       #hints.default_container_logs: true
     inputs:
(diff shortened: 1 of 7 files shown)


deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
```diff
@@ -28,8 +28,9 @@ data:
     providers.kubernetes:
       node: ${NODE_NAME}
       scope: node
-      #Uncomment to enable hints' support
+      #Uncomment to enable hints' support - https://www.elastic.co/guide/en/fleet/current/hints-annotations-autodiscovery.html
       #hints.enabled: true
+      #hints.default_container_logs: true
```
tetianakravchenko, line 33: is it required to set it explicitly? default value is `true` https://github.com/elastic/elastic-agent/blob/v8.14.3/internal/pkg/composable/providers/kubernetes/config.go#L69

gizas: Indeed the default is true. Not needed to be set.
It is something like the kubernetes provider that is enabled by default.

I want to avoid putting `hints.default_container_logs: false`, users accidentally uncomment and break default fucntionality

tetianakravchenko (approved):

PR #5145, opened 2024-07-16
[CI] Automatic retries for integration tests

pazone:
## What does this PR do?
Each failed integration test step will be retried once

## Why is it important?
Sometimes the tests are flaky. It's not the best practice but anyway we retry them manually 

## Checklist

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/3568

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -16,7 +16,7 @@ fi
 DOCKER_REGISTRY="docker.elastic.co"
 DOCKER_REGISTRY_SECRET_PATH="kv/ci-shared/platform-ingest/docker_registry_prod"
 CI_DRA_ROLE_PATH="kv/ci-shared/release/dra-role"
-CI_GCP_OBS_PATH="kv/ci-shared/platform-ingest/gcp-platform-ingest-ci-service-account"
+CI_GCP_OBS_PATH="kv/ci-shared/observability-ingest/cloud/gcp"
 CI_ESS_PATH="kv/ci-shared/platform-ingest/platform-ingest-ec-prod"
 CI_DRA_ROLE_PATH="kv/ci-shared/release/dra-role"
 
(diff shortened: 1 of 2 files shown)


michalpristas (approved):
limit 1 means, 1 failure will be retried right?

pazone:
> limit 1 means, 1 failure will be retried right?

It means it will be retried only once

jlind23:
Looks like this retry was meant to be only for OGC task according to the issue right?

rdner:
@jlind23 If I understand the diff correctly, now Buildkite retries every integration test step (Serverless integration test, Extended runtime leak tests, Integration tests, Serverless Beats Tests) if it fails, regardless the cause. Correct me if I'm wrong.

jlind23:
No you are right, I don't think this was the intent behind https://github.com/elastic/ingest-dev/issues/3568 according to the description wrote by @dli

dliappis:
> No you are right, I don't think this was the intent behind [elastic/ingest-dev#3568](https://github.com/elastic/ingest-dev/issues/3568) according to the description wrote by @dliappis

yes the intent was to retry OGC tasks only, as work around to out of capacity issues in an AZ e.g. https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1323#0190a43f-e423-470c-b73e-82a428a451dd/567-638
A well tested approach outside of BK (implemented e.g. in the corresponding tool used by elasticsearch performance) is that the tool itself (ogc) rotates across eligible AZs in the region, when there is an out of capacity issue (in addition to retries for a specific AZ).

pazone:
@dliappis @rdner If the BK retries are that bad idea I can revert it and reopen the initial issue

rdner:
reverted by https://github.com/elastic/elastic-agent/pull/5170

PR #5174, opened 2024-07-18
Update e2e-testing dependency

ycombinator:
This PR bumps up the `github.com/elastic/e2e-testing` dependency version.


cmacknz (approved):
We can probably get rid of this dependency by copying the parts we use, I think it's just file downloads.

ycombinator:
> We can probably get rid of this dependency by copying the parts we use, I think it's just file downloads.

Indeed. That was the discussion we had in Slack as well.  This PR is just the immediate fix.

michel-laterman (approved):

PR #5175, opened 2024-07-19
[Ci] Fix DRA artifacts permissions

pazone:
## What does this PR do?
Adds group read file permissions to artifacts.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/pipeline.elastic-agent-package.yml
+++ b/.buildkite/pipeline.elastic-agent-package.yml
@@ -112,12 +112,15 @@ steps:
     command: |
       echo "+++ Restoring Artifacts"
       echo "UMASK: $(umask)"
-      umask 0022
+      # umask 0022
 
       buildkite-agent artifact download "build/**/*" . --build 0190c8dd-5fa9-4894-ab74-61c817b930ab
       echo "+++ Changing permissions for the release manager"
 
-      sudo chown -R 1000:1000 build/distributions/
+      ls -lahR build/
+      sudo chmod -R g+r build/distributions/ 
+      sudo chown -R :1000 build/distributions/
+
       ls -lahR build/
       echo "+++ Running DRA publish step"
       if [[ -z "${MAGEFILE_VERBOSE}" ]]; then
--- a/.buildkite/pipeline.elastic-agent-package.yml
+++ b/.buildkite/pipeline.elastic-agent-package.yml
@@ -53,57 +53,57 @@ steps:
   - wait: ~
     if: build.env("MANIFEST_URL") == null
 
-  # - group: ":Packaging Artefacts"
-  #   key: "package"
-  #   steps:
-  #     - label: ":package: Cross Building and package elastic-agent"
-  #       key: package_elastic-agent
-  #       agents:
-  #         provider: "gcp"
-  #         machineType: "c2-standard-16"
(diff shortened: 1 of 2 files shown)


pazone:
go installation script is broken

pchila (approved):

.buildkite/pipeline.elastic-agent-package.yml
```diff
@@ -63,14 +61,14 @@ steps:
           machineType: "c2-standard-16"
           diskSizeGb: 400
         command: |
-          if [[ -z "${MANIFEST_URL}" ]]; then
+          if [[ -z "$${MANIFEST_URL}" ]]; then
```
rowlandgeoff, line 64: I'm confused. The switch to `$$` would output the PID, right? So everywhere this is used, the string being checked will be non-null.

.buildkite/scripts/common.sh
```diff
@@ -6,12 +6,11 @@ if [[ -z "${WORKSPACE-""}" ]]; then
...
     export WORKSPACE
 fi
-PIPELINE="${WORKSPACE}/.buildkite/pipeline.elastic-agent-package.yml"
 if [[ -z "${SETUP_MAGE_VERSION-""}" ]]; then
-    SETUP_MAGE_VERSION=$(grep -oe "SETUP_MAGE_VERSION\: [\"'].*[\"']" "$PIPELINE" | awk '{print $2}' | sed "s/'//g" )
+    SETUP_MAGE_VERSION="1.14.0"
 fi
 if [[ -z "${SETUP_GVM_VERSION-""}" ]]; then
-    SETUP_GVM_VERSION=$(grep -oe "SETUP_GVM_VERSION\: [\"'].*[\"']" "$PIPELINE" | awk '{print $2}' | sed "s/'//g" )
+    SETUP_GVM_VERSION="v0.5.0" # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
```
rowlandgeoff, line 13: How are these version changes related to the permissions problem?

.buildkite/pipeline.elastic-agent-package.yml
pazone: before any pipeline execution builkite agent uploads the pipeline yaml with `buildkite-agent pipeline upload`. It interpolates the yaml; if it finds `$somethig` it considers it a variable and tries to resolve it during the upload. `${MANIFEST_URL}` will be resolved to `""` (empty string). The `$$` is used to avoid variable substitution. 

* why did it work before?

rowlandgeoff: Ah, thanks for explaining. I guess I haven't noticed this because most pipelines have the command logic in separate .sh files.

.buildkite/scripts/common.sh
pazone: It's indirectly related to the issue. Go installation started to fail. It tried to extract the `SETUP_GVM_VERSION` from pipeline.elastic-agent-package.yml and it stopped working after my changes. I just decided to simplify it because it's actually a very simple thing that was done in an overly complex way

rowlandgeoff: Thanks 👍

PR #5201, opened 2024-07-26
Remove verbose debug log line.

cmacknz:
What the summary says, this log line at the debug level can be quite large, let's remove it.


michalpristas (approved):
Looks good.

PR #5208, opened 2024-07-27
[tests][integration] fix monitoring test cases

VihasMakwana:
- Bug

## What does this PR do?

- With introduction of status reporter for `metricbeat`,  [integration test cases](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1617#0190e937-4819-42b4-8175-b345052dd836) were failing 
- From diagnostics, it was observed that the status transitioned from `HEALTHY` to `DEGRADED` (momentarily) and again to `HEALTHY`.
- Previous behaviour:
    - During the above transition, `runner.agentFixture.ExecStatus(ctx)` returned an error because of the momentarily `DEGRADED` state and it caused entire test to fail (even though it returned to a `HEALTHY` state, its just how go tests function).
- This PR does the following:
    - Remove the check on error inside `require.Eventually`. 
       - If all the components return to a healthy state, the test would pass eventually.
       - If it remains in a unhealthy state, the test would fail.

## Checklist

closes https://github.com/elastic/elastic-agent/issues/5209


blakerouse (commented):
Maybe it would be better to change - https://github.com/elastic/elastic-agent/blob/2676a3f372b3aeca96c266fcb18d633a6c416496/pkg/testing/fixture.go#L717 instead of the actual test.

Reading the docstring on `ExecStatus` it says it can return an output even on exit of 1. Maybe we should only return an error when we are unable to read the output, otherwise as long as some output is returned it should not be an error.

That would probably fix more tests and not just these specific tests.

cmacknz:
Agree with Blake, the `ExecStatus` should return `nil` if we were able to get the output.

It feels like it should probably return three values: the command output, a boolean for whether  agent is healthy, and an error that is only not-nil if we couldn't get or parse the output of the command. 

https://github.com/elastic/elastic-agent/blob/03b5d31eeda94b2f3f7ab83bded45c03738fb5a7/internal/pkg/agent/cmd/status.go#L86-L92

If we were to keep it the way it is, the actual bug is that there is a `require.NoError` inside of a `require.Eventually` loop to immediately fail the test. The `require.Error` in the case below should have been `if err != nil { return false }` to let the eventually retry. 

https://github.com/elastic/elastic-agent/blob/03b5d31eeda94b2f3f7ab83bded45c03738fb5a7/testing/integration/agent_long_running_leak_test.go#L219

VihasMakwana:
@blakerouse @cmacknz Agree with you guys. Thanks for your inputs.

i've updated the `ExecStatus` and we return `nil` if we can parse the output.

pkg/testing/fixture.go
```diff
@@ -715,20 +715,14 @@ func (e *ExecErr) Unwrap() error {
 // determine if the returned AgentStatusOutput is empty or not.
 // It should work with any 8.6+ agent
```
blakerouse, line 716: update the docstring has this change affects the behavior

pkg/testing/fixture.go
```diff
@@ -715,20 +715,14 @@ func (e *ExecErr) Unwrap() error {
...
 func (f *Fixture) ExecStatus(ctx context.Context, opts ...process.CmdOption) (AgentStatusOutput, error) {
-	out, err := f.Exec(ctx, []string{"status", "--output", "json"}, opts...)
+	out, _ := f.Exec(ctx, []string{"status", "--output", "json"}, opts...)
+
 	status := AgentStatusOutput{}
 	if uerr := json.Unmarshal(out, &status); uerr != nil {
```
blakerouse, line 721: after parse you should verify that there is at least a `status.Message` set, if its empty then it parsed an empty JSON and that should be an error.

VihasMakwana: okay.
I think we can reuse `IsZero()` https://github.com/elastic/elastic-agent/blob/03b5d31eeda94b2f3f7ab83bded45c03738fb5a7/pkg/testing/fixture.go#L1245-L1247

blakerouse: Yes that should work.

pkg/testing/fixture.go
VihasMakwana: done

pkg/testing/fixture.go
VihasMakwana: PTAL

pkg/testing/fixture.go
```diff
@@ -709,26 +709,21 @@ func (e *ExecErr) Unwrap() error {
...
 		return AgentStatusOutput{},
-			fmt.Errorf("could not unmarshal agent status output: %w",
-				errors.Join(&ExecErr{
-					err:    err,
-					Output: out,
-				}, uerr))
+			fmt.Errorf("could not unmarshal agent status output: %w", uerr)
+	} else if status.IsZero() {
+		return status, fmt.Errorf("agent status output is empty")
```
cmacknz, line 724: I would keep the error and log it here in case it isn't empty, for example file not found or some permissions issue on the CI runner.

VihasMakwana: I didn't get you. Can you please elaborate?

cmacknz: You are throwing away the `error` return value from `out, _ := f.Exec(ctx, []string{"status", "--output", "json"}, opts...)`. The logic you have is correct to ignore it for deciding whether the command succeeded, but if the command failed and status is empty we should still wrap the original error in case it contains helpful context for why the command failed.

I want this to be `return status, fmt.Errorf("agent status output is empty: %w", err)` instead.

VihasMakwana: thanks craig.

pkg/testing/fixture.go
```diff
@@ -709,26 +709,20 @@ func (e *ExecErr) Unwrap() error {
...
 	status := AgentStatusOutput{}
 	if uerr := json.Unmarshal(out, &status); uerr != nil {
 		return AgentStatusOutput{},
-			fmt.Errorf("could not unmarshal agent status output: %w",
-				errors.Join(&ExecErr{
-					err:    err,
-					Output: out,
-				}, uerr))
+			fmt.Errorf("could not unmarshal agent status output: %w", uerr)
+	} else if status.IsZero() {
```
cmacknz, line 722: This check should be first otherwise `json.Unmarshal` is likely to fail with `unexpected end of json input` on an empty string as input and we'll never return the `err` from command.

blakerouse: @cmacknz This cannot be first, otherwise IsZero will always be true.

blakerouse (approved):
Looks good, thanks for changing this around.

VihasMakwana:
@blakerouse thanks for approving. So you know, I made a small change to combine `uerr` and `err`(just in case if it's not empty) moments after your approval 😅 
https://github.com/elastic/elastic-agent/blob/9e77ba4f48301578ea6771813a55fe99a4634dc9/pkg/testing/fixture.go#L719-L722

pkg/testing/fixture.go
VihasMakwana: @cmacknz this cannot be first. Although, I've combined `err` and `uerr`, if you can have a look.

cmacknz (approved):

blakerouse (approved):

pierrehilbert:
Looks like we now have failing tests that may be related to your change.

VihasMakwana:
@blakerouse @cmacknz  from the logs of failing tests, I can see that output is valid and **not** empty but `status.IsZero()` returns `false` because `Info.ID` field is not set somehow
Screenshot 2024-07-30 at 4 36 45 PM.

I think we should check also check for `Version`, `Message`, and `ID`.
So, something like:
```go
func IsZero() bool {
     return status.Message == "" && status.Info.Version == "" && status.Info.ID == ""
}
```

pkg/testing/fixture.go
```diff
@@ -1243,7 +1237,7 @@ type AgentStatusOutput struct {
 }
 
 func (aso *AgentStatusOutput) IsZero() bool {
-	return aso.Info.ID == ""
+	return aso.Info.ID == "" && aso.Message == "" && aso.Info.Version == ""
```
blakerouse, line 1240: This makes more sense actually. As `IsZero` should be if nothing is outputted at all.

blakerouse (approved):
Looks good. Thanks for the fixes.

Issue #5246, opened 2024-08-02
Docker image running Elastic Agent in Otel mode


michalpristas:
Currently agent start in normal mode by default.
Only way how to change that is overriding command that's executed.
This approach is not always viable.

For this reason we should create and additional image where agent starts in otel mode.

@cmacknz while we're at it it would be nice to consider making this image slim and get rid of all the agent boilerplate that is not needed when running otel (as this will be explicitly named `elastic-agent-otel`). 
What i have in mind is symlink to binary and all beats/endpoint...

strawgate:
> it it would be nice to consider making this image slim

Ideally we should keep all components in the image for now and minimize the differences between this and the existing image until we have a better idea of how long this will be around for and what the use-cases will be.

> For this reason we should create and additional image where agent starts in otel mode.

I wish I knew more about the build process but I would have thought that we could have just built from the existing image:

```
FROM .../elastic-agent:current_tag...

CMD ["--config", "otel.yml"]
```

And saved ourselves from adding a whole new image pipeline?

cmacknz:
It might actually be simpler if we just implement this as an environment variable like `ELASTIC_AGENT_OTEL=true` to save us the work of creating a new container.

Thinking more, one day we probably want to obsolete the elastic-agent-otel container because it will be the same as elastic-agent so we can just save ourselves the trouble.

blakerouse:
@rogercoll This is not in `8.15.1-SNAPSHOT`. It will only be present on `8.16.0-SNAPSHOT`, once a new snapshot is built. Possible that one has not been built yet.

rogercoll:
I can confirm that `8.16.0-SNAPSHOT` has been build and fixes the issue. Thanks team!

herrBez:
If someone needs to make it work with version 8.15.x, please note that we need to overwrite the command and also the entrypoint. Sharing a sample `docker-compose.yml` in the hope it can be helpful:

```yaml
services:
  elastic-agent:
    image: docker.elastic.co/beats/elastic-agent:8.15.1 
    container_name: elastic-agent # instead of 
    restart: no
    entrypoint: "elastic-agent"
    command: ["otel", "-c", "/etc/elastic-agent/otel.yaml"]
    volumes:
      - type: bind
        source: ./otel.yaml
        target: /etc/elastic-agent/otel.yaml
```

PR #5248, opened 2024-08-02
Docker image running Elastic Agent in Otel mode

michalpristas:
Adding new type of docker image for amd64 and arm64
Image is based on wolfi and contains agent binary without components resulting in 277MB
Using separate otel entrypoint and Dockerfile templates for simplification

fixes: #5246


blakerouse (commented):
I like the idea of a second dockerfile, as I know the other is complicated. On the other hand I do worry about the issues this brings of having to make updates in two places (like pcap adjustments not being consistent, or adjustments to deps, etc.).

My question is how much is this really saving in complexity in the file? It looks similar but without the synthetics stuff with is already conditional.

michalpristas:
@blakerouse if we agree on keeping components (still a question), i will merge this into a single docker file, 
with components out it seemed more clean to have separate dockerfile instead of nested conditional statements.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -265,3 +265,7 @@ RUN echo -e '#!/bin/sh\nexec /usr/local/bin/docker-entrypoint' > /app/apm.sh &&
 {{- else }}
 ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/docker-entrypoint"]
 {{- end }}
+
+{{- if contains .image_name "-otel" }}
```
michalpristas, line 269: starting with a default config

dev-tools/packaging/templates/docker/docker-entrypoint.elastic-agent-otel.tmpl
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Update umask to retain group write permissions on runtime directories: $root/tmp/default, $root/logs/default and $root/run/default
+umask 0007
+
+set -eo pipefail
+
+exec {{ .BeatName }} otel "$@"
```
michalpristas, line 8: kept entrypoint separate to avoid starting conditionally based on ENV vars

michalpristas:
failures on k8s integ tests. related: https://github.com/elastic/elastic-agent/pull/5248

strawgate:
Do we want to discuss the inclusion of components and the use of a separate dockerfile in the PR or in the issue? I have an open item in the issue

cmacknz:
Yeah, thinking more I think our life might be easier if we just support this with an environment variable. https://github.com/elastic/elastic-agent/issues/5246#issuecomment-2278070504

An elastic-agent-otel container is easier for users, but adding an env var that is equivalent to running `elastic-agent otel` in the container isn't much more work than changing the image name.

michalpristas:
Updated entrypoint to execute proper command based on env var. 
Hopefully we can unblock observability team soon

strawgate:
Did we confirm that the operator allows customizing the env vars?

michalpristas:
yes this is something i initially offered and they were happy with env vars

cmacknz:
This should probably have a basic smoke test. Probably something can be added using the tests in https://github.com/elastic/elastic-agent/blob/main/testing/integration/kubernetes_agent_standalone_test.go as a base.

blakerouse (approved):
Much simplier, but doesn't reduce the image size. That is something that will come with time, I believe.

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -190,7 +190,124 @@ func TestKubernetesAgentStandalone(t *testing.T) {
...
+		{
+			"run agent in otel mode",
+			[]corev1.EnvVar{
+				{Name: "ELASTIC_AGENT_OTEL", Value: "true"},
+			},
+			false,
+			map[string]bool{
+				"beat/metrics-monitoring": false,
+				"filestream-monitoring":   false,
+				"system/metrics-default":  false,
```
cmacknz, line 251: Is there actually a way you can tell we started the OTel collector?

This test would still pass if we started elastic agent as usual, but disabled monitoring and didn't use the system/metrics integration/

michalpristas: yes but scenarioyou mentioned is not the default config nor any config we test with fo false positive is really small, 
no way to tell agent is in otel besides logs that are with default config printed to stdout.

cmacknz: There isn't an HTTP port or something that returns otel specific data or something?

I don't feel that strongly about this

cmacknz: Improving this is something we could follow up with.

ycombinator:
@amolnater-qasource I have added testing instructions for this PR to the PR description.

amolnater-qasource:
Thank you!! @ycombinator

amolnater-qasource:
Hi Team,
We have tested this PR on Ubuntu 24.

Steps followed:
1. Run: `sudo docker run -e ELASTIC_AGENT_OTEL=true --rm --name agent docker.elastic.co/beats/elastic-agent-complete:8.16.0-SNAPSHOT`
2. On a new instance run: `sudo docker exec agent sh -c 'ps aux'`
3. Observe:  `elastic-agent otel` process is available.
4. Observe no Beats or other sub-processes are available.

**Screenshots:**
image
image

Hence, we are marking this ticket as **QA:Validated**.
Please let us know if we are missing anything here.

cc: @pierrehilbert 

Thanks!!

Issue #5259, opened 2024-08-06
Upgraded 8.15.0 BC6 agent from 7.17.23 falls back to 7.17.23 after sometime.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.15.0 BC6
BUILD: 76360
COMMIT: 8aa0b59da12c996e3048d8875446667ee6e15c7f
```

**Preconditions:**
1. 7.17.23>8.15.0-BC6 Kibana cloud environment should be available.
2. 7.17.23 agent should be installed.

**Steps to reproduce:**
1. Upgrade 7.17.23 kibana to 8.15.0 BC6.
2. Now navigate to Fleet tab.
3. Click Upgrade agent for 7.17.23 and upgrade to 8.15.0 BC6.
4. Observe agent gets upgraded successfully.
5. After sometime observe agent gets fall back to 7.17.23.

**Expected Result:**
Upgraded 8.15.0 BC6 agent from 7.17.23 should't fall back to 7.17.23 after sometime.

**Screenshot:**
![image](https://github.com/user-attachments/assets/97234cf0-2766-42d2-82cc-953dff3c1023)

**Logs:**
[elastic-agent-diagnostics-2024-08-06T10-17-18Z-00.zip](https://github.com/user-attachments/files/16509472/elastic-agent-diagnostics-2024-08-06T10-17-18Z-00.zip)
[elastic-agent-diagnostics-2024-08-06T11-06-50Z-00.zip](https://github.com/user-attachments/files/16509473/elastic-agent-diagnostics-2024-08-06T11-06-50Z-00.zip)

amolnater-qasource:
@karanbirsingh-qasource Please review.

michel-laterman:
I can't recreate this issue if I use 8.14.3->8.15.0, but I can recreate if I use 7.17.23->8.14.3; i'm not sure what the cause is but I think it's a bug within 7.17.23

pierrehilbert:
Do you have the same issue if you are removing the system integration?

pierrehilbert:
I did some tests and it seems that the problem is with 7.17.23 as I can't reproduce with 7.17.22 upgrades.
Not a blocker for 8.15.0 release but definitively something we need to solve for 7.17.24

ycombinator:
Current status, summarized from off-issue (Slack) discussion:
* @blakerouse is implementing a change in `7.17` Elastic Agent to start the new (post-upgrade Agent's) Watcher.  
[...]

PR #5270, opened 2024-08-08
Increase watcher in upgrade tests to timeout after 10m

michel-laterman:
## What does this PR do?

Increase the timeout we wait for the watcher to stop to 10m from 1m.

## Why is it important?

7.17.23 has a regression where it will upgrade an 8.x release then roll back after a few minutes. 
The expected fix is https://github.com/elastic/beats/pull/40452 which will affect 7.17.24.

## Checklist

## Related issues

- Relates #5259


blakerouse:
@michel-laterman Seems the change did cause it to fail, so that is a good sign.

blakerouse:
Seems we should merge https://github.com/elastic/beats/pull/40452 and wait for next snapshot to validate that it only fails on 7.17.23 and no longer 7.17.24-SNAPSHOT.

Need a +1 on https://github.com/elastic/beats/pull/40452 for merge.

michel-laterman:
@blakerouse do we want to revert the changes I made to `upgrader.go` or should we keep them in case there is a regression?

blakerouse (approved):
@michel-laterman Being that the watcher always reads the configuration to lower the timeout, I think its best to keep your change of the timeouts in the PR. This now means it will only slow the tests down in the case there is a regression and we would rather have that then miss a regression.

PR #5284, opened 2024-08-12
switch to monotonic clocks for component check-in

### LINKED ISSUE #5277: Use of wall clock in check-in calculations
In several of the check-in calculations we use `time.Now().UTC()` to do the comparison.  This is a problem, because as stated in https://pkg.go.dev/time

"Because t.In, t.Local, and t.UTC are used for their effect on the interpretation of the wall time, they also strip any monotonic clock reading from their results."

This can be seen in

https://cs.opensource.google/go/go/+/refs/tags/go1.22.6:src/time/time.go;l=1149
and
https://cs.opensource.google/go/go/+/refs/tags/go1.22.6:src/time/time.go;l=216

Place that need updating:
https://github.com/elastic/elastic-agent/blob/a1fd2c9d9a7cc3931f8f1d452a730b6d216f8bd3/pkg/component/runtime/command.go#L198

https://github.com/elastic/elastic-agent/blob/a1fd2c9d9a7cc3931f8f1d452a730b6d216f8bd3/pkg/component/runtime/command.go#L225

https://github.com/elastic/elastic-agent/blob/a1fd2c9d9a7cc3931f8f1d452a730b6d216f8bd3/pkg/component/runtime/service.go#L481

https://github.com/elastic/elastic-agent/blob/a1fd2c9d9a7cc3931f8f1d452a730b6d216f8bd3/pkg/component/runtime/service.go#L508

leehinman:
## What does this PR do?

Switch to montonic clocks for component check-in calculation.
Previously the wall clock was used and this caused components to
appear degraded if they "missed" check-ins because the time change
made it appear they hadn't checked in within the necessary time
window.

## Why is it important?

Necessary to prevent unnecessary restarting of components when the
wall time of the host changes significantly.

## Checklist

1. Install `elastic-agent` on a machine
2. continuously advance the clock
3. without patch observe that components go into degraded state due to failed check-in, and eventually are restarted
4. with patch observer that components stay healthy and are not restarted.  No missed check-ins are reported

## Related issues

- Closes #5277

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/component/runtime/command.go
+++ b/pkg/component/runtime/command.go
@@ -195,7 +195,7 @@ func (c *commandRuntime) Run(ctx context.Context, comm Communicator) error {
 				// first check-in
 				sendExpected = true
 			}
-			c.lastCheckin = time.Now().UTC()
+			c.lastCheckin = time.Now()
 			if c.state.syncCheckin(checkin) {
 				changed = true
 			}
@@ -222,7 +222,7 @@ func (c *commandRuntime) Run(ctx context.Context, comm Communicator) error {
 					}
 				} else {
 					// running and should be running
-					now := time.Now().UTC()
+					now := time.Now()
 					if now.Sub(c.lastCheckin) <= checkinPeriod {
 						c.missedCheckins = 0
 					} else {
--- a/pkg/component/runtime/command.go
+++ b/pkg/component/runtime/command.go
@@ -195,6 +195,11 @@ func (c *commandRuntime) Run(ctx context.Context, comm Communicator) error {
 				// first check-in
 				sendExpected = true
 			}
+			// Warning lastCheckin must contain a
+			// monotonic clock.  Functions like Local(),
+			// UTC(), Round(), AddDate(), etc. remove the
+			// monotonic clock.  See
+			// https://pkg.go.dev/time
 			c.lastCheckin = time.Now()
 			if c.state.syncCheckin(checkin) {
 				changed = true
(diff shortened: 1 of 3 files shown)


pkg/component/runtime/command.go
```diff
@@ -195,7 +195,7 @@ func (c *commandRuntime) Run(ctx context.Context, comm Communicator) error {
 				// first check-in
 				sendExpected = true
 			}
-			c.lastCheckin = time.Now().UTC()
+			c.lastCheckin = time.Now()
```
blakerouse, line 198: Interesting, didn't know that the `.UTC()` would prevent a monotonic clock usage in golang.

I think it would be good to add a comment over this line explaining the usage of `time.Now()` and how that ensures monotonic clock. Maybe linking to golang documentation for that would be nice to have.

blakerouse (approved):

pkg/component/runtime/command.go
cmacknz: > Because t.In, t.Local, and t.UTC are used for their effect on the interpretation of the wall time, they also strip any monotonic clock reading from their results. The canonical way to strip a monotonic clock reading is to use t = t.Round(0).

They document this in the package level comment of https://pkg.go.dev/time, but not actually on https://pkg.go.dev/time#Time.UTC

Agree with including this as an explicit comment.

It would also be great to be able to write a test for this, but I can't think of a way that doesn't involve manipulating the clock on the machine running the tests.

leehinman: Added comment.

https://github.com/elastic/elastic-agent/issues/5285 to track creating a static analysis to catch

blakerouse: Thanks for adding the comment.

faec (approved):

cmacknz (approved):
Thanks. I'm fine merging this without a test, considering how hard testing for this specifically is. Using static analysis is a better idea.

mauri870 (approved):

Issue #5275, opened 2024-08-09
Failing k8s integration test: TestKubernetesAgentStandalone/drop_ALL_add_CHOWN,_SETPCAP_capabilities_-_rootless_agent


ycombinator:
Example failure: https://buildkite.com/elastic/elastic-agent/builds/10958#01913488-f3f6-4c89-9b69-a9133824db90

```
=== FAIL: testing/integration TestKubernetesAgentStandalone/drop_ALL_add_CHOWN,_SETPCAP_capabilities_-_rootless_agent (99.69s)
  | kubernetes_agent_standalone_test.go:277: elastic-agent never reported healthy: command terminated with exit code 1
  | kubernetes_agent_standalone_test.go:278: stdout: ┌─ fleet
  | │  └─ status: (STOPPED) Not enrolled into Fleet
  | └─ elastic-agent
  | ├─ status: (DEGRADED) 1 or more components/units in a degraded state
  | └─ system/metrics-default
  | ├─ status: (HEALTHY) Healthy: communicating with pid '81'
  | └─ system/metrics-default-system-metrics
  | └─ status: (DEGRADED) Error fetching data for metricset system.process: Not enough privileges to fetch information: Not enough privileges to fetch information: non fatal error fetching PID some info for 13, metrics are valid, but partial: non-fatal error fetching PID metrics for 13, metrics are valid, but partial: Not enough privileges to fetch information: /io unavailable; if running inside a container, use SYS_PTRACE: error fetching IO metrics: open /proc/13/io: permission denied
  |  
  | kubernetes_agent_standalone_test.go:279: stderr:
  | kubernetes_agent_standalone_test.go:347: Wrote container elastic-agent-standalone of pod elastic-agent-standalone-xmt88 logs to /opt/buildkite-agent/builds/bk-agent-prod-gcp-1723161759733813737/elastic/elastic-agent/build/k8s-logs-v1.29.4-kubernetes-amd64--kubernetes.integration/TestKubernetesAgentStandalone/drop_ALL_add_CHOWN,_SETPCAP_capabilities_-_rootless_agent-elastic-agent-standalone-xmt88-elastic-agent-standalone.log
[...]

PR #5271, opened 2024-08-08
[docker][agentbeat]: add cap_sys_ptrace and cap_dac_override in permitted set

VihasMakwana:
## What does this PR do?

Grant cap_sys_ptrace and cap_dac_override to permitted set.
Required to operate in unprivileged mode.

This PR also updates k8s standalone tests and adds required permissions and disables the long running test case https://github.com/elastic/elastic-agent/issues/5279. 
I'm including all of these changes in the same PR because they are related to a same problem.

## Why is it important?

- Without these capabilities, agent will always remain in DEGRADED state inside the container

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## How to test this PR locally

- `PACKAGES=docker mage package` on this branch
```bash
docker run --entrypoint bash --user root --rm -it docker.elastic.co/beats/elastic-agent:8.16.0-SNAPSHOT
getcap data/elastic-agent-a1fd2c/components/agentbeat
data/elastic-agent-a1fd2c/components/agentbeat = cap_dac_override,cap_setuid,cap_net_raw,cap_sys_ptrace+p
``` 

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/5269
- Relates https://github.com/elastic/elastic-agent/issues/5279

PR #5290, opened 2024-08-13
fix: rootless k8s integration tests

pkoutsovasilis:
## What does this PR do?

This PR addresses the failures of k8s integrations tests:

1. Fix degraded agent status for rootless agent running with elastic-agent uid (1000). The main reason behind this was the setcap on the `agentbeat` binary in conjunction with it getting copied from a "builder" image during the building of the elastic-agent container image. I am not sure why this is the case, but I have observed in the past that copying files with capabilities across different docker layers results in an Effective set with no capabilities of the file at hand, ignoring completely the Ambient set of the parent which is elastic-agent in our case. This quirk got highlighted when metricbeat implemented status reporting. Now fun fact, we weren't bumping into this scenario for rootless agent with random uid:gid as in this case, elastic-agent is chowing everything and when you chown a file to a different user the capabilities are getting reset and hence the Ambient set of elastic-agent was getting applied. The solution is to remove setcap from agentbeat completely as it is redundant, elastic-agent will raise and pass down all capabilities in Bounding set and thus the former doesn't need to specify anything special in terms of capabilities.
2. Add extra required capabilities for rootless agent. The extra capabilities required to read /proc/${pid}/io are `DAC_READ_SEARCH` and `SYS_PTRACE`
3. increase memory limits for elastic-agent container as I observed some failures due to OOM when fixing the above issues
4. increase the timeout to 90 seconds while waiting elastic-agent to report healthy. As of now we get a degraded agent status until kube-state-metrics is properly initialised. Thus I feel like 60 seconds when running three kind clusters on the same machine with parallel integration k8s tests is kinda tight

## Why is it important?

Because it fixes the k8s integration tests

## Checklist

## Disruptive User Impact

N/A

```shell
DEV=true SNAPSHOT=true EXTERNAL=true PLATFORMS=linux/arm64 PACKAGES=docker mage -v package
mage integration:auth
mage integration:kubernetes
```

## Related issues

Closes https://github.com/elastic/elastic-agent/issues/5275

### DIFF
--- a/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
+++ b/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
@@ -56,8 +56,7 @@ RUN true && \
     true
 
 # Keep this after any chown command, chown resets any applied capabilities
-RUN setcap =p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/elastic-agent
-RUN setcap cap_net_raw,cap_setuid+p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/components/agentbeat && \
+RUN setcap =p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/elastic-agent && \
 {{- if .linux_capabilities }}
 # Since the beat is stored at the other end of a symlink we must follow the symlink first
 # For security reasons setcap does not support symlinks. This is smart in the general case
--- a/testing/integration/kubernetes_agent_standalone_test.go
+++ b/testing/integration/kubernetes_agent_standalone_test.go
@@ -30,6 +30,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -159,6 +160,10 @@ func TestKubernetesAgentStandalone(t *testing.T) {
 					// set ImagePullPolicy to "Never" to avoid pulling the image
 					// as the image is already loaded by the kubernetes provisioner
 					container.ImagePullPolicy = "Never"
+
+					container.Resources.Limits = corev1.ResourceList{
+						corev1.ResourceMemory: resource.MustParse("800Mi"),
+					}
 
 					if tc.capabilitiesDrop != nil || tc.capabilitiesAdd != nil || tc.runUser != nil || tc.runGroup != nil {
 						// set security context
--- a/testing/integration/kubernetes_agent_standalone_test.go
+++ b/testing/integration/kubernetes_agent_standalone_test.go
@@ -122,18 +122,18 @@ func TestKubernetesAgentStandalone(t *testing.T) {
 			int64Ptr(1000), // elastic-agent uid
 			nil,
 			[]corev1.Capability{"ALL"},
-			[]corev1.Capability{"CHOWN", "SETPCAP"},
+			[]corev1.Capability{"CHOWN", "SETPCAP", "DAC_READ_SEARCH", "SYS_PTRACE"},
 			true,
-			"https://github.com/elastic/elastic-agent/issues/5275",
+			"",
 		},
 		{
 			"drop ALL add CHOWN, SETPCAP capabilities - rootless agent random uid:gid",
 			int64Ptr(500),
 			int64Ptr(500),
 			[]corev1.Capability{"ALL"},
-			[]corev1.Capability{"CHOWN", "SETPCAP", "DAC_READ_SEARCH"},
+			[]corev1.Capability{"CHOWN", "SETPCAP", "DAC_READ_SEARCH", "SYS_PTRACE"},
 			true,
-			"https://github.com/elastic/elastic-agent/issues/5275",
+			"",
 		},
 	}
 


pkoutsovasilis:
cc @blakerouse @VihasMakwana

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -262,8 +267,8 @@ func deployK8SAgent(t *testing.T, ctx context.Context, client klient.Client, obj
 	command := []string{"elastic-agent", "status"}
 	var stdout, stderr bytes.Buffer
 	var agentHealthyErr error
-	// we will wait maximum 60 seconds for the agent to report healthy
-	for i := 0; i < 60; i++ {
+	// we will wait maximum 90 seconds for the agent to report healthy
+	for i := 0; i < 90; i++ {
```
blakerouse, line 271: 90 seconds still seems tight, how about 120 seconds?

blakerouse (commented):
Overall this change seems good, I just want to get a better understanding before I give it a +1.

Seems like this change removes the need for this PR as well https://github.com/elastic/elastic-agent/pull/5271.

Does this only work because of this PR in agentbeat? Or is that PR in agentbeat also not required? https://github.com/elastic/beats/pull/40466

testing/integration/kubernetes_agent_standalone_test.go
pkoutsovasilis: sure thing, "better safe than sorry" is my policy in this time checks 🥲 pushing

pkoutsovasilis:
> Overall this change seems good, I just want to get a better understanding before I give it a +1.
> 
> Seems like this change removes the need for this PR as well #5271.

Yep I don't think this one is needed and I would even propose to allow Agentbeat to get all capabilities through the Ambient set of elastic-agent

> Does this only work because of this PR in agentbeat? Or is that PR in agentbeat also not required? [elastic/beats#40466](https://github.com/elastic/beats/pull/40466)

I think that this PR is also not required https://github.com/elastic/beats/pull/40466#issuecomment-2286307998

pkoutsovasilis:
@rdner is the setcap removal from agentbeat somehow affecting the wolfi-based image? just checking

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -56,8 +56,7 @@ RUN true && \
     true
 
 # Keep this after any chown command, chown resets any applied capabilities
-RUN setcap =p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/elastic-agent
-RUN setcap cap_net_raw,cap_setuid+p {{ $beatHome }}/data/elastic-agent-{{ commit_short }}/components/agentbeat && \
```
cmacknz, line 60: The `cap_net_raw,cap_setuid+p` was added for Heartbeat which doesn't have any tests in this repository so not obvious if this broke anything CC @emilioalvap @vigneshshanmugam 

To solidify my understanding, with https://github.com/elastic/elastic-agent/pull/4925 when agent is root is has these capabilities and ensures sub-processes also have them so there's no need to set them explicitly anymore?

cmacknz:
Let's revert https://github.com/elastic/beats/pull/40466 and make sure this still passes, although it may be better for our CI runs if this merges first to ensure nothing is transiently broken.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
cmacknz: We just need to avoid a breaking change where these capbilities would now need to manually be set in the k8s security context explicitly when that wasn't necessary before.

cmacknz:
Also, be sure to revert https://github.com/elastic/elastic-agent/pull/5293 in this PR if it merges first.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
pkoutsovasilis: so `cap_net_raw,cap_setuid+p` means that agentbeat nowadays (instead of Heartbeat) will have these two capabilities in its permitted set. The code in Heartbeat tries to do a similar concept of what elastic-agent is doing after this https://github.com/elastic/elastic-agent/pull/4925 which is raise the capabilities from permitted to effective set. Now, in context of k8s, when a container is running under root you don't need to raise anything as the allowed capabilities of the parent process (container runtime) will be passed down without any reset to the entrypoint of the container elastic-agent and from there to any child processes of agent, such as agentbeat. Most of the time you wanna drop capabilities and not raise them under root. 

On the other hand, when a container is running rootless, the Effective set of the entrypoint process will be reset to 0 when you transition from the container runtime to the entrypoint (from uid = 0 you go to uid != 0). This is now the point where elastic-agent after https://github.com/elastic/elastic-agent/pull/4925 will raise again to all allowed capabilities and set the Ambient set (necessary as you are not running as root anymore) to allow its child subprocesses to get them as well. Thus there isn't much point setting any capability to any child process of agent as it will get any one that is allowed. Hopefully with that lengthy answer I helped 😅

blakerouse (approved):
Thank you for the detailed explanation's and for ensuring that the revert of the change for the agentbeat works with this change.

blakerouse:
Created PR for the revert in agentbeat.

@cmacknz See @pkoutsovasilis comment here about testing with that PR reverted - https://github.com/elastic/beats/pull/40466#issuecomment-2286307998

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
pkoutsovasilis: in simpler words, if the capabilities that Heartbeat required are allowed it will still get them without the need to specify anything at the binary level. If these are not allowed then as it would happen originally before this change it won't get them as it had them only in permitted set not in the effective one

rdner:
> @rdner is the setcap removal from agentbeat binary somehow affecting the wolfi-based image? just checking

it should not affect the work on Wolfi but I'll keep this change in mind from now on, thanks!

ycombinator:
The [PR to temporarily disable k8s integration tests around capabilities](https://github.com/elastic/elastic-agent/pull/5293) has now been merged.  Please remember to revert it once this PR here is merged.

pkoutsovasilis:
@cmacknz I have rebased this PR to re-enable the k8s test with the fix in place, after [this one](https://github.com/elastic/elastic-agent/pull/5293) is now merged. next steps? 🙂

cmacknz (approved):

cmacknz:
You are stuck behind some known flaky tests. I don't have a problem force merging past these since they do not even involve the container image.

rdner:
> Fix degraded agent status for rootless agent running with elastic-agent uid (1000). The main reason behind this was the setcap on the agentbeat binary in conjunction with it getting copied from a "builder" image during the building of the elastic-agent container image. I am not sure why this is the case, but I have observed in the past that copying files with capabilities across different docker layers results in an Effective set with no capabilities of the file at hand

@pkoutsovasilis could you elaborate more where this conclusion is coming from?

I've tested in my PR that capabilities were still on the binary after I built the final image with copying from the `builder` layer to the final layer. See the PR description for more information https://github.com/elastic/elastic-agent/pull/5073

Is there a quirk I'm not aware of?

pkoutsovasilis:
> #5073

As I mentioned in my comment @rdner this is an experimental observation. However, when you say you've tested this image, how have you tested this? there were no k8s tests back in the day, elastic-agent wasn't raising any capabilities, and most importantly there was nothing in Agentbeat that required that capabilities you were permitting to run successfully (aka no Heartbeat tests). What do I miss?

rdner:
@pkoutsovasilis is not it enough to test it with `getcap` like I did in my PR?
I'm not trying to say you're wrong, I'm trying to understand how one tests such a thing for the future. I have exactly the same problem to solve in the Heartbeat standalone Docker image.

pkoutsovasilis:
> @pkoutsovasilis is not it enough to test it with `getcap` like I did in my PR?
> I'm not trying to say you're wrong, I'm trying to understand how one tests such a thing for the future. I have exactly the same problem to solve in the Heartbeat standalone Docker image.

so the quirk I have observed is that the capabilities are not passed down to child processes through the parent's Ambient set. Specifically, elastic-agent sets the Ambient set and then spawns an Agentbeat child process, if the latter has been copied from another docker image and had capabilities set, the parent's capabilities of the Ambient set are dropped and not passed to it. 

With that said, in the Heartbeat standalone docker image I think you should set the capabilities at it, as there is no parent process that sets Ambient set in this case. 

Does the above help? 🙂

rdner:
@pkoutsovasilis so it's not about loosing capabilities between Docker layers while the image is built, it's about losing capabilities when spawning a sub-process, right?

Asking because I'm taking the same approach with setting capabilities in the `builder` layer here too https://github.com/elastic/beats/pull/40524

I'm trying to make sure that this will work with standalone Heartbeat.

pkoutsovasilis:
> @pkoutsovasilis so it's not about loosing capabilities between Docker layers while the image is built, it's about losing capabilities when spawning a sub-process, right?

yes no capabilities are lost from the binary, the quirk regards only the Ambient set of the parent getting passed at the child processes.

> Asking because I'm taking the same approach with setting capabilities in the `builder` layer here too https://github.com/elastic/beats/pull/40524
> 
> I'm trying to make sure that this will work with standalone Heartbeat.

yep this should have the setcap at the Heartbeat binary. 

In simple words, the issue here doesn't exist in the standalone Heartbeat image and you should keep the setcap at the Heartbeat binary

PR #5295, opened 2024-08-13
Replace `8.14.2-SNAPSHOT` with `8.14.4-SNAPSHOT` in testing versions list

ycombinator:
## What does this PR do?

This PR removes the expired `8.14.2-SNAPSHOT` from the testing versions list and replaces it with the `8.14.4-SNAPSHOT`.

## Why is it important?

So integration tests don't try to use an expired snapshot and use the latest available one for the same minor series instead.


cmacknz (approved):
Shouldn't this have auto-updated for us? Or is it only updating on main?

ycombinator:
> Shouldn't this have auto-updated for us? Or is it only updating on main?

It should've.  I do see PRs updating the versions file on branches other than main, e.g. https://github.com/elastic/elastic-agent/pull/5068.  Maybe the `8.14.2-SNAPSHOT` just expired and the [daily automation](https://github.com/elastic/elastic-agent/blob/06898a121c15eec442499df25466ef16e6537dcd/.github/workflows/bump-agent-versions.yml#L7C13-L7C24) hasn't had a chance to run yet on the `8.15` branch?

Issue #3847, opened 2023-11-30
Introduce a Helm chart for Kubernetes monitoring with Elastic Agent


joshdover:
### Description

Extracted from an internal discussion document:

We can solve many of the challenges with deploying Agent on Kubernetes through a Helm chart for Elastic Agent. This allows us to improve the current experience by (mapped to priorities above that are not already met and easy):
- Deploying separate cluster-level and node-level Agents
	- This allows us to eliminate the leader-election mechanism and the associated K8s API calls. These API calls currently scale horizontally with the number of Kubernetes nodes in the cluster.
	- This also allows separate resource planning and scaling for cluster-level and node-level Agents, which need different amount of resources depending on the horizontal and vertical scale of the cluster and its nodes, respectively.
- Provide “templates” or “presets” for common integrations
	- This moves us away from our current “one-size-fits-all” k8s manifest where users have to comment/uncomment certain sections based on which integrations they want to use
	- Along with this, we can configure the minimum security privileges needed for each integration’s container and deploy separate containers for integrations that require higher-privileges
	- Examples of presets we would provide:
		- P1  A containers.logs: true option would create a DaemonSet agent for container log collection
		- P1  A deployment.metrics: true option would create a ReplicaSet agent deployed with KSM bundled in the same pod. We’d also have similar options for Services and other workload types, all which would enable inputs on the same ReplicaSet
		- P1  An autodiscovery.metrics: true option would mount the required templates for hints-based autodiscovery
		- P2 A cloudDefend.enabled: true option would add the necessary container privileges and inputs to the DaemonSet agents for Cloud Defend
		- P2  A profiling.enabled: true option would add the necessary container privileges and inputs to the DaemonSet agents for Universal Profiling
		- P2  A autoInstrumentation.enabled: true option would deploy the apm-attacher hook as a Deployment for auto instrumenting application pods with APM Agents.
- P3 Simpler sharding of kube-state-metrics (KSM)
	- For any workload metrics that depend on kube-state-metrics, we could pseudo-automate sharding by adding a simple config like ksm.shards: 2 which would shard the ReplicaSet used for the agent collecting this data, along with KSM container in the same pod.

```[tasklist]
### Tasks
- [ ] https://github.com/elastic/elastic-agent/issues/3848
```

joshdover:
[...]

PR #5331, opened 2024-08-21
[k8s] Add elastic-agent helm-chart

pkoutsovasilis:
## What does this PR do?

This PR introduces a Helm Chart for deploying Elastic-Agent in Kubernetes 

## Why is it important?

By having an Elastic-Agent Helm Chart we get a simplified deployment on k8s

## Checklist

## Disruptive User Impact

N/A

Go to `deploy/helm/elastic-agent/examples` 

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/3610
- Closes https://github.com/elastic/elastic-agent/issues/3847


jlind23:
buildkite test it

ycombinator:
@pkoutsovasilis Just a general comment: we plan to use this Helm chart to create an AWS Marketplace listing for Elastic Agent.  To this end, I believe AWS has some [requirements](https://docs.aws.amazon.com/marketplace/latest/userguide/container-product-policies.html#preparing-eks-addon) that the Helm chart must fulfill.  Just wanted to bring that up in this PR in case it means some things in the chart need to change to meet these requirements.  Thanks!

pkoutsovasilis:
> @pkoutsovasilis Just a general comment: we plan to use this Helm chart to create an AWS Marketplace listing for Elastic Agent. To this end, I believe AWS has some [requirements](https://docs.aws.amazon.com/marketplace/latest/userguide/container-product-policies.html#preparing-eks-addon) that the Helm chart must fulfill. Just wanted to bring that up in this PR in case it means some things in the chart need to change to meet these requirements. Thanks!

Hi @ycombinator 👋 I have read these requirements before opening this PR and I think that all of them are satisfied:
1. it support both rootful and rootless deployments (we need the latter for AWS Marketplace)
2. the Helm chart does **not** utilise anything from Helm that isn't supported by the EKS addon framework, such Helm hooks, `lookup` function, etc.
3. although this is up for configuration but for the AWS Marketplace configuration we can offer deployment that have a Daemonset and thus meet the requirement of `Traceable Deployments or Daemonsets`

I don't see something that it isn't covered but maybe another extra pair of eyes on this PR would be helpful 🙂

swiatekm (commented):
I like all the examples! I was wondering if we could also render the final manifests using the example configurations and keep these in version control. The Otel Helm Charts do something like [this](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/Makefile#L6) and it acts partially as a test, and partially as code review aid, where it lets you visualize how a change to the templates impacts the rendered manifests.

pkoutsovasilis:
> I like all the examples! I was wondering if we could also render the final manifests using the example configurations and keep these in version control. The Otel Helm Charts do something like [this](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/Makefile#L6) and it acts partially as a test, and partially as code review aid, where it lets you visualize how a change to the templates impacts the rendered manifests.

oh that's a great idea @swiatekm , ty for bringing it up! 🙂 I say definitely let's add them; Initially I am gonna push them for each example and maybe afterwards we could investigate how to utilise them in a more meaningful way in CI

deploy/helm/elastic-agent/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: elastic-agent
+description: Elastic-Agent Helm Chart
+kubeVersion: ">= 1.27.0-0"
```
blakerouse, line 4: What is causing this version constraint? A comment on why that is the minimum version in the file would be very nice to have.

deploy/helm/elastic-agent/README.md.gotmpl
```diff
@@ -0,0 +1,58 @@
...
+- `ESSecretAuthAPI`: `elasticsearch` output with the connection details specified in a k8s secret
+- `ESECKRef`: `elasticsearch` output that references by name an Elasticsearch cluster managed by ECK operator
+{{ end }}
+
+{{ define "section.description.OutputObject.PostTable" }}
+Examples of Helm chart arguments to define an output with name `myOutput`:
+- `ESPlainAuth`: `--set outputs.myOutput.url=https://elasticsearch:9200 --set outputs.myOutput.api_key=token`
+- `ESSecretAuthBasic`: `--set outputs.myOutput.type=ESSecretAuthBasic --set outputs.myOutput.secretName=k8s_secret_name` (required keys in the k8s secret are `url`, `username`, `password`)
+- `ESSecretAuthAPI`: `--set outputs.myOutput.type=ESSecretAuthAPI --set outputs.myOutput.secretName=k8s_secret_name` (required keys in the k8s secret are `url`, `api_key`)
+- `ESECKRef`: `--set outputs.myOutput.type=ESECKRef --set outputs.myOutput.name=eck_es_cluster_name`
```
blakerouse, line 40: Does this require the cluster to be in the same namespace? Or can it be in a different namespace? If so how can referencing a cluster in another namespace work?

deploy/helm/elastic-agent/examples/kubernetes-only-logs/agent-kubernetes.yaml
```diff
@@ -0,0 +1,22 @@
...
+  controller_manager:
+    enabled: false
+  containers:
+    logs:
+      enabled: true
+  hints:
+    enabled: false
+
+agent:
+  unprivileged: true
```
blakerouse, line 22: I keep seeing this in the examples. Would it just be better to make `agent.unprivileged: true` as the default in the helm chart? This is net-new so it would be okay if it was the default.

deploy/helm/elastic-agent/values.yaml
```diff
@@ -0,0 +1,455 @@
...
+
+# -- extra [user-defined integrations](https://www.elastic.co/guide/en/fleet/current/elastic-agent-input-configuration.html) to be added to the Elastic Agent
+# An example can be found [here](./examples/nginx-custom-integration/README.md)
+# @section -- 3 - User Extra Integrations
+extraIntegrations: {}
+
+agent:
+  # -- enable elastic-agent
+  # @section -- 3 - Elastic-Agent Configuration
+  enabled: true
```
blakerouse, line 298: `enabled: false` would mean what here?

deploy/helm/elastic-agent/values.yaml
```diff
@@ -0,0 +1,455 @@
...
+  # -- elastic-agent version
+  # @section -- 3 - Elastic-Agent Configuration
+  version: 8.15.0
```
blakerouse, line 301: I assume this needs to get bump every version. We should ensure this is added to the automation for this repo so that it gets bumped when we do new releases.

deploy/helm/elastic-agent/values.yaml
```diff
@@ -0,0 +1,455 @@
...
+  # -- image configuration
+  # @section -- 3 - Elastic-Agent Configuration
+  image:
+    repository: docker.elastic.co/beats/elastic-agent-complete
```
blakerouse, line 305: Why is `-complete` being used? `synthetics`? Seems like it would be better to not use that one unless they are wanting to do `synthetics`.

deploy/helm/elastic-agent/values.yaml
```diff
@@ -0,0 +1,455 @@
...
+    pullPolicy: IfNotPresent
+    tag: ""
+  # -- generate kubernetes manifests or [ECK](https://github.com/elastic/cloud-on-k8s) CRDs
+  # @section -- 3 - Elastic-Agent Configuration
+  engine: k8s # k8s or eck
+  # -- enable unprivileged mode
+  # @section -- 3 - Elastic-Agent Configuration
+  unprivileged: false
```
blakerouse, line 313: `unprivileged: true`? Default?

blakerouse (commented):
Okay. I have a few questions.

I did not review the templates. They are so large and complex that it didn't seem like I should sit there and go through each and every line. I think it would be better to just merge this and start using it and make incremental fixes as issues are identified.

I am good with giving a +1 after and review some of the answers to my questions.

deploy/helm/elastic-agent/Chart.yaml
pkoutsovasilis: tbh nothing in particular, back at the time I was writing the Helm chart this was the older in life k8s version which is now EoL, so happy to change that 🙂

deploy/helm/elastic-agent/README.md.gotmpl
pkoutsovasilis: I think you are referring to `ESECKRef` and that's actually a good question; If I am not mistaken when you do an reference to an existing cluster the ECK operator can search for it in the namespaces the operator has visibilities, but I am going to test this and come back

deploy/helm/elastic-agent/examples/kubernetes-only-logs/agent-kubernetes.yaml
pkoutsovasilis: I started with `false` there as the capability to run rootless (properly) will be part of 8.16.0 version

deploy/helm/elastic-agent/values.yaml
pkoutsovasilis: maybe we want to support that?! if we don't this can be default to something else or maybe switch to wolfi-based images?!

deploy/helm/elastic-agent/values.yaml
pkoutsovasilis: so there was this "notion" back in the day of being able to deploy an elastic-based agent or another agent that the configs are gonna be adjusted to that, hence the `enabled` flag. But you are right, this is not needed and I will remove it

pkoutsovasilis: I removed this option, please feel free to unresolve if you spot anything else

deploy/helm/elastic-agent/README.md.gotmpl
pkoutsovasilis: ty for the proposal @blakerouse, I added a complete example about ECK with a cluster in a different namespace 🙂 feel free to unresolve if you spot anything else

deploy/helm/elastic-agent/templates/agent/cluster-role-binding.yaml
```diff
@@ -0,0 +1,28 @@
+{{- include "elasticagent.init" $ -}}
+{{- range $presetName, $presetVal := $.Values.agent.presets -}}
+{{- $serviceAccountName := printf "agent-%s" (lower $presetName) -}}
+{{/* cluster role binding is not namespace bound so let's try to give it a unique enough name */}}
+{{- $clusterRoleName := printf "agent-%s" (cat $presetName $.Release.Name $.Release.Namespace | sha1sum) -}}
```
swiatekm, line 5: I would prefer this to be more informative. The name should include the Helm Chart name, the release name and the namespace in plain text, imo. The full SHA-1 is also a bit too long.

deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
```diff
@@ -0,0 +1,91 @@
+{{- include "elasticagent.init" $ -}}
+{{- range $presetName, $presetVal := $.Values.agent.presets -}}
+{{/* cluster role binding is not namespace bound so let's try to give it a unique enough name */}}
+{{- $clusterRoleName := printf "agent-%s" (cat $presetName $.Release.Name $.Release.Namespace | sha1sum) -}}
```
swiatekm, line 4: I would prefer this to be more informative. The name should include the Helm Chart name, the release name and the namespace in plain text, imo. The full SHA-1 is also a bit too long.

deploy/helm/elastic-agent/templates/agent/eck/_pod_template.tpl
```diff
@@ -0,0 +1,82 @@
+{{- define "elasticagent.engine.eck.podTemplate" }}
```
swiatekm, line 1: Does this need to be a template function? Could we not make a *.yaml file for this and include that in the manifests? It'd be nicer to work with due to syntax highlighting and such.

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
```diff
@@ -0,0 +1,133 @@
+{{- define "elasticagent.engine.k8s.podTemplate" }}
```
swiatekm, line 1: Does this need to be a template function? Could we not make a *.yaml file for this and include that in the manifests? It'd be nicer to work with due to syntax highlighting and such.

swiatekm (commented):
Functionally, all of this looks good to me. I left some comments about specific issues, but those aren't that important.

My biggest issue with this Chart is that it uses a lot of named templates in tpl files. These make the Chart much harder to maintain in general. They're difficult to read, test and debug, and usually make things more difficult to manage than just plain duplication.

Can you try to cut down on these? Some things you could do, that I've commented on in specific cases:

* Use a values.schema.json for validation instead.
* If you have a large named template (for example for the Pod template), move it into a normal yaml file and include it where it needs to be.
* Try to avoid small named templates unless you use them in more than 2 places.

This is intentionally a somewhat vague request, because I don't feel sufficiently competent when it comes to the agent itself to prescribe what exactly needs changing here. Hopefully that's ok.

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
```diff
@@ -0,0 +1,412 @@
...
+{{- $ := . -}}
+{{- range $presetName, $presetVal := $.Values.agent.presets -}}
+{{- include "elasticagent.preset.mutate.unprivileged" (list $ $presetVal) -}}
+{{- include "elasticagent.preset.mutate.fleet" (list $ $presetVal) -}}
+{{- $presetMode := dig "mode" ("") $presetVal -}}
+{{- if not $presetMode -}}
+{{- fail (printf "mode is missing from preset \"%s\"" $presetName) -}}
+{{- else if eq $presetMode "deployment" -}}
+{{- else if eq $presetMode "statefulset" -}}
+{{- else if eq $presetMode "daemonset" -}}
```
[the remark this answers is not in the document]
pkoutsovasilis: sure thing, if Helm provided a way to automate this it would be way easier but as of now this is too much extra work and I would propose to be part of a follow up PR

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
pkoutsovasilis: how do you include dynamically at will *.yaml templates @swiatekm ? because even [here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_pod.tpl) the equivalent pod template needs to be a *.tpl to be included dynamically

deploy/helm/elastic-agent/templates/agent/eck/_pod_template.tpl
pkoutsovasilis: look at the comment below

deploy/helm/elastic-agent/templates/agent/cluster-role-binding.yaml
pkoutsovasilis: the issue I see with going directly with release name and namespace is that if the user has a long release name or/and a long namespace this will be a long-long cluster-role-binding name and it maybe fail as there is a max name limit.  However, If you insist no problem changing that. That said, objects managed by Helm is better not to be edited by the user so I am not entirely sure why you choose informative vs robustness here

deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
pkoutsovasilis: above comment

pkoutsovasilis:
> Functionally, all of this looks good to me. I left some comments about specific issues, but those aren't that important.
> 
> My biggest issue with this Chart is that it uses a lot of named templates in tpl files. These make the Chart much harder to maintain in general. They're difficult to read, test and debug, and usually make things more difficult to manage than just plain duplication.

This Helm chart has to adhere to kubernetes configuration, multiple and different agent presets, different agent modes (managed vs unmanaged), and many other knobs. So if you have an actual proposal how we could maintain the required "dynamicity" but not resort to named templates, more that happy to hear plan. 🙂

> Can you try to cut down on these? Some things you could do, that I've commented on in specific cases:
> 
> * Use a values.schema.json for validation instead.

Sure this would be beneficial and will get us away with some "checking/validation" templates but every other template (mutating an agent presets, etc.) will remain

> * If you have a large named template (for example for the Pod template), move it into a normal yaml file and include it where it needs to be.
> * Try to avoid small named templates unless you use them in more than 2 places.
> 
> This is intentionally a somewhat vague request, because I don't feel sufficiently competent when it comes to the agent itself to prescribe what exactly needs changing here. Hopefully that's ok.

This is indeed a vague request given what this Helm chart has to do

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
```diff
@@ -0,0 +1,412 @@
...
+{{- end -}}
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "elasticagent.labels" -}}
+helm.sh/chart: {{ include "elasticagent.chart" . }}
+{{ include "elasticagent.selectorLabels" . }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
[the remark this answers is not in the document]
pkoutsovasilis: I went with the default ones that Helm given you when you create a chart. I don't see the value in the context of this Helm chart for these labels `app.kubernetes.io/part-of` (we are not part of a higher-level application), `app.kubernetes.io/component` (we have only one component) so if you don't insist for this PR I won't extend the labels

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
```diff
@@ -0,0 +1,412 @@
...
+{{- if empty $.Values.agent.fleet.kibanaPassword -}}
+{{- fail "kibana password must be defined when fleet mode is enabled and a token is not supplied" -}}
+{{- end -}}
+{{- end -}}
+{{- if empty $.Values.agent.fleet.preset -}}
+{{- fail "preset must be defined when fleet mode is enabled" -}}
+{{- end -}}
+{{- if not (hasKey $.Values.agent.presets $.Values.agent.fleet.preset)}}
+{{- fail (printf "specified preset \"%s\" under fleet is not found" $.Values.agent.fleet.preset) -}}
+{{- end -}}
```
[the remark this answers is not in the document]
pkoutsovasilis: Some of the nested bit will remain as we have to validate the fields you mention only when the fleet mode is enabled. I am not a big fun of `required` in this particular case as it will indeed fail if the specified variable is empty but if it is not it will return it and I will have to do an empty assignment ($_) not to have it printed. However, I will switch to it and hopefully the one liner will look better

magefile.go
```diff
@@ -3174,3 +3181,112 @@ type otelDependencies struct {
...
+	}
+
+	for _, d := range dirEntries {
+		if !d.IsDir() {
+			continue
+		}
+
+		if !d.IsDir() {
+			return nil
+		}
```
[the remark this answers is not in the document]
pkoutsovasilis: nice catch! nope the second IsDir is not needed, I just pushed the removal of it

deploy/helm/elastic-agent/templates/agent/cluster-role-binding.yaml
swiatekm: We can truncate it if it's too long. I'm fine with, say, truncating it at 55 chars and then adding the first 8 chars of the hash. The reason I want to be informative is that K8s clusters have many users, some of whom may simply be looking at resources without knowing which Helm releases are installed. In that case, it's much easier to tell at a glance what a given resource is for by just looking at the name.

I don't feel too strongly about this, but if you want to keep the current format, I'd do something like `elastic-agent-{short_sha}` instead. The full hash is not very readable imo.

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
swiatekm: Here's an example, the otel-operator Chart has a templated CRD in the form of a yaml file: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/conf/crds/crd-opentelemetrycollector.yaml
and this is included dynamically via a combination of `tpl` and `Files.Get`: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/b8d11b63916bc40cd1aa2875a880f3995978d434/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L273.

pkoutsovasilis: @swiatekm this example essentially loads up the contents of the yaml file and then evaluates them as template which is practically the same as having directly a *.tpl. With the former approach though you have now to maintain also file structures where with the tpl approach you can reference the template from anywhere by name. What exactly do we gain by switching to this paradigm?

deploy/helm/elastic-agent/values.yaml
swiatekm: @pkoutsovasilis are we intentionally not putting this under `appVersion` in Chart.yaml? Normally that indicates the Chart works with a wide array of application versions.

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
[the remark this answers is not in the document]
swiatekm: Those aren't that useful, but we should set `app.kubernetes.io/version` to `.Values.agent.version`.

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
swiatekm: The benefit is that a *.yaml file can get YAML syntax highlighting from most tools, and often even K8s manifest linting. It's usually also obvious where the templating happens. In comparison tpl files get none of that, and non-tpl text in them is just treated as plain text.

In this very PR, reviewing the templated K8s manifest files is much easier than reviewing the tpl files, for this simple reason.

pkoutsovasilis: but @swiatekm the tpl under debate essentially fills up the spec of the podTemplate of either a `daemonset`, `statefulset` or `deployment`. Even if I switched this to *.yaml, how you would get linting since this yaml will still not feature any of the required yaml fields, such `apiVersion:`, `kind: ...`, etc., so your editor can give you proper linting?

pkoutsovasilis: I think what you are after is essentially duplicating this template in all daemonset, statefulset or deployment and having each k8s object as a whole. However there is an upper limit on the size of the releases and all these templates when you do a helm install are compressed and saved inside a kubernetes configmap. Thus since this pod spec is the same for all three deployment options it became a single tpl file

deploy/helm/elastic-agent/templates/agent/cluster-role-binding.yaml
pkoutsovasilis: just pushed your proposal there, feel free to unresolve if you spot anything else

deploy/helm/elastic-agent/values.yaml
pkoutsovasilis: yes I do that intentionally, because this Helm chart should be able to work with multiple elastic-agent versions. This was at least part of the old requirements, however maybe it is non-relevant any more?!

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
swiatekm: I don't actually want to duplicate this specific template, I'd just like to make it easier to read. This matters less for smaller templates, like the volumes or env variables, but for the pod template I think it's worth it to make it a separate yaml file.

If you're really against it, I won't push, but I do think doing this would make things easier for us in the long term.

pkoutsovasilis: @swiatekm I think you focused just on my latest reply, but what about the missing yaml fields required for linting that I mention [here](https://github.com/elastic/elastic-agent/pull/5331#discussion_r1740979067)? do you believe that by making this yaml without having the aforementioned fields will "help" your IDE/text editor to provide linting about pod spec?

deploy/helm/elastic-agent/templates/agent/cluster-role-binding.yaml
swiatekm: Turns out we don't need to truncate this or include the hash. ClusterRole and ClusterRoleBinding names can be up to 255 characters. I don't think this is documented anywhere explicitly, but I tested it in a kind cluster.

So we can have a descriptive name and not risk hash collisions.

swiatekm: Both release name and namespace name are capped at 63 characters, and I think preset name is reasonably small as well? So I don't think we need to truncate at all.

deploy/helm/elastic-agent/Chart.yaml
blakerouse: Sounds like a good reason. Just didn't know if a k8s feature was required at that level.

deploy/helm/elastic-agent/examples/kubernetes-only-logs/agent-kubernetes.yaml
blakerouse: Makes sense.

deploy/helm/elastic-agent/values.yaml
blakerouse: I would not default to `-complete` its much larger, just for synthetics. I would expect them to adjust the chart to use `-complete` if they want to run synthetics. Wolfi image could also be the default, no issues there from me.

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
pkoutsovasilis: not sure what IDE you are using but for my case - goland  + kubernetes plugin - YAML (not k8s specific object) highlighting works for templates. To this end, even when I did the change to yaml locally I couldn't find an option to enforce this yaml to be parsed as the spec of a pod template, and hence get k8s specific linting. For now I will resolve this but if you have more hassle-free steps that we could use to get former, for sure, we can switch to that

deploy/helm/elastic-agent/values.yaml
blakerouse: I think the open question here is how is this Helm chart getting released? Does it get updated out of sync of the Elastic Agent release or does it get released at the same time of the full stack release? If its stack release based then this should really sync with the version of the repository [here](https://github.com/elastic/elastic-agent/blob/main/version/version.go#L7) and when its the `main` repository it should really be `-SNAPSHOT` as that is the only released version for running code in actual `main`. When `8.16` is created, then it should not use `-SNAPSHOT`.

pkoutsovasilis: Just a quick thinking about that; I would like having this chart "following" the agent release to always have the appropriate agent image in it, but also be able to bump helm chart version to fix any agent-unrelated issue, e.g. issues with templating?!

swiatekm: Can we defer the release process question for now? It sounds like it could be its own major discussion, but whatever we decide is not going to require major changes to this Chart, which we're not publishing yet anyway. I think it might be better to merge this PR without answering this question, and do that in a separate issue. WDYT @blakerouse @pkoutsovasilis ?

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.tpl
swiatekm: That's all right. After we get this merged, I'll put up a draft PR to show you what change I had in mind exactly.

deploy/helm/elastic-agent/templates/agent/k8s/daemonset.yaml
```diff
@@ -0,0 +1,31 @@
+{{- include "elasticagent.init" $ -}}
+{{- range $presetName, $presetVal := $.Values.agent.presets -}}
+{{- if and (eq ($presetVal).mode "daemonset") (eq $.Values.agent.engine "k8s") -}}
+{{- $agentName := printf "agent-%s" (lower $presetName) -}}
```
swiatekm, line 4: These names should include the release name as well, imo. Otherwise, we'd get conflicts installing different agent releases in the same namespaces. What most Charts do is use the equivalent of `{{ template "elasticagent.fullname" . }}` here. That makes truncating a bit annoying, but it's necessary unless we want to explicitly forbid multiple installs in the same namespace.

This applies to every namespaced resource we name this way.

swiatekm (commented):
I have one final comment about names, but otherwise I think we're good to go.

I would still like to move away from named templates as much as possible, and definitely add a schema file for validation, but that can happen in a follow-up.

deploy/helm/elastic-agent/templates/agent/k8s/daemonset.yaml
pkoutsovasilis: oops, nice catch! yep this needs to change, working on it

deploy/helm/elastic-agent/values.yaml
pkoutsovasilis: yep I agree with deferring the release process question for now. That said, I added a helm:updateAgentVersion mage target that I deem it would be useful in this discussion

deploy/helm/elastic-agent/templates/agent/k8s/daemonset.yaml
pkoutsovasilis: these names now include the helm release name as well, feel free to unresolve if you spot anything else

swiatekm (approved):
:+1:

deploy/helm/elastic-agent/values.yaml
blakerouse: I agree we should defer. A follow up PR to change this is simple.

blakerouse (approved):
Looks good, I am ready for this to be merged.

pkoutsovasilis:
> Looks good, I am ready for this to be merged.

haha!! this phrase sounded like "release the Kraken"! 😄 ty both @swiatekm and @blakerouse for the comments and the reviews. That said, the agent-extended-testing CI step, which is required, is failing but it is not something caused by this PR?!

blakerouse:
@pkoutsovasilis It is very much that!

The failures are flaky tests that we are working on, also seems there is a conflict now. I would merge main into the PR and fix the conflict and have it re-run to get a green CI run.

ycombinator:
@pkoutsovasilis If we want to release the Helm chart as a beta initially, is there any way we can mention this as part of the chart itself, such that users would clearly see it before they use it or right as they start to use it?

pkoutsovasilis:
@ycombinator just thinking out loud the available options here:

1. I assume that we can push the helm chart with a beta suffix at the version and update the NOTES.txt accordingly here https://helm.elastic.co/
2. Somebody that wants to try this helm chart can clone the repo and install the helm chart from the folder `deploy/helm/elastic-agent`
3. We could utilise github pages on this repo and commit directly the helm chart package.tgz (that points to the SNAPSHOT agent image)  [my least favourite option]

that's why a discussion about the Helm chart release process is wise to have before any next step 🙂

swiatekm:
Helm Charts use semver 2.0 for versioning, so it's simple enough to just designate our releases as betas by setting it to `va.b.c-beta.d` or something similar. This makes Helm automatically treat it as a prerelease. This will cause them not to show up in searches without the `--devel` flag, amongst other consequences.

ycombinator:
Thanks @pkoutsovasilis and @swiatekm.  I've created https://github.com/elastic/elastic-agent/issues/5485 to continue the discussion on the beta marking and to track the implementation as well.  

Likewise, I have created https://github.com/elastic/elastic-agent/issues/5486 to resume the discussion about the release process that we [deferred](https://github.com/elastic/elastic-agent/pull/5331#discussion_r1742638157) from this PR.

Issue #4215, opened 2024-02-07
[Flaky Test]: TestComponentBuildHashInDiagnostics – build hash is empty


AndersonQ:
### Failing test case

TestComponentBuildHashInDiagnostics

### Error message

component beat/metrics-monitoring: VersionInfo.Meta.Commit mismatch: expected: "4e255e6fcc578d8628ff9d415a7a30504f5c13f1", actual  : ""

### Build

https://buildkite.com/elastic/elastic-agent/builds/6883#018d82da-5f3c-43be-914d-952a4b7216c5

### OS

Windows

### Stacktrace and notes

```markdown
When the diagnostics is taken the components seems to not be yet ready, therefore the build hash is empty


    package_version_test.go:166: 
        	Error Trace:	C:/Users/windows/agent/testing/integration/package_version_test.go:166
        	Error:      	Not equal: 
        	            	expected: "4e255e6fcc578d8628ff9d415a7a30504f5c13f1"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-4e255e6fcc578d8628ff9d415a7a30504f5c13f1
        	            	+
```
```

**Comment by rdner** [comment]
The issue seems to be back in https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1725#0191083f-be06-4e18-b33c-f4641f18385c:


```
package_version_test.go:161: unzipping file "C:\\Users\\windows\\AppData\\Local\\Temp\\TestComponentBuildHashInDiagnostics3173096240\\003\\logs\\elastic-agent-8.16.0-SNAPSHOT-633a16\\system"
    package_version_test.go:161: creating directory "C:\\Users\\windows\\AppData\\Local\\Temp\\TestComponentBuildHashInDiagnostics3173096240\\003\\logs\\elastic-agent-8.16.0-SNAPSHOT-633a16\\system"
    package_version_test.go:190: 
        	Error Trace:	C:/Users/windows/agent/testing/integration/package_version_test.go:190
        	Error:      	Not equal: 
        	            	expected: "a492c33e257b3b464d77b5f4156ca76043ce48f3"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-a492c33e257b3b464d77b5f4156ca76043ce48f3
        	            	+
[...]

PR #5333, opened 2024-08-21
Avoid reporting empty version_info for components that just started

AndersonQ:
## What does this PR do?

The state diagnostics hook now re-fetches once the state when necessary, improving the accuracy of component status reporting.

## Why is it important?

Given a design choice, the coordinator's reported state might be stale, leading to healthy components lacking version_info immediately after they start.

## Checklist

- ~~[ ] I have added an entry in `./changelog/fragments` using th~~e [changelog tool](https://github.com/elastic/elastic-agent#changelog)

Run the [`TestDiagnosticState`](https://github.com/AndersonQ/elastic-agent/blob/db1be7588b892df7d2ca902d5edd49727ae766d8/internal/pkg/agent/application/coordinator/diagnostics_test.go#L413) test

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/4215

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/package_version_test.go
+++ b/testing/integration/package_version_test.go
@@ -102,16 +102,23 @@ func TestComponentBuildHashInDiagnostics(t *testing.T) {
 		}
 
 		for _, c := range status.Components {
+			bs, err := json.MarshalIndent(status, "", "  ")
+			if err != nil {
+				stateBuff.WriteString(fmt.Sprintf("%s not health, could not marshal status outptu: %v",
+					c.Name, err))
+				return false
+			}
+
 			state := client.State(c.State)
 			if state != client.Healthy {
-				bs, err := json.MarshalIndent(status, "", "  ")
-				if err != nil {
-					stateBuff.WriteString(fmt.Sprintf("%s not health, could not marshal status outptu: %v",
-						c.Name, err))
-					return false
-				}
 
 				stateBuff.WriteString(fmt.Sprintf("%s not health, agent status output: %s",
+					c.Name, bs))
+				return false
+			}
+
+			if c.VersionInfo.Meta.Commit == "" {
+				stateBuff.WriteString(fmt.Sprintf("%s health, but no versionInfo. agent status output: %s",
 					c.Name, bs))
 				return false
 			}


blakerouse (changes requested):
I don't fully understand why this is being done? Do we really need it to have that guarantee? I don't feel like we do.

I don't like the idea that we are adjust core code to fix a condition that the test should just probably handle.

Why not have the test fetch the state twice, instead of doing it internally in agent?

AndersonQ:
> Why not have the test fetch the state twice, instead of doing it internally in agent?

my reasoning was that we choose to allow that to happens, so we might as well try to prevent it to happen in the diagnostics. As it's just the diagnostics hook code, I believe it's localised enough to do not lead to unexpected behaviour.

Also it's been happening on the tests, making it not that unlikely as expected

blakerouse:
I don't really see how this is really fixing anything, as it just calls the same function again in that one case. So its just reducing the chance that it could happen, but it still will happen.

AndersonQ:
This is the build info that should come during the check-in, the first one. Therefore most likely the 1st check-in happens, the coordinator internally have it all updated but because of the broadcaster when the diagnostics hook runs, the received state is stale. However a second call should return the right state because by the time the call happens again 1) the stale value has been read by the 1st call 2) the new value is on the broadcaster. Thus a 2nd call should be enough.

blakerouse:
@AndersonQ how are your confirming that the new value is present?

AndersonQ:
> @AndersonQ how are your confirming that the new value is present?

I'm not. If the component is healthy, the check-in happened and the versionInfo was sent. Thus there isn't much reason to believe several retries would be needed. Besides if it were the case the versionInfo isn't be correctly updated 1) the diagnostics should proceed and not get stuck anyway 2) we'd catch it on our tests. Also the diagnostic from the flaky test has all then versionInfo, shoing a retry might be all we need.

blakerouse:
> > @AndersonQ how are your confirming that the new value is present?
> 
> I'm not. If the component is healthy, the check-in happened and the versionInfo was sent. Thus there isn't much reason to believe several retries would be needed. Besides if it were the case the versionInfo isn't be correctly updated 1) the diagnostics should proceed and not get stuck anyway 2) we'd catch it on our tests. Also the diagnostic from the flaky test has all then versionInfo, shoing a retry might be all we need.

It is still unclear how a single retry provides the guarantee that versionInfo is now set. That is what I do not understand, and why I am not okay with this PR at the moment. Currently in my head this change is just reducing the chance slightly, but it is still possible.

AndersonQ:
> It is still unclear how a single retry provides the guarantee that versionInfo is now set. That is what I do not understand, and why I am not okay with this PR at the moment. Currently in my head this change is just reducing the chance slightly, but it is still possible.

The healthy status and the version_info should arrive together, therefore it should be pretty much instantaneous the update. That's why I believe it more than a slight reduction on the chance of getting the an empty version_info. 

Anyway I added more retries with a timeout between them. If all the retries run it's a 250ms + the normal code execution time, what should be a pretty much negligible on the total time to get a diagnostics. Does it looks more robust now?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -925,13 +924,19 @@ func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
...
+				var compStates []StateComponentOutput
+			outerLoop:
+				for i := 0; i < 5; i++ {
+					s, compStates = getState(c)
+					for _, cs := range compStates {
+						if cs.State.State == client.UnitStateHealthy &&
+							cs.State.VersionInfo.BuildHash != "" &&
+							cs.State.VersionInfo.Meta["commit"] != "" {
+							break outerLoop
+						}
```
blakerouse, line 937: This whole block would be much better as its own function either in `getState` or another wrapper around it.

I still am not a fan of this, again your just improving the odds and not solving the true issue. Could we some how guarantee that when status is updated to Healthy that the VersionInfo is set at the same time? Why are they not?

If there is a bad component that never provides the versioninfo it is going to delay this result as well indefinitely by 250 milliseconds.

AndersonQ:
> [...] again your just improving the odds and not solving the true issue

that's the whole idea. The broadcaster was built with the possibility of reporting stale states and the coordinator uses it that way. Changing it all isn't worth it, thus improving the odds seems good enough. 

the broadcaster might report a stale state:
https://github.com/elastic/elastic-agent/blob/729636ac6a4d83286f4bfca3e8e0d2a2b1a46a6f/pkg/utils/broadcaster/broadcaster.go#L194-L199

and how the coordinator uses it:
https://github.com/elastic/elastic-agent/blob/41ee2bb2f8e96ff902d0c88ed19ac5e3734973b9/internal/pkg/agent/application/coordinator/coordinator.go#L372-L387

the new state is sent on `refreshState()`
https://github.com/elastic/elastic-agent/blob/c929f79c7af4ea99a24104f155eb3a6e20910558/internal/pkg/agent/application/coordinator/coordinator_state.go#L127-L132

triggered by the coordinator's runLoop:
https://github.com/elastic/elastic-agent/blob/41ee2bb2f8e96ff902d0c88ed19ac5e3734973b9/internal/pkg/agent/application/coordinator/coordinator.go#L1095-L1099

the broadcaster will update the value its `Get` sends as soon as it receives a new value:
https://github.com/elastic/elastic-agent/blob/729636ac6a4d83286f4bfca3e8e0d2a2b1a46a6f/pkg/utils/broadcaster/broadcaster.go#L542

there is a lot of indirection until the new state finally arrives at the broadcaster to be consumed by the diagnostics hook. That's why I'm not trying to fix it completely, just avoid the problem with a good old retry. 

Our integration tests will say how good the fix is and we can adjust accordingly. That's why I rather have a single retry for now, which would have minimal impact on the time spent collecting a diagnostics and I believe should be enough to get an up-to-date state.

blakerouse:
I still don't like this change, even with the explanation. Honestly I don't believe it is even fixing the problem, just masking it with better odds.

If the component is being updated with healthy which the package_version_test.go is ensuring that it is healthy then it should have that version information.

Another option is to just adjust the test and make the `allHealthy` function to ensure that version information is present before reporting that the Elastic Agent is healthy. https://github.com/elastic/elastic-agent/blob/main/testing/integration/package_version_test.go#L94

You can use the `c.VersionInfo` to verify that the status information has the version information. If its present there then it should defiantly have it in the diagnostics.

AndersonQ:
> Another option is to just adjust the test and make the allHealthy function to ensure that version information is present before reporting that the Elastic Agent is healthy. https://github.com/elastic/elastic-agent/blob/main/testing/integration/package_version_test.go#L94

This does not address the problem at all, it just hides the issue. That's why I'm against that approach.

> Honestly I don't believe it is even fixing the problem, just masking it with better odds.

That's the idea ;)
It isn't supposed to be a perfect fix. The issue does not compromise the normal behaviour of the agent and a complete fix isn't worth it. That's why I chose a palliative fix, improving the odds of it succeeding with minimal overhead and added complexity.

We're seeing this test failing on CI and tweaking the test to ignore the issue isn't a good approach or a good precedent. If we merge this in the current state we'll be able to see if the problem persists or not. If it persists, another approach is needed, if it doesn't, problem solved.
This at least attempt to address the real issue instead of just ignoring it. Besides all the evidence (the failed tests) suggest a retry might be enough.

blakerouse:
> > Another option is to just adjust the test and make the allHealthy function to ensure that version information is present before reporting that the Elastic Agent is healthy. https://github.com/elastic/elastic-agent/blob/main/testing/integration/package_version_test.go#L94
> 
> This does not address the problem at all, it just hides the issue. That's why I'm against that approach.
> 
> > Honestly I don't believe it is even fixing the problem, just masking it with better odds.
> 
> That's the idea ;) It isn't supposed to be a perfect fix. The issue does not compromise the normal behaviour of the agent and a complete fix isn't worth it. That's why I chose a palliative fix, improving the odds of it succeeding with minimal overhead and added complexity.
> 
> We're seeing this test failing on CI and tweaking the test to ignore the issue isn't a good approach or a good precedent. If we merge this in the current state we'll be able to see if the problem persists or not. If it persists, another approach is needed, if it doesn't, problem solved. This at least attempt to address the real issue instead of just ignoring it. Besides all the evidence (the failed tests) suggest a retry might be enough.

Being that we know that its possible for the version information to be delayed (which even based on all the explanation in the PR, still doesn't make sense on why it would be separate from a single Healthy update), and we accepting that delay. My suggestion is to just fix the test to ensure that its always there before performing diagnostics as the test is explicitly testing contents of diagnostic output.

I don't think we actually care in the real world that it is missing on a freshly started running Elastic Agent. That is why I am suggesting to not to make a change inside the core of the Elastic Agent at all. I have not seen an argument that ensures that we need this guarantee, except in this one test case.

pchila (commented):
After reading the whole discussion and having a look at the code modification I tend to agree with @blakerouse about fixing this in test code rather than adding retried to the diagnostic hooks.

If we accept that the version_info will eventually be available and dumped in the diagnostics I would prefer to change the asserts on the test to reflect that.

Conversely, if we need some stronger guarantees about the state returned by the broadcaster it's a much bigger change but it's probably not worth it since the chance of diagnostics generated really quickly after components performed their first checkin is rather small and we can always ask for more diagnostics

AndersonQ:
There is something off, but let's get the flaky failures out of the way for now. I think this solution is good enough for now.

blakerouse (approved):
Thank you for just adjusting the test!

Looks good.

PR #5345, opened 2024-08-23
Improve watcher and TestWatcher_AgentErrorQuick logs

AndersonQ:
## What does this PR do?

It improves the watcher logs and prints the logs if TestWatcher_AgentErrorQuick fails.
If the test would fail, the logs would be like the following:
```
❯ go test -run TestWatcher_AgentErrorQuick ./internal/pkg/agent/application/upgrade/
--- FAIL: TestWatcher_AgentErrorQuick (1.00s)
    watcher_test.go:287: [info] Agent watcher started
    watcher_test.go:287: [info] Trying to connect to agent
    watcher_test.go:287: [info] Connected to agent
    watcher_test.go:287: [debug] received state: FAILED:force failure
    watcher_test.go:287: [info] Communicating with PID 0
    watcher_test.go:287: [debug] received state: HEALTHY:healthy
    watcher_test.go:287: [error] Agent reported failure (starting failed timer): agent reported failed state: force failure
    watcher_test.go:287: [info] Agent reported healthy (failed timer stopped)
    watcher_test.go:287: [debug] received state: error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
    watcher_test.go:287: [error] Lost connection: failed reading next state: rpc error: code = DeadlineExceeded desc = context deadline exceeded
FAIL
FAIL	github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade	1.351s
FAIL
```

## Why is it important?

`TestWatcher_AgentErrorQuick` was flaky before, but it hasn't happened again on CI. Even running it for 12 hours didn't reproduce the problem. 

## Checklist

- [x] My code follows the style guidelines of this project
- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Disruptive User Impact

None

## How to test this PR locally

make `TestWatcher_AgentErrorQuick` to fail by adding `t.Fail()` at the end of the test and then run the test.

```
❯ go test -count 43200 -run TestWatcher_AgentErrorQuick -timeout=0
PASS
[...]

PR #5346, opened 2024-08-23
Improve testing logger

AndersonQ:
## What does this PR do?

Adds a PrettyPrintf method to ObservedLogs to logger.NewTesting

## Why is it important?

The testing logger is helpful, however on every failed test it's necessary to manually format the logs before printing. That leads to several tests just printing the logs as they are which isn't the best for reading them and investigating why the test failed.

The new `PrettyPrintf` method provides an easy way to pretty print the logs
 
## Checklist

## Disruptive User Impact

None

run the test `TestObservedLogs_PrettyPrintf`

## Related issues

- Relates https://github.com/elastic/elastic-agent/pull/5345

## Questions to ask yourself

- ...


internal/pkg/agent/application/upgrade/watcher.go
```diff
@@ -133,12 +134,11 @@ LOOP:
 			// block on connection, don't retry connection, and fail on temp dial errors
 			// always a local connection it should connect quickly so the timeout is only 1 second
 			connectCtx, connectCancel := context.WithTimeout(ctx, 1*time.Second)
-			//nolint:staticcheck // requires changing client signature
 			err := ch.agentClient.Connect(connectCtx, grpc.WithBlock(), grpc.WithDisableRetry(), grpc.FailOnNonTempDialError(true))
```
mauri870, line 137: As a follow up to the linter message, seems like `WithBlock` and `FailOnNonTempDialError` are considered an [anti-pattern](https://github.com/grpc/grpc-go/blob/aae9e64/Documentation/anti-patterns.md).

Following up on the linter message, it appears that WithBlock and FailOnNonTempDialError are considered an [anti-pattern](https://github.com/grpc/grpc-go/blob/aae9e64/Documentation/anti-patterns.md).

It’s not entirely clear why these are used. They might be remnants of old code or included to work around an error or limitation. Could they be removed?

internal/pkg/agent/application/upgrade/watcher_test.go
```diff
@@ -270,14 +270,18 @@ func TestWatcher_AgentError(t *testing.T) {
...
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
 
 	errCh := make(chan error)
-	logger, _ := logger.NewTesting("watcher")
-	w := NewAgentWatcher(errCh, logger, 100*time.Millisecond)
+	log, obs := logger.NewTesting("watcher")
+	defer func() {
+		if t.Failed() {
+			obs.PrettyPrintf(t.Log)
```
mauri870, line 281: Maybe we can implement String() to ObserverLogs? Then the usage becomes `t.Log(obs)`

AndersonQ: What I don't like on this idea is that `ObservedLogs` contains all logs, so a `String()` method would have to decide how to deal with the multi-line nature of it and it'd not be possible to print one log at a time.

mauri870: You are right that String() would make this a bit more limited in use, but ObserverLogs has filters you can apply to limit the amount of logs that would be returned (see `Filter*` [methods](https://pkg.go.dev/go.uber.org/zap@v1.27.0/zaptest/observer)). I understand the limitation regarding multiline logs.

Another consideration is that PrettyPrintf calls the callback function with a string, so it is kinda like an inversion of control, wouldn't it read better to have `t.Log(obs.Pretty())`? This way it plays better with Logf, FatalF or any other Printer utility you might use. I believe it gives more control to the caller.

internal/pkg/agent/application/upgrade/watcher.go
```diff
@@ -133,12 +134,11 @@ LOOP:
 			// block on connection, don't retry connection, and fail on temp dial errors
 			// always a local connection it should connect quickly so the timeout is only 1 second
 			connectCtx, connectCancel := context.WithTimeout(ctx, 1*time.Second)
-			//nolint:staticcheck // requires changing client signature
```
andrzej-stencel, line 136: Why is this `nolint` clause being removed?

internal/pkg/agent/application/upgrade/watcher_test.go
pchila: +1 for having String() or Pretty() implemented for `ObservedLogs` or even for a single `LoggedEntry` as it would be easier and more flexible

pchila (commented):
Just a comment on the pretty-printing already pointed out by @mauri870 
The removal of `nolint` directive now introduces a lint failure, I would rather we fix the problem if possible, or maybe restore the `nolint` if it's too much work extraneous to this PR and fix it in a follow-up PR

internal/pkg/agent/application/upgrade/watcher_test.go
andrzej-stencel: My reservation about adding `ObserverLogs.String()` is that it would write all logs into one big string, which can use a lot of memory when there's a lot of logs. I like the approach of dumping the logs to strings one at a time.

On the other hand, my concern about the `ObserverLogs.TakeAndPrettyPrint()` method is that causes code like https://github.com/elastic/elastic-agent/pull/5346/files#diff-1881a0cf2ec8461140501dbbb913d9d91b93f00572c2e13977a6fca09b5ba92bR185:

```
obs = &logger.ObservedLogs{ObservedLogs: obs.Filter(func(entry observer.LoggedEntry) bool {
```

where you need to encapsulate the result of `obs.Filter` in a new `logger.ObservedLogs`. This notation is a bit verbose.

Another approach would be to have a function like `printTestLogs(t *testing.T, logs []observer.LoggedEntry)` like the one already defined [here](https://github.com/elastic/elastic-agent/blob/fd477ec3aa00afc77ae86e7fa9a4ce978a6958b6/internal/pkg/agent/application/upgrade/artifact/download/http/downloader_test.go#L341-L346):

```go
func printLogs(t *testing.T, logs []observer.LoggedEntry) {
	t.Helper()
	for _, entry := range logs {
		t.Logf("[%s] %s", entry.Level, entry.Message) // add dumping entry.ContextMap()
	}
}
```

This function could then be used like this:

```go
	defer func() {
		if t.Failed() {
			printTestLogs(t, obs.TakeAll())
		}
	}()
```

One benefit of `ObserverLogs.TakeAndPrettyPrint` over `printLogs` is that `TakeAndPrettyPrint` accepts a print function, giving the caller the choice where the logs are printed. But don't we always want to use `t.Log` in the tests?

AndersonQ:
> Just a comment on the pretty-printing already pointed out by @mauri870 The removal of `nolint` directive now introduces a lint failure, I would rather we fix the problem if possible, or maybe restore the `nolint` if it's too much work extraneous to this PR and fix it in a follow-up PR

it was removed by mistake, it's back now

AndersonQ:
@mauri870, @pchila, @andrzej-stencel how about now? I think it's a better compromise, following the idea @andrzej-stencel suggested.
Sorry about the package move, it changed a lot of files, but I believe it's better to put a cleat boundary between production code and code to support test

pkg/core/logger/loggertest/logger_test.go
```diff
@@ -0,0 +1,30 @@
...
+	logger.Debug("a debug message")
+	logger.Debugw("a debug message with keys", "key2", 42)
+	logger.Infow("an info message")
+	logger.Infow("an info message with keys", "key1", "value1")
+	logger.Warn("a warn message")
+	logger.Warnw("a warn message with keys", "key2", 42)
+	logger.Error("an error message")
+	logger.Errorw("an error message with keys", "key1", "value1")
+
+	pintFn := func(a ...any) { fmt.Println(a...) }
```
mauri870, line 17: s/pintFn/printFn/

internal/pkg/agent/application/upgrade/watcher_test.go
mauri870: I still think using `t.Log(obs.Pretty())` or even `obs.String()` would be a more idiomatic approach. If performance becomes a concern, we can always use strings.Builder.

I don’t want to prolong this discussion as I don't have strong arguments to favor one approach over the other; it seems more like a matter of preference. What do you think, @AndersonQ?

AndersonQ: I'd agree with you if it were a straight forward thing, but it isn't because the type isn't on our code.

It'd require to bring back the new type, which makes using the original methods a bit annoying. And again, ObserverLogs holds several lines of logs which I rather not put on a single string. Adding the method on LogEntry would mean more new types. So I kept the original idea, something to pretty print `ObserverLogs` but with a simpler implementation

pkg/core/logger/loggertest/logger.go
```diff
@@ -0,0 +1,44 @@
...
+		}))
+
+	return log, obs
+}
+
+// PrintObservedLogs consumes, formats and prints all log entries from obs,
+// one at a time, printFn. It calls `observer.ObservedLogs.TakeAll`,
+// therefore, after calling it, the ObservedLogs will be empty.
+func PrintObservedLogs(obs *observer.ObservedLogs, printFn func(a ...any)) {
+	rawLogs := obs.TakeAll()
```
andrzej-stencel, line 36: [suggestion/non-blocking] `TakeAll` modifies the state of `obs`, so ieally I would keep it out of this function. I understand that this makes the call to `PrintObservedLogs` a bit longer, user needs to pass `obs.TakeAll()` instead of just `obs` into the as the parameter. I'm leaving it up to you Anderson.

andrzej-stencel (approved):

pkg/core/logger/loggertest/logger_test.go
pchila: :beers:

PR #5347, opened 2024-08-23
[OTel]: enable persistence by default in our OTel distribution

VihasMakwana:
- Enhancement

## What does this PR do?
- By default, the filelog receiver stores offsets in memory rather than on disk. Since users might forget to enable on-disk storage themselves, this PR enables persistence by default in our sample configurations.

## Why is it important?
- If the collector is restarted, there could either be data duplication (if it starts re-reading files from the beginning) or data loss (if it starts reading files from the end and new data has been written to these files during the restart).

## Checklist

- Pull from this branch.
- Run `elastic-agent otel` 
- Try ingesting some data
- Shutdown the otel command
- Rerun the `elastic-agent otel` 
- Write new data
- You shouldn't see duplicated data, as offsets will be stored on disk.

## Related issues

-

## Questions to ask yourself

- ...


changelog/fragments/1724406973-enable-persistence-by-default.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: enable persistence by default
```
pierrehilbert, line 14: I would mention this is for OTel

VihasMakwana:
@elastic/elastic-agent-control-plane can someone take a look here?

michalpristas (approved):

strawgate:
Do we also need to update the sample yaml configurations that onboarding uses?

VihasMakwana:
@michalpristas @ycombinator those configs already use the filestorage extension. 
For eg. https://github.com/elastic/elastic-agent/blob/fd477ec3aa00afc77ae86e7fa9a4ce978a6958b6/internal/pkg/otel/samples/linux/platformlogs_hostmetrics.yml#L1-L9

strawgate:
It seems like `STORAGE_DIR` is probably too general to be using as an environment variable here?

VihasMakwana:
@strawgate 

> If this PR changes the default then can these changes be removed from the onboarding yaml?

Not really,

If you take a look here https://github.com/elastic/elastic-agent/blob/f32e874c92032f94677c22ba26894099a7152990/internal/pkg/otel/run.go#L91-L99
I only override `STORAGE_DIR` if it's not set by the user (by default I use `elastic-agent/data/registry/otelcol`, but we can change the default).

If a user has set the `STORAGE_DIR` env, then we will not override the env.

strawgate:
For the first part of my question though -- It seems like STORAGE_DIR is probably too general to be using as an environment variable here?

VihasMakwana:
> It seems like STORAGE_DIR is probably too general to be using as an environment variable here?

Sorry, I forgot to add it in above comment.

I agree with you. It's too general. I believe we should rename it in context of agent and otel.

strawgate:
> I only override `STORAGE_DIR` if it's not set by the user (by default I use `elastic-agent/data/registry/otelcol`, but we can change the default).

So doesn't this mean that we no longer need to define this directory in the onboarding yaml? Do we want to create this folder if it doesn't exist?

strawgate:
> @strawgate In the onboarding yaml, we use the env variable `STORAGE_DIR` (quite a general name though). We don't specify any directory explicitly.

This is no longer necessary though after this PR though because if it's not set, it defaults to `env:storage_dir`, so explicitly setting is not necessary anymore?

The onboarding flow actually replaces `env:STORAGE_DIR` with `./data/otelcol` which also shouldn't be needed anymore?

VihasMakwana:
> This is no longer necessary though after this PR though because if it's not set, it defaults to env:storage_dir, so explicitly setting it to env:storage_dir is redundant and is not necessary anymore?

I believe it's necessary to set it in the onboarding YAML.

If you review the [filestorage extension documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage#file-storage), it specifies that the default directory used is `/var/lib/otelcol/file_storage`.

If we omit the `env:storage_dir` in onboarding.yaml, the OpenTelemetry Collector will default to storing offsets in `/var/lib/otelcol/file_storage`, as per the documentation. This may not be the intended behavior if we are expecting it to use `env:storage_dir`. Therefore, it's important to set it explicitly to avoid this issue.

-----

Let me give an example to make this clear (hopefully ;)),

if a user has set `storage_dir` to `/path/data/otelcol` (because he wants to store offsets in this directory) and he runs the `elastic-agent` with the otel mode, then this PR will enusre that the `/path/data/otelcol` directory exists before booting up the otel collector.

However, if we remove the `env:storage_dir` setting from the onboarding YAML, the otel collector will revert to its default behavior and use the `/var/lib/otelcol/file_storage` directory for storage.

Therefore, omitting that setting means that the otel will not use the user-specified directory (`/path/data/otelcol`) and will instead use the default path, which may not align with the user's intended configuration.

Does that make sense?

VihasMakwana:
That being said, 

If the OpenTelemetry Collector used env:storage_dir by default, your point would be valid.

strawgate:
Sorry maybe I'm confused -- https://github.com/elastic/elastic-agent/blob/f32e874c92032f94677c22ba26894099a7152990/internal/pkg/otel/run.go#L91-L99

Doesn't this PR make it so that if the user does not explicitly set a storage_dir that we set it for them to `elastic-agent/data/registry/otelcol`?

Before this PR: 
Default storage: `/var/lib/otelcol/file_storage`

After this PR:
Default storage: `.../elastic-agent/data/registry/otelcol`

Is this not a correct understanding of the change in the PR?

VihasMakwana:
@strawgate your understanding is correct. 
But we also need to explicitly set `env:storage_dir` here https://github.com/elastic/elastic-agent/blob/fd477ec3aa00afc77ae86e7fa9a4ce978a6958b6/internal/pkg/otel/samples/linux/logs_metrics_traces.yml#L50-L52

VihasMakwana:
> so the remaining issue (from my perspective) is just us using the storage_dir environment variable

yes.

> though it probably makes more sense for the directory creation to be handled by the extension as in most cases this environment var will be unused anyway

I agree. But as of now, the extension doesn't handle that part. 
We can discuss this with upstream by opening an issue.

VihasMakwana:
@strawgate regarding the name, I believe we should switch to `OTEL_STORAGE_DIR` or `OTEL_REGISTRY_DIR` for the env variable. Let me know if you any suggestions.

strawgate:
> extensions: 
>    file_storage: 
>      directory


otel_ext_file_storage_dir:

internal/pkg/otel/run.go
```diff
@@ -80,3 +86,23 @@ func newSettings(version string, configPaths []string) (*otelcol.CollectorSettin
...
+		os.Setenv("STORAGE_DIR", storageDir)
+	}
+	if _, err := os.Stat(storageDir); err == nil {
+		// directory exists
+		return nil
+	} else if os.IsNotExist(err) {
+		return os.MkdirAll(storageDir, 0755)
+	} else {
+		return fmt.Errorf("error stating %s: %w", storageDir, err)
+	}
```
pchila, line 107: We can maybe restructure this block a bit
```suggestion

	err := os.Stat(storageDir);
	if errors.Is(err, fs.ErrNotExist) {
		// create directory if it doesn't exist
		return os.MkdirAll(storageDir, 0755)
	} 
	
	if err != nil  {
		// we have a generic error
		return fmt.Errorf("error stating %s: %w", storageDir, err)
	}
	
	return nil
```

internal/pkg/otel/run.go
```diff
@@ -33,6 +35,10 @@ func Run(ctx context.Context, stop chan bool, configFiles []string) error {
 		return err
 	}
 
+	if err := ensureRegistryExists(); err != nil {
+		return fmt.Errorf("error while creating registry: %w", err)
+	}
+
```
pchila, line 41: Shouldn't we check and inject the `storage` extension settings here? It may be conditional to a cli flag or some env variable or something else but I doubt very man people are gonna run the elastic otel collector with our sample config file and without that this change ensures only that a given dir exists/is created...

pchila (commented):
I see that we are ensuring the existence of a storage directory here but I am not sure that it's enough to enable persistence by default, without some default receiver/exporter/extensions/service configuration being injected.

Obviously the choice depends on what kind of compatibility we want to keep with upstream OTel collector we want to keep but in my view enabling persistence by default means that I can omit storage extensions settings in my config file and then still benefit from persistence by default but maybe I misunderstood the purpose of the change

andrzej-stencel:
This pull request is named "enable persistence by default", but it does not do what the description says.

I believe what this change is intending to achieve is "make sure the directory used by the File Storage extension exists", but I it does not actually do that either.

What this change actually does is:
- look at an arbitrary env var `STORAGE_DIR` (regardless of whether the env var is used in the collector's config or not),
- default this env var to a specific path if unset,
- create a directory at that path if it does not exist.

I see the following problems with this approach:

1. It creates a directory in an arbitrary location, regardless of whether this location is being used by the collector for persistence. This may be surprising and cause unwanted side effects for users that don't use the `STORAGE_DIR` env var in their File Storage extension configuration.
2. The collector fails to start if that arbitrary directory does not exist and cannot be created. This may be surprising to users that don't use File Storage extension and are running e.g. in a container with read-only file system.

A much better way to achieve the intention "make sure the directory used by the File Storage extension exists" would be to add an option to the upstream File Storage extension like `create_directory`, which, when set to `true`, would create the directory that is actually used by the extension, for each extension that is defined (and create no directory if no extension is defined). I suggest to create an issue in Contrib about it.

VihasMakwana:
Thank you for your feedback!

I agree that handling directory creation should ideally be managed upstream. I've submitted an enhancement request at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34939 to address this.

In the meantime, my current PR aimed to ensure the storage directory exists by using the `storage_dir` environment variable, as it's featured in our sample configs. However, this approach may lead to issues if the user doesn't include this setting in their otel config.

I propose repurposing the PR to focus solely on enabling `filestorage` in our sample configs. I will then submit a separate PR to address the directory creation flag.

VihasMakwana:
Thank you for your feedback!

I agree that handling directory creation should ideally be managed upstream. I've submitted an enhancement request to address this.

In the meantime, my current PR aimed to ensure the storage directory exists by using the storage_dir environment variable, as it's featured in our sample configs. However, this approach may lead to issues if the user doesn't include this setting in their otel config.

I propose repurposing the PR to focus solely on enabling `filestorage` in our sample configs. I will then submit a separate PR to address the directory creation flag.

internal/pkg/otel/run.go
```diff
@@ -80,3 +86,23 @@ func newSettings(version string, configPaths []string) (*otelcol.CollectorSettin
 		DisableGracefulShutdown: true,
 	}, nil
 }
+
+func ensureRegistryExists() error {
+	storageDir := os.Getenv("STORAGE_DIR")
```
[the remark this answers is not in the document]
cmacknz: I also think this should have a more specific name, we should also discuss if we should just reuse the existing STATE_PATH variable that controls the same thing.

internal/pkg/otel/run.go
```diff
@@ -80,3 +86,23 @@ func newSettings(version string, configPaths []string) (*otelcol.CollectorSettin
...
+	if storageDir == "" {
+
+		// by default use "${path.data}/registry/otelcol" to store offsets
+		storageDir = filepath.Join(paths.Data(), "registry", "otelcol")
```
cmacknz, line 95: I think we probably need some discussion on what our directory structure should be, considering https://github.com/elastic/ingest-dev/issues/3671 and https://github.com/elastic/elastic-agent/issues/5304 for example.

There are also things besides filelog that may want disk persistent, who don't use a registry concept. If we needed more than this, where would they go in the directory structure? Here? What if we decided we needed more than one storage extension?

cmacknz:
Taking a higher level look at this, I don't think we need to introduce the .go changes or the env var quite yet. These are not upstream functionality and if we expose them to users they will get hard to get rid of. 

Fundamentally, lack of persistence for log collection is sub-optimal but not fatal. As long as we are only considering our pure OTel collector distribution (i.e. not Beats receivers yet) we should solve this with documentation and reference examples while making or advocating for upstream changes to make the configuration changes needed simpler. 

+1 to the suggestion in https://github.com/elastic/elastic-agent/pull/5347#issuecomment-2320974215 to add a `create_directory` option to the storage extension, and as discussed elsewhere we should try to make it so that receivers that benefit from storage use it automatically if it is available.

For Beats receivers, we will have to store the Filebeat registry somewhere to maintain parity with what we do today, but we already have a place to put that we could reuse without adding anything new. We may opt to change where this state is stored, but until we have to do that let's avoid introducing any new configuration or env vars unless there is a really strong reason. Adding configuration is easy, it is much harder to take way later so let's not do it until we have to and I don't think we have to yet.

VihasMakwana:
Closing this PR.
Once https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34985 gets merged, I'll update the docs and raise a new PR as per https://github.com/elastic/elastic-agent/pull/5347#issuecomment-2327628042

PR #5352, opened 2024-08-23
[integration tests] Keep work directory if test fails

belimawr:
## What does this PR do?

This commit enables the work directory used by the integration tests framework to be kept in the filesystem if the test fails. The full path of the test directory is printed when the test fails.

## Why is it important?

It helps debugging failing integration tests

## Checklist

## Disruptive User Impact

There is no disruptive user impact because this PR changes our integration tests framework.

1. Add `t.FailNow()` at the end of any integration test
2. Run the integration test you modified with the `-v` flag
3. You will see a log entry like this:
```
fixture.go:242: Temporary directory saved: /tmp/TestEventLogOutputConfiguredViaFleet411647803
```

~~## Related issues~~

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -1205,7 +1205,6 @@ func createTempDir(t *testing.T) string {
 	tempDir, err := os.MkdirTemp("", strings.ReplaceAll(t.Name(), "/", "-"))
 	if err != nil {
 		t.Fatalf("failed to make temp directory: %s", err)
-		t.Logf("Created temp directory %q", tempDir)
 	}
 
 	cleanup := func() {
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -1205,6 +1205,7 @@ func createTempDir(t *testing.T) string {
 	tempDir, err := os.MkdirTemp("", strings.ReplaceAll(t.Name(), "/", "-"))
 	if err != nil {
 		t.Fatalf("failed to make temp directory: %s", err)
+		t.Logf("Created temp directory %q", tempDir)
 	}
 
 	cleanup := func() {
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -1214,7 +1214,7 @@ func createTempDir(t *testing.T) string {
 				t.Errorf("could not remove temp dir '%s': %s", tempDir, err)
 			}
 		} else {
-			t.Logf("Temporary directory saved: %s", tempDir)
+			t.Logf("Temporary directory %q preserved for investigation/debugging", tempDir)
 		}
 	}
 	t.Cleanup(cleanup)
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -206,7 +206,7 @@ func (f *Fixture) Prepare(ctx context.Context, components ...UsableComponent) er
 	if err != nil {
 		return err
 	}
-	workDir := f.t.TempDir()
+	workDir := createTempDir(f.t)
 	finalDir := filepath.Join(workDir, name)
 	err = ExtractArtifact(f.t, src, workDir)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mauri870:
I really like the idea. The Go project uses a similar flag, [-work](https://pkg.go.dev/cmd/go), to preserve files after a build. I wonder if it makes sense to add a env variable or flag to force this behavior even if a test passes.

pkg/testing/fixture.go
```diff
@@ -1196,6 +1196,31 @@ func performConfigure(ctx context.Context, c client.Client, cfg string, timeout
 	return nil
 }
 
+// createTempDir creates a temporary directory that will be
+// removed after the tests passes. If the test fails, the
+// directory is kept for further investigation.
+//
+// If the test is run with -v and fails the temporary directory is logged
```
mauri870, line 1203: Since you stated that the test must be run with the -v flag you might want to check for [`t.Verbose()`](https://pkg.go.dev/testing#Verbose) explicitly. I don't mind just rephrasing the comment to something like `If the test fails the temporary directory is logged. Use the -v flag to check the directory name.`

pkg/testing/fixture.go
```diff
@@ -1196,6 +1196,31 @@ func performConfigure(ctx context.Context, c client.Client, cfg string, timeout
...
+func createTempDir(t *testing.T) string {
+	tempDir, err := os.MkdirTemp("", strings.ReplaceAll(t.Name(), "/", "-"))
+	if err != nil {
+		t.Fatalf("failed to make temp directory: %s", err)
```
VihasMakwana, line 1207: Instead of `t.Fatalf` here, should we fallback to current implementation? i.e. `return f.t.TempDir()`

mauri870 (approved):

pkg/testing/fixture.go
```diff
@@ -1196,6 +1196,31 @@ func performConfigure(ctx context.Context, c client.Client, cfg string, timeout
...
+	}
```
pchila, line 1208: Add a log at creation so that it's possible to follow along as the test runs (without needing to wait for the end of the test)
```suggestion
		t.Logf("Created temp directory %q", tempDir)
	}
```

pchila (approved):
I think we should log the directory name as soon as it's created (useful in case of debugging or even just running with `-v` )

Looks ok, otherwise

belimawr:
> I really like the idea. The Go project uses a similar flag, [-work](https://pkg.go.dev/cmd/go), to preserve files after a build. I wonder if it makes sense to add a env variable or flag to force this behavior even if a test passes.

I wonder the use case for it, would that be developing new tests?

Maybe I'm just too used to add `t.FailNow()` at the end of a test that's passing but I'm still working on in and want to run things manually.

pkg/testing/fixture.go
belimawr: Because it uses `t.Logf` it is only shown when the `-v` flag is passed, so checking `t.Verbose()` seems redundant here.

Is there a corner case I'm not seeing?

I like the change in the copy, I'll implement it. Thanks!

pkg/testing/fixture.go
```diff
@@ -1196,6 +1196,31 @@ func performConfigure(ctx context.Context, c client.Client, cfg string, timeout
...
+
+	cleanup := func() {
+		if !t.Failed() {
+			if err := os.RemoveAll(tempDir); err != nil {
+				t.Errorf("could not remove temp dir '%s': %s", tempDir, err)
+			}
+		} else {
+			t.Logf("Temporary directory saved: %s", tempDir)
```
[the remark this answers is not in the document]
belimawr: Thanks!

pkg/testing/fixture.go
belimawr: I ended up accepting the suggestion, but I've just realised it will duplicate the message on failures.

Do you think it's that necessary to have it printed if someone wants to follow along? Tests usually last for a few seconds, how useful would that be?

Also, if someone wants to follow along, they can just look for the folder that will appear in `/tmp` and follow from there.

pkg/testing/fixture.go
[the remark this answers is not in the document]
belimawr: To reduce verbosity, also what is the use of knowing the directory/having it registered in the logs if it will be removed after the test?

pkg/testing/fixture.go
mauri870: That is exactly the point I was trying to get to, I think just rephrasing the comment is enough, I also don't see a need to check for t.Verbose explicitly.

mauri870:
> > I really like the idea. The Go project uses a similar flag, [-work](https://pkg.go.dev/cmd/go), to preserve files after a build. I wonder if it makes sense to add a env variable or flag to force this behavior even if a test passes.
> 
> I wonder the use case for it, would that be developing new tests?
> 
> Maybe I'm just too used to add `t.FailNow()` at the end of a test that's passing but I'm still working on in and want to run things manually.

On second thought, I wouldn't worry too much about my suggestion. Keeping things simple is probably best. Preserving the work directory is most useful when a test fails, which you've already addressed. We can always revisit and expand on this if necessary.

VihasMakwana (approved):

belimawr:
Because this PR is failing due to flaky tests that are being fixed by https://github.com/elastic/elastic-agent/pull/5375 and I need this PR to debug https://github.com/elastic/elastic-agent/pull/5375, I'll close this PR and add its changes to https://github.com/elastic/elastic-agent/pull/5375.

Issue #4502, opened 2024-03-29
[Flaky Test]: TestManager_StartStopComponent – panic: test timed out after 10m0s


rdner:
### Failing test case

TestManager_StartStopComponent

### Error message

panic: test timed out after 10m0s

### Build

https://buildkite.com/elastic/elastic-agent/builds/8017#018e8b05-4f91-450f-9c87-df9f94ba7f52

### OS

Mac

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: pkg/component/runtime  (0.00s)
panic: test timed out after 10m0s
running tests:
	TestFakeInputSuite (9m57s)
	TestFakeInputSuite/TestManager_StartStopComponent (6m20s)
goroutine 640896 [running]:
testing.(*M).startAlarm.func1()
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/testing/testing.go:2259 +0x259
created by time.goFunc
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/time/sleep.go:176 +0x45
goroutine 1 [chan receive, 10 minutes]:
testing.(*T).Run(0xc000092b60, {0x10119bfab, 0x12}, 0x1011e6b18)
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/testing/testing.go:1649 +0x871
testing.runTests.func1(0x0?)
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/testing/testing.go:2054 +0x85
testing.tRunner(0xc000092b60, 0xc000797ae8)
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/testing/testing.go:1595 +0x262
testing.runTests(0xc0002cc140?, {0x101ac2060, 0xe, 0xe}, {0x1000151f5?, 0xd0?, 0x101ad6740?})
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/testing/testing.go:2052 +0x8ae
testing.(*M).Run(0xc0002cc140)
	/Users/admin/.gvm/versions/go1.21.8.darwin.amd64/src/testing/testing.go:1925 +0xcd8
main.main()
	_testmain.go:107 +0x2e5
goroutine 9 [select]:
go.elastic.co/apm.(*Tracer).loop(0xc0002f8640)
	/Users/admin/go/pkg/mod/go.elastic.co/apm@v1.15.0/tracer.go:975 +0x15d6
[...]

PR #5353, opened 2024-08-23
TestManager_StartStopComponent add test logs, alternate manager ready signal

michel-laterman:
## What does this PR do?

Add some output to make it easier to see wheree the flakey `TestManager_StartStopComponent` test is getting stuck.
Use a chan instead of an atomic bool to signal when the manager is ready for tests as it is more idiomatic.

## Related issues

- Relates #4502

### DIFF
--- a/pkg/component/runtime/manager.go
+++ b/pkg/component/runtime/manager.go
@@ -16,7 +16,6 @@ import (
 	"path"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/gofrs/uuid/v5"
@@ -109,7 +108,7 @@ type Manager struct {
 	grpcConfig *configuration.GRPCConfig
 
 	// Set when the RPC server is ready to receive requests, for use by tests.
-	serverReady *atomic.Bool
+	serverReady chan struct{}
 
 	// updateChan forwards component model updates from the public Update method
 	// to the internal run loop.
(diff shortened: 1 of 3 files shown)


blakerouse:
@michel-laterman Are you looking to get this merged as is or are you looking to keep running this to try and reproduce?

michel-laterman:
@blakerouse, I can't reproduce at the moment so I think we should just merge and see if it reproduces on main or on other PRs

pkg/component/runtime/manager.go
```diff
@@ -192,7 +191,7 @@ func NewManager(
 		errCh:          make(chan error),
 		monitor:        monitor,
 		grpcConfig:     grpcConfig,
-		serverReady:    &atomic.Bool{},
+		serverReady:    make(chan struct{}, 1),
```
pchila, line 194: Why does this need to be a buffered channel? 
If it's only using as a signal for when we entered `serverLoop`, a boolean seemed more appropriate to represent a running state instead of something that can be read only once.

Moreover if it's only used in testing, isn't there a better way to know if the manager itself is running without always allocating a channel and always writing to it also for production code?

blakerouse: @michel-laterman @pchila has a point about being able to check it once. This change does prevent the `waitForReady` from being able to be called more than once.

michel-laterman: I'll change to closing an unbuffered channel on ready; this is the same approach we take in fleet-server with the policy monitor

pchila (approved):

michel-laterman:
buildkite test this

Issue #5159, opened 2024-07-17
[Flaky Test]: TestEventLogOutputConfiguredViaFleet – could not find event log file


rdner:
### Failing test case

TestEventLogOutputConfiguredViaFleet

### Error message

could not find event log file

### Build

* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1274
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1309
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1333
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1341
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1346
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/1373
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/2086#01917b0e-553b-4fd2-a2d1-4f224c11c6ed
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/2088#01917b0e-e4e4-4d20-af8a-2bd4b206b9e1

### OS

Linux

### Stacktrace and notes

```markdown
=== RUN   TestEventLogOutputConfiguredViaFleet
    container_cmd_test.go:75: Creating enrollment API key...
    fetcher.go:95: Using existing artifact elastic-agent-8.15.0-SNAPSHOT-linux-x86_64.tar.gz
    fixture.go:282: Extracting artifact elastic-agent-8.15.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestEventLogOutputConfiguredViaFleet188574346/001
    fixture.go:300: Completed extraction of artifact elastic-agent-8.15.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestEventLogOutputConfiguredViaFleet188574346/001
    fixture.go:906: Components were not modified from the fetched artifact
    fixture.go:657: >> running binary with: [/tmp/TestEventLogOutputConfiguredViaFleet188574346/001/elastic-agent-8.15.0-SNAPSHOT-linux-x86_64/elastic-agent status --output json]
    fixture.go:657: >> running binary with: [/tmp/TestEventLogOutputConfiguredViaFleet188574346/001/elastic-agent-8.15.0-SNAPSHOT-linux-x86_64/elastic-agent status --output json]
[...]

PR #5361, opened 2024-08-26
Fix TestEnvWithDefault and TestContainerCMDEventToStderr

belimawr:
## What does this PR do?

This PR fixes two tests:

TestEnvWithDefault
Now TestEnvWithDefault unsets all environment variables it sets,
allowing it to be run multiple times using `-count`.

TestContainerCMDEventToStderr
TestContainerCMDEventToStderr did not call agentFixture.Prepare early
enough leading to an empty STATE_PATH env var, so all state
information was in /usr/share/elastic-agent, which could make
subsequent tests to fail because they could read
/usr/share/elastic-agent/state/container-paths.yml and use a state
path different than the one set in the test. 

## Why is it important?

It fixes tests and prevents TestEventLogOutputConfiguredViaFleet from being flaky.

## Checklist

## Disruptive User Impact

There is no disruptive user impact because this PR only updates tests.

### Testing the fix in TestContainerCMDEventToStderr
1. Run the test
```
SNAPSHOT=true TEST_PLATFORMS="linux/amd64" mage integration:single TestContainerCMDEventToStderr
```
2. SSH into the VM
```
$(mage integration:ssh)
```
3. Ensure `/usr/share/elastic-agent` does not exist

### TestEnvWithDefault
```
cd internal/pkg/agent/cmd
go test -count=10 -v -run=TestEnvWithDefault
```
Make sure the test succeeds

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5159

## Questions to ask yourself

- ...


andrzej-stencel (approved):

testing/integration/container_cmd_test.go
```diff
@@ -327,6 +327,9 @@ func TestContainerCMDEventToStderr(t *testing.T) {
 	agentFixture, err := define.NewFixtureFromLocalBuild(t, define.Version())
 	require.NoError(t, err)
 
+	// We call agentFixture.Prepare to set the workdir
+	require.NoError(t, agentFixture.Prepare(ctx))
```
pchila, line 331: nit: add a little bit of context to the `require` (not mandatory but removes the need to look at the source to see what the test was doing when the unexpected error occurred)
```suggestion
	require.NoError(t, agentFixture.Prepare(ctx), "failed preparing agent fixture")
```

pchila (approved):

belimawr:
This PR is failing due to flaky tests, https://github.com/elastic/elastic-agent/pull/5375 should fix them.

PR #5366, opened 2024-08-27
Fix flaky test `TestActionDispatcher/Dispatch_multiple_events_returns_one_error`

### LINKED ISSUE #4937: [Flaky Test]: TestActionDispatcher/Dispatch_multiples_events_returns_one_error – Expected error
### Failing test case

TestActionDispatcher/Dispatch_multiples_events_returns_one_error

### Error message

dispatcher_test.go:393: Expected error

### Build

https://buildkite.com/elastic/elastic-agent/builds/9615#0190110f-0b1c-45ba-81d9-c1406e137e09

### OS

Mac

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: internal/pkg/agent/application/dispatcher TestActionDispatcher/Dispatch_multiples_events_returns_one_error (0.52s)
    dispatcher_test.go:393: Expected error
=== FAIL: internal/pkg/agent/application/dispatcher TestActionDispatcher (1.52s)
```

andrzej-stencel:
## What does this PR do?

Fixes the flaky test `TestActionDispatcher/Dispatch_multiple_events_returns_one_error`.

Here's my understanding of the test's intention:

```
When I create a dispatcher with a default handler
And I dispatch two actions that result in errors from handler
Then the default handler is called twice
And one error is placed on the dispatcher's error channel
```

I have introduced the following changes in the test's implementation:

- Removed the `time.Sleep` calls in favor of waiting on the completion of the `Dispatch` method.
- Changed the `Handle` method mocks to return two errors instead of one, and assert that only the second error is put on the errors channel.
- Narrowed down `mock.Anything` parameters in the `Handle` method mocks to specific variables `dispatchCtx`, `action1|2`, `ack`.

## Why is it important?

We don't want flaky tests.

## Checklist

## Related issues

- Closes #4937


pchila (approved):
LGTM

internal/pkg/agent/application/dispatcher/dispatcher_test.go
```diff
@@ -360,43 +360,49 @@ func TestActionDispatcher(t *testing.T) {
...
+			dispatchCompleted <- struct{}{}
+		}()
+
+		// First, assert that the Dispatch method puts one error - the second one - on the error channel.
 		select {
-		case <-d.Errors():
-		default:
+		case err := <-d.Errors():
+			assert.EqualError(t, err, "second error")
+		case <-dispatchCompleted:
```
blakerouse, line 396: If this never happens and nothing ever comes over d.Errors() doesn't this mean this test is now stuck?

Should a timeout be added to ensure that it doesn't deadlock if it doesn't work correctly.

internal/pkg/agent/application/dispatcher/dispatcher_test.go
```diff
@@ -360,43 +360,49 @@ func TestActionDispatcher(t *testing.T) {
...
 		}
-		time.Sleep(time.Millisecond * 200)
+
+		// Second, assert that the Dispatch method completes without putting anything else on the error channel.
 		select {
 		case <-d.Errors():
 			t.Fatal(err)
-		default:
+		case <-dispatchCompleted:
+			// Expecting the dispatch to complete.
```
blakerouse, line 405: Same as above.

internal/pkg/agent/application/dispatcher/dispatcher_test.go
andrzej-stencel: My assumption was the test timeout would ultimately break this wait. Is this incorrect?

blakerouse: Not incorrect, but also a very long time.

Issue #5092, opened 2024-07-09
[OTel] Additional components for the Elastic Distro


ChrsMark:
**Describe the enhancement:**

K8s is one of the most used envs according to https://opentelemetry.io/blog/2024/otel-collector-survey/#deployment-scale-and-environment, hence it makes sense to consult what the community distro for k8s includes.
We can evaluate the list of the k8s-distro components and consider including them as well in the Elastic distro: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-k8s/manifest.yaml

/cc @elastic/otel-devs 

**Describe a specific use case for the enhancement or feature:**

**What is the definition of done?**

_Taken from https://github.com/elastic/elastic-agent/issues/5092#issuecomment-2225517430_:

* [ ] The following additional OTel Collector components are included in the Elastic Distribution of the OTel Collector, i.e. Elastic Agent:
   * [ ] `prometheus` receiver
   * [ ] `memory_limiter` processor
   * [ ] `health_check` extension
* [ ] The `heml install ...` command mentioned in https://github.com/elastic/elastic-agent/issues/5092#issuecomment-2225517430 succeds.

ycombinator:
@strawgate @AlexanderWert What is the priority for this issue? Which release do we want to target it to?

AlexanderWert:
> We can evaluate the list of the k8s-distro components and consider including them as well in the Elastic distro

Looking at the list of components it seems like there are still many components that sound less relevant for our distro (like spanmetricsconnector, servicegraphconnector, etc.). As you wrote, I think we need to evaluate and come up with our own list inspired by that list.

> @strawgate @AlexanderWert What is the priority for this issue? Which release do we want to target it to?

I think 8.15.2 would be reasonable for that, since that's when we plan to have a broader k8s integration for OTel.

ycombinator:
> Looking at the list of components it seems like there are still many components that sound less relevant for our distro (like spanmetricsconnector, servicegraphconnector, etc.). As you wrote, I think we need to evaluate and come up with our own list inspired by that list.

What would the criteria be for deciding whether to include a component from the [list](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-k8s/manifest.yaml) in our distro vs. not include it?

ChrsMark:
[...]

PR #5369, opened 2024-08-28
Add `health_check` extension to bundled `otel.yml` file

ycombinator:
## What does this PR do?

This PR adds the `health_check` extension to the `otel.yml` file that's bundled with the Elastic Agent package.

## Why is it important?

Without this addition, if a user tries to install the `open-telemetry/opentelemetry-collector` Helm chart with Elastic's distribution of the OTel Collector (i.e. Elastic Agent), the installation eventually fails.  Specifically, as noted by @andrzej-stencel in https://github.com/elastic/elastic-agent/issues/5092#issuecomment-2309613481:

> One problem I'm seeing when running the chart with this configuration is that the Health Check extension is not used in the `otel.yml` file, which results in Kubernetes restarting the collector container and eventually resulting in a `CrashLoopBackOff`. We should add the `health_check` extension to `service::extensions` in the `otel.yml` file that's bundled with the image.
 
## Checklist

- [ ] ~My code follows the style guidelines of this project~

## Disruptive User Impact

If users are already running something that's listening on `localhost:13133`, there will be a port conflict.

See https://github.com/elastic/elastic-agent/issues/5092#issuecomment-2309509776 and https://github.com/elastic/elastic-agent/issues/5092#issuecomment-2309613481

### DIFF
--- a/otel.yml
+++ b/otel.yml
@@ -23,7 +23,7 @@ extensions:
     spike_limit_mib: 180
   health_check:
     endpoint: "localhost:13133"
-    path: "/liveness"
+    path: "/health/status"
 
 service:
   extensions: [health_check, memory_limiter]
--- a/otel.yml
+++ b/otel.yml
@@ -21,6 +21,9 @@ extensions:
     check_interval: 1s
     limit_mib: 700
     spike_limit_mib: 180
+  health_check:
+    endpoint: "localhost:13133"
+    path: "/health/status"
 
 service:
   extensions: [memory_limiter]
--- a/otel.yml
+++ b/otel.yml
@@ -23,7 +23,7 @@ extensions:
     spike_limit_mib: 180
   health_check:
     endpoint: "localhost:13133"
-    path: "/health/status"
+    path: "/liveness"
 
 service:
   extensions: [health_check, memory_limiter]
--- a/otel.yml
+++ b/otel.yml
@@ -26,7 +26,7 @@ extensions:
     path: "/health/status"
 
 service:
-  extensions: [memory_limiter]
+  extensions: [health_check, memory_limiter]
   pipelines:
     logs:
       receivers: [filelog]
--- a/changelog/fragments/1724804905-otel-health-check.yaml
+++ b/changelog/fragments/1724804905-otel-health-check.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Add health_check extension to otel.yml file bundled with Elastic Agent package
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/5369
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234


otel.yml
```diff
@@ -21,6 +21,9 @@ extensions:
     check_interval: 1s
     limit_mib: 700
     spike_limit_mib: 180
+  health_check:
+    endpoint: "localhost:13133"
```
ycombinator, line 25: I chose this value based on the documentation in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckextension/README.md#health-check.  I'm happy to change it to another value if others have a better suggestion.

otel.yml
```diff
@@ -21,6 +21,9 @@ extensions:
...
+    path: "/health/status"
```
ycombinator, line 26: The default value here, according to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckextension/README.md#health-check, is `/`.  But I think it's better to use a more specific path for the health check endpoint.  Again, happy to change this if others have a better suggestion.

otel.yml
```diff
@@ -21,6 +21,9 @@ extensions:
...
 
 service:
   extensions: [memory_limiter]
```
rogercoll, line 29: Should we add it in the pipeline too?

```suggestion
  extensions: [health_check, memory_limiter]
```

otel.yml
rogercoll: Maybe `/liveness`? In that sense, we will be more aligned to the `elastic-agent` health check endpoint https://github.com/elastic/elastic-agent/pull/4499

otel.yml
ycombinator: Yes, duh 🤦 - thanks for catching!

otel.yml
ycombinator: Done in 6f2306c6c8.  Thanks for the suggestion.

rogercoll: Just to give more context on the Helm configuration, any change on the default health_check configuration will need to be overrided in the `liveness and readinesss` proves of the Helm Chart config too:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L395

```yaml
livenessProbe:
  httpGet:
    port: 13133
    path: /health/status
readinessProbe:
  httpGet:
    port: 13133
    path: /health/status
```

rogercoll:
Sounds good to me on adding the `health_check` extension in the bundled `otel.yml`. :+1: 

> Without this addition, if a user tries to install the open-telemetry/opentelemetry-collector Helm chart with Elastic's distribution of the OTel Collector (i.e. Elastic Agent), the installation eventually fails.

I think that this will only happen if the user tries to use the `otel.yml` file as it overrides the configuration set by the Chart:

```
$ kubectl describe pod  elastic-otelcol-opentelemetry-collector-agent-d9s4k
...
Command:
      //usr/share/elastic-agent/elastic-agent
    Args:
      --config=/conf/relay.yaml
      otel
      --config=otel.yml
```

Note that the Chart added another [default config](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector#default-configuration) `--config=/conf/relay.yaml` (which contains the health_check extension). A workaround is to not use the `otel.yml` config and rely on the one configured by the Chart:

```
helm install elastic-otelcol open-telemetry/opentelemetry-collector --version=0.102.1 --set mode=daemonset --set image.repository="docker.elastic.co/beats/elastic-agent" --set image.tag="8.16.0-SNAPSHOT" --set image.pullPolicy=Always --set command.name="/usr/share/elastic-agent/elastic-agent" --set command.extraArgs="{otel}"
```

The recommened approach to provide a custom configuration for the collector is to manully create a [configuration configMap](https://github.com/elastic/opentelemetry-demo/blob/main/kubernetes/elastic-helm/deployment.yaml#L63). There is a PR upstream that will allow overriding the default configuration with a config option: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1301

otel.yml
ycombinator: Thanks @rogercoll.  

Does this mean we will need to specify additional `--set` arguments to `helm chart install` for the configuration keys you mentioned above, assuming we are using the bundled `otel.yml` file via `--set command.name="/usr/share/elastic-agent/elastic-agent" --set command.extraArgs="{otel, --config=otel.yml}"`?  

So something like `--set livenessProbe.httpGet.path=/liveness --set readinessProbe.httpGet.path=/liveness`?

Sorry if this is a basic question but I'm a n00b when it comes to Helm.

rogercoll: Correct, you will need to override the probes path with the --set flag. But as I was explaining in this [comment](https://github.com/elastic/elastic-agent/pull/5369#issuecomment-2314505406), the upstream Helm Charts override the arguments launched within the container and specify their custom configuration args. That means that the `otel.yml` file will not be taken into consideration and a custom configuration should be provided with a K8s config map.

I am not 100% sure about changing the default `health_check` path in the sample configuration, maybe @elastic/elastic-agent-control-plane can share their feedback based on the current elastic-agent health endpoint.

rogercoll (approved):

otel.yml
ycombinator: @blakerouse @andrzej-stencel @swiatekm do you have any opinions here?

blakerouse: The elastic-agent uses `/liveness` as the default endpoint for its own healthcheck. I don't know if we want to use the same for OTEL, as that would change the behavior of the default collector and probably don't want to do that.

I would like to also see v2 health check being used at some point, as it does a much better job of actually providing the true health of the collector.

ycombinator: Thanks @blakerouse. I will revert the `path` here in this PR to `/health/status` again for the reason you mentioned. 

Agreed on switching to healthcheckv2 in a future PR :)

blakerouse (approved):
Looks good, thanks for the updating the path.

PR #5373, opened 2024-08-28
Fix mage integration:auth target

michel-laterman:
`mage integration:auth` will exec `gcloud config get-value ...`

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -2721,7 +2721,7 @@ func authGCP(ctx context.Context) error {
 	// We only need the service account token to exist.
 	tokenPath, ok, err := getGCEServiceTokenPath()
 	if err != nil {
-		return err
+		return fmt.Errorf("local token lookup failed: %w", err)
 	}
 	if ok {
 		// exists, so nothing to do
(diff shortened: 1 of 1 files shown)


magefile.go
```diff
@@ -2792,7 +2792,7 @@ func authGCP(ctx context.Context) error {
 	}
 
 	// Check the authenticated account's project
-	cmd = exec.CommandContext(ctx, cliName, "config", "get", "core/project")
+	cmd = exec.CommandContext(ctx, cliName, "config", "get-value", "core/project")
```
ycombinator, line 2795: I wonder why this hasn't failed for others on the team.  @michel-laterman What version of the `gcloud` CLI are you using?  For comparison sake, here's what I'm using:

```
$ gcloud version
Google Cloud SDK 477.0.0
bq 2.1.4
core 2024.05.17
gcloud-crc32c 1.0.0
gke-gcloud-auth-plugin 0.5.8
gsutil 5.29
```

ycombinator: If it turns out that this change is due to a difference in the version of `gcloud` we should either make this change conditional on the version or document the suggested `gcloud` CLI version constraints in https://github.com/elastic/elastic-agent/blob/main/docs/test-framework-dev-guide.md#dependencies.

michel-laterman:
my glcoud was outdated

Issue #4976, opened 2024-06-21
[Windows] Service startup failing on CI with otel dependencies linked


michalpristas:
At the moment, Agent is unable to run in OTel mode on Windows — support for Windows was explicitly removed by adding build constraints as part of https://github.com/elastic/elastic-agent/pull/4908.   If we remove these constraints we see failures in CI like so:
`failed to start service (Elastic Agent): The service did not respond to the start or control request in a timely fashion.`

One of the reason may be late reaction to the control events from a service manager and invocation of `service.ProcessWindowsControlEvents` as described in #4971 

Failures are present even if we [move this to init function](https://github.com/elastic/elastic-agent/pull/4952) of alphabetically very early dependencies in main. 
Timing here is also tricky, as dependencies get initialized before package linking them. so calling something in init does not mean it's being called first. 

Event Log for elastic agent seems empty. Occasionally in linked draft above we see
`24304 Jun 21 05:05  Error       Elastic Agent                   1 The description for Event ID '1' in Source 'Ela...`
that could uncover some failures, but this is not present everywhere. But as i had this timeboxed i stopped here

https://techcommunity.microsoft.com/t5/iis-support-blog/the-description-for-event-id-1-from-source-getaccess-iis-runtime/ba-p/3744637

michalpristas:
cc @leehinman

leehinman:
Note to self https://pkg.go.dev/runtime the `inittrace=1` option should give insight into where time is being spent.

leehinman:
So I'm not entirely sure it is `init` that is causing the problem.  I set GODEBUG=inittrace=1.  And did `elastic-agent run` under Windows 2016.  

### Version before [4908](https://github.com/elastic/elastic-agent/pull/4908)

This took 406ms to get through init
```
tail -n 1 before_k8_4ccdd0.log
init github.com/elastic/elastic-agent/internal/pkg/agent/cmd @406 ms, 0 ms clock, 18208 bytes, 123 allocs
```

And one package took up over half that time, the `k8s.io/api/extensions/v1beta1` package

```
awk '$5 > 0 {print $5, $2}' before_k8_4ccdd0.log | sort -rn
244 k8s.io/api/extensions/v1beta1
18 go.elastic.co/apm
12 github.com/aws/aws-sdk-go/aws/endpoints
10 github.com/jedib0t/go-pretty/v6/text
8.6 github.com/shirou/gopsutil/process
[...]

PR #5382, opened 2024-08-29
WIP: otelcol: use ocb to autogenerate source files

mauri870:
## What does this PR do?

This PR uses the [OpenTelemetry collector builder](https://pkg.go.dev/go.opentelemetry.io/collector/cmd/builder) (ocb for short) to automate the generation of collector source files.

## Why is it important?

Updating otel collector files manually can be time consuming and error prone, we can let ocb generate the source files.


internal/pkg/otel/builder-config.yaml
```diff
@@ -0,0 +1,43 @@
+dist:
+  name: otelcol
+  description: Elastic Distribution of OpenTelemetry Collector
+  module: otel
+  output_path: .
+
+extensions:
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckextension v0.107.0
```
mauri870, line 8: Unfortunately, we can't manage versions in this file and sync them with a go.mod file, since the dependencies are defined in the elastic-agent go.mod located at the root of the repository. We could consider using Go workspaces and moving this code to `${REPO_ROOT/otel}`, for example, so it can be its own module with its own go.mod file. This approach would also clean up the otelcol dependencies in the root go.mod file.

internal/pkg/otel/Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: all
+
+all:
+	go run go.opentelemetry.io/collector/cmd/builder@v0.107.0 --skip-compilation --skip-get-modules --config builder-config.yaml
+	@rm main.go main_others.go main_windows.go go.mod
```
mauri870, line 5: This is a bit hacky since ocb expects to build a final binary. Having a otel collector embeded into another binary seems to be quite useful, I wonder if the folks at OpenTelemetry would be willing to support a ocb mode to build a library instead of a final binary.

internal/pkg/otel/Makefile
```diff
@@ -0,0 +1,6 @@
...
+	sed -i 's/package main/\/\/go:build \!windows\npackage otel/g' components.go
```
mauri870, line 6: I haven't looked into why there's a `!windows` build flag, but I would like to understand the reason behind it.

michalpristas: Having otel dependencies breaks agent on startup as a Windows service. Lee is working on finding out what's causing this precisely

internal/pkg/otel/Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: all
```
michalpristas, line 1: Make sure you include this in mage update target at root.
It would be nice if mage update could regenerate sources. Update go mod versions and then update readme file as it does today.

internal/pkg/otel/builder-config.yaml
ycombinator: We have dependabot currently bumping up the OTel dependency versions in `go.mod`: https://github.com/elastic/elastic-agent/blob/main/.github/dependabot.yml.  Perhaps we can add a new step or job to https://github.com/elastic/elastic-agent/blob/main/.github/workflows/post-dependabot.yml to do some parsing of the `go.mod` file and sync the versions in this file with the parsed results?

internal/pkg/otel/Makefile
pierrehilbert: More details here: https://github.com/elastic/elastic-agent/issues/4976

andrzej-stencel:
In general, I like the idea of keeping OTel dependencies in a canonical Builder config Yaml file. For engineers oriented around the OpenTelemetry Collector, this is the idiomatic way to understand what components are included.

However, given that the Elastic Agent is not simply an OTel Collector binary built by the OTel Collector Builder, but rather a binary that incorporates OTel Collector libraries into an existing code base, and given the current feature set of the Builder, my opinion is that the benefit of declaring OTel dependencies in Builder config file does not outweigh the added complexity.

Introducing this approach adds an additional level of abstraction, which affects all engineers working on the Elastic Agent code (Elasticians or not), some of them possibly not knowing anything about OTel Collector or the OTel Collector Builder. It creates new problems like "do we maintain a separate go.mod file for the OTel dependencies" or "how do we keep the OTel dependencies up to date".

The big advantage of the current approach is its simplicity. It is based on regular Go imports and nothing else. This makes it simpler to integrate with tools like Dependabot for dependency upgrades.

My recommendation is to leave the current approach unchanged for now for the sake of simplicity and revisit this decision when circumstances change.

ycombinator:
Thanks for weighing in, @andrzej-stencel.  You raise a good point about the added complexity and additional level of abstraction that may not be obvious to all contributors.  Unless others have strong opinions for favoring the approach in this PR, I'm OK with sticking with our current approach which is not as declarative as the one in this PR but is simpler to understand.

mauri870:
Just an update: I ended up opening an [issue](https://www.github.com/open-telemetry/opentelemetry-collector/issues/11059) upstream to suggest adding support for OCB library code generation. I don't expect this to be considered before the 1.0 release of the builder, but it is a valid feature request.

The changes proposed here aren't worth implementing, as they add unnecessary complexity compared to the current approach and require workarounds to bypass OCB's limitations.

With that in mind, I’m closing this for now.

PR #5385, opened 2024-08-29
Use buildkite test collector plugins for flaky tests analysis

jlind23:
Closes https://github.com/elastic/ingest-dev/issues/3888

Automate flaky tests analysis by using buildkite test collector plugin

- [x] This automation should only run on main


.buildkite/hooks/pre-exit
```diff
@@ -1,34 +0,0 @@
...
-  fi
-fi
-
-if [ -n "$TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE" ]; then
-  if test -f "$TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE"; then
-    rm $TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE
-  fi
-fi
-
-source .buildkite/scripts/unset-secrets.sh
```
alexsapran, line 29: This is what is caused the problem with the secret. Lets revert the file and remove this line only.

Because on like 13 we also perform clean up of the resources.

alexsapran (approved):
💯

jlind23:
Test failing is unrelated - https://github.com/elastic/elastic-agent/pull/5399

Issue #4818, opened 2024-05-24
[Integration test] Agent, running in unprivileged mode, can run the `nginx` integration


ycombinator:
We need to have confidence that Agent, running in `--unprivileged` mode, can successfully collect data for an integration that doesn't require privileged access to the system, e.g. the `nginx` integration.  

## Definition of done
* [ ] An integration test that installs an Agent in `--unprivileged` mode, uses a policy that contains the `nginx` integration, and verifies that data is being collected for this integration without any errors.

VihasMakwana:
@ycombinator I'm currently [following the guide](https://www.elastic.co/guide/en/integrations-developer/current/index.html) to test the integrations locally. Nginx integration already includes [system tests](https://github.com/elastic/integrations/tree/main/packages/nginx/data_stream/stubstatus/_dev/test/system) with [stubstatus](https://github.com/elastic/integrations/blob/main/packages/nginx/_dev/deploy/docker/nginx.conf) enabled.
I don't think we need separate tests in integrations repo as the [manifest.yaml](https://github.com/elastic/integrations/blob/main/packages/nginx/manifest.yml) doesn't require the user to be a root.

These tests are already present in `elastic/integrations` repository and they're run in unprivileged mode. https://github.com/elastic/elastic-agent/pull/5398#discussion_r1744513073

ycombinator:
@VihasMakwana I agree, now that we've switched to Wolfi-based images for system testing the integrations.  Closing this issue.

PR #5398, opened 2024-09-02
[testing/integration] - Add test cases covering unprivileged mode for nginx.stubstatus metricset

VihasMakwana:
- Enhancement

## What does this PR do?

- This PR adds test cases covering `unprivileged` mode for nginx metrics integration

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/4818

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/nginx_unprivileged_test.go
+++ b/testing/integration/nginx_unprivileged_test.go
@@ -21,30 +21,6 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
-// var defaultTextCfgUnprivileged = `
-// outputs:
-//   default:
-//     type: elasticsearch
-//     hosts: [127.0.0.1:9200]
-//     api_key: "example-key"
-//     preset: balanced
-//     allow_older_versions: true
-
-// inputs:
-//   - type: nginx/metrics
-//     id: unique-system-metrics-input
-//     data_stream.namespace: default
-//     use_output: default
-//     streams:
-//       - metricsets:
-//         - stubstatus
-//         data_stream.dataset: nginx.stubstatus
-//         data_stream.type: metrics
-
-// agent.logging.level: debug
-// agent.logging.to_stderr: true
-// `
-
 var nginxStatusModule string = `
 
 server {
@@ -75,7 +51,7 @@ server {
 `
 
 // ExtendedRunner is the main test runner
-type UnprivilegedRunner struct {
+type NginxUnprivilegedRunner struct {
 	suite.Suite
 	info                   *define.Info
 	agentFixture           *atesting.Fixture
@@ -93,7 +69,7 @@ func TestNginxUnprivileged(t *testing.T) {
 		Sudo:  true,  // requires Agent installation
 	})
 
-	runner := &UnprivilegedRunner{
+	runner := &NginxUnprivilegedRunner{
 		info:                   info,
 		healthCheckTime:        time.Minute * 5,
 		healthCheckRefreshTime: time.Second * 5,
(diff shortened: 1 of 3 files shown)


testing/integration/nginx_unprivileged_test.go
```diff
@@ -0,0 +1,164 @@
...
+	compDebugName := ""
+	require.Eventually(runner.T(), func() bool {
+		allHealthy := true
+		status, err := runner.agentFixture.ExecStatus(ctx)
+		if err != nil {
+			runner.T().Logf("agent status returned an error: %v", err)
+			return false
+		}
+
+		for _, comp := range status.Components {
```
blakerouse, line 149: Possible to be healthy without any components, should check that components list is not empty and if it is mark it not healthy.

testing/integration/nginx_unprivileged_test.go
```diff
@@ -0,0 +1,164 @@
...
+		healthCheckRefreshTime: time.Second * 5,
+	}
+
+	suite.Run(t, runner)
+}
+
+func (runner *NginxUnprivilegedRunner) SetupSuite() {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "apt-get", "install", "-v", "nginx-full")
```
blakerouse, line 88: This is not distro agnostic. You are making this debian specific, so you need to define this in the `define.Require` above.

See above.

testing/integration/nginx_unprivileged_test.go
```diff
@@ -0,0 +1,164 @@
...
+	healthCheckTime        time.Duration
+	healthCheckRefreshTime time.Duration
+}
+
+func TestNginxUnprivileged(t *testing.T) {
+	info := define.Require(t, define.Requirements{
+		Group: "fleet",
+		Stack: &define.Stack{},
+		OS: []define.OS{
+			{Type: define.Linux},
```
blakerouse, line 70:
```suggestion
			{Type: define.Linux, Distro: "ubuntu"},
```

blakerouse (commented):
I like these tests, but why are we placing this in the Elastic Agent repository and not in the integration repository?

testing/integration/nginx_unprivileged_test.go
cmacknz: It's also possible to be healthy while not yet having read any documents that would require accessing potentially privileged sources.

Just checking health doesn't actually guarantee we tested anything. You need to make sure we ingested data.

Potentially this entire test could be done in https://github.com/elastic/integrations/tree/main/packages/nginx, as long as we have a way to start unprivileged containers. The new Wolfi container would satisfy this by default. 

https://github.com/elastic/integrations/pull/10969 makes the switch to Wolfi. We may get significantly better test coverage by adding this health check to the end of every integration system test, with the advantage that we also won't be exposed to problems in the nginx integration here.

VihasMakwana: > Just checking health doesn't actually guarantee we tested anything. You need to make sure we ingested data.

Sure. Let me add that step.

blakerouse: Yeah I don't think we should have this test in our repository. I think this is ultimately a test for github.com/elastic/integrations repository.

VihasMakwana:
Closing this. See https://github.com/elastic/elastic-agent/issues/4818#issuecomment-2334564088

PR #5400, opened 2024-09-02
use Stringer to get string values for eaclient Unit types

mauri870:
## What does this PR do?

This PR simplifies getting string values for Unit type variants by using the fmt.Stringer impl provided by the elastic-agent-client.

Looking at the implementation in elastic-agent-client, it maps the exact same keys to the same strings, except for the 'unknown' variant, which we currently return as an empty string.

https://github.com/elastic/elastic-agent-client/blob/v7.15.0/pkg/client/unit.go#L91-L110
https://github.com/elastic/elastic-agent-client/blob/v7.15.0/pkg/client/unit.go#L27-L36

## Why is it important?

Just a cleanup. I'm not sure if it is better to map the values ourselves.


blakerouse (approved):
Looks good.

michel-laterman (approved):

PR #5408, opened 2024-09-03
[Flaky Test] TestComponentBuildHashInDiagnostics: save diagnostic and agent status on failure 

AndersonQ:
## What does this PR do?

Saves diagnostic used on TestComponentBuildHashInDiagnostics and prints the agent status output if the test fails

## Why is it important?

The test is flaky, it'll provide more data to investigate why it's failing

## Checklist

- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool]~~(https://github.com/elastic/elastic-agent#changelog)~~

## Disruptive User Impact

-N/A

- N/A
## Related issues

- Relates #5333

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/package_version_test.go
+++ b/testing/integration/package_version_test.go
@@ -131,7 +131,7 @@ func TestComponentBuildHashInDiagnostics(t *testing.T) {
 		5*time.Minute, 10*time.Second,
 		"agent never became healthy. Last status: %v", &stateBuff)
 	t.Cleanup(func() {
-		if t.Failed() {
+		if !t.Failed() {
 			return
 		}
 
--- a/testing/integration/package_version_test.go
+++ b/testing/integration/package_version_test.go
@@ -91,10 +91,11 @@ func TestComponentBuildHashInDiagnostics(t *testing.T) {
 		"failed to install start agent [output: %s]", string(output))
 
 	stateBuff := bytes.Buffer{}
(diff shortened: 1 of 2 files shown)


testing/integration/package_version_test.go
```diff
@@ -130,6 +130,13 @@ func TestComponentBuildHashInDiagnostics(t *testing.T) {
 		allHealthy,
 		5*time.Minute, 10*time.Second,
 		"agent never became healthy. Last status: %v", &stateBuff)
+	t.Cleanup(func() {
+		if t.Failed() {
```
mauri870, line 134: Isn't this supposed to be `!t.Failed() then t.Log`? Otherwise it will print the log if a test passes.

AndersonQ: fixed

blakerouse (approved):
This change is fine and I will approve it at your request as its not really hurting anything.

These changes should really just live in your own branch and you should be able to setup to just run this test to fail.

The integration suite was built specifically for this use-case. To run tests locally from a local machine or a dedicated cloud instance to make it easy to find issues and get results.

You can simply do the following and wait until it fails:

> `TEST_PLATFORMS="windows/amd64" TEST_RUN_UNTIL_FAILURE=true mage integration:single TestComponentBuildHashInDiagnostics`

PR #5418, opened 2024-09-04
test: check condition early for TestHTTPReloadEnableBehavior

kruskall:
## What does this PR do?

we're relying on assert.Eventually with a wait time of 3s we call it twice per test in 4 total tests
minimum test time should be 3*2*4=24s

Update assertion to check the condition before calling eventually

Test time went from 24.101s to 0.264s

## Why is it important?

slow test is bad

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


blakerouse (approved):
Nice!

PR #5419, opened 2024-09-04
fix: remove initial delay on restart

kruskall:
## What does this PR do?

restarting the agent on rollback uses an exponential backoff for retries but the initial wait starts before the first retry causing a 5s delay on every restart/rollback

this is particularly problematic in tests

update the code to use exponential rollback strting from the first retry

TestRollback time went from 36.281s to 6.263s

## Why is it important?

5s delay on restart and slow test

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


blakerouse (approved):
Nice catch!

Issue #5428, opened 2024-09-04
Remove automatic unenrollment after 7 Fleet authentication failures


cmacknz:
- Relates https://github.com/elastic/elastic-agent/issues/5433

Today Elastic Agent will unenroll itself automatically after receiving 7 consecutive 401 responses from Fleet when checking in. This was done to prevent agents that have been forced unenrolled (which revokes their API key) from checking in continuously until they can be re-installed. 

https://github.com/elastic/elastic-agent/blob/590c506aea6f278200d024e65d0bc7e1c8b5238a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L26-L28

https://github.com/elastic/elastic-agent/blob/590c506aea6f278200d024e65d0bc7e1c8b5238a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L360-L363

https://github.com/elastic/elastic-agent/blob/590c506aea6f278200d024e65d0bc7e1c8b5238a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L329-L341

This prevents force unenrolled agents from continuing to contact Fleet Server, but represents an edge case that can be hit in disaster recovery situations. To eliminate the chance that users recovering their cluster need to manually intervene on machines, we should stop unenrolling and instead greatly increase the checkin interval.

The initial proposal is that instead of unenrolling, we should switch to checking in once per hour. A successful checkin must return the agent to its original checkin interval.

elpang1201:
May I know if there are any plan on the release date? As this is troublesome on keep reload all those agents

PR #5438, opened 2024-09-05
Fix state store SetAction panic

### LINKED ISSUE #5434: Agent panics when trying to persist a `nil` action
When the agent tries to auto unenroll due too many authentication failures, the unenroll handler sets the action to `nil` before saving it. However the state store panics when setting a `nil` action.

 - the handler sets the action to `nil`:
   - https://github.com/elastic/elastic-agent/blob/b4877f52a39c53591634394be57e156d79ed57e0/internal/pkg/agent/application/actions/handlers/handler_action_unenroll.go#L86-L89
 - the handler tries to save the action
   - https://github.com/elastic/elastic-agent/blob/b4877f52a39c53591634394be57e156d79ed57e0/internal/pkg/agent/application/actions/handlers/handler_action_unenroll.go#L95-L101
 - the store tries to access properties of a `nil` pointer
   - https://github.com/elastic/elastic-agent/blob/b4877f52a39c53591634394be57e156d79ed57e0/internal/pkg/agent/storage/store/state_store.go#L183-L188

AndersonQ:
The state store SetAction did not correctly cover the case where a nil action is passed as parameter. The unenroll handler might pass a nil action if the unenroll is a auto-unenroll, that means, it does not come from fleet

Also the unenroll handler does checks for a nil state store anymore, it assumes it's valid just as it does for all other dependencies

## What does this PR do?

Fixes the state store panicking when receiving an `nil` action

## Why is it important?

Panic is bad, the agent should not panic
Also the unenroll handler sets an `nil` action on the state store if the unenroll is self-inflicted. The agent auto unenroll after the 7th authentication failure against fleet.

## Checklist

## Disruptive User Impact

- N/A

run the tests added on the PR, all work
keep the tests, rever the changes on the state store, they will panic

## Related issues

- Closes #5434

## Questions to ask yourself

- ...


internal/pkg/agent/storage/store/state_store.go
```diff
@@ -169,6 +170,18 @@ func (s *StateStore) SetAction(a fleetapi.Action) {
 	s.mx.Lock()
 	defer s.mx.Unlock()
 
+	if a == nil || reflect.ValueOf(a).IsNil() {
+		s.log.Debugf("trying to set an nil '%T' action, ignoring the action", a)
```
pchila, line 174: Shouldn't this rather be an error log ? We're never gonna see this in logs collected via diagnostics otherwise...

pchila (commented):
Left a comment about the log level used to signal the error in action store when trying to save a nil action.
Moreover if the unenroll action is generated internally, shouldn't we fix that too somehow ?

internal/pkg/agent/storage/store/state_store.go
AndersonQ: no, we're supposed to ignore `nil`. I'm just fixing the panic, not changing the behaviour . We have another issue to deal with the self-unenroll https://github.com/elastic/elastic-agent/issues/5428. We can address that there

pchila (approved):

PR #5443, opened 2024-09-05
mage: check that links are live in the OTel README

### LINKED ISSUE #5117: Automatically verify links in OTel README file
it looks good, i was wondering if we should do some kind of validation to check remote part exists, but this could be problematic in CI in case if there are traffic filters.

_Originally posted by @michalpristas in https://github.com/elastic/elastic-agent/pull/5087#pullrequestreview-2168896679_
            
The [OTel README file](https://github.com/elastic/elastic-agent/blob/47e87690bf5134aa60c1c6b99ac2b364363ab9fb/internal/pkg/otel/README.md) has links to components that are created automatically by the `mage otel:readme` command. We should have a tool to automatically check that the links are valid (point to something that exists).

We could use https://github.com/tcort/markdown-link-check/ or similar solution for this.

mauri870:
This PR adds a utility based on the `link-patrol` command to check if links in a given file are live. This utility should be reusable so we can use it to validate other files in the repository. 

Also added a check in the `check` and `otel:readme` mage targets to validate all links in the Otel Readme are valid.

## Why is it important?

You can either run `mage check` or `mage otel:readme` to trigger the link check. 

To test it indeed fails if a link is invalid you can introduce a typo and it should fail in subsequent runs.

`sed -i 's/featuregate/fff/' internal/pkg/otel/templates/README.md.tmpl` 

## Related issues

- Closes #5117


swiatekm (commented):
The changes look good to me. I do have three questions about the semantics of this check:

1. Are we checking all the links every time? Is there a mode available where we only check the PR diff?
2. What happens if the site we check is slow or unavailable? How does the tool behave in that case?
3. Do we have special cases (I imagine mostly around releases and changelog generation) where we'd run into a chicken-and-egg situation where links will only become valid after the PR is merged?

mauri870:
> The changes look good to me. I do have three questions about the semantics of this check:
> 
>     1. Are we checking all the links every time? Is there a mode available where we only check the PR diff?
> 
>     2. What happens if the site we check is slow or unavailable? How does the tool behave in that case?
> 
>     3. Do we have special cases (I imagine mostly around releases and changelog generation) where we'd run into a chicken-and-egg situation where links will only become valid after the PR is merged?

Thanks for the feedback, I'll try to reply to the points in order.

1. Yes,it checks the links every time. The tool I'm relying on ([link-patrol](https://pkg.go.dev/github.com/rednafi/link-patrol@v0.0.0-20240826150821-057643e74d4d)) needs to scan the entire file every time. Now that you mentioned this, this tool only works with markdown files, which would not work well in diffs, the parser can fail with broken mardown in a partial diff. I could not find any other alternatives written in Go. I found https://github.com/untitaker/hyperlink which seems to be more robust. I don't think this task can work well in PR diffs, but I could be mistaken.

2. It retries once. If the request times out, it doesn’t count as a failure. Only explicit errors, such as 404, will cause the check to fail.

3. Perhaps, but in this case (the OTel README), the file is generated via a template. As far as I know, the resulting file is always valid and isn’t meant to be changed manually.

swiatekm (approved):
:+1:

swiatekm:
> Yes,it checks the links every time. The tool I'm relying on ([link-patrol](https://pkg.go.dev/github.com/rednafi/link-patrol@v0.0.0-20240826150821-057643e74d4d)) needs to scan the entire file every time. Now that you mentioned this, this tool only works with markdown files, which would not work well in diffs, the parser can fail with broken mardown in a partial diff. I could not find any other alternatives written in Go. I found https://github.com/untitaker/hyperlink which seems to be more robust. I don't think this task can work well in PR diffs, but I could be mistaken.

We could only run it on files changed in the PR, but I'm not sure that would really improve much. There's not a lot of markdown files in this repo. I was mostly concerned about run times. If we're annoyed by the time it takes to run or by general flakiness, we can look into making it better.

mauri870:
> We could only run it on files changed in the PR, but I'm not sure that would really improve much. There's not a lot of markdown files in this repo. I was mostly concerned about run times. If we're annoyed by the time it takes to run or by general flakiness, we can look into making it better.

Definitely. For now it will only check the OTel README, but if we want to extend it to other files we might have to look closer to runtime performance, timeouts and the noisy output that the tool generates.

andrzej-stencel (approved):

Issue #5437, opened 2024-09-05
[Flaky Test]: TestStandaloneUpgradeRollback – upgrade details in the state cannot be nil


rdner:
### Failing test case

TestStandaloneUpgradeRollback

### Error message

upgrade details in the state cannot be nil

### Build

* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/2401
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/2423
* https://buildkite.com/elastic/elastic-agent-extended-testing/builds/2307

### OS

Linux

### Stacktrace and notes

    upgrader.go:347: waiting for upgrade watcher to start
    upgrader.go:352: upgrade watcher started
    upgrader.go:357: Checking upgrade details state while Upgrade Watcher is running
    fixture.go:657: >> running binary with: [/opt/Elastic/Agent/elastic-agent version --binary-only --yaml]
    fixture.go:657: >> running binary with: [/opt/Elastic/Agent/elastic-agent status --output json]
    fixture.go:657: >> running binary with: [/opt/Elastic/Agent/elastic-agent status --output json]
    upgrader.go:514: waiting for healthy agent and proper version: versions don't match: got 8.16.0, want 8.15.0
    fixture.go:657: >> running binary with: [/opt/Elastic/Agent/elastic-agent status --output json]
    upgrade_rollback_test.go:138: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/upgrade_rollback_test.go:138
        	Error:      	Expected value not to be nil.
        	Test:       	TestStandaloneUpgradeRollback
        	Messages:   	upgrade details in the state cannot be nil
    fixture_install.go:265: [test TestStandaloneUpgradeRollback] Inside fixture cleanup function
    fixture_install.go:281: collecting diagnostics; test failed
    fixture.go:657: >> running binary with: [/opt/Elastic/Agent/elastic-agent diagnostics -f /home/ubuntu/agent/build/diagnostics/TestStandaloneUpgradeRollback-diagnostics-2024-09-04T21:45:16Z.zip]
    fixture.go:657: >> running binary with: [/opt/Elastic/Agent/elastic-agent uninstall --force]
    fixture.go:1028: Dumping running processes in /home/ubuntu/agent/build/diagnostics/TEST-TestStandaloneUpgradeRollback-linux-amd64-ProcessDump-cleanup.json
[...]

PR #5445, opened 2024-09-05
Fix TestStandaloneUpgradeRollback

belimawr:
## What does this PR do?
This commit fixes two problems:
1. When getting the Elastic-Agent status to check for state.UpgradeDetails, it could happen that the upgrade had already finished, thus state.UpgradeDetails was nil. Now we only check for state.UpgradeDetails if the status is upgrading
2. When CheckHealthyAndVersion checks for the version during the rollback it could get the previous version, thus failing. This is fixed by ignoring version mismatch error.

## Why is it important?

It fixes flaky tests

## Checklist

## Disruptive User Impact

There is no disruptive user impact

Run TestStandaloneUpgradeRollback
```
SNAPSHOT=true TEST_PLATFORMS="linux/amd64" mage -v integration:single TestStandaloneUpgradeRollback
```

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/5437

## Questions to ask yourself

- ...

### DIFF
--- a/testing/upgradetest/upgrader.go
+++ b/testing/upgradetest/upgrader.go
@@ -421,6 +421,8 @@ func PerformUpgrade(
 	return nil
 }
 
+var ErrVerMismatch = errors.New("versions don't match")
+
 func CheckHealthyAndVersion(ctx context.Context, f *atesting.Fixture, versionInfo atesting.AgentBinaryVersion) error {
 	checkFunc := func() error {
 		status, err := f.ExecStatus(ctx)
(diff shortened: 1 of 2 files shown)


testing/upgradetest/upgrader.go
```diff
@@ -507,6 +513,12 @@ func WaitHealthyAndVersion(ctx context.Context, f *atesting.Fixture, versionInfo
 			return ctx.Err()
 		case <-t.C:
 			err := CheckHealthyAndVersion(ctx, f, versionInfo)
+			// If we're in an upgrade process, the versions might not match
+			// so we wait to see if we get to a stable version
+			if errors.Is(err, ErrVerMismatch) {
```
belimawr, line 518: The one thing I'm not 100% sure is if this could slow down some other tests that now will need to wait for the timeout. If that's a problem, I can add an "IgnoreVersionMismatch" flag.

blakerouse (approved):
Looks good!

Issue #5451, opened 2024-07-26
[Fleet]: Log level is not set automatically on re-assigning from policy with Tamper protection enabled to policy with no Elastic Defend.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.15.0 BC4
BUILD: 76261
COMMIT: 9d62937675e62265342e86d8f0db601dc75498b8
```

**Preconditions:**
1. 8.15.0-BC4 Kibana cloud environment should be available.
2. Agent should be installed.

**Steps to reproduce:**
1. Navigate to Fleet>Agent policies tab.
2. Add Elastic Defend to the policy with Agent and enable Tamper protection.
3. Set global logging level to `debug`.
4. Now wait for sometime and observe debug logs are generated for the agent.
5. Now create a new policy with only System integration with logging level set to default- `info`.
6. Assign the agent to this policy,
7. Wait for more than 10 minutes.
8. Observe logging level is not set to info for the agent.

**Logs:**
[elastic-agent-diagnostics-2024-07-26T06-59-05Z-00.zip](https://github.com/user-attachments/files/16389161/elastic-agent-diagnostics-2024-07-26T06-59-05Z-00.zip)

**Expected Result:**
Log level should be set automatically on assigning from policy with Tamper protection enabled to policy with no Elastic Defend.

**What's working fine:**
- No issue is observed without the tamper protection, and logging level is set automatically on policy reassignment.
- Issue is fixed on manually clicking Reset to policy under Agent logs tab.

**Screen Recording:**

https://github.com/user-attachments/assets/8bfb50be-8083-4959-8512-33cab0793869

**Feature:**
https://github.com/elastic/kibana/issues/158861

pchila:
> Though weirdly I see the local agent storage with the log level set to info which is a bit odd:
@cmacknz that is the default value agent writes for its own (specific to the agent) log level, due to the unique way agent manages the agentinfo, removing that was impossible without some big refactoring 😓 

> I think this could be fixed on Agent side to set the log level to the default info when nil is received from the policy.
> This is already how it works in settings action: https://github.com/pchila/elastic-agent/blob/db92d839bb237ba06bb10c633ade1b0047aecbdc/internal/pkg/agent/application/actions/handlers/handler_action_settings.go#L89-L91

[...]

PR #5452, opened 2024-09-06
set default log level when not set in policy

juliaElastic:
## What does this PR do?

Fix fallback to default log level when reassigning an agent to a policy without explicit log level set.

## Why is it important?

It is unexpected that on reassign agent policy, the agent doesn't use the log level of the agent policy.

## Checklist

- Create an agent policy and set `debug` log level on it
- Enroll agent (build from this pr) to the agent policy
- Verify the agent logging on debug level
- Create another agent policy (with default log level)
- Reassign agent to the second agent policy
- Verify the agent logging is set back to default (info) level

image

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5451


internal/pkg/agent/application/actions/handlers/handler_action_policy_change_test.go
```diff
@@ -1103,10 +1115,10 @@ func TestPolicyChangeHandler_handlePolicyChange_LogLevelSet(t *testing.T) {
 }
 
 func nilLogLevelSet(t *testing.T) *mockhandlers.LogLevelSetter {
-	// nilLogLevel is a variable used to match nil policy log level being set
-	var nilLogLevel *logger.Level = nil
+	// default log level should be set when policy log level is nil
+	defaultLogLevel := logger.DefaultLogLevel
```
pchila, line 1119: If you set Info here, how do you clear a log level coming from policy to allow fallback onto the agent default level ? Having a nil log level value from the policy should be handled in the loglevel setter since that's the object that has the logic for proper priority and fallback, this will force the agent to always have a log level coming from policy even when there is non specified

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -400,10 +400,14 @@ func (h *PolicyChangeHandler) applyLoggingConfig(ctx context.Context, loggingCon
...
 
-	h.log.Infof("Setting fallback log level %v from policy", policyLogLevel)
-	return h.policyLogLevelSetter.SetLogLevel(ctx, policyLogLevel)
+		h.log.Infof("Setting fallback log level %v from policy", policyLogLevel)
+		return h.policyLogLevelSetter.SetLogLevel(ctx, policyLogLevel)
+	}
+	// use default log level
+	defaultLogLevel := logger.DefaultLogLevel
+	h.log.Infof("Setting fallback log level to default %s", defaultLogLevel)
+	return h.policyLogLevelSetter.SetLogLevel(ctx, &defaultLogLevel)
```
pchila, line 410: If you set Info here, how do you clear a log level coming from policy to allow fallback onto the agent default level ? Having a nil log level value from the policy should be handled in the loglevel setter since that's the object that has the logic for proper priority and fallback, this will force the agent to always have a log level coming from policy even when there is non specified

juliaElastic: Before this change, when the policy log level was not set (nil), there wasn't any change made to the agent log level (that's why we had the bug on reassign).
This logic sets the fallback log level to info, which only applies if the agent doesn't have log level set.

I tested setting an agent log level to debug, and then making a policy change on a policy which has no log level. The agent in that case kept the debug log level, so I don't think there is any issue with this.

juliaElastic: Thanks for the explanation, I moved the logic to SetLogLevel, and tested that it works.

pchila (approved):
Just a couple of suggestion for the changelog fragment... LGTM otherwise

PR #5454, opened 2024-09-06
Don't use the default port in runtime manager tests

swiatekm:
## What does this PR do?

Uses random ports in unit tests where we actually bind to them.

## Why is it important?
Using defaults runs into conflicts with an actual agent running on the system.

## Checklist

Normally run unit tests via `mage test:unit`.


blakerouse (approved):
+1

PR #5457, opened 2024-09-06
Add integration test upgrading latest release to a build from the current PR

### LINKED ISSUE #4560: Add an upgrade integration test upgrading to the PR build
**Describe the enhancement:**

The integration test should install and enroll the latest official release Elastic Agent, then upgrade it to the agent built from the PR.

The following is required:
 - Sing the agent package built from the PR. (use [`pgptest.Sing`](https://github.com/elastic/elastic-agent/blob/main/testing/pgptest/pgp.go#L19)
 - Setup a http server serving the ASCII armoured public key used to sign the agent artifact.
   - the agent downloads it from `https://artifacts.elastic.co/GPG-KEY-elastic-agent`.
   - See [PGP fallback URL in air-gapped environment workaround](https://github.com/elastic/elastic-agent/blob/26947871957102cd972210202f15217789ff5cb9/docs/pgp-workaround.md) for details.
 - setup a http server to serve the agent binary, its signature and checksum file for the upgrade
 - configure the policy to use the custom agent download binary URL

To test [the migration of the state store](https://github.com/elastic/elastic-agent/pull/4441) during upgrade

The test is created and successfully runs together with the current integration tests

AndersonQ:
## What does this PR do?

Adds an integration test which upgrades the latest Elastic Agent release to the version built form the current PR being tested.

## Why is it important?

It'll allow us to catch any bug preventing the agent from upgrading before merging the PR.

## Checklist

## Disruptive User Impact

- N/A

```shell
AGENT_KEEP_INSTALLED=true INSTANCE_PROVISIONER="multipass" SNAPSHOT=true TEST_PLATFORMS="linux/amd64" TEST_PACKAGES="tar.gz" mage integration:single TestFleetPRBuildSelfSigned
```

## Related issues

- Closes #4560

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -50,6 +50,7 @@ type Fixture struct {
 	binaryName      string
 	runLength       time.Duration
 	additionalArgs  []string
+	env             []string
 
 	srcPackage string
 	workDir    string
@@ -136,6 +137,15 @@ func WithRunLength(run time.Duration) FixtureOpt {
 func WithAdditionalArgs(args []string) FixtureOpt {
 	return func(f *Fixture) {
 		f.additionalArgs = args
+	}
+}
+
+// WithEnv sets the environment when running the Elastic Agent.
+// Each entry is of the form "key=value". os.Environ() is added by default.
+// See exec.Cmd.Env for details.
+func WithEnv(env []string) FixtureOpt {
+	return func(f *Fixture) {
+		f.env = append(os.Environ(), env...)
 	}
 }
 
@@ -673,6 +683,11 @@ func (f *Fixture) PrepareAgentCommand(ctx context.Context, args []string, opts .
 		if err := o(cmd); err != nil {
 			return nil, fmt.Errorf("error adding opts to Exec: %w", err)
 		}
+	}
+
+	if len(f.env) > 0 {
+
+		cmd.Env = f.env
 	}
 	return cmd, nil
 }
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -140,15 +140,6 @@ func WithAdditionalArgs(args []string) FixtureOpt {
 	}
 }
 
-// WithEnv sets the environment when running the Elastic Agent.
-// Each entry is of the form "key=value". os.Environ() is added by default.
-// See exec.Cmd.Env for details.
-func WithEnv(env []string) FixtureOpt {
-	return func(f *Fixture) {
-		f.env = append(os.Environ(), env...)
-	}
-}
-
 // NewFixture creates a new fixture to setup and manage Elastic Agent.
 func NewFixture(t *testing.T, version string, opts ...FixtureOpt) (*Fixture, error) {
 	// we store the caller so the fixture can find the cache directory for the artifacts that
--- a/pkg/testing/fixture.go
+++ b/pkg/testing/fixture.go
@@ -1044,6 +1044,29 @@ func (f *Fixture) DumpProcesses(suffix string) {
 	}
 }
 
+func (f *Fixture) KeepFileByMoving(file string) {
+	filename := filepath.Base(file)
+
+	dir, err := findProjectRoot(f.caller)
+	if err != nil {
+		f.t.Logf("failed to keep file; failed to find project root: %s", err)
+		return
+	}
+
+	destFile := filepath.Join(dir, "build", "diagnostics", filename)
+	fileDir := path.Dir(destFile)
+	if err := os.MkdirAll(fileDir, 0777); err != nil {
+		f.t.Logf("failed to keep file; failed to create directory %s: %s", fileDir, err)
+		return
+	}
+
+	f.t.Logf("moving %q to %q", file, destFile)
+	err = os.Rename(file, destFile)
+	if err != nil {
+		f.t.Logf("failed to move %q to %q: %v", file, destFile, err)
+	}
+}
+
 // validateComponents ensures that the provided UsableComponent's are valid.
 func validateComponents(components ...UsableComponent) error {
 	for idx, comp := range components {
--- a/pkg/testing/fixture_install.go
+++ b/pkg/testing/fixture_install.go
@@ -165,7 +165,7 @@ func (f *Fixture) Install(ctx context.Context, installOpts *InstallOpts, opts ..
 
 	// check for running agents before installing, but only if not installed into a namespace whose point is allowing two agents at once.
 	if installOpts != nil && !installOpts.Develop && installOpts.Namespace == "" {
-		assert.Empty(f.t, getElasticAgentProcesses(f.t), "there should be no running agent at beginning of Install()")
+		// assert.Empty(f.t, getElasticAgentProcesses(f.t), "there should be no running agent at beginning of Install()")
 	}
 
 	switch f.packageFormat {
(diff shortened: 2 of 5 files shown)


pchila (commented):
I may have misunderstood the intent of the PR so please correct me if I am wrong but the purpose here is to have a fleet-managed agent (installed using the latest release) upgrade to the version built from source on the branch, correct?

Why do we need to impersonate the artifact server, messing with /etc/hosts? Wouldn't a simple `httptest.Server` serve the package we have available locally? We should be able to set that on the policy using Fleet API...

Am I missing something?

pchila:
> * one served by fleet-server, which on cloud we cannot changed
> 
> * one that is built-in in the agent
> 
> * one it downloads from the artifacts API, which the address is hard coded

Ok, so for a limitation on ESS fleet-server side (not being able to send a `skip-verify` or a setting a custom PGP key for validation) we are impersonating the API server itself by limiting this test to Linux and playing with `/etc/hosts`, did I get the gist ?

I am not sure this is the best way to get a fleet-managed agent to upgrade to a locally signed package (we already support the test for standalone agents since we can pass the additional options). Maybe it's time to have integration tests run with something other than ESS deployments, using a fleet-server we can configure?

@cmacknz do we still want to persevere in using ESS deployments in integration tests even if it forces us to implement these sort of workarounds ?

@AndersonQ thanks for the explanation, I see now what constraints we have on integration tests involving fleet-managed agents, I will have another look

AndersonQ:
> I am not sure this is the best way to get a fleet-managed agent to upgrade to a locally signed package (we already support the test for standalone agents since we can pass the additional options). Maybe it's time to have integration tests run with something other than ESS deployments, using a fleet-server we can configure?

I agree 100% with you, but it's the current limitations we have. 


> @cmacknz do we still want to persevere in using ESS deployments in integration tests even if it forces us to implement these sort of workarounds ?

being able to deploy a fleet-server would also help us with the mTLS tests

pkg/testing/fixture.go
```diff
@@ -1044,6 +1043,33 @@ func (f *Fixture) DumpProcesses(suffix string) {
 	}
 }
 
+// KeepFileByMoving moves file to 'build/diagnostics' which contents are
+// available on CI if the test fails or on the agent's 'build/diagnostics'
+// if the test is run locally.
+// 'prefix is added to the file name when moving
+func (f *Fixture) KeepFileByMoving(file, prefix string) {
```
pchila, line 1050: Nit: a clearer name would help, maybe even splitting the 2 things that this function is doing in multiple functions
```suggestion
func (f *Fixture)BuildArtifactsDir() {...}
func (f *Fixture)MoveFile(src, dst string) {...}
func (f *Fixture)MoveToArtifacts(file string) {
	dstFileDir := f.BuildArtifactsDir()
	dstFileName := f.FileNamePrefix() + filepath.Base(file)
	f.MoveFile(file, filepath.Join(dstFileDir, dstFileName))
}
```

pkg/testing/fixture_install.go
```diff
@@ -699,6 +691,27 @@ func (f *Fixture) collectDiagnostics() {
...
+
+	stamp := time.Now().Format(time.RFC3339)
+	// on Windows a filename cannot contain a ':' as this collides with disk
+	// labels (aka. C:\)
+	stamp = strings.ReplaceAll(stamp, ":", "-")
+
+	// Subtest names are separated by "/" characters which are not valid
+	// filenames on Linux.
+	sanitizedTestName := strings.ReplaceAll(f.t.Name(), "/", "-")
+	prefix := fmt.Sprintf("%s-%s", sanitizedTestName, stamp)
```
pchila, line 709: Nit: this could be grouped in a generic utility function `SanitizePath` or whatever other name...

testing/integration/upgrade_fleet_test.go
```diff
@@ -127,6 +132,146 @@ func TestFleetAirGappedUpgradePrivileged(t *testing.T) {
...
+	require.NoError(t, err, "failed to get fixture with PR build")
+
+	prBuildPkgPath, err := toFixture.SrcPackage(ctx)
+	require.NoError(t, err, "could not get path to PR build artifact")
+
+	agentPkg, err := os.Open(prBuildPkgPath)
+	require.NoError(t, err, "could not open PR build artifact")
+
+	// sign the build
+	pubKey, ascData := pgptest.Sing(t, agentPkg)
```
pchila, line 187:
```suggestion
	pubKey, ascData := pgptest.Sign(t, agentPkg)
```

testing/integration/upgrade_fleet_test.go
```diff
@@ -247,6 +393,7 @@ func testUpgradeFleetManagedElasticAgent(
 		nonInteractiveFlag = true
 	}
 	installOpts := atesting.InstallOpts{
+		Insecure:       true,
```
pchila, line 396: Could this be passed as parameter so that we use insecure: true only on the new test ?

testing/integration/upgrade_fleet_test.go
```diff
@@ -127,6 +132,146 @@ func TestFleetAirGappedUpgradePrivileged(t *testing.T) {
...
+func TestFleetUpgradeToPRBuild(t *testing.T) {
+	stack := define.Require(t, define.Requirements{
+		Group: FleetUpgradeToPRBuild,
+		Stack: &define.Stack{},
+		OS:    []define.OS{{Type: define.Linux}}, // The test uses /etc/hosts.
+		Sudo:  true,                              // The test uses /etc/hosts.
+		// The test requires:
+		//   - bind to port 443 (HTTPS)
+		//   - edit /etc/hosts
+		//   - agent installation
```
pchila, line 144:
```suggestion
		Sudo:  true,                              // The test uses /etc/hosts.
		// The test requires:
		//   - bind to port 443 (HTTPS)
		//   - edit /etc/hosts
		//   - changes to /etc/ssl/certs
		//   - agent installation
```

testing/integration/upgrade_fleet_test.go
```diff
@@ -426,3 +573,89 @@ func agentUpgradeDetailsString(a *kibana.AgentExisting) string {
 
 	return fmt.Sprintf("%#v", *a.UpgradeDetails)
 }
+
+// prepareFileServer prepares and returns a started HTTPS file server serving
+// files from rootDir and using cert as its TLS certificate.
+func prepareFileServer(t *testing.T, rootDir string, cert tls.Certificate) *httptest.Server {
```
pchila, line 579: it's better if we say what we actually do in this function
```suggestion
func startHttpsFileServer(t *testing.T, rootDir string, cert tls.Certificate) *httptest.Server {
```

testing/integration/upgrade_fleet_test.go
```diff
@@ -426,3 +573,89 @@ func agentUpgradeDetailsString(a *kibana.AgentExisting) string {
...
+	require.NoError(t, err, "could not create net listener for port 443")
+
+	server.TLS = &tls.Config{Certificates: []tls.Certificate{cert}}
+	server.StartTLS()
+	t.Logf("file server running on %s", server.URL)
+
+	return server
+}
+
+// prepareTLSCerts generates a CA and a child certificate for the given host and
```
pchila, line 606: I am pretty sure there's already a function in test code that does this: can we avoid the duplication ?

testing/integration/upgrade_fleet_test.go
pchila: I guess this line wasn't needed after all ? I cannot find it in the latest version of the PR

testing/integration/upgrade_fleet_test.go
pchila: If the other function is only missing the hostname I would rather extend the existing function, generating certificates is tedious and error prone, I would rather avoid the duplication...

Not blocking, anyways

pchila (approved):

PR #5458, opened 2024-09-06
Fix minor issues with the test framework guide

swiatekm:
## What does this PR do?

In the integration test framework guide, mention that authenticating with Elastic's docker registry is necessary to run integration tests. Change `integration:listStacks` to `integration:stacks`.

## Related issues

- https://github.com/elastic/ingest-dev/issues/3759

### DIFF
--- a/docs/test-framework-dev-guide.md
+++ b/docs/test-framework-dev-guide.md
@@ -38,6 +38,11 @@ version without SNAPSHOT.
 One-time setup is required to run any integration and E2E tests. Run
 `mage integration:auth` to perform this setup.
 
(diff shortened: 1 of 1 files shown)


cmacknz (approved):

blakerouse (approved):
Used to not be the case, was missed when the OGC docker image was moved.

swiatekm:
Looks like SonarQube doesn't run on PRs that only touch docs, but the check is still required to pass in Github.

ycombinator:
> Looks like SonarQube doesn't run on PRs that only touch docs, but the check is still required to pass in Github.

Ugh, https://github.com/elastic/ingest-dev/issues/3525 rears its head again. Force merging since other checks are green.

PR #5462, opened 2024-09-06
Log when `container-paths.yml` is loaded

belimawr:
## What does this PR do?

See title

## Why is it important?

It adds logs when the Elastic-Agent loads `container-paths.yml`

## Checklist

## Disruptive User Impact

None

Export the following environmet variables:
- `FLEET_ENROLL`
- `FLEET_URL`
- `FLEET_ENROLLMENT_TOKEN`

Run the container command as root
```
./elastic-agent container
```

Stop the Elastic-Agent and run the status command:
```
./elastic-agent status
```

You should see the logs:
```
2024/09/06 16:54:33 container path file '/usr/share/elastic-agent/state/container-paths.yml' found
2024/09/06 16:54:33 state Path: '/usr/share/elastic-agent/state', config path: '/usr/share/elastic-agent/state', logs path: '', socket path: 'unix:///usr/share/elastic-agent/state/data/Td8I7R-Zby36_zF_IOd9QVNlFblNEro3.sock'
```

~~## Related issues~~

## Questions to ask yourself

- ...


internal/pkg/agent/cmd/container.go
```diff
@@ -903,6 +905,8 @@ func tryContainerLoadPaths() error {
 	if err != nil {
 		return fmt.Errorf("failed to unpack %s: %w", pathFile, err)
 	}
+	log.Printf("state Path: '%s', config path: '%s', logs path: '%s', socket path: '%s'", paths.StatePath, paths.ConfigPath, paths.LogsPath, paths.SocketPath)
```
cmacknz, line 908: Shouldn't agent be logging this regardless of if it is in a container? Also isn't this in diagnostics, why do we need to log it (from the perspective that every log statement we add adds to the size and cost of the remote monitoring indices)?

belimawr: This PR was triggered by my work on fixing a flaky test (https://github.com/elastic/elastic-agent/issues/5159), I don't recall if the diagnostics were being successfully collected, but I believe they were not because the failing tests do not install the Elastic-Agent, they enrol and run the agent, hence the need to have it in the logs.

This is actually logged in any run of the agent where `tryContainerLoadPaths` sets the paths. On a correct execution it should be only inside a container.

Thinking more broadly, I agree it makes sense to log it every time the agent runs. Should I update the PR for it?

cmacknz: I diagnostics we get the following in variables.yaml which omits the state path which is container specific, so let's just leave the scope of this one to here.

```yaml
      path:
        config: C:\Program Files\Elastic\Agent
        data: C:\Program Files\Elastic\Agent\data
        home: C:\Program Files\Elastic\Agent\data\elastic-agent-8.14.1-1348b9
        logs: C:\Program Files\Elastic\Agent
```

I'm fine with this log.

cmacknz: Actually, it seems like instead of both `log.Printf("container path file '%s' found", pathFile)` and `log.Printf("state Path: '%s', config path: '%s', logs path: '%s', socket path: '%s'", paths.StatePath, paths.ConfigPath, paths.LogsPath, paths.SocketPath)` we should have a single log line just unconditionally logging the state path we are using.

Reading https://github.com/elastic/elastic-agent/pull/5361 it also seems like the state path is the problem.

cmacknz: We should log the state path right after this block above:

```go
    statePath := envWithDefault("", "STATE_PATH")
	if statePath == "" {
		statePath = defaultStateDirectory
	}
```

belimawr: The `STATE_PATH` mentioned on https://github.com/elastic/elastic-agent/pull/5361 was what made that test leave state behind, which made a new execution/installation of the Elastic-Agent not to behave as it should have, even the `--force` flag does not overwrite the `container-paths.yml`. That's is, in my opinion, one of the issues, hence this PR to be explicit about which paths the Elastic-Agent is using.

belimawr: Reading the code and thinking more about it, the data in `container-paths.yml` can overwrite the state path set by `STATE_PATH`, so I believe we still need both logs, one indicating which file we read and another indicating the final paths we're using after reading it. I'd keep both log lines.

belimawr:
I cannot reproduce those CI failures, I'll just re-run/update the branch

belimawr:
buildkite test this

belimawr:
The tests are failing due to issues creating a 9.0.0 stack.

swiatekm (approved):

cmacknz:
I still don't love logging this unconditionally, instead of putting this diagnostics, or having it respect a log level, but it's not a hill I'm going to die on if we think it's useful.

I also think the true root cause that caused us to need this is our attempts to run the `elastic-agent container` command outside of an actual container, which is something we no longer need to do now that we have a way to run integration tests on k8s directly.

cmacknz:
I am talking about https://github.com/elastic/elastic-agent/blob/3b1f793a0d5e4fa016d148f7931d48af50e0c506/testing/integration/container_cmd_test.go#L171 specifically that should probably be rewritten to actually run in a container, for reference.

belimawr:
> I also think the true root cause that caused us to need this is our attempts to run the `elastic-agent container` command outside of an actual container, which is something we no longer need to do now that we have a way to run integration tests on k8s directly.

That is true. The reason for this PR was really to help debugging tests or other weird situations we might fall into. The tests didn't have diagnostics because the agent wasn't installed, logs are the primary source of information on those cases.

I'm not opposed to closing this PR, I haven't had time to come back to it in a while :/.

belimawr:
> I am talking about
> 
> https://github.com/elastic/elastic-agent/blob/3b1f793a0d5e4fa016d148f7931d48af50e0c506/testing/integration/container_cmd_test.go#L171
> 
> specifically that should probably be rewritten to actually run in a container, for reference.

I agree, now that we have the infrastructure in place, that's definitely a better option.

jlind23:
Cleaning up the elastic agent repo, closing this PR for now please reopen if required.

PR #5464, opened 2024-09-06
Correctly package with the ELv2 license

cmacknz:
Elastic Agent should have the Elastic License 2.0 as that is the one included in the root of the repository. 

https://github.com/elastic/elastic-agent/blob/781f5296ae27d368ca0f5a14ae2cd34d21210a3a/LICENSE.txt#L1

However, the license header on source files and the license packaged as part of the build was Elastic License 1.0. Update the source and packaging to correctly include the 2.0 version of the license.

This PR touches pretty much every file in the repository because of the header update, the actual changes to the build are in the dev-tools directory.

To check the license:

```sh
EXTERNAL=true SNAPSHOT=true PACKAGES=targz PLATFORMS=darwin/arm64 mage package
cd build/distributions
tar xvfz elastic-agent-8.16.0-SNAPSHOT-darwin-aarch64.tar.gz
cat elastic-agent-8.16.0-SNAPSHOT-darwin-aarch64/LICENSE.txt
```


changelog/fragments/1725655844-Include-correct-ELv2-license-file-in-elastic-agent-packages.yaml
```diff
@@ -0,0 +1,27 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11: Is this a feature? Maybe enhancement? Or just other?

blakerouse (approved):
+1 from control plane member, just a comment about the changelog entry (non-blocker).

MichaelKatsoulis (approved):

changelog/fragments/1725655844-Include-correct-ELv2-license-file-in-elastic-agent-packages.yaml
cmacknz: I did not think about this at all, probably enhancement would be better. There's no "other" section of the [changelog](https://www.elastic.co/guide/en/fleet/current/release-notes-8.15.0.html) so it's either feature or enhancement.

blakerouse: Says there is an "other" option on line 10 there, if that is not the case probably should get it removed from the changelog tool so it doesn't generate that line.

PR #5473, opened 2024-09-09
Add JUnit annotate BK plugin

alexsapran:
## What does this PR do?

This PR adds the JUnit annotate Buildkite plugin

## Why is it important?

This is goint to help with easier triaging what failed on each build.

I pushed a temp commit to cause some errors. See the screenshot
Screenshot 2024-10-07 at 1 53 12 PM

Now, when a test fails, you see the failure at the top with an additional link to the step that caused that error.

Example build with a failure to show this in action https://buildkite.com/elastic/elastic-agent/builds/12866

This is a Buildkite change, only testable on BK.

## Related issues

Related: https://github.com/elastic/ingest-dev/issues/3773


dev-tools/mage/gotest.go
```diff
@@ -47,7 +47,17 @@ type TestBinaryArgs struct {
 }
 
 func makeGoTestArgs(name string) GoTestArgs {
-	fileName := fmt.Sprintf("build/TEST-go-%s", strings.Replace(strings.ToLower(name), " ", "_", -1))
+	fileName := ""
+	_, isCI := os.LookupEnv("CI")
+	if isCI {
```
alexsapran, line 52: When running inside CI, the BK plugin will try to link the failure to the step if the generated file has this pattern.

see: https://github.com/buildkite-plugins/junit-annotate-buildkite-plugin#job-uuid-file-pattern-optional

.buildkite/scripts/steps/unit-tests.sh
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
-set -uo pipefail
-
 source .buildkite/scripts/common.sh
+set +euo pipefail
```
alexsapran, line 3: We set pipefail in the `common.sh` so the script was never run and performed the proper coverage files.

.buildkite/pipeline.yml
```diff
@@ -154,6 +154,32 @@ steps:
...
 
+
+  - label: ":junit: Junit annotate"
+    plugins:
+      - junit-annotate#v2.4.1:
+          artifacts: "build/*.xml"
+    agents:
+      provider: "gcp"
+    depends_on:
+      - step: "unit-tests-2204"
```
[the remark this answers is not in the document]
alexsapran: Now, failures will be reported at the top with a link to the step that produced this error.
See 
Screenshot 2024-10-07 at 1 53 12 PM

.buildkite/scripts/steps/unit-tests.sh
pazone: What script was never run?

alexsapran: Because of the order in which we set the bash flags, the' pipe fail` was overridden inside the `common.sh`, so the wrapping of the exit code was never run.

pazone (approved):

jlind23 (approved):

PR #5476, opened 2024-09-09
[CI] Correct Go version when running the coverage report

alexsapran:
## What does this PR do?

Updates the coverage report step to use the proper Golang version

## Why is it important?

I am not sure if this is related, but recent builds show that the `golang.org/x/tools/cover` suddenly fails when pulling.

We should ensure the version is pulled from the right place `.go-version` or similar.  This is not in scope in this PR, this is to help unblock main.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -171,7 +171,7 @@ steps:
     artifact_paths:
       - "build/TEST-go-unit.cov"
     agents:
-      image: "golang:1.20.10"
+      image: "golang:1.22.6"
     depends_on:
       - unit-tests
       - extended-windows


rowlandgeoff (approved):
👍

.buildkite/pipeline.yml
```diff
@@ -171,7 +171,7 @@ steps:
     artifact_paths:
       - "build/TEST-go-unit.cov"
     agents:
-      image: "golang:1.20.10"
+      image: "golang:1.22.6"
```
cmacknz, line 174: Related, I think this is being omitted from the action that auto-bumps the go version https://github.com/elastic/elastic-agent/blob/main/.ci/updatecli/updatecli-bump-golang.yml

alexsapran: IMHO, we should not have a pinned version in there like that or if we endup doing it like that, to have it be bumped with the automation.

pchila (approved):

.buildkite/pipeline.yml
rowlandgeoff: Tracking with https://github.com/elastic/ingest-dev/issues/3969

Issue #4185, opened 2024-02-02
APM config is not correctly marshaled in diagnostic files


pchila:
Please post all questions and issues concerning the Elastic Agent on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For security vulnerabilities please only send reports to security@elastic.co.
See https://www.elastic.co/community/security for more information.

Please include configurations and logs if available.

APM configuration is not correctly marshaled when collecting diagnostics, see https://github.com/elastic/elastic-agent/pull/3223/files#diff-fc6b455c3e3a0055095709cc4138f976720bf752acce1ac872505bcf94bf9322R100-R113

This is happening because the APMConfig struct has only `config` tags and doesn't specify `yaml:` tags to set the names correctly when marshaling objects with a yaml.Marshal (for example in diagnostics)

Setting the proper tags should fix this issue.

PR #5492, opened 2024-09-10
Add apm trace sampling

### LINKED ISSUE #5211: Add missing tracing sampling APM config section
**Describe the enhancement:**

[APM tracing config](https://github.com/elastic/elastic-agent/blob/03b5d31eeda94b2f3f7ab83bded45c03738fb5a7/internal/pkg/core/monitoring/config/config.go#L126-L134) should additionally include a section for [trace sampling](https://www.elastic.co/guide/en/observability/current/apm-configure-tail-based-sampling.html) to enable APM Server monitoring be configured with custom sampling rate when managed by Elastic Agent. 

Allow to specify a custom trace sampling rate to control the ingestion rate when self monitoring running APM Server.

The custom sampling APM config section is handled correctly and used to configure tracing in APM Server.

pchila:
## What does this PR do?
This PR adds the possibility to specify apm sample rate in elastic agent APM configuration

## Why is it important?
Specifying a sample rate allows controlling the ingestion rate, reducing overhead and storage requirements when self-monitoring is active

## Checklist

## Related issues

- Closes #5211 

## Questions to ask yourself

- ...

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1031,11 +1031,11 @@ Contents of probable licence file $GOMODCACHE/github.com/elastic/elastic-agent-a
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/elastic-agent-client/v7
-Version: v7.15.0
+Version: v7.16.0
 Licence type (autodetected): Elastic
 --------------------------------------------------------------------------------
 
-Contents of probable licence file $GOMODCACHE/github.com/elastic/elastic-agent-client/v7@v7.15.0/LICENSE.txt:
+Contents of probable licence file $GOMODCACHE/github.com/elastic/elastic-agent-client/v7@v7.16.0/LICENSE.txt:
 
 ELASTIC LICENSE AGREEMENT
 


blakerouse (approved):
More testing code then actual change code. I appreciate the attention to coverage in this change, well done!

internal/pkg/agent/application/coordinator/diagnostics_test.go
```diff
@@ -106,15 +106,14 @@ agent:
         - host1
         - host2
       environment: diag-unit-test
-      apikey: apikey
-      secrettoken: secret
-      globallabels:
+      api_key: apikey
```
andrzej-stencel, line 109: Oh I think I see now, the `yaml` tag additions in `core/monitoring/config/config.go` are to fix the way the fields are marshaled into Yaml here.

I don't know the context of this config, I wonder if this might be treated as a breaking change? Is it possible that a user currently depends on the `apikey` not being `api_key`, or on `skipverify` to always exist? Or is the structure of this config an internal detail that user's aren't supposed to rely on?

andrzej-stencel (commented):
I'm not seeing any validation for the value of SamplingRate. Should there be any? I assume the value of SamplingRate should be a positive number lower than 1?

internal/pkg/core/monitoring/config/config.go
```diff
@@ -125,20 +125,21 @@ func DefaultConfig() *MonitoringConfig {
 
 // APMConfig configures APM Tracing.
 type APMConfig struct {
-	Environment  string            `config:"environment"`
-	APIKey       string            `config:"api_key"`
-	SecretToken  string            `config:"secret_token"`
-	Hosts        []string          `config:"hosts"`
-	GlobalLabels map[string]string `config:"global_labels"`
-	TLS          APMTLS            `config:"tls"`
+	Environment  string            `config:"environment" yaml:"environment,omitempty"`
```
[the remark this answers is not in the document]
pchila: This configuration gets dumped in a yaml file when we are capturing diagnostics, without the yaml tags the keys in the yaml files do not correspond to the keys we use in configuration and that's confusing...

See https://github.com/elastic/elastic-agent/issues/4185 for an example

internal/pkg/agent/application/coordinator/diagnostics_test.go
pchila: Exactly, I already spotted that the APM config fields used different keys when dumped into yaml, and I tried to correct (as much as possible) in this PR

It's not a breaking change as diagnostics are used internally to debug SDHs and weird agent behaviors, the consumer of these files is a human.

I tried to align the keys to facilitate copy/paste from diagnostics to config files/policies when that becomes necessary.

It would be a breaking change if I changed the `config:` tags as that would change the expected keys in config file/policy ;), changing the diagnostics output we are just trying to align those without any breakage :wink:

andrzej-stencel: Awesome, thanks for explaining this! :+1:

pchila:
> I'm not seeing any validation for the value of SamplingRate. Should there be any? I assume the value of SamplingRate should be a positive number lower than 1?

According to the APM documentation https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-transaction-sample-rate it should be a float between 0.0 and 1.0

If we have to add a validation also on agent side we would need to rely on `ucfg` module as we do for all the rest of the config although I would prefer to leave that to any validation performed in the apm module

andrzej-stencel (approved):

ycombinator:
SonarQube is complaining about lines in `run.go` being not tested; code in this file is not easy to unit test.  Force merging...

PR #5522, opened 2024-09-11
Clarify purpose of `.agent-versions` file

ycombinator:
## What does this PR do?

This PR clarifies the purpose of the `.agent-versions` file. It does so by making the following changes:
* Moves and renames the file from `.agent-versions.json` to `testing/integration/testdata/.upgrade-test-agent-versions.yml` and adjust the contents to match.  The reason for using YAML is so we can add a header comment about the contents of the file in the file itself.
* Updates the upgrade integration tests that use this file to reference the new one.
* Updates the commit message in PRs created by automation to keep this file updated.

### Before this PR

Before this PR, we had an `.agent-versions.json` file and it looked like this:

```json
{
  "testVersions": [
    "8.15.2-SNAPSHOT",
    "8.15.1",
    "8.15.0",
    "7.17.25-SNAPSHOT"
  ]
}
```

### After this PR

After this PR, we instead have an `testing/integration/testdata/.upgrade-test-agent-versions.yml` file and it looks like this:

```yml
# This file is generated automatically. Please do not manually edit it.

# The testVersions list in this file specifies Elastic Agent versions to be used as
# the starting (pre-upgrade) or ending (post-upgrade) versions of Elastic Agent in
# upgrade integration tests.

testVersions:
  - 8.15.2-SNAPSHOT
  - 8.15.1
  - 8.15.0
  - 7.17.25-SNAPSHOT
```

## Why is it important?

To make it easier to understand the purpose of the `.agent-versions` file and to prevent accidental manual edits to it.

1. Create a dummy, valid `testing/integration/testdata/.upgrade-test-agent-versions.yml` file.
2. Run `integration:updateVersions`.
3. Make sure the `testing/integration/testdata/.upgrade-test-agent-versions.yml` file is updated with the expected contents.


pchila (approved):
LGTM

.github/workflows/bump-agent-versions.sh
```diff
@@ -22,7 +22,7 @@ else
     git add .agent-versions.json .package-version
 
     nl=$'\n' # otherwise the new line character is not recognized properly
-    commit_desc="These files are used for picking agent versions in integration tests.${nl}${nl}The content is based on responses from https://www.elastic.co/api/product_versions and https://snapshots.elastic.co${nl}${nl}The current update is generated based on the following requirements:${nl}${nl}Package version: ${package_version}${nl}${nl}\`\`\`json${nl}${version_requirements}${nl}\`\`\`"
+    commit_desc="These files are used for picking the starting (pre-upgrade) agent versions in upgrade integration tests.${nl}${nl}The content is based on responses from https://www.elastic.co/api/product_versions and https://snapshots.elastic.co${nl}${nl}The current update is generated based on the following requirements:${nl}${nl}Package version: ${package_version}${nl}${nl}\`\`\`json${nl}${version_requirements}${nl}\`\`\`"
```
rdner, line 25: This is not entire true, the upgrade tests run both directions and these versions are not used in just upgrade tests.

ycombinator: Thanks, I see now that the versions from this file are not just used as starting versions in the upgrade test.  For example, `TestStandaloneUpgradeRetryDownload` calls `upgradetest.PreviousMinor()` to set the end version and that function ends up looking at the versions from the file.  So I will fix the text in this PR accordingly.

But I'm not seeing where the versions from this file are being used outside of upgrade tests.  Could you point me to an example?

rdner: @ycombinator yes, I meant the usage of `upgradetest.PreviousMinor()` in other tests, there is a few of them https://github.com/search?q=repo%3Aelastic%2Felastic-agent%20upgradetest.PreviousMinor()&type=code

rdner (approved):

swiatekm:
Since this file is only used for tests, can we move it to the `testing` directory, and maybe change the name to be more indicative of the purpose? Say `.upgrade-test-agent-versions.yaml`? In general, it almost seems like an implementation detail of the `upgradetest` package, minus the fact that a mage target is used to update it (I assume that's why it's a separate file and not just some constants in Go code.).

AndersonQ (approved):

ycombinator:
> Since this file is only used for tests, can we move it to the `testing` directory, and maybe change the name to be more indicative of the purpose? Say `.upgrade-test-agent-versions.yaml`? In general, it almost seems like an implementation detail of the `upgradetest` package, minus the fact that a mage target is used to update it (I assume that's why it's a separate file and not just some constants in Go code.).

I'm a +1 to this idea.  The generic name of the file and it's current location has confused me in the past (which is part of the reason I raised this PR to switch to YAML so we could add comments in the file).  

Other reviewers (@rdner @pchila @AndersonQ @blakerouse @cmacknz), any objections to this move? I'll make sure that everyone on the team is aware of it, in case they go looking for the file.

pchila:
> I'm a +1 to this idea. The generic name of the file and it's current location has confused me in the past, which is part of the reason I raised this PR to switch to YAML so we could at least add comments in the file.
> 
> Other reviewers (@rdner @pchila @AndersonQ @blakerouse @cmacknz), any objections to this move? I'll make sure that everyone on the team is aware of it, in case they go looking for the file.

If we are discussing moving files, .agent-versions should live under `testing/integration/testdata` because that is what it effectively is, data for the upgrade tests to use.

In any case I am in favor of moving the file away from the repo root. :+1:

rdner:
I'm fine with moving the file. It might require changes in the automation code here https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-agent-versions.sh

rdner:
@ycombinator actually this script requires changes even for this original PR here https://github.com/elastic/elastic-agent/blob/80ad645335d66a620211220257d9e2873762d391/.github/workflows/bump-agent-versions.sh#L6

I missed that 🙂

ycombinator:
Based on the discussion, I've moved the `.agent-versions` file to `testing/integration/testdata/.upgrade-test-agent-versions.yml` and updated code to look at this new location and filename.  Please re-review when you get a chance.  Thanks!

swiatekm (approved):

AndersonQ (approved):

pchila (approved):

testing/integration/testdata/.upgrade-test-agent-versions.yml
```diff
@@ -0,0 +1,12 @@
+# This file is generated automatically. Please do not manually edit it.
```
cmacknz, line 1: We do manually edit this occasionally when we need to temporarily block a version because of a bug, but in general that is the exception.

cmacknz (approved):

PR #5527, opened 2024-09-11
Change maintainer label to opencontainers in Dockerfile

mauri870:
## Proposed commit message

The old way of specifying maintainers in a Dockerfile was to use the MAINTAINER instruction. This is now deprecated and the maintainer label was used instead.

Nowadays the recommended way to set a maintainer is through the `org.opencontainers.image.authors` label.

See https://docs.docker.com/reference/build-checks/maintainer-deprecated/.

### DIFF
--- a/changelog/fragments/1726238750-change-deprecated-maintainer-label-in-Dockerfile-to-org.opencontainers.image.authors.yaml
+++ b/changelog/fragments/1726238750-change-deprecated-maintainer-label-in-Dockerfile-to-org.opencontainers.image.authors.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: other
+
+# Change summary; a 80ish characters long description of the change.
+summary: change deprecated maintainer label in Dockerfile to org.opencontainers.image.authors
+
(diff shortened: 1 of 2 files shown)


swiatekm (commented):
Shouldn't this merit a changelog entry? It is a user-facing change, if a small one.

mauri870:
> Shouldn't this merit a changelog entry? It is a user-facing change, if a small one.

I had considered this when making the same change in Beats (https://github.com/elastic/beats/pull/40736). While I agree that it is a user-facing change, I highly doubt any tool would rely on a maintainer label, especially since it's not standardized anywhere and the value is a free form string. That said, I could be wrong, so I'll wait for someone with more experience to weigh in.

swiatekm:
> I had considered this when making the same change in Beats (https://github.com/elastic/beats/pull/40736). While I agree that it is a user-facing change, I highly doubt any tool would rely on a maintainer label, especially since it's not standardized anywhere and the value is a free form string. That said, I could be wrong, so I'll wait for someone with more experience to weigh in.

Are there any downsides to including a changelog entry here?

michel-laterman (approved):

mauri870:
> Are there any downsides to including a changelog entry here?

I guess there is not, I have added a changelog entry.

swiatekm (approved):

mauri870:
The build failure in k8s tests seems to be unrelated, I'm merging the changes.

pierrehilbert:
The failures seem to be related to the 9.0 version bump preparation.

mauri870:
> The failures seem to be related to the 9.0 version bump preparation.

Thanks Pierre, should we wait for the tests to be passing on main or can you force merge this change?

pierrehilbert:
I don't think we have a huge pressure about merging this PR so probably best to wait and avoid abusing the force merge option. WDYT?

pierrehilbert:
@mauri870 shouldn't we backport to 8.x branch?

mauri870:
> @mauri870 shouldn't we backport to 8.x branch?

Yes, sorry. I went through my PRs adding the backport label, seems like I missed this one.

PR #5535, opened 2024-09-12
[Helm]  Implement values.schema.json

pkoutsovasilis:
## What does this PR do?

This PR does the following:
* introduces a `value.schema.json` inside the chart which is used by Helm to validate the input values.yaml
* splits ESPlainAuth output type into ESPlainAuthBasic and ESPlainAuthAPI types, favouring explicitness, for better validation and
* introduces `hostNetwork`  as an option in an Agent preset

## Why is it important?

By having a `value.schema.json` we can reduce validation checks inside the templates of the Helm chart (less ifs inside templates). Furthermore, many text editors/IDEs offer the capability to import the schema and hence get hints and proposal write writing a `values.yaml` for the Helm chart

## Checklist

## Disruptive User Impact

N/A

### Checking validation

run the following which is actually missing required values 
```console
helm template test ./deploy/helm/elastic-agent --set kubernetes.enabled=true  --set outputs.default.type=ESSecretAuthBasic  --set agent.fleet.enabled=true --set agent.fleet.token=""
```

you should see the following errors
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
elastic-agent:
- agent.fleet: Must validate "then" as "if" was valid
- agent.fleet: Must validate at least one schema (anyOf)
- agent.fleet.url: Does not match format 'uri'
- agent.fleet.kibanaHost: Does not match format 'uri'
- agent.fleet.kibanaUser: String length must be greater than or equal to 1
- agent.fleet.kibanaPassword: String length must be greater than or equal to 1
- outputs.default: Must validate "then" as "if" was valid
- outputs.default.secretName: String length must be greater than or equal to 1
- outputs.default: Must validate all the schemas (allOf)
```

### Checking hints
1. Create a new values.yaml file with your text editor/IDE (tested with vscode)
2. Add on top this comment 
    ```yaml
    # yaml-language-server: $schema=elastic-agent/deploy/helm/elastic-agent/values.schema.json
    ```
3. For any fields you add below the above line you should get hints

## Related issues

N/A


pchila (approved):
I had a quick glance at the `values.schema.json` for a couple of validations that were moved from the helm template itself and it looks consistent.

LGTM

swiatekm (approved):
Nice!

PR #5536, opened 2024-09-12
Allow setting ES api key via env variable in container mode

### LINKED ISSUE #97: elastic-agent container output API key support
**Describe the enhancement:**

When running the elastic-agent in a container, in standalone mode, there is currently no way to specify an API key as a credential instead of username/password.

The `elastic-agent.docker.yml` file only shows username/password:
https://github.com/elastic/elastic-agent/blob/5664fa47b8103942136b3d8f18632020a4356f66/elastic-agent.docker.yml#L4-L10

swiatekm:
## What does this PR do?

Remove defaults for username and password when running in container mode, and also allow api_key to be set in that mode via the `ELASTICSEARCH_API_KEY` environment variable.

Note that the first of these changes is breaking. I've added two separate changelog entries to make this clear.

## Why is it important?

It's very useful for users to be able to use an ES api key for authentication without needing to override the whole config file.

## Checklist

## Disruptive User Impact

This is a breaking change for users who rely on the default username and password in container mode. These users will need to explicitly set these values after this change is rolled out. After the upgrade, the agent will simply fail to start, and will complain about lack of authentication credentials.

```bash
DEV=true SNAPSHOT=true EXTERNAL=true PACKAGES=docker mage package
docker run -e ELASTICSEARCH_API_KEY="***" -e ELASTICSEARCH_HOSTS="***" --rm docker.elastic.co/beats/elastic-agent-complete:9.0.0-SNAPSHOT
```

This should successfully start if the host and api key are valid.

## Related issues

Closes #97.

### DIFF
--- a/_meta/config/elastic-agent.docker.yml.tmpl
+++ b/_meta/config/elastic-agent.docker.yml.tmpl
@@ -5,8 +5,9 @@ outputs:
   default:
     type: elasticsearch
     hosts: '${ELASTICSEARCH_HOSTS:http://elasticsearch:9200}'
-    username: '${ELASTICSEARCH_USERNAME:elastic}'
-    password: '${ELASTICSEARCH_PASSWORD:changeme}'
+    username: '${ELASTICSEARCH_USERNAME:}'
+    password: '${ELASTICSEARCH_PASSWORD:}'
+    api_key: '${ELASTICSEARCH_API_KEY:}'
     preset: balanced
 
 inputs:
--- a/changelog/fragments/1726144746-container-default-credentials.yaml
+++ b/changelog/fragments/1726144746-container-default-credentials.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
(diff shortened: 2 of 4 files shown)


jlind23:
@ycombinator @nimarezainia if this is considered a breaking shouldn't make it land only in 9.x?

swiatekm:
@jlind23 that was my intent, though I believe @cmacknz mentioned wanting to consider it for 8.x as well.

FYI, I'm waiting for E2E tests to be successfully runnable on this change before submitting it properly.

jlind23:
Thanks @swiatekm! @cmacknz do we really need it for `8.x`?

> FYI, I'm waiting for E2E tests to be successfully runnable on this change before submitting it properly.

This should be shortly solved probably tomorrow EOD the latest.

pchila (approved):

jlind23:
@swiatekm Are we good to merge this in 9.0?
cc @ycombinator

swiatekm:
> @swiatekm Are we good to merge this in 9.0? cc @ycombinator

I think so. I considered adding an E2E test for it, but it ended up being a lot of setup for such a trivial check.

ycombinator:
Yes, this is a breaking change and is OK to merge (only) in `9.0`.  @swiatekm go ahead and merge whenever you get the chance.

Issue #5490, opened 2024-09-10
Pass the decrypted mTLS client certificate key to Elastic Defend


AndersonQ:
Elastic Defend currently does not support passphrase-protected private keys for mTLS communication with Fleet Server. To enable secure communication between Elastic Defend and Fleet Server when using passphrase-protected keys, the elastic-agent should decrypt the private key and pass the decrypted key to Elastic Defend. This should be done without requiring Elastic Defend to implement support for passphrase handling. Instead, leverage the existing `fleet.ssl.*` configuration block within Elastic Defend to provide the decrypted key directly. They are following the Filebeat ssl configuration as a spec: https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#client-key

**Acceptance Criteria:**

* The Elastic Agent injects the decrypted private key into the `fleet.ssl.key` configuration setting within Elastic Defend's configuration.
* Unit tests verify that the decrypted key is correctly injected into the Elastic Defend configuration.
  - extend the [`TestEndpointSignedComponentModifier`](https://github.com/elastic/elastic-agent/blob/8c43ad333837380f1646c3b007ff6339b1593a94/internal/pkg/agent/application/endpoint_signed_component_modifier_test.go#L20) to include a check for the mTLS client certificate key decryption.
* Integration tests: **Out of scope for this ticket** as it's require https://github.com/elastic/elastic-agent/issues/4903 to be done first.

**Additional Notes:**

* This issue is blocked until https://github.com/elastic/elastic-agent/issues/5489 is completed.

cmacknz:
FYI @nfritts, no work from endpoint should be required here since it appears that endpoint already accepts client certificates and keys for mTLS.

Issue #4903, opened 2024-06-11
Improve test proxy/mock fleet server to proxy HTTPS connections


pchila:
During development of #4770 it was not possible to test a scenario where elastic-agent installs and enrolls to an http mock fleet server using TLS going through an https proxy because:
- Using an http fleet endpoint requires specifying `--insecure` at install even when we have a correctly configured https proxy (no TLS validation would then happen) (see #4896)
- I tried having the proxy support https requests doing tunneling similar to [goproxy](https://github.com/elazarl/goproxy/blob/7cc037d33fb57d20c2fa7075adaf0e2d2862da78/https.go#L127-L156), it didn't work but here's the [code](https://github.com/elastic/elastic-agent/commit/ea748115e579dfd8326f58de563ae4021ded26f6) for reference

The testcase is [`TLSEnrollProxy-mTLSProxyInThePolicy`](https://github.com/elastic/elastic-agent/pull/4770/files#diff-1fbc90643f8ebd353bb15ab983de102a2d80a949dffdb9e5ba118548341c6bc6R523-R666) where we try to test this scenario.

In order to test such scenario we would need at least one of 2 things:
- an https proxy that can terminate TLS and forward plain http requests to the mock fleet server (either extend proxy test or swap out for some other proxy implementation)
- have an https fleet endpoint (either mocked or the real thing in a cloud deployment), ideally where we can set up (m)TLS with custom CAs (this is not possible at the moment on ESS)

### Conclusion:
[testproxy](https://github.com/elastic/elastic-agent/tree/main/testing/proxytest) needs to proxy HTTPS requests. 
Once https://github.com/elastic/elastic-agent/issues/5715 is completed, implement the chosen solution on [testproxy](https://github.com/elastic/elastic-agent/tree/main/testing/proxytest)

AndersonQ:
I might be worth looking at how [Caddyserver](https://caddyserver.com/) [does that](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/reverseproxy/reverseproxy.go#L391) as well

PR #5542, opened 2024-09-13
Decrypt client TLS certificate key for Elastic Defend

AndersonQ:
## What does this PR do?

It adds EndpointTLSComponentModifier which will check if the client certificate key is encrypted, if so, it'll decrypt the key and pass it decrypted to endpoint (Elastic Defend)

## Why is it important?

Elastic Defend does not support passphrase-protected certificate key, but the agent does. It'll allow the agent to receive passphrase protected client certificate key and still work when Elastic Defend is installed.

## Checklist

## Author's checklist

**Acceptance Criteria:**

- [x] The Elastic Agent injects the decrypted private key into the `fleet.ssl.key` configuration setting within Elastic Defend's configuration.
- [x] Unit tests verify that the decrypted key is correctly injected into the Elastic Defend configuration.
    - [x] extend the [`TestEndpointSignedComponentModifier`](https://github.com/elastic/elastic-agent/blob/8c43ad333837380f1646c3b007ff6339b1593a94/internal/pkg/agent/application/endpoint_signed_component_modifier_test.go#L20) to include a check for the mTLS client certificate key decryption.
    - [x] Integration tests: **Out of scope for this ticket** as it's require https://github.com/elastic/elastic-agent/issues/4903 to be done first.

## Disruptive User Impact

 - None

adjust the IPs/hostnames as needed

### you might need to build a 8.16 agent out of main:
```
AGENT_PACKAGE_VERSION=8.16.0-SNAPSHOT BEATS_VERSION=8.16.0-SNAPSHOT DEV=true SNAPSHOT=true EXTERNAL=true PLATFORMS="linux/amd64" PACKAGES="tar.gz" mage package
```

### create 2 TLS certificates
use `elastic-agent-libs/testing/certutil/cmd` to create the certificates. Make sure to use `elastic-agent-libs@v0.12.0+` which supports generating RSA certificates as Elastic Defend only accepts RSA certificates.

```
go run main.go --prefix server --pass server --ips 10.80.40.162,127.0.0.1 --rsa
go run main.go --prefix client --pass client  --ips 10.80.40.42,127.0.0.1 --rsa
```
you should have:
```
-rw------- 1 ainsoph ainsoph  288 Sep 20 18:35 client-ca_key.pem
-rw------- 1 ainsoph ainsoph  899 Sep 20 18:35 client-ca.pem
-rw------- 1 ainsoph ainsoph  399 Sep 20 18:35 client-localhost_enc-key.pem
-rw------- 1 ainsoph ainsoph  288 Sep 20 18:35 client-localhost_key.pem
-rw------- 1 ainsoph ainsoph    6 Sep 20 18:35 client-localhost-passphrase
-rw------- 1 ainsoph ainsoph  916 Sep 20 18:35 client-localhost.pem

-rw------- 1 ainsoph ainsoph  288 Sep 20 18:27 server-ca_key.pem
-rw------- 1 ainsoph ainsoph  904 Sep 20 18:27 server-ca.pem
-rw------- 1 ainsoph ainsoph  399 Sep 20 18:27 server-localhost_enc-key.pem
-rw------- 1 ainsoph ainsoph  288 Sep 20 18:27 server-localhost_key.pem
-rw------- 1 ainsoph ainsoph    6 Sep 20 18:27 server-localhost-passphrase
-rw------- 1 ainsoph ainsoph  916 Sep 20 18:27 server-localhost.pem
```

### add a fleet server with mTLS

```
elastic-agent install -nf \
--url=https://10.80.40.162:8220 \
--fleet-server-es=https://fleet.elastic-cloud.com:443 \
--fleet-server-service-token=a-token \
--fleet-server-policy=fleet-server-policy \
--certificate-authorities=/root/certs/server-ca.pem,/root/certs/client-ca.pem,/etc/ssl/certs/ca-certificates.crt \
--fleet-server-cert=/root/certs/server-localhost.pem \
--fleet-server-cert-key=/root/certs/server-localhost_enc-key.pem \
--fleet-server-cert-key-passphrase=/root/certs/server-localhost-passphrase \
--elastic-agent-cert=/root/certs/client-localhost.pem \
--elastic-agent-cert-key=/root/certs/client-localhost_enc-key.pem \
--elastic-agent-cert-key-passphrase=/root/certs/client-localhost-passphrase \
--fleet-server-client-auth=required \
--fleet-server-port=8220
```

### add an agent to that policy
```
elastic-agent install -nf \
--url=https://10.80.40.162:8220 \
--enrollment-token=a-token
--certificate-authorities=/root/certs/server-ca.pem,/etc/ssl/certs/ca-certificates.crt \
--elastic-agent-cert=/root/certs/client-localhost.pem \
--elastic-agent-cert-key=/root/certs/client-localhost_enc-key.pem \
--elastic-agent-cert-key-passphrase=/root/certs/client-localhost-passphrase
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5490

## Questions to ask yourself

- ...


blakerouse (changes requested):
I am worried about the performance of this component modifier. Most modifiers should just do some small movements or inject a little bit of extra configration.

This modifier is read a file and decrypting it every time the component model changes, which can be a lot. When using say the kubernetes provider with composable inputs it will recompute the model on every new pod state.

I believe this needs to be changed to only do that work once and only inject the cached content.

AndersonQ:
> I am worried about the performance of this component modifier. Most modifiers should just do some small movements or inject a little bit of extra configration.
> 
> This modifier is read a file and decrypting it every time the component model changes, which can be a lot. When using say the kubernetes provider with composable inputs it will recompute the model on every new pod state.
> 
> I believe this needs to be changed to only do that work once and only inject the cached content.

good point, even thought I'm not sure if defend would be used on such environments. Anyway I added a local cache for the component modifier and it's also thread-safe as the modifiers might be invoked from the coordinator main goroutine and for diagnostics as well.

AndersonQ:
@blakerouse, just to let you know I added a cache and it's ready for re-review

internal/pkg/agent/application/endpoint_component_modifier.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+}
+
+func loadCertificatesWithCache(log *logger.Logger, cache *tlsCache, keyPassPath string, certPath string, keyPath string) (string, string, error) {
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+
+	// cache hit
+	// hot reload of TLS files isn't supported, thus using the path as key is
+	// fine.
```
blakerouse, line 188: How is this not supported? Isn't it possible for the certificate and key to come from Fleet where those values can be changed?

Even if it is the case that they con't change, would it be difficult here to make this just a little bit smarter? Like if it does change it does the correct thing. That would future proof this code path so if they can be updated to be changed this code path doesn't also need to be updated.

blakerouse (commented):
Thanks for adding the cache. I do have another question related.

internal/pkg/agent/application/endpoint_component_modifier.go
AndersonQ: > How is this not supported? Isn't it possible for the certificate and key to come from Fleet where those values can be changed?

I think you misunderstood it. What I'm saying to be unsupported is hot reload of the files, changing the file on disk and without restarting the agent, the agent pickup the new file content. Given that isn't supported, using the file path as cache key is fine.

If fleet-server sends a new config, a different file path, then it'd be a cache miss, the new file would be read, decrypted and added to the cache.

VihasMakwana:
@AndersonQ I think you also need to update all occurrences of `certutil.GenerateChildCert` to make CI green.

AndersonQ:
> @AndersonQ I think you also need to update all occurrences of `certutil.GenerateChildCert` to make CI green.

there was a bug in the test. I'm opening a PR to fix it

AndersonQ:
> @AndersonQ I think you also need to update all occurrences of `certutil.GenerateChildCert` to make CI green.

the issue is with the test, I'm fixing it on https://github.com/elastic/elastic-agent/pull/5741

internal/pkg/agent/application/endpoint_component_modifier.go
AndersonQ: @blakerouse, so, what do u think?

blakerouse: @AndersonQ can you point to the code where if the file path changes that it would result in it being used? What I see from the code if the passcode file path doesn't change then it will not be read again at all.

https://github.com/elastic/elastic-agent/pull/5542/files#diff-10a613fe5fe9a1e7985f230c74d11378c33f842e882e21bb3af5be2bea680e53R192

Maybe I am miss-understanding the flow here, but to me that means it will just check that and move on.

andrzej-stencel (approved):

blakerouse (approved):
Thanks for fixing the caching key!

Looks good now.

cmacknz:
Is there anywhere in our logs or diagnistcs .yaml files where the decrypted key could be logged accidentally (expected-config.yaml maybe)? Did we look through the diagnostics to check for this?

The file path likely won't be marked as a secret explicitly because it is obscured by default.

cmacknz:
> However I must say I haven't done explicit test for that

I don't think we need a test, but a manual spot check searching through diagnostics should be done. Leaked TLS private keys is a bad time for everybody.

AndersonQ:
> a manual spot check searching through diagnostics should be done. Leaked TLS private keys is a bad time for everybody.

we could add it to as a requirement on one of the manual tests for new releases. It'd cover this and any other change we might make

PR #5547, opened 2024-09-16
Add ability to run the same tests with multi docker image variants

blakerouse:
## What does this PR do?

Extends the integration testing framework for Kubernetes to allow multiple docker images to be used for the tests.

## Why is it important?

Providers better coverage of our docker variants and it allows the tests author to write tests targeting different image variants.

## Checklist

- `INSTANCE_PROVISIONER="kind" SNAPSHOT="true" mage integration:kubernetes`
- `INSTANCE_PROVISIONER="kind" SNAPSHOT="true" mage integration:kubernetesMatrix`
- `INSTANCE_PROVISIONER="kind" SNAPSHOT="true" TEST_PLATFORMS="kubernetes/arm64/1.30.2/wolfi" mage integration:kubernetesMatrix`


cmacknz (commented):
Controlling+filtering by docker image should be added to https://github.com/elastic/elastic-agent/blob/main/docs/test-framework-dev-guide.md for the case where you only want to test one of these.

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -60,7 +62,11 @@ func TestKubernetesAgentStandaloneKustomize(t *testing.T) {
 		Local: false,
 		Sudo:  false,
 		OS: []define.OS{
-			{Type: define.Kubernetes},
+			{Type: define.Kubernetes, DockerImage: "docker.elastic.co/beats/elastic-agent"},
+			{Type: define.Kubernetes, DockerImage: "docker.elastic.co/beats/elastic-agent-complete"},
+			{Type: define.Kubernetes, DockerImage: "docker.elastic.co/beats/elastic-agent-ubi"},
```
cmacknz, line 67: For 9.0 the elastic-agent and elastic-agent-ubi will eventually be the same, so this is redundant there but not in 8.x where Ubuntu will continue to exist.

pkg/testing/kubernetes/kind/provisioner.go
```diff
@@ -72,46 +76,33 @@ func (p *provisioner) SetLogger(l runner.Logger) {
 }
 
 func (p *provisioner) Supported(batch define.OS) bool {
-
-	supported := batch.Type == define.Kubernetes && batch.Arch == runtime.GOARCH && (batch.Distro == "" || batch.Distro == "kind")
```
pkoutsovasilis, line 76: the Distro logical check was here to allow a k8s test run only for a specified provider; "" means undefined thus it can run under any provider, "kind" runs only run under a Kind-based k8s cluster. Do we want to remove this specification?

pkg/testing/kubernetes/kind/provisioner.go
```diff
@@ -72,46 +76,33 @@ func (p *provisioner) SetLogger(l runner.Logger) {
...
-
-	if supported && batch.Version != "" {
```
pkoutsovasilis, line 78: the Version logical check here was to allow a k8s test to run only for a specific k8s version. I admit that being able to define version thresholds and not fixed one would be more handy, but with that said, do we want remove this capability?

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -60,7 +62,11 @@ func TestKubernetesAgentStandaloneKustomize(t *testing.T) {
 		Local: false,
 		Sudo:  false,
 		OS: []define.OS{
-			{Type: define.Kubernetes},
+			{Type: define.Kubernetes, DockerImage: "docker.elastic.co/beats/elastic-agent"},
```
pkoutsovasilis, line 65: I think it would be easier to maintain if the list of  all images Docker to test was part of the testing framework and only when a test needs to runs under specific ones define them explicitly. In simple terms a test that has `{Type: define.Kubernetes }` should be running under all Docker images.

Also maybe we want to exclude the repository part of the Docker image as this can change and go only by image name?!

pkoutsovasilis:
Overall I like this PR! I left some comments which are more points for a quick discussion around which fields a k8s test could define, namely Distro, Version, etc. but I am happy for +1 as soon as we agree on whether we do or don't need them 🙂

testing/integration/kubernetes_agent_standalone_test.go
blakerouse: I can do a followup in main to merge those into the same.

pkg/testing/kubernetes/kind/provisioner.go
blakerouse: True. I will add that back, as I could see a tests saying specifically AKS or GKE or something.

pkg/testing/kubernetes/kind/provisioner.go
blakerouse: I didn't understand why it was there. That makes more sense. I will add it back, with a code comment so the next developer doesn't remove it.

testing/integration/kubernetes_agent_standalone_test.go
blakerouse: I didn't do that mainly because I didn't want to hard code the image names in the testing framework. By placing them directly into the tests it removed the need for me to hard-code anything.

`{Type: define.Kubernetes}` in the testing framework case would not auto run the test on every docker variant it would just pick the first one that is defined as supported. Only in the case of `mage integration:matrix` would it then perform the tests on all versions of the image.

I lean towards keeping it this way, let me know your opinion.

pkg/testing/kubernetes/kind/provisioner.go
blakerouse: I take that back, my change auto creates a kind cluster for that version of k8s even in the environment variable has a `K8S_VERSION` that is different. `K8S_VERSION` in this case is only the default in the case that a test is not specific.

blakerouse:
> Controlling+filtering by docker image should be added to https://github.com/elastic/elastic-agent/blob/main/docs/test-framework-dev-guide.md for the case where you only want to test one of these.

This is currently not supported. It will be even more work to allow filtering to a specific docker image for running tests. I am looking to just to get this done for the `elastic-agent-server` container image. Adding that will make this take even longer, its already snowballed into this work.

pkg/testing/kubernetes/kind/provisioner.go
pkoutsovasilis: ok based on this comment and https://github.com/elastic/elastic-agent/pull/5547#discussion_r1765159296 I just want to understand which layer holds responsibility for these testing knobs, for example; the "supported" k8s version(s) is mandated by the env vars and if a test doesn't support them it is just skipped [CI pipeline responsibility] or based on the defined versions of the k8s tests these are the versions of the clusters that are gonna be defined [developer writing the test responsibility]?! The same goes for k8s providers and docker images. Personally I more accustomed to defining what is allowed or not through env vars [CI pipeline responsibility] and tests either getting skipped or running based on the former. But here it is seems to me that we are mix matching them. Maybe I am missing something!? 🙂

blakerouse: The more I think about this, the kubernetes mode of the integration testing suite is very different then how the standard mode works. The standard modes defines everything in code, and then a caller can say what they want to allow to run. That is defined by the testing runner even before the provider is invoked.

That allows for things like `mage integration:matrix` to perform running a larger set of tests across a whole matrix of versions. It think I need to adjust this PR to take that same approach and `mage integration:matrix` would then run multiple versions of kubernetes, but the standard `mage integration:test` would only run a single version.

That would make this align correctly with the standard operating mode of the integration suite.

pkoutsovasilis: ok from this comment I understand that the knobs are controlled by the defines in each test and, if that's the case indeed, then yes the k8s integration tests need some shifting towards that direction. You wanna leave that for a subsequent PR or tackle it in this one?

blakerouse:
@pkoutsovasilis Please give this another look, I took a different path. This now more aligns with OS testing.

pkg/testing/kubernetes/supported.go
```diff
@@ -0,0 +1,77 @@
...
+// versions defines the list of supported version of Kubernetes.
+var versions = []define.OS{
+	Kubernetes_1_30,
+	Kubernetes_1_29,
+	Kubernetes_1_28,
+}
+
+// variantToImage defines the mapping of the variants to image name.
+var variantToImage = map[string]string{
+	"basic":          "elastic.docker.co/beats/elastic-agent",
```
pkoutsovasilis, line 46:
```suggestion
	"basic":          "docker.elastic.co/beats/elastic-agent",
```

🙂

cmacknz:
> This is currently not supported. It will be even more work to allow filtering to a specific docker image for running tests. I am looking to just to get this done for the elastic-agent-server container image. Adding that will make this take even longer, its already snowballed into this work.

Thanks, I think we'll want this but it can be a follow up addition.

blakerouse:
@cmacknz Actually my latest commit does add this and is now supported. Example add to dev docs and in PR description.

pkoutsovasilis:
So after some offline syncing with @blakerouse I think that the code changes are looking good. Now some things I noticed while running these tests locally (with 8.16.0-SNAPSHOT version) as our CI still needs some love is the following:
* all ubi-image based tests are failing because agent pod is getting OOMKilled, we need to bump the memory limits?!
* `integration:kubernetesMatrix` mage target, which will be the default now when this PR gets merged, takes ~ 2 hours to complete, is this causing any inconvenience for us?!

blakerouse:
@pkoutsovasilis I have reduced the tests to only test basic and wolfi at the moment. This is to reduce the time spent and get around the issue with UBI. Being that the UBI is going away in 9.0, and we haven't really been testing. The complete images are just additions to the original so they should basically work the same, so removing it for now to improve performance should be okay.

I think the overall solution to reducing the time it takes is to split up the work across multiple runners, but I want that done in a way where its not hard-coded in the YAML. I am already working on that here with buildkite, where it uses the batches to generate a buildkite pipeline. I think the same approach should be taken here, so its only required to update the golang testing framework for new versions and not also adjust matching YAML pipelines.

See: https://github.com/elastic/elastic-agent/pull/5391

cmacknz (approved):
Tried it locally and works as expected. LGTM.

.buildkite/pipeline.yml
```diff
@@ -198,23 +198,6 @@ steps:
         retry:
           manual:
             allowed: true
-      - label: "K8s integration tests"
-        key: "k8s-integration-tests"
-        env:
-          K8S_VERSION: "v1.30.2,v1.29.4,v1.28.9"
```
mauri870, line 204: Noticed that you changed this to only v1.30.2, does that mean we are not testing against the other stable versions of kubernetes? If so, is that desired? Thanks.

pkoutsovasilis: so this change means that `mage integration:kubernetes` indeed will run the test against a cluster that its version is defined in the env var of  K8S_VERSION. However this PR added also `mage integration:kubernetesMatrix`, which is what is invoked in our CI, that runs the k8s integration tests against multiple k8s versions defined [here](https://github.com/elastic/elastic-agent/blob/main/pkg/testing/kubernetes/supported.go#L22)

mauri870: Ha, it makes sense now. Thanks for the explanation!

PR #5549, opened 2024-09-17
[chore][otel] Enable persistence in otel.yml

VihasMakwana:
- Enhancement
- Docs

## What does this PR do?

This PR adds the `filestorage` extension in `otel.yaml` and updates the documentation.
Updates `otel` dependencies to 0.110.0

## Why is it important?

By default, the OpenTelemetry Collector operates in a stateless mode, meaning it does not save file offsets to disk. To resolve this, we enable persistence in the default configuration files.

## Checklist

- Build from this branch
- Keep the config default and run `elastic-agent otel --config otel.yaml`
- Ingest some data and stop the process.
- Restart the process and ingest some more data.
- You shouldn't see any data loss/duplication.
## Related issues

- Closes https://github.com/elastic/opentelemetry-dev/issues/281


VihasMakwana:
@elastic/elastic-agent-control-plane 

 I propose two options for default directory:

1. Set it as per current sample configurations and modify them during the onboarding process. Currently, the sample configs use STORAGE_DIR as the default directory option
2. Follow Filebeat's [directory layout](https://www.elastic.co/guide/en/beats/filebeat/8.15/directory-layout.html) and set the default to a path under `path.data`.

For the later option, we need to implement a config converter which maps `path.data` to `/opt/Elastic/Agent/data` (in case of Linux).

Let me know how do you feel about this proposal.

cc: @ycombinator @pierrehilbert @andrzej-stencel

VihasMakwana:
Introducing a new config converter would be straightforward and could be easily integrated into our distribution.

cmacknz:
As discussed let's leave this as an environment variable. How the collector is run and installed (if it is installed as a service at all) is currently completely up to the user and this goes along with that.

The only place we likely want a default is in our reference Kubernetes configurations, which can set the environment variable or set a default in the configuration it includes directly.

In general we do not want to use any of the paths that are also used by Beats or Agent, because a standalone OTel collector must be able to run alongside them without issue. 

In our OTel distribution I think we solve this by having an `install` command that puts this in whatever base install directory we pick (/opt/Elastic, C:\Program Files\Elastic, or /Library/Elastic) but we need an installer that runs as a privileged user to do that. You can't do it with config alone.

pchila (approved):

otel.yml
```diff
@@ -24,6 +25,9 @@ extensions:
   health_check:
     endpoint: "localhost:13133"
     path: "/health/status"
+  file_storage/filelogreceiver:
+    create_directory: true
+    directory: ${env:STORAGE_DIR}
 
 service:
   extensions: [health_check, memory_limiter]
```
andrzej-stencel, line 33: The `file_storage/filelogreceiver` extension must also be added to `service::extensions` list, right?

internal/pkg/otel/README.md
```diff
@@ -78,3 +78,35 @@ This section provides a summary of components included in the Elastic Distributi
...
+    storage: file_storage/filelogreceiver
+extensions:
+  file_storage/filelogreceiver:
+    directory: ${env:STORAGE_DIR}
+exporters:
+  ...
+processors:
+  ...
+service:
+  extensions: [file_storage]
```
andrzej-stencel, line 102:
```suggestion
  extensions: [file_storage/filelogreceiver]
```

changelog/fragments/1726572104-enable-persistence-by-default.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Enable persistence in the configuration provided with our OTeL distribution.
```
andrzej-stencel, line 14:
```suggestion
summary: Enable persistence in the configuration provided with our OTel Collector distribution.
```

go.mod
```diff
@@ -87,49 +87,49 @@ require (
 
 require (
 	github.com/distribution/reference v0.6.0 // indirect
-	github.com/open-telemetry/opentelemetry-collector-contrib/connector/spanmetricsconnector v0.109.0
```
andrzej-stencel, line 90: This update is not related to this pull request, right? Let's do this separately if possible.

otel.yml
```diff
@@ -24,6 +25,9 @@ extensions:
   health_check:
     endpoint: "localhost:13133"
     path: "/health/status"
+  file_storage/filelogreceiver:
+    create_directory: true
+    directory: ${env:STORAGE_DIR}
```
andrzej-stencel, line 30: The addition of the File Storage extension may cause problems for users using this configuration with read-only filesystems, like in Docker, in Kubernetes, unless they configure a writable storage layer in their containers/pods. Should we be concerned about this?

go.mod
VihasMakwana: Yeah, you're right but I needed to update the `filestorage` extension to the latest version, so I updated all the OTeL dependencies to the latest version as well to maintain consistency. Otherwise, some would dependencies be at `0.109.0` and others at `0.110.0`.

otel.yml
VihasMakwana: > The addition of the File Storage extension may cause problems for users using this configuration with read-only filesystems, like in Docker, in Kubernetes, unless they configure a writable storage layer in their containers/pods. Should we be concerned about this?

I believe we will have a separate recommendation in our k8s configs based on @cmacknz's [comment](https://github.com/elastic/elastic-agent/pull/5549#issuecomment-2361000323).

go.mod
andrzej-stencel: Oh, OK then.

otel.yml
andrzej-stencel: This change breaks running Elatic Agent as OTel Collector in Docker. I'd rather have this change applied in a way that doesn't break running the EDOT Collector in Docker. This should "just work", just like it "just works" with the regular Elastic Agent.

```console
$ docker run --rm -ti --entrypoint="elastic-agent" -v ${HOME}/elastic/elastic-agent/otel.yml:/usr/share/elastic-agent/otel.yml docker.elastic.co/beats/elastic-agent:9.0.0-SNAPSHOT otel
Starting in otel mode
2024-09-30T09:06:39.260Z	info	service@v0.110.0/service.go:137	Setting up own telemetry...
2024-09-30T09:06:39.261Z	info	service@v0.110.0/service.go:186	Skipped telemetry setup.
2024-09-30T09:06:39.261Z	info	builders/builders.go:26	Development component. May change in the future.	{"kind": "exporter", "data_type": "logs", "name": "debug"}
2024-09-30T09:06:39.262Z	info	builders/extension.go:50	Development component. May change in the future.	{"kind": "extension", "name": "memory_limiter"}
2024-09-30T09:06:39.262Z	info	memorylimiter/memorylimiter.go:75	Memory limiter configured	{"kind": "extension", "name": "memory_limiter", "limit_mib": 700, "spike_limit_mib": 180, "check_interval": 1}
2024-09-30T09:06:39.252Z	warn	envprovider@v1.16.0/provider.go:59	Configuration references unset environment variable	{"name": "STORAGE_DIR"}
failed to build extensions: failed to create extension "file_storage/filelogreceiver": mkdir /var/lib/otelcol: permission denied
```

VihasMakwana: @andrzej-stencel I see what you mean. The error is due to an empty `env:STORAGE_DIR` and hence, the `filestorage` extension defaults to `/var/lib/otelcol`. Since the default user in the Docker container is `elastic-agent`, this causes a permission denied error.

If I use any of following commands, it works.
```bash
docker run --rm -ti --entrypoint="elastic-agent" --mount type=bind,source=$(pwd)/data,target=/usr/share/otel -e STORAGE_DIR=/usr/share/otel/  docker.elastic.co/beats/elastic-agent:9.0.0-SNAPSHOT otel
```
or
```bash
docker run --user root --rm -ti --entrypoint="elastic-agent" docker.elastic.co/beats/elastic-agent:9.0.0-SNAPSHOT otel
```

--- 

I wonder how should we approach this. Maybe add documentation? or update EDOT according to @cmacknz's [comment](https://github.com/elastic/elastic-agent/pull/5549#issuecomment-2361000323). Or do both.

> In our OTel distribution I think we solve this by having an install command that puts this in whatever base install directory we pick (/opt/Elastic, C:\Program Files\Elastic, or /Library/Elastic) but we need an installer that runs as a privileged user to do that. You can't do it with config alone.

@cmacknz @andrzej-stencel I'd appreciate your opinions over this!

VihasMakwana: I agree with you that is should "just work".

cmacknz: In the container we control the paths, and can pick a default that is writable by our user in the container. Elastic agent uses the STATE_PATH and puts it under /usr/share/elastic-agent.

https://github.com/elastic/elastic-agent/blob/95021e1287a2a637df9c84c0e4bff0bfc661f90c/internal/pkg/agent/cmd/container.go#L48-L49

We could add an `/otelcol` sub-directory to `/usr/share/elastic-agent` for one possibility.

cmacknz: Possibly we can reuse the `STATE_PATH` variable and change which sub-directory we use based on whether the entrypoint is the otel command or the elastic-agent container command.

VihasMakwana: Makes sense.

VihasMakwana:
@andrzej-stencel @cmacknz PTAL!

internal/pkg/agent/cmd/otel.go
```diff
@@ -118,3 +124,18 @@ func runCollector(cmdCtx context.Context, configFiles []string) error {
...
+
+func prepareEnv() error {
+	if _, ok := os.LookupEnv("STORAGE_DIR"); !ok {
+		// STORAGE_DIR is not set. Set it to ${path.Top()}/otel_registry because we do not want to use any of the paths, that are also used by Beats or Agent
+		// because a standalone OTel collector must be able to run alongside them without issue.
+
+		// The filestorage extension will handle directory creation since create_directory: true is set by default.
+		// If the user hasn’t specified the env:STORAGE_DIR in filestorage, they may have opted for a custom path, and the extension will create the directory accordingly.
+		// In this case, setting env:STORAGE_DIR will have no effect.
+		if err := os.Setenv("STORAGE_DIR", filepath.Join(paths.Top(), "otel_registry")); err != nil {
```
VihasMakwana, line 136: We do this to avoid introducing a breaking change, as we expect the collector to function smoothly without issues.

VihasMakwana:
@andrzej-stencel @cmacknz PTAL 😅

andrzej-stencel (approved):
I'm not fond of the `prepareEnv` function setting the `STORAGE_DIR` env var irrespective of whether the env var is used in configuration or not. On the other hand, I don't have an alternative "perfect" solution, so let's go with this approach now and perhaps change if we come up with a better way.

internal/pkg/otel/templates/README.md.tmpl
```diff
@@ -74,3 +74,53 @@ This section provides a summary of components included in the Elastic Distributi
 | [{{ .Name }}]({{ .Link }}) | {{ .Version }} |
 {{ end -}}
 {{ end -}}
+
+
+## Persistense in OpenTelemetry Collector
```
cmacknz, line 79:
```suggestion
## Persistence in OpenTelemetry Collector
```

internal/pkg/otel/templates/README.md.tmpl
```diff
@@ -74,3 +74,53 @@ This section provides a summary of components included in the Elastic Distributi
...
+      exporters: [...]
+```
+
+> [!WARNING]  
+Removing the storage key from the filelog section will disable persistence, which will lead to data duplication or loss when the collector restarts.
+
+> [!IMPORTANT]  
+If you remove the `create_directory: true` option, you'll need to manually create a directory to store the data. You can ignore this option if the directory already exists.
+
+### Persistense in standalone Docker mode
```
cmacknz, line 115:
```suggestion
### Persistence in standalone Docker mode
```

internal/pkg/agent/cmd/otel.go
cmacknz: nit: I would just call the path `otel` instead of `otel_registry` since there is no guarantee we only use this directory to store file log registries long term.

cmacknz: Also, why can't we just reuse the STATE_PATH environment variable the container already supports so we don't need to add a new one? The default value is also writable by the elastic agent container user.

One reason to do this is it won't be a breaking change if we decide to get rid of STORAGE_DIR later.

VihasMakwana: That makes sense. I'll update PR.

VihasMakwana:
@cmacknz I've reused the STATE_PATH and changed directory name. PTAL!

internal/pkg/agent/cmd/otel.go
```diff
@@ -118,3 +123,22 @@ func runCollector(cmdCtx context.Context, configFiles []string) error {
 	return goerrors.Join(errs...)
 
 }
+
+func prepareEnv() error {
+	if _, ok := os.LookupEnv("STORAGE_DIR"); !ok {
```
cmacknz, line 128: Can you use STATE_PATH here as well?

What I am trying to do is avoid introducing a new environment variable we may decide to remove in the future, trying to keep our public API as small as possible while we figure out how everything fits together. 

STATE_PATH isn't going anywhere without making it a breaking change for many people, so it doesn't increase the surface area of configuration we need to maintain.

cmacknz (approved):
Thanks!

PR #5580, opened 2024-09-20
Reduce the amount of stored ReplicaSet data

### LINKED ISSUE #4729: Kubernetes metadata overwhelms memory limits in the Agent process
Diagnostics from production Agents running on Kubernetes show:
- The elastic-agent process itself uses more memory than all its configured inputs combined.
- Within the elastic-agent process, more than 90% of memory use is in Kubernetes helpers. 70% of that is from `elastic-agent-autodiscover` and the other 20% is from helpers internal to `elastic-agent`.

We need to understand why the Kubernetes helpers are using so much memory, and find a way to mitigate it.

## Definition of done
- [ ] Provide steps for a reproducible setup that can demonstrate the aforementioned memory usage with an Agent diagnostic
- [ ] Attach Agent diagnostic to this issue to use as a baseline, so we can compare against it when improvements are made
- [ ] Reduce memory use by Kubernetes helpers from 90% to TBD% (TBD, at the moment, until we've done more investigation)

swiatekm:
## What does this PR do?

Use a transform function to drop all ReplicaSet data in the kubernetes provider, except for the owner references, which we need to find the Deployment name.

Unfortunately, the autodiscovery library doesn't let us pass the transform function in, so I had to copy their implementation and make the change myself. The plan is to upstream this change later.

## Why is it important?

In clusters with a lot of Deployments, we end up storing a lot of ReplicaSets in the local cache of each agent, resulting in significant unnecessary memory consumption.

## Checklist

There isn't an easy way. You can start a local K8s cluster, create a lot of Deployments, deploy Agent using the Helm Chart [here](https://github.com/elastic/elastic-agent-service/tree/main/helm), and check the memory consumption.

## Related issues

- Partially resolves #4729 
- Closes https://github.com/elastic/elastic-agent/issues/5623


swiatekm:
Note: I'd like to verify that this actually improves the situation in a real cluster. I'm working with our SRE team to do this. Until then, I'm keeping this PR as Draft and not adding unit tests or changelog entries.

internal/pkg/composable/providers/kubernetes/pod.go
```diff
@@ -539,3 +542,19 @@ func hintsCheck(annotations mapstr.M, container string, prefix string, validate
...
+func removeUnnecessaryReplicaSetData(obj interface{}) (interface{}, error) {
+	old, ok := obj.(*v1.ReplicaSet)
+	if !ok {
+		return nil, fmt.Errorf("obj is not a ReplicaSet")
+	}
+	transformed := v1.ReplicaSet{}
+	transformed.ObjectMeta = kubernetes.ObjectMeta{
+		Name:            old.GetName(),
+		Namespace:       old.GetNamespace(),
+		OwnerReferences: old.GetOwnerReferences(),
```
pkoutsovasilis, line 557: I think that if you don't include `ResourceVersion: old.GetResourceVersion()` here then you effectively make the `opts.IsUpdated` check inside UpdateFunc of the informer to be always true

pkoutsovasilis:
Observations:

1. I like the introduction of the transform func and I left a comment on it
2. I am wondering the following; since it seems we want only the Metadata why aren't we using something like a metadata shared informer?! e.g. snippet for replicasets
    ```
    metadataClient, err := metadata.NewForConfig(config) // "k8s.io/client-go/metadata"
    if err != nil {
      panic(err.Error())
    }

    // Define the resource to watch metadata for
    gvr := schema.GroupVersionResource{
        Group:    "apps",
	Version:  "v1",
	Resource: "replicasets",
    }

    resyncPeriod := 10 * time.Second

    indexers := cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}

    // Create the SharedIndexInformer using the provided code snippet
    informer := cache.NewSharedIndexInformer(
        &cache.ListWatch{
            ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
                return metadataClient.Resource(gvr).List(ctx, options)
            },
            WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
                return metadataClient.Resource(gvr).Watch(ctx, options)
            },
        },
        &metav1.PartialObjectMetadata{},
        resyncPeriod,
        indexers,
    )
    ```
    after a quick local-testing because PartialObjectMetadata struct is way smaller than ReplicaSet struct used in this snippet the allocations to unmarshall the former are way less than the ones required for the latter and this is evident in List and Stream funcs of the informer.

internal/pkg/composable/providers/kubernetes/watcher.go
```diff
@@ -0,0 +1,285 @@
...
+	// DeletionHandlingMetaNamespaceKeyFunc that we get a key only if the resource's state is not Unknown.
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	if err != nil {
+		return
+	}
+	if deleted, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		w.logger.Debugf("Enqueued DeletedFinalStateUnknown contained object: %+v", deleted.Obj)
+		obj = deleted.Obj
+	}
+	w.queue.Add(&item{key, obj, state})
```
henrikno, line 229: Adding a comment here but can be it's own issue/PR. 
According to https://github.com/topfreegames/maestro/pull/473
https://github.com/kubernetes/kubernetes/issues/103789
https://github.com/kubernetes/sample-controller/blob/master/docs/controller-client-go.md
https://github.com/kubernetes/client-go/blob/master/examples/workqueue/main.go
we're supposed to only put the `key` on the work queue, and retrieve it from the indexer when processing it. Here we're storing the object and later retrieving it by key, so we're storing extra copies.

swiatekm: We should file that against https://github.com/elastic/elastic-agent-autodiscover. To be honest, I'm not convinced this queue is even necessary, but in this PR I just wanted to make the smallest amount of changes necessary.

henrikno: https://github.com/elastic/elastic-agent-autodiscover/issues/108

swiatekm:
Note: This code is going to move to the autodiscovery library, because we need to make a similar change in beats as well. I'm going to keep it here for now, and rebase once that's done. See #5580 for more details.

swiatekm:
Closing in favor of https://github.com/elastic/elastic-agent-autodiscover/pull/111. I'm going to recreate this PR after releasing a new version of autodiscover and updating go.mod in agent.

Issue #5558, opened 2024-09-18
Unable to disable Kubernetes Watchers and Leader Election for Fleet Managed Agents


btrieger:
For confirmed bugs, please report:
- Version: 8.15.1
- Operating System: Linux on Kuberentes
- Steps to Reproduce:

Attempting to deploy elastic agent on kubernetes to run threat intel integration or other api integrations. I am receiving errors related to missing permissions in kubernetes. As I am not running the kubernetes integration or monitoring kubernetes itself I shouldn't need access to the kubernetes api server to watch nodes, namespaces, and pods. I also should not need to create a lease without the kubernetes integration.

I am attempting to run fleet managed agents on kubernetes. I have deployed the following yamls:

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: elastic-agent-k8s-test
  namespace: elastic
  labels:
    app: elastic-agent-k8s-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: elastic-agent-k8s-test
  template:
    metadata:
      labels:
        app: elastic-agent-k8s-test
    spec:
      serviceAccountName: elastic-agent
      hostNetwork: false
      dnsPolicy: ClusterFirst
      securityContext:
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
      containers:
        - name: elastic-agent
          image: docker.elastic.co/beats/elastic-agent:8.15.1
          args: ["-c", "/etc/elastic-agent/agent.yml"]
          env:
            # Set to 1 for enrollment into Fleet server. If not set, Elastic Agent is run in standalone mode
            - name: FLEET_ENROLL
              value: "1"
            # Set to true to communicate with Fleet with either insecure HTTP or unverified HTTPS
            - name: FLEET_INSECURE
              value: "false"
            # Fleet Server URL to enroll the Elastic Agent into
            # FLEET_URL can be found in Kibana, go to Management > Fleet > Settings
            - name: FLEET_URL
              value: "https://076433a0edff4346b220c692d2e9c56a.fleet.us-central1.gcp.cloud.es.io:443"
            # Elasticsearch API key used to enroll Elastic Agents in Fleet (https://www.elastic.co/guide/en/fleet/current/fleet-enrollment-tokens.html#fleet-enrollment-tokens)
[...]

PR #5593, opened 2024-09-23
[helm] support using user-created serviceAccount and clusterRole

pkoutsovasilis:
## What does this PR do?

This PR introduces the capability for users to bind agent presets with custom service accounts and cluster roles, rather than relying solely on auto-generated or default configurations. This allows greater control over security and permissions for agent interactions. Additionally, it provides support for adding annotations to these bindings, allowing for custom metadata that can be leveraged by observability and monitoring tools to enhance tracking, auditing, and configuration management.

## Why is it important?

This added flexibility ensures that the Helm chart can be better aligned with organizational policies and infrastructure requirements of users.

## Checklist

## Disruptive User Impact

N/A

```
mage helm:renderExamples
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/5558 as this PR exposes `automountServiceAccountToken` for each preset which by setting it to false a user can effectively disable kubernetes provider completely

### DIFF
--- a/deploy/helm/elastic-agent/templates/agent/_helpers.tpl
+++ b/deploy/helm/elastic-agent/templates/agent/_helpers.tpl
@@ -284,10 +284,12 @@ app.kubernetes.io/version: {{ .Values.agent.version}}
 {{- $ := index . 0 -}}
 {{- $preset := index . 1 -}}
 {{- $templateName := index . 2 -}}
-{{- $presetRules := dig "rules" (list) $preset -}}
+{{- if eq ($preset).clusterRole.create true -}}
+{{- $presetClusterRoleRules := dig "rules" (list) ($preset).clusterRole -}}
 {{- $rulesToAdd := get (include $templateName $ | fromYaml) "rules" -}}
-{{- $presetRules = uniq (concat $presetRules $rulesToAdd) -}}
-{{- $_ := set $preset "rules" $presetRules -}}
+{{- $presetClusterRoleRules = uniq (concat $presetClusterRoleRules $rulesToAdd) -}}
+{{- $_ := set ($preset).clusterRole "rules" $presetClusterRoleRules -}}
+{{- end -}}
 {{- end -}}
 
 {{- define "elasticagent.preset.mutate.annotations" -}}
--- a/deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
+++ b/deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
@@ -1,5 +1,6 @@
 {{- include "elasticagent.init" $ -}}
 {{- range $presetName, $presetVal := $.Values.agent.presets -}}
+{{- if eq $presetVal.clusterRole.create true -}}
 {{/* cluster role binding is not namespace bound so let's try to give it a unique enough name */}}
 {{- $clusterRoleName := printf "agent-%s-%s-%s" $presetName $.Release.Name $.Release.Namespace -}}
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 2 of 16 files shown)


deploy/helm/elastic-agent/templates/agent/service-account.yaml
```diff
@@ -11,9 +12,10 @@ metadata:
     {{- with ($presetVal).labels -}}
     {{ toYaml . | nindent 4 }}
     {{- end }}
-  {{- with ($presetVal).annotations }}
+  {{- with ($presetVal).serviceAccount.annotations }}
```
swiatekm, line 15: Shouldn't this be both the preset-wide annotations *and* the ServiceAccount-specific ones?

deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
```diff
@@ -11,9 +12,9 @@ metadata:
     {{- with ($presetVal).labels -}}
     {{ toYaml . | nindent 4 }}
     {{- end }}
-  {{- with ($presetVal).annotations }}
+  {{- with ($presetVal).clusterRole.annotations -}}
```
swiatekm, line 15: Shouldn't this be both the preset-wide annotations *and* the ClusterRole-specific ones?

swiatekm (commented):
Overall this looks good, had questions about a few parts.

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
```diff
@@ -284,10 +284,12 @@ app.kubernetes.io/version: {{ .Values.agent.version}}
 {{- $ := index . 0 -}}
 {{- $preset := index . 1 -}}
 {{- $templateName := index . 2 -}}
-{{- $presetRules := dig "rules" (list) $preset -}}
+{{- if eq ($preset).clusterRole.create true -}}
+{{- $presetClusterRoleRules := dig "rules" (list) ($preset).clusterRole -}}
 {{- $rulesToAdd := get (include $templateName $ | fromYaml) "rules" -}}
-{{- $presetRules = uniq (concat $presetRules $rulesToAdd) -}}
-{{- $_ := set $preset "rules" $presetRules -}}
+{{- $presetClusterRoleRules = uniq (concat $presetClusterRoleRules $rulesToAdd) -}}
```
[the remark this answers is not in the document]
pkoutsovasilis: just gave this a spin and it doesn't... thx. That is not what I had in mind and I will remove it but I was feeling "safer" when I thought that this was indeed deduplicating the list entries 🙂

deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
pkoutsovasilis: I debated in my head about it, and I chose to not "merge" the two as I deemed it cleaner to have separate annotations between {deployment|statefulset|daemonset}, clusterrole, and serviceaccount. That said, no strong feelings, hence, sure I will add also the preset annotations 🙂

swiatekm: Hm, maybe this isn't that useful? In hindsight, I was thinking of labels, where having a common set for all resources makes a lot more sense than annotations, which tend to be tool-specific. If you don't want to merge them, I'm ok with it.

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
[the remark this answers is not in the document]
swiatekm: For reference, it looks like it uses [reflect.DeepEqual](https://github.com/Masterminds/sprig/blob/e708470d529a10ac1a3f02ab6fdd339b65958372/list.go#L338) for comparisons, so it should work if it's just comparing standard maps. Who know what Helm actually puts in there though.

deploy/helm/elastic-agent/templates/agent/cluster-role.yaml
pkoutsovasilis: oops missed this one and I merged them 🙂 have a look and if you don't like it I will revert it

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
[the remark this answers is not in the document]
pkoutsovasilis: I did a mistake before, `uniq` actually deduplicates even complex list entries, reverting my latest commit 🙂

swiatekm (approved):
Nice!

PR #5598, opened 2024-09-23
[Debugging] Revert "Add ability to run the same tests with multi docker image variants (#5547)

ycombinator:
This reverts commit 67b709f6e6ed24efcd3d9f191dad15893a845f33.


ycombinator:
No effect from reverting #5547 so it's not the culprit.  Closing without merging.

PR #5599, opened 2024-09-23
Add daily cron for elastic-agent

rowlandgeoff:
Adds the [daily cron that was manually configured](https://buildkite.com/elastic/elastic-agent/settings/schedules/621c1260-ba0a-4ab1-874e-0fe8b2a2797b) to the yml config.

### DIFF
--- a/.buildkite/integration.pipeline.yml
+++ b/.buildkite/integration.pipeline.yml
@@ -95,7 +95,6 @@ steps:
     command: |
       buildkite-agent artifact download build/distributions/** . --step 'package-it'
       .buildkite/scripts/steps/beats_tests.sh
-    # if: "build.env('CRON') == 'yes'"
     agents:
       provider: "gcp"
       machineType: "n1-standard-8"
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -47,6 +47,11 @@ spec:
         filter_enabled: true
         filter_condition: >-
           build.pull_request.id == null || (build.creator.name == 'elasticmachine' && build.pull_request.id != null)
+      schedules:
+        Daily main:
+          branch: main
+          cronline: "0 0 * * 1-5"
+          message: Mon-Fri daily build
       cancel_intermediate_builds: true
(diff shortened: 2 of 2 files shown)


.buildkite/integration.pipeline.yml
```diff
@@ -95,7 +95,6 @@ steps:
     command: |
       buildkite-agent artifact download build/distributions/** . --step 'package-it'
       .buildkite/scripts/steps/beats_tests.sh
-    # if: "build.env('CRON') == 'yes'"
```
pazone, line 98: Let's uncomment this line

rowlandgeoff: Agent team, can you confirm the behavior you want? Currently the serverless beats tests are running with every single build, and it's been like that for 4 months. Do you only want the serverless beats tests running once per day?

rowlandgeoff: I found the history. It looks like @pchila intentionally enabled them to run with every build - https://github.com/elastic/elastic-agent/pull/4501

blakerouse (approved):

rowlandgeoff:
I have removed the manually configured cron at https://buildkite.com/elastic/elastic-agent/settings/schedules

.buildkite/integration.pipeline.yml
pchila: Sorry, just found this PR via its backport but I can still add a bit of context:

This was supposed to be temporary to make sure that agent/beats changes on PRs would not break serverless beats tests.

The whole step was supposed to be moved to the beats CI pipeline at some point but here it still is 6 months ago.

@pierrehilbert @jlind23 do we have an idea when these tests can be moved off the agent pipeline and into beats CI?

jlind23: We were "only" waiting for buildkite to be available in the beats repository so I think we should be able to do it by now.

rowlandgeoff: Tracking with https://github.com/elastic/ingest-dev/issues/4128, currently set for sprint 37

PR #4625, opened 2024-04-26
Kubernetes standalone manifest: remove STATE_PATH, replace initcontainer image

tetianakravchenko:
## What does this PR do?

- remove STATE_PATH
- in case hints are used - `inputs.d` with all templates will be stored under `/usr/share/elastic-agent/state`
- replaced `busybox` image with the image used for the main container `docker.elastic.co/beats/elastic-agent` to avoid TLS errors

## Why is it important?

- remove STATE_PATH - to use the default config location - `/usr/share/elastic-agent/state`, in this location `hostPath` is mounted to store the agent state
- replaced `busybox` - to remove extra dependency and use the same image that is already used for the main container. With the busybox:1.28 image I got this error `wget: TLS error from peer (alert code 80): 80`, after updating the image to the latest version (1.36) I still got `wget: note: TLS certificate validation not implemented`, even though it is not the error, but as suggested in this issue - https://github.com/docker-library/busybox/issues/64#issuecomment-510194111 busybox shouldn't be used.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Author's Checklist

- [ ]

## How to test this PR locally
1. Deploy elastic-agent-standalon with uncommented hints related comments
2. deploy redis pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: redis
  annotations:
    co.elastic.hints/package: redis
    co.elastic.hints/data_streams: info, key
    co.elastic.hints/host: '${kubernetes.pod.ip}:6379'
    co.elastic.hints/info.period: 5s
    co.elastic.hints/key.period: 5s
  labels:
    k8s-app: redis
    app: redis
spec:
  containers:
  - image: redis
    imagePullPolicy: IfNotPresent
    name: redis
    ports:
    - name: redis
      containerPort: 6379
      protocol: TCP
```

3. run to check templates are present:
```
kubectl exec -it elastic-agent-standalone-6bmrm bash
...
Defaulted container "elastic-agent-standalone" out of: elastic-agent-standalone, k8s-templates-downloader (init)
[...]

Issue #3063, opened 2023-07-12
[Meta] Improving Hints Autodiscovery with Elastic Agent


gizas:
## Synopsis

Hints based Autodiscovery for Elastic Agent has been released as beta since 8.5. Information about the functionality and configuration of this feature can be found [here](https://www.elastic.co/guide/en/fleet/master/hints-annotations-autodiscovery.html) 

This is the Meta issue that tries to summarise a list of enhancements, bug fixes and documentation updates we have gathered so far in order to improve the general UX and functionality of Hints based Autodiscovery.

## List of Issues

- [ ] https://github.com/elastic/package-storage-infra/issues/479
- [ ] https://github.com/elastic/elastic-agent/issues/2959
- [ ] https://github.com/elastic/elastic-agent/issues/3057
- [ ] https://github.com/elastic/elastic-agent/issues/3062

ChrsMark:
Added https://github.com/elastic/elastic-agent/issues/3158 after discussed at https://github.com/elastic/elastic-agent/issues/2959 with @gsantoro.

cmacknz:
We are starting to look at https://github.com/elastic/elastic-agent/issues/2460 where the root problem is that the agent configurations generated by autodiscovery are gigantic when it is configured to monitor all the pods on a node or all the nodes in a cluster.

We haven't decided on a solution yet, so the exact impact isn't known. I don't expect it to affect any of the work here, but wanted to note it here for visibility anyway since it may affect how agent autodiscovery generates configurations.

ChrsMark:
@gizas sth also related to this feature: https://github.com/elastic/elastic-agent/issues/2200

PR #5643, opened 2024-10-01
[Kubernetes] Add kustomize template for hints auto discover 

MichaelKatsoulis:
## What does this PR do?

This PR 
1. adds a Kustomize template to enable hints based auto discovery by default in standalone elastic agent.
It adds a init container block in elastic agent daemonset.
For all the rest resources needed for the deployment of elastic agent it relies on the resources from the default kustomize templates.
2. Updates the volume mount of the init container and adds a shared volume between init and main container for storing the input templates. This allows for removal of root privileges in init container which can be a security vulnerability. 

## Why is it important?

To enable users of the Observability Kubernetes quickstart  to make use of hints auto discovery without the need of configuring and redeploying their manifests.

## Checklist

Follow instructions of https://github.com/elastic/elastic-agent/tree/main/deploy/kubernetes/elastic-agent-kustomize/default#remote-usage-of-kustomize-templates

Use https://github.com/MichaelKatsoulis/elastic-agent/deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone\?ref\=kustomize_templates_hints_autodiscover  for the kustomization file link.

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/3760


deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+        spec:
+          initContainers:
+            - name: k8s-templates-downloader
+              image: docker.elastic.co/beats/elastic-agent:9.0.0
+              command: ['bash']
+              args:
+                - -c
+                - >-
+                  mkdir -p /usr/share/elastic-agent/state/inputs.d &&
+                  curl -sL https://github.com/elastic/elastic-agent/archive/9.0.tar.gz | tar xz -C /usr/share/elastic-agent/state/inputs.d --strip=5 "elastic-agent-9.0/deploy/kubernetes/elastic-agent-standalone/templates.d"
```
pkoutsovasilis, line 27: does this url https://github.com/elastic/elastic-agent/archive/9.0.tar.gz exist?

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+              securityContext:
+                runAsUser: 0
```
pkoutsovasilis, line 29: I remember back in the day that the hints approach was using an emptydir volume that the init container would download and move the templates there and then the agent container would mount it on the appropriate path. This approach allowed the init container to run as non root. Any reason this approach isn't still relevant?

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+              volumeMounts:
+                - name: elastic-agent-state
+                  mountPath: /usr/share/elastic-agent/state
```
pkoutsovasilis, line 32: Could you extend the integration k8s tests by adding a new one with hints enabled? it would be nice to  check that the agent and all components are starting and reporting healthy with this feature enabled

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: It is the same approach, only that now we use the already mounted `/usr/share/elastic-agent/state/` path now and we dont create another folder
This changed since this [PR](https://github.com/elastic/elastic-agent/pull/4625)

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: No it does not exist, you are right. This is something that we always do,  we can not escape.

The caclulcation of the version is made based on the makefile https://github.com/elastic/elastic-agent/pull/5643/files/d1051cc2b11833d31df535b98d074ad76a3eea83#diff-18ea46196e4e61745238c819aae128261e8c81b1025bc9e455092d3e9f22817bR123

So we can merge this PR once the image is ready , or someone first to produce the manifest, change manually and then reapply

I dont have any preference

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: Which tests are you referring to Panos?

pkoutsovasilis: I had in mind these tests https://github.com/elastic/elastic-agent/blob/main/testing/integration/kubernetes_agent_standalone_test.go#L63 🙂

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
pkoutsovasilis: I am wondering how difficult would be to include these templates in the agent image (or have a separate wolfi-based flavor that includes them in the first place?!). The whole `deploy/kubernetes/elastic-agent-standalone/templates.d` directory is 960KB in total which imo is negligible compared to the image size of agent. That said, it is not my intention to postpone this PR, so if this is how it is being done so far, I don't see any harm of doing it for one more time. But let's discuss how we could get away from this in the future

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
pkoutsovasilis: what I have in mind is the following; First running as root with all capabilities an init container seems kinda excessive, with emptydir this wasn't necessary. Second, if the pod gets restarted, now that you write the templates directly in the state of the agent which a host mountpath that is already populated by the previous pod execution, it means that the init container is prone to face permission errors, as elastic-agent container could have changed the permissions of its state dir (agent does that after [this PR](https://github.com/elastic/elastic-agent/pull/4925)).

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: We have a CI (see the command here https://github.com/elastic/elastic-agent/blob/e06e78666dc774b60ec0a2d398ec320a32feb601/.buildkite/scripts/steps/sync-k8s.sh#L26) which actually produces those templates under `deploy/kubernetes/elastic-agent-standalone/templates.d`.

Initially the idea was that we dont want to bundle the autodiscovery template process update with the actual beats release cycle. But considering the fact that we dont update them so often we can recosnider it.

It makes sense overal what you say

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: Yes it is a scenario. Of course this approach with the init container was a general solution. We document [here](https://www.elastic.co/guide/en/fleet/current/hints-annotations-autodiscovery.html#_configure_hints_autodiscovery) that this can be done manually too. 

I guess what you propose above to already copy them in the image will solve our problems

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: Thank you !! I wanted to double check as we dont actually own them. 

But I created an issue here: https://github.com/elastic/elastic-agent/issues/5660. We can discuss internally for prioritisation

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
gizas: https://github.com/elastic/elastic-agent/issues/5661 for tracking this

(FYI we have a META issue for hints [here](https://github.com/elastic/elastic-agent/issues/3063))

gizas:
@swiatekm no the main standalone manifest should not be affected

The users should run a command like the following and make an installation of agent+k8s integration+system integration+kube-state-metrics up and running:

```bash
kubectl kustomize https://github.com/MichaelKatsoulis/elastic-agent/deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone\?ref\=kustomize_templates_hints_autodiscover | sed -e "s/JUFQSV9LRVkl/UTWamNtcGFkdz09/g" -e "s/%ES_HOST%/https:\/\/e56ee37c53644df6940a8ae217ab5960.us-central1.gcp.cloud.es.io:443/g" -e "s/%CA_TRUSTED%/ca_trusted_fingerprint/g" -e "s/%ONBOARDING_ID%/12345/g" | kubectl apply -f```

Please read some more info [here](https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-kustomize/default/README.md)

File: deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml (hunk 3 above)
Comment (pkoutsovasilis) [reply deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml:32]: AFAICT the k8s ownership is shared, and the agent repo now has k8s integration tests. So ideally anything that has to do with different k8s deployments should be covered by a test but sure discuss https://github.com/elastic/elastic-agent/issues/5660 and try to couple this integration test with this PR

**Comment by gizas** [comment]
Tested successfully the PR with a current Elastic Cloud setup.


Prereq: Go to Stack Management > API Keys and create a new key

Important use the **beats** format key

Screenshot 2024-10-02 at 4 55 46 PM


Create the encoded value by running:

```bash
echo -n oB1lTZIBdpMiLQUgtFM4:Z1bs7_ufTuOj5O8pg0Sxmg | base64
b0IxbFRaSUJkcE1pTFFVZ3RGTTQ6WjFiczdfdWZUdU9qNU84cGcwU3htZw==
```

Then run the following:

```bash
kubectl kustomize https://github.com/MichaelKatsoulis/elastic-agent/deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone\?ref\=kustomize_templates_hints_autodiscover | sed -e "s/JUFQSV9LRVkl/b0IxbFRaSUJkcE1pTFFVZ3RGTTQ6WjFiczdfdWZUdU9qNU84cGcwU3htZw==/g" -e "s/%ES_HOST%/https:\/\/<CloudURL>:443/g" -e "/{CA_TRUSTED}/c\ "  > k2.yaml
```

Replace the image referneces:
9.0 > 8.15
9.0.0 > 8.15.0

Then run `kubectl apply -f k2.yaml`

Metrics and logs collection is successfull. I will run some more tests with autodiscovery templates and annotations in pods. 
But main fucntionality is ok

gizas:
Continuation of testing, the autodiscovery works as expected: 

Once the pod gets annotated with:

```yaml
annotations:
        co.elastic.hints/enabled: 'true'
        co.elastic.hints/package: "nginx"
```

The datastream changes and no duplication happens:

Screenshot 2024-10-03 at 3 50 11 PM

I can see two issues with current template:
- We need to comment out / remove the dataset: kubernetes.container_logs that is produced 
- We need to uncomment following lines 
```
providers.kubernetes:
      hints.enabled: true
      hints.default_container_logs: true
```

deploy/kubernetes/elastic-agent-kustomize/ksm-hints/elastic-agent-standalone/kustomization.yaml
MichaelKatsoulis: I think I got what we can do here. 

In the initContainer we can set the 
```
volumeMounts:
        - mountPath: /etc/elastic-agent/inputs.d
          name: external-inputs
```
where
```
- name: external-inputs
        emptyDir: {}
```
and untar the templates in there. It will be started with default permissions (not root)

The main container will just have
```
- name: external-inputs
          mountPath: /usr/share/elastic-agent/state/inputs.d
```

I think that covers all scenarios. WDYT?

pkoutsovasilis: @MichaelKatsoulis I am still wondering why this needs to write in a hostpath and thus require "root"; AFAICT the init container everytime that the pod restarts will re-download the templates and because of that this path can be easily considered ephemeral. I think what you are taking for granted here is that the user will allow all the capabilities to the root user but this is not always the case, especially for an init container. So still an emptydir that gets mounted at `/usr/share/elastic-agent/state/inputs.d` seems more ideal to me and it doesn't require your init container to run as root. Also bear in mind that there is an active push to run elastic-agent as rootless and not rootful. Now my in-the-future suggestion, is to add these templates in the agent image (or create a variant of it at least) as their size is really small. Feel free to ask more question, more than happy to reply 🙂

pkoutsovasilis: > I think that covers all scenarios. WDYT?
this is what I had in mind @MichaelKatsoulis indeed! If you could could test that everything works fine with the emptyDir approach I think it would be better 🙂 (please for both rootful and rootless agent invocations)

MichaelKatsoulis: Running elastic-agent without root permissions is trickier as agent tries to create data folder in a host path. Maybe it needs some specific psp. Do you know if there is such an example? I can just set the state volume to an empty dir instead. 
It works in both scenarios. Rootful and rootless agent. I will update the configuration in this PR

MichaelKatsoulis:
@gizas  and @pkoutsovasilis  could you please review this PR again after latest changes ?

changelog/fragments/1727790202-kustomize-templates-with-hints.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Provide a Kustomize template which enables hints based autodiscovery by default when deploying standalone elastic-agent into a Kubernetes cluster.
```
gizas, line 14: nit: Updates the mount path of autodiscovery init container to /etc/elastic-agent/inputs.d in order to eliminate root access needs

To add sth like that in the reasoning as well

gizas:
LGTM Mike! Great work

gizas (approved):

pkoutsovasilis (approved):
LGTM! For the future I strongly recommend to extent the agent k8s integration tests to include testing these templates (tracked by this issue https://github.com/elastic/elastic-agent/issues/5660)

swiatekm (approved):

Issue #5631, opened 2024-09-30
[Flaky Test]: TestLogIngestionFleetManaged – Condition never satisfied


blakerouse:
### Failing test case

TestLogIngestionFleetManaged

### Error message

Condition never satisfied

### Build

https://buildkite.com/elastic/elastic-agent-extended-testing/builds/3139

### OS

Linux

### Stacktrace and notes

```
=== RUN   TestLogIngestionFleetManaged/Monitoring_logs_are_shipped
    logs_ingestion_test.go:138: Making sure metricbeat logs are populated
    logs_ingestion_test.go:268: 
        	Error Trace:	C:/Users/windows/agent/testing/integration/logs_ingestion_test.go:268
        	            				C:/Users/windows/agent/testing/integration/logs_ingestion_test.go:139
        	            				C:/Users/windows/agent/testing/integration/logs_ingestion_test.go:104
        	Error:      	Condition never satisfied
        	Test:       	TestLogIngestionFleetManaged/Monitoring_logs_are_shipped
--- FAIL: TestLogIngestionFleetManaged/Monitoring_logs_are_shipped (180.02s)
```

cmacknz:
```yaml
            input-beat/metrics-monitoring-metrics-monitoring-beats:
                message: 'Error fetching data for metricset beat.stats: error making http request: Get "http://npipe/stats": open \\.\pipe\PGwsYWcynGUYZEjD872Gs-npqbv-30jS.sock: The system cannot find the file specified.'
                payload:
                    streams:
                        metrics-monitoring-filebeat:
                            error: 'Error fetching data for metricset beat.stats: error making http request: Get "http://npipe/stats": open \\.\pipe\PGwsYWcynGUYZEjD872Gs-npqbv-30jS.sock: The system cannot find the file specified.'
                            status: DEGRADED
                        metrics-monitoring-metricbeat:
                            error: ""
                            status: HEALTHY
```

```json
[...]

PR #5648, opened 2024-10-01
[testing/integration]: fix TestLogIngestionFleetManaged

VihasMakwana:
Looking at diagnostics of https://github.com/elastic/elastic-agent/issues/5631 failure, it was clear that the `monitoring` units were not healthy. This is a known issue and it takes some time for them to be healthy.

This PR aims to fix that by making sure the components are healthy before proceeding with the test case.

- Closes https://github.com/elastic/elastic-agent/issues/5631


testing/integration/logs_ingestion_test.go
```diff
@@ -127,6 +131,37 @@ func startMockES(t *testing.T) string {
...
+				}
+			}
+			t.Logf("component state: %s", comp.Message)
+			if comp.State != int(cproto.State_HEALTHY) {
+				compDebugName = comp.Name
+				allHealthy = false
+			}
+		}
+		return allHealthy
+	}, 3*time.Minute, 15*time.Second, "install never became healthy: components did not return a healthy state: %s", compDebugName)
```
swiatekm, line 162: Is there a reason not to check more frequently than 15 seconds? I'd rather the test succeed faster if it succeeds,

VihasMakwana: No reason tbh. Let me drop it to a couple of seconds.

VihasMakwana:
succeeded 2/2 times.

VihasMakwana:
TestLogIngestionFleetManaged  succeeded 3/3

cmacknz:
I think we should also decrease the metrics interval used in the test to 1s, this will effectively be speeding up and allowing for retries if we hit this failure on the first attempt to collect metrics.

https://github.com/elastic/elastic-agent/issues/5631#issuecomment-2387047468

testing/integration/logs_ingestion_test.go
```diff
@@ -127,6 +131,37 @@ func startMockES(t *testing.T) string {
...
+					allHealthy = false
+				}
+			}
+			t.Logf("component state: %s", comp.Message)
+			if comp.State != int(cproto.State_HEALTHY) {
+				compDebugName = comp.Name
+				allHealthy = false
+			}
+		}
+		return allHealthy
```
pchila, line 161: Nitpicking here: can we log the full status of the agent instead of having multiple separate logs per invocation? Something like:

```suggestion
		t.Logf("Received agent status:\n%+v\n", status) // this can be re-marshaled to JSON if we prefer that notation
		for _, comp := range status.Components {
			// make sure the components include the expected integrations
			for _, v := range comp.Units {
				if v.State != int(cproto.State_HEALTHY) {
					allHealthy = false
				}
			}
			if comp.State != int(cproto.State_HEALTHY) {
				compDebugName = comp.Name
				allHealthy = false
			}
		}
		return allHealthy
```

VihasMakwana:
> I think we should also decrease the metrics interval used in the test to 1s, this will effectively be speeding up and allowing for retries if we hit this failure on the first attempt to collect metrics.
> 
> [#5631 (comment)](https://github.com/elastic/elastic-agent/issues/5631#issuecomment-2387047468)

This particular test creates policy in fleet and the schema doesn't really have any `metrics_period` field.

VihasMakwana:
> I think we should also decrease the metrics interval used in the test to 1s, this will effectively be speeding up and allowing for retries if we hit this failure on the first attempt to collect metrics.
> 
> [#5631 (comment)](https://github.com/elastic/elastic-agent/issues/5631#issuecomment-2387047468)

This was the first thing I wanted to do, as we've done the same with `Monitoring*` integration test cases by overriding the `elastic-agent.yml`

blakerouse (approved):
Looks good.

cmacknz:
> EDIT2: We can override elastic-agent.yml, but doing so would contradict the purpose of this fleet-managed test.

This is fine, if you want to properly do it through Fleet you can use the policy overrides API.   See https://support.elastic.dev/knowledge/view/06b69893 for an example.

cmacknz:
(I don't think it matters for the purpose of what we are trying to test here whether you change the initial local agent configuration or use the overrides API to do this).

Checking for health at startup makes sure we identify the right error, but it roughly equivalent to just waiting longer for us to get healthy. If we increase the metrics interval the whole test should execute faster.

swiatekm (approved):

pchila (approved):

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/dolmen-go/contextio v0.0.0-20200217195037-68fc5150bcd5
 	github.com/elastic/elastic-agent-autodiscover v0.8.2
 	github.com/elastic/elastic-agent-client/v7 v7.16.0
-	github.com/elastic/elastic-agent-libs v0.10.1
+	github.com/elastic/elastic-agent-libs v0.12.1-0.20241004133534-196dace11604
```
VihasMakwana, line 17: I'll create a minor release once the CI is green on this commit.

AndersonQ:
> CI is red. BLOCKED on #5542

hey, my PR might take some time to get merged, why is it blocking you? Is it the update of `elastic-agent-libs`?

PR #5653, opened 2024-10-02
feat: drop hashstructure dependency

kruskall:
## What does this PR do?

we are only using the library to hash a struct.
marshal the struct by manually calling the hasher.

## Why is it important?

hashstructure has been archived and is no longer maintained

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/pkgtypes.go
+++ b/dev-tools/mage/pkgtypes.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"io/fs"
 	"log"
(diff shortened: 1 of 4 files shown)


dev-tools/mage/pkgtypes.go
```diff
@@ -500,12 +501,15 @@ func copyInstallScript(spec PackageSpec, script string, local *string) error {
 }
 
 func (s PackageSpec) hash() string {
-	h, err := hashstructure.Hash(s, nil)
+	out, err := yaml.Marshal(s)
```
swiatekm, line 504: How big are these on average? Relative to `hashstructure`, we'd be using more memory here, as we need to hold the full yaml encoding of the spec in memory.

blakerouse: This is only used for packaging the elastic-agent and is not something even used by a running elastic-agent.  Even with the increase in memory here it would be very minimal, most package specifications are not that large.

blakerouse (approved):
Always nice to see less dependencies.

swiatekm (approved):

michel-laterman (approved):

PR #5663, opened 2024-10-02
Update Makefile to remove reviewers

jlind23:
This team is not required as a reviewer anymore as it is already part of Kibana codeowners.

This was causing the synck8s step to fail on main.


gizas:
@jlind23 we use the deploy/kubernetes/makefile to update our manifests.
Are you sure that we are going to be notified when it is changed?

gizas (approved):

PR #5666, opened 2024-10-02
Add filebeat receiver to OTEL components

blakerouse:
## What does this PR do?

Adds the `fbreceiver` (aka. filebeat receiver) to OTel receivers that the Elastic Agent OTel supports.

## Why is it important?

Adds the ability to run filebeat inputs as an OTel receiver.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation
- ~~[ ] I have made corresponding change to the default configuration files
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (will be tested in integration test)
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Disruptive User Impact

None.

## How to test this PR locally

Run `elastic-agent otel` with an OTel configuration (otel.yml) like the following:

```
receivers:
  filebeatreceiver:
    filebeat:
      inputs:
        - type: benchmark
          enabled: true
          message: "test"
		  eps: 5
    output:
      otelconsumer:
    logging:
      level: debug
      selectors:
        - '*'
    processors:
      - add_fields:
          target: project
          fields:
            name: myproject
processors:
  batch:
exporters:
  debug:
    verbosity: detailed
    sampling_initial: 100
service:
  pipelines:
    logs:
      receivers: [filebeatreceiver]
      processors: [batch]
      exporters: [debug]
```

PR #5677, opened 2024-10-02
Shorten allowlist for `replace`able dependencies

ycombinator:
## What does this PR do?

This PR cleans up the list for allowed dependencies in the `go.mod` `replace` directive.  It removes dependencies from the allowlist that are no longer being `replace`d in `go.mod`.

## Why is it important?

To help minimize the number of dependencies we `replace` with forks.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -112,13 +112,8 @@ linters-settings:
     # Forbid any `replace` directives that are intended temporarily only during
     # development. The modules listed below are intended to be replaced permanently.
     replace-allow-list:
-      - github.com/Microsoft/go-winio
       - github.com/Shopify/sarama
-      - github.com/dop251/goja
-      - github.com/dop251/goja_nodejs
-      - github.com/fsnotify/fsnotify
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -112,13 +112,8 @@ linters-settings:
     # Forbid any `replace` directives that are intended temporarily only during
     # development. The modules listed below are intended to be replaced permanently.
     replace-allow-list:
-      - github.com/Microsoft/go-winio
       - github.com/Shopify/sarama
-      - github.com/dop251/goja
-      - github.com/dop251/goja_nodejs
-      - github.com/fsnotify/fsnotify
```
blakerouse, line 119: I believe these will still be needed for enabling the filebeat receiver - https://github.com/elastic/elastic-agent/pull/5666

ycombinator: Thanks.  Added them back in bbefe5e9bf.

leehinman (approved):

PR #5748, opened 2024-10-09
Don't run any agent service when running `otel` subcommand

blakerouse:
## What does this PR do?

Previously the control protocol was started when the subcommand `otel` was ran. This should not be done as the subcommand `otel` should operate it in a pure OTel mode.

This is a cleanup that comes out of working on Hybrid mode.

## Why is it important?

Cleanup to make the code cleaner and less messy as the control protocol should not really be running when running the otel subcommand.

## Checklist

## Disruptive User Impact

Can no longer call `elastic-agent status` when a `elastic-agent otel` is running.

Run `elastic-agent otel`, then try to run `elastic-agent status` and see that it doesn't connect.

### DIFF
--- a/testing/integration/otel_test.go
+++ b/testing/integration/otel_test.go
@@ -20,10 +20,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
-	"github.com/elastic/elastic-agent/pkg/control/v2/client"
 	aTesting "github.com/elastic/elastic-agent/pkg/testing"
-	atesting "github.com/elastic/elastic-agent/pkg/testing"
-	integrationtest "github.com/elastic/elastic-agent/pkg/testing"
 	"github.com/elastic/elastic-agent/pkg/testing/define"
 	"github.com/elastic/elastic-agent/pkg/testing/tools/estools"
 	"github.com/elastic/elastic-agent/pkg/testing/tools/testcontext"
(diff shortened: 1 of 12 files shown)


swiatekm (approved):
LGTM, but it'd be good for @andrzej-stencel to have a look at this as well, imo.

testing/integration/otel_test.go
```diff
@@ -463,22 +426,6 @@ func TestOtelAPMIngestion(t *testing.T) {
 	)
 	require.NoError(t, err, "APM not initialized")
 
-	// wait for otel collector to start
```
cmacknz, line 466: Is it worth replacing this with a check of the upstream health endpoint?

cmacknz (approved):
I like this as a simplification now that the direction we are going with elastic-agent is clearer than it was when the `otel` command was first implemented.

Once https://github.com/elastic/elastic-agent/pull/5767 is implemented we can probably remove the `elastic-agent otel` command entirely.

testing/integration/otel_test.go
blakerouse: I don't think its worth it at the moment. The work I am doing in #5767 will expose status information over the `elastic-agent status` command again.

blakerouse:
> I like this as a simplification now that the direction we are going with elastic-agent is clearer than it was when the `otel` command was first implemented.
> 
> Once #5767 is implemented we can probably remove the `elastic-agent otel` command entirely.

I think it depends on if we want a true 100% otel implementation or something that has some of the Elastic Agent code running. I think the longer term goal is to make the `run` become `otel` sub-command, but at the moment we are not that far along.

ycombinator:
Force merging to get around the failed SonarQube check.

PR #5753, opened 2024-10-10
Add more details about Docker image variants

jlind23:
Add more details about the different Docker image variants and their specificities.


swiatekm (approved):

pchila (approved):

jlind23:
Force merging - Sonar is required but will never run and there was no code change.

PR #5767, opened 2024-10-11
Add support for running EDOT inside of running Elastic Agent

### LINKED ISSUE #5796: [EDOT Collector] Allow Elastic Agent to run in hybrid mode
**Describe the enhancement:**

Elastic Agent should be able to run in hybrid mode — that is, it should be able to accept a policy containing OTel Collector configuration, start up an embedded OTel Collector, and pass the OTel Collector configuration from the policy to the embedded OTel Collector.

To be able to run "classic" Elastic Agent configuration and OTel Collector configuration concurrently, to allow a more seamless transition from the former to the latter.

Elastic Agent should be able to:
* [ ] accept a policy containing OTel Collector configuration
* [ ] start up an embedded OTel Collector, and 
* [ ] pass the OTel Collector configuration from the policy to the embedded OTel Collector

blakerouse:
## What does this PR do?

Adds the ability to run the EDOT along side the running Elastic Agent.

This connects the EDOT into the coordinator of the Elastic Agent. At any point if any of these top-level keys (`receivers`, `processors`, `exporters`, `extensions`, `service`) exist in the configuration or policy for the elastic-agent the EDOT is started. If all of those keys are removed from the configuration or policy then the EDOT is automatically stopped. If any configuration change occurs the updated configuration is passed along to the EDOT to handle.

## Why is it important?

This allows EDOT configuration to exist inside of the configuration or policy and allow it to work as expected.

## Checklist

## Disruptive User Impact

This is an addition and doesn't affect the way the current Elastic Agent runs at all.

Place OTel configuration into the `elastic-agent.yml`:

```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
processors:
  batch:

exporters:
  otlp:
    endpoint: otelcol:4317

extensions:
  health_check:
  pprof:

service:
  extensions: [health_check, pprof]
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
    metrics:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
```

Run `elastic-agent run -e`.

## Related issues

Closes #5796


blakerouse:
This is ready for a review. Don't be scare too much by the size of this change, most of that comes from NOTICE.txt, the generate control protocol and the rename of `Unpack` to `UnpackTo`.

To answer the incoming question of why rename `Unpack` to `UnpackTo` is that the `Unpacker` interface used by go-ucfg has the function named `Unpack` so that collides. This required the rename for the configuration to work correctly.

internal/pkg/agent/cmd/inspect.go
```diff
@@ -389,7 +389,7 @@ func getMonitoringFn(ctx context.Context, cfg map[string]interface{}) (component
 	}
 
 	agentCfg := configuration.DefaultConfiguration()
-	if err := config.Unpack(agentCfg); err != nil {
+	if err := config.UnpackTo(agentCfg); err != nil {
```
michalpristas, line 392: this change makes PR much less readable, generates a lot of unrelated file mods

internal/pkg/agent/cmd/otel.go
```diff
@@ -2,8 +2,6 @@
 // or more contributor license agreements. Licensed under the Elastic License 2.0;
 // you may not use this file except in compliance with the Elastic License 2.0.
 
-//go:build !windows
```
michalpristas, line 5: have you verified service starts properly without any timing issues? it was a bit difficult to verify last time

internal/pkg/config/config.go
```diff
@@ -39,14 +51,20 @@ var DefaultOptions = []interface{}{
 	ucfg.VarExp,
 	VarSkipKeys("inputs"),
 	ucfg.IgnoreCommas,
+	OTelKeys("receivers", "processors", "exporters", "extensions", "service"),
```
michalpristas, line 54: missing `connectors` 
refer to unmarshaller https://github.com/open-telemetry/opentelemetry-collector/blob/8e522ad950de6326a0841d7e1bef808bbc0d3537/otelcol/unmarshaler.go#L18

internal/pkg/config/config.go
```diff
@@ -123,12 +154,23 @@ func MustNewConfigFrom(from interface{}) *Config {
...
+	if c.Agent == nil {
+		c.Agent = ucfg.New()
+	}
+	return c.Agent.Merge(val)
 }
 
-// Unpack unpacks a struct to Config.
-func (c *Config) Unpack(to interface{}, opts ...interface{}) error {
+// UnpackTo unpacks this config into to with the given options.
+func (c *Config) UnpackTo(to interface{}, opts ...interface{}) error {
```
michalpristas, line 173: as said PR would be o simpler without this one, if we could extract this to refactoring PR we could get reviews for this one faster

internal/pkg/otel/agentprovider/provider.go
```diff
@@ -0,0 +1,112 @@
...
+	p.replaceCanceller(nil)
+	return nil
+}
+
+// URI returns the URI to be used for this provider.
+func (p *Provider) URI() string {
+	return p.uri
+}
+
+func (p *Provider) replaceCanceller(replace context.CancelFunc) {
```
michalpristas, line 104: this looks wild

internal/pkg/otel/run.go
```diff
@@ -49,34 +44,69 @@ func Run(ctx context.Context, stop chan bool, configFiles []string) error {
...
-			URIs: configPaths,
-			ProviderFactories: []confmap.ProviderFactory{
-				fileprovider.NewFactory(),
-				envprovider.NewFactory(),
-				yamlprovider.NewFactory(),
-				httpprovider.NewFactory(),
-				httpsprovider.NewFactory(),
-			},
-			ConverterFactories: []confmap.ConverterFactory{
-				expandconverter.NewFactory(),
```
michalpristas, line 69: 👍 for dropping, we should pay more attention to upstream changelog

michalpristas (approved):
I think this looks good, the mentioned change makes this PR a bit uglier that it needs to be.

I imagined our first take on hybrid approach pretty much like this, i have no major things to point out besides missing keyword and verifying windows works. 

Tested this locally, seems fine. Status reported nicely on component level

internal/pkg/agent/cmd/otel.go
michalpristas: linking issue https://github.com/elastic/elastic-agent/issues/4976

blakerouse: I am not seeing those issues in this PR. This PR is running on Windows with the integration testing framework and all of those are passing just fine. Those would not be passing if this was not working correctly.

internal/pkg/config/config.go
blakerouse: Yeah, I probably should have split it out. Being that you already gave a +1 I would prefer to leave it as it. Splitting that out is going to take more work, then just getting this merged.

blakerouse:
@michalpristas For some reason GitHub is not saying your approval counted, could you possibly approve again to see if that lets me have an approval.

pierrehilbert:
You need more than @michalpristas's approval.
@MichaelKatsoulis / @gizas we will need your approval here please.

blakerouse:
@pierrehilbert That is fine, but for some reason even @michalpristas is not showing as approved to GitHub.

cmacknz:
I tried this locally and the first thing I noticed is that the format of the JSON output by `elastic-agent status` is different.

In isolation in the context of this PR, this isn't a problem. It will be a problem for Beats receivers though, as unless Fleet knows how to parse the new format we are going to lose the input health reporting features.

So we either have to:
1. Have Fleet understand collector status reporting and have it integrate with the UI similar to what we have today.
2. Have the agent translate the collector health into the existing `components` based health reporting structure.

I think I'd prefer 2 since the point of the Beats receivers project is to make the agent transparently use the OTel collector. Additional this has a better chance of getting the state reporting that is happening directly in Filebeat inputs working, see https://github.com/elastic/beats/pull/39209 for where this was added.

I'll for this into a separate tracking issue unless we can resolve it in this PR.

`sudo elastic-development-agent status --output=json`
<details>

```json
{
    "info": {
        "id": "8c884936-b6d7-4ff7-af1f-c15da3752384",
        "version": "9.0.0",
        "commit": "e787a19a6a6cb0f570cdf4f16c8839e20b0feb23",
        "build_time": "2024-11-18 19:12:02 +0000 UTC",
        "snapshot": true,
        "pid": 22023,
        "unprivileged": false,
        "is_managed": false
    },
    "state": 2,
    "message": "Running",
    "components": [],
    "FleetState": 6,
    "FleetMessage": "Not enrolled into Fleet",
    "collector": {
        "status": 2,
        "timestamp": "2024-11-18T14:52:56.483851-05:00",
        "components": {
            "extensions": {
                "status": 2,
                "timestamp": "2024-11-18T14:52:56.482935-05:00",
                "components": {
                    "extension:health_check": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.482525-05:00"
                    },
                    "extension:pprof": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.482935-05:00"
                    }
                }
            },
            "pipeline:logs": {
                "status": 2,
                "timestamp": "2024-11-18T14:52:56.483683-05:00",
                "components": {
                    "exporter:otlp": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483204-05:00"
                    },
                    "processor:batch": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483252-05:00"
                    },
                    "receiver:otlp": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483683-05:00"
                    }
                }
            },
            "pipeline:metrics": {
                "status": 2,
                "timestamp": "2024-11-18T14:52:56.483851-05:00",
                "components": {
                    "exporter:otlp": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483369-05:00"
                    },
                    "processor:batch": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483702-05:00"
                    },
                    "receiver:otlp": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483851-05:00"
                    }
                }
            },
            "pipeline:traces": {
                "status": 2,
                "timestamp": "2024-11-18T14:52:56.483829-05:00",
                "components": {
                    "exporter:otlp": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.4838-05:00"
                    },
                    "processor:batch": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483812-05:00"
                    },
                    "receiver:otlp": {
                        "status": 2,
                        "timestamp": "2024-11-18T14:52:56.483829-05:00"
                    }
                }
            }
        }
    }
}
```

</details>

I enrolled an agent running a collector configuration into Fleet, which was not interesting because the collector configuration was immediately replaced with the standard agent configuration from Fleet.

cmacknz:
Looking in the diagnostics I see the otel configuration is separated out, and pre-config.yaml and computed-config.yaml are both mostly empty:

```yaml
cat diag/computed-config.yaml
host:
    id: 48DA13D6-B83B-5C71-A4F3-494E674F9F37
path:
    config: /Library/Elastic/Agent-Development
    data: /Library/Elastic/Agent-Development/data
    home: /Library/Elastic/Agent-Development/data/elastic-agent-9.0.0-SNAPSHOT-e787a1
    logs: /Library/Elastic/Agent-Development
runtime:
    arch: arm64
    native_arch: arm64
    os: darwin
    osinfo:
        family: darwin
        major: 14
        minor: 7
        patch: 1
        type: macos
        version: 14.7.1
```

What would we expect computed-config to look like if we used variables in the otel configuration? Is that even supported here? If it is, how would we debug it?

I also see `state.yaml` is missing all of the collector status. This we need to address somehow for sure, the collector runtime state needs to be in diagnostics:

```
cat diag/state.yaml
components: []
fleet_message: Not enrolled into Fleet
fleet_state: 6
log_level: info
message: Running
state: 2
```

internal/pkg/agent/application/monitoring/liveness.go
```diff
@@ -79,6 +83,11 @@ func livenessHandler(coord CoordinatorState) func(http.ResponseWriter, *http.Req
 				unhealthyComponent = true
 			}
 		}
+		if state.OTelStatus != nil {
```
cmacknz, line 86: I've seen this `OtelStatus != nil` twice and I worry it might be something we can forget to check easily. Would be simpler to have a no-op status reporter when there isn't an otel collector component to eliminate the need for this check everywhere?

cmacknz: Also, perhaps worth an IsFailed or IsDegraded otelhelper to simplify the mapping of otel errors to agent states.

internal/pkg/otel/agentprovider/provider.go
cmacknz: +1, why is this necessary? If there is no way to get rid of it, please put the explanation in the code as comments.

internal/pkg/otel/manager/extension.go
```diff
@@ -0,0 +1,185 @@
...
+	}()
+	select {
+	case as.eventCh <- &evtPair{source: source, event: event}:
+	case <-as.ctx.Done():
+	}
+}
+
+func (as *AgentStatusExtension) eventLoop(ctx context.Context) {
+	// prevent aggregate statuses from flapping between StatusStarting and StatusOK
+	// as components are started individually by the service.
```
cmacknz, line 131: Could we just not do this filtering? Was it just annoying or causing an actual problem?

internal/pkg/otel/manager/force_extension_converter.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+type forceExtension struct {
+	name string
+}
+
+func (fe *forceExtension) Convert(ctx context.Context, conf *confmap.Conf) error {
```
cmacknz, line 18: What is this doing and why is it necessary? Please add the explanation as a comment in the code directly.

internal/pkg/otel/manager/manager.go
```diff
@@ -0,0 +1,225 @@
...
+func (m *OTelManager) Run(ctx context.Context) error {
+	var err error
+	var cancel context.CancelFunc
+	var provider *agentprovider.Provider
+
+	// signal that the run loop is ended to unblock any incoming update calls
+	defer close(m.doneChan)
+
+	runErrCh := make(chan error)
+	for {
```
cmacknz, line 65: Should there be debug logging in this loop to cover the case that we accidentally deadlock or block writing to `errCh`?

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -280,6 +287,36 @@ func (f *FleetGateway) convertToCheckinComponents(components []runtime.Component
 		checkinComponents = append(checkinComponents, checkinComponent)
 	}
 
+	// OTel status is placed as a component for each top-level component in OTel
+	// and each subcomponent is a unit.
+	if otelStatus != nil {
```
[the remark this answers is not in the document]
blakerouse: I don't understand how this would be broken in Fleet. This is sending it exactly in the same format that Fleet understand today, so this just works transparently without any changes to Fleet. Which is what we discussed and what was implemented here.

internal/pkg/agent/application/monitoring/liveness.go
blakerouse: Please elaborate, I do not understand the ask here.

internal/pkg/otel/manager/extension.go
blakerouse: It causes the reporting to not really be in sync and looks very strange. This is also following the same pattern of the healthcheckv2extension.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckv2extension/extension.go#L168

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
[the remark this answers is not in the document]
cmacknz: Fleet won't reject it, I am worried about how Fleet maps inputs back to integrations, since Fleet is integration focused and agent is only inputs.

We would need to test an integration implemented with Beats receivers and make sure the status reporting still works the same way.

internal/pkg/agent/application/monitoring/liveness.go
cmacknz: Nevermind, I was misreading this. My head was still a few files earlier.

internal/pkg/otel/manager/manager.go
blakerouse: When the collector is start and stopped it already logs, that is obvious in the logs. The configuration change is also logged, so in all cases those events are covered. I don't think its worth adding more redundant logging.

internal/pkg/otel/manager/extension.go
cmacknz: Ah, can you link to that code directly?

blakerouse:
> I tried this locally and the first thing I noticed is that the format of the JSON output by `elastic-agent status` is different.

It is only different as at the top-level there is a new field called `collector` that provides all the status information about the collector.

> 
> In isolation in the context of this PR, this isn't a problem. It will be a problem for Beats receivers though, as unless Fleet knows how to parse the new format we are going to lose the input health reporting features.

As you see later in your code review this is not sent to Fleet this way. The Fleet information is not in any way decoded from JSON marshalling.

> 
> So we either have to:
> 
> 1. Have Fleet understand collector status reporting and have it integrate with the UI similar to what we have today.

Nope, this is already done in this PR.

> 2. Have the agent translate the collector health into the existing `components` based health reporting structure.

We could do that, but I don't think that is a good idea long term. The reason I made a new `collector` top-level key with the OTel status information, is that in the long run we will drop the other format and switch to the new format.

> 
> I think I'd prefer 2 since the point of the Beats receivers project is to make the agent transparently use the OTel collector. Additional this has a better chance of getting the state reporting that is happening directly in Filebeat inputs working, see [elastic/beats#39209](https://github.com/elastic/beats/pull/39209) for where this was added.

They should 100% only be reporting status using otel. The goal is to one day only run the collector, they should add status reporting through the collector.

> 
> I'll for this into a separate tracking issue unless we can resolve it in this PR.
> 
> `sudo elastic-development-agent status --output=json`

I don't see an issue. Adding new fields is not considered breaking. API's add new fields all the time and that is not considered breaking.

blakerouse:
> Looking in the diagnostics I see the otel configuration is separated out, and pre-config.yaml and computed-config.yaml are both mostly empty:
> 
> ```yaml
> cat diag/computed-config.yaml
> host:
>     id: 48DA13D6-B83B-5C71-A4F3-494E674F9F37
> path:
>     config: /Library/Elastic/Agent-Development
>     data: /Library/Elastic/Agent-Development/data
>     home: /Library/Elastic/Agent-Development/data/elastic-agent-9.0.0-SNAPSHOT-e787a1
>     logs: /Library/Elastic/Agent-Development
> runtime:
>     arch: arm64
>     native_arch: arm64
>     os: darwin
>     osinfo:
>         family: darwin
>         major: 14
>         minor: 7
>         patch: 1
>         type: macos
>         version: 14.7.1
> ```
> 
> What would we expect computed-config to look like if we used variables in the otel configuration? Is that even supported here? If it is, how would we debug it?

Variables are not supported in otel configuration. That would be extending otel outside of the Elastic Agent which I don't think is a good idea.  You will notice in this PR that otel configuration is extracted as early as possible. This is done to ensure that the Elastic Agent in no way affects the otel configuration.

> 
> I also see `state.yaml` is missing all of the collector status. This we need to address somehow for sure, the collector runtime state needs to be in diagnostics:
> 
> ```
> cat diag/state.yaml
> components: []
> fleet_message: Not enrolled into Fleet
> fleet_state: 6
> log_level: info
> message: Running
> state: 2
> ```

Good catch. I will add that now.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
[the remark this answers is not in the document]
blakerouse: I didn't know there was a link between those two. I was just under the impression that the UI just rendered the content of the `components` field that is shipped to Fleet. If it does some referencing between the two (which I don't think it does), then it would be better to just let Fleet understand the `component` field.

Being this is experimental that is something that can easily change in the future.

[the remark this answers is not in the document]
cmacknz: I think it looks into the meta information in the inputs, possibly it only needs the input ID to be what was in the agent policy for the component and it could figure it out.

I agree this isn't blocking and is something we could check easily with a test.

I don't love the `elastic-agent status output=json` format diverging from what gets sent to Fleet, I have used the fact that it is in the checkin format a few times for debugging, but I won't block the PR on it.

internal/pkg/otel/manager/extension.go
blakerouse: Done.

internal/pkg/otel/manager/force_extension_converter.go
blakerouse: Done.

cmacknz:
> They should 100% only be reporting status using otel. The goal is to one day only run the collector, they should add status reporting through the collector.

Long term yes, my actual goal is to discover the point where we need to commit engineering time from the UI team to rewriting or updating the input and component health implementation to account for the collector health status.

I don't think just Beats receivers is that point, it is probably the point at which there are integrations with OTel native configurations. That said, we could plan UI work to account for the status changes for Beats receivers, but it will block us shipping them.

cmacknz:
> If you mean more like transparent taking the otel status and translating it to a component status to report to Fleet, then that is something different and could be done.

Yes the interface to the Fleet and the eventual content of the .fleet-agents datastream there is all I care about. It's our API contract with the UI team.

Everything happening inside the agent as far as getting status from the collector LGTM.

michalpristas (approved):
reapproving, probably force push did the magic

MichaelKatsoulis (approved):

alexsapran (approved):
Reviewing `.github/workflows/golangci-lint.yml` LGTM

PR #5773, opened 2024-10-14
Exclude the coverage files from the annotate step

alexsapran:
## What does this PR do?

We noticed that the JUnit annotate step needed to take more time. This PR addresses that issue by not requiring it to parse the coverage files.

The generated coverage XML file was meant to be used in Jenkins to show the coverage report stats. The generated XML files were large, resulting in a problem with BK's JUnit annotate plugin. 

Given that we don't operate any Jenkins instance, this PR removes the code that generated the XML coverage reports.
Now, coverage is reported only to Sonar.

## Why is it important?

Reduce the time of CI step

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -158,7 +158,7 @@ steps:
   - label: ":junit: Junit annotate"
     plugins:
       - junit-annotate#v2.5.0:
-          artifacts: "*/TEST-*.xml"
+          artifacts: "**TEST-*.xml"
           always-annotate: true
     agents:
       provider: "gcp"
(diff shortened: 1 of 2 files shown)


rowlandgeoff (approved):
From 90min to 39sec on the annotate step, I'd say that fixes it. :) Thanks!

I had to look at the individual commit history to understand why this was the fix. You might want to update the PR description to keep the history easier to understand.

alexsapran:
The problem was that the annotation plugin was reading the coverage files, which are converted to XML format with the help of the `gocover-cobertura` go package. 

That was needed back when we combined Jenkins with some Jenkins plugins. We could keep creating that XML file, but we don't use it anymore, thus I removed it. 
For now I am waiting for the full CI to finish. Also the PR is not yet ready, as I think I found another minor issue.

rowlandgeoff (approved):
LGTM 👍

jlind23 (approved):

PR #4832, opened 2024-05-31
[DO NOT MERGE] - PoC for orchestrating remote tests from inside the package using terraform

pchila:
## What does this PR do?

This PR contains a simple PoC to show what a TestMain for remote tests looks like when provisioning is offloaded to a third-party tool like Terraform.

**Note: the code is really rough and in no way, shape or form close to production quality. This has been tested only on Linux even if it should work with minimal changes on MacOS. Windows has not been harmed during the writing of this code but it's also not been taken into account for compatibility. This PR is there just to explore what is possible without getting distracted by something like support for all OSes**

Goals:
- Test if it's possible to orchestrate remote tests (both provisioning and running) without using `mage` or other tools
- Support and pass along any go test flags the user may specify when launching the tests to be run on a remote machine
- Limit the usage of environment variables as much as possible
- Test is terraform can simplify the test code and abstract away provisioning complexity
- Check compatibility with `define.*` constructs already used for elastic-agent integration tests
- Provision an environment where a `go test` can be readily executed without setup getting too much in the way of the tests to be run

Non-Goals:
- Implement a drop-in replacement for the current integration framework
- Support multiple OSes both on local or remote side

What has been implemented:
- Provisioning of a single linux ubuntu 2204 VM using elastic ubuntu image
- Spin up a small ESS deployment
- Sync local code with remote host "automagically" using terraform and rsync
- Collect the information about the provisioning and expose it to a TestMain in an agnostic way
- Support running local and remote tests with a simple `go test` command that supports custom command line arguments
- Add small tests to be run remotely to verify that we can run them and connect to an ESS deployment

What is still missing (and would probably be the next steps)
- A bunch of command line options and terraform variables
- Properly abstract `Provisioner`, `Runner` via interfaces so that we can swap different implementations (the ones needed by the current code are at least `TerraformProvisioner` and `SSHRunner`)
- Implement a pluggable system for various provisioners each with their own flags (handled within the specific provisioner type) so that we can implement Docker , `elastic-package` (and others...) provisioning
- Spin up multiple VMs in terraform based on arguments passed as variables (by the user or by the TestMain itself)
- Implement some smart batching and parallel test execution for the remote tests on different hosts.
- Debug support by launching tests with `dlv` on the remote while port forwarding the debug port on the local machine
- I am sure I am forgetting a lot of other things but the ones above are what springs to mind...

I would also like to extend special thanks to @alexsapran and @pierrehilbert which helped me get through some of the terraform stuff

## Checklist

[...]

PR #5777, opened 2024-10-14
Devmachine mage target

pazone:
## What does this PR do?

`devmachine:create` mage target. Creates a dev machine in GCE from the specified image. 
The VM will be automatically deleted in 6 hours.

I tried to make `devmachine:create` self-documented and it prompts instructions how to connect the VM, etc.

[Full Docs](https://github.com/elastic/ingest-dev/pull/4248) 

## Why is it important?

Allows to create a dev machine equivalent to a CI runner. 
## Checklist

## Related issues

-

## Questions to ask yourself

- ...


pazone:
> I only had a quick first look but I was wondering why cannot we use terraform invoked from the mage target? Something similar to https://github.com/elastic/elastic-agent/pull/4832/files#diff-d6bb41b0e2013dce4632f1d552f264ccfdecb0fcae4f09c28e7212e3af2b5bc8R102-R134

It's great that we have an alternative approach using Terraform. Currently, the Mage target only runs a single VM, which is automatically deleted after the TTL. This solution is intentionally kept simple, as gcloud allows us to connect to the VM without needing to specify an IP address or SSH key. As a result, this approach involves less imperative code compared to using Mage to invoke Terraform.

However, if we need to manage multiple types of resources, we'll need to switch to using Terraform.

.tool-versions
```diff
@@ -0,0 +1,3 @@
+mage 1.14.0
```
alexsapran, line 1: Since we are adding this `.tool-versions` we might need to also update https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-golang.yml

Alternativly, have the addition of the `.tool-versions` + workflow on a dedicated PR

pazone: Do you mean we have to create a dedicated PR that with `.tool-version` update automation?

alexsapran: There is an automation that bumps the golang version (mainly).

We need to ensure that it will also include this file when this happens.

dev-tools/devmachine/devmachine.go
```diff
@@ -0,0 +1,116 @@
...
+		log.Fatalf("Failed to create instance: %v", err)
+	}
+
+	log.Printf("Instance creation initiated: %s\n", op.Name)
+	log.Printf("Instance name: %s\n", instanceName)
+
+	retriesLimit := 24
+	retryCount := 0
+	for {
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
pazone: Nice catch!

.tool-versions
pazone: Created a follow-up [issue](https://github.com/elastic/ingest-dev/issues/4302)

dev-tools/devmachine/devmachine.go
```diff
@@ -0,0 +1,115 @@
...
+	projectID := "elastic-platform-ingest"
+	log.Println("Authenticating with GCP...")
+	computeService, err := compute.NewService(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create compute service: %v. \nRun `gcloud auth login` to make it work", err)
+	}
+
+	imageProject := "elastic-images-prod"
+	machineType := fmt.Sprintf("zones/%s/machineTypes/n1-standard-4", zone)
+	sourceImage := fmt.Sprintf("projects/%s/global/images/%s", imageProject, imageName)
```
swiatekm, line 28: All of these strings should be constants imo.

magefile.go
```diff
@@ -844,6 +848,23 @@ func (Cloud) Push() error {
 	return nil
 }
 
+// Creates a new devmachine that will be auto-deleted in 6 hours.
+// Example: mage devmachine:create "pavel-dev-machine" "family/platform-ingest-elastic-agent-ubuntu-2204" "us-central1-a"
+func (Devmachine) Create(instanceName string, imageName string, zone string) error {
```
swiatekm, line 853: Do we have a way of enumerating what images are available?

dev-tools/devmachine/devmachine.go
swiatekm, line 28: All of these strings should be constants imo.

dev-tools/devmachine/devmachine.go
```diff
@@ -0,0 +1,115 @@
...
+					{
+						Name: "External NAT",
+						Type: "ONE_TO_ONE_NAT",
+					},
+				},
+				Network: "global/networks/default", // Use default network
+			},
+		},
+		Tags: &compute.Tags{
+			Items: []string{"http-server", "https-server"},
```
swiatekm, line 54: Why this tag specifically? Shouldn't we have some tags here identifying this instance as a dev machine as well?

swiatekm (commented):
The mage code looks fine to me, left some minor comments.

magefile.go
pazone: Good question. I can implement it, but there are 2 ways:

- a simple gcloud command where you need to specify your email
- I can wrap it in a mage target like `mage devmachine:list`

What's would be better? 🤔

swiatekm: Honestly, I think the best user experience would be to provide a reasonable default here (the ubuntu 22.04 one sounds fine to me), and if someone wants a different image, they can run the gcloud command. We should include that command in the help text though.

pazone: We have a lot of different prebuilt images. `family/platform-ingest-elastic-agent-ubuntu-2204` stands for Ubuntu 22.04. 
Unfortunately, mage doesn't support default parameter values. But it can be achieved with env vars. For example, 
```
MACHINE_IMAGE="family/platform-ingest-elastic-agent-ubuntu-2204" ZONE="us-central1-a" mage devmachine:create "machine_name"
```
WDYT?

pazone: The full list of available platform-ingest images is [here](https://docs.elastic.dev/ingest-dev-docs/engineer-productivity/custom-ci-images#available-vm-images). The image is specified in the [pipeline step definition](https://github.com/elastic/elastic-agent/blob/main/.buildkite/bk.integration.pipeline.yml#L52). It is better explained in the [Full doc](https://github.com/elastic/ingest-dev/pull/4248/files#diff-a95e8cfcce5a66414ad9917e12cadcbff18d2feff2a92da35a76367900cb2c8fR13) mentioned in the PR description

dev-tools/devmachine/devmachine.go
pazone: I searched through the [code entries](https://github.com/search?q=zones%2F%25s%2FmachineTypes+language%3AGo&type=code&l=Go) and these strings are usually inlined. That's reasonable because it helps better understand the final string

magefile.go
swiatekm:
```
MACHINE_IMAGE="family/platform-ingest-elastic-agent-ubuntu-2204" ZONE="us-central1-a" mage devmachine:create "machine_name"
```
Looks good to me.

dev-tools/devmachine/devmachine.go
pazone: Good catch! Replaced tags with required labels

dev-tools/devmachine/devmachine.go
pazone: Moved the templates to constants

swiatekm (approved):

magefile.go
pazone: Now we specify `ZONE` and `MACHINE_IMAGES` as optional env vars. For example, `ZONE="europe-west1-b" mage -v devmachine:create "pavel-dev-machine"` will run a machine from `family/platform-ingest-elastic-agent-ubuntu-2204` image in `europe-west1-b` zone.

pazone: Thanks, Mikołaj!

PR #5790, opened 2024-10-15
Add all Docker images to K8s integration tests

rdner:
So, we validate all produced images.

Test runtime:
* before this PR – [~57 minutes](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/3760#019290b0-a3b7-4600-b3a3-673bc796005b)
* after this PR – [~1h 37 minutes](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/3756#0192908e-d36e-4588-ba19-262921ad51ca)

Let me know if the runtime increase is sensible.

### DIFF
--- a/testing/integration/kubernetes_agent_standalone_test.go
+++ b/testing/integration/kubernetes_agent_standalone_test.go
@@ -66,9 +66,12 @@ func TestKubernetesAgentStandaloneKustomize(t *testing.T) {
 		Local: false,
 		Sudo:  false,
 		OS: []define.OS{
-			// only test the basic and the wolfi container with otel
+			// test all produced images
 			{Type: define.Kubernetes, DockerVariant: "basic"},
 			{Type: define.Kubernetes, DockerVariant: "wolfi"},
+			{Type: define.Kubernetes, DockerVariant: "ubi"},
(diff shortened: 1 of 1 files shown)


jlind23:
Looks like the overall time is still lower then the integration tests execution time so we should be good on that end.
Starting 9.0 ubi and basic will be the same so this would mean minus one variant that has to be removed from main.

ycombinator:
> Looks like the overall time is still lower then the integration tests execution time so we should be good on that end. Starting 9.0 ubi and basic will be the same so this would mean minus one variant that has to be removed from main.
> 
> @ycombinator what do you think?

I tend to agree but I'm also wondering: how much more effort would it be to run `TestKubernetesAgentStandaloneKustomize` for each of the docker variants concurrently?

rdner:
@jlind23 @ycombinator looks like we all agree that the overall runtime of the tests won't change after merging this (because the agent integration tests are running for even longer).

I think changing the test framework to run these images concurrently is out of the scope of this task, here is the issue to follow up https://github.com/elastic/elastic-agent/issues/5801

jlind23 (approved):

PR #5844, opened 2024-10-23
Add notification for when the packaging pipeline fails

alexsapran:
## What does this PR do?

This PR adds a notification for the packaging pipeline when it fails.
The notification is for all the branches this pipeline runs and only on failures.
The notification is sent to the team on their Slack channel.

## Why is it important?

Yes, this is related to https://github.com/elastic/ingest-dev/issues/3625


rowlandgeoff (approved):
Thanks!

PR #5854, opened 2024-10-25
[helm]: implement support for defining image pull secrets

pkoutsovasilis:
## What does this PR do?

This PR that adds support for defining `pullImageSecrets` support in the chart

## Why is it important?

Users mights need to deploy an agent image from a private repository

## Checklist

## Disruptive User Impact

N/A

```
```

### DIFF
--- a/deploy/helm/elastic-agent/README.md
+++ b/deploy/helm/elastic-agent/README.md
@@ -133,6 +133,7 @@ The chart built-in [kubernetes integration](https://docs.elastic.co/integrations
 |-----|------|---------|-------------|
 | agent.version | string | `"9.0.0"` | elastic-agent version |
 | agent.image | object | `{"pullPolicy":"IfNotPresent","repository":"docker.elastic.co/beats/elastic-agent","tag":"9.0.0-SNAPSHOT"}` | image configuration |
+| agent.imagePullSecrets | list | `[]` | image pull secrets |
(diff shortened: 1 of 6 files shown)


blakerouse (approved):
Looks good.

swiatekm (approved):
I think it's more idiomatic to attach these to a ServiceAccount instead of the Pod, but in this case it doesn't matter too much.

andrzej-stencel (approved):

pkoutsovasilis:
> I think it's more idiomatic to attach these to a ServiceAccount instead of the Pod, but in this case it doesn't matter too much.

I don't have a preference of one way (ServiceAccount attachment) over the other (Pod attachment), but since there is this [draft PR](https://github.com/elastic/elastic-agent/pull/5593) (planning on working on that soon) that will allow the users to complete override service accounts I chose the latter to deterministically always apply the `imagePullSecrets` and not allow it to be overidden. 


> Also, can you actually set this in one of the examples?

I can mention it but having it actually in the example cli commands I don't think so as I don't have a guaranteed elastic-agent image under a private registry and I would like to keep the examples as simple as a straight-forward copy-paste. Would that work or you had something else in mind?

swiatekm:
> I can mention it but having it actually in the example cli commands I don't think so as I don't have a guaranteed elastic-agent image under a private registry and I would like to keep the examples as simple as a straight-forward copy-paste. Would that work or you had something else in mind?

That's fine. I mostly just want a sanity check that the template works as expected, but for a change as simple as this one it's not that important.

PR #5855, opened 2024-10-25
[helm]: implement system integration as chart built-in

pkoutsovasilis:
## What does this PR do?

This PR introduces support for the system integration 

## Why is it important?

system integration is usually used in the context of k8s, thus having it as a built-in integration improves the user configuration experience

## Checklist

```
```

### DIFF
--- a/deploy/helm/elastic-agent/README.md
+++ b/deploy/helm/elastic-agent/README.md
@@ -123,12 +123,21 @@ The chart built-in [kubernetes integration](https://docs.elastic.co/integrations
 | kubernetes.system.metrics.enabled | bool | `true` | enable system metric stream (kubelet) [ref](https://www.elastic.co/docs/current/integrations/kubernetes/kubelet#system) |
 | kubernetes.system.metrics.vars | object | `{}` | system metric stream vars |
 
-### 3 - User Extra Integrations
+### 4 - System integration
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| system.enabled | bool | `false` | enable System integration. |
+| system.output | string | `"default"` | name of the output used in kubernetes integration. Note that this output needs to be defined in [outputs](#1-outputs) |
+| system.namespace | string | `"default"` | output namespace |
+| system.logs.enabled | bool | `true` | enable logs |
+| system.metrics.enabled | bool | `true` | enable metrics |
+
+### 5 - User Extra Integrations
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | extraIntegrations | object | `{}` | extra [user-defined integrations](https://www.elastic.co/guide/en/fleet/current/elastic-agent-input-configuration.html) to be added to the Elastic Agent An example can be found [here](./examples/nginx-custom-integration/README.md) |
 
-### 3 - Elastic-Agent Configuration
+### 6 - Elastic-Agent Configuration
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | agent.version | string | `"9.0.0"` | elastic-agent version |
@@ -138,7 +147,7 @@ The chart built-in [kubernetes integration](https://docs.elastic.co/integrations
 | agent.unprivileged | bool | `false` | enable unprivileged mode |
 | agent.presets | map[string]{} | `{ "perNode" : {...}, "clusterWide": {...}, "ksmSharded": {...} }` | Map of deployment presets for the Elastic Agent. The key of the map is the name of the preset. See more for the presets required by the built-in Kubernetes integration [here](./values.yaml) |
 
-### 3.1 - Elastic-Agent Managed Configuration
+### 6.1 - Elastic-Agent Managed Configuration
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | agent.fleet.enabled | bool | `false` | enable elastic-agent managed |
--- a/deploy/helm/elastic-agent/README.md
+++ b/deploy/helm/elastic-agent/README.md
@@ -127,9 +127,12 @@ The chart built-in [kubernetes integration](https://docs.elastic.co/integrations
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | system.enabled | bool | `false` | enable System integration. |
-| system.output | string | `"default"` | name of the output used in kubernetes integration. Note that this output needs to be defined in [outputs](#1-outputs) |
+| system.output | string | `"default"` | name of the output used in System integration. Note that this output needs to be defined in [outputs](#1-outputs) |
 | system.namespace | string | `"default"` | output namespace |
-| system.logs.enabled | bool | `true` | enable logs |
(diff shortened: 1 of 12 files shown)


deploy/helm/elastic-agent/README.md
```diff
@@ -123,12 +123,21 @@ The chart built-in [kubernetes integration](https://docs.elastic.co/integrations
 | kubernetes.system.metrics.enabled | bool | `true` | enable system metric stream (kubelet) [ref](https://www.elastic.co/docs/current/integrations/kubernetes/kubelet#system) |
 | kubernetes.system.metrics.vars | object | `{}` | system metric stream vars |
 
-### 3 - User Extra Integrations
+### 4 - System integration
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| system.enabled | bool | `false` | enable System integration. |
```
blakerouse, line 129: Why is the default `false`? I feel like most will want this on, it brings immediate value to the Observability section in Kibana.

deploy/helm/elastic-agent/templates/integrations/_system/_system_logs.tpl
```diff
@@ -0,0 +1,45 @@
...
+  use_output: {{ $.Values.system.output }}
+  data_stream:
+    namespace: {{ $.Values.system.namespace }}
+  streams:
+    - data_stream:
+        dataset: system.auth
+        type: logs
+      paths:
+        - /var/log/auth.log*
+        - /var/log/secure*
```
blakerouse, line 22: Will users want to adjust these paths?

deploy/helm/elastic-agent/templates/integrations/_system/_system_logs.tpl
```diff
@@ -0,0 +1,45 @@
...
+        match: after
+      processors:
+        - add_locale: null
+      ignore_older: 72h
+    - data_stream:
+        dataset: system.syslog
+        type: logs
+      paths:
+        - /var/log/messages*
+        - /var/log/syslog*
```
blakerouse, line 36: Will users want to adjust these paths?

deploy/helm/elastic-agent/values.yaml
```diff
@@ -287,55 +287,74 @@ kubernetes:
       # -- system metric stream vars
       # @section -- 2 - Kubernetes integration
       vars: {}
+system:
+  # -- enable System integration.
+  # @section -- 4 - System integration
+  enabled: false
```
blakerouse, line 293: I do feel like default here is a better choice.

blakerouse (approved):
I don't see any issue, just really a comment on it being default and extra options. Not really blockers, they can be adjusted later if we see that it should be default or those settings need options.

swiatekm (approved):
Looks good to me. I don't have an informed opinion on whether this integration should be enabled by default. But if it's not, can we add an example where we enable it?

deploy/helm/elastic-agent/README.md
pkoutsovasilis: More than happy to have it by default as `true`, it also makes sense to me. @nimarezainia what do you think on that? 🙂

deploy/helm/elastic-agent/templates/integrations/_system/_system_logs.tpl
pkoutsovasilis: I pushed an update that now allows the user to adjust these paths, and thank you for the inspiration, I added an example showcasing that 🙂 Feel free to unresolve if you spot anything else

deploy/helm/elastic-agent/templates/integrations/_system/_system_logs.tpl
pkoutsovasilis: same as above 🙂

deploy/helm/elastic-agent/values.yaml
pkoutsovasilis: Resolving in favour of https://github.com/elastic/elastic-agent/pull/5855#discussion_r1816742053 🙂

pkoutsovasilis:
> Looks good to me. I don't have an informed opinion on whether this integration should be enabled by default. But if it's not, can we add an example where we enable it?

thanks for the review @swiatekm and the example proposal, I took inspiration from one of @blakerouse 's comments and added one but also did some other changes, would you mind having another look? 🙂

deploy/helm/elastic-agent/examples/system-custom-auth-paths/agent-system-values.yaml
```diff
@@ -0,0 +1,9 @@
+system:
```
swiatekm, line 1: We should have rendered examples for this, right? I thought the mage target would error in this case.

pkoutsovasilis: hmm, I thought I added it but this never happened, but the CI should fail yes

pkoutsovasilis: oh I suspect that this is because the file was never versioned and the `check-ci` apparently fails on changes of versioned files?!

swiatekm: Shouldn't it fail anyway if it can't find the file? I didn't know the examples mage target was aware of versioning.

Anyway, this isn't a blocker for this PR, but we should make sure the check actually catches issues.

deploy/helm/elastic-agent/templates/integrations/_system/_system_metrics.tpl
```diff
@@ -0,0 +1,103 @@
...
+      cpu.metrics:
+        - percentages
+        - normalized_percentages
+      metricsets:
+        - cpu
+    - data_stream:
+        dataset: system.diskio
+        type: metrics
+      period: 10s
+      diskio.include_devices: null
```
swiatekm, line 30: Are these nulls intentional? I would expect this to either be an empty list or for the key to not exist at all.

swiatekm (approved):
I really like the added configurability. Had a small question about the configuration template, but it shouldn't block merging this PR. Nice work. :muscle:

deploy/helm/elastic-agent/examples/system-custom-auth-paths/agent-system-values.yaml
pkoutsovasilis: the current CI checks flow is the following:

* render helm examples
* check for git diffs; the assumption here was that it failed also on unversioned files but evidently that's not the case

100% agreed this needs to be extended

deploy/helm/elastic-agent/templates/integrations/_system/_system_metrics.tpl
pkoutsovasilis: this crossed my mind as well, but then I wondered why this config was rendered like that from Kibana; thus, I figured not to mess with it 😄 but sure this can be further optimised in the future

PR #5864, opened 2024-10-26
Feature/4890 detect fail early upgrade

### LINKED ISSUE #4890: Detect and fail early if user attempts to upgrade Agent using the CLI in unsupported scenarios
- Relates https://github.com/elastic/fleet-server/pull/3991

**Version: 8.14.0**
**Operating System: Ubuntu 24.04 LTS**
**Platform: arm64**

While working on [this issue](https://github.com/elastic/elastic-agent/issues/4705) comparing root and unprivileged elastic agents, I encountered an error when upgrading the agent.

**Steps to Reproduce:**
1. Deploy ESS v8.14.0
2. Create agent policy with system integration
3. Install fleet managed agent v8.13.4 with or without the unprivileged flag `sudo ./elastic-agent install --url=<fleet url> --enrollment-token=<token>` 
4. Run `sudo elastic-agent upgrade 8.14.0`
5. Upgrade seems to work; however, when checking the status we get the following error
```
┌─ fleet
│  └─ status: (FAILED) status code: 500, fleet-server returned an error: BadRequest, message: failed to update upgrade_details: upgrade_details no action for id "" found
├─ elastic-agent
└─ upgrade_details
   ├─ target_version: 8.14.0
   ├─ state: UPG_WATCHING
   └─ metadata
```
6. Uninstall and unenroll the agent and install v8.13.4 again
7. Upgrade through the fleet ui. This should work. 

This bug occurs for both privileged and unprivileged agents. 

## Definition of Done

Synthesized from https://github.com/elastic/elastic-agent/issues/4890#issuecomment-2433337815:

- [x] If an unprivileged user attempts to upgrade a Fleet-managed unprivileged Agent from the CLI, Agent should refuse to upgrade and output a message explaining why and additionally mention that the upgrade cannot be performed because the command was not executed with root/Administrator permissions (same as the `isAdmin` check in `cmd/install.go`).
- [x] If a privileged user attempts to upgrade a Fleet-managed Agent (privileged or unprivileged) from the CLI (`elastic-agent upgrade ...`), Agent should refuse to upgrade and output a message explaining why. The message should NOT mention anything about a `--force` flag (explained below).
- [x] However, if a user additionally provides a `--force` flag in the previous scenario, Agent should present a warning message and proceed with the upgrade anyway.  This `--force` flag should be hidden; it should NOT show up in the output of `elastic-agent help upgrade`.

kaanyalti:
- Enhancement
## What does this PR do?
This PR restricts using the cli to upgrade fleet managed elastic agents. With this implementation, the only way to upgrade a fleet managed agent using the cli is only if the agent is running in privileged mode and the user provides a `force` flag.

## Checklist

## Disruptive User Impact
If users are upgrading fleet-managed agents using the cli for some reason, they should be aware that the upgrade command will stop working for agents running in unprivileged mode. For any fleet managed agents running in privileged mode, users will need to go through an interactive step and use the `--force` flag to be able to upgrade an agent.

### Setup
- Create an ESS deployment 
- Go to fleet and click add agent
- Create a new policy if there isn't already one
- Build the agent locally by running `DEV=true SNAPSHOT=true mage build`
- Uncompress the tar
- If there isn't an agent on your machine already just execute the provided command on fleet ui
- If there is already an agent on your machine add `--develop` flag to your installation command

#### Unprivileged
- Add `--unprivileged` flag to your installation command
- Ensure that the agent is enrolled
- Run `sudo -u elastic-agent-user elastic-development-agent upgrade <VERSION>`. The command should not execute.
- Make sure that `--force` flag does not change the behavior

#### Privileged
- Install the agent and make sure that it is enrolled
- Run `sudo elastic-development-agent upgrade <VERSION>`. The command should not execute
- Run the same command with `--force` flag `sudo elastcit-development-agent upgrade <VERSION> --force`.
- Confirm the prompted message. Agent upgrade should start
## Related issues
- Closes elastic/elastic-agent#4890 
- Closes elastic/ingest-dev#4414


changelog/fragments/1729971565-detect-fail-early-fleet-managed-cli-upgrade.yaml
```diff
@@ -0,0 +1,31 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: detect-fail-early-fleet-managed-cli-upgrade
```
andrzej-stencel, line 14: This should be a human readable text, right?

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,20 +57,85 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
 	cmd.Flags().String(flagPGPBytes, "", "PGP to use for package verification")
 	cmd.Flags().String(flagPGPBytesURI, "", "Path to a web location containing PGP to use for package verification")
 	cmd.Flags().String(flagPGPBytesPath, "", "Path to a file containing PGP to use for package verification")
+	cmd.Flags().BoolP(flagForce, "f", false, "Advanced option to force an upgrade on a fleet managed agent")
```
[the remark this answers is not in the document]
kaanyalti: I can remove it, I just thought it wouldn't hurt to provide a shorthand way to execute it, please let me know if you want it removed.

changelog/fragments/1729971565-detect-fail-early-fleet-managed-cli-upgrade.yaml
kaanyalti: Yes, updating it right away

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,20 +57,85 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
...
+	force, err := cmd.Flags().GetBool(flagForce)
+	if err != nil {
+		return false, fmt.Errorf("failed to retrieve command flag information while trying to upgrade the agent: %w", err)
+	}
+
+	if !force {
+		return false, fmt.Errorf("upgrading fleet managed agents is not supported")
+	}
+
+	cf, err := cFunc("Upgrading fleet managed agents is not supported. Would you still like to proceed?", false)
```
andrzej-stencel, line 115: @cmacknz Are we sure we want to introduce this confirmation (if I understand correctly, you proposed it [here](https://github.com/elastic/elastic-agent/issues/4890#issuecomment-2414789873))? This will make the command unusable in unattended scenarios, like Ansible mentioned in https://github.com/elastic/elastic-agent/issues/4890#issuecomment-2337891708.

It seems to me that passing the `--force` flag means that the user has already opted in and we should expect at this point that the user knows what they're doing. The additional confirmation sounds like we're asking the user the same thing twice: "Are you sure? But are you sure you're sure?".

On a separate but related note, I'm not fond of the `--force` flag being hidden from docs. I would rather have it not hidden, and have the docs clearly describe that using it is not recommended, what can go wrong and when it can be useful.

internal/pkg/agent/cmd/upgrade.go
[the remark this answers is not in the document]
andrzej-stencel: I would remove it, but having it won't stop me from approving this pull request. It's a non-blocking suggestion. If you'd rather have it, I'm fine with it.

internal/pkg/agent/cmd/upgrade_test.go
```diff
@@ -68,6 +82,245 @@ func TestUpgradeCmd(t *testing.T) {
...
+
+		s := grpc.NewServer()
+		defer s.Stop()
+
+		// Define mock server and agent information
+		upgradeCh := make(chan struct{})
+		mock := &mockServer{upgradeStop: upgradeCh}
+		mockAgentInfo := mockinfo.NewAgent(t)
+		mockAgentInfo.On("IsStandalone").Return(false) // Simulate fleet-managed agent
+		mockAgentInfo.On("Unprivileged").Return(true)  // Simulate unprivileged mode
```
pchila, line 99: Why not setting expectations on the mock using something like:
```suggestion
		mockAgentInfo.EXPECT().IsStandalone().Return(false) // Simulate fleet-managed agent
		mockAgentInfo.EXPECT().Unprivileged().Return(true)  // Simulate unprivileged mode
```

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,20 +57,85 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
...
+	if err != nil {
+		return false, fmt.Errorf("failed while confirming action: %w", err)
+	}
+
+	if !cf {
+		return false, fmt.Errorf("upgrade not confirmed")
```
pchila, line 121: Please define a package variable with this error, using `errors.New`.
Here we are not formatting a new error, wrapping an inner error or some variable values in the error message so the call to `fmt.Errorf` is superfluous.

Bonus: we can assert the error in the test using `errors.Is` or its equivalent assertion function

internal/pkg/agent/cmd/upgrade.go
cmacknz: +1 to getting rid of the confirmation for force.

The intent for Fleet managed agents, which already have an automated bulk upgrade mechanism, is for this to be a "use in case of emergency" option where the Fleet upgrades are broken. 

The official docs already state that this is not recommended for Fleet managed agents. https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-upgrade-command

Today using `elastic-agent upgrade` with a Fleet managed agent does not work at all because of a separate bug, so nobody is even able to use this.

The thing this is protecting us against is a situation where an agent has gone offline in Fleet because of a bug, and the way out is to upgrade the agent (or there is a bug in the Fleet upgrade process itself). Upgrading the agent should not be impossible in this circumstance.

The intent is to discourage use of `elastic-agent upgrade` outside of these circumstances, but if somebody is root, and they can discover this `--force` option exists, and they really want to do it, they can. That is fine.

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,20 +58,73 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
 	cmd.Flags().String(flagPGPBytes, "", "PGP to use for package verification")
 	cmd.Flags().String(flagPGPBytesURI, "", "Path to a web location containing PGP to use for package verification")
 	cmd.Flags().String(flagPGPBytesPath, "", "Path to a file containing PGP to use for package verification")
+	cmd.Flags().BoolP(flagForce, "", false, "Advanced option to force an upgrade on a fleet managed agent")
+	err := cmd.Flags().MarkHidden(flagForce)
```
kaanyalti, line 62: @cmacknz Could you please confirm whether we keep this flag hidden or whether to remove the hidden attribute and document it?

cmacknz: Keep it hidden, it should not be the primary way to upgrade agents that are Fleet managed, and it doesn't work today regardless.

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,20 +58,73 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
...
+	if err != nil {
+		return false, fmt.Errorf("failed to retrieve command flag information while trying to upgrade the agent: %w", err)
+	}
+
+	if !force {
+		return false, unsupportedUpgrade
+	}
+
+	return true, nil
+}
```
pchila, line 114: If I am not missing something this function can only return:
- true, nil
- false, <some error>

If that is the case the boolean return value is superfluous and you can just return error and check that for success in the caller; maybe the function could change name to something like `checkUpgradable()` and we can pass a boolean for the force flag instead of the whole cmd,  something like:
```suggestion
func checkUpgradable(force bool, agentInfo info.Agent) error {
	if agentInfo.IsStandalone() {
		return nil
	}

	if agentInfo.Unprivileged() {
		return fmt.Errorf("upgrade command needs to be executed as root for fleet managed agents")
	}

	if !force {
		return unsupportedUpgrade
	}

	return nil
}
```

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,20 +58,73 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
...
+	cmd := input.cmd
+	c := input.c
+	version := input.args[0]
 	sourceURI, _ := cmd.Flags().GetString(flagSourceURI)
 
-	err := c.Connect(context.Background())
+	su, err := shouldUpgrade(cmd, input.agentInfo)
+	if !su {
+		return fmt.Errorf("aborting upgrade: %w", err)
+	}
```
pchila, line 125: The error wasn't even checked here :sweat_smile:, we can change this to 
```suggestion
	force, err := cmd.Flags().GetBool(flagForce)
	if err != nil {
		return fmt.Errorf("failed to retrieve command flag information while trying to upgrade the agent: %w", err)
	}

	err := checkUpgradable(force, input.agentInfo)
	if err != nil {
		return fmt.Errorf("aborting upgrade: %w", err)
	}
```

andrzej-stencel (approved):

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -31,15 +32,19 @@ const (
 	flagPGPBytes       = "pgp"
 	flagPGPBytesPath   = "pgp-path"
 	flagPGPBytesURI    = "pgp-uri"
+	flagForce          = "force"
 )
 
+var unsupportedUpgrade error = errors.New("upgrading fleet managed agents is not supported")
```
cmacknz, line 38:
```suggestion
var unsupportedUpgrade error = errors.New("this agent is Fleet managed and must be upgraded using Fleet")
```

Upgrading fleet managed agents is supported, just from fleet. Also let's avoid the word "unsupported" since there may be cases where we explicitly tell users to do this because of bugs and we don't want to think they are doing something fundamentally risky.

cmacknz:
I tried to manually test that upgrading as a non-root user fails, I did the following:

1. `AGENT_PACKAGE_VERSION=1.0.0 EXTERNAL=true SNAPSHOT=true PACKAGES=targz PLATFORMS=darwin/arm64 mage package` to force the agent version to 1.0.0 since in Fleet if you are 9.0 you can't upgrade again to 9.0 or a lower version.
2. Install a development, unprivileged agent:
```sh
sudo ./elastic-agent install -f --unprivileged --develop
Unprivileged installation mode enabled; this feature is currently in beta.
[=== ] Service Started  [6s] Elastic Agent successfully installed, starting enrollment.
[=== ] Done  [6s]
```
3. Check the version:
```sh
sudo -u elastic-agent-user elastic-development-agent version
Binary: 1.0.0-SNAPSHOT (build: 269031487ee30d3ec60058c1743e01eedd21dd2b at 2024-11-01 18:23:57 +0000 UTC)
Daemon: 1.0.0-SNAPSHOT (build: 269031487ee30d3ec60058c1743e01eedd21dd2b at 2024-11-01 18:23:57 +0000 UTC)
```
4. Upgrade to 8.15.3 and it lets me do it:
```
sudo -u elastic-agent-user elastic-development-agent upgrade 8.15.3
Upgrade triggered to version 8.15.3, Elastic Agent is currently restarting

sudo -u elastic-agent-user elastic-development-agent version
Binary: 8.15.3 (build: 61975895b1409449db21ddca0405e7b71bfc1c46 at 2024-10-08 14:25:20 +0000 UTC)
Daemon: 8.15.3 (build: 61975895b1409449db21ddca0405e7b71bfc1c46 at 2024-10-08 14:25:20 +0000 UTC)
```

I was expecting to be stopped, but maybe this is a bug in how we detect we are root when doing `sudo -u elastic-agent-user` which should run the command as a non-zero UID.

cmacknz:
OK, if I enroll in Fleet it works I see what has happened:

```sh
❯ sudo -u elastic-agent-user elastic-development-agent upgrade 8.15.3
Error: aborting upgrade: upgrade command needs to be executed as root for fleet managed agents
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
```

cmacknz:
We only require being root to use the CLI with --force if enrolled with Fleet, the standalone check comes first.

We could require being root to upgrade an unprivileged agent, it makes sense to me as a system management action, but it would be a breaking change for it today.

It does appear that the escape hatch for upgrades doesn't work on a Mac though for unprivileged agents:

```sh
❯ sudo -u elastic-agent-user elastic-development-agent upgrade 8.15.3 --force
Error: aborting upgrade: upgrade command needs to be executed as root for fleet managed agents
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
~/Downloads/elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64 ·································· 02:58:47 PM
❯ sudo elastic-development-agent upgrade 8.15.3 --force
Error: failed to retrieve agent info while tring to upgrade the agent: could not get agent info from store: fail to read configuration /Library/Elastic/Agent-Development/fleet.enc for the agent: fail to decode bytes: cipher: message authentication failed
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
```

I can't upgrade as root or as `elastic-agent-user`. I can't upgrade at all through the CLI when installed as a unprivileged agent on a Mac and enrolled in Fleet.

kaanyalti:
> I can't upgrade as root or as `elastic-agent-user`. I can't upgrade at all through the CLI when installed as a unprivileged agent on a Mac and enrolled in Fleet.
> 

Looking into this right away

cmacknz:
It may be a bug or limitation in how we store the key for for unprivileged agents on MacOS, it is probably tied to the `elastic-agent-user` and root doesn't inherit any capability in the key store. CC @pchila I recall you did some work for unprivileged agents on MacOS.

If we can't fix this (make both root and elastic-agent-user able to access the key explicitly?), we could relax this restriction on Mac.

@kaanyalti see if you can reproduce this on Linux to try and confirm if this is MacOS specific.

pchila (approved):
LGTM

kaanyalti:
> It may be a bug or limitation in how we store the key for for unprivileged agents on MacOS, it is probably tied to the `elastic-agent-user` and root doesn't inherit any capability in the key store. CC @pchila I recall you did some work for unprivileged agents on MacOS.
> 
> If we can't fix this (make both root and elastic-agent-user able to access the key explicitly?), we could relax this restriction on Mac.
> 
> @kaanyalti see if you can reproduce this on Linux to try and confirm if this is MacOS specific.

I can confirm that this does not happen on linux. 
- I installed agent on linux in unprivileged mode. 
- Upgrading the agent did not work when ran as `elastic-agent-user` even with the force flag. 
- I then tried upgrading as an admin user without the force flag. Upgrade got reject
- Tried upgrading as an admin user with the force flag which worked.

I ran through the same steps on mac, and got the following error output
image

pchila:
> It may be a bug or limitation in how we store the key for for unprivileged agents on MacOS, it is probably tied to the `elastic-agent-user` and root doesn't inherit any capability in the key store. CC @pchila I recall you did some work for unprivileged agents on MacOS.
> 

> If we can't fix this (make both root and elastic-agent-user able to access the key explicitly?), we could relax this restriction on Mac.
> 
> @kaanyalti see if you can reproduce this on Linux to try and confirm if this is MacOS specific.



> > It may be a bug or limitation in how we store the key for for unprivileged agents on MacOS, it is probably tied to the `elastic-agent-user` and root doesn't inherit any capability in the key store. CC @pchila I recall you did some work for unprivileged agents on MacOS.
> > If we can't fix this (make both root and elastic-agent-user able to access the key explicitly?), we could relax this restriction on Mac.
> > @kaanyalti see if you can reproduce this on Linux to try and confirm if this is MacOS specific.
> 
> I can confirm that this does not happen on linux.
> 
>     * I installed agent on linux in unprivileged mode.
> 
>     * Upgrading the agent did not work when ran as `elastic-agent-user` even with the force flag.
> 
>     * I then tried upgrading as an admin user without the force flag. Upgrade got reject
> 
>     * Tried upgrading as an admin user with the force flag which worked.
> 
> 
> I ran through the same steps on mac, and got the following error output image

@kaanyalti @cmacknz 
I imagine you have 2 agents running on your Mac ? If the second agent is running unprivileged but the first is privileged the `unprivileged` check on uninstall/upgrade may fail (since we didn't want to add specific CLI switches to upgrade/uninstall or metadata during install)

What I think is happening:
- we run the `upgrade` command always with root/administrator privileges
- we try to detect if agent was installed as unprivileged on Mac (in uninstall this is done by [checking if the file-based vault exists](https://github.com/elastic/elastic-agent/blob/0a965617920874e740b85aad7b383ea370804401/internal/pkg/agent/install/uninstall.go#L75)) not too sure what is happening here
- elastic-agent gets confused between the 2 installs and tries to decrypt what looks like the main (infosec) agent using the file based key and fails...

kaanyalti:
> > It may be a bug or limitation in how we store the key for for unprivileged agents on MacOS, it is probably tied to the `elastic-agent-user` and root doesn't inherit any capability in the key store. CC @pchila I recall you did some work for unprivileged agents on MacOS.
> 
> > If we can't fix this (make both root and elastic-agent-user able to access the key explicitly?), we could relax this restriction on Mac.
> > @kaanyalti see if you can reproduce this on Linux to try and confirm if this is MacOS specific.
> 
> > > It may be a bug or limitation in how we store the key for for unprivileged agents on MacOS, it is probably tied to the `elastic-agent-user` and root doesn't inherit any capability in the key store. CC @pchila I recall you did some work for unprivileged agents on MacOS.
> > > If we can't fix this (make both root and elastic-agent-user able to access the key explicitly?), we could relax this restriction on Mac.
> > > @kaanyalti see if you can reproduce this on Linux to try and confirm if this is MacOS specific.
> > 
> > 
> > I can confirm that this does not happen on linux.
> > ```
> > * I installed agent on linux in unprivileged mode.
> > 
> > * Upgrading the agent did not work when ran as `elastic-agent-user` even with the force flag.
> > 
> > * I then tried upgrading as an admin user without the force flag. Upgrade got reject
> > 
> > * Tried upgrading as an admin user with the force flag which worked.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I ran through the same steps on mac, and got the following error output image
> 
> @kaanyalti @cmacknz I imagine you have 2 agents running on your Mac ? If the second agent is running unprivileged but the first is privileged the `unprivileged` check on uninstall/upgrade may fail (since we didn't want to add specific CLI switches to upgrade/uninstall or metadata during install)
> 
> What I think is happening:
> 
> * we run the `upgrade` command always with root/administrator privileges
> * we try to detect if agent was installed as unprivileged on Mac (in uninstall this is done by [checking if the file-based vault exists](https://github.com/elastic/elastic-agent/blob/0a965617920874e740b85aad7b383ea370804401/internal/pkg/agent/install/uninstall.go#L75)) not too sure what is happening here for the upgrade...
> * elastic-agent gets confused between the 2 installs and tries to decrypt what looks like the main (infosec) agent using the file based key and fails...
> 
> Edit: if I had to guess I think this check is failing in case of 2 agents installed on the same Mac (1 with keychain secret and one with the file one)
> 
> https://github.com/elastic/elastic-agent/blob/8c43ad333837380f1646c3b007ff6339b1593a94/internal/pkg/agent/application/info/agent_info.go#L71
> 
> Edit no. 2: @kaanyalti if you are testing on your Mac and you have 2 agent installed shouldn't you invoke `sudo elastic-development-agent upgrade ...` ? (I noticed in the screenshot above that the command doesn't look right)

@pchila I did run through the same testing steps with the development agent as well. To test the non-dev agent I first uninstalled the agent through the elastic shield. Just to  be extra sure, I also run the same tests with non-dev agent in a mac vm as well.

pchila:
@kaanyalti ok that excludes interactions between two agents install on the same MacOS  bit I still think we need a check for update similar as what we do in uninstall...

Moreover the elastic-agent key in the keychain is not removed when agent is uninstalled so there's a good chance we are still loading the wrong (privileged install) one

cmacknz:
Following from @pchila's suggestions I see what is causing this:

https://github.com/kaanyalti/elastic-agent/blob/1b109bde54d419c9f92e868e58bbdb66262c7103/internal/pkg/agent/cmd/upgrade.go#L81-L84
```go
	agentInfo, err := info.NewAgentInfoWithLog(cmd.Context(), "error", false)
	if err != nil {
		return fmt.Errorf("failed to retrieve agent info while tring to upgrade the agent: %w", err)
	}
```

This leads us to the code below to read the encrypted store which has a direct check for being root:

https://github.com/elastic/elastic-agent/blob/9d2c3c92069894db0bbea34a7c5ca66f4b632377/internal/pkg/agent/storage/encrypted_disk_store.go#L42-L53

This happens because we need to do `sudo elastic-development-agent upgrade` but the `utils.IsRoot()` call is happening for the upgrade command we are executing in the terminal, which is run as root, but this is a command targeting a different agent process running as the non-root `elastic-agent-user`.

You load the agent info in the upgrade command here to:
1. Determine if the command is running as root, but you can just call `utils.IsRoot()` for that which is all agent info does: https://github.com/elastic/elastic-agent/blob/9d2c3c92069894db0bbea34a7c5ca66f4b632377/internal/pkg/agent/application/info/agent_info.go#L71-L74
2. Determine if the agent is managed by Fleet, which does legitimately require loading the agent info.

So, within the `upgrade` command you can have it figure out it is targeting an unprivileged agent even though it is running at root, or we can have the actual running agent binary tell us if it is managed by Fleet.

The agent commands make gRPC requests to the running agent service using a local protocol defined in https://github.com/elastic/elastic-agent/blob/main/control_v2.proto.

There is conveniently already a `State` RPC call you could execute that contains some of the agent info, if you also had it tell you if it was standalone you wouldn't need to load agent info at all and this problem would completely go away.

https://github.com/elastic/elastic-agent/blob/a3385439b5cfc739e1bd1cf9840c2a87ef08d28f/control_v2.proto#L157-L178

cmacknz:
I have an additional request that is a small increase in scope, can be handled here or in a follow up PR. When we know agent is managed by Fleet, we need to forbid using the `--skip-verify` argument to the `upgrade` command. We do no expose this in Fleet managed upgrades in the UI at all.

For standalone agents, we need to require that you be root to use `--skip-verify`. Unprivileged users should not be able to use `--skip-verify` at all.

cmacknz:
I filed https://github.com/elastic/elastic-agent/issues/5947 to track that, makes no difference to me if it is done in this PR or another one.

internal/pkg/agent/cmd/upgrade_test.go
```diff
@@ -67,6 +82,426 @@ func TestUpgradeCmd(t *testing.T) {
 		close(upgradeCh)
 		// this makes sure all client assertions are done
 		<-clientCh
+		c.Disconnect()
+	})
+
+	t.Run("fail if fleet managed and unprivileged", func(t *testing.T) {
+		var wg sync.WaitGroup
+		// Set up mock TCP server for gRPC connection
+		tcpServer, err := net.Listen("tcp", "127.0.0.1:")
```
cmacknz, line 91: Only investigate this if you want to, but I think these tests can be simplified because you don't actually need a server to test gRPC server implementations. You can just invoke the `mock.Upgrade()` RPC implementation directly without needing a network transport to connect you to it.

You would probably have to make a test client type that wraps the server type, so that calling `client.Upgrade()` just immediately calls `mock.Upgrade()`.

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,24 +62,116 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
...
+	checkManaged := func() error {
+		if cond.skipVerify {
+			return skipVeriftNotAllowedError
+		}
+
+		if !cond.isRoot {
+			return nonRootExecutionError
+		}
+
+		if !cond.force {
```
cmacknz, line 129: This needs to come first so that users see that it is unsupported regardless of what they do unless they give `--force`.

This is what I see doing this locally:

```sh
❯ sudo -u elastic-agent-user elastic-development-agent upgrade 8.15.3
Error: aborting upgrade: upgrade command needs to be executed as root for fleet managed agents
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
```

cmacknz: Users should be told it is unsupported, until they supply `--force`, then they can work through the other checks.

cmacknz (commented):
Tested again and it works, although I would like the order in which the Fleet managed checks done re-ordered.

Besides that LGTM. I'm on PTO until next Tue, if the order of the checks if fixed you can merge without waiting for my explicit approval.

internal/pkg/agent/cmd/upgrade_test.go
kaanyalti: This makes much more sense. I just looked at the already implemented test case and followed the same approach for the new ones that I added. Comparing the test cases I added to the one that was already implemented, I see that the new ones don't necessarily need an actual server running as we are not testing for connection interruptions. Will look into this one; however, it is not a priority for me. I may open a new PR to refactor these test cases.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -2,6 +2,8 @@
 // or more contributor license agreements. Licensed under the Elastic License 2.0;
 // you may not use this file except in compliance with the Elastic License 2.0.
 
+// you may not use this file except in compliance with the Elastic License 2.0.
```
andrzej-stencel, line 5: merge artifact?

internal/pkg/agent/cmd/upgrade_test.go
kaanyalti: Updating this now actually. Instead of mock server, I am using mock client

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -31,6 +32,14 @@ const (
 	flagPGPBytes       = "pgp"
 	flagPGPBytesPath   = "pgp-path"
 	flagPGPBytesURI    = "pgp-uri"
+	flagForce          = "force"
+)
+
+var (
+	unsupportedUpgradeError   error = errors.New("this agent is Fleet managed and must be upgraded using Fleet")
+	nonRootExecutionError           = errors.New("upgrade command needs to be executed as root for fleet managed agents")
+	skipVerifyNotAllowedError       = errors.New(fmt.Sprintf("\"%s\" flag is not allowed when upgrading a fleet managed agent using the cli", flagSkipVerify))
```
andrzej-stencel, line 41: Inconsistent casing: should "Fleet" start with a capital "F" or not?

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -53,24 +63,119 @@ func newUpgradeCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Comman
...
+
 func upgradeCmd(streams *cli.IOStreams, cmd *cobra.Command, args []string) error {
 	c := client.New()
-	return upgradeCmdWithClient(streams, cmd, args, c)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	err := c.Connect(ctx)
+	if err != nil {
+		return errors.New(err, "Failed communicating to running daemon", errors.TypeNetwork, errors.M("socket", control.Address()))
```
andrzej-stencel, line 92: Errors should start with lowercase https://go.dev/wiki/Errors
```suggestion
		return errors.New(err, "failed communicating to running daemon", errors.TypeNetwork, errors.M("socket", control.Address()))
```

andrzej-stencel (approved):

pkg/control/v1/proto/control_v1_grpc.pb.go
pchila, line 1: Has this been edited by hand even if it's generated code, or is it just a side effect of a newer version of `protoc` ?

pkg/control/v1/proto/control_v1_grpc.pb.go
```diff
@@ -20,8 +25,15 @@ import (
 
 // This is a compile-time assertion to ensure that this generated file
 // is compatible with the grpc package it is being compiled against.
-// Requires gRPC-Go v1.32.0 or later.
```
pchila, line 23: Not sure if we document anywhere the minimum version we require for `gRPC-Go` because we jumped a bit forward here

internal/pkg/agent/application/coordinator/coordinator.go
kaanyalti: Ah, good catch, most likely, I didn't add it manually, removing it

internal/pkg/agent/cmd/upgrade.go
kaanyalti: I'll resort to using lower case

pkg/control/v1/proto/control_v1_grpc.pb.go
kaanyalti: This is a side effect of using a newer version of protoc

pkg/control/v1/proto/control_v1_grpc.pb.go
kaanyalti: that is quite a bit of a jump yes, I didn't find anything that's pinning the grpc-go version. It maybe a good idea for mage commands to attempt to install dependencies rather.

leandrojmp:
Hello @cmacknz, @kaanyalti 

In which version this starting working? This makes agent management a complete nightmare in some use cases.

We are just finishing a migration from on-premises to Elastic Cloud after months working with Elastic and the last step is to re-enroll thousands of agents, because we cannot migrate it, after that we would ask all teams to upgrade the agents to the target version using `elastic-agent upgrade VERSION`.

I've just discovered now that this is not possible anymore, what is the reasoning behind this change? 

We have thousands of hosts, that are managed by dozens of different teams, for compliance reasons I cannot start the upgrade of something on a machine that my team is not the administrator, so to upgrade the agents we have to tell each team to run the upgrade command and this is automated.

This change breaks this flow, I'm not even sure how I would be able to upgrade the agents now, this severily limits the ways we have to manage the agent, I cannot see what problem this solves, having to confirm a prompt without the option to skip it is extremely unfriendly to users that have to manage hundreds, sometime thousands of agents.

leandrojmp:
Hello @cmacknz @kaanyalti 

To which version was this backported?

This change is very unfriendly to users and makes the management of elastic agents even more complicated than they already are.

We rely on the `elastic-agent upgrade` command so many different teams can upgrade their agents to a specific target version when asked to do so, this is done using different kinds of automation tools.

This change basically breaks this flow as it seems that even using `--force` you would still be required to do this in an interactive way, the use of `--force` should be enough no extra confirmation should've be required.

What problem this solve? What was the reasoning behind this change? At a first glance this just adds more work on the shoulders of users.

Fleet UI has a lot of limitations and we rely on the CLI commands and API Endpoints to work around these limitations, changes like this makes every thing more complicated.

Is it possible for this to be reverted or at least not require interaction?

leandrojmp:
Hello @cmacknz and @kaanyalti ,

It is possible to provide more context about this change? We just upgraded our cloud deployment from `8.17.3` to `8.18.1`, we rely on manually upgrading some agents using `elastic-agent upgrade <target-version>` and now we need to use the `--force`, but we just know that we need to use `--force` because we found this issue before as this is not present on the documentation.

All our agents are installed in privileged mode, will this have any impact in the future? Why the `--force` parameter is hidden from the documetnation? For agents installed using rpm/deb, the upgrade steps are still the same, right? Just upgrade the packages?

PR #5877, opened 2024-10-30
fix(5595): added logic to use development binary path for uninstall

### LINKED ISSUE #5595: `--force` ignores `--develop` and uninstalls production agent instead
When installing an development elastic-agent, using `--develop`, if it's done to override an existing development agent, using `--force`, the development agent will uninstall the production agent.

It seems the `--force` ignores the `--develop` and runs the uninstall command against the production agent. It creates two problems:
 - the develop agent uninstalls the production agent
 - if the uninstall succeeds, the install fails as the development agent hasn't been removed.

- Version: 8.15.0, main
- Operating System: linux, mac, needs test on Windows, but it should be affected as well
- Discuss Forum URL: N/A
- Steps to Reproduce:
  - install an production agent
  - install a development agent
  - try to override the development agent: `sudo elastic-agent install -nf [...]`
  - it'll uninstall the production agent and fail:
```
root@elastic-agent:~# ./elastic-agent-8.15.0-linux-x86_64/elastic-agent install -fn --develop --url=https://fleet.elastic-cloud.com:443 --enrollment-token=a-token==
[    ] Done  [39s]                                    
Elastic Agent has been uninstalled. ## <---- the production agent was uninstalled
[ ===] Error copying files  [39s] Error uninstalling. Printing logs
Error: error installing package: failed to copy source directory (data/elastic-agent-25075f) to destination (data/elastic-agent-8.15.0-25075f): open /opt/Elastic/Agent-Development/data/elastic-agent-8.15.0-25075f/components/agentbeat: text file busy # <---- the develop agent hasn't been removed
```

kaanyalti:
- Bug

## What does this PR do?

Updates how the `upgrade` command works when used with the `--force` and `--develop` flags. Currently if a development agent is already installed, and if a user wants to force install a new development agent, the command uninstalls the production agent, if it exists, and tries to install the development agent again. 

## Checklist

- Install and enroll production agent 
- Install and enroll development agent
- Run the development installation command with `--force` flag and confirm that at the end of the operation both `elastic-agent` and `elastic-development-agent` binaries are installed and that both agents are healthy 

## Related issues

- Closes #5595

### DIFF
--- a/internal/pkg/agent/application/paths/paths_darwin.go
+++ b/internal/pkg/agent/application/paths/paths_darwin.go
@@ -10,6 +10,10 @@ const (
 	// BinaryName is the name of the installed binary.
 	BinaryName = "elastic-agent"
 
+	// DevelopmentBinaryName is the name of the installed binary when --develop
+	// flag is used
+	DevelopmentBinaryName = "elastic-development-agent"
+
 	// DefaultBasePath is the base path used by the install command
 	// for installing Elastic Agent's files.
 	DefaultBasePath = "/Library"
--- a/internal/pkg/agent/cmd/install.go
+++ b/internal/pkg/agent/cmd/install.go
@@ -234,7 +234,8 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 		// Uninstall the agent
 		progBar.Describe("Uninstalling current Elastic Agent")
 		if !runUninstallBinary {
-			err := execUninstall(streams)
+			bn := getBinaryName(isDevelopmentMode)
+			err := execUninstall(bn, streams)
 			if err != nil {
 				progBar.Describe("Uninstall failed")
 				return err
@@ -328,13 +329,20 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 	return nil
 }
 
+func getBinaryName(isDev bool) string {
+	if isDev {
+		return paths.DevelopmentBinaryName
+	}
+	return paths.BinaryName
+}
+
 // execUninstall execs "elastic-agent uninstall --force" from the elastic agent installed on the system (found in PATH)
-func execUninstall(streams *cli.IOStreams) error {
+func execUninstall(binaryName string, streams *cli.IOStreams) error {
 	args := []string{
 		"uninstall",
 		"--force",
 	}
-	execPath, err := exec.LookPath(paths.BinaryName)
+	execPath, err := exec.LookPath(binaryName)
 	if err != nil {
 		return fmt.Errorf("unable to find %s on path: %w", paths.BinaryName, err)
 	}
--- a/internal/pkg/agent/cmd/install.go
+++ b/internal/pkg/agent/cmd/install.go
@@ -344,7 +344,7 @@ func execUninstall(binaryName string, streams *cli.IOStreams) error {
 	}
 	execPath, err := exec.LookPath(binaryName)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


michalpristas:
please check integration test failures, these are related

testing/integration/install_test.go
```diff
@@ -408,8 +407,64 @@ func TestRepeatedInstallUninstall(t *testing.T) {
...
+	require.NoError(t, installtest.CheckSuccess(ctx, devFix, devTopPath, &installtest.CheckOpts{Privileged: devInstOpt.Privileged, Namespace: devInstOpt.Namespace}))
+
+	devFix2, err := define.NewFixtureFromLocalBuild(t, define.Version())
+	require.NoError(t, err)
+	err = devFix2.Prepare(ctx)
+	require.NoError(t, err)
+	devOut2, err := devFix2.Install(ctx, &devInstOpt)
+	if err != nil {
+		t.Logf("install output: %s", devOut2)
+		require.NoError(t, err)
```
michalpristas, line 458: this will always fail if err is not nil

internal/pkg/agent/cmd/install.go
```diff
@@ -328,13 +329,20 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
...
-func execUninstall(streams *cli.IOStreams) error {
+func execUninstall(binaryName string, streams *cli.IOStreams) error {
 	args := []string{
 		"uninstall",
 		"--force",
 	}
-	execPath, err := exec.LookPath(paths.BinaryName)
+	execPath, err := exec.LookPath(binaryName)
 	if err != nil {
 		return fmt.Errorf("unable to find %s on path: %w", paths.BinaryName, err)
```
michalpristas, line 347: update to proper exec path

testing/integration/install_test.go
kaanyalti: I was following similar implementation in the prior test cases, but I can remove the if statement and just assert agains the error.

AndersonQ: just in case you don't know `assert.X` will just add a failure to the test, whereas `require.X` will abort the test. So the idea is to use `assert.X` for the actual test assertions and `require.X` during setup to avoid the test continuing after critical failure

kaanyalti: I did not know that, thank you

internal/pkg/agent/application/paths/paths_darwin.go
```diff
@@ -10,6 +10,10 @@ const (
 	// BinaryName is the name of the installed binary.
 	BinaryName = "elastic-agent"
 
+	// DevelopmentBinaryName is the name of the installed binary when --develop
+	// flag is used
+	DevelopmentBinaryName = "elastic-development-agent"
+
```
blakerouse, line 16: I don't think you can do this, as the `--develop` is just an alias for the `--namespace` flag. That allows multiple agents to be installed on the same host at the same time. The name of the executable here should not be hard coded as it should either come from the `--namespace` flag or from infering the namespace by reading the namespace from the path that is loaded using the full filepath from `os.Args[0]`.

blakerouse (commented):
I don't think this is taking the correct approach currently. I would expect the usage of the following would allow you to get the correct path to the binary that should be called to perform the uninstall.

https://github.com/elastic/elastic-agent/blob/b6a608d950dfbe9dbf41eed5a199d31555310d2d/internal/pkg/agent/application/paths/common_namespace.go#L43

kaanyalti:
Closing, will be changing implementation drastically.

PR #5879, opened 2024-10-30
Assume installed agent on upgrade

### LINKED ISSUE #5872: Windows agent fall backs to 7.17.25 on upgrade to 8.16.0 BC2.
**Kibana Build details:**
```
VERSION: 8.16.0 BC2
BUILD: 79434
COMMIT: 59220e984f2e3ca8b99fe904d077a5979f5f298d
```
Artifact: https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-7.17.25-windows-x86_64.zip

**Preconditions:**
1. 8.16.0-BC2 Kibana cloud environment should be available.
2. 7.17.25 Windows agent should be installed with System and Elastic Defend integration.

**Steps to reproduce:**
1. Navigate to Fleet>Agents tab.
2. Upgrade Windows 7.17.25 agent to 8.16.0.
3. Observe agent starts upgrading and falls back to 7.17.25.

**What's working fine:**
- Issue is not observed on Linux and MAC agent upgrades from 7.17.25>8.16.0 BC2.
[elastic-agent-diagnostics-2024-10-29T10-36-01Z-00.zip](https://github.com/user-attachments/files/17555509/elastic-agent-diagnostics-2024-10-29T10-36-01Z-00.zip)

- Issue is not observed on Windows agent upgrade from 8.15.3>8.16.0 BC2.
[elastic-agent-diagnostics-2024-10-29T10-41-37Z-00.zip](https://github.com/user-attachments/files/17555597/elastic-agent-diagnostics-2024-10-29T10-41-37Z-00.zip)

**Expected Result:**
Windows agent shouldn't fall back to 7.17.25 on upgrade to 8.16.0 BC2.

**Logs:**
[elastic-agent-diagnostics-2024-10-29T10-23-14Z-00.zip](https://github.com/user-attachments/files/17555459/elastic-agent-diagnostics-2024-10-29T10-23-14Z-00.zip)
[elastic-agent-diagnostics-2024-10-29T10-03-01Z-00.zip](https://github.com/user-attachments/files/17555461/elastic-agent-diagnostics-2024-10-29T10-03-01Z-00.zip)

michalpristas:
## What does this PR do?

This PR assumes installed agent on upgrade.
Detecting installed based on marker file does not work prior to 8.8 as marker file is not there leading to incorrect control path. 
During upgrade we resolve paths to proper value assuming upgrading agent is installed (as this is requirement anyways). 

Other way would be to keep `runningInstalled` evaluation based on paths as before OR have two ways of detecting installed agent during upgrade based on previous agent version. 

## Why is it important?

To make ugprade from <8.8 work

Fixes #5872


internal/pkg/agent/application/paths/paths_windows.go
```diff
@@ -65,8 +65,13 @@ func initialControlSocketPath(topPath string) string {
 // RunningInstalled will always be false, even when it is an installed version. Once
 // that is fixed from the upgrade process the control socket path needs to be updated.
 func ResolveControlSocket() {
+	ResolveControlSocketWithInstalledState(RunningInstalled())
+}
+
+// ResolveControlSocketWithInstalledState updates the control socket path.
+func ResolveControlSocketWithInstalledState(runningInstalled bool) {
```
pchila, line 72: Can we get some unit tests to maintain and document the behavior of `paths.ResolveControlSocketWithInstalledState(true)` ?  (we probably need to abstract the calls to `TopPath()` with a parameter and some overloading in order to use temp directories in test)

pchila (commented):
Fix looks reasonable.

I wish we would have more tests that document behavior of some functions (unit tests should be enough for this) and find out why this problem slipped through integration tests.

Are we missing a test? 
The current tests are not run correctly ? 
Are we not waiting long enough to see a rollback as @cmacknz was describing in https://github.com/elastic/elastic-agent/issues/5872#issuecomment-2445223213 ?

cmacknz:
Why doesn't our existing 7.17.x -> 8.16 integration test fail? We need to identify that and fix it as part of this work, it's the root cause for why we are finding this bug so late.

internal/pkg/agent/application/paths/paths_test.go
```diff
@@ -92,3 +94,123 @@ func TestHasPrefixWindows(t *testing.T) {
 		})
 	}
 }
+
+func TestResolveControlSocketWithInstalledState(t *testing.T) {
```
pchila, line 98: Thank you, I appreciate that you added a unit test.

So, we only have one case where `ResolveControlSocketWithInstalledState` returns a different result than `ControlSocketFromPath` and it's for Windows when the agent is detected as installed...

I was more thinking about having a small UT for [ResolveControlSocket()](https://github.com/elastic/elastic-agent/pull/5879/files#diff-8260d5183b15bf6f7e77073338ec4994ce3a058aa618b4b015993f2f56b18474R67) where we can prepare a temp directory to be the top path with and without the `.installed` marker and having explicit expected results.

michalpristas:
updated grace period configuration for fast watcher to be 100s
why we did not detect failure for this is that we fail on 5th connection error. 5*15s for fast watcher is > `60s` grace configured for it. so we did not encounter any connection issue within the period and assumed everything is fine.

100s gives some space for connections to fail and detect failures. 
I verified that with this config without a fix introduced in this PR test fails

internal/pkg/agent/application/paths/paths_test.go
michalpristas: i understand what you're trying to achieve.
i'd rather address this in a followup as we may hit the window for next release with actual state of the PR and tests which should be good enough to get us covered.

michalpristas:
integration tests failing on missing template fixed in [this PR](https://github.com/elastic/elastic-agent/pull/5910) created by @ycombinator

internal/pkg/agent/application/paths/paths_test.go
cmacknz: `ResolveControlSocket()` is only called from a single place, and that code already calls `paths.RunningInstalled()` right before `ResolveControlSocket()`.

https://github.com/michalpristas/elastic-agent/blob/be8773ff18061ec0a7570fa43db902ff86e709fe/internal/pkg/agent/cmd/run.go#L713-L736

So IMO `ResolveControlSocket` just should always take whether we know we are installed as an argument and we can avoid having two functions. Then having the installed detection is never a part of this functions job, so we don't need to test that part here at all.

Just reading the test cases here it is really hard to get a picture of what is supposed to happen, fundamentally:
1. When on Windows, whether you are installed or not changes the result.
2. On Darwin and Linux, being installed makes no difference.

It's not clear why we have 3 test cases for each platform to me, when there are naturally only 2.

It feels like the tests could be simpler.

P.S. 8.16 is likely going to be delayed and there will be another BC sometime next week. If there isn't, this PR has already missed the release. So let's not worry about landing this and instead focus on whether we are happy with it as that is already out of our control until we know the next BC date.

michalpristas: will update test just a note we have 4 test cases per each platform not 3. 
they are following matrix of values installed vs control socket differs to expected
- installed same paths,
- installed different paths,
- not installed same paths
- not installed different paths

what we track is if control socket persists the same except for windows use case when it's installed and paths are different when we fix the path to be `npipe:///elastic-agent-system`

michalpristas: test logic stayed the same, hope that change for test to be more matrix like will make it more clear/readable
```go
		{"darwin", false, false, false, ""},
		{"darwin", true, false, false, ""},
		{"darwin", false, true, false, ""},
		{"darwin", true, true, false, ""},

		{"linux", false, false, false, ""},
		{"linux", true, false, false, ""},
		{"linux", false, true, false, ""},
		{"linux", true, true, false, ""},

		{"windows", false, false, false, ""},
		{"windows", true, false, false, ""},
		{"windows", false, true, false, ""},
		{"windows", true, true, false, WindowsControlSocketInstalledPath},
```

pchila (approved):
LGTM

Issue #5835, opened 2024-10-23
Elastic agent uses too much memory per Pod in k8s


swiatekm:
In its default configuration, agent has the kubernetes provider enabled by default. In DaemonSet mode, which is our default, this provider keeps track of data about Pods scheduled on the Node the agent is running on. This issue concerns the fact that the agent process itself uses an excessive amount of memory if the number of these Pods is high (for the purpose of this issue, this will mean close to the default Kubernetes limit of 110). This was originally discovered while troubleshooting #4729.

This effect is visible even if we disable all inputs and self-monitoring, leaving agent to run as a single process without any components. This strongly implies it has to do with configuration variable providers. I used this empty configuration in my testing to limit confounding variables from beats, but the effect *is* more pronounced when components using variables are present in the configuration.

Here's a graph of agent memory consumption as the number of Pods on the Node increases from 10 to 110:

![Image](https://github.com/user-attachments/assets/3de2f058-e585-4cd7-b6ea-9a2b85cb5a25)

A couple of observations from looking at how configuration changes affect this behaviour:

* Making the garbage collector more aggressive makes most of the effect disappear, as does restarting the Pod. Very likely that this is caused by allocation churn more than steady-state heap utilization.
* Disabling the host provider also reduces the effect greatly. When creating variables, each entry from a dynamic provider gets its own copy of data from context providers. On a large node, there can be quite a bit of host provider data. This is also visible when looking at variables.yml in diagnostics.
* Increasing the debounce time on variable emission in the composable coordinator doesn't help much.

## Test setup

* Single node KiND cluster, default settings.
* Default standalone kubernetes manifest, all inputs removed from configuration.
* A single nginx Deployment, starting at 0 replicas, and later scaled up to the maximum the Node allows.

## More data

I resized my Deployment a couple times and looked at a heap profile of the agent process:

![Image](https://github.com/user-attachments/assets/3d5612f0-20f8-4c55-8faf-933cdb41f171)

The churn appears to be coming primarily from needing to recreate all the variables whenever a Pod is updated. The call to `composable.cloneMap` is where we copy data from the host provider.

## Root cause

The root cause appears to be a combination of behaviours in the variable pipeline:

1. All variables are recreated and emitted whenever there's any change to the underlying data. In Kubernetes, with a lot of Pods on a Node, changes can be quite frequent, and the amount of data is non-trivial.
[...]

PR #5884, opened 2024-10-30
Share host info between packages

swiatekm:
## What does this PR do?

Lets packages share the same host info struct. We add a thread-safe wrapper about `sysinfo.Host` and keep a singleton that callers can transparently request and use.

The host provider is exempted from this change, as it explicitly wants to update this information periodically. I added a comment explaining why that is.

## Why is it important?

There's several packages where we need the host information. Thus far, each of them just computed it on their own, which was wasted effort. Normally this wouldn't be a big deal, as it only happens once at startup. However, this call can be expensive in some circumstances - most notably when the host has a lot of network interfaces. See https://github.com/elastic/elastic-agent/issues/5835#issuecomment-2441544789 for an example in Kubernetes. The result is agent having unnecessarily high memory consumption after startup, which is annoying.

## Checklist

Starting agent with any configuration and looking at the host information emitted in the logs is good enough. Unit testing this functionality is quite awkward unfortunately.

## Related issues

- Relates #5835

### DIFF
--- a/internal/pkg/composable/providers/host/host.go
+++ b/internal/pkg/composable/providers/host/host.go
@@ -132,6 +132,7 @@ func ContextProviderBuilder(log *logger.Logger, c *config.Config, _ bool) (corec
 
 func getHostInfo(log *logger.Logger) func() (map[string]interface{}, error) {
 	return func() (map[string]interface{}, error) {
+		// We don't use the shared host info from util here, as we explicitly want the latest host information on every call.
 		sysInfo, err := sysinfo.Host()
 		if err != nil {
 			return nil, err
--- a/internal/pkg/util/host.go
+++ b/internal/pkg/util/host.go
@@ -6,9 +6,11 @@ package util
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/elastic/elastic-agent/pkg/core/logger"
+	"github.com/elastic/go-sysinfo"
 	"github.com/elastic/go-sysinfo/types"
 )
 
@@ -30,4 +32,62 @@ func GetHostName(isFqdnFeatureEnabled bool, hostInfo types.HostInfo, host types.
 	}
 
 	return fqdn
+}
+
+var _ types.Host = &threadSafeHost{}
+
+// threadSafeHost is a thread-safe wrapper around types.Host.
+// It exists so we can only create it once, as some of the setup it does is relatively expensive.
+type threadSafeHost struct {
+	sync.Mutex
+	inner types.Host
+}
+
+func newThreadSafeHost(inner types.Host) *threadSafeHost {
+	return &threadSafeHost{inner: inner}
+}
+
+func (s *threadSafeHost) CPUTime() (types.CPUTimes, error) {
+	s.Lock()
+	defer s.Unlock()
+	return s.inner.CPUTime()
+}
+
+func (s *threadSafeHost) Info() types.HostInfo {
+	s.Lock()
+	defer s.Unlock()
+	return s.inner.Info()
+}
+
+func (s *threadSafeHost) Memory() (*types.HostMemoryInfo, error) {
+	s.Lock()
+	defer s.Unlock()
+	return s.inner.Memory()
+}
+
+func (s *threadSafeHost) FQDNWithContext(ctx context.Context) (string, error) {
+	s.Lock()
+	defer s.Unlock()
+	return s.inner.FQDNWithContext(ctx)
+}
+
+func (s *threadSafeHost) FQDN() (string, error) {
+	s.Lock()
+	defer s.Unlock()
+	return s.inner.FQDN()
+}
+
+var (
+	sharedHost types.Host
+	once       sync.Once
+	hostErr    error
+)
+
+func GetHost() (types.Host, error) {
+	once.Do(func() {
+		var innerHost types.Host
+		innerHost, hostErr = sysinfo.Host()
+		sharedHost = newThreadSafeHost(innerHost)
+	})
+	return sharedHost, hostErr
 }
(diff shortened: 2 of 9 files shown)


internal/pkg/util/host.go
```diff
@@ -31,3 +33,61 @@ func GetHostName(isFqdnFeatureEnabled bool, hostInfo types.HostInfo, host types.
...
+	sync.Mutex
+	inner types.Host
+}
+
+func newThreadSafeHost(inner types.Host) *threadSafeHost {
+	return &threadSafeHost{inner: inner}
+}
+
+func (s *threadSafeHost) CPUTime() (types.CPUTimes, error) {
+	s.Lock()
```
pkoutsovasilis, line 51: my first reaction/thinking was that this thread-safety code most probably belongs to the go-sysinfo codebase?! but then looking again in the exposed functions, and from a quick look in go-sysinfo they do not mutate anything, so I am wondering do we need to hold a mutex to call them?! did I miss a mutation?

pierrehilbert:
With @blakerouse in PTO, probably better to assign someone else from the Control Plane team to this review.
cc @ycombinator

internal/pkg/util/host.go
swiatekm: go-sysinfo doesn't guarantee that these are thread-safe, so I don't have much of a choice here. We could try to make them so in the library, but I'm not sure it's worth it. These functions aren't called very often, so in practice there shouldn't be any contention.

pkoutsovasilis: since these functions are not mutating anything I would propose to remove the mutex. Or again you may have spotted a mutation that I missed?! If not, I am not entirely convinced that losing the concurrency of calling these functions, given that this is now a shared instance across multiple places, is wise. But if you insist sure go with it 🙂

ycombinator:
> With @blakerouse in PTO, probably better to assign someone else from the Control Plane team to this review. cc @ycombinator

I already chatted with @swiatekm about this.  This PR is not urgent so we are OK keeping the reviewers as is.

pkoutsovasilis (approved):
LGTM. I still hold some reservations regarding the necessity of a thread-safe inner host in this context, given that these functions aren’t mutating any state and introducing a mutex could impact concurrent usage. However, since the usage frequency is low, I’m fine proceeding with this for now. We can revisit if contention becomes an issue. Nice work overall! 🙂

PS: The CI failures seem unrelated to this PR and I think that pulling from main will make the CI happy again

internal/pkg/util/host.go
swiatekm: The functions are not mutating anything *right now*, but go-sysinfo doesn't guarantee this interface is thread-safe, so this is an implementation detail that may change. If we want to remove the locks, we should make the change in go-sysinfo, and I don't think that's worth the hassle. This PR is a bit borderline already in my opinion when it comes to complexity introduced vs performance gained.

At the very least, I'd have to see evidence of actual contention before looking into this further.

internal/pkg/composable/providers/host/host.go
```diff
@@ -132,6 +132,7 @@ func ContextProviderBuilder(log *logger.Logger, c *config.Config, _ bool) (corec
 
 func getHostInfo(log *logger.Logger) func() (map[string]interface{}, error) {
 	return func() (map[string]interface{}, error) {
+		// We don't use the shared host info from util here, as we explicitly want the latest host information on every call.
```
cmacknz, line 135: I would prefer if we didn't have two different ways to get Host information where callers need to think about which one to use, or whether they actually need the parts of it that can change at runtime (FQDN for an obvious example).

Can the periodic updates be pushed into `util.GetHost()`? They could be unconditional updates at a fixed rate, or it could behave as a rate limit where it updates at most every X seconds.

As a possible simplification for all of this, perhaps the network address information isn't valuable at all for containerized applications and we can add a way to filter it into https://github.com/elastic/go-sysinfo. There is an [IsContainerized()](https://github.com/elastic/go-sysinfo/blob/750e2c7a2db68de43d0b2ad7158f4e061de1a200/providers/linux/container.go#L30) already.

cmacknz: For some related context, `add_kubernetes_metadata` stopped including the host IPs by default:
- https://github.com/elastic/integrations/issues/667

cmacknz: Perhaps we could just filter out the link local addresses (assuming those are the problem here too) if just not having the addresses at all doesn't work, or its too risky to think being in a container always means you are in k8s (it doesn't). https://github.com/elastic/integrations/issues/6674#issuecomment-1609000789

swiatekm: I feel like the periodicity of the host info checks in the provider should be a part of the provider's logic, as opposed to being hidden in the utils package. I could see adding a parameter to this API indicating if a cached value is fine, if that helps.

Honestly, what I'm really trying to do here is avoid recomputing the host info, which is only done at the start. All the other methods on this struct fetch the latest data anyway. Maybe the solution is to compute host info at call time, same as everything else? Then every component could have its own `types.Host` (as creating one would be cheap), and we'd separately cache host info for the packages that want it at startup.

kaanyalti (approved):
Code looks good however I’m less familiar with the underlying context, so I can’t fully judge the implementation’s accuracy.

internal/pkg/composable/providers/host/host.go
cmacknz: I think all I really care about is that it's obvious when a call to get host info will potentially contain stale data. Keeping the periodicity of the host info check in the provider is fine if that is clear.

swiatekm:
I'm closing this for now, we should change the underlying library to always recompute host info, and then cache that. https://github.com/elastic/go-sysinfo/pull/246 reduces the cost of this call significantly, so sharing is less important now.

Issue #5716, opened 2024-10-07
Add an integration test using a proxy with mTLS for control plane and Elastic Defend installed


AndersonQ:
This test should validate the mTLS flow for control-plane (communication with fleet-server) through an proxy with mTLS and having Elastic Defend integration installed.

The test should:
 - use RSA TLS certificates (Elastic Defend restriction)
 - validate passphrase protected certificate key for the client works. (As the server is the proxy, it's pointless to test it supports passphrase protected key)
 - use [proxytest](https://github.com/elastic/elastic-agent/tree/main/testing/proxytest) for proxy
 - use Elastic Cloud for the Elastic Stack
 - configure a wrong fleet-server address, which the proxy should correct
 - validate connections are flowing through the proxy. Ideally the proxy will assert it. However just the fact the configured host is invalid is enough
   - for the Elastic Agent
   - for Elastic Defend

### Test scenarios:
- [x] mTLS configuration passed throughout the CLI at install/enroll time. No configuration in the policy
  - [x] with passphrase protected key
  - [x] without passphrase protected key
- [x] mTLS configuration passed through CLI and present in the policy. The policy configuration should take precedence
  - [x] both (cli and policy) using plain certificate key
  - ~~[ ] both (cli and policy) using passphrase protected key~~ not possible, the proxy UI does not allow to add a passphrase or passphrase path
  - [x] cli configuration with passphrase protected key, policy configuration **without** passphrase protected key
  - [x] cli with mTLS and passphrase protected key, policy with on-way TLS. The policy must take precedence and remove the client TLS configuration
- ~~[x] mTLS configuration only in the policy:~~ test skipped, see https://github.com/elastic/elastic-agent/issues/5888
  ~~- fleet-server is reachable without proxy~~
  ~~- the fleet-server host configured in the policy is only reachable through the proxy~~
- [x] removing the proxy from the policy does not remove it from the agent:
  - it's the same as "an empty proxy from the policy does not change the current configuration" 

### Out of scope:
 - data plane connections (with Elasticsearch or any other output)
 - any test on the fleet-server side of things
 - certificate rotation

Issue #5491, opened 2024-09-10
Add an integration test for `--elastic-agent-cert-key-passphrase` with Elastc Defend installed


AndersonQ:
In order to fully validate https://github.com/elastic/elastic-agent/issues/4515 a integration test should be added. The integration test should run the Elastic Agent using mTLS for control plane communication, either through a proxy or directly connecting to fleet-server.
The test needs to validate:
 - the Elastic Agent is running and successfully communicating with fleet-server
 - a client certificate in the policy takes precedence over the CLI configuration (`--elastic-agent-cert-key-passphrase`). That means, a policy client TLS configuration fully overrides the one coming from the CLI, the `cert-key-passphrase` is cleared if the CLI define it but not the policy
 - a policy with no client TLS configuration does not override the CLI configuration
 - Elastic Defend is installed and successfully connecting to fleet-server
   - open question: is relying on Elastic Defend status enough to validate that?

This issue is blocked by: 
 - https://github.com/elastic/elastic-agent/issues/4903

PR #5889, opened 2024-10-30
Add integration tests using a proxy with mTLS for control plane with Elastic Defend installed

AndersonQ:
## What does this PR do?

Add integration tests for the Elastic Agent running Elastic Defend with a mTLS proxy
See https://github.com/elastic/elastic-agent/issues/5716 for tje covered test cases

## Why is it important?

To have automated tests covering mTLS and Elastic Defend

## Checklist

- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool]~~(https://github.com/elastic/elastic-agent#changelog)

## Disruptive User Impact

 - N/A

In short, manually reproduce the tests described on https://github.com/elastic/elastic-agent/issues/5716.

The long version:
 - run 3 proxies:
   - mTLScli -> a proxy with mTLS to be passed as cli arguments
   - mTLSpolicy -> a proxy with mTLS configured through the policy
   - oneWayTLSpolicy -> a proxy with simple/one way TLS which defines its CA in the policy
   - all the certificates must be RSA because of an Elastic Defend restriction

 - set up the proxies:
   - add a test to `testing/integration/endpoint_security_test.go` to configure and run the needed proxies:
```go
func TestDebugTestProxies(t *testing.T) {
	prepareProxies(t, &url.URL{Host: "a.wrong.fleet.host"}, "the.real.fleet.host")

	t.Logf("CTRL + C to exit")
	<-make(chan struct{})
}
```
 
   - run the test with:
```
go test -v -timeout 0 -tags integration -run TestDebugTestProxies ./testing/integration/

=== RUN   TestDebugTestProxies
    endpoint_security_test.go:1333: [mtlsCLI] certificates saved on: /tmp/TestDebugTestProxies-mtlsCLI2475114678
    endpoint_security_test.go:1333: [mtlsPolicy] certificates saved on: /tmp/TestDebugTestProxies-mtlsPolicy1317235685
    endpoint_security_test.go:1333: [oneWayTLSPolicy] certificates saved on: /tmp/TestDebugTestProxies-oneWayTLSPolicy327431283
```

   - check the logs like `[%s] certificates saved on: %s`. they will show where the certificates are.
   - set up the necessary fleet hosts and proxies on fleetUI
   - install/enroll the agent using the configured fleet hosts and proxies. Again, check the test cases on https://github.com/elastic/elastic-agent/issues/5716

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/5491
- Closes https://github.com/elastic/elastic-agent/issues/5716

## Questions to ask yourself

- ...


AndersonQ:
folks (@blakerouse, @pchila, @kaanyalti), all the tests are passing now. In case you were waiting the tests to pass to review it

blakerouse (commented):
Overall this looks good.

I do question the change from returning errors to using `require.NoError` directly. Why was that change made?

AndersonQ:
> I do question the change from returning errors to using `require.NoError` directly. Why was that change made?

to follow the convention we have on our tests. The majority of the tests use this approach

testing/integration/endpoint_security_test.go
```diff
@@ -894,3 +901,536 @@ func TestForceInstallOverProtectedPolicy(t *testing.T) {
...
+	fleetKibanaHostmTLS, err := stack.KibanaClient.CreateFleetServerHosts(ctx, kibana.ListFleetServerHostsRequest{
+		ID:        "proxyPolicymTLS" + testUUID,
+		Name:      "proxyPolicymTLS" + testUUID,
+		HostURLs:  []string{fleethostWrong.String()},
+		IsDefault: false,
+		ProxyID:   fleetProxymTLS.Item.ID,
+	})
+	require.NoError(t, err, "error creating fleet host with mTLS proxy")
+
+	// add new fleet-server host with oen way TLS proxy
```
pchila, line 961:
```suggestion
	// add new fleet-server host with one way TLS proxy
```

testing/integration/endpoint_security_test.go
```diff
@@ -894,3 +901,536 @@ func TestForceInstallOverProtectedPolicy(t *testing.T) {
...
+					buff.Reset()
+
+					got, err := f.ExecInspect(ctx)
+					if err != nil {
+						buff.WriteString(fmt.Sprintf("error running inspect cmd: %v", err))
+						return false
+					}
+
+					return proxyPolicymTLS.URL == got.Fleet.ProxyURL
+				}, time.Minute, time.Second, "inspect never showed proxy from policy: %s", buff)
```
pchila, line 1150: `buff` is really the last error encountered in the assert.Eventually, maybe we can define and treat it as such:



```suggestion
				// feel free to change the variable name
				var buff error
				assert.Eventuallyf(t, func() bool {

					got, err := f.ExecInspect(ctx)
					if err != nil {
						buff = fmt.Errorf("error running inspect cmd: %w", err)
						return false
					}
					// (optional) reset the error
					buff = nil
					return proxyPolicymTLS.URL == got.Fleet.ProxyURL
				}, time.Minute, time.Second, "inspect never showed proxy from policy: %s", buff)
```

testing/integration/endpoint_security_test.go
```diff
@@ -894,3 +901,536 @@ func TestForceInstallOverProtectedPolicy(t *testing.T) {
...
+	defaultFleetHost := fleetServerURL[8:]
+	fleethostWrong, err := url.Parse("https://fixme.elastic.co:443")
+	require.NoError(t, err, "failed parsing fleethostWrong")
+
+	// ================================= proxy =================================
+	mtlsCLI, mtlsPolicy, oneWayTLSPolicy, proxyCLI, proxyPolicymTLS, proxyPolicyOneWayTLS := prepareProxies(t, fleethostWrong, defaultFleetHost)
+
+	// =================== Prepare fleet hosts and proxies ===================
+
+	// create mTLS proxy
```
pchila, line 931: Is it possible to create the policies and enrollment tokens as part of each testcase instead of preparing a bunch of different policies and tokens upfront and only have the testcases specify the token use for enrollment? I know that there's a 1-1 relationship between the enrollment token and the policy but this setup makes the test kinda hard to follow:
- the policy definition is far removed from the assertions on the agent behavior
- each test case enroll exactly once with a given token and the a big assertion function is called to wait for the policy to apply and check the results
- it's kinda difficult to follow which test uses which policy (I know, I know, enrollment token... but still...)

testing/integration/endpoint_test_tools.go
```diff
@@ -128,6 +128,6 @@ func installElasticDefendPackage(t *testing.T, info *define.Info, policyID strin
 		t.Logf("Error installing fleet package: %v", err)
 		return r, fmt.Errorf("error installing fleet package: %w", err)
 	}
-	t.Logf("Endpoint package Policy Response:\n%+v", pkgResp)
```
pchila, line 131: was this log not useful for debugging when tests fail ?

pchila (commented):
A few comments here and there, mostly about readability

testing/integration/endpoint_security_test.go
```diff
@@ -894,3 +901,536 @@ func TestForceInstallOverProtectedPolicy(t *testing.T) {
 	out, err := fixture.Exec(ctx, args)
 	require.Errorf(t, err, "No error detected, command output: %s", out)
 }
+
+func TestInstallDefendWithMTLSandEncCertKey(t *testing.T) {
```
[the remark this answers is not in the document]
AndersonQ: what about mTLS -> `TestInstallDefendWithmTLSAndEncCertKey`?

testing/integration/endpoint_security_test.go
AndersonQ: I did that way to minimise the code for each test case and let it be reusable.
Just the `enrollmentTokenOneWayProxyResp` is used only once. the others are used at least by 2 tests.

I get your point, but I believe it'd be worse having each test to set it up. It'd be more repetitive as each test would need to set it up and the test cases would be a lot longer, making it harder to see what each test do.

testing/integration/endpoint_test_tools.go
AndersonQ: the response is returned regardless, even on failure. So it's better to log the response if and only if an error happens. Here it was being logged on success as well.

pchila (approved):
LGTM

Issue #5890, opened 2024-10-30
[Failing Test]: TestAPMConfig – no such index [traces-apm-default]


ycombinator:
### Failing test case

TestAPMConfig

### Error message

```
apm_propagation_test.go:184: received ES response: {"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index [traces-apm-default]","resource.type":"index_or_alias","resource.id":"traces-apm-default","index_uuid":"_na_","index":"traces-apm-default"}],"type":"index_not_found_exception","reason":"no such index [traces-apm-default]","resource.type":"index_or_alias","resource.id":"traces-apm-default","index_uuid":"_na_","index":"traces-apm-default"},"status":404}     
```

### Build

https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4075#0192dd03-00ea-41c0-a56a-a5edb99318eb
https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4077#0192dd14-e4c3-4901-b0d9-f50065d445b5
https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4086#0192dda7-21c4-4d6a-9ce7-2c9a9ac4a3e3

### OS

Linux, Windows

### Stacktrace and notes

```
=== RUN   TestAPMConfig
    fetcher.go:93: Using existing artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz
    fixture.go:288: Extracting artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestAPMConfig4001669197
    fixture.go:306: Completed extraction of artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestAPMConfig4001669197
    fixture.go:963: All component specifications where removed
    fixture.go:1004: Placed component specifications: fake
    apm_propagation_test.go:72: Rendered agent config:
        
        outputs:
          default:
            type: fake-output
        inputs:
          - id: fake-apm
            type: fake-apm
        agent.monitoring:
          traces: true
          apm:
            hosts:
              - https://9c26ecf86b6347b1b8363bce0edebab6.apm.us-west2.gcp.elastic-cloud.com:443
            environment: nsi3gya7ccprbvh2xzveb5ytfy0beb0csqxcuftfqu
            secret_token: AeKIQKeIi5tKM2Wt9s
            global_labels:
              test_name: TestAPMConfig
[...]

PR #5891, opened 2024-10-30
Skip TestAPMConfig 

ycombinator:
## What does this PR do?

This PR skips the `TestAPMConfig` test until we figure out root cause and fix it (see https://github.com/elastic/elastic-agent/issues/5890).

## Why is it important?

To get CI on `main` to green again.


swiatekm (approved):

leehinman:
Added backport-8.x tag because 8.x backports are failing for the same reason

PR #5939, opened 2024-11-05
[helm] derive the lease name from the release and preset names

pkoutsovasilis:
## What does this PR do?

This PR disables by default the kubernetes leader election unless a user explicitly marks it as enabled. More importantly this PR sets the lease name from the Helm release and the agent preset names and thus different agent deployments won't utilise the same lease now.

## Why is it important?

Because it guarantees correct lease isolation across different agent deployments in the same namespace

## Checklist

## Disruptive User Impact

N/A

follow [the kubernetes example](https://github.com/elastic/elastic-agent/tree/main/deploy/helm/elastic-agent/examples/kubernetes-default) and do two installations of the Helm chart with a different release name and under the same namespace (leases are namespace-scoped). With this PR you will see that two leases are created for the `clusterWide` preset where before there would be only one

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/3633 as the user facing issues with the lease name is mentioned there

### DIFF
--- a/deploy/helm/elastic-agent/values.yaml
+++ b/deploy/helm/elastic-agent/values.yaml
@@ -450,8 +450,6 @@ agent:
           logs: true
           metrics: true
       providers:
-        kubernetes_leaderelection:
-          enabled: false
         kubernetes:
           node: ${NODE_NAME}
           scope: node
@@ -481,7 +479,5 @@ agent:
           logs: true
           metrics: true
       providers:
-        kubernetes_leaderelection:
-          enabled: false
         kubernetes:
           enabled: false
--- a/deploy/helm/elastic-agent/examples/eck/rendered/manifest.yaml
+++ b/deploy/helm/elastic-agent/examples/eck/rendered/manifest.yaml
@@ -76,6 +76,7 @@ stringData:
         scope: cluster
       kubernetes_leaderelection:
         enabled: true
+        leader_lease: example-clusterwide
     inputs:
       - data_stream:
           namespace: default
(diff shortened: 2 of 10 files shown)


deploy/helm/elastic-agent/values.yaml
```diff
@@ -450,8 +450,6 @@ agent:
           logs: true
           metrics: true
       providers:
-        kubernetes_leaderelection:
-          enabled: false
```
swiatekm, line 454: I wonder if keeping this explicitly disabled here isn't better. This provider is enabled by default, and I think it's better to be more explicit when diverging from the agent's default behaviour. It would also make the code in `_helpers.tpl` nicer, as we wouldn't need to handle the edge cases.

pkoutsovasilis: @cmacknz any opinions on that? I have no issue reverting this to be enabled by default, however, I had the impression we want to minimise the usage of leader election to solely where it is needed.

swiatekm (approved):

deploy/helm/elastic-agent/values.yaml
cmacknz: I think the way the Helm chart deploys agent makes this much less necessary (e.g. DaemonSet + Deployment instead of just DaemonSet).

IMO we should turn leaderelection on unless it is necessary for correctness.

pkoutsovasilis: > IMO we should turn leaderelection on unless it is necessary for correctness.

I have a feeling that here you mean turn leaderelection **off**, right? 🙂

cmacknz: I meant to write "we should **not** turn leaderelection on unless it is necessary for correctness"

This would mean it is off by default.

PR #5941, opened 2024-11-05
Add `os_family`, `os_platform` and `os_version` to host provider

belimawr:
## What does this PR do?
This commit adds `os_family`, `os_platform` and `os_version` to the host provider, enabling differentiating Linux distributions. This is required to support Debian 12 and other distributions that are moving away from traditional log files in favour of Journald.

## Why is it important?
It is required to support Debian 12 and other Linux distributions that are moving away from traditional log files in favour of Journald.

## Checklist

- [ ] I have made corresponding change to the default configuration files~~

## Disruptive User Impact

None

1. Package the Elastic-Agent
2. Deploy it with any policy, even an empty policy with no monitoring will work
3. Collect the diagnostics
4. Look at `variables.yml` and ensure `os_family`, `os_platform` and `os_version` are populated under `host`. Here is an example from Debian 11:
```
      host:
        architecture: x86_64
        id: 9a7aa218396d47b1b17218a99918bae4
        ip:
            - 127.0.0.1/8
            - 10.0.2.15/24
            - ::1/128
            - fe80::a00:27ff:fe8d:c04d/64
        mac:
            - 08:00:27:8d:c0:4d
        name: bullseye
        os_family: debian
        os_platform: debian
        os_version: 11 (bullseye)
        platform: linux
```

## Related issues

- Relates https://github.com/elastic/integrations/issues/10797
- Is required by https://github.com/elastic/integrations/pull/11618

## Questions to ask yourself

- ...


pkoutsovasilis:
@belimawr could you add a changelog fragment? 🙂

swiatekm (approved):

internal/pkg/composable/providers/host/host_test.go
```diff
@@ -167,3 +168,20 @@ func returnHostMapping(log *logger.Logger) infoFetcher {
 		return host, nil
 	}
 }
+
+func TestGetHostInfoReturnsSomeKeys(t *testing.T) {
+	// this is a simple test to ensure the host provicer returns the new keys
```
mauri870, line 173:
```suggestion
	// this is a simple test to ensure the host provider returns the new keys
```

internal/pkg/composable/providers/host/host_test.go
```diff
@@ -167,3 +168,20 @@ func returnHostMapping(log *logger.Logger) infoFetcher {
...
+		t.Fatal(err)
+	}
+
+	expectedKeys := []string{"os_family", "os_platform", "os_version"}
+
+	for _, key := range expectedKeys {
+		if _, exist := osInfo[key]; !exist {
+			t.Errorf("expecting key '%s' from host provider.", key)
+		}
+	}
```
mauri870, line 186: I'm not sure if this is needed, but should we also check that the values are not empty?

belimawr: That's highly dependant on the OS/Linux distro it's running on, for the few I've tested, only Arch had an empty string, but overall it's very inconsistent across different distros (some examples: https://github.com/elastic/integrations/issues/10797#issuecomment-2452212459).

internal/pkg/composable/providers/host/host_test.go
```diff
@@ -167,3 +168,20 @@ func returnHostMapping(log *logger.Logger) infoFetcher {
 		return host, nil
 	}
 }
+
+func TestGetHostInfoReturnsSomeKeys(t *testing.T) {
+	// this is a simple test to ensure the host provider returns the new keys
+	// needed to add support to Debian 12.
+	osInfo, err := getHostInfo(logp.L())()
```
AndersonQ, line 175: There is a test logger on https://github.com/elastic/elastic-agent/blob/d8125be16f8bf369738113861f958c59441bc696/pkg/core/logger/loggertest/logger.go#L20. Better to use it. Or at least the logger from `pkg/core/logger/`

belimawr:
> @belimawr could you add a changelog fragment? 🙂

Done on 2a789ba764

mauri870 (approved):

strawgate:
Per ECS -- https://www.elastic.co/guide/en/ecs/current/ecs-os.html these should be `os.family`, `os.platform`, `os.version`?

strawgate:
Oh, if these don't end up in metadata for documents then you're right it doesnt need to be ecs

Issue #5869, opened 2024-10-28
[Flaky Test]: TestHTTPSProxy – data race


mauri870:
### Failing test case

TestHTTPSProxy

### Error message

data race

### Build

https://buildkite.com/elastic/elastic-agent/builds/13456#0192d3f2-11d6-4f7e-a646-9e5c48fa8463

### OS

Mac

### Stacktrace and notes

This was introduced in https://github.com/elastic/elastic-agent/pull/5802.

```
PASS
==================
WARNING: DATA RACE
Read at 0x00c00007c863 by goroutine 80:
  testing.(*common).logDepth()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/testing/testing.go:1017 +0xcd
  testing.(*common).log()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/testing/testing.go:1010 +0xa4
  testing.(*common).Logf()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/testing/testing.go:1061 +0x6a
  testing.(*common).Logf-fm()
      <autogenerated>:1 +0x69
  github.com/elastic/elastic-agent/testing/proxytest.prepareMTLSProxyAndTargetServer.WithRequestLog.func3.1()
      /Users/admin/builds/bk-agent-prod-orka-1730132740729123083/elastic/elastic-agent/testing/proxytest/proxytest.go:100 +0xc3
  github.com/elastic/elastic-agent/testing/proxytest.New.func2()
      /Users/admin/builds/bk-agent-prod-orka-1730132740729123083/elastic/elastic-agent/testing/proxytest/proxytest.go:191 +0x75c
  net/http.HandlerFunc.ServeHTTP()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/net/http/server.go:2171 +0x47
  net/http.serverHandler.ServeHTTP()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/net/http/server.go:3142 +0x2a1
  net/http.(*conn).serve()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/net/http/server.go:2044 +0x13c4
  net/http.(*Server).Serve.gowrap3()
      /Users/admin/.gvm/versions/go1.22.8.darwin.amd64/src/net/http/server.go:3290 +0x4f

Previous write at 0x00c00007c863 by goroutine 57:
  testing.tRunner.func1()
[...]

PR #5950, opened 2024-11-06
[proxytest] Wait all requests to finish before closing the server

AndersonQ:
## What does this PR do?

Makes the proxytest to wait all requests to finish before closing the underlying HTTP server

## Why is it important?

The request log sometimes happens after the tests has finished, what causes the test to panic

## Checklist

```
go test -count 15000 -timeout 0 -run TestHTTPSProxy ./testing/proxytest/
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5869

## Questions to ask yourself

- ...

### DIFF
--- a/testing/proxytest/proxytest.go
+++ b/testing/proxytest/proxytest.go
@@ -251,9 +251,11 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func (p *Proxy) Close() {
-	p.requestsWG.Wait()
-
 	p.Server.Close()
+
+	// it's necessary as sometimes the request log happens after the test
+	// finishes. See https://github.com/elastic/elastic-agent/issues/5869
+	p.requestsWG.Wait()
 }
 
 func (p *Proxy) serveHTTP(w http.ResponseWriter, r *http.Request) {
--- a/testing/proxytest/proxytest.go
+++ b/testing/proxytest/proxytest.go
@@ -38,6 +38,7 @@ type Proxy struct {
 	// proxiedRequests is a "request log" for every request the proxy receives.
 	proxiedRequests   []string
 	proxiedRequestsMu sync.Mutex
+	requestsWG        *sync.WaitGroup
 
 	opts options
 	log  *slog.Logger
@@ -164,8 +165,9 @@ func New(t *testing.T, optns ...Option) *Proxy {
 		lv = slog.LevelDebug
 	}
 	p := Proxy{
-		opts:   opts,
-		client: opts.client,
+		requestsWG: &sync.WaitGroup{},
+		opts:       opts,
+		client:     opts.client,
 		log: slog.New(slog.NewTextHandler(logfWriter(opts.logFn), &slog.HandlerOptions{
 			Level: lv,
 		})),
@@ -176,6 +178,9 @@ func New(t *testing.T, optns ...Option) *Proxy {
 
 	p.Server = httptest.NewUnstartedServer(
 		http.HandlerFunc(func(ww http.ResponseWriter, r *http.Request) {
+			p.requestsWG.Add(1)
+			defer p.requestsWG.Done()
+
 			w := &proxyResponseWriter{w: ww}
 
 			requestID := uuid.Must(uuid.NewV4()).String()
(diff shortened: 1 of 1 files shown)


testing/proxytest/proxytest.go
```diff
@@ -176,6 +178,9 @@ func New(t *testing.T, optns ...Option) *Proxy {
 
 	p.Server = httptest.NewUnstartedServer(
 		http.HandlerFunc(func(ww http.ResponseWriter, r *http.Request) {
+			p.requestsWG.Add(1)
```
pkoutsovasilis, line 181: just a maybe stupid question, but my understanding of it is that we increment the waitgroup on each request. Don't we know how many request we are gonna perform on this test so we can add them beforehand and leave only `requestsWG.Done()` inside the handler func?

AndersonQ: no, we don't. The proxy is used on tests and each test might vary, or even not be possible to know for sure how many requests will go through the proxy

pkoutsovasilis: just thinking out loud here; 🙂  how do we cover for the scenario where there are still incoming requests but for some reason one of them manages to call Done() on the syncgroup and cause the Wait() in Close() to exit while other incoming ones haven't executed the Add() on the syncgroup yet?!

AndersonQ: I updated it to close the server first.

pkoutsovasilis: moving the close before the wait is effectively the same code as before though right?! I mean from the description of the issue "Makes the proxytest to wait all requests to finish before closing the underlying HTTP server" I understand that the reason behind the issue is that we call `p.Server.Close()` before all requests have been done, right?!

AndersonQ: no, the point is that `p.Server.Close()` should wait all requests to finish, then return. 
The last request log should happen before the server closes, but, as we see in the test failing, some how it isn't the case. 
That's why I added another barrier to try to force the request log to be logged before the test ends.

pkoutsovasilis: I still don't see how `p.Server.Close()` waits for all requests to finish especially now that it is before `p.requestsWG.Wait()`; `p.Close()` will wait probably for some requests to finish until a `syncgroup.Done()` inside the HandlerFunc manages to lower the syncgroup enough to cause the `p.requestsWG.Wait()` to exit. 🙂

AndersonQ: That's right, the wait has to come before.
The main idea here is to try to delay a bit the end of the test, just enough for the last request log to be logged before the test end.

That's why it isn't a huge problem if a request slips over the wait group.

pkoutsovasilis: I am far from an expert in this particular piece of code, but the case that we have here, reminds me a lot some other rare cases where we need close a channel on the receivers end without knowing if the writers are done writing to it. Usually we treat such cases with first shutting down the writers (here should be the components that write to this proxy) then, if we have no shutdown confirmation from the writers, we wait as much as we think the writers should take to shutdown (not ideal) and then we close the channel (the proxy server in this case). Would a similar approach work here?! could we at each test "shutdown" the components that make requests to the proxy first and then close the proxy server?

pchila (commented):
Since we are only interested in making sure that the server completes any outstanding requests before we complete the test I can see multiple options:

- create a `serverch chan struct{}`  and close it after [server.Close()](https://pkg.go.dev/net/http/httptest#Server.Close) has completed while blocking with a `<-serverch` so that we know that the shutdown has completed
- uninstall the agent in each testcase and then close the server: in that case the only client of the server has disappeared so closing it should be safe (even if some synchronization would be better
-  close the server in the test (blocking) and set it to nil *and* try to close it also in a `defer/Cleanup()` if not nil (only for tests that will panic/not complete for some reason)
- have each testcase start and stop its own proxy server(s) in a synchronized manner to reduce the lifetime of the servers and let each test decide when to stop it

pkoutsovasilis:
> * have each testcase start and stop its own proxy server(s) in a synchronized manner to reduce the lifetime of the servers and let each test decide when to stop it

Thanks for chiming in @pchila 🙂  @AndersonQ do you want to consider if any of the proposals here is easily applicable without causing heavy changes?

AndersonQ:
as this is a flaky test, I'd rather merge it and see if it fixes the issue. If don't we might move for another approach or even remove the request log at all

AndersonQ:
so far CI has been pretty good at triggering it, so we should know soon if this fixes the issue or not

pkoutsovasilis (approved):
I saw that the unit-tests in CI are back to green with this PR, so 👍 That said, in the future we should try to implement a different solution to prevent in full consciousness the same hiccup from happening

PR #5988, opened 2024-11-11
Add support for pre existing Active Directory user

### LINKED ISSUE #4585: Allow using pre-existing user and group when installing unprivileged agents
Quoting https://github.com/elastic/elastic-agent/issues/3868#issuecomment-2053871719 for a Windows use case for this:

> For example, if I have a service account in active dieectory that works on the system already, can I use that or is elastic creating an account and group locally?
> 
> The use case is integrations that need network access from specific service accounts in active directory to be the service context for filebeat (api, file shares, etc).
> 
> Today system is the only option but this gives me hope for 1 non system which is great, but having custom users would be excellent.
> 
> In the past you could simply install filebeat or winglogbeat with the service account you needed as a service and it would use that user context when it runs.

michalpristas:
Waiting on custom windows image for testing with AD e2e.

New flags are introduced 
- user
- group
- password (windows only)

These flags are taken into account only when `--unprivileged` is used.

New user is added same permissions as elastic-agent user when created in order to be able to log on as a service (otherwise agent won't start) 

Custom user won't be created and needs to be present

##### Run script 1 that will prepare Active Directory

Script will also wait for AD services to start

```powershell
$domainname = 'testing.local'
$addsSecurePass = ConvertTo-SecureString 'Changeme123+' -AsPlainText -Force

# install ad domain services
Install-WindowsFeature -name AD-Domain-Services -IncludeManagementTools


# setup pass for administrator to conform to forest requirements
net user Administrator 'Changeme123+' /passwordreq:yes

# add a new forest testing.com
Install-ADDSForest -DomainName $domainname -InstallDNS -ErrorAction Stop -NoRebootOnCompletion -SafeModeAdministratorPassword $addsSecurePass -Confirm:$false 

# reboot needs to be performed so changes are effective
```
##### Run script that will prepare user

```powershell
$domainname = 'testing.local'
$addsSecurePass = ConvertTo-SecureString 'Changeme123+' -AsPlainText -Force

# wait for AD to boot up
$domainFound = $true

$newObjectParameters = @{
    TypeName = 'System.DirectoryServices.ActiveDirectory.DirectoryContext'
    ArgumentList = @(
        "Domain"
        $domainname
    )
}

$DirectoryContext = New-Object @newObjectParameters

$startDate = Get-Date
do {
    try
    {
        [System.DirectoryServices.ActiveDirectory.DomainController]::FindOne($DirectoryContext)
        $domainFound = $true
    }
    catch
    {
        $domainFound = $false
        Start-Sleep -s 2
    }
} until ($domainFound -and $startDate.AddMinutes(5) -gt (Get-Date))


# add user
New-ADUser -Name 'TestingUser' -AccountPassword $addsSecurePass -Enabled $true -ChangePasswordAtLogon $false -PasswordNeverExpires $true
```
#### Scenario 1
- install as unpriviledged `elastic-agent install --unprivileged  --user="testing.local\TestUser" --password=Changeme123+`
- check permissions are set to user and service is running as this user

#### Scenario 2
- install as admin
- switch to unpriviledged `elastic-agent unprivileged --user="testing.local\TestUser" --password=Changeme123+`
- check permissions are set to user and service is running as this user

Resolves #4585


internal/pkg/agent/cmd/install.go
```diff
@@ -28,6 +29,10 @@ const (
 	flagInstallDevelopment            = "develop"
 	flagInstallNamespace              = "namespace"
 	flagInstallRunUninstallFromBinary = "run-uninstall-from-binary"
+
+	flagInstallADUser  = "ad-user"
+	flagInstallADGroup = "ad-group"
+	flagInstallADPass  = "ad-password"
```
blakerouse, line 35: Why make this a specific Windows feature? Wouldn't it be better to just allow `user` and `group` for all platforms? `password` is only needed for Windows.

michalpristas: i agree, let's see if i can squeeze this in as this is planned for 8.17. if not i can do it iteratively

michalpristas: added support for mac and linux, tested on both including switch from privileged to unprivileged. 
tests TBD still

internal/pkg/agent/cmd/install.go
```diff
@@ -61,6 +66,13 @@ would like the Agent to operate.
 	cmd.Flags().Bool(flagInstallDevelopment, false, "Install into a standardized development namespace, may enable development specific options. Allows multiple Elastic Agents to be installed at once. (experimental)")
 	_ = cmd.Flags().MarkHidden(flagInstallDevelopment) // For internal use only.
 
+	// Active directory user specification
+	cmd.Flags().String(flagInstallCustomUser, "", "Custom user used to run Elastic Agent")
+	cmd.Flags().String(flagInstallCustomGroup, "", "Custom group used to access Elastic Agent files")
+	if runtime.GOOS == "windows" {
+		cmd.Flags().String(flagInstallCustomPass, "", "Password for Active directory user used to run Elastic Agent")
```
blakerouse, line 73: On Windows a password is required for a user event if its a custom user that is not in Active Directory.

```suggestion
		cmd.Flags().String(flagInstallCustomPass, "", "Password for user used to run Elastic Agent")
```

internal/pkg/agent/cmd/install.go
```diff
@@ -249,7 +261,19 @@ func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
...
-		ownership, err = install.Install(cfgFile, topPath, unprivileged, log, progBar, streams)
+		customUser, _ := cmd.Flags().GetString(flagInstallCustomUser)
+		customGroup, _ := cmd.Flags().GetString(flagInstallCustomGroup)
+		customPass := ""
+		if runtime.GOOS == "windows" {
+			customPass, _ = cmd.Flags().GetString(flagInstallCustomPass)
+
+			if (customUser != "" || customPass != "") &&
+				(customUser == "" || customPass == "") {
+				return fmt.Errorf("error installing package: all Active Directory parameters must be provided")
```
blakerouse, line 272: Why? I think this could work by just providing say the user, then still allow the group to be created.

Or maybe you want to use the default user and password, but you want to just set the password with `--password`?

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -44,6 +45,13 @@ unprivileged it will still perform all the same work, including stopping and sta
 	cmd.Flags().BoolP("force", "f", false, "Do not prompt for confirmation")
 	cmd.Flags().DurationP("daemon-timeout", "", 0, "Timeout waiting for Elastic Agent daemon restart after the change is applied (-1 = no wait)")
 
+	// Custom  user specification
+	cmd.Flags().String(flagInstallCustomUser, "", "Custom user used to run Elastic Agent")
+	cmd.Flags().String(flagInstallCustomGroup, "", "Custom group used to access Elastic Agent files")
+	if runtime.GOOS == "windows" {
+		cmd.Flags().String(flagInstallCustomPass, "", "Password for Active directory user used to run Elastic Agent")
```
blakerouse, line 52: On Windows a password is required for a user event if its a custom user that is not in Active Directory.

```suggestion
		cmd.Flags().String(flagInstallCustomPass, "", "Password for user used to run Elastic Agent")
```

internal/pkg/agent/cmd/unprivileged.go
```diff
@@ -56,6 +64,18 @@ func unprivilegedCmd(streams *cli.IOStreams, cmd *cobra.Command) (err error) {
...
 
+	customUser, _ := cmd.Flags().GetString(flagInstallCustomUser)
+	customGroup, _ := cmd.Flags().GetString(flagInstallCustomGroup)
+	customPass := ""
+	if runtime.GOOS == "windows" {
+		customPass, _ = cmd.Flags().GetString(flagInstallCustomPass)
+
+		if (customUser != "" || customPass != "") &&
+			(customUser == "" || customPass == "") {
+			return fmt.Errorf("error installing package: all Active Directory parameters must be provided")
```
blakerouse, line 75: Same as above.

internal/pkg/agent/install/install_windows.go
```diff
@@ -55,12 +55,17 @@ func fixInstallMarkerPermissions(markerFilePath string, ownership utils.FileOwne
...
-func withServiceOptions(username string, groupName string) ([]serviceOpt, error) {
+func withServiceOptions(username string, groupName string, password string) ([]serviceOpt, error) {
 	if username == "" {
 		// not installed with --unprivileged; nothing to do
 		return []serviceOpt{}, nil
 	}
 
+	if password != "" {
+		// existing user
+		return []serviceOpt{withUserGroup(username, groupName), withPassword(password)}, nil
```
blakerouse, line 66: At this point the `username` must either have `.\` at the start or the DOMAIN NAME like `AD\username-in-ad`. You are going to need to handle username having `AD\` or not having `.\` in the name. Otherwise the service will not start.

internal/pkg/agent/cmd/install.go
michalpristas: Yeah but we need to be vocal about that and raise a warning that we are adding user somewhere or creating group that existing user will be part of

internal/pkg/agent/install/install_windows.go
```diff
@@ -109,3 +121,17 @@ func serviceConfigure(ownership utils.FileOwner) error {
...
 	return nil
 }
+
+func isWindowsDomainUsername(username string) (bool, error) {
+	if !strings.Contains(username, `\`) {
+		// fail fast
+		return false, nil
+	}
+
+	match, err := regexp.MatchString(`^.*(\\)(.*)$`, username)
```
andrzej-stencel, line 131: Isn't this regex match equivalent to the `strings.Contains`?

michalpristas: yeah basically. but i'd rather change regexp to be more strict and not allow multiple `\`s 
need to check windows reqs for domain and user names

blakerouse (approved):
This looks good. Thanks for the fixes on checking the domain name in the user.

changelog/fragments/1731314919-Added-support-for-custom-Active-Directory-user-for-unprivileged-mode.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Added support for pre-existing Active Directory user for unprivileged mode
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: User can specify custom pre-existing user for running unprivileged mode. This user will be gived permissions to log on as a service.
```
strawgate, line 19: > This user will be **given** permissions to log on as a service

internal/pkg/agent/install/install_windows.go
```diff
@@ -21,6 +22,13 @@ import (
 	"github.com/elastic/elastic-agent/version"
 )
 
+const (
+	// Conforming to https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/naming-conventions-for-computer-domain-site-ou#domain-names
+	// domain names can contain all alphanumeric characters except for the extended characters that appear in the Disallowed characters list. Names can contain a period, but names can't start with a period.
+	// Disallowed characters: [, ~ : @ # $ % ^ ' . ( ) { } _ {whitespace} \ / ]
+	activeDirectoryUsername = `^[A-Za-z0-9]+(?:\.[A-Za-z0-9]+)*\\[A-Za-z0-9.-]{1,104}$`
```
cmacknz, line 29: Why do we need to validate this instead of letting Windows tell us the characters are invalid when we try to install?

internal/pkg/agent/install/install_windows.go
```diff
@@ -55,12 +63,23 @@ func fixInstallMarkerPermissions(markerFilePath string, ownership utils.FileOwne
...
 	if username == "" {
 		// not installed with --unprivileged; nothing to do
 		return []serviceOpt{}, nil
 	}
 
+	if password != "" {
+		if isFullDomainName, err := isWindowsDomainUsername(username); err != nil {
+			return nil, fmt.Errorf("failed to parse username: %w", err)
+		} else if !isFullDomainName {
+			return nil, fmt.Errorf("username is not in proper format 'domain\\username'")
```
cmacknz, line 76: This error message could be clearer, one of the failure reasons is an illegal character and that isn't listed here.

internal/pkg/agent/install/install_windows_test.go
```diff
@@ -0,0 +1,36 @@
...
+		{`user`, false},
+		{`domain\user`, true},
+		{`domain/user`, false},
+		{`domain/domain\user`, false},
+		{`domain\subdomain\user`, false},
+		{`dom,ain\user`, false},
+		{`doma~in\user`, false},
+		{`domai:n\user`, false},
+		{`dom.ain\user`, true},
+		{`dom.ain\.user`, true},
```
cmacknz, line 28: Include a test using a non-english character or accent, é for example

testing/integration/install_test.go
```diff
@@ -250,6 +258,50 @@ func TestInstallPrivilegedWithBasePath(t *testing.T) {
...
+	}
+
+	// Check that Agent was installed in default base path
+	topPath := installtest.DefaultTopPath()
+	checks := &installtest.CheckOpts{
+		Privileged: opts.Privileged,
+		Username:   customUsername,
+		Group:      customGroup,
+	}
+	require.NoError(t, installtest.CheckSuccess(ctx, fixture, topPath, checks))
```
cmacknz, line 285: It looks to me like the test for Windows is slightly weaker than the test for Linux here, because Windows does not try to execute the Elastic Agent status command as the relevant user, because that requires more than just a single command on Windows. 

Is there a way for us to do this? The test right now has us using a privileged user to create a less privileged user, but doesn't seem to try to interact with the agent as that less privileged user. 

* Windows just checks the permissions SIDs https://github.com/elastic/elastic-agent/blob/434986ed9fa6fbed6257557a4c6773d519742f66/testing/installtest/checks_windows.go#L55-L60
* Linux checks the permissions, UID+GID, and makes sure the commands work from the perspective of the relevant user https://github.com/elastic/elastic-agent/blob/434986ed9fa6fbed6257557a4c6773d519742f66/testing/installtest/checks_unix.go#L25-L34

cmacknz (commented):
When this merges we will need to document the feature, make sure there is a follow up issue in ingest-docs for this.

Along the lines of documentation:
1. We need to document if we store the password. It doesn't look like we do but we will be asked about this regardless.
2. How would a user rotate the password of a custom Elastic Agent user? If we don't support this we need to document it, and probably we should pre-emptively create an issue to make it possible.

ycombinator:
@michalpristas 
Thanks for writing out the test steps in the PR description.  I've added the `QA:Ready for Testing` label for @amolnater-qasource and team to test this feature.

As mentioned in https://github.com/elastic/elastic-agent/issues/4585#issuecomment-2501659147, please:
* [ ] get this PR merged without the AD integration test
* [ ] create a follow up issue for adding the AD integration test (with a reverse link from the private repo PR for the Windows VM image with AD)
* [ ] work with @kilfoyle on documentation around this feature, marking the AD part as Technical Preview

kilfoyle:
Thanks for the summary @ycombinator!
@michalpristas, [here's a PR](https://github.com/elastic/ingest-docs/pull/1505) for the documentation. Please let me know whatever I've missed.

testing/integration/install_test.go
blakerouse: > Is there a way for us to do this? The test right now has us using a privileged user to create a less privileged user, but doesn't seem to try to interact with the agent as that less privileged user.

It is not possible to execute as the `elastic-agent-user` because the user is created without the ability to login or get a real execution shell. The only way to actually test this would be to create another user, add that user to the `elastic-agent` group and then run the command as that user.

internal/pkg/agent/install/install_windows.go
michalpristas: because I find it good practice to validate on input level rather than let system find failures for you

cmacknz: Yes, with the trade off that we have to encode the rules of the system into our code, and the two things can get out of sync. Probably this one is OK, but if the error message from the system were reasonably clear, this is some code we wouldn't need to maintain.

ycombinator:
Not sure why SonarQube is reporting 0% coverage when there is a unit test added in this PR.  Perhaps it's because it's a Windows-only test file? 

The remaining CI checks are green. Force merging as we're trying to get this feature into `8.17.0`.

amolnater-qasource:
Hi @ycombinator 

We have revalidated this on latest 8.17.0 BC6 kibana cloud environment and had below observations:

**Observations:**
**Scenario 1:**
- Install agent as unpriviledged: `elastic-agent install --unprivileged  --user="testing.local\TestUser" --password=Changeme123+`
- Checked permissions are set to TestUser and service is running as TestUser.



**Scenario 2:**
- Installed agent as admin.
- Then switched to unpriviledged: `elastic-agent unprivileged --user="testing.local\TestUser" --password=Changeme123+`
- Checked permissions are set to TestUser and service is running as TestUser.
image
image





**Logs:**
[elastic-agent-diagnostics-2024-12-11T08-50-00Z-00.zip](https://github.com/user-attachments/files/18092180/elastic-agent-diagnostics-2024-12-11T08-50-00Z-00.zip)

**Build details:**
VERSION: 8.17.0 BC6
BUILD: 80521
COMMIT: e8a820624a03a412433584d3e3df951838e4c63c
Artifact Link: https://staging.elastic.co/8.17.0-6b31e673/downloads/beats/elastic-agent/elastic-agent-8.17.0-windows-x86_64.zip

Hence we are  marking this ticket as **QA:Validated**.

Please let us know if we are missing anything here.

cc: @michalpristas 
Thanks!

PR #6002, opened 2024-11-12
[helm] add support for creating services for agent presets

pkoutsovasilis:
## What does this PR do?

This PR introduces support for automatically creating Kubernetes services for specific Elastic Agent presets through the Helm chart. It adds a structured and scalable mechanism for managing and exposing agent instances within the Kubernetes cluster. This enhancement allows users to define service configurations (e.g., type, ports) for push-based integrations like NetFlow. 

Key changes:
1. Added logic in the Helm chart to generate services for agent presets when specified in `values.yaml`.
2. Introduced configuration examples for creating a `NodePort` service linked to a NetFlow agent preset.
3. Updated schema definitions in `values.schema.json` to support new service and port configuration options.

## Why is it important?

This enhancement simplifies the deployment and operational management of push-based integrations by allowing users to define and scale agent services directly through Helm configurations. By enabling Kubernetes service generation, this change:
- Streamlines access to agent instances.
- Supports various service types (e.g., `ClusterIP`, `NodePort`, `LoadBalancer`) to accommodate diverse user needs.

## Checklist

## Disruptive User Impact

This PR does not introduce disruptive changes, as service creation is optional and requires explicit configuration by users in `values.yaml`. By default, no services will be created unless configured.

Follow the [example](https://github.com/elastic/elastic-agent/tree/ccb05817e856be3c2d30c74ff144b801afadb325/deploy/helm/elastic-agent/examples/netflow-service) introduced in this PR

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6012

### DIFF
--- a/deploy/helm/elastic-agent/examples/netflow-service/README.md
+++ b/deploy/helm/elastic-agent/examples/netflow-service/README.md
@@ -1,0 +1,28 @@
+# Example: Netflow Custom Integration
+
+In this example we define a `netflow` custom integration alongside a custom agent preset defined in [agent-netflow-values.yaml](agent-netflow-values.yaml). Also, we disable all `kubernetes` related providers and creation of cluster role and service account, as they are not required for this example.
+
+## Prerequisites:
+1. A k8s secret that contains the connection details to an Elasticsearch cluster such as the URL and the API key ([Kibana - Creating API Keys](https://www.elastic.co/guide/en/kibana/current/api-keys.html)):
+    ```console
+    kubectl create secret generic es-api-secret \
+       --from-literal=api_key=... \
+       --from-literal=url=...
+    ```
+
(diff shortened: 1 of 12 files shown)


swiatekm (approved):
Looks good to me.

Something that jumps out at me in this PR is that we're making essentially the same change in the ECK and K8s templates. Could we unify that?

pkoutsovasilis:
This is true there is enough overlapping between the two, but there are enough small differences that make me wonder if merging them is gonna add value; namely deployment, statefulset, daemonset are ECK CRDs, the agent config (secret) is referenced in the CRD and all the mounts of that are handled by the ECK operator, thus, the pod template has a different flow. But sure in the future we could investigate merging them, does this make sense? 🙂

Issue #5991, opened 2024-11-11
Memory Consumption Issue with Elastic Agent on Kubernetes with high number of resources


rhr323:
### Issue Summary

In our testing on the serverless platform, we aimed to assess the maximum number of projects that can be supported on a single MKI cluster. We were using the Elastic Agent version 8.15-4-SNAPSHOT to mitigate previously identified memory issues.

Most Elastic Agent instances functioned without issues. However, on nodes hosting vector search projects, where a larger number of Elasticsearch instances and their associated Kubernetes resources (e.g., pods, deployments, services, secrets) are allocated, we observed the Elastic Agent running out of memory. This typically occurred when these nodes were hosting around 100 Elasticsearch instances.

### Observed Behavior

- Elastic Agent on high-density nodes (around 100 Elasticsearch instances) experienced memory exhaustion and got stuck in a crash loop.
- Diagnostic data was collected from an Elastic Agent on a node with ~70 allocated projects at the time of capture.

### Environment

- Elastic Agent version: 8.15-4-SNAPSHOT
- Kubernetes environment: Serverless platform, MKI cluster
- Node allocation: ~100 Elasticsearch instances per node for vector search projects

swiatekm:
Having looked at the diagnostics and telemetry for one of the agent Pods encountering this problem, my initial thoughts are as follows:

1. The metricbeat elasticsearch module itself is, unsurprisingly, using a good chunk of memory. We can probably optimize that.
2. There's a lot of configuration churn on the elastic-agent side caused by the kubernetes variable provider, and that's most likely what leads to the agent itself using more memory than expected. I'm tackling that in #5835 .
3. In beats themselves, needing to reload configuration frequently also adds up to additional memory consumption and possibly other kinds of disruption (If we have a scraper that is supposed to fetch metrics every 10 seconds, but we reload config every 5 seconds, then we're effectively scraping every 5 seconds instead). If the beats config manager could avoid restarting units it doesn't need to restart, this effect could be mitigated.

swiatekm:
@elastic/stack-monitoring do we have a synthetic benchmark for the elasticsearch metricbeat module? It would help a lot, as reproducing the environment this issue came up in is a huge pain.

consulthys:
@swiatekm Honestly not to my knowledge, but since we've taken over SM very recently, maybe @miltonhultgren has a more insightful answer to give you.

miltonhultgren:
👋🏼 
[...]

PR #6011, opened 2024-11-13
Emit Pod data only for running Pods in the Kubernetes provider

swiatekm:
## What does this PR do?

The kubernetes provider emits data each time a Pod gets updated, even if the Pod is not yet running. As a result, every time a new Pod is spawned, we get multiple updates as it is created, scheduled, containers are created, and so on. Instead, emit the data only if the Pod is actually running.

## Why is it important?

Configuration reloading can be quite expensive when there are a lot of Pods on the Node. We should avoid doing so unnecessarily. This change should help #5835 and #5991.

Note that in principle this exposes us to a new failure mode. It's possible for the Pod to successfully finish running and be removed before we push the new configuration to beats. This was much less likely when we included the Pod when it was scheduled, but still possible with our 100ms debounce timer in the coordinator. I think the tradeoff is worth it, considering the issues with config reloading on large Nodes.

## Checklist

Deploy the agent in a local kind cluster using either the default manifests or the Helm Chart.

## Related issues

- Relates #5835
- Relates #5991

### DIFF
--- a/changelog/fragments/1731500837-pod-status-updates.yaml
+++ b/changelog/fragments/1731500837-pod-status-updates.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Emit Pod data only for running Pods in the Kubernetes provider
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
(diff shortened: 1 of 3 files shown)


pkoutsovasilis (approved):
This change makes sense to me; LGTM

gizas:
@swiatekm thanks for this. This makes sense.

I want to make sure some corner cases like:
- pods created from cronjobs
- pods that are restarting (eg. OOM or Imagepull error) how this will behave in such cases?

blakerouse:
> > @swiatekm thanks for this. This makes sense.
> > I want to make sure some corner cases like:
> > ```
> > * pods created from cronjobs
> > ```
> 
> There isn't really any difference. It's possible for the Pod to be finished running before we update the configuration, but this is possible right now too. I'm not sure if this is a problem either way - scrapers probably wouldn't get any metrics, and container logs should remain for filelog to read until the Pod is deleted.

We actually delay the removal of a pod after it is stopped to ensure that it exists for a period of time. This is needed for reading the log files for a reason (think nginx access logs from inside of the container). I think this should handle the case where the status is basically greater than > Running. As in crashed, stopping, stopped, etc.

swiatekm:
For the record, I considered other approaches to this problem, but trying to debounce the updates in a more clever way made things more difficult to reason about and ultimately not much better than the current solution.

blakerouse (changes requested):

gizas (approved):

PR #6028, opened 2024-11-14
Add var generation benchmark

swiatekm:
## What does this PR do?

Add a benchmark for variable generation from provider mappings. We specifically profile having a lot of Pods from the Kubernetes provider, which is something that has recently become a problem, and which the configuration generation pipeline doesn't deal with very well. I used test data from a real environment where the problems outlined in #5991 and #5835 could be reproduced.

## Why is it important?

I'd like to make some improvements to both the var generation and the configuration generation in the coordinator, and want a straightforward way of demonstrating their impact.

## Checklist

## Related issues

- Related #5835
- Relates #5991


swiatekm:
> Looks good. I imagine we are going to track the results of this benchmark in PRs aimed to improve k8s variable performance but do we have plans to track changes to the benchmark results for the long term ? Are we going to rely on CI to keep track of these results ?

My impression is that this code changes infrequently enough that this wouldn't give us much benefit vs the work necessary, unless we already have a benchmark result tracking solution that would be easy to use. I was mostly hoping that whenever someone makes a change to var generation, they could run this manually.

swiatekm:
I thought I could avoid backporting this, but my other changes in #6058 and #6114 depend on it, so I'll have to do it.

PR #6038, opened 2024-11-14
enhancement(4889): added check to abort enrolling if user is privileg…

### LINKED ISSUE #4889: Actionable error message when attempting to `enroll` an unprivileged Agent as a privileged user
**Version: 8.14.0**
**Operating System: Ubuntu 24.04 LTS**
**Platform: arm64**

While working on [this issue](https://github.com/elastic/elastic-agent/issues/4705) comparing root and unprivileged elastic agents, I encountered an error with the `enroll` command when using unprivileged agent.

**Steps to Reproduce:**
1. Deploy ESS v8.14.0
2. Create agent policy with system integration
3. Install fleet managed agent with the unprivileged flag `sudo ./elastic-agent install --unprivileged --url=<fleet url> --enrollment-token=<token>` 
4. Unenroll the agent 
5. Enroll the agent `sudo elastic-agent enroll --url=<fleet url> --enrollment-token=<token>`
6. Agent enrollment gets confirmed, but incoming data does not get confirmed
7. Agent status in fleet ui is stuck in `updating`
8. Running `sudo elastic-agent status` we get the following error

```
Error: failed to communicate with Elastic Agent daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /opt/Elastic/Agent/elastic-agent.sock: connect: connection refused"
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.14/fleet-troubleshooting.html
```

## Definition of Done

_Based on discussion in the comments:_

* [ ] When a user `install`s an Agent using `--unprivileged`, then tries to `enroll` this Agent with a privileged user, an error message with the suggestions in https://github.com/elastic/elastic-agent/issues/4889#issuecomment-2158635347 is shown to the user.

kaanyalti:
- Enhancement

## What does this PR do?

This PR updates the enroll command so that it provides a more descriptive error message to the user in the case a root user tries to enroll an unprivileged agent.

## Why is it important?

Provides users with clearer information.

## Checklist

- Create an agent policy, install and enroll and agent in unprivileged mode
- Unenroll the agent
- Try to enroll the agent with root user, validate the error message

## Related issues
- Closes #4889

### DIFF
--- a/testing/integration/escaping_secrets_in_policy_test.go
+++ b/testing/integration/escaping_secrets_in_policy_test.go
@@ -30,6 +30,7 @@ func TestEscapingSecretsInPolicy(t *testing.T) {
 		Stack: &define.Stack{},
 		Sudo:  true,
 	})
+	t.Skip("flaky test: https://github.com/elastic/elastic-agent/issues/6107")
 	ctx := context.Background()
 	fixture, err := define.NewFixtureFromLocalBuild(t, define.Version())
 	require.NoError(t, err)


internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -328,26 +360,21 @@ func (c *enrollCmd) writeDelayEnroll(streams *cli.IOStreams) error {
...
-			}
-		} else {
-			agentRunning = false
+	if !c.options.FleetServer.SpawnAgent {
+		// wait longer to try and communicate with the Elastic Agent
+		if agentConErr != nil {
+			return "", errors.New("failed to communicate with elastic-agent daemon; is elastic-agent running?")
 		}
+	} else {
+		agentRunning = false
```
michalpristas, line 374: i think this does not really keep the logic. 
we should check probably
```go
if agentConnErr != nil && !c.options.FleetServer.SpawnAgent {
    return "", errors.New("failed to communicate with elastic-agent daemon; is elastic-agent running?")
} else {
    agentRunning = agentConnErr == nil
}
```

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -195,11 +216,22 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
...
+	c.log.Debug("verifying communication with running Elastic Agent daemon")
+	// attempt to get the agent state, if there is an error, proceed without
+	// handling the error and provide the fleetServerBootstrap with the error
+	as, asErr := c.getAgentStateWB(ctx)
+	if asErr == nil {
+		// skip this check if enroll is run by the install command
+		if hasRoot && as.Info.Unprivileged && !c.options.FromInstall {
+			return unprivilegedAgentRootUserError
+		}
+	}
```
blakerouse, line 233: I did not see the original filled issue for this and I would have made a comment there. I know this is coming late, but I believe this is the wrong approach.

If you call `enroll` as root, you can easily tell if Elastic Agent is installed without communicating with the daemon. You can read the user/group from the executing binary from the filesystem, because you are root. 

You can then just `ReExec` the enroll command directly using the `user` you retrieved from reading from the disk. You need to read `user` from the filesystem as @michalpristas is adding the ability to use a custom user, so you cannot rely on the idea that the user is always the same.

This change also will allow a root user to call enroll and have the enroll work correctly as the unprivileged user.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -174,16 +178,33 @@ func newEnrollCmd(
...
 		options:          options,
 		configStore:      store,
 		configPath:       configPath,
 		daemonReloadFunc: daemonReload,
+		getAgentStateWB:  agentStateBackoffFunc(getDaemonState, waitForAgentState, daemonTimeout),
+		hasRoot:          hr,
 	}, nil
 }
 
+var unprivilegedAgentRootUserError = errors.New("cannot execute this command as root, agent is unprivileged, use \"sudo -u elastic-agent-user enroll\"")
```
blakerouse, line 196: See below.

But if you don't change the way this works then `sudo -u elastic-agent-user` cannot be hard-coded because the username will be able to be customized. @michalpristas Is adding that ability.

This also doesn't work on Windows, you would need to provide an error message that world work on Windows.

blakerouse (changes requested):

internal/pkg/agent/cmd/enroll_cmd.go
kaanyalti: yes, you're right, updating right away

internal/pkg/agent/cmd/enroll_cmd.go
kaanyalti: Yes it would, and I am now thinking that the current implementation is tricky to get working with delayed enrolls. I started looking into your suggestion and will be replacing this PR.

kaanyalti:
Closed in favor of #6144

PR #6058, opened 2024-11-18
Don't deepcopy AST when generating vars

swiatekm:
## What does this PR do?

We generate vars from mappings from both context providers and dynamic providers. Dynamic providers can produce multiple var entries - the Kubernetes provider will do so for each Pod, for example. When that happens, we currently make a deep copy of the context provider mapping, and add the dynamic provider mapping as a key.

As we don't modify the mappings, this deep copy can be replaced by a shallow copy. This PR introduces shallow copying to the AST and uses it when generating vars, leading to a major performance improvement.

## Why is it important?

All these deep copies can be quite expensive when there's a lot of mappings from the dynamic provider. See the benchstat report below:

```
goos: linux
goarch: amd64
pkg: github.com/elastic/elastic-agent/internal/pkg/composable
                       │ bench_main.txt │          bench_shallow.txt          │
GenerateVars100Pods-20    15247.8µ ± 4%   778.8µ ± 7%  -94.89% (p=0.000 n=10)

                       │ bench_main.txt │          bench_shallow.txt           │
GenerateVars100Pods-20   10273.4Ki ± 0%   779.1Ki ± 0%  -92.42% (p=0.000 n=10)

                       │ bench_main.txt │          bench_shallow.txt          │
GenerateVars100Pods-20     201.28k ± 0%   22.13k ± 0%  -89.00% (p=0.000 n=10)
```

## Checklist

## Related issues

- Relates #5835
- Relates #5991

### DIFF
--- a/internal/pkg/agent/transpiler/ast.go
+++ b/internal/pkg/agent/transpiler/ast.go
@@ -52,6 +52,9 @@ type Node interface {
 	//Close clones the current node.
 	Clone() Node
 
+	// ShallowClone makes a shallow clone of the node.
+	ShallowClone() Node
+
 	// Hash compute a sha256 hash of the current node and recursively call any children.
 	Hash() []byte
 
@@ -132,6 +135,19 @@ func (d *Dict) Clone() Node {
 			continue
 		}
 		nodes = append(nodes, i.Clone())
+
+	}
+	return &Dict{value: nodes}
+}
+
+// ShallowClone makes a shallow clone of the node.
+func (d *Dict) ShallowClone() Node {
+	nodes := make([]Node, 0, len(d.value))
+	for _, i := range d.value {
+		if i == nil {
+			continue
+		}
+		nodes = append(nodes, i)
 
 	}
 	return &Dict{value: nodes}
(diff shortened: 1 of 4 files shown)


andrzej-stencel (approved):

blakerouse (approved):
Really nice! Looks good.

michalpristas (approved):

amitkanfer:
🚀

pierrehilbert:
@swiatekm I think we want to backport this to 8.17 also. WDYT?

amitkanfer:
> @swiatekm I think we want to backport this to 8.17 also. WDYT?

Are we 100% sure there is no hidden bug here?

pierrehilbert:
My two cents: 
- We don't have the first BC for 8.17 yet
- We are backporting this to 8.15 and 8.16 so I hope we don't have an hidden bug here as the timeframe to test for those will be shorter than the 8.17 branch.

swiatekm:
If anything, maybe we should skip the 8.15 backport? It's pretty close, and this isn't that critical of a fix. For 8.17, I think we should backport either way, but I could wait until 8.17.0 is out before merging the PR.

> > @swiatekm I think we want to backport this to 8.17 also. WDYT?
> 
> Are we 100% sure there is no hidden bug here?

I'm never going to claim 100% for a codebase I haven't been working on for at least a year. I am as sure as I can reasonably be, I think. In practice, the change only has an effect when one of the dynamic variable providers is used. For these:

* The local dynamic provider is used in unit tests
* I've done fairly extensive manual e2e testing on the kubernetes provider

I could also manually test the docker provider, but I can't think of any reason it would be affected when the others aren't.

Issue #5952, opened 2024-11-06
Agent throws exception when uninstalling on windows


marc-gr:
For confirmed bugs, please report:
- Version: 8.16.0 bc3
- Operating System: windows
- Steps to Reproduce: 

Install and enroll an agent, then uninstall the agent.

On windows, intermittently elastic-agent will throw an exception when uninstalling. Logs in https://gist.github.com/Trinity2019/adddeec9a5ea7314e34c7d443a6bdc2f

cmacknz:
@mauri870 does anything jump out at you in the linked stack trace, it doesn't look like a normal panic, rather some kind of execution violation considering the only error is an invalid program counter: `runtime: g 1 gp=0xc00007c000: unknown pc 0x2`

cmacknz:
There is a similar but different error in https://github.com/elastic/elastic-agent/issues/3703 which was ultimately closed as not reproducible.

That one gave us the following:

```
unexpected fault address 0x1f86480
fatal error: fault
[signal 0xc0000005 code=0x1 addr=0x1f86480 pc=0x639c20]

goroutine 6 [running]:
fatal error: unexpected signal during runtime execution
panic during panic
[signal 0xc0000005 code=0x1 addr=0x63a425 pc=0x660b40]

runtime stack:
fatal error: unexpected signal during runtime execution
stack trace unavailable
```

ycombinator:
More occurrences:
* 2024-11-07: https://gist.github.com/Trinity2019/88aaaed9b5cf58cea8565572036dc9ff
* 2024-11-12: https://gist.github.com/Trinity2019/f7c6a15a11f425f3da3f44e7318f22f5
* 2024-11-13: https://gist.github.com/Trinity2019/48bb399243f9843108dd1e9150530ff8

blakerouse:
I was able to create an integration tests that reproduces this. The test runs install/uninstall in a loop with the Elastic Agent enrolled into Fleet each time. It normally fails on the 6th to 8th loop.

```
>>> (windows-amd64-2022-default) Test output (sudo) (stdout): === FAIL: testing/integration TestRepeatedInstallUninstallFleet/TestRepeatedInstallUninstallFleet-6 (5.93s)
[=== ] Attempting to notify Fleet of uninstall  [1s] runtime: pcdata is 39 and 4 locals stack map entries for github.com/elastic/elastic-agent-libs/transport/tlscommon.verifyCertsWithOpts (targetpc=0x8ecf77)
[...]

PR #6065, opened 2024-11-18
Skip notifyFleetAuditUninstall on Windows

blakerouse:
## What does this PR do?

Skips calling `notifyFleetAuditUninstall` on Windows.

When enabled it makes it much more likely that a panic will occur, but this doesn't actually seem like the culprit it just makes it more likely to happen.

**NOTE**: This doesn't actually fix the issue on Windows, it just makes it less likely to occur. Even with this disabled I am still seeing uninstall failing on like the 68th retry, but with this enabled it more likely happens on the 6-8th try. This is just reducing the likely that this issue occurs, it is not actually fixing the issue. (Just want to make that clear).

## Why is it important?

Reduces changes that uninstall happens on Windows.

## Checklist

## Related issues

- Relates #5952

### DIFF
--- a/internal/pkg/agent/install/uninstall.go
+++ b/internal/pkg/agent/install/uninstall.go
@@ -144,7 +144,9 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
 	}
 	pt.Describe("Removed install directory")
 
-	if notifyFleet {
+	// Skip on Windows because of https://github.com/elastic/elastic-agent/issues/5952
+	// Once the root-cause is identified then this can be re-enabled on Windows.
+	if notifyFleet && runtime.GOOS != "windows" {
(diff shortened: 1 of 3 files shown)


blakerouse:
See **NOTE** above about it not actually fix it, just for clarity if you missed it.

cmacknz (approved):

internal/pkg/agent/install/uninstall.go
```diff
@@ -144,7 +144,9 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
 	}
 	pt.Describe("Removed install directory")
 
-	if notifyFleet {
+	// Skip on Windows because of https://github.com/elastic/elastic-agent/issues/5952
+	// Once the root-cause is identified then this can be re-enabled on Windows.
+	if notifyFleet && runtime.GOOS != "windows" {
```
ycombinator, line 149: FYI @michel-laterman

Issue #4771, opened 2024-05-16
Remove legacy fields `Protocol`, `Host`, `Path` from Config struct in internal/pkg/remote/config.go


pchila:
`remote.Config` struct contains fields `Protocol`, `Host`, `Path` that are initialized from the enroll command using the `--url` CLI option.
Fleet sends connection parameters in policy change action only using `Hosts` as a `[]string ` containing full URLs without splitting them into URL parts.

The goal of this issue is to refactor the `enroll` command to use the same field as we receive in the policy change action and remove the legacy fields to reduce the noise about fleet client configuration.

PR #6085, opened 2024-11-19
Fix audit/unenroll calls when agent runs fleet-server

### LINKED ISSUE #5752: elastic-agent installed as Fleet server remains stuck in uninstalling with retry errors for over 4-5 minutes.
**Kibana Build details:**
```
VERSION: 8.16.0 SNAPSHOT
BUILD: 78993
COMMIT: 6eb8471c3124046eca03cccf20e0cc4f9706bcd5
```
Artifact: https://snapshots.elastic.co/8.16.0-106cdbc2/downloads/beats/elastic-agent/elastic-agent-8.16.0-SNAPSHOT-windows-x86_64.zip

Host: Windows Server 2022- Test Signing ON

**Preconditions:**
1. 8.16.0 SNAPSHOT Kibana self-managed environment should be available.
2. Agent should be installed using Quickstart or Production mode of security.

**Steps to reproduce:**
1. Observe Fleet Server is installed.
2. Run `.\Elastic\Agent\elastic-agent.exe uninstall`.
3. Observe notify: Fleet network error[retry] error is stuck for more than 4 minutes.
4. Observe folder is already blank before the completion.

**Expected Result:**
elastic-agent installed as Fleet server should not remain stuck in  uninstalling with retry errors for over 4-5 minutes.

**Screenshots:**
Image
Image
Image
Image
Image

michel-laterman:
## What does this PR do?

Call audit/unenroll (during uninstall) before components are stopped if the elastic-agent is running a fleet-server instance. 

## Why is it important?

Agents running fleet-server instances would always get an error because fleet-server would be stopped by the time the notify call is attempted.

## Checklist

Enroll an agent in a policy with the fleet-server integration and uninstall the agent

## Related issues

- Closes #5752


internal/pkg/agent/install/uninstall.go
```diff
@@ -59,6 +60,48 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
...
+	}
+
+	// Skip on Windows because of https://github.com/elastic/elastic-agent/issues/5952
+	// Once the root-cause is identified then this can be re-enabled on Windows.
+	// Notify fleet-server while it is still running if it's running locally
+	if notifyFleet && localFleet && runtime.GOOS != "windows" {
+		if cfg.Fleet.Client.Transport != nil {
+			cfg.Fleet.Client.Transport.TLS.VerificationMode = tlscommon.VerifyCertificate
+		}
+		cfg.Fleet.Client.Hosts = []string{cfg.Fleet.Client.Host}
```
michel-laterman, line 101: I'm not sure what's going on here; but `Client.Hosts` had the remote (cloud) fleet-server address, and `Client.Host` was the internal port.
I'm going to take some time tomorrow to verify that our normal agent operations still use the internal port when the agent is running fleet-server.

michel-laterman: agent does connect to `localhost:8221`; but strangely enough fleet-server does not populate the `server.address` attribute in HTTP logs; i'll open a fleet-server PR to fix this

internal/pkg/agent/install/uninstall.go
```diff
@@ -59,6 +60,48 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
...
+			pt.Describe(fmt.Sprintf("unable to read agent info, Fleet will not be notified of uninstall: %v", err))
+		} else {
+			notifyFleet = true
+		}
+		if cfg.Fleet != nil && cfg.Fleet.Server != nil {
+			localFleet = true
+		}
+	}
+
+	// Skip on Windows because of https://github.com/elastic/elastic-agent/issues/5952
```
swiatekm, line 94: Can we link to that issue in a more visible way? I'm concerned we're going to forget to re-enable this once that issue is resolved.

internal/pkg/agent/install/uninstall.go
```diff
@@ -59,6 +60,48 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
...
+	// Once the root-cause is identified then this can be re-enabled on Windows.
+	// Notify fleet-server while it is still running if it's running locally
+	if notifyFleet && localFleet && runtime.GOOS != "windows" {
+		if cfg.Fleet.Client.Transport.TLS != nil {
+			cfg.Fleet.Client.Transport.TLS.VerificationMode = tlscommon.VerifyCertificate
```
swiatekm, line 99: Why are we changing TLS settings here?

michel-laterman: The certificate does not have `localhost` as a DNS entry.

swiatekm: Shouldn't this be part of normal configuration parsing then? It seems like whatever additional adjustment we make here should always be applied.

pierrehilbert:
It's a detail for now but we should probably also plan to backport this to 8.17

cmacknz:
This needs a regression test :)

internal/pkg/agent/install/uninstall.go
michel-laterman: We inject the setting when we detect fleet-server will be running on enrollment: https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/cmd/enroll_cmd.go#L615-L620

michel-laterman: turns out this setting is passed correctly, but the `host` vs `hosts` was not

testing/integration/fleetserver_test.go
```diff
@@ -165,4 +163,6 @@ func TestInstallFleetServerBootstrap(t *testing.T) {
 		require.Error(t, err, "uninstall should have failed")
 		require.Containsf(t, string(out), "uninstall must be run from outside the installed path", "expected error string not found in: %s err: %s", out, err)
 	}
+
+	t.Run("Test audit/unenroll", testUninstallAuditUnenroll(ctx, fixture, info))
```
cmacknz, line 167: This is the regression test for this correct?

michel-laterman: yes, I've refactored `TestInstallUninstallAudit` to also be able to test the audit/unenroll notification for both a standard agent, and one running fleet-server

internal/pkg/agent/install/uninstall.go
```diff
@@ -59,6 +59,45 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
 		return fmt.Errorf("uninstall must be run from outside the installed path '%s'", topPath)
 	}
 
+	// check if the agent was installed using --unprivileged by checking the file vault for the agent secret (needed on darwin to correctly load the vault)
+	unprivileged, err := checkForUnprivilegedVault(ctx)
```
cmacknz, line 63: Instead of doing all this here you could just ask the agent if it bootstrapped a fleet server via an RPC call.

You could add something to the agent info response on if it has bootstrapped a fleet server or just look at the components in the state response. https://github.com/elastic/elastic-agent/blob/a4bf7d07bf441e279f4da3d826faf5fe6314edc0/control_v2.proto#L169

testing/integration/install_test.go
```diff
@@ -338,32 +338,41 @@ func TestInstallUninstallAudit(t *testing.T) {
 		return waitForAgentAndFleetHealthy(ctx, t, fixture)
 	}, time.Minute, time.Second, "agent never became healthy or connected to Fleet")
 
-	agentID, err := getAgentID(ctx, fixture)
-	require.NoError(t, err, "error getting the agent inspect output")
-	require.NotEmpty(t, agentID, "agent ID empty")
+	t.Run("run uninstall", testUninstallAuditUnenroll(ctx, fixture, info))
```
cmacknz, line 341: Am I blind or is this the exact same test logic before that was passing before this bug fix? Was this test actually catching the problem? Or was something changed I just can't see yet?

cmacknz: OK nevermind reading the wrong file, it's called in a new place now.

internal/pkg/agent/install/uninstall.go
cmacknz: `cfg.Fleet.Client.Hosts = []string{cfg.Fleet.Client.Host}` is a bit weird. This is perhaps something else that could be encapsulated into the output of an RPC call to the running agent service since agent itself must know that these represent two different things.

internal/pkg/agent/install/uninstall.go
```diff
@@ -59,6 +59,45 @@ func Uninstall(ctx context.Context, cfgFile, topPath, uninstallToken string, log
...
+			localFleet = true
+		}
+	}
+
+	// Skip on Windows because of https://github.com/elastic/elastic-agent/issues/5952
+	// Once the root-cause is identified then this can be re-enabled on Windows.
+	// Notify fleet-server while it is still running if it's running locally
+	if notifyFleet && localFleet && runtime.GOOS != "windows" {
+		cfg.Fleet.Client.Hosts = []string{cfg.Fleet.Client.Host}
+		notifyFleetAuditUninstall(ctx, log, pt, cfg, ai) //nolint:errcheck // ignore the error as we can't act on it
```
cmacknz, line 98: The Fleet API call we make doesn't return until the uninstall reason is persisted in ES right? Or just when it is queued in the bulker?

The very next thing we are going to do is SIGTERM all the agent processes, so if we haven't persisted the update into ES for sure, the result of this call isn't deterministic. This will probably show up in the integration test eventually if it were run enough times.

michel-laterman: It returns after writing to ES https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/handleAudit.go#L116

internal/pkg/agent/install/uninstall.go
michel-laterman: Using a `client.State` call to determine if we need to notify fleet or are running a fleet-server also adds the assumption that the agent is running during uninstall. Also we would need to load the config in order to build the fleet-server client to make the api request.

blakerouse: We have never had the requirement that the Elastic Agent is running to uninstall. I don't think we want to add that requirement for this to work. I would not use RPC to determine this, just like we don't do that to determine if endpoint-security is running.

internal/pkg/agent/install/uninstall.go
michel-laterman: We should avoid using an RPC call because we don't want to introduce the requirement that an agent is running for the uninstall to succeed.

internal/pkg/agent/install/uninstall.go
cmacknz: Agree, uninstall needs to work regardless of if the agent is running.

In this specific case, if the RPC fails because the agent isn't running, the local fleet server isn't running either. We would have to treat the RPC failure as non-fatal and make the call to a potentially remote fleet server anyway.

When I made this comment, I was thinking of a problem like https://github.com/elastic/elastic-agent/pull/5864#issuecomment-2457943974 which involved use of `sudo` to interact with an unprivileged agent and was broken in an unexpected way. For uninstall, it has to execute as root regardless of the privilege level of the agent, so I don't think we can have a problem like that here.

cmacknz:
I built this PR, installed it on my Mac as an unprivileged agent, and got the following output:

```sh
❯ ./elastic-agent version --binary-only
Binary: 9.0.0-SNAPSHOT (build: 1de9ae4b5a58dec89e5221418f7f5b31df855eb6 at 2024-12-02 20:24:03 +0000 UTC)

❯ sudo ./elastic-agent install --url=https://a64e3e715dc44617aac7175197185ca2.fleet.eastus2.staging.azure.foundit.no:443 --enrollment-token=XXXXX --develop --unprivileged

❯ sudo elastic-development-agent uninstall
Elastic Agent will be uninstalled from your system at /Library/Elastic/Agent-Development. Do you want to continue? [Y/n]:y
[  ==] unable to read agent info, Fleet will not be notified of uninstall: could not get agent info from store: fail to read configuration /Library/Elastic/Agent-Development/fleet.enc for the agent: fail to                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        [   =] Done  [1s]



















```

Note the `unable to read agent info` error, also there was a ridiculous amount of whitespace logged to the console as part of the uninstall logs.

I didn't have a fleet server input in the policy because most of the new permissions logic doesn't actually depend on that.

michel-laterman:
Using an inline function to return fixed the whitespace issue:
```
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo ./elastic-agent install --url=https://33b335c59053499a9d8ba70e7120f801.fleet.us-central1.gcp.qa.cld.elstc.co:443 --enrollment-token=XXX --develop --unprivileged
...
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent uninstall
Elastic Agent will be uninstalled from your system at /Library/Elastic/Agent-Development. Do you want to continue? [Y/n]:y
[====] Done  [0s]
```

cmacknz:
I tried this again locally and got the same agent info error and the crazy whitespace again. Including a screenshot because I can't copy/paste the whitespace.

Screenshot 2024-12-06 at 4 21 48 PM

There's so much whitespace I to scroll to see the eventual uninstall success.

michel-laterman:
I'm unable to recreate the failure locally, i've tried with an agent that has no integrations in the policy, and one that has the system+self monitoring, on my arm mac (using the `--develop --unprivileged` flags, and another attempt with just the `--develop` flag).

But my output lacks the block of white space and ends up looking like:
```
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo ./elastic-agent install --url=https://c7fc95a7c1aa4d018242d27b093d2635.fleet.us-central1.gcp.qa.cld.elstc.co:443 --enrollment-token=REDACTED --develop --unprivileged
Unprivileged installation mode enabled; this feature is currently in beta.
Elastic Agent will be installed at /Library/Elastic/Agent-Development and will run as a service. Do you want to continue? [Y/n]:y
[  ==] Service Started  [4s] Elastic Agent successfully installed, starting enrollment.
[    ] Waiting For Enroll...  [7s] {"log.level":"info","@timestamp":"2024-12-06T15:43:50.858-0800","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).enrollWithBackoff","file.name":"cmd/enroll_cmd.go","file.line":520},"message":"Starting enrollment to URL: https://c7fc95a7c1aa4d018242d27b093d2635.fleet.us-central1.gcp.qa.cld.elstc.co:443/","ecs.version":"1.6.0"}
[==  ] Waiting For Enroll...  [8s] {"log.level":"info","@timestamp":"2024-12-06T15:43:52.306-0800","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).daemonReloadWithBackoff","file.name":"cmd/enroll_cmd.go","file.line":483},"message":"Restarting agent daemon, attempt 0","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2024-12-06T15:43:52.308-0800","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).Execute","file.name":"cmd/enroll_cmd.go","file.line":301},"message":"Successfully triggered restart on running Elastic Agent.","ecs.version":"1.6.0"}
Successfully enrolled the Elastic Agent.
[==  ] Done  [8s]
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent version
Binary: 9.0.0-SNAPSHOT (build: 79b3645f4a0b8b76da76003eed22fbe54ebc80ce at 2024-12-06 23:21:40 +0000 UTC)
Daemon: 9.0.0-SNAPSHOT (build: 79b3645f4a0b8b76da76003eed22fbe54ebc80ce at 2024-12-06 23:21:40 +0000 UTC)
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent uninstall
Elastic Agent will be uninstalled from your system at /Library/Elastic/Agent-Development. Do you want to continue? [Y/n]:y
[====] Done  [0s]
```

I also tested using a policy with fleet-server in a VM (trying on my mac gave me an issue with the self-signed cert) and didn't get the same output.

michel-laterman:
From what I can tell, the out difference in output (whitespace) is caused by progressbar having strange output when the text is too wide

testing/integration/fleetserver_test.go
cmacknz, line 167: This test only runs for privileged agents, because of this block:

```go
	// Run `elastic-agent install` with fleet-server bootstrap options.
	// We use `--force` to prevent interactive execution.
	opts := &atesting.InstallOpts{
		Force:      true,
		Privileged: true,
```

There shouldn't be a reason for this test to only run privileged, tests should ideally be unprivileged by default as that is the harder case to get right.

michel-laterman: I think i've fixed it; I no longer try to load the agent info through the info package; i just use the agent ID from config. With these changes my local testing (on a mac) shows the uninstalled audit reason, and no more wall of whitespace. I'll see about adding a test that does not run in privileged mode

internal/pkg/agent/install/uninstall.go
pchila: What I *think* is happening here:
- elastic agent sets `Host` from the [enroll command](https://github.com/elastic/elastic-agent/blob/b07566b96a95daa11897c7fbf0b9320618e2540b/internal/pkg/agent/cmd/enroll_cmd.go#L984) 
- in the policy_change handler elastic-agent uses [only `Hosts` field coming from Fleet](https://github.com/elastic/elastic-agent/blob/8c43ad333837380f1646c3b007ff6339b1593a94/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go#L202-L209) and tries to reset the other fields. 

I guess that the merged result of that is not quite right in this scenario and we end up with the discrepancy between `Host` and `Hosts`

From discussion with @nchaulet  at the time, it was determined that Fleet will send back only `Hosts` field while the separate `Host`, `Protocol`, `Path` is only handled by agent when initializing config to store command line args.

I filed an [issue](https://github.com/elastic/elastic-agent/issues/4771) to have a look and possible remote Host, Protocol and Path fields from the agent config at the time but maybe this needs further evaluation/discussion taking into account fleet bootstrap scenarios

pchila (approved):
Tested on Linux with and without the fix and it solves the issue with audit/unenroll when running a local Fleet server

I left a small [comment](https://github.com/elastic/elastic-agent/pull/6085#discussion_r1879481233) for the `Host`/`Hosts` discrepancy we have trying to give more context, not sure if we want to follow that up with a dedicated issue

internal/pkg/agent/install/uninstall.go
cmacknz: Can we link to that issue and add this explanation to the code? This is going to be really not obvious to someone reading this code without the context of this PR.

cmacknz (approved):
Tried it with an unprivileged agent on my Mac and it worked this time. Thanks for the follow up and additional test coverage!

```
❯ sudo elastic-development-agent uninstall
Elastic Agent will be uninstalled from your system at /Library/Elastic/Agent-Development. Do you want to continue? [Y/n]:y
[=== ] Done  [2s]
```

swiatekm (approved):
LGTM

michel-laterman:
@cmacknz @ycombinator, can one of you please force this in once buildkite completes? we have integration test coverage but not unit tests

michel-laterman:
buildkite test this

cmacknz:
Force merging now that CI has passed.

PR #6109, opened 2024-11-21
Make redaction of common keys in diagnostics case insensitive

andrzej-stencel:
## What does this PR do?

Changes the redaction of common keys like "password" to be case insensitive. This means keys like "PASSWORD" and "PassWord" will also be redacted.

## Why is it important?

To prevent leaking secrets into diagnostics.

## Checklist

   - I couldn't find any docs on diagnostics redaction

## Disruptive User Impact

This may cause keys that were previously not redacted to be redacted after the change. For example, the "Key" substring may be common in configuration keys that don't need redaction.

```console
cd internal/pkg/diagnostics
go test .
```


pkoutsovasilis (approved):
LGTM

pkoutsovasilis:
AFAICT unit-test fail because otel collector never reports healthy (unrelated to this PR) and the serverless ones are due to the BK agent (Error: error creating test runner: error checking region setting: Non-201 status code returned by server: 503)?!

swiatekm (approved):

ycombinator:
> Should this be backported into 8.16? I marked this as an "enhancement", but maybe it is a bug fix?

I agree, this feels more like a bug than an enhancement.  Added backport labels and bug label instead of enhancement.

Issue #4148, opened 2024-01-25
[Flaky Test]: TestOtelAPMIngestion – there should be apm logs by now


cmacknz:
### Failing test case

TestOtelAPMIngestion

### Error message

there should be apm logs by now: map[string]bool{"This is a test debug message 2":false, "This is a test debug message 3":false, "This is a test debug message 4":false, "This is a test error message":false}

### Build

https://buildkite.com/elastic/elastic-agent/builds/6574

### OS

Linux

### Stacktrace and notes

```markdown
=== RUN   TestOtelAPMIngestion
    fetcher.go:91: Using existing artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz
    fixture.go:261: Extracting artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestOtelAPMIngestion3888313451/002
    fixture.go:274: Completed extraction of artifact elastic-agent-8.13.0-SNAPSHOT-linux-arm64.tar.gz to /tmp/TestOtelAPMIngestion3888313451/002
    fixture.go:886: All component specifications where removed
    fixture.go:927: Placed component specifications: fake, fake-shipper
    log_watcher.go:40: {"log.level":"info","@timestamp":"2024-01-25T19:19:13.716Z","log.origin":{"function":"github.com/elastic/apm-server/internal/beatcmd.(*Beat).init","file.name":"beatcmd/beat.go","file.line":143},"message":"Home path: [/tmp/TestOtelAPMIngestion3888313451/002/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/data/elastic-agent-d93eb0/components] Config path: [/tmp/TestOtelAPMIngestion3888313451/002/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/data/elastic-agent-d93eb0/components] Data path: [/tmp/TestOtelAPMIngestion3888313451/002/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/data/elastic-agent-d93eb0/components/data] Logs path: [/tmp/TestOtelAPMIngestion3888313451/002/elastic-agent-8.13.0-SNAPSHOT-linux-arm64/data/elastic-agent-d93eb0/components/logs]","service.name":"apm-server","ecs.version":"1.6.0"}
[...]

PR #6114, opened 2024-11-21
Store AST directly in provider mappings

swiatekm:
## What does this PR do?

For all variable provider types, we store mappings as untyped dictionaries and only convert them to AST when we're about to generate actual vars. This means we have to recompute all of it every time there's a change to any of the provider data, even if all the other data is unaffected. For example, if a Kubernetes Pod gains a new label, we need to regenerate vars for all Pods.

Instead, we can store mappings as AST to begin with. Then the only cost of emitting new vars after a change is updating a bunch of references.

## Why is it important?

This unnecessary recomputation can be quite expensive when there's a lot of mappings from the dynamic provider. See the benchstat report below:

```
goos: linux
goarch: amd64
pkg: github.com/elastic/elastic-agent/internal/pkg/composable
                       │ bench_main.txt │          bench_dynamic.txt          │
GenerateVars100Pods-20     755.49µ ± 3%   76.87µ ± 2%  -89.82% (p=0.000 n=10)

                       │ bench_main.txt │          bench_dynamic.txt           │
GenerateVars100Pods-20    735.00Ki ± 0%   98.69Ki ± 0%  -86.57% (p=0.000 n=10)

                       │ bench_main.txt │          bench_dynamic.txt          │
GenerateVars100Pods-20     21.335k ± 0%   1.364k ± 0%  -93.61% (p=0.000 n=10)
```

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/5835
- Relates https://github.com/elastic/elastic-agent/issues/5991

### DIFF
--- a/internal/pkg/agent/transpiler/ast.go
+++ b/internal/pkg/agent/transpiler/ast.go
@@ -825,6 +825,9 @@ func (a *AST) HashStr() string {
 
 // Equal check if two AST are equals by using the computed hash.
 func (a *AST) Equal(other *AST) bool {
+	if a.root == nil || other.root == nil {
+		return a.root == other.root
+	}
 	return bytes.Equal(a.Hash(), other.Hash())
 }
 
--- a/internal/pkg/composable/controller.go
+++ b/internal/pkg/composable/controller.go
@@ -79,10 +79,12 @@ func New(log *logger.Logger, c *config.Config, managed bool) (Controller, error)
 		if err != nil {
 			return nil, errors.New(err, fmt.Sprintf("failed to build provider '%s'", name), errors.TypeConfig, errors.M("provider", name))
 		}
+		emptyMapping, _ := transpiler.NewAST(nil)
 		contextProviders[name] = &contextProviderState{
 			// Safe for Context to be nil here because it will be filled in
 			// by (*controller).Run before the provider is started.
 			provider: provider,
+			mapping:  emptyMapping,
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


internal/pkg/agent/transpiler/ast.go
```diff
@@ -825,6 +825,9 @@ func (a *AST) HashStr() string {
 
 // Equal check if two AST are equals by using the computed hash.
 func (a *AST) Equal(other *AST) bool {
+	if a.root == nil || other.root == nil {
+		return a.root == other.root
+	}
```
pkoutsovasilis, line 830: nit 🙂 

```suggestion
	}
	// nil check above is mandatory as calling Hash() on a nil may cause a panic
```

internal/pkg/composable/controller.go
```diff
@@ -324,30 +323,25 @@ func (c *contextProviderState) Signal() {
...
+	ast, err := transpiler.NewAST(mapping)
 	if err != nil {
 		return err
 	}
 
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	if reflect.DeepEqual(c.mapping, mapping) {
+	if c.mapping != nil && c.mapping.Equal(ast) {
```
pkoutsovasilis, line 334: ok now I see what you meant that we don't need to check if something has indeed been updated when at Insert. Ty

pkoutsovasilis (approved):
This code change LGTM. the CI failures are due to `TestOtelAPMIngestion` (https://github.com/elastic/elastic-agent/issues/4148) and `TestRpmFleetUpgrade`. The last one fails because it doesn't download in time the `elastic-agent-8.17.0-SNAPSHOT-x86_64.rpm` artifact with a `context deadline exceeded`

ycombinator:
> the CI failures are due to `TestOtelAPMIngestion` (#4148) and `TestRpmFleetUpgrade`. The last one fails because it doesn't download in time the `elastic-agent-8.17.0-SNAPSHOT-x86_64.rpm` artifact with a `context deadline exceeded` ([code ref](https://github.com/elastic/elastic-agent/blob/main/pkg/testing/fetcher_artifact.go#L200-L202))
> 
> cc @ycombinator should we create an issue about the latter?

I'm not seeing these tests failing on `main` so either something in this PR is causing them to fail (unlikely) or they are flaky. 
I just re-ran the integration tests step.  If it passes, it means those tests are flaky and we should [create an issue](https://github.com/elastic/elastic-agent/issues/new?template=flaky-test.yml) for each of them.

blakerouse (approved):
This is a great performance improvement! Nicely done.

PR #6121, opened 2024-11-21
Temporarily pin versions for integration tests to `8.17.0`

ycombinator:
Follow up to https://github.com/elastic/elastic-agent/pull/6098.

While the Agent version has been set to `8.18.0` on the `8.x` branch, it's not yet possible to create `8.18.0-SNAPSHOT` deployments in ESS.  As such, Agent integration tests — which use deployments in ESS — will fail on the `8.x` branch.

To work around this problem, this PR temporarily builds Elastic Agent with the `8.17.0` version and uses this build in integration tests.  Likewise, it asks the integration tests to spin up `8.17.0-SNAPSHOT` deployments in ESS.

---

We should revert this PR once it becomes possible to create `8.18.0-SNAPSHOT` deployments in ESS.


jlind23 (approved):

ycombinator:
> We should revert this PR once it becomes possible to create 8.18.0-SNAPSHOT deployments in ESS.

It is now possible to create `8.18.0-SNAPSHOT` deployments in the ESS production CFT region, so reverting this PR...

PR #6167, opened 2024-11-28
[CI] Source BUILDKITE_ANALYTICS_TOKEN in the commands

pazone:
## What does this PR do?
Sources `BUILDKITE_ANALYTICS_TOKEN` in the target script instead of `pre-command`.

## Why is it important?
Simplifies the variable declaration conditions. The build step declares that the `BUILDKITE_ANALYTICS_TOKEN` is required by sourcing the `.buildkite/scripts/set-bk-analytics-token.sh`

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/set-bk-analytics-token.ps1
+++ b/.buildkite/scripts/set-bk-analytics-token.ps1
@@ -6,4 +6,4 @@ Write-Host "--- Set BUILDKITE_ANALYTICS_TOKEN :vault:"
 $BUILDKITE_ANALYTICS_TOKEN = Retry-Command -ScriptBlock {
   vault kv get -field=token kv/ci-shared/platform-ingest/buildkite_analytics_token
 }
-[System.Environment]::SetEnvironmentVariable("BUILDKITE_ANALYTICS_TOKEN", $BUILDKITE_ANALYTICS_TOKEN, [System.EnvironmentVariableTarget]::Machine)
+[System.Environment]::SetEnvironmentVariable("BUILDKITE_ANALYTICS_TOKEN", $BUILDKITE_ANALYTICS_TOKEN, [System.EnvironmentVariableTarget]::User)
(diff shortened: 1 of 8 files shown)


.buildkite/scripts/set-bk-analytics-token.sh
```diff
@@ -0,0 +1,4 @@
+# !/bin/bash
+echo "--- Set BUILDKITE_ANALYTICS_TOKEN :vault:"
```
alexsapran, line 2: should we define the shebang here at least the pipefail so if this fails, it would stop the execution

.buildkite/scripts/set-bk-analytics-token.ps1
```diff
@@ -0,0 +1,3 @@
+Write-Host "--- Prepare BK test analytics token :vault:"
+$BUILDKITE_ANALYTICS_TOKEN = & vault kv get -field=token kv/ci-shared/platform-ingest/buildkite_analytics_token
```
alexsapran, line 2: similar to the other comment, should we define some flags that would prevent execution if there is a failure

alexsapran: and potentially add a retry?

alexsapran (commented):
As we discussed, I'm not sure why we want to move the logic of injecting tokens out of the hook. 
For this repo, we can have this, and we can discuss the rest of the repositories.

I have left some comments, onced adressed i will approve

.buildkite/scripts/set-bk-analytics-token.ps1
alexsapran: I was wrong. Retrying is a bit more complicated, so it might make sense to follow up and add it later.

pazone: Added retries

PR #6169, opened 2024-11-28
Only run providers that are referenced in the policy

### LINKED ISSUE #3609: Run only providers that have references in the policy
**Describe the enhancement:**

Currently the Elastic Agent runs all of its providers all the time. Context providers are rather low overhead, but dynamic providers are not and when they are not even being used its adding load for no benefit to the running policy.

We should adjust the Elastic Agent to only run providers that are referenced in the policy. Variable referencing is always prefixed with a provider name aka `${kubernetes.*}` that means its rather easy to identify which providers should be enabled based on what is in the policy.

This also removes the need to add any configuration to Fleet to disable or enable a provide for the Elastic Agent. If the policy from Fleet doesn't use a provider then the Elastic Agent will not run it.

Reduce load of an Elastic Agent, prevents it from trying to communicate with Docker or Kubernetes in the case that its not even running a Docker or Kubernetes inputs in the policy.

I propose the following change:

1. Parse the policy into the AST.
2. Read the AST to determine the providers that are referenced.
3. Start/stop the providers that are referenced or no longer referenced.

### LINKED ISSUE #4648: Elastic Agent is waiting for initial configuration forever when all providers are disabled
In case of no enabled providers, the elastic agent stalls forever.
It seems to be a bug here:
https://github.com/elastic/elastic-agent/blob/0c7212f2d92021a9e008de4abe362d0c77f78638/internal/pkg/composable/controller.go#L188-L203
where there is no way to break to `DEBOUNCE` if no provider is updating.

A possible solution could be based on the providers config length or the waiting group which should return immediately in this case since if there are no providers it is equal to 0.
https://github.com/elastic/elastic-agent/blob/0c7212f2d92021a9e008de4abe362d0c77f78638/internal/pkg/composable/controller.go#L128

The bug was discovered as part of the work on agentless controller. Currently we use a [workaround](https://github.com/elastic/agentless-controller/pull/343) to solve this issue.

Bug details:
- Version: 8.13.2
- Operating System: all
- Steps to Reproduce:

Setup (`agent-bug` is the directory name which shows up before every command)
```
➜  test cd agent-bug
➜  agent-bug docker pull docker.elastic.co/beats/elastic-agent:8.13.2
8.13.2: Pulling from beats/elastic-agent
c93e5d1261d3: Pull complete
9204e5b4f4d9: Pull complete
77db57972e9d: Pull complete
5a09faecb150: Pull complete
ef1d475c705b: Pull complete
9be0bc4d4489: Pull complete
97bac83776bc: Pull complete
594c586edc1b: Pull complete
b45e1922fc73: Pull complete
a9c1a4bc09dd: Pull complete
83a50bca82ec: Pull complete
4ca545ee6d5d: Pull complete
Digest: sha256:1b1346f6228c4cfcc8bd6b05e0eb24f15bcfd616935d0f2fffe0754d7d3fe31b
Status: Downloaded newer image for docker.elastic.co/beats/elastic-agent:8.13.2
docker.elastic.co/beats/elastic-agent:8.13.2
```
Get the original config file
```
➜  agent-bug docker run --rm -d --name elastic-agent docker.elastic.co/beats/elastic-agent:8.13.2 container
76af237adeb219eb7591c6b7647c3bfe523e4d73121bf1677388720b79e29d85
➜  agent-bug docker cp elastic-agent:/usr/share/elastic-agent/elastic-agent.yml ./
➜  agent-bug docker stop elastic-agent
```
Modify it to disable all providers
```
➜  agent-bug cat <<EOF >> elastic-agent.yml
providers:
  agent:
    enabled: false
  docker:
    enabled: false
  env:
    enabled: false
  host:
    enabled: false
  kubernetes:
    enabled: false
  kubernetes_leaderelection:
    enabled: false
  kubernetes_secrets:
    enabled: false
  local:
    enabled: false
  local_dynamic:
    enabled: false
  path:
    enabled: false
EOF
```
Start the agent with the modified config
```
➜  agent-bug docker run --rm -d --name elastic-agent -v ./elastic-agent.yml:/usr/share/elastic-agent/elastic-agent.yml docker.elastic.co/beats/elastic-agent:8.13.2 container
decb5e6e6d443e11d228a9ff32e8a9ad3ed78f466b7c830e85aec0a3818b9aa4
```
Enter the container
```
➜  agent-bug docker exec -it elastic-agent /bin/bash
```
Agent stuck on waiting for initial configuration
```
elastic-agent@decb5e6e6d44:~$ elastic-agent status
┌─ fleet
└─ elastic-agent
   └─ status: (STARTING) Waiting for initial configuration and composable variables
```
... waiting ...
```
elastic-agent@decb5e6e6d44:~$ elastic-agent status
┌─ fleet
└─ elastic-agent
   └─ status: (STARTING) Waiting for initial configuration and composable variables
```
Agent configuration (for some reason inspect stalls so I have to kill it)
```
elastic-agent@decb5e6e6d44:~$ elastic-agent inspect
agent:
  logging:
    to_stderr: true
inputs:
- data_stream.namespace: default
  streams:
  - data_stream.dataset: system.cpu
    metricsets:
    - cpu
  - data_stream.dataset: system.memory
    metricsets:
    - memory
  - data_stream.dataset: system.network
    metricsets:
    - network
  - data_stream.dataset: system.filesystem
    metricsets:
    - filesystem
  type: system/metrics
  use_output: default
outputs:
  default:
    hosts: http://elasticsearch:9200
    password: changeme
    preset: balanced
    type: elasticsearch
    username: elastic
providers:
  agent:
    enabled: false
  docker:
    enabled: false
  env:
    enabled: false
  host:
    enabled: false
  kubernetes:
    enabled: false
  kubernetes_leaderelection:
    enabled: false
  kubernetes_secrets:
    enabled: false
  local:
    enabled: false
  local_dynamic:
    enabled: false
  path:
    enabled: false

^CError: could not load agent info: could not get agent info from store: failed to load from ioStore: failed to ensure key during encrypted disk store Load: could not get agent key: failed to acquire exclusive lock: /usr/share/elastic-agent/state/vault/.lock, err: context canceled
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.13/fleet-troubleshooting.html
elastic-agent@decb5e6e6d44:~$ exit
```
Cleanup
```
➜  agent-bug docker stop elastic-agent
elastic-agent
```

blakerouse:
## What does this PR do?

This updates the Elastic Agent to only run composable providers if they are referenced in the policy.

## Why is it important?

This change provides the following benefits:

- Reduces the memory and load of the Elastic Agent, by only running the providers it references.
- No longer requires access to the kubernetes API or resources just to run an Elastic Agent (as long as its not using any ${kubernetes*` references in the policy).
- Fixes an issue where if no providers are running that a set of empty variables is passed to the coordinator.

## Checklist

## Disruptive User Impact

None. As it still runs all those that are referenced from the user.

Run the elastic-agent container with just the system integration and notice that it doesn't require access to the Kubernetes API any more.

## Related issues

- Closes #3609 
- Closes #4648

### DIFF
--- a/internal/pkg/agent/transpiler/ast.go
+++ b/internal/pkg/agent/transpiler/ast.go
@@ -58,6 +58,9 @@ type Node interface {
 	// Hash compute a sha256 hash of the current node and recursively call any children.
 	Hash() []byte
 
+	// Vars return a list of referenced vars.
+	Vars([]string) []string
+
 	// Apply apply the current vars, returning the new value for the node.
 	Apply(*Vars) (Node, error)
 
@@ -160,6 +163,15 @@ func (d *Dict) Hash() []byte {
 		h.Write(v.Hash())
 	}
 	return h.Sum(nil)
+}
+
+// Vars returns a list of all variables referenced in the dictionary.
+func (d *Dict) Vars(vars []string) []string {
+	for _, v := range d.value {
+		k := v.(*Key)
+		vars = k.Vars(vars)
+	}
+	return vars
 }
 
 // Apply applies the vars to all the nodes in the dictionary.
@@ -275,6 +287,14 @@ func (k *Key) Hash() []byte {
 		h.Write(k.value.Hash())
 	}
 	return h.Sum(nil)
+}
+
+// Vars returns a list of all variables referenced in the value.
+func (k *Key) Vars(vars []string) []string {
+	if k.value == nil {
+		return vars
+	}
+	return k.value.Vars(vars)
 }
 
 // Apply applies the vars to the value.
@@ -397,6 +417,14 @@ func (l *List) ShallowClone() Node {
 	return &List{value: nodes}
 }
 
+// Vars returns a list of all variables referenced in the list.
+func (l *List) Vars(vars []string) []string {
+	for _, v := range l.value {
+		vars = v.Vars(vars)
+	}
+	return vars
+}
+
 // Apply applies the vars to all nodes in the list.
 func (l *List) Apply(vars *Vars) (Node, error) {
 	nodes := make([]Node, 0, len(l.value))
@@ -472,6 +500,40 @@ func (s *StrVal) Hash() []byte {
 	return []byte(s.value)
 }
 
+// Vars returns a list of all variables referenced in the string.
+func (s *StrVal) Vars(vars []string) []string {
+	value := s.value
+	matchIdxs := varsRegex.FindAllSubmatchIndex([]byte(value), -1)
+	if !validBrackets(value, matchIdxs) {
+		// brackets are not valid; unable to pull vars (computing the policy will fail)
+		return vars
+	}
+	for _, r := range matchIdxs {
+		for i := 0; i < len(r); i += 4 {
+			if value[r[i]+1] == '$' {
+				// match on an escaped var, this is not a real variable
+				continue
+			}
+			// match on a non-escaped var
+			extractedVars, err := extractVars(value[r[i+2]:r[i+3]])
+			if err != nil {
+				// variable parsing failed (computing the policy will fail)
+				return vars
+			}
+			for _, val := range extractedVars {
+				switch val.(type) {
+				case *constString:
+					// not a variable
+				case *varString:
+					// found variable add it to the array
+					vars = append(vars, val.Value())
+				}
+			}
+		}
+	}
+	return vars
+}
+
 // Apply applies the vars to the string value.
 func (s *StrVal) Apply(vars *Vars) (Node, error) {
 	return vars.Replace(s.value)
@@ -521,6 +583,11 @@ func (s *IntVal) Clone() Node {
 // ShallowClone makes a shallow clone of the node.
 func (s *IntVal) ShallowClone() Node {
 	return s.Clone()
+}
+
+// Vars does nothing. Cannot have variable in an IntVal.
+func (s *IntVal) Vars(vars []string) []string {
+	return vars
 }
 
 // Apply does nothing.
@@ -584,6 +651,11 @@ func (s *UIntVal) Hash() []byte {
 	return []byte(s.String())
 }
 
+// Vars does nothing. Cannot have variable in an UIntVal.
+func (s *UIntVal) Vars(vars []string) []string {
+	return vars
+}
+
 // Apply does nothing.
 func (s *UIntVal) Apply(_ *Vars) (Node, error) {
 	return s, nil
@@ -639,6 +711,11 @@ func (s *FloatVal) ShallowClone() Node {
 // Hash return a string representation of the value, we try to return the minimal precision we can.
 func (s *FloatVal) Hash() []byte {
 	return []byte(strconv.FormatFloat(s.value, 'f', -1, 64))
+}
+
+// Vars does nothing. Cannot have variable in an FloatVal.
+func (s *FloatVal) Vars(vars []string) []string {
+	return vars
 }
 
 // Apply does nothing.
@@ -701,6 +778,11 @@ func (s *BoolVal) Hash() []byte {
 		return trueVal
 	}
 	return falseVal
+}
+
+// Vars does nothing. Cannot have variable in an BoolVal.
+func (s *BoolVal) Vars(vars []string) []string {
+	return vars
 }
 
 // Apply does nothing.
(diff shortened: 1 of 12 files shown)


internal/pkg/agent/transpiler/ast.go
```diff
@@ -982,6 +1064,11 @@ func attachProcessors(node Node, processors Processors) Node {
 
 // Lookup accept an AST and a selector and return the matching Node at that position.
 func Lookup(a *AST, selector Selector) (Node, bool) {
+	// Be defensive and ensure that the ast is usable.
+	if a == nil || a.root == nil {
+		return nil, false
+	}
```
swiatekm, line 1070: We should probably write some tests for these kinds of cases. Not necessarily in this PR, but I recall recently adding similar conditions.

internal/pkg/composable/controller.go
```diff
@@ -192,8 +171,13 @@ func (c *controller) Run(ctx context.Context) error {
...
 			select {
 			case <-ctx.Done():
-				cleanupFn()
 				return ctx.Err()
+			case observed := <-c.observedCh:
+				c.handleObserved(localCtx, &wg, stateChangedChan, fetchContextProviders, observed)
+				t.Reset(100 * time.Millisecond)
+				c.logger.Debugf("Observed state changed for composable inputs; debounce started")
+				drainChan(stateChangedChan)
+				break DEBOUNCE
```
swiatekm, line 180: This is so we always update vars after `Observe` is called? If so, we should make sure we only emit if anything changed.

internal/pkg/composable/controller.go
```diff
@@ -35,18 +35,24 @@ type Controller interface {
...
-	dynamicProviders map[string]*dynamicProviderState
+	logger          *logger.Logger
+	ch              chan []*transpiler.Vars
+	observedCh      chan map[string]bool
+	errCh           chan error
+	restartInterval time.Duration
+
+	managed          bool
+	contextProviders map[string]contextProvider
+	dynamicProviders map[string]dynamicProvider
```
swiatekm, line 52: These are more like provider *factories* rather than providers themselves, right?

swiatekm (commented):
Looks good overall, had some questions about the design and intended control flow.

internal/pkg/agent/transpiler/ast.go
```diff
@@ -58,6 +58,9 @@ type Node interface {
 	// Hash compute a sha256 hash of the current node and recursively call any children.
 	Hash() []byte
 
+	// Vars return a list of referenced vars.
+	Vars([]string) []string
```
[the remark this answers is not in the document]
blakerouse: Correct.

internal/pkg/composable/controller.go
blakerouse: Correct.

internal/pkg/agent/transpiler/ast.go
blakerouse: I can add a test.

internal/pkg/composable/controller.go
blakerouse: This should only happen when the policy changes which should not be often, at least not as often compared to a variable changing in a dynamic provider like kubernetes. I will update it to only emit if something actually changed.

internal/pkg/agent/transpiler/ast.go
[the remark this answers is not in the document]
swiatekm: We should add a comment explaining that, imo. It wasn't obvious to me, and I was a bit confused looking at the implementations. Maybe change the name to `VarsInto`?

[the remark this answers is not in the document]
blakerouse: I added a comment in latest commit.

internal/pkg/agent/transpiler/ast.go
```diff
@@ -472,6 +500,40 @@ func (s *StrVal) Hash() []byte {
...
+				case *constString:
+					// not a variable
+				case *varString:
+					// found variable add it to the array
+					vars = append(vars, val.Value())
+				}
+			}
+		}
+	}
+	return vars
```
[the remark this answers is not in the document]
blakerouse: I updated this to share the code.

internal/pkg/composable/controller.go
blakerouse: Updated in latest commit.

internal/pkg/composable/controller.go
blakerouse: Updated variable naming to make it easier to read.

internal/pkg/agent/transpiler/ast.go
blakerouse: Test added.

swiatekm (approved):
:+1:

cmacknz:
Nice! We need to update the docs to indicate that providers are no longer enabled by default for the first release that will have this change (8.18 I believe): https://www.elastic.co/guide/en/fleet/current/providers.html#disable-providers-by-default

Some of the individual providers also indicate they are enabled by default like https://www.elastic.co/guide/en/fleet/current/kubernetes_leaderelection-provider.html

CC @kilfoyle

PR #6174, opened 2024-12-02
feat: drop init.d support and go-daemon binaries

kruskall:
## What does this PR do?

remove init.d scripts
drop go-daemon dep and daemon binaries
cleanup mage tasks

## Why is it important?

Similar to https://github.com/elastic/beats/pull/41845

init.d should not be needed on distros we support
According to go-daemon readme: "Go daemon is not needed on systems with [systemd](http://www.freedesktop.org/wiki/Software/systemd/)."

See https://www.elastic.co/support/matrix#matrix_os

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


kaanyalti (approved):

kaanyalti (dismissed):
Code changes look good to me, but I don't have enough context here, if someone else can look at it as well it would be better.

cmacknz:
This is fine as long as the DEB and RPM packages still run as services. That there is no replacement systemd unit file added here makes me think we may have broken something. I see the DEB tests are failing now, so that is one potential reason.

We create out existing systemd service in Go code right now so there isn't an easy way to extract it without refactoring that. https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/agent/install/svc.go#L84

blakerouse (changes requested):
I think this change is possible because it is true that on DEB/RPM that we do not support any OS that uses sysvinit. But there is clearly an issue with the upgrade path for the deb packages.

Interested to determine what you find on why that is not working with these changes. I am going to mark this as "Request changes" only, because I was to review it again before its merged with that fixed.

cmacknz:
There is an actual bug related to us continuing to include init.d files now https://github.com/elastic/elastic-agent/issues/6704

michalpristas:
closing in favor of https://github.com/elastic/elastic-agent/pull/8896
i picked that up and missed work done here. 
thanks anyways @kruskall for following up with agent repo as well

PR #6178, opened 2024-12-02
Do external package manager check only at startup

swiatekm:
## What does this PR do?

We were doing the check for external package manager use when validating vars for the platform, which happens on every configuration reload. I've made this an attribute of the platform instead, and it's now only checked on startup.

## Why is it important?

Reading a bunch of files every time configuration is reloaded is an unnecessary expense.

## Checklist

## Related issues

- Relates #5835 #5991

### DIFF
--- a/changelog/fragments/1733131399-vars-for-platform.yaml
+++ b/changelog/fragments/1733131399-vars-for-platform.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: enhancement
 
 # Change summary; a 80ish characters long description of the change.
 summary: Do the external package manager check only at startup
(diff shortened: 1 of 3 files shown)


kaanyalti (approved):

changelog/fragments/1733131399-vars-for-platform.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11:
```suggestion
kind: enhancement
```

blakerouse (approved):

PR #6180, opened 2024-12-02
Add config generation benchmark

swiatekm:
## What does this PR do?

Add a benchmark for config generation in the coordinator. We specifically profile having a lot of Pods from the Kubernetes provider, which is something that has recently become a problem, and which the configuration generation pipeline doesn't deal with very well. I used test data from a real environment where the problems outlined in https://github.com/elastic/elastic-agent/issues/5991 and https://github.com/elastic/elastic-agent/issues/5835 could be reproduced.

From profiling the live agent experiencing these issues, this benchmark covers the most expensive parts, so it should be representative.

Current results on `main`:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator
cpu: 13th Gen Intel(R) Core(TM) i7-13700H
BenchmarkCoordinator_generateComponentModel-20    	      30	  36693604 ns/op	35875285 B/op	  810689 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      28	  37448198 ns/op	35886685 B/op	  810670 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      27	  37193185 ns/op	35867024 B/op	  810650 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      22	  47190285 ns/op	35865402 B/op	  810526 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      31	  59598417 ns/op	35889725 B/op	  810712 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      31	  38748898 ns/op	35890251 B/op	  810715 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      28	  38338585 ns/op	35878466 B/op	  810667 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      27	  38126388 ns/op	35889337 B/op	  810647 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      31	  37849710 ns/op	35884377 B/op	  810712 allocs/op
BenchmarkCoordinator_generateComponentModel-20    	      27	  37850712 ns/op	35882968 B/op	  810648 allocs/op
PASS
ok  	github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator	24.152s
```

## Why is it important?

I'd like to make some improvements to the configuration generation in the coordinator, and want a straightforward way of demonstrating their impact.

## Checklist

- [x] My code follows the style guidelines of this project
[...]

PR #6184, opened 2024-12-02
Avoid unnecessary copies during config generation

swiatekm:
## What does this PR do?

Remove some unnecessary copies when generating component configuration.

The first copy avoided is done because we want to replace the `inputs` key with the rendered inputs. A shallow copy is sufficient for this. I did find and fix a bug in Dict shallow copies while at it.

The rest of the copies have to do with applying vars to each AST node. We currently make a copy of each Node we do this for, but that's not necessary, as `Apply` creates a new Node with all the data recursively copied anyway. I've added a test to verify that `Apply` actually works this way.

Benchstat result vs `main`, using the benchmark from #6180:

```
goos: linux
goarch: amd64
                                      │ bench_main.txt │          bench_noclone.txt          │
Coordinator_generateComponentModel-20     37.99m ± 24%   32.36m ± 4%  -14.82% (p=0.000 n=10)

                                      │ bench_main.txt │          bench_noclone.txt           │
Coordinator_generateComponentModel-20     34.22Mi ± 0%   25.38Mi ± 0%  -25.84% (p=0.000 n=10)

                                      │ bench_main.txt │          bench_noclone.txt          │
Coordinator_generateComponentModel-20      810.7k ± 0%   580.4k ± 0%  -28.40% (p=0.000 n=10)
```

## Why is it important?

More performance is good! More broadly, this is an attempt to incrementally improve the problem from #5991.

## Checklist

## Related issues

- Related #5991
- Relates #5835

### DIFF
--- a/internal/pkg/agent/transpiler/utils.go
+++ b/internal/pkg/agent/transpiler/utils.go
@@ -26,7 +26,7 @@ func RenderInputs(inputs Node, varsArray []*Vars) (Node, error) {
 	nodesMap := map[string]*Dict{}
 	for _, vars := range varsArray {
 		for _, node := range l.Value().([]Node) {
-			dict, ok := node.Clone().(*Dict)
+			dict, ok := node.(*Dict)
 			if !ok {
 				continue
 			}
@@ -34,6 +34,7 @@ func RenderInputs(inputs Node, varsArray []*Vars) (Node, error) {
 			if streams := getStreams(dict); streams != nil {
 				hadStreams = true
 			}
+			// Apply creates a new Node with a deep copy of all the values
 			n, err := dict.Apply(vars)
 			if errors.Is(err, ErrNoMatch) {
 				// has a variable that didn't exist, so we ignore it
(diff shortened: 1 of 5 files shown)


kaanyalti (approved):
Code changes look good to me, but I don't have much context here. It would be better for someone else to look at it as well.

internal/pkg/agent/transpiler/utils.go
```diff
@@ -26,14 +26,15 @@ func RenderInputs(inputs Node, varsArray []*Vars) (Node, error) {
...
-			dict, ok := node.Clone().(*Dict)
+			dict, ok := node.(*Dict)
 			if !ok {
 				continue
 			}
 			hadStreams := false
 			if streams := getStreams(dict); streams != nil {
 				hadStreams = true
 			}
+			// Apply creates a new Node with a deep copy of all the values
```
blakerouse, line 37: Thanks for the comment here, helps with understanding that Apply is going to return a copy.

blakerouse (approved):
Nice improvement! Looks good.

PR #6192, opened 2024-12-03
Use xxhash instead of sha256 for hashing AST nodes

swiatekm:
## What does this PR do?

We currently use sha256 for hashing AST nodes when generating configuration. This is used both for checking equality between AST instances and for avoiding recomputing inputs unnecessarily. This PR changes this to xxHash, which is much faster. It also changes the interface, letting the caller supply their own hasher instance, and avoiding needing to allocate one for each Node in the tree.

Of note is that I'm using the `xxhash.Digest` struct as an argument instead of the generic `hash.Hash64` interface from the standard library. The reason is that the former has an optimized `WriteString` method, which avoid needing to cast the string to a byte slice. I also don't expect to need to actually supply different hash implementations to this method.

Note also that there's already fairly exhaustive tests for this, some unexpected behaviour included.

## Why is it important?

It's a pretty significant performance improvement. See the benchstat results using the benchmark from #6180:

```
goos: linux
goarch: amd64
                                      │ bench_main.txt │           bench_hash.txt            │
                                      │     sec/op     │    sec/op     vs base               │
Coordinator_generateComponentModel-20     37.99m ± 24%   34.29m ± 14%  -9.73% (p=0.009 n=10)

                                      │ bench_main.txt │           bench_hash.txt            │
                                      │      B/op      │     B/op      vs base               │
Coordinator_generateComponentModel-20     34.22Mi ± 0%   31.53Mi ± 0%  -7.88% (p=0.000 n=10)

                                      │ bench_main.txt │           bench_hash.txt            │
Coordinator_generateComponentModel-20      810.7k ± 0%   713.0k ± 0%  -12.05% (p=0.000 n=10)
```

## Checklist

## Related issues

- Relates #5835 
- Related #5991

### DIFF
--- a/internal/pkg/agent/transpiler/ast.go
+++ b/internal/pkg/agent/transpiler/ast.go
@@ -5,7 +5,6 @@
 package transpiler
 
 import (
-	"bytes"
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
@@ -13,6 +12,8 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+
+	"github.com/cespare/xxhash/v2"
 
 	"github.com/elastic/elastic-agent/internal/pkg/eql"
 )
@@ -57,6 +58,9 @@ type Node interface {
 
 	// Hash compute a sha256 hash of the current node and recursively call any children.
 	Hash() []byte
+
+	// Hash64With recursively computes the given hash for the Node and its children
+	Hash64With(h *xxhash.Digest) error
 
 	// Vars adds to the array with the variables identified in the node. Returns the array in-case
 	// the capacity of the array had to be changed.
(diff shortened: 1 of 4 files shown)


andrzej-stencel (approved):

internal/pkg/agent/transpiler/ast.go
```diff
@@ -58,6 +59,9 @@ type Node interface {
 	// Hash compute a sha256 hash of the current node and recursively call any children.
 	Hash() []byte
 
+	// Hash64With recursively computes the given hash for the Node and its children
+	Hash64With(h *xxhash.Digest) error
```
blakerouse, line 63: Why not just replace the usage of `Hash()`? Why do you have both? Do we need both?

swiatekm: I don't think we need both. If there were a bunch of callers of `Hash`, then I'd be in favor of keeping it and implementing it using `Hash64With`, but there aren't, so we can get rid of it. I was planning to do so in a follow up.

blakerouse (approved):
Okay, removal in follow-up is good.

PR #6218, opened 2024-12-04
Remove SonarQube badge from README

ycombinator:
This PR removes the currently-broken SonarQube badge from the Elastic Agent README file.  

Elastic's SonarQube instance is private (hosted at sonar.elastic.dev instead of the public sonarcloud.io).  As such a security token is needed as part of the badge image URL.  Since the elastic-agent repository is public, it would not be a good idea to expose this token.  So we remove the badge instead of trying to replace it.


swiatekm (approved):

jlind23:
> SonarQube Code Analysis Expected — Waiting for status to be reported
This will never be reported as we are not changing any code hence force merging this.

PR #6228, opened 2024-12-05
otel: add test for filebeatreceiver delivery guarantees during a collector restart

mauri870:
## What does this PR do?

This commit adds a integration test that restarts the collector a couple times during the log ingestion and confirms that the data was fully written to elasticsearch.

## Checklist

## Related issues

- For https://github.com/elastic/ingest-dev/issues/3677


testing/integration/otel_test.go
```diff
@@ -1099,3 +1100,186 @@ service:
...
+
+	var fixtureWg sync.WaitGroup
+	fixtureWg.Add(1)
+	go func() {
+		defer fixtureWg.Done()
+
+		for {
+			if ingestDone.Load() {
+				break
+			}
```
leehinman, line 1244: I know it isn't likely, but it is possible that the go routine that is writing the log files could be done on the first iteration.

I'm wondering if we should have a go routine that always writes, but exposes a counter of how many have been written.  Then the go routine that is restarting can make sure that this increases between restarts.  And then at the end we stop the log writer and make sure that however many it did write end up in Elasticsearch.

mauri870: Thanks, I updated the logic to check that line count increases between restarts and I also asserted that the collector restarts at least once during the test.

leehinman (approved):
LGTM

testing/integration/otel_test.go
```diff
@@ -1099,3 +1100,191 @@ service:
...
+			findCtx, findCancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer findCancel()
+
+			docs, err := estools.GetLogsForIndexWithContext(findCtx, info.ESClient, ".ds-"+index+"*", map[string]interface{}{
+				"log.file.path": inputFilePath,
+			})
+			require.NoError(t, err)
+
+			actualHits.Hits = docs.Hits.Total.Value
+			return actualHits.Hits == int(inputLinesCounter.Load())
```
cmacknz, line 1282: This doesn't guarantee we read each unique log line, only that we read an amount of log lines equal to the number that was written, regardless of if some of them are duplicates?

The test we need is "every log line in the file was shipped at least once" and I don't think this is that.

If you gave the log lines you are shipping an _id field Elasticsearch would deduplicate them for us and then the count would be enough. There is a blog post explaining this process and some sample configurations for Beats and Elastic Agent https://www.elastic.co/blog/log-deduplication-with-elasticsearch

testing/integration/otel_test.go
```diff
@@ -1099,3 +1100,191 @@ service:
...
+			}
+		}
+
+		// start the collector again for the remaining of the test
+		fCtx, cancel := context.WithDeadline(ctx, time.Now().Add(5*time.Minute))
+		defer cancel()
+		err = fixture.RunOtelWithClient(fCtx)
+	}()
+
+	time.Sleep(5 * time.Second)
```
cmacknz, line 1264: This might be better as a counter of the number of times the collector has restarted. Then you only need to maintain a single arbitrary time interval, the delay between restarts.

cmacknz: Instead of a require.True that restartCount is greater than zero, you would just delay until it cross a threshold.

Perhaps better ways to do that than using an atomic and busy waiting on it as well: number of writes to a channel, condition variable, etc.

testing/integration/otel_test.go
```diff
@@ -1099,3 +1100,191 @@ service:
...
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
+	index := "logs-integration-default"
+	otelConfigTemplate := `receivers:
+  filebeatreceiver:
+    filebeat:
+      inputs:
+        - type: filestream
```
cmacknz, line 1145: This is perhaps a bit beyond the scope of this test, but it would be interesting to confirm that the registry is actually updating as expected and we aren't re-reading the file from the beginning every time we restart.

This would validate that acknowledgements flow back to the receiver properly.

This would perhaps best be done as a separate test that writes a file, confirms we read it all, writes a single additional line, then confirms only that one line is resent.

mauri870: Acknowledged. Agree that this should be a follow up test.

testing/integration/otel_test.go
mauri870: Thanks for the input. I think this test should be refactored to use  the dynamic id in the elasticsearchexporter once it is available (collector v0.119.0), this would align to how we tackle deduplication in beats today (the link you posted).

For now, idea suggested by Lee is to use a unique field in the log message to verify we only ingested log lines once. We can then count the number of documents and it should match the number of lines. Additionally, the unique field in each log line must be present in a matching document.

testing/integration/otel_test.go
mauri870: Thanks, I have updated the logic.

leehinman (approved):
LGTM

Thank You

testing/integration/otel_test.go
mauri870: Would you mind giving the code here another look to see if it aligns with what you had in mind? Thanks.

testing/integration/otel_test.go
```diff
@@ -1099,3 +1100,245 @@ service:
...
+	restartCh := make(chan int, 1)
+	go func() {
+		defer fixtureWg.Done()
+		var restartCount int
+
+		for {
+			// restart the collector every couple seconds while new data is being written
+			prevCount := inputLinesCounter.Load()
+			fCtx, cancel := context.WithDeadline(ctx, time.Now().Add(2*time.Second))
+			err = fixture.RunOtelWithClient(fCtx)
```
cmacknz, line 1261: This assumes that the collector can start up and being ingesting log lines with 2 seconds. Probably it can, but this isn't guaranteed in this test.

The test would be stronger if you had a way to observe that the collector had read and shipped at least one log line to ES before restarting it.

cmacknz: That said, this test has reached "good enough" if there are other things to make progress on.

mauri870: Thanks, I just made changes following your guidelines.

cmacknz (approved):

testing/integration/otel_test.go
cmacknz: Thanks for following up!

PR #6229, opened 2024-12-05
Cache conditions when applying variables to config

swiatekm:
## What does this PR do?

Our configurations can have conditional sections. The conditions are expressed in an EQL-like syntax, and we parse the expressions lazily on evaluation. However, we don't cache the parsed expression, and parse it every time it's evaluated. This change instead only parses the condition expression once, and it's then cached in the AST node.

## Why is it important?

When there are a lot of variables from dynamic providers - notably in Kubernetes, on a Node with a lot of Pods - we end up spending significant resources on re-parsing the condition expressions for every var entry.

See benchstat showing the effect of this change on the benchmark from #6180 (not that this includes #6184, as otherwise we just make a copy before applying and never actually use the cache):

```
goos: linux
goarch: amd64
                                      │ bench_noclone.txt │            bench_eql.txt            │
                                      │      sec/op       │   sec/op     vs base                │
Coordinator_generateComponentModel-20         32.36m ± 4%   27.45m ± 2%  -15.17% (p=0.000 n=10)

                                      │ bench_noclone.txt │            bench_eql.txt             │
                                      │       B/op        │     B/op      vs base                │
Coordinator_generateComponentModel-20        25.38Mi ± 0%   20.51Mi ± 0%  -19.17% (p=0.000 n=10)

                                      │ bench_noclone.txt │            bench_eql.txt            │
                                      │     allocs/op     │  allocs/op   vs base                │
Coordinator_generateComponentModel-20         580.4k ± 0%   490.1k ± 0%  -15.57% (p=0.000 n=10)
```

## Checklist

Unit tests are sufficient, the benchmark in #6180 helps as well.

## Related issues

- Relates #5835
- Relates #5991

### DIFF
--- a/internal/pkg/agent/transpiler/ast_test.go
+++ b/internal/pkg/agent/transpiler/ast_test.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/elastic/elastic-agent-libs/mapstr"
 
+	"github.com/elastic/elastic-agent/internal/pkg/eql"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


internal/pkg/agent/transpiler/ast_test.go
```diff
@@ -1071,6 +1073,29 @@ func TestLookup(t *testing.T) {
 	}
 }
 
+func TestCondition(t *testing.T) {
+	vars := mustMakeVars(map[string]interface{}{
+		"other": map[string]interface{}{
+			"data": "info",
+		}})
+
+	input := NewKey("condition", NewStrVal("${other.data} == 'info'"))
```
cmacknz, line 1082: There should be a test for the case where the condition is edited for the same input.

As an implementation question does the cache need to be invalidated when this happens, or is a new AST node created so that we don't need to care?

blakerouse (approved):
Looks good.

internal/pkg/agent/transpiler/ast_test.go
swiatekm: The ability to modify an AST is very limited to begin with. The only public API to do it is the `Insert` function, which only works on Dicts and Lists. I can add a test where I replace a key in a dict and verify the condition doesn't get saved.

swiatekm: https://github.com/elastic/elastic-agent/pull/6229/commits/b5ac1c4176c7f423468bf191508f527fc81a2240

cmacknz (approved):

PR #6273, opened 2024-12-10
Log warning on same version upgrade to prevent failed status

### LINKED ISSUE #6186: 8.13.0+ Fleet managed agents upgraded to the currently running version report failed upgrades until a successful upgrade occurs
8.13.0+ Fleet managed agents upgraded to the currently running version report failed upgrades indefinitely until a successful upgrade occurs. If the agent is running the latest version at the time of the duplicate upgrade attempt, it will stay in the upgrade failed state until the next version is released as there is no newer version to upgrade to.

Prior to v8.13.0 we logged a warning when instructed to upgrade to the currently running version: https://github.com/elastic/elastic-agent/blob/v8.12.2/internal/pkg/agent/application/upgrade/upgrade.go#L182-L184

```go
	if strings.HasPrefix(release.Commit(), newHash) {
		u.log.Warn("Upgrade action skipped: upgrade did not occur because its the same version")
		return nil, nil
	}
```

In v8.13.0 the agent upgrade process was refactored to account for the situation where the agent package version changed, but the agent binary version and commit hash was unchanged. When this was done an attempt to upgrade to the same version was changed to return an error. This error ultimately ends up in the upgrade details reported to fleet, but because there is no alternative upgrade path if you are upgrading to the currently running version and that version is the latest, the upgrade details persist the failed state indefinitely.

https://github.com/elastic/elastic-agent/blob/56901edc3f2c6d1fc0c1271fad374cdbb24ed7f2/internal/pkg/agent/application/upgrade/upgrade.go#L215-L218

There are two improvements to make here:
1. We make upgrades to the currently running version a no-op and implicitly successful. The upgrade process becomes idempotent in this case. Logging a warning was the correct approach in the original implementation.
2. We should move this check before the point in time where we download the agent package when it is possible. For official stack releases and independent agent releases, the version in the downloaded agent artifact is always unique between versions. The edge case and exception is snapshots which can have the same version (e.g. 8.16.0-SNAPSHOT) but be a different build, in which case the snapshot ID will be different once it is obtained.

michel-laterman:
## What does this PR do?

Log warning on same version upgrade and filter error to prevent upgrade status reporting.
If a user attempts to upgrade to the same version, the `elastic-agent upgrade` call will exit with an error, but the agent's status will be cleared of any upgrade markers.

Include an earlier check for same-version upgrades before artifacts are downloaded to prevent unnecessary downloads

## Why is it important?

Currently upgrading to the same version results in the agent reporting a failed state until a successful upgrade can be done.

## Testing instructions

Install a stand-alone agent and use `elastic-agent upgrade 9.0.0-SNAPSHOT --source-uri file://path/to/dir --skip-verification` to upgrade the agent

For example:
```
# Install the agent
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo ./elastic-agent install --develop
Elastic Agent will be installed at /Library/Elastic/Agent-Development and will run as a service. Do you want to continue? [Y/n]:y
[   =] Service Started  [4s] Elastic Agent successfully installed, starting enrollment.
[   =] Done  [4s]

# Check status and version
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent version
Binary: 9.0.0-SNAPSHOT (build: 14e78753e6c06295e3348a00de01ae54905fb2c0 at 2024-12-18 18:30:32 +0000 UTC)
Daemon: 9.0.0-SNAPSHOT (build: 14e78753e6c06295e3348a00de01ae54905fb2c0 at 2024-12-18 18:30:32 +0000 UTC)
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent status
┌─ fleet
└─ elastic-agent
   └─ status: (HEALTHY) Running

# attempt to upgrade to the same version
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent upgrade 9.0.0-SNAPSHOT --source-uri file:///Users/mlaterman/git/elastic-agent/build/distributions/ --skip-verify
Upgrade triggered to version 9.0.0-SNAPSHOT, Elastic Agent is currently restarting
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ echo $?
0
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent status
┌─ fleet
└─ elastic-agent
   └─ status: (HEALTHY) Running
elastic-agent-9.0.0-SNAPSHOT-darwin-aarch64|⇒ sudo elastic-development-agent version
Binary: 9.0.0-SNAPSHOT (build: 14e78753e6c06295e3348a00de01ae54905fb2c0 at 2024-12-18 18:30:32 +0000 UTC)
Daemon: 9.0.0-SNAPSHOT (build: 14e78753e6c06295e3348a00de01ae54905fb2c0 at 2024-12-18 18:30:32 +0000 UTC)
```

## Checklist

## Disruptive User Impact

N/A

## Related issues

- Closes #6186


internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -162,6 +163,19 @@ func (av agentVersion) String() string {
...
+	currentVersion := agentVersion{
+		version:  release.Version(),
+		snapshot: release.Snapshot(),
+		hash:     release.Commit(),
+	}
+
+	// check version before download
+	same, _ := isSameVersion(u.log, currentVersion, packageMetadata{}, version)
+	if same {
+		u.log.Warnf("Upgrade action skipped: %v", ErrUpgradeSameVersion)
```
pchila, line 175: Please log the version we tried to upgrade to, something like 
```suggestion
		u.log.Warnf("Upgrade action skipped because agent is already at version %s", currentVersion)
```

Same at line 226

testing/integration/upgrade_standalone_same_commit_test.go
```diff
@@ -74,7 +74,7 @@ func TestStandaloneUpgradeSameCommit(t *testing.T) {
 			upgradetest.WithUnprivileged(unprivilegedAvailable),
 			upgradetest.WithDisableHashCheck(true),
 		)
-		assert.ErrorContainsf(t, err, fmt.Sprintf("agent version is already %s", currentVersion), "upgrade should fail indicating we are already at the same version")
+		assert.ErrorContains(t, err, "failed to find watcher:")
```
pchila, line 77: Why is it expected to have an error "failed to find watcher:" here? Shouldn't we assert that we end up with a "Completed " upgrade and no error reported in the upgradeDetails ?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -162,6 +163,19 @@ func (av agentVersion) String() string {
...
 	u.log.Infow("Upgrading agent", "version", version, "source_uri", sourceURI)
 
+	currentVersion := agentVersion{
+		version:  release.Version(),
+		snapshot: release.Snapshot(),
+		hash:     release.Commit(),
+	}
+
+	// check version before download
+	same, _ := isSameVersion(u.log, currentVersion, packageMetadata{}, version)
```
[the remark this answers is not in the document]
michel-laterman: Yes, but it should work for non-snapshot attempts, right?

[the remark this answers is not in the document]
michalpristas: @pchila  so in this case this will fail because hash is unpopulated right? but then later it is extracted and checked again on line 224. so we lose a bit of time but will not perform upgrade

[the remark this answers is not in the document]
pchila: Here we are relying on the fact that the current version has a not empty hash compared to the zero value in `packageMetadata{}`  so I don't see how it can ever detect this to be the same version if the has in packageMetadata is ""

I guess we will never get true from here... Did you test this ?

testing/upgradetest/upgrader.go
```diff
@@ -332,11 +332,15 @@ func PerformUpgrade(
 
 	upgradeOutput, err := startFixture.Exec(ctx, upgradeCmdArgs)
 	if err != nil {
+		logger.Logf("Upgrade err: %v, output: %s", err, upgradeOutput)
 		// Sometimes the gRPC server shuts down before replying to the command which is expected
 		// we can determine this state by the EOF error coming from the server.
 		// If the server is just unavailable/not running, we should not succeed.
 		// Starting with version 8.13.2, this is handled by the upgrade command itself.
 		outputString := string(upgradeOutput)
+		if strings.Contains(outputString, "upgrade did not occur because it is the same version") {
```
pchila, line 341: Why is the check and the specific error needed here ?

michel-laterman: If we don't check the output here then our integration tests will wait for the watcher to start (which fails)

internal/pkg/agent/application/upgrade/upgrade.go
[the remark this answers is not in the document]
michel-laterman: you're correct, I forgot that comparing currentVersion vs version would also compare the hashes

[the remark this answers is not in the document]
michel-laterman: I've added a new func that does this comparison correctly.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -575,6 +575,11 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 	cb, err := c.upgradeMgr.Upgrade(ctx, version, sourceURI, action, det, skipVerifyOverride, skipDefaultPgp, pgpBytes...)
 	if err != nil {
 		c.ClearOverrideState()
+		if errors.Is(err, upgrade.ErrUpgradeSameVersion) {
+			// Set upgrade state to completed, but return an error if a same version-upgrade is attempted.
+			det.SetState(details.StateCompleted)
+			return err
```
cmacknz, line 581: I don't think we want to consider this an error, that is what lead to the bug this is fixing in the first place. It needs to be like an idempotent API call. It returns success if the action has already completed successfully once.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -626,3 +633,17 @@ func IsInProgress(c client.Client, watcherPIDsFetcher func() ([]int, error)) (bo
...
 	return state.State == cproto.State_UPGRADING, nil
 }
+
+// isSameReleaseVersion will if upgradeVersion and currentVersion are equal using only release numbers.
+// They are not equal if either are a SNAPSHOT, or if the semver numbers (including prerelease and build identifiers) differ.
+func isSameReleaseVersion(current agentVersion, upgradeVersion string) bool {
+	if current.snapshot {
+		return false
+	}
+	target, _ := agtversion.ParseVersion(upgradeVersion)
```
pchila, line 643: This will be the first time that the upgradeVersion is parsed, please handle the error.
`IsSameVersion` had an explicit comment about being able to skip the error of parsing the version string again because it was already parsed and handled in lines 198-201 but here we have no such guarantee as this is called on line 172

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -626,3 +633,17 @@ func IsInProgress(c client.Client, watcherPIDsFetcher func() ([]int, error)) (bo
 
 	return state.State == cproto.State_UPGRADING, nil
 }
+
+// isSameReleaseVersion will if upgradeVersion and currentVersion are equal using only release numbers.
+// They are not equal if either are a SNAPSHOT, or if the semver numbers (including prerelease and build identifiers) differ.
```
pchila, line 638:
```suggestion
// isSameReleaseVersion will return true if upgradeVersion and currentVersion are equal using only release numbers and SNAPSHOT prerelease qualifiers.
// They are not equal if either are a SNAPSHOT, or if the semver numbers (including prerelease and build identifiers) differ.
```

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -983,3 +1000,60 @@ func Test_selectWatcherExecutable(t *testing.T) {
...
+		expect: false,
+	}, {
+		name: "target version is different version",
+		current: agentVersion{
+			version: "1.2.3",
+		},
+		target: "1.2.4",
+		expect: false,
+	}, {
+		name: "target version is same with pre-release",
```
pchila, line 1033: Maybe we should avoid "is same" in testcase names for a function that is called `isSame*` where the expected result is in fact that those versions are *not* the same
```suggestion
		name: "target version has same major.minor.patch, different pre-release",
```

testing/integration/upgrade_standalone_same_commit_test.go
```diff
@@ -74,7 +74,7 @@ func TestStandaloneUpgradeSameCommit(t *testing.T) {
 			upgradetest.WithUnprivileged(unprivilegedAvailable),
 			upgradetest.WithDisableHashCheck(true),
 		)
-		assert.ErrorContainsf(t, err, fmt.Sprintf("agent version is already %s", currentVersion), "upgrade should fail indicating we are already at the same version")
+		assert.NoError(t, err)
```
pchila, line 77: Please add a message to explain why we expect no error in this case (otherwise this test is functionally equal to the next one in line 80)

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -206,15 +218,10 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
-		version:  release.Version(),
-		snapshot: release.Snapshot(),
-		hash:     release.Commit(),
-	}
-
 	same, newVersion := isSameVersion(u.log, currentVersion, metadata, version)
 	if same {
-		return nil, fmt.Errorf("agent version is already %s", currentVersion)
+		u.log.Warnf("Upgrade action skipped because agent is already at version %s", currentVersion)
+		return nil, ErrUpgradeSameVersion
```
swiatekm, line 224: As someone new to this part of the codebase, a comment here explaining why we're doing a second version check would help a lot.

swiatekm (approved):
LGTM, but someone with more familiarity with the domain should review as well before merging.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -575,6 +575,12 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 	cb, err := c.upgradeMgr.Upgrade(ctx, version, sourceURI, action, det, skipVerifyOverride, skipDefaultPgp, pgpBytes...)
 	if err != nil {
 		c.ClearOverrideState()
+		if errors.Is(err, upgrade.ErrUpgradeSameVersion) {
+			c.logger.Info("Same ver upgrade detected.")
```
michalpristas, line 579: dont't we end up with 2 log levels for same thing one Warning coming from `internal/pkg/agent/application/upgrade/upgrade.go:L175` the other one here

michel-laterman:
buildkite test this

Issue #5660, opened 2024-10-02
Include hints autodiscovery configuration inside integration tests for k8s


gizas:
Currently Elastic Agent has the following integration tests for K8s: https://github.com/elastic/elastic-agent/blob/main/testing/integration/kubernetes_agent_standalone_test.go#L63

As part of the ksutomize template enhancement we added the hints autodiscovery template here: https://github.com/elastic/elastic-agent/pull/5643

It is a good idea to align the tests with what we have in the ksutomize templates

PR #6277, opened 2024-12-11
[k8s]: transition integration tests to adapter pattern

pkoutsovasilis:
## What does this PR do?

This PR continues the groundwork for Kubernetes hints integration tests. The next and last episode of this PR-series can be seen here https://github.com/elastic/elastic-agent/commit/2ea1f2ff2cdf57442ea29c1e5c269b907971c312. Again these were split up because I want to keep each PR around [500 lines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md?plain=1#L47). 

Key changes in this PR:
1. **Transition Kubernetes integration tests to an "adapter" pattern**:
   - Refactors Kubernetes integration tests to follow a more modular and reusable adapter pattern.
   - Introduces `k8sTestStep` functions that encapsulate individual test steps for clarity and flexibility.
   - Simplifies the test flow by eliminating repetitive code and leveraging reusable functions like `k8sStepCreateNamespace`, `k8sStepDeployKustomize`, and `k8sStepRunInnerTests`.
   - Improves readability and maintainability.

2. **Align all kubernetes integration tests**:
   - All kubernetes integration tests got an empty for now `skipReason`.
   - All kubernetes integration tests check for schedulable k8s nodes and got more deterministic in the number of agent pods expected to be running.

## Why is it important?

The adapter pattern for Kubernetes integration tests enhances readability, maintainability, making it easier to add or modify test cases.

## Checklist

- ~~[ ] I have made corresponding changes to the default configuration files~~

## Disruptive User Impact

- None expected. The changes affect only the internal kubernetes integration tests without altering user-facing configurations or behaviours.

```
mage integration:auth
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/5660

### DIFF
--- a/testing/integration/kubernetes_agent_standalone_test.go
+++ b/testing/integration/kubernetes_agent_standalone_test.go
@@ -77,64 +77,87 @@ func TestKubernetesAgentStandaloneKustomize(t *testing.T) {
 		Group: define.Kubernetes,
 	})
 
+	ctx := context.Background()
 	kCtx := k8sGetContext(t, info)
 
-	renderedManifest, err := renderKustomize(agentK8SKustomize)
-	require.NoError(t, err, "failed to render kustomize")
+	nodeList := corev1.NodeList{}
+	err := kCtx.client.Resources().List(ctx, &nodeList)
+	require.NoError(t, err)
+
+	totalK8SNodes := len(nodeList.Items)
+	require.NotZero(t, totalK8SNodes, "No Kubernetes nodes found")
 
 	testCases := []struct {
-		name             string
-		runUser          *int64
-		runGroup         *int64
-		capabilitiesDrop []corev1.Capability
-		capabilitiesAdd  []corev1.Capability
-		runK8SInnerTests bool
-		skipReason       string
+		name       string
+		skipReason string
+		steps      []k8sTestStep
 	}{
 		{
-			"default deployment - rootful agent",
-			nil,
-			nil,
-			nil,
-			nil,
-			false,
-			"",
+			name: "default deployment - rootful agent",
+			steps: []k8sTestStep{
+				k8sStepCreateNamespace(),
+				k8sStepDeployKustomize(agentK8SKustomize, "elastic-agent-standalone", k8sKustomizeOverrides{
+					agentContainerMemoryLimit: "800Mi",
+				}, nil),
+				k8sStepCheckAgentStatus("app=elastic-agent-standalone", totalK8SNodes, "elastic-agent-standalone", nil),
+			},
 		},
 		{
-			"drop ALL capabilities - rootful agent",
-			int64Ptr(0),
-			nil,
-			[]corev1.Capability{"ALL"},
-			[]corev1.Capability{},
-			false,
-			"",
+			name: "drop ALL capabilities - rootful agent",
+			steps: []k8sTestStep{
+				k8sStepCreateNamespace(),
+				k8sStepDeployKustomize(agentK8SKustomize, "elastic-agent-standalone", k8sKustomizeOverrides{
+					agentContainerRunUser:          int64Ptr(0),
+					agentContainerCapabilitiesAdd:  []corev1.Capability{},
+					agentContainerCapabilitiesDrop: []corev1.Capability{"ALL"},
+					agentContainerMemoryLimit:      "800Mi",
+				}, nil),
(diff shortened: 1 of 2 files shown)


testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -77,64 +77,87 @@ func TestKubernetesAgentStandaloneKustomize(t *testing.T) {
...
+	ctx := context.Background()
 	kCtx := k8sGetContext(t, info)
 
-	renderedManifest, err := renderKustomize(agentK8SKustomize)
-	require.NoError(t, err, "failed to render kustomize")
+	nodeList := corev1.NodeList{}
+	err := kCtx.client.Resources().List(ctx, &nodeList)
+	require.NoError(t, err)
+
+	totalK8SNodes := len(nodeList.Items)
```
pchila, line 87: Take care of filtering nodes here: we need to take out the tainted nodes with a `NoSchedule` effect  if agent does not have a toleration for that (ex. https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane-taint for control-plane nodes)

pkoutsovasilis: more than a valid point! thanks for catching that @pchila 🙂

pkoutsovasilis:
@blakerouse @swiatekm could you give this PR a review? 🙏

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -1118,3 +934,233 @@ func k8sGetContext(t *testing.T, info *define.Info) k8sContext {
...
+func k8sStepCreateNamespace() k8sTestStep {
+	return func(t *testing.T, ctx context.Context, kCtx k8sContext, namespace string) {
+		k8sNamespace := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespace,
+			},
+		}
+
+		t.Cleanup(func() {
+			err := k8sDeleteObjects(ctx, kCtx.client, k8sDeleteOpts{wait: true}, k8sNamespace)
```
swiatekm, line 952: Maybe worth skipping finalizers here? Deleting namespaces can sometimes take a long time.

swiatekm (commented):
Looks good to me overall, left some comments.

testing/integration/kubernetes_agent_standalone_test.go
pkoutsovasilis: I added this cleanup to accommodate running the kubernetes integration tests locally. In CI our env is provisioned at every CI run. But locally the k8s cluster might and most probably be already there. Thus if we don't cleanup the namespace we are gonna get an error at the creation of the next run. We could always ignore error exists but I don't consider that the best idea. Also because the namespace here is one that only the test deploys the corresponding k8s objects which are cleaned up in other steps, it's not gonna be that slow most of the times it is empty. And as an extra delay minimisation here the default timeout for waiting to delete k8s object is just 20 secs. Lastly we never fail a test because of an error at delete. Thus I would propose to leave this delete. Does it make sense? 🙂

pkoutsovasilis: > then making the namespace unique per test run is better imo.

It is already unique across the different tests but if you rerun the same test it will get the same namespace (it derives from the test name). And now that I think about it again, we do need that cleanup for our CI as well, because we test different docker variants on the same k8s cluster. That said, I share similar experiences to you, but whenever I resorted to force delete a namespace it was never faster than normal delete and the reason behind a namespace being slow at delete, is weird finalisers at the objects contained in this namespace, which we don't have. So coming back to my original opinion I believe this delete should stay here 🙂

swiatekm: Allright, how about attaching a timestamp to the namespace name then? Or even https://github.com/dustinkirkland/golang-petname if we want to be fancy. Would be nice if `testing.T` exposed a run id, but unfortunately no such luck.

pchila: I see no problem in cleaning up the test environment when the test is done, unless we have metrics showing that this is heavily impacting our test run time

pchila (approved):
Looks good, test are more readable by refactoring common boilerplate in separate functions.

swiatekm (approved):

Issue #5485, opened 2024-09-09
Mark the Elastic Agent Helm Chart as beta


ycombinator:
**Describe the enhancement:**

In https://github.com/elastic/elastic-agent/pull/5331, we added a Helm Chart for Elastic Agent.  For the moment, we would like to mark this chart as beta.

**Describe a specific use case for the enhancement or feature:**

To set appropriate expectations around stability and supportability of the Helm Chart with its users.

**What is the definition of done?**

TBD — we need to first decide how best to mark the chart as beta.

Related: https://github.com/elastic/elastic-agent/issues/5486

ycombinator:
_Copied from https://github.com/elastic/elastic-agent/pull/5331#issuecomment-2338320617, where @pkoutsovasilis said:_

Just thinking out loud the available options here:

1. I assume that we can push the helm chart with a beta suffix at the version and update the NOTES.txt accordingly here https://helm.elastic.co/
2. Somebody that wants to try this helm chart can clone the repo and install the helm chart from the folder `deploy/helm/elastic-agent`
3. We could utilise github pages on this repo and commit directly the helm chart package.tgz (that points to the SNAPSHOT agent image)  [my least favourite option]

that's why a discussion about the Helm chart release process is wise to have before any next step 🙂

ycombinator:
_Copied from https://github.com/elastic/elastic-agent/pull/5331#issuecomment-2338484088, where @swiatekm said:_

Helm Charts use [semver 2.0](https://semver.org/) for versioning, so it's simple enough to just designate our releases as betas by setting it to `va.b.c-beta.d` or something similar. This makes Helm automatically treat it as a prerelease. This will cause them not to show up in searches without the `--devel` flag, amongst other consequences.

ycombinator:
cc: @cmacknz

swiatekm:
I like @pkoutsovasilis's option 1. I don't see any reason to make things complicated if we already have a Helm repo we can push to.

ycombinator:
Seems like we have consensus on this option:

> push the helm chart with a beta suffix at the version and update the NOTES.txt accordingly here https://helm.elastic.co/

@pkoutsovasilis Could you please close this issue when the above is done?

Issue #5486, opened 2024-09-09
[Discuss] Elastic Agent Helm Chart release process


ycombinator:
In https://github.com/elastic/elastic-agent/pull/5331, we added a Helm Chart for Elastic Agent.  At the time, we [deferred](https://github.com/elastic/elastic-agent/pull/5331#discussion_r1742638157) the discussion on how this chart should be released.  Let's continue the discussion in this issue and try to arrive at a decision.

ycombinator:
cc: @pkoutsovasilis @swiatekm @blakerouse @cmacknz

swiatekm:
As I see it, we have two basic choices here:

1. Release a Helm Chart version every time we release an agent version. The default agent image version is set to the Chart version.
2. Release on a different schedule. There's either no default agent image (user needs to set it themself), or there's a default, but it may not be the latest agent image.

The important questions that should guide this choice are:

1. How often do we expect to have semantic changes to the Chart, not related to the agent version?
2. Are we ok with the user having to set the agent image?
  a) If we want a default, are we ok with the latest Chart version *not* having the latest agent version as the default?
  b) Are we confident in users' ability to upgrade the agent image independently of the Helm Chart version?
3. Is it important for users to be able to use the Chart with agent versions older than the first Chart release?
4. Is it a pain to release the Chart in lockstep with agent?

I don't feel like I know enough about the domain to have informed opinions about these myself. I think I'd default to option 1 just because it's simple, but it does sacrifice some flexibility.

swiatekm:
@pkoutsovasilis do you mean literally just `main`, or also the release branches? If we don't do the latter, we won't have Helm Chart releases for agent patch versions.

pkoutsovasilis:
@swiatekm I was thinking `main` but sure this can be adjusted to any branch if we deem such a wiser choice. The way I am thinking of it 

> imo the helm chart should support the user specifying different agent versions/images but provide guarantees of nominal operation only with the default image that comes with it

the helm chart will support agent versions incrementally thus when there is a patch version to the latest agent release I see no issue creating a PR that points the aforementioned `version.json` to the patch version and triggering all the necessary CI steps. But maybe I am missing something, so please do elaborate 😃

cmacknz:
> Changes to the built-in kubernetes integration as it mostly reflects [this package](https://github.com/elastic/integrations/tree/main/packages/kubernetes)

[...]

PR #6293, opened 2024-12-12
chore: update version and appVersion in the elastic-agent helm chart

pkoutsovasilis:
## What does this PR do?

This PR extends the logic of `mage helm:updateAgentVersion` to include updating the `version` (chart version) and `appVersion` keys in the Chart.yaml of the Elastic-Agent Helm chart. Note that since the chart is not yet GA for the chart version we always append `-SNAPSHOT` to mark it as devel. This will have to change when we reach the version that the Helm chart is GA.

## Why is it important?

It is important because it sets correctly the chart and the app version in the Chart.yaml of the Elastic-Agent Helm chart and this is required before being able to release the Helm chart.

## Checklist

## Disruptive User Impact

N/A

change either or both `version` and `appVersion` in `deploy/helm/elastic-agent/Chart.yaml` and then run `mage helm:updateAgentVersion`. The two fields will get updated to reflect the correct values based on the branch you are.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5485
- Relates https://github.com/elastic/elastic-agent/issues/5486

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -114,25 +114,6 @@ var (
 	goIntegTestTimeout = 2 * time.Hour
 	// goProvisionAndTestTimeout is the timeout used for both provisioning and running tests.
 	goProvisionAndTestTimeout = goIntegTestTimeout + 30*time.Minute
-
-	helmChartsValues = []struct {
-		path        string
-		versionKeys []string
-		tagKeys     []string
-	}{
-		// elastic-agent Helm Chart
-		{
-			helmChartPath,
-			[]string{"agent", "version"},
-			[]string{"agent", "image", "tag"},
-		},
-		// edot-collector values file for kube-stack Helm Chart
-		{
-			helmOtelChartPath,
-			[]string{"defaultCRConfig", "image", "tag"},
-			nil,
-		},
-	}
 )
 
 func init() {
@@ -3423,6 +3404,9 @@ type otelDependencies struct {
 
 type Helm mg.Namespace
 
+// RenderExamples runs the equivalent of `helm template` and `helm lint`
+// for the examples of the Elastic Helm chart which are located at
+// `deploy/helm/elastic-agent/examples` directory.
 func (Helm) RenderExamples() error {
 	settings := cli.New() // Helm CLI settings
 	actionConfig := &action.Configuration{}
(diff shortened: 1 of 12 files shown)


magefile.go
```diff
@@ -3503,71 +3487,43 @@ func (Helm) RenderExamples() error {
...
-			return fmt.Errorf("failed to check if tag contains commit: %w", err)
-		}
-
-		if !isTagged {
-			isTagged = os.Getenv(snapshotEnv) != ""
-		}
-
-		agentVersion := getVersion()
+	agentVersion := getVersion()
+	agentSnapshotVersion := agentVersion + "-SNAPSHOT"
```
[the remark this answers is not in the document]
pkoutsovasilis: ty @michalpristas! that's more than a valid point. I though that the `getVersion()` was always returning `major.minor.patch` version of the agent which is what I need here. Pushed an update that does that 🙂

swiatekm (commented):
The code changes look good to me. I had a question about the Chart versioning:

> Note that since the chart is not yet GA for the chart version we always append -SNAPSHOT to mark it as devel. This will have to change when we reach the version that the Helm chart is GA.

Do we plan to keep it this way while the Chart is in technical preview? Wouldn't that cause problems for users, who will have no way of distinguishing the Chart version for a given release (say `9.0.0`) from the dev one on main?

pkoutsovasilis:
> Do we plan to keep it this way while the Chart is in technical preview? Wouldn't that cause problems for users, who will have no way of distinguishing the Chart version for a given release (say `9.0.0`) from the dev one on main?

The dev Helm chart will have a version of `9.0.0-SNAPSHOT` (requires `--devel` flag to install it through helm) while the one that we will release will be `9.0.0`? Am I missing something? 🤔

swiatekm:
> > Do we plan to keep it this way while the Chart is in technical preview? Wouldn't that cause problems for users, who will have no way of distinguishing the Chart version for a given release (say `9.0.0`) from the dev one on main?
> 
> The dev Helm chart will have a version of `9.0.0-SNAPSHOT` (requires `--devel` flag to install it through helm) while the one that we will release will be `9.0.0`? Am I missing something? 🤔

Correct me if I'm wrong, but with the current changes in this PR, the Chart version will always have a `-SNAPSHOT` suffix. So we won't actually have a `9.0.0` version, just `9.0.0-SNAPSHOT`.

pkoutsovasilis:
> Correct me if I'm wrong, but with the current changes in this PR, the Chart version will always have a `-SNAPSHOT` suffix. So we won't actually have a `9.0.0` version, just `9.0.0-SNAPSHOT`.

you 100% correct, and before we release the version that the Helm chart needs to be GA another PR will open to remove this always "-SNAPSHOT" suffix. So then we will get non-devel Helm charts

swiatekm:
> > Correct me if I'm wrong, but with the current changes in this PR, the Chart version will always have a `-SNAPSHOT` suffix. So we won't actually have a `9.0.0` version, just `9.0.0-SNAPSHOT`.
> 
> you 100% correct, and before we release the version that the Helm chart needs to be GA another PR will open to remove this always "-SNAPSHOT" suffix. So then we will get non-devel Helm charts

Right, we're on the same page then. My understanding was that we want the Helm Chart to be in technical preview in 9.0.0, and become GA in a later release. Users will still want to install it while it's in technical preview - are we going to tell them to install `9.0.0-SNAPSHOT`?

pkoutsovasilis:
> Right, we're on the same page then. My understanding was that we want the Helm Chart to be in technical preview in 9.0.0, and become GA in a later release. Users will still want to install it while it's in technical preview - are we going to tell them to install `9.0.0-SNAPSHOT`?

isn't that what was agreed [here](https://github.com/elastic/elastic-agent/issues/5485#issuecomment-2339339231)? I mean if  the suffix `-SNAPSHOT` is what causes the confusion here it can easily change to `-beta`. My end-goal was to make the Helm chart installable only by using `--devel` flag which itself means that this chart is not GA yet

swiatekm:
> isn't that what was agreed https://github.com/elastic/elastic-agent/issues/5485#issuecomment-2339339231? I mean if the suffix -SNAPSHOT is what causes the confusion here it can easily change to -beta. My end-goal was to make the Helm chart installable only by using --devel flag which itself means that this chart is not GA yet

If we want to follow agent versioning, then `X.Y.Z-SNAPSHOT` is the version for the latest commit on the `X.Y` branch. This is a moving target, as it changes whenever we merge into that branch. I was thinking that we should differentiate these versions from the beta versions we actually want users to install.

Let's say for example that we're releasing agent `9.0.0`. With your current changes in this PR, we'd get Helm Chart `9.0.0-SNAPSHOT`. Then for the next commit to the `9.0` branch, we'd get agent `9.1.0-SNAPSHOT` and Helm Chart `9.1.0-SNAPSHOT`. My point was that it'd be better for the Helm Chart for agent `9.0.0` to be `9.0.0-beta` instead, to differentiate it from the snapshot versions. Does that make sense?

pkoutsovasilis:
> If we want to follow agent versioning, then `X.Y.Z-SNAPSHOT` is the version for the latest commit on the `X.Y` branch. This is a moving target, as it changes whenever we merge into that branch. I was thinking that we should differentiate these versions from the beta versions we actually want users to install.
> 
> Let's say for example that we're releasing agent `9.0.0`. With your current changes in this PR, we'd get Helm Chart `9.0.0-SNAPSHOT`. Then for the next commit to the `9.0` branch, we'd get agent `9.0.1-SNAPSHOT` and Helm Chart `9.0.1-SNAPSHOT`. My point was that it'd be better for the Helm Chart for agent `9.0.0` to be `9.0.0-beta` instead, to differentiate it from the snapshot versions. Does that make sense?

yes it does. I have to admit that I have been looking long enough in elastic-agent repo to a point where not -SNAPSHOT translates to -beta in my head, but this is not the general case. Thus I changed that to `-beta` thanks for catching that 🙏

magefile.go
```diff
@@ -3503,71 +3487,43 @@ func (Helm) RenderExamples() error {
...
-			return fmt.Errorf("root node is not a document node")
-		} else if len(rootNode.Content) == 0 {
-			return fmt.Errorf("root node has no content")
-		}
-
-		if err := updateYamlNodes(rootNode.Content[0], agentVersion, chart.versionKeys...); err != nil {
+	agentVersion := bversion.GetParsedAgentPackageVersion().CoreVersion()
+	agentSnapshotVersion := agentVersion + "-SNAPSHOT"
+	// until the Helm chart reaches GA this remains with -beta suffix
+	agentChartVersion := agentVersion + "-beta"
```
swiatekm, line 3495: I *think* what we want to do here is something like: If the agent version has the `-SNAPSHOT` suffix, then Chart version should be the same as agent version. Otherwise, Chart version should be agent version + "beta". Does that make sense?

swiatekm: That sounds good to me. We can do that in a separate PR though, this one looks good to me. Thanks for being patient with my feedback, in hindsight I probably should've brought this up in the issue.

swiatekm (approved):

PR #6315, opened 2024-12-12
[otel] add loadbalancing exporter component

rogercoll:
## What does this PR do?

Adds the [Otel loadbalancing exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter) to the elastic-agent.

## Why is it important?

It gives the ability to route logs, metrics or traces to different OTLP endpoints depending on the configured `routing_key` . This feature can be helpful when distributing the processing load among different collectors. For example, current APM OTLP data is directly processed with the [lsminterval, elastictrace and signaltometrics components](https://github.com/elastic/elastic-agent/blob/main/deploy/helm/edot-collector/kube-stack/values.yaml#L902), but all those components could be moved from the collection/edge collectors to a set of collector that only handle data transformations. One of the main benefits of the latest approach is the separation of concerns, processing collectors would be part of another resource entity that would be managed independetly. 

Sample configuration that routes traces based on their service.name key:

```yaml
      exporters:
        loadbalancing/traces:
          routing_key: "service"
          protocol:
            otlp:
          resolver:
            # use k8s service resolver, if collector runs in kubernetes environment
            k8s:
              service: lb-svc.kube-public
```

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


swiatekm (approved):
:+1: 

Just a clarification, you say:

> It gives the ability to route logs, metrics or traces to different OTLP endpoints depending on the configured routing_key . This feature can be helpful when distributing the processing load among different collectors. For example, current APM OTLP data is directly processed with the [lsminterval, elastictrace and signaltometrics components](https://github.com/elastic/elastic-agent/blob/main/deploy/helm/edot-collector/kube-stack/values.yaml#L902), but all those components could be moved from the collection/edge collectors to a set of collector that only handle data transformations. One of the main benefits of the latest approach is the separation of concerns, processing collectors would be part of another resource entity that would be managed independetly.

For routing data to different OTLP endpoints, the routing connector is sufficient, and much simpler. What you really want the load balancing exporter for, is ensuring that data for a given partition (based on an attribute, or context) is always routed to the same instance. That's required for tail sampling traces, and for global metrics aggregations, but not for data routing in general.

Is that correct, or am I misunderstanding what we need this component for?

rogercoll:
> What you really want the load balancing exporter for, is ensuring that data for a given partition (based on an attribute, or context) is always routed to the same instance.

Correct, thanks for the clarification! At the moment, we are doing some tests using the "service.name" resource attribute as routing key, so their metrics aggregation happens in the same "backend/collector".  

This is the configuration:

```yaml
        loadbalancing/other:
          routing_key: "service"
          protocol:
            otlp:
              # all options from the OTLP exporter are supported
              # except the endpoint
              timeout: 1s
          resolver:
            # use k8s service resolver, if collector runs in kubernetes environment
            k8s:
              service: opentelemetry-kube-stack-lb-collector-headless
              ports:
                - 4317
```

What we would expect by using this loadbalancing configuration is to forward each service's data to a specific instance of the collector behind the headless K8s service.

swiatekm:
@rogercoll could you add a changelog fragment and rebase? We've had some issues with the CI today, and the current failure should clear if you do that.

michalpristas (approved):

PR #5113, opened 2024-07-10
Ess terraform + run integration tests on BK agents

pazone:
## What does this PR do?
A dedicated pipeline that runs Ubuntu integration tests directly on Buildkite runners without OGC. Currently it's not triggered by any of Github events.
Sudo tests [build example](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/3489#_)

- Improves groupping transparency. We can now immediately see long-running groups;
- Adds the ability to rerun failed groups instead of the entire pipeline build;
- Introduces a simple terraform configuration for ESS stack instead of a complex stack provisioner;
- Uses prebuilt VM images `platform-ingest-elastic-agent-ubuntu-2204`. All required packages are pre-installed and we don't need to install them in runtime; 
- Adds the ability to use all the gobld-supported CSPs and VMs incl. ARM64 and MacOS.

#### Shared ESS stack and test reruns

The `Start ESS stack for integration tests` spins up an ESS stack that is used by following integration tests. When the tests are complete, the stack will be destroyed regardless of the test results. When we **restart** a single test group, it will start a dedicated stack and destroy it when the tests are complete.

Screenshot 2024-10-07 at 11 53 03

This type of pipeline is supposed to replace the existing single-step approach. Further upcoming changes a listed in the [the meta-issue](https://github.com/elastic/ingest-dev/issues/2589).

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #3547, opened 2023-10-05
ESS Staging integ test improvements

pchila:
## What does this PR do?

- refactor deployment creation and support aws templates
- support tags to be assigned to integration tests stack deployments

## Why is it important?
Support all 3 major CSP and tag stack deployments appropriately

## Checklist

- [x] My code follows the style guidelines of this project
- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Author's Checklist

- [ ]

## Related issues

- 

## Questions to ask yourself

- How are we going to support this in production? 
- How are we going to measure its adoption? 
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #6323, opened 2024-12-13
[CI] Backport buildkite integration tests to 8.17

pazone:
## What does this PR do?
Synchronizes the new integration tests execution and the `devmachine` mage target with `main`

This is a manual backport. All changes related to BK integration tests are collectied in one pull request

## Checklist

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/4585

## Questions to ask yourself

- ...


.buildkite/scripts/steps/integration_tests_tf.sh
```diff
@@ -2,20 +2,63 @@
...
+echo "~~~ Building test binaries"
 mage build:testBinaries
 
-ess_up $OVERRIDE_TEST_AGENT_VERSION || echo "Failed to start ESS stack" >&2
-trap 'ess_down' EXIT
+# If the step is retried, we start the stack again.
+# BUILDKITE_RETRY_COUNT == "0" for the first run
+# BUILDKITE_RETRY_COUNT > 0 for the retries
+if [[ "${BUILDKITE_RETRY_COUNT}" -gt 0 ]]; then
+  echo "~~~ The steps is retried, starting the ESS stack again"
```
v1v, line 36: Can you confirm if `buildkite-agent meta-data get` can be used for secret access?

pazone: The management of metadata and secrets differs from that of GH Actions. It's like a cross-step key-value storage with a job scope. You have access to what you put there. You can potentially get a secret from Vault and put it in metadata. We don't do that because it might cause a leak.

v1v: > You can potentially get a secret from Vault and put it in metadata.  We don't do that because it might cause a leak.

IIUC, https://github.com/elastic/elastic-agent/pull/5113 does put a secret in metadata.

pazone: Raised a follow-up issue

test_infra/ess/deployment.tf
```diff
@@ -59,20 +59,30 @@ resource "ec_deployment" "integration-testing" {
   version                = local.deployment_version
 
   elasticsearch = {
-    autoscale = false
+    autoscale                 = false
+    instance_configuration_id = "gcp.es.datahot.n2.68x10x45"
```
kuisathaverat, line 63: Is there any reason to pass the instance_configuration_id? By default, the `deployment_template_id` selected will get this setting from the template override. It will stop working if the 'instance_configuration_id` disappears; that could happen on every new ms version.

test_infra/ess/deployment.tf
```diff
@@ -59,20 +59,30 @@ resource "ec_deployment" "integration-testing" {
...
+      size        = "8g"
       zone_count  = 1
     }
   }
   kibana = {
-    size       = "1g"
-    zone_count = 1
+    size                      = "1g"
+    zone_count                = 1
+    instance_configuration_id = "gcp.kibana.n2.68x32x45"
```
kuisathaverat, line 74: https://github.com/elastic/elastic-agent/pull/6323/files#r1939841825

test_infra/ess/deployment.tf
pazone: Thanks for noticing that. This pull request just migrates ESS cluster management from a custom go-based solution that uses [this configuration](https://github.com/elastic/elastic-agent/blob/6b9ff26acd583e055b3683b0aae2f8a0bdf337cb/pkg/testing/ess/create_deployment_csp_configuration.yaml#L2) to terraform. I can discuss the issue with the team and raise a follow-up issue.

kuisathaverat: I'm not sure why that file is used, but fixing the instance size seems pretty wrong because you are using a template from ECH and overriding the instances used (and tested) with another type without any justification. This cannot be addressed in a follow-up.

kuisathaverat (changes requested):
We cannot override ECH hardware templates without justification. Doing so implies cost and can cause issues in the tests and the ECH service.

test_infra/ess/deployment.tf
pazone: 1. This is a backport from the `main` to `8.17`. That means the problem you've mentioned is located in the main branch. 
2. The initial change that was made in the main branch doesn't introduce any new changes in the stack configuration. It just migrates it as it is to terraform.
3. When we want to raise a fixing PR - first we fix it in the `main` branch and then backport the fix to other branches like the `8.17`

v1v: Can you help me to understand what PRs are actually the ones that are backported from `main`? I can only see some changes, but I cannot find the original PRs? In other words, what's the reason `mergify` is not used for backporting the PRs from `main` to `8.17`? 

I think tracking the original PR/PRs can help with the context. Otherwise, keeping track of all the changes in this PR is hard and the original reason for those changes.

Mergify plays quite well with the backports; if this PR aggregates PRs, I feel it should be decomposed in small batches.

kuisathaverat: >This is a backport from the main to 8.17. That means the problem you've mentioned is located in the main branch.

Spread a wrong configuration because is in main does not justify the changes, and it will force you to fix it later when you can do it now.

>The initial changes that were made in the main branch don't introduce any new changes in the stack configuration. It just migrates it as it is to terraform. (The initial PR that introduces the deployment templates https://github.com/elastic/elastic-agent/pull/3547)
When we want to raise a fixing PR - first we fix it in the main branch and then backport the fix to other branches like the 8.17

Could it be better to fix the issue in the main and make only one backport?

pazone: @v1v Unfortunately, we couldn't backport multiple selective pull requests with mergify. You can find the list of the backported pull requests in the related issue. We discussed it previously, and this is the most effective way that we could find to backport multiple cherry-picked pull requests in terms of CI costs, speed, and reversibility

v1v: > we couldn't backport multiple selective pull requests with mergify

It's better to backport individual PRs instead of a massive one with multiple PRs.

Otherwise, it's hard/impossible to know if a branch has all the PRs from `main`. I understand the concerns about solving conflicts, but from the git history point of view there are plenty of benefits. Git blame/log will not help to know rationale for a change in a release branch that didn't use the standard approach for using one backport per PR and target branch

pazone: @pchila Do you happen to remember why did we use [this configuration](https://github.com/elastic/elastic-agent/blob/6b9ff26acd583e055b3683b0aae2f8a0bdf337cb/pkg/testing/ess/create_deployment_csp_configuration.yaml#L2)?

pchila: > @pchila Do you happen to remember why did we use [this configuration](https://github.com/elastic/elastic-agent/blob/6b9ff26acd583e055b3683b0aae2f8a0bdf337cb/pkg/testing/ess/create_deployment_csp_configuration.yaml#L2)?


This was coming from the original GCP template https://github.com/elastic/elastic-agent/pull/3547/files#diff-160183c98bcf9ae79293f7555054448bf030a6066bc2c0bb5ce31458d580373fL336
which was initially used to spin up clusters in ESS QA environment for which we had to add aws and azure as we went  though the environments from QA to Staging to Prod.
I think the original json has been derived from the ESS UI by selecting the `Equivalent API request` on the `Create deployment screen` which contains the `instance_configuration_id` attribute.
image

@kuisathaverat 
If this attribute is not needed to create ESS deployments we can make a quick PR on main to remove it without any problems

blakerouse (approved):

kuisathaverat (approved):

test_infra/ess/deployment.tf
kuisathaverat: it is not needed

PR #6328, opened 2024-12-13
Enable Terraform for all the BK integration steps

alexsapran:
## What does this PR do?

This PR enables the Terraform to be available in all the steps running the integration tests

## Why is it important?

When we retry a step, we need Terraform available so we can re-create the cluster for that step.


alexsapran:
As we discussed on Slack, we are exploring an alternative solution. To prevent further disruptions due to conflicts, I will be closing this issue.

PR #6334, opened 2024-12-13
otel: add test for batch splitting with the elasticsearchexporter

mauri870:
## What does this PR do?

Adds a test that the elasticsearchexporter does batch splitting properly on large batches comming from a beats receiver.

## Why is it important?

This guarantees that the scenario covered in https://github.com/elastic/beats/pull/41911 does not happen again.

## Checklist

## Related issues

- For https://github.com/elastic/ingest-dev/issues/3677

### DIFF
--- a/testing/integration/otel_test.go
+++ b/testing/integration/otel_test.go
@@ -1099,3 +1099,168 @@ service:
 	fixtureWg.Wait()
 	require.True(t, err == nil || errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded), "Retrieved unexpected error: %s", err.Error())
 }
+
+func TestOtelFBReceiverBatchSplitE2E(t *testing.T) {
+	// This test asserts that the batcher configuration from the elasticseachexporter
+	// splits batches to avoid a 413 Request Entity Too Large error from Elasticsearch.
+	info := define.Require(t, define.Requirements{
+		Group: Default,
+		Local: true,
+		OS: []define.OS{
+			{Type: define.Windows},
+			{Type: define.Linux},
+			{Type: define.Darwin},
+		},
+		Stack: &define.Stack{},
+	})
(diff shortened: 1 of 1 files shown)


jlind23:
@mauri870 does this PR fixes https://github.com/elastic/ingest-dev/issues/3677? After reading the description I am not sure what the `For` keyword means here.

testing/integration/otel_test.go
```diff
@@ -1099,3 +1099,168 @@ service:
 	fixtureWg.Wait()
 	require.True(t, err == nil || errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded), "Retrieved unexpected error: %s", err.Error())
 }
+
+func TestOtelFBReceiverBatchSplitE2E(t *testing.T) {
+	// This test asserts that the batcher configuration from the elasticseachexporter
+	// splits batches to avoid a 413 Request Entity Too Large error from Elasticsearch.
+	info := define.Require(t, define.Requirements{
```
cmacknz, line 1106: Just double checking why we need this to be an integration test that runs the entire collector. What additional test coverage is this adding? Why can't this be covered by a unit test in the exporter?

I like integration tests, they catch a lot of problems, but they are also driving our CI execution time to infinity :)

mauri870: Agree, I think this is better suited to be in the exporter. Actually, it might as well be covered in the exporter already by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36396. I will double check.

mauri870:
So I looked into the tests and the elasticsearchexporter has coverage of test options, including flush::interval and flush::bytes [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/87f39c4/exporter/elasticsearchexporter/bulkindexer_test.go#L72), I don't think an integration test in this case adds any additional coverage.

PR #6401, opened 2024-12-19
Split packaging into separate steps

oakrizan:
## What does this PR do?
Splitting centralized packaging step into separate steps to speed up builds.
Currently packagin takes up to 30 mins to run:
image
With thae changes in current PR packaging in total takes up to 15 mins:
image

Buildkite builds: https://buildkite.com/elastic/elastic-agent-extended-testing/builds?branch=split-packaging

NB! Failure of `Win2022:sudo:upgrade` test is not related to changes in current PR.

## Why is it important?
Since the integration test steps are divided by platforms, we can divide packaging into the corresponding steps and run integration tests for the platform as soon as the required artifacts are built without waiting for the rest.

## Checklist

## Related issues

- Relates: https://github.com/elastic/ingest-dev/issues/4712

## Questions to ask yourself

- ...


.buildkite/integration.pipeline.yml
```diff
@@ -4,28 +4,84 @@ env:
...
+              - linux/arm64
+            ext:
+              - rpm
+              - deb
 
   - label: "Serverless integration test"
     key: "serverless-integration-tests"
     depends_on:
-      - package-it
+      - packaging-ubuntu-x86-64
```
pazone, line 73: NIT: Can we use the group key instead of mentioning all the steps?

.buildkite/integration.pipeline.yml
```diff
@@ -4,28 +4,84 @@ env:
   DOCKER_REGISTRY: "docker.elastic.co"
   VAULT_PATH: "kv/ci-shared/observability-ingest/cloud/gcp"
 
+  # Packaging
+  SNAPSHOT: true
```
pazone, line 8: I'd recommend returning those variables to the packaging script to reduce the scope

.buildkite/integration.pipeline.yml
```diff
@@ -39,14 +89,14 @@ steps:
...
+      - int-packaging
     concurrency_group: elastic-agent-extended-testing/leak-tests
     concurrency: 8
     env:
       TEST_INTEG_AUTH_GCP_DATACENTER: "us-central1-b"
-    command: |
-      buildkite-agent artifact download build/distributions/** . --step 'package-it'
-      .buildkite/scripts/steps/integration_tests.sh stateful integration:TestForResourceLeaks
+    commands:
+      - buildkite-agent artifact download "build/distributions/**" . --build $BUILDKITE_BUILD_ID
```
pazone, line 98: NIT: we can omit `--build` flag when we download artifacts from the current

.buildkite/integration.pipeline.yml
```diff
@@ -59,23 +109,26 @@ steps:
...
     key: "serverless-beats-integration-tests"
     concurrency_group: elastic-agent-extended-testing/beats-integration
     concurrency: 8
     env:
       TEST_INTEG_AUTH_GCP_DATACENTER: "us-central1-a"
-    command: |
-      buildkite-agent artifact download build/distributions/** . --step 'package-it'
-      .buildkite/scripts/steps/beats_tests.sh
+    commands:
+      - buildkite-agent artifact download "build/distributions/**" . --build $BUILDKITE_BUILD_ID
```
pazone, line 130: NIT: we can omit --build flag when we download artifacts from the current

pazone (approved):

.buildkite/integration.pipeline.yml
```diff
@@ -59,23 +109,23 @@ steps:
...
+      - int-packaging
     trigger: "elastic-agent-extended-testing-bk"
     build:
       commit: "${BUILDKITE_COMMIT}"
       branch: "${BUILDKITE_BRANCH}"
 
   - label: "Serverless Beats Tests"
     depends_on:
-      - package-it
+      - packaging-ubuntu-x86-64
```
v1v, line 120: what's the reason for using `packaging-ubuntu-x86-64` instead of `int-packaging`?

.buildkite/scripts/steps/integration-package.sh
```diff
@@ -2,5 +2,11 @@
...
 source .buildkite/scripts/common.sh
+
 # Remove AGENT_PACKAGE_VERSION pinning as soon as 9.0.0 is released
-AGENT_PACKAGE_VERSION=9.0.0 PACKAGES=tar.gz,zip,rpm,deb PLATFORMS=linux/amd64,linux/arm64,windows/amd64  SNAPSHOT=true EXTERNAL=true  mage package
+export AGENT_PACKAGE_VERSION=9.0.0
+
+export SNAPSHOT="true"
+export EXTERNAL="true"
+
+mage package
```
v1v, line 12: What's the reason for using exports instead of the former approach with in-line variables?

```suggestion
AGENT_PACKAGE_VERSION=9.0.0 SNAPSHOT=true EXTERNAL=true mage package
```

.buildkite/bk.integration.pipeline.yml
```diff
@@ -78,7 +78,7 @@ steps:
 
       - label: "x86_64:sudo: {{matrix}}"
         command: |
-          buildkite-agent artifact download build/distributions/** . --step 'package-it' --build ${BUILDKITE_TRIGGERED_FROM_BUILD_ID}
+          buildkite-agent artifact download build/distributions/** . --build ${BUILDKITE_TRIGGERED_FROM_BUILD_ID}
```
v1v, line 81: what's the reason we don't use `-step 'packaging-ubuntu-x86-64'` as done above?

.buildkite/integration.pipeline.yml
oakrizan: For this particular step only artifact produced from `packaging-ubuntu-x86-64` is required. No need to spending more time to download all artifacts. Besides once `packaging-ubuntu-x86-64` is completed "Serverless Beats Tests" step will start without awaiting for other packaging steps.

.buildkite/bk.integration.pipeline.yml
oakrizan: Tests executed in this step require not only ubuntu, but debian as well.

v1v: Can you add an in-line comment to explain the rationale?

v1v: > Tests executed in this step require not only ubuntu, but debian as well.

I'm not sure if I follow, IIUC, `ubuntu` and `debian` are `deb` packages. I cannot find how those packages are generated now.

OTOH,  `label: "x86_64:non-sudo: {{matrix}}"` uses `--step 'packaging-ubuntu-x86-64'`, so this tasks is `sudo`, I cannot understand what's the difference though

oakrizan: In `https://github.com/elastic/elastic-agent/blob/f0b810acdba94c3c9cf0609367ca94572f3fe1a1/.buildkite/integration.pipeline.yml` we create packages for Ubuntu x86_64, Ubuntu arm64, Windows and RPM amd64/amr64 and DEB amd64/arm64.  Then those are used in subsequenty triggered Integration test pipeline.

v1v: IIUC, https://github.com/elastic/elastic-agent/blob/f0b810acdba94c3c9cf0609367ca94572f3fe1a1/.buildkite/integration.pipeline.yml#L47-L65 is actually the task creating the deb/rpm files (it honours `PACKAGES`), but the name says `Packaging: Containers`. 

Do you mean that `label: "x86_64:non-sudo: {{matrix}}"` uses  [PACKAGES: "tar.gz"](https://github.com/elastic/elastic-agent/blob/f0b810acdba94c3c9cf0609367ca94572f3fe1a1/.buildkite/integration.pipeline.yml#L14C22-L14C28) while `"x86_64:sudo: {{matrix}}"` uses `deb` packages?

oakrizan: Step `"x86_64:sudo: {{matrix}}"` uses `tar.gz` and `deb` packages, while step `"x86_64:non-sudo: {{matrix}}"` uses only `tar.gz` packages.

Please check the build for the `"x86_64:sudo: {{matrix}}"` using --step 'packaging-ubuntu-x86-64', which causes build `x86_64:sudo: deb` to failhttps://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/1306#0194d10a-653b-4002-83c1-56c5a7a95eed/326-1014

.buildkite/scripts/steps/integration-package.sh
oakrizan: to improve readability, since previously there were a lot of params and command was too long

.buildkite/bk.integration.pipeline.yml
v1v: > Step "x86_64:sudo: {{matrix}}" uses tar.gz and deb packages, while step "x86_64:non-sudo: {{matrix}}" uses only tar.gz packages.

Thanks for confirming. Can you add an inline comment explaining the rationale?

.buildkite/integration.pipeline.yml
pazone: IMHO it's obvious. If a step depends on `packaging-ubuntu-x86-64` that means that the dependent step requires only the artifacts from `packaging-ubuntu-x86-64`. I don't think that it worth retriggering the CI pipeline

v1v: > . If a step depends on packaging-ubuntu-x86-64 that means that the dependent step requires only the artifacts from packaging-ubuntu-x86-64

It's implicit, however, the more context about those dependencies, the better for the reviewers and the people who don't know anything. It's the only way we can avoid tribal knowledge or the need to use `git blame` and search for when and how that change happened.

.buildkite/integration.pipeline.yml
```diff
@@ -5,27 +5,77 @@ env:
...
+        key: "packaging-ubuntu-x86-64"
+        env:
+          PACKAGES: "tar.gz"
+          PLATFORMS: "linux/amd64"
+        command: ".buildkite/scripts/steps/integration-package.sh"
+        artifact_paths:
+        - build/distributions/**
+        agents:
+          provider: "gcp"
+          machineType: "n1-standard-8"
```
dliappis, line 21: nit: `n1` is the oldest general purpose machine generation on GCP. With `n2-standard-8` we'd get much better performance with a very small [diff in price](https://cloud.google.com/compute/vm-instance-pricing?hl=en#n2_predefined) (+6$ for continuous month use for n2, not including discounts.)

.buildkite/integration.pipeline.yml
```diff
@@ -5,27 +5,77 @@ env:
...
     concurrency_group: elastic-agent-extended-testing/serverless-integration
     concurrency: 8
     env:
       # we run each step in a different data center to spread the load
       TEST_INTEG_AUTH_GCP_DATACENTER: "us-central1-a"
-    command: |
-      buildkite-agent artifact download build/distributions/** . --step 'package-it'
-      .buildkite/scripts/steps/integration_tests.sh serverless integration:single TestLogIngestionFleetManaged #right now, run a single test in serverless mode as a sort of smoke test, instead of re-running the entire suite
+    commands:
+      - buildkite-agent artifact download "build/distributions/**" . $BUILDKITE_BUILD_ID
```
dliappis, line 77: any reason we are we moving away from the block scalar (`|`) to arrays? Is the reason that you want to rely on buildkite's built-in failure handling (

We've established the pattern with `|` in other repos which IMHO enhances readability and helps with copying/pasting for reproductions which why I am bringing this up.

dliappis (commented):
Basically LGTM, one question

dliappis (approved):
LGTM

v1v (approved):

PR #6408, opened 2024-12-19
Build only cloud image variant when it's requested

rdner:
Before this change we were building all possible Docker images to produce the cloud image. Which is unnecessary and takes a long time.

Now only the `cloud` variant is built. 

**On my machine this reduces this build time from 27 to 20 minutes.**

For testing:

```sh
mage cloud:image
```

Or follow this guide https://github.com/elastic/elastic-agent?tab=readme-ov-file#testing-on-elastic-cloud


jlind23 (approved):

jlind23:
Looks good to me.

michalpristas (approved):

PR #6413, opened 2024-12-19
[ESS Testing] Update ESS region and deployment template ID

ycombinator:
## What does this PR do?

This PR updates the GCP region and ESS deployment template ID in the Terraform variables file to values that are currently acceptable in ESS, as per https://www.elastic.co/guide/en/cloud/current/ec-regions-templates-instances.html.

## Why is it important?

So ESS deployments can be spun for testing Elastic Agent changes in ESS, as per https://github.com/elastic/elastic-agent?tab=readme-ov-file#testing-on-elastic-cloud

## Checklist

- [ ] ~My code follows the style guidelines of this project~
- [ ] ~I have added tests that prove my fix is effective or that my feature works

## Disruptive User Impact

None, as this change only impacts Elastic Agent developers.

1. Run `EC_API_KEY=your_api_key make apply`, replacing `your_api_key` with an [API key from your ESS account](https://cloud.elastic.co/account/keys).
2. You may see failures about the Elastic Agent version and the command may not succeed, but you should **not** see failures like `api error: deployments.elasticsearch_using_legacy_dt: The referenced deployment template in elasticsearch plan [...] is no longer available.`

## Questions to ask yourself

- ...

### DIFF
--- a/testing/environments/cloud/variables.tf
+++ b/testing/environments/cloud/variables.tf
@@ -1,13 +1,13 @@
 ## Deployment configuration
 
 variable "ess_region" {
-  default     = "gcp-us-west2"
-  description = "Optional ESS region where the deployment will be created. Defaults to gcp-us-west2"
+  default     = "gcp-us-west1"
+  description = "Optional ESS region where the deployment will be created. Defaults to gcp-us-west1"
   type        = string
 }
 
 variable "deployment_template" {
-  default     = "gcp-compute-optimized-v2"
+  default     = "gcp-compute-optimized"
   description = "Optional deployment template. Defaults to the CPU optimized template for GCP"
   type        = string
 }


cmacknz (approved):

testing/environments/cloud/variables.tf
```diff
@@ -1,13 +1,13 @@
 ## Deployment configuration
 
 variable "ess_region" {
-  default     = "gcp-us-west2"
-  description = "Optional ESS region where the deployment will be created. Defaults to gcp-us-west2"
+  default     = "gcp-us-west1"
```
rdner, line 4: We cannot use `gcp-us-west1` region because we don't publish snapshots there. Any reason why it was switched from `gcp-us-west2`?

ycombinator: Evidently, `gcp-us-west2` is not a supported region: https://www.elastic.co/guide/en/cloud/current/ec-regions-templates-instances.html.  I tried it anyway initially and was getting the same error as before from the ESS API (via `terraform apply`):

```
│ Error: failed creating deployment: 9 errors occurred:
│ 	* api error: deployments.elasticsearch_using_legacy_dt: The referenced deployment template in elasticsearch plan [gcp-compute-optimized] is no longer available.
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.coordinating.1] in topology element [coordinating] is no longer available .
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.data.highcpu.1] in topology element [hot_content] is no longer available .
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.data.highstorage.1] in topology element [cold] is no longer available .
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.data.highstorage.1] in topology element [warm] is no longer available .
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.es.datafrozen.n1.64x10x95] in topology element [frozen] is no longer available .
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.master.1] in topology element [master] is no longer available .
│ 	* api error: deployments.topology_element_using_legacy_ic: The referenced instance configuration [gcp.ml.1] in topology element [ml] is no longer available .
```

So I had to switch it to `gcp-us-west1`.

rdner: @ycombinator did it change recently, I used this Terraform script with `gcp-us-west2` a couple of months ago and it worked just fine.

Using any region that does not have snapshots (any other than `gcp-us-west2`) is not helpful because we always test the latest snapshot.

ycombinator: > did it change recently?

@gersonsosa might be able to answer this question.

gersonsosa: @ycombinator these deployment templates were deprecated some time ago, recently we released a validation that doesn't allow these deployments to be created.

I just checked and the deployment template "gcp-compute-optimized" is marked as legacy on both regions which is odd given only one of them is giving the error, are you executing the command using an internal user with an `@elastic` email address? if so any internal user will be allowed to create deployments with any template, maybe that's the case here

PR #6428, opened 2024-12-23
[go.mod] update elastic-agent-system-metrics

VihasMakwana:
This PR updates `beats/v7` to point to https://github.com/elastic/beats/commit/58af1537830bf314698d760f17ba0b051f2b1dc0. 

Because of new config changes in beats and elastic-agent-system-metrics, we need to update both of them. This would also include Lee's `fbreceiver` work so it should be alright.


VihasMakwana:
this is replacement for https://github.com/elastic/elastic-agent/pull/6320

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/elastic/elastic-agent
 
-go 1.22.9
+go 1.22.10
```
pkoutsovasilis, line 3: I believe we shouldn't bump golang version in this PR, right? 🙂

VihasMakwana: yeah, not needed. I've removed it.

blakerouse (approved):
+1

andrzej-stencel (approved):

PR #6444, opened 2024-12-27
[otel/kube-stack]: Add gateway collector

rogercoll:
## What does this PR do?

This PR adds a new K8s deployment of the EDOT collector named "gateway". The main purpose of this new deployment is to simplify the daemonset collector configuration and unify managed/self-managed scenarios. The gateway collector configuration contains all Elastic's custom Otel components needed for the signals transformations in self-managed scenarios, which are currently configured in the daemonset collector. 

image

Elastic configured components in the "Gateway" collectors (previously in the daemonset):
- signaltometrics
- elasticinframetrics
- elastictrace
- lsminterval

Another important change is that the Gateway collectors will forward the Otel data to Elasticsearch, the daemonset and cluster collectors configurations have been updated to export all collected data to the corresponding Gateway OTLP endpoint. Although the daemonset collectors are still configured to collect the auto-instrumentation OTLP data, the data is load balanced (loadbalancing exporter) to the gateway collectors based on the service name.

Additional context: https://github.com/elastic/opentelemetry-dev/issues/587

## Why is it important?

The key benefit of this architecture is to decouple data collection from data transformations (e.g. APM enrichment and aggregations), for managed scenarios, users would _just_ need to remove (or comment) the "gateway" collector configuration. Note that moving the data processing from a k8s "daemonset" to a "deployment", it eases its horizontal scaling. 

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


michalpristas:
please link an issue if such exists

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,75 +518,418 @@ collectors:
...
+            key: elastic_api_key
+      - name: GOMAXPROCS
+        valueFrom:
+          resourceFieldRef:
+            resource: limits.cpu
+      - name: GOMEMLIMIT
+        value: "1025MiB"
+    config:
+      connectors:
+        routing:
```
rogercoll, line 597: @felixbarny for context, this configuration routes the incoming OTLP metrics based on the scope name:
- kubeletstats metrics: Need to be exported to both, Elasticsearch in `ecs` (inventory) and `otel` mode (otel native dashboards).
- hostmetrics: Only `ecs` mode is supported.

These two pipelines are needed to avoid `hostmetrics` metrics deduplication. The `elasticinframetrics` processor is configured to drop the original metrics.  Reference: https://github.com/elastic/opentelemetry-lib/issues/97

rogercoll:
This new configuration requires the `loadbalancing` and `routing` components that were added in https://github.com/elastic/elastic-agent/pull/6210 and https://github.com/elastic/elastic-agent/pull/6315

deploy/helm/edot-collector/kube-stack/values.yaml
felixbarny: I think we should unify and simplify the configuration so that we always export the original data in `otel` mode and both translated host and k8s metrics are additionally sent in `ecs` mode. That way users can already build their custom dashboards with OTel native host metrics and we're set up for changing our UIs to also use OTel native host metrics. I don't see a good reason why we should have different strategies for host vs k8s metrics, and I assume this was more accidental rather than intentional.

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -204,344 +181,50 @@ collectors:
...
-          endpoints:
-            - ${env:ELASTIC_ENDPOINT}
-          api_key: ${env:ELASTIC_API_KEY}
-          # tls:
-          #   insecure_skip_verify: true
-          mapping:
-            mode: ecs
+        # [Loadbalancing exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter)
+        loadbalancing/gateway:
+          routing_key: service
```
felixbarny, line 211: I suppose you're implementing the load balancing in order to help with spreading the load for the metric aggregations. However, I'm not sure if load balancing on service name is doing more harm or good. The risk is that this could lead to hot-spotting specific instances if there's a service that's a lot more chatty than others. For APM Server, we've just been doing a random load-balancing so far.

I'd like to get opinions from @axw @simitt @lahsivjar on what you think we should start with for routing OTLP traffic from daemonsets to a set of gateway collectors.

I think that for very large-scale deployments, we need something more akin to what we're working on for HOTel, and that will probably be managed via a separate operator.

axw: I would also be inclined to keep it randomly distributed, unless there's a need for some stateful processing that requires it to be partitioned.

pkoutsovasilis: Hi all 👋 Just my 2 cents here! Based on my experience, nodes within the same zone as the requester often tend to be favored. Thus, I would also vote starting with a randomly distributed approach as it seems like the safer option initially and it helps avoid potential hot-spotting issues 🙂

testing/integration/otel_helm_test.go
```diff
@@ -99,10 +101,8 @@ func TestOtelKubeStackHelm(t *testing.T) {
 
 				// override secrets reference with env variables
 				JSONValues: []string{
-					fmt.Sprintf(`collectors.cluster.env[1]={"name":"ELASTIC_ENDPOINT","value":"%s"}`, kCtx.esHost),
-					fmt.Sprintf(`collectors.cluster.env[2]={"name":"ELASTIC_API_KEY","value":"%s"}`, kCtx.esAPIKey),
-					fmt.Sprintf(`collectors.daemon.env[2]={"name":"ELASTIC_ENDPOINT","value":"%s"}`, kCtx.esHost),
-					fmt.Sprintf(`collectors.daemon.env[3]={"name":"ELASTIC_API_KEY","value":"%s"}`, kCtx.esAPIKey),
+					fmt.Sprintf(`collectors.gateway.env[1]={"name":"ELASTIC_ENDPOINT","value":"%s"}`, kCtx.esHost),
```
pkoutsovasilis, line 104: I am sorry for the request but how hard is it to align the code of this test with [these one](https://github.com/elastic/elastic-agent/blob/main/testing/integration/kubernetes_agent_standalone_test.go#L267). Feel free to say no if it too much of an ask, we can do that in a follow-up PR 🙂

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,75 +518,418 @@ collectors:
...
+            processors:
+              - elasticinframetrics
+              - attributes/dataset
+              - batch
+            exporters: [debug, elasticsearch/ecs]
+          metrics/infra/otel:
+            receivers: [routing]
+            processors: [batch]
+            exporters: [debug, elasticsearch/otel]
+          metrics/otlp:
```
felixbarny, line 921:
```suggestion
          metrics/otel:
```

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,75 +518,418 @@ collectors:
...
+            receivers: [routing]
+            processors:
+              - elasticinframetrics
+              - attributes/dataset
+              - batch
+            exporters: [debug, elasticsearch/ecs]
+          metrics/infra/otel:
+            receivers: [routing]
+            processors: [batch]
+            exporters: [debug, elasticsearch/otel]
```
felixbarny, line 920: Why don't we send metrics to `metrics/otel`, so that the `signaltometrics` connector is invoked for these?

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Makes sense to me, the reasoning behind dropping the original data within the EIMP was to prevent exporting the same metrics twice, in `ecs` and `otel` elasticsearch exporter modes. The following could be a simplification:

```yaml
      processors:
          elasticinframetrics:
            add_system_metrics: true
            add_k8s_metrics: true
            drop_original: true
      ...
      connectors:
        routing:
          default_pipelines: [metrics/otlp]
          error_mode: ignore
          match_once: true
          table:
            - context: metric
              statement: route() where instrumentation_scope.name == "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver" or IsMatch(instrumentation_scope.name, "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/*")
              pipelines: [metrics/infra/ecs, metrics/otlp]
```

The previous would do the following data routings:

- Original host and k8s data -> elasticsearch exporter otel mode
- **Only** ecs remapped data from host and k8s  -> elasticsearch exporter ecs mode

Should we push the original data to the elasticsearch in ecs mode as well `drop_original: false`? What would be the benefit? E.g:

- Original host and k8s data -> elasticsearch exporter otel mode
- Original and ecs remapped data from host and k8s  -> elasticsearch exporter ecs mode

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Good point, fixed in https://github.com/elastic/elastic-agent/pull/6444/commits/8d859d62048a7349fe394cdd1aa3f37bace0f874

Should we route the ecs infra metrics to the `signaltometrics` connector too?

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Changed https://github.com/elastic/elastic-agent/commit/8d859d62048a7349fe394cdd1aa3f37bace0f874

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,75 +518,414 @@ collectors:
...
+              from_attribute: data_stream.dataset
+              action: upsert
+        batch:
+          send_batch_size: 1000
+          timeout: 1s
+          send_batch_max_size: 1500
+        # [Elastic Trace Processor](https://github.com/elastic/opentelemetry-collector-components/tree/main/processor/elastictraceprocessor)
+        elastictrace: {} # The processor enriches traces with elastic specific requirements.
+        # [LSM Interval Processor](https://github.com/elastic/opentelemetry-collector-components/tree/main/processor/lsmintervalprocessor)
+        lsminterval:
```
lahsivjar, line 868: We should a `directory` here, do we have provision for a volume mount?

deploy/helm/edot-collector/kube-stack/values.yaml
felixbarny: > The previous would do the following data routings:
> 
> * Original host and k8s data -> elasticsearch exporter otel mode
> * **Only** ecs remapped data from host and k8s -> elasticsearch exporter ecs mode

This makes sense to me. I don't think we should send both original and remapped data in ecs mode.

deploy/helm/edot-collector/kube-stack/values.yaml
felixbarny: Good question. Since we already route the `otel` metrics through it, it feels like that would duplicate the counting. That's why I'd lean towards not sending `ecs` metrics through the `signaltometrics` connector.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Great, infra pipelines unified https://github.com/elastic/elastic-agent/pull/6444/commits/1652f1483670bdab80173dac95a59e1da6904722

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Thanks for the feedback, looks like there's an agreement on using random distribution for now. Removed the service based approach to use the default (traceID based) https://github.com/elastic/elastic-agent/pull/6444/commits/30512425f40de321ca9e6ceeeead569e35bf3339

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Yes, we can define volumes/volumes mount for each set of collectors. If persistence is not a requirement, we could use an `emptyDir` [volume type](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir)

Does it need to be persistent over pod restarts? And unique per container?

lahsivjar: Persistence is not necessary for now but we would need to have some headroom to scale the volume. If the node size is big enough then it should be good to start with emptyDir.

rogercoll: > If the node size is big enough then it should be good to start with emptyDir.

The issue is that we cannot do any assumption on the node size, we are not currently setting any hardware limit aside from the resource memory limits. The `emptyDir` volume size can be limited (I am a bit lost of which should be the limit), but if we don't set a limit it will be memory backed (max size == pod.resource.memory.limit) which we might not get any benefit out of it.

lahsivjar: > but if we don't set a limit it will be memory backed (max size == pod.resource.memory.limit) which we might not get any benefit out of it.

Hmm, we definitely don't want memory backed. I think we can start with 20GB of disk space if the node will allow that, does that sound reasonable for now?

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -204,344 +181,49 @@ collectors:
...
-        elasticsearch/ecs:
-          endpoints:
-            - ${env:ELASTIC_ENDPOINT}
-          api_key: ${env:ELASTIC_API_KEY}
-          # tls:
-          #   insecure_skip_verify: true
-          mapping:
-            mode: ecs
+        # [Loadbalancing exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter)
+        loadbalancing/gateway:
```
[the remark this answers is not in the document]
axw: ++ came to say the same thing. If we use a non-headless Service, then I think we can just export to that and let k8s randomly distribute the load as normal.

It could be useful to do trace ID-based load balancing if we want to perform trace analytics, but I'd defer on that for now. Otherwise we may need to think about misbehaving agents that repeat trace IDs, e.g. https://github.com/elastic/apm-server/issues/3922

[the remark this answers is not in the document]
rogercoll: That was my assumption as well, but I run some tests to verify the load distribution and it seems each `otlp/gateway` exporter sends to the same OTLP pod/endpoint during its lifetime. This is a simplified workflow of the tests that I run:

-  Traces load generator → Non-headless OTLP daemonset Service → Non-headless OTLP Gateway Service (`otlp/gateway`)
-  Traces load generator → Non-headless OTLP daemonset Service → Headless OTLP Gateway Service (`loadbalancing/gateway`)

Each Gateway Service is on top of 2 Pods K8s deployment, thus the load should be distributed among the 4 gateway pods.  The following screenshot are the results of the total number of received spans on each gateway pod and its memory usage (colors do not match over diagrams).  

image

From both visualization, there is one Pod behind the Non-headless OTLP Gateway Service that is not getting any load (no spans, no increase on memory usage `lsminterval`).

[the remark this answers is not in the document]
swiatekm: K8s Services do load balancing at the TCP level. In fact, they're actually just firewall rules on the Node. The OTLP exporter uses GRPC, which establishes a persistent TCP connection to the host, so it will always send to the same gateway collector. I think this should actually work with a headless Service, as that's just a DNS record, and the GRPC library is aware of DNS and capable of doing client-side load balancing on it.

Alternatively, we could do OTLP over HTTP and explicitly disable keep alives.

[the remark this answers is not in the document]
rogercoll: > I think this should actually work with a headless Service, as that's just a DNS record, and the GRPC library is aware of DNS and capable of doing client-side load balancing on it.
Alternatively, we could do OTLP over HTTP and explicitly disable keep alives.

Good point, that works for load balancing incoming connections (otlp GRPC defaults `round_robin` mode works on connection level), but not for each request/trace. I was only able to get the same load distribution (for a single connection) as using the `loadbalancing` exporter by switching to OTLP over HTTP with keep alives disabled:

image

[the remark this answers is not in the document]
felixbarny: > it seems each otlp/gateway exporter sends to the same OTLP pod/endpoint during its lifetime.

Yeah, that's what I would expect as well. This is usually fine if you have more daemonset collectors than gateway collector instances. I'd expect a single collector instance being able to handle traffic from multiple daemonset collectors.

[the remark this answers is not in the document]
rogercoll: Since using `otlphttp` with disabled keep-alives isn't ideal (as it creates a new connection for every request), what do you think about continuing with the `otlp/gateway` approach for this iteration (as proposed)? The load tests were conducted using a single connection to a specific DaemonSet collector. In a real-world scenario, I would expect multiple agents (and therefore multiple connections), which each `otlp/gateway` should distribute across the gateways using a round-robin approach.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: I'm a bit concerned about adding this hardware requirement at the moment. While 20GB of disk space sounds reasonable in principle, the configuration is deploying two instances of the gateway collector (40GB in total). Also, I think we should add this disk requisite on the onboarding page (Kibana) as well. 
Do you think we could explore an opt-in approach for the next iteration?

deploy/helm/edot-collector/kube-stack/values.yaml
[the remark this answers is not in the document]
felixbarny: Just to validate my understanding - you're proposing to remove the `loadbalancing/gateway` exporter in favor of `otlp/gateway`, with the assumption that load balancing in real-world scenarios where you have multiple agents evenly distributes load. Did I get that right? If so, I agree.

testing/integration/otel_helm_test.go
```diff
@@ -55,22 +54,36 @@ func TestOtelKubeStackHelm(t *testing.T) {
...
-			// (1 EDOT collector pod per node)
-			// - A Cluster wide Deployment to collect K8s metrics and
-			// events (1 EDOT collector pod per cluster)
-			// - An OpenTelemetry Operator Deployment (1 pod per
-			// cluster)
-			atLeastValidatedPodsNumber: 3,
+			name: "helm kube-stack operator standalone agent kubernetes privileged",
+			steps: []k8sTestStep{
+				k8sStepCreateNamespace(),
+				k8sStepHelmDeployWithValues(chartLocation, "kube-stack-otel",
```
pkoutsovasilis, line 64:
```suggestion
				k8sStepHelmDeployWithValueOptions(chartLocation, "kube-stack-otel",
```

testing/integration/otel_helm_test.go
```diff
@@ -79,82 +92,47 @@ func TestOtelKubeStackHelm(t *testing.T) {
...
+		helmValues, err := values.MergeValues(providers)
+		if err != nil {
+			require.NoError(t, err, "failed to helm values")
+		}
 
-				_, err = uninstallAction.Run(tc.helmReleaseName)
-				if err != nil {
-					require.NoError(t, err, "failed to uninstall helm chart")
-				}
+		k8sStepHelmDeploy(chartPath, releaseName, helmValues)
```
pkoutsovasilis, line 115:
```suggestion
		k8sStepHelmDeploy(chartPath, releaseName, helmValues)(t, ctx, kCtx, namespace)
```

testing/integration/otel_helm_test.go
```diff
@@ -79,82 +92,47 @@ func TestOtelKubeStackHelm(t *testing.T) {
...
-
-				checkedAgentContainers := 0
-
-				for _, pod := range podList.Items {
-					if strings.HasPrefix(pod.GetName(), tc.helmReleaseName) && pod.Status.Phase == corev1.PodRunning {
-						checkedAgentContainers++
-					}
+			for _, pod := range perNodePodList.Items {
+				if pod.Status.Phase == corev1.PodRunning {
+					checkedAgentContainers++
```
pkoutsovasilis, line 132:
```suggestion
			for _, pod := range perNodePodList.Items {
				if pod.Status.Phase != corev1.PodRunning {
					continue
				}

				for _, container := range pod.Status.ContainerStatuses {
					if container.Name != containerName {
						continue
					}

					if container.RestartCount == 0 && container.State.Running != nil {
						checkedAgentContainers++
					}
```

testing/integration/otel_helm_test.go
rogercoll: renamed in https://github.com/elastic/elastic-agent/pull/6444/commits/ab64abd4ce2600177c9be5f230f9eca2ce71f8b8

deploy/helm/edot-collector/kube-stack/values.yaml
lahsivjar: > Do you think we could explore an opt-in approach for the next iteration?

Hmm, it is already kinda opt-in (users can define the directory, if not defined we use in-memory).

> While 20GB of disk space sounds reasonable in principle, the configuration is deploying two instances of the gateway collector (40GB in total).

Even 40GB sounds reasonable to me since disk is cheap but I might be missing something. Overall, opt-in sounds okay to me for now.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Resolving the thread as the `loadbalancing` was removed in favor of a headless service otlp endpoint.

testing/integration/otel_helm_test.go
rogercoll: refactored during https://github.com/elastic/elastic-agent/pull/6444/commits/c5c52aa53918fc9d074798a738a888303a4900af

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,75 +480,418 @@ collectors:
...
+            - key: process.executable.path
+              action: delete
+        batch:
+          send_batch_size: 1000
+          timeout: 1s
+          send_batch_max_size: 1500
+        # [Elastic Trace Processor](https://github.com/elastic/opentelemetry-collector-components/tree/main/processor/elastictraceprocessor)
+        elastictrace: {} # The processor enriches traces with elastic specific requirements.
+        # [LSM Interval Processor](https://github.com/elastic/opentelemetry-collector-components/tree/main/processor/lsmintervalprocessor)
+        lsminterval:
```
lahsivjar, line 833:
```suggestion
        batch/aggs:
          send_batch_size: 16384 # 2x the default
          timeout: 10s
        lsminterval:
```

Adds batching for lsm interval processor as per https://github.com/elastic/opentelemetry-collector-components/issues/211

We would also need to add the `batch/aggs` processor in the `metrics/aggregated-otel-metrics` pipeline. Note that it already seems to have a batch processor but that seems to be too small for aggregation purposes.

lahsivjar (commented):
Adding bigger batch processor for lsm interval processor. The processor aggregates over interval 1m, 10m, and 60m so I have chosen a timeout of 10s.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Thanks @lahsivjar , pushed these changes in https://github.com/elastic/elastic-agent/pull/6444/commits/af831c8a1d922ce3f22103cfbbfce96fee07b9e1

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -33,44 +33,21 @@ collectors:
...
-            enabled: true
-          # Enable in order to skip the SSL certificate Check
-          # tls:
-          #   insecure_skip_verify: true
-          mapping:
-            mode: otel
+        otlp/gateway:
+          # !!! Elastic APM https endpoint WITHOUT the "https://" prefix
+          endpoint: "http://opentelemetry-kube-stack-gateway-collector:4317"
+          compression: none
```
swiatekm, line 48: Why are we disabling compression here?

rogercoll: Ah, nice catch. I think this was a leftover from a configuration that was connecting to the `apm-server`. Now the otlp connections are within the cluster (otlp receiver/exporters), we can enable it again https://github.com/elastic/elastic-agent/pull/6444/commits/d538b4b4f057eec188083ce628a8974875fd990f

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,80 +476,426 @@ collectors:
...
+          default_pipelines: [metrics/otel]
+          error_mode: ignore
+          match_once: true
+          table:
+            - context: metric
+              statement: route() where instrumentation_scope.name == "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver" or IsMatch(instrumentation_scope.name, "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/*")
+              pipelines: [metrics/infra/ecs, metrics/otel]
+        # [Signal To Metrics Connector](https://github.com/elastic/opentelemetry-collector-components/tree/main/connector/signaltometricsconnector)
+        signaltometrics: # Produces metrics from all signal types (traces, logs, or metrics).
+          logs:
```
gizas, line 565: shall we inlcude the orchestrator.cluster.name ? 
It is something that in general needed to filter with.

I added the comment in logs but same for metrics as well

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -33,44 +33,19 @@ collectors:
...
-          api_key: ${env:ELASTIC_API_KEY} # API key for Elasticsearch authentication.
-          logs_dynamic_index:
-            enabled: true
-          # Enable in order to skip the SSL certificate Check
-          # tls:
-          #   insecure_skip_verify: true
-          mapping:
-            mode: otel
+        otlp/gateway:
+          endpoint: "http://opentelemetry-kube-stack-gateway-collector:4317"
```
swiatekm, line 46: Is this name always going to be the same? If it depends on the Helm release name, then we should find a more dynamic way to inject it. Or are we assuming it will always be the same based on Kibana install instructions?

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -204,344 +179,38 @@ collectors:
...
   #  Auto-instrumentation SDKs will use this endpoint.
   daemon:
-    # Configure the pods resources to control CPU and memory usage.
-    resources:
-      limits:
-        cpu: 1500m
-        memory: 1500Mi
-      requests:
-        cpu: 100m
-        memory: 500Mi
```
swiatekm, line 219: Shouldn't we have some resource limits and requests set for the daemonset?

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: @felixbarny Reopening this thread as we have noticed that non apm signals (hostmetrics, filelog, k8sobjects) going through the signaltometrics connector were generating a service entity in Kibana. For example, if `filelog receiver` logs were sent through the `signaltometrics` connector, a service was seen in Kibana for each K8s pod in the cluster (that produces logs).

The workaround that we applied to prevent creating those service is to route filelog data to another pipeline, similar to what we do for hostmetrics. https://github.com/elastic/elastic-agent/pull/6444/commits/ffa6620e1f38dd9d301be48c331d93fbe6cf26ef

cc @gizas

felixbarny: > service was seen in Kibana for each K8s pod in the cluster (that produces logs).

That's because a default `service.name` is extracted based on the pod name, right?

I'm torn whether we should work around that or not. We were working towards the service list to not be APM specific and to be able to have logs/metrics-only services. But I'm not sure if the current approach of signal to metrics is still how we want to get there. Let me ask around and get back to you.

felixbarny: Whatever we end up doing, I think the @elastic/obs-ds-managed-services team should own the exact configuration of the gateway collector, to make sure it's aligned with how our managed OTel collector is configured. Until we have figured out the details, I would propose to revert the workaround in ffa6620e1f38dd9d301be48c331d93fbe6cf26ef and merge this PR. We can tweak the configuration later.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: > If it depends on the Helm release name, then we should find a more dynamic way to inject it.

Totally agree. The current values file is dependent not only on the Helm release name but on the namespace (opentelemetry-operator-system) too:

```yaml
instrumentation:
  name: elastic-instrumentation
  enabled: true # Enable/disable auto-instrumentation.
  exporter:
    endpoint: http://opentelemetry-kube-stack-daemon-collector.opentelemetry-operator-system.svc.cluster.local:4318 # The daemonset OpenTelemetry Collector endpoint where telemetry data will be exported.
```

> Or are we assuming it will always be the same based on Kibana install instructions?

I would say that at the moment the current supported installation approach is with the Kibana instructions.  The purpose of this file is for new users that want to try out OpenTelemetry in K8s with a few commands (new users). Nonetheless, I agree that making these fields dynamic would improve the long term maintenance and usability. Wdyt about this proposal?

1. Replace instrumentation endpoint with node IP, the daemonset opens an OTLP port on each node and the operator injects the `OTEL_NODE_IP` on each instrumented pod:

```yaml
instrumentation:
  name: elastic-instrumentation
  enabled: true # Enable/disable auto-instrumentation.
  exporter:
    endpoint: http://$(OTEL_NODE_IP):4318 # The daemonset OpenTelemetry Collector endpoint where telemetry data will be exported.
```

2. Set a specific name for each set of collectors (it won't be dynamically created):

```yaml
daemon:
  fullnameOverride: opentelemetry-kube-stack-daemon-collector
gateway:
  fullnameOverride: opentelemetry-kube-stack-gateway-collector
```

K8s services are later created using this field.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: > I think the @elastic/obs-ds-managed-services team should own the exact configuration of the gateway collector, to make sure it's aligned with how our managed OTel collector is configured. 

Totally agree.

> I would propose to revert the workaround in https://github.com/elastic/elastic-agent/commit/ffa6620e1f38dd9d301be48c331d93fbe6cf26ef and merge this PR. We can tweak the configuration later.

The problem that I see is that if we revert https://github.com/elastic/elastic-agent/commit/ffa6620e1f38dd9d301be48c331d93fbe6cf26ef and gets released on 8.18.0, the user experience will be different from the previous version (services with no data):

image

What about not generating these new signals until we have a decision?

deploy/helm/edot-collector/kube-stack/values.yaml
swiatekm: For instrumentation, the best approach would be to use a Service with internal traffic policy imo: https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/. But I'm not sure of the Chart lets us set this.

For the gateway, pinning the gateway Service works reasonably, although it prevents multiple installation in the same namespace (maybe multiple installations altogether?). I thought you might be able to get the Helm release name from the `meta.helm.sh/release-name` annotation via the downward API, but it looks like this annotation doesn't appear on collector Pods.

deploy/helm/edot-collector/kube-stack/values.yaml
felixbarny: I think this is in line with what the managed OTel collector will produce as well. Also, the workaround seems a bit brittle as it only works for a specific set of receivers and can't be replicated for logs that are sent from non-EDOT collectors to the managed OTLP endpoint. I'd lean towards not including the workaround but strive for a decision before the release.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: > For instrumentation, the best approach would be to use a Service with internal traffic policy imo: https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/. But I'm not sure if the Chart lets us set this.

This sounds like a good improvement. But I am not sure how we can get rid of the namespace from the daemonset endpoint without modifying the upstream Chart (any namespace should be able to reach the OTLP endpoint). 


> For the gateway, pinning the gateway Service works reasonably, although it prevents multiple installation in the same namespace (maybe multiple installations altogether?)

That's correct, but deploying this values file in the same cluster multiple times would require changes either way (e.g `hostNetwork: false` for the daemonset collector, to prevent port collision).

deploy/helm/edot-collector/kube-stack/values.yaml
felixbarny: Could you share a screenshot of how the service overview of one of these logs-only services look like? Does the logs tab show the logs for that service?

rogercoll: Sure, unfortunately the logs tab does not show up:

image

deploy/helm/edot-collector/kube-stack/values.yaml
swiatekm: > This sounds like a good improvement. But I am not sure how we can get rid of the namespace from the daemonset endpoint without modifying the upstream Chart (any namespace should be able to reach the OTLP endpoint).

The proper way to fix this would actually be to provide some mechanism to indicate that a given Instrumentation CR should export data to a specific Collector CR. This is a feature we've wanted to have in the operator for a while, but it can also be implemented in the upstream Helm Chart by injecting the right environment variables into the Instrumentation CR.

In any case, this PR is fine as-is.

deploy/helm/edot-collector/kube-stack/values.yaml
roshan-elastic: I think showing the services makes sense. It's the lack of 'actionability' that I think is a bit clunky for the user I think.

For example, as a user I would be asking:

- Why are metrics like latency missing on these services and how do I add them?
- I see a lot of empty charts, is this service OK? How do can I use the data I am collecting?

The work we did to adapt the service inventory and service views to work with only logs but both required the entities API to work (so are behind FF) and we deprecated the services inventory work (because it inefficient to iterate on this and the new inventory at the same time):

_Service Inventory w logs (deprecated)_
image


_Adaptive Service Views w logs (there is logs-focused overview page but I can't find the PR)_
service views
 (hidden behind `entityCentricExperience` FF - off by default)_

I'm not sure how much we can do here without significant investment.

@alex-fedotyev @smith what do you think?

felixbarny: Likely, the reason why the `konnectivity-agent` service in the screenshot above shows no data at all is because it only emits metrics (no logs, no traces). The service overview page isn't a great fit for metrics-only services at the moment. So we were discussing to remove the configuration that "spawns a service" based in it emitting just metrics. That should solve a lot of the issues.

> The work we did to adapt the service inventory and service views to work with only logs but both required the entities API to work (so are behind FF) and we deprecated the services inventory work (because it inefficient to iterate on this and the new inventory at the same time):

Can we decouple the entity-based service inventory page from adapting the service overview page to be useful for logs only services? The latter seems to be useful to have even without the rest of the services inventory work.

rogercoll: What it is being used to identify a service in Kibana?  With the previous screenshot context, we are sending logs for the `konnectivity-agent` with a resource attribute `service.name: konnectivity-agent`.

Following the Kibana error message (`cannot get properties of undefined setting agent.name`), I extended the configuration to add the `agent.name` resource attribute based on the `service.name` value into all the logs gathered by the filelog receiver and it unblocked the logs and metrics tab within the Services view:

image

felixbarny: I think this has to do with the fact that we enrich the `agent.name` with the elastictraceprocessor for traces but not for metrics or logs. Adding a default value for `agent.name` is probably the right short-term workaround. Even better if we could make the UI not require the field to be present.

rogercoll: Workaround reverted and agent.name set as `edot-collector` (open to any other naming suggestion) https://github.com/elastic/elastic-agent/pull/6444/commits/b563651c26faec9a098b1af6f39bde56566c423d

The initial idea was to set something more similar to what [we do for traces](https://github.com/elastic/opentelemetry-lib/blob/main/enrichments/trace/internal/elastic/resource.go#L78) (e.g. opentelemetry/collector/elastic) but then is detected as a language in the UI:

image

felixbarny: Looks good! However, I'd suggest not adding an `agent.name` attribute to the original data (logs/metrics) but instead apply a default value in the signaltometrics configuration, to avoid spoiling the data unnecessarily with APM/Elastic-specific things.

That means replacing usages of 
```yaml
include_resource_attributes:
  - key: agent.name # set via elastictraceprocessor
```

with 
```yaml
include_resource_attributes:
  - key: agent.name
    default_value: unknown
```

rogercoll: That makes a lot of sense to me. Also, `unknown` seems a reasonable value to prevent users from configuring dasboards/alerts with the agent.name `edot-collector`.

cc @AlexanderWert @gregkalapos

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -841,80 +476,437 @@ collectors:
...
+            processors:
+              - elasticinframetrics
+              - attributes/dataset
+              - resource/process
+              - batch
+            exporters: [debug, elasticsearch/ecs]
+          metrics/otel:
+            receivers: [routing]
+            processors: [batch]
+            exporters: [debug, signaltometrics, elasticsearch/otel]
```
felixbarny, line 896: I'm a bit on the fence whether to send metrics through the signaltometrics connector as we don't have a great experience for metrics-only services. I think I'd lean towards removing it now as we can always easily add it back in later. But either way, I don't see this as blocking.

felixbarny (approved):
LGTM! Let's get it in and iterate from there.

deploy/helm/edot-collector/kube-stack/values.yaml
rogercoll: Let's remove then for now https://github.com/elastic/elastic-agent/pull/6444/commits/c83cd8f1550101dea5bcd9ac5885a495e4c49561

Thanks for the review!

rogercoll:
@elastic/elastic-agent-control-plane PR back to ready for review

pkoutsovasilis (approved):

swiatekm (approved):

PR #6461, opened 2024-12-31
cleanup(2489): removed path.install option from the paths command, removed any usage of this command

### LINKED ISSUE #2489: Remove `--path.install` option
As per the conversation starting from https://github.com/elastic/ingest-dev/issues/1609#issuecomment-1503384831, the `--path.install` option is no longer being used by Elastic Agent.  We should remove it but, since that would be a breaking change, we should do this in 9.0.

Copying relevant bits of the conversation here for easy reference:

@blakerouse said:
> I believe with the changes in the V2 architecture (aka 8.6+) we missed removing some command-line arguments. `--path.install` has a different meaning then what this issue is seeking to solve. It is documented as `Install path contains binaries Agent extracts`, which meant the binaries that Elastic Agent extracted to run (aka. filebeat, metricbeat, etc..). Obviously with V2 (8.6+) we no longer extract the binaries they come pre-extracted so we no longer need or use the `--path.install`.

@ycombinator asked:
> Shouldn't we wait to remove the --path.install option in 9.0 since it would technically be a breaking change to remove it in 8.x? 

@jlind23 confirmed:
> Let's wait for [9.0] indeed.

kaanyalti:
- Breaking change

## What does this PR do?

Removes `--path.install` flag declaration from the paths command and removes its use from container and enroll commands

## Why is it important?

This command is currently ignored and deprecated

## Checklist

## Disruptive User Impact

Any client using `--path.install` flag will fail

- Deploy ess
- Prepare necessary packages for testing
  - tar.gz
  - docker
- Try to install tar.gz with `path.install` and verify that the installation fails
- Install tar.gz without the flag and validate that the installation is successful
- Unenroll the agent on fleet ui, and then enroll the agent again and validate that enrollment is successful
- Run agent in docker and validate that it is successfully run

## Related issues

- Closes #2489

### DIFF
--- a/changelog/fragments/1735664420-remove-deprecated-path-install-cli-flag.yaml
+++ b/changelog/fragments/1735664420-remove-deprecated-path-install-cli-flag.yaml
@@ -1,0 +1,31 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: breaking-change
+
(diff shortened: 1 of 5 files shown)


michalpristas (approved):

jlind23:
Removing the backport label as this should remain on main.

Issue #6469, opened 2025-01-02
Cloud Defend binary missing from 8.17.0+ containers


cmacknz:
```
docker run -it --entrypoint /bin/bash docker.elastic.co/elastic-agent/elastic-agent:8.17.0
elastic-agent@f048b436b7c0:~$ ls data/elastic-agent-96f2b9/components/
LICENSE.txt                       bundle.tar.gz                    java-attacher.jar
NOTICE.pf-elastic-collector.txt   certs                            lenses
NOTICE.pf-elastic-symbolizer.txt  checksum.yml                     module
NOTICE.pf-host-agent.txt          cloudbeat                        osquery-extension.ext
NOTICE.txt                        cloudbeat.spec.yml               osqueryd
README.md                         cloudbeat.yml                    pf-elastic-collector
agentbeat                         endpoint-security                pf-elastic-collector.spec.yml
agentbeat.spec.yml                endpoint-security-resources.zip  pf-elastic-symbolizer
apm-server                        endpoint-security.spec.yml       pf-elastic-symbolizer.spec.yml
apm-server.spec.yml               fleet-server                     pf-host-agent
apm-server.yml                    fleet-server.spec.yml            pf-host-agent.spec.yml
```

It exists in 8.16.x but is missing from 8.17.0, 8.17.0-SNAPSHOT, and 8.18.0-SNAPSHOT.

It was removed intentionally in 9.0 but the change in elastic agent was not backported as far as I can tell: https://github.com/elastic/elastic-agent/pull/5481

cmacknz:
Reproduces locally. cloud-defend was removed from the list of expected binaries to package in https://github.com/elastic/elastic-agent/pull/6042/files, likely as part of merge conflict resolution as it no longer exists on main.

swiatekm:
I wonder what kind of test could've caught this. It's easy to miss changes in these kinds of lists during review. Maybe if we had separate test cases checking the container contents, it'd be harder to remove them accidentally this way?

PR #6470, opened 2025-01-02
[8.x] restore cloud defend as an expected binary

cmacknz:
The cloud-defend binary was accidentally removed from 8.x starting with 8.17.0 in the backport for https://github.com/elastic/elastic-agent/pull/6042/files, likely as part of merge conflict resolution.

## What does this PR do?

Includes the cloud-defend binary in the agent docker container again.

## Why is it important?

Defend for containers which is implemented by cloud-defend is still a supported product in 8.x.

```sh
DOCKER_VARIANTS=basic EXTERNAL=true SNAPSHOT=true PACKAGES=docker PLATFORMS=linux/arm64 mage package

docker run -it --entrypoint /bin/bash docker.elastic.co/beats/elastic-agent:8.18.0-SNAPSHOT
elastic-agent@240c9b878938:~$ ls data/elastic-agent-1f89c7/
components  elastic-agent  logs  otelcol  package.version
elastic-agent@240c9b878938:~$ ls data/elastic-agent-1f89c7/components/
LICENSE.txt                       bundle.tar.gz          lenses
NOTICE.pf-elastic-collector.txt   certs                  module
NOTICE.pf-elastic-symbolizer.txt  checksum.yml           osquery-extension.ext
NOTICE.pf-host-agent.txt          cloud-defend           osqueryd
NOTICE.txt                        cloud-defend.spec.yml  pf-elastic-collector
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6469

### DIFF
--- a/changelog/fragments/1735833843-Restore-cloud-defend..yaml
+++ b/changelog/fragments/1735833843-Restore-cloud-defend..yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Restore the cloud-defend binary which was accidentally removed in 8.17.0.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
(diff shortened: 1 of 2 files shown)


cmacknz:
We have package tests that could catch this. https://github.com/cmacknz/elastic-agent/blob/622b4f09d7dad52f55ea1b2c029a554260f24dc9/dev-tools/packaging/package_test.go#L287

My first instinct was to add a method on the [ExpectedBinaries](/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/dev-tools/mage/manifest/manifest.go) list that would return the set of binaries expected for each platform and then check that the files exist in a test. This sounds great except that it would not have caught this.

We would need to duplicate the list into the tests with a comment that it is duplicated on purpose or something. Really I think this type of problem should be getting caught at the integration level (as in Fleet integration package tests not agent integration tests).

swiatekm (approved):

cmacknz:
Tests for this would be better placed in https://github.com/elastic/integrations/tree/main/packages/cloud_defend

PR #6475, opened 2025-01-03
[helm] reload chart for each example folder

pkoutsovasilis:
## What does this PR do?

This PR fixes an issue with the Helm chart examples rendering. Specifically, the Helm chart is now reloaded for each example folder. Previously, the chart was loaded once and reused, which leads to stale data when different examples enable or disable subchart dependencies (spoiler alert for the upcoming kube-state-metrics subchart dependency PR I am working on).

## Why is it important?

Reloading the Helm chart for each example folder ensures correctness in the rendered k8s manifests.

## Checklist

```
mage helm:renderExamples
```

## Related issues

N/A

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -3417,12 +3417,7 @@ func (Helm) RenderExamples() error {
 	settings := cli.New() // Helm CLI settings
 	actionConfig := &action.Configuration{}
 
-	helmChart, err := loader.Load(helmChartPath)
-	if err != nil {
-		return fmt.Errorf("failed to load helm chart: %w", err)
-	}
-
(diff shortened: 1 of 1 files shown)


swiatekm:
Does running init mean that we do a dependency pull on each invocation?

> Previously, the chart was loaded once and reused, which leads to stale data when different examples enable or disable subchart dependencies (spoiler alert for the upcoming kube-state-metrics subchart dependency PR I am working on).

Does this also happen if the dependency is controlled via a condition in the values file?

swiatekm (approved):
I don't think this change has any negative side effects, but it feels like we're working around a bug in Helm.

PR #6477, opened 2025-01-04
Added loggers for Fleet enrollment call failures

### LINKED ISSUE #6287: `Remote server failed to handle the request, will retry in a moment` needs to be better handled
When trying to communicate with Fleet-Server (e.g: during enrol), Elastic-Agent might log the error
```
Remote server failed to handle the request, will retry in a moment
```
However this can be the result of 3 different HTTP status codes:
https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/fleetapi/enroll_cmd.go#L35-L40

Bad Gateway and Gateway Timeout are likely connectivity problems, however Service Unavailable can be a real problem with Fleet-Server.

We need to be more clear in the logs which error has happened and likely better handle them individually. E.g: Bad Gateway might be a configuration issue, while Gateway Timeout is a connectivity issue. It might not make sense to retry Bad Gateway, while Gateway Timeout should be retried with an exponential backoff.

Rohit-code14:
- Enhancement

## What does this PR do?
Added logger to print the status and code when enrollment call to fleet failed.

## Why is it important?
As mentioned in the issue(#6287), It will be difficult to debug enrollment failures in such cases. Therefore adding the actual status code and error in logs will be helpful for debugging.

## Checklist

## Disruptive User Impact

No disruptive user impact

## Related issues

- Closes #6287


pkoutsovasilis:
buildkite test this

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -524,6 +524,8 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
 	}
 
 	c.log.Infof("1st enrollment attempt failed, retrying enrolling to URL: %s with exponential backoff (init %s, max %s)", c.client.URI(), enrollBackoffInit, enrollBackoffMax)
+	c.log.Warn(err.Error())
```
pkoutsovasilis, line 527:
```suggestion
	c.log.Warn("1st enrollment attempt failed: %s", err.Error())
	c.log.Infof("Retrying enrolling to URL: %s with exponential backoff (init %s, max %s)", c.client.URI(), enrollBackoffInit, enrollBackoffMax)
```

pkoutsovasilis: Also I would propose to extend the `if` - `else if` branches below in the retrying for loop (maybe change that to switch case for readability?!) and add a `err != nil` last `else if` that logs the error. This will log it only once and we will know why the retry loop broke 🙂

Rohit-code14: Thanks for the suggestions. Added appropriate changes. Have made failure logger as Info print for better understanding for the user.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -523,22 +523,23 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
 		return nil
 	}
 
-	c.log.Infof("1st enrollment attempt failed, retrying enrolling to URL: %s with exponential backoff (init %s, max %s)", c.client.URI(), enrollBackoffInit, enrollBackoffMax)
-	c.log.Warn(err.Error())
+	c.log.Infof("1st enrollment attempt failed: %s", err.Error())
```
pkoutsovasilis, line 526: this probably should be `Warnf` right? 🙂

Rohit-code14: Wouldn't it be better to show the failure reason in Info? better visibility.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -523,22 +523,23 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 			retry = true
-		} else if errors.Is(err, fleetapi.ErrConnRefused) {
+		case fleetapi.ErrConnRefused.Error():
 			c.log.Warn("Remote server is not ready to accept connections, will retry in a moment.")
 			retry = true
-		} else if errors.Is(err, fleetapi.ErrTemporaryServerError) {
+		case fleetapi.ErrTemporaryServerError.Error():
 			c.log.Warn("Remote server failed to handle the request, will retry in a moment.")
 			retry = true
 		}
```
pkoutsovasilis, line 545: how about this approach? 🙂 

```go
switch {
case errors.Is(err, fleetapi.ErrTooManyRequests):
	c.log.Warn("Too many requests on the remote server, will retry in a moment.")
	retry = true
case errors.Is(err, fleetapi.ErrConnRefused):
	c.log.Warn("Remote server is not ready to accept connections, will retry in a moment.")
	retry = true
case errors.Is(err, fleetapi.ErrTemporaryServerError):
	c.log.Warn("Remote server failed to handle the request, will retry in a moment.")
	retry = true
case err != nil:
	c.log.Warn("Enrollment failed: %s", err.Error())
}
```

internal/pkg/agent/cmd/enroll_cmd.go
pkoutsovasilis: Aren't warnings more visible than info messages 🙂

Rohit-code14: Got it! Will make the changes

internal/pkg/agent/cmd/enroll_cmd.go
Rohit-code14: Missed it, Thanks.. will work on your suggestions.

Rohit-code14: `		if err != nil {
			c.log.Warnf("Enrollment attempt failed: %s", err.Error())
		}
`

I think it would be better to keep this if condition and remove the last case in the suggested switch, as the actual error will not be printed if it matches any of the known cases which does not align with our goal here.

Rohit-code14: Also we could break in case err is nil

pkoutsovasilis: I would still go with the switch case but change the `c.log.Warn(...)` to include the `err.Error()` for the `case errors.Is(err, fleetapi.ErrTemporaryServerError)`. However, after looking again at the existing code, there is something missing; we should break out the loop if `backExp.Wait()` returns false which means that it is time to give up 🤔 

```go
// Wait blocks until either the exponential backoff timer is completed or the
// done channel is closed.
// Wait returns true until done is closed. When done is closed, wait returns
// immediately, therefore callers should always check the return value.
func (b *ExpBackoff) Wait() bool {
	b.duration = b.NextWait()

	select {
	case <-b.done:
		return false
	case <-time.After(b.duration):
		b.last = time.Now()
		return true
	}
}
```

Rohit-code14: Not only here but I could see multiple places where return value is not checked when Wait() is invoked.

Rohit-code14: @pkoutsovasilis  I have made the suggested changes. Kindly check the same once. Regarding the Wait() method, Can we have it as a separate issue?

internal/pkg/fleetapi/enroll_cmd.go
```diff
@@ -24,19 +24,19 @@ import (
 type EnrollType string
 
 // ErrTooManyRequests is received when the remote server is overloaded.
-var ErrTooManyRequests = errors.New("too many requests received (429)")
+var ErrTooManyRequests = errors.New("Received status 429(TooManyRequest)")
```
pkoutsovasilis, line 27: I kinda like this guideline 🙂 https://go.dev/wiki/CodeReviewComments#error-strings 

> Error Strings
> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -524,21 +524,25 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 	defer close(signal)
 	backExp := backoff.NewExpBackoff(signal, enrollBackoffInit, enrollBackoffMax)
 
 	for {
 		retry := false
-		if errors.Is(err, fleetapi.ErrTooManyRequests) {
-			c.log.Warn("Too many requests on the remote server, will retry in a moment.")
+		switch {
+		case errors.Is(err, fleetapi.ErrTooManyRequests):
+			c.log.Warnf("%s. Too many requests on the remote server, will retry in a moment.", err.Error())
```
pkoutsovasilis, line 536: No need to print err.Error() the message is already self-describing? 🙂

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -524,21 +524,25 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 			retry = true
-		} else if errors.Is(err, fleetapi.ErrConnRefused) {
-			c.log.Warn("Remote server is not ready to accept connections, will retry in a moment.")
+		case errors.Is(err, fleetapi.ErrConnRefused):
+			c.log.Warnf("%s. Remote server is not ready to accept connections, will retry in a moment.", err.Error())
```
pkoutsovasilis, line 539: Again I don't see the reason to log err.Error() this error maps to exactly one root cause; maybe reword the previous message to something like "Remote server is not ready to accept connections (Connection Refused), will retry in a moment."

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -524,21 +524,25 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 			retry = true
-		} else if errors.Is(err, fleetapi.ErrTemporaryServerError) {
-			c.log.Warn("Remote server failed to handle the request, will retry in a moment.")
+		case errors.Is(err, fleetapi.ErrTemporaryServerError):
+			c.log.Warnf("%s. Remote server failed to handle the request, will retry in a moment.", err.Error())
```
pkoutsovasilis, line 542: Ok here we want err.Error() as this error can have multiple reasons; maybe something like "Remote server failed to handle the request (%s), will retry in a moment."?

internal/pkg/agent/cmd/enroll_cmd.go
pkoutsovasilis: after looking at the Wait() use-case once more I believe this is not meant to give up it must continuously try to enroll with a maximum backoff delay, thus no need for another reason 🙂

pkoutsovasilis:
buildkite test this

pkoutsovasilis:
buildkite test this

Rohit-code14:
@pkoutsovasilis  not sure why the buildkite test is failing..shall i merge latest changes to my branch once again? And will i be able to trigger test by commenting `buildkite test this`  or is it only for elastic team members?

michel-laterman:
buildkite test this

pkoutsovasilis (approved):
LGTM thanks for the contribution @Rohit-code14

PR #6485, opened 2025-01-07
[k8s] fix hints stream missing ids

pkoutsovasilis:
## What does this PR do?

After [this PR](https://github.com/elastic/beats/pull/41954) got merged our k8s [CI test started failing](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/5958#01943e1b-bb24-43f8-8da0-2da5f265e1ab/60-358). After investigating I realised that all the hint-based inputs are lacking streams IDs in the respective filestream inputs resulting in all of them registering with an empty ID and thus colliding. Also, during this investigation, I realised that by default when hints are enabled are capturing the logs of containers that do not feature a hints-related annotation. Thus I updated the Helm chart to explicitly disable the former when both the contailer logs of kubernetes integration and hints are enabled. 

A fix should also happen in the actual job that fabricates and introduces these templates

## Why is it important?

Because it fixes an actual bug and it brings the CI back to green 

## Checklist

## Disruptive User Impact

None

```
mage integration:auth
```

## Related issues

N/A

### DIFF
--- a/deploy/helm/elastic-agent/examples/kubernetes-hints-autodiscover/rendered/manifest.yaml
+++ b/deploy/helm/elastic-agent/examples/kubernetes-hints-autodiscover/rendered/manifest.yaml
@@ -564,6 +564,7 @@ stringData:
     providers:
       kubernetes:
         hints:
+          default_container_logs: false
           enabled: true
         node: ${NODE_NAME}
         scope: node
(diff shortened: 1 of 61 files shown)


pkoutsovasilis:
buildkite test extended

swiatekm:
Looks like we need to add the ID [here](https://github.com/elastic/package-storage-infra/blob/f891ec0d01a6ea3794eddcd000ac5fabb4ccead8/jobs/refreshk8stemplates/job.go#L550). I'll take care of it.

swiatekm (approved):

PR #6486, opened 2025-01-07
Add feature flags for running beats as OTel receivers

andrzej-stencel:
## What does this PR do?

Adds new Agent feature flag "beats_as_otel_receivers". This is for internal usage only. No documentation or changelog is added for this.

## Checklist

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -20,10 +20,12 @@ import (
 // 8.10  - default is disabled
 // 8.11+ - default is enabled
 const defaultTamperProtection = true
+const defaultBeatsAsOtelReceivers = false
 
 var (
 	current = Flags{
-		tamperProtection: defaultTamperProtection,
+		tamperProtection:     defaultTamperProtection,
+		beatsAsOtelReceivers: defaultBeatsAsOtelReceivers,
 	}
 )
 
@@ -37,6 +39,8 @@ type Flags struct {
 	fqdnCallbacks map[string]BoolValueOnChangeCallback
 
 	tamperProtection bool
+
+	beatsAsOtelReceivers bool
 }
 
 type cfg struct {
@@ -48,6 +52,9 @@ type cfg struct {
 			TamperProtection *struct {
 				Enabled bool `json:"enabled" yaml:"enabled" config:"enabled"`
 			} `json:"tamper_protection,omitempty" yaml:"tamper_protection,omitempty" config:"tamper_protection,omitempty"`
+			BeatsAsOtelReceivers *struct {
+				Enabled bool `json:"enabled" yaml:"enabled" config:"enabled"`
+			} `json:"beats_as_otel_receivers,omitempty" yaml:"beats_as_otel_receivers,omitempty" config:"beats_as_otel_receivers,omitempty"`
 		} `json:"features" yaml:"features" config:"features"`
 	} `json:"agent" yaml:"agent" config:"agent"`
 }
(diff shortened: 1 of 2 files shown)


andrzej-stencel:
> I worry that this feature flag is too broad. We might want to define a few flags to enable this. We want to start with monitoring only so maybe `monitoring_as_beatsreceivers` would be better. I see `beats_as_otel_receivers` as meaning that all beat inputs are converted to otel beatsreceivers. I do think we want to get to that level, but we might want to do it in stages. A user might just want to turn on monitoring for now, test it out then turn it on the others once we support those as well. I think if we don't add some levels of enablement here, it because an all on or nothing and that is probably not something we would want.
> 
> I was leaning this more to a flag in the monitoring configuration so it just turned on specifically to monitoring.

I agree we might want separate feature flags for monitoring and for data collection. Given that this feature flag is not documented, I believe we can always change, rename, delete, add new flags later, right?

I think it would be best to add a specific feature flag together with the functionality it implements, as a single task. A separate task for just creating a feature flag in isolation from the functionality it guards is prone to deliberations like this.

I propose to close this task and add new feature flag(s) as part of one or both of the tasks:
- https://github.com/elastic/ingest-dev/issues/3517
- https://github.com/elastic/ingest-dev/issues/3984

Alternatively, I can create two separate feature flags for monitoring and data collection as part of this task. As I said, we can always modify the approach later.

pkg/features/features.go
```diff
@@ -37,6 +41,12 @@ type Flags struct {
 	fqdnCallbacks map[string]BoolValueOnChangeCallback
 
 	tamperProtection bool
+
+	// beatsAsOtelReceivers indicates if the Agent run Beats as OpenTelemetry receivers like filebeatreceiver etc.
+	beatsAsOtelReceivers bool
+
+	// monitoringWithOtel indicates if the Agent should use OpenTelemetry components for self-monitoring instead of Beats.
+	monitoringWithOtel bool
```
blakerouse, line 49: I think there is a case where we want to run monitoring with native otel in the future. So we might just want to use a little different name. What about `monitoringWithOtelReceivers`? Just so in the future we want to add `monitoringWithOtel` where it is using native filelog receiver and a native metric receiver.

pkg/features/features.go
```diff
@@ -37,6 +41,12 @@ type Flags struct {
 	fqdnCallbacks map[string]BoolValueOnChangeCallback
 
 	tamperProtection bool
+
+	// beatsAsOtelReceivers indicates if the Agent run Beats as OpenTelemetry receivers like filebeatreceiver etc.
+	beatsAsOtelReceivers bool
```
blakerouse, line 46: I don't think this should just be a boolean. I think this needs to be an array of input types to run as otel beatsreceivers. That way the feature can be enable per input type at a time for the user. They will want to test out 1 input, see if it works then enable the next type. They will not want to be a change all. Possible that a `*` can be used to mean all of them, but I think we need a way to limit it to specific input types.

andrzej-stencel: Right. Let me [repeat myself](https://github.com/elastic/elastic-agent/pull/6486#issuecomment-2576948261):

> I think it would be best to add a specific feature flag together with the functionality it implements, as a single task. A separate task for just creating a feature flag in isolation from the functionality it guards is prone to deliberations like this.

I propose to close this pull request and create the feature flag(s) together with the functionality.

blakerouse: I agree it would just be better to add these when the features are each implemented. I don't know how that will affect the size of the PR when adding the feature, so having it seperate could help with reviewing.

I am good with just closing it and doing it in the implementation.

andrzej-stencel:
Closing this PR. The relevant feature flags will be created as part of the functionality.

PR #6494, opened 2025-01-08
Don't restart agent after enrolling if installed via deb or rpm

swiatekm:
## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/4084

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/cmd/enroll.go
+++ b/internal/pkg/agent/cmd/enroll.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/elastic/elastic-agent/internal/pkg/agent/install/pkgmgr"
+
 	"github.com/spf13/cobra"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application"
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/cmd/enroll.go
```diff
@@ -478,6 +480,12 @@ func enroll(streams *cli.IOStreams, cmd *cobra.Command) error {
 		fixPermissions = nil
 	}
 
+	// If agent was installed via a deb or rpm package, we want to avoid restarting because it may not be running yet.
+	// In this case, the user is expected to (re)start agent via systemd instead.
+	if pkgmgr.InstalledViaExternalPkgMgr() {
+		skipDaemonReload = true
+	}
```
blakerouse, line 487: Why is this only for external package manager? Wouldn't it be good to make this change for all installation types?

swiatekm:
This was an attempt to make the agent work in-line with the documentation, but it turns out the documentation itself was fixed instead. Closing for now, we can revisit this later if we need to, possibly making a more systemic change.

PR #6498, opened 2025-01-08
Add ability to enroll with defined ID and replace_token

### LINKED ISSUE #6361: Add ability to define the specific agent-id during enroll
**Describe the enhancement:**

A new flag `--agent-id` should be added to the `enroll` command to allow the Elastic Agent to define the ID it wants to use when enrolling with Fleet Server. An `ELASTIC_AGENT_ID` container environment variable also needs to work the same way as the `--agent-id` on the `enroll` command.

This allows an Elastic Agent to tell Fleet Server that this Elastic Agent is this specific agent.

Most use-cases are described in the sibling issue on fleet-server: https://github.com/elastic/fleet-server/issues/4226

When `--agent-id` and `ELASTIC_AGENT_ID` will enroll to the Fleet Server with the provided ID.

blakerouse:
## What does this PR do?

Adds new `--id` (`ELASTIC_AGENT_ID` env for container) and `--replace-token` (`FLEET_REPLACE_TOKEN` env for container) to enrollment.

## Why is it important?

Allows an Elastic Agent to enroll with a defined ID and replacement token to allow it to replace an existing Elastic Agent. The original Elastic Agent must have also been enrolled with the same `--replace-token` or it will not be allow to enroll if the `--id` collides with an existing Elastic Agent.

## Checklist

- ~~[ ] I have made corresponding change to the default configuration files~~ (no config changes)
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (cannot unit test)

## Disruptive User Impact

None

You need this PR (https://github.com/elastic/fleet-server/pull/4290) from Fleet Server and have that Fleet Server running to enroll. Then you can build this PR and run enroll with `--id` and `--replace-token`.

## Related issues

- Closes #6361


pkg/testing/fixture_install.go
```diff
@@ -601,8 +609,9 @@ func (f *Fixture) installRpm(ctx context.Context, installOpts *InstallOpts, shou
 }
 
 type UninstallOpts struct {
-	Force          bool // --force
-	UninstallToken string
+	Force          bool   // --force
+	UninstallToken string // --uninstall-token
+	SkipFleetAudit bool   // --skip-fleet-audit
```
pkoutsovasilis, line 614: this param `--skip-fleet-audit` is not mentioned int he PR description. Is this a left-over? if not, please do update the description to capture it

testing/integration/enroll_replace_token_test.go
```diff
@@ -0,0 +1,118 @@
...
+	fixture, err = define.NewFixtureFromLocalBuild(t, define.Version())
+	require.NoError(t, err)
+
+	output, err = fixture.Install(ctx, &installOpts)
+	if err != nil {
+		t.Log(string(output))
+		t.Fatalf("failed installing the agent again: %s", err)
+	}
+
+	t.Logf(">>> Enroll succeeded again. Output: %s", output)
```
pkoutsovasilis, line 108: you don't know if this enrollment is actually a different/newer one compared to the previous enrollment you did above. IMO, you need to extend `FleetAgentStatusByAgentID` to also expose and check enrolled_at ([api ref](https://www.elastic.co/docs/api/doc/serverless/operation/operation-get-fleet-agents-agentid#operation-get-fleet-agents-agentid-200-body-application-json-elastic-api-version-2023-10-31-item-enrolled_at)

testing/integration/enroll_replace_token_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.Eventually(
+		t,
+		check.FleetAgentStatusByAgentID(ctx, t, info.KibanaClient, agentID, "online"),
+		10*time.Minute,
+		10*time.Second,
+		"Elastic Agent status is not online",
+	)
+
+	// Force uninstall the Agent (skipping information Fleet)
+	output, err = fixture.Uninstall(ctx, &atesting.UninstallOpts{
```
pkoutsovasilis, line 88: what if this agent is running through a container and replace-token and/or agent-id change?! shouldn't it re-enroll?

testing/integration/enroll_replace_token_test.go
blakerouse: The `enrolled_at` doesn't change because in the eyes of Fleet. This Elastic Agent didn't become unenrolled and then enrolled so the `enrolled_at` is not reset, see here - https://github.com/elastic/fleet-server/pull/4417/files#diff-b3d2d3a49ef214fba08e3aa10772162cbb65c378e24ef6c4a0c5df65b051f71aR372

I should be able to set the headers on enrollment, to identify the difference.

blakerouse: I don't think it should because it is not a new agent, it is just replacing an existing agent and becoming the same agent.

pkoutsovasilis: I still have that in my head as something that should be updated or at least the comment in the code on the fleet-side makes me anticipate it 🙂 But I guess you know the intricacies of Fleet and Kibana UI better so if we did that we might even break something there. Let's skip it for now, and if required we might introduce even a new field to capture the replacements in the future?!

internal/pkg/agent/cmd/container.go
```diff
@@ -1031,7 +1045,21 @@ func shouldFleetEnroll(setupCfg setupConfig) (bool, error) {
...
 	}
 
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Check if enrolling with a specifically defined Elastic Agent ID.
+	// If the ID's don't match then it needs to enroll.
+	if setupCfg.Fleet.ID != "" {
+		ai, err := getAgentInfo(ctx, false)
```
pkoutsovasilis, line 1054: `getAgentInfo` seems to be loading the config exactly like we do below. Since you need only the ID can't we just use  `storedConfig.Settings.ID` right after the `storedConfig, err := configuration.NewFromConfig(cfg)`  below?

blakerouse: Duh, I was so used to using the `AgentInfo` flow all over the code that I didn't even look at how it was working under the hood and see that it was loading the same encrypted store.

I have updated it to just check the results from the encrypted store.

jlind23:
@theletterf @nimarezainia do we have any follow up actions here in order to document this for Kubernetes use case at least?
cc @ebeahan

nimarezainia:
> @theletterf @nimarezainia do we have any follow up actions here in order to document this for Kubernetes use case at least? cc @ebeahan

@jlind23 I don't see an issue if we wanted to document this option for users who want to reuse the agent.id - but we haven't documented this yet as far as I know.

theletterf:
I take this is for Elastic Agent docs, not EDOT docs, right? I'd be happy to review a docs PR on this.

nimarezainia:
yes at the moment only applies to elastic agent docs.

jlind23:
@nimarezainia are you taking this one or shall I kindly ask @ebeahan to find someone?

nimarezainia:
@jlind23  maybe I'm confused. This ask is for just documentation of the install flag `--replace-token` and the description above by Blake is explaining it well.

theletterf:
I guess Julien was hinting precisely at who would open the documentation PR. :)

I'll check with the team and see to it.

vishaangelova:
I’ve created a [docs issue](https://github.com/elastic/docs-content/issues/2491) to document this, so it’s on our radar.

PR #6542, opened 2025-01-17
Changes to binary distribution

michalpristas:
> Ready for review but DO NOT MERGE before spec files of dependencies are updated in other repos

## What does this PR do?

This PR changes logic of installation and ugprades in following manner.

Default `install` installs only:
  - agentbeat
  - endpoint-security
  - pf-host-agent

additional flag is added that includes components above and:
  - cloudbeat
  - apm-server
  - fleet-server
  - pf-elastic-symbolizer
  - pf-elastic-collector

Upon installation user choice is backed into `.flavor` file. 

Part of elastic agent package is `.flavors` file defining flavor for this particular version. 
Upon upgrade or install `.flavors` is read and only components specified in desired flavor (using flag with `install` or content of `.flavor` in case of `upgrade`) are copied. 

In case of any issues with parsing any of the above files we fall back to previous way and copy everything.

Each components needs to define files in addition to binary, spec and config in a form 
```yml
component_files:
 - file1
 - dir1/*
 ```
in their spec file. 

During upgrade target version spec file is parsed. not the current installation.

## Why is it important?

Reducing space on disk by omitting not needed components.

## Checklist

- [ ] I have made corresponding changes to the documentation TODO: @michalpristas 

#### Scenario 1 and 2 (installation)
Build package and install without and with flag `--install-server`
check components present

#### Scenario 3 - upgrade to future version
Prepare same agent but with different commit hash (do small change and commit locally)
Run `upgrade 9.0.0-SNAPSHOT --source-uri="file://....zip" --skip-verify` and check only desired components are present

#### Scenario 4 - upgrade from stale version
Upgrade from previous version to this one and check ALL components are present

### CODE REVIEW

**Comment by cmacknz** [comment]
You will need to add integration tests to make sure upgrades respect the flavor or lack of one.

For the existing upgrade tests you could probably add a check to ensure every component exists in both versions easily. For flavors you'll need new tests for those case most likely.

File: internal/pkg/agent/application/upgrade/step_unpack_test.go [hunk 1]
```diff
@@ -190,7 +190,7 @@ func TestUpgrader_unpackTarGz(t *testing.T) {
 			archiveFile, err := tt.args.archiveGenerator(t, tt.args.archiveFiles)
 			require.NoError(t, err, "creation of test archive file failed")
 
-			got, err := untar(log, archiveFile, testDataDir)
+			got, err := untar(log, archiveFile, testDataDir, false, "")
```
cmacknz, line 193: You should be able to unit test that the logic to respect the flavors when unpacking actually works.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -231,7 +231,12 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	u.log.Infow("Unpacking agent package", "version", newVersion)
 
 	// Nice to have: add check that no archive files end up in the current versioned home
-	unpackRes, err := u.unpack(version, archivePath, paths.Data())
+	// default to extended flavor to avoid breaking behavior
+
+	// no default flavor, keep everything in case flavor is not
```
cmacknz, line 236: > keep everything in case flavor is not

Is not what?

internal/pkg/agent/install/.flavors
```diff
@@ -0,0 +1,13 @@
+basic:
```
cmacknz, line 1: What value are we getting out of making this a file and having to deal with opening+parsing it vs just embedded this list into the agent binary?

michalpristas: upgrade scenario. 
if you want to change content of flavor in between upgrades you need somehow know what should be present in a new version. upgrade code is executed from stale version not the new one so embedded content would reflect state of things of version-1. if you add a new component or change layout of existing one in a new version it would be stripped and break

internal/pkg/agent/application/upgrade/upgrade.go
michalpristas: right :-)

michalpristas:
Spec file updated for agentbeat here: https://github.com/elastic/beats/pull/42348

michalpristas:
Fixing windows

internal/pkg/agent/install/.flavors
michalpristas: it should be ok to merge if you're ok with mixing concerns. the only difference is that manifest is in versioned home after installation, but it should not matter in this case.

cmacknz: Keeping it in versioned home is a benefit IMO so that it is a property of an individual agent version. I'd rather not introduce something new for this if we can handle this with an existing mechanism.

CC @pchila for opinions on use of the manifest file for defining the flavors per agent version since he implemented it originally.

michalpristas: updated to use manifest in a meanwhile

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -231,7 +231,12 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	u.log.Infow("Unpacking agent package", "version", newVersion)
 
 	// Nice to have: add check that no archive files end up in the current versioned home
-	unpackRes, err := u.unpack(version, archivePath, paths.Data())
+	// default to extended flavor to avoid breaking behavior
+
+	// no default flavor, keep everything in case flavor is not wspecified
```
andrzej-stencel, line 236:
```suggestion
	// no default flavor, keep everything in case flavor is not specified
```

internal/pkg/agent/install/flavors.go
```diff
@@ -0,0 +1,258 @@
...
+	components, found := flavorsRegistry[detectedFlavor]
+	if !found {
+		return FlavorDefinition{}, ErrUnknownFlavor
+	}
+
+	return FlavorDefinition{detectedFlavor, components}, nil
+}
+
+// SpecsForFlavor returns spec files associated with specific flavor
+func SpecsForFlavor(flavor FlavorDefinition) ([]string, error) {
```
andrzej-stencel, line 77: This function always returns `nil` as error, is the error needed?

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -204,6 +215,61 @@ func getPackageMetadataFromZip(archivePath string) (packageMetadata, error) {
...
+			// unknown flavor fallback to copy all
+			return func(relPath string) bool { return false }, nil
+		}
+		return nil, err
+	}
+	specsInFlavor, _ := install.SpecsForFlavor(flavor) // ignoring error flavor exists, it was loaded before
+
+	// fix versionedHome
+	versionedHome = strings.ReplaceAll(versionedHome, "\\", "/")
+	log.Errorf("using %q", versionedHome)
```
andrzej-stencel, line 236: Is this needed? It's not an error, right?

michalpristas: leftover from previous design, thanks for pointing it out

internal/pkg/agent/install/flavors.go
michalpristas: leftover from previous design, thanks for pointing it out

internal/pkg/agent/install/.flavors
pchila: I am not against using the manifest... In the original RFC where the manifest was introduced, there was already a proposal to have  a list of inclusion/exclusion of file paths during install/upgrade (never implemented as the RFC changed based on feedback) so package manifest was already supposed to be used for this purpose.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -34,15 +35,15 @@ type UnpackResult struct {
 }
 
 // unpack unpacks archive correctly, skips root (symlink, config...) unpacks data/*
-func (u *Upgrader) unpack(version, archivePath, dataDir string) (UnpackResult, error) {
+func (u *Upgrader) unpack(version, archivePath, dataDir string, skipComponents bool, flavor string) (UnpackResult, error) {
```
swiatekm, line 38: What is `skipComponents` for? It doesn't look like it's used in the function body.

Weird that none of our linters catches this.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -204,6 +215,61 @@ func getPackageMetadataFromZip(archivePath string) (packageMetadata, error) {
 	return getPackageMetadataFromZipReader(r, fileNamePrefix)
 }
```
swiatekm, line 217: A docstring would be nice here.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -95,15 +96,21 @@ func unzip(log *logger.Logger, archivePath, dataDir string) (UnpackResult, error
...
 	if metadata.manifest != nil {
 		pm.mappings = metadata.manifest.Package.PathMappings
 		versionedHome = filepath.FromSlash(pm.Map(metadata.manifest.Package.VersionedHome))
+		registry = metadata.manifest.Package.Flavors
 	} else {
 		// if at this point we didn't load the manifest, set the versioned to the backup value
 		versionedHome = createVersionedHomeFromHash(hash)
 	}
 
+	skipFn, err := skipFnFromZip(log, r, flavor, fileNamePrefix, createVersionedHomeFromHash(hash), registry)
```
swiatekm, line 109: Shouldn't we use `versionedHome` instead of `createVersionedHomeFromHash(hash)` here?

cmacknz:
Manually testing this this and it works so far:

- `elastic-agent install` omits all server components
- `elastic-agent install --install-servers` includes server components
- `elastic-agent upgrade 9.0.0-SNAPSHOT` from 8.15.x with no flavor to this branch kept all the components.

Things that still need to be tested:
- Upgrading from a version with a defined flavor to another version that supports flavors and ensuring it respects the flavor, ideally for both basic and server flavors. We can set this up by messing with the package version during build but we might need to merge this first to allow automated tests of it.

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -204,6 +215,61 @@ func getPackageMetadataFromZip(archivePath string) (packageMetadata, error) {
...
+		return nil, err
+	}
+	specsInFlavor := install.SpecsForFlavor(flavor) // ignoring error flavor exists, it was loaded before
+
+	// fix versionedHome
+	versionedHome = strings.ReplaceAll(versionedHome, "\\", "/")
+	log.Errorf("using %q", versionedHome)
+
+	readFile := func(specFilePath string) ([]byte, error) {
+		log.Errorf("reading %q", specFilePath)
```
cmacknz, line 239: The errorf's are probably better as debugfs so we don't log every file in the upgrade every time, if you are planning to keep them.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -231,7 +231,12 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	u.log.Infow("Unpacking agent package", "version", newVersion)
 
 	// Nice to have: add check that no archive files end up in the current versioned home
-	unpackRes, err := u.unpack(version, archivePath, paths.Data())
+	// default to extended flavor to avoid breaking behavior
+
+	// no default flavor, keep everything in case flavor is not specified
+	// in case of error fallback to keep-all
+	detectedFlavor, _ := install.UsedFlavor(paths.Top(), "")
```
cmacknz, line 238: We probably want to log the flavor we detected so we can tell which code paths we went down if something went wrong.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -231,7 +231,12 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	u.log.Infow("Unpacking agent package", "version", newVersion)
 
 	// Nice to have: add check that no archive files end up in the current versioned home
-	unpackRes, err := u.unpack(version, archivePath, paths.Data())
+	// default to extended flavor to avoid breaking behavior
```
cmacknz, line 234:
```suggestion
	// default to no flavor to avoid breaking behavior
```

internal/pkg/agent/cmd/install.go
```diff
@@ -57,6 +58,7 @@ would like the Agent to operate.
 	cmd.Flags().BoolP("non-interactive", "n", false, "Install Elastic Agent in non-interactive mode which will not prompt on missing parameters but fails instead.")
 	cmd.Flags().String(flagInstallBasePath, paths.DefaultBasePath, "The path where the Elastic Agent will be installed. It must be an absolute path.")
 	cmd.Flags().Bool(flagInstallUnprivileged, false, "Install in unprivileged mode, limiting the access of the Elastic Agent. (beta)")
+	cmd.Flags().Bool(flagInstallServers, false, "Install extended version of agent that includes server components")
```
cmacknz, line 61:
```suggestion
	cmd.Flags().Bool(flagInstallServers, false, "Install larger version of agent that includes server components")
```

internal/pkg/agent/install/flavors.go
```diff
@@ -0,0 +1,259 @@
...
+			err = removeErr
+		}
+	}
+
+	return err
+}
+
+// ParseComponentFiles parses spec files and returns list of associated paths with component.
+// Default set consisting of binary, spec file and default config file is always present
+func ParseComponentFiles(content []byte, filename string, includeDefaults bool) ([]string, error) {
```
cmacknz, line 128: nit: we already have a file encapsulating spec files, this should probably live there and not with the install logic https://github.com/elastic/elastic-agent/blob/af2a1c2cf9ccfb06c95ca51ec91fb5d83413e1e8/pkg/component/load.go#L241

internal/pkg/agent/install/flavors.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func skipComponentsPath(relPath string, allowedSubpaths []string) bool {
+	if allowedSubpaths == nil {
+		return false
+	}
+	if runtime.GOOS == "windows" {
+		relPath = strings.ReplaceAll(relPath, "\\", "/")
+	}
+	componentsDir := "/components/" // fmt.Sprintf("%ccomponents%c",  .PathSeparator, os.PathSeparator)
```
cmacknz, line 185:
```
	componentsDir := "/components/" // fmt.Sprintf("%ccomponents%c",  .PathSeparator, os.PathSeparator)
```

Is the commented out code a TODO?

internal/pkg/agent/install/install.go
```diff
@@ -75,12 +75,29 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
...
+	}
+
+	err = copyFiles(copyConcurrency, pathMappings, dir, topPath, skipFn)
 	if err != nil {
 		pt.Describe("Error copying files")
 		return utils.FileOwner{}, err
 	}
 
+	if err := markFlavor(topPath, flavor); err != nil {
+		return utils.FileOwner{}, err
```
cmacknz, line 98: nit: wrap the error with some context, same goes for the other new locations we can return errors from the install command.

cmacknz (commented):
No big comments besides nits.

I wonder if you can use the "upgrade to same commit" test to add a basic test of upgrading from an agent installed with a flavor to another one installed with the same flavor to make sure the unpacking logic works properly. https://github.com/michalpristas/elastic-agent/blob/af2a1c2cf9ccfb06c95ca51ec91fb5d83413e1e8/testing/integration/upgrade_standalone_same_commit_test.go#L83

michalpristas:
@cmacknz i did not want to add same commit upgrade as it may not uncover some issues e.g if there are leftovers from servers install and only basic is applied. but test would evaluate properly. i will add tests once we have snapshot.
but i was testing this manually by adding some artificial commit (e.g logline) and using update with custom source uri

internal/pkg/agent/application/upgrade/step_unpack.go
michalpristas: i thought i removed it, thanks

internal/pkg/agent/application/upgrade/step_unpack.go
michalpristas: versionedHome at this point can have form of `version-snapshot-hash` as defined in a mapping here https://github.com/elastic/elastic-agent/blob/2985c322f5d383dc1b3b89e4d1c227225873d850/dev-tools/mage/settings.go#L361

this is not what is physical path in a zip file. this is resulting after unpacking

internal/pkg/agent/application/upgrade/step_unpack.go
michalpristas: no these are just debug logs, i use error because it's easier to grep, removed

internal/pkg/agent/install/flavors.go
michalpristas: previous version.

pkg/component/load.go
```diff
@@ -9,8 +9,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 
 	"github.com/elastic/go-ucfg/yaml"
+
+	yamlv3 "gopkg.in/yaml.v3"
```
cmacknz, line 17: nit: we are using both `github.com/elastic/go-ucfg/yaml` and `gopkg.in/yaml.v3` at the same time

cmacknz (approved):
LGTM, it works and the only comments are nits. Make sure this is followed up with an integration test for an upgrade from a basic->basic flavor and servers->servers flavor.

Also address any outstanding comments from others before merge if you haven't already.

pkg/component/load.go
michalpristas: i wanted to reduce ucfg step. and keeping ucfg one for NewConfigFrom use cases i don't want to address in this PR if that makes sense.
otherwise insteadn of unmarshal it would be
`unmarshal -> ucfg.NewFrom -> Unpack`

michalpristas:
https://github.com/elastic/beats/pull/42348
merged, i'll merge this once beats snapshot is ready

cmacknz:
We also need to test that the spec file changes to include additional files for every integration work correctly. We have an integration test that installs defend but nothing for osquery or apm server (those are the ones I remember off the top of my head).

cmacknz:
Follow up testing work captured into https://github.com/elastic/elastic-agent/issues/6631 so we don't forget.

v1v:
@michalpristas ,  I've no have enough context, but I cannot see any `8.x` backports, is that intended? I've seen https://github.com/elastic/elastic-agent/pull/6870 depends on this PR, hence it's gonna be backported to `9.0` but, I don't know if the improvements in the CI should also tackle other active branches such as `8.18`, `8.17` or `8.x`. Can you confirm?

PR #6565, opened 2025-01-22
FIPS Build

michel-laterman:
## What does this PR do?

Adds `FIPS` env var to magefile to enable FIPS compliant builds using the microsoft/go toolchain.

This PR will not be sufficient to ensure that every artifact made with these changes are compliant, we still need to verify our crypto use.

## Why is it important?

FIPS artifacts must be built with compliant toolchains.

## Checklist

## Disruptive User Impact

None

Assuming microsoft go is available, run `FIPS=true mage build:binary`


dev-tools/mage/settings.go
```diff
@@ -35,6 +35,8 @@ const (
 	beatsFPMImage = "docker.elastic.co/beats-dev/fpm"
 	// BeatsCrossBuildImage is the image used for crossbuilding Beats.
 	BeatsCrossBuildImage = "docker.elastic.co/beats-dev/golang-crossbuild"
+	//FIPSBuildImage is the image used for building FIPS compliant artifacts
+	FIPSBuildImage = "mcr.microsoft.com/oss/go/microsoft/golang"
```
michel-laterman, line 39: Currently using this image directly fails; i think it's because `mage` is not installed.
The error when running `FIPS=true PACKAGES="tar.gz" PLATFORMS=linux/arm64 mage package` is:
```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "--build-cmd": executable file not found in $PATH: unknown.
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "--build-cmd": executable file not found in $PATH: unknown.
package ran for 14.597428834s
Error: multiple failures: running "docker run --env EXEC_UID=501 --env EXEC_GID=20 -v /Users/mlaterman/go/pkg/mod:/go/pkg/mod:ro --rm --env GOFLAGS=-mod=readonly --env MAGEFILE_VERBOSE= --env MAGEFILE_TIMEOUT= --env SNAPSHOT=false --env DEV=false --env EXTERNAL=false --env FIPS=true -v /Users/mlaterman/git/elastic-agent:/go/src/github.com/elastic/elastic-agent -w /go/src/github.com/elastic/elastic-agent mcr.microsoft.com/oss/go/microsoft/golang:1.22.10-1-fips-bookworm --build-cmd build/mage-linux-arm64 buildGoDaemon --platforms linux/arm64" failed with exit code 127
multiple failures: running "docker run --env EXEC_UID=501 --env EXEC_GID=20 -v /Users/mlaterman/go/pkg/mod:/go/pkg/mod:ro --rm --env GOFLAGS=-mod=readonly --env MAGEFILE_VERBOSE= --env MAGEFILE_TIMEOUT= --env SNAPSHOT=false --env DEV=false --env EXTERNAL=false --env FIPS=true -v /Users/mlaterman/git/elastic-agent:/go/src/github.com/elastic/elastic-agent -w /go/src/github.com/elastic/elastic-agent mcr.microsoft.com/oss/go/microsoft/golang:1.22.10-1-fips-bookworm --build-cmd build/mage-linux-arm64 golangCrossBuild --platforms linux/arm64" failed with exit code 127
```

cmacknz:
This should come with a test that the binary we produce actually works, ideally one that can confirm it properly uses the FIPS OpenSSL in the expected way. 

You'll also need to update CI to actually build the FIPS variant, the package steps for testing are in https://github.com/elastic/elastic-agent/blob/a29af648901e3025eee6bccb45b321c83775d227/.buildkite/integration.pipeline.yml#L8 right now

simitt (commented):
Similar to what @cmacknz already suggested, there should be some tests. IMO there would ideally also be some unit tests for the mage functionality, which shouldn't be too hard given it is regular go code.

dev-tools/mage/build.go
```diff
@@ -34,6 +35,39 @@ type BuildArgs struct {
...
 
+// buildTagRE is a regexp to match strings like "-tags=abcd"
+// but does not match "-tags= "
+var buildTagRE = regexp.MustCompile(`-tags=([\S]+)`)
+
+// ParseExtraFlags returns the ExtraFlags param where all flags that are go build tags are joined by a comma.
+//
+// For example if given -someflag=val1 -tags=buildtag1 -tags=buildtag2
+// It will return -someflag=val1 -tags=buildtag1,buildtag2
+func (b BuildArgs) ParseExtraFlags() []string {
```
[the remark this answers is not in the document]
michel-laterman: This is a safety check to avoid passing two `-tags` flags; this occures in testing in the beats repo (when producing agentbeat).
Having this defined and used here will prevent others in the future from encountering the same issue

.buildkite/bk.integration.pipeline.yml
```diff
@@ -157,6 +157,21 @@ steps:
...
+      - label: "x86_64:fips"
+        command: |
+          buildkite-agent artifact download build/distributions/** . --step 'packaging-ubuntu-x86-64-fips' --build ${BUILDKITE_TRIGGERED_FROM_BUILD_ID}
+          .buildkite/scripts/steps/integration_tests_tf.sh fips false
+        artifact_paths:
+          - build/**
+          - build/diagnostics/**
+        agents:
+          provider: "gcp"
+          machineType: "n1-standard-8"
```
dliappis, line 172: do we actually need a VM with 30Gi of ram? If yes, I'd still recommend to switch to `n2-standard-8`, n1 is really an old relic that can end up using very old (and slow) hardware.

.buildkite/integration.pipeline.yml
```diff
@@ -20,6 +20,19 @@ steps:
...
+        env:
+          PACKAGES: "tar.gz"
+          PLATFORMS: "linux/amd64"
+          FIPS: "true"
+        command: ".buildkite/scripts/steps/integration-package.sh"
+        artifact_paths:
+        - build/distributions/**
+        agents:
+          provider: "gcp"
+          machineType: "n2-standard-8"
```
dliappis, line 34: same question regarding the size, do we need 30Gi or is a machine with ~16Gi line `n2-standard-4` enough?

.buildkite/integration.pipeline.yml
```diff
@@ -32,6 +45,19 @@ steps:
...
+        env:
+          PACKAGES: "tar.gz"
+          PLATFORMS: "linux/arm64"
+          FIPS: "true"
+        command: ".buildkite/scripts/steps/integration-package.sh"
+        artifact_paths:
+          - build/distributions/**
+        agents:
+          provider: "gcp"
+          machineType: "n2-standard-8"
```
dliappis, line 59: same question as earlier

dev-tools/mage/build.go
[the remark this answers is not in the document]
simitt: Is there a specific reason why the parsing should only be applied to build tags? If not, how about implementing this a bit more generic to apply the logic to all flags, or rename the function to something like `ParseBuildTags` or something similar?

E.g. (although this slightly changes the logic to allow flags with a space)
```
flags := make([]string, 0)
flagsMap := make(map[string][]string)
for _, flag := range b.ExtraFlags {
	arr := strings.Split(flag, "=")
	if len(arr) != 2 || arr[0] == "" || arr[1] == "" {	
		log.Printf("Unexpected format found for buildargs.ExtraFlags, ignoring value  %q", flag)
		continue
	}
	flagsMap[arr[0]] = append(flagsMap[arr[0]], arr[1])
}
for k,v := range flagsMap {
	flags = append(flags, k + "=" + strings.Join(v, ","))
}
return flags
```

magefile.go
```diff
@@ -802,6 +806,14 @@ func (Cloud) Image(ctx context.Context) {
 		devtools.Snapshot = true
 	}
 
+	if f, err := strconv.ParseBool(fips); err == nil && !f {
+		os.Setenv(fipsEnv, "false")
+		devtools.FIPSBuild = false
+	} else {
+		os.Setenv(fipsEnv, "true")
+		devtools.FIPSBuild = true
+	}
```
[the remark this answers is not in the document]
simitt: This reads a bit complicated - how about something like:
```
	f, err := strconv.ParseBool(fips) 
        if err != nil {
		f = false
	}
	os.Setenv(fipsEnv, strconv.FormatBool(f))
	devtools.FIPSBuild = f
```

testing/integration/fips_verification_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	checkLinks := false
+	for _, setting := range info.Settings {
+		switch setting.Key {
+		case "-tags":
+			require.Contains(t, setting.Value, "requirefips")
+			continue
+		case "GOEXPERIMENT":
+			require.Contains(t, setting.Value, "systemcrypto")
+			continue
```
simitt, line 48: For a FIPS specific test, this should test that `-tags=requirefips` and `GOEXPERIMENT=systemcrypto`. Currently it only checks for an expected value _if_ the key is there, but not that the key exists.

magefile.go
```diff
@@ -788,6 +789,9 @@ func (Cloud) Image(ctx context.Context) {
 	variant := os.Getenv(dockerVariants)
 	defer os.Setenv(dockerVariants, variant)
 
+	fips := os.Getenv(fipsEnv)
+	defer os.Setenv(fipsEnv, fips)
```
simitt, line 793: Why is the env set back to the original - potentially invalid - value with `defer` when also you specifically set it to `true` or `false` in line (809:818)?

internal/pkg/release/version.go
```diff
@@ -20,6 +20,9 @@ const (
 // snapshot is a flag marking build as a snapshot.
 var snapshot = ""
 
+// fips is a flag for marking a FIPS compliant build.
+var fips = "false"
```
simitt, line 24: Where is this ever set to a different value? This whole version logic seems overly complex.

simitt (commented):
Left some more comments, mostly trying to follow the complexity in the mage/release logic or seeing whether it can be reduced a bit to reduce the surface for bugs.

internal/pkg/release/version.go
michel-laterman: Yes, a FIPS build will replace the symbol with "true"

.buildkite/bk.integration.pipeline.yml
michel-laterman: I just copied from the template above

dev-tools/mage/build.go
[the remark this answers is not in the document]
michel-laterman: We know that build tags are the issue, i'll rename the func to reflect this

magefile.go
michel-laterman: I'm being consistent with how Image handles the `SNAPSHOT` env var

simitt (approved):

simitt:
Let's wait for a review from @pchila as this is related to packaging.

dliappis (commented):
Thanks for iterating.
There seem to be a few failing tests in CI e.g. https://buildkite.com/elastic/elastic-agent-extended-testing/builds/7357#019520ad-6465-47f3-8b18-7fca000d61ba could you please take a look?

michel-laterman:
The current failure error is:
```
.../elastic-agent-9.0.0-SNAPSHOT-linux-x86_64-fips/manifest.yaml: no such file or directory
```
When I download the artifact and extract it, i can see the folder it creates is missing the `-fips` suffix; i'll try to get this working but I'm having issues stemming from running package on an arm macbook

pchila (approved):
Good enough to get a first FIPS compliant agent package in order to start testing.
There are some parts that will need modification to integrate with elastic-agent packaging code but that will be done in a follow-up PR that will produce FIPS artifacts along with the current ones.

Left a couple of comments as a reminder of what needs modifying

michel-laterman:
buildkite test this

michel-laterman:
buildkite test this

dev-tools/mage/build.go
```diff
@@ -151,6 +190,12 @@ func Build(params BuildArgs) error {
 	if params.CGO {
 		cgoEnabled = "1"
 	}
+
+	if FIPSBuild {
+		cgoEnabled = "1"
+		env["GOEXPERIMENT"] = "systemcrypto"
```
swiatekm, line 196: I think this would be more visible if we set it using a build tag instead. According to https://pkg.go.dev/internal/goexperiment, the tag would be `goexperiment.systemcrypto`.

swiatekm (approved):

dev-tools/mage/build.go
simitt: The alternative of using build tags is also mentioned in https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md#build-tags. Functionality wise either approach should be valid.

michel-laterman: I tried changing this to add the build tag, but it changed the behaviour of the build command.

Currently if you run `FIPS=true mage build` without microsoft/go, the build fails.
Changing from an env var to another build tag allows it to succeed.
I think having a failure here is desirable as it functions as another check to ensure the binary is built correctly

.buildkite/integration.pipeline.yml
```diff
@@ -20,6 +20,19 @@ steps:
           provider: "gcp"
           machineType: "n2-standard-8"
 
+      - label: "Packaging: Ubuntu x86_64 FIPS"
```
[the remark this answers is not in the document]
michel-laterman: I need to mark this as resolved in order to merge

dev-tools/mage/build.go
```diff
@@ -34,6 +35,39 @@ type BuildArgs struct {
...
+	buildTags := make([]string, 0)
+	for _, flag := range b.ExtraFlags {
+		if buildTagRE.MatchString(flag) {
+			arr := buildTagRE.FindStringSubmatch(flag)
+			if len(arr) != 2 || arr[1] == "" {
+				log.Printf("Unexpected format found for buildargs.ExtraFlags, ignoring value  %q", flag)
+				continue
+			}
+			buildTags = append(buildTags, arr[1])
+		} else {
```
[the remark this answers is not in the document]
michel-laterman: I need to mark this as resolved in order to merge

michel-laterman:
@dliappis, @v1v can I please get a review?

dliappis (approved):
LGTM

Issue #3586, opened 2023-10-11
Elastic Agent doesn't update the enrollment token in Kubernetes Deployment statefile


ch9hn:
- Duplicates https://github.com/elastic/elastic-agent/issues/5187

When a new enrollment token is updated as env or envFrom in Kubernetes Manifest, this new token is not reflected in Elastic Agent.
Reason for that is probably the fact, that Elastic Agent saves the state locally on every Kubernetes Node and doesn't update the new token.
This leads to Unauthorised issues on the Agent - a redeploy with a new token is not possible anymore.

For confirmed bugs, please report:
- Version: 8.10
- Operating System: Ubuntu Linux / Kubernetes 1.27
- Discuss Forum URL: 

- Steps to Reproduce:

1. Install Elastic Agent on Kubernetes Cluster as described in the docs with enrollment-token "ABC"
2. Expire "ABC" and add new token "DEF" 
3. Restart Elastic Agent Daemonset
4. Result: Old token "ABC" is persisted and used for the communication to Elastic Fleet Server

Error logs:

`"Failed to connect to backoff(elasticsearch(https://xxxx.xxxx.cloud.es.io:443)): 401 Unauthorized: {\"error\":{\"root_cause\":[{\"type\":\"security_exception\",\"reason\":\"unable to authenticate with provided credentials and anonymous access is not allowed for this request\",\"additional_unsuccessful_credentials\":\"API key: api key [xxxxxxx] has been invalidated\",\"header\":{\"WWW-Authenticate\":[\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\",\"Bearer`

How to temporary fix:
When using Kustomize deployment, the hostPath can be overwritten quite easily with the following DaemonSet overwrite:

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: elastic-agent
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app: elastic-agent
  template:
    metadata:
      labels:
        app: elastic-agent
    spec:
      containers:
        - name: elastic-agent
          env:
            - name: FLEET_URL
              $patch: delete
            - name: FLEET_ENROLLMENT_TOKEN
              $patch: delete
            - name: FLEET_INSECURE
              value: "false"
            - name: KIBANA_HOST
              $patch: delete
            - name: KIBANA_FLEET_USERNAME
              $patch: delete
            - name: KIBANA_FLEET_PASSWORD
              $patch: delete
          envFrom:
[...]

PR #6568, opened 2025-01-22
Fix Fleet Enrollment Handling for Containerized Agent

pkoutsovasilis:
## What does this PR do?

This PR introduces a fix for containerized Fleet-managed Elastic Agents to handle scenarios where:
- The Fleet url changes.
- The Fleet enrollment token changes.
- The agent is unenrolled from Kibana Fleet.

It achieves the above by enhancing the agent's logic inside the `container` cmd to:
- Verify enrollment conditions against stored state and only re-enroll when necessary.
- Validate API token validity with Fleet, ensuring correct authentication even when the configuration hasn't changed.

The PR also adds a Kubernetes integration tests to verify the proper behaviour of enrollment and re-enrollment under various scenarios, including:
- Re-deployment of agent with an updated fleet enrollment token.
- Re-deployment of agent that is unenrolled from Kibana Fleet with the same enrollment token.
- Re-deployment of agent across an older version and this one that ensures that no re-enrollment happens.

Key changes include:
- Introduction of `shouldFleetEnroll` to centralize logic for fleet enrollment decisions.
- Creation of PBKDF2-based enrollment token hashing to save in the agent state and be able to track token changes securely.
- Utilise the `agent/{id}/acks` path of Fleet server with **empty events** to check if the agent API token is still valid. More than happy to introduce a separate path just for this cause on Fleet server, although it will be the same the ACKs one with empty events (as of now at least) 

You can easily see [here](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/6193#01948d9d-25f1-4d51-85ed-6494e8f0da2d/57-489) in the CI run of the first commit in this PR that the Elastic Agent, before this PR, doesn't handle enrollment correctly and always resorts to using the Fleet token and URL stored in its state. This PR addresses that issue and ensures enrollment uses the correct configuration and token.

PS: the actual changes of this PR are this commit https://github.com/elastic/elastic-agent/commit/b6596d015138349c4edde154c4528207a64c3404 which is `+305 -25` thus I consider this PR aligned with the team policies 🙂 

## Why is it important?

This fix ensures robust and predictable behavior for containerized Fleet-managed Elastic Agents and enhance user experience of managing elastic-agent in Kubernetes.

## Checklist

- [x]  I have made corresponding changes to the documentation

## Disruptive User Impact

The changes introduced in this PR are non-disruptive. They improve fleet enrollment handling and maintain backward compatibility. 

```
mage integration:auth
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/3586


internal/pkg/agent/cmd/container.go
```diff
@@ -988,3 +991,129 @@ func isContainer(detail component.PlatformDetail) component.PlatformDetail {
...
+
+	storedConfig, err := configuration.NewFromConfig(cfg)
+	if err != nil {
+		return false, fmt.Errorf("failed to read from disk store: %w", err)
+	}
+
+	storedFleetHosts := storedConfig.Fleet.Client.GetHosts()
+	if len(storedFleetHosts) == 0 || !slices.Contains(storedFleetHosts, setupCfg.Fleet.URL) {
+		// fleet url changed thus elastic-agent should enroll
+		return true, nil
```
blakerouse, line 1048: What if they are just updating the endpoint at which Fleet Server is running but its the same Fleet Server? I would expect just checking the token hash is enough, because if it is a different Fleet Server then token would 100% be different.

internal/pkg/agent/cmd/container.go
```diff
@@ -988,3 +991,129 @@ func isContainer(detail component.PlatformDetail) component.PlatformDetail {
...
+			// the enrollment token hash does not match thus elastic-agent should enroll
+			return true, nil
+		case err != nil:
+			return false, fmt.Errorf("failed to compare hash: %w", err)
+		}
+	}
+
+	// Although fleet url and enrollment token are the same, we must also validate the api token.
+	// The agent might have been unenrolled from kibana which doesn't require for fleet url and
+	// enrollment token need to be changed
```
blakerouse, line 1072: Who unenrolled it? Like it was a manual action in the UI to unenroll it? Should it then auto re-enroll?

internal/pkg/agent/cmd/container.go
```diff
@@ -988,3 +991,129 @@ func isContainer(detail component.PlatformDetail) component.PlatformDetail {
...
+		Events: nil,
+	}
+	ackCMD := fleetapi.NewAckCmd(&agentInfo{storedConfig.Fleet.Info.ID}, fc)
+	_, err = ackCMD.Execute(ctx, ackRequest)
+	switch {
+	case errors.Is(err, fleetclient.ErrInvalidAPIKey):
+		// this elastic-agent has not valid api token thus it should enroll
+		return true, nil
+	case err != nil:
+		return false, fmt.Errorf("failed to validate api token: %w", err)
```
blakerouse, line 1094: A simple network error could cause this, should there be some retries in here?

internal/pkg/agent/cmd/container.go
```diff
@@ -988,3 +991,129 @@ func isContainer(detail component.PlatformDetail) component.PlatformDetail {
...
+	}
+
+	if len(storedConfig.Fleet.EnrollmentTokenHash) == 0 {
+		// previous EnrollmentTokenHash was empty, but we just validated
+		// that this elastic-agent is still eligible to authenticate with fleet.
+		// Update the config with the validated EnrollmentTokenHash.
+		enrollmentHashBytes, err := crypto.GeneratePBKDF2FromPassword([]byte(setupCfg.Fleet.EnrollmentToken))
```
blakerouse, line 1101: What happens if they have updated the container to remove the EnrollmentToken but are using the same storage of an already enrolled container? Maybe this should check that an EnrollmentToken is set in the `setupCfg` before it just hashes it?

internal/pkg/agent/cmd/container.go
pkoutsovasilis: ok I hear you, so I go I unenroll agent from Kibana and then I restart the pods, what would you expect the agent to do? Personally I think that since I give it a valid fleet url and a token and this is unenrolled, it should re-enroll. Also this user [here](https://github.com/elastic/elastic-agent/issues/3586#issuecomment-2602306451) 🙂

blakerouse: Yeah that does make sense. Could you maybe add more of that context to the comment?

internal/pkg/agent/cmd/container.go
pkoutsovasilis: if the enrollmentToken is empty and the one we have in our state isn't wouldn't that cause the two hashes to be different and then the agent to re-enroll? What am I missing 🤔

blakerouse: I don't think your missing anything, just wanted to be sure that it would work.

internal/pkg/agent/cmd/container.go
pkoutsovasilis: are you talking of changing solely the DNS records?! because if they actually change the endpoint IP then using the fleet url from the state at run the agent won't be able to connect to it right?!

blakerouse: Not if they setup a redirect and then also updated the policy with a new URL. Then it will start connect be redirected, connect get new policy revision and then update the local state. Then they update the container manifest and then remove the redirect.

pkoutsovasilis: If I follow the flow correctly this is something that will happen through the Kibana-Fleet UI where the agent will persist the changes to it's state without restarting. So the debate here is what happens when the user restarts the pod as the manifest will still point to the old url?

blakerouse: The flow is both happen. They update the URL in Kibana, then they update the manifest for the Elastic Agent. They would not want it to re-enroll its already enrolled.

pkoutsovasilis: > They update the URL in Kibana, then they update the manifest for the Elastic Agent.

if they add the same URL from Kibana and then in the manifest, then they won't re-enroll. The re-enrollment will happen only if the URL defined in the manifest doesn't exist in the hosts saved in the state, right?

blakerouse: That is correct. I was mis-reading this. I was thinking that it was caching the enrollment URL, but actually this is coming from the policy hosts. This makes more sense, sorry for the noise.

internal/pkg/agent/cmd/container.go
pkoutsovasilis: I update the comment, feel free to unresolve if you believe that more context is needed 🙂

internal/pkg/agent/cmd/container.go
pkoutsovasilis: thanks for the comment, actually you made me think of another case where the fleet enrollment token from the setup can be empty and the agent should fallback to kibana to extract a new one. I updated the code to capture that case as well. Feel free to unresolve if you spot anything else 🙂

internal/pkg/agent/cmd/container.go
pkoutsovasilis: I updated the comment on this part in an attempt to make that more visible. Feel free to unresolve if you spot anything else 🙂

internal/pkg/agent/cmd/container.go
pkoutsovasilis: I added a really simple retry logic, let me know if that's align with what you had on mind 🙂

pchila (approved):
LGTM
Just a couple of nitpicks on the usage of mocks in tests but that's not a blocker for merging

internal/pkg/agent/cmd/container_test.go
```diff
@@ -241,3 +257,272 @@ func TestKibanaFetchToken(t *testing.T) {
...
+		cfg                  setupConfig
+		statFn               func(path string) (os.FileInfo, error)
+		encryptedDiskStoreFn func(savedConfig *configuration.Configuration) storage.Storage
+		fleetClientFn        func() client.Sender
+		expectedSavedConfig  func(t *testing.T, savedConfig *configuration.Configuration)
+		expectedShouldEnroll bool
+		expectedErr          error
+	}{
+		"should not enroll if fleet enroll is disabled": {
+			cfg: setupConfig{Fleet: fleetConfig{Enroll: false}},
```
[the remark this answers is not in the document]
pkoutsovasilis: nice point I do also prefer explicitness, I just pushed a commit for it. Please feel free to unresolve if you spot anything else 🙂

internal/pkg/agent/cmd/container_test.go
```diff
@@ -241,3 +257,272 @@ func TestKibanaFetchToken(t *testing.T) {
...
+    - host1
+    - host2
+    - host3
+  agent:
+  protocol: "https"`)), nil)
+				return m
+			},
+			fleetClientFn: func() client.Sender {
+				tries := 0
+				m := &mockFleetClient.Sender{}
```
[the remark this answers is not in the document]
pkoutsovasilis: I have no idea why I didn't go with it from the start, but I just pushed a commit that does it as you propose, ty. Please feel free to unresolve if you spot anything else 🙂

internal/pkg/agent/cmd/container.go
pkoutsovasilis: @blakerouse any objections merging this PR?

blakerouse: I am happy with the added retry loop. Would have been a little cleaner to place it in its own function, but that is okay. Feel free to merge.

blakerouse (approved):

internal/pkg/agent/cmd/container.go
```diff
@@ -989,3 +994,152 @@ func isContainer(detail component.PlatformDetail) component.PlatformDetail {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to create logger: %w", err)
+	}
+	fc, err := newFleetClient(log, storedConfig.Fleet.AccessAPIKey, storedConfig.Fleet.Client)
+	if err != nil {
+		return false, fmt.Errorf("failed to create fleet client: %w", err)
+	}
+
+	// Perform an ACK request with **empty events** to verify the validity of the API token.
+	// If the agent has been manually un-enrolled through the Kibana UI, the ACK request will fail due to an invalid API token.
```
cmacknz, line 1091: File an issue in https://github.com/elastic/ingest-docs to document this and other new behavior.

This makes so that if you want to unenroll an containerized agent, you can't do it from the UI in the normal way. You either need to revoke the enrollment token and/or remove the Fleet URL and Token from the environment variables.

Someone force uenrolling an agent in the UI now has the behavior that the next container restart will cause the same agent to re-enroll itself with a different ID.

pkoutsovasilis: roger that! just to be sure, the behaviour of this PR is what we are after, right? In other words, should I revert this one and change it?

cmacknz: The behavior is different for containers now, but I think in a way that is OK. We just need to document it. The user left the container with a valid enrollment token and essentially told it to enroll at startup if not enrolled, we just do a better job of detecting that.

PR #6583, opened 2025-01-23
Switch k8s input paths to /var/log/pods/* to ingest rotated container logs

### LINKED ISSUE #5164: [k8s provider] Look up metadata based on file path
**Describe the enhancement:**

Have the [Kubernetes provider](https://www.elastic.co/guide/en/fleet/current/kubernetes-provider.html) look up metadata based on file path, similar to what the `add_kubernetes_metadata` provider in Beats is [capable](https://github.com/elastic/beats/pull/7072) of doing.

From @henrikno:

> We have this bug where we lose logs from short lived containers (e.g. if they're crashing). It seems to be like it happens when there's some delay in autodiscover, i.e. by the time it has discovered the container via k8s, and updated the policy to filebeat, the file has been rotated to the new "attempt", which has a new container id.
> So I was looking if it's possible to "reverse" it, and look for the files, and look up metadata based on the file path. And that looks like what `add_kubernetes_metadata` is doing.

pchila:
## What does this PR do?

This PR changes the input definition for ingesting k8s container logs from `/var/log/containers/*` to `/var/log/pods/*` which contains all available logs (including rotated ones) for pods running on the k8s node. 

Before this change elastic-agent would configure the input using the currently running container id and locating the corresponding log file under `/var/log/containers/*` which would point to the current log file, ignoring previously rotated log files for the container.

This change only impacts standalone agents. To have the same behaviour for managed agents a separate PR must be opened for the kubernetes integration in https://github.com/elastic/integrations/tree/main/packages/kubernetes

## Why is it important?
This allow ingesting logs also for containers that rotate logs quickly (for example containers that are crashing very early in their execution) instad of trying to ingest only the last log file for the container.

## Checklist

## Related issues

- Closes #5164 

## Questions to ask yourself

- ...

### DIFF
--- a/deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
+++ b/deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
@@ -16,12 +16,12 @@ Config input for container logs
     namespace: {{ .Values.kubernetes.namespace }}
   use_output: {{ .Values.kubernetes.output }}
   streams:
-  - id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+  - id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
     data_stream:
       dataset: kubernetes.container_logs
       type: logs
     paths:
-      - '/var/log/containers/*${kubernetes.container.id}.log'
+      - '/var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log'
     prospector.scanner.symlinks: {{ dig "vars" "symlinks" true .Values.kubernetes.containers.logs }}
     parsers:
       - container:
--- a/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
+++ b/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
@@ -354,7 +354,7 @@ data:
             condition: '${host.platform} == ''windows'''
             ignore_older: 72h
       # Input ID allowing Elastic Agent to track the state of this input. Must be unique.
-      - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.id}
+      - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
         type: filestream
         use_output: default
         meta:
@@ -366,7 +366,7 @@ data:
         streams:
           # Stream ID for this data stream allowing Filebeat to track the state of the ingested files. Must be unique.
           # Each filestream data stream creates a separate instance of the Filebeat filestream input.
-          - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.id}
+          - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
             data_stream:
               dataset: kubernetes.container_logs
               type: logs
--- a/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
+++ b/deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
@@ -354,7 +354,7 @@ data:
             condition: '${host.platform} == ''windows'''
             ignore_older: 72h
       # Input ID allowing Elastic Agent to track the state of this input. Must be unique.
-      - id: container-log-${kubernetes.pod.name}-${kubernetes.container.id}
+      - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.id}
         type: filestream
         use_output: default
         meta:
@@ -366,7 +366,7 @@ data:
         streams:
           # Stream ID for this data stream allowing Filebeat to track the state of the ingested files. Must be unique.
           # Each filestream data stream creates a separate instance of the Filebeat filestream input.
-          - id: container-log-${kubernetes.pod.name}-${kubernetes.container.id}
+          - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.id}
             data_stream:
               dataset: kubernetes.container_logs
               type: logs
@@ -381,7 +381,7 @@ data:
               #     negate: true
               #     match: after
             paths:
-              - /var/log/containers/*${kubernetes.container.id}.log
+              - /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
       - id: audit-log
         type: filestream
         use_output: default
--- a/deploy/helm/elastic-agent/examples/eck/rendered/manifest.yaml
+++ b/deploy/helm/elastic-agent/examples/eck/rendered/manifest.yaml
@@ -321,13 +321,13 @@ stringData:
         - data_stream:
             dataset: kubernetes.container_logs
             type: logs
-          id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+          id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
           parsers:
           - container:
               format: auto
               stream: all
           paths:
-          - /var/log/containers/*${kubernetes.container.id}.log
+          - /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
           processors:
           - add_fields:
               fields:
--- a/deploy/helm/elastic-agent/examples/kubernetes-default/rendered/manifest.yaml
+++ b/deploy/helm/elastic-agent/examples/kubernetes-default/rendered/manifest.yaml
@@ -320,13 +320,13 @@ stringData:
         - data_stream:
             dataset: kubernetes.container_logs
             type: logs
-          id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+          id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
           parsers:
           - container:
               format: auto
               stream: all
           paths:
-          - /var/log/containers/*${kubernetes.container.id}.log
+          - /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
           processors:
           - add_fields:
               fields:
@@ -888,7 +888,7 @@ spec:
       labels:
         name: agent-pernode-example
       annotations:
-        checksum/config: 233affcd72143e637a130b5f099c30e194d90042eb00a26512f51c844c65a821
+        checksum/config: 0a02c8b3783f5e072f6eb62d87c3fefb2df9cdaf8efb7dbda9bcc11050246c3d
     spec:
       automountServiceAccountToken: true
       containers:
--- a/deploy/helm/elastic-agent/examples/kubernetes-hints-autodiscover/rendered/manifest.yaml
+++ b/deploy/helm/elastic-agent/examples/kubernetes-hints-autodiscover/rendered/manifest.yaml
@@ -320,13 +320,13 @@ stringData:
         - data_stream:
             dataset: kubernetes.container_logs
             type: logs
-          id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+          id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
           parsers:
           - container:
               format: auto
               stream: all
           paths:
-          - /var/log/containers/*${kubernetes.container.id}.log
+          - /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
           processors:
           - add_fields:
               fields:
@@ -891,7 +891,7 @@ spec:
       labels:
         name: agent-pernode-example
       annotations:
-        checksum/config: daca0d998edb3afa587d96e69b0833f6919ca6ba72f58f3a1f83b22d7e5ffaf6
+        checksum/config: 440922798e9423e250298ce729018751f6a9d3343a03f3e54f93380e932acdfe
     spec:
       automountServiceAccountToken: true
       containers:
--- a/deploy/helm/elastic-agent/examples/kubernetes-ksm-sharding/rendered/manifest.yaml
+++ b/deploy/helm/elastic-agent/examples/kubernetes-ksm-sharding/rendered/manifest.yaml
@@ -132,13 +132,13 @@ stringData:
         - data_stream:
             dataset: kubernetes.container_logs
             type: logs
-          id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+          id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
           parsers:
           - container:
               format: auto
               stream: all
           paths:
-          - /var/log/containers/*${kubernetes.container.id}.log
+          - /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log
           processors:
           - add_fields:
               fields:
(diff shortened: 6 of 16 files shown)


cmacknz:
Other references to /var/log/containers to possibly correct:

- https://github.com/search?q=repo%3Aelastic%2Fingest-docs%20%2Fvar%2Flog%2Fcontainers&type=code
- https://github.com/search?q=repo%3Aelastic%2Fbeats%20%2Fvar%2Flog%2Fcontainers&type=code

blakerouse:
@pchila @pkoutsovasilis Do we need to worry about this change causing the files to be re-indexed? How filebeat stores files in the registry is not 100% clear in my mind, I would hope that it would not do that, but maybe?

cmacknz:
> @pchila @pkoutsovasilis Do we need to worry about this change causing the files to be re-indexed? How filebeat stores files in the registry is not 100% clear in my mind, I would hope that it would not do that, but maybe?

It shouldn't https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#_file_identity_2

The default is either device ID + inode but we are switching everything to fingerprint because inode reuse is more common than we originally thought. CC @rdner in case there is some subtlety here that I'm missing.

pkoutsovasilis:
> It shouldn't https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#_file_identity_2
> 
> The default is either device ID + inode but we are switching everything to fingerprint because inode reuse is more common than we originally thought. CC @rdner in case there is some subtlety here that I'm missing.

Just echoing the same message; AFAICT nothing should be re-indexed as fingerprint [is enabled](https://github.com/elastic/integrations/blob/main/packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs#L10) in filestream 🙂

deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
```diff
@@ -16,12 +16,12 @@ Config input for container logs
     namespace: {{ .Values.kubernetes.namespace }}
   use_output: {{ .Values.kubernetes.output }}
   streams:
-  - id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+  - id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
     data_stream:
       dataset: kubernetes.container_logs
       type: logs
```
rdner, line 22: As long as this translates to a running `filestream` input and we don't switch from 8.x to 9.x the change in the PR should cause no re-indexing.

Starting with 9.0 we will have fingerprint as a default file identity which is based on the file content, not its filesystem metadata.

rdner:
> Just echoing the same message; AFAICT nothing should be re-indexed as fingerprint [is enabled](https://github.com/elastic/integrations/blob/main/packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs#L10) in filestream 🙂

@pkoutsovasilis 

There is a problem with the template you linked, fingerprint is default now but this template expects the opposite, probably needs to be changed.

pkoutsovasilis:
> @pkoutsovasilis
> 
> There is a problem with the template you linked, fingerprint is default now but this template expects the opposite, probably needs to be changed.

@rdner  I believe you refer to `{{#if useFingerprint}}` part which by default is `true` thus I would say that the template expects this to be enabled, what am I missing? 🙂

rdner:
@pkoutsovasilis since 9.0 fingerprint is enabled by default. If the template has `useFingerprint = false` it will still be enabled. I don't think this is what expected there.

pchila:
@rdner Re: fingerprint mechanism I performed a quick test with a container that would output a very short log before exiting and I noticed that having fingerprint enabled (with default length of 1024) would result in logs not being ingested even after the container terminated... Is this a known issue/limitation/design feature ?

deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
belimawr: > Does this mean that an upgrade from 8.x to 9.x will result in a re-index of all files? That would be very bad.

No, we also implemented a feature to migrate the state from the old default (`native` that uses inode + device ID) to the new default `fingerprint`. If you're curious, [here is the PR](https://github.com/elastic/beats/pull/41762).

deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
```diff
@@ -16,12 +16,12 @@ Config input for container logs
     namespace: {{ .Values.kubernetes.namespace }}
   use_output: {{ .Values.kubernetes.output }}
   streams:
-  - id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
+  - id: kubernetes-container-logs-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}
```
belimawr, line 19: The new ID does not seem to be unique, this will cause problems and lead to data duplication.
Are you sure that 
`${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}` will always be unique and will make this block be rendered only once?

The previous implementation used the `${kubernetes.container.id}` because that's guaranteed to always be unique.
```suggestion
  - id: kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}
```

The key thing we have to ensure here is that there will never be two Filestream inputs started with the same ID.

belimawr: I see this file is in the helm chart folder, how does upgrade here work? Is there any change an user will upgrade and have the input ID changing? If so, that will lead to data duplication because the new ID will make all files be considered new, thus fully re-ingested.

deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
```diff
@@ -354,7 +354,7 @@ data:
             condition: '${host.platform} == ''windows'''
             ignore_older: 72h
       # Input ID allowing Elastic Agent to track the state of this input. Must be unique.
-      - id: container-log-${kubernetes.pod.name}-${kubernetes.container.id}
+      - id: container-log-${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.id}
```
belimawr, line 357: Same thing here and in the next ones about updating the ID. I see it is in the kustomize folder, however I'm not 100% sure how this file is used. Is there any chance that during an (automated) upgrade a user will go from the old ID to the new one?

belimawr:
> > It shouldn't https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#_file_identity_2
> > The default is either device ID + inode but we are switching everything to fingerprint because inode reuse is more common than we originally thought. CC @rdner in case there is some subtlety here that I'm missing.
> 
> Just echoing the same message; AFAICT nothing should be re-indexed as fingerprint [is enabled](https://github.com/elastic/integrations/blob/main/packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs#L10) in filestream 🙂

While that is correct, it is still incomplete. The input ID is part of the state ID, so changing the input ID (I mean the Filestream input ID, not the Elastic-Agent unit/input ID) will cause all files to be re-ingested regardless of the file identity used.

There is no way to migrate state from one input instance to another.

@blakerouse's concern is very valid and real here
> @pchila @pkoutsovasilis Do we need to worry about this change causing the files to be re-indexed?

If we change the IDs, like we're doing in `deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl` files will be re-ingested.

The question for me is more about how (if any) an update happens. My understanding of helm charts is that they can be updated, however if we're only talking about static examples in our documentation ([like our example manifest](https://raw.githubusercontent.com/elastic/elastic-agent/8.17/deploy/kubernetes/elastic-agent-managed-kubernetes.yaml) linked in [our docs](https://www.elastic.co/guide/en/fleet/current/running-on-kubernetes-managed-by-fleet.html#_step_1_download_the_elastic_agent_manifest)), then there is no issue because users won't be downloading/updating their manifests every new release. Well, that's what I'd expect users to do.

deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
pkoutsovasilis: @belimawr some things I have in mind, the combination of `${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name}` should always be unique in a k8s cluster, right?! 

also about the 
> how does upgrade here work

indeed when we restart we might re-ingest the logs but if we have to do it, we should do it now that helm chart is not GA yet. 🙂

belimawr: > the combination of ${kubernetes.namespace}-${kubernetes.pod.name}-${kubernetes.container.name} should always be unique in a k8s cluster, right?!

I hope so, however I'm not sure. It would be nice to back it with some K8s documentation. I know for sure `${kubernetes.container.id}` is always unique, hence it's the one I've always used/recommended.

> indeed when we restart shouldn't fingerprint be the same no matter the ID?!

pchila: Quoting from k8s docs https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#when-to-use-multiple-namespaces

> Namespaces provide a scope for names. Names of resources need to be unique within a namespace, but not across namespaces.
hence, given a namespace + podname pair it's guaranteed to be unique in the cluster (as opposed to the previous id that was NOT namespaced)

Within a pod spec container names must be unique so once the pod is created the tuple (namespace, pod name, container name) is unique and the input definition does not change with container restarts.

deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
pchila: please look at my previous response

deploy/helm/elastic-agent/templates/integrations/_kubernetes/_kubernetes_logs_containers.tpl
belimawr: Thanks for checking the docs and clarifying @pchila !!

Then ignore the change I suggested.

The only issue that persists is the ID changing when the helm chart is upgraded... That will cause re-ingestion of files, even with fingerprint.

deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
belimawr: > Does this mean that filebeat will not keep track of the actual file metadata in the registry ?

TL;DR: Not in the way you expect, if you change the input ID, that input will re-ingest all files.

Longer version:
The registry for Filestream is "namespaced" by the input ID, an instance of the Filestream input will only have access to entries in the registry that match its ID. See the registry entry I mentioned on https://github.com/elastic/elastic-agent/pull/6583#discussion_r1932505269. The state key is
```
filestream::my-filestream-id::native::8787338-65029
```
The format is:
```
filestream::<input ID>::<file identity name>::<state ID>
```
On the example above:
- Input ID: `my-filestream-id`
- File identity name: `native`
- State ID: `8787338-65029`.

If any of those change, it's a new entry in the registry and the file is re-ingested.

We can jump on a call tomorrow if you have any questions.

pchila: I just tested switching from the new to the old input id using the standalone k8s manifest and I observed a reingestion of container logs because of the change of input id.

Changing datastream id does not look like it triggers any re-ingestion.

This means that people upgrading from agents installed using plain k8s manifests to installing agent with the helm chart will experience the same reingestion since the input id is different there as well :disappointed: 

To be clear this will happen only once during the switch and only for the logs that are present on the node at that moment.

@nimarezainia @cmacknz @jlind23 what is your opinion on reingestion due to filestream input id change? Is it acceptable to reingest "once" when the input id changes (when we switch from the old input definition to the new or from the plan k8s manifest to the helm chart) ?

pchila: Because the input is by pod and container name (withing the pod spec) now, we don't track the files by container id anymore. 
For all intents and purposes we can even use a constant input id `container-logs` and have only the datastream id changing.
The input id containing the container id is gonna change very fast when we start having containers in CrashLoop.

If it's purely a matter of giving a unique string that has no relation with the paths that the input is actually specifying, we can keep using a containerid that changes with every container restart, it would just be strange and less readable

pchila: Adding on my previous response: if the input id changes and the path matches all the available logs for the container in the pod (including the rotated ones) we will have log duplication at every container restart.

If we don't target all the logs for the container in the pod under `/var/log/pods/* ` we are gonna miss rotated logs and  not implement a solution for https://github.com/elastic/elastic-agent/issues/5164

cmacknz: Missing logs is worse than one time duplication of logs IMO. We changed the ID before in the past as it was not unique and there wasn't much backlash.

If we convinced this is 100% more correct it is probably worth doing. If we are really worried about duplication only make the change in 9.0 where the file identity is changing anyway.

The way most hosted k8s providers do log rotation, we are looking at re-ingesting 10 MB of logs per node at worst. This is because the log rotation is 5 10 MB files but the oldest 4 are gzipped compressed and we can't read them. We can only read the file actively being written to.

cmacknz:
> There is no way to migrate state from one input instance to another.

We could build one if we really needed it. We could add an `old_id` field with the old pattern and then we could have the input pick up from there and write updates with the new one or something. This isn't the first time we've needed to change the input ID in the k8s integration so maybe it's worth thinking about this.

deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset-configmap.yaml
blakerouse: I understand, thanks for the explination @pchila.

Is it possible to tell filebeat that it used to have an old ID so it can use it as a basis and then migrate to the new ID? I don't know if that is possible or if its hard to add. Maybe having something like:

```
registry:
  previous_id: container-log-${kubernetes.pod.name}-${kubernetes.container.id}
```

Just a thought to provide the best experience for our customers. Doing this in 9.0 only is probably safest, but I think we would need to place a note in the release notes that upgrade from 8.x to 9.x would result in a re-ingestion of all container logs. I think that is just something that is crazy to accept for an upgrade path.... but it is a path...

cmacknz: Blake I both thought of the `previous_id` idea at about the same time (https://github.com/elastic/elastic-agent/pull/6583#issuecomment-2621859914), I think we should definitely explore this.

I created https://github.com/elastic/beats/issues/42472 to track and discuss this idea.

belimawr: Just for context, when we change the file identity in 9.0 there will be **no data re-ingestion** because we can migrate the state from `native` (the current default) to `fingerprint` (the new default), that was all implemented by https://github.com/elastic/beats/pull/41762.

Regarding the `previous_id` idea, I'll comment on the issue itself.

pchila:
@belimawr I see that https://github.com/elastic/beats/issues/42472 is done, does it mean that we can resume this and https://github.com/elastic/integrations/pull/12500 PRs ?

What are the new settings that agent needs to use ?

belimawr:
> @belimawr I see that [elastic/beats#42472](https://github.com/elastic/beats/issues/42472) is done, does it mean that we can resume this and [elastic/integrations#12500](https://github.com/elastic/integrations/pull/12500) PRs ?
> 
> What are the new settings that agent needs to use ?

Hi @pchila, yes, https://github.com/elastic/integrations/pull/12500 can resume.

You can look at the [latest documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#filebeat-input-filestream-take-over), the TL;DR is:
```
type: filestream
id: new-id
take_over:
  enabled: true
  from_ids:
    - old-id-1
    - old-id-2
```

Remember that for this feature under Elastic-Agent to work, all inputs (old and new) need to came from the same component/data folder.

Ping me if you have any more questions ;)

belimawr:
You also need to ensure this is limited to the correct version of Filebeat.

ebeahan:
Cleaning up the repo's open PR list by closing stale item.

Feel free to reopen when work is ready to continue.

PR #6585, opened 2025-01-23
Change the default gRPC port to 0 when in a container

cmacknz:
## What does this PR do?

Changes the default port used when agent runs in a container to port 0. There is at least one user whose rollout of Elastic Agent is experiencing collisions with another application on the same port (nobody else could ever pick 6789 as a default port surely...) and I thought this would be straight forward to fix quickly. I was wrong. I expected ~2 hours and spent 3 days:

1. It looked like we already have a separate default config file for Docker containers I could just update: https://github.com/elastic/elastic-agent/blob/9b8a25f8feb5919965f1e5480d26e0324dd88736/_meta/config/elastic-agent.docker.yml.tmpl#L88-L89
2. This didn't work because the Helm chart defines its own configuration separately. I also didn't want to define this more than once because we'll probably change it again later (see **note** below): https://github.com/elastic/elastic-agent/blob/9b8a25f8feb5919965f1e5480d26e0324dd88736/deploy/helm/elastic-agent/templates/agent/k8s/_secret.tpl#L5
3. The `run()` function that is the entrypoint of the agent already has a hook for overriding configuration https://github.com/elastic/elastic-agent/blob/4199196174f1361e14c3ed6e6315b97bf98c3b81/internal/pkg/agent/cmd/run.go#L426-L428 that is already used in the container command to override the logging configuration https://github.com/elastic/elastic-agent/blob/4199196174f1361e14c3ed6e6315b97bf98c3b81/internal/pkg/agent/cmd/container.go#L769 but of course updating this to change the gRPC port also didn't work.
4. It turns out we load the configuration from disk at least twice at startup, once in https://github.com/elastic/elastic-agent/blob/9b8a25f8feb5919965f1e5480d26e0324dd88736/internal/pkg/agent/cmd/run.go#L415 and again in https://github.com/elastic/elastic-agent/blob/4199196174f1361e14c3ed6e6315b97bf98c3b81/internal/pkg/agent/application/application.go#L99 The second use didn't apply the overrides, the overrides only applied for the first case because that's when we created our logger and the only current overrides are for logging 🤦. Probably we could tidy this up but I've already spent enough time on this one.

**note** We should eventually stop using local TCP at all for the control protocol between sub-processes, and switch to unix sockets / named pipes. The capability for this was added in https://github.com/elastic/elastic-agent/pull/4249, but it was left disabled because endpoint-security doesn't support it yet (because the upstream gRPC C++ client doesn't support it on Windows). We have now removed endpoint-security from our containers which would allow us to switch to unix sockets there, but this change required an elastic-agent-client package update and we need to test that every client we have has it first. This was more testing effort than I wanted to take on now, but I will create a follow up issue to do this.

## Why is it important?

This automatically avoids port collisions between Elastic Agents using `hostNetwork: true` on Kubernetes as our DaemonSet does by default and other applications (or other Elastic Agents). 

## Disruptive User Impact

None, but in case I'm wrong about this I made it possible to choose a specific port with an environment variable.

```
INSTANCE_PROVISIONER=kind SNAPSHOT=true GOTEST_FLAGS="-test.run TestKubernetesAgentHelm" TEST_PLATFORMS="kubernetes/arm64/1.31.0/basic" mage -v integration:kubernetes
```

### DIFF
--- a/testing/integration/kubernetes_agent_standalone_test.go
+++ b/testing/integration/kubernetes_agent_standalone_test.go
@@ -266,7 +266,10 @@ func TestKubernetesAgentHelm(t *testing.T) {
 		steps      []k8sTestStep
 	}{
 		{
-			name: "helm standalone agent default kubernetes privileged",
+			// Configure the perNode and clusterWide agents to both use host networking. On the node that
+			// runs the clusterWide agent, this tests that two agents do not try to bind to the same
+			// gRPC control protocol port by default preventing one from starting.
+			name: "helm standalone agent default kubernetes privileged without host network port collision",
 			steps: []k8sTestStep{
 				k8sStepCreateNamespace(),
 				k8sStepHelmDeploy(agentK8SHelm, "helm-agent", map[string]any{
@@ -279,6 +282,14 @@ func TestKubernetesAgentHelm(t *testing.T) {
 							"repository": kCtx.agentImageRepo,
 							"tag":        kCtx.agentImageTag,
 							"pullPolicy": "Never",
+						},
+						"presets": map[string]any{
+							"clusterWide": map[string]any{
+								"hostNetwork": true,
+							},
+							"perNode": map[string]any{
+								"hostNetwork": true,
+							},
 						},
 					},
 					"outputs": map[string]any{
--- a/changelog/fragments/1737657890-Use-a-random-free-port-for-sub-process-communication-in-containers.yaml
+++ b/changelog/fragments/1737657890-Use-a-random-free-port-for-sub-process-communication-in-containers.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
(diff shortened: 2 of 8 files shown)


testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -266,7 +266,10 @@ func TestKubernetesAgentHelm(t *testing.T) {
 		steps      []k8sTestStep
 	}{
 		{
-			name: "helm standalone agent default kubernetes privileged",
+			// Configure the perNode and clusterWide agents to both use host networking. On the node that
+			// runs the clusterWide agent, this tests that two agents do not try to bind to the same
+			// gRPC control protocol port by default preventing one from starting.
+			name: "helm standalone agent default kubernetes privileged without host network port collision",
```
cmacknz, line 272: @pkoutsovasilis I just added host networking to the clusterWide agent in an existing k8s test to test this, let me know if it would be better as a separate test.

blakerouse (approved):
Looks good.

The day sockets/npipe can be used for all installation types with be a glorious day!

testing/integration/kubernetes_agent_standalone_test.go
pkoutsovasilis: yy that makes total sense, I do also believe that there is no need to have a separate one just for this case 🙂 (our k8s integration tests already take a lot of time)

pkoutsovasilis (approved):
LGTM 👍 this change will help a lot with k8s deployments

michalpristas (approved):

PR #6587, opened 2025-01-23
Add filesource provider

### LINKED ISSUE #6362: Add filesource provider to composable inputs
**Describe the enhancement:**

There are cases where the ability to get information from files and place that information into a policy. This could be because this information is secret and shouldn't be exposed in an environment variable, or that the information could change while the Elastic Agent is running and an environment variable cannot but the contents of the file can.

Imagine that inside of the Elastic Agent container that you want to read a secret from kubernetes but you don't want the Elastic Agent to communicate directly with kubernetes because of security reasons. In this environment you can only mount the secret value as content of a file in a container, and this value can change at any time.

To enable this it must be clear that the contents of files can only be read from files that have been whitelisted in the provides configuration, and the provider must watch for changes of these files.

- When the contents of a file can be read from disk and used inside of `inputs`. (eg. `keyvalue: ${filesource.secretfile}`)
- When the contents of that file are changed that the `${filesource.secretfile}` is updated and the policy is updated with the new value.

**Example configuration**

```
providers:
  filesource:
   secretfile:
     path: /etc/secretfile
   otherfile:
     path: /etc/anotherfile
```

The nested `path:` sub-key is in-place to allow future capabilities in the case that more features are added to the provider like reading JSON, YAML content and extracting values from keys. At the moment the default of basically `type: raw` can be assumed.

blakerouse:
## What does this PR do?

Adds the `filesource` provide to composable.

The provider watches for changes of the files and updates the values of the variables when the content of the file changes.

## Why is it important?

This allows information from the filesystem to be used as variables in the Elastic Agent configuration. This information is only allowed by explicitly configuring the provider to read this information from the disk. The policy cannot just read any file, it has to be explicitly configured to allow it.

## Checklist

- ~~[ ] I have added an integration test or an E2E test~~ (covered by unit tests well)

## Disruptive User Impact

None

Create a configuration:

`elastic-agent.yml`

```
provides:
  filesource:
    sources:
      file1:
        path: ./file1

inputs:
 - id: filestream
   type: filestream
   paths:
     - ${filesource.file1}
```

`echo "/var/log/syslog" > ./file1`

Observe that syslog is set to paths.

## Related issues

- Closes #6362

### DIFF
--- a/internal/pkg/composable/providers/filesource/filesource.go
+++ b/internal/pkg/composable/providers/filesource/filesource.go
@@ -1,0 +1,177 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package filesource
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/elastic/elastic-agent/internal/pkg/composable"
+	"github.com/elastic/elastic-agent/internal/pkg/config"
+	corecomp "github.com/elastic/elastic-agent/internal/pkg/core/composable"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+)
+
+func init() {
+	// filesource provider reads and watches for changes on files that are defined in the provider configuration.
+	//
+	// To be notified when a file is change the provider will watch the parent directory of the file so if the file
+	// is replaced that it will read the new contents. If a file doesn't exist or the provider is unable to read
+	// the file then it will report the value as an empty string.
+	//
+	// If the provided path happens to be a directory then it just report the value as an empty string.
+	composable.Providers.MustAddContextProvider("filesource", ContextProviderBuilder)
+}
+
+type fileSourceConfig struct {
+	Type string `config:"type"`
+	Path string `config:"path"`
+}
+
+type contextProvider struct {
+	logger *logger.Logger
+
+	sources map[string]fileSourceConfig
+}
+
+// Run runs the filesource context provider.
+func (c *contextProvider) Run(ctx context.Context, comm corecomp.ContextProviderComm) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("failed to create watcher: %w", err)
+	}
+	defer watcher.Close()
+
+	// invert the mapping to map paths to source names
+	inverted := make(map[string][]string, len(c.sources))
+	for sourceName, sourceCfg := range c.sources {
+		sources, ok := inverted[sourceCfg.Path]
+		if !ok {
+			sources = []string{sourceName}
+		} else {
+			sources = append(sources, sourceName)
+		}
+		inverted[sourceCfg.Path] = sources
+	}
+
+	// determine the paths to watch (watch is performed on the directories that contain the file)
+	//
+	// you cannot register the same directory multiple times so this ensures its only registered once
+	paths := make([]string, 0, len(c.sources))
+	for _, cfg := range c.sources {
+		parent := filepath.Dir(cfg.Path)
+		if !slices.Contains(paths, parent) {
+			paths = append(paths, parent)
+		}
+	}
+	for _, path := range paths {
+		err = watcher.Add(path)
+		if err != nil {
+			return fmt.Errorf("failed to watch path %q: %w", path, err)
+		}
+	}
+
+	// read the initial values after the watch has started
+	// this ensures that if the value changed between this code and the loop below
+	// the updated file changes will not be missed
+	current := make(map[string]interface{}, len(c.sources))
+	for path, sources := range inverted {
+		value := c.readContents(path)
+		for _, source := range sources {
+			current[source] = value
+		}
+	}
+	err = comm.Set(current)
+	if err != nil {
+		return fmt.Errorf("failed to set current context: %w", err)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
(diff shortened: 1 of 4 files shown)


internal/pkg/composable/providers/filesource/filesource.go
```diff
@@ -0,0 +1,177 @@
...
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				// watcher was closed
+				return nil
+			}
+			c.logger.Errorf("file watcher errored: %s", err)
+		case e, ok := <-watcher.Events:
+			if !ok { // Channel was closed (i.e. Watcher.Close() was called).
+				// watcher was closed
+				return nil
```
pkoutsovasilis, line 109: just checking; what happens when a watcher is closed, for each own reasons, and we return `nil` but the provider still needs to run, as there are inputs depending on it?

internal/pkg/composable/providers/filesource/filesource.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+
+			path := filepath.Clean(e.Name)
+			sources, ok := inverted[path]
```
andrzej-stencel, line 113: This searches for `path` that is processed by `filepath.Clean` among paths in `inverted` that are not cleaned.

If for some reason user configures 'path: ././file.txt` then this path will never be found, right?

andrzej-stencel, line 113: On Windows, paths are case insensitive. If user configures `path: myfile.txt`, but the path in Event.Name is `MyFile.txt`, then the file won't be found in the list, right? Maybe worth having a separate Windows test for this?

internal/pkg/composable/providers/filesource/filesource.go
```diff
@@ -0,0 +1,177 @@
...
+				if err != nil {
+					return fmt.Errorf("failed to set current context from notify event: %w", err)
+				}
+			}
+		}
+	}
+}
+
+func (c *contextProvider) readContents(path string) string {
+	data, err := os.ReadFile(path)
```
andrzej-stencel, line 137: Should we guard against very big files? I'm not sure if this is something we should worry about.

internal/pkg/composable/providers/filesource/filesource.go
andrzej-stencel: Oh I can see now that below in `ContextProviderBuilder` function the relative paths are replaced with absolute ones and cleaned in the process. :+1:

internal/pkg/composable/providers/filesource/filesource.go
blakerouse: If the channel is closed then the watcher is closed and the only way the watcher is closed is if `defer watcher.Close` occurs. So this really isn't ever going to happen, I will update the code to just run the code in the select/case when the channel is okay. That will make it more clear and then it will return on the `ctx.Done()` and return the `ctx.Err()` which is more appropriate the the caller of `func Run()`

internal/pkg/composable/providers/filesource/filesource.go
```diff
@@ -0,0 +1,177 @@
...
+			if !ok {
+				// watching the directory, it can contain files that we are not watching
+				// ignore these events unless we are actively watching this file
+				continue
+			}
+
+			switch {
+			case e.Op&(fsnotify.Create|fsnotify.Write|fsnotify.Remove) != 0:
+				// file was created, updated, or deleted (update the value)
+				value := c.readContents(path)
```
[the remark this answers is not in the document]
blakerouse: I don't believe that events will be lost, there is a buffer already held by the each file watcher implementation depending on OS type. I also don't expect this to be a large list of files, it should be very small.

internal/pkg/composable/providers/filesource/filesource.go
blakerouse: Good catch, thanks! I will add a test for this that will run on Windows and see what needs to be added.

internal/pkg/composable/providers/filesource/filesource.go
blakerouse: It is not a bad idea, I think I will put a guard on it. These files are meant to be very small. Maybe I will put a maximum of like 4k in size.

internal/pkg/composable/providers/filesource/filesource.go
blakerouse: Updated.

internal/pkg/composable/providers/filesource/filesource.go
blakerouse: Added.

internal/pkg/composable/providers/filesource/filesource.go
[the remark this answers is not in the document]
pkoutsovasilis: I essentially raised the same comment as [this one](https://github.com/elastic/elastic-agent/pull/6587#discussion_r1928586206) but for some reason, your answer is different 🙂 Now you can definitely lose events depending on how fast we read from the underlying event buffer (utilised in your code as Events channel); just take a look at [ErrEventOverflow](https://github.com/fsnotify/fsnotify/blob/main/fsnotify.go#L243) and it's usages. So either be explicit with a comment that we are best-effort here and fingers crossed 🤞 or you could try to recover from such cases by reading the latest value of each file and resubscribing for events.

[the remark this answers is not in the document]
blakerouse: I took that comment differently as I was thinking file size more related to memory usage then I was the queue filling up. I have added a way to handle ErrEventOverflow in the code, let me know what you think. I honestly don't see it happening as I would expect the usage of this provide to be only 1 to 5 files max.

cmacknz:
You can file an issue in https://github.com/elastic/ingest-docs describing how this works to get it included in the provider documentation, assuming we want this available outside of internal use.

blakerouse:
@cmacknz https://github.com/elastic/ingest-docs/issues/1712

PR #6598, opened 2025-01-24
Emit vars even if provider data is empty from the start

swiatekm:
## What does this PR do?

We only signal from providers if the emitted data is different from the current state. This caused a bug where the data wasn't ever emitted if it was empty. The provider controller now distinguishes between no data (yet) and empty data. As a result, it's again possible to effectively disable all providers by using a local provider with no data.

This is a regression caused by https://github.com/elastic/elastic-agent/pull/6114. It doesn't apply to the main and 8.x branches due to https://github.com/elastic/elastic-agent/pull/6169 refactoring provider initialization in those.

## Why is it important?

Fixes a regression where agent wouldn't generate any configuration if the only enabled context providers only emitted empty data. This is something users do if they want to effectively disable providers.

## Checklist

### DIFF
--- a/changelog/fragments/1737735780-fix-empty-provider-data.yaml
+++ b/changelog/fragments/1737735780-fix-empty-provider-data.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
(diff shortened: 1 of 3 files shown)


blakerouse (approved):
Nice catch. Looks good.

swiatekm:
> Nice catch. Looks good.

On a related note, should we make it possible to run agent with all providers disabled? I think https://github.com/elastic/elastic-agent/pull/6169 effectively achieves that, but I'm not sure if intentionally or not.

michalpristas (approved):

blakerouse:
@swiatekm it was intentional, a requirement of not having any variables.

PR #6611, opened 2025-01-27
Integration tests: use the mage target to get agent version

pazone:
## What does this PR do?
The way how we defined the agent version for integration tests wasn't reliable. It works correctly on the main branch but might work incorrectly on 8.17. That blocks the backports to [8.17](https://github.com/elastic/elastic-agent/pull/6323) and [8.16](https://github.com/elastic/elastic-agent/pull/6325). The new approach utilizes version.go which is automatically updated. 
* Created the `printVersion` target
* Modified integration test script

## Checklist

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/4585

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/buildkite-integration-tests.sh
+++ b/.buildkite/scripts/buildkite-integration-tests.sh
@@ -45,7 +45,7 @@ fully_qualified_group_name="${GROUP_NAME}${root_suffix}_${os_data}"
 outputXML="build/${fully_qualified_group_name}.integration.xml"
 outputJSON="build/${fully_qualified_group_name}.integration.out.json"
 set +e
-TEST_BINARY_NAME="elastic-agent" AGENT_VERSION="${PACKAGE_VERSION}" SNAPSHOT=true gotestsum --no-color -f standard-quiet --junitfile "${outputXML}" --jsonfile "${outputJSON}" -- -tags integration -test.shuffle on -test.timeout 2h0m0s github.com/elastic/elastic-agent/testing/integration -v -args -integration.groups="${GROUP_NAME}" -integration.sudo="${TEST_SUDO}"
+TEST_BINARY_NAME="elastic-agent" AGENT_VERSION=$(mage -v printVersion) SNAPSHOT=true gotestsum --no-color -f standard-quiet --junitfile "${outputXML}" --jsonfile "${outputJSON}" -- -tags integration -test.shuffle on -test.timeout 2h0m0s github.com/elastic/elastic-agent/testing/integration -v -args -integration.groups="${GROUP_NAME}" -integration.sudo="${TEST_SUDO}"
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -26,15 +26,15 @@ fi
...
 
-PACKAGE_VERSION="$(cat .package-version)"
-if [[ -n "$PACKAGE_VERSION" ]]; then
-    PACKAGE_VERSION=${PACKAGE_VERSION}"-SNAPSHOT"
-fi
+# Parsing version.go. Will be simplified here: https://github.com/elastic/ingest-dev/issues/4925
+AGENT_VERSION=$(grep "const defaultBeatVersion =" version/version.go | cut -d\" -f2)
+AGENT_VERSION="${AGENT_VERSION}-SNAPSHOT"
+export AGENT_VERSION
+echo "~~~ Agent version: ${AGENT_VERSION}"
```
v1v, line 37: I see this in two places, do you mind moving this logic to a make goal or something else? Otherwise, it's hard to maintain.

For instance, the APM Server uses:

- https://github.com/elastic/apm-server/blob/42e0cbf5b049a92b5536f34090b15f5eb0a7167e/go.mk#L11
- and https://github.com/elastic/apm-server/blob/42e0cbf5b049a92b5536f34090b15f5eb0a7167e/Makefile#L146-L149

Therefore the call is as simple as this https://github.com/elastic/apm-server/blob/42e0cbf5b049a92b5536f34090b15f5eb0a7167e/.buildkite/scripts/dra.sh#L39

pazone: What is the second place? If it's the PS file for Windows then unfortunately it's hard to avoid. 
I also added a comment above. The mentioned [issue](https://github.com/elastic/ingest-dev/issues/4925) will replace it with `AGENT_VERSION=$(cat .agent-version)`. 
I would like to ask you a QQ about your example if you don't mind. Wouldn't it be easier to create a similar `.version` file and embed it to the `version.go`? Then getting the version will be as simple as it mentioned above

v1v: > What is the second place? If it's the PS file for Windows then unfortunately it's hard to avoid(Possible but the solution is controversial).

Yep

> Wouldn't it be easier to create a similar .version file and embed it to the version.go? 

I think so, thanks for bringing this to my attention, if there is a plan, then all good.

v1v (approved):

oakrizan (approved):

oakrizan:
https://github.com/Mergifyio backport 8.17

v1v:
@pazone , I cannot find this backport in `9.0`, can you confirm whether it's needed?

PR #6619, opened 2025-01-28
enhancement(5423): added logic to replaces scheduler with long-wait scheduler in case of exceeded unauth response limit

### LINKED ISSUE #5428: Remove automatic unenrollment after 7 Fleet authentication failures

kaanyalti:
- Enhancement

## What does this PR do?

Removes the forced unenroll from fleet gateway. Adds logic in the fleet gateway to switch out the scheduler used for checkins. If the unauthorized response limit is exceeded, a the scheduler is replaced with one that has a long wait duration. When the gateway receives a successful response, it switches back to using the regular scheduler with the shorter wait duration.

## Why is it important?

Currently the agent unenrolls after 7 unauthorized error responses. This can causes problems in disaster recovery scenarios where users may have to manually intervene. 

## Checklist

## Disruptive User Impact

None

- Create ESS deployment
- Build the agent locally
- Enroll the agent
- In dev tools find the access token and delete it
```
GET /.security-7/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "name": "AGENT ID"
          }
        }
      ]
    }
  }
}
```
```
DELETE /_security/api_key
{
  "ids": ["KEY ID"]
}
```
- Follow the agent logs `sudo elastic-agent logs -f`
- After a while you will see `retrieved an invalid api key error '10' times. will use long scheduler` error message in the logs

 Due to the backoff algorithm used, this test can take a long time. In order to see immediate results comment out the following code block
```
			if !bo.Wait() {
				if ctx.Err() != nil {
					// if the context is cancelled, break out of the loop
					break
				}

				// This should not really happen, but just in-case this error is used to show that
				// something strange occurred and we want to log it and report it.
				err := errors.New(
					"checkin retry loop was stopped",
					errors.TypeNetwork,
					errors.M(errors.MetaKeyURI, f.client.URI()),
				)

				f.log.Error(err)
				f.errCh <- err
				return nil, err
			}
```

## Related issues

- Closes #5428 
- Relates #6760

### DIFF
--- a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
+++ b/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
@@ -39,9 +39,9 @@ const (
 
 // Default Configuration for the Fleet Gateway.
 var defaultGatewaySettings = &fleetGatewaySettings{
-	Duration:    1 * time.Second,        // time between successful calls
-	Jitter:      500 * time.Millisecond, // used as a jitter for duration
-	ErrDuration: 1 * time.Hour,          // time between calls when the agent exceeds unauthorized response limit
+	Duration:                     1 * time.Second,        // time between successful calls
+	Jitter:                       500 * time.Millisecond, // used as a jitter for duration
+	ErrConsecutiveUnauthDuration: 1 * time.Hour,          // time between calls when the agent exceeds unauthorized response limit
 	Backoff: backoffSettings{ // time after a failed call
 		Init: 60 * time.Second,
 		Max:  10 * time.Minute,
@@ -49,10 +49,10 @@ var defaultGatewaySettings = &fleetGatewaySettings{
 }
 
 type fleetGatewaySettings struct {
-	Duration    time.Duration   `config:"checkin_frequency"`
-	Jitter      time.Duration   `config:"jitter"`
-	Backoff     backoffSettings `config:"backoff"`
-	ErrDuration time.Duration
+	Duration                     time.Duration   `config:"checkin_frequency"`
+	Jitter                       time.Duration   `config:"jitter"`
+	Backoff                      backoffSettings `config:"backoff"`
+	ErrConsecutiveUnauthDuration time.Duration
 }
 
 type backoffSettings struct {
@@ -361,16 +361,14 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse, 
 		f.unauthCounter++
 		if f.shouldUseLongSched() {
 			f.log.Warnf("retrieved an invalid api key error '%d' times. will use long scheduler", f.unauthCounter)
-			f.scheduler = scheduler.NewPeriodic(defaultGatewaySettings.ErrDuration)
+			f.scheduler.SetDuration(defaultGatewaySettings.ErrConsecutiveUnauthDuration)
 			return &fleetapi.CheckinResponse{}, took, nil
 		}
 
 		return nil, took, err
 	}
 
-	if _, ok := f.scheduler.(*scheduler.PeriodicJitter); !ok {
-		f.scheduler = scheduler.NewPeriodicJitter(defaultGatewaySettings.Duration, defaultGatewaySettings.Jitter)
-	}
+	f.scheduler.SetDuration(defaultGatewaySettings.Duration)
 
 	f.unauthCounter = 0
 	if err != nil {
--- a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
+++ b/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
@@ -74,7 +74,6 @@ type FleetGateway struct {
 	log                *logger.Logger
 	client             client.Sender
 	scheduler          scheduler.Scheduler
-	isLongSched        bool
 	settings           *fleetGatewaySettings
 	agentInfo          agentInfo
 	acker              acker.Acker
@@ -127,7 +126,6 @@ func newFleetGatewayWithScheduler(
 		acker:        acker,
 		stateFetcher: stateFetcher,
 		stateStore:   stateStore,
-		isLongSched:  false,
 		errCh:        make(chan error),
 		actionCh:     make(chan []fleetapi.Action, 1),
 	}, nil
@@ -135,27 +133,6 @@ func newFleetGatewayWithScheduler(
 
 func (f *FleetGateway) Actions() <-chan []fleetapi.Action {
 	return f.actionCh
-}
-
-// Replaces scheduler depending on whether the fleetgateway needs to wait longer
-// or shorter between checkins. Regular Periodic scheduler is used for long wait
-// times as the jitter is not significant compared to the wait time, and
-// PeriodicJitter is used otherwise
-func (f *FleetGateway) tryReplaceScheduler() {
-	switch f.scheduler.(type) {
-	case *scheduler.Periodic:
-		if !f.isLongSched {
-			// If the current scheduler is Periodic and the wait time needs to be
-			// short, switch to using PeriodicJitter with short duration
-			f.scheduler = scheduler.NewPeriodicJitter(defaultGatewaySettings.Duration, defaultGatewaySettings.Jitter)
-		}
-	case *scheduler.PeriodicJitter:
-		if f.isLongSched {
-			// If the current scheduler is PeriodicJitter and the wait time needs to
-			// be long, switch to using Periodic with long duration
-			f.scheduler = scheduler.NewPeriodic(defaultGatewaySettings.ErrDuration)
-		}
-	}
 }
 
 func (f *FleetGateway) Run(ctx context.Context) error {
@@ -181,8 +158,6 @@ func (f *FleetGateway) Run(ctx context.Context) error {
 			if err != nil {
 				continue
 			}
-
-			f.tryReplaceScheduler()
 
 			actions := make([]fleetapi.Action, len(resp.Actions))
 			copy(actions, resp.Actions)
@@ -385,16 +360,17 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse, 
 	if isUnauth(err) {
 		f.unauthCounter++
 		if f.shouldUseLongSched() {
-			// If we exceed the unauth response
-			f.isLongSched = true
 			f.log.Warnf("retrieved an invalid api key error '%d' times. will use long scheduler", f.unauthCounter)
+			f.scheduler = scheduler.NewPeriodic(defaultGatewaySettings.ErrDuration)
 			return &fleetapi.CheckinResponse{}, took, nil
 		}
 
 		return nil, took, err
 	}
 
-	f.isLongSched = false
+	if _, ok := f.scheduler.(*scheduler.PeriodicJitter); !ok {
+		f.scheduler = scheduler.NewPeriodicJitter(defaultGatewaySettings.Duration, defaultGatewaySettings.Jitter)
+	}
 
 	f.unauthCounter = 0
 	if err != nil {
--- a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
+++ b/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
@@ -6,7 +6,6 @@ package fleet
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -356,17 +384,18 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
...
 		f.unauthCounter++
-
-		if f.shouldUnenroll() {
-			f.log.Warnf("retrieved an invalid api key error '%d' times. Starting to unenroll the elastic agent.", f.unauthCounter)
-			return &fleetapi.CheckinResponse{
-				Actions: []fleetapi.Action{&fleetapi.ActionUnenroll{ActionID: "", ActionType: "UNENROLL", IsDetected: true}},
-			}, took, nil
+		if f.shouldUseLongSched() {
+			// If we exceed the unauth response
+			f.isLongSched = true
```
swiatekm, line 389: Is there a reason why we set this variable here, but actually change the scheduler later? It'd be significantly simpler to just directly change the scheduler here, we wouldn't even need `tryReplaceScheduler`.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -30,25 +30,29 @@ import (
...
+	fleetStateStarting = "starting"
+)
 
 // Default Configuration for the Fleet Gateway.
 var defaultGatewaySettings = &fleetGatewaySettings{
-	Duration: 1 * time.Second,        // time between successful calls
-	Jitter:   500 * time.Millisecond, // used as a jitter for duration
+	Duration:    1 * time.Second,        // time between successful calls
+	Jitter:      500 * time.Millisecond, // used as a jitter for duration
+	ErrDuration: 1 * time.Hour,          // time between calls when the agent exceeds unauthorized response limit
```
swiatekm, line 44: This seems a bit high for a default. I wouldn't want to go higher than 5 minutes.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
kaanyalti: I added it here because this code path gets executed with each retry whereas the the switch case in the Run function gets run once per checkin. Aside from this, this function is already doing quite a bit, that's why I added it in Run. I don't have any strong opinions on where scheduler switch should be though. I can implement it the way you're suggesting if you want.

swiatekm: Correct me if I'm wrong, but if we do need to switch scheduler to long, then we return from the function right after. And we do the check every request anyway. I'd prefer to do it immediately because it's easier to understand. Also, we know what scheduler we want to use here, so we can avoid the clunkiness of `tryReplaceScheduler`, as well as get rid of the `isLongSched` attribute.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
kaanyalti: The duration for error case was mentioned in the issue by @cmacknz, so that's what I went with, but I can use something shorter.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
kaanyalti: Yes that is true, in the case we need to switch to long scheduler, we return immediately. I'll update right away

swiatekm (approved):
LGTM, although I still think the default should be lower than 1 hour.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -356,17 +359,19 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
...
 		f.unauthCounter++
-
-		if f.shouldUnenroll() {
-			f.log.Warnf("retrieved an invalid api key error '%d' times. Starting to unenroll the elastic agent.", f.unauthCounter)
-			return &fleetapi.CheckinResponse{
-				Actions: []fleetapi.Action{&fleetapi.ActionUnenroll{ActionID: "", ActionType: "UNENROLL", IsDetected: true}},
-			}, took, nil
+		if f.shouldUseLongSched() {
+			f.log.Warnf("retrieved an invalid api key error '%d' times. will use long scheduler", f.unauthCounter)
+			f.scheduler = scheduler.NewPeriodic(defaultGatewaySettings.ErrDuration)
```
cmacknz, line 364: This should still be using jitter, just on a longer request duration. You do not want 50,000 agents accidentally checking it at the exact same moment unnecessarily.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -356,17 +359,19 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
...
+			return &fleetapi.CheckinResponse{}, took, nil
 		}
 
 		return nil, took, err
 	}
 
+	if _, ok := f.scheduler.(*scheduler.PeriodicJitter); !ok {
```
cmacknz, line 371: This interface check seems dangerous, if somebody updates a method signature or something this will fail at runtime but not compile time, leaving us stuck on the long scheduler.

Why was this necessary?

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
cmacknz: Some context: This is explicitly handling a force unenroll from the Fleet UI, which just revokes API keys but leaves the agent running. 

When this happens agent keeps checkin in indefinitely until the service stops, which might not ever happen. This pollutes our telemetry with very rapid retries of requests that will never succeed and places unnecessary load on Fleet Server. So we tried to detect API key revocation and unenroll when that happens. 

This didn't consider the case where some other bug or disaster unintentionally caused mass API key revocation or unavailability (e.g. Fleet Server can't reach Elasticsearch for 1+ hour). We recently had a support case where exactly this happened and once an agent is unenrolled there is no way to recover it.

Since unenroll is destructive and unrecoverable, then the next best thing is reducing the request frequency. This is where my arbitrary number of 1 hour came from after 7 unauthorized request came from.

We should still try to protect ourselves from accidentally checking in at 1 hour intervals. I think this logic now will reset every time the agent restarts, so just rebooting the machine or agent service should put us back on the fast path.

It would probably be better to gradually ramp up the duration to 1 hour instead of just jumping to it immediately, but considering we completely unenrolled before, this is still a net improvement. The threshold (seven consecutive unauthorized requests) that kicks us into this state is also very uncommon.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -356,17 +359,19 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
...
+		f.scheduler = scheduler.NewPeriodicJitter(defaultGatewaySettings.Duration, defaultGatewaySettings.Jitter)
```
cmacknz, line 372: When switching this out (both here and above) call the `Stop()` method first. It doesn't  matter now from what I can tell, but if it ever gets changed to say start a goroutine the goroutine might leak.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
cmacknz: I would suggest renaming `ErrDuration` to `ErrConsecutiveUnauthDuration` to be as specific as possible. We do not want all errors to have a 1 hour retry, only errors after 7 consecutive unauthorized errors.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
cmacknz: Possibly it is easier to adjust the duration up and down than to replace the whole scheduler.

internal/pkg/agent/application/gateway/fleet/fleet_gateway_test.go
```diff
@@ -564,3 +565,74 @@ func TestAgentStateToString(t *testing.T) {
 		})
 	}
 }
+
+func TestFleetGatewaySchedulerSwitch(t *testing.T) {
+	agentInfo := &testAgentInfo{}
+	settings := &fleetGatewaySettings{
+		Duration: 5 * time.Second,
```
cmacknz, line 572: This test takes 22s to run, lower this to speed it up.

```
go test ./internal/pkg/agent/application/gateway/... -run TestFleetGatewaySchedulerSwitch
ok      github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet   22.178s
```

internal/pkg/agent/application/gateway/fleet/fleet_gateway_test.go
```diff
@@ -564,3 +565,74 @@ func TestAgentStateToString(t *testing.T) {
...
+
+		for i := 0; i <= maxUnauthCounter; i++ {
+			<-clientWaitFn
+		}
+
+		cancel()
+		err := <-errCh
+		require.NoError(t, err)
+
+		_, ok = g.scheduler.(*scheduler.Periodic)
```
cmacknz, line 604: This only checks that the type of the scheduler changed not that the retry interval actually changed as expected. That is the behavior you want to test for.

cmacknz: This test would still pass if the retry duration got shorter for example.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
kaanyalti: if I were to adjust the duration instead of replacing the schedulers as you mentioned here https://github.com/elastic/elastic-agent/pull/6619#discussion_r1939939254, then I will not be able to call Stop as it closes the done channel. I'll still replace the schedulers, but will use jitter scheduler for both long and short wait times.

cmacknz: You don't need to call stop if you don't replace the scheduler IMO. Also, this is a preventative fix for a problem that doesn't quite exist yet.

I don't have a problem with replacing the schedulers fundamentally, there isn't a need to stick with that just because of this though if that's the case.

cmacknz (approved):

pierrehilbert:
@cmacknz I think we should also backport this in 9.0 but would like to get your opinion here.

pchila (approved):
LGTM
I agree with @cmacknz that we should progressively ramp up the polling interval up to the maximum duration of 1h but that can be done in a follow-up PR

kaanyalti:
Created this issue as a follow to add gradual ramp up to 1 hour 
https://github.com/elastic/elastic-agent/issues/6760

kalramani:
tagging the corresponding  elastic case  #01815174

Issue #6340, opened 2024-12-13
Deletion race condition when fetching secret from the Kubernetes secrets provider cache


cmacknz:
- Relates https://github.com/elastic/elastic-agent/issues/6187

The Kubernetes secrets provider caches secrets and updates them on a configurable schedule to avoid placing unnecessary load on the Kubernetes API.

The cache is updated in a goroutine: https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L88-L90

The cache is copied, the latest value for each non-expired secret is fetched, and then the copy is merged into the active map. https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L164-L199

The logic to get a value from the cache follows below.

https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L204-L225

1. A read lock on the cache is taken to check if the value exists in the cache.
2. If it isn't, the secret is fetched which involves holding a write lock again to update the cache.
3. A write lock on the cache is taken and the cache is updated with the value from step 2 and the last access time is set.

The race condition is that after 1 and 2 have completed, the `updateCache()` method that can delete secrets from the cache could have run and deleted the secret right before 3 below where the value is returned directly from the cache with no check for whether the secret still exists.

https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L218-L222

cmacknz:
There is another potential problem here, where time since last access is evaluated twice:

Once here where we refresh the value if the TTL hasn't expired, but the problem is if the value changed and agent should update the component model with the new value, we don't update the last access.

https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L174-L182

The TTL is then evaluated again when merging the maps which can lead to deletion even though the value updated:

[...]

PR #6623, opened 2025-01-29
[k8s] Fix logical race conditions in kubernetes_secrets provider

pkoutsovasilis:
## What does this PR do?

This PR refactors the implementation to eliminate logical race conditions of the `kubernetes_secrets` provider alongside with brand new unit-tests.

Initially, the issue seemed to stem from misuse or lack of synchronisation primitives, but after deeper analysis, it became evident that the "race" conditions were logical rather than concurrency-related. The existing implementation was structured in a way that led to inconsistencies due to overlapping responsibilities of different actors managing the secret lifecycle.

To address this, I restructured the logic while keeping in mind the constraints of the existing provider, specifically:

- **Using a Kubernetes reflector (watch-based mechanism) is not an option** because it would require listing and watching all **secrets**, which is often a non-starter for users.
- Instead, we must maintain our own caching mechanism that periodically refreshes only the referenced Kubernetes secrets.

With this in mind, the provider behaviour is now as follows:

#### Cache Disabled Mode:
- When caching is disabled, the provider simply reads secrets directly from the Kubernetes API server.

#### Cache Enabled Mode:
- When caching is enabled, the provider stores secrets in a cache where entries expire based on the configured **TTL** (time-to-live) and a **lastAccess** field of each cache entry.
- The provider has two primary actors: **cache actor** and **fetch actor**, each with well-defined responsibilities.

#### Cache Actor Responsibilities:
1. **Signal expiration of items**: When a secret expires, the cache actor signals that a fetch should occur to reinsert the key into the cache, ensuring continued refreshing.
2. **Detect secret updates and signal changes**: When the cache actor detects a secret value change, it signals the ContextProviderComm.
3. **Conditionally update `lastAccess`**:
   - If the secret has changed, update `lastAccess` to prevent premature expiration and give the fetch actor time to pick up the new value.
   - In any other case, do **not** update `lastAccess` and let the entry "age" as it should.

#### Fetch Actor Responsibilities:
1. **Retrieve secrets from the cache**:
   - If present, return the value.
   - If missing, fetch from the Kubernetes API.
2. **Insert fetched secrets into the cache** if there isn't a more recent version of the secret already in it (can happen by the cache actor or a parallel fetch actor).
3. **Always update `lastAccess`** when an entry is accessed to prevent unintended expiration.

#### Considerations:
- **No global locks**: Store operations are the only part of the critical path, ensuring that neither cache nor fetch actors block each other.
- **Conditional updates**: Since cache state can change between the time an actor reads and writes, all updates use **conditional store operations** that are part of the critical path.
- **Custom store implementation**: The existing `ExpirationCache` from `k8s.io/client-go/tools/cache` does not suit our needs, as it lacks the aforementioned conditional insertion required to handle these interactions correctly.
- **Optimized memory management**: The prior implementation copied the cache map on every update to prevent Golang map bucket retention. However, I believe this was a **misunderstanding of Golang internals** and a premature optimisation. If needed in the future, this can be revisited in a more controlled manner.

PS: as the main changes of this PR are captured by the commit https://github.com/elastic/elastic-agent/commit/a54972875c2c33e2978fee26f3a2a2f0d36115ab, I consider this PR to be aligned with the [Pull Requests policy](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#as-a-reviewee-ie-author-of-a-pr)

## Why is it important?

This refactor significantly improves the correctness of the `kubernetes_secrets` provider by ensuring:
- Secrets do not expire prematurely due to logical race conditions.
- Updates are properly signaled to consuming components.
- Performance is optimised with minimal locking and unnecessary memory allocations.

## Checklist

- [x] I have made corresponding changes to the default configuration files
- [x] I have added an entry in `./changelog/fragments` using the [change-log tool](https://github.com/elastic/elastic-agent#changelog)

## Disruptive User Impact

This change does **not** introduce breaking changes but ensures that the `kubernetes_secrets` provider operates correctly in cache-enabled mode. Users relying on cache behaviour may notice improved stability in secret retrieval.

1. Run unit tests to validate the new caching behaviour:
   ```sh
   go test ./internal/pkg/composable/providers/kubernetessecrets/...
   ```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6340


internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// expirationCache is a store that expires items after time.Now - secret.lastAccess > ttl (if ttl > 0) at Get or List.
+// expirationCache works with *cacheEntry, a pointer struct that wraps secret, instead of secret directly because map
+// structure in standard go library never removes the buckets from memory even after removing all the elements from it.
+// However, since *cacheEntry is a pointer it can be garbage collected when no longer referenced by the GC, such as
+// when deleted from the map. More importantly working with a pointer makes the entry in the map bucket, that doesn't
+// get deallocated, to utilise only 8 bytes on a 64-bit system.
+type expirationCache struct {
+	sync.RWMutex
```
michalpristas, line 19: we sure we need RWMutex, this does not add any performance gain and if write is frequent it can even result in a bad performance.
i'm leaning more towards normal Mutex, or we should benchmark to see benefits of having RWMutex.

michalpristas (approved):

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
pkoutsovasilis: the choice of RWMutex, was solely done on the expectation that this would more read-based as each secret value would be inserted once either during the refresh of the cache or during a fetch, and then everything will hit for the ttl. I mean having a cache should make us read-heavy otherwise on each write we would be also fetching from the API server, which kinda defeats the purpose? I can still switch to plain Mutex just let me know 🙂

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -21,96 +21,183 @@ import (
...
+	}
+
+	ctx := context.Background()
+
+	secretNamespace := tokens[1]
+	secretName := tokens[2]
+	secretKey := tokens[3]
+
+	// Wait for the provider to be initialized
+	<-p.running
```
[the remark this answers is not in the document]
pkoutsovasilis: `<-p.running` is only here to make the any fetch wait for the p.client to be initialised. We don't need to wait for the first iteration to finish because during the first iteration our cache would be empty, keys are inserted to the cache by Fetch. Does that make sense? 🙂

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -21,96 +21,183 @@ import (
...
 	}
-	p.clientMx.Lock()
+	p.clientMtx.Lock()
 	p.client = client
-	p.clientMx.Unlock()
+	p.clientMtx.Unlock()
 
 	if !p.config.DisableCache {
-		go p.updateSecrets(ctx, comm)
+		go p.refreshCache(ctx, comm)
```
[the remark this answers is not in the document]
pkoutsovasilis: Again, `<-p.running` is only to make any fetch wait for the `p.client` to be initialised. Now we do close the `p.running` after we invoke the `go p.refreshCache(ctx, comm)`, and during the first iteration cache will be empty so there is no extra safety added by closing it after it, right?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -21,96 +21,183 @@ import (
...
+		value:        apiSecretValue,
+		apiExists:    apiExists,
+		apiFetchTime: now,
+	}
+	p.store.AddConditionally(key, sd, true, func(existing secret, exists bool) bool {
+		if !exists {
+			// no existing secret in the cache thus add it
+			return true
+		}
+		if existing.value != apiSecretValue && !existing.apiFetchTime.After(now) {
```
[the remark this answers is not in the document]
pkoutsovasilis: `sd.apiFetchTime.After(existing.apiFetchTime)` if I go with that, and the time between sd and existing are the same (I know it is way too hard but can happen) then I am gonna lose a secret even if the value has changed. makes sense? 🙂

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
michalpristas: by the name i thought you wanted to prevent simultanious initialisation and fetching.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
michalpristas: then we also could reverse and do !before. but this is just a readability nit

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
michalpristas: i leave it up to you as RWMutex vs Mutex is a bit controversial topic

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
```diff
@@ -0,0 +1,124 @@
...
+		if c.isExpired(entry.lastAccess) {
+			continue
+		}
+		list = append(list, entry.s)
+	}
+	return list
+}
+
+// newExpirationStore creates and returns an expirationCache
+func newExpirationStore(ttl time.Duration) *expirationCache {
```
pchila, line 119: Is there a specific reason for calling the constructor for the `expirationCache` `newExpirationStore` ?
```suggestion
func newExpirationCache(ttl time.Duration) *expirationCache {
```

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
```diff
@@ -0,0 +1,124 @@
...
+	entry, exists := c.items[key]
+	if !exists {
+		return secret{}, false
+	}
+	if updateAccess {
+		entry.lastAccess = time.Now()
+	} else if c.isExpired(entry.lastAccess) {
+		delete(c.items, key)
+		return secret{}, false
+	}
```
pchila, line 46: This is a modification of the map while holding only a Read lock, this is not concurrent-safe.
There's also a possible concurrent access on `lastAccess` at the entry level.

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
```diff
@@ -0,0 +1,124 @@
...
+	items map[string]*cacheEntry
+}
+
+type cacheEntry struct {
+	s          secret
+	lastAccess time.Time
+}
+
+// Get returns the secret associated with the given key from the store if it exists and is not expired. If updateAccess is true
+// and the secret exists, essentially the expiration check is skipped and the lastAccess timestamp is updated to time.Now().
```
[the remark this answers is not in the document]
pkoutsovasilis: Thanks for the question 🙂 so the point of having such a `Get` function with such a flag `updateAccess` is to perform all operations at a single time point under the same critical path. If we split up the `Refresh` and `Get`  then we can't make a "correct", in terms of timing, decision as the entry might have changed between the Get and Refresh. Thus it needs to happen in a single go. Does it make sense? 🙂

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
pkoutsovasilis: I did rename it, please feel free to unresolve if you spot anything else 🙂

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
pchila: Makes sense to me. "premature optimization is the root of all evil" :wink:

internal/pkg/composable/providers/kubernetessecrets/store_test.go
pchila, line 1: If this is a generated mock (as implied by the header) why is this not reflected in the file name? (it should be `store_mock.go` or similar)
Moreover, at this time all the generated mocks are kept under `/mocks` directory, reflecting the original interface package structure (although this can easily be changed if needed).

Is this mock here because it's mocking an internal interface ?

pkoutsovasilis: yes that is indeed generated by mockery but I had to bring it manually inside the same package for the reason you already identified, private interface and private struct that we pass to the functions of the interface. I tried having both of these public and follow the standard practices.of mocks of this repo but then I got cyclic import errors thus that was not that helpful. Actually I was wondering why we generate our mocks in a different package than the actual interfaces 🤔

kaanyalti (approved):

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -21,96 +21,183 @@ import (
...
-
 // Run initializes the k8s secrets context provider.
-func (p *contextProviderK8sSecrets) Run(ctx context.Context, comm corecomp.ContextProviderComm) error {
+func (p *contextProviderK8SSecrets) Run(ctx context.Context, comm corecomp.ContextProviderComm) error {
 	client, err := getK8sClientFunc(p.config.KubeConfig, p.config.KubeClientOptions)
 	if err != nil {
-		p.logger.Debugf("kubernetes_secrets provider skipped, unable to connect: %s", err)
-		return nil
+		// signal that the provider has initialized
+		close(p.running)
```
swiatekm, line 105: Is this guard necessary? I thought it wasn't legal to call `Fetch` before `Run` returns.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -21,96 +21,183 @@ import (
...
 	return comm.Err()
 }
 
-func getK8sClient(kubeconfig string, opt kubernetes.KubeClientOptions) (k8sclient.Interface, error) {
-	return kubernetes.GetKubernetesClient(kubeconfig, opt)
+// Fetch returns the secret value for the given key
+func (p *contextProviderK8SSecrets) Fetch(key string) (string, bool) {
+	// Make sure the key has the expected format "kubernetes_secrets.somenamespace.somesecret.value"
+	tokens := strings.Split(key, ".")
+	if len(tokens) > 0 && tokens[0] != "kubernetes_secrets" {
```
swiatekm, line 132: We should make "kubernetes_secrets" a constant, and use that everywhere we use the string literal.

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
```diff
@@ -0,0 +1,124 @@
...
+func (c *expirationCache) Get(key string, updateAccess bool) (secret, bool) {
+	c.Lock()
+	defer c.Unlock()
+
+	entry, exists := c.items[key]
+	if !exists {
+		return secret{}, false
+	}
+	if updateAccess {
+		entry.lastAccess = time.Now()
```
swiatekm, line 42: What if we did the expiration check first, would that cause any problems? Intuitively, I'd expect expired entries to always be deleted after TTL, but here we extend that potentially indefinitely. My understanding is that the cache refresh routine doubles up as a garbage collector for expired entries, but I think the code would be clearer if that implicit dependency didn't exist. What do you think?

swiatekm (commented):
Overall this looks good to me, had some questions about the implementation details.

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
```diff
@@ -0,0 +1,124 @@
...
+		if condition != nil && condition(secret{}, false) {
+			c.items[key] = &cacheEntry{in, time.Now()}
+		}
+		return
+	}
+
+	if condition != nil && condition(entry.s, true) {
+		entry.s = in
+		entry.lastAccess = time.Now()
+	} else if updateAccess {
```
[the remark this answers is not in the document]
pkoutsovasilis: Yes this is intentional, because this function is intended to be used with an actual `condition`. Now because of semantics someone can call this function with a `nil` condition, thus with a nil condition is always false condition. So it is just a guarding here, do you see an issue with that?

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
pkoutsovasilis: >What if we did the expiration check first, would that cause any problems?

So imagine that the `Get` is happening from a `Fetch` (updateAccess = true in this case) where essentially we should *not* expire the entry and still keep the reference in the cache. If we did introduce the expiration before this won't be the case anymore. Thus I vote for the expiration to stay where it currently is. What do you think?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
pkoutsovasilis: In the happy path I see what you say and indeed `Fetch` can't happen before `Run` but if the `Run` is too slow, since this a detached goroutine if I am correct [here](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/composable/controller.go#L392), then a `Fetch` can happen in parallel, right?

pchila (approved):
LGTM

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
pkoutsovasilis: ok fair point const introduced! feel free to unresolve if you spot anything else 🙂

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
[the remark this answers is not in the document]
swiatekm: That's fine, could make a note in the docstring to be more explicit.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
swiatekm: Allright, makes sense. If this is how we want this to work, we should consider changing `Run` to be non-blocking so we don't need to defensively code in every provider. But that's material for a separate issue.

pkoutsovasilis:
@cmacknz since I am not sure that this PR should be categorised as a bug fix; please advice on the backport labels we should go with 🙂

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
swiatekm: Why shouldn't we expire the entry in Fetch if it's actually expired according to the TTL? That sounds like it'd make the logic easier to understand, and would at best remove a micro-optimization from the provider.

I'm also wondering if this wouldn't be clearer if we had separate methods for updates and non-updates. This cache essentially has two types of users:

* The fetcher, whose every action should result in a `lastAccess` update.
* The cache refresher, none of whose actions should result in a `lastAccess` update.

This might be easier to understand if they used different methods, even if the implementations were similar.

pkoutsovasilis: >Why shouldn't we expire the entry in Fetch if it's actually expired according to the TTL? That sounds like it'd make the logic easier to understand, and would at best remove a micro-optimization from the provider.

so the mechanism of TTL is for the cache to call signal and re-trigger the Fetch of each key that needs to be in the cache. What's the point of going all the way around this, since we are already calling Get() from a Fetch. 

> I'm also wondering if this wouldn't be clearer if we had separate methods for updates and non-updates. This cache essentially has two types of users:
> * The fetcher, whose every action should result in a lastAccess update.
> * The cache refresher, none of whose actions should result in a lastAccess update.
> This might be easier to understand if they used different methods, even if the implementations were similar.

isn't that exactly what I write in the description of PR, have you read it? 🤔 Essentially TTL needs to effective only to the cache actor. Now I see no value of splitting the current Get with updateAccess bool parameter into two different functions that one does Get with expiration and the other do a Get without it.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -21,96 +21,186 @@ import (
...
-func (p *contextProviderK8sSecrets) Run(ctx context.Context, comm corecomp.ContextProviderComm) error {
+func (p *contextProviderK8SSecrets) Run(ctx context.Context, comm corecomp.ContextProviderComm) error {
 	client, err := getK8sClientFunc(p.config.KubeConfig, p.config.KubeClientOptions)
 	if err != nil {
-		p.logger.Debugf("kubernetes_secrets provider skipped, unable to connect: %s", err)
-		return nil
+		// signal that the provider has initialized
+		close(p.running)
+		p.logger.Debug(k8sSecretsProviderName, " provider skipped, unable to connect: ", err.Error())
+		return err
```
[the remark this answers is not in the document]
cmacknz: Otherwise, it potentially makes sense to have Run implement the retry loop under the assumption we'll be able to create a client eventually.

cmacknz:
This looks great, and given it's an action item from a production incident, we should backport to the branches that allow it to be deployed to guard against those incidents occurring again. Backport to the oldest branch that infra observability will deploy and any newer branches.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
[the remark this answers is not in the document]
pkoutsovasilis: looking at this code [here](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/composable/controller.go#L451) we constantly retry to run a provider unless it returns an context.Canceled error. Thus I believe we are covered?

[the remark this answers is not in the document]
cmacknz: Yeah it looks like it retries periodically, if this function makes k8s API calls we may want an actual backoff algorithm here.

cmacknz (approved):

internal/pkg/composable/providers/kubernetessecrets/expiration_cache.go
swiatekm: > isn't that exactly what I write in the description of PR, have you read it? 🤔 Essentially TTL needs to effective only to the cache actor. Now I see no value of splitting the current Get with updateAccess bool parameter into two different functions that one does Get with expiration and the other does a Get without it.

Yeah, I wrote it out to make sure I understand correctly. My point is basically that when I see:

```
cache.Get(key, true)

```
it's less immediately clear to me than:

```
cache.GetAndUpdateAccessTime(key)
```

But that's a relatively minor thing in an out of itself. If you don't want to change it, that's fine.

> IMO, not quite right, it should update lastAccess on a secret change to provide enough grace period for the Fetch to happen and not expired if the cache refreshes sooner

True. What I meant was that the fetcher always sets `updateAccess=true` in its calls, while the refreshes always sets `updateAccess=false`. They're essentially separate use cases.
```
```

swiatekm (approved):

Issue #6637, opened 2025-01-30
DEB/RPM not defaulting to basic agent


michalpristas:
With https://github.com/elastic/elastic-agent/pull/6542 in RPM and DEB are not respecting behavior used during normal install.

Currently RPM/DEB installs agent with all components

Expected
- RPM/DEB installs basic agent by default
- installation respects value of ELATIC_AGENT_FLAVOR
- on upgrade
  - in case previous installlation without flavor -> copy all
  - in case previous installation with flavor -> respect flavor

jlind23:
Waiting for the release eng team to chime in here.
cc @michalpristas

PR #6638, opened 2025-01-30
[RPM/DEB] Changes to binary distribution 

michalpristas:
## What does this PR do?

Fixes behavior for RPM

 - RPM/DEB installs basic agent by default
 - installation respects value of ELATIC_AGENT_FLAVOR
 - on upgrade
   - in case previous installlation without flavor -> copy all
   - in case previous installation with flavor -> respect flavor

This takes a bit different approach as we cannot control what's being copied over during installations.
Before install we mark desired flavor into flavor file.
After install we execute new undocumented command `apply-flavor` that removes all unnecessary component from current version home.

## Why is it important?

Reducing space on disk by omitting not needed components.

## Checklist

##### install scenario
- compile deb
- run `dpkg -i elastic-agent-{rest of filename}.deb`
- `cat /var/lib/elastic-agent/.flavor` (expected basic)
- `ls -la /var/lib/elastic-agent/data/elastic-agent-*/componentes` expected limited set
- uninstall and purge
- run `ELASTIC_AGENT_FLAVOR=servers dpkg -i elastic-agent-{rest of filename}.deb`
- `cat /var/lib/elastic-agent/.flavor` (expected servers)
- `ls -la /var/lib/elastic-agent/data/elastic-agent-*/componentes` expected apm fleet-server present
- uninstall and purge

##### upgrade scenario
- compile deb
- install 8.17.2 snapshot `dpkg -i elastic-agent-{rest of filename}.deb`
- `cat /var/lib/elastic-agent/.flavor` (expected file not found)
- `ls -la /var/lib/elastic-agent/data/elastic-agent-*/componentes` expected whole set
- install 9.0.0 snapshot `dpkg -i elastic-agent-{rest of filename}.deb`
- `cat /var/lib/elastic-agent/.flavor` expected not found
- `ls -la /var/lib/elastic-agent/data/elastic-agent-*/componentes` expected whole set

- uninstall and purge
- compile additional deb with different hash
- install additional deb `ELASTIC_AGENT_FLAVOR=servers dpkg -i elastic-agent-{rest of filename}.deb`
- `cat /var/lib/elastic-agent/.flavor` (expected servers)
- `ls -la /var/lib/elastic-agent/data/elastic-agent-*/componentes` expected servers set
- install original build `dpkg -i elastic-agent-{rest of filename}.deb`
- `cat /var/lib/elastic-agent/.flavor` expected servers
- `ls -la /var/lib/elastic-agent/data/elastic-agent-*/componentes` expected servers set

same steps for rpm but use yum

Fixes: #6637


leehinman:
> Is an environment variable the canonical way to provide options to these package types? If we do something that doesn't follow the system convention we'll be explaining it forever.

An env var is **NOT** what you would expect.  For both RPM & DEB you would expect multiple RPM & DEB files that are built from the same source.  I would expect 2 RPM & DEB files, one that installs the minimal, and one that installs the extra components.  The extra components one would depend on the minimal.  Any upgrade scenario would upgrade both.


vim is a good example.

## RPM 

You can install `vim-minimal`, and have a working `vi`, here are the package contents.

```shell
$ rpm -ql vim-minimal
/etc/virc
/usr/bin/ex
/usr/bin/rvi
/usr/bin/rview
/usr/bin/vi
/usr/bin/view
/usr/lib/.build-id
/usr/lib/.build-id/5c
/usr/lib/.build-id/5c/ca9c420c7d69ce13f7cfe7ea0e7fdc7c18bc3f
/usr/libexec/vi
/usr/share/man/man1/ex.1.gz
/usr/share/man/man1/rvi.1.gz
/usr/share/man/man1/rview.1.gz
/usr/share/man/man1/vi.1.gz
/usr/share/man/man1/view.1.gz
/usr/share/man/man5/virc.5.gz
```

If you then want a more "full" vim experience you install `vim-enhanced` along side `vim-minimal` and you have:

``` shell
$ rpm -ql vim-enhanced
/usr/bin/rvim
/usr/bin/vim
/usr/bin/vimdiff
/usr/bin/vimtutor
/usr/lib/.build-id
/usr/lib/.build-id/cc
/usr/lib/.build-id/cc/b0b26d3af0feb447e3f786a6b7be4f79f854bd
```

But both of these are built from the same RPM source:

vim-minimal

``` shell
$ dnf info vim-minimal
Last metadata expiration check: 0:00:40 ago on Fri 31 Jan 2025 03:49:34 PM CST.
Installed Packages
Name         : vim-minimal
Epoch        : 2
Version      : 8.2.2637
Release      : 21.el9
Architecture : aarch64
Size         : 1.4 M
Source       : vim-8.2.2637-21.el9.src.rpm
Repository   : @System
From repo    : baseos
Summary      : A minimal version of the VIM editor
URL          : http://www.vim.org/
License      : Vim and MIT
Description  : VIM (VIsual editor iMproved) is an updated and improved version of the
             : vi editor.  Vi was the first real screen-based editor for UNIX, and is
             : still very popular.  VIM improves on vi by adding new features:
             : multiple windows, multi-level undo, block highlighting and more. The
             : vim-minimal package includes a minimal version of VIM, providing
             : the commands vi, view, ex, rvi, and rview. NOTE: The online help is
             : only available when the vim-common package is installed.
```			 

vim-enhanced

``` shell
$ dnf info vim-enhanced
Last metadata expiration check: 0:00:29 ago on Fri 31 Jan 2025 03:49:34 PM CST.
Available Packages
Name         : vim-enhanced
Epoch        : 2
Version      : 8.2.2637
Release      : 21.el9
Architecture : aarch64
Size         : 1.7 M
Source       : vim-8.2.2637-21.el9.src.rpm
Repository   : appstream
Summary      : A version of the VIM editor which includes recent enhancements
URL          : http://www.vim.org/
License      : Vim and MIT
Description  : VIM (VIsual editor iMproved) is an updated and improved version of the
             : vi editor.  Vi was the first real screen-based editor for UNIX, and is
             : still very popular.  VIM improves on vi by adding new features:
             : multiple windows, multi-level undo, block highlighting and more.  The
             : vim-enhanced package contains a version of VIM with extra, recently
             : introduced features like Python and Perl interpreters.
             :
             : Install the vim-enhanced package if you'd like to use a version of the
             : VIM editor which includes recently added enhancements like
             : interpreters for the Python and Perl scripting languages.  You'll also
             : need to install the vim-common package.
```

## Deb

You can install `vim-tiny` to get a minimal vi clone.

```shell
 dpkg -L vim-tiny
/.
/etc
/etc/vim
/etc/vim/vimrc.tiny
/usr
/usr/bin
/usr/bin/vim.tiny
/usr/share
/usr/share/bug
/usr/share/bug/vim-tiny
/usr/share/bug/vim-tiny/presubj
/usr/share/bug/vim-tiny/script
/usr/share/doc
/usr/share/doc/vim-tiny
/usr/share/doc/vim-tiny/changelog.Debian.gz
/usr/share/doc/vim-tiny/copyright
/usr/share/lintian
/usr/share/lintian/overrides
/usr/share/lintian/overrides/vim-tiny
/usr/share/vim
/usr/share/vim/vim91
/usr/share/vim/vim91/doc
/usr/share/vim/vim91/doc/README.Debian
/usr/share/vim/vim91/doc/help.txt
diverted by vim-runtime to: /usr/share/vim/vim91/doc/help.txt.vim-tiny
/usr/share/vim/vim91/doc/tags
diverted by vim-runtime to: /usr/share/vim/vim91/doc/tags.vim-tiny
```

If you want the full vim experience you would install `vim-nox` along side `vim-tiny`

```shell
$ dpkg -L vim-nox
/.
/usr
/usr/bin
/usr/bin/vim.nox
/usr/share
/usr/share/bug
/usr/share/bug/vim-nox
/usr/share/bug/vim-nox/presubj
/usr/share/bug/vim-nox/script
/usr/share/doc
/usr/share/doc/vim-nox
/usr/share/doc/vim-nox/changelog.Debian.gz
/usr/share/doc/vim-nox/copyright
/usr/share/lintian
/usr/share/lintian/overrides
/usr/share/lintian/overrides/vim-nox
```

Deb packages use the alternates system to set symlink so `vi` could point to either `/usr/bin/vim.tiny` or `/usr/bin/vim-nox`

But as you can see from the `apt info` for each, they are built from the same source.

vim-tiny

```shell
$ apt info vim-tiny
Package: vim-tiny
Version: 2:9.1.0016-1ubuntu7.6
Priority: important
Section: editors
Source: vim
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian Vim Maintainers <team+vim@tracker.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 1,850 kB
Provides: editor
Depends: vim-common (= 2:9.1.0016-1ubuntu7.6), libacl1 (>= 2.2.23), libc6 (>= 2.34), libselinux1 (>= 3.1~), libtinfo6 (>= 6)
Suggests: indent
Homepage: https://www.vim.org/
Task: minimal
Download-Size: 770 kB
APT-Manual-Installed: no
APT-Sources: http://ports.ubuntu.com/ubuntu-ports noble-updates/main arm64 Packages
Description: Vi IMproved - enhanced vi editor - compact version
 Vim is an almost compatible version of the UNIX editor Vi.
 .
 This package contains a minimal version of Vim compiled with no GUI and
 a small subset of features. This package's sole purpose is to provide
 the vi binary for base installations.
 .
 If a vim binary is wanted, try one of the following more featureful
 packages: vim, vim-nox, vim-motif, or vim-gtk3.
```

vim-nox

```shell
$ apt info vim-nox
Package: vim-nox
Version: 2:9.1.0016-1ubuntu7.6
Priority: extra
Section: universe/editors
Source: vim
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian Vim Maintainers <team+vim@tracker.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 4,950 kB
Provides: editor, vim (= 2:9.1.0016-1ubuntu7.6), vim-lua (= 2:9.1.0016-1ubuntu7.6), vim-perl (= 2:9.1.0016-1ubuntu7.6), vim-python3 (= 2:9.1.0016-1ubuntu7.6), vim-ruby (= 2:9.1.0016-1ubuntu7.6), vim-tcl (= 2:9.1.0016-1ubuntu7.6)
Depends: vim-common (= 2:9.1.0016-1ubuntu7.6), vim-runtime (= 2:9.1.0016-1ubuntu7.6), libacl1 (>= 2.2.23), libc6 (>= 2.38), libgpm2 (>= 1.20.7), liblua5.1-0, libperl5.38t64 (>= 5.38.2), libpython3.12t64 (>= 3.12.1), libruby3.2 (>= 3.2.2), libselinux1 (>= 3.1~), libsodium23 (>= 1.0.14), libtcl8.6 (>= 8.6.0), libtinfo6 (>= 6)
Suggests: cscope, vim-doc
Homepage: https://www.vim.org/
Download-Size: 2,040 kB
APT-Manual-Installed: yes
APT-Sources: http://ports.ubuntu.com/ubuntu-ports noble-updates/universe arm64 Packages
Description: Vi IMproved - enhanced vi editor - with scripting languages support
 Vim is an almost compatible version of the UNIX editor Vi.
 .
 Many new features have been added: multi level undo, syntax
 highlighting, command line history, on-line help, filename
 completion, block operations, folding, Unicode support, etc.
 .
 This package contains a version of vim compiled with support for
 scripting with Lua, Perl, Python 3, Ruby, and Tcl but no GUI.
```

blakerouse:
I agree with @leehinman that for linux distributions it is done with multiple deps/rpms. I normally see things like:

- `elastic-agent-base` - only the Elastic Agent (no components)
- `elastic-agent-components-basic` - all basic components
- `elastic-agent-components-server` - all server components
- `elastic-agent` - (meta package) that depends on `elastic-agent-base` and `elastic-agent-component-basic`
- `elastic-agent-server` - (meta package) that depends on `elastic-agent-base`, `elastic-agent-components-basic`, and `elastic-agent-components-server`

The issue with this type of design, even if there is two packages and not as many as I said above, is that it no longer becomes a single package to download and install. A better install path for that is to add it to a APT source or a DPK source then allow it easy to install and have it install the dependencies.

I think the question here is, if this design (even if more correct) is how we want to require our users to install Elastic Agent from DEB/RPM packages.

leehinman:
Might just be me but when I did `ELASTIC_AGENT_FLAVOR=servers dpkg -i elastic-agent-9.0.0-arm64.deb` the `.flavor` file did indeed have "servers", but none of the extra components were installed.

One thing I don't like about this approach is that the DEB and RPM file databases will be out of sync with what is on disk.  Running `elastic-agent apply-flavor` will change what is on disk, but it doesn't update the DEB or RPM databases.  This means if anyone is using those databases for things like "making sure the binaries on disk are the ones RPM/DEB installed" it will fail.

michalpristas:
thanks for your insights. 
if that's ok i will then create base images like we discussed with docker images that represent either full or base version of agent. to avoid breaking changes and staying consistent in between types i'd rather add additional base to current all. behavior would stay the same except flavor file is now part of the rpm/deb package

cc @cmacknz

cmacknz:
Considering the timing, and the potential inconsistency between RPM/DEB/MIS and plain `elastic-agent install` if we don't change this before 9.0 FF, the decision is to keep the approach of `ELATIC_AGENT_FLAVOR`. It isn't the canonical way to do this, but we'd rather drift from the system convention than end up with our installation methods not being in a consistent state in 9.0.

We could still introduce meta packages later if we think they result in a better experience here.

internal/pkg/agent/install/.flavors
```diff
@@ -0,0 +1,13 @@
+basic:
```
cmacknz, line 1: This is duplicating https://github.com/elastic/elastic-agent/blob/main/_meta/.flavors

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -43,4 +44,19 @@ if test -L "$symlink"; then
     fi
 else
     echo "no previous installation found"
+
+    # create dir in case it does not exist
+    mkdir -p "$new_agent_dir" 
+    
+    # 2 is upgrade for Fedora, do not upgrade file when upgrading and file exists
+    if [[ "$1" != "2" ]]; then
+        if [[ -n "${ELASTIC_AGENT_FLAVOR}" ]]; then
```
cmacknz, line 53: There should be a test proving this does what it is supposed to do.

We already have RPM and DEB tests:
- https://github.com/elastic/elastic-agent/blob/main/testing/integration/linux_rpm_test.go
- https://github.com/elastic/elastic-agent/blob/main/testing/integration/linux_deb_test.go

You should be able to modify those to prove the default installation type changed and is maintained as expected. 

If you pushed understanding the ELASTIC_AGENT_FLAVOR variable into the `apply-flavor` command you could test a lot of this independent of the package type.

One way is having a `set-flavor` command and passing the env var value to it (or just having an argument to read it from the environment you unconditionally set here). 

Then you'd just need to test that `set-flavor` and `apply-flavor` chain together properly, which basic smoke testing that the behavior of the RPM and DEB changed appropriate as a result.

internal/pkg/agent/install/flavors.go
```diff
@@ -48,7 +48,7 @@ func UsedFlavor(topPath, defaultFlavor string) (string, error) {
 		return "", fmt.Errorf("failed reading flavor marker file: %w", err)
 	}
 
-	return string(content), nil
+	return strings.TrimSpace(string(content)), nil
```
cmacknz, line 51: What bug is this fixing? It's not clear there's a test for it.

internal/pkg/agent/install/flavors.go
```diff
@@ -118,7 +118,7 @@ func ApplyFlavor(versionedHome string, flavor FlavorDefinition) error {
 	}
 
 	for _, ftr := range filesToRemove {
-		if removeErr := os.RemoveAll(ftr); !os.IsNotExist(removeErr) {
+		if removeErr := os.RemoveAll(ftr); removeErr != nil && !os.IsNotExist(removeErr) {
```
cmacknz, line 121: Same as above.

What bug is this fixing? It's not clear there's a test for it.

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
cmacknz: We would also want to make sure that an RPM and DEB upgrade do not reset this accidentally.

internal/pkg/agent/install/flavors.go
michalpristas: this is path that was not hit before. in case removeErr is nil it will still go inside this condition

internal/pkg/agent/install/flavors.go
michalpristas: i've seen weird formatting in some test, i do not recall which one

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
michalpristas: what kind of commands do you have in mind, elastic agent commands?

cmacknz: Yes, `apply-flavor` already is a command. Thinking more, these commands only make sense if you start from an unflavored agent as it can't add binaries back.

So my suggestion would be to have `apply-flavor` take the flavor to apply as an argument and encapsulate updating the file into it (or you could add a `--env` argument into it to push the env var there as well).

cmacknz: I am much more interested in getting correct test coverage than I am about the exact design that leads to that happening at this point.

Doing more in apply-flavor would let you test outside of the RPM and DEB package but if it's easier to just test those directly do that instead.

michalpristas: added tests

internal/pkg/agent/cmd/apply_flavor.go
```diff
@@ -0,0 +1,59 @@
...
+func applyCmd() error {
+	topPath := paths.Top()
+	detectedFlavor, err := install.UsedFlavor(topPath, "")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return err
+	}
+	if detectedFlavor == "" {
```
cmacknz, line 47: Same here, we should log that we detected no flavor and did nothing.

In general this function should log what it did, neither install.Flavor or install.ApplyFlavor seem to log anything either. We can't tell "successfully changed flavor" from "did nothing because the flavor was already applied" for example.

testing/integration/linux_deb_test.go
```diff
@@ -49,6 +50,75 @@ func TestDebLogIngestFleetManaged(t *testing.T) {
 	agentFixture, err := define.NewFixtureFromLocalBuild(t, define.Version(), atesting.WithPackageFormat("deb"))
 	require.NoError(t, err)
 
+	installOpts := atesting.InstallOpts{
+		NonInteractive: true,
+		Force:          true,
+		InstallServers: false,
+	}
+
+	testDebLogIngestFleetManagedWithCheck(ctx, t, agentFixture, info, installOpts,
```
cmacknz, line 59: This check should go at the start and end of the upgrade test as well, perhaps just the end since we know from this test that installation behaves properly.

The same should apply to the RPM upgrade variant.

cmacknz:
I am testing the deb manually and it can't seem to get it to install the servers variant.

```sh
ELASTIC_AGENT_FLAVOR=servers sudo dpkg -i /tmp/elastic-agent-9.1.0-SNAPSHOT-arm64.deb
Selecting previously unselected package elastic-agent.
(Reading database ... 77355 files and directories currently installed.)
Preparing to unpack .../elastic-agent-9.1.0-SNAPSHOT-arm64.deb ...
no previous installation found
Unpacking elastic-agent (9.1.0) ...
Setting up elastic-agent (9.1.0) ...
create symlink /usr/share/elastic-agent/bin/elastic-agent to /var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe/elastic-agent
systemd enable/restart elastic-agent

ubuntu@feminine-oarfish:~$ ls /var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe/components/
agentbeat           endpoint-security                lenses                 osqueryd
agentbeat.spec.yml  endpoint-security-resources.zip  module                 pf-host-agent
certs               endpoint-security.spec.yml       osquery-extension.ext  pf-host-agent.spec.yml

ubuntu@feminine-oarfish:~$ cat /var/lib/elastic-agent/.flavor
basic
```


Switching to `export ELASTIC_AGENT_FLAVOR=servers` doesn't change anything either.

There is no logging about why it is doing what it is doing. The first time I ran this, I had a typo in the env var and it installed the basic version with no indication that's what happened. This is a bit of a limitation for env vars but we can at least log what env var value we found.

I've also noticed that you need `dpkg -P` to remove the flavors file because `dpkg -r` leaves it there. This will have to be clearly documented.

cmacknz:
OK another documentation update, `sudo ELASTIC_AGENT_FLAVOR=servers` instead of `ELASTIC_AGENT_FLAVOR=servers sudo` otherwise `sudo` discards env vars which is a mistake I have had in the past.  This is a suggestion we could put right in the logging.

cmacknz:
The original changelog in https://github.com/elastic/elastic-agent/blob/main/changelog/fragments/1737131552-Changes-to-binary-distribution.yaml seems like it would work for this, though it doesn't actually specify how to choose flavors so we should edit it to specify this.

cmacknz:
DEB upgrade works:

```
sudo apt-get install -y -qq -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold ./elastic-agent-9.2.0-SNAPSHOT-arm64.deb
(Reading database ... 77795 files and directories currently installed.)
Preparing to unpack .../elastic-agent-9.2.0-SNAPSHOT-arm64.deb ...
previous installation directory /var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe
migrate state from /var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe to /var/lib/elastic-agent/data/elastic-agent-9.2.0-SNAPSHOT-3fb4fe
didn't find /var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe/state.yml
didn't find /var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe/state.enc
Unpacking elastic-agent (9.2.0) over (9.1.0) ...
stopping elastic-agent
Warning: The unit file, source configuration file or drop-ins of elastic-agent.service changed on disk. Run 'systemctl daemon-reload' to reload units.
found symlink /usr/share/elastic-agent/bin/elastic-agent, unlink
systemd daemon-reload
dpkg: warning: unable to delete old directory '/var/lib/elastic-agent/data/elastic-agent-9.1.0-SNAPSHOT-3fb4fe': Directory not empty
Setting up elastic-agent (9.2.0) ...
create symlink /usr/share/elastic-agent/bin/elastic-agent to /var/lib/elastic-agent/data/elastic-agent-9.2.0-SNAPSHOT-3fb4fe/elastic-agent
systemd enable/restart elastic-agent
Scanning processes...
Scanning linux images...

Running kernel seems to be up-to-date.

No services need to be restarted.

No containers need to be restarted.

No user sessions are running outdated binaries.

No VM guests are running outdated hypervisor (qemu) binaries on this host.
```

michalpristas:
> OK another documentation update, `sudo ELASTIC_AGENT_FLAVOR=servers` instead of `ELASTIC_AGENT_FLAVOR=servers sudo` otherwise `sudo` discards env vars which is a mistake I have had in the past. This is a suggestion we could put right in the logging.

you can also use `ELASTIC_AGENT_FLAVOR=servers sudo -E` as used in tests. this will preserve env variables.

blakerouse (approved):
Code looks good, install tests looks good. Please create a follow up to add more to the upgrade test to verify that the flavor is used on upgrade (that is what is missing).

internal/pkg/agent/cmd/apply_flavor.go
```diff
@@ -0,0 +1,59 @@
...
+
+	return cmd
+}
+
+func applyCmd() error {
+	topPath := paths.Top()
+	detectedFlavor, err := install.UsedFlavor(topPath, "")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
```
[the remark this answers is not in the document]
michalpristas: followup

testing/integration/linux_deb_test.go
michalpristas: followup

PR #6662, opened 2025-01-30
[CI] Fix throttled windows test failures

pazone:
## What does this PR do?
In some cases windows tests failures were throttled and the step was false-passing

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/integration-tests.ps1
+++ b/.buildkite/scripts/integration-tests.ps1
@@ -38,6 +38,9 @@ try {
     Get-Ess-Stack -StackVersion $PACKAGE_VERSION
     Write-Output "~~~ Running integration test group: $GROUP_NAME as user: $env:USERNAME"
     gotestsum --no-color -f standard-quiet --junitfile "${outputXML}" --jsonfile "${outputJSON}" -- -tags=integration -shuffle=on -timeout=2h0m0s "github.com/elastic/elastic-agent/testing/integration" -v -args "-integration.groups=$GROUP_NAME" "-integration.sudo=$TEST_SUDO"
+} catch {
+    Write-Output "There are test failures: $_"
+    throw
 } finally {
     ess_down
     


.buildkite/scripts/integration-tests.ps1
```diff
@@ -49,3 +51,7 @@ try {
         Write-Output "Cannot generate HTML test report: $outputXML not found"
     }
 }
+
+if ($TestsExitCode -ne 0) {
```
dwhyrock, line 55: It's probably worth adding checks like this to all of your external commands.  (Annoying i know).  Otherwise it's like you're doing a `set +x` before each command and ignoring the results.

You have commands on lines 12, 13, and 28 in here that may or may not fail without you properly knowing.

dwhyrock (commented):
A few comments about properly catching the errors, and properly calling commands.

dwhyrock:
I'll put this link I also put in a slack thread here:

https://github.com/elastic/endpoint-dev/blob/main/.buildkite/scripts/common/base.ps1

We created some powershell functions for common commands that we call in CI.  We do the check for the return code in the function.  It may be something to look into, as it can also clean up the code where it's called.

For example, we have this function:

```
function Fn-Endpoint-Pip-Install {
    & python -m pip install wheel
    if ( $LASTEXITCODE -ne 0 ) {
        throw "Unable to install wheel"
    }
    & python -m pip install -r requirements.txt
    if ( $LASTEXITCODE -ne 0 ) {
        throw "Unable to install pip requirements.txt"
    }
}
```

And in the CI script where we call it, it simply looks like this (with a preceding `echo` statement):

```
echo "--- Install pip requirement"
Fn-Endpoint-Pip-Install
```

.buildkite/scripts/integration-tests.ps1
```diff
@@ -26,6 +26,10 @@ $env:SNAPSHOT = $true
 
 echo "~~~ Building test binaries"
 mage build:testBinaries
```
dwhyrock, line 28:
```suggestion
& mage build:testBinaries
```
Helps be consistent and makes sure it doesn't happen to call some cmdlet named the same thing.

dwhyrock (approved):

pkoutsovasilis (commented):
LGTM in the sense that I can see windows failures now in the CI, not an expert in powershell though 🙂

jlind23 (approved):

v1v:
@mergify backport 9.0

PR #6665, opened 2025-01-31
Update branch filters to include/exclude 9.*

dliappis:
## What does this PR do?

This commit updates branch_configuration and branch filters in CI pipeline definitions to support the new 9.* branches.

## Why is it important?

Enables automated runs of important CI jobs like DRA on new 9.* branches

Can only tested in BK after merge.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -37,7 +37,7 @@ spec:
       name: elastic-agent
       description: Buildkite pipeline for the Elastic Agent project
     spec:
-      branch_configuration: "main 7.* 8.* 9.* v7.* x8.* v9.*"
+      branch_configuration: "main 7.* 8.* 9.* v7.* v8.* v9.*"
       repository: elastic/elastic-agent
       pipeline_file: ".buildkite/pipeline.yml"
       provider_settings:
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -37,7 +37,7 @@ spec:
       name: elastic-agent
       description: Buildkite pipeline for the Elastic Agent project
     spec:
-      branch_configuration: "main 7.* 8.* v7.* v8.*"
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       name: elastic-agent
       description: Buildkite pipeline for the Elastic Agent project
     spec:
-      branch_configuration: "main 7.* 8.* v7.* v8.*"
+      branch_configuration: "main 7.* 8.* 9.* v7.* x8.* v9.*"
```
v1v, line 40: What's the reason for `v[0-9].*`? IIRC, we don't have branches with that naming

```suggestion
      branch_configuration: "main 7.* 8.* 9.* v7.* v8.* v9.*"
```

pchila (approved):
LGTM

alexsapran (approved):

catalog-info.yaml
dliappis: Great question. I was wondering why we even have v7* / v8* so just copied over the same pattern. IIRC @mrodm added those quite some time ago, not sure if there was context he remembers?

dliappis: @pazone do you remember what was the reason for using `v7*`/`v8*` historically for branch filters (in addition to the usual `7.*/8.*` etc) in elastic-agent pipeline configs?

mrodm: @v1v @dliappis IIRC this branch configuration setting is also applied for both branches and tags that Buildkite is going to trigger a build.

If not set, I think it would trigger a build for every tag pushed to the repository (upstream).

For instance , in `package-spec` these are the values 
image

dliappis: Thanks @mrodm for the clarification. In this case it makes sense to keep this pattern, as this PR does.

v1v (approved):

Issue #6209, opened 2024-12-04
Injected proxy variables into endpoint elasticsearch does not include NO_PROXY variable


janniten:
Hello,
We did an upgrade from 8.14.2 to 8.16.1.
In our environment the agent has the HTTP_PROXY, HTTPS-PROXY and NO_PROXY configured.
NO_PROXY variable **is not being injected into endpoint** . 
As a result, the elastic endpoint is trying to connect the ES cluster Via proxy, which is not correct.

Thank you

- Version: 8.16.1
- Operating System: Windows / linux
- Discuss Forum URL: https://elasticstack.slack.com/archives/C02J2JBS0FP/p1733218957317399
- Steps to Reproduce: Upgrade from 8.14.2 to 8.16.1

michel-laterman:
HTTP_PROXY/HTTPS_PROXY injection was added with: https://github.com/elastic/elastic-agent/pull/5044
We'll probably need to alter the `injectProxyURL` func to add `NO_PROXY`.

@intxgo, would the expected value in the config be `no_proxy: string`, or `no_proxy: []string`?

intxgo:
Since the env vars are injected as a regular output proxy settings, there's no place for NO_PROXY (endpoint won't recognize it, ignore it). However since the output URL is known, I think the NO_PROXY can be matched against the output URL at Agent side before deciding to inject HTTP_PROXY/HTTPS_PROXY variable to the output config.

ManojPraveeen:
Hi Team,

The issue is related to case 01791476. From what I understand from the @intxgo  statement, you’re suggesting including these proxy configurations directly in the agent's environment settings, right?

It would look something like the configuration below: 
   ```yaml
   proxy_settings:
     http_proxy: 
     https_proxy:
     no_proxy: 
   ```
Adding the above parameters to the configuration should fix the issue, correct me if I’m wrong.

intxgo:
@ManojPraveeen it's more complicated. Since quite some time we're not forwarding the config between components anymore, instead we have a GRPC base communication, see demo client https://github.com/elastic/elastic-agent-client

[...]

PR #6675, opened 2025-01-31
resolve proxy to inject using http package

intxgo:
## What does this PR do?

Agent is injecting proxy env variables into components configs if no proxy is already specified by the config. Unfortunately with this approach the NO_PROXY variable is lost, so the target URL has to be evaluated against the environment proxy to determine if the proxy should be used. Go [http](https://pkg.go.dev/net/http#ProxyFromEnvironment) provides such functionality.

## Why is it important?

The environment proxy may be defined with exclusion for Fleet and Output addresses.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6209
- Relates https://github.com/elastic/elastic-agent/issues/2602


internal/pkg/agent/application/inject_proxy_component_modifier.go
```diff
@@ -79,28 +77,20 @@ func injectProxyURL(m map[string]interface{}, hosts []string) {
...
+		if err != nil {
+			continue
 		}
-	}
-	// if no hosts are specified, or it could not match a host prefix prefer HTTPS_PROXY over HTTP_PROXY
-	if proxyURL == "" && !matched {
-		proxyURL = os.Getenv("HTTPS_PROXY")
-		if proxyURL == "" {
-			proxyURL = os.Getenv("HTTP_PROXY")
+		proxyURL, err := http.ProxyFromEnvironment(request)
```
michel-laterman, line 87: 👍 didn't know this existed!

michel-laterman (approved):

intxgo:
I probably found the culprit of the unit tests failure. The `t.Setenv()` function sets the actual environment variables on the executing process and clears them after test. It's documented as unsafe for parallel test execution.

Indeed, even locally when I run all tests I get the same failure. Experimenting I've found that environment proxy set in the shell before test execution is not overridden by `t.Setenv()`, so if I set them to values expected by my test it PASS even when invoked with all tests. 

```
export NO_PROXY="do.not.inject.proxy.for.me"
export HTTPS_PROXY="https://localhost:8080"
export HTTP_PROXY="https://localhost:8081"
```

Could it be that these env variables are set on the testing nodes but with empty strings?

@cmacknz @michel-laterman do you know if we could set those env variables on the test nodes like above? If yes, how to do this?

cmacknz:
The `t.SetEnv` call is used elsewhere in agent without issue and used to work here without issue as well. I suspect it by itself isn't the problem, the parallel warning is just that env vars are global. This is no different that calling `os.Setenv` I think. This is the only test setting one of the _PROXY vars so execution should be sequential here.

Something else is the problem, I can't reproduce the failure locally, so I suspect you are onto something with the CI environment being somehow different in a way that matters. I don't see us setting any of the proxy environment variables in the buildkite jobs and I can't see it in any of the buildkite output which is curious.

cmacknz:
OK so the problem isn't `t.Setenv` not being safe it's that the switch to `http.ProxyFromEnvironment(request)` I think.

It includes a do.Once to setup the proxy func and I think this only reads the env vars one time. https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/net/http/transport.go;l=907-914

This has something to do with at which point in the tests the variables are read I think. 

`go test ./internal/pkg/agent/application/... -run Test_injectProxyURL` passes.

`go test ./internal/pkg/agent/application/...` fails which is closer to what CI does.

intxgo:
> It includes a do.Once to setup the proxy func and I think this only reads the env vars one time. https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/net/http/transport.go;l=907-914

Thanks for checking for the root cause. 

Do you have a suggestion then how to handle it in tests? Obviously this works correctly in the product as Agent process will just see environment it was tarted with. Failure to write a test doesn't justify keeping a bug, but I'm out of ideas. (I don't think re-implementing this function is the right approach).

internal/pkg/agent/application/inject_proxy_component_modifier.go
cmacknz: I think this function in its current form is the problem in the tests. If you look at the implementation it is just calling https://cs.opensource.google/go/go/+/master:src/net/http/transport.go?q=FromEnvironment&ss=go%2Fgo

```go
func ProxyFromEnvironment(req *Request) (*url.URL, error) {
	return envProxyFunc()(req.URL)
}


// envProxyFunc returns a function that reads the
// environment variable to determine the proxy address.
func envProxyFunc() func(*url.URL) (*url.URL, error) {
	envProxyOnce.Do(func() {
		envProxyFuncValue = httpproxy.FromEnvironment().ProxyFunc()
	})
	return envProxyFuncValue
}
```

If you directly get a reference to httpproxy.FromEnvironment().ProxyFunc() yourself I think the problem goes away as you skip the `do.Once`. The following fixes the tests. You could maybe be smarter and not re-read from the environment on every configuration change if you wanted to, I'm not sure that efficiency matters here.

```
diff --git a/internal/pkg/agent/application/inject_proxy_component_modifier.go b/internal/pkg/agent/application/inject_proxy_component_modifier.go
index d8eb9f1e98..2f68e7452c 100644
--- a/internal/pkg/agent/application/inject_proxy_component_modifier.go
+++ b/internal/pkg/agent/application/inject_proxy_component_modifier.go
@@ -10,6 +10,7 @@ import (
        "github.com/elastic/elastic-agent-client/v7/pkg/client"
        "github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator"
        "github.com/elastic/elastic-agent/pkg/component"
+       "golang.org/x/net/http/httpproxy"
 )

 // InjectProxyEndpointModifier injects a proxy_url value into endpoint's output config if one is not set.
@@ -84,7 +85,7 @@ func injectProxyURL(m map[string]interface{}, hosts []string) {
                if err != nil {
                        continue
                }
-               proxyURL, err := http.ProxyFromEnvironment(request)
+               proxyURL, err := httpproxy.FromEnvironment().ProxyFunc()(request.URL)
                if err != nil {
                        continue
                }
```

intxgo: Thanks. This will theoretically departure a tiny bit from the desired goal to have Endpoint behaving exactly as Agent, because Agent will read the environment only once for it's http operations. However since the environment on a process can be only modified from within the process, by debugger, etc, but not just by modifying system configuration LGTM.

intxgo: updated 894f882d3b8c1a7391891a4c2c02ab86fb7d2dda

cmacknz (approved):

kaanyalti (approved):

PR #6681, opened 2025-01-31
otel: add test for document equivalence between agent filestream input and filebeat receiver

mauri870:
## What does this PR do?

This PR adds a test that compares a log line ingested by a filestream agent input and a filebeat OTel receiver with equivalent configuration.

## Checklist

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/4736
- For https://github.com/elastic/ingest-dev/issues/3676


testing/integration/otel_test.go
```diff
@@ -1099,3 +1102,265 @@ service:
...
+	fbDoc := fbDocs.Hits.Hits[0].Source
+	fbReceiverDoc := fbReceiverDocs.Hits.Hits[0].Source
+	ignoredFields := []string{
+		// Expected to change between filebeat and fbreceiver
+		"@timestamp",
+		"agent.ephemeral_id",
+		"agent.id",
+
+		// Missing from fbreceiver doc
+		"data_stream.dataset",
```
mauri870, line 1322: I'm a bit concerned by all these fields that are missing. I assume these are added by the agent itself. Still, I'm not sure exactly what this means for the document equivalence.

leehinman: This is because we haven't done https://github.com/elastic/ingest-dev/issues/4893 yet.  Your fbreceiver config above is missing processors.  Take a look at https://github.com/elastic/beats/blob/76f40863c9918ee1a691a2bb29cbab98a5d83878/x-pack/libbeat/management/generate.go#L225. That will give you some idea of the processors that are "magically" added to the config when it comes from elastic-agent.   Once you figure out what processors need to be added, can you add the info to https://github.com/elastic/ingest-dev/issues/4890 that will make it much easier when we get there.

leehinman (commented):
I think if you add the default processors listed at https://github.com/elastic/beats/blob/76f40863c9918ee1a691a2bb29cbab98a5d83878/x-pack/filebeat/cmd/root.go#L48 and add the processors listed in the generate.go below you should get a much shorter list of fields to ignore.

testing/integration/otel_test.go
mauri870: Got it, thanks! I’ll check the processors and update https://github.com/elastic/ingest-dev/issues/4890 once I have a working set of processors.

mauri870: So I managed to add some `add_*` processors and the add_host_metadata made all the host.* related fields match.

Now I have this list:

```
		"data_stream.dataset",
		"data_stream.namespace",
		"data_stream.type",
		"elastic_agent.id",
		"elastic_agent.snapshot",
		"elastic_agent.version",
		"event.dataset",
```

Is my understanding correct that these fields cannot be added via the test config (e.g., using add_fields processor) and instead need to be added to the otelconsumer/fbreceiver code at runtime, much like what Mikolaj is doing at https://github.com/elastic/elastic-agent/pull/6697?

khushijain21: Here, you can add it like this. The fields use default values here when not set.

```
        - add_fields:
            fields:
                dataset: elastic_agent
                namespace: default
                type: logs
            target: data_stream
        - add_fields:
            fields:
                dataset: elastic_agent
            target: event
``` 

This would also mean that all events from this input will be available in <type>-<dataset>-<namespace> index. And you may want to query THIS index for comparing the documents further in the code. 

If this is not very important, you may ignore this fields I believe. @Lee can confirm

mauri870: Huh, I thought of just add the missing add_fields with the same values as the original doc ingested by filebeat+agent but it seemed like a hack of sorts 🤔

khushijain21 (approved):

testing/integration/otel_test.go
khushijain21: you may also have to ignore cloud related fields

```
+- 	"cloud.account.id":        string("elastic-ci-prod")
+- 	"cloud.availability_zone": string("us-central1-a"),
+- 	"cloud.instance.id":       string("437430177250243523"),
+- 	"cloud.instance.name":     string("bk-agent-prod-gcp-1738853675336047652"),
+- 	"cloud.machine.type":      string("n1-standard-8"),
+- 	"cloud.project.id":        string("elastic-ci-prod"),
+- 	"cloud.provider":          string("gcp"),
+- 	"cloud.region":            string("us-central1"),
+- 	"cloud.service.name":      string("GCE"),
```

pierrehilbert:
I don't think the failing tests are related to this PR:
```
=== RUN   TestDownloadFile
    utils_test.go:21: 
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-aws-1738851441003538166/elastic/elastic-agent/dev-tools/mage/downloads/utils_test.go:21
        	Error:      	Expected nil, but got: &fmt.wrapError{msg:"downloading file https://www.elastic.co/robots.txt: GET request failed with 504", err:(*errors.errorString)(0xc00085a970)}
        	Test:       	TestDownloadFile
--- FAIL: TestDownloadFile (20.57s)
```
Will update the branch to have them running again and hope the gateway failure was temporary

testing/integration/otel_test.go
khushijain21: All `host` and `cloud` related fields should have been set by the default processors [here](https://github.com/elastic/beats/blob/95b5dce16c07df1c2f1aa88dc4aa507441504e78/x-pack/filebeat/fbreceiver/factory.go#L40).

Created an issue for the same https://github.com/elastic/beats/issues/42631.

mauri870: Thanks, the cloud fields missing are new to me, I think is because I'm running the tests locally before pushing.

khushijain21 (approved):

leehinman (approved):
LGTM

PR #6697, opened 2025-02-03
Add ability to run components in the Otel manager

swiatekm:
## What does this PR do?

Allows to run components in the Otel manager instead of the runtime manager. Currently, only filestream is supported. This implements option 2 from https://github.com/elastic/ingest-dev/issues/4931.

Some notes about the implementation:

* We use the same otel collector as hybrid mode. We add a prefix to each component and pipeline name in an effort to avoid conflicts.
* The Otel configuration is generated from the component model agent already uses for beats processes. It is easier to manipulate and lets us emit the same status information in the future.
* This behaviour is controlled by a single attribute of the Coordinator, the only way to enable it is by changing the code. This is a temporary measure until we decide what feature flags we want for it.
* Status reporting uses the otel collector status, and doesn't try to be compatible with beats processes status.
* Which input types are supported is hardcoded.

## Why is it important?

This is a preprequisite for migrating to Otel within agent, which is our long-term plan. In the short-term, it will make it easier for us to run tests.

## Checklist

1. Change the boolean value in https://github.com/elastic/elastic-agent/blob/0c79718f1ce259cf0980d8085da74732535246de/internal/pkg/agent/application/coordinator/coordinator.go#L402
2. Build an agent package
3. Run agent with a configuration containing a filestream input

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/3530
- Relates https://github.com/elastic/ingest-dev/issues/3517

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -1460,6 +1460,8 @@ func (c *Coordinator) updateManagersWithConfig(model *component.Model) error {
 	return c.updateOtelManagerConfig(otelModel)
 }
 
+// updateOtelManagerConfig updates the otel collector configuration for the otel manager. It assembles this configuration
+// from the component model passed in and from the hybrid-mode otel config set on the Coordinator.
 func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
 	finalOtelCfg := confmap.New()
 	var componentOtelCfg *confmap.Conf
@@ -1496,8 +1498,7 @@ func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
 	return nil
 }
 
-// updateRuntimesWithModel updates runtime managers with the component model.
-// Components may be sent to different runtimes depending on various criteria.
+// splitModelBetweenManager splits the model components between the runtime manager and the otel manager.
 func (c *Coordinator) splitModelBetweenManagers(model *component.Model) (runtimeModel *component.Model, otelModel *component.Model) {
 	if !c.runComponentsInOtelManager {
 		// Runtime manager gets all the components, this is the default
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -1487,7 +1487,6 @@ func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
 		}
 	}
 
-	c.logger.With("otel config", finalOtelCfg.ToStringMap()).Debug("Updating otel config")
 	if len(finalOtelCfg.AllKeys()) == 0 {
 		// if the config is empty, we want to send nil to the manager, so it knows to stop the collector
 		finalOtelCfg = nil
@@ -1516,7 +1515,7 @@ func (c *Coordinator) splitModelBetweenManagers(model *component.Model) (runtime
 	}
 	otelModel = &component.Model{
 		Components: otelComponents,
-		Signed:     model.Signed,
+		// the signed portion of the policy is only used by Defend, so otel doesn't need it for anything
 	}
 	runtimeModel = &component.Model{
 		Components: runtimeComponents,
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -784,7 +784,7 @@ func (c *Coordinator) Run(ctx context.Context) error {
 // information about the state of the Elastic Agent.
 // Called by external goroutines.
 func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
-	return diagnostics.Hooks{
+	hooks := diagnostics.Hooks{
 		{
 			Name:        "agent-info",
 			Filename:    "agent-info.yaml",
@@ -1024,7 +1024,9 @@ func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
 				return o
 			},
 		},
-		{
+	}
+	if c.runComponentsInOtelManager {
+		otelComponentHook := diagnostics.Hook{
 			Name:        "otel-final",
 			Filename:    "otel-final.yaml",
 			Description: "Final otel configuration used by the Elastic Agent. Includes hybrid mode config and component config.",
@@ -1039,8 +1041,10 @@ func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
 				}
 				return o
 			},
-		},
+		}
+		hooks = append(hooks, otelComponentHook)
 	}
+	return hooks
 }
 
 // runner performs the actual work of running all the managers.
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -217,6 +217,12 @@ type Coordinator struct {
 
 	otelMgr OTelManager
 	otelCfg *confmap.Conf
+	// the final config sent to the manager, contains both config from hybrid mode and from components
+	finalOtelCfg *confmap.Conf
+
+	// This variable controls whether we run supported components in the Otel manager instead of the runtime manager.
+	// It's a temporary measure until we decide exactly how we want to control where specific components run.
+	runComponentsInOtelManager bool
 
 	caps      capabilities.Capabilities
 	modifiers []ComponentsModifier
@@ -384,21 +390,22 @@ func New(logger *logger.Logger, cfg *configuration.Configuration, logLevel logp.
 		LogLevel:     logLevel,
 	}
 	c := &Coordinator{
-		logger:     logger,
-		cfg:        cfg,
-		agentInfo:  agentInfo,
-		isManaged:  isManaged,
-		specs:      specs,
-		reexecMgr:  reexecMgr,
-		upgradeMgr: upgradeMgr,
-		monitorMgr: monitorMgr,
-		runtimeMgr: runtimeMgr,
-		configMgr:  configMgr,
-		varsMgr:    varsMgr,
-		otelMgr:    otelMgr,
-		caps:       caps,
-		modifiers:  modifiers,
-		state:      state,
+		logger:                     logger,
+		cfg:                        cfg,
+		agentInfo:                  agentInfo,
+		isManaged:                  isManaged,
+		specs:                      specs,
+		reexecMgr:                  reexecMgr,
+		upgradeMgr:                 upgradeMgr,
+		monitorMgr:                 monitorMgr,
+		runtimeMgr:                 runtimeMgr,
+		configMgr:                  configMgr,
+		varsMgr:                    varsMgr,
+		otelMgr:                    otelMgr,
+		runComponentsInOtelManager: false, // change this to run supported components in the Otel manager
+		caps:                       caps,
+		modifiers:                  modifiers,
+		state:                      state,
 		// Note: the uses of a buffered input channel in our broadcaster (the
 		// third parameter to broadcaster.New) means that it is possible for
 		// immediately adjacent writes/reads not to match, e.g.:
@@ -1015,6 +1022,22 @@ func (c *Coordinator) DiagnosticHooks() diagnostics.Hooks {
 				return o
 			},
 		},
+		{
+			Name:        "otel-final",
+			Filename:    "otel-final.yaml",
+			Description: "Final otel configuration used by the Elastic Agent. Includes hybrid mode config and component config.",
+			ContentType: "application/yaml",
+			Hook: func(_ context.Context) []byte {
+				if c.finalOtelCfg == nil {
+					return []byte("no active OTel configuration")
+				}
+				o, err := yaml.Marshal(c.finalOtelCfg.ToStringMap())
+				if err != nil {
+					return []byte(fmt.Sprintf("error: failed to convert to yaml: %v", err))
+				}
+				return o
+			},
+		},
 	}
 }
 
@@ -1227,7 +1250,6 @@ func (c *Coordinator) runLoopIteration(ctx context.Context) {
 func (c *Coordinator) processConfig(ctx context.Context, cfg *config.Config) (err error) {
 	if c.otelMgr != nil {
 		c.otelCfg = cfg.OTel
-		c.otelMgr.Update(cfg.OTel)
 	}
 	return c.processConfigAgent(ctx, cfg)
 }
@@ -1413,15 +1435,87 @@ func (c *Coordinator) refreshComponentModel(ctx context.Context) (err error) {
 		c.logger.Debugf("Continue with missing \"signed\" properties: %v", err)
 	}
 
-	model := component.Model{
+	model := &component.Model{
 		Components: c.componentModel,
 		Signed:     signed,
 	}
 
 	c.logger.Info("Updating running component model")
 	c.logger.With("components", model.Components).Debug("Updating running component model")
-	c.runtimeMgr.Update(model)
+	return c.updateManagersWithConfig(model)
+}
+
+// updateManagersWithConfig updates runtime managers with the component model and config.
+// Components may be sent to different runtimes depending on various criteria.
+func (c *Coordinator) updateManagersWithConfig(model *component.Model) error {
+	runtimeModel, otelModel := c.splitModelBetweenManagers(model)
+	c.logger.With("components", runtimeModel.Components).Debug("Updating runtime manager model")
+	c.runtimeMgr.Update(*runtimeModel)
+	return c.updateOtelManagerConfig(otelModel)
+}
+
+func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
+	finalOtelCfg := confmap.New()
+	var componentOtelCfg *confmap.Conf
+	if len(model.Components) > 0 {
+		var err error
+		c.logger.With("components", model.Components).Debug("Updating otel manager model")
+		componentOtelCfg, err = getOtelConfig(model, c.agentInfo)
+		if err != nil {
+			c.logger.Errorf("failed to generate otel config: %v", err)
+		}
+	}
+	if componentOtelCfg != nil {
+		err := finalOtelCfg.Merge(componentOtelCfg)
+		if err != nil {
+			c.logger.Error("failed to merge otel config: %v", err)
+		}
+	}
+
+	if c.otelCfg != nil {
+		err := finalOtelCfg.Merge(c.otelCfg)
+		if err != nil {
+			c.logger.Error("failed to merge otel config: %v", err)
+		}
+	}
+
+	c.logger.With("otel config", finalOtelCfg.ToStringMap()).Debug("Updating otel config")
+	if len(finalOtelCfg.AllKeys()) == 0 {
+		// if the config is empty, we want to send nil to the manager, so it knows to stop the collector
+		finalOtelCfg = nil
+	}
+
+	c.otelMgr.Update(finalOtelCfg)
+	c.finalOtelCfg = finalOtelCfg
 	return nil
+}
+
+// updateRuntimesWithModel updates runtime managers with the component model.
+// Components may be sent to different runtimes depending on various criteria.
+func (c *Coordinator) splitModelBetweenManagers(model *component.Model) (runtimeModel *component.Model, otelModel *component.Model) {
+	if !c.runComponentsInOtelManager {
+		// Runtime manager gets all the components, this is the default
+		otelModel = &component.Model{}
+		runtimeModel = model
+		return
+	}
+	var otelComponents, runtimeComponents []component.Component
+	for _, comp := range model.Components {
+		if isComponentOtelSupported(&comp) {
+			otelComponents = append(otelComponents, comp)
+		} else {
+			runtimeComponents = append(runtimeComponents, comp)
+		}
+	}
+	otelModel = &component.Model{
+		Components: otelComponents,
+		Signed:     model.Signed,
+	}
+	runtimeModel = &component.Model{
+		Components: runtimeComponents,
+		Signed:     model.Signed,
+	}
+	return
 }
 
 // generateComponentModel regenerates the configuration tree and
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -13,6 +13,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/elastic/elastic-agent/internal/pkg/otel/configtranslate"
+
 	"go.opentelemetry.io/collector/component/componentstatus"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/status"
@@ -1460,7 +1462,7 @@ func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
 	if len(model.Components) > 0 {
 		var err error
 		c.logger.With("components", model.Components).Debug("Updating otel manager model")
-		componentOtelCfg, err = getOtelConfig(model, c.agentInfo)
+		componentOtelCfg, err = configtranslate.GetOtelConfig(model, c.agentInfo)
 		if err != nil {
 			c.logger.Errorf("failed to generate otel config: %v", err)
 		}
@@ -1501,7 +1503,7 @@ func (c *Coordinator) splitModelBetweenManagers(model *component.Model) (runtime
 	}
 	var otelComponents, runtimeComponents []component.Component
 	for _, comp := range model.Components {
-		if isComponentOtelSupported(&comp) {
+		if configtranslate.IsComponentOtelSupported(&comp) {
 			otelComponents = append(otelComponents, comp)
 		} else {
 			runtimeComponents = append(runtimeComponents, comp)
--- a/internal/pkg/agent/application/coordinator/otelconfig.go
+++ b/internal/pkg/agent/application/coordinator/otelconfig.go
@@ -116,9 +116,6 @@ func getCollectorConfigForComponent(comp *component.Component, info info.Agent) 
 		return nil, err
 	}
 	exportersConfig, err := getExportersConfigForComponent(comp)
-	processorsConfig := map[string]any{
-		getTransformProcessorName(): getTransformProcessorConfig(),
-	}
 	if err != nil {
 		return nil, err
 	}
@@ -128,16 +125,14 @@ func getCollectorConfigForComponent(comp *component.Component, info info.Agent) 
 	}
 	pipelinesConfig := map[string]any{
 		pipelineID.String(): map[string][]string{
-			"exporters":  maps.Keys(exportersConfig),
-			"receivers":  maps.Keys(receiversConfig),
-			"processors": maps.Keys(processorsConfig),
+			"exporters": maps.Keys(exportersConfig),
+			"receivers": maps.Keys(receiversConfig),
 		},
 	}
 
 	fullConfig := map[string]any{
-		"receivers":  receiversConfig,
-		"exporters":  exportersConfig,
-		"processors": processorsConfig,
+		"receivers": receiversConfig,
+		"exporters": exportersConfig,
 		"service": map[string]any{
 			"pipelines": pipelinesConfig,
 		},
@@ -334,31 +329,4 @@ func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
 	// for compatibility with beats, we want bodymap mapping
 	esConfig["mapping"] = map[string]any{"mode": "bodymap"}
 	return esConfig, nil
-}
-
-// getTransformProcessorConfig returns a transform processor configuration needed for all pipelines involving beats
-// receivers and the elasticsearch exporter.
-// beats receivers currently put all event fields into the log body, but we need some of them in log record attributes
-// instead for ES exporter
-func getTransformProcessorConfig() map[string]any {
-	return map[string]any{
-		"error_mode": "ignore",
-		"log_statements": []map[string]any{
-			{
-				"context": "log",
-				"conditions": []string{
-					`IsMap(body) and IsMap(body["data_stream"])`,
-				},
-				// move dataset fields to log record attributes, so ES exporter can use them for index selection
-				"statements": []string{
-					`set(attributes["data_stream.dataset"], body["data_stream"]["dataset"])`,
-					`set(attributes["data_stream.namespace"], body["data_stream"]["namespace"])`,
-				},
-			},
-		},
-	}
-}
-
-func getTransformProcessorName() string {
-	return "transform"
 }
--- a/internal/pkg/agent/application/coordinator/otelconfig.go
+++ b/internal/pkg/agent/application/coordinator/otelconfig.go
@@ -147,12 +147,8 @@ func getReceiversConfigForComponent(comp *component.Component, info info.Agent) 
 	if err != nil {
 		return nil, err
 	}
-	signal, err := getSignalForComponent(comp)
-	if err != nil {
-		return nil, err
-	}
 	// this is necessary to convert policy config format to beat config format
-	defaultDataStreamType, err := signalToDefaultDatastreamType(signal)
+	defaultDataStreamType, err := getDefaultDatastreamTypeForComponent(comp)
 	if err != nil {
 		return nil, err
 	}
@@ -215,6 +211,7 @@ func getExportersConfigForComponent(comp *component.Component) (map[string]any, 
 	return exportersConfig, nil
 }
 
+// getBeatNameForComponent returns the beat binary name that would be used to run this component.
 func getBeatNameForComponent(comp *component.Component) string {
 	// TODO: Add this information directly to the spec?
 	if comp.InputSpec == nil || comp.InputSpec.BinaryName != "agentbeat" {
@@ -223,6 +220,8 @@ func getBeatNameForComponent(comp *component.Component) string {
 	return comp.InputSpec.Spec.Command.Args[0]
 }
 
+// getSignalForComponent returns the otel signal for the given component. Currently, this is always logs, even for
+// metricbeat.
 func getSignalForComponent(comp *component.Component) (pipeline.Signal, error) {
 	beatName := getBeatNameForComponent(comp)
 	switch beatName {
@@ -235,6 +234,7 @@ func getSignalForComponent(comp *component.Component) (pipeline.Signal, error) {
 	}
 }
 
+// getReceiverTypeForComponent returns the receiver type for the given component.
 func getReceiverTypeForComponent(comp *component.Component) (otelcomponent.Type, error) {
 	beatName := getBeatNameForComponent(comp)
 	switch beatName {
@@ -247,6 +247,7 @@ func getReceiverTypeForComponent(comp *component.Component) (otelcomponent.Type,
 	}
 }
 
+// getExporterTypeForComponent returns the exporter type for the given component.
 func getExporterTypeForComponent(comp *component.Component) (otelcomponent.Type, error) {
 	switch comp.OutputType {
 	case "elasticsearch":
@@ -256,6 +257,7 @@ func getExporterTypeForComponent(comp *component.Component) (otelcomponent.Type,
 	}
 }
 
+// unitToExporterConfig translates a component.Unit to an otel exporter configuration.
 func unitToExporterConfig(unit component.Unit, exporterType otelcomponent.Type) (map[string]any, error) {
 	if unit.Type == client.UnitTypeInput {
 		return nil, fmt.Errorf("unit type is an input, expected output: %v", unit)
@@ -304,19 +306,21 @@ func getInputsForUnit(unit component.Unit, info info.Agent, defaultDataStreamTyp
 	return inputs, nil
 }
 
-// signalTypeToDefaultDatastreamType returns the default datastream type for a given otel signal type.
+// getDefaultDatastreamTypeForComponent returns the default datastream type for a given component.
 // This is needed to translate from the agent policy config format to the beats config format.
-func signalToDefaultDatastreamType(signal pipeline.Signal) (string, error) {
-	switch signal {
-	case pipeline.SignalLogs:
+func getDefaultDatastreamTypeForComponent(comp *component.Component) (string, error) {
+	beatName := getBeatNameForComponent(comp)
+	switch beatName {
+	case "filebeat":
 		return "logs", nil
-	case pipeline.SignalMetrics:
+	case "metricbeat":
 		return "metrics", nil
 	default:
-		return "", fmt.Errorf("signal type not supported by Beats receivers: %s", signal)
+		return "", fmt.Errorf("input type not supported by Otel: %s", comp.InputType)
 	}
 }
 
+// translateEsOutputToExporter translates an elasticsearch output configuration to an elasticsearch expoter configuration.
 func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
 	esConfig, err := elasticsearchtranslate.ToOTelConfig(cfg)
 	if err != nil {
--- a/internal/pkg/agent/application/coordinator/otelconfig.go
+++ b/internal/pkg/agent/application/coordinator/otelconfig.go
@@ -1,0 +1,364 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package coordinator
+
+import (
+	"fmt"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	otelcomponent "go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/pipeline"
+	"golang.org/x/exp/maps"
+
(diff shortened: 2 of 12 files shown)


internal/pkg/agent/application/coordinator/otelconfig.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+	return comp.InputSpec.Spec.Command.Args[0]
+}
+
+func getSignalForComponent(comp *component.Component) (pipeline.Signal, error) {
+	beatName := getBeatNameForComponent(comp)
+	switch beatName {
+	case "filebeat":
+		return pipeline.SignalLogs, nil
+	case "metricbeat":
```
khushijain21, line 236: The `metricbeat` also uses `pipeline.SignalLogs`.

internal/pkg/agent/application/coordinator/otelconfig.go
```diff
@@ -0,0 +1,364 @@
...
+	// for compatibility with beats, we want bodymap mapping
+	esConfig["mapping"] = map[string]any{"mode": "bodymap"}
+	return esConfig, nil
+}
+
+// getTransformProcessorConfig returns a transform processor configuration needed for all pipelines involving beats
+// receivers and the elasticsearch exporter.
+// beats receivers currently put all event fields into the log body, but we need some of them in log record attributes
+// instead for ES exporter
+func getTransformProcessorConfig() map[string]any {
```
khushijain21, line 343: This is being taken care by the `otelconsumer`. Related issue here https://github.com/elastic/beats/issues/42623.

You may skip this transformation

internal/pkg/agent/application/coordinator/otelconfig.go
```diff
@@ -0,0 +1,332 @@
...
+		return nil, err
+	}
+	return inputs, nil
+}
+
+// signalTypeToDefaultDatastreamType returns the default datastream type for a given otel signal type.
+// This is needed to translate from the agent policy config format to the beats config format.
+func signalToDefaultDatastreamType(signal pipeline.Signal) (string, error) {
+	switch signal {
+	case pipeline.SignalLogs:
```
khushijain21, line 311: Can we take `beatName` as input here and switch cases accordingly?

This is because this function will now return `logs` for both metricbeat and filebeat type (since they have matching signalTypes)

The expectation is to return `metrics` for metricbeat

cmacknz:
The failure on Linux is:

```
    agent_long_running_leak_test.go:210: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/agent_long_running_leak_test.go:210
        	            				/home/ubuntu/agent/testing/integration/agent_long_running_leak_test.go:141
        	Test:       	TestLongRunningAgentForLeaks/TestHandleLeak
        	Messages:   	install never became healthy: components did not return a healthy state: 
```

The failure is coming from this block in the test that tries to make sure all the components that should be running actually are running: https://github.com/elastic/elastic-agent/blob/99696b87673c84c37d7a5bf5e04e582b11afe9ee/testing/integration/agent_long_running_leak_test.go#L208-L246

That `compDebugName` is empty in the log means that nothing was unhealthy, but one of the expected units is missing or has an unexpected name. 

https://github.com/elastic/elastic-agent/blob/99696b87673c84c37d7a5bf5e04e582b11afe9ee/testing/integration/agent_long_running_leak_test.go#L240-L241

We've got to be failing on the `foundApache` or `foundSystem` checks: https://github.com/elastic/elastic-agent/blob/99696b87673c84c37d7a5bf5e04e582b11afe9ee/testing/integration/agent_long_running_leak_test.go#L245

I see a `log-default` but not `log-apache` so that is probably what is missing or changed looking at the units list. https://github.com/elastic/elastic-agent/blob/99696b87673c84c37d7a5bf5e04e582b11afe9ee/testing/integration/agent_long_running_leak_test.go#L218

leehinman (approved):

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+		slices.Contains(OtelSupportedInputTypes, comp.InputType)
+}
+
+// getSupportedComponents returns components from the given model that can be run in an Otel Collector.
+func getSupportedComponents(model *component.Model) []*component.Component {
+	var supportedComponents []*component.Component
+
+	for _, comp := range model.Components {
+		if IsComponentOtelSupported(&comp) {
+			comp := comp
```
mauri870, line 81: I don't think this loop var assignment is needed after Go 1.22 https://go.dev/wiki/LoopvarExperiment

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+	case "filebeat":
+		return "logs", nil
+	case "metricbeat":
+		return "metrics", nil
+	default:
+		return "", fmt.Errorf("input type not supported by Otel: %s", comp.InputType)
+	}
+}
+
+// translateEsOutputToExporter translates an elasticsearch output configuration to an elasticsearch expoter configuration.
```
mauri870, line 326:
```suggestion
// translateEsOutputToExporter translates an elasticsearch output configuration to an elasticsearch exporter configuration.
```

blakerouse (approved):
This is really well done and has code comments that really explain what is happening and what open questions still exist in this code path.

I think this is a great start on this journey and its only the beginning.

internal/pkg/otel/configtranslate/otelconfig_test.go
```diff
@@ -0,0 +1,335 @@
...
+						"user":      "elastic",
+						"retry": map[string]any{
+							"enabled":          true,
+							"initial_interval": 1 * time.Second,
+							"max_interval":     1 * time.Minute,
+							"max_retries":      3,
+						},
+						"logs_dynamic_index": map[string]any{
+							"enabled": true,
+						},
```
mauri870, line 270: Needs logs_dynamic_id as well

mauri870 (approved):
LGTM. I left some minor feedback. Thanks for working on this—it looks like a huge effort.

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+			}
+			inputs = append(inputs, unitInputs...)
+		}
+	}
+
+	receiverId := getReceiverID(receiverType, comp.ID)
+	// Beat config inside a beat receiver is nested under an additional key. Not sure if this simple translation is
+	// always safe. We should either ensure this is always the case, or have an explicit mapping.
+	beatName := strings.TrimSuffix(receiverType.String(), "receiver")
+	beatDataPath := filepath.Join(paths.Home(), "run", comp.ID)
```
pchila, line 174: This could be simplified using https://github.com/elastic/elastic-agent/blob/38bcd03d6bfe043d1e9cbbe393092dd9600fd14e/internal/pkg/agent/application/paths/common.go#L199-L201
```suggestion
	beatDataPath := filepath.Join(paths.Run(), comp.ID)
```

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+				exportersConfig[k] = v
+			}
+		}
+	}
+	return exportersConfig, nil
+}
+
+// getBeatNameForComponent returns the beat binary name that would be used to run this component.
+func getBeatNameForComponent(comp *component.Component) string {
+	// TODO: Add this information directly to the spec?
```
pchila, line 221: Is this TODO still relevant? Maybe is it worth opening a follow-up issue for it ?

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+	if unit.Type == client.UnitTypeInput {
+		return nil, fmt.Errorf("unit type is an input, expected output: %v", unit)
+	}
+	configTranslationFunc, ok := configTranslationFuncForExporter[exporterType]
+	if !ok {
+		return nil, fmt.Errorf("no config translation function for exporter type: %s", exporterType)
+	}
+	// we'd like to use the same exporter for all outputs with the same name, so we parse out the name for the unit id
+	// these will be deduplicated by the configuration merging process at the end
+	outputName := strings.Split(unit.ID, "-")[1]
```
pchila, line 274: Is this the only way to get the output name? What happens if the input already contains a '-' ?

internal/pkg/otel/configtranslate/otelconfig_test.go
```diff
@@ -0,0 +1,335 @@
...
+				assert.Error(t, actualError)
+				assert.EqualError(t, actualError, tt.expectedError.Error())
+			} else {
+				assert.NoError(t, actualError)
+			}
+		})
+	}
+}
+
+// TODO: Add unit tests for other config generation functions
```
pchila, line 335: What other tests are missing here? Are they going to be added in this or in a follow-up PR ?

swiatekm:
> @leehinman @blakerouse @mauri870 should we worry about these CI failures 🤔 ? https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/1827

> Asking because since you approved I assume that you deem these as flaky tests? Should we create issues about these?

I don't think they're flaky, but the way the failures follow from my changes is not very obvious at the moment. I'm debugging it right now. I suspect the approvals are for the general semantics of this change.

internal/pkg/otel/configtranslate/otelconfig.go
swiatekm: The output name is not saved directly anywhere in the component model, so parsing it out of ids is the only way right now. I'd like to explicitly save it in the component model in a follow-up, to avoid any ambiguities.
But yeah, as both the input type and the output name themselves can contain dashes, the only safe way to do this is to explicitly trim the (known) type from the ID.

internal/pkg/otel/configtranslate/otelconfig.go
swiatekm: Yeah, that's the plan.

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+			"inputs": inputs,
+		},
+		// the output needs to be otelconsumer
+		"output": map[string]any{
+			"otelconsumer": map[string]any{},
+		},
+		// just like we do for beats processes, each receiver needs its own data path
+		"path": map[string]any{
+			"data": beatDataPath,
+		},
```
khushijain21, line 186: some of the args with which agentbeat is started should also be part of the receiverConfig. 
such as `setup.ilm.enabled=false`, `setup.template.enabled=false` and `http.host`

These can be accessed on this level `comp.InputSpec.Spec.Command.Args`

swiatekm: Do we actually need to set all of them? I thought default configuration was different between a beats binary and a beats receiver. What does a beats receiver need `http.host` for?

internal/pkg/otel/configtranslate/otelconfig.go
swiatekm: It's not, but we have a linter enabled (exportloopref) that flags it. I'd rather disable that linter and fix all the incidences of this in a separate PR.

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,366 @@
...
+// getExporterID returns the exporter id for the given exporter type and output name.
+func getExporterID(exporterType otelcomponent.Type, outputName string) otelcomponent.ID {
+	exporterName := fmt.Sprintf("%s%s", OtelNamePrefix, outputName)
+	return otelcomponent.NewIDWithName(exporterType, exporterName)
+}
+
+// getCollectorConfigForComponent returns the Otel collector config required to run the given component.
+// This function returns a full, valid configuration that can then be merged with configurations for other components.
+func getCollectorConfigForComponent(comp *component.Component, info info.Agent) (*confmap.Conf, error) {
+	outputQueueConfig := getOutputQueueConfig(comp)
```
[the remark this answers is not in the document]
swiatekm: You mean that `ToOtelConfig` mutates its input?

cmacknz:
I changed the value of the `runComponentsInOtelManager` flag to true, built this, and installed it with the default agent configuration. It worked (mostly) but as expected a lot of follow up issues will need to be addressed.

Disclaimer that I did this after only skimming the code, perhaps there are already TODOs for some of this. I will do a more in depth review now that I've experimented with it a bit.

1. There are no logs from the collector generated at all.
2. We dump the verbose status even without specifying the `--output=full` format specifier so you always see the otel status. Status in general we already know we need to think about more.
```
sudo elastic-development-agent status
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ pipeline:logs/_agent-component/filestream-default
   │  ├─ status: StatusOK
   │  ├─ exporter:elasticsearch/_agent-component/default
   │  │  └─ status: StatusOK
   │  └─ receiver:filebeatreceiver/_agent-component/filestream-default
   │     └─ status: StatusOK
   └─ pipeline:logs/_agent-component/filestream-monitoring
      ├─ status: StatusOK
      ├─ exporter:elasticsearch/_agent-component/monitoring
      │  └─ status: StatusOK
      └─ receiver:filebeatreceiver/_agent-component/filestream-monitoring
         └─ status: StatusOK
```
3. If you collect diagnostics, it dumps an ugly unmarshalling error at you right away:
```
sudo elastic-development-agent diagnostics
[WARNING] Could not redact otel.yaml due to unmarshalling error: yaml: unmarshal errors:
line 1: cannot unmarshal !!str `no acti...` into map[string]interface {}
Created diagnostics archive "elastic-agent-diagnostics-2025-02-18T22-20-57Z-00.zip"
***** WARNING *****
Created archive may contain plain text credentials.
Ensure that files in arch
```
4. Looking at the logs I see that the filebeat receiver gets created, but it isn't actually running. One of the last logs it dumps out is that it thinks it is running the deprecated log input which is strange because the policy only has filestream in it.
```json
    {"log.level":"error","@timestamp":"2025-02-18T22:17:38.424Z","log.origin":{"function":"github.com/elastic/beats/v7/x-pack/filebeat/fbreceiver.(*filebeatReceiver).Start.func1","file.name":"fbreceiver/receiver.go","file.line":31},"message":"filebeat receiver run error","log":{"source":"elastic-agent"},"kind":"receiver","name":"filebeatreceiver/_agent-component/filestream-default","data_type":"logs","error":{"message":"Failed to start crawler: starting input failed: error while initializing input: Found log input configuration: Log input is deprecated. Use Filestream input instead. Follow our migration guide https://www.elastic.co/guide/en/beats/filebeat/current/migrate-to-filestream.html\n{\n  \"data_stream\": {\n    \"dataset\": \"generic\"\n  },\n  \"id\": \"your-filestream-stream-id\",\n  \"index\": \"logs-generic-default\",\n  \"paths\": [\n    \"/var/log/*.log\"\n  ],\n  \"processors\": [\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"input_id\": \"your-input-id\"\n        },\n        \"target\": \"@metadata\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"dataset\": \"generic\",\n          \"namespace\": \"default\",\n          \"type\": \"logs\"\n        },\n        \"target\": \"data_stream\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"dataset\": \"generic\"\n        },\n        \"target\": \"event\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"stream_id\": \"your-filestream-stream-id\"\n        },\n        \"target\": \"@metadata\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"id\": \"273e0c3c-8e68-4340-abde-c085d9614091\",\n          \"snapshot\": true,\n          \"version\": \"9.1.0\"\n        },\n        \"target\": \"elastic_agent\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"id\": \"273e0c3c-8e68-4340-abde-c085d9614091\"\n        },\n        \"target\": \"agent\"\n      }\n    }\n  ]\n}"},"ecs.version":"1.6.0","log.origin.stack_trace":"github.com/elastic/beats/v7/x-pack/filebeat/fbreceiver.(*filebeatReceiver).Start.func1\n\tgithub.com/elastic/beats/v7@v7.0.0-alpha2.0.20250210163254-7514fa92158a/x-pack/filebeat/fbreceiver/receiver.go:31"}
```
5. Looking at the otel-final.yml configuration, I can see this is because we are not include the `type` field of the input and it is default to the log input which is the default for Filebeat:
```yaml
receivers:
    filebeatreceiver/_agent-component/filestream-default:
        filebeat:
            inputs:
                - data_stream:
                    dataset: generic
                  id: your-filestream-stream-id
                  index: logs-generic-default
                  paths:
                    - /var/log/*.log
                  processors:
                    - add_fields:
                        fields:
                            input_id: your-input-id
                        target: '@metadata'
                    - add_fields:
                        fields:
                            dataset: generic
                            namespace: default
                            type: logs
                        target: data_stream
                    - add_fields:
                        fields:
                            dataset: generic
                        target: event
                    - add_fields:
                        fields:
                            stream_id: your-filestream-stream-id
                        target: '@metadata'
                    - add_fields:
                        fields:
                            id: 273e0c3c-8e68-4340-abde-c085d9614091
                            snapshot: true
                            version: 9.1.0
                        target: elastic_agent
                    - add_fields:
                        fields:
                            id: 273e0c3c-8e68-4340-abde-c085d9614091
                        target: agent
        output:
            otelconsumer: {}
        path:
            data: /Library/Elastic/Agent-Development/data/elastic-agent-9.1.0-SNAPSHOT-469358/run/filestream-default
```

So it runs a Beat receiver, but one that doesn't actually run. I suspect this means we need some better testing put into place to make sure what we create actually works. This could actual functional testing that it can ship data somewhere, or we could hook the beat receivers up to the status reporting properly, etc.

cmacknz:
Also, we seem to have lost the component fields to tell us which input in a policy generated a log line. The beat receiver logs are now `"log":{"source":"elastic-agent"},` as if they were the agent's own logs, which makes complete sense given that's where they run now, but you can't grep specific input logs out anymore. We'll need to think about how to improve that.

cmacknz:
I've also noticed it takes a few seconds for the OTel components to show up in the status output. Is that expected? Is the internal collector starting up slower than the sub-processes?

blakerouse:
> @leehinman  @blakerouse @mauri870 should we worry about these CI failures 🤔 ? https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/1827 
> 
> 
> 
> Asking because since you approved I assume that you deem these as flaky tests? Should we create issues about these?

Approval doesn't mean this PR can bypass CI. An approval means that I reviewed the code and that I like what I see and everything that was changed or added was acceptable. I don't think we should block PR approvals on the fact that CI is not passing exactly. I trust my coworkers to do the correct thing and only merge an acceptable PR.

cmacknz:
I enrolled the agent into Fleet which leaves it as healthy (🚀) with only the monitoring filestream converted to an otel component:

```
sudo elastic-development-agent status
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   └─ pipeline:logs/_agent-component/filestream-monitoring
      ├─ status: StatusOK
      ├─ exporter:elasticsearch/_agent-component/monitoring
      │  └─ status: StatusOK
      └─ receiver:filebeatreceiver/_agent-component/filestream-monitoring
         └─ status: StatusOK
```


Looking at the status information in the .fleet-agents document I see the otel status is making it to Fleet but we are translating all the statuses set to Stopped.

```json
{
  "id": "ed75c79e-7e32-4fa9-9a1b-9ae9ca7b7810",
  "type": "PERMANENT",
  "active": true,
  "enrolled_at": "2025-02-18T22:34:30Z",
  "policy_id": "49ff7ddb-5af6-4ed2-bd32-592e2b4afe96",
  "last_checkin": "2025-02-18T22:38:37Z",
  "last_checkin_status": "online",
  "last_checkin_message": "Running",
  "policy_revision": 4,
  "packages": [],
  "components": [
    {
      "id": "pipeline:logs/_agent-component/filestream-monitoring",
      "type": "otel",
      "status": "STOPPED",
      "message": "Stopped",
      "units": [
        {
          "id": "exporter:elasticsearch/_agent-component/monitoring",
          "type": "",
          "status": "STOPPED",
          "message": "Stopped"
        },
        {
          "id": "receiver:filebeatreceiver/_agent-component/filestream-monitoring",
          "type": "",
          "status": "STOPPED",
          "message": "Stopped"
        }
      ]
    }
  ],
```

For some reason Fleet won't let me install the custom filestream integration, probably it isn't marked as compatible with 9.x yet.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1413,17 +1441,90 @@ func (c *Coordinator) refreshComponentModel(ctx context.Context) (err error) {
...
+// from the component model passed in and from the hybrid-mode otel config set on the Coordinator.
+func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
+	finalOtelCfg := confmap.New()
+	var componentOtelCfg *confmap.Conf
+	if len(model.Components) > 0 {
+		var err error
+		c.logger.With("components", model.Components).Debug("Updating otel manager model")
+		componentOtelCfg, err = configtranslate.GetOtelConfig(model, c.agentInfo)
+		if err != nil {
+			c.logger.Errorf("failed to generate otel config: %v", err)
```
cmacknz, line 1473: Eventually, this error (perhaps any error in this chain) should change the overall agent status to failed or degraded. That will make these failures much more obvious and visible in Fleet.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1413,17 +1441,90 @@ func (c *Coordinator) refreshComponentModel(ctx context.Context) (err error) {
...
+	}
+
+	if c.otelCfg != nil {
+		err := finalOtelCfg.Merge(c.otelCfg)
+		if err != nil {
+			c.logger.Error("failed to merge otel config: %v", err)
+		}
+	}
+
+	c.logger.With("otel config", finalOtelCfg.ToStringMap()).Debug("Updating otel config")
```
cmacknz, line 1490: The `ToStringMap` will leak credentials in the input configurations (e.g. for HTTP based inputs), won't it?

Leaving this in here long term is dangerous, we recently had exactly this InfoSec incident from a debug log of part of the component model.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1413,17 +1441,90 @@ func (c *Coordinator) refreshComponentModel(ctx context.Context) (err error) {
...
+	for _, comp := range model.Components {
+		if configtranslate.IsComponentOtelSupported(&comp) {
+			otelComponents = append(otelComponents, comp)
+		} else {
+			runtimeComponents = append(runtimeComponents, comp)
+		}
+	}
+	otelModel = &component.Model{
+		Components: otelComponents,
+		Signed:     model.Signed,
```
cmacknz, line 1519: The signed portion of the policy is only used by Defend AFAIK, if it's simpler to omit it here. The collector has no use for it yet.

internal/pkg/otel/configtranslate/otelconfig.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+// translateEsOutputToExporter translates an elasticsearch output configuration to an elasticsearch exporter configuration.
+func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
+	esConfig, err := elasticsearchtranslate.ToOTelConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+	// we want to use dynamic indexing
+	esConfig["logs_dynamic_index"] = map[string]any{"enabled": true}
```
cmacknz, line 332: We may want to provide a way to supplement the ES exporter configuration directly from the elastic-agent.yml configuration. Inevitably we will want to change some part of it, and we may be annoyed waiting 12h for the snapshot build job to propagate it everywhere.

internal/pkg/otel/configtranslate/otelconfig_test.go
```diff
@@ -0,0 +1,338 @@
...
+			} else {
+				assert.NoError(t, actualError)
+			}
+		})
+	}
+}
+
+func TestGetOtelConfig(t *testing.T) {
+	agentInfo := &info.AgentInfo{}
+	fileStreamConfig := map[string]any{
```
cmacknz, line 142: You probably want a configuration in this test that includes the "input with streams" syntax, for example:

```yaml
  - type: filestream
    id: input-id-a
    use_output: default
    streams:
      - id: filestream-stream-id-a
        data_stream:
          dataset: generic.first
        paths:
          - /var/log/*.log
      - id: filestream-stream-id-b
        data_stream:
          dataset: generic.second
        paths:
          - /var/log/*.log
```

I tested this and it works because the underlying Agent to Beat configuration parsing handles it, but it doesn't cost much to confirm it here.

Multiple outputs works properly as well I think because the component model is handling the grouping before we even get into this code 👍

internal/pkg/otel/manager/manager.go
```diff
@@ -141,10 +141,12 @@ func (m *OTelManager) Run(ctx context.Context) error {
 				}
 				// ensure that the coordinator knows that there is no error
 				// as the collector is not running anymore
-				select {
-				case m.errCh <- nil:
-				case <-ctx.Done():
-				}
+				go func() {
```
cmacknz, line 144: Why was this necessary? The write to the error channel blocking forever or taking a while is a hint that something else is wrong IMO, assuming that's why this changed.

Perhaps tied to my repeatable observation that the otel components take several seconds to show up in the output of `elastic-agent status`.

blakerouse:
> I enrolled the agent into Fleet which leaves it as healthy (🚀) with only the monitoring filestream converted to an otel component:
> 
> ```
> sudo elastic-development-agent status
> ┌─ fleet
> │  └─ status: (HEALTHY) Connected
> └─ elastic-agent
>    └─ pipeline:logs/_agent-component/filestream-monitoring
>       ├─ status: StatusOK
>       ├─ exporter:elasticsearch/_agent-component/monitoring
>       │  └─ status: StatusOK
>       └─ receiver:filebeatreceiver/_agent-component/filestream-monitoring
>          └─ status: StatusOK
> ```
> 
> Looking at the status information in the .fleet-agents document I see the otel status is making it to Fleet but we are translating all the statuses set to Stopped.
> 
> ```json
> {
>   "id": "ed75c79e-7e32-4fa9-9a1b-9ae9ca7b7810",
>   "type": "PERMANENT",
>   "active": true,
>   "enrolled_at": "2025-02-18T22:34:30Z",
>   "policy_id": "49ff7ddb-5af6-4ed2-bd32-592e2b4afe96",
>   "last_checkin": "2025-02-18T22:38:37Z",
>   "last_checkin_status": "online",
>   "last_checkin_message": "Running",
>   "policy_revision": 4,
>   "packages": [],
>   "components": [
>     {
>       "id": "pipeline:logs/_agent-component/filestream-monitoring",
>       "type": "otel",
>       "status": "STOPPED",
>       "message": "Stopped",
>       "units": [
>         {
>           "id": "exporter:elasticsearch/_agent-component/monitoring",
>           "type": "",
>           "status": "STOPPED",
>           "message": "Stopped"
>         },
>         {
>           "id": "receiver:filebeatreceiver/_agent-component/filestream-monitoring",
>           "type": "",
>           "status": "STOPPED",
>           "message": "Stopped"
>         }
>       ]
>     }
>   ],
> ```
> 
> For some reason Fleet won't let me install the custom filestream integration, probably it isn't marked as compatible with 9.x yet.
> 
> We must be massaging the raw collector status a bit since what I get out of `elastic-agent status --output=json` doesn't match exactly with what is sent to Fleet anymore. The component is getting extracted out of the collector structure.
> 
> ```json
>     "FleetState": 2,
>     "FleetMessage": "Connected",
>     "collector": {
>         "status": 2,
>         "timestamp": "2025-02-18T17:36:46.84802-05:00",
>         "components": {
>             "pipeline:logs/_agent-component/filestream-monitoring": {
>                 "status": 2,
>                 "timestamp": "2025-02-18T17:36:46.84802-05:00",
>                 "components": {
>                     "exporter:elasticsearch/_agent-component/monitoring": {
>                         "status": 2,
>                         "timestamp": "2025-02-18T17:36:46.847945-05:00"
>                     },
>                     "receiver:filebeatreceiver/_agent-component/filestream-monitoring": {
>                         "status": 2,
>                         "timestamp": "2025-02-18T17:36:46.84802-05:00"
>                     }
>                 }
>             }
>         }
>     }
> ```

This output is not related to this PR and is related to the PR I did for hybrid OTel mode. This is all great feedback and as I have discussed with @swiatekm that all the status reporting will need to be cleaned up to make this how we want it to be. We have had many discussions on how we want this to work, hopefully now that we are actually seeing results from running code it will help us better understand what is the best path forward. Before it was all theory now its becoming reality.

As you see from the output that the collector current has its own status block over the control protocol, its questionable if we want to keep that or emulate that into how the Elastic Agent reports today. Except emulating affects the ability to make the Elastic Agent the true upstream collector in the future. A decision needs to be made on the path forward.

Reporting to Fleet is emulated (as in mapped into the components model, so as not to initial change Fleet), but maybe we should change Fleet (or maybe we shouldn't). We need to make a decision that provides something that works today but also leads us into the future.

internal/pkg/otel/manager/manager.go
blakerouse: @swiatekm Has identified this as a deadlock and changing this removes that deadlock. I am looking into why this is the case and if this can be removed/changed into a something different.

blakerouse (changes requested):
With the change to the `go func() {` for the possible deadlock and I am moving my review to request changes, until I can understand why this is happening. Not currently a fan of the addition go routine to make this work.

@swiatekm Thank you for bring this to my attention even after I approved it.

cmacknz:
In addition to the deadlock investigation, the only one of my comments I would say is blocking is the filebeat receivers existing but not actually running, which I suspect can be trivially fixed by preserving the `type` field of the input.

I think this is missing because each Beat has some un-exported transformations.
- For Filebeat, the one you are missing is [here](https://github.com/elastic/beats/blob/fb79d4957e02ed5db3a8a700ba0097712965aab0/x-pack/filebeat/cmd/agent.go#L16-L38) which is called via the [SetTransform hook](https://github.com/elastic/beats/blob/fb79d4957e02ed5db3a8a700ba0097712965aab0/x-pack/filebeat/cmd/root.go#L41-L43)
- For Metricbeat, it's similar and [here](https://github.com/elastic/beats/blob/fb79d4957e02ed5db3a8a700ba0097712965aab0/x-pack/metricbeat/cmd/agent.go#L17-L38). Metricbeat inputs are called "modules" because reasons.
- In general you can find the one for each Beat by searching for calls to `management.ConfigTransform.SetTransform`, which will also lead you to some [brief documentation](https://github.com/elastic/beats/blob/fb79d4957e02ed5db3a8a700ba0097712965aab0/x-pack/libbeat/management/devguide.asciidoc#L55-L67).

It would be nice to figure out how to test for this, so we don't make changes that completely break the beat receivers ability to even start. As I mentioned probably setting the status to something noticeable is the best way, otherwise you'll be search for log lines or looking for events to be written.

Perhaps you can find a way to support configuring the file or other convenient test exporter purely from agent.

internal/pkg/otel/configtranslate/otelconfig.go
khushijain21: `http.host` exposes its internal metrics on this endpoint, which is required by metricbeat to monitor beats.

swiatekm:
>     1. There are no logs from the collector generated at all.

You mean the non-receiver

>     3. If you collect diagnostics, it dumps an ugly unmarshalling error at you right away:
> 
> 
> ```
> sudo elastic-development-agent diagnostics
> [WARNING] Could not redact otel.yaml due to unmarshalling error: yaml: unmarshal errors:
> line 1: cannot unmarshal !!str `no acti...` into map[string]interface {}
> Created diagnostics archive "elastic-agent-diagnostics-2025-02-18T22-20-57Z-00.zip"
> ***** WARNING *****
> Created archive may contain plain text credentials.
> Ensure that files in arch
> ```

This is actually a known issue even without this change: https://github.com/elastic/elastic-agent/issues/6783.

>     4. Looking at the logs I see that the filebeat receiver gets created, but it isn't actually running. One of the last logs it dumps out is that it thinks it is running the deprecated log input which is strange because the policy only has filestream in it.
> 
> ```json
>     {"log.level":"error","@timestamp":"2025-02-18T22:17:38.424Z","log.origin":{"function":"github.com/elastic/beats/v7/x-pack/filebeat/fbreceiver.(*filebeatReceiver).Start.func1","file.name":"fbreceiver/receiver.go","file.line":31},"message":"filebeat receiver run error","log":{"source":"elastic-agent"},"kind":"receiver","name":"filebeatreceiver/_agent-component/filestream-default","data_type":"logs","error":{"message":"Failed to start crawler: starting input failed: error while initializing input: Found log input configuration: Log input is deprecated. Use Filestream input instead. Follow our migration guide https://www.elastic.co/guide/en/beats/filebeat/current/migrate-to-filestream.html\n{\n  \"data_stream\": {\n    \"dataset\": \"generic\"\n  },\n  \"id\": \"your-filestream-stream-id\",\n  \"index\": \"logs-generic-default\",\n  \"paths\": [\n    \"/var/log/*.log\"\n  ],\n  \"processors\": [\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"input_id\": \"your-input-id\"\n        },\n        \"target\": \"@metadata\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"dataset\": \"generic\",\n          \"namespace\": \"default\",\n          \"type\": \"logs\"\n        },\n        \"target\": \"data_stream\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"dataset\": \"generic\"\n        },\n        \"target\": \"event\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"stream_id\": \"your-filestream-stream-id\"\n        },\n        \"target\": \"@metadata\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"id\": \"273e0c3c-8e68-4340-abde-c085d9614091\",\n          \"snapshot\": true,\n          \"version\": \"9.1.0\"\n        },\n        \"target\": \"elastic_agent\"\n      }\n    },\n    {\n      \"add_fields\": {\n        \"fields\": {\n          \"id\": \"273e0c3c-8e68-4340-abde-c085d9614091\"\n        },\n        \"target\": \"agent\"\n      }\n    }\n  ]\n}"},"ecs.version":"1.6.0","log.origin.stack_trace":"github.com/elastic/beats/v7/x-pack/filebeat/fbreceiver.(*filebeatReceiver).Start.func1\n\tgithub.com/elastic/beats/v7@v7.0.0-alpha2.0.20250210163254-7514fa92158a/x-pack/filebeat/fbreceiver/receiver.go:31"}
> ```
> 
>     5. Looking at the otel-final.yml configuration, I can see this is because we are not include the `type` field of the input and it is default to the log input which is the default for Filebeat:
> 
> 
> ```yaml
> receivers:
>     filebeatreceiver/_agent-component/filestream-default:
>         filebeat:
>             inputs:
>                 - data_stream:
>                     dataset: generic
>                   id: your-filestream-stream-id
>                   index: logs-generic-default
>                   paths:
>                     - /var/log/*.log
>                   processors:
>                     - add_fields:
>                         fields:
>                             input_id: your-input-id
>                         target: '@metadata'
>                     - add_fields:
>                         fields:
>                             dataset: generic
>                             namespace: default
>                             type: logs
>                         target: data_stream
>                     - add_fields:
>                         fields:
>                             dataset: generic
>                         target: event
>                     - add_fields:
>                         fields:
>                             stream_id: your-filestream-stream-id
>                         target: '@metadata'
>                     - add_fields:
>                         fields:
>                             id: 273e0c3c-8e68-4340-abde-c085d9614091
>                             snapshot: true
>                             version: 9.1.0
>                         target: elastic_agent
>                     - add_fields:
>                         fields:
>                             id: 273e0c3c-8e68-4340-abde-c085d9614091
>                         target: agent
>         output:
>             otelconsumer: {}
>         path:
>             data: /Library/Elastic/Agent-Development/data/elastic-agent-9.1.0-SNAPSHOT-469358/run/filestream-default
> ```
> 
> So it runs a Beat receiver, but one that doesn't actually run. I suspect this means we need some better testing put into place to make sure what we create actually works. This could actual functional testing that it can ship data somewhere, or we could hook the beat receivers up to the status reporting properly, etc.

Huh. I swear this used to actually work. I'm not sure why it doesn't, surely the policy-to-beats conversion code should take care of it? I'll investigate.

internal/pkg/otel/manager/manager.go
swiatekm: I added a comment explaining this is only here to verify that integration tests pass with it present. I'll remove it once that's done, and we can fix the deadlock in a different PR.

internal/pkg/agent/application/coordinator/coordinator.go
swiatekm: Yeah, good point. This was useful for debugging during development, but as a general purpose debug log, I think it's fine if we just list components and pipelines here.

internal/pkg/otel/configtranslate/otelconfig_test.go
swiatekm: Good catch, done.

swiatekm:
@cmacknz I fixed the input type problem in https://github.com/elastic/elastic-agent/pull/6697/commits/fa371f8c996652fc75991f04e887a98ae7ff9258, filestream now runs correctly. I swear this used to work for me locally without this change, but in hindsight it really shouldn't have, so I'm not sure what happened there. For now, this bit of config translation needs to take place in agent code because the beats functions aren't exported.

internal/pkg/agent/application/coordinator/coordinator.go
swiatekm: On second thought, we already log components in the same function, so this isn't necessary. If someone wants the full config, they can get a diagnostic.

internal/pkg/otel/configtranslate/otelconfig.go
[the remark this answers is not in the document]
swiatekm: We should add this information to the docstring of `ToOtelConfig` imo, ideally return the modified config as well. It's not obvious at all that this happens.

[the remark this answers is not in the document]
swiatekm: I'm also going to leave the preset problem unresolved in this PR, and address it in a follow-up issue. It's not critical right now and it's really annoying to address it with how I have the config generation organized right now.

blakerouse:
PR for the deadlock issue - https://github.com/elastic/elastic-agent/pull/6927

internal/pkg/agent/application/coordinator/coordinator.go
swiatekm: Removed it.

internal/pkg/otel/configtranslate/otelconfig.go
swiatekm: I tried to implement this in a naive way and ran into breakage. This PR is already quite complex, so I'd rather deal with this in a follow-up issue. As-is, the filestream input is functional. I'm also going to have a separate issue for dealing with self-monitoring specifically.

cmacknz:
> What do you mean? I can see lines from the collector when I run agent with the flag enabled:

Yup I see them now, they look like they came from agent itself so they were hard to pick up on the first pass.

internal/pkg/otel/configtranslate/otelconfig.go
cmacknz: I suspect that just giving each beat receiver their own http.host for internal metrics collection as happens now isn't going to work quite right. 

One of the things we scrape from that endpoint are CPU and memory stats, those would now all be for the same process, and shouldn't be duplicated. 

Things like filebeat input metrics could still be scraped this way, but there are things we are going to have to change.

Deferring this to a follow up issue makes the most sense.

cmacknz (approved):
I can confirm filestream now runs as expected. LGTM, pending the deadlock fix incoming from Blake and CI passing as expected.

cmacknz:
Everything else can be addressed in follow up issues.

I think we want this and related functionality in 8.x so that the stack version a user chooses to use doesn't block any of our architectural changes. I've added the 8.x backport label.

internal/pkg/otel/configtranslate/otelconfig.go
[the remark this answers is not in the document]
swiatekm: Opened an issue here: https://github.com/elastic/ingest-dev/issues/5120.

internal/pkg/otel/configtranslate/otelconfig.go
swiatekm: Filed an issue here: https://github.com/elastic/ingest-dev/issues/5128.

internal/pkg/otel/configtranslate/otelconfig.go
swiatekm: Filed an issue here: https://github.com/elastic/ingest-dev/issues/5129.

khushijain21 (approved):

blakerouse (approved):

PR #6710, opened 2025-02-05
Redacted nested secrets path value

michel-laterman:
## What does this PR do?

Redact secrets within complex nested paths.

## Why is it important?

some secrets are getting skipped within diagnostics bundles

## Checklist

## Disruptive User Impact

None

### DIFF
--- a/internal/pkg/diagnostics/diagnostics.go
+++ b/internal/pkg/diagnostics/diagnostics.go
@@ -398,7 +398,8 @@ func redactKey(k string) bool {
 		strings.Contains(k, "passphrase") ||
 		strings.Contains(k, "password") ||
 		strings.Contains(k, "token") ||
-		strings.Contains(k, "key")
+		strings.Contains(k, "key") ||
+		strings.Contains(k, "secret")
 }
 
 func zipLogs(zw *zip.Writer, ts time.Time, topPath string, excludeEvents bool) error {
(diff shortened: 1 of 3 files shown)


blakerouse (approved):
Looks good.

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -593,19 +594,21 @@ func RedactSecretPaths(mapStr map[string]any, errOut io.Writer) map[string]any {
...
 		}
 
-		if ok, _ := cfg.Has(key, -1, ucfg.PathSep(".")); ok {
+		if ok, err := cfg.Has(key, -1, ucfg.PathSep(".")); ok {
 			err := cfg.SetString(key, -1, REDACTED, ucfg.PathSep("."))
 			if err != nil {
 				fmt.Fprintf(errOut, "No output redaction for %q: %v.\n", key, err)
 			}
+		} else {
+			fmt.Fprintf(errOut, "Unable to find secret path %q for redaction: %v.\n", key, err)
```
pkoutsovasilis, line 611: can't err be nil here?

michel-laterman: yes, if the path just does not exist it would a log similar to`Unable to find secret path "non.existent.key" for redaction: <nil>.`

pkoutsovasilis: not the biggest fan of printing nils. On the contrary if ok is true but err is not nil (buggy library) we missed completely logging it, right?

michel-laterman: yes. the current implementation for `Has` makes that impossible though

pkoutsovasilis: then if you have strong feeling on handling an err correctly (imo) and, then please revert your change and make it explicitly ignored.

michel-laterman: I've seperated out the key not found, from the error output

pkoutsovasilis (approved):
LGTM, ty for taking care of my comment

Issue #6631, opened 2025-01-29
Follow up testing work from the addition of installation flavors


cmacknz:
Follow up testing work from https://github.com/elastic/elastic-agent/pull/6542

### Upgrades for agents with an installed flavor
* Start with an agent with the default, basic flavor and upgrade it to another version that supports flavors. After the upgrade none of the binaries from the servers flavor should be on disk.
* Start with an agent with the servers flavor and upgrade to another version that supports flavors. After the upgrade all of the binaries from the servers flavor should be on disk.

### Integration test for components with additional file dependencies
* Ensure the component_files lists of osquerybeat, apm-server, endpoint-security, and cloudbeat are correctly preserved for the appropriate flavors. Ideally, confirm the integration can be installed and remain healthy as a basic sanity check. 
  * endpoint-security is already covered by an integration test but the others are not.

michalpristas:
note: checking for files defined in spec file is not an option, component files is a filter and does not require files to be present. e.g some files present only on linux, exe paths only on windows. this information is not part of the filter definition

pierrehilbert:
My two cents: you shouldn't test all combinations but should ensure the most commons are working (in the same way we are running some tests to ensure we can upgrade with Endpoint enabled).

pkoutsovasilis:
@jlind23 I definitely agree in the sense that all combinations of all flavours in the shape of integrations tests is not realistic. I had a brief discussion about my thinking with @michalpristas . So, I was thinking of a `divide and conquer`-style approach and having components team doing extra testing on those is only helpful, so definitely this can happen. That said, I do believe that within our responsibilities are:
* flavours should contain the expected components and specs; I am thinking that more in the sense of packaging tests than integration ones that run locally and do check files existence and file structure right after packaging
* with the above in place, and if we wanna be extra-cautious, I believe that if we run only 1 integration test that installs a flavour with all components in it (I think it is called unflavoured 😆 ) and test them all (or at least the ones that we wanna test, e.g. endpoint) we should be covered. In terms of code footprint the same code that unpacks, installs, invokes elastic-agent will be the same for all flavours thus no need to test for each one individually

my 2 cents, please @michalpristas do tell me if I missed anything above 🙂

cmacknz:
> flavours should contain the expected components and specs; I am thinking that more in the sense of packaging tests rather than integration ones which run locally and do check files existence and file structure right after packaging

[...]

PR #6712, opened 2025-02-05
Update tests for flavors

michalpristas:
Followup to #6542 

Adding 3 integration tests:
- upgrading from `basic` flavor
- upgrading from `servers` flavor
- upgrading from version not supporting flavors

Related: #6631


blakerouse (approved):
Looks good.

swiatekm (approved):
Looks good to me as well.

Issue #6656, opened 2025-01-30
9.0: Direct users to install with `--install-servers` when `--fleet-server-*` options are used without it


cmacknz:
On the latest 9.0.0-SNAPSHOT, if you attempt to install an agent as a Fleet Server but do not specify the servers flavor, you get a very vague and unhelpful message. We should detect use of the `--fleet-server-*` options at install time and immediately error if `--install-servers` was not also provided on the command line.

Here is example error output:

```sh
sudo ./elastic-agent install \
  --fleet-server-es=http://localhost:9200 \
  --fleet-server-service-token=XXXXXXXX \
  --fleet-server-policy=fleet-server-policy \
  --fleet-server-port=8220 --develop
Installing into development namespace; this is an experimental and currently unsupported feature.
Elastic Agent will be installed at /Library/Elastic/Agent-Development and will run as a service. Do you want to continue? [Y/n]:y
[    ] Service Started  [3s] Elastic Agent successfully installed, starting enrollment.
[    ] Waiting For Enroll...  [8s] {"log.level":"info","@timestamp":"2025-01-30T13:13:38.943-0500","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).prepareFleetTLS","file.name":"cmd/enroll_cmd.go","file.line":437},"message":"Generating self-signed certificate for Fleet Server","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [8s] {"log.level":"info","@timestamp":"2025-01-30T13:13:39.598-0500","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).daemonReloadWithBackoff","file.name":"cmd/enroll_cmd.go","file.line":483},"message":"Restarting agent daemon, attempt 0","ecs.version":"1.6.0"}
[ ===] Waiting For Enroll...  [10s] {"log.level":"info","@timestamp":"2025-01-30T13:13:41.613-0500","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.waitForFleetServer.func1","file.name":"cmd/enroll_cmd.go","file.line":794},"message":"Waiting for Elastic Agent to start Fleet Server","ecs.version":"1.6.0"}
[=   ] Waiting For Enroll...  [2m8s] Error: fleet-server failed: timed out waiting for Fleet Server to start after 2m0s
[...]

PR #6727, opened 2025-02-06
Fail install in case Fleet Server flags are used and `--install-server` flag is missing

michalpristas:
Small change makes install command to fail in case `--fleet-server` flags are used

Fixes: #6656


internal/pkg/agent/cmd/install.go
```diff
@@ -84,6 +86,12 @@ would like the Agent to operate.
 func installCmd(streams *cli.IOStreams, cmd *cobra.Command) error {
 	var err error
 
+	if isFleetServerFlagProvided(cmd) {
+		if installServers, _ := cmd.Flags().GetBool(flagInstallServers); !installServers {
+			return fmt.Errorf("to install fleet server `--%s` flag must be used", flagInstallServers)
```
blakerouse, line 91: Or it just is automatic? Why make the user have to add --install-servers when we know it is needed?

michalpristas: i like that.

cmacknz: This would make the change non-breaking which is much nicer, just make sure to log that this is happening when we do this.

blakerouse (approved):
Looks good!

PR #6730, opened 2025-02-06
Add conditions to copy_fields processor

### LINKED ISSUE #5299: Event log file gets flooded with `copy_fields` processor error
- Version: 8.15.0

The `filestream-monitoring` deployed by Elastic-Agent to collect its own logs is flooding the event logs with an `copy_fields` processor error:
```json
{
  "log.level": "debug",
  "@timestamp": "2024-08-14T13:40:27.920Z",
  "message": "Failed to copy fields in copy_fields processor: target field data_stream.dataset already exists, drop or rename this field first",
  "component": {
    "binary": "filebeat",
    "dataset": "elastic_agent.filebeat",
    "id": "filestream-monitoring",
    "type": "filestream"
  },
  "log": {
    "source": "filestream-monitoring"
  },
  "log.type": "event",
  "ecs.version": "1.6.0",
  "log.logger": "copy_fields",
  "log.origin": {
    "file.line": 81,
    "file.name": "actions/copy_fields.go",
    "function": "github.com/elastic/beats/v7/libbeat/processors/actions.(*copyFields).Run"
  },
  "service.name": "filebeat"
}
```

On my test, that accounted for about 38% of the entries in the event log.

This seems to be coming from the following chain of processors (from `components/filestream-monitoring/beat-rendered-config.yml`):
```yaml
        - copy_fields:
            fields:
                - from: data_stream.dataset
                  to: data_stream.dataset_original
        - drop_fields:
            fields:
                - data_stream.dataset
        - copy_fields:
            fail_on_error: false
            fields:
                - from: component.dataset
                  to: data_stream.dataset
            ignore_missing: true
        - copy_fields:
            fail_on_error: false
            fields:
                - from: data_stream.dataset_original
                  to: data_stream.dataset
```

The third and forth processors copy different fields to the same destination, so if the third runs successfully, then the forth will always fail and generate the log message above.

While this can be the intended behaviour (trying to set `data_stream.dataset` from multiple sources), it is flooding our logs.

We can use some [processor conditions](https://www.elastic.co/guide/en/beats/filebeat/current/defining-processors.html#conditions) to avoid running the `copy_fields` processor if the field is already present, thus avoiding the flood of debug logs.

## Steps to reproduce
- Generate logs with a reasonable rate for testing, like 1 event per second, keep it running
- Add the Custom Logs integration to an Elastic-Agent policy to ingest the log file
- Let the Elastic-Agent run for a minute or so
- Collect the diagnostics
- Inspect the event log file (`logs/elastic-agent-8.15.0-25075f/events/*.ndjosn`)

belimawr:
## What does this PR do?
This commit adds conditions to the `copy_fields` processor from the monitoring Filebeat to prevent it from failing and spamming the event logger at debug level with:
`target field xxx already exists, drop or rename this field first`

## Why is it important?

It makes the debug logs more useful by remove unnecessary entries from our monitoring Filebeat

## Checklist

~~## Disruptive User Impact~~

1. Create a log file with more than 1kb: `docker run -it --rm mingrammer/flog -n 20 > /tmp/flog.log`
2. Package the Elastic-Agent from this PR
3. Start the Elastic-Agent with the following configuration
    <details><summary>elastic-agent.yml</summary>
    <p>
    
    ```
    outputs:
      default:
        type: elasticsearch
        hosts:
          - https://localhost:9200
        protocol: https
        username: elastic
        password: changeme
        preset: latency
        ssl.verification_mode: none
    
    inputs:
      - type: filestream
        id: your-input-id
        log_level: debug
        streams:
          - id: your-filestream-stream-id
            data_stream:
              dataset: generic
            paths:
              - /tmp/flog.log
    
    agent.monitoring:
      enabled: true
      logs: true
      metrics: false
      pprof.enabled: false
      use_output: default
    
    # Needed if you already have an Elastic-Agent running on your machine
    # That's very helpful for running the tests locally
    agent.monitoring:
      http:
        enabled: false
        port: 7002
    
    agent.grpc:
      address: localhost
      port: 7001
    
    ```
    
    </p>
    </details> 

5. Ensure the event logs (`data/elastic-agent*/logs/events/*.ndjson`) do not contain any messages from the `copy_fields` processor. The following command must return `0`:
    ```
    cat data/elastic-agent*/logs/events/*.ndjson| grep copy_fields |wc -l
    ```

## Related issues

- Closes #5299

## Questions to ask yourself

- ...


testing/integration/logs_ingestion_test.go
```diff
@@ -128,7 +128,7 @@ func startMockES(t *testing.T) string {
 		uid,
 		clusterUUID,
 		nil,
-		time.Now().Add(time.Hour), 0, 0, 0, 100, 0))
+		time.Now().Add(time.Hour), 0, 0, 0, 0, 0))
```
belimawr, line 131: This causes the mock to accept all the events, instead of returning an error. The test that relied on the error has been updated.

belimawr:
I looked at the test failures, they seem unrelated to the PR, likely some flakiness communicating with the Elastic-Agent during the test.

pkoutsovasilis (commented):
@belimawr could you write some unit-tests around that to increase the coverage?

belimawr:
> @belimawr could you write some unit-tests around that to increase the coverage?

I can if you insist. The only unit test for it is to test if the conditions are introduced to the policy, but not their effectiveness. I'd test something very generic like: "all `copy_fields` processor contain a `when..not.has.fields` condition matching the `fields.to` for the Filebeat monitoring.

I'm not convinced that this is what we want to enforce.

There is an integration test that ensures the correct behaviour: we do not spam our logs with "failed processors" message from processors we control and can avoid their failure.

What do you think?

pkoutsovasilis:
> I can if you insist. The only unit test for it is to test if the conditions are introduced to the policy, but not their effectiveness. I'd test something very generic like: "all `copy_fields` processor contain a `when..not.has.fields` condition matching the `fields.to` for the Filebeat monitoring.
>
> I'm not convinced that this is what we want to enforce.
> 
> There is an integration test that ensures the correct behaviour: we do not spam our logs with "failed processors" message from processors we control and can avoid their failure.
> 
> What do you think?

I kinda see multiple reasons to have unit-tests:
1. check that your code, ideally your function(s), work as expected without any external dependencies
2. are actually helpful to a newcomer to this code to understand how the code should work
3. having unit-tests can help identify if an integration test is failing because of your code or another service that is part of the integration test

now if you insist on not writing any, I am not here to enforce anything 🙂

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -443,13 +450,27 @@ func (b *BeatsMonitor) injectLogsInput(cfg map[string]interface{}, components []
 								"to":   "data_stream.dataset",
 							},
 						},
+						"when": map[string]any{
```
pkoutsovasilis, line 453: isn't the above `drop_fields` making this when condition redundant?

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -424,6 +424,13 @@ func (b *BeatsMonitor) injectLogsInput(cfg map[string]interface{}, components []
 								"to":   "data_stream.dataset_original",
 							},
 						},
+						"when": map[string]any{
```
pkoutsovasilis, line 427: do we anticipate something to have written in the target `data_stream.dataset_original` field? As far as I can this condition missing isn't the reason of the reported issue, right?

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -471,6 +492,13 @@ func (b *BeatsMonitor) injectLogsInput(cfg map[string]interface{}, components []
 				// update event.dataset with the now used data_stream.dataset
 				map[string]interface{}{
 					"copy_fields": map[string]interface{}{
+						"when": map[string]any{
```
pkoutsovasilis, line 495: do we anticipate something to have written in the target event.dataset field? As far as I can tell this condition missing isn't the reason of the reported issue, right?

internal/pkg/agent/application/monitoring/v1_monitor.go
belimawr: > As far as I can tell this condition missing isn't the reason of the reported issue, right?

You're right, it's not on the reported issue.

Honestly, I don't know by heart all the possible ways `data_stream.dataset_original` could be set. So out of extra caution I added the `when` to all `copy_fields` processors to avoid any of them flooding the debug logs.

internal/pkg/agent/application/monitoring/v1_monitor.go
belimawr: As I mentioned on the other ones, that's just extra caution.

internal/pkg/agent/application/monitoring/v1_monitor.go
belimawr: Done on 7f3b86f3db

internal/pkg/agent/application/monitoring/v1_monitor.go
belimawr: Done on 7f3b86f3db

internal/pkg/agent/application/monitoring/v1_monitor.go
belimawr: Done on 7f3b86f3db

belimawr:
@pkoutsovasilis I tried the `if/then/else` possibility we discussed, however it does not work under Elastic-Agent :/ It breaks somewhere while the policy is being parsed. I got the logic working on a standalone Filebeat, but the Elastic-Agent could not get it sent to Filebeat :/

pkoutsovasilis:
> @pkoutsovasilis I tried the `if/then/else` possibility we discussed, however it does not work under Elastic-Agent :/ It breaks somewhere while the policy is being parsed. I got the logic working on a standalone Filebeat, but the Elastic-Agent could not get it sent to Filebeat :/

hmmm that is really interesting 🤔 could you please create an issue with your findings to not lose track of such behaviour. Also I did notice the build failures and this is related to the unit-test of actually my own PR that got merged yesterday. I have opened a PR that mitigates this failures [here](https://github.com/elastic/elastic-agent/pull/6805) but you would have once more to merge main when this goes in

pkoutsovasilis (dismissed):

belimawr:
@pkoutsovasilis here is the issue: https://github.com/elastic/elastic-agent/issues/6820

kaanyalti (dismissed):

pkoutsovasilis:
hey @belimawr 👋 are you sure that the CI failure doesn't relate with [this one](https://github.com/elastic/elastic-agent/pull/6730/files#r1945418228)? 🤔

belimawr:
@pkoutsovasilis @kaanyalti, CI is finally green, could you take a last look at this PR?

kaanyalti (approved):
Changes look good to me

pkoutsovasilis (approved):

testing/integration/logs_ingestion_test.go
belimawr: I ended up updating the code so whomever is creating the mock can pass the desired probabilities.

PR #6801, opened 2025-02-11
[e2e] Add tests for self monitoring agent logs in otel mode

khushijain21:
## What does this PR do?

It adds tests for comparing docs ingested by `agent.monitoring` vs otel.yml

Notes:
Some changes made to otel.yml produced by hand [here](https://github.com/elastic/ingest-dev/issues/4875#issuecomment-2640536930):

## Checklist

## Related issues

- Closes [#4877](https://github.com/elastic/ingest-dev/issues/4877)


khushijain21:
Waiting on https://github.com/elastic/elastic-agent/pull/6837

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,323 @@ service:
...
+		ESEndpoint     string
+		ESApiKey       string
+		BeatsESApiKey  string
+		SocketEndpoint string
+	}
+	esEndpoint, err := getESHost()
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
```
michel-laterman, line 1630:
```suggestion
	require.NotEmptyf(t, esApiKey.Encoded, "api key is invalid %q", esApiKey)
```

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,323 @@ service:
...
+      exporters:
+        - elasticsearch/log
+`
+
+	beatsApiKey, err := base64.StdEncoding.DecodeString(esApiKey.Encoded)
+	require.NoError(t, err, "error decoding api key")
+
+	var configBuffer bytes.Buffer
+	require.NoError(t,
+		template.Must(template.New("config").Parse(configTemplate)).Execute(&configBuffer,
```
michel-laterman, line 1803: `template.Must` will panic on an error, you don't need it as a part of this call

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,323 @@ service:
...
+
+	if len(files) == 0 {
+		t.Errorf("No matching files found in path: %s", pattern)
+	}
+
+	var agentDocs estools.Documents
+	var otelDocs estools.Documents
+	require.Eventually(t,
+		func() bool {
+			findCtx, findCancel := context.WithTimeout(context.Background(), 10*time.Second)
```
michel-laterman, line 1865:
```suggestion
			findCtx, findCancel := context.WithTimeout(ctx, 10*time.Second)
```

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,323 @@ service:
...
+			require.NoError(t, err)
+			otelDocs, err = estools.GetLogsForIndexWithContext(findCtx, info.ESClient, ".ds-"+fbReceiverMonitoringIndex+"*", map[string]interface{}{
+				"data_stream.dataset": "otel",
+			})
+			require.NoError(t, err)
+
+			// Atleast 1 Document in each index should be present
+			return otelDocs.Hits.Total.Value > 1 && agentDocs.Hits.Total.Value > 1
+		},
+		2*time.Minute, 1*time.Second, "the number of logs in both index are not same")
```
michel-laterman, line 1881:
```suggestion
		2*time.Minute, 1*time.Second, "a datastream is empty")
```
We don't validate counts are equal

michel-laterman (commented):
I have nitpicks regarding the test, the test is also failing,

I think you're missing some ignore fields:
```
+- 	"cloud.instance.id":       string("6692864320899608549"),
--
  | ++ 	"cloud.instance.id":       string("5902608334348578789"),
  | +  	"cloud.instance.name": strings.Join({
  | +  		"bk-agent-prod-gcp-1740034313",
  | +- 		"605048021",
  | ++ 		"705811855",
  | +  	}, ""),
```

michel-laterman:
There is still a failure with the test:
```
=== FAIL: testing/integration TestMonitoringAgentE2E (42.33s)
--
  | fetcher.go:93: Using existing artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz
  | fixture.go:306: Extracting artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestMonitoringAgentE2E1798530589
  | fixture.go:324: Completed extraction of artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestMonitoringAgentE2E1798530589
  | fixture.go:924: Components were not modified from the fetched artifact
  | fixture.go:681: >> running binary with: [/tmp/TestMonitoringAgentE2E1798530589/elastic-agent-9.0.0-SNAPSHOT-linux-x86_64/elastic-agent status --output json]
  | otel_test.go:1819: waiting for agent healthy: agent status returned an error: could not unmarshal agent status output: invalid character 'E' looking for beginning of value
  | exit status 1:
  | Error: failed to communicate with Elastic Agent daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /tmp/elastic-agent/eF0-NlU2jTIS15SmZ3bn9S5SBuGINFf4.sock: connect: no such file or directory"
  | For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
  | fixture.go:681: >> running binary with: [/tmp/TestMonitoringAgentE2E1798530589/elastic-agent-9.0.0-SNAPSHOT-linux-x86_64/elastic-agent status --output json]
  | otel_test.go:1819: waiting for agent healthy: agent status returned an error: could not unmarshal agent status output: invalid character 'E' looking for beginning of value
```

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,287 @@ service:
...
+	err = fixture.Configure(ctx, configContents)
+	require.NoError(t, err)
+
+	cmd, err := fixture.PrepareAgentCommand(ctx, nil)
+	require.NoError(t, err)
+	cmd.WaitDelay = 1 * time.Second
+
+	var output strings.Builder
+	cmd.Stderr = &output
+	cmd.Stdout = &output
```
leehinman, line 1804: The framework should already logs on error.  Was that not working?

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,287 @@ service:
...
+	// This is important for the test to succeed
+	pattern := inputPath + "/data/elastic-agent-*/logs/elastic-agent-*.ndjson"
+	files, err := filepath.Glob(pattern)
+	if err != nil {
+		t.Errorf("error checking files: %v", err)
+	}
+
+	if len(files) == 0 {
+		t.Errorf("No matching files found in path: %s", pattern)
+	}
```
leehinman, line 1835: I'd recommend changing to a `require.Eventually`, just in case the logs need some extra time to arrive.  Also, that first "*" really opens you up to matching logs that aren't part of this test.  So if some other test left a log file, you would match on that, not the ones from the current run.  Can you make the match more specific to just files that would appear from this run?

leehinman (commented):
Overall, I'd suggest that we run the `elastic-agent` command twice instead of once.  Once with beats receivers and once without.  By doing them at the same time we run the risk that the two modes interact in a way that we don't expect.

testing/integration/otel_test.go
khushijain21: no, it did not output the cause of the error.

testing/integration/otel_test.go
khushijain21: > Can you make the match more specific to just files that would appear from this run?

The input path returns the agent directory that is specific to this test so we are sure it will not interfere with other left over runs.

testing/integration/otel_test.go
```diff
@@ -1595,3 +1597,287 @@ service:
...
+		Local: true,
+		OS: []define.OS{
+			{Type: define.Linux},
+			{Type: define.Darwin},
+		},
+		Stack: &define.Stack{},
+	})
+
+	fbMonitoringIndex := "logs-elastic_agent-notdefault" // the namespace here is different to avoid reading stale logs from previous runs
+	fbReceiverMonitoringIndex := "logs-otel-default"
```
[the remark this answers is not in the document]
leehinman: We could run into problems with Index settings being different and potentially pipelines being different.  We should at least be able to separate them time.

testing/integration/otel_test.go
```diff
@@ -1595,3 +1599,334 @@ service:
...
+  port: 6792	
+receivers:
+  filebeatreceiver/filestream-monitoring:
+    filebeat:
+      inputs:
+        - type: filestream
+          enabled: true
+          id: filestream-monitoring-agent
+          paths:
+            - {{.InputPath}}/data/elastic-agent-*/logs/elastic-agent-*.ndjson
```
leehinman, line 1723: Is there any chance you could be picking up logs from the previous run with this?

testing/integration/otel_test.go
```diff
@@ -1595,3 +1599,334 @@ service:
...
+	otel := otelDocs.Hits.Hits[0].Source
+	ignoredFields := []string{
+		// Expected to change between agentDocs and OtelDocs
+		"@timestamp",
+		"agent.ephemeral_id",
+		"agent.id",
+		"agent.version",
+		"data_stream.dataset",
+		"data_stream.namespace",
+		"event.dataset",
```
leehinman, line 1879: shouldn't these be the same?  We didn't change agent version, and we need the data_stream to be the same.

cmacknz: Yes if the data_stream fields aren't right we are not writing to the correct index in Elasticsearch and the elastic_agent integration and parts of the Fleet UI that rely on this data won't work.

testing/integration/otel_test.go
khushijain21: no, this input path is unique to every fixture.

testing/integration/otel_test.go
khushijain21: `dataset` is different in this test. Again, for same reasons as [here](https://github.com/elastic/elastic-agent/pull/6801#discussion_r1974909800)

`agent.version` is different in a similar test here too. 
https://github.com/elastic/elastic-agent/pull/6837/files#diff-6a6ea8fd6d7a7e32e68b936b4e5377c511620b54da40b01243fea2f2a67fe992R1329

mauri870: About agent.version, I think it differs for the fbreceiver because we always require beats latest, depending on the beats commit it will reflect another version. I have the feeling that this should match the elastic-agent version, but how can we achieve that? This is also a problem in 8.x backports were beats receivers will report 9.x for example.

khushijain21:
As discussed, I have changed the "namespace" part of the datastream to ensure we don't run into problems arising with different index settings.

pierrehilbert:
```

otel_test.go:1893:
--
  | Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1741673427581617733/elastic/elastic-agent/testing/integration/otel_test.go:1364
  | /opt/buildkite-agent/builds/bk-agent-prod-gcp-1741673427581617733/elastic/elastic-agent/testing/integration/otel_test.go:1893
  | Error:      	expected documents to be equal
  | Test:       	TestMonitoringAgentE2E
  | Messages:   	ignored field "log.offset" is equal in both maps, please remove it from the ignored fields
  | otel_test.go:1893:


```

cc @khushijain21

khushijain21:
Waiting for https://github.com/elastic/elastic-agent/pull/7367 to be merged before I refactor to make it work with the main [test](https://github.com/elastic/elastic-agent/pull/7367/files#diff-449de09ddd8542557a06d9cec89af187be217c88c46fe1cba90ad7d824936c81R34)

pkg/testing/fixture.go
```diff
@@ -228,6 +228,22 @@ func (f *Fixture) Prepare(ctx context.Context, components ...UsableComponent) er
 	return nil
 }
 
+// Return path of running directory, should only be called after Prepare()
+func (f *Fixture) GetRunningDir(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
khushijain21: yes, we would still need the path to the logs.

khushijain21:
The `check-ci` fails because integration tests here are running inside `t.Run()` blocks for readability and keeping sub tests isolated. Can we redefine this check to accommodate this? 

The failure -
```
Running dependency: main.Integration.Check
>> check: Checking for define.Require in integration tests
Error: test TestAgentMonitoring first statement must be a function call to define.Require
make: *** [Makefile:29: check-ci] Error 1
```

pchila:
> The `check-ci` fails because integration tests here are running inside `t.Run()` blocks for readability and keeping sub tests isolated. Can we redefine this check to accommodate this?
> 
> The failure -
> 
> ```
> Running dependency: main.Integration.Check
> >> check: Checking for define.Require in integration tests
> Error: test TestAgentMonitoring first statement must be a function call to define.Require
> make: *** [Makefile:29: check-ci] Error 1
> ```

The check for `define.Require()` at the beginning of the outer test function has been introduced because the initial integration test framework rely on it to skip the whole test during the discovery phase while dumping the requirements without any extra logs.
See 
https://github.com/elastic/elastic-agent/blob/b011c25412801880bda7cd051a3096ca29a49b4d/pkg/testing/define/define_define.go#L15
and
https://github.com/elastic/elastic-agent/blob/b011c25412801880bda7cd051a3096ca29a49b4d/pkg/testing/define/batch.go#L81

The check itself is defined here https://github.com/elastic/elastic-agent/blob/b011c25412801880bda7cd051a3096ca29a49b4d/magefile.go#L1951
and it parses the integration tests directory for `*_test.go` files containing `Test*` functions and [validating that the first statement in the ast  is a `define.Require` call](https://github.com/elastic/elastic-agent/blob/b011c25412801880bda7cd051a3096ca29a49b4d/pkg/testing/define/parser.go#L40-L60)

Your PR is not passing this check since the `define.Require` call has been wrapped into a `t.Run()` in `TestAgentMonitoring`, just move it out of `t.Run` and the check will pass. This is a limitation of the initial integration framework implementation that does not have full support for subtests with different requirements. Since you have different requirements across the subtests (like [here](https://github.com/elastic/elastic-agent/pull/6801/files#diff-449de09ddd8542557a06d9cec89af187be217c88c46fe1cba90ad7d824936c81R51) and [here](https://github.com/elastic/elastic-agent/pull/6801/files#diff-449de09ddd8542557a06d9cec89af187be217c88c46fe1cba90ad7d824936c81R222) ) you may need to split the TestAgentMonitoring test function into multiple ones.

Maybe @blakerouse as the original implementer of the check and the batching may have another solution.

pchila:
Add-on to my previous comment: after chatting with @khushijain21 it seems that the only difference between the 2 sub-tests is that [this subtest](https://github.com/elastic/elastic-agent/pull/6801/files#diff-449de09ddd8542557a06d9cec89af187be217c88c46fe1cba90ad7d824936c81R221) must not be run on windows.

The same thing can be achieved by using a single `define.Require()` call as it was defined before this change and add a `t.Skip()` on the second sub-test if the OS is windows. It's not very elegant but it should achieve what you need without triggering the integration test mage check.

pkg/testing/fixture.go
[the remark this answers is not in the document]
khushijain21: no, we wouldn't

testing/integration/beat_receivers_test.go
```diff
@@ -7,180 +7,462 @@
...
+			BeatsESApiKey  string
+			SocketEndpoint string
+		}
+		esEndpoint, err := getESHost()
+		require.NoError(t, err, "error getting elasticsearch endpoint")
+		esApiKey, err := createESApiKey(info.ESClient)
+		require.NoError(t, err, "error creating API key")
+		require.NotEmptyf(t, esApiKey.Encoded, "api key is invalid %q", esApiKey)
+
+		var inputPath string
```
leehinman, line 250: I really don't like the inputPath based on GOOS.  Once it is installed `fixture.WorkDir()` should point to the correct top directory.  I wonder if we should first install with a blank `elastic-agent.yml`, then write the correct `elastic-agent.yml` and then do a restart.  That way we would know the correct paths.  We can use a filter on our searches based on time to exclude any logs that might be generated with the blank `elastic-agent.yml`

testing/integration/beat_receivers_test.go
```diff
@@ -7,180 +7,462 @@
...
+			err = fixture.IsHealthy(ctx)
+			if err != nil {
+				t.Logf("waiting for agent healthy: %s", err.Error())
+				return false
+			}
+			return true
+		}, 30*time.Second, 1*time.Second)
+
+		var otelDocs estools.Documents
+		require.Eventually(t,
```
leehinman, line 423: Can we do the query like above with table checking the different type-datastream-namespace above?  I know the metrics will be commented out for now, but doing so will make it easier to add when the metrics are possible.

testing/integration/beat_receivers_test.go
```diff
@@ -7,180 +7,462 @@
...
+			2*time.Minute, 1*time.Second, "could not find monitoring log")
+	})
+
+	t.Run("compare logs ingested by agent monitoring vs otel monitoring", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip("skipping this test on windows for now")
+		}
+
+		// Not proceed with this test if monitoring logs from elastic-agent does not exist
+		require.NotPanics(
```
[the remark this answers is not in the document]
khushijain21: It is in case the above test fails - we should not proceed with this subtest since it depends on it.

testing/integration/beat_receivers_test.go
khushijain21: Right, for now I have added a query for logs only. And this can be extended when we write the metrics part

khushijain21: The logs from both agent and otel are now being sent to the same index. We are uniquely identifying otel logs based on timestamp field.

leehinman (approved):

PR #6818, opened 2025-02-11
Adding unit tests for debian 12

ev1yehor:
## What does this PR do?
Adding unit tests for Debian 12 image that was created previously.

## Checklist


v1v (dismissed):

.buildkite/pipeline.yml
```diff
@@ -53,6 +53,21 @@ steps:
...
+          - "build/TEST-*.html"
+          - "build/TEST-*.xml"
+          - "build/diagnostics/*"
+          - "coverage-*.out"
+        agents:
+          provider: "gcp"
+          image: "family/platform-ingest-elastic-agent-debian-12"
+        retry:
+          manual:
+            allowed: true
```
dliappis, line 69: FYI auto retries have been added as part of https://github.com/elastic/elastic-agent/pull/6982

@pkoutsovasilis do you want to have auto retries added here too?

dliappis (dismissed):
LGTM, but if @pkoutsovasilis feels we should have auto retries here too, let's wait for his comment before mergin

pkoutsovasilis (commented):
Hey 👋 IMO having another operating system running just the unit-tests isn't that beneficial. Ideally we would want to run the integration tests on it, but I don't believe that we can invoke them in parallel with ubuntu both targeting the same ECH stack. Thus this requires a little bit more thinking

.buildkite/pipeline.yml
ev1yehor: Updated.

v1v:
> So should I merge this changes and may be create a new issue about integration tests?

@ev1yehor , IIUC, UTs are not something valuable but the Integrations Tests. I'm not an expert, but if UTs don't actually do much in terms of using the underlying OS, I'd defer to creating a new PR for the ITs.

PR #5309, opened 2024-08-16
feat: bump go to 1.23.6

kruskall:
## What does this PR do?

Update dockerfile go versions
Update documentation go version
Update go.mod
Remove go version from golangci config so it
defaults to the one used in go.mod

## Why is it important?

Bump to the latest go version: https://go.dev/doc/go1.23
Also adds some nice feature that allows us to drop two more dependency (x/exp and copy)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #6825, opened 2025-02-11
Windows: Fix Windows job assignment to work on Go 1.23+

cmacknz:
## What does this PR do?

https://github.com/elastic/elastic-agent/pull/5309 has been open for some time with failing Windows tests. Now that Go 1.24 is available and 1.22 is EOL, we are forced to fix the problems and upgrade to Go 1.23.

This PR includes two changes:
1. https://github.com/elastic/elastic-agent/commit/ecc27a5316eeca9cdd25753f77e2dc7f48410404 A debugging change to dump out the local agent status when the remote status in Fleet isn't what is in fixes. This allows the underlying error to be identified as `Failed: failed job assignment "C:\\Program Files\\Elastic\\Agent\\data\\elastic-agent-9.0.0-SNAPSHOT-7a8a5b\\components\\agentbeat.exe": The handle is invalid`.
2. https://github.com/elastic/elastic-agent/commit/c7077ad116fc36536030dd186f3598bc260cccb3 A fix for the problem, which is to stop using struct memory aliasing to get a Windows process handle suitable for use in Job assignment. Instead we use [OpenProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess) which is what both [upstream Go](https://github.com/golang/go/blob/dcbdc1a2f7368ad8a9193e969cc76c7ffd2f7685/src/os/exec_windows.go#L84-L91) and the [Buildkite agent](https://github.com/buildkite/agent/blob/440b8fb9c8988c88c421a7d6f0eeadf9cf489943/process/signal_windows.go#L58-L74) use to do the same thing.

The exact details of the problem are that we get at the process handle in `os.Process` by aliasing a struct with the same layout on top of it: https://github.com/elastic/elastic-agent/blob/45e3abf15c0fd1a351c5820646b5bd2eff6ce46b/pkg/core/process/job_windows.go#L75-L87

In Go 1.22 the layout of `os.Process` was https://github.com/golang/go/blob/5817e650946aaa0ac28956de96b3f9aa1de4b299/src/os/exec.go#L21-L26 

```go
type Process struct {
	Pid    int
	handle uintptr      // handle is accessed atomically on Windows
	isdone atomic.Bool  // process has been successfully waited on
	sigMu  sync.RWMutex // avoid race between wait and signal
}
```

In Go 1.23, the layout of `os.Process` changed such that this doesn't work anymore. It has also been updated with comments saying not to do what we are doing. https://github.com/golang/go/blob/6885bad7dd86880be6929c02085e5c7a67ff2887/src/os/exec.go#L58-L103
```go
type Process struct {
	Pid int

	mode processMode

        // ...
	state atomic.Uint64

	// Used only in modePID.
	sigMu sync.RWMutex // avoid race between wait and signal

	// handle is the OS handle for process actions, used only in
	// modeHandle.
	//
	// handle must be accessed only via the handleTransientAcquire method
	// (or during closeHandle), not directly! handle is immutable.
	//
	// On Windows, it is a handle from OpenProcess.
	// On Linux, it is a pidfd.
	// It is unused on other GOOSes.
	handle uintptr
```

This does not appear in the release notes because changes to unexported struct layout are not a change anyone should have to note, because they shouldn't be depending on it. And yet, here we are.

## Why is it important?

We have to update to Go 1.23 now that 1.22 is EOL.

Build the agent for Windows, install it in a Windows VM, and use Task Manager to confirm stopping the agent also stops the sub-processes.


cmacknz:
Opening a draft to sanity check CI, I tested this manually already.

It should be possible to write a test (or hijack an existing test) to make sure that stopping the agent also stops the sub-processes. I'll see if I can do that before taking this out of draft.

cmacknz:
We already check that there are no elastic-agent processes running at the end of our tests, so I just extended it to also check that there are no agentbeat sub-processes either as the test for that this didn't break anything. https://github.com/elastic/elastic-agent/pull/6825/commits/22f9f816a6d3a5e12c379b0d0e7691a17f922aa0

pkg/core/process/external_windows.go
```diff
@@ -34,16 +36,17 @@ func externalProcess(proc *os.Process) {
 
 func isWindowsProcessExited(pid int) bool {
 	const desiredAccess = syscall.STANDARD_RIGHTS_READ | syscall.PROCESS_QUERY_INFORMATION | syscall.SYNCHRONIZE
-	h, err := syscall.OpenProcess(desiredAccess, false, uint32(pid))
+	h, err := windows.OpenProcess(desiredAccess, false, uint32(pid)) //nolint:gosec // G115 Conversion from int to uint32 is safe here.
 	if err != nil {
 		// failed to open handle, report exited
 		return true
 	}
+	defer windows.CloseHandle(h) //nolint:errcheck // No way to handle errors returned here so safe to ignore.
```
cmacknz, line 44: Running the linter with GOOS=windows showed me this file and I noticed we were not closing the handle we got from OpenProcess, so I fixed it and also silenced the lint warnings.

pkg/core/process/job_windows.go
```diff
@@ -76,12 +77,24 @@ func (job Job) Assign(p *os.Process) error {
...
-type process struct {
-	Pid    int
-	Handle uintptr
+	// To assign a process to a job, you need a handle to the process. Since os.Process provides no
+	// way to obtain it's underlying handle safely, get one with OpenProcess().
+	//   https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess
+	// This requires at least the PROCESS_SET_QUOTA and PROCESS_TERMINATE access rights. Various
+	// examples also add VM_READ or the safer PROCESS_QUERY_LIMITED_INFORMATION so include that too.
+	//   https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject
+	desiredAccess := uint32(windows.PROCESS_QUERY_LIMITED_INFORMATION | windows.PROCESS_SET_QUOTA | windows.PROCESS_TERMINATE)
```
swiatekm, line 87: Are we asking for new access rights here, in effect, which would make this a breaking change?

mauri870: I don't think that is the case. [Here](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject#parameters) it states that the process handle MUST have PROCESS_SET_QUOTA and PROCESS_TERMINATE. The fact that it was working before confirms that that was the case. The only additional flag is PROCESS_QUERY_LIMITED_INFORMATION which I don't have much information about.

mauri870 (approved):

pkg/core/process/job_windows.go
cmacknz: https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject
> A handle to the process to associate with the job object. The handle must have the PROCESS_SET_QUOTA and PROCESS_TERMINATE access rights. 

PROCESS_SET_QUOTA and PROCESS_TERMINATE are required, we have to set them. 

Golang asks for `PROCESS_QUERY_INFORMATION` in its handle, and Buildkite asks for `PROCESS_QUERY_LIMITED_INFORMATION`. I can try removing it, but `PROCESS_QUERY_LIMITED_INFORMATION` is quite safe and we are able to get it for processes that aren't child processes for system metrics collection https://github.com/search?q=repo%3Aelastic/elastic-agent-system-metrics%20PROCESS_QUERY_LIMITED_INFORMATION&type=code

This is what it allows from https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights:

> Required to retrieve certain information about a process (see [GetExitCodeProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess), [GetPriorityClass](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getpriorityclass), [IsProcessInJob](https://learn.microsoft.com/en-us/windows/win32/api/jobapi/nf-jobapi-isprocessinjob), [QueryFullProcessImageName](https://learn.microsoft.com/en-us/windows/desktop/api/WinBase/nf-winbase-queryfullprocessimagenamea)). A handle that has the PROCESS_QUERY_INFORMATION access right is automatically granted

Mostly I added this initially to avoid having to fiddle too much with Windows permissions.

pierrehilbert (approved):

swiatekm (approved):

pkg/core/process/job_windows.go
cmacknz: I removed `PROCESS_QUERY_LIMITED_INFORMATION` to see if it works, now that I'm confident the tests will detect if I broke this.

mauri870 (approved):

leehinman (approved):

cmacknz:
Removing backport from 8.16 and 8.17 as those are likely to stay on Go 1.22 until they have no further planned releases.

PR #6837, opened 2025-02-12
chore(go.mod): update beats dependency

mauri870:
This PR updates beats/v7 to point to https://github.com/elastic/beats/commit/ef79280d53a29f011d6f4b5f78bbaec10e8b7221.

This is needed for the recent fixes and work on the beats receivers, for example https://github.com/elastic/beats/pull/42412, https://github.com/elastic/beats/pull/42658, https://github.com/elastic/beats/pull/42707 and https://github.com/elastic/beats/pull/42725.


swiatekm (approved):

mauri870:
Decided to update to https://github.com/elastic/beats/commit/d8133114f109461a0e39d44a88a76a1998d389c0 instead, since it includes a fix for default processors in fbreceiver.

testing/integration/otel_test.go
```diff
@@ -887,6 +887,8 @@ func TestOtelFBReceiverE2E(t *testing.T) {
           enabled: true
           paths:
             - {{.InputPath}}
+          prospector.scanner.fingerprint.enabled: false
```
pkoutsovasilis, line 890: just wondering is this a valid `chore` PR if we do changes of configuration in our integration tests? Is this change required to happen because the previous config is no longer valid with the prior version?

mauri870: This is a leftover, it is like this for other tests. Long story short filebeat now has [fingerprint fiile identity by default](https://github.com/elastic/beats/pull/41762), which causes issues in tests that ingest a small file that is not big enough to produce a valid fingerprint. These options are used in most tests because of this.

PR #6840, opened 2025-02-12
Remove .package-version file and its usage

pazone:
## What does this PR do?
Removes the `.package-version` file and all usages. 

## Why is it important?
The `.package-version` file is not required anymore and brings unnecessary complexity to the CI and related codebase

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/steps/integration_tests.sh
+++ b/.buildkite/scripts/steps/integration_tests.sh
@@ -7,15 +7,8 @@ STACK_PROVISIONER="${1:-"stateful"}"
 MAGE_TARGET="${2:-"integration:test"}"
 MAGE_SUBTARGET="${3:-""}"
 
-
-# Override the stack version from `.package-version` contents
-# There is a time when the current snapshot is not available on cloud yet, so we cannot use the latest version automatically
-# This file is managed by an automation (mage integration:UpdateAgentPackageVersion) that check if the snapshot is ready.
-
-STACK_VERSION="$(cat .package-version)"
-if [[ -n "$STACK_VERSION" ]]; then
-    STACK_VERSION=${STACK_VERSION}"-SNAPSHOT"
-fi
+STACK_VERSION=$(grep "const defaultBeatVersion =" version/version.go | cut -d\" -f2)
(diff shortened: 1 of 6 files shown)


.buildkite/scripts/steps/integration_tests.sh
```diff
@@ -7,15 +7,8 @@ STACK_PROVISIONER="${1:-"stateful"}"
 MAGE_TARGET="${2:-"integration:test"}"
 MAGE_SUBTARGET="${3:-""}"
 
-
-# Override the stack version from `.package-version` contents
-# There is a time when the current snapshot is not available on cloud yet, so we cannot use the latest version automatically
-# This file is managed by an automation (mage integration:UpdateAgentPackageVersion) that check if the snapshot is ready.
-
-STACK_VERSION="$(cat .package-version)"
```
cmacknz, line 15: OK reading everything this does, we actually want to keep this as a way to have a lagging update of the stack version. Since we will bump the agent version from N to N+1 but need to test against stack version N until the first snapshot for N+1 is produced.

Let's keep that part of this.

We should not need to change the actual agent version anymore, the fact that we do this without consequence is a hint we can just relax the version constraints for convenience where necessary. We already did this for minors, but didn't handle majors properly yet.

cmacknz: I would rename the file to .stack-version (and the associated mage target) and remove anything that sets the agent package version (which is the version agent reports itself as to the stack).

pazone: Then the `.stack-version` should also be auto-updated, right?

cmacknz: Yes

pazone: Do I understand it correctly that this PR doesn't make much sense?

cmacknz: I think Paolo's work to allow pinning the stack version will be a better way to specify the stack version. https://github.com/elastic/elastic-agent/pull/7061

We can close this and reevaluate later.

PR #6841, opened 2025-02-13
[k8s] improve kubernetes_secrets provider secret logging

pkoutsovasilis:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...


internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -175,11 +176,13 @@ func (p *contextProviderK8SSecrets) Fetch(key string) (string, bool) {
 	p.store.AddConditionally(key, sd, true, func(existing secret, exists bool) bool {
 		if !exists {
 			// no existing secret in the cache thus add it
+			p.logger.Info("Fetch: \"", key, "\" inserted. Resource Version of secret: \"", apiSecretResourceVersion, "\"")
```
swiatekm, line 179: It would be nicer to use a format string instead I think
```suggestion
			p.logger.Infof("Fetch: \"%s\" inserted. Resource Version of secret: \"%s"\"", key, apiSecretResourceVersion)
```

Same for a few other log lines added in this PR.

swiatekm (approved):
LGTM

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
pkoutsovasilis: No problem changing into Infof but since everything we need to log is already a string, I don't see the point of paying the extra price of calling [fmt.Sprintf](https://github.com/uber-go/zap/blob/6d482535bdd97f4d97b2f9573ac308f1cf9b574e/sugar.go#L378) instead of fmt.Sprint. Does this make sense?

swiatekm: That's a low price for keeping the code more readable imo, but it's not that big a deal. Feel free to ignore this comment if you disagree.

changelog/fragments/1739437836-Improve-kubernetes_secrets-provider-secret-logging.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11:
```suggestion
kind: enhancement
```

blakerouse (commented):
Overall this looks good, just a question about why the use of backslash when golang offers a way to not have to do that.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -175,11 +176,13 @@ func (p *contextProviderK8SSecrets) Fetch(key string) (string, bool) {
 	p.store.AddConditionally(key, sd, true, func(existing secret, exists bool) bool {
 		if !exists {
 			// no existing secret in the cache thus add it
+			p.logger.Infof("Fetch: \"%s\" inserted. Resource Version of secret: \"%s\"", key, apiSecretResourceVersion)
```
blakerouse, line 179: Why not the following?

```suggestion
			p.logger.Infof(`Fetch: "%s" inserted. Resource Version of secret: "%s"`, key, apiSecretResourceVersion)
```

blakerouse (commented):
Missed the inline comment.

blakerouse (approved):
Thanks!

pkoutsovasilis:
@cmacknz if we follow our policies this shouldn't be backported on any active releases branches. But I believe such enhancement will be helpful for users that use `kubernetes_secrets` provider on active release such 8.17.x. WDYT? 🙂

cmacknz:
I would backport this everywhere the secrets provider implementation has changed.

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -136,7 +136,7 @@ func (p *contextProviderK8SSecrets) Fetch(key string) (string, bool) {
 		return "", false
 	}
 	if len(tokens) != 4 {
-		p.logger.Warn("Invalid secret key format: ", key, ". Secrets should be of the format kubernetes_secrets.namespace.secret_name.value")
+		p.logger.Warnf(`Invalid secret key format: "%s". Secrets should be of the format kubernetes_secrets.namespace.secret_name.value`, key)
```
michel-laterman, line 139: fyi `%q` is a formatting directive to quote a string

michel-laterman (approved):

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
pkoutsovasilis: TIL thanks @michel-laterman , I think I am gonna push this change

pkoutsovasilis: just pushed a commit about it, please feel free to unresolve if you spot anything else 🙂

pkoutsovasilis:
So @cmacknz I did get a hands-on experience with it. As of now, you could debug the `kubernetes_secrets` provider by just downloading the diagnostics and looking at the `computed-config.yaml`. This will contain the actual secret values where you can deem if the provider has stale secrets or not. Although this might seem somewhat convenient, IMO this is an issue and we should  change it. I opened [this one](https://github.com/elastic/elastic-agent/issues/6964) as a follow-up.

pkoutsovasilis:
created issue for one new flaky test here https://github.com/elastic/elastic-agent/issues/6965

pkoutsovasilis:
created another issue for a new flaky test [here](https://github.com/elastic/elastic-agent/issues/6967)

pchila (approved):

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
```diff
@@ -151,7 +151,8 @@ func (p *contextProviderK8SSecrets) Fetch(key string) (string, bool) {
 
 	if p.config.DisableCache {
 		// cache disabled - fetch secret from the API
-		return p.fetchFromAPI(ctx, secretName, secretNamespace, secretKey)
+		val, _, ok := p.fetchFromAPI(ctx, secretName, secretNamespace, secretKey)
```
adamkasztenny, line 154: @pkoutsovasilis I'm guessing that adding logging here would add too much noise?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
```diff
@@ -935,6 +937,69 @@ func Test_Config(t *testing.T) {
...
+			name: "key in secret not found",
+			k8sClient: k8sfake.NewClientset(
+				buildK8SSecret("secret_namespace", "secret_name", "secret_key", "secret_value"),
+			),
+			secretName:      "secret_name",
+			secretNamespace: "secret_namespace",
+			secretKey:       "secret_key_not_found",
+		},
+		{
+			name: "key in secret not found",
```
adamkasztenny, line 972: Isn't this the happy path case, where the key _is_ found?

pkoutsovasilis: this is the almost happy path 😆  so the secret is found in the api server, there is no error reading it, but the key inside the secret we are trying to fetch doesn't exist. does it make sense?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
adamkasztenny: Would a debug-level log make sense?

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
adamkasztenny: I must be missing something 😕 IIUC, the key we're looking for is `secret_key` (that's what we're passing to the `fetchFromAPI` function on line 978), and the mock secret we're building also has a key of `secret_key` (line 974). So to me it seems like the key is found

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go
pkoutsovasilis: https://github.com/elastic/elastic-agent/pull/7057 addressed in this PR 🙂

internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets_test.go
pkoutsovasilis: https://github.com/elastic/elastic-agent/pull/7057 addressed in this PR 🙂

PR #6866, opened 2025-02-14
fix install prevention on Linux

intxgo:
## What does this PR do?

`go-sysinfo` determines native architecture based on `/proc` filesystem which could be disabled, therefore the prevention should err on allowing installation when native architecture is unknown.

The fixed prevention was not included in any released version yet.

## Why is it important?

Agent prevents adding Elastic Defend when it cannot determine if it's being emulated. However it's fine to install Elastic Defend in such case, as it does emulation check on it's own upon installation returning installation error when appropriate.

## Checklist

## Related issues

- Relates https://github.com/elastic/go-sysinfo/pull/252

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/elastic/elastic-transport-go/v8 v8.6.0
 	github.com/elastic/go-elasticsearch/v8 v8.17.0
 	github.com/elastic/go-licenser v0.4.2
-	github.com/elastic/go-sysinfo v1.15.0
+	github.com/elastic/go-sysinfo v1.15.1
 	github.com/elastic/go-ucfg v0.8.8
 	github.com/elastic/mock-es v0.0.0-20241101195702-0a41fa3d30d9
 	github.com/elastic/opentelemetry-collector-components/connector/signaltometricsconnector v0.3.0
--- a/go.sum
+++ b/go.sum
@@ -484,8 +484,8 @@ github.com/elastic/go-sfdc v0.0.0-20241010131323-8e176480d727 h1:yuiN60oaQUz2PtN
 github.com/elastic/go-sfdc v0.0.0-20241010131323-8e176480d727/go.mod h1:sw1pzz4pIqzDQxFWt3dFoG2uIUFAfThxlMfWpjH590E=
 github.com/elastic/go-structform v0.0.12 h1:HXpzlAKyej8T7LobqKDThUw7BMhwV6Db24VwxNtgxCs=
 github.com/elastic/go-structform v0.0.12/go.mod h1:CZWf9aIRYY5SuKSmOhtXScE5uQiLZNqAFnwKR4OrIM4=
-github.com/elastic/go-sysinfo v1.15.0 h1:54pRFlAYUlVNQ2HbXzLVZlV+fxS7Eax49stzg95M4Xw=
-github.com/elastic/go-sysinfo v1.15.0/go.mod h1:jPSuTgXG+dhhh0GKIyI2Cso+w5lPJ5PvVqKlL8LV/Hk=
+github.com/elastic/go-sysinfo v1.15.1 h1:zBmTnFEXxIQ3iwcQuk7MzaUotmKRp3OabbbWM8TdzIQ=
+github.com/elastic/go-sysinfo v1.15.1/go.mod h1:jPSuTgXG+dhhh0GKIyI2Cso+w5lPJ5PvVqKlL8LV/Hk=
 github.com/elastic/go-ucfg v0.8.8 h1:54KIF/2zFKfl0MzsSOCGOsZ3O2bnjFQJ0nDJcLhviyk=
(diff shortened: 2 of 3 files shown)


dev-tools/mage/target/common/notice.go
```diff
@@ -99,7 +99,7 @@ func Notice() (err error) {
 	}
 	defer f.Close()
 
-	out, err := os.OpenFile(outfn, os.O_WRONLY|os.O_APPEND, 0644)
+	out, err := os.OpenFile(outfn, os.O_RDWR|os.O_APPEND, 0644)
```
pkoutsovasilis, line 102: why this needs to change? what are you trying to solve here @intxgo

intxgo: the file has inconsistent line endings. Now that I am on windows it writes the new lines with windows endings causing CI diff failure, but when I convert the NOTICE.txt file into unix style there are also some changes because the file convent is inconsistent.

pkoutsovasilis: could you tune your git config to transform CLRF to LF endings on commit?

NOTICE.txt
```diff
@@ -2375,11 +2375,11 @@ Contents of probable licence file $GOMODCACHE/github.com/elastic/go-licenser@v0.
 
 --------------------------------------------------------------------------------
 Dependency : github.com/elastic/go-sysinfo
-Version: v1.15.0
+Version: v1.15.1
 Licence type (autodetected): Apache-2.0
 --------------------------------------------------------------------------------
 
-Contents of probable licence file $GOMODCACHE/github.com/elastic/go-sysinfo@v1.15.0/LICENSE.txt:
+Contents of probable licence file $GOMODCACHE/github.com/elastic/go-sysinfo@v1.15.1/LICENSE.txt:
```
intxgo, line 2382: The rest of the changes here is just a clean-up of the line endings, done by the fix in `notice.go`, you can review it by omitting whitespace from diff
https://github.com/elastic/elastic-agent/pull/6866/files?diff=split&w=1

dev-tools/mage/target/common/notice.go
intxgo: yes, but the problem was that the CI machine didn't produce clean all Unix-style line endings file to compare with the checked-in NOTICE.txt

pkoutsovasilis: it is green because you added the dos2unix. This wasn't required before, why it is needed now, if you push all the file with LF endings to begin with?

kaanyalti (approved):

dev-tools/mage/target/common/notice.go
```diff
@@ -115,5 +115,21 @@ func Notice() (err error) {
 		return fmt.Errorf("failed to append file %s: %w", outfn, err)
 	}
 
+	// dos2unix NOTICE.txt
+	fmt.Printf(">> %s\n", "dos2unix NOTICE.txt")
```
pkoutsovasilis, line 119: opening my comment here so it is clearly visible, I don't believe this step is necessary if you tune your git config properly and make it change CLRF endings to LF, which I believe it asks you to do so by default in windows. @intxgo any reason you think this isn't sufficient?

intxgo: This is a fix for the CI side of NOTICE.txt generation. The build machine in CI is producing NOTICE.txt with inconsistent line endings, so when I commit a file with all Unix-style it won't match. You can download the NOTICE.txt from my PR, it's clean all unix-style so should be green, agree?

pkoutsovasilis: Sorry but on my end, I don't see what you say above @intxgo, as shown below, clearly you pushed a NOTICE.txt with CLRF endings at some point which you tried to fix with dos2unix step you introduced afterwards. 

image

Something that could definitely help in our review is you to rewrite your commit history and initially keep only the commits that are necessary for the actual issue. If then the check-ci step fails because the CI machine produces a NOTICE.txt with CLRF ending you can do one extra commit with your dos2unix step

intxgo: please download the NOTICE.txt from main, and examine it exactly where the diff on my PR points you

image

intxgo: then download the file from my branch and view/compare

pkoutsovasilis: @intxgo AFAICT your PR affects only go-sysinfo, thus I would expect only this thing to be updated in the NOTICE.txt. Again please try to make this easier for the reviewers and tidy up your commit history, as is now it is not clear that the issue you are trying to fix in mage exists indeed. Do one commit with all the changes related to this issue only to make it clear that the CI machine is producing a wrong NOTICE.txt file and then push your dos2unix fix. I don't believe this is much of an ask here

pkoutsovasilis (commented):
code changes look good to me. Waiting for the CI to complete just to strengthen my understanding of `specs/endpoint-security.spec.yml` that we are not breaking any IT (integration tests)

pkoutsovasilis (approved):

PR #6874, opened 2025-02-14
feat(elastic-agent-services): add another tag to docker image for agentless use

olegsu:
## What does this PR do?
This pr will add additional tag to docker image `elastic-agent-service` in form of `elastic-agent-service:git-{SHORT_COMMIT}` 

## Why is it important?
Elastic-Agent is used in Agentless in both ECH and ES3.
While in ECH we must a version that will match the ELK stack, in ES3 we dont have that restriction.
Two reasons why this is important for ES3:
1. Automatic release flow 
2. Continuous feature release

I do not want to elaborate on the details in a public repo, please check the [RFC](https://docs.google.com/document/d/1U0LMCTA1c3eV4O5ADtH5Th-dRRk30zz6b6VJ2avjDoI/edit?tab=t.0#heading=h.w8zo9ze00qw8) and https://github.com/elastic/security-team/issues/9273 with full context 

## Checklist

## Related issues
- https://github.com/elastic/security-team/issues/9273

-

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/dockerbuilder.go
+++ b/dev-tools/mage/dockerbuilder.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"maps"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -53,13 +54,13 @@ func (b *dockerBuilder) Build() error {
 		return fmt.Errorf("failed to prepare build: %w", err)
 	}
 
-	tag, err := b.dockerBuild()
+	tag, additionalTags, err := b.dockerBuild()
 	tries := 3
 	for err != nil && tries != 0 {
 		fmt.Println(">> Building docker images again (after 10 s)")
 		// This sleep is to avoid hitting the docker build issues when resources are not available.
 		time.Sleep(time.Second * 10)
-		tag, err = b.dockerBuild()
+		tag, additionalTags, err = b.dockerBuild()
 		tries--
 	}
 	if err != nil {
@@ -68,6 +69,16 @@ func (b *dockerBuilder) Build() error {
 
 	if err := b.dockerSave(tag); err != nil {
 		return fmt.Errorf("failed to save docker as artifact: %w", err)
+	}
+
+	// additional tags should not be created with
+	for _, tag := range additionalTags {
+		if err := b.dockerSave(tag, map[string]interface{}{
+			// effectively override the name used from b.ImageName() to the tag
+			"Name": strings.ReplaceAll(tag, ":", "-"),
+		}); err != nil {
+			return fmt.Errorf("failed to save docker with tag %s as artifact: %w", tag, err)
+		}
 	}
 
 	return nil
@@ -173,25 +184,43 @@ func (b *dockerBuilder) expandDockerfile(templatesDir string, data map[string]in
 	return nil
 }
 
-func (b *dockerBuilder) dockerBuild() (string, error) {
-	tag := fmt.Sprintf("%s:%s", b.imageName, b.Version)
+// dockerBuild runs "docker build -t t1 -t t2 ... buildDir"
+// returns the main tag additional tags if specified as part of extra_tags property
+// the extra tags are not push to the registry from b.ExtraVars["repository"]
+// returns an error if the command fails
+func (b *dockerBuilder) dockerBuild() (string, []string, error) {
+	mainTag := fmt.Sprintf("%s:%s", b.imageName, b.Version)
 	// For Independent Agent releases, replace the "+" with a "." since the "+" character
 	// currently isn't allowed in a tag in Docker
 	// E.g., 8.13.0+build202402191057 -> 8.13.0.build202402191057
-	tag = strings.Replace(tag, "+", ".", 1)
+	mainTag = strings.Replace(mainTag, "+", ".", 1)
 	if b.Snapshot {
-		tag = tag + "-SNAPSHOT"
+		mainTag = mainTag + "-SNAPSHOT"
 	}
 	if b.FIPS {
-		tag = tag + "-fips"
+		mainTag = mainTag + "-fips"
 	}
 	if repository := b.ExtraVars["repository"]; repository != "" {
-		tag = fmt.Sprintf("%s/%s", repository, tag)
+		mainTag = fmt.Sprintf("%s/%s", repository, mainTag)
 	}
-	return tag, sh.Run("docker", "build", "-t", tag, b.buildDir)
+
+	args := []string{
+		"build",
+		"-t", mainTag,
+	}
+	extraTags := []string{}
+	for _, tag := range b.ExtraTags {
+		extraTags = append(extraTags, fmt.Sprintf("%s:%s", b.imageName, tag))
+	}
+	for _, t := range extraTags {
+		args = append(args, "-t", t)
+	}
+	args = append(args, b.buildDir)
+
+	return mainTag, extraTags, sh.Run("docker", args...)
 }
 
-func (b *dockerBuilder) dockerSave(tag string) error {
+func (b *dockerBuilder) dockerSave(tag string, templateExtraArgs ...map[string]interface{}) error {
 	if _, err := os.Stat(distributionsDir); os.IsNotExist(err) {
 		err := os.MkdirAll(distributionsDir, 0750)
 		if err != nil {
@@ -201,9 +230,13 @@ func (b *dockerBuilder) dockerSave(tag string) error {
 	// Save the container as artifact
 	outputFile := b.OutputFile
 	if outputFile == "" {
-		outputTar, err := b.Expand(defaultBinaryName+".docker.tar.gz", map[string]interface{}{
+		args := map[string]interface{}{
 			"Name": b.imageName,
-		})
+		}
+		for _, extraArgs := range templateExtraArgs {
+			maps.Copy(args, extraArgs)
+		}
+		outputTar, err := b.Expand(defaultBinaryName+".docker.tar.gz", args)
 		if err != nil {
 			return err
 		}
--- a/dev-tools/mage/pkg_test.go
+++ b/dev-tools/mage/pkg_test.go
@@ -20,6 +20,9 @@ func testPackageSpec() PackageSpec {
 		Snapshot: true,
 		OS:       "windows",
 		Arch:     "x86_64",
+		ExtraTags: []string{
+			"git-{{ substring commit 0 12 }}",
+		},
 		Files: map[string]PackageFile{
 			"brewbeat.yml": PackageFile{
 				Source: "./testdata/config.yml",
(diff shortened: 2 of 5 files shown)


dev-tools/mage/dockerbuilder.go
```diff
@@ -173,22 +180,42 @@ func (b *dockerBuilder) expandDockerfile(templatesDir string, data map[string]in
...
 	}
 	if repository := b.ExtraVars["repository"]; repository != "" {
-		tag = fmt.Sprintf("%s/%s", repository, tag)
+		mainTag = fmt.Sprintf("%s/%s", repository, mainTag)
+	}
+
+	moreTags := []string{}
+	if commit, found := b.ExtraVars["commit"]; found && b.DockerVariant == Service {
+		if len(commit) >= 12 {
+			moreTags = append(moreTags, fmt.Sprintf("%s:%s", b.imageName, "git-"+commit[:12]))
```
michalpristas, line 202:
```suggestion
			moreTags = append(moreTags, fmt.Sprintf("%s:git-%s", b.imageName, commit[:12]))
```

dev-tools/mage/dockerbuilder.go
```diff
@@ -173,22 +180,42 @@ func (b *dockerBuilder) expandDockerfile(templatesDir string, data map[string]in
...
+		}
+	}
```
michalpristas, line 204:
```suggestion
	if commit, found := b.ExtraVars["commit"]; found && b.DockerVariant == Service && len(commit) >= 12 {
		moreTags = append(moreTags, fmt.Sprintf("%s:git-%s", b.imageName, commit[:12]))
	}
```

dev-tools/mage/dockerbuilder.go
```diff
@@ -53,23 +53,30 @@ func (b *dockerBuilder) Build() error {
...
 
-	if err := b.dockerSave(tag); err != nil {
+	if err := b.dockerSave(tag, defaultBinaryName); err != nil {
 		return fmt.Errorf("failed to save docker as artifact: %w", err)
 	}
 
+	// additional tags should not be created with
+	for _, tag := range additionalTags {
+		if err := b.dockerSave(tag, "{{.Name}}{{if .OS}}-{{.OS}}{{end}}{{if .Arch}}-{{.Arch}}{{end}}"); err != nil {
+			return fmt.Errorf("failed to save docker as artifact: %w", err)
```
michalpristas, line 76: specify a tag for narrowing down specifics in case of failure

michalpristas (approved):
i think it looks ok.

dev-tools/mage/dockerbuilder.go
michalpristas, line 204: ExtraVars contains raw values, expanded values are inside evalContext, use that one and replace also one for repository above 

```suggestion
	moreTags := []string{}
	if commitI, found := b.evalContext["commit"]; found && b.DockerVariant == Service {
		commit, ok := commitI.(string)
		if ok && len(commit) >= 12 {
			moreTags = append(moreTags, fmt.Sprintf("%s:git-%s", b.imageName, commit[:12]))
		}
	}
```

pkoutsovasilis (commented):
hey @olegsu 👋 thanks for opening this PR. I believe that already @michalpristas has left some comments that capture some issues with this PR (ty @michalpristas ). However, I kinda have another thinking that affects the approach of this PR. I am not a fan of implicit actions, and here the new tag you try to introduce happens ambiguously if commit is found in extra vars. However, IMO, adding an extra vars shouldn't affect the tags as the two are not profoundly coupled. That said, would it make sense to you as well to alternate this PR, promoting explicitness, e.g. introduce `extraTags` in package.yml and do utilise them from there? in this way if in the future we need to change/remove/add another tag this will be really straight-forward

olegsu:
> hey @olegsu 👋 thanks for opening this PR. I believe that already @michalpristas has left some comments that capture some issues with this PR (ty @michalpristas ). However, I kinda have another thinking that affects the approach of this PR. I am not a fan of implicit actions, and here the new tag you try to introduce happens ambiguously if commit is found in extra vars. However, IMO, adding an extra vars shouldn't affect the tags as the two are not profoundly coupled. That said, would it make sense to you as well to alternate this PR, promoting explicitness, e.g. introduce `extraTags` in package.yml and do utilise them from there? in this way if in the future we need to change/remove/add another tag this will be really straight-forward

Thank you for the feedback! 
It makes sense, let me update the pr

blakerouse (approved):

blakerouse:
Please wait for approval from @pkoutsovasilis before merging, but this looks good to me with the latest updates.

jlind23:
@blakerouse Thanks for the approval but I am a bit confused. Didn't we agreed on giving approval only once CI is green to avoid code changes to make CI happy after approval that would not dismiss your review?

dev-tools/mage/dockerbuilder.go
```diff
@@ -53,23 +53,30 @@ func (b *dockerBuilder) Build() error {
...
 	}
 
-	if err := b.dockerSave(tag); err != nil {
+	if err := b.dockerSave(tag, defaultBinaryName); err != nil {
 		return fmt.Errorf("failed to save docker as artifact: %w", err)
 	}
 
+	// additional tags should not be created with
+	for _, tag := range additionalTags {
+		if err := b.dockerSave(tag, "{{.Name}}{{if .OS}}-{{.OS}}{{end}}{{if .Arch}}-{{.Arch}}{{end}}"); err != nil {
```
pkoutsovasilis, line 75: `b.dockerSave(tag, "{{.Name}}{{if .OS}}-{{.OS}}{{end}}{{if .Arch}}-{{.Arch}}{{end}}")` if we add multiple extra tags going with this static name won't we overwrite the archives of the different tags? Should we utilise instead `fmt.Sprintf("%s-%s", defaultBinaryName, tag)`

olegsu: Should we avoid at all the OS and arch in the name? I am not familiar with the unified release internals, but it seems that this will be executed twice ( [amd64](https://github.com/elastic/elastic-agent/blob/88d1289aa03bd79bb415233765e6400c2fadf394/dev-tools/packaging/packages.yml#L1028), [arm](https://github.com/elastic/elastic-agent/blob/88d1289aa03bd79bb415233765e6400c2fadf394/dev-tools/packaging/packages.yml#L1042) ) and one will get overridden on the filesystem. 
To avoid duplication we can add the tag into the name `{.Name}}-{{ .AdditionalTag }}{{if .OS}}-{{.OS}}{{end}}{{if .Arch}}-{{.Arch}}{{end}}`
And later on
```go
func (b *dockerBuilder) dockerSave(tag string, binaryNameTemplate string) error {
  outputTar, err := b.Expand(binaryNameTemplate+".docker.tar.gz", map[string]interface{}{
	  "Name":          b.imageName,
	  "AdditionalTag": tag,
  })
}
```

The remaining unknown for me is will this artifact result in the DRA so the unified release can pick it and publish the image?

pkoutsovasilis: so `	defaultBinaryName = "{{.Name}}{{if .Qualifier}}-{{.Qualifier}}{{end}}-{{.Version}}{{if .Snapshot}}-SNAPSHOT{{end}}{{if .OS}}-{{.OS}}{{end}}{{if .Arch}}-{{.Arch}}{{end}}"` it already contains the OS and the ARCH. Essentially you need to leave this format for the default image we produce and then for every tag in the extraTags you need to save the image with the tag init, thus no archive overwriting will happen.

>The remaining unknown for me is will this artifact result in the DRA so the unified release can pick it and publish the image?

I think it won't be picked up automatically and just to be on the safe side you should reach out to somebody from the release-eng team and ask about it 🙂

pkoutsovasilis: @olegsu this `[elastic-agent-service-elastic-agent-service](elastic-agent-service-elastic-agent-service:git-5a12ef920669-9.1.0-SNAPSHOT-linux-arm64.docker.tar.gz)` seems kinda off right?! tbh I was expecting something like `elastic-agent-service-git-5a12ef920669-9.1.0-SNAPSHOT-linux-arm64.docker.tar.gz`, does that make sense also to you? 🤔

olegsu: Right, missed the semicolon escaping

pkoutsovasilis: I still see the archive name containing two times the service name in the archive `elastic-agent-service-elastic-agent-service` which I would expect to see only one.

pkoutsovasilis: after another look, the container image that gets produced from your example above @olegsu is this
```
❯ docker images
REPOSITORY              TAG                IMAGE ID       CREATED         SIZE
elastic-agent-service   git-5a12ef920669   f8fd4bc26da2   3 minutes ago   2.84GB
```

shouldn't it be

```
❯ docker images
REPOSITORY              TAG                IMAGE ID       CREATED         SIZE
docker.elastic.co/beats-ci/elastic-agent-service   git-5a12ef920669   f8fd4bc26da2   3 minutes ago   2.84GB
```
?

olegsu: > after another look, the container image that gets produced from your example above @olegsu is this

Based on [log](https://buildkite.com/elastic/unified-release-snapshot/builds/5016#_) (to see the log, download the whole file) sample from the [unified-release-snapshot pipeline](https://github.com/elastic/unified-release/blob/ae4d5979642a6b19eaf7ed523e058b1c45787e83/catalog-info.yaml#L92), the image name that was built is not used.


The step `Push container images - Elastic-Agent-Package` runs the [script](https://github.com/elastic/unified-release/blob/ae4d5979642a6b19eaf7ed523e058b1c45787e83/scripts/push-multi-arch-container-image.sh#L12) `push-multi-arch-container-image.sh` with `elastic-agent-service` as a parameter, not the the full image name:
`scripts/push-container-image.sh https://snapshots.elastic.co/9.1.0-d86fb820/downloads/beats/elastic-agent/elastic-agent-service-9.1.0-SNAPSHOT-docker-image-linux-arm64.tar.gz elastic-agent-service 9.1.0-SNAPSHOT-arm64 cloud-release

Eventually pushed to as `docker.elastic.co/cloud-release/elastic-agent-service:9.1.0-SNAPSHOT-arm64`
So I assume the same will happen with any extra tag.


> I still see the archive name containing two times the service name in the archive elastic-agent-service-elastic-agent-service which I would expect to see only one.

Agree, checking how to remove it nicely

olegsu: Update, now we dont have another template as it overrides the `Name` value

pkoutsovasilis (approved):
Thanks for going through my comments @olegsu this LGTM. I had a look in the CI failures and they all seem to be related to already reported flaky tests and a new one which I have raised with eng prod team. That said, I invoked retries to get this PR merged 🙂

ingress-gateway-ca.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
pkoutsovasilis, line 1: I believe this file is not required, sorry I need to request changes until this file is removed

PR #6893, opened 2025-02-17
ci(buildkite): use go-version env variable in the pipeline.yml

v1v:
## What does this PR do?

Use https://github.com/elastic/elastic-agent/blob/08048a900da9e73f782c6ae65b034cf45ae1a05b/.buildkite/hooks/pre-command#L12-L14 to avoid a hardcoded value

## Why is it important?

Otherwise, the updatecli pipeline will need to be adapted:
- https://github.com/elastic/elastic-agent/blob/main/.ci/updatecli/updatecli-bump-golang.yml

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -196,7 +196,8 @@ steps:
     artifact_paths:
       - "build/TEST-go-unit.cov"
     agents:
-      image: "golang:1.22.6"
+      # NOTE: use the go-version env variable that's set in the pre-command hook
+      image: "golang:${GO_VERSION}"
     depends_on:
       - unit-tests
       - extended-windows


jlind23 (dismissed):

.buildkite/pipeline.yml
```diff
@@ -196,7 +196,8 @@ steps:
     artifact_paths:
       - "build/TEST-go-unit.cov"
     agents:
-      image: "golang:1.22.6"
+      # NOTE: use the go-version env variable that's set in the pre-command hook
+      image: "golang:${GO_VERSION}"
```
dliappis, line 200: I don't think variables in the precommand hook get evaluated here (only after an agent has started running in the context of a shell)

v1v: That's indeed an excellent point!

I guess I'll need to use updatecli for this file then

PR #6907, opened 2025-02-18
Enhancement/6394 allow deb rpm to upgrade with endpoint tamper protection

### LINKED ISSUE #6394: Allow DEB and RPM packages to supply an uninstall token when upgrading with Elastic Defend installed
- Relates https://github.com/elastic/elastic-agent/issues/6356
- Relates https://github.com/elastic/elastic-agent/issues/4720

**Describe the enhancement:**

For Fleet managed agents that are running Elastic Defend, upgrading the RPM or DEB packages does not send a signed upgrade action to Defend which is what allows Defend to upgrade without believing it is being tampered with. The tamper protection feature of Defend is based on all operations requiring a digital signature signed by a private key in Kibana. There is no way to provide a signed token for DEB or RPM upgrades, so Defend legitimately believes it was being tampered with and orphaned from agent here. 

Specifically using the CLI skips this block https://github.com/elastic/elastic-agent/blob/54932dc7b46cdd0d16791e07b8fc5dd596efb129/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L58-L74

To make the upgrade command work with tamper protection properly we'd have to allow it to accept an uninstall token to allow endpoint to unprotect itself for a CLI upgrade.

Upgrading an Elastic Agent installed with either the RPM or DEB packages when Elastic Defend is used. Without this change, the only way to upgrade in this situation is to temporarily disable tamper protection.

- Tests exists proving that the RPM and DEB packages can be upgraded when Elastic Defend is installed without disabling tamper protection.

**The two bullet points below are not relevant anymore as the user does not have to provide an uninstall token**
- ~~These packages types can ideally indicate to users that they should supply an uninstall token when defend is installed.~~
- ~~The documentation for the RPM and DEB packages is updated to indicate that an uninstall token is required in this situation.~~

kaanyalti:
- Enhancement

## What does this PR do?

Updates the preinstall script for deb and rpm to stop the endpoint security service and remove the vault

## Why is it important?

Currently when users upgrade their deb or rpm agents using their respective package managers, if there is endpoint running and if endpoint has tamper protection enabled, endpoint will become unhealthy. This PR fixes it by replicating what `enpoint-security uninstall --uninstall-token <token>`  does in the deb/rpm preinstall script.

## Checklist

## Disruptive User Impact

There shouldn't be anything disruptive for the users.

- Start [endpoint compatible](https://www.elastic.co/support/matrix) VMs, one for deb one for rpm
- Build the agent for these two platforms
- Create an ESS deployment
- Create a policy, add endpoint integration
- Follow the "Add Agent" instructions and install the version that is shown on the instructions.
- Validate that both the agent and endpoint are healthy
- Upgrade the agent running either `sudo apt install <Agent that you built>` or `sudo dnf install <Agent that you built>`
- After the upgrade is done, validate that both the agent and the endpoint are healthy
- Just to be sure, validate that the new endpoint is still tamper protected
  - Try to uninstall agent without a token and see it fail
  - Try to uninstall agent with an uninstall token and validate that it is uninstalled
  - Beware that removing the agent using the package managers will indeed uninstall the agent but will leave endpoint. Use the elastic-agent command to uninstall.

Repeat the steps above but test same version upgrades.
- Install the agent you packaged from the branch.
- Enroll into fleet
- Install the same agent again with `dpkg -i` or `rpm -Uvh --force`
- Check the status of the agent and endpoint
- Validate that tamper protection is enabled

Note: When testing you may run into the following error 
```
┌─ fleet
│  └─ status: (STARTING)
└─ elastic-agent
   ├─ status: (DEGRADED) 1 or more components/units in a failed state
   └─ endpoint-default
      ├─ status: (FAILED) Elastic Defend cannot run on an emulated architecture
      ├─ endpoint-default
      │  └─ status: (FAILED) Elastic Defend cannot run on an emulated architecture
      └─ endpoint-default-b73c5b0f-3b0a-4877-a531-35fdd9f9204f
         └─ status: (FAILED) Elastic Defend cannot run on an emulated architecture
```
This is related to the following bug https://github.com/elastic/elastic-agent/pull/6866 

## Related issues

- Closes #6394 
- Relates #6866


pkg/testing/fixture_install.go
```diff
@@ -445,6 +445,58 @@ func getProcesses(t *gotesting.T, regex string) []runningProcess {
...
+func (f *Fixture) simpleInstallRPM(ctx context.Context) ([]byte, error) {
+	f.t.Logf("[test %s] Inside fixture simpleInstallRPM function", f.t.Name())
+	// Prepare so that the f.srcPackage string is populated
+	err := f.EnsurePrepared(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare: %w", err)
+	}
+
+	// sudo apt-get install the deb
+	out, err := exec.CommandContext(ctx, "sudo", "rpm", "-Uvh", f.srcPackage).CombinedOutput() // #nosec G204 -- Need to pass in name of package
```
pchila, line 468: Is the difference between this and the existing `installRpm` function just that the new function upgrades and doesn't enroll the agent ?

pkg/testing/fixture_install.go
```diff
@@ -445,6 +445,58 @@ func getProcesses(t *gotesting.T, regex string) []runningProcess {
...
+func (f *Fixture) simpleInstallDeb(ctx context.Context) ([]byte, error) {
+	f.t.Logf("[test %s] Inside fixture simpleInstallDeb function", f.t.Name())
+	// Prepare so that the f.srcPackage string is populated
+	err := f.EnsurePrepared(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare: %w", err)
+	}
+
+	// sudo apt-get install the deb
+	out, err := exec.CommandContext(ctx, "sudo", "DEBIAN_FRONTEND=noninteractive", "apt-get", "-y", "-q", "-o", "Dpkg::Options::=--force-confdef", "-o", "Dpkg::Options::=--force-confold", "install", f.srcPackage).CombinedOutput() // #nosec G204 -- Need to pass in name of package
```
pchila, line 488: here we are installing rather than upgrading but it still has a lot of overlap with `installDeb` (minus the enrollment)

pkg/testing/tools/tools.go
```diff
@@ -35,6 +35,25 @@ func IsPolicyRevision(ctx context.Context, t *testing.T, client *kibana.Client,
 	}
 }
 
+func GetandSetUninstallTokens(ctx context.Context, t *testing.T, kibClient *kibana.Client, fixture *atesting.Fixture, policyId string) error {
```
pchila, line 38: Maybe a clearer name `SetPolicyUninstallTokenInFixture` would help readability :sweat_smile:

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +61,212 @@ var protectionTests = []struct {
...
+	})
+	testTamperProtectedDebRpmUpgrades(t, info, "rpm")
+}
+
+func getEndpointVersion(t *testing.T) string {
+	cmd := exec.Command("sudo", "/opt/Elastic/Endpoint/elastic-endpoint", "version")
+	output, err := cmd.CombinedOutput()
+	require.NoError(t, err)
+	// version: 8.18.0-SNAPSHOT, compiled: Wed Feb 19 01:00:00 2025, branch: HEAD, commit: c450b50f91507c3166b072df8557f5efd871103a
+	return strings.Split(strings.Split(string(output), "version: ")[1], ",")[0]
```
pchila, line 100: What about something like this (optionally you can check for `found` return value:
```suggestion
	endpointVersionFragment, _, _ := strings.Cut(string(output), ",")
	endpointVersion, _ := strings.CutPrefix(endpointVersionFragment, "version: ")
```

pkg/testing/fixture_install.go
kaanyalti: Yes exactly, this one just upgrades and does not enroll

pkg/testing/fixture_install.go
kaanyalti: What I had implemented at first was I was using this new install function in the installDeb that is already there. Just because with rpm I had to call it with upgrade flags, I decided to completely separate the simpleInstall and regular install functions completely thinking that it would lead to a less convoluted code. I am open to suggestions here though.

pkg/testing/tools/tools.go
kaanyalti: Yes, I agree, thank you

pkg/testing/tools/tools.go
```diff
@@ -35,6 +35,25 @@ func IsPolicyRevision(ctx context.Context, t *testing.T, client *kibana.Client,
 	}
 }
 
+func SetPolicyUninstallTokenInFixture(ctx context.Context, t *testing.T, kibClient *kibana.Client, fixture *atesting.Fixture, policyId string) error {
+	resp, err := kibClient.GetPolicyUninstallTokens(ctx, policyId)
+	if err != nil {
+		return fmt.Errorf("failed to fetch uninstal tokens: %w", err)
```
pchila, line 41:
```suggestion
		return fmt.Errorf("failed to fetch uninstall tokens: %w", err)
```

pkg/testing/tools/tools.go
```diff
@@ -53,22 +72,9 @@ func InstallAgentWithPolicy(ctx context.Context, t *testing.T,
 	}
 
 	if createPolicyReq.IsProtected {
-		// If protected fetch uninstall token and set it for the fixture
-		resp, err := kibClient.GetPolicyUninstallTokens(ctx, policy.ID)
-		if err != nil {
-			return policy, fmt.Errorf("failed to fetch uninstal tokens: %w", err)
+		if err := SetPolicyUninstallTokenInFixture(ctx, t, kibClient, agentFixture, policy.ID); err != nil {
```
pchila, line 75: Is there a reason why we ignore the returned error here? In line 78 `err` is overwritten by the result of the `InstallAgentForPolicy` call

kaanyalti: After calling SetPolicyUninstallTokenInFixture, the function returns the policy and the error, if error is not nil. So execution does not reach InstallAgentForPolicy. I may be missing something though.

pchila: Oh I missed the `return err` clause in line 76, closing this...

pkg/testing/fixture_install.go
pchila: In principle I am not against a bit of duplication if there's a good reason for it and it seems that you have one. As a first iteration it is ok to keep the 2 functions, it can always be refactored later.
Could you please add a comment on top of the function explaining the rationale for it so that in 6 months time we can still figure out why the code is written in this way ? ;)

pchila (commented):
Code looks good, however there are still some failures in CI about `TestUpgradeAgentWithTamperProtectedEndpoint_RPM` since we try to upgrade from version 9.0.0-SNAPSHOT to 9.0.0-SNAPSHOT and `rpm` is not happy about it.
This is not a problem when installing using `elastic-agent upgrade/install` commands but here we are missing something I guess.

@cmacknz could you please chime in here: are we supposed to support rpm upgrades between two builds of 9.0.0-SNAPSHOT ?

```
    fixture_install.go:460: [test TestUpgradeAgentWithTamperProtectedEndpoint_RPM] Inside fixture simpleInstallRPM function
    endpoint_security_test.go:233:
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1740982239855044947/elastic/elastic-agent/testing/integration/endpoint_security_test.go:233
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1740982239855044947/elastic/elastic-agent/testing/integration/endpoint_security_test.go:92
        	            	apt install failed: exit status 1 output:Verifying...                          ########################################
        	            	Preparing...                          ########################################
        	            		package elastic-agent-9.0.0-1.x86_64 is already installed
        	            		file /etc/elastic-agent/.elastic-agent.active.commit from install of elastic-agent-9.0.0-1.x86_64 conflicts with file from package elastic-agent-9.0.0-1.x86_64
        	            		file /lib/systemd/system/elastic-agent.service from install of elastic-agent-9.0.0-1.x86_64 conflicts with file from package elastic-agent-9.0.0-1.x86_64
        	            		file /usr/share/elastic-agent/.build_hash.txt from install of elastic-agent-9.0.0-1.x86_64 conflicts with file from package elastic-agent-9.0.0-1.x86_64
        	            		file /usr/share/elastic-agent/NOTICE.txt from install of elastic-agent-9.0.0-1.x86_64 conflicts with file from package elastic-agent-9.0.0-1.x86_64
        	Test:       	TestUpgradeAgentWithTamperProtectedEndpoint_RPM
    fixture.go:1286: Temporary directory "/tmp/TestUpgradeAgentWithTamperProtectedEndpoint_RPM3576483495" preserved for investigation/debugging
```

pkg/testing/fixture_install.go
kaanyalti: I reverted back to using PreviousMinor. When locally triggering tests, it fixed this particular error. Now hopefully tests will pass.

kaanyalti: Added a comment for `SimpleInstall` explaining the situation

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -2,6 +2,57 @@
...
+    cleanup() {
+        rm -f "$cleanup_state"
+        rm -rf "$temp_vault"
+    }
+
+    # restore endpoint and vault if installation fails
+    restore_endpoint() {
+        if [ -f "$cleanup_state" ]; then
+            # Check if we need to restore the vault
+            if grep -q "vault_removed" "$cleanup_state"; then
```
cmacknz, line 21: Can endpoint function if we don't restore the vault? Wondering if on failure it's less risky to just run `systemctl start ElasticEndpoint` with the goal of just restoring endpoint at all instead of risking that is left not running.

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +61,217 @@ var protectionTests = []struct {
...
+				Type:   define.Linux,
+				Distro: "rhel",
+			},
+		},
+	})
+	testTamperProtectedDebRpmUpgrades(t, info, "rpm")
+}
+
+func getEndpointVersion(t *testing.T) string {
+	cmd := exec.Command("sudo", "/opt/Elastic/Endpoint/elastic-endpoint", "version")
```
cmacknz, line 96: Does endpoint not send us it's version on the control protocol? This should be in the output of `elastic-agent status --output=json` for us to parse but I see it's missing.

```json
        {
            "id": "endpoint-default",
            "name": "endpoint",
            "state": 2,
            "message": "Healthy: communicating with endpoint service",
            "units": [
                {
                    "unit_id": "endpoint-default",
                    "unit_type": 1,
                    "state": 2,
                    "message": "Applied policy {d6a361f2-aa77-4a84-ad0f-9249873f3293}",
                    "payload": {
                        "error": {
                            "code": 0,
                            "message": "Success"
                        }
                    }
                },
                {
                    "unit_id": "endpoint-default-d6a361f2-aa77-4a84-ad0f-9249873f3293",
                    "unit_type": 0,
                    "state": 2,
                    "message": "Applied policy {d6a361f2-aa77-4a84-ad0f-9249873f3293}",
                    "payload": {
                        "error": {
                            "code": 0,
                            "message": "Success"
                        }
                    }
                }
            ],
            "version_info": {
                "name": "Endpoint"
            }
        },
```

It somehow is in the state.yaml of diagnostics though:

```yaml
        version_info:
            build_hash: 72f40668ce48bcd9911ff7d402cad2f5c807d944
            name: Endpoint
```

This suggests it'd be possible to pull this out of the status command instead of directly talking to endpoint here. Worth a quick check to see if we can avoid needing new code to do this.

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +61,217 @@ var protectionTests = []struct {
...
+		"Endpoint component or units are not healthy prior to upgrade.",
+	)
+
+	t.Log("The initial installation of both the agent and endpoint are healthy")
+
+	initEndpointVersion := getEndpointVersion(t)
+	t.Logf("The initial endpoint version is %s", initEndpointVersion)
+
+	// try to uninstall endpoint without token and assert that endpoint is not removed
+	_, err = exec.Command("sudo", "/opt/Elastic/Endpoint/elastic-endpoint", "uninstall", "--log", "stdout").CombinedOutput()
```
cmacknz, line 221: Shouldn't this be trying to uninstall the deb/rpm and ensuring that actually gets blocked? We are testing agents packaging, not endpoint itself.

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +61,217 @@ var protectionTests = []struct {
...
+			return
+		}
+
+		out, err := exec.Command("sudo", "systemctl", "stop", "ElasticEndpoint").CombinedOutput()
+		if err != nil {
+			t.Log(string(out))
+			t.Logf("error while stopping Elastic Endpoint: %s", err.Error())
+		}
+
+		if atesting.KeepInstalledFlag() {
```
cmacknz, line 187: You don't set this in the test, do you need to this logic or can you just remove it. Unless I'm missing something this is dead code.

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +61,217 @@ var protectionTests = []struct {
...
+	tools.InstallAgentForPolicyWithToken(ctx, t, opts, startFixture, info.KibanaClient, policyResp.ID, enrollKeyResp)
+
+	t.Cleanup(func() {
+		_, err := os.Stat("/opt/Elastic/Endpoint/elastic-endpoint")
+		if os.IsNotExist(err) {
+			t.Log("Endpoint binary does not exist, aborting endpoint cleanup")
+			return
+		}
+
+		out, err := exec.Command("sudo", "systemctl", "stop", "ElasticEndpoint").CombinedOutput()
```
cmacknz, line 181: Again, why not just uninstall agent via the deb/rpm? 

If you aren't uninstall the agent, agent is going to install endpoint again immediately after you do this if endpoint is still in the policy.

pkg/testing/fixture_install.go
cmacknz: IMO just move this out of the fixture and directly into your test, then you don't need to explain it in a general way.

The only things here that depends on being a method on Fixture is creating and setting the protocol client. Just add a public method for that.

pkg/testing/fixture.go
```diff
@@ -261,6 +261,10 @@ func (f *Fixture) SetUninstallToken(uninstallToken string) {
 	f.uninstallToken = uninstallToken
 }
 
+func (f *Fixture) UninstallToken() string {
```
cmacknz, line 264: This needs to document that you need to call SetPolicyUninstallTokenInFixture first or it does nothing.

It seems like you could just have a function to get an uninstall token and then hold onto it in your tests as a test local variable. I'm not following the reason it needs to be part of the fixture. You have a method to inject it and then extract it out again, the fixture itself doesn't actually set this variable.

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
cmacknz: You aren't testing the recovering path so simplifying to something that can be viewed as correct by inspection (e.g. single call to systemctl start) seems better here. Otherwise, we should look at testing that this actually works.

pchila:
@kaanyalti  could you please fix conflicts and reply to @cmacknz comments ?

testing/integration/endpoint_security_test.go
kaanyalti: Using deb/rpm does make sense as well. I had just focused on using the package managers for upgrades as that's what the PR is about. I can switch to using package managers instead of endpoint.

testing/integration/endpoint_security_test.go
kaanyalti: You're right, I'm not using this in the test. I added this so that endpoint does not get cleaned up for debugging purposes and the behaviour is consistent with that of the agent. Otherwise the cleanup func removes endpoint regardless of the `keep installed` flag.

testing/integration/endpoint_security_test.go
kaanyalti: I tried using deb/rpm and had a problem because the agent gets removed but endpoint remains if uninstalled with deb/rpm. Uninstalling with `sudo elastic-agent uninstall --uninstall-token` and then purging package metadata works though, so I'm going to implement that.

pkg/testing/fixture.go
kaanyalti: You're right, I had just refactored what was already implemented. I wanted to be able to execute certain commands on their own. Instead of writing them from scratch I refactored pieces from functions that are already implemented. 

That being said I can definitely just fetch it once and store it.

testing/integration/endpoint_security_test.go
kaanyalti: Uninstalling the agent via deb/rpm unfortunately does not remove endpoint. After removing agent, endpoint remains running. That's why I had to uninstall agent with `sudo elastic-agent uninstall --uninstall-token` first, then remove package metadata. The code here is to cleanup endpoint before agent cleanup executes otherwise endpoint will remain running after the cleanup functions.

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
kaanyalti: Added test case for the recovery. Additionally simplified the recovery logic.

pkg/testing/fixture.go
kaanyalti: Removed UninstallToken(). As suggested I'm now fetching the token once and storing it in the test function.

changelog/fragments/1740166208-allow-deb-rpm-upgrade-with-tamper-protected-endpoint.yaml
```diff
@@ -0,0 +1,30 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Updated the preinstall script to allow users to upgrade deb or rpm agents when using Elastic Defend with tamper protection.
```
pchila: we can maybe keep it short and sweet (it's a proposal, not a must-do)  :wink: 
```suggestion
summary: Allow upgrading deb or rpm agents when using Elastic Defend with tamper protection.
```

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -6,13 +6,44 @@ commit_hash="{{ commit_short }}"
...
 # delete $symlink if exists
 if test -L "$symlink"; then
     echo "found symlink $symlink, unlink"
     unlink "$symlink"
 fi
 
+SERVICE_NAME="ElasticEndpoint"
+
+echo "Checking if ElasticEndpoint is installed"
+if systemctl list-unit-files --type=service | grep -q "^${SERVICE_NAME}.service"; then
```
michalpristas, line 20: i'm not sure, do we support any other init system?

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -6,13 +6,44 @@ commit_hash="{{ commit_short }}"
...
+    archive_endpoint_version="$("$new_endpoint_component_bin" version)"
+
+    echo "ElasticEndpoint is installed"
+    if ! systemctl is-active --quiet "$SERVICE_NAME"; then
+        echo "$SERVICE_NAME is installed but not running"
+        if [[ "$installed_endpoint_version" == "$archive_endpoint_version" ]]; then
+            echo "New endpoint version is the same as the one that's already
+            installed"
+            echo "Starting $SERVICE_NAME"
+            sudo systemctl start ElasticEndpoint
```
pchila: let's be consistent in using the variables
```suggestion
            sudo systemctl start ${SERVICE_NAME}
```

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -6,13 +6,44 @@ commit_hash="{{ commit_short }}"
...
 
+SERVICE_NAME="ElasticEndpoint"
+
+echo "Checking if ElasticEndpoint is installed"
+if systemctl list-unit-files --type=service | grep -q "^${SERVICE_NAME}.service"; then
+
+    installed_endpoint_version="$(/opt/Elastic/Endpoint/elastic-endpoint version)"
+    archive_endpoint_version="$("$new_endpoint_component_bin" version)"
+
+    echo "ElasticEndpoint is installed"
```
pchila:
```suggestion
    echo "${SERVICE_NAME} is installed"
```

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -6,13 +6,44 @@ commit_hash="{{ commit_short }}"
...
+    if ! systemctl is-active --quiet "$SERVICE_NAME"; then
+        echo "$SERVICE_NAME is installed but not running"
+        if [[ "$installed_endpoint_version" == "$archive_endpoint_version" ]]; then
+            echo "New endpoint version is the same as the one that's already
+            installed"
+            echo "Starting $SERVICE_NAME"
+            sudo systemctl start ElasticEndpoint
+        else
+            echo "New endpoint version is different than the one that's already
+            installed"
```
pchila: Nit: we can add a bit more info about versions here
```suggestion
            echo "New endpoint version  \"${archive_endpoint_version}\" is different than the one that's already
            installed  \"${installed_endpoint_version}\""
```

pkg/testing/fixture_install.go
```diff
@@ -593,6 +603,10 @@ func (f *Fixture) installRpm(ctx context.Context, installOpts *InstallOpts, shou
 		return out, fmt.Errorf("systemctl start elastic-agent failed: %w", err)
 	}
 
+	socketPath := "unix:///var/lib/elastic-agent/elastic-agent.sock"
```
michalpristas: isnt this implemented in SetClient?

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -6,13 +6,44 @@ commit_hash="{{ commit_short }}"
...
+
+    installed_endpoint_version="$(/opt/Elastic/Endpoint/elastic-endpoint version)"
+    archive_endpoint_version="$("$new_endpoint_component_bin" version)"
+
+    echo "ElasticEndpoint is installed"
+    if ! systemctl is-active --quiet "$SERVICE_NAME"; then
+        echo "$SERVICE_NAME is installed but not running"
+        if [[ "$installed_endpoint_version" == "$archive_endpoint_version" ]]; then
+            echo "New endpoint version is the same as the one that's already
+            installed"
```
pchila:
```suggestion
            echo "New endpoint and installed endpoint versions are the same: \"${installed_endpoint_version}\""
```

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -2,6 +2,44 @@
 
 set -e
 
+endpoint_active=false
+
+handle_endpoint() {
+  echo "Checking if ElasticEndpoint is an active service"
+	if systemctl --quiet is-active ElasticEndpoint; then
```
pchila: We could use a SERVICE_NAME variable like in postinstall script

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -2,6 +2,44 @@
...
+
+			rm -rf "/opt/Elastic/Endpoint/state/vault"
+
+      echo "Stopping ElasticEndpoint"
+		  systemctl --quiet stop ElasticEndpoint
+    else
+      echo "ElasticEndpoint tamper protection is not enabled"
+		fi
+	fi
+}
```
pchila: Nit: would it be possible to fix the indentation of this block ? It would improve readability (a bit)

pkg/testing/tools/tools.go
```diff
@@ -36,6 +36,22 @@ func IsPolicyRevision(ctx context.Context, t *testing.T, client *kibana.Client,
 	}
 }
 
+func GetUninstallToken(ctx context.Context, kibClient *kibana.Client, policyId string) (string, error) {
+	resp, err := kibClient.GetPolicyUninstallTokens(ctx, policyId)
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch uninstall tokens: %w", err)
+	}
+	if len(resp.Items) == 0 {
+		return "", fmt.Errorf("expected non-zero number of tokens: %w", err)
```
pchila: In this instance `err == nil` so we shouldn't wrap a nil error 
```suggestion
		return "", fmt.Errorf("expected non-zero number of tokens")
```

pkg/testing/tools/tools.go
```diff
@@ -36,6 +36,22 @@ func IsPolicyRevision(ctx context.Context, t *testing.T, client *kibana.Client,
...
+	}
+
+	if len(resp.Items[0].Token) == 0 {
+		return "", fmt.Errorf("expected non-empty token: %w", err)
```
pchila:
```suggestion
		return "", fmt.Errorf("expected non-empty token")
```

pchila (commented):
Still need to look at a couple of files, these are the comments so far...

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +62,452 @@ var protectionTests = []struct {
...
+
+	if len(envVars) != 0 {
+		args = append(args, envVars...)
+	}
+
+	switch packageFormat {
+	case "deb":
+		args = append(args, "dpkg", "-i")
+	case "rpm":
+		args = append(args, "rpm", "-Uvh", "--force")
```
pchila, line 156: Is there a specific reason why we have to specify `--force` on the rpm install command ? Is that for replacing an installed package with the same version ?

pchila (commented):
Just one more question about forcing rpm installation on the rest of files

testing/integration/endpoint_security_test.go
kaanyalti: Yes exactly. I'm reinstalling an already installed package version in one of the test cases, and `--force` is for that.

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
kaanyalti: I checked the [support matrix](https://www.elastic.co/support/matrix) and found this [support case](https://discuss.elastic.co/t/elastic-agent-fails-under-sysvinit-due-to-dying-endpoint-security/257921/2) which suggest all distributions supported by endpoint use systemd.

In addition to this the deb/rpm scripts already have `systemctl` calls, for example [these lines](https://github.com/elastic/elastic-agent/pull/6907/files#diff-44f36381f1d27390fa33105c4cb4adb73800d433e78cfddd20aceeb4fe892018L24-L26) in the postinstall script.

cmacknz: Just systemd is fine

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
cmacknz: You might want to include `${installed_endpoint_version}` in the output here as well so that you know what the output of both command was e.g. `ElasticEndpoint was installed at version 9.0.2`

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
```diff
@@ -6,13 +6,44 @@ commit_hash="{{ commit_short }}"
...
+            echo "Starting $SERVICE_NAME"
+            sudo systemctl start ${SERVICE_NAME}
+        else
+            echo "New endpoint version  \"${archive_endpoint_version}\" is different than the one that's already
+            installed  \"${installed_endpoint_version}\""
+        fi
+        ls -li /opt/Elastic/Endpoint/elastic-endpoint
```
cmacknz: This statement has no effect right? Leftover from debugging?

pkg/testing/fixture_install.go
```diff
@@ -502,6 +508,10 @@ func (f *Fixture) installDeb(ctx context.Context, installOpts *InstallOpts, shou
...
```
cmacknz: This is the same as the `SetClient` method above that could be called here.

pkg/testing/fixture_install.go
```diff
@@ -445,6 +445,12 @@ func getProcesses(t *gotesting.T, regex string) []runningProcess {
 	return processes
 }
 
+func (f *Fixture) SetClient() {
+	socketPath := "unix:///var/lib/elastic-agent/elastic-agent.sock"
```
cmacknz: It looks like there is a function to determine the client address you can reuse 

https://github.com/elastic/elastic-agent/blob/77669137ea1f9c0c55244e1a03a22ee04ede3aa1/pkg/testing/fixture.go#L534-L537

https://github.com/elastic/elastic-agent/blob/77669137ea1f9c0c55244e1a03a22ee04ede3aa1/pkg/testing/fixture.go#L566-L572

dev-tools/packaging/templates/linux/postinstall.sh.tmpl
kaanyalti: Correct, used it for debugging, removing it

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -2,6 +2,45 @@
...
+    }
+
+    trap restore_endpoint ERR
+
+    # Remove the vault directory if it exists
+    if [ -d "/opt/Elastic/Endpoint/state/vault" ]; then
+      echo "$SERVICE_NAME is tamper protected"
+      echo "Removing $SERVICE_NAME vault"
+
+      rm -rf "/opt/Elastic/Endpoint/state/vault"
```
cmacknz: Wouldn't you want to do this even if the unit was stopped or disabled already? In the postinstall you handle this case with `systemctl list-unit-files --type=service` but you are only doing an `is-active` check here.

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -2,6 +2,45 @@
...
+SERVICE_NAME="ElasticEndpoint"
+endpoint_active=false
+
+handle_endpoint() {
+  echo "Checking if $SERVICE_NAME is an active service"
+  if systemctl --quiet is-active $SERVICE_NAME; then
+    endpoint_active=true
+
+    # restore endpoint and vault by restarting the service
+    restore_endpoint() {
```
cmacknz: Does the postinstall script run regardless of whether there was an error in this preinstall script? If it does, you might not need this because it already handles starting endpoint if it is stopped.

Just wondering if this can simplify the error handling cases a little bit.

cmacknz: It is a bit hard to find documentation on this as the preinstall and postinstall seem to get converted into prein and in triggers in the RPM package.

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +62,452 @@ var protectionTests = []struct {
...
+			},
+		},
+	})
+
+	t.Run("Upgrade from older version to newer version", func(t *testing.T) {
+		testTamperProtectedDebRpmUpgrades(t, info, "deb")
+	})
+
+	t.Run("Install same version over the installed agent", func(t *testing.T) {
+		testTamperProtectedSameVersionInstall(t, info, "deb")
```
cmacknz: What is this testing that `testTamperProtectedDebRpmUpgrades` is not? Reading the two tests side by side they look very similar, with the difference mostly being that in the upgrade case the version of agent is expected to different.

Does anything different happen in the test logic besides the versions involved?

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +62,452 @@ var protectionTests = []struct {
...
+		}
+	})
+	t.Log("Installing the second agent, upgrading from the older version")
+	installCmd, err := getInstallCommand(ctx, fixture.PackageFormat(), srcPkg, nil)
+	require.NoError(t, err)
+
+	t.Log("Installing the second agent")
+	out, err = installCmd.CombinedOutput()
+	t.Log(string(out))
+	require.Error(t, err)
```
cmacknz: Shouldn't there be some assertions about the state of agent and endpoint in this case to prove the recovery happened?

This is just verifying that the install command exited with a non-zero exit code, not that the agent and endpoint services are in the state they are expected to be in. If they were both broken I think this test would still pass.

testing/integration/endpoint_tamper_protected_mock_systemctl_debrpm.sh.tmpl
```diff
@@ -0,0 +1,48 @@
...
+    echo "[mock] Simulating failure after stop"
+    exit 1
+
+elif [[ "$2" == "start" && "$3" == "$SERVICE_NAME" ]]; then # systemctl --quiet start ElasticEndpoint
+    echo "[mock] Actually starting $2"
+    "$REAL_SYSTEMCTL" start "$SERVICE_NAME"
+    if $SHOULD_WRITE ; then
+        echo "$SERVICE_START_ASSERTION" >> "$ASSERTIONS"
+    fi
+    exit 0
```
cmacknz: This block just looks like it calls through to `systemctl start` unconditionally? Does it need to exist in that case? What's the value of piping a message into the assertions file?

The `systemctl stop` case is clearly simulating an error which makes sense.

kaanyalti: I wanted to assert that endpoint gets stopped and gets started. I guess if I just check if it got stopped and then later check if it is running it would be the same. I can simplify this part yes.

testing/integration/endpoint_security_test.go
```diff
@@ -58,6 +62,452 @@ var protectionTests = []struct {
...
+
+	err = createMockSystemctl(mockSystemctlScriptPath, assertionsPath, serviceStart, serviceStop)
+	require.NoError(t, err, "creating mock systemctl failed")
+
+	systemctlPath := "/usr/bin/systemctl"
+	backupSystemctl := "/usr/bin/systemctl.real"
+
+	if _, err := os.Stat(backupSystemctl); os.IsNotExist(err) {
+		err = os.Rename(systemctlPath, backupSystemctl)
+		require.NoError(t, err, "failed to move real systemctl")
```
pchila: Aside from the fact that I am not too fond of renaming systemctl binary to substitute it with a script, please put the `t.Cleanup()` call restoring the original systemctl just after this so we keep a useable system also in case of failure (for example the `os.Symlink()` just below.

We can have the 2 Cleanup for removing the mock and restoring the original separate and those will execute in the correct order (just like `defer`s)

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
kaanyalti: That is true yes, will update this

testing/integration/endpoint_security_test.go
kaanyalti: You're right, they are almost the same, will refactor

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
kaanyalti: Deb and rpm both mention that scripts must exit with a zero error code. Only postrm seems to be run if preinstall fails for deb [source](https://www.debian.org/doc/debian-policy/ap-flowcharts.html). I couldn't find such a doc for rpm but I found this [source](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/)

```
All scriptlets MUST exit with the zero exit status. Because RPM in its default configuration does not execute shell scriptlets with the -e argument to the shell, excluding explicit exit calls (frowned upon with a non-zero argument!), the exit status of the last command in a scriptlet determines its exit status. Most commands in the snippets in this document have a “|| :” appended to them, which is a generic trick to force the zero exit status for those commands whether they worked or not. Usually the most important bit is to apply this to the last command executed in a scriptlet, or to add a separate command such as plain “:” or “exit 0” as the last one in a scriptlet. Note that depending on the case, other error checking/prevention measures may be more appropriate.

Non-zero exit codes from scriptlets can break installs/upgrades/erases such that no further actions will be taken for that package in a transaction (see [Ordering](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering)), which may for example prevent an old version of a package from being erased on upgrades, leaving behind duplicate rpmdb entries and possibly stale, unowned files on the filesystem. There are some cases where letting the transaction to proceed when some things in scriptlets failed may result in partially broken setup. It is however often limited to that package only whereas letting a transaction to proceed with some packages dropped out on the fly is more likely to result in broader system wide problems.
```
Which tells me rpm exits only with an explicit `exit` call regardless of whether `set -e` is used or not.

I removed the recovery logic from preinstall to simplify error handling. It was not robust, didn't account for other error cases and interruptions and most importantly it did not recover the agent. Preinstall also now checks if endpoint is installed rather than if it is running to check for the endpoint vault. If there are any transient errors during installation, running the install again fixes the issue.

testing/integration/endpoint_security_test.go
kaanyalti: Removed this test case because the recovery logic is removed from preinstall

testing/integration/endpoint_tamper_protected_mock_systemctl_debrpm.sh.tmpl
kaanyalti: Removed this file because recovery logic is removed from preinstall

testing/integration/endpoint_security_test.go
kaanyalti: Removed this test case because recovery logic is removed from preinstall

pkg/testing/fixture_install.go
```diff
@@ -445,6 +446,19 @@ func getProcesses(t *gotesting.T, regex string) []runningProcess {
...
+	if err != nil {
+		return fmt.Errorf("failed to get control protcol address: %w", err)
+	}
+
+	c := client.New(client.WithAddress(socketPath))
+	f.setClient(c)
+
+	return nil
+}
+
```
pchila, line 461: Does this need to be public ? This is called within installDeb and installRpm functions and, aside from a different, constant workdir, it should be instantiated in the same way it's done for .tar.gz/.zip installs.
This is non-blocking obviously, just a thought for future refactor/cleanup

pchila (approved):
Small nitpick about agent fixture Client setting, non-blocking
The rest of the code LGTM

pkg/testing/fixture_install.go
kaanyalti: This gets called in the test functions as well that's why I made it public. The initial agent gets installed using the built in fixture install function. I had created a separate install function `SimpleInstallDebRpm`, but moved that logic into the tests following Craig's suggestion. I have to call `SetDebRpmClient` after those installs.

cmacknz (approved):

cmacknz:
Let's only put this in 9.1 and 8.19 since it is more like a new feature. We can add it to the maintenance branches once we confirm it has no introduced any problems.

I see the 9.0 backport has already merged, we should revert that. This is on the line of being an enhancement vs fixing something that should have worked already, and we lost the bet last time on that one so let's not risk anything again :)

cmacknz:
Revert for 9.0:

- https://github.com/elastic/elastic-agent/pull/8638

Issue #6119, opened 2024-11-21
[Flaky Test]: TestOTelManager_Run – otel collector never got healthy


michalpristas:
### Failing test case

TestOTelManager_Run

### Error message

otel collector never got healthy

### Build

https://buildkite.com/elastic/elastic-agent/builds/14279

### OS

Linux

### Stacktrace and notes

```markdown
=== RUN   TestOTelManager_Run
    manager_test.go:123: 
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1732147373514009164/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:123
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1732147373514009164/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:169
        	Error:      	Condition never satisfied
        	Test:       	TestOTelManager_Run
        	Messages:   	otel collector never got healthy
    manager_test.go:146: otel collector never got healthy: <nil> (latest err: <nil>)
```

swiatekm:
This happened again in https://buildkite.com/elastic/elastic-agent/builds/17563#0195426f-9bc9-43ce-8f7e-94880f4adb80

```
=== RUN   TestOTelManager_Run
    manager_test.go:121: 
        	Error Trace:	C:/buildkite-agent/builds/bk-agent-prod-gcp-1740576392206160015/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:121
        	            				C:/buildkite-agent/builds/bk-agent-prod-gcp-1740576392206160015/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:171
        	Error:      	Condition never satisfied
        	Test:       	TestOTelManager_Run
        	Messages:   	otel collector never got healthy
    manager_test.go:144: otel collector never got healthy: &{Event:0xc000972f90 ComponentStatusMap:map[pipeline:logs:0xc000cfcba0 pipeline:metrics:0xc000cfcc60 pipeline:traces:0xc000cfcc00]} (latest err: <nil>)
--- FAIL: TestOTelManager_Run (301.05s)
```

Looks like we should improve the debug information in this test first, as the status map doesn't tell us anything useful the way it is right now.

swiatekm:
[...]

PR #6927, opened 2025-02-19
Fix deadlock in error handling in OTelManager

blakerouse:
## What does this PR do?

Fixes an issue where a deadlock would cause the the OTelManager to get stuck.

This also seems to be the cause of the flaky test `TestOTelManager_Run`. I have ran this over 100 times and hit no more errors with this PR.

## Why is it important?

Deadlock causes the running internal OTel collector from getting any more updates or actions.

## Checklist

## Disruptive User Impact

None

`go test -race -v -count=100 github.com/elastic/elastic-agent/internal/pkg/otel/manager`

## Related issues

- Closes #6119

### DIFF
--- a/changelog/fragments/1739981369-Fix-deadlock-in-OTelManager.yaml
+++ b/changelog/fragments/1739981369-Fix-deadlock-in-OTelManager.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
 summary: Fix deadlock in OTelManager
--- a/changelog/fragments/1739981369-Fix-deadlock-in-OTelManager.yaml
+++ b/changelog/fragments/1739981369-Fix-deadlock-in-OTelManager.yaml
@@ -1,0 +1,34 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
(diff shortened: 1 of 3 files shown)


swiatekm:
I think we should do the same thing with status reporting.

blakerouse:
@swiatekm We could do that and it could result in less chance of deadlock. But we do not do that with the runtime manager for a reason, and I think the same reason applies here. The reason we do not do that is for logging and `StateWatch` over the control protocol. If we don't ensure that every state change is sent over the channel, then logging or watching of that state could be missed. At the moment the runtime manager doesn't do that for that exact reason.

swiatekm:
> @swiatekm We could do that and it could result in less chance of deadlock. But we do not do that with the runtime manager for a reason, and I think the same reason applies here. The reason we do not do that is for logging and `StateWatch` over the control protocol. If we don't ensure that every state change is sent over the channel, then logging or watching of that state could be missed. At the moment the runtime manager doesn't do that for that exact reason.

I'm fine merging this PR as is, but I also think we should ensure we don't create the same problem with status updates. If all status updates need to be delivered, maybe it's worth having a reasonably large buffered channel for them? I don't think we're really concerned about the Coordinator just not ever consuming the updates, more about unlucky streaks of select choices in the Coordinator loop.

What I really do not want to do is leave another potential race condition in this code just because it's much less probable (right now).

swiatekm (approved):
The changes themselves look good to me. I would feel better about them if we added a simple test to verify that it's possible to, say, submit two empty config changes to the Otel manager without blocking. Especially since it's a bit undertested in general.

blakerouse:
@swiatekm I added to the existing `TestOTelManager_ConfigError` test to send an invalid configuration multiple times to ensure that a deadlock is not caused.

PR #6958, opened 2025-02-20
Fix panic on fetch provider initialization

### LINKED ISSUE #6957: panic adding fetch provider after change to only run provides referenced in policy
When running the Elastic Agent with the `${kubernetes_secrets.*}` provider the following panic will occur. This doesn't seem to have been caught by any integration tests or unit tests.

Caused by my PR here: https://github.com/elastic/elastic-agent/pull/6169

```
panic: assignment to entry in nil map

goroutine 145 [running]:
github.com/elastic/elastic-agent-libs/mapstr.M.Put(0x0, {0x40000bbd40, 0x12}, {0x9a8db80, 0x4000444320})
        /Users/pkoutsovasilis/go/pkg/mod/github.com/elastic/elastic-agent-libs@v0.18.2/mapstr/mapstr.go:383 +0x140
github.com/elastic/elastic-agent/internal/pkg/composable.(*controller).Run.func2()
        /Users/pkoutsovasilis/repos/elastic-agent/internal/pkg/composable/controller.go:204 +0x164
created by github.com/elastic/elastic-agent/internal/pkg/composable.(*controller).Run in goroutine 131
        /Users/pkoutsovasilis/repos/elastic-agent/internal/pkg/composable/controller.go:194 +0x440
```

Version: 8.18+

blakerouse:
## What does this PR do?

Fixes an issue where the initialization of a fetch provider resulted in a `mapstr.Put` on a `nil` `mapstr` and a race condition that can result in the fetch provider not be included in the context variable set.

## Why is it important?

Using any context provider that is a fetch context provider will cause the Elastic Agent to crash.

## Checklist

## Disruptive User Impact

None

## Related issues

- Closes #6957


pkoutsovasilis (commented):
I did test this PR change and I can confirm that I didn't experience the reported panic with it. However I did observe that sometimes (actually most of them) the `Fetch` method of the provider is not called at all. Looked like a race condition but I can't pin it down to a specific place. With your knowledge around these code parts @blakerouse could you see a flow where [this code](https://github.com/elastic/elastic-agent/blob/800d19e780fc03383ab4978c8809ff96428fbb1c/internal/pkg/agent/transpiler/vars.go#L95) is missing a provider due to race conditions?

pkoutsovasilis:
the path so far I have found to make Fetch not getting called is by adding an artificial delay of 1 sec in the "constructor" of the kubernetes_secrets provider [here](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/composable/providers/kubernetessecrets/kubernetes_secrets.go#L83). This makes me suspect that there is somewhere a `<-time.After(...)` with a time limit that when exceeded a provider is lost in the void. To last bit is just a suspicion at the moment I don't have code proof. Looking more into that

swiatekm:
It's definitely possible for the observe call [here](https://github.com/elastic/elastic-agent/blob/93b3138a5a4222054f23e9600ca8ed0a9995d865/internal/pkg/agent/application/coordinator/coordinator.go#L1353) to return without all the providers having been started. Providers are started asynchronously, and the fetch providers map is also updated asynchronously [here](https://github.com/elastic/elastic-agent/blob/93b3138a5a4222054f23e9600ca8ed0a9995d865/internal/pkg/composable/controller.go#L200), but this doesn't trigger new variables. So if this happens, and nothing triggers a new config reload in the coordinator, we'll never get variables with the new providers.

Does that sound possible @blakerouse ?

blakerouse:
@swiatekm It is not that the providers need to be fully started, they publish a new variable once fully started which will signal a new set of variables including the contents.

The second link is defiantly more related to the issue. The issue does seem to be that the fetch providers are not sent over the channel until they are initialized if that takes time (which kubernetes_secrets will to create the client to kubernetes) then it will not be included in the vars.

I have been working on a proper fix, just wanted to inform that I know the issue just not at the proper solution just yet.

blakerouse:
@pkoutsovasilis @swiatekm I have updated this to handle the case for the fetch providers not always being present. I actually like this much better as it removes a lock. I was thinking in my mind this is what it was already doing, but I was incorrect. This change works because each context provider is always started in its own goroutine allowing for a blocking channel to be used for sending the fetch providers. Now a fetch provider results in a new observed set of variables, which is what should have been occurring.

The delay added in the test (which I added a comment on), ensures this code is working correctly. Without the change and the delay present I was seeing what @pkoutsovasilis was describing.

pkoutsovasilis (commented):
yes I just tried again reproducing that but with the latest commits I am not able to do so. Code changes LGTM but I will have to wait for our CI to become again healthy to make sure nothing somewhere is affected by the changes. That said, I wonder if the race condition issue about the providers should be backported also to version prior to 8.18, wdyt? 🤔

blakerouse:
buildkite test this

pkoutsovasilis (approved):

Issue #6965, opened 2025-02-21
[Flaky Test]: TestOTelManager_Run – runtime errored


pkoutsovasilis:
### Failing test case

TestOTelManager_Run

### Error message

Received unexpected error:         	            	failed to create SDK: binding address localhost:8888 for Prometheus exporter: listen tcp 127.0.0.1:8888: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.

### Build

https://buildkite.com/elastic/elastic-agent/builds/17278#019527c5-fdbd-4001-8139-e8c103935195/821-1137

### OS

Windows

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: internal/pkg/otel/manager TestOTelManager_Run (1.00s)
    manager_test.go:149:
        	Error Trace:	C:/buildkite-agent/builds/bk-agent-prod-gcp-1740129062708748931/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:149
        	            				C:/buildkite-agent/builds/bk-agent-prod-gcp-1740129062708748931/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:169
        	Error:      	Received unexpected error:
        	            	failed to create SDK: binding address localhost:8888 for Prometheus exporter: listen tcp 127.0.0.1:8888: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.
        	Test:       	TestOTelManager_Run
        	Messages:   	runtime errored
DONE 2617 tests, 25 skipped, 1 failure in 1424.308s
```

PR #6966, opened 2025-02-21
Fix config used in Otel manager test

swiatekm:
## What does this PR do?

Otel collector exposes metrics on port 8888 by default, which can cause conflicts with other applications in tests. Disable this.

Use the debugexporter and nopreceiver in the test instead of the otlpexporter and otlpreceiver. This is faster and doesn't require binding to any ports as well.

## Why is it important?

Tests should not be flaky.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6965

### DIFF
--- a/internal/pkg/otel/manager/manager_test.go
+++ b/internal/pkg/otel/manager/manager_test.go
@@ -24,38 +24,36 @@ import (
 var (
 	testConfig = map[string]interface{}{
 		"receivers": map[string]interface{}{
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (approved):
To me such a change from `otlp` to `nop` makes sense for this particular test, thus LGTM

blakerouse (approved):
Thanks for improving this.

PR #6982, opened 2025-02-24
Automatic retry for all steps prone to flakiness

pkoutsovasilis:
## What does this PR do?

This PR enables the automatic retry up to 1 time for all steps that are prone to flakiness, such as packaging and testing steps

## Why is it important?

Helps in accelerating PR velocity by preventing unnecessary disruptions caused by transient flakiness, while ensuring that flakiness gets reported back to the author

## Disruptive User Impact

None


jlind23 (approved):

dliappis:
Thank you for raising this.
For posterity, we can already see the result of this change, the GH comment reports that it succeeeded but was flaky:

image

and looking at the steps we can see that there was an auto retry:

image

dliappis (approved):
LGTM from eng prod pov too

Issue #6761, opened 2025-02-07
Indefinite retries on delayed enrollment retry too quickly


cmacknz:
https://github.com/elastic/elastic-agent/pull/4727 made us retry indefinitely when delayed enrollment is used. This works, and does include an exponential backoff, but with a very short duration.

The backoff for delayed enrollment is implemented in https://github.com/elastic/elastic-agent/blob/345f2aeb5e55972177b1caf7ef1f1c0d8d9df2e8/internal/pkg/agent/cmd/enroll_cmd.go#L277

https://github.com/elastic/elastic-agent/blob/345f2aeb5e55972177b1caf7ef1f1c0d8d9df2e8/internal/pkg/agent/cmd/enroll_cmd.go#L532

https://github.com/elastic/elastic-agent/blob/be179d886663a8b0d57a53fe56290bcbc78aebea/internal/pkg/agent/cmd/enroll_cmd.go#L57-L58

The initial delay is 1s ramping up to 10s. We have a user report that delayed enrollment when there are many agent VM images starting before Fleet Server is available and ready to accept connections can DDoS their network infrastructure.

We should make the following changes:
1. There must be a random delay added before the first connection attempt to avoid each agent making it's initial request concurrnetly. The Fleet Gateway uses 500ms for this jitter duration. https://github.com/elastic/elastic-agent/blob/345f2aeb5e55972177b1caf7ef1f1c0d8d9df2e8/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L39-L46
2. The maximum backoff duration when using delayed enrollment should be increased. The Fleet gateway for checkin requests uses 10 minutes for the maximum period.

Possibly, the delayed enrollment and fleet gateway checkins should use the same backoff algorithm since they are both critical operations that reach out to Fleet Server indefinitely.

swiatekm:
Note for posterity

> There must be a random delay added before the first connection attempt to avoid each agent making it's initial request concurrnetly. The Fleet Gateway uses 500ms for this jitter duration.

The jitter here is actually the checkin jitter. For request retries, fleet gateway uses https://github.com/elastic/elastic-agent/blob/a4325c8f5e8cfc1dae44deb6402c875bb90a718c/internal/pkg/core/backoff/equal_jitter.go#L27, which sets jitter to be half the wait time.

PR #6983, opened 2025-02-24
Make enroll command backoff more conservative

swiatekm:
## What does this PR do?

We make two changes to the backoff behaviour of the enroll command:

* Plain enroll now has an initial delay of 5s, a max delay of 10m, and uses jitter.
* Delayed enroll now has the same backoff behaviour as other requests to Fleet. This means an initial delay of 1m, a max delay of 10m, and jitter equal to half the delay time.

## Why is it important?

We've had a report where a lot of agents trying to enroll before Fleet was ready resulted in a DDOS with the previous backoff settings. Any situation where we can retry forever should use very conservative settings to avoid these kinds of situations.

## Checklist

There's a unit test for this behaviour. Testing the actual binary e2e is annoying, because we need to induce fleet server to return 503s.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6761

### DIFF
--- a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
+++ b/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
@@ -37,21 +37,24 @@ const (
 	fleetStateStarting = "starting"
 )
 
+// Default backoff settings for connecting to Fleet
+var defaultFleetBackoffSettings = backoffSettings{
+	Init: 60 * time.Second,
+	Max:  10 * time.Minute,
+}
+
 // Default Configuration for the Fleet Gateway.
 var defaultGatewaySettings = &fleetGatewaySettings{
 	Duration:                     1 * time.Second,        // time between successful calls
 	Jitter:                       500 * time.Millisecond, // used as a jitter for duration
 	ErrConsecutiveUnauthDuration: 1 * time.Hour,          // time between calls when the agent exceeds unauthorized response limit
-	Backoff: backoffSettings{ // time after a failed call
-		Init: 60 * time.Second,
-		Max:  10 * time.Minute,
-	},
+	Backoff:                      &defaultFleetBackoffSettings,
 }
 
 type fleetGatewaySettings struct {
-	Duration                     time.Duration   `config:"checkin_frequency"`
-	Jitter                       time.Duration   `config:"jitter"`
-	Backoff                      backoffSettings `config:"backoff"`
+	Duration                     time.Duration    `config:"checkin_frequency"`
+	Jitter                       time.Duration    `config:"jitter"`
+	Backoff                      *backoffSettings `config:"backoff"`
 	ErrConsecutiveUnauthDuration time.Duration
 }
 
@@ -136,11 +139,18 @@ func (f *FleetGateway) Actions() <-chan []fleetapi.Action {
 }
 
 func (f *FleetGateway) Run(ctx context.Context) error {
-	backoff := backoff.NewEqualJitterBackoff(
-		ctx.Done(),
-		f.settings.Backoff.Init,
-		f.settings.Backoff.Max,
-	)
+	var requestBackoff backoff.Backoff
+	if f.settings.Backoff == nil {
+		requestBackoff = RequestBackoff(ctx.Done())
+	} else {
+		// this is only used in tests
+		requestBackoff = backoff.NewEqualJitterBackoff(
+			ctx.Done(),
+			f.settings.Backoff.Init,
+			f.settings.Backoff.Max,
+		)
+	}
+
 	f.log.Info("Fleet gateway started")
 	for {
 		select {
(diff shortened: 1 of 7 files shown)


internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -37,21 +37,24 @@ const (
 	fleetStateStarting = "starting"
 )
 
+// Default backoff settings for connecting to Fleet
+var defaultFleetBackoffSettings = backoffSettings{
+	Init: 60 * time.Second,
```
blakerouse, line 42: The default of the initial enrollment during installation could be drastically slower because of this change. The original value for this was 1 second, so if it was just something like a network hickup or the server was very close to being ready it will now take 60 seconds versus the original 1 second.

blakerouse (changes requested):
See my comment inline. I think this has an overall effect on installation.

A change to the defaults for delayed enrollment make sense, but not globally for all enrollment.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
swiatekm: Hm, are confident we can't get a thundering herd problem with plain enroll? It can definitely happen with delayed.

The 60s initial delay primarily affects checkins in fleet gateway, where it's not really a problem, but that may be too long of a wait for an interactive command.

blakerouse: I think it would be possible to get a thundering herd problem if Elastic Agent is being deployed out to a fleet of machines through some external orchestration platform.

It was more of a highlight of the fact that for an interactive command that is too long for the initial delay.

swiatekm: Do you think it'd be ok to use the same initial delay (60s) for delayed enroll as we do for Fleet checkins, and set a shorter initial time (let's say 5s) while also adding jitter for the plain enroll command?

swiatekm: Done.

internal/pkg/agent/cmd/run.go
```diff
@@ -545,6 +547,7 @@ func tryDelayEnroll(ctx context.Context, logger *logger.Logger, cfg *configurati
 		&options,
 		paths.ConfigFile(),
 		store,
+		fleetgateway.RequestBackoff, // for delayed enroll, we want to use m
```
blakerouse, line 550: I think this comment got cut off or something. `use m`?

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
blakerouse: Thanks.

blakerouse (commented):
Looks good with the latest changes. Waiting for green CI to approve.

internal/pkg/agent/cmd/run.go
swiatekm: Oops, fixed now.

pkoutsovasilis:
all retries so far are known issue thus I am gonna retry them

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -178,13 +179,20 @@ func newEnrollCmd(
 	options *enrollCmdOption,
 	configPath string,
 	store saver,
+	backoffFactory func(done <-chan struct{}) backoff.Backoff,
```
pkoutsovasilis, line 182: nit; I understand that this might be a personal preference but the "Factory" here doesn't seem 100% appropriate to me. I may be missing something but AFAIK the field `backoffFactory` can seamlessly become `backoff backoff.Backoff` and then inside `enrollWithBackoff` if it is nil assign it to `backoff.NewEqualJitterBackoff(signal, enrollBackoffInit, enrollBackoffMax)`. In this way we keep "close" at the code the signal that this backoff depends on and we don't hold a ref from a function that the enrollCmd instantiator might provide. No strong feelings, just as an extra approach here 🙂

pkoutsovasilis (commented):
@blakerouse these changes LGTM but I see you request for changes, is there something that I am missing or this is the case until you saw our CI green and happy? 🙂

blakerouse (approved):
Yes I was waiting for green CI in accordance to the new team PR requirements.

andrzej-stencel (approved):

internal/pkg/agent/cmd/enroll_cmd.go
pkoutsovasilis: Sorry my intention wasn't to delay this PR from getting merged. I was initially under the impression that we use as signal the `ctx context.Context` passed in `enrollWithBackoff` thus I believed that since the caller is giving us this context it can instantiate properly the backoff and pass it as a parameter in this function as well; I don't see any point having the backoff inside enrollCmd struct as it is used only in `enrollWithBackoff`. However, after looking again at the code, I did notice that we use a `chan struct{}` as the signal thus Factory makes more sense now to me. Resolving this no need to do anything about my comment in this PR 🙂

PR #6997, opened 2025-02-24
Remove unnecessary loop variable copies

swiatekm:
## What does this PR do?

Removes unnecessary loop variable copies. These are not necessary anymore as of Go 1.22. It also replaces the deprecated `exportloopref` linter which would find places where these copies are necessary, with the `copyloopvar` linter which instead flags the copies as unnecessary.

## Why is it important?

We shouldn't unnecessarily copy loop variables, it's confusing and was only necessary due to older Go behaviour.

## Checklist


blakerouse (approved):
Nice, thanks for this!

pchila (approved):
LGTM

Issue #6873, opened 2025-02-14
CI: usage: grep when running the Unit tests in *nix


v1v:
```

2025-02-14 15:07:39 UTC | Unit tests |  
-- | -- | --
  | 2025-02-14 15:07:39 UTC | installing golang 1.24.0 for darwin/arm64
  | 2025-02-14 15:07:46 UTC | ~/builds/bk-agent-prod-orka-1739545555141498457/elastic/elastic-agent ~/builds/bk-agent-prod-orka-1739545555141498457/elastic/elastic-agent
  | 2025-02-14 15:07:46 UTC | go version go1.24.0 darwin/arm64
  | 2025-02-14 15:07:46 UTC | ~/builds/bk-agent-prod-orka-1739545555141498457/elastic/elastic-agent
  | 2025-02-14 15:07:46 UTC | usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
  | 2025-02-14 15:07:46 UTC | [-e pattern] [-f file] [--binary-files=value] [--color=when]
  | 2025-02-14 15:07:46 UTC | [--context[=num]] [--directories=action] [--label] [--line-buffered]
  | 2025-02-14 15:07:46 UTC | [--null] [pattern] [file ...]
  | 2025-02-14 15:07:46 UTC | Installing mage.
```

```

2025-02-14 15:07:21 UTC | Unit tests |  
-- | -- | --
  | 2025-02-14 15:07:21 UTC | installing golang 1.24.0 for linux/amd64
  | 2025-02-14 15:07:25 UTC | ~/builds/bk-agent-prod-gcp-1739545555528492919/elastic/elastic-agent ~/builds/bk-agent-prod-gcp-1739545555528492919/elastic/elastic-agent
  | 2025-02-14 15:07:25 UTC | go version go1.24.0 linux/amd64
  | 2025-02-14 15:07:25 UTC | ~/builds/bk-agent-prod-gcp-1739545555528492919/elastic/elastic-agent
  | 2025-02-14 15:07:25 UTC | Usage: grep [OPTION]... PATTERNS [FILE]...
  | 2025-02-14 15:07:25 UTC | Try 'grep --help' for more information.



```

[...]

PR #7003, opened 2025-02-25
fix: grep is not working correctly

v1v:
## Checklist

## Related issues

Fixes https://github.com/elastic/elastic-agent/issues/6873

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/steps/check-ci.sh
+++ b/.buildkite/scripts/steps/check-ci.sh
@@ -4,6 +4,13 @@ set -euo pipefail
 
 source .buildkite/scripts/common.sh
 
+set +e
+set +x
+echo "--- Check grep"
+grep -oE '[0-9]+\.[0-9]+\.[0-9]+(\-[a-zA-Z]+[0-9]+)?' "version/version.go"
+set -e
+set -x
+
 echo "--- Check CI"
 go version
 mage --version


oakrizan (approved):

Makefile
```diff
@@ -3,20 +3,13 @@ COVERAGE_DIR=$(BUILD_DIR)/coverage
...
-export MAGE_IMPORT_PATH
 
 ## mage : Sets mage
 .PHONY: mage
 mage:
-ifndef MAGE_PRESENT
-	@echo Installing mage.
-	@go install ${MAGE_IMPORT_PATH}
+	@echo Installing mage
+	@go install github.com/magefile/mage
```
blakerouse, line 11: This change is removing an optimization for using the integration testing framework from a developers machine. When running the `mage integration:test` where they already have prepared hosts and they are working on debugging flaky tests or writing new tests, this will slow down the running of those tests.

Because the start of the running of the test will run `make mage` and this will add time. The previous check would cause this to not be ran every time.

blakerouse (changes requested):

Makefile
v1v: > This change is removing an optimization for using the integration testing framework from a developers machine. When running the mage integration:test where they already have prepared hosts and they are working on debugging flaky tests or writing new tests, this will slow down the running of those tests.

If it's installed, it will do nothing as far as I see

```bash
$ time make mage
Installing mage
/Users/vmartinez/.magefile cleaned
make mage  0.41s user 0.46s system 62% cpu 1.378 total

$ time make mage      
Installing mage
/Users/vmartinez/.magefile cleaned
make mage  0.14s user 0.23s system 250% cpu 0.147 total
```

IMO, the time it takes to install the tool, just a couple of seconds if mage is not installed, is undoubtedly worth it for simplicity. IIUC, the toolchain is cached in the CI runners.

However, can you help me understand the impact of slowness?

v1v: I've just found that for the UTs, it took between 7 and 10 seconds:

for x86:
image

for aarch64:
image

It's not 2 seconds, however, I still think it's still valid in terms of what we get compared to the cost of maintenance. 

Let me double-check if I can do something like, if the tool is not installed, install it.

v1v: One last thing currently, it's permanently installing it:
- https://buildkite.com/elastic/elastic-agent-extended-testing/builds/7660#01953f8f-a3ce-4aa3-8a7f-5f385c901d9b/72-98
- https://buildkite.com/elastic/elastic-agent-extended-testing/builds/7660#01953f8f-a3ca-402c-8b1f-54c2980ff2d9/64-72

pchila (approved):
LGTM
Prefer simplicity over a very minor speedup when debugging a test.
I replicated @v1v results when installing mage so I don't think this will heavily affect developers

pchila (commented):
LGTM
Prefer simplicity over a very minor speedup when debugging a test.
I replicated @v1v results when installing mage so I don't think this will heavily affect developers.

Waiting to approve after @blakerouse 's comments have been resolved

Makefile
blakerouse: I was expecting the slowdown to be more, being the Makefile had it explicitly in place. The added time is to minimal to matter. I am good with this change.

blakerouse (approved):

Issue #6917, opened 2025-02-18
[enroll] Windows can return err, even when operation is successful


leehinman:
From CI we have at least one instance of:

```
[=== ] Waiting For Enroll...  [4s] {"log.level":"info","@timestamp":"2025-02-17T22:04:43.470Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).enrollWithBackoff","file.name":"cmd/enroll_cmd.go","file.line":524},"message":"Starting enrollment to URL: https://079ff92028de13e7d02afc20ee027b5f.fleet.us-west2.gcp.elastic-cloud.com:443/","ecs.version":"1.6.0"}
        Error: failed to fix permissions: The operation completed successfully.
```

That error is coming from https://github.com/elastic/elastic-agent/blob/1eefbe0c7b5515b45e2dc71cd543ed2c0a83c476/internal/pkg/agent/cmd/enroll_cmd.go#L287

So apparently on Windows `perms.FixPermissions` can return an "Error" of "The operation completed successfully".  So we need to handle this.

pkoutsovasilis:
@michel-laterman since you are the author of [this PR](https://github.com/elastic/elastic-agent/pull/5317) could the root reason behind this issue be the fact that we don't check for Errno(0) here https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/perms/windows.go#L103?

Related CI failures are [here](https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/1830#019515e2-045a-4ecd-9bac-3797a67fc63f) and they are regarding `sudo` and `windows`, thus, I expect the the `if isAdmin` branch to be executed

michel-laterman:
Yes, this is likely a case that we forgot to test for `errors.Is(err, syscall.Errno(0))` for the error that can be returned from
https://github.com/elastic/elastic-agent/blob/1eefbe0c7b5515b45e2dc71cd543ed2c0a83c476/internal/pkg/agent/perms/windows.go#L102-L104

pkoutsovasilis:
@jlind23 @ebeahan here is an update; we will now return the correct error and not a misleading `Error: failed to fix permissions: The operation completed successfully.`. However, until this re-surfaces we don't have the necessary amount of info to see what might go wrong with our fix permissions on Windows. That said, given the description of the issue, I believe we can close it as this no longer happends, but I leave it up to you 🙂

[...]

PR #7059, opened 2025-02-27
[bug] handle syscall errno in windows permissions

pkoutsovasilis:
## What does this PR do?

This PR fixes an issue in Windows permission handling where `FixPermissions` could return an incorrect error message: `"The operation completed successfully."` The issue was caused by improper handling of the return value from Windows API calls, such as in `SetNamedSecurityInfoW`. 

Previously, the code was relying on `GetLastError`, which does not always provide relevant information in this context. Instead, the correct approach is to check the primary return value (`ret`) from the system call itself. The fix ensures that errors are propagated correctly, preventing misleading error messages.

To achieve the above, this PR removes the dependency on `github.com/hectane/go-acl` and replaces it with an internal implementation of `acl.Apply`.

## Why is it important?

- Prevents misleading error messages.
- Ensures proper error handling for Windows API calls.

## Checklist

- [ ] My code follows the style guidelines of this project.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] I have made corresponding changes to the default configuration files.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog).
- [ ] I have added an integration test or an E2E test.

## Disruptive User Impact

This fix is not expected to cause any disruptions.

N/A

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/6917

### DIFF
--- a/internal/pkg/acl/acl_windows.go
+++ b/internal/pkg/acl/acl_windows.go
@@ -180,9 +180,14 @@ func TakeOwnership(name string, owner *windows.SID, group *windows.SID) error {
 // SetEntriesInAcl creates a new access control list (ACL) by merging new access
 // control or audit control information into an existing ACL structure.
 func SetEntriesInAcl(entries []ExplicitAccess, oldAcl windows.Handle, newAcl *windows.Handle) error {
+	var entriesPtr unsafe.Pointer
+	if len(entries) > 0 {
+		entriesPtr = unsafe.Pointer(&entries[0])
+	}
+
 	ret, _, _ := procSetEntriesInAclW.Call(
 		uintptr(len(entries)),
-		uintptr(unsafe.Pointer(&entries[0])),
+		uintptr(entriesPtr),
 		uintptr(oldAcl),
 		uintptr(unsafe.Pointer(newAcl)),
 	)
--- a/internal/pkg/acl/acl_windows.go
+++ b/internal/pkg/acl/acl_windows.go
@@ -158,8 +158,26 @@ func Chmod(name string, fileMode os.FileMode) error {
 		name,
 		true,
 		false,
+		// For owner permissions:
+		// (mode&0700)<<23: Extracts the rwx bits for owner (bits 8-6) and shifts them
+		// to bits 31-29 in the Windows ACL
+		// (mode&0200)<<9: Extracts the write bit for owner (bit 7) and shifts it
+		// to bit 16 in the Windows ACL (for additional write permissions)
+		// https://learn.microsoft.com/en-us/windows/win32/secauthz/access-mask
 		GrantSid(((mode&0700)<<23)|((mode&0200)<<9), creatorOwnerSID),
+		// For group permissions:
+		// (mode&0070)<<26: Extracts the rwx bits for group (bits 5-3) and shifts them
+		// to bits 31-29 in the Windows ACL
+		// (mode&0020)<<12: Extracts the write bit for group (bit 4) and shifts it
+		// to bit 16 in the Windows ACL (for additional write permissions)
+		// https://learn.microsoft.com/en-us/windows/win32/secauthz/access-mask
 		GrantSid(((mode&0070)<<26)|((mode&0020)<<12), creatorGroupSID),
+		// For other/everyone permissions:
+		// (mode&0007)<<29: Extracts the rwx bits for others (bits 2-0) and shifts them
+		// to bits 31-29 in the Windows ACL
+		// (mode&0002)<<15: Extracts the write bit for others (bit 1) and shifts it
+		// to bit 16 in the Windows ACL (for additional write permissions)
+		// https://learn.microsoft.com/en-us/windows/win32/secauthz/access-mask
 		GrantSid(((mode&0007)<<29)|((mode&0002)<<15), everyoneSID),
 	)
 }
--- a/internal/pkg/acl/acl_windows.go
+++ b/internal/pkg/acl/acl_windows.go
@@ -1,0 +1,279 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+//go:build windows
+
+package acl
+
+import (
+	"fmt"
+	"os"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa379638.aspx
+const (
+	TRUSTEE_IS_SID = iota
+	TRUSTEE_IS_NAME
+	TRUSTEE_BAD_FORM
+	TRUSTEE_IS_OBJECTS_AND_SID
+	TRUSTEE_IS_OBJECTS_AND_NAME
+)
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa379593.aspx
+const (
+	SE_UNKNOWN_OBJECT_TYPE = iota
+	SE_FILE_OBJECT
+	SE_SERVICE
+	SE_PRINTER
+	SE_REGISTRY_KEY
+	SE_LMSHARE
+	SE_KERNEL_OBJECT
+	SE_WINDOW_OBJECT
+	SE_DS_OBJECT
+	SE_DS_OBJECT_ALL
+	SE_PROVIDER_DEFINED_OBJECT
+	SE_WMIGUID_OBJECT
+	SE_REGISTRY_WOW64_32KEY
+)
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa374899.aspx
+const (
+	NOT_USED_ACCESS = iota
+	GRANT_ACCESS
+	SET_ACCESS
+	DENY_ACCESS
+	REVOKE_ACCESS
+	SET_AUDIT_SUCCESS
+	SET_AUDIT_FAILURE
+)
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa446627.aspx
+const (
+	NO_INHERITANCE                     = 0x0
+	SUB_OBJECTS_ONLY_INHERIT           = 0x1
+	SUB_CONTAINERS_ONLY_INHERIT        = 0x2
+	SUB_CONTAINERS_AND_OBJECTS_INHERIT = 0x3
+	INHERIT_NO_PROPAGATE               = 0x4
+	INHERIT_ONLY                       = 0x8
+
+	OBJECT_INHERIT_ACE       = 0x1
+	CONTAINER_INHERIT_ACE    = 0x2
+	NO_PROPAGATE_INHERIT_ACE = 0x4
+	INHERIT_ONLY_ACE         = 0x8
+)
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa379573.aspx
+const (
+	OWNER_SECURITY_INFORMATION               = 0x00001
+	GROUP_SECURITY_INFORMATION               = 0x00002
+	DACL_SECURITY_INFORMATION                = 0x00004
+	SACL_SECURITY_INFORMATION                = 0x00008
+	LABEL_SECURITY_INFORMATION               = 0x00010
+	ATTRIBUTE_SECURITY_INFORMATION           = 0x00020
+	SCOPE_SECURITY_INFORMATION               = 0x00040
+	PROCESS_TRUST_LABEL_SECURITY_INFORMATION = 0x00080
+	BACKUP_SECURITY_INFORMATION              = 0x10000
+
+	PROTECTED_DACL_SECURITY_INFORMATION   = 0x80000000
+	PROTECTED_SACL_SECURITY_INFORMATION   = 0x40000000
+	UNPROTECTED_DACL_SECURITY_INFORMATION = 0x20000000
+	UNPROTECTED_SACL_SECURITY_INFORMATION = 0x10000000
+)
+
+var (
+	advapi32 = windows.MustLoadDLL("advapi32.dll")
+	// procGetNamedSecurityInfoW (https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getnamedsecurityinfow)
+	//   If the function succeeds, the return value is ERROR_SUCCESS.
+	//   If the function fails, the return value is a nonzero error code defined in WinError.h.
+	//   Doesn't set last error.
+	procGetNamedSecurityInfoW = advapi32.MustFindProc("GetNamedSecurityInfoW")
+	// procSetNamedSecurityInfoW (https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setnamedsecurityinfow)
+	//   If the function succeeds, the return value is ERROR_SUCCESS.
+	//   If the function fails, the return value is a nonzero error code defined in WinError.h.
+	//   Doesn't set last error.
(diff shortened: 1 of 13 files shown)


internal/pkg/agent/perms/windows.go
```diff
@@ -83,52 +92,41 @@ func FixPermissions(topPath string, opts ...OptFunc) error {
...
-						inherit = false
-					}
+					inherit := topPath == name
 
-					err = acl.Apply(name, true, inherit, grants...)
-					if err != nil {
-						// Check for Errno = 0 which indicates success
-						// https://pkg.go.dev/golang.org/x/sys/windows#Errno
-						if errors.Is(err, syscall.Errno(0)) {
-							return nil
```
pkoutsovasilis, line 98: watch here, previously if this was syscall.Errno(0), we were returning which IMO shouldn't be the case, and we should continue below trying to take onwership

internal/pkg/agent/perms/windows.go
```diff
@@ -21,6 +21,15 @@ import (
 	"github.com/elastic/elastic-agent/pkg/utils"
 )
 
+func errHandler(err error) error {
```
blakerouse, line 24: A clearer name would be better here. Being names `errHandler` doesn't really provide context to what it is doing, when reading the code where it is called from. Something like `filterSuccessErr` would be more appropriate.

```suggestion
func filterSuccessErr(err error) error {
```

blakerouse (dismissed):
Other than the naming suggestion this looks good.

internal/pkg/agent/perms/windows.go
cmacknz: Should we put this on every windows sycall we make? What is special about `acl.Apply` here?

pkoutsovasilis: I believe that this error code (ERROR_SUCCESS) handling should happen for all functions that call directly the Win32 low level API. Now acl.Apply is actually calling this API [here](https://github.com/hectane/go-acl/blob/master/api/secinfo.go#L71) but doesn't check for it. Thus I don't believe we should add this check in every windows syscall 🙂

pkoutsovasilis: thank you @blakerouse I was clearly lacking inspiration when naming this function but I do like your proposal! I just committed it 🙂

pkoutsovasilis: @blakerouse @cmacknz should I consider this conversation as resolved? 🙂

blakerouse: Resolved from my side. I will let @cmacknz respond for his question on doing this with every Windows syscall we make.

cmacknz: It looks to me like we are directly calling through to Windows DLLs in other places, like https://github.com/elastic/elastic-agent/blob/d6687876cb0ac8671719aecca7336cdeff671710/internal/pkg/agent/install/user_windows.go#L26-L32

pkoutsovasilis: ok I believe then we should handle the errors, under this issue or a separate one?

pkoutsovasilis: and after taping inside acl.Apply I believe we should re-implement that in our own. Specifically this is the body of acl.Apply

```
func Apply(name string, replace, inherit bool, entries ...api.ExplicitAccess) error {
	var oldAcl windows.Handle
	if !replace {
		var secDesc windows.Handle
		api.GetNamedSecurityInfo(
			name,
			api.SE_FILE_OBJECT,
			api.DACL_SECURITY_INFORMATION,
			nil,
			nil,
			&oldAcl,
			nil,
			&secDesc,
		)
		defer windows.LocalFree(secDesc)
	}
	var acl windows.Handle
	if err := api.SetEntriesInAcl(
		entries,
		oldAcl,
		&acl,
	); err != nil {
		return err
	}
	defer windows.LocalFree((windows.Handle)(unsafe.Pointer(acl)))
	var secInfo uint32
	if !inherit {
		secInfo = api.PROTECTED_DACL_SECURITY_INFORMATION
	} else {
		secInfo = api.UNPROTECTED_DACL_SECURITY_INFORMATION
	}
	return api.SetNamedSecurityInfo(
		name,
		api.SE_FILE_OBJECT,
		api.DACL_SECURITY_INFORMATION|secInfo,
		nil,
		nil,
		acl,
		0,
	)
}
```

SetEntriesInAcl [can return a ERROR_SUCCESS](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setentriesinaclw) which is not handled and instead it returns. Which means that we will miss calling `SetNamedSecurityInfoW` 🤦

leehinman: I could be wrong, but I think with that code snippet, `oldAcl` could have received a pointer to the DACL, in which case it would need to be freed when you exit.

pkoutsovasilis: the more I read about it, the more I start to believe that we are mislead here gents. So our starting point is this error message `Error: failed to fix permissions: The operation completed successfully.` and I believe that exactly this one is derailing us. Specifically, this error comes from within `github.com/hectane/go-acl` from a function like this

```go
func SetNamedSecurityInfo(objectName string, objectType int32, secInfo uint32, owner, group *windows.SID, dacl, sacl windows.Handle) error {
	ret, _, err := procSetNamedSecurityInfoW.Call(
		uintptr(unsafe.Pointer(windows.StringToUTF16Ptr(objectName))),
		uintptr(objectType),
		uintptr(secInfo),
		uintptr(unsafe.Pointer(owner)),
		uintptr(unsafe.Pointer(group)),
		uintptr(dacl),
		uintptr(sacl),
	)
	if ret != 0 { 
		return err // <-- attention here actual ret is not returned
	}
	return nil
}
```

the godoc of the `.Call(...)` func states the following

```go
// Call executes procedure p with arguments a. It will panic, if more than 15 arguments
// are supplied.
//
// The returned error is always non-nil, constructed from the result of GetLastError (<-- pay attention here).
// Callers must inspect the primary return value to decide whether an error occurred
// (according to the semantics of the specific function being called) before consulting
// the error. The error will be guaranteed to contain windows.Errno.
```

Based on the above, I do believe that we indeed experience *an actual error* because `ret != 0`. In conjunction with the [official documentation of the function](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setnamedsecurityinfow) which doesn't mention that you should read the last error to see more details why `ret != 0`, I truly believe that the lastErr that `Call` returns and is propagated to us on `ret != 0` from the library regards another operation and not our error, thus it says `operation completed successfully.`. 

Based on all the above my proposal is actually do something similar we do for the other windows `Calls` in the code and propagate an error with the actual retval in it in the hope of being able understand better what is failing when this occurs

cc @cmacknz @blakerouse

pkoutsovasilis: @leehinman for the code snippet comment; according to the [GetNamedSecurityInfoW documentation](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getnamedsecurityinfow) you shouldn't free `oldAcl`

```
[out, optional] ppDacl

A pointer to a variable that receives a pointer to the DACL in the returned security descriptor or NULL if the security descriptor has no DACL. The returned pointer is valid only if you set the DACL_SECURITY_INFORMATION flag. Also, this parameter can be NULL if you do not need the DACL.
```

where you should free `secDesc`

```
[out, optional] ppSecurityDescriptor

A pointer to a variable that receives a pointer to the security descriptor of the object. When you have finished using the pointer, free the returned buffer by calling the [LocalFree](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-localfree) function.
```

internal/pkg/util/errno_windows.go
```diff
@@ -0,0 +1,22 @@
...
+
+package util
+
+import (
+	"errors"
+	"syscall"
+)
+
+// FilterErrno filters out Errno = 0 which indicates success
+func FilterErrno(err error) error {
```
blakerouse, line 15: This is now only being used in testing, I think it would be better to just remove it.

testing/installtest/checks_windows.go
```diff
@@ -151,9 +155,11 @@ func getAllowedSIDs(secInfo *windows.SECURITY_DESCRIPTOR) ([]*windows.SID, error
 	aceCount := rs.Field(3).Uint()
 	for i := uint64(0); i < aceCount; i++ {
 		ace := &accessAllowedAce{}
-		ret, _, _ := procGetAce.Call(uintptr(unsafe.Pointer(dacl)), uintptr(i), uintptr(unsafe.Pointer(&ace)))
+		ret, _, err := procGetAce.Call(uintptr(unsafe.Pointer(dacl)), uintptr(i), uintptr(unsafe.Pointer(&ace)))
 		if ret == 0 {
-			return nil, fmt.Errorf("while getting ACE: %w", windows.GetLastError())
+			if err = util.FilterErrno(err); err != nil {
```
blakerouse, line 160: Is this even needed here? Why would it return a 0 error and the last error be set as success?

blakerouse: I think the reason we where seeing this before was because of the issue you found in this work. That golang was reading last error when it was not being set.

pkoutsovasilis: actually you are right the return value is just a bool for this one thus there is not point to check GetLastError(), that golang always returns, with ERROR_SUCCESS. removing it now

internal/pkg/agent/perms/windows.go
pkoutsovasilis: gents after cross-validating my understanding with @blakerouse I think we both agree that there is an error happening and the first step to get more awareness of the root cause, is to propagate the "correct" error that captures the actual ret code that the Win32 API calls fail with. I have updated my PR to achieve that. So, I am resolving this one and let's kick off a new round of reviews please 🙂

testing/installtest/checks_windows.go
pkoutsovasilis: just removed, please have another pass 🙂

blakerouse (approved):
Thanks for the cleanup and researching this issue. Looks good.

leehinman (dismissed):
Code changes look fine.  I think it would be nice if there was a unit test in acl package for `Chmod` verifying that when it is supposed to fail it does, and the error message is "reasonable".  I'll let you decide if that is reasonable or not.

pkoutsovasilis:
Always happy to lean more on unit-test thus thank you for raising this in this PR @leehinman 🙂. I added unit-tests for all Win32 API calls affected in this PR. During this effort I also spotted an issue with `SetEntriesInAcl` func where we were accessing `&entries[0]` without checking the length of entries (fixed). Also I wrapped any error during the file walking to capture exactly which part of that went wrong. Last but not least, with unit-tests in place that test the unhappy path I could validate if this whole analysis of not relying on the LastError() for these Win32 API calls is legit and yes I do report that it is. This is what the unit-test report if I use the previous code, which obviously is not correct 🙂 

```
--- PASS: TestChmod/should_succeed (0.00s)
=== RUN   TestChmod/non_existing_file
    acl_windows_test.go:65: 
        	Error Trace:	C:/Users/kouts/elastic-agent/internal/pkg/acl/acl_windows_test.go:65
        	Error:      	Not equal: 
        	            	expected: "call to SetNamedSecurityInfoW failed: ret=3"
        	            	actual  : "The operation completed successfully."
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-call to SetNamedSecurityInfoW failed: ret=3
        	            	+The operation completed successfully.
        	Test:       	TestChmod/non_existing_file
        	Messages:   	errors do not match
--- FAIL: TestChmod/non_existing_file (0.00s)

Expected :call to SetNamedSecurityInfoW failed: ret=3
Actual   :The operation completed successfully.
<Click to see difference>


=== RUN   TestChmod/invalid_path
    acl_windows_test.go:65: 
        	Error Trace:	C:/Users/kouts/elastic-agent/internal/pkg/acl/acl_windows_test.go:65
        	Error:      	Not equal: 
        	            	expected: "call to SetNamedSecurityInfoW failed: ret=123"
        	            	actual  : "The operation completed successfully."
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-call to SetNamedSecurityInfoW failed: ret=123
        	            	+The operation completed successfully.
        	Test:       	TestChmod/invalid_path
        	Messages:   	errors do not match
--- FAIL: TestChmod/invalid_path (0.00s)

Expected :call to SetNamedSecurityInfoW failed: ret=123
Actual   :The operation completed successfully.
<Click to see difference>


--- FAIL: TestChmod (0.00s)
FAIL
```

pkoutsovasilis:
I have reported the re-occurrences of the flaky issues

internal/pkg/acl/acl_windows.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return err
+	}
+
+	mode := uint32(fileMode)
+	return Apply(
+		name,
+		true,
+		false,
+		GrantSid(((mode&0700)<<23)|((mode&0200)<<9), creatorOwnerSID),
```
michalpristas, line 161: please comment these shifts for context

michalpristas (approved):

internal/pkg/acl/acl_windows.go
pkoutsovasilis: although I "inherited" these from the external library that now lives inside agent code, I just added comments 🙂 feel free to un-resolve if you spot anything 🙂

blakerouse (approved):
Thanks for working on this and thanks for adding the unit tests. Overall this looks good.

blakerouse:
Thought I observed the CI as passing, so its a little early on approval. I don't see any issues that would affect an early approval.

pkoutsovasilis:
ok I reported all the flaky tests occurrences in the last CI run of this PR

internal/pkg/acl/acl_windows_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestChmod(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("skipping test on non-windows platform")
+	}
+
+	// Create a temporary file
+	tmpFile, err := os.CreateTemp("", "testfile")
```
leehinman, line 25:
```suggestion
	tmpFile, err := os.CreateTemp(t.TempDir(), "testfile")
```

The TempDir is automatically deleted, so you don't have to worry about the cleanup and more importantly it has [removeAll](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/testing/testing.go;drc=3aa7c5ef01e147fb482f4b3e79c6f875a4b1b9fb;l=1290) that handles Windows giving a temporary "Access Denied" error message.

leehinman (changes requested):

internal/pkg/acl/acl_windows_test.go
leehinman: Hit the wrong button, should be a comment so optional.  Switching to `t.TempDir` is shorter, uses `testing` package in a more idiomatic way and has more flaky test protections.  Up to you if you want to switch.

pkoutsovasilis: ok if this comes up as I flaky I can provide guarantees that I will open such a fix on the next instant 🙂

PR #7145, opened 2025-03-03
Add a changelog note that the agent Docker container is not longer available from docker.elastic.co/beats/elastic-agent

cmacknz:
## What does this PR do?

What the summary says.

## Why is it important?

We haven't documented this anywhere else and it's a breaking change.


pkoutsovasilis (approved):
LGTM, do we need to add the `backport-9.0` label?

cmacknz:
I made this PR directly against 9.0 so no backport needed

Issue #6969, opened 2025-02-21
[Flaky Test]: TestForceInstallOverProtectedPolicy – Policy Response was: {{ 0 false [] { }     }}


pkoutsovasilis:
### Failing test case

TestForceInstallOverProtectedPolicy

### Error message

Policy Response was: {{ 0 false [] { }     }}

### Build

https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/2011#019525dd-d7fc-4048-adc9-1ddaaa41bb90/336-1001

### OS

Linux

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: testing/integration TestForceInstallOverProtectedPolicy (107.61s)
    endpoint_security_test.go:854: Enrolling the agent in Fleet
    endpoint_security_test.go:854: Protected with uninstall token: e30df3f877deab8d335e40198c2e3eef
    tools.go:104: Creating enrollment API key...
    tools.go:95: Creating enrollment API key...
    fixture_install.go:199: [test TestForceInstallOverProtectedPolicy] Inside fixture install function
    fixture_install.go:227: [test TestForceInstallOverProtectedPolicy] Inside fixture installNoPkgManager function
    fixture.go:290: Extracting artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestForceInstallOverProtectedPolicy3080064164
    fixture.go:308: Completed extraction of artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestForceInstallOverProtectedPolicy3080064164
    fixture.go:908: Components were not modified from the fetched artifact
    fixture.go:665: >> running binary with: [/tmp/TestForceInstallOverProtectedPolicy3080064164/elastic-agent-9.0.0-SNAPSHOT-linux-x86_64/elastic-agent install --force --non-interactive --url https://b8e86b692001e78b7fe16bbdc3e61a2f.fleet.us-west2.gcp.elastic-cloud.com:443 --enrollment-token UnpEb0paVUJSTV9ad2ZxOHQwcGo6Y1N3Q2ZhYVFwNGtacDFTVVE1b2dTdw==]
    tools.go:95: >>> Enroll succeeded. Output: Installing in non-interactive mode.
[    ] Service Started  [4s] Elastic Agent successfully installed, starting enrollment.
[...]

Issue #4883, opened 2024-06-07
[Flaky Test]: TestEndpointSecurityNonDefaultBasePath, TestEndpointSecurityUnprivileged – version conflict, document already exists


rdner:
### Failing test case

TestEndpointSecurityNonDefaultBasePath, TestEndpointSecurityUnprivileged

### Error message

version conflict, document already exists

### Build

https://buildkite.com/elastic/elastic-agent-extended-testing/builds/441

### OS

Linux

### Stacktrace and notes

```markdown
endpoint_security_test.go:614: Installing Elastic Defend
    endpoint_security_test.go:615: Templating endpoint package policy request
    endpoint_security_test.go:615: POST /api/fleet/package_policies
    endpoint_security_test.go:615: Error installing fleet package: Error installing endpoint 8.11.0: version_conflict_engine_exception
        	Root causes:
        		version_conflict_engine_exception: [data_frame_transform_state_and_stats-endpoint.metadata_current-default-8.11.0]: version conflict, document already exists (current version [1])
    endpoint_security_test.go:616: 
        	Error Trace:	/home/ubuntu/agent/testing/integration/endpoint_security_test.go:616
        	Error:      	Received unexpected error:
        	            	error installing fleet package: Error installing endpoint 8.11.0: version_conflict_engine_exception
        	            		Root causes:
        	            			version_conflict_engine_exception: [data_frame_transform_state_and_stats-endpoint.metadata_current-default-8.11.0]: version conflict, document already exists (current version [1])
        	Test:       	TestEndpointSecurityUnprivileged
        	Messages:   	Policy Response was: {{ 0 false [] { }     }}
```

rdner:
Possibly related to:

* https://github.com/elastic/elastic-agent/issues/4158
* https://github.com/elastic/elastic-agent/issues/3524

rdner:
@elastic/fleet could you have a look?

rdner:
Another failure in https://buildkite.com/elastic/elastic-agent-extended-testing/builds/727#01902afb-0d73-4154-b29f-1353b660d38c

criamico:
[...]

PR #7165, opened 2025-03-04
[flakiness] Correct group for TestEndpointAgentServiceMonitoring

pkoutsovasilis:
## What does this PR do?

This PR relocates TestEndpointAgentServiceMonitoring to the FleetEndpointSecurity execution group to prevent test failures caused by concurrent installation attempts of the Endpoint package when tests run in different groups.

## Why is it important?

Reduces flakiness of Endpoint related tests that run in our CI

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/6969
- Relates https://github.com/elastic/elastic-agent/issues/4883
- Relates https://github.com/elastic/elastic-agent/issues/7221

### DIFF
--- a/testing/integration/monitoring_endpoint_test.go
+++ b/testing/integration/monitoring_endpoint_test.go
@@ -34,7 +34,7 @@ type EndpointMetricsMonRunner struct {
 
 func TestEndpointAgentServiceMonitoring(t *testing.T) {
 	info := define.Require(t, define.Requirements{
-		Group: Fleet,
+		Group: FleetEndpointSecurity,
 		Stack: &define.Stack{},
 		Local: false, // requires Agent installation
 		Sudo:  true,  // requires Agent installation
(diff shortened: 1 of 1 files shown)


swiatekm:
Is this supposed to be a temporary fix? We may eventually want to run these tests concurrently, in which case I think we'd want to check if the package is installed before trying to install it?

pkoutsovasilis:
> Is this supposed to be a temporary fix? We may eventually want to run these tests concurrently, in which case I think we'd want to check if the package is installed before trying to install it?

I feel that we should first address the flakiness of how test are currently running, with the minimal blast radius (avoid adding extra flakiness) and then in a follow-up decide how these tests can be altered to support running concurrently. wdyt? 🙂

swiatekm (approved):

pkoutsovasilis:
@dliappis @pazone since this regards a change that was introduced by the latest concurrent groups effort in the CI test execution I would appreciate a review from either of you 🙂

andrzej-stencel (approved):

Issue #4588, opened 2024-04-17
Allow different package contents by package type


pchila:
Elastic Agent already packages different components for different platforms (os/cpu architecture pairs) but as shown in PR #4584 we have the need to include different components depending also on the package format.

The goal of this issue is to modify the packaging process so we can describe what exactly a given package should contain or exclude and exert more control over the packaging process overall,

Once this issue is implemented the hack introduced in PR #4584 should be replaced with a proper solution that leverages the package specs/component manifests.

PR #7173, opened 2025-03-04
EDOT and Slim docker images added

### LINKED ISSUE #6764: EDOT-Collector Docker Images
As a follow-on to https://github.com/elastic/elastic-agent/pull/6542

Implementing this for the elastic-agent images is a breaking change, but we also want to introduce new `edot-collector` images for Elastic Agent.

For Elastic Agent we want there to be two docker images: `elastic-agent` which includes all components and `elastic-agent-slim`, which would be a container built with only agentbeat, osqueryd, and pf-host-agent.

For EDOT Collector, we want there to be one docker image: `edot-collector` which includes only Agentbeat and osqueryd. Ideally this would be the same image as `elastic-agent-slim`, simply aliased to `edot-collector`.

As a workaround to enable availability of the `edot-collector` image starting with 9.0, without making a breaking change later, we could base our edot-collector image on the existing elastic-agent image, and simply delete the extra components from the image. The addition/removal of the components would still occupy space in the image as they would be done in separate layers, but the binaries would be unavailable for a customer to leverage.

When we are ready to take on work to produce the slim image from scratch we will be able to swap it in without a breaking change for users.

michalpristas:
## What does this PR do?

This PR adds 2 new docker images
- `slim` with, agentbeat and profiling
- `edot-collector` with agentbeat

It also modifies packaging so in case spec file is specified with option `expand_spec` everything from components directory conforming to filter specified in a spec file is copied over not just a spec file. 

EDOT image defaults to otel mode 

Closes #6764


dev-tools/packaging/packages.yml
```diff
@@ -299,6 +305,67 @@ shared:
         source: '{{ elastic_beats_dir }}/specs/connectors.spec.yml'
         mode: 0644
 
+  - &agent_docker_edot_spec
+    <<: *agent_binary_spec_no_components
+    docker_variant: 'edot-collector'
+    extra_vars:
+      dockerfile: 'Dockerfile.elastic-agent.tmpl'
+      docker_entrypoint: 'docker-entrypoint.elastic-agent.tmpl'
```
cmacknz, line 313: I think we would want the entrypoint to explicitly be the otel sub-command for this image

cmacknz: I see this is getting handled by an env var.

cmacknz (commented):
You need some tests that the images do what they are supposed to do. We have the following requirements to prove actually happened:

1. The elastic-agent-slim image runs as an elastic-agent, is smaller than the basic image, and only includes agentbeat, osqueryd, and pf-host-agent
2. The edot-collector images runs as an otel collector, is smaller than the slim images, and includes agentbeat.

You have added the additional capability which we should confirm we actually need, which is that the edot-collector will run as an Elastic Agent
if you add `ELASTIC_AGENT_OTEL=false`. We can remove a couple of files if we don't need this. People who would do this
can just use the slim image instead. As is, you may see people choose this image if they want the smallest possible elastic agent container we
have and they use Beat based integrations that are not OSQuerybeat. This might create some confusion for us when triaging problems as people
will be using EDOT collector to run integrations with Beats as sub-processes (vs the long term approach of Beats receivers without the agentbeat
binary). `docker run --env ELASTIC_AGENT_OTEL=false docker.elastic.co/elastic-agent/elastic-agent-edot-collector:9.1.0-SNAPSHOT` right now
runs a standard Elastic Agent.

Some of this could be done as packaging tests, I think we need at least one test that makes sure these images actually run. If we can avoid just
feeding these contains through every single Kubernetes test we have that would be ideal. @pkoutsovasilis happy to get your opinion on smarter ways to test these besides just running the existing k8s tests on them as another variant.

Checking the sizes the size relationships are as expected with respect to the elastic-agent image, but the slim and edot images are still huge and
edot is barely smaller than slim making me question if it should be a different container spec vs just the same one renamed.

```sh
docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/elastic-agent:9.1.0-SNAPSHOT
1586975035 # 1.58 GB

docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/elastic-agent-slim:9.1.0-SNAPSHOT
988323176 # 988 MB, 62.2% of elastic-agent

>>> 988323176 / 1586975035.0
0.6227717224297734

docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/elastic-agent-edot-collector:9.1.0-SNAPSHOT
903081204 # 903 MB, 56.9% of elastic-agent
```

Looking at where the size is coming from, it's the giant chown command that has always caused all our problems. Do we still need this
in the exact same form for edot and slim images, or are we stuck with it?

```sh
docker history docker.elastic.co/elastic-agent/elastic-agent-edot-collector:9.1.0-SNAPSHOT
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
964421276311   26 minutes ago   ENTRYPOINT ["/usr/bin/tini" "--" "/usr/local…   0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   WORKDIR /usr/share/elastic-agent                0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   ENV LIBBEAT_MONITORING_CGROUPS_HIERARCHY_OVE…   0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   USER elastic-agent                              0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   COPY /usr/share/elastic-agent/NOTICE.txt /li…   5.89MB    buildkit.dockerfile.v0
<missing>      26 minutes ago   COPY /usr/share/elastic-agent/LICENSE.txt /l…   3.86kB    buildkit.dockerfile.v0
<missing>      26 minutes ago   RUN /bin/sh -c mkdir /licenses # buildkit       0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   RUN /bin/sh -c chmod 0777 /usr/share/elastic…   101kB     buildkit.dockerfile.v0

# Ouch
<missing>      26 minutes ago   COPY --chown=elastic-agent:elastic-agent /us…   828MB     buildkit.dockerfile.v0

<missing>      26 minutes ago   RUN /bin/sh -c groupadd --gid 1000 elastic-a…   4.18kB    buildkit.dockerfile.v0
<missing>      26 minutes ago   COPY docker-entrypoint /usr/local/bin/docker…   509B      buildkit.dockerfile.v0
<missing>      26 minutes ago   RUN /bin/sh -c set -e ;   TINI_BIN="";   TIN…   23.9kB    buildkit.dockerfile.v0
<missing>      26 minutes ago   ENV ELASTIC_AGENT_OTEL=true                     0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   ENV GODEBUG=madvdontneed=1                      0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   ENV PATH=/usr/share/elastic-agent:/usr/local…   0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   ENV ELASTIC_CONTAINER=true                      0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   LABEL org.label-schema.build-date=2025-03-10…   0B        buildkit.dockerfile.v0
<missing>      26 minutes ago   RUN /bin/sh -c for iter in {1..10}; do      …   55.3MB    buildkit.dockerfile.v0
<missing>      26 minutes ago   ENV BEAT_SETUID_AS=elastic-agent                0B        buildkit.dockerfile.v0
<missing>      2 days ago       apko                                            13.5MB    This is an apko single-layer image
```

dev-tools/packaging/packages.yml
```diff
@@ -988,6 +1055,59 @@ specs:
           '{{.BeatName}}{{.BinaryExt}}':
             source: ./build/golang-crossbuild/{{.BeatName}}-{{.GOOS}}-{{.Platform.Arch}}{{.BinaryExt}}
     #### wolfi-based ####
+    #### Slim docker images ####
+    - os: linux
+      arch: amd64
+      types: [ docker ]
+      spec:
+        # The service image is always based on Wolfi
```
cmacknz, line 1063: This comment doesn't look like applies and is repeated for times

dev-tools/packaging/packages.yml
```diff
@@ -988,6 +1055,59 @@ specs:
...
+        files:
+          '{{.BeatName}}{{.BinaryExt}}':
+            source: ./build/golang-crossbuild/{{.BeatName}}-{{.GOOS}}-{{.Platform.Arch}}{{.BinaryExt}}
+    #### EDOT collector docker images ####
+    - os: linux
+      arch: amd64
+      types: [ docker ]
+      spec:
+        # The service image is always based on Wolfi
+        <<: *docker_wolfi_spec
```
cmacknz, line 1091: Using only wolfi here (for both edot and slim but I think it's worse for EDOT) I think limits people's ability to customize the image. Probably we will need UBI variants as well to handle that case.

pkoutsovasilis:
@cmacknz while thinking of our viable approach here I am thinking of some sort of spec where it would capture for each container image (and for any package artefact in general tbh) what should exist in it, what permissions and capabilities it should have, pretty much like a `catalog`. If we had that then testing all of these "flavors/variants" right after packaging would be pretty much straight forward. Also with some unit-test apprach we could even invoke upgrades only and make sure that we transition properly from catalog `A` to `B`. 

@pchila since I have read the latest RFC, please correct me if I am wrong but I think I recall reading something similar, to what I describe above, in it, right?

pchila:
> @pchila since I have read the latest RFC, please correct me if I am wrong but I think I recall reading something similar, to what I describe above, in it, right?

@pkoutsovasilis I guess you are referring to the original version of the RFC that originated https://github.com/elastic/elastic-agent/pull/4193
In that document the PackageManifest contained sets of globs matching expected files for both install and update operations, we could reuse the idea to collect components/input types/ full file lists during packaging and checking those during tests.

dev-tools/packaging/templates/docker/Dockerfile.edot.tmpl
pchila, line 1: Why is another Dockerfile template needed? What are the main differences with the regular Dockerfile.tmpl ?

dev-tools/mage/copy.go
```diff
@@ -63,6 +70,15 @@ func (t *CopyTask) isExcluded(src string) bool {
 			return true
 		}
 	}
+
+	return false
+}
+
+func (t *CopyTask) isDestinationExcluded(dst string) bool {
```
pchila, line 77: Is this function needed to exclude some paths? I remember seeing the same for the install flavors PR.
If this is the purpose shouldn't we try to do something that is useful in general?
See https://github.com/elastic/elastic-agent/issues/4588 for example.
Upcoming FIPS work will need some restriction on components included in packages as well.

cc @cmacknz

dev-tools/mage/dockerbuilder.go
```diff
@@ -103,8 +106,35 @@ func (b *dockerBuilder) exposePorts() []string {
...
+				}
+				return fmt.Errorf("failed reading spec file for component %q: %w", specFilename, err)
+			}
+
+			// create filter
+			allowedPaths, err := component.ParseComponentFiles(specContent, specFilename, true)
+			if err != nil {
+				return fmt.Errorf("failed computing component files %q: %w", specFilename, err)
+			}
+			checkFn, err = install.SkipComponentsPathWithSubpathsFn(allowedPaths)
```
pchila, line 128: We are introducing a dependency here between `dev-tools` and `elastic-agent` install package here.
Same with the `component.ParseComponentFiles` call a few lines above this.
I am not sure that this is the correct dependency direction between build/package code and the code being packaged

dev-tools/packaging/package_test.go
```diff
@@ -301,11 +306,15 @@ func checkNpcapNotices(pkg, file string, contents io.Reader) error {
 	return nil
 }
 
-func checkDocker(t *testing.T, file string) {
+func checkDocker(t *testing.T, file string) (string, int64) {
+	if !strings.HasPrefix(file, "edot-collector") {
+		return checkEdotCollectorDocker(t, file)
+	}
```
pchila, line 312: aside from the "special checks" that an Edot collector docker image has here, shouldn't the condition be 
```suggestion
	if strings.HasPrefix(file, "edot-collector") {
		return checkEdotCollectorDocker(t, file)
	}
```

dev-tools/mage/copy.go
cmacknz: Good point, this is excluding specific binaries which is exactly what the FIPS variant will require.

+1 to generalizing.

dev-tools/mage/dockerbuilder.go
michalpristas: we do have flavors so besides duplicating logic there needs to be some kind of dependency.

dev-tools/mage/copy.go
michalpristas: this is checkFn not bound with flavors. you can pass whatever you like here. i do pass one from flavors to spare myself logic of parsing manifest and extracting paths out of it. but if you have checkFn provider for fips use case you can have that. you can have compundCheckFnProvider accepting more checks and returning skip if any of them fails. this can be enhanced. but please let's not do it here as it's not really related and i'm not sure how many `mage package` runs i have left

dev-tools/packaging/packages.yml
```diff
@@ -988,6 +1053,56 @@ specs:
           '{{.BeatName}}{{.BinaryExt}}':
             source: ./build/golang-crossbuild/{{.BeatName}}-{{.GOOS}}-{{.Platform.Arch}}{{.BinaryExt}}
     #### wolfi-based ####
+    #### Slim docker images ####
+    - os: linux
+      arch: amd64
+      types: [ docker ]
+      spec:
+        <<: *docker_ubi_spec
```
cmacknz, line 1061: We should have both UBI and Wolfi, not either or. For both Slim and EDOT.

Wolfi has license restrictions but will have no OS CVEs because they are distroless.

UBI is OSS but will have OS CVEs in it.

AFAIK, we need both variants for every image we intend to be widely used.

dev-tools/packaging/templates/docker/docker-entrypoint.edot.tmpl
```diff
@@ -0,0 +1,14 @@
...
+
+set -eo pipefail
+
+# For information on the possible environment variables that can be passed into the container. Run the following
+# command for information on the options that are available.
+#
+# `./elastic-agent container --help`
+#
+
+exec {{ .BeatName }} otel "$@"
```
cmacknz, line 14: I think we can probably start using the `edot-colletor` image directly in our existing tests where ever we use ELASTIC_AGENT_OTEL: 

For example: https://github.com/michalpristas/elastic-agent/blob/29051e1bb7b5086beca0b453a0ac527dfeca7ead/testing/integration/kubernetes_agent_standalone_test.go#L207-L210

That will give you testing of the edot image. I would say we will eventually deprecate the ELASTIC_AGENT_OTEL variable in place of this container image rather than maintain it.

Let's add a log message that ELASTIC_AGENT_OTEL is deprecated and that you should use the edot-collector image directly instead so we can just delete it in 18 months or whatever time we consider reasonable since it will have been formally deprecated.

cmacknz: The Helm charts can also be updated to use the edot-collector image directly https://github.com/michalpristas/elastic-agent/blob/29051e1bb7b5086beca0b453a0ac527dfeca7ead/deploy/helm/edot-collector/kube-stack/values.yaml#L36-L38

pkg/testing/kubernetes/supported.go
```diff
@@ -72,6 +72,14 @@ var variants = []struct {
 		Name:  "service",
 		Image: "docker.elastic.co/beats-ci/elastic-agent-service",
 	},
+	{
+		Name:  "edot-collector",
```
cmacknz, line 76: If this has the same test coverage as the ELASTIC_AGENT_OTEL=true tests that exist already, instead of using edot in their place as https://github.com/elastic/elastic-agent/pull/7173/files#r1999460268 suggest perhaps we just need to delete those extra tests.

cmacknz:
The new docker containers got slightly larger after the latest change. Possibly just the change of base image is responsible for this.

```
❯ docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/elastic-agent:9.1.0-SNAPSHOT
1586579673

❯ docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/elastic-agent-slim:9.1.0-SNAPSHOT
1034258193

❯ docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/edot-collector:9.1.0-SNAPSHOT
948386470
```

Addressing the size still being ridiculous would be nice but might have to happen as a follow up.

dev-tools/packaging/packages.yml
```diff
@@ -301,6 +307,65 @@ shared:
         source: '{{ elastic_beats_dir }}/specs/connectors.spec.yml'
         mode: 0644
 
+  - &agent_docker_edot_spec
+    <<: *agent_binary_spec_no_components
+    docker_variant: 'edot-collector'
```
cmacknz, line 312: The third name change is the charm: https://github.com/elastic/elastic-agent/issues/6764#issuecomment-2730506335

`elastic-otel-collector`

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
```diff
@@ -16,7 +16,7 @@ crds:
   create: true # Install the OpenTelemetry Operator CRDs.
 defaultCRConfig:
   image:
-    repository: "docker.elastic.co/beats/elastic-agent"
+    repository: "docker.elastic.co/beats/elastic-otel-collector"
```
michalpristas, line 19: @pkoutsovasilis do we have test for helm charts?

pkoutsovasilis: So there are already some that utilise the EDOT variant [here](https://github.com/elastic/elastic-agent/blob/main/testing/integration/otel_helm_test.go#L31) and [here](https://github.com/elastic/elastic-agent/blob/main/testing/integration/kubernetes_agent_standalone_test.go#L206)

and there is also [this issue](https://github.com/elastic/elastic-agent/issues/7381) to bring in more 🙂

jlind23:
@michalpristas once this PR is merged we should follow up with an infra related PR just like this one https://github.com/elastic/infra/pull/42160/files

deploy/kubernetes/elastic-agent-kustomize/default/elastic-agent-standalone/base/elastic-agent-standalone-daemonset.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       #        mountPath: /etc/elastic-agent/inputs.d
       containers:
         - name: elastic-agent-standalone
-          image: docker.elastic.co/elastic-agent/elastic-agent:9.1.0
+          image: docker.elastic.co/elastic-agent/elastic-otel-collector:9.1.0
```
pkoutsovasilis, line 43: should we change this one to the otel-collector? 🤔 this is a k8s manifest of the "traditional" elastic-agent

.buildkite/bk.integration.pipeline.yml
```diff
@@ -277,7 +277,7 @@ steps:
         matrix:
           setup:
             variants:
-            - "basic,wolfi,complete,complete-wolfi,service,cloud"
+            - "basic,wolfi,complete,complete-wolfi,service,elastic-otel-collector-wolfi,elastic-otel-collector"
```
cmacknz, line 280: This is missing slim and slim-wolfi?

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
cmacknz, line 19: I think these might be used for demos and are the reference for how to use EDOT, so once this merges but before the first snapshot or release with these images are available this might be broken.

cmacknz: Updating them like this is definitely what we want to do, but this has some chance of causing a bit of chaos until the unified release changes we need actually happen everywhere

pkoutsovasilis: ok so if I decode correctly the above, we leave the manifests intact but we extend only the tests to use the new images?

dev-tools/packaging/package_test.go
```diff
@@ -115,9 +115,21 @@ func TestZip(t *testing.T) {
...
+
+	if len(dockers) > 0 {
+		// run only as part of packaging testing
+		require.Less(t, sizeMap["elastic-otel-collector"], sizeMap["elastic-agent-slim"], "unexpected size: %v", sizeMap)
+		require.Less(t, sizeMap["elastic-agent-slim"], sizeMap["elastic-agent"], "unexpected size: %v", sizeMap)
+
+		// run wolfi checks
+		require.Less(t, sizeMap["elastic-otel-collector-wolfi"], sizeMap["elastic-agent-slim-wolfi"], "unexpected size: %v", sizeMap)
+		require.Less(t, sizeMap["elastic-agent-slim-wolfi"], sizeMap["elastic-agent-wolfi"], "unexpected size: %v", sizeMap)
```
cmacknz, line 132: These checks need to be conditional on the selected variants.

```sh
DOCKER_VARIANTS=slim,edot-collector,basic EXTERNAL=true SNAPSHOT=true PACKAGES=docker PLATFORMS=linux/arm64 mage package
>> Testing package contents
package ran for 5m20.579235333s
Error: error running package_test.go: running "go test /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/dev-tools/packaging/package_test.go -root-owner -files /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/distributions/*" failed with exit code 1, stdout: --- FAIL: TestDocker (17.86s)
    package_test.go:120: /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/distributions/elastic-agent-9.1.0-SNAPSHOT-linux-arm64.docker.tar.gz
    package_test.go:120: /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/build/distributions/elastic-agent-slim-9.1.0-SNAPSHOT-linux-arm64.docker.tar.gz
    package_test.go:131:
                Error Trace:    /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/dev-tools/packaging/package_test.go:131
                Error:          "0" is not less than "0"
                Test:           TestDocker
                Messages:       unexpected size: map[elastic-agent:417474174 elastic-agent-slim:277013799]
FAIL
FAIL    command-line-arguments  18.339s
FAIL
```

The failure above happens because I didn't include the wolfi images.

cmacknz:
Taking another look at where the container size comes from, I misread the docker file and it isn't us chowning things it's the command that copies dependencies into the image.

```sh
docker inspect -f "{{ .Size }}" docker.elastic.co/elastic-agent/elastic-agent-slim:9.1.0-SNAPSHOT
1034520505 # 1.034 GB
```

Looking at the slim image it's got 872M in the elastic-agent directory of which 866MB is just the binaries themselves.

```sh
du -h /usr/share/elastic-agent/
872M    /usr/share/elastic-agent/

bash-5.1$ du -h data/elastic-agent-a77a7d/components
501M    data/elastic-agent-a77a7d/components/
bash-5.1$ du -h data/elastic-agent-a77a7d/elastic-agent
365M    data/elastic-agent-a77a7d/elastic-agent
```

Not much we can do in the container image itself besides making the actual binaries smaller. A lot of the elastic-agent binary size increase is from beats receivers.

Using the history command the only thing that jumps out as crazy is 103 MB from the `LABEL` command or things surrounding it. It doesn't show me exactly where this is coming from.

```sh
docker history docker.elastic.co/elastic-agent/elastic-agent-slim:9.1.0-SNAPSHOT
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
# The notice file is 6MB!
<missing>      27 minutes ago   COPY /usr/share/elastic-agent/NOTICE.txt /li…   5.9MB        buildkit.dockerfile.v0

# This is the actual binary files being copied to the image.
<missing>      27 minutes ago   COPY --chown=elastic-agent:elastic-agent /us…   912MB     buildkit.dockerfile.v0
       buildkit.dockerfile.v0

# This us installing runtime dependencies via microdnf
<missing>      23 hours ago     RUN /bin/sh -c for iter in {1..10}; do      …   12.8MB    buildkit.dockerfile.v0

# Where is this 103 MB coming from? The LABEL command ?!?
<missing>      5 days ago       /bin/sh -c #(nop) LABEL "build-date"="2025-0…   103MB
<missing>      5 days ago       /bin/sh -c . /cachi2/cachi2.env &&     rm -r…   0B
<missing>      5 days ago       /bin/sh -c #(nop) CMD ["/bin/bash"]             0B
<missing>      5 days ago       /bin/sh -c #(nop) COPY file:b37d593713ee21ad…   0B
<missing>      5 days ago       /bin/sh -c #(nop) COPY dir:b7138ac36fc7710c1…   0B
<missing>      5 days ago       /bin/sh -c #(nop) ENV container oci             0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL io.openshift.tags="m…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL io.openshift.expose-…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL io.k8s.display-name=…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL io.k8s.description="…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL description="The Uni…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL summary="Provides th…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL com.redhat.license_t…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL com.redhat.component…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL url="https://www.red…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL vendor="Red Hat, Inc…   0B
<missing>      5 days ago       /bin/sh -c #(nop) LABEL maintainer="Red Hat,…   0B
```

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
cmacknz: In this PR, yes. We should update these manifests, but only after there is a publicly available docker image to point them at. Otherwise, these charts will not work by default from main for however long it takes for the unified release job to publish the new container images.

cmacknz: Undo this change here and we'll do it as a follow up.

.buildkite/bk.integration.pipeline.yml
cmacknz: We need to smoke test it somehow to make sure it isn't completely broken. If there's a better way that's great but if this is the only way then we need it.

pkoutsovasilis: at the time being I cannot offer a better solution than introducing these variants in the k8s integration tests. However, I split up the variants into two groups (non-wolfi and wolfi ones) so k8s integration tests duration won't be hindered

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
pkoutsovasilis: ok no image changes in deploy. For now we do the necessary (kinda hacky for the env vars) changes only in our tests and we test both `edot-collector` and `edot-collector-wolfi`. Feel free to unresolve if you spot anything else 🙂

.buildkite/bk.integration.pipeline.yml
pkoutsovasilis: ok `slim` and `slim-wolfi` variants are not being tested under both kustomize and helm chart k8s integration tests. Feel free to unresolve if you spot anything else 🙂

dev-tools/packaging/packages.yml
pkoutsovasilis: I believe this has already been addressed and now the docker_variant for edot points to `elastic-otel-collector` and `elastic-otel-collector-wolfi`. Please unresolve if you spot anything else 🙂

dev-tools/packaging/templates/docker/Dockerfile.edot.tmpl
pkoutsovasilis: ok AFAICT this has been altered and now we use the same Dockerfile template as before. Please unresolve if you spot anything else 🙂

dev-tools/packaging/package_test.go
pkoutsovasilis: nice catch! I just push a commit that applies these checks only for variants that are actually built (https://github.com/elastic/elastic-agent/pull/7173/commits/56ab883490c3d3863356c0b1e33b6425cfe4f8f9). Feel free to unresolve if you spot anything else 🙂

dev-tools/mage/copy.go
pkoutsovasilis: guys have we reached a consensus here? maybe we should have a quick sync about it? 🙂

pchila: The checkFn is not the ideal solution for the FIPS upcoming work but since this is under time constraints to get in 9.0, I guess we can merge it as-is and rework it later for FIPS

dev-tools/mage/dockerbuilder.go
pchila: We also have a declarative YAML where package specs can be configured and that would not introduce a dev-tool -> elastic-agent code dependency.

Since this is under time constraints to get in 9.0, I guess we can merge it as-is and rework it later

dev-tools/mage/copy.go
michalpristas: cc @ebeahan

jlind23 (approved):

cmacknz (approved):

PR #7186, opened 2025-03-04
Speed up serverless Beat tests by not testing Metricbeat or Auditbeat

cmacknz:
## What does this PR do?

Stops testing Metricbeat and Auditbeat in the serverless Beat tests. The way this step was written it re-invoked the mage integration test target serially 3 times in a row which re-does provisioning unnecessarily. The tests should be rewritten eventually.

These tests also primarily test libbeat setup functionality that is shared amongst Beats, and we already aren't testing all the Beats, so let's just smoke test the one that covers the most functionality.

## Why is it important?

These tests take too long.

## Related issues

-

### DIFF
--- a/.buildkite/scripts/steps/beats_tests.sh
+++ b/.buildkite/scripts/steps/beats_tests.sh
@@ -63,15 +63,5 @@ popd
 
 # exit $TESTS_EXIT_STATUS
 
-echo "testing metricbeat..."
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (approved):

dliappis (approved):
LGTM

Issue #7060, opened 2025-02-27
[Flaky Test]: Kubernetes platform-related integration test failures


swiatekm:
We suspect that various K8s integration test failures caused by problems with Kubernetes itself, are in fact caused by resource contention due to running too many kind clusters on a single host. This issue exists to group these tests, so we can avoid the overhead of trying to debug them individually.

The short-term solution to this problem is using a larger instance for existing tests. The long-term solution is running these tests in parallel with finer-grained concurrency.

swiatekm:
https://github.com/elastic/elastic-agent/issues/7048 probably belongs here. The problem manifests as mysterious 403 responses from the API Server, but is most likely caused by the API Server not having enough resources.

swiatekm:
In #7050, we encountered the following:
```
=== Failed
=== FAIL: testing/integration TestOtelKubeStackHelm/helm_kube-stack_operator_standalone_agent_kubernetes_privileged (334.06s)
    otel_helm_test.go:122:
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1740576462608313681/elastic/elastic-agent-extended-testing/testing/integration/otel_helm_test.go:122
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1740576462608313681/elastic/elastic-agent-extended-testing/testing/integration/otel_helm_test.go:96
        	Error:      	Condition never satisfied
        	Test:       	TestOtelKubeStackHelm/helm_kube-stack_operator_standalone_agent_kubernetes_privileged
        	Messages:   	at least 4 agent containers should be checked
=== FAIL: testing/integration TestOtelKubeStackHelm (342.70s)
```

in https://buildkite.com/elastic/elastic-agent-extended-testing/builds/7694#01954270-d4c5-4926-8442-4861f27bedea.

swiatekm:
In #7044, just a plain runner failure without any additional information:

```
>>> (kubernetes-amd64-1289-complete-wolfi-kubernetes) Failed to execute tests on instance: %!s(<nil>)
```
in Kubernetes tests: https://buildkite.com/elastic/elastic-agent-extended-testing/builds/7684#01954204-430a-4656-864b-2496f3cc9ef8/56.

pkoutsovasilis:
thanks you for capturing all of this info @swiatekm , cc @dliappis @pazone

dliappis:
[...]

PR #7202, opened 2025-03-05
Fix a bug preventing kind cluster deletion in K8s tests

swiatekm:
## What does this PR do?

Fixes a bug preventing the kind provisioner for integration tests from deleting the clusters it creates. The provisioner was trying to delete the cluster using the wrong name.

This wasn't visible in logs because the `kind delete cluster` command always returns success, even if the cluster doesn't exist. I added a check for this to detect this situation in the future.

## Why is it important?

We should clean up resources after integration tests.

## Checklist

```
SNAPSHOT=true EXTERNAL=true PACKAGES=docker PLATFORMS="linux/amd64" mage -v  package
TEST_INTEG_CLEAN_ON_EXIT=true INSTANCE_PROVISIONER=kind STACK_PROVISIONER=stateful SNAPSHOT=true mage integration:kubernetesMatrix
```

## Related issues

Possibly related to #7060.

### DIFF
--- a/pkg/testing/kubernetes/kind/provisioner.go
+++ b/pkg/testing/kubernetes/kind/provisioner.go
@@ -212,13 +212,22 @@ func (p *provisioner) Clean(ctx context.Context, cfg common.Config, instances []
 	// doesn't execute in parallel for the same reasons in Provision
 	// multipass just cannot handle it
 	for _, instance := range instances {
-		func(instance common.Instance) {
-			err := p.deleteCluster(instance.ID)
-			if err != nil {
-				// prevent a failure from stopping the other instances and clean
-				p.logger.Logf("Delete instance %s failed: %s", instance.Name, err)
-			}
-		}(instance)
+		// manually check if the cluster exists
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (commented):
ty for highlighting and fixing this one @swiatekm Code changes LGTM, I don't expect any CI failures, but let's wait for it 🙂

swiatekm:
I wonder if this will affect #7060. I'm not sure if it's possible to leak kind clusters on buildkite runners, but if it is, then those failures may be caused by this bug. @dliappis @pazone do you think it could be related?

pazone:
> I wonder if this will affect #7060. I'm not sure if it's possible to leak kind clusters on buildkite runners, but if it is, then those failures may be caused by this bug. @dliappis @pazone do you think it could be related?

From the CI perspective, it's not critical to leave kind clusters on the CI runner. From the tests perspective, the clusters may consume memory and CPU and affect the tests execution.

pkoutsovasilis (approved):

dliappis (approved):
LGTM

PR #7267, opened 2025-03-07
Test integrations with basic installation

michalpristas:
CI check


testing/integration/network_traffic_monitoring_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func TestNetworkTraffic(t *testing.T) {
+	info := define.Require(t, define.Requirements{
+		Group: Fleet,
+		Stack: &define.Stack{},
+		Local: false, // requires Agent installation
+		Sudo:  true,  // requires Agent installation
+		OS: []define.OS{
+			{Type: define.Linux},
+			{Type: define.Windows},
```
pkoutsovasilis, line 42: if these two tests run concurrently at the same time the fact that they install the same package is it gonna affect any of them? I am not sure if the same as with endpoint test will happen here or whether kibana should handle that just wondering

testing/integration/auditd_monitoring_test.go
```diff
@@ -0,0 +1,120 @@
...
+		Privileged:     true,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	policyResp, err := tools.InstallAgentWithPolicy(ctx, runner.T(), installOpts, runner.agentFixture, runner.info.KibanaClient, basePolicy)
+	require.NoError(runner.T(), err)
+
+	_, err = tools.InstallPackageFromDefaultFile(ctx, runner.info.KibanaClient, "auditd_manager", "1.18.3", "auditd_package.json", uuid.Must(uuid.NewV4()).String(), policyResp.ID)
```
pkoutsovasilis, line 76: how to we plan to keep this `1.18.3` up to date?

testing/integration/network_traffic_monitoring_test.go
michalpristas: should use different policy right?

testing/integration/auditd_monitoring_test.go
michalpristas: i looked at other tests and we dont', this being up to date (very recent) is not a requirement as long as we use supported versions.

testing/integration/network_traffic_monitoring_test.go
blakerouse: Installing a package should work concurrently. If installing a package doesn't work concurrently and it errors back to the caller, then the caller should retry. Have we added retries?

michalpristas: @blakerouse retries in a test itself?

blakerouse: I wouldn't expect it to be something the test itself should do. Probably something the kibana client that the test uses should do. If it gets something like a conflict error (or what ever error that is reported by kibana, hopefully not a 500 error) then it should retry those calls automatically.

pkoutsovasilis (commented):
This looks good to me. Waiting for all comments to be resolved and then I will approve

testing/integration/network_traffic_monitoring_test.go
pkoutsovasilis: ok to me then it means that this PR doesn't have to include anything else, we just need to make sure that kibana client does retry. @blakerouse should we resolve this one?

blakerouse: Yeah I didn't mean for this to block this PR. Just a comment that packing installation should add retries in the case of failures working with Kibana API.

blakerouse (approved):

pkoutsovasilis (approved):

PR #7312, opened 2025-03-11
Enhancement/4811 fips to fips upgrade

kaanyalti:
- Enhancement

## What does this PR do?

Updates the upgrade process to prevent fips to non-fips upgrades

## Why is it important?

upgrading to non-fips agent is not going to be supported

## Checklist

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/4811

### DIFF
--- a/internal/pkg/agent/application/upgrade/upgrade.go
+++ b/internal/pkg/agent/application/upgrade/upgrade.go
@@ -54,8 +54,11 @@ var agentArtifact = artifact.Artifact{
 	Artifact: "beats/" + agentName,
 }
 
-var ErrWatcherNotStarted = errors.New("watcher did not start in time")
-var ErrUpgradeSameVersion = errors.New("upgrade did not occur because it is the same version")
+var (
+	ErrWatcherNotStarted     = errors.New("watcher did not start in time")
+	ErrUpgradeSameVersion    = errors.New("upgrade did not occur because it is the same version")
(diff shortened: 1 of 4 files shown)


pchila (approved):
LGTM

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -54,8 +54,11 @@ var agentArtifact = artifact.Artifact{
 	Artifact: "beats/" + agentName,
 }
 
-var ErrWatcherNotStarted = errors.New("watcher did not start in time")
-var ErrUpgradeSameVersion = errors.New("upgrade did not occur because it is the same version")
+var (
+	ErrWatcherNotStarted     = errors.New("watcher did not start in time")
+	ErrUpgradeSameVersion    = errors.New("upgrade did not occur because it is the same version")
+	ErrFipsNotUpgradedToFips = errors.New("cannot upgrade from a fips compliant agent to a non-compliant one")
```
simitt, line 60: The wording of the error should cover both cases:
(1) fips -> non-fips
(2) non-fips -> fips

In `checkUpgrade` it actually fails for both cases, but the error that will be logged would be confusing if someone tries to upgrade from `non-fips` to `fips`. 

Also, we should be careful with wording and avoid `fips compliant`. I would suggest using a more neutral language, something along the lines of `cannot switch fips mode when upgrading`.

kaanyalti: I'll create a PR to update the wording here

PR #7327, opened 2025-03-11
[otel/kube-stack] Update EDOT SDK k8s auto-instrumentation images to their latest versions

trentm:
This also updates the vanilla OTel Go image version to its latest.

---

The versions of the EDOT language SDK images being used with the OTel Operator were quite old.
This *manually* updates them to the current latest.

/cc @jackshirazi, @xrmx, @dmathieu (as my best guess for the authority on the OTel Operator auto-instrumentation image for each of the languages I've updated here)

---

Is there a good place to discuss how/when/who should be **regularly** updating these versions?
Or perhaps use the `:latest` Docker tag, though that, of course, changes over time.


trentm:
> This pull request does not have a backport label. Could you fix it @trentm? 🙏

I'd appreciate guidance on this. I don't know what backport targets would be appropriate for this PR.

pkoutsovasilis:
hey @trentm 👋 so `backport-active-all` will open backport PRs, when this one gets merged, to `9.0`, `8.x`, `8.18`, `8.17` and `8.16` branches. I have a feeling that this change shouldn't go that far back but I am not 100% sure. As this PR affects the edot collector helm chart, I will need a chime in from the devs that have contributed to it @rogercoll @carsonip @swiatekm

carsonip:
Sorry I would love to be more helpful here but I'm not familiar with the expectation of auto instrumentation SDK version bump risk and reward balance in backported versions.
cc @AlexanderWert

jackshirazi:
yeah, these can be backported to any/every version and will work fine (and be better for the customer)

swiatekm:
I'm reluctant to do this right before the final RC for 9.0.0 and 8.18.0 unless someone exhaustively tests these first. We don't have enough automated testing for this configuration to be confident we're not breaking something. See https://github.com/elastic/elastic-agent/issues/7381 for a very recent example.

jackshirazi:
To be clear what this change is, it is changing the default agent docker images that are downloaded into the kubernetes cluster that the customer is running, which can then be applied for auto-instrumentation of customer applications. The customer is able to override these docker images very easily. Sticking with the existing values means that customers onboarding will install old versions of the agents, applying the change means the customers install the latest released (and definitely tested) versions of the agents

When we defined these values, we considered using `latest` as the version for all agents instead of defined versions, and this would have been fine, but decided that it was more transparent for the customers to be on a specific default version rather than getting the latest release. But that was only for onboarding transparency, not safety. Having `latest` would have been perfectly safe.

All of which is to say, this not a risk on the Elastic infrastructure side, it only affects agent versions downloaded into customer clusters, and the more recent version the better where this is concerned

swiatekm:
@jackshirazi I understand what this change accomplishes. I'm saying I'm not confident this won't break something for customers. If @elastic/ingest-otel-data, who are effectively the owners of this config, are confident the change is safe, I don't mind merging it.

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
```diff
@@ -605,12 +605,12 @@ instrumentation:
...
-    image: docker.elastic.co/observability/elastic-otel-javaagent:1.0.0
+    image: docker.elastic.co/observability/elastic-otel-javaagent:1.3.0
   nodejs:
-    image: docker.elastic.co/observability/elastic-otel-node:0.4.1
+    image: docker.elastic.co/observability/elastic-otel-node:0.7.0
   dotnet:
     image: docker.elastic.co/observability/elastic-otel-dotnet:edge
   python:
-    image: docker.elastic.co/observability/elastic-otel-python:0.3.0
+    image: docker.elastic.co/observability/elastic-otel-python:0.6.1
```
xrmx, line 614:
```suggestion
    image: docker.elastic.co/observability/elastic-otel-python:0.8.0
```

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -935,12 +935,12 @@ instrumentation:
...
-    image: docker.elastic.co/observability/elastic-otel-javaagent:1.0.0
+    image: docker.elastic.co/observability/elastic-otel-javaagent:1.3.0
   nodejs:
-    image: docker.elastic.co/observability/elastic-otel-node:0.4.1
+    image: docker.elastic.co/observability/elastic-otel-node:0.7.0
   dotnet:
     image: docker.elastic.co/observability/elastic-otel-dotnet:edge
   python:
-    image: docker.elastic.co/observability/elastic-otel-python:0.3.0
+    image: docker.elastic.co/observability/elastic-otel-python:0.6.1
```
xrmx, line 944:
```suggestion
    image: docker.elastic.co/observability/elastic-otel-python:0.8.0
```

trentm:
@elastic/ingest-otel-data can we merge this?  I'm not clear on what versions this should be backported to.

girodav (approved):
I am fine with merging this, after syncing with the other team members. I think the risk is relatively low, considering this PR only affects auto-instrumentation and not the collector configuration or elastic-agent image used.

We should backport to at least 8.19 and 9.1 at the very least, but also 9.0.1 and 8.18.1.

deploy/helm/edot-collector/kube-stack/values.yaml
```diff
@@ -935,12 +935,12 @@ instrumentation:
     type: parentbased_traceidratio # Sampler type
     argument: "1.0" # Sampling rate set to 100% (all traces are sampled).
   java:
-    image: docker.elastic.co/observability/elastic-otel-javaagent:1.0.0
+    image: docker.elastic.co/observability/elastic-otel-javaagent:1.3.0
   nodejs:
-    image: docker.elastic.co/observability/elastic-otel-node:0.4.1
+    image: docker.elastic.co/observability/elastic-otel-node:1.0.0
   dotnet:
     image: docker.elastic.co/observability/elastic-otel-dotnet:edge
```
girodav, line 942: Any reasons why we do not use versioned tags here as done for the other SDKs?

stevejgordon: We'll have a look at the tagging process for the release. Does this block merging this PR? Or can we follow up by including a version once we resolve the process during the build? Do we need to bump this manually after every release? Would `latest` not also be suitable? I'm not familiar with how these images are used.

girodav: > Does this block merging this PR? Or can we follow up by including a version once we resolve the process during the build?

I'd prefer to update all versions within this PR, if feasible, considering that we need to backport it as well. Are we also sure that `edge` points to the latest SDK anyway?

> Do we need to bump this manually after every release? 

These configuration files are bundled with the Elastic Agent, so they follow the regular Elastic stack release cycle. We should just ensure that the latest active stack versions are using the latest SDKs.

> Would latest not also be suitable?

I would avoid using latest to maintain consistency with the versioned tags used for other SDKs. Additionally, it can make debugging harder if an issue arises.

stevejgordon: Okay, thanks @girodav. I've identified the fix on the release flow, so hopefully we'll have a correctly tagged image today, and we can update this to use that tag.

> These configuration files are bundled with the Elastic Agent, so they follow the regular Elastic stack release cycle. We should just ensure that the latest active stack versions are using the latest SDKs.

Perhaps we can automate that part at some stage.

stevejgordon: @trentm There's now a `docker.elastic.co/observability/elastic-otel-dotnet:1.0.1` image if you want to use that tag as part of this PR.

xrmx:
```suggestion
    image: docker.elastic.co/observability/elastic-otel-dotnet:1.0.1
```

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
```diff
@@ -605,12 +605,12 @@ instrumentation:
     type: parentbased_traceidratio # Sampler type
     argument: "1.0" # Sampling rate set to 100% (all traces are sampled).
   java:
-    image: docker.elastic.co/observability/elastic-otel-javaagent:1.0.0
+    image: docker.elastic.co/observability/elastic-otel-javaagent:1.3.0
   nodejs:
-    image: docker.elastic.co/observability/elastic-otel-node:0.4.1
+    image: docker.elastic.co/observability/elastic-otel-node:1.0.0
   dotnet:
     image: docker.elastic.co/observability/elastic-otel-dotnet:edge
```
xrmx, line 612:
```suggestion
    image: docker.elastic.co/observability/elastic-otel-dotnet:1.0.1
```

girodav (approved):
LGTM

xrmx (approved):

PR #7328, opened 2025-03-11
Allow switching inputs to the otel runtime manager

swiatekm:
## What does this PR do?

Allows switching individual inputs to the Otel runtime manager. This is done by introducing a new configuration field: `_runtime_experimental` with possible values of `process` (the default) and `otel`. It is no longer necessary to modify agent code and build a local binary to run inputs as beats receivers in an otel collector.

This change is only intended for testing, and the configuration is not stable. See https://github.com/elastic/ingest-dev/issues/5117 for more details. Its only effect on users not opting in to using the otel runtime is that diagnostics now unconditionally emit an additional file containing the final otel collector configuration.

In terms of implementation, we add a new field to Components in the component model, and ensure inputs with a different runtime manager setting end up in different Components.

## Why is it important?

It will make testing beats receivers in agent a lot easier. The component model machinery is also easy to repurpose even if we ultimately decide this needs to be configured differently.

## Checklist

Build the agent binary and use the following configuration:

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [localhost:9200]
    username: elastic
    password: ""
    preset: "balanced"

inputs:
  # Collecting log files
  - type: filestream
    # Input ID allowing Elastic Agent to track the state of this input. Must be unique.
    id: your-input-id
    use_output: default
    _runtime_experimental: otel
    streams:
      - id: input2
        data_stream:
          dataset: myfile
        paths:
          - /tmp/input.txt
        prospector.scanner.fingerprint.enabled: false
        file_identity.native: ~
```

## Related issues

Implements the agent part of https://github.com/elastic/ingest-dev/issues/5117. It's possible the actual config key and values will change, but the changes to the component model should be universal.


testing/integration/otel_test.go
```diff
@@ -979,6 +979,142 @@ service:
...
+	var fixtureWg sync.WaitGroup
+	cmd, err := fixture.PrepareAgentCommand(ctx, nil)
+	require.NoError(t, err, "cannot prepare Elastic-Agent command: %w", err)
+
+	output := strings.Builder{}
+	cmd.Stderr = &output
+	cmd.Stdout = &output
+
+	err = cmd.Start()
+	require.NoError(t, err)
```
swiatekm, line 1080: Is this the right way to run a local agent in an integration test? I tried using `fixture.Run()`, but it resulted in errors involving the control protocol. I copied this from the event logging test, but I'm not sure why the naive approach failed.

testing/integration/otel_test.go
```diff
@@ -979,6 +979,138 @@ service:
...
+
+	cmd, err := fixture.PrepareAgentCommand(ctx, nil)
+	require.NoError(t, err, "cannot prepare Elastic-Agent command: %w", err)
+
+	output := strings.Builder{}
+	cmd.Stderr = &output
+	cmd.Stdout = &output
+
+	err = cmd.Start()
+	require.NoError(t, err)
```
swiatekm, line 1079: Is this the right way to run a local agent in an integration test? I tried using `fixture.Run()`, but it resulted in errors involving the control protocol. I copied this from the event logging test, but I'm not sure why the naive approach failed.

blakerouse: `Run` does work. See here - https://github.com/elastic/elastic-agent/blob/a283180e6dfacbcc52118171b63f9b28d9fb9b10/testing/integration/otel_test.go#L302

swiatekm: Ah, I missed that you need to pass the configuration in via a state. It's actually written in the method docstring too, I just didn't read it fully. :sweat_smile: Thanks for the tip!

pkoutsovasilis (commented):
since the following is the case 

> This change is only intended for testing, and the configuration is not stable. See https://github.com/elastic/ingest-dev/issues/5117 for more details. Its only effect on users not opting in to using the otel runtime is that diagnostics now unconditionally emit an additional file containing the final otel collector configuration.

any reason we aren't guarding this behind a [Feature Flag](https://www.elastic.co/guide/en/fleet/current/elastic-agent-standalone-feature-flags.html)? cc @cmacknz

swiatekm:
> since the following is the case
> 
> > This change is only intended for testing, and the configuration is not stable. See [elastic/ingest-dev#5117](https://github.com/elastic/ingest-dev/issues/5117) for more details. Its only effect on users not opting in to using the otel runtime is that diagnostics now unconditionally emit an additional file containing the final otel collector configuration.
> 
> any reason we aren't guarding this behind a [Feature Flag](https://www.elastic.co/guide/en/fleet/current/elastic-agent-standalone-feature-flags.html)? cc @cmacknz

I'm not against adding an undocumented feature flag for it. I was trying to minimize the set of changes I'd potentially need to redo in the future, but a flag isn't that much work.

swiatekm:
> since the following is the case
> 
> > This change is only intended for testing, and the configuration is not stable. See [elastic/ingest-dev#5117](https://github.com/elastic/ingest-dev/issues/5117) for more details. Its only effect on users not opting in to using the otel runtime is that diagnostics now unconditionally emit an additional file containing the final otel collector configuration.
> 
> any reason we aren't guarding this behind a [Feature Flag](https://www.elastic.co/guide/en/fleet/current/elastic-agent-standalone-feature-flags.html)? cc @cmacknz

We discussed this during a meeting today, and decided on the following:

* We don't need a flag specifically to enable the new setting at all, but we decided to make it less accessible. It's not called `_runtime_experimental` to discourage accidental use.
* We do want to have a feature flag controlling the default runtime, but that can come in a separate PR.

Are you ok with this approach @pkoutsovasilis ?

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
```diff
@@ -1009,6 +1008,7 @@ inputs:
   - id: test-input
     type: filestream
     use_output: default
+    _runtime_experimental: otel
```
pchila, line 1011:
```suggestion
    # This is temporary and it's gonna be removed/renamed before beats receivers are GA
    _runtime_experimental: otel
```
as a personal preference I think the simple `runtime` key was a better choice

testing/integration/otel_test.go
```diff
@@ -979,6 +981,143 @@ service:
...
+		InputPath  string
+		ESEndpoint string
+		ESApiKey   string
+	}
+	esEndpoint, err := getESHost()
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
+	configTemplate := `inputs:
```
pchila, line 1032: nit: this could be a constant

testing/integration/otel_test.go
```diff
@@ -979,6 +981,143 @@ service:
...
+				return false
+			},
+		})
+	}()
+
+	// Make sure find the logs
+	actualHits := &struct{ Hits int }{}
+	assert.EventuallyWithT(t, func(ct *assert.CollectT) {
+		output, execErr := fixture.ExecStatus(ctx)
+		assert.NoError(ct, execErr, "expected health check to succeed: %w, status: %v", execErr, output)
```
pchila, line 1087: The `assert.CollectT` is there only to capture the error from the assert on execErr? I don't see it used for the `assert.Equal()`

testing/integration/otel_test.go
swiatekm: Could be, but it's only used once in this test. I could see moving it to a file, it's a bit annoying to work with as a multiline string.

testing/integration/otel_test.go
swiatekm: Good catch, that's a mistake.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1473,6 +1465,7 @@ func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
 		if err != nil {
 			c.logger.Errorf("failed to generate otel config: %v", err)
 		}
+		c.logger.Warnf("The Otel runtime manager is HIGHLY EXPERIMENTAL and only intended for testing. Use at your own risk.")
```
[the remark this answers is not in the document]
swiatekm: Added the ids to this warning.

testing/integration/otel_test.go
swiatekm: Fixed.

pkoutsovasilis:
> We discussed this during a meeting today, and decided on the following:
> 
> * We don't need a flag specifically to enable the new setting at all, but we decided to make it less accessible. It's now called `_runtime_experimental` to discourage accidental use.
> * We do want to have a feature flag controlling the default runtime, but that can come in a separate PR.
> 
> Are you ok with this approach @pkoutsovasilis ?

ok @swiatekm since this was discussed and agreed to be in this way, no argument on my end. Waiting for the CI to be green 🙂

testing/integration/otel_test.go
```diff
@@ -979,6 +981,143 @@ service:
...
+	}, 1*time.Minute, time.Second)
+	assert.Eventually(t,
+		func() bool {
+			findCtx, findCancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer findCancel()
+
+			docs, err := estools.GetLogsForIndexWithContext(findCtx, info.ESClient, index, map[string]interface{}{
+				"log.file.path": inputFilePath,
+			})
+			require.NoError(t, err)
```
pkoutsovasilis, line 1098: qq: if this require fails will the assert.Eventually retry again?! and even if it does, is the test failed now or not? 🤔

swiatekm: The whole test will fail. I should probably use EventuallyWithT here, which would give a more reasonable behaviour.

swiatekm: Done.

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
swiatekm: I like this one better, as it immediately communicates that the name is temporary. My primary concern was stopping users from depending on it.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1473,6 +1465,11 @@ func (c *Coordinator) updateOtelManagerConfig(model *component.Model) error {
 		if err != nil {
 			c.logger.Errorf("failed to generate otel config: %v", err)
 		}
+		componentIDs := make([]string, len(model.Components))
```
pkoutsovasilis, line 1468: if you want to use `append` 
```suggestion
		componentIDs := make([]string, 0, len(model.Components))
```

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
pkoutsovasilis: @pchila are we good to resolve this issue? 🙂

testing/integration/otel_test.go
pkoutsovasilis: @pchila are we good to resolve this issue? 🙂

pchila: sure this is not blocking

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
pchila: Not blocking, resolving...

pkoutsovasilis (approved):

pchila (approved):

Issue #6763, opened 2025-02-07
Issues running Elastic-Agent OTel Mode in Docker


strawgate:
It looks like `./otelcol` is not executable in our docker image:

```
bill.easton@billeastons-MBP OTel Benchmark % docker run --entrypoint /bin/bash -v ./cases/edot-es.yml:/usr/share/elastic-agent/elastic-agent.yml -it elastic/elastic-agent:8.17.1
elastic-agent@145dfd9960a7:~$ ./otelcol
bash: ./otelcol: Permission denied
```

And it looks like the two ways I would expect to launch OTel do not work:

Passing `otel` as the command when running the elastic-agent container does not start OTel mode.
```
bill.easton@billeastons-MBP OTel Benchmark % docker run -it elastic/elastic-agent:8.17.1 otel                                      
agent container initialisation - effective capabilities
agent container initialisation - ambient capabilities
agent container initialisation - chown paths
{"log.level":"info","@timestamp":"2025-02-07T17:35:34.308Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.runElasticAgent","file.name":"cmd/run.go","file.line":195},"message":"Elastic Agent started","log":{"source":"elastic-agent"},"process.pid":7,"agent.version":"8.17.1","agent.unprivileged":true,"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2025-02-07T17:35:34.522Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade.InvokeWatcher","file.name":"upgrade/rollback.go","file.line":149},"message":"agent is not upgradable, not starting watcher","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2025-02-07T17:35:34.523Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.runElasticAgent","file.name":"cmd/run.go","file.line":284},"message":"APM instrumentation disabled","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2025-02-07T17:35:34.526Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/application.New
```

Ensuring that `elastic-agent.yml` only contains OTel config keys also does not result in running in OTel mode.

andrzej-stencel:
[...]

PR #7345, opened 2025-03-12
fix(docker): make `otelcol` executable

andrzej-stencel:
## What does this PR do?

Makes the `otelcol` script executable in the Docker image.

## Why is it important?

The `otelcol` script exists in packaged Elastic Agent tarball when downloaded from https://www.elastic.co/downloads/elastic-agent. For consistency, it would be good for the script to also be available in the Docker image, so that the user can do one of the following:

```console
$ docker run -it --rm --entrypoint=/bin/bash docker.elastic.co/elastic-agent/elastic-agent:x.y.z
bash-5.1$ ./otelcol
```

or

```console
$ docker run -it --rm --entrypoint=otelcol docker.elastic.co/elastic-agent/elastic-agent:x.y.z
```

On the other hand, one could argue that this makes it possible to do one thing (running EDOT Collector in Docker) in a couple different ways - by using the ELASTIC_AGENT_OTEL env var or by using the `otelcol` script. If the decision is to not make the `otelcol` script available in the Docker image, we should remove it from the image.

## Checklist

In terms of tests: Do we have a step in CI where we can run a test on the built docker image? I'd like to run a custom `docker run` command on the built container and verify the command's output.

```console
DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64 PACKAGES=docker mage -v package
docker run -it --rm --entrypoint=otelcol docker.elastic.co/elastic-agent/elastic-agent:9.1.0-SNAPSHOT
```

The `docker run` command should successfully start EDOT Collector. The logs from the container should start with "Starting in otel mode".

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/6763


swiatekm (commented):
Could we add a test to verify this? Seems like it'd be straightforward here: https://github.com/elastic/elastic-agent/blob/6b1691205757b83a799a3ca328b1ef55a244b165/dev-tools/packaging/package_test.go#L311.

swiatekm (approved):

PR #7360, opened 2025-03-12
FIPS complaint agent file vault

michel-laterman:
## What does this PR do?

Add elastic file vault seed implementation to allow variable length salt sizes.
FIPS distributions will strictly use the new `.seedV2` format with the salt size set to 16 for FIPS compliance.
When not in FIPS only the (V1) `.seed` file is used.

Seed file version is as follows:
- `.seed` files remain unchanged from the current released implementation. Used in non-FIPS distribution.
- `.seedV2` files are new; they contain the seed followed by the salt size (as a little endian uint32 value). This allows specify the salt size in use by the agent vault. Used by the FIPS distribution.

## Why is it important?

Current file vault does not meet FIPS compliance.

## Checklist

## Disruptive User Impact

No replacement path from a non-FIPS to a FIPS-agent (or vice-versa).


internal/pkg/agent/vault/seed.go
```diff
@@ -16,14 +17,23 @@ import (
 )
 
 const (
+	// seedFile is len(aesgcm.AES256) and contains only the random seed
+	// A default salt size of 8 is used with this seed file
 	seedFile = ".seed"
+	// seedFileV2 is len(aesgcm.AES256+4) and contains the random seed followed by a non-zero salt size (little endian uint32)
+	seedFileV2 = ".seedV2"
```
michel-laterman, line 24: Currently, we have hard-coded values for iterations, and keyLen. Should we add these to the seed file to allow for changes?

michel-laterman (commented):
This is currently a draft as I need to do testing.
I may need to have a different default salt size for fips/nonfips distributions to start.

My thought is that this approach will give us the ability to migrate from saltSize 8 to 16 which will be useful if customers want to migrate agents (they will need to go from agent version `N-1 -> N -> N-fips`).

Migration is not part of this PR.

blakerouse:
> This is currently a draft as I need to do testing. I may need to have a different default salt size for fips/nonfips distributions to start.
> 
> My thought is that this approach will give us the ability to migrate from saltSize 8 to 16 which will be useful if customers want to migrate agents (they will need to go from agent version `N-1 -> N -> N-fips`).

Can you explain this more? Why not just `N-1 -> N-fips`? Could it be made that way?

michel-laterman:
> > This is currently a draft as I need to do testing. I may need to have a different default salt size for fips/nonfips distributions to start.
> > My thought is that this approach will give us the ability to migrate from saltSize 8 to 16 which will be useful if customers want to migrate agents (they will need to go from agent version `N-1 -> N -> N-fips`).
> 
> Can you explain this more? Why not just `N-1 -> N-fips`? Could it be made that way?

I'll admit that i'm not 100% sure on this; I'm assuming that a FIPS agent can not run the migration as it would need to use non-compliant crypto settings in order to do so

michel-laterman:
Right now the integration tests fail as one of the upgrade tests tries to go from an artifact that is built from this PR to the latest snapshot.
In terms of changes from this PR it goes from a vault with the V2 seed file + vault, to no seed file detected.
After we "upgrade" the vault contains:

```
drwxr-x--- 2 root root 4096 Mar 12 17:14 .
drwxrwx--- 5 root root 4096 Mar 12 17:14 ..
-rw------- 1 root root    0 Mar 12 17:11 .lock
-rw------- 1 root root   32 Mar 12 17:14 .seed
-rw------- 1 root root   36 Mar 12 17:11 .seedV2
-rw------- 1 root root  125 Mar 12 17:14 8c6b8035489e04adf7302525047eb100b54b791916b09a3cc5c73b8b59bb09ee # V1 vault
-rw------- 1 root root  133 Mar 12 17:11 8fee45c194974431fa8a0cfa3355ed15e6c0e742fdfb4992767198a77dd16bd3 # V2 vault
```

The upgraded agent fails to start with `could not load agent info: could not get agent info from store: fail to read configuration /opt/Elastic/Agent/fleet.enc for the agent: fail to decode bytes: cipher: message authentication failed`

internal/pkg/agent/vault/seed.go
```diff
@@ -16,62 +17,116 @@ import (
...
+	// V1 seed fallback
+	// getSeedV1 will return ErrNotExist when in FIPS mode.
+	pass, err = getSeedV1(path)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return nil, 0, err
+		}
+	}
+	if len(pass) != 0 {
+		return pass, saltSizeV1, nil
```
blakerouse, line 104: Based on my reading here this means that if v2 seed is not present and only v1 is present then it will only read the v1 seed and never create the v2 seed. I am reading that correctly, yes?

How does that work when upgrading from a non-fips to a fips agent? Does it re-create the vault with a new large seed and move all of the data over from non-fips to fips compliant?

blakerouse (commented):
I really want to just understand what happens here so

internal/pkg/agent/vault/seed.go
```diff
@@ -16,62 +17,116 @@ import (
 )
 
 const (
+	// seedFile is len(aesgcm.AES256) and contains only the random seed
+	// A default salt size of 8 is used with this seed file
 	seedFile = ".seed"
+	// seedFileV2 is len(aesgcm.AES256)+4 and contains the random seed followed by a non-zero salt size (little endian uint32)
+	seedFileV2     = ".seedV2"
+	seedFileV2Size = int(aesgcm.AES256) + 4
```
pkoutsovasilis, line 25: nit: define a const of `int(aesgcm.AES256)` and derive seedFileV2Size from it + 4. Then use the AES256 const when reading the file in `getSeedV2` func

internal/pkg/agent/vault/seed.go
simitt: >I will pose the question, should we support upgrades from a non-fips to a fips agent? 

+1 on this question; I believe @nimarezainia already confirmed in a different conversation that this is not required - @nimarezainia keep me honest here.

michel-laterman: I'll remove all v2 support from non-fips agents

blakerouse (approved):
With product saying that we don't need an upgrade path then I am good with this PR. I think this decision will be re-visited in the future. I could see an organization with 100+ agents having to re-deploy to get FIPS as being an issue.

I would think we could support upgrade to non-FIPS that adjusts the size of the salt to FIPS compliant, then upgrade to a FIPS compliant binary that is now using the salt. It is interesting topic if a FIPS binary upgrade using non-FIPS compliant path to support an upgrade is not non-FIPS if then it uses FIPS from that upgrade. To me this is an interesting discussion.

pkoutsovasilis:
interestingly enough I do see a CI failure with this error

```
POST git-upload-pack (305 bytes)
--
  | From https://github.com/elastic/elastic-agent
  | * branch                  main       -> FETCH_HEAD
  | = [up to date]            main       -> origin/main
  | Warning: you are leaving 9 commits behind, not connected to
  | any of your branches:
  |  
  | c1beeb9557 Remove V2 seed functions from non-FIPS agents
  | 83a2014771 Provide v1 seedfile when v2 size matches
  | 18493241fa Use saltSize of 8 for non-fips
  | e2440300fe Fix unit tests
  | ... and 5 more.
  |  
  | If you want to keep them by creating a new branch, this may be a good time
  | to do so with:
  |  
  | git branch <new-branch-name> c1beeb9557
  |  
  | HEAD is now at 2336fb88bf [CI] Grant build and read permission for serverless beats pipeline (#7449)
  | Current branch: HEAD
  | Switched to a new branch 'pr_merge_7360'
  | New branch created: pr_merge_7360
  | Auto-merging internal/pkg/agent/vault/vault_file_notwindows.go
  | CONFLICT (content): Merge conflict in internal/pkg/agent/vault/vault_file_notwindows.go
  | Automatic merge failed; fix conflicts and then commit the result.
  | Merge failed: 1
``` 

but GH reports no conflicts 🤔

simitt (changes requested):
>Add elastic file vault seed implementation to allow variable length salt sizes.
>.seedV2 files are new; they contain the seed followed by the salt size (as a little endian uint32 value). This allows specify the salt size in use by the agent vault. Used by the FIPS distribution.

I don't see how the vault salt size can be set, it always uses the default of size 16 for fips. It's also not clear to me why it would need to be able to deal with variable salt size? 

Looking at the current code, are all these changes really necessary and relevant? 
- now that non-fips is unchanged, what is the reasoning behind introducing a `.seedV2` file for fips rather than also using the `.seed` file? 
- The only place where I see a salt size before the changes is in `vault_file_notwindows.go`. It can easily be that I am overlooking something, so please push back if that is the case, but wouldn't the simplest required change be to only switch the `saltSize=16` in this file for the `fips` case and keep everything else untouched? Can you elaborate why this wouldn't be sufficient or desired?

simitt (approved):

internal/pkg/agent/vault/seed.go
nimarezainia: yes agreed. I can't see a scenario where the user goes between the two flavours. That is not to say someone won;t come up with one (complaining they don;t have access to the machines to do a reinstall). I think instinctively users would understand that fip is a new binary and needs a new install.  For Fleet managed agents, upgrading to a different falvor is not even possible. For standalone users the 'upgrade' command doesn't allow a different binary to be applied. 

I would say it's safe to assume moving to a fips enabled image needs a new installation.

PR #7367, opened 2025-03-12
Add AgentMonitoring Test for Beat Receivers work

leehinman:
## What does this PR do?

Adds Integration test that:

1. Creates and install policy with just monitoring
2. Download the policy, add the API key
3. Installs elastic-agent without enrolling in fleet
4. Make sure logs and metrics for agent monitoring are being received

This is a baseline test that will be expanded to compare agent
monitoring when using beats vs beat receivers.

## Why is it important?

Provide a baseline to compare Beats receivers vs Beats for agent monitoring

## Checklist

## Disruptive User Impact

None.  Integration test only

``` shell
mage integration:single TestAgentMonitoring
```

## Related issues

- Closes elastic/ingest-dev#4874

### DIFF
--- a/testing/integration/beat_receivers_test.go
+++ b/testing/integration/beat_receivers_test.go
@@ -1,0 +1,186 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+//go:build integration
+
+package integration
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"net/http"
+	"testing"
+	"time"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/elastic/elastic-agent-libs/kibana"
+	"github.com/elastic/elastic-agent-libs/testing/estools"
+	atesting "github.com/elastic/elastic-agent/pkg/testing"
+	"github.com/elastic/elastic-agent/pkg/testing/define"
+	"github.com/elastic/elastic-agent/pkg/testing/tools/testcontext"
+
+	"github.com/gofrs/uuid/v5"
+	"github.com/stretchr/testify/require"
+)
+
+// TestAgentMonitoring is a test to provide a baseline for what
+// elastic-agent monitoring looks like with classic monitoring.  It
+// will be expanded in the future to compare with beats receivers for
+// elastic-agent monitoring.
+func TestAgentMonitoring(t *testing.T) {
+	// Flow
+	// 1. Create and install policy with just monitoring
+	// 2. Download the policy, add the API key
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (commented):
code changes LGTM, gonna wait for CI to be green. Also do we wanna add the `skip-changelog` label here?

khushijain21:
So there is already a test for "metrics" part of monitoring [here](https://github.com/elastic/elastic-agent/blob/main/testing/integration/metrics_monitoring_test.go). We could extend this for "logs" as well - instead of having two tests doing the same thing. wdyt?

pkoutsovasilis:
> So there is already a test for "metrics" part of monitoring [here](https://github.com/elastic/elastic-agent/blob/main/testing/integration/metrics_monitoring_test.go). We could extend this for "logs" as well - instead of having two tests doing the same thing. wdyt?

thanks for raising this point @khushijain21 , this makes sense to me and I can't think of any reason of not doing so, @leehinman wdyt?

leehinman:
> > So there is already a test for "metrics" part of monitoring [here](https://github.com/elastic/elastic-agent/blob/main/testing/integration/metrics_monitoring_test.go). We could extend this for "logs" as well - instead of having two tests doing the same thing. wdyt?
> 
> thanks for raising this point @khushijain21 , this makes sense to me and I can't think of any reason of not doing so, @leehinman wdyt?

2 main reasons why I did a separate test.  1) This test will change over time, the next step is to add another fixture in the test that runs agent monitoring as beat receivers, then the test will be modified to compare the docs between beats & beats receivers.  2) The other test enrolls the agent in fleet, the SRE use case is for stand-alone.

Once the SRE use case project is done we should go through and see if we can consolidate or eliminate  redundant tests.  I updated https://github.com/elastic/ingest-dev/issues/4899 to include that task.

khushijain21 (approved):

testing/integration/beat_receivers_test.go
```diff
@@ -0,0 +1,186 @@
...
+		dsType      string
+		dsDataset   string
+		dsNamespace string
+	}
+
+	tests := []test{
+		{dsType: "logs", dsDataset: "elastic_agent", dsNamespace: info.Namespace},
+		{dsType: "metrics", dsDataset: "elastic_agent.elastic_agent", dsNamespace: info.Namespace},
+		{dsType: "metrics", dsDataset: "elastic_agent.filebeat", dsNamespace: info.Namespace},
+		{dsType: "metrics", dsDataset: "elastic_agent.metricbeat", dsNamespace: info.Namespace},
```
khushijain21, line 154: We also have `elastic_agent.filebeat_input` ds.
https://github.com/khushijain21/elastic-agent/blob/main/internal/pkg/agent/application/monitoring/v1_monitor.go#L796

pkoutsovasilis (approved):

mauri870 (approved):

PR #7368, opened 2025-03-12
pkg/component/runtime: limit scope of updateInProgress status flag

efd6:
## What does this PR do?

```
pkg/component/runtime: limit scope of updateInProgress status flag

The updateDoneChan and upDateInProgress status variables were not accessed or
updated outside the scope of (*Manager).runLoop except during initialisation,
and the scope is only in existence once, so make them local to that scope.
```

## Why is it important?

Reduces code and limits range for bug in doing so.

## Checklist

This is unclear to me. Running `go test` in the package at main (currently 456bcdad83e58b34286af4cb137fac7529b29b42) fails, so what a correct test set-up is is unclear.

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/component/runtime/manager.go
+++ b/pkg/component/runtime/manager.go
@@ -115,23 +115,11 @@ type Manager struct {
 	// to the internal run loop.
 	updateChan chan component.Model
 
-	// Component model update is run asynchronously and pings this channel when
-	// finished, so the runtime manager loop knows it's safe to advance to the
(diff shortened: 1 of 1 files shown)


swiatekm (approved):
Looks reasonable to me, but I'd feel more confident if @blakerouse had a look as well.

blakerouse (approved):
Looks like a safe change to me. Not clear why it wasn't done this way initially.

PR #7369, opened 2025-03-12
pkg/component/runtime: use atomic.Pointer rather than locks for current component

efd6:
## What does this PR do?

```
pkg/component/runtime: use atomic.Pointer rather than locks for current component

Locking to obtain write boundaries when obtaining the current component.Component
makes use of a read-write mutex. This protection only protects the access and does
not ensure that the current is not mutated with a logical race. The same semantics
can be obtained more cheaply and without the risk of recursive read-lock deadlock
by using an atomic value. This could either be via atomic.Value or atomic.Pointer;
atomic.Pointer is chosen since it can be type-defined in code, rather than by the
addition of a comment.
```

## Why is it important?

Simpler code that is more closely aligned with the semantics of the apparently intended behaviour. Reduced potential for incorrect use of locks.

## Checklist

Unclear, `go test` in pkg/component/runtime fails at main, so I have no ground truth with which to compare.

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/pkg/component/runtime/runtime.go
+++ b/pkg/component/runtime/runtime.go
@@ -80,10 +80,9 @@ type componentRuntimeState struct {
 	logger  *logger.Logger
 	comm    *runtimeComm
 
-	id         string
-	currCompMx sync.RWMutex
-	currComp   component.Component
-	runtime    componentRuntime
+	id       string
+	currComp atomic.Pointer[component.Component]
+	runtime  componentRuntime
 
 	shuttingDown atomic.Bool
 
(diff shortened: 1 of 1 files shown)


pkg/component/runtime/runtime.go
```diff
@@ -160,15 +159,11 @@ func (s *componentRuntimeState) runLoop() {
 }
 
 func (s *componentRuntimeState) getCurrent() component.Component {
-	s.currCompMx.RLock()
-	defer s.currCompMx.RUnlock()
-	return s.currComp
+	return *s.currComp.Load()
```
pchila, line 162: This will break when the atomic pointer holds a nil value.
I guess we need a nil check here and some zero value to be returned instead

efd6: Please look [up](https://github.com/elastic/elastic-agent/pull/7369/files#diff-25a0cbe41ced249892ab5090a7a2dacc81349647ea4d99361df6c0d5fb0079c0R119). The invariant that is held is that this always holds a non-nil value.

PR #7411, opened 2025-03-13
[9.0] Switch to fork of opentelemetry-collector-contrib and restore beat receivers

cmacknz:
Currently pinned to the head of https://github.com/cmacknz/opentelemetry-collector-contrib/commits/9.0-revert-prometheus-0.300/ to allow us to downgrade prometheus to a version that Beats can support.


internal/pkg/otel/README.md
```diff
@@ -35,32 +35,50 @@ This section provides a summary of components included in the Elastic Distributi
 
 | Component | Version |
 |---|---|
+| [filelogreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/receiver/filelogreceiver/=&gt; github.com/elastic/opentelemetry-collector-contrib/receiver/filelogreceiver v0.0.0-20250317163643-19cd4e80024f/receiver/filelogreceiver/README.md) | =&gt; github.com/elastic/opentelemetry-collector-contrib/receiver/filelogreceiver v0.0.0-20250317163643-19cd4e80024f |
```
cmacknz, line 38: Having this twice seems like an artifact of the replace directive, it's not entirely wrong but it is unnecessary duplication.

internal/pkg/otel/README.md
```diff
@@ -48,6 +48,7 @@ This section provides a summary of components included in the Elastic Distributi
 | [nopreceiver](https://github.com/open-telemetry/opentelemetry-collector/blob/receiver/nopreceiver/v0.119.0/receiver/nopreceiver/README.md) | v0.119.0 |
 | [otlpreceiver](https://github.com/open-telemetry/opentelemetry-collector/blob/receiver/otlpreceiver/v0.120.0/receiver/otlpreceiver/README.md) | v0.120.0 |
 | [prometheusreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/receiver/prometheusreceiver/v0.120.1/receiver/prometheusreceiver/README.md) | v0.120.1 |
+| [prometheusreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/receiver/prometheusreceiver/=&gt; github.com/elastic/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.0.0-20250317163643-19cd4e80024f/receiver/prometheusreceiver/README.md) | =&gt; github.com/elastic/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.0.0-20250317163643-19cd4e80024f |
```
cmacknz, line 51: The replace directive causes us to include both versions of the receiver here which is a bit annoying.

andrzej-stencel: Right, apparently the README generator needs an adjustment.

swiatekm: I fixed it here: https://github.com/elastic/elastic-agent/pull/7443/commits/032519e96a96a2c5ff59bc5e8c8623e89c41b488.

swiatekm (approved):

PR #7413, opened 2025-03-14
catalog-info: permissions for triggering builds does not work

v1v:
## What does this PR do?

Add permissions for triggering a build from another build. Even though it's a merge commit event.

## Why is it important?

This [BK build](https://buildkite.com/elastic/elastic-agent-binary-dra/builds/2711) failed

image

See https://buildkite.com/docs/pipelines/configure/step-types/trigger-step#permissions

The current security model is a little bit complicated now:
- https://github.com/elastic/elastic-agent/blob/a439f4edecaf176d463bf10a9057f6f54a58fd45/.buildkite/pull-requests.json#L35-L50
- https://github.com/elastic/elastic-agent/blob/a439f4edecaf176d463bf10a9057f6f54a58fd45/catalog-info.yaml#L369-L370
- https://github.com/elastic/elastic-agent/blob/a439f4edecaf176d463bf10a9057f6f54a58fd45/catalog-info.yaml#L244-L245

In my view, we should not need this... since it's a merge commit hence it should not take into account the author at all... however, that's me

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -366,8 +366,10 @@ spec:
           access_level: BUILD_AND_READ
         release-eng:
           access_level: BUILD_AND_READ
+        # See https://buildkite.com/docs/pipelines/configure/step-types/trigger-step#permissions
+        # Otherwise even if merge commits in main or any branch will fail
         everyone:
-          access_level: READ_ONLY
+          access_level: BUILD_AND_READ
 
 
 ---


pkoutsovasilis:
> In my view, we should not need this... since it's a merge commit hence it should not take into account the author at all... however, that's me

@v1v trying to decode if we should do this change or not 😄 we should right?

v1v:
> @v1v trying to decode if we should do this change or not 😄 we should right?

That's the only way at the moment, unless we want to get a more sophisticated approach using the BK API and an Service Machine API Token.

However, I plan to ask the BK team if they are aware of anything a bit better for this...

But for now, our CI experts could not find an alternative besides what I mentioned above.

This should be merged

pkoutsovasilis (approved):

PR #7441, opened 2025-03-17
[CI] Get stable ESS version for branches

pazone:
### Pipeline and scripts chages
Gets stable ESS stack version from the stable channels. 
The `ess_start.sh` script is now inline. The `Start ESS stack for integration tests` saves the `stable.ess.version` to buildkite metadata for further usage by retried steps.

### Terraform changes
Adds a `stable_snapshot_version` variable. Now we have variables:

* `stack_version` - represents the deployment version corresponding to [version_regex](https://registry.terraform.io/providers/elastic/ec/latest/docs/resources/deployment#basic) format.
* `stable_snapshot_version` - if it's not empty, we override `integration_server_image`, `elasticsearch_image` and `kibana_image` by images with `stable_snapshot_version`.
 
If `stable_snapshot_version ` is empty - it uses `stack_version` as deployment version and doesn't override the images

The varables :

- `elasticsearch_docker_image`
- `integration_server_docker_image`
- `kibana_docker_image`

have higher priority than `stable_snapshot_version`. 

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


.buildkite/scripts/common-integration.sh
```diff
@@ -51,3 +51,29 @@ google_cloud_auth() {
...
 }
+
+# Prints stack version for current or target release branch without '-SNAPSHOT' suffix 
+# example: 
+# BUILDKITE_PULL_REQUEST_BASE_BRANCH=8.x .buildkite/scripts/test.sh
+# 8.19.0-64846b77
+getStableEssSnapshotForBranch() {
+  local baseStackBranch
+  # If Pull request
+  if [ -n "${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-}" ]; then
```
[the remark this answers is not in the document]
pazone: Thanks for the suggestion. The big problem is that I must implement the same functionality in PowerShell. Ideally, we need something cross-platform. I thought about mage, but its output is not stable

.buildkite/scripts/steps/ess_down.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-source .buildkite/scripts/common2.sh
+source .buildkite/scripts/common-integration.sh
```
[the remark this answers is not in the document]
pazone: Because we have `common.sh` and `common2.sh`. The `common2.sh` is used for integration tests. The renaming makes a lot of sense

.buildkite/scripts/common-integration.sh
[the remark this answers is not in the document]
v1v: I understand your concern is about `stdout` and mage. Do you think it could be used an `.override-package-version` file that it's created on the file using Mage. And hence it's always read it when running the ITs?

In other words, `OVERRIDE_STACK_VERSION` will point to the content of `.override-package-version` and there will be a mage call initially to create the `.override-package-version` on the fly using the same logic as done in bash but in a Magefile. 

By doing so, ITs for *nix or windows will always run the mage command for creating the `.override-package-version` file and then set the env variable `OVERRIDE_STACK_VERSION` to the content of that file.

WDYT?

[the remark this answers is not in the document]
pazone: Yeah, that's right. We can create a temporary `stack_version` file, and the `OVERRIDE_STACK_VERSION` will always get its value. 

It can be done by bash only. The `Start ESS` step prerequisites the integration tests with a single shared stack. We can execute `getStableEssSnapshotForBranch` within this step and place the stack version to buildkite metadata.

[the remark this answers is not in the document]
v1v: In my opinion, less obfuscation is preferable. Using a magefile function to provide details on the fly seems easier than relying on the BK idiosyncrasies (hard to reproduce it locally and debugging)


For instance, I created a quick function in mage:

```go


// GetStableEssSnapshotForBranch retrieves the stable ESS snapshot for a given branch.
func GetStableEssSnapshotForBranch() error {
    baseStackBranch := os.Getenv("BUILDKITE_PULL_REQUEST_BASE_BRANCH")
    if baseStackBranch == "" {
        baseStackBranch = os.Getenv("BUILDKITE_BRANCH")
    }

    if baseStackBranch == "" {
        content, err := os.ReadFile(".package-version")
        if err != nil {
            return fmt.Errorf("failed to read .package-version: %v", err)
        }
        fmt.Print(string(content))
        return nil
    }

    stableChannelURL := fmt.Sprintf("https://storage.googleapis.com/artifacts-api/channels/%s.json", baseStackBranch)

	resp, err := http.Get(stableChannelURL)
	if err != nil || resp.StatusCode != http.StatusOK {
		content, err := os.ReadFile(".package-version")
		if err != nil {
			return fmt.Errorf("failed to read .package-version: %v", err)
		}
		fmt.Print(string(content))
		return nil
	}
	defer resp.Body.Close()

    body, err := io.ReadAll(resp.Body)
	if err != nil {
		return fmt.Errorf("failed to read response body: %v", err)
	}

	var result struct {
		Build string `json:"build"`
	}
	if err := json.Unmarshal(body, &result); err != nil {
		return fmt.Errorf("failed to parse JSON: %v", err)
	}

	fmt.Println(strings.TrimSpace(result.Build))
	return nil
}
```


image

I have not done much to tweak it for creating a file instead of printing in the stdout or adding UTs to help with the maintainability.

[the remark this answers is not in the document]
pazone: Created a mage target that writes the obtained version to the `.override_stack_version` file. Need to test it with the old stack provisioner

.buildkite/scripts/steps/ess.sh
```diff
@@ -13,6 +12,8 @@ function ess_up() {
     return 1
   fi
 
+  echo "~~~ Staring ESS Stack ${STACK_VERSION}"
```
v1v, line 15:
```suggestion
  echo "~~~ Starting ESS Stack ${STACK_VERSION}"  
```

pkg/testing/ess/create_deployment_csp_configuration.yaml
```diff
@@ -14,6 +14,6 @@ aws:
   elasticsearch_deployment_template_id: "aws-storage-optimized"
   kibana_instance_configuration_id: "aws.kibana.c5d"
 docker:
-  integration_server_image: "docker.elastic.co/cloud-release/elastic-agent-cloud:9.1.0-6afab230-SNAPSHOT"
-  elasticsearch_image: "docker.elastic.co/cloud-release/elasticsearch-cloud-ess:9.1.0-6afab230-SNAPSHOT"
-  kibana_image: "docker.elastic.co/cloud-release/kibana-cloud:9.1.0-6afab230-SNAPSHOT"
\ No newline at end of file
+  integration_server_image: "docker.elastic.co/cloud-release/elastic-agent-cloud:${OVERRIDE_STACK_VERSION}"
+  elasticsearch_image: "docker.elastic.co/cloud-release/elasticsearch-cloud-ess:${OVERRIDE_STACK_VERSION}"
+  kibana_image: "docker.elastic.co/cloud-release/kibana-cloud:${OVERRIDE_STACK_VERSION}"
```
v1v, line 19: I'm not sure about the changes, IIUC, https://github.com/elastic/elastic-agent/pull/7098/files#diff-38acbfb3e60b4cc22598e92b85dda23797dec02f252687606910b58687a89d60R65-R79 is the one actually removing the pinned version.

What's the reason `stack_version` is not override with the value of `OVERRIDE_STACK_VERSION`?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -13,7 +13,28 @@ steps:
     command: |
       #!/usr/bin/env bash
       set -euo pipefail
-      source .buildkite/scripts/steps/ess_start.sh
+      source .buildkite/scripts/steps/stable_ess_version.sh
```
v1v, line 16: What's the reason for using all those inline commands instead of using the original script called `.buildkite/scripts/steps/ess_start.sh`?

I found a good pattern to use pipelines as orchestrators without any business logic

.buildkite/bk.integration.pipeline.yml
```diff
@@ -335,15 +356,15 @@ steps:
...
+              - "basic,slim,complete,service,elastic-otel-collector"
+              - "wolfi,slim-wolfi,complete-wolfi,elastic-otel-collector-wolfi"
             version:
-            - v1.27.16
-            - v1.28.9
-            - v1.29.8
-            - v1.30.8
-            - v1.31.0
-            - v1.32.0
+              - v1.27.16
```
v1v, line 362: let's avoid these kind of changes, they are unrelated to the nature of the PR.

v1v (commented):
Some early feedback 👼

.buildkite/bk.integration.pipeline.yml
pazone: The idea is to keep our shell scripts CI-agnostic. That means the scripts should suit CI and development/testing purposes. The `ess_start.sh` contained mostly `buildkite-agent ...` commands that connect it with buildkite.  The `ess.sh` is sufficient for CI and development purposes.

v1v: it can also be agnostic by using:

```BASH
# only if running on the CI
if [ "${CI:=false}" == 'true' ]; then
  buildkite-agent ....
fi
```

Therefore, the same script can be used locally and in the CI, and also behave as the CI by running `CI=true .buildkite/scripts/steps/ess_start.sh`

If we do so, we can keep the BK pipeline as an orchestrator without business logic for simplicity.

pazone: That's an interesting idea. I'm not quite sure about if-conditions, but I can review the `ess_start.sh` and leave there only ess logic. The pipeline script will be like:
```
source ...ess_start.sh
buildkite-agent meta-data set "es.host" $ELASTICSEARCH_HOST
buildkite-agent meta-data set "es.username" $ELASTICSEARCH_USERNAME
...
buildkite-agent meta-data set "stable.ess.version" $STABLE_ESS_VERSION
```
WDYT?

.buildkite/bk.integration.pipeline.yml
pazone: Fixed

.buildkite/bk.integration.pipeline.yml
v1v: I prefer to keep things simple in the BK pipelines; I have learned that it pays off in the long run. Additionally, we are discouraged from using the `buildkite-agent meta-data set` commands for several reasons.

magefile.go
```diff
@@ -2844,6 +2844,70 @@ func (Integration) TestOnRemote(ctx context.Context) error {
 	return nil
 }
 
+// Writes stable ESS version to .override_stack_version file
+func (Integration) GetStableEssSnapshotForBranch() error {
```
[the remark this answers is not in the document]
pazone: I just removed it

.buildkite/bk.integration.pipeline.yml
pazone: Applied the proposed condition. It indeed looks more simple

pazone:
We have to merge this PR first: https://github.com/elastic/elastic-agent/pull/7919

v1v:
@pazone, this PR is in draft, shall it be changed to `ready for review`?

pazone:
@v1v I'm waiting for tests to pass

.buildkite/scripts/steps/ess_start.sh
```diff
@@ -1,21 +1,25 @@
...
 
-OVERRIDE_STACK_VERSION="$(cat .package-version)"
-OVERRIDE_STACK_VERSION=${OVERRIDE_STACK_VERSION}"-SNAPSHOT"
-
-ess_up $OVERRIDE_STACK_VERSION
+echo "~~~ Getting stable stack version"
+DEFAULT_STACK_VERSION="$(cat .package-version)-SNAPSHOT"
+STABLE_ESS_VERSION="$(getStableEssSnapshotForBranch)-SNAPSHOT"
+ess_up $DEFAULT_STACK_VERSION $STABLE_ESS_VERSION
+echo "ES_HOST: ${ELASTICSEARCH_HOST}"
```
v1v, line 12: Is this the place to print `ELASTICSEARCH_HOST`?

.buildkite/scripts/steps/stable_ess_version.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
v1v, line 1: Can you change the shebang to be:

```suggestion
#!/usr/bin/env bash
```

**why**? Because `/usr/bin/env` interprets the $PATH, which makes scripts more portable. And it's the behaviour in some other sh scripts

.buildkite/scripts/steps/ess_start.sh
```diff
@@ -1,21 +1,25 @@
...
 source .buildkite/scripts/steps/ess.sh
 source .buildkite/scripts/steps/fleet.sh
 
-OVERRIDE_STACK_VERSION="$(cat .package-version)"
-OVERRIDE_STACK_VERSION=${OVERRIDE_STACK_VERSION}"-SNAPSHOT"
-
-ess_up $OVERRIDE_STACK_VERSION
+echo "~~~ Getting stable stack version"
+DEFAULT_STACK_VERSION="$(cat .package-version)-SNAPSHOT"
+STABLE_ESS_VERSION="$(getStableEssSnapshotForBranch)-SNAPSHOT"
```
v1v, line 10: IIUC, `getStableEssSnapshotForBranch` adds the `-SNAPSHOT` suffix, according the to the docs in the function.

.buildkite/scripts/steps/integration_tests_tf.sh
```diff
@@ -35,8 +29,12 @@ mage build:testBinaries
 # BUILDKITE_RETRY_COUNT > 0 for the retries
 if [[ "${BUILDKITE_RETRY_COUNT}" -gt 0 ]]; then
   echo "~~~ The steps is retried, starting the ESS stack again"
+  echo "~~~ Getting stable stack version"
+  DEFAULT_STACK_VERSION="$(cat .package-version)-SNAPSHOT"
+  STABLE_ESS_VERSION="$(getStableEssSnapshotForBranch)-SNAPSHOT"
```
v1v, line 34: ditto

.buildkite/scripts/steps/integration_tests_tf.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-source .buildkite/scripts/common2.sh
-
+source .buildkite/scripts/common-integration.sh
 source .buildkite/scripts/steps/ess.sh
+source .buildkite/scripts/steps/stable_ess_version.sh
```
v1v, line 6: What's the reason for creating a step called `stable_ess_version` instead of adding the function to the `common-integration.sh` script itself?

.buildkite/scripts/integration-tests.ps1
```diff
@@ -44,7 +40,12 @@ $outputXML = "build/${fully_qualified_group_name}.integration.xml"
 $outputJSON = "build/${fully_qualified_group_name}.integration.out.json"
 $TestsExitCode = 0
 try {
-    Get-Ess-Stack -StackVersion $PACKAGE_VERSION
+    echo "~~~ Getting stable stack version"
+    mage integration:getStableEssSnapshotForBranch
```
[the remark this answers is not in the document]
pazone: That's a good catch!

.buildkite/scripts/steps/integration_tests_tf.sh
pazone: That's also an option

.buildkite/scripts/integration-tests.ps1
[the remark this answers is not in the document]
v1v: BTW, it did not fail:
- https://buildkite.com/elastic/elastic-agent/builds/20383#01966320-5e32-4d70-9483-88a026ba22e3/1745420461238/Getting%20stable%20stack%20version

image


Is that the expected behaviour? Or should it fail regardless?

.buildkite/scripts/steps/ess_start.sh
pazone: It used several times for debugging. I think I can delete it

.buildkite/scripts/steps/ess_start.sh
pazone: That's right. Updated the doc for the function

.buildkite/scripts/steps/integration_tests_tf.sh
pazone: That's right. Updated the doc for the function

.buildkite/scripts/steps/ess.ps1
```diff
@@ -14,6 +15,8 @@ function ess_up {
       return 1
   }
 
+  Write-Output "~~~ Starting ESS Stack $StackVersion"
```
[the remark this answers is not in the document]
pazone: removed

.buildkite/scripts/integration-tests.ps1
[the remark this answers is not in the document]
pazone: It is expected, but now I think it's not very good. It shouldn't throttle stack-related errors. Will move it out of try block and add a check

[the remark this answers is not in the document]
pazone: Now it looks better to me

.buildkite/scripts/steps/integration_tests_tf.sh
pazone: Moved to common-integration.sh

.buildkite/scripts/steps/stable_ess_version.sh
pazone: done

v1v (dismissed):
LGTM. Waiting for validating if the new changes run all the ITs without any hidden errors

.buildkite/scripts/steps/ess.ps1
```diff
@@ -1,10 +1,11 @@
 function ess_up {
   param (
       [string]$StackVersion,
+      [string]$StableSnapshotVersion,
       [string]$EssRegion = "gcp-us-west2"
   )
 
-  Write-Output "~~~ Starting ESS Stack"
+  Write-Output "~~~ Starting ESS. Version: $StackVersion. Stable snapshot: $StableSnapshotVersion"
```
v1v, line 8: I'm not much familiar with the ITs, but I cannot see this trace in one of the Windows ITs:
- https://buildkite.com/elastic/elastic-agent/builds/20447#01966817-7b5b-4f0a-8cb4-56fe4bda1983/1745503548222/Running%20global%20environment%20hook


image

.buildkite/scripts/integration-tests.ps1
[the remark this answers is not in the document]
v1v: I can see the `misisng meta-data key`
image

https://buildkite.com/elastic/elastic-agent/builds/20447#01966817-7b5b-4f0a-8cb4-56fe4bda1983

It was not clear to me whether that's the expected behaviour

pazone:
Tested retries on:
- [Ubuntu](https://buildkite.com/elastic/elastic-agent/builds/20543/summary/annotations?jid=01966d7a-aadd-4418-8f20-cac1aecb669e)
-  [windows](https://buildkite.com/elastic/elastic-agent/builds/20543/summary/annotations?jid=01966d7c-b5a0-4819-912f-b59d97b75289): The stack is successfully recreated, tests are skipped.

pazone:
@v1v @pchila @pkoutsovasilis the experts are welcome for review

pazone:
Moving this PR to draft. Upcoming changes:
* Return back the ability to override the docker images for the ESS deployment 
* Have a static `.override-stack-version` file with the latest known stable snapshot
* Create a updateCLI-managed PR creation that updates the `.override-stack-version` file and runs the tests

PR #7443, opened 2025-03-17
[8.18] Update OTel components to v0.120.x

swiatekm:
## What does this PR do?

Updates Otel components to 0.120.x. It uses a forked version of prometheus receiver with the prometheus dependency bump reverted, just like https://github.com/elastic/elastic-agent/pull/7411.

It also updates the mage target for generating the otel readme to handle replace directives and pseudo versions correctly.

## Why is it important?

We want these changes in 8.18.0, as they contain important fixes to elasticsearch exporter, but we don't want to remove beats receivers to resolve the dependency conflict between Prometheus packages.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)

## Related issues

- https://github.com/elastic/elastic-agent/pull/7411
- https://github.com/elastic/elastic-agent/pull/6912

PR #7456, opened 2025-03-18
Support replace directives in otel readme mage command

swiatekm:
## What does this PR do?

Makes the `otel:readme` mage command support replace directives. It now also supports pseudo-versions that Go uses for commit hashes in go.mod.

The implementation now actually parses the go.mod file using x/mod.

## Why is it important?

In order to resolve a dependency conflict between beats and otel 0.120.x, we're going to use our own fork of prometheusreceiver. See https://github.com/elastic/elastic-agent/pull/7443 for this change made in the 8.18 branch.

## Checklist

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -33,6 +33,9 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/mod/modfile"
+	"golang.org/x/mod/module"
+
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/otiai10/copy"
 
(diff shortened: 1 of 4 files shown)


magefile.go
```diff
@@ -3324,6 +3329,14 @@ func getOtelDependencies() (*otelDependencies, error) {
 		sort.Slice(list, func(i, j int) bool { return list[i].Name < list[j].Name })
 	}
 
+	// take care of replaces
+	for _, rep := range modFile.Replace {
```
cmacknz, line 3333: This looks correct, but since there are no replaces in the go.mod on main there's no proof it works. This feels like it shouldn't be too horrible to unit test, with a reference go.mod file. Are you able to add a simple test for at least the replace case?

Probably this entire set of functions should move into a sub-package of dev-tools to enable doing this.

andrzej-stencel (approved):

magefile.go
cmacknz: Thank you! LGTM

PR #7483, opened 2025-03-20
Update golangci-lint.yml

ismisepaul:
Pinning action to a full length commit SHA [see](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)


v1v (approved):
Thanks Paul! 

After joining forces with ingest, we actually need to revisit where we are at with the pinned actions. Thanks again

PR #7486, opened 2025-03-20
Skip FIPS components when packaging elastic-agent from a Manifest

pchila:
## What does this PR do?

ChecksumsWithManifest function assumes that all (actually only one) packages matching a component binary name prefix and the target platform are downloaded and available but since -fips packages are available in manifests this is no longer true (see endpoint-security-fips-9.1.0-SNAPSHOT-linux-x86_64.tar.gz and endpoint-security-9.1.0-SNAPSHOT-linux-x86_64.tar.gz which are both present in recent manifest versions).

This is a short term hack to avoid breaking elastic-agent packaging now that -fips package variants are available by skipping them.

## Why is it important?

To be able to generate elastic-agent packages again

## Checklist

- ~~[ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.~~

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/checksums.go
+++ b/dev-tools/mage/checksums.go
@@ -110,7 +110,11 @@ func ChecksumsWithManifest(requiredPackage string, versionedFlatPath string, ver
 				// Iterate over the external binaries that we care about for packaging agent
 				for _, spec := range manifest.ExpectedBinaries {
 					// If the individual package doesn't match the expected prefix, then continue
-					if !strings.HasPrefix(pkgName, spec.BinaryName) {
+					// FIXME temporarily skip fips packages until elastic-agent FIPS is in place
+					if !strings.HasPrefix(pkgName, spec.BinaryName) || strings.Contains(pkgName, "-fips-") {
+						if mg.Verbose() {
+							log.Printf(">>>>>>> Package [%s] skipped", pkgName)
+						}
 						continue
 					}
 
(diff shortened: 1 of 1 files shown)


dwhyrock (approved):

pchila:
> What is the underlying issue?
> 
> As of yesterday, Endpoint is publishing the FIPS artifacts with Release Manager: https://artifacts-snapshot.elastic.co/endpoint-dev/9.1.0-6ed2da2c/summary-9.1.0-SNAPSHOT.html
> 
> Is the Agent Packaging step reading the manifest but not downloading the `-fips` endpoint artifact?

The underlying issue is in these checks
https://github.com/elastic/elastic-agent/blob/931134f3fca32305d002d1042bfc296cc5f9c36b/dev-tools/mage/checksums.go#L113

https://github.com/elastic/elastic-agent/blob/931134f3fca32305d002d1042bfc296cc5f9c36b/dev-tools/mage/checksums.go#L200

that will match ` endpoint-security-fips-9.1.0-SNAPSHOT-linux-x86_64.tar.gz` (even though it's **not** been downloaded from the manifest) and try to copy files from a non-existing directory. Since the copy obviously fails, the mage target panics and aborts the packaging.

The issue here is that in those 2 location the manifest is scanned (twice) and it makes the assumption that anything having the right prefix has been downloaded, even when it hasn't

pkoutsovasilis (commented):
code changes LGTM waiting for the CI to get green, although `Extended runtime leak tests` continues to be in a bad mood

ebeahan (approved):
LGTM.

Will force merge due to an unrelated, ongoing CI issue with the extended runtime leak testing.

simitt:
@pchila could you create a follow up card to remove this temporary solution again once possible.

pchila:
> @pchila could you create a follow up card to remove this temporary solution again once possible.

Created https://github.com/elastic/elastic-agent/issues/7610 to track the removal of this hack

Issue #7506, opened 2025-03-21
[bug] Ensure generated User passwords comply with Windows password policy


pkoutsovasilis:
We are seeing failures in CI, for example in this [build](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/2084#01917ae9-918f-414b-b9d5-a80b523129b4/102-821):

```
| [==  ] Failed to install service  [3s] Error uninstalling. Printing logs
| Error: error installing package: error getting service installation options: failed to set user elastic-agent-user password for service: call to NetUserSetInfo failed: status=2245 error=4294967295
| For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
```

The root cause translates to:
```
The password is shorter than required. (The password could also be too long, be too recent in its change history, not have enough unique characters, or not meet another password policy requirement.)
```

We need to update the password generation logic to ensure it consistently meets Windows password policy constraints.

PR #7507, opened 2025-03-21
[bug] rework Windows user password generation to match security constraints

pkoutsovasilis:
## What does this PR do?

This PR improves the password generation logic used when installing the Elastic Agent on Windows by reducing the likelihood of generating invalid passwords that violate Windows security policies.

The new implementation introduces:
- Passwords are now generated with a random length between 32 and 64 characters to avoid fixed-length assumptions.
- At least one character from each required category (uppercase, lowercase, digit, special) is guaranteed to be included.
- Consecutive duplicate characters are avoided, further improving compliance with stricter policy checks.

## Why is it important?

This PR can't completely eliminate the possibility of a password policy mismatch, since policies can vary per machine/domain and user configuration. However, it greatly reduces the chance of generating an invalid password by aligning more closely with the Windows password requirements.

## Checklist

## Disruptive User Impact

None expected. The password generation change is internal to the installation process on Windows systems and should only improve robustness.

- Run `mage unitTest` to validate that passwords generated over many iterations conform to the policy constraints.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/7506

## Questions to ask yourself

- ...


pkoutsovasilis:
gonna request a review also from @blakerouse as he is familiar with Windows intricacies 🙂

internal/pkg/agent/install/user_windows.go
```diff
@@ -20,7 +19,16 @@ import (
 )
 
 const (
-	passwordLength = 127 // maximum length allowed by Windows
+	// Reference: https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements
+	passwordMinLength = 32 // Minimum length for better security
+	passwordMaxLength = 64 // Upper limit to avoid policy issues
```
leehinman, line 24: Do you have a reference for what issues could happen with too long of a password?  That would be a nice breadcrumb for our future selves if we have to change this.

internal/pkg/agent/install/user_windows.go
```diff
@@ -238,19 +246,81 @@ func SetUserPassword(name string, password string) error {
...
-		n, err := rand.Int(rand.Reader, maxN)
+	// Generate a random length within the allowed range
+	length, err := rand.Int(rand.Reader, big.NewInt(passwordMaxLength-passwordMinLength+1))
+	if err != nil {
+		return "", fmt.Errorf("failed to generate random length: %w", err)
+	}
+	passwordLength := int(length.Int64()) + passwordMinLength
+	password := make([]rune, passwordLength)
+
+	// Ensure Windows password constraints are met by guaranteeing at least one character from each category
```
leehinman, line 259: On the one hand I like that this is guaranteed to produce a candidate
password that has one of each class, without adjacent duplicates.  On
the other, when we inevitably need to expand the conditions, it is
going to require a bit of thought.

What if we switched to a for loop that generated a random password and
then we tested that password for our requirements?  We can break when
we find one that meets all of them.

We know from experience that the random password usually meets the
requirements, so we would only go through the loop a few times.  The
checks for the conditions would be easy to write and would be easy to
write both positive and negative unit tests.  And when another check
needs to be added (like can't be similar to username, or 2 upper case,
must contain emoji, etc).  It just becomes another check that we write
and it will be independent of the existing ones.

What do you think?

pkoutsovasilis: Thanks for the feedback, Lee 🙂 

I get where you’re coming from but if we were to follow the check-retry paradigm, we could just rely on an error from NetUserSetInfo and retry there, rather than implementing our own local validation checks. This would achieve the same goal with less code complexity.

The main reason I prefer the direct password fabrication approach is that, mathematically, a check function is essentially the reverse of the password generation function. Given this equivalence, it makes more sense to optimise the generation itself to maximise the probability of producing a valid password upfront rather than generating arbitrarily and filtering afterward, which maintains the same probability.

To fully understand, do you find the current approach to be overly complicated? If so, I am more than happy to discuss specific parts that could be simplified further. Let me know what you think

internal/pkg/agent/install/user_windows.go
blakerouse: Why the 64 cap? Windows GUI allows up to 127 character length and AD allows 256 characters.

internal/pkg/agent/install/user_windows.go
blakerouse: I am actually okay with the implementation and think its relatively easy to follow. It is commented very well.

internal/pkg/agent/install/user_windows_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+		for _, char := range password {
+			switch {
+			case strings.ContainsRune(passwordCharsLower, char):
+				hasLower = true
+			case strings.ContainsRune(passwordCharsUpper, char):
+				hasUpper = true
+			case strings.ContainsRune(passwordCharsDigits, char):
+				hasDigit = true
+			case strings.ContainsRune(passwordCharsSpecial, char):
```
leehinman, line 42: Can you add a check that the password doesn't contain the zero value of a rune?  If for some reason we got the record keeping wrong in the `filledPositions` map we could end up with a zero value in the password.  The length check won't catch this.  This might result in the zero value for the rune being turned into a NULL character in the Win32 call, but even if it doesn't we should still catch it.

internal/pkg/agent/install/user_windows_test.go
```diff
@@ -0,0 +1,65 @@
...
+				hasSpecial = true
+			default:
+				t.Fatalf("password %q contains an invalid character %q", password, string(char))
+			}
+		}
+
+		if !hasLower || !hasUpper || !hasDigit || !hasSpecial {
+			t.Fatalf("password %q does not contain all required character categories", password)
```
leehinman, line 50: Can you add the values for `hasLower`, `hasUpper`, `hasDigit`, and `hasSpecial` to the error.  If it ever does fail it will make it easier on the person looking at the CI failure.

leehinman (changes requested):

internal/pkg/agent/install/user_windows.go
leehinman: These were "comments" not "blockers", so no big deal if you want to keep what is in the PR currently.

>  if we were to follow the check-retry paradigm, we could just rely on an error from NetUserSetInfo and retry there, rather than implementing our own local validation checks

That made me think of the fact that we don't really know what the password policy is.  That is something that can be changed by the local admin.  So it is possible that even with the checks you have, we could fail setting the password.  What is our fallback if that happens?

cmacknz: If the user has changed the default, at that point we could let the user generate the password. Presumably they have a way to do that if they tweaked the rules.

Otherwise we would have to let users customize the password generation rules, not sure we want to get into doing that.

pkoutsovasilis: According to the [Microsoft documentation on password complexity requirements](https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements), Windows enforces a fixed set of checks when the policy is enabled—specifically: inclusion of upper/lowercase letters, digits, and special characters, etc.. The logic in this PR ensures that we always meet those predefined criteria.

If a system administrator wants to enforce additional custom password rules (e.g., disallowing certain substrings, requiring emoji, etc.), they need to supply a custom `Passfilt.dll` to implement those checks. In those cases, I had assumed the recommended approach would be for them to use the `--password` CLI argument to provide a password of their own that meets the organization’s custom rules.

Given that, I’m not sure this PR needs to implement any additional fallbacks. But maybe I’m missing a scenario—happy to hear thoughts if there’s a corner case we should be handling more explicitly!

blakerouse (approved):
Thanks for updating the password max length and thanks for working on this change!

pkoutsovasilis:
@leehinman since you requested changes would you mind having another look with the latest commits here? 🙂

leehinman (approved):

PR #7521, opened 2025-03-21
Followup to #6638: Added more integration tests and logging

### LINKED ISSUE #7478: Follow up tweaks to RPM and DEB flavor support
This is a follow up to https://github.com/elastic/elastic-agent/pull/6638 now that the feature has landed.

1. The DEB upgrade integration test needs to be updated to confirm that the installation flavor is preserved regardless of the ELASTIC_AGENT_FLAVOR environment variables value. https://github.com/elastic/elastic-agent/blob/5d4b00ecc2e721733b059dc112a1ff2185d4431f/testing/integration/linux_deb_test.go#L251
2. The RPM upgrade integration test needs to be updated to confirm that the installation flavor is preserved regardless of the ELASTIC_AGENT_FLAVOR environment variables value. https://github.com/elastic/elastic-agent/blob/5d4b00ecc2e721733b059dc112a1ff2185d4431f/testing/integration/linux_rpm_test.go#L251
3. The block that determines the flavor at installation time needs to log which flavor it is going to install. The log message must show up in the output of the package installation command https://github.com/elastic/elastic-agent/blob/5d4b00ecc2e721733b059dc112a1ff2185d4431f/dev-tools/packaging/templates/linux/preinstall.sh.tmpl#L51-L60

michalpristas:
Followup for #6638 Added more integration tests and logging

Integration tests are testing :
- Pre 9.0 to current upgrade: flavor is not applied
- 9.0 to current: flavor is respected

Closes #7478


michalpristas:
version pinning making this fail. need to figureout workaround fo now

testing/integration/linux_deb_test.go
```diff
@@ -187,12 +188,41 @@ func TestDebFleetUpgrade(t *testing.T) {
...
+	for _, tc := range testCases {
+		if !tc.upgradeFromVersion.Less(*currentVersion) {
+			// allow only upgrades to higher versions
+			continue
+		}
+
+		t.Run(fmt.Sprintf("Upgrade DEB from %s - %q", tc.upgradeFromVersion.String(), tc.name), func(t *testing.T) {
+			t.Cleanup(func() {
+				// cleanup after ourselves
+				_ = os.Remove("/var/lib/elastic-agent/.flavor")
```
blakerouse, line 214: Does `apt-get remove --purge` not remove this file?

cmacknz: `--purge` does but a plain` --remove` doesn't, are we using remove instead?

testing/integration/linux_deb_test.go
```diff
@@ -187,12 +188,41 @@ func TestDebFleetUpgrade(t *testing.T) {
...
 
+	testCases := []struct {
+		name               string
+		upgradeFromVersion *version.ParsedSemVer
+		installingServers  bool
+		expectingServers   bool
+	}{
+		{"legacy installation", version.NewParsedSemVer(8, 17, 3, "", ""), false, true},      // in case of legacy we don't apply flavor, expecting all to be preserved
+		{"9.0 snapshot with basic flavor", upgradetest.Version_9_0_0_SNAPSHOT, false, false}, // TODO: replace with PreviousMinor once 9.1 is released
+		{"9.0 snapshot with basic flavor", upgradetest.Version_9_0_0_SNAPSHOT, true, true},   // TODO: replace with PreviousMinor once 9.1 is released
```
cmacknz, line 199: The test name here is the same as the one above.

testing/integration/linux_rpm_test.go
```diff
@@ -187,12 +188,41 @@ func TestRpmFleetUpgrade(t *testing.T) {
...
 
+	testCases := []struct {
+		name               string
+		upgradeFromVersion *version.ParsedSemVer
+		installingServers  bool
+		expectingServers   bool
+	}{
+		{"legacy installation", version.NewParsedSemVer(8, 17, 3, "", ""), false, true},      // in case of legacy we don't apply flavor, expecting all to be preserved
+		{"9.0 snapshot with basic flavor", upgradetest.Version_9_0_0_SNAPSHOT, false, false}, // TODO: replace with PreviousMinor once 9.1 is released
+		{"9.0 snapshot with basic flavor", upgradetest.Version_9_0_0_SNAPSHOT, true, true},   // TODO: replace with PreviousMinor once 9.1 is released
```
cmacknz, line 199: Same as deb, test name for servers is copy/pasted from basic.

testing/integration/linux_deb_test.go
michalpristas: we are using remove to keep state on upgrade right

michalpristas: i added this mostly because when i ran tests on a same machine over again and it failed for some reason sometimes it left flavor but agent was not installed (because of failure) and next run was affected.

cmacknz: At the end of a test we should be able to use `--purge` to get rid of everything shouldn't we?

leehinman: > At the end of a test we should be able to use `--purge` to get rid of everything shouldn't we?

If `--purge` isn't removing everything than that is a bug and we need to fix it.

```
 -P, --purge package...|-a|--pending
           Purge an installed or already removed package. This removes
	   everything, including conffiles, and anything else cleaned
	   up from postrm. If -a or --pending is given instead of a
	   package name, then all packages unpacked or removed, but
	   marked to be purged in file /var/lib/dpkg/status, are purged.

           Note: Some configuration files might be unknown to dpkg
           because they are created and handled separately through
	   the configuration scripts. In that case, dpkg won't remove
	   them by itself, but the package's postrm script (which is
	   called by dpkg), has to take care of their removal during
	   purge. Of course, this only applies to files in system
	   directories, not configuration files written to individual
	   users' home directories.
```

cmacknz: It does, I've tried it, most likely we just are using --purge.

michalpristas: does it remove flavor as well? because we don't specify this in any config so dpkg should not have a notion of it. if it does, it's probably side effect of a dir cleanup

cmacknz: Leaving this code in would allow us to break --purge without noticing, I would rather see this happen conditionally if `--purge` fails or package install fails.

blakerouse: Yeah I agree, I would rather not see this code in the test. Feels like its covering up something that could be broken.

testing/integration/linux_rpm_test.go
```diff
@@ -187,12 +180,37 @@ func TestRpmFleetUpgrade(t *testing.T) {
...
+	currentVersion, err := version.ParseVersion(define.Version())
+	require.NoError(t, err)
+
+	for _, tc := range testCases {
+		if !tc.upgradeFromVersion.Less(*currentVersion) {
+			// allow only upgrades to higher versions
+			continue
+		}
+
+		t.Run(fmt.Sprintf("Upgrade DEB from %s - %q", tc.upgradeFromVersion.String(), tc.name), func(t *testing.T) {
```
pchila, line 203: shouldn't this be rpm ?
```suggestion
		t.Run(fmt.Sprintf("Upgrade RPM from %s - %q", tc.upgradeFromVersion.String(), tc.name), func(t *testing.T) {
```

pkg/testing/fixture.go
```diff
@@ -1241,19 +1246,28 @@ func getCacheDir(caller string, name string) (string, error) {
 }
 
 // FindComponentsDir identifies the directory that holds the components.
-func FindComponentsDir(dir string) (string, error) {
+func FindComponentsDir(dir, version string) (string, error) {
```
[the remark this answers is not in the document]
michalpristas: version is needed just for upgrade scenarios so this should be fine right

leehinman (approved):

testing/integration/linux_deb_test.go
```diff
@@ -187,12 +180,37 @@ func TestDebFleetUpgrade(t *testing.T) {
...
 	})
 
+	testCases := []struct {
+		name               string
+		upgradeFromVersion *version.ParsedSemVer
+		installingServers  bool
+		expectingServers   bool
+	}{
+		{"legacy installation", version.NewParsedSemVer(8, 17, 3, "", ""), false, true},      // in case of legacy we don't apply flavor, expecting all to be preserved
+		{"9.0 snapshot with basic flavor", upgradetest.Version_9_0_0_SNAPSHOT, false, false}, // TODO: replace with PreviousMinor once 9.1 is released
```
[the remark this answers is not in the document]
michalpristas: issue here: https://github.com/elastic/elastic-agent/issues/7636

cmacknz (approved):
Thanks for following up with this!

PR #7548, opened 2025-03-25
feat(fips): disable otel components pulling in x/crypto deps

kruskall:
## What does this PR do?

some otel components are pulling several x/crypto packages under the hood

disable them in fips mode

## Why is it important?

fips

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/otel/components.go
+++ b/internal/pkg/otel/components.go
@@ -93,7 +93,9 @@ func components(extensionFactories ...extension.Factory) func() (otelcol.Factori
 			jmxreceiver.NewFactory(),
 			nopreceiver.NewFactory(),
 		}
-		receivers = addExtraReceivers(receivers)
+		// some receivers should only be available when
+		// not in fips mode due to restrictions on crypto usage
+		receivers = addNonFipsReceivers(receivers)
 		factories.Receivers, err = receiver.MakeFactoryMap(receivers...)
 		if err != nil {
 			return otelcol.Factories{}, err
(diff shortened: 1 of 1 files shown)


internal/pkg/otel/components.go
```diff
@@ -95,9 +91,10 @@ func components(extensionFactories ...extension.Factory) func() (otelcol.Factori
 			fbreceiver.NewFactory(),
 			mbreceiver.NewFactory(),
 			jmxreceiver.NewFactory(),
-			kafkareceiver.NewFactory(),
 			nopreceiver.NewFactory(),
-		)
+		}
+		receivers = addExtraReceivers(receivers)
```
swiatekm, line 96: Can we use a more descriptive name for this, like `addNonFipsReceivers`? The function itself should also have a docstring explaining what it does, and ideally there should be a comment here explaining why this is necessary.

internal/pkg/otel/components.go
```diff
@@ -122,15 +121,18 @@ func components(extensionFactories ...extension.Factory) func() (otelcol.Factori
...
 			elasticsearchexporter.NewFactory(),
 			loadbalancingexporter.NewFactory(),
 			otlphttpexporter.NewFactory(),
-			kafkaexporter.NewFactory(),
-		)
+		}
+		// some exporters should only be available when
+		// not in fips mode due to restrictions on crypto usage
+		exporters = addNonFipsExporters(exporters)
+		factories.Exporters, err = exporter.MakeFactoryMap(exporters...)
```
cmacknz, line 135: The prometheus and kafkareceivers are pretty functionally important. Is it even worth doing this vs just documenting they are potentially not compliant under some configurations? Or are they always non compliant?

We will need to figure out how to make these FIPS compliant eventually, and hopefully be able to push changes upstream. 

I'm not strictly opposed to this I just want to note it is kicking the can down the road.

simitt: How valuable is the otel mode when disabling kafka and prometheus receivers right now? What we have discussed was to potentially completely disable otel when using fips for now, and then have a dedicated workstream (outside of current scope) for fixing problematic libriaries in otel mode as this might be a larger undertaking.

simitt: Discussed offline that we want to keep anything from otel that is not problematic enabled and create follow up cards (https://github.com/elastic/elastic-agent-libs/issues/300) for fixing anything that is disabled here.

simitt (approved):

swiatekm (approved):

Issue #6970, opened 2025-02-21
[Flaky Test]: TestSwitchPrivilegedWithBasePath – Received unexpected error: exit status 1


pkoutsovasilis:
### Failing test case

TestSwitchPrivilegedWithBasePath

### Error message

Received unexpected error: exit status 1

### Build

https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/1988#01952460-12d0-4c7c-a666-1d79053f086c/288-934

### OS

Linux

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: testing/integration TestSwitchPrivilegedWithBasePath (57.95s)
    fetcher.go:93: Using existing artifact elastic-agent-9.0.0-SNAPSHOT-windows-x86_64.zip
    fixture.go:290: Extracting artifact elastic-agent-9.0.0-SNAPSHOT-windows-x86_64.zip to C:\Users\BUILDK~1\AppData\Local\Temp\TestSwitchPrivilegedWithBasePath3915442953
    fixture.go:308: Completed extraction of artifact elastic-agent-9.0.0-SNAPSHOT-windows-x86_64.zip to C:\Users\BUILDK~1\AppData\Local\Temp\TestSwitchPrivilegedWithBasePath3915442953
    fixture.go:908: Components were not modified from the fetched artifact
    fixture_install.go:199: [test TestSwitchPrivilegedWithBasePath] Inside fixture install function
    fixture_install.go:227: [test TestSwitchPrivilegedWithBasePath] Inside fixture installNoPkgManager function
    fixture.go:665: >> running binary with: [C:\Users\BUILDK~1\AppData\Local\Temp\TestSwitchPrivilegedWithBasePath3915442953\elastic-agent-9.0.0-SNAPSHOT-windows-x86_64\elastic-agent.exe install --base-path C:\ --force --unprivileged]
    fixture.go:665: >> running binary with: [C:\Elastic\Agent\elastic-agent.exe privileged -f]
[  ==] Failed to install service  [1s] Error: failed to install service (Elastic Agent): service Elastic Agent already exists
        For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/9.0/fleet-troubleshooting.html
    switch_privileged_test.go:134:
        	Error Trace:	C:/buildkite-agent/builds/bk-agent-prod-gcp-1740072248710733418/elastic/elastic-agent-extended-testing-bk/testing/integration/switch_privileged_test.go:134
[...]

PR #7589, opened 2025-03-26
[bug] wait for service to be reported as not found during switch

pkoutsovasilis:
## What does this PR do?

This PR introduces a platform-specific fix to ensure the Elastic Agent properly waits for the relevant Windows service to be fully removed before attempting to reinstall it.

Previously, attempting to reinstall the service immediately after an uninstall could result in a "service already exists" error, as the system may not have completed the deletion process. To address this, a new function `EnsureServiceRemoved` has been added. It connects to the Windows service manager and actively polls until the service is confirmed to be removed, ensuring that the uninstall has fully completed before proceeding with installation.

On non-Windows platforms, EnsureServiceRemoved is a no-op.

## Why is it important?

According to [Microsoft's documentation](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-deleteservice#remarks), calling `DeleteService` marks a service for deletion but does not guarantee immediate removal. If a process still has an open handle to the service or if the service is still stopping, the deletion will be deferred.

This lack of immediate consistency has been causing flakiness in CI for the test `TestSwitchPrivilegedWithBasePath`. The error indicates the service is still present when the agent tries to reinstall it, resulting in failure.

By explicitly waiting for the service to be fully removed, this PR avoids race conditions and improves test reliability on Windows.

## Checklist

## Disruptive User Impact

None expected. This change only impacts the internal reinstallation logic for Windows services and should be transparent to end users.

Run the flaky test `TestSwitchPrivilegedWithBasePath` multiple times on a Windows machine or CI runner to confirm that the issue does not reoccur.

If we have a way to invoke unit-tests with elevated privileged on Windows more than happy try and write a unit-test about that 🙂 

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6970

## Questions to ask yourself

- ...


internal/pkg/agent/install/switch.go
```diff
@@ -76,6 +77,12 @@ func SwitchExecutingMode(topPath string, pt *progressbar.ProgressBar, username s
 		pt.Describe(err.Error())
 	}
 
+	err = EnsureServiceRemoved(30*time.Second, 250*time.Millisecond, paths.ServiceName())
```
cmacknz, line 80: I see this in `SwitchExecutingMode`, shouldn't it also be in the normal install path, especially the force install case where it has to uninstall the existing agent?

pkoutsovasilis: yep more than a valid point, ty 🙂 This belongs also in the `Uninstall`. I missed completely that install can uninstall an existing agent as well

cmacknz: It doesn't necessarily need to go in Uninstall, it only matters there if agent is going to be installed again. To me, handling it in install makes more sense to avoid waiting on both ends of the process. 

We know this is going to be deleted eventually, and we only care about when it is specifically deleted if we are trying to install the service again.

pkoutsovasilis: I thought of the same thing, but what if a user uninstalls and then calls install right away?! 🙂 I mean this has to be really tight to happen but still. But with that said, I will make it only be part of the install

cmacknz: You could put it in uninstall, just make any error there non-fatal. It will make uninstall take longer unnecessarily if agent is being completely removed from a machine, so my preference is to keep it install where it is necessary but it's not a strong preference.

pkoutsovasilis: it makes sense to be in install and I added it there as well with dc2e259ee7008d5c142bf692945f77f86f15dbe8 🙂

pkoutsovasilis: I am gonna resolve this one, please feel free to unresolve if you spot anything else 🙂

internal/pkg/agent/install/install.go
```diff
@@ -166,6 +166,13 @@ func Install(cfgFile, topPath string, unprivileged bool, log *logp.Logger, pt *p
 
 	// install service
 	pt.Describe("Installing service")
+	// ensure that service is removed
+	err = EnsureServiceRemoved(30*time.Second, 250*time.Millisecond, paths.ServiceName())
+	if err != nil {
+		// error context already added by EnsureServiceRemoved
+		pt.Describe(err.Error())
```
ycombinator, line 173: I think we could use some additional context here, saying something like "failed to ensure service is removed". Otherwise, the progress tracker would say something like:
```
Installing service
failed to connect to service manager
```
IMO, it's not obvious that the second line comes from Agent ensuring that the service is removed.

ycombinator (approved):
Just a minor comment (not a blocker).  Otherwise, LGTM.

blakerouse (approved):
Looks good to me. Nice job!

internal/pkg/agent/install/install.go
pkoutsovasilis: more than a fair point @ycombinator , we should be able to distinguish them apart. I did push this 389cb5aeb19cb24d6e808313f2af87312b9b5ea1 I will wait for the CI to become green and the I will resolve this. That said, please feel free to mention/unresolve if you spot anything else 🙂

internal/pkg/agent/install/wait_service_windows.go
```diff
@@ -58,3 +60,41 @@ func isStopped(timeout time.Duration, interval time.Duration, service string) er
...
+				_ = s.Close()
+			case errors.Is(err, windows.ERROR_SERVICE_DOES_NOT_EXIST):
+				// The service is no longer installed
+				return nil
+			default:
+				// An unknown error occurred trying to open the service
+				return fmt.Errorf("error opening service (%s): %w", service, err)
+			}
+		case <-timer.C:
+			return fmt.Errorf("timed out after %s waiting for service (%s) to uninstall", timeout, service)
```
cmacknz, line 97: This path doesn't call `s.Close()`

cmacknz: That will leak a service handle:

```go
// Close relinquish access to the service s.
func (s *Service) Close() error {
	return windows.CloseServiceHandle(s.Handle)
}
```

pkoutsovasilis: hmm 🤔 I don't see this problem, I do open the service and close it when there is no error; according to [this](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-openservicew) (underlying windows call of OpenService)  on error the handle is NULL but I guess you say `better safe than sorry always close it`?!

However your comment made me look again, and I am not disconnecting from the `service control manager` when I go out of the scope of this function, so thanks 🙂

blakerouse: Yeah this looks to be calling `s.Close()` in the case that there is no error, which looks correct. Glad it help find them missing call to `m.Disconnect()`. I missed that as well in my review. I always forget that it also has to be called.

blakerouse (approved):
Looks good with the updated call to `m.Disconnect` and the `s.Close` looks correct.

cmacknz (approved):

PR #7600, opened 2025-03-27
Refactor self-monitoring config generation

swiatekm:
## What does this PR do?

Refactor how input configuration is generated for agent self-monitoring. This was primarily focused on readability. Assembling the configuration involved handling large maps of parameters, which are quite unwieldy to deal with in Go code.

## Why is it important?

## Checklist

## Related issues


ycombinator:
@swiatekm would it be possible to split up this PR into smaller PRs to make it easier to review?  Even excluding the golden file, it's still over 1k LOC, mostly coming from `internal/pkg/agent/application/monitoring/v1_monitor.go`.  Would it be possible to introduce the test first and then the refactorings in follow up PR(s)? Not sure if the sorting logic to make the test work is intimately tied to the refactoring, though.

swiatekm:
@ycombinator I can easily submit the test and the ordering change separately, as it's already its own commit. I could try to split the refactor into smaller changes as well, though that's going to be a bit more difficult, and might make understanding the change harder. I'll see what I can do there.

swiatekm:
Changes from this PR were split between #7734 and #7854, closing.

Issue #7592, opened 2025-03-26
[Windows] Shrink the disk footprint of the .zip package by including only one copy of elastic-agent.exe


cmacknz:
The Elastic Agent windows package is distributed as a .zip file. The agent package structure includes a symlink to the active version of the elastic-agent version at the root of package. The .zip format does not support symlinks, and so includes another copy of the elastic-agent.exe executable in place of the symlink that is then replaced with the expected symlink at install time. This increases the disk footprint prior to installation and download size of the windows .zip unnecessarily.

In the darwin-aarch64 .tar.gz, the root level elastic-agent is a symlink into the elastic-agent.app. The symlink shows as 0B and the elastic-agent executable is 175M for a total footprint of 175M after extraction.

```sh
ls -l ../builds/elastic-agent-9.0.0-rc1-darwin-aarch64/elastic-agent
lrwxr-xr-x@ 1 cmackenzie  staff  72 Mar 12 09:40 ../builds/elastic-agent-9.0.0-rc1-darwin-aarch64/elastic-agent -> data/elastic-agent-a7a8a1/elastic-agent.app/Contents/MacOS/elastic-agent

du -h ../builds/elastic-agent-9.0.0-rc1-darwin-aarch64/elastic-agent
  0B    ../builds/elastic-agent-9.0.0-rc1-darwin-aarch64/elastic-agent

du -h ../builds/elastic-agent-9.0.0-rc1-darwin-aarch64/data/elastic-agent-a7a8a1/elastic-agent.app/Contents/MacOS/elastic-agent
175M    ../builds/elastic-agent-9.0.0-rc1-darwin-aarch64/data/elastic-agent-a7a8a1/elastic-agent.app/Contents/MacOS/elastic-agent
```

In the windows-x86_64 .zip the root level elastic-agent.exe is not a symlink but a copy of the binary. The root elastic-agent.exe is 524M after extraction.

```sh
❯ du -h elastic-agent-8.18.0-windows-x86_64/elastic-agent.exe
262M    elastic-agent-8.18.0-windows-x86_64/elastic-agent.exe

❯ du -h elastic-agent-8.18.0-windows-x86_64/data/elastic-agent-63b735/elastic-agent.exe
262M    elastic-agent-8.18.0-windows-x86_64/data/elastic-agent-63b735/elastic-agent.exe
```

Change the package structure to eliminate the duplication on Windows, either as a special case or as a change to the installation process for all architectures for consistency.

[...]

PR #7540, opened 2025-03-24
updatecli: update other files

v1v:
## What does this PR do?

Add new files to be changed by the updatecli pipeline

## Why is it important?

See https://github.com/elastic/elastic-agent/pull/7454#issuecomment-2749210248

## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## How to test this PR locally

```diff
diff --git a/.ci/updatecli/values.d/scm.yml b/.ci/updatecli/values.d/scm.yml
index 2708262728..26308c9e94 100644
--- a/.ci/updatecli/values.d/scm.yml
+++ b/.ci/updatecli/values.d/scm.yml
@@ -1,8 +1,8 @@
 scm:
   enabled: true
-  owner: elastic
+  owner: v1v
   repository: elastic-agent
-  branch: main
+  branch: test/support-updatecli
   commitusingapi: true
   # begin updatecli-compose policy values
   user: 'github-actions[bot]'
```

Then I ran

```bash
$ GITHUB_TOKEN=$(gh auth token) \
  GITHUB_ACTOR=v1v \
  updatecli apply --config .ci/updatecli/updatecli-bump-golang.yml --values .ci/updatecli/values.d/scm.yml
```

<details><summary>Expand to view</summary>
<p>

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline ".ci/updatecli/updatecli-bump-golang.yml"

SCM repository retrieved: 1


++++++++++++++++++
+ AUTO DISCOVERY +
++++++++++++++++++



++++++++++++
+ PIPELINE +
++++++++++++



#########################################
# BUMP GOLANG-VERSION TO LATEST VERSION #
#########################################

source: source#minor
------------
[...]

PR #7601, opened 2025-03-27
[packaging] proxy binary to be for the root dir of windows archive 

pkoutsovasilis:
## What does this PR do?

This PR introduces a Windows-specific proxy binary (elastic-agent-archive-root.exe) that is placed at the root of the Elastic Agent .zip archive. The proxy acts as a thin wrapper that forwards execution to the actual agent binary located under the data/elastic-agent-{commit-sha} directory.

The proxy is built during the packaging step using a new Mage target (WindowsArchiveRootBinary), and replaces the duplicated agent binary that previously resided in the archive root due to lack of symlink support in .zip archives. This approach mimics the behavior seen on other platforms like Linux/macOS, where the root-level agent is a symlink into the versioned binary path.

This change avoids the need to duplicate the agent binary in the archive, reducing both the archive size and post-extraction disk footprint.

## Why is it important?

The Windows .zip archive currently includes a full copy of the elastic-agent.exe binary at the root of the archive due to the lack of symlink support in the zip format. This duplication, increases the download size and disk footprint unnecessarily

## Checklist

## Disruptive User Impact

There should be no disruptive user impact. The proxy binary is fully backward-compatible with existing workflows that invoke the elastic-agent.exe binary from the root of the archive on Windows. Users should not notice any difference in behavior.

1. Package the windows archive
   ```
   PACKAGES=zip EXTERNAL=true PLATFORMS=windows/amd64 SNAPSHOT=true mage -v package
   ```

2. Unzip the resulting .zip file and verify that:
* The root elastic-agent.exe is a lightweight proxy (~2MB instead of 260MB+)
* The actual binary lives under data/elastic-agent-<commit-sha>/elastic-agent.exe
* Running the root elastic-agent.exe launches the nested agent as expected

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/7592

## Manual Testing
1. DRA dry-run here and artifacts are produced successfully ([link](https://buildkite.com/elastic/elastic-agent-package/builds/5005#0195dc26-37b3-4e09-b6a7-9535b5f92914)) (ty @pchila 🙏 )
1. With the windows artifact from the above I directly installed a fleet-managed and agent got installed fine and reported healthy
2. I installed a standalone agent at the version v8.17.4. Then again from the windows artifact of the above DRA I did an upgrade and this was successful


magefile.go
```diff
@@ -1022,6 +1070,15 @@ func packageAgent(ctx context.Context, platforms []string, dependenciesVersion s
...
 	mg.Deps(Update)
 	mg.Deps(agentBinaryTarget, CrossBuildGoDaemon)
+
+	// compile the elastic-agent.exe proxy binary for the windows archive
+	for _, platform := range platforms {
+		if platform == "windows/amd64" {
+			mg.Deps(Build.WindowsArchiveRootBinary)
+			break
+		}
+	}
```
pchila, line 1080:
```suggestion
	if slices.Contains(platforms, "windows/amd64") {
		mg.Deps(Build.WindowsArchiveRootBinary)
	}
```

pchila (approved):
A couple of nitpicks, nothing blocking
Good to merge.

hack/windows/archive-proxy/main.go
```diff
@@ -0,0 +1,96 @@
...
+	exePath, err := os.Executable()
+	if err != nil {
+		log.Fatalf("Error getting executable path: %v\n", err)
+		return
+	}
+
+	exeAbsPath, err := filepath.Abs(exePath)
+	if err != nil {
+		log.Fatalf("Error getting executable absolute path: %v\n", err)
+		return
```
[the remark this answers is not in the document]
pkoutsovasilis: nice catch, removed this here 521e771eeeb0991f5d4ca26a43c9213f7cf0e726, please feel free to unresolve if you spot anything else 🙂

hack/windows/archive-proxy/main.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Fabricate the elastic-agent.exe path that reside inside the data/elastic-agent-{commit-short-sha} directory
+	exeTopPath := filepath.Dir(exeAbsPath)
+	nestedAgentBinaryPath := filepath.Join(exeTopPath, "data", fmt.Sprintf("elastic-agent-%s", CommitSHA), "elastic-agent.exe")
+	if _, err := os.Stat(nestedAgentBinaryPath); err != nil {
+		log.Fatalf("Unable to stat nested agent binary %q: %v\n", nestedAgentBinaryPath, err)
```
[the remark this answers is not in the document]
pkoutsovasilis: thanks for the rewording proposal, I pushed it here 6f5df45c300ccb81f93b601b76d47e56d4a78654, please feel free to unresolve if you spot anything else 🙂

magefile.go
pkoutsovasilis: sure why not, let's utilise standard library 🙂 pushed a commit with it here fe889d436efbbcbaf43820b3a2711c9acf09919e please feel free to unresolve

ebeahan (commented):
I'm unsure about introducing a top level dir named `hack` 😅

pkoutsovasilis:
> hack

😄 I am not afraid to admit that choosing where this would live was kinda of puzzle, but then I recalled [this one](https://github.com/kubernetes/kubernetes/tree/master/hack) and it seems appropriate to me. That said, more than happy to accept more "appropriate" proposals

hack/windows/archive-proxy/main.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/elastic/elastic-agent/pkg/core/process"
+)
+
+var CommitSHA string
```
swiatekm, line 18: Worth adding a comment that this is set by the linker at build time.

swiatekm (dismissed):
LGTM, comment not blocking.

hack/windows/archive-proxy/main.go
pkoutsovasilis: I am always up for improving readability, thanks for the point 🙂 I added it here bc4c38c8b1f51e1d7e79d16008e8ea6497daa2da, please feel free to unresolve if you spot anything else

swiatekm (dismissed):

ebeahan:
| more than happy to accept more "appropriate" proposals.

I don't have strong feelings about using `hack`, but I worry it could be seen and taken out of context.

`wrapper(s)`? If we think it's something we might address different in the future (famous words): `interim`?

blakerouse:
This change looks good overall. I agree with @ebeahan that `hack` as the top-level name is not correct. I don't think this is a hack, its more of a workaround due to limitations of zip on Windows. I like the idea of `wrapper` that @ebeahan suggested. 

Will wait for green CI before approving.

magefile.go
```diff
@@ -294,6 +294,54 @@ func (Build) GenerateConfig() error {
 	return sh.Copy(filepath.Join(buildDir, configFile), filepath.Join(metaDir, configFile))
 }
 
+// WindowsArchiveRootBinary compiles a binary to be placed at the root of the windows elastic-agent archive. This binary
+// is a thin proxy to the actual elastic-agent binary that resides in the data/elastic-agent-{commit-short-sha}
+// directory of the archive.
+func (Build) WindowsArchiveRootBinary() error {
```
cmacknz, line 300: To ease future development and testing, we could make it possible to build this for the other OSes. There's no reason it wouldn't work, it's just unnecessary. The job object APIs are already stubbed.

The archive binary is larger than a symlink so I'm not sure we want to use it by default, but perhaps a flag to switch to this so Windows can be taken out of the primary edit code/develop loop and only used for final testing. Dealing with Windows on Mac is a pain currently.

pkoutsovasilis:
ok folks, I hear you both. Let's avoid giving the "wrong" impression 🙂 I will resort to `wrapper` in a sec

hack/windows/archive-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	exeAbsPath, err := filepath.Abs(exePath)
+	if err != nil {
+		log.Fatalf("Error getting executable absolute path: %v\n", err)
+	}
+
+	// Fabricate the elastic-agent.exe path that reside inside the data/elastic-agent-{commit-short-sha} directory
+	exeTopPath := filepath.Dir(exeAbsPath)
+	nestedAgentBinaryPath := filepath.Join(exeTopPath, "data", fmt.Sprintf("elastic-agent-%s", CommitSHA), "elastic-agent.exe")
+	if _, err := os.Stat(nestedAgentBinaryPath); err != nil {
+		log.Fatalf("Unable to stat inner agent binary at %q: %v\n", nestedAgentBinaryPath, err)
```
cmacknz:
```suggestion
		log.Fatalf("Unable to execute elastic-agent.exe at %q: %v\n", nestedAgentBinaryPath, err)
```

stat has no meaning to Windows users who have never used Linux :)

cmacknz (commented):
I built this and installed it on a Windows VM. It works! I see the binary is replaced with a symlink at install time and the archive binary works as expected.

I tried to break things by changing paths and restricting permissions and noticed that the main error message has a Linux-ism in it. I left a suggestion to fix that (yes I know the underlying call is os.Stat but Windows people don't know that). It is also interesting that the root error has `CreateFile `in it because that's how Go checks for existence. It may be better to just try to execute it so the sycall in the error is correct. If the file doesn't exist we may get a clearer error like "cannot execute file not found"

```
2025/03/28 14:20:05 Unable to stat inner agent binary at "C:\\Users\\craig_mackenzie\\Downloads\\elastic-agent-9.1.0-SNAPSHOT-windows-x86_64\\data\\elastic-agent-fe889d\\elastic-agent.exe": CreateFile C:\Users\craig_mackenzie\Downloads\elastic-agent-9.1.0-SNAPSHOT-windows-x86_64\data\elastic-agent-fe889d\elastic-agent.exe: The system cannot find the path specified.

2025/03/28 14:22:04 Unable to stat inner agent binary at "C:\\Users\\craig_mackenzie\\Downloads\\elastic-agent-9.1.0-SNAPSHOT-windows-x86_64\\data\\elastic-agent-fe889d\\elastic-agent.exe": CreateFile C:\Users\craig_mackenzie\Downloads\elastic-agent-9.1.0-SNAPSHOT-windows-x86_64\data\elastic-agent-fe889d\elastic-agent.exe: Access is denied.
```

Otherwise, we are using `log` which puts timestamps in what we print to console which we otherwise don't do, but that small inconsistency isn't mandatory to fix.

Finally, as a follow up we should look at directing all of these errors into the windows event log as well. There are examples of how to do that but it should be fine to avoid this now.

magefile.go
pkoutsovasilis: that's the reason I didn't add a windows tag in the main.go of this wrapper subpackage. I can compile that from mac and linux, since this has CGO disabled, it compiles just fine from any OS but it targets Windows only. we want something more out of it? 🤔

hack/windows/archive-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		log.Fatalf("Error getting executable path: %v\n", err)
+	}
+
+	exeAbsPath, err := filepath.Abs(exePath)
+	if err != nil {
+		log.Fatalf("Error getting executable absolute path: %v\n", err)
+	}
+
+	// Fabricate the elastic-agent.exe path that reside inside the data/elastic-agent-{commit-short-sha} directory
```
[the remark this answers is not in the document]
pkoutsovasilis: thanks for catching this, fixed in 0a4702703a721290cc3a9f52e5c38fbe95400f71 please feel free to unresolve if you spot anything else 🙂

hack/windows/archive-proxy/main.go
pkoutsovasilis: valid point, reworded according to proposal here 316fc118304a26ff4e267a6dca316fd2890bd888, please feel free to unresolve if you spot anything else with it 🙂

hack/windows/archive-proxy/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/elastic/elastic-agent/hack/windows/archive-proxy
+
+go 1.24.1
```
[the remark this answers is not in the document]
pkoutsovasilis: I did push this c9a698b0ee1a0ecef6c70cd8042085b12576fbff which I believe takes care of this point, but I am not an fully confident. @pazone @v1v does the former change make sense to you? 🙂

magefile.go
cmacknz: You should be able to run this executable on non-Windows. We could allow building an agent package for Linux/Mac with the proxy binary but I'm fine just doing that with copy paste.

This is something we can address as a follow up, I don't want to block progress here with changes to this. I just want to highlight there is no actual cross-platform dependency here, besides our choice to use the proxy binary only on Windows.

hack/windows/archive-proxy/go.mod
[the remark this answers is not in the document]
v1v: LGTM, you can test it too, see the section `How to test this PR locally
` in https://github.com/elastic/elastic-agent/pull/7540


1. install https://www.updatecli.io/docs/prologue/quick-start/
2. create a test branch in your forked repository and push it.
3. update `.ci/updatecli/values.d/scm.yml` to point to your fork and branch
4. run

```bash
  GITHUB_ACTOR=v1v \
  updatecli apply \
    --config .ci/updatecli/updatecli-bump-golang.yml \
```

you should see a PR in your forked with those changes. As long as your branch is based on an older golang version, hence you can replace it too.

pkoutsovasilis:
> It may be better to just try to execute it so the sycall in the error is correct. If the file doesn't exist we may get a clearer error like "cannot execute file not found"

ok so if my understanding is correct remove os.Stat entirely and allow the error from `command.Start()` to be raised instead?

> Otherwise, we are using log which puts timestamps in what we print to console which we otherwise don't do, but that small inconsistency isn't mandatory to fix.
>
> Finally, as a follow up we should look at directing all of these errors into the windows event log as well. There are examples of how to do that but it should be fine to avoid this now.

ok I did push this e7c6ed69dee872b3bcfdccb07938af7252e7120f which uses fmt and also ships logs to eventlog using similar code to the elastic-agent one.

hack/windows/archive-proxy/go.mod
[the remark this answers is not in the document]
pkoutsovasilis: thanks for the immediate response @v1v I will create a follow up PR to handle just this case when this one makes it in 🙂

cmacknz:
> ok so if my understanding is correct remove os.Stat entirely and allow the error from command.Start() to be raised instead?

Try this and see if the error is clearer. The way I tested this was to rename the `data` sub-directory to something else then run `./elastic-agent.exe install` for reference to check what the command did.

cmacknz:
Screenshot 2025-03-28 at 12 38 52 PM

Something we changed is making Defender want to scan us before executing. That didn't happen the first time. My bet is it was the event log change. That's not worth this so I'd vote we pull that out.

cmacknz:
(there is a chance this is unrelated but I am no MS Defender whisperer)

pkoutsovasilis:
> (there is a chance this is unrelated but I am no MS Defender whisperer)

hmmm I did get one similar notification but I assumed that windows defender just decided to wake up 🙂 ok `better safe than sorry` removing the eventlog

cmacknz:
I can confirm writing to the event log does work, and doesn't happen when not running as an admin as expected.

cmacknz:
Yeah that's the first time I've ever seen defender pop up like that so I think we must have done something to cause it.

pkoutsovasilis:
ok resorted back to the initial log implementation but removed the timestamp inconsistency 🙂

cmacknz:
Tested latest changes again, still works. Curiously I got the Defender pop up again so it wasn't the event log. There seems to be no consequences to this as far as I can tell and I have no idea what is actually causing it. Perhaps the os.Stat implicitly doing CreateFile was part of it? Not sure how much we want to churn on this.

cmacknz:
https://support.microsoft.com/en-us/topic/what-is-a-cloud-security-scan-75112696-7660-4450-9194-d717f72a8ad8

There is a chance the defender pop up goes away once we test a version of the file that is properly signed as part of the unified release.

cmacknz (approved):
LGTM, we should probably include a changelog if the disk size is smaller than 8.17 noting how much this saved (if not, we don't need to highlight how much bigger it got)

pkoutsovasilis:
> LGTM, we should probably include a changelog if the disk size is smaller than 8.17 noting how much this saved (if not, we don't need to highlight how much bigger it got)

So I did a quick comparison we are still producing a ~ 50 MB larger archive compared to 8.17 🙂 But no more explosion when you decompress

blakerouse (approved):
CI passed, looks good!

pchila (approved):

pkoutsovasilis:
@ebeahan should we merge this? 🙂

Issue #6733, opened 2025-02-06
[Flaky Test]: TestStandaloneUpgradeRollbackOnRestarts – failed waiting for healthy agent and version (context deadline exceeded): could not unmarshal agent status output: unexpected end of JSON input


pchila:
### Failing test case

TestStandaloneUpgradeRollbackOnRestarts

### Error message

failed waiting for healthy agent and version (context deadline exceeded): could not unmarshal agent status output: unexpected end of JSON input

### Build

https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/1350#0194daae-65c2-4d84-83d3-99c13e0c7e13

### OS

Windows

### Stacktrace and notes

```markdown
According to the watcher logs the upgrade is not rolled back as expected
In the diagnostics we only find 9.0.0-SNAPSHOT logs (the test performs 8.18.0-SNAPSHOT -> 9.0.0-SNAPSHOT upgrade, rolled back to 8.18.0-SNAPSHOT) so it seems that 8.18.0 has been cleaned up confirmed by this watcher log
  
  {"log.level":"info","@timestamp":"2025-02-06T10:21:27.713Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade.cleanup","file.name":"upgrade/rollback.go","file.line":132},"message":"Removing hashed data directory","file.path":"C:\\Program Files\\Elastic\\Agent\\data\\elastic-agent-8.18.0-SNAPSHOT-eb771e","ecs.version":"1.6.0"}
  
  The interesting bit is that the watcher stated that the grace period is passed after about a minute, during which there are some connection errors, HEALTHY and STOPPED states, but not enough to trigger a rollback.
```

pchila:
New occurrence (although slightly different)
https://buildkite.com/elastic/elastic-agent/builds/19363#0195ec35-a401-4150-89db-8c09f0dfe75a/336-1054

In this case the test did not detect an `UpgradeDetails` section in the elastic-agent state output.

pchila:
This test is no longer flaky by extending the [watcher duration](https://github.com/elastic/elastic-agent/blob/48379406bf29bb50e8267032c900b886436f923c/testing/integration/ess/upgrade_rollback_test.go#L38) from 1m to 2m making sure that the watcher would still have time to see all the restarts.

[...]

PR #7602, opened 2025-03-27
Make components in packages configurable

### LINKED ISSUE #7610: Refactor required packages definition for packaging to correctly select dependencies
Elastic Agent packaging is using a list of ExpectedBinaries [defined in go code](https://github.com/elastic/elastic-agent/blob/12c5ef7e5ba594f8d422ecda99eed6c4c75fc0cd/dev-tools/mage/manifest/manifest.go#L100-L110).

The expected binaries list is then used when packaging in multiple places:

in `checksums.go` there is some prefix/suffix checks to match the necessary dependencies:
https://github.com/elastic/elastic-agent/blob/12c5ef7e5ba594f8d422ecda99eed6c4c75fc0cd/dev-tools/mage/checksums.go#L111-L119
https://github.com/elastic/elastic-agent/blob/12c5ef7e5ba594f8d422ecda99eed6c4c75fc0cd/dev-tools/mage/checksums.go#L202-L207

more assumptions are made for the package name based on the binary name, we can see that here
https://github.com/elastic/elastic-agent/blob/12c5ef7e5ba594f8d422ecda99eed6c4c75fc0cd/dev-tools/mage/manifest/manifest.go#L138-L143

This kind of assumptions created issues when `*-fips*`  package names started appearing in manifests since the code above started selecting the wrong packages since those would have a matching prefix and suffix.
This has led to a quick hack implemented in https://github.com/elastic/elastic-agent/pull/7486 to filter out fips packages for dependencies and allow packaging of elastic-agent to succeed.

The goal of this issue is to allow for greater control and more precise definition of necessary dependencies for elastic-agent packages, replacing the hacky FIPS dependency exclusion with better package selection checks.

This is a pre-requisite for implementing Elastic Agent FIPS packaging, where a different set of packages for the same elastic agent dependencies needs to be selected.

pchila:
## What does this PR do?

This PR allows declarative definitions of elastic-agent components in `dev-tools/packaging/packages.yml`.
In the declaration of such components it's now possible to specify a `packageName` template that will be rendered using platform and version during packaging.

## Why is it important?

This is a prerequisite for selecting different component packages for specific packaging specs.

## Checklist

## Disruptive User Impact

No disruptive user impact as neither the mage  build targets nor the final packages changed

### Package using a manifest
```shell
MANIFEST_URL="https://snapshots.elastic.co/9.1.0-15b6399f/agent-package/agent-artifacts-9.1.0-SNAPSHOT.json" AGENT_DROP_PATH=build/elastic-agent-drop  PLATFORMS="linux/amd64 windows/amd64 darwin/amd64" mage clean downloadManifest package
```

### Package without using a manifest
```shell
EXTERNAL=true PLATFORMS="linux/amd64" mage clean package
```

## Related issues

- Closes #7610

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/manifest/manifest.go
+++ b/dev-tools/mage/manifest/manifest.go
@@ -158,7 +158,13 @@ func (p Platform) Platform() string {
 	return p.OS + "/" + p.Arch
 }
 
-var AllPlatforms = []Platform{{"linux", "x86_64"}, {"linux", "arm64"}, {"windows", "x86_64"}, {"darwin", "x86_64"}, {"darwin", "aarch64"}}
+var AllPlatforms = []Platform{
+	{"linux", "x86_64"},
+	{"linux", "arm64"},
+	{"windows", "x86_64"},
+	{"darwin", "x86_64"},
+	{"darwin", "aarch64"},
+}
 
 // DownloadManifest is going to download the given manifest file and return the ManifestResponse
 func DownloadManifest(ctx context.Context, manifest string) (Build, error) {
--- a/dev-tools/mage/manifest/manifest.go
+++ b/dev-tools/mage/manifest/manifest.go
@@ -191,13 +191,16 @@ func resolveManifestPackage(project Project, spec packaging.BinarySpec, version 
 
 	// Try the normal/easy case first
 	packageName := spec.GetPackageName(version, platform)
+	if mg.Verbose() {
+		log.Printf(">>>>>>>>>>> Got packagename [%s], looking for exact match", packageName)
+	}
 	val, ok = project.Packages[packageName]
 	if !ok {
 		// If we didn't find it, it may be an Independent Agent Release, where
 		// the opted-in projects will have a patch version one higher than
 		// the rest of the projects, so we need to seek that out
 		if mg.Verbose() {
-			log.Printf(">>>>>>>>>>> Looking for package [%s] of type [%s]", spec.BinaryName, PlatformPackages[platform])
+			log.Printf(">>>>>>>>>>> Couldn't find exact match, looking for package [%s] of type [%s]", spec.BinaryName, PlatformPackages[platform])
 		}
 
 		var foundIt bool
--- a/dev-tools/mage/manifest/manifest.go
+++ b/dev-tools/mage/manifest/manifest.go
@@ -288,8 +288,15 @@ func relaxVersion(version string) (string, error) {
 	// check if there's more characters after the patch version
 	remainderIndex := matchIndices[3]
 	if remainderIndex < len(version) {
+		// This is a stricter regexp that allows flexibility ONLY on the patch version (prerelease and build metadata must be the same)
 		// if we have a remainder from the original version, add it escaping it once more
-		relaxedVersion += regexp.QuoteMeta(version[remainderIndex:])
+		// relaxedVersion += regexp.QuoteMeta(version[remainderIndex:])
+
+		// This is a looser regexp that allows anything beyond the major version to change (while still enforcing a valid patch version though)
+		// TODO: check with @dwhyrock why we allow more than just the patch version to change
+		// see TestResolveManifestPackage/Independent_Agent_Staging_8.14_apm-server and TestResolveManifestPackage/Independent_Agent_Staging_8.14_endpoint-dev
+		// Be more relaxed and allow for any character sequence after this
+		relaxedVersion += `.*`
 	}
 	return relaxedVersion, nil
 }
--- a/dev-tools/mage/manifest/manifest.go
+++ b/dev-tools/mage/manifest/manifest.go
@@ -19,7 +19,7 @@ import (
 	"github.com/magefile/mage/mg"
 	"golang.org/x/sync/errgroup"
 
-	"github.com/elastic/elastic-agent/dev-tools/mage/pkgcommon"
+	"github.com/elastic/elastic-agent/dev-tools/packaging"
 	"github.com/elastic/elastic-agent/pkg/version"
 )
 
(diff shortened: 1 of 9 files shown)


pchila:
Packaging pipeline run https://buildkite.com/elastic/elastic-agent-package/builds/5070

pkoutsovasilis:
@pchila bring in main, I believe all the CI failures reported here are caused by stale test-versions and it should be fixed now 🙂

pchila:
CI flakiness looks unrelated, opened issues for tracking
https://github.com/elastic/elastic-agent/issues/7640
https://github.com/elastic/elastic-agent/issues/6733#issuecomment-2766545780
https://github.com/elastic/elastic-agent/issues/7641
https://github.com/elastic/elastic-agent/issues/7642

dev-tools/mage/manifest/manifest_test.go
```diff
@@ -143,22 +146,77 @@ func TestResolveManifestPackage(t *testing.T) {
...
+			args: args{
+				version: "1.2.3-SNAPSHOT",
+			},
+			want:    `1\.2\.(?:0|[1-9]\d*).*`,
+			wantErr: assert.NoError,
+		},
+		{
+			name: "major-minor-patch-snapshot-buildmeta",
+			args: args{
+				version: "1.2.3-SNAPSHOT+build20250328112233",
```
dwhyrock, line 199: FWIW, the Independent Agent Release never builds any SNAPSHOT builds.  It only builds BC/Staging builds, so no need to check for anything with both `SNAPSHOT` and the `+build` metadata string in it.

dev-tools/packaging/packages.yml
```diff
@@ -584,6 +584,115 @@ shared:
...
+  - &linux-arm64
+    os: linux
+    arch: arm64
+  - &windows-amd64
+    os: windows
+    arch: x86_64
+  - &darwin-amd64
+    os: darwin
+    arch: x86_64
+  - &darwin-arm
```
dwhyrock, line 600: I thought we also produced a `darwin-universal` package, but it looks like Agent does not produce one.  We produce one for Endpoint for DRA.  Looks like we can possibly get rid of that one at some point if Agent isn't planning on ever producing one.

dev-tools/mage/manifest/manifest.go
```diff
@@ -251,76 +186,119 @@ func DownloadComponents(ctx context.Context, manifest string, platforms []string
...
+	relaxedVersion += anyPatchVersionRegexp
+	// check if there's more characters after the patch version
+	remainderIndex := matchIndices[3]
+	if remainderIndex < len(version) {
+		// This is a stricter regexp that allows flexibility ONLY on the patch version (prerelease and build metadata must be the same)
+		// if we have a remainder from the original version, add it escaping it once more
+		// relaxedVersion += regexp.QuoteMeta(version[remainderIndex:])
+
+		// This is a looser regexp that allows anything beyond the major version to change (while still enforcing a valid patch version though)
+		// TODO: check with @dwhyrock why we allow more than just the patch version to change
```
dwhyrock, line 296: Not fully sure what is meant by this.

For the IAR test version you have (`8.14.0+build202406201002`), most of the components ("opted-out") should have the same `Major.Minor.Patch` version as the IAR version (`8.14.0`).  

Any "opted-in" projects (`endpoint-dev`) should be exactly 1 patch version higher (`8.14.1`).  Nothing else should change.

It's also worth noting that none of the components in the IAR manifest will have the build metadata (`+build202406201002`).  That's only for the top level of the manifest.

Does that clear some things up?  Or have I misunderstood what you were saying here?

dwhyrock (commented):
A few comments/questions.

dev-tools/mage/manifest/manifest_test.go
pchila: This is a general test for the `relaxVersion` function which will replace the patch version with a regular expression matching any other patch expression.
Even if Independent Agent Release does not use a version with prerelease and buildmetadata right now, it's worth checking that the resulting regular expression still makes sense.

dev-tools/packaging/packages.yml
pchila: Elastic Agent doesn't produce a`darwin-universal` package.
In this PR I just moved the dependency listing from go code to YAML without changing it.
I also don't think this dependency is ever used but I would prefer if we perform a cleanup in a separate change, to reduce risk :sweat_smile:

dev-tools/packaging/packages.yml
```diff
@@ -584,6 +584,115 @@ shared:
...
+      - *linux-amd64
+      - *linux-arm64
+    pythonWheel: true
+    packageTypes: *all-package-types
+  - &comp-endpoint
+    projectName: endpoint-dev
+    packageName: endpoint-security-{{.Version}}-{{.Platform}}.{{.Ext}}
+    binaryName: endpoint-security
+    platforms: *all-platforms
+    packageTypes: *all-package-types
```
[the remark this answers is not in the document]
pchila: `packageType` indicate the *elastic-agent* package types that will need the dependency.
The line above means that `endpoint-security` will be included in all elastic-agent package types (tar.gz, zip, deb, rpm, docker images).
During packaging elastic-agent will only download the `tar.gz` and `zip` archives for endpoint :smiley:

dev-tools/mage/manifest/manifest.go
```diff
@@ -251,76 +186,119 @@ func DownloadComponents(ctx context.Context, manifest string, platforms []string
...
+	// Find the original version in the filename
+	versionIndex := strings.Index(packageName, dependencyVersion)
+	if versionIndex == -1 {
+		return nil, fmt.Errorf("no exact match and filename %q does not seem to contain dependencyVersion %q to try a fallback", packageName, dependencyVersion)
+	}
+
+	// TODO move relaxVersion to the version package so we can rewrite the version like so
+	//parseVersion, _ := version.ParseVersion(dependencyVersion)
+	//parseVersion.GetRelaxedPatchRegexp()
+	relaxedVersion, err := relaxVersion(dependencyVersion)
```
pkoutsovasilis, line 227: nit: wrap the logic of relaxedPackageName in a separate function, I believe this will help highlight that in `ResolveManifestPackage` we first try to do exact matching and then fallback to the relaxed one. Thus if any of the two has any issue we can work our fix in a more scoped footprint 🙂

pkoutsovasilis (commented):
In general the code changes look good to me and I can see that we are starting to embrace a more declarative with this PR which I am in favour. Waiting for @dwhyrock comments to be resolved 🙂

magefile.go
```diff
@@ -1441,7 +1460,7 @@ func downloadDRAArtifacts(ctx context.Context, manifestUrl string, downloadDir s
 	draDownloadDir := filepath.Join(downloadDir, build.BuildID)
 	err = os.MkdirAll(draDownloadDir, 0o770)
 	if err != nil {
-		return nil, fmt.Errorf("creating %q directory: %w", err)
+		return nil, fmt.Errorf("creating %q directory: %w", draDownloadDir, err)
```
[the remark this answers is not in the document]
pchila: kudos to @pkoutsovasilis for spotting that :smile:

dev-tools/packaging/packages.yml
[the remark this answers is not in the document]
dwhyrock: ah cool, thanks for the explanation.

dev-tools/packaging/packages.yml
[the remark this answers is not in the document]
pchila: Well spotted! will put arm64 there

dev-tools/packaging/settings.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Converts to the format expected on the mage command line "linux", "x86_64" = "linux/amd64"
+func (p Platform) Platform() string {
+	if p.Arch == "x86_64" {
+		p.Arch = "amd64"
+	}
+	if p.Arch == "aarch64" {
+		p.Arch = "arm64"
+	}
```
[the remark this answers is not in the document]
pchila: +1 Going to replace this with a `switch`

dev-tools/mage/manifest/manifest.go
pchila: Agreed, `ResolveManifestPackage` has become a bit messy, I will refactor it in the upcoming FIPS packaging PR as there are other modification in the same area of code

dev-tools/packaging/settings.go
```diff
@@ -0,0 +1,146 @@
...
+	return p.OS + "/" + p.Arch
+}
+
+//go:embed packages.yml
+var packageSpecBytes []byte
+
+func init() {
```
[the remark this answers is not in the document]
pchila: Done in 39e766ded898d398bc6ff69c7cbf50cabd9fd8e3

dev-tools/mage/manifest/manifest.go
pchila: At this time the loose implementation works in the Independent Agent Release scenario.
For the unified release, all dependencies have the same version so we always get an exact match.

I don't really have a preference and it's not gonna be a problem if we don't have to deal with multiple versions of the same package (today this is not allowed in IAR or unified release manifests)

I am going to remove the TODO and leave the loose implementation in place.
With the `relaxVersion` function in place, we can fine-tune the regexp generation to suit more advanced cases if we need to.

dwhyrock: Sounds great to me.

dwhyrock (commented):
LGTM

dwhyrock (approved):

pkoutsovasilis (approved):
LGTM

dev-tools/mage/manifest/manifest.go
pchila: Refactor done in https://github.com/elastic/elastic-agent/pull/7690

PR #7622, opened 2025-03-28
otel: add agent input metrics test for Beat receivers

mauri870:
## What does this PR do?

This commit adds a test to serve as a baseline for comparing agent metrics with Beats receivers. The test uses the default metricsets for the metrics input and asserts that they were ingested into Elasticsearch.

## Checklist

## Related issues

<!-- Recommended
Link related issues below. Insert the issue link or reference after the word "Closes" if merging this should automatically close it.

- Closes https://github.com/elastic/ingest-dev/issues/4882

### DIFF
--- a/testing/integration/beat_receivers_test.go
+++ b/testing/integration/beat_receivers_test.go
@@ -14,6 +14,7 @@ import (
 	"io"
 	"net/http"
 	"runtime"
+	"strings"
 	"testing"
 	"text/template"
 	"time"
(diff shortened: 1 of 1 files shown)


testing/integration/beat_receivers_test.go
```diff
@@ -480,3 +481,159 @@ service:
...
+		index := fmt.Sprintf(".ds-metrics-system.%s-%s*", mset, info.Namespace)
+		require.Eventuallyf(t,
+			func() bool {
+				findCtx, findCancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer findCancel()
+
+				docs, err := estools.PerformQueryForRawQuery(findCtx, rawQuery, index, info.ESClient)
+				require.NoError(t, err)
+
+				return docs.Hits.Total.Value > 0
```
swiatekm, line 631: Could we also verify the shape of the data here?

mauri870: Yes, this is part of a future task. This PR is similar to https://github.com/elastic/elastic-agent/pull/7367. This is a baseline test that will be expanded upon to include an equivalent otel configuration and a comparison between the documents ingested by normal agent and otel.

leehinman (dismissed):

mauri870:
@ycombinator @pchila I would appreciate a review. Thanks!

testing/integration/beat_receivers_test.go
```diff
@@ -480,3 +481,159 @@ service:
...
+		ESApiKey        string
+		BeatsESApiKey   string
+		FBReceiverIndex string
+		Namespace       string
+	}
+	esEndpoint, err := getESHost()
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
```
pchila, line 515: It would be a good practice not to log an esApiKey (even if it's just a test one).
Furthermore, if you wanna format your assertion failure message, you probably need `require.Truef`

testing/integration/beat_receivers_test.go
```diff
@@ -480,3 +481,159 @@ service:
...
+				ESEndpoint:    esEndpoint,
+				ESApiKey:      esApiKey.Encoded,
+				BeatsESApiKey: string(beatsApiKey),
+				Namespace:     info.Namespace,
+			}))
+	configContents := configBuffer.Bytes()
+	t.Cleanup(func() {
+		if t.Failed() {
+			t.Log("Contents of agent config file:\n")
+			println(string(configContents))
```
pchila, line 562: printing more secrets here

pchila (commented):
Aside from some logs that contain some ES API Keys that are created and never removed once the test is finished, I wonder why is this test here?

By the looks of it this is a simple metricbeat test collecting system metrics: why is this test necessary here in elastic-agent repository ?

mauri870:
> By the looks of it this is a simple metricbeat test collecting system metrics: why is this test necessary here in elastic-agent repository ?

This is a baseline test. It may seem simple now, but it will be part of a larger test. This test only persists default agent metrics to Elasticsearch, but it will be expanded to include a similar test for metrics ingested with OTel, followed by a comparison to ensure they are equivalent. These additions will be introduced separately and will build upon the current test.

pchila:
> This is a baseline test. It may seem simple now, but it will be part of a larger test. This test only persists default agent metrics to Elasticsearch, but it will be expanded to include a similar test for metrics ingested with OTel, followed by a comparison to ensure they are equivalent. These additions will be introduced separately and will build upon the current test.

So what is the actual value of this test in its current form? Maybe we should wait until we have the rest of the Otel receiver stuff to compare it to something. 
As it's written now it has no relevance to agent features but we are still gonna run it? What is the added value here ?

testing/integration/beat_receivers_test.go
mauri870: I was under the assumption that CI test logs were private. Almost every test in this file and in otel_test.go print the message in the same way, we would have to fix those as well.

testing/integration/beat_receivers_test.go
mauri870: Almost every single test prints the config for easier debugging in case of test failures. If this is really a concern then we would have to do a global cleanup like I mentioned in the other thread.

mauri870:
> So what is the actual value of this test in its current form? Maybe we should wait until we have the rest of the Otel receiver stuff to compare it to something. As it's written now it has no relevance to agent features but we are still gonna run it? What is the added value here?

You have a solid point, I cannot justify the cost of running this test in every build without any added benefit of coverage since it is a very simple test. The value of splitting this tasks is to simplify development as the e2e tests are very tedious to write and test locally so having simpler build blocks actually speeds up development as multiple people could work on them in parallel. Since I'm assigned to both tasks I can live without merging this and only push a single PR.

pchila:
> > So what is the actual value of this test in its current form? Maybe we should wait until we have the rest of the Otel receiver stuff to compare it to something. As it's written now it has no relevance to agent features but we are still gonna run it? What is the added value here?
> 
> You have a solid point, I cannot justify the cost of running this test in every build without any added benefit of coverage since it is a very simple test. The value of splitting this tasks is to simplify development as the e2e tests are very tedious to write and test locally so having simpler build blocks actually speeds up development as multiple people could work on them in parallel. Since I'm assigned to both tasks I can live without merging this and only push a single PR.

If there are issues with the development as the e2e tests are very tedious to write and test locally, that is something we should address then, feel free to file an issue under elastic/elastic-agent where you describe the pain points and any proposal you might have.
Using CI as an e2e test runner is not a sustainable solution, there are dedicated mage targets for testing locally that can be used. If those are not usable in their current form we should probably fix them or replace them altogether.

pkoutsovasilis:
Thanks for the discussion so far @pchila @mauri870 - it’s been helpful to follow the rationale.

Please correct me if I’m wrong, but the way I understand the discussion above is that we’re aiming to introduce a base test that will help expedite the development of more complex scenarios down the line (e.g., comparing default agent metrics vs. OTel-ingested ones)?! If that’s the case, and the purpose is purely to support local development workflows, I wonder if we should avoid running this in CI altogether.

Perhaps we could still keep the test code in the repo - so that other developers can run it locally while working on their respective parts - but mark it with a skip so it doesn’t run in CI? This way, we can still benefit from the simplified development setup without incurring the CI cost unnecessarily.

Happy to hear other thoughts on this! 🙂

leehinman:
> Thanks for the discussion so far @pchila @mauri870 - it’s been helpful to follow the rationale.
> 
> Please correct me if I’m wrong, but the way I understand the discussion above is that we’re aiming to introduce a base test that will help expedite the development of more complex scenarios down the line (e.g., comparing default agent metrics vs. OTel-ingested ones)?! If that’s the case, and the purpose is purely to support local development workflows, I wonder if we should avoid running this in CI altogether.
> 
> Perhaps we could still keep the test code in the repo - so that other developers can run it locally while working on their respective parts - but mark it with a skip so it doesn’t run in CI? This way, we can still benefit from the simplified development setup without incurring the CI cost unnecessarily.
> 
> Happy to hear other thoughts on this! 🙂

Long term, we want to get to the point that we have tests that show that data (monitoring as well as from integrations) ingested with beats receivers is functionally equivalent to data ingested with classic beats as individual processes.  This is not a one time test, it needs to happen continuously because changes in beats, elastic-agent, elastic-agent-libs, elastic-agent-autodiscover, elastic-agent-system-metrics, and opentelemetry-collector-contrib/exporter/elasticsearch all impact this.  These tests will have to be run in CI until such a time as we no longer expect customers to be switching from classic beats to beat receivers (years).  This is not to "purely support local development workflows".  This is to provide a level of confidence that when we ask users to switch from classic beats as processes to beat receivers that their data will remain intact and be usable.  That all of their detections, dashboards and alerts will work as expected.

Short term, not everything with beats receivers is 100% ready to go, but classic beats as individual process is.  So the idea was to start on the parts of the tests that use classic beats as individual processes.  Then as beats receivers gain the needed bits, we add to the these existing tests.  The fact that these look like existing tests is both expected and by design.

How about we add a `Skip` to these tests until we can add the beat receiver bits?  That way it still provides a level of parallelism of the work, but won't make CI more costly/complicated until they have unique value.

mauri870:
Apologies for the delay in replying. Now I believe it makes sense to merge this once it's fully implemented. I already have a test for agent metrics input and another one with the equivalent OTel configuration, so what's left is to combine the two tests and assert that the resulting documents match. I plan to focus on this next.

testing/integration/beat_receivers_test.go
```diff
@@ -356,6 +360,1666 @@ func TestAgentMonitoring(t *testing.T) {
...
+			Namespace       string
+			OS              string
+		}
+		esEndpoint, err := getESHost()
+		require.NoError(t, err, "error getting elasticsearch endpoint")
+		esApiKey, err := createESApiKey(info.ESClient)
+		require.NoError(t, err, "error creating API key")
+		require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
+
+		configTemplate := `agent.logging.level: info
```
swiatekm, line 1335: You can use the same template as for the non-otel case and just set `_runtime_experimental: otel` in the input.

mauri870: Thanks! Done.

mauri870: I had to revert this—I forgot that the OTel translation doesn't work for metrics input. That should be covered by https://github.com/elastic/ingest-dev/issues/4886.

mauri870:
Abandoning this PR, tests were added as part of https://github.com/elastic/elastic-agent/pull/8148.

Issue #6875, opened 2025-02-14
Beats shutdown issue


bjmcnic:
For confirmed bugs, please report:
- Version: At least 8.17.1  (likely earlier)-> 9.0.0-beta
- Operating System: Windows Server (tested on 2019)
- Steps to Reproduce:
      * Setup the Microsoft DNS server
      * Install Elastic Agent
      * Add the Microsoft DNS Server Integration with defaults
      * Stop the Elastic Agent service
      * Notice the `Elastic-DNSServer-Analytical` ETW trace remains

```
PS C:\> logman query -ets

Data Collector Set                      Type                          Status
-------------------------------------------------------------------------------
AppModel                                Trace                         Running
Elastic-DNSServer-Analytical            Trace                         Running
DiagLog                                 Trace                         Running
EventLog-Application                    Trace                         Running
EventLog-System                         Trace                         Running
NtfsLog                                 Trace                         Running
...
```

This is just a specific testable scenario that speaks to a potentially broader issue.  The integration within the filebeat portion of the `agentbeat.exe` process has code to stop that trace upon the integration being stopped, but that code never runs when Agent stops.  Additionally, the code does not run when Agent is left running, but the integration is removed.  

Through debugging, it's been observed during the Elastic Agent service stop that the `elastic-agent.exe` service process calls `NtTerminateProcess()` upon the `agentbeat.exe` process hosting the filebeat integration.  The `agentbeat.exe` process does not itself run its cleanup code and cleanly exit.

It would seem that IPC between the `elastic-agent.exe` process and the subordinate `agentbeat.exe` process is not occurring in a way to trigger clean shutdowns, at least in this instance.

blakerouse:
The Elastic Agent calls proc.Kill where proc is `*os.Process` on Windows. - https://github.com/elastic/elastic-agent/blob/main/pkg/core/process/cmd.go#L35

There is no `Stop` on golang, and on Windows golang only supported signal is `Kill` which calls `syscall.TerminateProcess`

jlind23:
@blakerouse does it mean this is an expected behavior and not a bug?

belimawr:
The documentation for `os.Process.Kill` states:
> Kill causes the [Process](https://pkg.go.dev/os#Process) to exit immediately. Kill does not wait until the Process has actually exited. This only kills the Process itself, not any other processes it may have started.

[...]

PR #7624, opened 2025-03-28
6875 do not kill process on windows

belimawr:
## What does this PR do?

Calling proc.Kill causes the process to exit immediately, which
prevents Beats from executing some clean up functions.

Now windows.GenerateConsoleCtrlEvent to send a CTRL+C signal that is
haled as a SIGINT, thus allowing the Beat to execute all necessary
clean up.

## Why is it important?

It allows Beats or any subprocess from the Elastic-Agent to gracefully terminate

## Checklist

~~## Disruptive User Impact~~

TODO

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/6875

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1743191647-gracefully-terminate-windows-process.yaml
+++ b/changelog/fragments/1743191647-gracefully-terminate-windows-process.yaml
@@ -1,0 +1,38 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: gracefully terminate windows process
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: |
+  Calling proc.Kill causes the process to exit immediately, which
+  prevents Beats from executing some clean up functions.
+  
+  Now windows.GenerateConsoleCtrlEvent to send a CTRL+C signal that is
+  haled as a SIGINT, thus allowing the Beat to execute all necessary
+  clean up.
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/7624
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: https://github.com/elastic/elastic-agent/issues/6875
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -7,6 +7,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"os/signal"
@@ -356,6 +357,36 @@ func runElasticAgent(ctx context.Context, cancel context.CancelFunc, override ap
 	// listen for signals
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT, syscall.SIGHUP)
+	//HERE
+	go func() {
+		logger := logp.L().Named("debug-server")
+		print := func(msg ...string) {
+			final := strings.Join(msg, " ")
+			logger.Info(final)
+			fmt.Println("--------------------", final)
+		}
+		mux := http.NewServeMux()
+		mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+			print("Request:", r.URL.String(), "Query:", r.URL.RawQuery)
+			terminateQs := r.URL.Query().Get("terminate")
+			terminate, err := strconv.ParseBool(terminateQs)
+			if err != nil {
+				msg := fmt.Sprintf("Invalid query parameter %q: %s", terminateQs, err)
+				print(msg)
+				w.Write([]byte(msg))
+				return
+			}
+
+			if terminate {
+				print("Sending SIGTERM into the channel")
+				signals <- syscall.SIGTERM
+			}
+		})
+
+		print("Starting server on :4000")
+		http.ListenAndServe(":4000", mux)
+	}()
+
 	isRex := false
 	logShutdown := true
 LOOP:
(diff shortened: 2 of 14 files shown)


cmacknz:
It does look to me like Go's OS package automatically converts `CTRL_BREAK_EVENT` to `SIGINT` for us so that the Beats and other client's won't need any other changes to handle this. I didn't expect this when I suggested `CTRL_BREAK_EVENT`, I thought we'd need to implement something on the client side as well.

https://github.com/golang/go/blob/b9cbb65384f6bebd58f7a8354759b8c7b1bbb13f/src/runtime/os_windows.go#L1097-L1102

cmacknz:
Also, to note the outcome of https://github.com/elastic/elastic-agent/issues/6875 is not "Beats get SIGINT" it is "Beat graceful shutdown logic executes" which is slightly different. 

This is the only change necessary if graceful shutdown on windows is correctly hooked up to SIGINT and I wouldn't trust that without explicitly testing it. There is another bug report about lack of graceful shutdown in APM server on Linux that is a hint to me that even if you got SIGINT, you might not be done fixing the problems.

cmacknz:
Also on Linux, we use SIGTERM: https://github.com/elastic/elastic-agent/blob/6b4b0dc69bca88fbdd4e5dca869645c05755e734/pkg/core/process/cmd_linux.go#L55-L58

So using SIGINT on Windows creates inconsistency, looking at the Go source we'd want one of these three signals to mimic SIGINT on Windows.

https://github.com/golang/go/blob/b9cbb65384f6bebd58f7a8354759b8c7b1bbb13f/src/runtime/os_windows.go#L1097-L1102
```go
	case _CTRL_CLOSE_EVENT, _CTRL_LOGOFF_EVENT, _CTRL_SHUTDOWN_EVENT:
		s = _SIGTERM
```

If we for some reason want to switch to SIGINT, we should do it everywhere if the behavior is different.

belimawr:
> Also, to note the outcome of #6875 is not "Beats get SIGINT" it is "Beat graceful shutdown logic executes" which is slightly different.

Yes, I'm well aware of it. I tested your suggestion of sending `CTRL_C_EVENT` and it worked perfectly with the current small change on this PR. I tested with Filebeat and the whole shutdown process works well, I even tested with the EWT integration as described by #6875, and the ETW trace is removed as expected, as well as I can see all the shutdown logs from Filebeat.

> This is the only change necessary if graceful shutdown on windows is correctly hooked up to SIGINT and I wouldn't trust that without explicitly testing it. There is another bug report about lack of graceful shutdown in APM server on Linux that is a hint to me that even if you got SIGINT, you might not be done fixing the problems.

Once again, yes, that is the case. All Beats handle SIGINT correctly, we even log it as the cause for the shutdown. I'm writing integration tests that ensures the Beats (Filebeat) shutdown process works correctly, so not only the Elastic-Agent is sending the correct signal, but also ensure Beats is handling it as expected.

cmacknz:
Awesome thanks for confirming!

Then my main comment is to prefer keeping to SIGTERM for consistency. We have several agent clients that are not Beats or even written in Go and introducing SIGINT instead introduces inconsistency that might not be handled properly everywhere.

belimawr:
> Awesome thanks for confirming!
> 
> Then my main comment is to prefer keeping to SIGTERM for consistency. We have several agent clients that are not Beats or even written in Go and introducing SIGINT instead introduces inconsistency that might not be handled properly everywhere.

The Elastic-Agent code already uses SIGTERM to all OSes but Windows, I'll look on how to send one of the Windows signals Go converts to SIGTERM on Windows.

Looking at the [Go doc](https://pkg.go.dev/os/signal):
> Additionally, if Notify is called, and Windows sends CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT or CTRL_SHUTDOWN_EVENT to the process, Notify will return syscall.SIGTERM. Unlike Control-C and Control-Break, Notify does not change process behavior when either CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT or CTRL_SHUTDOWN_EVENT is received - the process will still get terminated unless it exits. But receiving syscall.SIGTERM will give the process an opportunity to clean up before termination.

That might not be exactly what we want. It is a bit unclear to me what *process will still get terminated* exactly means for the process (is it a grace period?), anyways I'll do some more research and testing to find a reliable solution.

belimawr:
Quick update on this PR: There are two problems preventing the graceful shutdown of Beats:
1. The signal was not correctly sent on Windows, we were killing the Beat instead of signalling for a graceful shutdown.
2. Even if we send the correct signal, Elastic-Agent does not wait for the components to gracefully shutdown and exit when the agent itself is shutting down. Which makes the Elastic-Agent process to terminate before the components/Beats have time to actually gracefully exit.

I have a fix for `#1` (315c33f) and a lot of print debug (1f6834d) investigating `#2`

cmacknz:
> Even if we send the correct signal, Elastic-Agent does not wait for the components to gracefully shutdown and exit when the agent itself is shutting down. Which makes the Elastic-Agent process to terminate before the components/Beats have time to actually gracefully exit.

Fix 2 in a separate PR or just defer it to a separate issue, this will need to be a net new feature that can be configured. Not all components are Beats or well behaved, and in the case of Defend has a lifetime separate from agent.

The only case I would consider tackling here is uninstall where we have a nice place to hook into (only handles defend right now) and the configuration for wait time can be handled with a command line parameter. This still feels like a candidate for a separate PR.

https://github.com/elastic/elastic-agent/blob/3229be39c543ea278bebaac99520b476268e19c6/internal/pkg/agent/install/uninstall.go#L124-L134

Solving this sounds like it will require multiple steps I'd rather treat the original issue more like a parent issue to the sub-issue steps then try to get it all right at once.

cmacknz:
For the change in https://github.com/elastic/elastic-agent/commit/315c33fee7cce1669a3d29cfd448573460945ec7, make sure you have no defeated the job hierarchy that let's the sub-processes terminate when the agent as parent process terminates.

You want to be able to signal the child processes independent of the parent, but if the parent is terminated the child processes need to terminate as well.

belimawr:
> For the change in [315c33f](https://github.com/elastic/elastic-agent/commit/315c33fee7cce1669a3d29cfd448573460945ec7), make sure you have no defeated the job hierarchy that let's the sub-processes terminate when the agent as parent process terminates.
> 
> You want to be able to signal the child processes independent of the parent, but if the parent is terminated the child processes need to terminate as well.

@cmacknz I've been research about it and doing some testing, and even without setting any flag (a plain `exec.Command`), on Windows, if the parent exits, the child stays executing. Setting `CREATE_NEW_PROCESS_GROUP` does not directly affect the behaviour of the child once the parent process exits.

The good news is that when Elastic-Agent runs and exits, it does a good job signalling Beats to exit and Beats do a good job setting the correct signal handlers and responding to them.

The Elastic-Agent even launches a goroutine that kills the child process if it does not exit in a pre-defined timeout (default is 30s).

If anything, the Elastic-Agent process is doing a pretty good job at ensure all of its child process stop running once it exits. Hence, Beats not running their teardown.

I haven't dug deep enough to find exactly the point in the code the Beats get killed during the shutdown. Tomorrow I'll look a little more into it.

belimawr:
I finally found it! The TL;DR: (I'll create an issue describing the problem in details)
At startup we create a "JobObject", then whenever a child process is created, we assign it to this JobObject, once the process holding the JobObject (the Elastic-Agent process) exits, all other process associated with it are killed.

belimawr:
I created a detailed issue: https://github.com/elastic/elastic-agent/issues/7756

belimawr:
Closing this PR in favour of https://github.com/elastic/elastic-agent/pull/7738 and the follow up issue https://github.com/elastic/elastic-agent/issues/7756

cmacknz:
> At startup we create a "JobObject", then whenever a child process is created, we assign it to this JobObject, once the process holding the JobObject (the Elastic-Agent process) exits, all other process associated with it are killed.

Yes, the JobObject is like the Windows version of a Linux cgroup.

Issue #7309, opened 2025-03-10
Investigate a panic error on agentless


orouz:
the [`too many restarts`](https://overview.elastic-cloud.com/app/observability/alerts/3102ca98-4741-4db7-b44b-ba5e267996c5) alert was triggered and going over the logs i can see that:

1. agent starts shutting down
2. after some work, we get this error:
```
failed to load metadata: failed to create new agent info: could not get agent info from store: failed to load from ioStore: failed to ensure key during encrypted disk store Load: could not get agent key: failed to acquire exclusive loc
```

3. then there’s a stack trace, one line in each log that reads:
```
github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go:177 +0xc4
	github.com/elastic/elastic-agent/internal/pkg/agent/application/managed_mode.go:231 +0x68
	github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go:304 +0x1c0
github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet.(*FleetGateway).doExecute(0x4001ba81b0, {0xaaaacc468f48, 0x40001e1090}, {0xaaaacc4423e0, 0x4002ad2080})
github.com/elastic/elastic-agent/internal/pkg/runner.Start.func1()
	github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet.(*FleetGateway).Run(0x4001ba81b0, {0xaaaacc468f48, 0x40001e1090})
created by github.com/elastic/elastic-agent/internal/pkg/runner.Start in goroutine 138
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xaaaac3f57da0]
github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go:151 +0x158
github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet.(*FleetGateway).execute(0x4001ba81b0, {0xaaaacc468f48, 0x40001e1090})
github.com/elastic/elastic-agent/internal/pkg/agent/application.(*managedConfigManager).Run.func4({0xaaaacc468f48?, 0x40001e1090?})
goroutine 152 [running]:

github.com/elastic/elastic-agent/internal/pkg/runner/runner.go:69 +0x3c
	panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #7629, opened 2025-03-31
[bug] do not panic when info.Metadata errors in FleetGateway execute

pkoutsovasilis:
## What does this PR do?

This PR fixes a bug in the `FleetGateway.execute` method where a call to `info.Metadata` could return an error that was not being propagated. If this error occurred during shutdown (e.g. due to context is done/cancelled), the subsequent logic would attempt to use `nil` values, resulting in a panic.

The fix ensures that the `execute` function returns early if `info.Metadata` fails, avoiding a `nil` dereference and the resulting crash.

## Why is it important?

This bug was causing panics during agent shutdown, as reported in [issue #7309](https://github.com/elastic/elastic-agent/issues/7309). The error originated from the failure to load agent metadata, which led to a `SIGSEGV` (segmentation fault) because the error was logged but not returned. This led to `nil` being used downstream, triggering a crash.

Fixing this ensures more graceful shutdown behaviour and prevents unexpected restarts of the Elastic Agent under these failure conditions.

## Checklist

## Disruptive User Impact

No disruptive impact expected. The change improves error handling during shutdown, preventing panics and unnecessary agent restarts. It does not alter runtime behaviour under normal conditions.

Deploy a k8s agent pod and delete the pod, observe that it no longer panics with this reason

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/7309

### DIFF
--- a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
+++ b/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
@@ -337,6 +337,7 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse, 
 	ecsMeta, err := info.Metadata(ctx, f.log)
 	if err != nil {
 		f.log.Error(errors.New("failed to load metadata", err))
+		return nil, 0, err
 	}
 
 	// retrieve ack token from the store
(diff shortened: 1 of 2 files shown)


internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -337,6 +337,7 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
 	ecsMeta, err := info.Metadata(ctx, f.log)
 	if err != nil {
 		f.log.Error(errors.New("failed to load metadata", err))
+		return nil, 0, err
```
cmacknz, line 340: I noticed we did this by looking at an SDH in 8.16 where we probably send empty or malformed metadata.

I think this approach is incorrect because it will prevent agent from checking in at all. This includes checking in to get upgrade actions to fix this bug. We need to not panic but degrade the content of the checkin payload instead.

PR #7190, opened 2025-03-05
Increase timeout for assertions in TestSetLogLevelFleetManaged

pchila:
## What does this PR do?

Fleet checkins may take up to 5 minutes so we need to increase timeouts for assertions in TestSetLogLevelFleetManaged to account for it as the agent may apply log level settings while there's already a Fleet check-in in progress.
Settings timeouts for retrieving the correct log level in Fleet to 6 minutes should give enough time to agent to set the log level and communicate it to Fleet in a new check-in request.

## Why is it important?
This PR should remove flakiness coming from elastic-agent check-in timing in TestSetLogLevelFleetManaged.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Related issues

- Closes #5193

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #7635, opened 2025-03-31
Fix upgrade for same versions

michalpristas:
## What does this PR do?

What we're seeing here is upgrade marked uncomplete in 2 scenarios.
First one is when upgrade action for same version arrives for the first time.
We go through logic up until the steps and then we set nil to details. What we're not doing is Acking the action though. 
This can lead to having upgraded at property still populated and hence upgrade showing as ongoing.

Second case is when we trigger the action second time, maybe in a hope to fix the label/state.
In this case we have already first one in a local state and we detect that this action is upgrade to same version, althought it's not the same action. We append local queue of actions and do nothing. Again there is a need to ack because upgraded at property is set.

Logic determining Updating label looks like this

```javacript
function isAgentUpgrading(agent: Agent) {
  if (agent.upgrade_details) {
    return agent.upgrade_details.state !== 'UPG_FAILED';
  }
  return agent.upgrade_started_at && !agent.upgraded_at;
}
```

## Why is it important?

Leads to weird state being displayed in a UI

## Checklist

- [ x I have added tests that prove my fix is effective or that my feature works

Fixes: #6186


testing/integration/upgrade_fleet_test.go
```diff
@@ -341,6 +341,135 @@ func testFleetAirGappedUpgrade(t *testing.T, stack *define.Info, unprivileged bo
...
+		5*time.Minute, time.Second,
+		"last error: %v. agent.UpgradeDetails: %s",
+		err, agentUpgradeDetailsString(agent))
+
+	t.Log("Waiting for agent to have empty upgrade details")
+	require.Eventuallyf(t, func() bool {
+		agent, err = kibClient.GetAgent(ctx, kibana.GetAgentRequest{ID: agentID})
+		return err == nil && agent.UpgradeDetails == nil
+	},
+		20*time.Minute, time.Second,
```
michalpristas, line 468: the reason for such a long wait is that we need to wait for Fleet to start another round of checkin to recalculate state.

michalpristas: also i need to add check for upgraded at/upgrade started fields but for that i need change in elastic-agent-libs. so followup in case i don't make the change in libs in time

blakerouse: Yeah that is a long time, any ideas on how to shorten it? Reduce the check-in window?

blakerouse (commented):
Overall this code looks good. Changes are clear and easy to understand. Will hold off on +1 until passing CI. Also, see inline comment. Maybe that could be reduced?

testing/integration/upgrade_fleet_test.go
cmacknz: We had the same problem in https://github.com/elastic/elastic-agent/pull/7190 where all the timeouts got bumped to 5 minutes.

Shortening the checkin window works and I suspect is the fastest path. I like the idea of being able to trigger a checkin to happen and return on demand but that's a bit more work. Any action triggers the long poll to return so stuff like changing the log level triggers this in a hacky way, maybe a NOOP action would be the tidiest way to get an on-demand checkin response without changing too much.

cmacknz: Also, we should evaluate if we can test this fix without an integration test. We already have really good upgrade integration test coverage so adding more isn't strictly necessary.

Would calling Upgrade() with a mock acker implementation be enough to validate this instead?

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
```diff
@@ -125,8 +125,7 @@ func (h *Upgrade) getAsyncContext(ctx context.Context, action fleetapi.Action, a
 		h.log.Errorf("invalid type, expected ActionUpgrade and received %T", action)
 		return nil, false
 	}
-	if (upgradeAction.Data.Version == bkgAction.Data.Version) &&
-		(upgradeAction.Data.SourceURI == bkgAction.Data.SourceURI) {
```
cmacknz, line 129: Is removing the source URI check actually correct? In air gapped if someone changes the source URI from one that doesn't work to one that does, the source URI changing means it is not the same upgrade action anymore.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -585,7 +592,7 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 		if errors.Is(err, upgrade.ErrUpgradeSameVersion) {
 			// Set upgrade state to completed so update no longer shows in-progress.
 			det.SetState(details.StateCompleted)
-			return nil
+			return c.upgradeMgr.AckAction(ctx, c.fleetAcker, action)
```
cmacknz, line 595: What timeout and retry policy does this have? Is their one? 

If Fleet Server is unavailable for an hour and all acks are failing what is the behavior of this call?

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
michalpristas: yep wrong replace

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: i went with simple case with no retries for readability and that the probability with immediate ack is low. 
it can still happen though. i refactored code to pass things around from the top but it don't like it that much, it makes this a mess

testing/integration/upgrade_fleet_test.go
michalpristas: remove Integ test and added unit test for coordinator and upgrade handler

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
cmacknz: Not necessarily, I don't think Fleet guarantees actions with the same content have the same ID if they are triggered separately in the UI.

This does call out there are three cases to handle properly:
1. Getting the exact same action twice, detectable by the action ID.
2. Getting different action IDs with the same content. For example a user spamming the upgrade agent button or a retry from a the upgrade agent Fleet API.
3. Getting an upgrade action for an upgrade that has no work to do. Having a single action in the queue telling an 8.16.3 agent it should upgrade to 8.16.3.

The 3rd case is the one I was focusing on in https://github.com/elastic/elastic-agent/issues/6186. InfoSec seems to just bulk upgrade all agents to 8.16.3 until they all are upgraded so we can run into all 3 of these cases as they do that. That is what is bringing these problems on.

The changes I see here now are only addressing case 1, when before it looked like we tried to handle both case 1 and 2 by looking only at the content, and the bug was in case 3 where we actually got into the upgrade handler. In that case we'd hit the code in https://github.com/elastic/elastic-agent/blob/56901edc3f2c6d1fc0c1271fad374cdbb24ed7f2/internal/pkg/agent/application/upgrade/upgrade.go#L215-L218

That returns an error that possibly causes us to not ack the action or ack it with an error.

michalpristas: yes different action with same content can be delivered, and i don;t think this is a problem. upgrade handles them sequentially.
in case of snapshots it is a valid case, it can be 2 different snapshots. 
the thing is if we don't ack all of them agent will still be shown as upgrading.

changes here addresses deduplication of actions that are identical. 
deduplication on a content level is done later when we download a package. we determine that hash is the same as installed, ack and abort. this covers cases 2 and 3. (3 was not acked before)

here we ack on restart. that's valid for normal flow
but in case we run into a case where we receive action that differs in hash 
 - this adds to queue and ack that action in ugprade handler (now)
 - then we receive something that has same content 
 - we compare (old flow) add to queue never kick upgrade, never kick ack
 
the only way how to ack this one is to receive upgrade and not restart in between (queue is not persisted)
also not acking can lead to resending action to agent causing us to grow queue more and more.

michalpristas: to make it more robust and avoid cancelling first action in favor of second one i added one additional handling case.
checking source uri and version. in case these are the same we ack immediately but we don't append action to the queue to avoid queue getting too big in case of spamming.

blakerouse (approved):
This looks good. I believe you are now covering all 3 cases that Craig mentioned, and the code change is clear. Plus green CI!

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
```diff
@@ -91,17 +92,31 @@ func (h *Upgrade) Handle(ctx context.Context, a fleetapi.Action, ack acker.Acker
 // ackActions Acks all the actions in bkgActions, and deletes entries from bkgActions.
 // User is responsible for obtaining and releasing bkgMutex lock
 func (h *Upgrade) ackActions(ctx context.Context, ack acker.Acker) {
+	h.ackLock.Lock()
+	defer h.ackLock.Unlock()
```
pkoutsovasilis, line 96: based on the godoc of `ackAction` I would expect here `ackLock` to be `bkgMutex`. Is `ackLock` intentional here, and if yes, could you please help me understand what concurrent pattern we are trying to prevent here? 🙂

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
```diff
@@ -125,11 +140,22 @@ func (h *Upgrade) getAsyncContext(ctx context.Context, action fleetapi.Action, a
...
-		(upgradeAction.Data.SourceURI == bkgAction.Data.SourceURI) {
+	if upgradeAction.ActionID == bkgAction.ActionID {
 		h.log.Infof("Duplicate upgrade to version %s received",
 			bkgAction.Data.Version)
-		h.bkgActions = append(h.bkgActions, action)
+		return nil, false
+	}
+
+	if upgradeAction.Data.Version == bkgAction.Data.Version &&
+		upgradeAction.Data.SourceURI == bkgAction.Data.SourceURI {
```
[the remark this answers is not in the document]
cmacknz: Actually we already check for version equality before downloading the package in upgrade.go:

https://github.com/elastic/elastic-agent/blob/894ef9385efddcff69cd77ab0da3635c1db31f6a/internal/pkg/agent/application/upgrade/upgrade.go#L198-L203

So that part is already done. I think the ack goroutine here is unnecessary. I am trying to reason through if there is a possible deadlock between `ackLock` and `bkgMutex` and it's simpler if we just don't need `ackLock` at all.

`isSameReleaseVersion` doesn't consider sourceURI like this does, and it could, but it's hard for me to think through a real case where the version wouldn't change. The exception is probably just major.minor.patch-SNAPSHOT where the snapshot ID is different but not captured in the version.

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
cmacknz: I think we can probably get rid of the need for the second lock: https://github.com/elastic/elastic-agent/pull/7635/files#r2027652932

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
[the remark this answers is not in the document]
michalpristas: yes the check before let snapshots pass otherwise it's early termination of processing without downloading a package, then the check in https://github.com/elastic/elastic-agent/blob/894ef9385efddcff69cd77ab0da3635c1db31f6a/internal/pkg/agent/application/upgrade/upgrade.go#L167-L174
takes hash into account to handle SNAPSHOT cases

dup lock was removed

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
```diff
@@ -125,11 +137,22 @@ func (h *Upgrade) getAsyncContext(ctx context.Context, action fleetapi.Action, a
...
+		// not the same action this one needs to be acked
+		go func() {
+			// kick it off and don't block, lock to prevent race with ackActions from finished upgrade
+			h.bkgMutex.Lock()
+			defer h.bkgMutex.Unlock()
+
+			h.ackAction(ctx, ack, upgradeAction, true)
+		}()
```
cmacknz, line 155: Is there an actual reason for this to be async? Using the same lock, this should most of the time execute immediately after `getAsyncContext` returns. If another upgrade action came in, it would then block again in `getAsyncContext` until this completes anyway (or `getAsyncContext` wins and this keeps waiting).

There's a synchronous call to ack actions right after this one in https://github.com/michalpristas/elastic-agent/blob/3229be39c543ea278bebaac99520b476268e19c6/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L164-L165

```
	// Versions must be different, cancel the first upgrade and run the new one
	h.log.Infof("Canceling upgrade to version %s received",
		bkgAction.Data.Version)
	h.bkgCancel()

	// Ack here because we have the lock, and we need to clear out the saved actions
	h.ackActions(ctx, ack)
```

You could also just enqueue this action and then the upgrade handler will deal with detecting that the action has no work to do and ack it as a duplicate.

cmacknz:
Since we are only unit testing this one, which I agree makes the most sense, did you manually trigger duplicate actions via the Fleet API to make sure this actually resolved the problem in the way we expect?

The whole reason for this change is we never manually tested this with a Fleet managed agent so let's make sure we don't do the same thing again.

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
michalpristas: i did it async to allow early exit in favor of actually useful actions.
also uprade is performed in a separate goroutine syncing on return which makes it consistent in my point of view.

sync in an example you shared because you need to flush the queue (cancelled upgrades in favor of new one) so you cannot risk flushing action that sneaked in. 

this line acks just a single action, so risk of acking sneaked in action is non existent

michalpristas:
i did handful of scenarios when manually testing this

cmacknz (approved):

Issue #7529, opened 2025-03-21
Improve package install error handling for parallel tests.


ebeahan:
Continuing thread from https://github.com/elastic/elastic-agent/issues/4883#issuecomment-2732574092. 

A current limitation of using a single ESS stack during testing are conflicts when attempting to install an already installed package.

To avoid needing multiple ESS stacks for CI runs, can we change the behavior when trying to install an existing package. Discussed options:

* Implement some type of retry if Fleet returns a conflict error.
* Preinstall necessary packages in ESS prior and remove step from testing.
* Fleet side is exploring improvements in locking  https://github.com/elastic/kibana/issues/213337.

pkoutsovasilis:
did some digging [here](https://github.com/elastic/kibana/issues/213337#issuecomment-2745206919) 🙂

ebeahan:
Thanks @pkoutsovasilis - also saw your notes from https://github.com/elastic/elastic-agent/issues/7221#issuecomment-2745216784.

> the safest solution for our CI is to pre-install all packages we need.

This seems like the simplest path forward, IIUC.

ycombinator:
If I'm following the discussion so far correctly, I'm thinking we need:
1. a way for tests to declare which packages they need (probably via a new [define.Requirements](https://github.com/elastic/elastic-agent/blob/b364d3d4a760de44fbc691d11e3650fc3198a154/pkg/testing/define/requirements.go#L98) field)
2. the test framework to expose this information to the runner, so
3. it can gather all the packages needed by a group of tests and pre-install them before running the tests.

Does this seem like the right approach to you @pkoutsovasilis @blakerouse @pazone?

pkoutsovasilis:
hey @ycombinator 👋 TBH since the root cause of this issue is in Kibana Fleet, I would split this effort into two parts:

Stop-gap solution: In the ESS provisioning CI step, install all the packages that are gonna be used in the follow-up integration test steps by using just curl cmds. Also write comments in the tests that actually require a package to be installed with their policy that for the moment there is this static pre-installing happening. Thus if a developer bumps the version of such a package will know to go to the curl commands and update it there as well.

Long term fix (we might not even need this if Kibana Fleet fixes this on their end, thus I wouldn't dive into this one directly):
[...]

PR #7651, opened 2025-03-31
Pre-install EPM packages needed by integration tests in Fleet

ycombinator:
## What does this PR do?

This PR installs EPM packages needed by integration tests into Fleet, prior to the tests being executed.

## Why is it important?
Several integration tests rely on EPM packages being installed in Fleet.  If two tests depending on the same package run concurrently, one test may try to install the package while the other has already started the process.  This results in conflicts and flaky tests. 

This PR provides a stop-gap solution in which the test runner installs the packages right after the stack has been provisioned and before any integration tests are executed against that stack.

## Checklist

- ~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~

## Disruptive User Impact

None, as this is impacts integration testing, not the built product.

## Related issues

Relates to #7529


ycombinator:
This PS is ready for review now.  

I've implemented pre-installing of packages (using `endpoint` and `system` for now but it's easy to add more) for the CI integration testing path.  

Do we want to also implement the same for the local testing path, where the stack provisioning is done in the [runner](https://github.com/elastic/elastic-agent/blob/cf16186003b7146e8a88de4c3268a9658d94b660/pkg/testing/runner/runner.go#L610)?

testing/integration/testdata/preinstalled_packages.json
```diff
@@ -0,0 +1,6 @@
+{
+    "packages": [
+        { "name": "endpoint", "version": "9.0.0" },
+        { "name": "system", "version": "1.67.3" }
```
pkoutsovasilis, line 4: we also need here the following

```json
{ "name": "auditd_manager", "version": "1.18.3" },
{ "name": "fleet_server", "version": "1.5.0" },
{ "name": "network_traffic", "version": "1.32.1" },
{ "name": "osquery_manager", "version": "1.16.0" },
{ "name": "apache", "version": "1.17.0" },
{ "name": "log", "version": "2.3.0" }
```

Also the `endpoint` package in use in our code is at `8.11.0` and the `system` is at `1.53.0`. So definitely we need to align all of these versions to be the same in our code. 

some of the above are statically defined in json files under the path `testing/integration/testdata` and `testing/integration/`. Some others are part of the go code and also we need to align the versions some 

and

ycombinator: > Also the endpoint package in use in our code is at 8.11.0 and the system is at 1.53.0. So definitely we need to align all of these versions to be the same in our code.

The reason I bumped the versions of these two packages is that the versions currently in our code are not compatible with 9.1.0 stacks, so CI in this PR was failing.

Will add the other packages to the list, thanks.

ycombinator: > some of the above are statically defined in json files under the path testing/integration/testdata and testing/integration/. Some others are part of the go code

I have aligned the versions in code using the `preinstalledPackages` map introduced in this PR.  However, I don't have a good approach to aligning the ones defined in JSON files unless we want to templatize them?  Happy to do that, but I'm not sure if that's overkill for a stop-gap solution? WDYT?

ycombinator: > that's interesting 🤔 then how we are not seeing these failures when the integration tests are running?! Are we using force under the hood?

Yes, I am wondering the same thing as well.  

In fact, this seems to be the case with some of the packages I just added to the list as well:

```
{"name":"auditd_manager","statusCode":400,"error":"auditd_manager-1.18.3 is out-of-date and cannot be installed or updated"}
{"name":"network_traffic","statusCode":400,"error":"network_traffic-1.32.1 is out-of-date and cannot be installed or updated"}
{"name":"apache","statusCode":400,"error":"apache-1.17.0 is out-of-date and cannot be installed or updated"}
{"name":"log","statusCode":400,"error":"log-2.3.0 is out-of-date and cannot be installed or updated"}
```

I'll bump their versions in this PR and try to figure out why this is not currently a problem on the `main` branch.

pkoutsovasilis: ok I just tested it and this extra param in curl does the trick `-d '{ "force": true }'` so please stay with the same package versions as we had them for now but we should definitely create an issue of properly handling these versions, wdyt? 🙂

ycombinator: Since we are testing against 9.1.0 stacks in `main` and 9.0.0 stacks in the `9.0` branch, I think we should use the versions of packages that are intended to work with 9.x.  But when we backport this PR to `8.*` branches, we can add the `force` parameter to the curl request.  WDYT? 🙂

pkoutsovasilis: oops I didn't test this in time and you went the extra mile to bump the versions everywhere. Then yes what you say makes sense. Waiting for the CI to be 🟢 as this looks good to me 🙂

ycombinator:
@pkoutsovasilis Thanks for your review.  There is one [pending question](https://github.com/elastic/elastic-agent/pull/7651#issuecomment-2772995257) I'd like your or other reviewers' opinion on:

> I've implemented pre-installing of packages... for the CI integration testing path... Do we want to also implement the same for the local testing path, where the stack provisioning is done in the [runner](https://github.com/elastic/elastic-agent/blob/cf16186003b7146e8a88de4c3268a9658d94b660/pkg/testing/runner/runner.go#L610)?

.buildkite/scripts/steps/fleet.sh
```diff
@@ -0,0 +1,65 @@
...
+  retcode=$?
+
+  return $retcode
+}
+
+# install_fleet_packages installs the EPM packages defined in the given file into Fleet. The file
+# must define the packages as the request body of the Bulk install packages Fleet API:
+# https://www.elastic.co/docs/api/doc/kibana/v8/operation/operation-bulk-install-packages
+function install_fleet_packages() {
+  local install_api_request_file=$1
```
v1v, line 21: Do you think we should avoid any 'verbose output' by forcing `set +x`?

```suggestion
  set +x
  local install_api_request_file=$1
```

v1v (dismissed):

testing/integration/logs_ingestion_test.go
```diff
@@ -498,7 +499,7 @@ var policyJSON = `
   "policy_id": "{{.PolicyID}}",
   "package": {
     "name": "log",
-    "version": "2.3.0"
+    "version": {{.LogPackageVersion}}"
```
pkoutsovasilis, line 502:
```suggestion
    "version": "{{.LogPackageVersion}}"
```

pkoutsovasilis:
> @pkoutsovasilis Thanks for your review. There is one [pending question](https://github.com/elastic/elastic-agent/pull/7651#issuecomment-2772995257) I'd like your or other reviewers' opinion on:
> 
> > I've implemented pre-installing of packages... for the CI integration testing path... Do we want to also implement the same for the local testing path, where the stack provisioning is done in the [runner](https://github.com/elastic/elastic-agent/blob/cf16186003b7146e8a88de4c3268a9658d94b660/pkg/testing/runner/runner.go#L610)?

Sorry I missed this question @ycombinator. IMO we aren't required to do this for the local testing part; the issue you are tackling is observed only for the integration tests that run with Buildkite runners and test groups are in parallel. Also we have upcoming work to migrate other tests to buildkite runners that will use the same ESS stack provisioned from the step you added the preinstallation, so things will converge 🙂

.buildkite/scripts/steps/fleet.sh
ycombinator: My preference would be to keep the verbose output for now, while this change is still somewhat new, in case it becomes helpful for debugging any issues.  Once things start running smoothly, we can add `set +x`.

v1v (approved):

pkoutsovasilis (approved):
LGTM

PR #7661, opened 2025-04-01
Update golangci-lint to `v2.0.2`

belimawr:
## What does this PR do?

This PR updates golangci-lint to `v2.0.2`. The configuration file, `.golangci.yml` and the GitHub action are all updated to use the same golangci-lint version.

## Why is it important?

It updates our lint version

## Checklist

## Related issues
- https://github.com/elastic/elastic-agent-libs/pull/299

## Questions to ask yourself

- ...


.github/workflows/golangci-lint.yml
```diff
@@ -25,10 +25,12 @@ jobs:
           go-version-file: .go-version
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@2226d7cb06a077cd73e56eedd38eecad18e5d837  # v6.5.0
+        env:
+          CGO_ENABLED: 1
+        uses: golangci/golangci-lint-action@v7
```
v1v, line 30: Please use the pinned version, as explained at https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

```suggestion
        uses: golangci/golangci-lint-action@1481404843c368bc19ca9406f87d6e0fc97bdcfd # v7.0.0
```

belimawr: Done, I also added the link there.

v1v: Thanks 👍

v1v (approved):

pkoutsovasilis (commented):
@belimawr ideally I would have expected the golangci-lint v2 to be a different PR and from bumping the agent libs dependency. Please correct me if I am missing something but this should be feasible. With that aside, I am fine with keeping them both in this PR, but please update your PR description to reflect all the changes, ty

belimawr:
> @belimawr ideally I would have expected the golangci-lint v2 to be a different PR and from bumping the agent libs dependency. Please correct me if I am missing something but this should be feasible. With that aside, I am fine with keeping them both in this PR, but please update your PR description to reflect all the changes, ty

The reason I'm updating elastic-agent-libs is to get the `v2.0.2` golangci-lit, that is versioned there, because Elastic-Agent was not using it on the GitHub Action, I updated both.

It's better to have both updated in one go, so the configuration, the golangci-lint used by mage and the GitHub action are all in sync.

I'll update the PR title and description.

.github/workflows/golangci-lint.yml
```diff
@@ -25,10 +25,14 @@ jobs:
           go-version-file: .go-version
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@2226d7cb06a077cd73e56eedd38eecad18e5d837  # v6.5.0
+        env:
+          CGO_ENABLED: 1
```
andrzej-stencel, line 29: Why is `CGO_ENABLED=1` added with this change?

pkoutsovasilis:
@belimawr could you please take care of the conflicts? 🙂

.github/workflows/golangci-lint.yml
belimawr: Maybe because I copied it from Beats :thinking:. I'll remove it.

andrzej-stencel (commented):
The `main` branch already is on elastic-agent-libs `v0.19.0`, so this PR is now only about upgrading linter, right? Can you update the PR title and description?

kaanyalti (approved):

PR #7062, opened 2025-02-27
Enrich otel logs to match beat process logs 

khushijain21:
## What does this PR do?
This PR adds `component.*` [fields](https://github.com/elastic/elastic-agent/blob/a4325c8f5e8cfc1dae44deb6402c875bb90a718c/pkg/component/runtime/command.go#L545-L555) for components run by the OTelManager. For more reasoning check https://github.com/elastic/ingest-dev/issues/5139

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## How to test this PR locally
Logs from normal mode

```
{"log.level":"info","@timestamp":"2025-02-24T10:59:43.243Z","message":"X","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"filestream-monitoring","type":"filestream"},"log":{"source":"filestream-monitoring"},"log.logger":"centralmgmt.V2-manager","log.origin":{"file.line":505,"file.name":"management/managerV2.go","function":"github.com/elastic/beats/v7/x-pack/libbeat/management.(*BeatV2Manager).unitListen"},"service.name":"filebeat","ecs.version":"1.6.0","ecs.version":"1.6.0"}
```

Logs from components run by Otel Manager
```
{"log.level":"error","@timestamp":"2025-02-27T17:24:15.152+0530","log.logger":"modules","message"X","log":{"source":"elastic-agent"},"kind":"receiver","name":"filebeatreceiver/_agent-component/filestream-default","data_type":"logs","log":{"source":"filestream-default"},"component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"filestream-default"},"service.name":"Khushis-MacBook-Pro.local","ecs.version":"1.6.0","ecs.version":"1.6.0"}
```

Some Differences:
-  We have two `log.source` fields in the latter. It is not misleading, but two fields with same key can be confusing to users. 

## Related issues

[...]

PR #7686, opened 2025-04-03
[Chore] Update otel components to v0.127.0/v0.121.0

khushijain21:
## What does this PR do?
It updates all otel components except `prometheus-receiver` due to its dependency on higher prometheus version (which conflicts with beats code). We should update `prometheus-receiver` once the following issue is resolved. See https://github.com/elastic/beats/issues/42916

This is a temporary solution to unblock us with otel related work on main branch. 

This update was performed using following command
```
.ci/scripts/otel-update.sh v0.121.0 v1.27.0 v0.121.0 
```

and manually degraded the version of `prometheus-receiver`.
Skipping changelog for now.

## Checklist

```
go build .
sudo ./elastic-agent otel
```

## Related issues

- Supersedes https://github.com/elastic/beats/pull/42748, https://github.com/elastic/elastic-agent/pull/7663


testing/integration/beat_receivers_test.go
```diff
@@ -370,7 +370,7 @@ exporters:
       enabled: true
     batcher:
       enabled: true
-      flush_timeout: 0s
+      flush_timeout: 0.5s
```
mauri870, line 373: I'm curious as to why this is needed. This started breaking in the PRs that upgraded to 0.120.0, but from the elasticsearchexporter changelog I couldn't spot any breaking changes.

pkoutsovasilis: I agree this look suspicious 🙂

khushijain21: Looks like it is coming from here https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queue_sender.go#L137


```
    OTel manager failed: invalid configuration: exporters::elasticsearch/log::batcher: `flush_timeout` must be greater than zero
    exporters::elasticsearch/log::batcher::config: `flush_timeout` must be greater than zero
```

pkoutsovasilis: I believe you are right @khushijain21 , any reason not to go with the new default of 0.2 seconds ([ref](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queue_sender.go#L158))? To be honest, I am not saying that 0.2s is better than 0.5s, I am just wondering if we want to be fully aligned here with the new defaults 🙂

khushijain21: `elasticsearchexporter` sets it to `30s` though. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticsearchexporter#batching

I am okay setting it to `0.2s` either way

mauri870 (approved):
I tested this manually: elastic-agent compiles, the collector appears to work as before, and all tests pass. I think it's fine to merge this.

One concern is the prometheusreceiver. The collector compiles, but does it actually work with the different version of the receiver? I suppose the best way to be sure is to create a configuration using prometheusreceiver and confirm that it works.

testing/integration/beat_receivers_test.go
pkoutsovasilis: you are @mauri870 are the beats as otel representantives @khushijain21 in this PR, so if you are okay with 0.5s just resolve the comment 🙂

khushijain21:
> One concern is the prometheusreceiver. The collector compiles, but does it actually work with the different version of the receiver? I suppose the best way to be sure is to create a configuration using prometheusreceiver and confirm that it works.

Well, one reason I wouldn't want to test this is because this is a temporary solution (to unblock us). We should soon be able to ramp the `prometheus-receiver` version and any concerns regarding this can rest

pkoutsovasilis:
@khushijain21 @mauri870 what backports we need here?

mauri870:
> @khushijain21 @mauri870 what backports we need here?

Based on the [original PR](https://github.com/elastic/beats/pull/42748), only 8.x which I have already added.

pkoutsovasilis:
Although this PR has a lot of dependency bumps, I did a quick cross-check and they check out. However, I will wait for an approval from @cmacknz, or anybody else from control-plane team that feels more confident about approving this, as there are other PRs ([example](https://github.com/elastic/elastic-agent/pull/7062#issuecomment-2773393799)) that should have a backport to `9.0` but it was missed 🙂

khushijain21:
I've created an issue to upgrade `prometheus-receiver` when prometheus version does not conflict with beats code anymore - here https://github.com/elastic/elastic-agent/issues/7712

go.mod
```diff
@@ -34,36 +34,36 @@ require (
...
+	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/loadbalancingexporter v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckextension v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/extension/observer/k8sobserver v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/extension/pprofextension v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/status v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/jaegerreceiver v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/jmxreceiver v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver v0.121.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/nginxreceiver v0.121.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.119.0
```
cmacknz, line 56: Use the replace into our fork that is in use in 9.0 and 8.18 so everything matches up. This will make sure we have bug fixes from v0.120 here and won't accidentally go backwards in 8.19 and 9.1 depending on the timing of the fixes in beats.

https://github.com/elastic/elastic-agent/blob/b375b9c234ad17259722c84014eba0128a3cb59c/go.mod#L664-L665

khushijain21: Just one question, should this PR add the changelog entry or defer it to the PR that will bump `prometheus-receiver`? 

Since we are not updating every otel component yet.

cmacknz:
> Skipping changelog for now.

I think there should be a changelog for the bump to otel 0.121.0. There are tons of changes included in upstream releases so we shouldn't silently do these version bumps.

cmacknz (approved):
Thanks for the updates!

jlind23 (approved):

PR #7702, opened 2025-04-03
Enable buildkite step for unit testing with GODEBUG=fips140=only

michel-laterman:
## What does this PR do?

Add a buildkite step to enable unit tests with `GODEBUG=fips140=only` set.

## Why is it important?

Testing with `fips140=only` will help to ensure that we do not mistakenly use non-compliant algorithms in the future. 

## Checklist

## Disruptive User Impact

N/A

`FIPS=true mage test:fipsOnlyUnit`


internal/pkg/core/authority/ca.go
```diff
@@ -96,6 +100,23 @@ func NewCA() (*CertificateAuthority, error) {
 	}, nil
 }
 
+func generateSubjectKeyID(pub crypto.PublicKey) []byte {
```
michel-laterman, line 103: Generating our own subject key like this avoids the default SHA1 id
Copied from https://github.com/elastic/elastic-agent-libs/pull/263

internal/pkg/agent/application/endpoint_component_modifier_test.go
```diff
@@ -29,6 +30,7 @@ import (
 )
 
 func TestEndpointComponentModifier(t *testing.T) {
+	testutils.SkipIfFIPSOnly(t, "generating an encryped priave key for failure testing results in a MD5 violation.")
```
ycombinator, line 33: Typos.
```suggestion
	testutils.SkipIfFIPSOnly(t, "generating an encrypted private key for failure testing results in a MD5 violation.")
```

.buildkite/scripts/steps/fips140-only-tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+source .buildkite/scripts/common.sh
+set +euo pipefail
+
+echo "--- fips140=only Unit tests"
+RACE_DETECTOR=true TEST_COVERAGE=true mage test:fipsOnlyUnit
```
ycombinator, line 6: AFAICT, this file is identical to `unit-tests.sh`, with the only difference being the mage target.   Similarly, the `func (Test) FIPSOnlyUnit` method is almost the same as the `func (Test) Unit` method, with the exception that the former sets the `GODEBUG` env var.  Perhaps we could just keep the `Unit` method and have it pass through the `GODEBUG` env var if set?  That would allow us to also reuse the `unit-tests.sh` by setting `GODEBUG=fips140=only` in this file here and then just sourcing in `unit-tests.sh`.  WDYT?

.buildkite/pipeline.yml
```diff
@@ -53,6 +53,25 @@ steps:
           manual:
             allowed: true
 
+      - label: "Unit tests - fips140=only Ubuntu 22.04"
```
ycombinator, line 56: In the step above this one, we run unit tests with `FIPS=true`.  Is there some reason not to run that step with `GODEBUG=fips140=only` instead of introducing this new step? In other words, when we run unit tests with `FIPS=true` wouldn't we want any calls to FIPS-noncompliant crypto functions to fail, so we can identify and fix those usages until all the tests start passing?

internal/pkg/testutils/testutils.go
```diff
@@ -44,3 +46,17 @@ func NewErrorLogger(t *testing.T) *logger.Logger {
...
 }
+
+// SkipIfFIPSOnly will mark the passed test as skipped if GODEBUG=fips140=only is detected.
+// If GODBUG=fips140=on, go may call non-compliant algorithms and the test does not need to be skipped.
+func SkipIfFIPSOnly(t *testing.T, msg string) {
+	// NOTE: This only checks env var; at the time of writing fips140 can only be set via env
+	// other GODEBUG settings can be set via embedded comments or in go.mod, we may need to account for this in the future.
+	ss := strings.Split(os.Getenv("GODEBUG"), ",")
+	for _, s := range ss {
+		if s == "fips140=only" {
```
swiatekm, line 57: This should probably be a `string.Contains` check instead for robustness. Other options could be set as well.

swiatekm (dismissed):

swiatekm (dismissed):

swiatekm (approved):

kaanyalti (approved):

internal/pkg/core/authority/ca.go
ycombinator: It seems a bit risky to me to copy the implementation over, as implementations could diverge between here and elastic-agent-libs.  Could we not export this function in elastic-agent-libs and call it from here instead?

swiatekm (approved):

v1v (dismissed):
LGTM - just reviewed the BK pipeline

internal/pkg/core/authority/ca.go
michel-laterman: Most of the time this function is just used in test code

kaanyalti (approved):

v1v (approved):
LGTM - I only reviewed the BK pipeline

PR #7734, opened 2025-04-07
Refactor self-monitoring metrics config generation

swiatekm:
## What does this PR do?

Refactor how metrics input configuration is generated for agent self-monitoring. This was primarily focused on readability. Assembling the configuration involved handling large maps of parameters, which are quite unwieldy to deal with in Go code.

This PR covers the metrics part of #7600. The logs part will be covered in a follow-up.

## Why is it important?

## Checklist

## Related issues

- Split out from #7600.

### DIFF
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -627,8 +627,6 @@ func (b *BeatsMonitor) injectMetricsInput(
 		componentListWithMonitoring[k] = v
 	}
 
-	// beatStreams and streams MUST be []interface{} even if in reality they are []map[string]interface{}:
-	// if those are declared as slices of maps the message "proto: invalid type: []map[string]interface{}" will pop up
 	beatsStreams := b.getBeatsStreams(componentListWithMonitoring, failureThreshold, metricsCollectionIntervalString)
 	httpStreams := b.getHttpStreams(componentListWithMonitoring, failureThreshold, metricsCollectionIntervalString)
 
@@ -675,6 +673,7 @@ func (b *BeatsMonitor) injectMetricsInput(
 }
 
 // getHttpStreams returns stream definitions for http/metrics inputs.
+// Note: The return type must be []any due to protobuf serialization quirks.
 func (b *BeatsMonitor) getHttpStreams(
 	componentIDToBinary map[string]string,
 	failureThreshold *uint,
@@ -715,7 +714,7 @@ func (b *BeatsMonitor) getHttpStreams(
 		}
 
 		endpoints := []interface{}{prefixedEndpoint(utils.SocketURLWithFallback(unit, paths.TempDir()))}
-		name := strings.ReplaceAll(strings.ReplaceAll(binaryName, "-", "_"), "/", "_") // conform with index naming policy
+		name := sanitizeName(binaryName)
 
 		httpStream := map[string]interface{}{
 			idKey: fmt.Sprintf("%s-%s-1", monitoringMetricsUnitID, name),
@@ -769,6 +768,7 @@ func (b *BeatsMonitor) getHttpStreams(
 }
 
 // getBeatsStreams returns stream definitions for beats inputs.
+// Note: The return type must be []any due to protobuf serialization quirks.
 func (b *BeatsMonitor) getBeatsStreams(
 	componentIDToBinary map[string]string,
 	failureThreshold *uint,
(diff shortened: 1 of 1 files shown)


internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -616,249 +616,137 @@ func (b *BeatsMonitor) injectMetricsInput(
...
+	componentListWithMonitoring := map[string]string{
+		fmt.Sprintf("beat/%s", monitoringMetricsUnitID): "metricbeat",
+		fmt.Sprintf("http/%s", monitoringMetricsUnitID): "metricbeat",
+		monitoringFilesUnitsID:                          "filebeat",
+	}
+	for k, v := range componentIDToBinary {
+		componentListWithMonitoring[k] = v
+	}
+
 	// beatStreams and streams MUST be []interface{} even if in reality they are []map[string]interface{}:
```
ycombinator, line 630: Probably should update this comment to match the code changes below.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -616,249 +616,137 @@ func (b *BeatsMonitor) injectMetricsInput(
...
 	// if those are declared as slices of maps the message "proto: invalid type: []map[string]interface{}" will pop up
-	beatsStreams := make([]interface{}, 0, len(componentIDToBinary))
-	streams := []interface{}{
+	beatsStreams := b.getBeatsStreams(componentListWithMonitoring, failureThreshold, metricsCollectionIntervalString)
+	httpStreams := b.getHttpStreams(componentListWithMonitoring, failureThreshold, metricsCollectionIntervalString)
+
+	inputs := []interface{}{
 		map[string]interface{}{
-			idKey: fmt.Sprintf("%s-agent", monitoringMetricsUnitID),
+			idKey:        fmt.Sprintf("%s-beats", monitoringMetricsUnitID),
```
ycombinator, line 637: Was this change from `%s-agent` to `%s-beats` intentional?

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -616,249 +616,137 @@ func (b *BeatsMonitor) injectMetricsInput(
...
+			"name":       fmt.Sprintf("%s-beats", monitoringMetricsUnitID),
+			"type":       "beat/metrics",
```
ycombinator, line 639: Was this change from `%s-agent` to `%s-beats` intentional?  Note that, further below, when setting up the "http/metrics" type input, it's still `%s-agent` there.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -867,203 +755,258 @@ func (b *BeatsMonitor) injectMetricsInput(
...
-			"data_stream": map[string]interface{}{
-				"namespace": monitoringNamespace,
+			"drop_fields": map[string]interface{}{
+				"fields": []interface{}{
+					"http",
+				},
+				"ignore_missing": true,
 			},
-			"streams": streams,
 		},
```
ycombinator, line 910: Nit: I realize these two processors are not reused elsewhere, so that's why you probably didn't extract their creation into separate functions like you did for the other processors.  I do really like how readable the processor list has become with the function calls, since the function names now tell us the _purpose_ of the processor.  So I'd personally be +1 to extracting these two two processor definitions into their own functions as well.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -867,203 +755,258 @@ func (b *BeatsMonitor) injectMetricsInput(
...
 		}
-		for _, s := range beatsStreams {
-			if streamMap, ok := s.(map[string]interface{}); ok {
-				streamMap[failureThresholdKey] = *failureThreshold
-			} else {
-				return fmt.Errorf("unable to set %s: %d in monitoring stream %q: unexpected type %T", failureThresholdKey, *failureThreshold, s, s)
-			}
+
+		endpoints := []interface{}{prefixedEndpoint(utils.SocketURLWithFallback(unit, paths.TempDir()))}
+		name := strings.ReplaceAll(strings.ReplaceAll(binaryName, "-", "_"), "/", "_") // conform with index naming policy
```
ycombinator, line 789: I think you could call `sanitizeName` here?

ycombinator (commented):
Built Agent from this PR and checked that metrics in Fleet UI are unaffected. 👍 

Left a few minor code comments.  Love how much more readable this code is now. 👍 👍

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: I'll move this comment to the functions themselves, since you don't need to know the slice type here.

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: Sure, can do.

swiatekm: Turns out they were reused, I just missed it. Good catch.

PR #6708, opened 2025-02-04
mergify: support backport-active-8, backport-active-9 and backport-active-all 

v1v:
## What does this PR do?

Enable `backport-active-all`, `backport-active-8` and `backport-active-9` GitHub labels in Mergify to backport to a set of active branches in addition to the existing automation with the `backport-<major>.<minor>` labels. This can help with the current 6 active branches, which we have today.

See https://docs.mergify.com/workflow/actions/backport/#parameters

image

I can find `regexes` too:

image

But I'm not sure whether that can solve our current release model easily

## Why is it important?

> Maybe this is solved already, but is there a way we can get an automatically kept up to date “backport-active-minors” label that auto backports to the set of minor releases that are still supported? I’d want this in Beats, Elastic Agent, and Fleet Server if it already exists. I’ve noticed an increase in forgotten backports because there are more releases to keep track of. If we can abstract this it’d make life easier, I think Kibana does something like this

## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #7735, opened 2025-04-07
Update way of backporting

ev1yehor:
## What does this PR do?

Adds github workflow to automatically create comment for mergify to create backport PRs based labels.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


.github/workflows/backport-active.yml
```diff
@@ -0,0 +1,24 @@
...
+  contents: read
+
+jobs:
+  backport:
+    # Only run if the PR was merged (not just closed) and has one of the backport labels
+    if: |
+      github.event.pull_request.merged == true && 
+      (contains(github.event.pull_request.labels.*.name, 'backport-active-all') || 
+       contains(github.event.pull_request.labels.*.name, 'backport-active-9') || 
+       contains(github.event.pull_request.labels.*.name, 'backport-active-8'))
```
pkoutsovasilis, line 20: not a github actions expert but what if I want to backport something only to 8.x (label `backport-8.x`), and 8.18 (label `backport-8.18`), would that be supported? with the previous way it was

v1v, line 20: @ev1yehor , could you simplify the condition to search for any labels that start with `backport-active`? Currently, we have `backport-active-8`, `backport-active-9`, and `backport-active-all` in two locations: here and at https://github.com/elastic/oblt-actions/blob/df86975aed33be75b475e216ba621151dfc9b8b4/github/backport-active/backport_script.py#L86-L97.

If we later decide to add support for `backport-active-10`, we will need to modify both locations.

.github/workflows/backport-active.yml
```diff
@@ -0,0 +1,24 @@
+name: Automatic Backport
```
v1v, line 1: I prefer a more clear name, something like `backport to active branches` or something along those lines

.github/workflows/backport-active.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  pull_request:
```
v1v, line 4: `pull_request` does not grant permissions with `write`, hence you need to use `pull_request_target` instead.

I'm not a big fan of using `pull_request_target`, but we do not use `checkout` or anything along those lines, hence we are safe to use it

```suggestion
  pull_request_target:
```

.github/workflows/backport-active.yml
v1v: > so we deliberately invest in two different mechanisms to manage different backport labels? 🤔

No, we use `mergify` in both cases. However, there is no support for dynamism in the `.mergify.yml` file; hence, we either change the release scripts in the `elastic/ingest-dev` to update the `.mergify.yml` or use an alternative that avoid the complexity per consumer (we support the `active-backport-*` in different repositories such as `logstash`, `apm-server`, `fleet-server`, `elastic-agent`, `beats`, `observability-docs`).

https://github.com/elastic/elastic-agent/pull/6708/files#diff-6a30e47dd51449847422017b16878889b848f6602ac56d5a2e67ce8a19afade2R418-R426 is the section that will require to be changed if we use `mergify` with the `rules` in `.mergify.yml`. 

I think @ev1yehor , added a link to the private GH issue, you can search for `Update branches when FF for the new GitHub backport labels` and you will find further details about the different alternatives

v1v: > hence, we either change the release scripts in the elastic/ingest-dev to update the .mergify.yml

For the APM Server, Logstash or observability-docs, that will be a different script

ev1yehor: Updated.

.github/workflows/backport-active.yml
ev1yehor: Updated.

.github/workflows/backport-active.yml
```diff
@@ -0,0 +1,22 @@
+name: Backport to active branches
+
+on:
+  pull_request_target:
+    types: [closed]
+    branches:
+      - main  # Only run for PRs targeting the main branch
```
v1v, line 7: **nit**: we don't need that comment, it does not say the rationale, but what

```suggestion
      - main
```

v1v (approved):

pkoutsovasilis (dismissed):

v1v (approved):

pkoutsovasilis (approved):

PR #7738, opened 2025-04-07
Correctly handle sending signal to child process

belimawr:
## What does this PR do?

Calling proc.Kill causes the process to exit immediately, which
prevents components from executing some clean up functions.

To gracefully terminate the components, on Linux and MacOS SIGTERM
is sent.

On Windows signals are sent to process groups, child process are created
in the same group as the parent, so sending a signal to a child process
causes the Elastic-Agent to receive the signal as well. Therefore child
process need to be created in their own process group.

To do so, the child process needs to be created setting the creation flag
CREATE_NEW_PROCESS_GROUP, which has the side effect of disbling
the CTRL+C signal.

Then to correctly terminate the child process `windows.GenerateConsoleCtrlEvent`
is called to send CTRL+BREAK signal that is translated by go Go runtime into a
`syscall.SIGINT`, which is already correctly handled by Beats.

## Why is it important?
It correctly signals components to exit, allowing them to run their cleanup.

## Checklist

~~## Disruptive User Impact~~

```
mage build:TestBinaries
cd pkg/core/process
go test .
```

## Related issues
- Partially fixes https://github.com/elastic/elastic-agent/issues/6875
- Relates to https://github.com/elastic/elastic-agent/issues/7756

## Questions to ask yourself

- ...


belimawr:
buildkite test this

blakerouse (approved):
Thanks for working on this. Windows hard-stopping everything is not good, this was really needed! Nice work.

Issue #7381, opened 2025-03-13
Add functional tests for the opentelemetry-kube-stack configuration


swiatekm:
https://github.com/elastic/elastic-agent/pull/5822 added an opentelemetry-kube-stack configuration that Kibana uses to install the EDOT Collector in Kubernetes. We currently only have very basic tests for this functionality, and they only verify if the EDOT containers start successfully.

https://github.com/elastic/elastic-agent/pull/6444 accidentally broke this config in a way these tests didn't catch. The fix is in https://github.com/elastic/elastic-agent/pull/7380, but we should also add more functional tests for the config to avoid similar regressions in the future.

The new tests should:

1. Start an ESS cluster.
1. Install opentelemetry-kube-stack using the checked-in configuration.
1. Install some applications to test autoinstrumentation with.
    -  This should either be otel-demo, or a different set of applications we already use for testing.
1. Assert that expected data reached Elasticsearch.
    - This should check all the data types and sources.
    - It should check both content and metadata.

The main lift here will be point 4. Points 1 and 2 are already done in other tests, and point 3 shouldn't be too onerous.

PR #7754, opened 2025-04-08
Set fullnameOverride and validate Otel kube-stack deployment datastreams are being written (metrics, traces)

rogercoll:
## What does this PR do?

Adds tests to ensure the Helm chart is correctly deployed and data pushed to Elasticsearch:

- Validates metrics collected with the `kubeletstats` receiver are written to Elasticsearch.
- Deploys a Java application to validate auto-instrumentation works properly and the corresponding traces documents are available in Elasticsearch. As an alternative, I tried to install the OpenTelemetry demo Helm Chart and assert its traces, but it is too heavy (big images, e.g. flagd → 1.2GB) for the testing timeouts.

## Why is it important?

Fixes https://github.com/elastic/elastic-agent/issues/7381


deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
```diff
@@ -33,6 +33,7 @@ collectors:
   #  Cluster is a K8s deployment EDOT collector focused on gathering telemetry
   #  at the cluster level (Kubernetes Events and cluster metrics).
   cluster:
+    fullnameOverride: "opentelemetry-kube-stack-cluster"
```
rogercoll, line 36: For reviewers: Explicitly setting the pods/services name to ensure uniqueness across all Helm releases. These changes should not affect deployments that followed our Kibana instructions, indeed, it ensures the correct data flow even if users use a different Helm release name and/or namespace as the values contains some related hard coded values. Examples:

1.  Auto instrumentation daemon collector endpoint:
```yaml
instrumentation:
  name: elastic-instrumentation
  enabled: true # Enable/disable auto-instrumentation.
  exporter:
    endpoint: http://opentelemetry-kube-stack-daemon-collector.opentelemetry-operator-system.svc.cluster.local:4318 # The daemonset OpenTelemetry Collector endpoint where telemetry data will be exported.
```

2. Gateway endpoint in daemon/cluster collectors:
```yaml
        otlp/gateway:
          endpoint: "http://opentelemetry-kube-stack-gateway-collector:4317"
```

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -1109,35 +1111,8 @@ func k8sRenderKustomize(kustomizePath string) ([]byte, error) {
 }
 
 // generateESAPIKey generates an API key for the given Elasticsearch.
-func generateESAPIKey(esClient *elasticsearch.Client, keyName string) (string, error) {
```
rogercoll, line 1112: Removed in favor of `estools.APIKeyResponse`, the `elasticsearch` exporter requires the encoded version.

testing/integration/otel_helm_test.go
```diff
@@ -128,108 +147,6 @@ func TestOtelKubeStackHelm(t *testing.T) {
...
 
-func TestOtelKubeStackHelmEDOTImage(t *testing.T) {
-	info := define.Require(t, define.Requirements{
-		Stack: &define.Stack{},
-		Local: false,
-		Sudo:  false,
-		OS: []define.OS{
-			// only test the basic and the wolfi container with otel
-			{Type: define.Kubernetes, DockerVariant: "elastic-otel-collector"},
-			{Type: define.Kubernetes, DockerVariant: "elastic-otel-collector-wolfi"},
```
rogercoll, line 139: Moved these two cases to `TestOtelKubeStackHelm` to reduce duplicated code. Let me know if they still need to run in a separate test.

testing/integration/kubernetes_agent_standalone_test.go
```diff
@@ -1337,6 +1312,8 @@ type k8sContext struct {
 	esHost string
 	// esAPIKey is the API key of the elasticsearch to use in the test
 	esAPIKey string
+	// esAPIKey is the API key of the elasticsearch to use in the test
```
ChrsMark, line 1315:
```suggestion
	// esEncodedAPIKey is the encoded API key of the elasticsearch to use in the test
``` 

?

testing/integration/otel_helm_test.go
```diff
@@ -277,3 +194,90 @@ func k8sStepCheckRunningPods(podLabelSelector string, expectedPodNumber int, con
 		}, 5*time.Minute, 10*time.Second, fmt.Sprintf("at least %d agent containers should be checked", expectedPodNumber))
 	}
 }
+
+func k8sStepDeployJavaApp(info *define.Info, dsType, dataset, datastreamNamespace string) k8sTestStep {
```
ChrsMark, line 198: Since this does 2 different things, deploying the target object and then checks the ES data-stream, should that be renamed or better split in 2 separate ones? I guess `k8sStepCheckDatastreamsHits` could be reused for the second part right? So I would just keep the deployment of the target in this one and then call from the main flow the  `k8sStepCheckDatastreamsHits`. WDYT?

ChrsMark (commented):
I'm not quite familiar with this code-base, but overall looks good to me with 2 small suggestions.

testing/integration/otel_helm_test.go
rogercoll: good point, wrapped both steps in a single main function https://github.com/elastic/elastic-agent/pull/7754/commits/788eef8c4a36b3ae2f8184c7d50ca0e6d75e7467

ChrsMark (dismissed):
lgtm

deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
swiatekm: So, is this an actual semantic change to the values?

rogercoll: It is in the sense that the k8s collector's resources (e.g. service) names will now be fixed (not using release or namespace name). Maybe it is not a chore anymore and a changelog might be safer in case a user overrode some values configs. Wdyt?

swiatekm: I'd prefer if this PR only had the tests if you don't want to add a changelog entry. This change deserves one imo.

swiatekm: Also, what happens if a user installs the Helm Chart using the previous values, and then upgrades to these?

testing/integration/otel_helm_test.go
```diff
@@ -277,3 +197,83 @@ func k8sStepCheckRunningPods(podLabelSelector string, expectedPodNumber int, con
 		}, 5*time.Minute, 10*time.Second, fmt.Sprintf("at least %d agent containers should be checked", expectedPodNumber))
 	}
 }
+
+func k8sStepDeployJavaApp() k8sTestStep {
+	return func(t *testing.T, ctx context.Context, kCtx k8sContext, namespace string) {
+		objects, err := testK8s.LoadFromYAML(bufio.NewReader(strings.NewReader(fmt.Sprintf(`apiVersion: apps/v1
+kind: Deployment
```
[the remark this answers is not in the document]
rogercoll: Definitely, moved in https://github.com/elastic/elastic-agent/pull/7754/commits/17cef05349b8232ea6c52cac33b6cbc68fddbd56

swiatekm (approved):

rogercoll:
@swiatekm Do you know to which Elasticsearch stack version are this tests sending data into? Given our [compatibility matrix](https://elastic.github.io/opentelemetry/compatibility/collectors.html) it would be great testing 9.x with 8.18/19 clusters. Is the testing infrastructure aligned with the branch version?

ChrsMark (dismissed):

ChrsMark:
@rogercoll @swiatekm I'm enabling auto-merge for this one. Not sure if we should wait for anybody else here.

rogercoll:
> @rogercoll @swiatekm I'm enabling auto-merge for this one. Not sure if we should wait for anybody else here.

Looks like it needs for another review from @elastic/elastic-agent-control-plane :pray:

swiatekm (approved):

ChrsMark (approved):

PR #7760, opened 2025-04-08
Running mage controlProto

ycombinator:
## What does this PR do?

This PR simply runs `mage controlProto` on the codebase.

## Why is it important?

It appears the control protocol `.go` files have not been regenerated in a while.

Also, I want to add a `fips` boolean field to the `VersionResponse` message in the control protocol, which I will do in a separate follow up PR to this one here.


pkoutsovasilis (commented):
@ycombinator thanks for that, I just realised that we should keep these fresh. I believe the CI failures in the check-ci step is because the generated files are getting alternated by `devtools.GoImports`. So would it make sense to add this `mg.Deps(devtools.GoImports)` in the `ControlProto` mage target? Second question, would it make sense to invoke `mage controlProto` in the check-ci script? 🙂

ycombinator:
> @ycombinator thanks for that, I just realised that we should keep these fresh. I believe the CI failures in the check-ci step is because the generated files are getting alternated by `devtools.GoImports`. So would it make sense to add this `mg.Deps(devtools.GoImports)` in the `ControlProto` mage target? Second question, would it make sense to invoke `mage controlProto` in the check-ci script? 🙂

I think we should keep the `ControlProto` mage target as is, so it has a single purpose: to update the control protocol files. However, to your point about the imports getting screwed up (and also the license headers, BTW), I think it would be better to run `mage update` from now on instead and add `ControlProto`, `devtools.GoImports`, and `devtools.AddLicenseHeaders` as _its_ dependencies.  I've done that in e06e980241b693c8d73322f7a87c4b8f44619d3e.  Let me know what you think.  Thanks!

ycombinator:
> I think we should keep the ControlProto mage target as is, so it has a single purpose: to update the control protocol files. However, to your point about the imports getting screwed up (and also the license headers, BTW), I think it would be better to run mage update from now on instead and add ControlProto, devtools.GoImports, and devtools.AddLicenseHeaders as its dependencies. I've done that in https://github.com/elastic/elastic-agent/commit/e06e980241b693c8d73322f7a87c4b8f44619d3e. Let me know what you think. Thanks!

Okay, this idea isn't going to work, at least not easily in CI, since `mage ControlProto` requires the `protoc` binary to be present:
```
Error: failed to run "protoc --go_out=pkg/control/v2/cproto --go_opt=paths=source_relative --go-grpc_out=pkg/control/v2/cproto --go-grpc_opt=paths=source_relative control_v2.proto: exec: "protoc": executable file not found in $PATH"
```

And it looks like the installation for this binary is OS-dependent 🙁.  So let's go with your suggestion of going the other way: having `ControlProto` depend on `devtools.GoImports` and `devtools.AddLicenseHeaders`.  I've implemented this in 1ae2e3fe90.

pkoutsovasilis (approved):
LGTM 🙂

PR #7690, opened 2025-04-03
Fips packaging

pchila:
## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #7795, opened 2025-04-09
Correctly inject dependency list when packaging using DROP_PATH

pchila:
## What does this PR do?
Fix a bug introduced by https://github.com/elastic/elastic-agent/pull/7690 where dependency list is not correctly passed to `isPlatformIndependentPackage()`.
Without dependency list `isPlatformIndependentPackage()` did not recognize `connectors-9.1.0-SNAPSHOT.zip` as a platform independent package  so it wasn't copied in the `archives` directory.

This in turn would make the `elastic-agent-service` docker image spec fail because of these [lines](https://github.com/elastic/elastic-agent/blob/6abd585e55cbb27500b9821ddd94b80263ca150b/dev-tools/packaging/packages.yml#L582-L584) that expects the connectors zip to exist in the archive directory.

## Why is it important?

To fix unified release for elastic-agent

## Checklist

## Disruptive User Impact
None

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -1333,7 +1333,7 @@ func collectPackageDependencies(platforms []string, packageVersion string, packa
 			}
 		}
 	} else {
-		archivePath = movePackagesToArchive(dropPath, platforms, packageVersion, nil)
+		archivePath = movePackagesToArchive(dropPath, platforms, packageVersion, dependencies)
 	}
 	return archivePath, dropPath
 }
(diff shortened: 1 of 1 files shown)


pchila:
elastic-agent-package pipeline dry-run: https://buildkite.com/elastic/elastic-agent-package/builds/5251

magefile.go
```diff
@@ -1333,7 +1333,7 @@ func collectPackageDependencies(platforms []string, packageVersion string, packa
 			}
 		}
 	} else {
-		archivePath = movePackagesToArchive(dropPath, platforms, packageVersion, nil)
+		archivePath = movePackagesToArchive(dropPath, platforms, packageVersion, dependencies)
```
pkoutsovasilis, line 1336: I see 👀

pkoutsovasilis (commented):
Ok I see what was the issue before and why we hit that only in the DRA manifest building. But I believe that this is addressed now (looking [here](https://buildkite.com/elastic/elastic-agent-package/builds/5251#_)), waiting for the CI to be green 🙂

magefile.go
pchila: Yep :sweat_smile: 
Bad refactoring of `movePackagesToArchive` passed `nil ` instead of the actual list of dependencies

pkoutsovasilis (approved):

PR #7812, opened 2025-04-10
Refactor the monitoring injection interface

swiatekm:
## What does this PR do?

Change the monitoring injection interface to only accept a list of components. Up until now, it also accepted a map of component id to binary name, and it wasn't clear which one took priority or why this was even necessary, given that the map can easily be computed from the list.

Internally, the monitoring config injector now defines a `componentInfo` struct containing all the information about components necessary to generate monitoring configuration for them.

## Why is it important?

Monitoring configuration generation is slightly awkward, in that it gets a list of components to be monitored as an input, but also adds its own configurations which can spawn their own components, which should also be monitored. To deal with this, the interface would additional accept a map of component id to binary name, and add its own components to that map. This was all quite ad-hoc and spread all over the different injection functions.

To make this cleaner and easier to understand, the config injection now operates on `componentInfo` structs, which contain all the necessary information, and it's easy to add them for monitoring components.

This also makes it easier to differentiate between Otel and non-Otel components down the line.

## Checklist

## Related issues

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -94,7 +94,6 @@ type MonitorManager interface {
 	// args:
 	// - the existing config policy
 	// - a list of the expected running components
-	// - a map of component IDs to binary names
 	// - a map of component IDs to the PIDs of the running components.
 	MonitoringConfig(map[string]interface{}, []component.Component, map[string]uint64) (map[string]interface{}, error)
 }


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -96,7 +96,7 @@ type MonitorManager interface {
 	// - a list of the expected running components
 	// - a map of component IDs to binary names
```
khushijain21, line 97: [Nit] This should also be deleted.

blakerouse (dismissed):
It took me a minute to really understand why `componentInfo` struct was added. I finally got it that it was because you will also monitor the component that is doing the monitoring as well.

Overall this looks good, and makes it easier to understand what is going on with the monitoring part of the code.

blakerouse (approved):
Still looks good!

khushijain21 (approved):

PR #7840, opened 2025-04-14
Disable ironbank target when packaging with FIPS=true

pchila:
## What does this PR do?

This PR skips `ironBank` mage target when packaging with `FIPS=true` as `ironBank` does not have a specific packaging in that case, it would only end up duplicating the packaging for no benefit

## Why is it important?

To avoid generating the same package twice

## Checklist

## Disruptive User Impact
None

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/scripts/steps/package.sh
+++ b/.buildkite/scripts/steps/package.sh
@@ -13,13 +13,15 @@ fi
 export AGENT_DROP_PATH=build/elastic-agent-drop
 mkdir -p $AGENT_DROP_PATH
 
-# Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
-mage clean downloadManifest packageUsingDRA
+MAGE_TARGETS=(clean downloadManifest packageUsingDRA)
 if [ "$FIPS" != "true" ]; then
-mage ironbank
+  # Build ironbank only on non-FIPS builds
+  MAGE_TARGETS+=("ironbank")
 fi
-mage fixDRADockerArtifacts
+MAGE_TARGETS+=("fixDRADockerArtifacts")
 
+# Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
+mage "${MAGE_TARGETS[@]}"
 
 echo  "+++ Generate dependencies report"
 BEAT_VERSION_FULL=$(curl -s -XGET "${MANIFEST_URL}" |jq '.version' -r )
--- a/.buildkite/scripts/steps/package.sh
+++ b/.buildkite/scripts/steps/package.sh
@@ -13,13 +13,13 @@ fi
 export AGENT_DROP_PATH=build/elastic-agent-drop
 mkdir -p $AGENT_DROP_PATH
 
-if [ "$FIPS" = "true" ]; then
 # Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
-mage clean downloadManifest packageUsingDRA fixDRADockerArtifacts
-else
-# Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
-mage clean downloadManifest packageUsingDRA ironbank fixDRADockerArtifacts
+mage clean downloadManifest packageUsingDRA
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/steps/package.sh
```diff
@@ -13,8 +13,13 @@ fi
 export AGENT_DROP_PATH=build/elastic-agent-drop
 mkdir -p $AGENT_DROP_PATH
 
+if [ "$FIPS" = "true" ]; then
+# Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
+mage clean downloadManifest packageUsingDRA fixDRADockerArtifacts
+else
 # Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
 mage clean downloadManifest packageUsingDRA ironbank fixDRADockerArtifacts
+fi
```
v1v, line 22: How about we explore using something that is less commonly duplicated?

```suggestion
mageIronbankCommand=
if [ "$FIPS" = "true" ]; then
  mageIronbankCommand=ironbank
fi
echo "Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH"
mage clean downloadManifest packageUsingDRA $mageIronbankCommand fixDRADockerArtifacts
```

I decided to remove the `comment` for an `echo`. The above proposal removes the duplicated mage goals

v1v: We could use an array:

```suggestion
mageOptions=(clean downloadManifest packageUsingDRA fixDRADockerArtifacts)

# Stop building ironbank for FIPS
if [ "$FIPS" != "true" ]; then
    mageOptions+=("ironbank")
fi

# Download the components from the MANIFEST_URL and then package those downloaded into the $AGENT_DROP_PATH
mage "${mageOptions[@]}"
```

v1v (approved):

pchila:
Using `ironbank` target by itself has a curious side effect in the version used in the artifacts file names: https://buildkite.com/elastic/elastic-agent-package/builds/5352#019634ad-89d7-4f28-a221-7e1a984e913b/187-244

The docker context tar.gz use `9.1.0` instead of `9.1.0-SNAPSHOT` so it seems that `ironBank` target relies on other targets in the same mage invocation for some setup.

v1v (approved):

pkoutsovasilis (approved):

pchila:
> > The docker context tar.gz use 9.1.0 instead of 9.1.0-SNAPSHOT so it seems that ironBank target relies on other targets in the same mage invocation for some setup.
> 
> IIUC, the `SNAPSHOT=true` should be available to generate the artifacts with the `-SNAPSHOT` suffix, according to
> 
> https://github.com/elastic/elastic-agent/blob/2f3c36024968a3d1ee2cbec83aaa03dd18da85aa/magefile.go#L2120
> 
> As far as I see, `SNAPSHOT=true` is not set when running the `DRA_WORKFLOW="snapshot"`:
> image
> 
> Maybe the function `FixDRADockerArtifacts` ->
> 
> https://github.com/elastic/elastic-agent/blob/2f3c36024968a3d1ee2cbec83aaa03dd18da85aa/magefile.go#L710
> is the one doing the magic?

I am not sure if some other mage target is initializing the [`mage.Snapshot`](https://github.com/elastic/elastic-agent/blob/abb571448713e2b6f48ef21b4e01e2b79bc4c6a4/dev-tools/mage/settings.go#L143) flag or if the [spec rendering for ironBank](https://github.com/elastic/elastic-agent/blob/3347dc2c7d09452646a55c64519bafcf128864fd/magefile.go#L2118-L2126) is misbehaving here.

This is not a behavior introduced by this PR though, it seems to be a pre-existing bug that should be investigated separately

.buildkite/scripts/steps/package.sh
pchila: @v1v  for the moment I went with the mage target array in the package.sh script (your second suggestion).
We can always refactor in a follow-up PR, feel free to open an issue with your proposal so we can discuss on it.

PR #7841, opened 2025-04-14
Update helm.sh/helm/v3 to v3.17.3

michel-laterman:
## What does this PR do?

Update helm.sh/helm/v3 to v3.17.3


pkoutsovasilis:
Thanks for this PR @michel-laterman , I expect no surprises from the CI 🙂 That said, AFAICT this dependency is solely using for our testing needs and it's not part of the elastic-agent as a product, but still, having up-to-date dependencies that address CVEs is a good thing

pkoutsovasilis (approved):

PR #7854, opened 2025-04-15
Refactor monitoring log config generation

swiatekm:
## What does this PR do?

Refactor how logs input configuration is generated for agent self-monitoring. This was primarily focused on readability. Assembling the configuration involved handling large maps of parameters, which are quite unwieldy to deal with in Go code.

This PR covers the logs part of #7600. The metrics part was covered by #7734.

## Why is it important?

## Checklist

## Related issues

- Split out from #7600.

### DIFF
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -994,6 +994,7 @@ func dropPeriodicMetricsLogsProcessor() map[string]any {
 // routed to the elastic-agent logger get sent to the component-specific dataset.
 func useComponentDatasetProcessors() []any {
 	return []any{
+		// copy original dataset so we can drop the dataset field
 		map[string]any{
 			"copy_fields": map[string]any{
 				"fields": []any{
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -409,215 +409,11 @@ func (b *BeatsMonitor) getComponentInfos(components []component.Component, compo
 
 // injectLogsInput adds logging configs for component monitoring to the `cfg` map
 func (b *BeatsMonitor) injectLogsInput(cfg map[string]interface{}, componentInfos []componentInfo, monitoringOutput string) error {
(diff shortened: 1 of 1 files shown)


internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -1032,6 +958,136 @@ func dropFieldsProcessor(fields []any, ignoreMissing bool) map[string]any {
...
+		},
+	}
+}
+
+// useComponentDatasetProcessors returns a list of processors which replace data_stream.dataset with component.dataset
+// if the latter is set. It also sets event.dataset to the same value. This is used to ensure logs from components
+// routed to the elastic-agent logger get sent to the component-specific dataset.
+func useComponentDatasetProcessors() []any {
+	return []any{
+		map[string]any{
```
pkoutsovasilis, line 997: nit
```suggestion
                 // copy original dataset so we can drop the dataset field
		map[string]any{
```

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -897,6 +772,57 @@ func (b *BeatsMonitor) getServiceComponentProcessMetricInputs(
...
+	processors := []any{
+		// drop all events from monitoring components (do it early)
+		// without dropping these events the filestream gets stuck in an infinite loop
+		// if filestream hits an issue publishing the events it logs an error which then filestream monitor
+		// will read from the logs and try to also publish that new log message (thus the infinite loop).
+		dropEventsFromMonitoringComponentsProcessor(),
+		// drop periodic metrics logs (those are useful mostly in diagnostic dumps where we collect log files)
+		dropPeriodicMetricsLogsProcessor(),
+	}
+	// if the event is from a component, use the component's dataset
```
pkoutsovasilis, line 786: I see an `if` in the comment but the following appends happen always, am I missing something?

swiatekm: The append always happens, but some of the added processors are conditional. The comment describes what all the added processors do collectively. Do you think it's unclear?

pkoutsovasilis (approved):
LGTM

Issue #6667, opened 2025-01-31
Review TestPreviousMinor() (and version in general) unit tests


pchila:
TestPreviousMinor() fails when bumping from 9.0.0 to 9.1.0 as in https://github.com/elastic/elastic-agent/pull/6649

The goal of this issue is to fix the unit test by removing the `t.Skip()` and have predictable behavior using test data.

VihasMakwana:
@pchila can you take a look at https://github.com/elastic/elastic-agent/pull/7888 if it is an expected fix?

pchila:
> [@pchila](https://github.com/pchila) can you take a look at [#7888](https://github.com/elastic/elastic-agent/pull/7888) if it is an expected fix?

@VihasMakwana PR [#7888](https://github.com/elastic/elastic-agent/pull/7888)  doesn't address two fundamental problems of PreviousMinor() unit tests:
- result of the tests still depend on content of module [globals](https://github.com/elastic/elastic-agent/blob/1ab183db9539c2d30d91ab19274d2e0fef26f32f/testing/upgradetest/versions.go#L82) initialized with [init() function from a yaml file](https://github.com/elastic/elastic-agent/blob/1ab183db9539c2d30d91ab19274d2e0fef26f32f/testing/upgradetest/versions.go#L85-L93) that is changed using Github automation.
- the test just checks that we are able to extract a previous minor from that data but it doesn't formalize any expected behavior of PreviousMinor() function

For these reasons [#7888](https://github.com/elastic/elastic-agent/pull/7888) cannot be considered a full fix for this issue, rather a quick way to reenable the test on 9.0 branch with limited value as a general solution.

swiatekm:
> result of the tests still depend on content of module [globals](https://github.com/elastic/elastic-agent/blob/1ab183db9539c2d30d91ab19274d2e0fef26f32f/testing/upgradetest/versions.go#L82) initialized with [init() function from a yaml file](https://github.com/elastic/elastic-agent/blob/1ab183db9539c2d30d91ab19274d2e0fef26f32f/testing/upgradetest/versions.go#L85-L93) that is changed using Github automation.

Is the problem here the logic of keeping and updating this information in the repository, or secretly doing I/O in an init function, in your view?

> the test just checks that we are able to extract a previous minor from that data but it doesn't formalize any expected behavior of PreviousMinor() function

[...]

PR #7888, opened 2025-04-17
[chore][testing] fix `PreviousMinor` function

VihasMakwana:
`PreviousMinor` is supposed to return previous minor version available for upgrade . 
1. For `9.0.0` it should return `8.19.0-SNAPSHOT`
2. For `9.1.0` it should return `9.0.0`

It works as expected for above two scenarios, but it fails when patch release comes into play for a first major release (x.0.x):
1. For `9.0.1`, `PreviousMinor(..)` should return `8.19.0-SNAPSHOT`, but it errors out. 

This PR fixes the function and test. I tested it on `9.0` and `8.19` branches and it worked as expected. Let me know if I missed anything.

Related: https://github.com/elastic/elastic-agent/issues/6667
Related bk failure: https://buildkite.com/elastic/elastic-agent/builds/20152#01964228-cba6-4907-9453-c01c488d8980


pkoutsovasilis (commented):
@VihasMakwana I will be honest and say that I am not fully aware what the expected test versions should be, I can't figure out the logic behind this; maybe @michalpristas @rdner you guys know better what are the requirements here? 

That said, @VihasMakwana please run `mage integration:updateVersions` and push the changes in `.upgrade-test-agent-versions.yml` so the CI is tested with them. After running your changes this is the content of `.upgrade-test-agent-versions.yml`

```yaml
testVersions:
  - 9.0.1-SNAPSHOT
  - 9.0.0
  - 8.18.0
  - 7.17.29-SNAPSHOT
```

Also having draft PRs with your changes for both `9.0` and `8.19` would be extremely helpful to see how your changes affect the `.upgrade-test-agent-versions.yml` under each branch.

swiatekm:
@pkoutsovasilis I don't think the function @VihasMakwana is changing - `PreviousMinor` - affects how we determine which versions to test against. Rather, it looks at these versions (as represented in `.upgrade-test-agent-versions.yml`), and tries to find the previous minor version to run upgrade tests against.

pkoutsovasilis:
> @pkoutsovasilis I don't think the function @VihasMakwana is changing - `PreviousMinor` - affects how we determine which versions to test against. Rather, it looks at these versions (as represented in `.upgrade-test-agent-versions.yml`), and tries to find the previous minor version to run upgrade tests against.

yep after looking again thx @swiatekm I see that you are right. That said we still need to test this with the latest `.upgrade-test-agent-versions.yml` file that `mage integration:updateVersions` is giving us, right?

swiatekm:
> > @pkoutsovasilis I don't think the function @VihasMakwana is changing - `PreviousMinor` - affects how we determine which versions to test against. Rather, it looks at these versions (as represented in `.upgrade-test-agent-versions.yml`), and tries to find the previous minor version to run upgrade tests against.
> 
> yep after looking again thx @swiatekm I see that you are right. That said we still need to test this with the latest `.upgrade-test-agent-versions.yml` file that `mage integration:updateVersions` is giving us, right?

Ideally, but I'm not fully sure that command isn't wrong as well. :unamused:

pkoutsovasilis:
> Ideally, but I'm not fully sure that command isn't wrong as well. 😒

I am not fully sure of that either 😄 but ideally I would want to be sure of this command first so I can build up confidence to review this PR. That said, if you are confident with the changes of this PR @swiatekm, feel free to approve 🙂 no blockers on my end

swiatekm:
I experimentally ported this change to 9.0 and CI passed there: https://github.com/elastic/elastic-agent/pull/7907. @pchila @pkoutsovasilis you reckon we could merge this change to unblock the 9.0 CI while @kaanyalti works out a proper fix?

ebeahan:
> you reckon we could merge this change to unblock the 9.0 CI while @kaanyalti works out a proper fix?

++ as a workaround while @kaanyalti finds a more permanent solution.

pkoutsovasilis (approved):

PR #7894, opened 2025-04-17
[otel] Disable process scraper of hostmetrics receiver

ChrsMark:
## What does this PR do?

This PR disables the `process` scraper of the `hostmetrics` receiver for existing configurations. This affects the helm values files and the distro specific sample configs.

## Why is it important?

This scraper collects metrics for all available processes of a host without an easy way to limit this to only report top N process for example. This results in quite big amount of timeseries. Since this is not quite critical for any of the available UIs or dashboards we decide to disable it temporarily until we find a better solution. Users that specifically need these metrics can also enable it back manually. 

## Checklist

## Disruptive User Impact

Metrics won't be collected for individual processes.

## Related issues

Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39423


rogercoll (approved):

swiatekm (approved):

ChrsMark:
I plan to merge this by Thursday (EoD, EEST).

smith:
@ChrsMark is it correct that this will make it so the process list in the processes tab will be empty? Will this be resolved before GA? cc @mlunadia

roshan-elastic:
> @ChrsMark is it correct that this will make it so the process list in the processes tab will be empty? Will this be resolved before GA? cc @mlunadia

Hey @mlunadia, just following up on this. Are you able to confirm the above?

We've [had a question](https://elastic.slack.com/archives/C031284ANNL/p1753101148320909?thread_ts=1752867874.531609&cid=C031284ANNL) in the OTel Slack channel asking about this.

My understanding is that the only way for us to show process metrics is for the [process scraper](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/samples/linux/logs_metrics_traces.yml#L27-L40)  to be turned on which is disabled by default.

cc @alaa-mallah

mlunadia:
> My understanding is that the only way for us to show process metrics is for the process scraper to be turned on which is disabled by default.

@roshan-elastic your team probably has a better chance of answering the impact on UIs by enabling it and/or checking which of the processes scraper metrics (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processesscraper/documentation.md) are being used or planned for use in the UI.

As for why it was disabled, the reasoning is explained in the issue. An escape hatch is available for users who want to enable it. I’ll confirm the latest status with the team next week, but as long as the cardinality problems described exist, it cannot be enabled by default.

roshan-elastic:
Thanks @mlunadia - that's really helpful.

I'll go back to @alaa-mallah and confirm that our UIs will work but they'll need to enable that setting by default.

It might be worth us thinking about referring to this in the documentation (both for the UI and ingest) so that users can understand that's what they need to do.

Do you know if this scraper is already mentioned anywhere?

PR #7924, opened 2025-04-22
Self-monitoring for beats receivers

swiatekm:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...


leehinman (dismissed):

khushijain21 (dismissed):

testing/integration/otel_test.go
```diff
@@ -1042,23 +1043,32 @@ func TestOtelFilestreamInput(t *testing.T) {
         file_identity.native: ~
 outputs:
   default:
+    type: elasticsearch
+    hosts: [{{.ESEndpoint}}]
+    api_key: "{{.ESApiKeyEncoded}}"
```
swiatekm, line 1048: These key encoding shenanigans will be unnecessary once we incorporate https://github.com/elastic/beats/pull/44031.

rdner: FYI: https://github.com/elastic/beats/pull/44031 has been merged.

swiatekm: I'll fix the tests in a follow-up.

internal/pkg/otel/configtranslate/otelconfig_test.go
```diff
@@ -277,6 +277,16 @@ func TestGetOtelConfig(t *testing.T) {
 			},
 		}
 	}
+
+	getBeatMonitoringConfig := func(_, _ string) map[string]any {
+		return map[string]any{
+			"http": map[string]any{
+				"enabled": true,
+				"host":    "localhost:9999",
```
leehinman, line 285: I don't think this will change anything, but shouldn't this be:
```golang
"http": map[string]any{
        "enabled": true
        "host": "localhost"
        "port": 9999
}
```
I think the port is supposed to be it's own entry in the map.  https://www.elastic.co/docs/reference/beats/metricbeat/http-endpoint

leehinman (approved):
LGTM

one question on a test, but I don't think it changes the test.

internal/pkg/otel/configtranslate/otelconfig_test.go
swiatekm: Yeah, that's a bit misleading. Agent actually never sets the port, because we use unix domain sockets, but I'll correct it for clarity.

cmacknz:
```yaml
inputs:
# Collecting log files
- type: filestream
  id: your-input-id
  use_output: default
  _runtime_experimental: otel
 # ...

agent:
  grpc:
    port: 16789
  monitoring:
    metrics: true
    logs: true
```

The way we want to release this, we are going to want the ability to run the monitoring Beats as OTel receivers regardless of if any input is being run as an OTel receiver.


What we want in the end is:

```yaml
inputs:
# Collecting log files
- type: filestream
  id: your-input-id
  use_output: default
 # ... still a Filebeat sub-process

agent:
  grpc:
    port: 16789
  monitoring:
    metrics: true
    logs: true
    _runtime_experimental: otel # Only monitoring Beats are in the collector.
````

This will let us turn on Beats receivers everywhere agent is used while containing their impact only to monitoring data. I don't think this needs to be addressed as part of this PR specifically but we do need to do this.

File: testing/integration/otel_test.go [hunk 3]
```diff
@@ -1111,6 +1121,27 @@ agent:
...
+			docs, err := estools.GetLogsForIndexWithContext(findCtx, info.ESClient, metricsIndex, map[string]interface{}{
+				"component.id": "filestream-default",
+			})
+			require.NoError(ct, err)
+
+			actualHits.Hits = docs.Hits.Total.Value
+			output, execErr := fixture.ExecStatus(context.Background())
+			require.NoError(ct, execErr)
+			t.Logf("status output: %v", output)
+			assert.Greater(ct, actualHits.Hits, 0)
```
cmacknz, line 1140: This only checks that we wrote some documents, not that the document content is what it was supposed to be. I think this test would probably pass if we only wrote a JSON document that container `@timestamp` (maybe with some other minimum requirements to be a valid document).

Where are we validating that the documents this produces are what we expect and are compatible with the elastic_agent integration and the calculations the Fleet UI does for CPU and memory usage? Is that going to need a follow up?

cmacknz: I see this might be coming in via https://github.com/elastic/elastic-agent/pull/7622 separately CC @mauri870

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -835,25 +864,49 @@ func processorsForProcessMetrics(binaryName, unitID, namespace, dataset string,
...
+	runtimeManager component.RuntimeManager,
+) []any {
+	processors := []any{
 		addDataStreamFieldsProcessor(dataset, namespace),
 		addEventFieldsProcessor(dataset),
 		addElasticAgentFieldsProcessor(binaryName, agentInfo),
 		addAgentFieldsProcessor(agentInfo.AgentID()),
 		addComponentFieldsProcessor(binaryName, unitID),
 	}
+	if runtimeManager == component.OtelRuntimeManager { // we don't want process metrics for beats receivers
```
cmacknz, line 879: These metrics are duplicated, but the UI may still want them.

We need to take a look at the [elastic_agent](https://github.com/elastic/integrations/blob/main/packages/elastic_agent) integration and make sure things still work.

What should happen is that `metrics-elastic_agent-*` is enough to calculate this and you no longer need to also account for `metrics-elastic-agent.$beat-*` in queries and dashboards. It should be possible for this to work but we need to explicitly check it.

cmacknz: The Fleet CPU and Metric aggregations are going to be querying into the per beat indexes I think, we need to make sure we didn't break that https://github.com/elastic/elastic-agent/blob/main/docs/cpu-metrics-in-fleet.md

testing/integration/otel_test.go
swiatekm: I think @mauri870's tests cover collecting data using beats receivers, not metrics emitted by the receivers themselves.

> Where are we validating that the documents this produces are what we expect and are compatible with the elastic_agent integration and the calculations the Fleet UI does for CPU and memory usage? Is that going to need a follow up?

We're not validating it in any E2E tests added in this PR. I did do a manual test of the agent dashboard, you can see the screenshots in the PR description. I'd like to add those as a follow-up, since they're going to be big and tedious. Here, I only validated that we are, in fact, getting metrics from beats receivers at all. Is that ok?

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: At a glance, the Fleet aggregations don't care about the index, so that part should be fine. I'm not completely sure about the beats-specific parts of the agent integration - I'll have a look at those.

I'd be inclined to file a follow-up issue and investigate there. Modifying what exactly these events contain isn't particularly difficult, we can always fix it later if it turns out the process metrics need to exist after all.

swiatekm:
> ```yaml
> inputs:
> # Collecting log files
> - type: filestream
>   id: your-input-id
>   use_output: default
>   _runtime_experimental: otel
>  # ...
> 
> agent:
>   grpc:
>     port: 16789
>   monitoring:
>     metrics: true
>     logs: true
> ```
> 
> The way we want to release this, we are going to want the ability to run the monitoring Beats as OTel receivers regardless of if any input is being run as an OTel receiver.
> 
> What we want in the end is:
> 
> ```yaml
> inputs:
> # Collecting log files
> - type: filestream
>   id: your-input-id
>   use_output: default
>  # ... still a Filebeat sub-process
> 
> agent:
>   grpc:
>     port: 16789
>   monitoring:
>     metrics: true
>     logs: true
>     _runtime_experimental: otel # Only monitoring Beats are in the collector.
> ```
> 
> This will let us turn on Beats receivers everywhere agent is used while containing their impact only to monitoring data. I don't think this needs to be addressed as part of this PR specifically but we do need to do this.

Yes, this is a follow up I already have prepared. For various reasons, it's simpler to merge this change first.

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: I tested this change (alongside the follow-up change enabling the use of beats receivers for self-monitoring) with Fleet, and didn't see any problems related to CPU and memory metrics. Both on the agent list, and in the Elastic Agent integration dashboard, these values were shown correctly.

I did see some issues, but these are related to beats internal metrics and are already tracked in https://github.com/elastic/beats/issues/44030.

testing/integration/otel_test.go
```diff
@@ -1111,6 +1121,27 @@ agent:
 		2*time.Minute, 5*time.Second,
 		"Expected %d logs, got %v", numEvents, actualHits)
 
+	metricsIndex := ".ds-metrics-elastic_agent*"
```
cmacknz, line 1124: I setup what this test is doing manually and I still see the beat sub-processes for monitoring running. Note the agent commit is https://github.com/elastic/elastic-agent/pull/7924/commits/a087f2803c0a658b50f5b33906ee7821979102bb from this branch.

```sh
sudo elastic-development-agent status --output=full
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: 23052443-e6b7-4e7e-8f66-e574471b6f9e
   │  ├─ version: 9.1.0
   │  └─ commit: a087f2803c0a658b50f5b33906ee7821979102bb
   ├─ beat/metrics-monitoring
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '62414'
   │  ├─ beat/metrics-monitoring
   │  │  └─ type: OUTPUT
   │     └─ type: INPUT
   ├─ filestream-monitoring
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '62413'
   │  ├─ filestream-monitoring
   │  │  └─ type: OUTPUT
   │  └─ filestream-monitoring-filestream-monitoring-agent
   │     └─ type: INPUT
   ├─ http/metrics-monitoring
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '62415'
   │  ├─ http/metrics-monitoring
   │  │  └─ type: OUTPUT
   │  └─ http/metrics-monitoring-metrics-monitoring-agent
   │     └─ type: INPUT
   └─ pipeline:logs/_agent-component/filestream-default
      ├─ status: StatusOK
      ├─ exporter:elasticsearch/_agent-component/default
      │  └─ status: StatusOK
      └─ receiver:filebeatreceiver
```

If this is also the case here, you need this test to pass when monitoring is done exclusively with beats receivers or you aren't testing them. Add a condition to confirm the standard monitoring sub-processes aren't running at the same time or this test passes regardless of whether the monitoring beat receivers do anything.

This is the config I used:

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    api_key: "example-key"
    preset: balanced
inputs:
  - type: filestream
    id: your-input-id
    _runtime_experimental: otel
    streams:
      - id: your-filestream-stream-id
        data_stream:
          dataset: generic
        paths:
          - /var/log/*.log
agent.monitoring:
  enabled: true
  logs: true
  metrics: true
agent.logging.to_stderr: true

```

swiatekm: I think you might be misunderstanding what this change is meant to accomplish. It's about monitoring the beats receivers themselves. Using beats receivers for self-monitoring will happen in a follow-up.

The output you're seeing from this test is what is expected. We have one component - `filestream-default` running as a beat receiver, and the monitoring components running as beats processes. The test checks that we get metrics from `filestream-default`, which is what this PR enables. Does that make sense?

cmacknz (approved):
LGTM, thanks for the clarifications.

PR #7932, opened 2025-04-23
[testing] - minor enhancements to perform testing against local changes 

VihasMakwana:
## What does this PR do?

If we want to run agent integration tests with local beats changes, we're required to do two things:
1. Add a replace directive 
2. Update docker cross builder to mount the local beats directory
    - This is necessary or else packaging will fail with `replacement directory does not exist`

We can automate the 2nd step if `EXTERNAL=false` and add a guide on how to perform these steps.

Note: This is only applicable for otel related changes on filebeat and metricbeat receivers. We don't directly use beats anywhere else AFAIK.

## Why is it important?

It wasn't straightforward to me that the 2nd step was required. Adding steps and automating it make it much easier. Without perfoming the 2nd step, packaging fails with:

```replacement directory /Users/vihasmakwana/Desktop/Vihas/elastic/beats does not exist```

## Checklist

To run this locally, run 
```sh
LOCAL=true PLATFORMS=darwin/arm64 mage -v package
```

## Related issues

-

## Questions to ask yourself

- ...


swiatekm:
I like the change, hacking the build system to do achieve this is quite annoying. I don't like the env variable name - `LOCAL` is not specific enough, I'd prefer something like `MOUNT_LOCAL_BEATS_REPOSITORY`. But I think it would be fine to skip the new variable altogether and always mount this when EXTERNAL=false, because in that case we're building beats locally anyway and the repository needs to exist. Honestly, I'd be fine with just always doing this if the directory exists. WDYT @pchila @pkoutsovasilis ?

pkoutsovasilis (commented):
@VihasMakwana @swiatekm the change conceptually makes sense to me, but the newly introduced `LOCAL` env vars feels kinda bit off to me. How about the following, when `EXTERNAL=false`, we extend the respective building code to invoke through cli `go mod edit -replace ....` and do the appropriate mounts to the container, wdyt?

swiatekm:
> @VihasMakwana @swiatekm the change conceptually makes sense to me, but the newly introduced `LOCAL` env vars feels kinda bit off to me. How about the following, when `EXTERNAL=false`, we extend the respective building code to invoke through cli `go mod edit -replace ....` and do the appropriate mounts to the container, wdyt?
> 
> UPDATE: now I re-read your comment @swiatekm , so we are fully aligned and I think that the go mod edit proposal above in my comment is not prohibitive 🙂

I don't know if I like editing the go.mod just because someone wants to build with EXTERNAL=false. In my experience, these kinds of changes are easy to commit by accident. I think most of the time you don't actually want the replace, though maybe this will change with time, as Otel is used more.

WDYT about checking if the beats directory exists, and always mounting if it does? The mount by itself doesn't have any effect on the build.

VihasMakwana:
@swiatekm @pkoutsovasilis 

> I don't know if I like editing the go.mod just because someone wants to build with EXTERNAL=false. In my experience, these kinds of changes are easy to commit by accident. I think most of the time you don't actually want the replace, though maybe this will change with time, as Otel is used more.

In my opinion we have two options,

1. We can accomplish what Panos suggested, without worrying about committing. 
We add a new flag `MOUNT_LOCAL_BEATS_REPOSITORY` and if it set to `true`:
   - we save a backup of `go.mod` under `.agent-testing`.
   - we invoke `go mod edit -replace ... && go mod tidy` before packaging
   - we also add a defer call that restores the backed up `go.mod` once everything is finished
     - if this fails, we can call `go mod edit -dropreplace ... && go mod tidy`
2. we can leave this to user and document that they need to explicitly add a replace, if they want to test local changes. If we decide to go with option 2, we don't need to add any new flag and we can mount the beats if `EXTERNAL=false`.

Let me know what you guys think

VihasMakwana:
Yeah you're right. I missed that initially.
We can keep new variables out for now. 

---

Let me know which one of the above options sound good to you 😄

swiatekm (dismissed):

khushijain21:
@VihasMakwana you might want to update the PR description as well

docs/test-framework-dev-guide.md
```diff
@@ -292,6 +292,22 @@ until it reports a failure.
...
 - `TEST_RUN_UNTIL_FAILURE=true mage integration:single [testName]`
 
+### Running integration tests with local changes in beats (for otel receivers)
+
+If you've made local changes to Beats-OTeL and want to run the Agent's integration tests against those changes, follow these steps:
+1. Update go.mod and add a replace directive as follows (change the path to match your local machine):
+```
+replace github.com/elastic/beats/v7 => /Users/vihasmakwana/Desktop/Vihas/elastic/beats
+```
+2. Package the Agent with `LOCAL=true`:
```
pkoutsovasilis, line 302: is this still valid? 🙂

VihasMakwana: Missed this. Fixed it now.

pkoutsovasilis (approved):

swiatekm (approved):

PR #7948, opened 2025-04-24
Removing unreachable files

ycombinator:
## What does this PR do?

This PR removes files related to developer/CI tooling that are no longer being used.

## Why is it important?

To reduce noise when trying to navigate our developer and CI tooling scripts.


swiatekm (approved):

pkoutsovasilis (approved):
Thanks for the tidying up @ycombinator LGTM

v1v:
> LGTM. However, I would like feedback from @elastic/observablt-ci and @elastic/observablt-ci-contractors to determine if those files are used in the ci-agents-images.


Damn! I didn't know it was `auto-merged`. Regardless, let's wait for the feedback and we can revert if needed

PR #7957, opened 2025-04-24
[ci][linkcheck] - Change GOBIN env

VihasMakwana:
I'm not really sure why, but our CI is failing to run `link-patrol` and I observed two failures (there can be more):
- https://buildkite.com/elastic/elastic-agent/builds/20404#0196662a-957c-4809-af2d-201ef6cb9598
- https://buildkite.com/elastic/elastic-agent/builds/20407#01966642-a791-4f27-868b-307c215dc749

In this PR, I'm adding a new `BIN()` option which forces `go install ...` to use `$(pwd)/bin` directory to install binaries and I've updated `runGoLinkCheck` to make use of local binary.


swiatekm:
This is most likely due to a buildkite base image change, though we're not sure what the root cause is exactly. Investigation is underway.

PR #7995, opened 2025-04-25
Enable Filebeat's Journald input in some docker image variants

belimawr:
## What does this PR do?

Filebeat's Journald input requires `journalctl`, so the systemd package is added to all docker image variants.

#### Images built from `main` (f1086184d87ed755b9275081c3b42f58c54feb63):
```
docker.elastic.co/elastic-agent/elastic-agent-complete         9.1.0-SNAPSHOT         123d3a5d0a22   2 hours ago   3.57GB
docker.elastic.co/elastic-agent/elastic-otel-collector         9.1.0-SNAPSHOT         a5182ef34ae0   2 hours ago   1.2GB
docker.elastic.co/elastic-agent/elastic-agent                  9.1.0-SNAPSHOT         cf9309afdc3f   2 hours ago   1.89GB
```

#### Images built from this PR:
```
docker.elastic.co/elastic-agent/elastic-agent-complete         9.1.0-SNAPSHOT         18845462f221   16 minutes ago   3.58GB
docker.elastic.co/elastic-agent/elastic-otel-collector         9.1.0-SNAPSHOT         2f29d6c0f17a   18 minutes ago   1.25GB
docker.elastic.co/elastic-agent/elastic-agent                  9.1.0-SNAPSHOT         d5b99148748a   18 minutes ago   1.94GB
```

#### Increase in size
 - `docker.elastic.co/elastic-agent/elastic-agent-complete`: 10MB
 - `docker.elastic.co/elastic-agent/elastic-otel-collector`: 50MB
 - `docker.elastic.co/elastic-agent/elastic-agent`: 40MB
 
## Why is it important?

Currently it is only possible to use integrations that requires the journald input with the complete docker image variant, this PR allows it to be used with any docker image variant. The complete docker image is the largest with more than 3Gb, while the standard Elastic-Agent image is less than 2Gb.

## Checklist

- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~

~~## Disruptive User Impact~~

## Author's Checklist
 - [x] Add integration test for each docker image: mount the journal into the container and ensure the Journald input can read it.

1. Package the Elastic-Agent:
    ```
    DEV=true SNAPSHOT=true EXTERNAL=true PACKAGES="docker" PLATFORMS=linux/amd64 mage -v package
    ```
2. Check the version of `journalctl` on each docker image by running the following script/commands:
    ```
    #!/bin/bash
    
    echo "############################## Testing images with journalctl ##############################"
    
    set -o xtrace
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent-complete:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent-ubi:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-otel-collector:9.1.0-SNAPSHOT  --version
    
    set +o xtrace
    echo "############################## Testing images without journalctl ##############################"
    
    set -o xtrace
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent-complete-wolfi:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent-slim-wolfi:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent-slim:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-agent-wolfi:9.1.0-SNAPSHOT  --version
    docker run --rm -it --entrypoint journalctl docker.elastic.co/elastic-agent/elastic-otel-collector-wolfi:9.1.0-SNAPSHOT  --version
    ```
3. Alternatively, check the Dockerfile for each of the images:
    ```sh
    grep "systemd" build/package/elastic-*/*/docker-build/Dockerfile
    ```
## Related issues

- Relates https://github.com/elastic/beats/issues/44040
- Docs update: https://github.com/elastic/docs-content/pull/1272

## Questions to ask yourself

- ...


rdner:
@belimawr please add the image size comparison to the description.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -73,15 +73,15 @@ ENV BEAT_SETUID_AS={{ .user }}
 {{- if contains .from "ubi-minimal" }}
 RUN for iter in {1..10}; do \
         microdnf update -y && \
-        microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz && \
+        microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz systemd && \
```
cmacknz, line 76: How much bigger does this make the images? 

Also, we don't want it in every container, it would be in elastic-agent, elastic-agent-complete but not elastic-agent-slim. This is really dependent on how much the size increase is as we do not want them getting bigger since they are already huge.

cmacknz: The downside of using the elastic-agent-complete image is that people who don't use synthetics but want to read node logs on K8S now have to deploy a bunch of browser dependencies and such.

Perhaps this would be best addressed with an elastic-agent-journald container. We already have a lot of containers though so I'm not sure this clarifies anything with respect to which one someone should use.

cmacknz (changes requested):
If we keep this (size dependent), it definitely shouldn't be in the slim variant at least.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -73,15 +73,15 @@ ENV BEAT_SETUID_AS={{ .user }}
...
         microdnf clean all && \
         exit_code=0 && break || exit_code=$? && echo "microdnf error: retry $iter in 10s" && sleep 10; done; (exit $exit_code)
 {{- end }}
 
 {{- if contains .from "wolfi" }}
 RUN for iter in {1..10}; do \
         apk fix && \
-        apk add --no-cache ca-certificates curl gawk shadow bash && \
+        apk add --no-cache ca-certificates curl gawk shadow bash systemd && \
```
cmacknz, line 84: We shouldn't include systemd in wolfi because it is going to significantly broaden it's exposure to CVEs, the point of Wolfi is to minimize the OS level CVEs we are exposed to. Same rationale as Filebeat: https://github.com/elastic/beats/pull/44056/files#r2064612269

pkoutsovasilis:
So I went ahead and found my notes from when we did the initial switch from ubuntu-based to ubi-based images, and I remembered the quirks with journalctl and systemd package. To this end, `journalctl` is only part of the `systemd` package and it's not something that we can install standalone.  

I built all images with this PR and this is the comparison in terms of size:

arm64:
```
docker.elastic.co/elastic-agent/elastic-agent-complete         with-systemd-PR-9.1.0-SNAPSHOT   693b6e89332b   2 hours ago     3.33GB
docker.elastic.co/elastic-agent/elastic-agent-complete         9.1.0-SNAPSHOT                   002667864129   12 hours ago    3.33GB
docker.elastic.co/beats-ci/elastic-agent-service               with-systemd-PR-9.1.0-SNAPSHOT   bf4763072e2f   2 hours ago     2.3GB
docker.elastic.co/beats-ci/elastic-agent-service               9.1.0-SNAPSHOT                   bf4763072e2f   2 hours ago     2.3GB
docker.elastic.co/beats-ci/elastic-agent-cloud                 with-systemd-PR-9.1.0-SNAPSHOT   134ef9b6eed6   2 hours ago     2.21GB
docker.elastic.co/beats-ci/elastic-agent-cloud                 9.1.0-SNAPSHOT                   134ef9b6eed6   2 hours ago     2.21GB
docker.elastic.co/elastic-agent/elastic-agent-ubi              with-systemd-PR-9.1.0-SNAPSHOT   f6f367b91ffc   2 hours ago     1.67GB   (~ 40MB increase)
docker.elastic.co/elastic-agent/elastic-agent-ubi              9.1.0-SNAPSHOT                   c39e45552f7c   12 hours ago    1.63GB
docker.elastic.co/elastic-agent/elastic-agent                  with-systemd-PR-9.1.0-SNAPSHOT   f6f367b91ffc   2 hours ago     1.67GB   (~ 40MB increase)
docker.elastic.co/elastic-agent/elastic-agent                  9.1.0-SNAPSHOT                   c39e45552f7c   12 hours ago    1.63GB
docker.elastic.co/elastic-agent/elastic-agent-slim             with-systemd-PR-9.1.0-SNAPSHOT   7311176930d6   2 hours ago     1.12GB   (~ 50MB increase)
docker.elastic.co/elastic-agent/elastic-agent-slim             9.1.0-SNAPSHOT                   fc65146d79d6   12 hours ago    1.07GB
docker.elastic.co/elastic-agent/elastic-otel-collector         with-systemd-PR-9.1.0-SNAPSHOT   5eb9b4234489   2 hours ago     1.03GB   (~ 50MB increase)
docker.elastic.co/elastic-agent/elastic-otel-collector         9.1.0-SNAPSHOT                   10f3e676c678   12 hours ago    983MB
docker.elastic.co/elastic-agent/elastic-agent-wolfi            with-systemd-PR-9.1.0-SNAPSHOT   4d3de9efd830   2 hours ago     1.75GB   (~ 170MB increase)
docker.elastic.co/elastic-agent/elastic-agent-wolfi            9.1.0-SNAPSHOT                   c85d302eb13d   12 hours ago    1.58GB
docker.elastic.co/elastic-agent/elastic-agent-slim-wolfi       with-systemd-PR-9.1.0-SNAPSHOT   ac34c798b016   2 hours ago     1.2GB    (~ 180MB increase)
docker.elastic.co/elastic-agent/elastic-agent-slim-wolfi       9.1.0-SNAPSHOT                   d5cd36078fdf   12 hours ago    1.02GB
docker.elastic.co/elastic-agent/elastic-agent-complete-wolfi   with-systemd-PR-9.1.0-SNAPSHOT   1ef4f0086615   2 hours ago     2.97GB   (~ 180MB increase)
docker.elastic.co/elastic-agent/elastic-agent-complete-wolfi   9.1.0-SNAPSHOT                   147828688b56   12 hours ago    2.81GB
docker.elastic.co/elastic-agent/elastic-otel-collector-wolfi   with-systemd-PR-9.1.0-SNAPSHOT   9a7f5b346b99   2 hours ago     1.11GB   (~ 180MB increase)
docker.elastic.co/elastic-agent/elastic-otel-collector-wolfi   9.1.0-SNAPSHOT                   d2a00044c610   12 hours ago    933MB
```

amd64:
```
(amd64)
docker.elastic.co/elastic-agent/elastic-agent-complete         with-systemd-PR-9.1.0-SNAPSHOT   6bb84a795044   13 minutes ago  3.37GB   
docker.elastic.co/elastic-agent/elastic-agent-complete         9.1.0-SNAPSHOT                   0000e9ed5f23   12 hours ago    3.37GB   
docker.elastic.co/beats-ci/elastic-agent-service               with-systemd-PR-9.1.0-SNAPSHOT   95e3bfe0941c   13 minutes ago  2.38GB   
docker.elastic.co/beats-ci/elastic-agent-service               9.1.0-SNAPSHOT                   95e3bfe0941c   13 minutes ago  2.38GB   
docker.elastic.co/beats-ci/elastic-agent-cloud                 with-systemd-PR-9.1.0-SNAPSHOT   b889bd89bee4   15 minutes ago  2.31GB   
docker.elastic.co/beats-ci/elastic-agent-cloud                 9.1.0-SNAPSHOT                   b889bd89bee4   15 minutes ago  2.31GB   
docker.elastic.co/elastic-agent/elastic-otel-collector         with-systemd-PR-9.1.0-SNAPSHOT   e77a5648b478   15 minutes ago  1.06GB   (~ 30MB increase)  
docker.elastic.co/elastic-agent/elastic-otel-collector         9.1.0-SNAPSHOT                   46aedf7de7e1   12 hours ago    1.03GB   
docker.elastic.co/elastic-agent/elastic-agent-ubi              with-systemd-PR-9.1.0-SNAPSHOT   fa6b2f661969   15 minutes ago  1.75GB   (~ 30MB increase)  
docker.elastic.co/elastic-agent/elastic-agent-ubi              9.1.0-SNAPSHOT                   b6e665ec531c   12 hours ago    1.72GB   
docker.elastic.co/elastic-agent/elastic-agent                  with-systemd-PR-9.1.0-SNAPSHOT   fa6b2f661969   15 minutes ago  1.75GB   (~ 30MB increase)  
docker.elastic.co/elastic-agent/elastic-agent                  9.1.0-SNAPSHOT                   b6e665ec531c   12 hours ago    1.72GB   
docker.elastic.co/elastic-agent/elastic-agent-slim             with-systemd-PR-9.1.0-SNAPSHOT   58d3dd1d12d1   15 minutes ago  1.16GB   (~ 30MB increase) 
docker.elastic.co/elastic-agent/elastic-agent-slim             9.1.0-SNAPSHOT                   a58865886ade   12 hours ago    1.13GB   
docker.elastic.co/elastic-agent/elastic-agent-wolfi            with-systemd-PR-9.1.0-SNAPSHOT   fdaf4fa8a0ac   15 minutes ago  1.82GB   (~ 160MB increase)
docker.elastic.co/elastic-agent/elastic-agent-wolfi            9.1.0-SNAPSHOT                   b8f9b0d336ad   12 hours ago    1.66GB   
docker.elastic.co/elastic-agent/elastic-agent-slim-wolfi       with-systemd-PR-9.1.0-SNAPSHOT   464367c5c29c   15 minutes ago  1.24GB   (~ 170MB increase)
docker.elastic.co/elastic-agent/elastic-agent-slim-wolfi       9.1.0-SNAPSHOT                   a4007ffd9dea   12 hours ago    1.07GB   
docker.elastic.co/elastic-agent/elastic-agent-complete-wolfi   with-systemd-PR-9.1.0-SNAPSHOT   4824ac942ad1   13 minutes ago  3GB      (~ 140MB increase)
docker.elastic.co/elastic-agent/elastic-agent-complete-wolfi   9.1.0-SNAPSHOT                   1dcefb2b9c7c   12 hours ago    2.86GB   
docker.elastic.co/elastic-agent/elastic-otel-collector-wolfi   with-systemd-PR-9.1.0-SNAPSHOT   366fc969e3df   15 minutes ago  1.14GB   (~ 170MB increase)
docker.elastic.co/elastic-agent/elastic-otel-collector-wolfi   9.1.0-SNAPSHOT                   2b85fbbb4e34   12 hours ago    973MB   
```

Just from the above comparisons, and the security-based reasons, I would say that wolfi-based images should *not* feature systemd. Now for the ubi-based ones, if the on average 50MB of an increase is prohibitive, we could try out having an intermediate step that installs systemd package and copies the bare minimum of journalctl dependencies to the final image
```
$ du -sh /usr/bin/journalctl
132K    /usr/bin/journalctl
$ du -sh /usr/lib/aarch64-linux-gnu/systemd/libsystemd-shared-255.so
3.8M    /usr/lib/aarch64-linux-gnu/systemd/libsystemd-shared-255.so
$ readelf -d /usr/bin/journalctl | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libsystemd-shared-255.so]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-aarch64.so.1]
```
I have to test the above out but AFAIK `journalctl` either connects to the `/run/systemd/journal/socket`, host path that needs to be mounted, and/or reads persistent journal files on disk `/var/log/journal/*`

cmacknz:
IMO based on the above we keep systemd out of the slim, service, and wolfi variants. W

We would add systemd/journald to the plain `elastic-agent` (and `elastic-agent-ubi`) and the `elastic-otel-collector` since the collector needs journald for the same reason. This removes the need to deploy the complete variant and preserve the main properties of slim (smallest size) and wolfi (minimal dependencies and CVEs) variants.

cmacknz:
(service has no need for journald/systemd it's the agentless container).

pchila (commented):
I agree with @cmacknz assessment, adding `systemd` to all docker images is not what we want to do, please try to limit the addition to only some of the variants and not to the security/size focused ones.

belimawr:
Thanks for the input folks!

I believe we should add systemd/journalctl in the slim variant as well as the size increase is very small. How slim does the slim image need to be?

Based on @pkoutsovasilis numbers, adding systemd in the slim image is a 30Mb increase on a 1.13Gb, this means 2.65% increase in size, however if a user needs the journald input and migrates from the slim to the ubi variant, that means jumping from 1.13Gb to 1.75Gb, a 54.88% increase, that's a significant change.

If the user wants to monitor their Kubernetes cluster, I'd expect them to also want to collect the system logs, and if the logs are on journald, then they will need the journald input. So it looks to be this is a very important use case.

Do we have any numbers on this use case and which image was being used before we introduced the need for journalctl in the docker image?

In the mean time, I'll look into adding systemd based on the image variant.

cmacknz:
> I believe we should add systemd/journalctl in the slim variant as well as the size increase is very small. How slim does the slim image need to be?

It needs to be as small as possible while still being useful.  That including systemd is a marginal increase is a hint that we need to remove other things not that we can include any arbitrary thing that is 10-50 MB in size because the image is too huge as a starting point.

Everyone with this use case is already using the plain elastic-agent image because -slim was only available in 9.0 so we aren't breaking anything.

We can put journald in later if people tell us it is critical requirement, but removing it later based on an assumption will be a breaking change if we do it too early.

pierrehilbert:
@belimawr is it ready for another round of review?

belimawr:
> @belimawr is it ready for another round of review?

Yes, I was waiting for CI to be green before requesting another round of review.

changelog/fragments/1745609163-journalctl-on-all-docker-variants.yaml
```diff
@@ -0,0 +1,33 @@
...
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: |
+  Ship journalctl on the following Docker images variants: basic, complete, ubi and otel collector
```
cmacknz, line 15:
```suggestion
  Ship journalctl in the elastic-agent, elastic-agent-complete, and elastic-otel-collector UBI based Docker images to enable reading journald logs. It is omitted from elastic-agent-slim and all Wolfi based images.
```

Basic isn't a user facing container name, also it's clearer to just name the containers directly and include where we didn't include it.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -247,7 +259,7 @@ RUN for iter in {1..10}; do \
         microdnf -y install fontconfig freetype cairo glib2 gtk3 pango xorg-x11-fonts-misc xorg-x11-fonts-Type1 \
         at-spi2-atk atk at-spi2-core alsa-lib cups-libs dbus-libs libdrm mesa-libEGL mesa-libgbm nspr nss libX11 \
         libX11-xcb libxcb libXcomposite libXdamage libXext libXfixes libXrandr libxkbcommon libxshmfence glib2 \
-        dbus-glib libicu mesa-libGL unzip iptables systemd && \
```
cmacknz, line 250: I think doing this above instead reduces duplication, but adds another layer to the container.

cmacknz (commented):
Changes look to be in the right places?

What is the size change of each container after these changes to confirm the correct things changed? 

P.S. once this merges I think we can simplify the integration tests in the system package https://github.com/elastic/integrations/blob/9f744b6f150d31dbbc1a8cd8b367c70e0fbbb5f3/packages/system/data_stream/syslog/_dev/test/system/test-journald-config.yml#L9

belimawr:
> Changes look to be in the right places?

Yes, they do. I tested all docker images generated by `mage package`. The script I used for testing is in the "How to test this PR locally" section of the description.

> What is the size change of each container after these changes to confirm the correct things changed?

I did not compare the sizes before and after the last commit I pushed to this branch/PR. To ensure the changes are in the correct place I checked:
- The `Dockerfile` generated for each image variant
- If `journalct` is installed by trying to run it from the container.

Both procedures are detailed on "How to test this PR locally".

If you want I can re-run them all to post the output here and before/after sizes here.

belimawr:
Thanks @cmacknz and @pkoutsovasilis ! I'll look into the K8s integration tests.

> I'm not sure I 100% understand what shipping the Docker packages to the host VM means
I meant building the docker image from the PR and sending/adding it to the VM running the tests.

belimawr:
Some of the failures are not related to this PR:
- https://github.com/elastic/elastic-agent/issues/8282

The others I'm investigating

dev-tools/packaging/settings.go
```diff
@@ -219,9 +218,6 @@ func parsePackageSettings(r io.Reader) (*packagesConfig, error) {
 		return nil, fmt.Errorf("unmarshalling package spec yaml: %w", err)
 	}
 
-	if mg.Verbose() {
```
pchila, line 222: Any specific reason for removing this verbose log ?

belimawr: Yes, having (I'm adding it here as a single line comment to better reflect how it looks on a terminal when using mage)
`
2025/06/03 09:17:50 Read packages config: &{Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64} {OS:windows Arch:x86_64} {OS:darwin Arch:x86_64} {OS:darwin Arch:aarch64}] PackageTypes:[1 2 3 4 5] Components:[{BinaryName:agentbeat PackageName:agentbeat-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:agentbeat-{{.Version}}-{{.Platform}} ProjectName:beats FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64} {OS:windows Arch:x86_64} {OS:darwin Arch:x86_64} {OS:darwin Arch:aarch64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:apm-server PackageName:apm-server-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:apm-server-{{.Version}}-{{.Platform}} ProjectName:apm-server FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64} {OS:windows Arch:x86_64} {OS:darwin Arch:x86_64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:cloudbeat PackageName:cloudbeat-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:cloudbeat-{{.Version}}-{{.Platform}} ProjectName:cloudbeat FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:connectors PackageName:connectors-{{.Version}}.zip RootDir:elasticsearch_connectors-{{beat_version}} ProjectName:connectors FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:true PackageTypes:[1 2 3 4 5]} {BinaryName:endpoint-security PackageName:endpoint-security-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:endpoint-security-{{.Version}}-{{.Platform}} ProjectName:endpoint-dev FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64} {OS:windows Arch:x86_64} {OS:darwin Arch:x86_64} {OS:darwin Arch:aarch64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:fleet-server PackageName:fleet-server-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:fleet-server-{{.Version}}-{{.Platform}} ProjectName:fleet-server FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64} {OS:windows Arch:x86_64} {OS:darwin Arch:x86_64} {OS:darwin Arch:aarch64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:pf-elastic-collector PackageName:pf-elastic-collector-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:pf-elastic-collector-{{.Version}}-{{.Platform}} ProjectName:prodfiler FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:pf-elastic-symbolizer PackageName:pf-elastic-symbolizer-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:pf-elastic-symbolizer-{{.Version}}-{{.Platform}} ProjectName:prodfiler FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:pf-host-agent PackageName:pf-host-agent-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:pf-host-agent-{{.Version}}-{{.Platform}} ProjectName:prodfiler FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:elastic-agent PackageName:elastic-agent-core-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:elastic-agent-core-{{.Version}}-{{.Platform}} ProjectName:elastic-agent-core FIPS:false Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64} {OS:windows Arch:x86_64} {OS:darwin Arch:x86_64} {OS:darwin Arch:aarch64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:agentbeat PackageName:agentbeat-fips-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:agentbeat-{{.Version}}-{{.Platform}} ProjectName:beats FIPS:true Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:apm-server PackageName:apm-server-fips-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:apm-server-fips-{{.Version}}-{{.Platform}} ProjectName:apm-server FIPS:true Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:fleet-server PackageName:fleet-server-fips-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:fleet-server-fips-{{.Version}}-{{.Platform}} ProjectName:fleet-server FIPS:true Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]} {BinaryName:elastic-agent PackageName:elastic-agent-core-fips-{{.Version}}-{{.Platform}}.{{.Ext}} RootDir:elastic-agent-core-fips-{{.Version}}-{{.Platform}} ProjectName:elastic-agent-core FIPS:true Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}] PythonWheel:false PackageTypes:[1 2 3 4 5]}] Settings:{FIPS:{Compile:{CGO:true Env:map[GOEXPERIMENT:systemcrypto] Tags:[requirefips ms_tls13kdf] Platforms:[{OS:linux Arch:x86_64} {OS:linux Arch:arm64}]}}}}
`

Printed every time I ran some mage command like `mage -v integration:kubernetes`, did not seem helpful, it looked like some debug left over.

If it is helpful, I'm fine with reverting this change, but at the moment I cannot see a use for it.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
cmacknz: How much larger does the container get with the extra layer?

testing/integration/journald_test.go
```diff
@@ -0,0 +1,265 @@
...
+			}
+
+			// Query the index and filter by the input type
+			docs := findESDocs(t, func() (estools.Documents, error) {
+				return estools.GetLogsForIndexWithContext(
+					ctx,
+					info.ESClient,
+					fmt.Sprintf("logs-generic.otel-%s", namespace),
+					map[string]any{
+						"event.dataset": "generic.otel",
```
cmacknz: The test passes if the collector ships any logs at all? That doesn't seem like a strong enough condition, it has to ship a journald log.

cmacknz: If you configure the ES exporter's dynamic routing mode, then have a processor set the journald dataset for in the attributes of the events from the filebeatreceiver, the test condition can be similar to what you did for plain Elastic Agent. 

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticsearchexporter#elasticsearch-document-routing

You might be able to get the two tests to merge into one: container is a parameter, whether to check status is a parameter, the target index is a parameter. The rest looks like it is the same.

belimawr:
I added the difference in size to the PR description.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
belimawr: Do you mean the image size? I've added that to the PR description.

testing/integration/journald_test.go
belimawr: > The test passes if the collector ships any logs at all? 

No, I'm querying directly the index/datastream generated by my specific configuration, that contains a single input: journald.

I first set `data_stream.namespace` using an env var: https://github.com/elastic/elastic-agent/blob/5861ff3f76bf1329434c794acc4538e741db902a/testing/integration/testdata/journald-otel.yml#L22-L24

Then I set the env var with a test-unique name: https://github.com/elastic/elastic-agent/blob/5861ff3f76bf1329434c794acc4538e741db902a/testing/integration/journald_test.go#L190-L193 
 
And finally I query the data_stream for the specific namespace I set before: https://github.com/elastic/elastic-agent/blob/5861ff3f76bf1329434c794acc4538e741db902a/testing/integration/journald_test.go#L255

> You might be able to get the two tests to merge into one: container is a parameter, whether to check status is a parameter, the target index is a parameter. The rest looks like it is the same.

The configuration YAML is very different for Elastic-Agent and OTelCollector, hence I ended up with different tests, also the container used is set in the `define.Require`, which cannot be changed during the test runtime.

What I can consolidate is the actual test function run by `t.Run`, I can parameterise that with the index and query map to reduce a bit the code duplication.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
cmacknz: I meant how much of a size difference does this layer make?

testing/integration/journald_test.go
cmacknz: Changing the namespace makes this index specific to the test which is good, but you are using the default index any log from edot would land in which was my concern.

The configuration only contains the journald receiver so I don't see how we'd end up with an event from a different place.

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
cmacknz: `docker image history --format=json --no-trunc docker.elastic.co/elastic-agent/elastic-agent:9.1.0-SNAPSHOT`

```json
{"Comment":"buildkit.dockerfile.v0","CreatedAt":"2025-06-05T16:13:03-04:00","CreatedBy":"RUN /bin/sh -c for iter in {1..10}; do           microdnf update -y \u0026\u0026           microdnf install -y systemd \u0026\u0026           microdnf clean all \u0026\u0026           exit_code=0 \u0026\u0026 break || exit_code=$? \u0026\u0026 echo \"microdnf error: retry $iter in 10s\" \u0026\u0026 sleep 10; done; (exit $exit_code) # buildkit","CreatedSince":"26 minutes ago","ID":"\u003cmissing\u003e","Size":"59.2MB"}
{"Comment":"buildkit.dockerfile.v0","CreatedAt":"2025-06-05T16:12:56-04:00","CreatedBy":"RUN /bin/sh -c for iter in {1..10}; do         microdnf update -y \u0026\u0026         microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz \u0026\u0026         microdnf clean all \u0026\u0026         exit_code=0 \u0026\u0026 break || exit_code=$? \u0026\u0026 echo \"microdnf error: retry $iter in 10s\" \u0026\u0026 sleep 10; done; (exit $exit_code) # buildkit","CreatedSince":"26 minutes ago","ID":"\u003cmissing\u003e","Size":"16.8MB"}
```

cmacknz: If I combine them I get 61.8 MB

```json
{"Comment":"buildkit.dockerfile.v0","CreatedAt":"2025-06-05T16:47:50-04:00","CreatedBy":"RUN /bin/sh -c for iter in {1..10}; do           microdnf update -y \u0026\u0026           microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz systemd \u0026\u0026           microdnf clean all \u0026\u0026           exit_code=0 \u0026\u0026 break || exit_code=$? \u0026\u0026 echo \"microdnf error: retry $iter in 10s\" \u0026\u0026 sleep 10; done; (exit $exit_code) # buildkit","CreatedSince":"13 minutes ago","ID":"\u003cmissing\u003e","Size":"61.8MB"}
```

Compared to the above which is 59.2+16.8 = 76 MB. So this extra layer adds 14.2 MB which is pretty huge for putting a single package on its own line. 

Can we just combine these? There will be some duplication in the packages but that is fine to minimize the size increase.

```diff
diff --git a/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl b/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
index 1ef1745198..046828e53f 100644
--- a/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
+++ b/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
@@ -78,14 +78,6 @@ FROM {{ .from }}

 ENV BEAT_SETUID_AS={{ .user }}

-{{- if (and (contains .from "redhat/ubi") (contains .from "-minimal")) }}
-RUN for iter in {1..10}; do \
-        microdnf update -y && \
-        microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz && \
-        microdnf clean all && \
-        exit_code=0 && break || exit_code=$? && echo "microdnf error: retry $iter in 10s" && sleep 10; done; (exit $exit_code)
-{{- end }}
-
 {{- if contains .from "wolfi" }}
 RUN for iter in {1..10}; do \
         apk fix && \
@@ -101,7 +93,13 @@ RUN for iter in {1..10}; do \
   {{- if (or (contains .Variant "basic") (contains .Variant "ubi") (contains .Variant "elastic-otel-collector") (contains .Variant "complete"))  }}
   RUN for iter in {1..10}; do \
           microdnf update -y && \
-          microdnf install -y systemd && \
+          microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz systemd && \
+          microdnf clean all && \
+          exit_code=0 && break || exit_code=$? && echo "microdnf error: retry $iter in 10s" && sleep 10; done; (exit $exit_code)
+  {{- else }}
+  RUN for iter in {1..10}; do \
+          microdnf update -y && \
+          microdnf install -y tar gzip findutils shadow-utils ca-certificates gawk libcap xz && \
           microdnf clean all && \
           exit_code=0 && break || exit_code=$? && echo "microdnf error: retry $iter in 10s" && sleep 10; done; (exit $exit_code)
   {{- end }}

```

dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl
```diff
@@ -95,6 +87,24 @@ RUN for iter in {1..10}; do \
     (exit $exit_code)
 {{- end }}
 
+# Add Systemd only to: elastic-agent, elastic-agent-complete, ubi, and otel collector.
+# Systemd is not add to any of the wolfi or slim images/variants
```
cmacknz:
```suggestion
# Systemd is not added to any of the wolfi or slim images/variants
```

testing/integration/journald_test.go
```diff
@@ -0,0 +1,227 @@
...
+				"journald": true,
+			}),
+	}
+
+	journaldTest(
+		t,
+		info.ESClient,
+		kCtx,
+		steps,
+		fmt.Sprintf("logs-%s-default", namespace),
```
cmacknz: The namespace is going where the dataset should be in the index name.

belimawr: The namespace variable there is the K8s namespace. Indeed it's confusing using it as dataset, I updated the PR to use it as the data_stream namespace.

cmacknz (approved):

testing/integration/journald_test.go
cmacknz: Just quickly unresolving to double check, you want the datastream namespaces to be different per test so that test groups (and things like docker variants) can run the same test in parallel without colliding.

If the k8s namespace here provides this isolation you are good, if not the fixture makes a Namespace that is unique per run https://github.com/elastic/elastic-agent/blob/17557993d5e7be78efcb55faef3724d862b50d77/pkg/testing/define/define.go#L57-L62

belimawr: Exactly, the K8s namespace I'm using is also unique per run/docker variant, actually every time `getNamespace` is called, a new namespace is returned.

I didn't remember at the time that `require.Info` had an unique namespace.

cmacknz (approved):

belimawr:
@cmacknz, regarding backports, I'm thinking about only backporting to 8.18 as all the other active branches have journalctl at least in the complete image, which makes this PR more of a feature addition than bug fix.

Some context: https://github.com/elastic/elastic-agent/issues/8445

What do you think?

cmacknz:
> @cmacknz, regarding backports, I'm thinking about only backporting to 8.18 as all the other active branches have journalctl at least in the complete image, which makes this PR more of a feature addition than bug fix.

I think that makes sense, if users ask about fixing it in the other versions we can consider going back even further.

Though by the same logic, shouldn't 8.18 just have it in the complete image to match 9.0 and this change only lands in 9.1 and 8.19?

belimawr:
> Though by the same logic, shouldn't 8.18 just have it in the complete image to match 9.0 and this change only lands in 9.1 and 8.19?

That makes sense.

I'll change the backport to 8.18 to only include it in the complete image. And create one to 8.19 with all other images.

Issue #7772, opened 2025-04-09
Handle `Migrate` action in agent


michalpristas:

cmacknz:
@michalpristas I think we likely want to make sure migrate actions are signed by Kibana, since if you can spoof fleet server you can now hijack an agent and re-enroll it.

I think this would be a good way to introduce action signature verification to make sure it actually works before turning it on by default in more places. The migrate action failing with signature verification errors isn't going to be fatal, but policy change or upgrade would be.

michalpristas:
i see this as a two step process not necessarily interconnected. do we have issue for re-enabling verification after it was disabled in 8.8? we can split that one and plan accordingly

cmacknz:
No we don't have a separate issue for it, I'm fine with handling it separately from the main implementation. I had not even realized this was never enabled until you started asking questions about it :)

I do think we definitely do need to use a properly signed action for this to avoid a security flaw, regardless of exactly what order it gets implemented in.

cmacknz:
https://github.com/elastic/elastic-agent/issues/8081

PR #8014, opened 2025-04-29
Migrate agent to a different cluster

michalpristas:
## What does this PR do?

This PR adds a behavior for MIGRATE action.
So far migrate action is not implemented on fleet API so testing this is difficult and cannot be automated in any way. 

Fleet server counterpart in a PR https://github.com/elastic/fleet-server/pull/4786
Fleet API implementation not started

## Why is it important?

Ability to migrate agent to different cluster

## Checklist

Fixes: #7772


internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+// Handle handles MIGRATE action.
+func (h *Migrate) Handle(ctx context.Context, a fleetapi.Action, ack acker.Acker) error {
+	h.log.Debugf("handlerMigrate: action '%+v' received", a)
+
+	action, ok := a.(*fleetapi.ActionMigrate)
+	if !ok {
+		return fmt.Errorf("invalid type, expected ActionSettings and received %T", a)
```
blakerouse:
```suggestion
		return fmt.Errorf("invalid type, expected ActionMigrate and received %T", a)
```

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,101 @@
...
+		if ucs := findMatchingUnitsByActionType(state, a.Type()); len(ucs) > 0 {
+			// tamper protected agents are unsupported, fail fast
+			err := errors.New("unsupported action: tamper protected agent")
+			h.ackFailure(ctx, err, action, ack)
+			return err
+		}
+	}
+
+	if err := h.coord.Migrate(ctx, action, fleetgateway.RequestBackoff); err != nil {
+		if errors.Is(err, coordinator.ErrNotManaged) {
```
blakerouse, line 73: This seems to just be defensive correct? How could an unmanaged agent get an action?

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,568 @@
...
+
+type saver interface {
+	Save(io.Reader) error
+}
+
+func CheckRemote(ctx context.Context, c fleetclient.Sender) error {
+	ctx, cancel := context.WithTimeout(ctx, apiStatusTimeout)
+	defer cancel()
+
+	// TODO: a HEAD should be enough as we need to test only the connectivity part
```
blakerouse: This depends if your checking that it can be reached or if you checking that it is healthy and working. A `GET` to `/api/status` is verifying that it is also healthy. I think the implementation is better than this comment, and probably should just be removed.

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,568 @@
...
+}
+
+// BackupConfig creates a backup of currently used fleet config
+func BackupConfig() error {
+	configFile := paths.AgentConfigFile()
+	backup := configFile + backupSuffix
+
+	err := copy.Copy(configFile, backup, copy.Options{
+		PermissionControl: copy.AddPermission(0600),
+	})
```
blakerouse: What happens if a copy already exists? I hope this doesn't fail, we would not want that to happen.

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
michalpristas: yeah this is just a defensive coding. i can remove that but i prefer this to be here to make it explicit

internal/pkg/agent/application/enroll/enroll.go
michalpristas: it uses os.Create so it truncates already existing content

specs/endpoint-security.spec.yml
```diff
@@ -14,6 +14,7 @@ inputs:
     proxied_actions: &proxied_actions
       - UNENROLL
       - UPGRADE
+      - MIGRATE
```
ycombinator: Just curious: why does Endpoint need to be aware of it's Agent migrating to another cluster?

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/elastic/elastic-agent/pkg/features"
+)
+
+type migrateCoordinator interface {
+	Migrate(_ context.Context, _ *fleetapi.ActionMigrate, _ func(done <-chan struct{}) backoff.Backoff) error
+	ReExec(callback reexec.ShutdownCallbackFn, argOverrides ...string)
+	State() coordinator.State
+}
+
+// Settings handles settings change coming from fleet and updates log level.
```
ycombinator: Godoc needs update.

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,101 @@
...
+type Migrate struct {
+	log       *logger.Logger
+	agentInfo info.Agent
+	coord     migrateCoordinator
+
+	tamperProtectionFn func() bool // allows to inject the flag for tests, defaults to features.TamperProtection
+}
+
+// NewSettings creates a new Settings handler.
```
ycombinator: Godoc needs update.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1601,6 +1714,48 @@ func (c *Coordinator) splitModelBetweenManagers(model *component.Model) (runtime
...
+	}
+
+	if err := acker.Commit(ctx); err != nil {
+		return fmt.Errorf("failed to commit migrate action: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Coordinator) computeEnrollOptions(ctx context.Context, cfgPath string, cfgFleetPath string) (enroll.EnrollOptions, error) {
```
ycombinator: Nothing in this method is using the receiver, `c *Coordinator`.  Could we make this is just a regular function (not a method) and unit test it?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -547,6 +559,107 @@ func (c *Coordinator) ReExec(callback reexec.ShutdownCallbackFn, argOverrides ..
...
+	}
+
+	// ACK success to source fleet server
+	if err := c.ackMigration(ctx, action, c.fleetAcker); err != nil {
+		c.logger.Infof("failed to ACK success: %v", err)
+	}
+
+	// Best effort: call unenroll on source cluster once done
+	if err := c.unenroll(ctx, originalClient); err != nil {
+		c.logger.Infof("failed to unenroll from original cluster: %v", err)
```
ycombinator: Should this be a warning?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -547,6 +559,107 @@ func (c *Coordinator) ReExec(callback reexec.ShutdownCallbackFn, argOverrides ..
...
+	// Agent enrolls to target cluster
+	err = enroll.EnrollWithBackoff(ctx, c.logger,
+		persistentConfig,
+		enrollDelay,
+		options,
+		store,
+		backoffFactory,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to enroll: %w", err)
```
ycombinator: If enrollment fails, should the config be restored?

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,567 @@
...
+	return persistentMap, nil
+}
+
+func delay(ctx context.Context, d time.Duration) {
+	t := time.NewTimer(rand.N(d))
+	defer t.Stop()
+	select {
+	case <-ctx.Done():
+	case <-t.C:
+	}
```
ycombinator: I know this code is not new, but just a result of refactoring, but there is an opportunity for a nitpicky improvement, if you wish to take it 😄.

```suggestion
	select {
	case <-ctx.Done():
	case <-time.After(rand.N(d)):
	}
```

internal/pkg/fleetapi/action.go
```diff
@@ -429,6 +433,77 @@ func (a *ActionSettings) String() string {
...
+	Insecure         bool     `json:"insecure,omitempty"`
+	ProxyDisabled    bool     `json:"proxy_disabled,omitempty"`
+	ProxyHeaders     []string `json:"proxy_headers,omitempty"`
+	ProxyURL         string   `json:"proxy_url,omitempty"`
+	Staging          string   `json:"staging,omitempty"`
+	Tags             []string `json:"tags,omitempty"`
+	ReplaceToken     string   `json:"replace_token,omitempty"`
+}
+
+func ParseMigrateSettings(raw json.RawMessage) (ActionMigrateSettings, error) {
```
ycombinator: Is this function used anywhere?

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,101 @@
...
+	h.log.Debugf("handlerMigrate: action '%+v' received", a)
+
+	action, ok := a.(*fleetapi.ActionMigrate)
+	if !ok {
+		return fmt.Errorf("invalid type, expected ActionMigrate and received %T", a)
+	}
+
+	if h.tamperProtectionFn() {
+		state := h.coord.State()
+		if ucs := findMatchingUnitsByActionType(state, a.Type()); len(ucs) > 0 {
```
cmacknz: Are we expecting that endpoint will handle the migrate action when we forward it?

michalpristas: no, but this is to figure out if we have endpoint running

specs/endpoint-security.spec.yml
michalpristas: we don't forward it but we check for presense of endpoint in `findMatchingUnitsByActionType` and then fail if endpoint is found. kinda hacky maybe(?)

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: restore is performed at the start to be sure. but i added it here as well

internal/pkg/fleetapi/action.go
michalpristas: no now :-) removed

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
blakerouse: No I am okay with it. I just wanted it to be clear. Maybe adding a comment here on why this is here and that it is only defensive would be good to add.

internal/pkg/agent/application/enroll/enroll.go
blakerouse: Okay thanks for the clarity.

specs/endpoint-security.spec.yml
blakerouse: Does it need to be here in `proxied_actions` for this to work? This field in the specification does mean that the action will be proxied to them. This seems wrong and should not be here.

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: I see you converted this to a regular function — 👍.   Could you add some unit tests for it, please? Thanks.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -547,6 +559,108 @@ func (c *Coordinator) ReExec(callback reexec.ShutdownCallbackFn, argOverrides ..
...
+	if err := enroll.CleanBackupConfig(); err != nil {
+		// when backup is present, it will be restored on next start.
+		// do not ack success
+		return fmt.Errorf("failed to clean backup config: %w", err)
+	}
+
+	// ACK success to source fleet server
+	if err := c.ackMigration(ctx, action, c.fleetAcker); err != nil {
+		c.logger.Infof("failed to ACK success: %v", err)
+	}
```
ycombinator: Would it make sense to clean the backup config only after we've ack'd?

internal/pkg/fleetapi/action.go
```diff
@@ -429,6 +433,68 @@ func (a *ActionSettings) String() string {
...
+	Headers          []string `json:"headers,omitempty"`
+	Insecure         bool     `json:"insecure,omitempty"`
+	ProxyDisabled    bool     `json:"proxy_disabled,omitempty"`
+	ProxyHeaders     []string `json:"proxy_headers,omitempty"`
+	ProxyURL         string   `json:"proxy_url,omitempty"`
+	Staging          string   `json:"staging,omitempty"`
+	Tags             []string `json:"tags,omitempty"`
+	ReplaceToken     string   `json:"replace_token,omitempty"`
+}
+
```
ycombinator: Is this type being used anywhere?

ycombinator (commented):
Left a couple more comments and unresolved a comment on the endpoint spec file because it had an unanswered question.  Also, it looks like `internal/pkg/agent/cmd/enroll_cmd.go` is now in conflict.

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: yes i have no issues with suggested ordering

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: see `func TestComputeEnrollOptions(t *testing.T)`

specs/endpoint-security.spec.yml
michalpristas: i overthought it, tried to make it generic. modified to check for endpoint

internal/pkg/fleetapi/action.go
michalpristas: nope, removed

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,574 @@
...
+	statusPath                     = "/api/status"
+	apiStatusTimeout               = 15 * time.Second
+	backupSuffix                   = ".enroll.bak"
+)
+
+type saver interface {
+	Save(io.Reader) error
+}
+
+func CheckRemote(ctx context.Context, c fleetclient.Sender) error {
```
ycombinator: Nit: Could we move this out of the `enroll` package, seeing that it doesn't do anything specific to enrollment and also that it's called from the policy change handler code?

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,109 @@
...
+			if component.Component.InputType == endpoint {
+				err := errors.New("unsupported action: tamper protected agent")
+				h.ackFailure(ctx, err, action, ack)
+				return err
+			}
+		}
+	}
+
+	if err := h.coord.Migrate(ctx, action, fleetgateway.RequestBackoff); err != nil {
+		// this should not happen, not managed agent should not receive the action
```
ycombinator: Nit:
```suggestion
		// this should not happen, unmanaged agent should not receive the action
```

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,109 @@
...
+		// defensive coding to avoid misbehavior
+		if errors.Is(err, coordinator.ErrNotManaged) {
+			return errors.New("unmanaged agent, use Enroll instead")
+		}
+
```
ycombinator, line 76: Could we add another `errors.Is` check here for `ErrFleetServer` and specify in the returned error that Fleet Server cannot be migrated?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -547,6 +559,108 @@ func (c *Coordinator) ReExec(callback reexec.ShutdownCallbackFn, argOverrides ..
...
+	originalRemoteConfig, err := options.RemoteConfig(false)
+	if err != nil {
+		return fmt.Errorf("failed to construct original remote config: %w", err)
+	}
+
+	originalClient, err := fleetapiClient.NewAuthWithConfig(
+		c.logger, options.EnrollAPIKey, originalRemoteConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create original fleet client: %w", err)
+	}
```
ycombinator: Nit: could we move this block down to just before `originalClient` is used?  Makes it a bit easier to follow the code and also avoids executing this code if something else fails before it.  I realize that means renaming `options` to something like `originalOptions` and `newOptions` but I actually think that's even more readable as it fits the rest of the pattern.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -547,6 +559,108 @@ func (c *Coordinator) ReExec(callback reexec.ShutdownCallbackFn, argOverrides ..
...
+		backoffFactory,
+	)
+	if err != nil {
+		restoreErr := enroll.RestoreConfig()
+		return errors.Join(fmt.Errorf("failed to enroll: %w", err), restoreErr)
+	}
+
+	// ACK success to source fleet server
+	if err := c.ackMigration(ctx, action, c.fleetAcker); err != nil {
+		c.logger.Infof("failed to ACK success: %v", err)
```
ycombinator: Should this be an `Errorf` or at least `Warnf`?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -547,6 +559,108 @@ func (c *Coordinator) ReExec(callback reexec.ShutdownCallbackFn, argOverrides ..
...
+	}
+
+	if err := enroll.CleanBackupConfig(); err != nil {
+		// when backup is present, it will be restored on next start.
+		// do not ack success
+		return fmt.Errorf("failed to clean backup config: %w", err)
+	}
```
ycombinator: Would it make sense to move this block below the best-effort unenroll block since that one does not error out and it gives us a chance to unenroll even if we fail to clean up the backup config?

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -0,0 +1,109 @@
...
+	if h.tamperProtectionFn() {
+		state := h.coord.State()
+		// if endpoint is present do not proceed
+		for _, component := range state.Components {
+			if component.Component.InputType == endpoint {
+				err := errors.New("unsupported action: tamper protected agent")
+				h.ackFailure(ctx, err, action, ack)
+				return err
+			}
+		}
```
ycombinator: Elsewhere in this PR, you introduce a `func (c *Coordinator) isFleetServer() bool` function.  I think it would be nice to introduce a similar `func (c *Coordinator) HasEndpoint() bool` function and call it from here.  I could see that function being useful in other places too.  It also makes the check here much more concise and readable, something like:

```golang
if h.tamperProtectionFn() && h.coord.HasEndpoint() {
  // return error
}
```

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,574 @@
...
+
+	// discard body for proper cancellation and connection reuse
+	_, _ = io.Copy(io.Discard, resp.Body)
+	resp.Body.Close()
+
+	return nil
+}
+
+// BackupConfig creates a backup of currently used fleet config
+func BackupConfig() error {
```
ycombinator: Is there any reason for this function to be in the `enroll` package and exported?  AFAICT, it's only being called from the coordinator, so perhaps it could live as an unexported function in the `coordinator` package?

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,574 @@
...
+		PermissionControl: copy.AddPermission(0600),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to backup config file %s -> %s: %w", configFile, backup, err)
+	}
+
+	return nil
+}
+
+// RestoreConfig restores from backup if needed and signals restore was performed
```
ycombinator: Is there any reason for this function to be in the `enroll` package  AFAICT, it's  called from the coordinator and cmd/run. Perhaps it could live in the `coordinator` package?

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -0,0 +1,574 @@
...
+
+	if err := file.SafeFileRotate(configFile, backup); err != nil {
+		return fmt.Errorf("failed to safe rotate backup config file: %w", err)
+	}
+
+	return nil
+}
+
+// CleanBackupConfig removes backup config file
+func CleanBackupConfig() error {
```
ycombinator: Is there any reason for this function to be in the `enroll` package and exported?  AFAICT, it's only being called from the coordinator, so perhaps it could live as an unexported function in the `coordinator` package?

ycombinator (commented):
Left a few comments but they're around code organization and readability more than anything else.

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: yes, warn level seems right

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: if we fail to cleanup backup config it may happen that agent will restart and recover from backup, in case it is unenrolled, there's no way to retry action or spot this in UI

PR #8009, opened 2025-04-28
[otel]: Add e2e test for monitoring metrics in otel mode

khushijain21:
## What does this PR do?
This PR adds e2e tests for self-monitoring metrics exposed using beatreceivers. It also asserts document equivalency for metrics exposed by normal mode vs otel mode. 

## Why is it important?
Required to safely transition running elastic-agent in otel mode.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/4877
- Closes https://github.com/elastic/ingest-dev/issues/4879

PR #8031, opened 2025-04-29
Allow using beats receivers for self-monitoring

swiatekm:
## What does this PR do?

Adds the ability to use beats receivers for agent self-monitoring. To do so, we add a new configuration key to `agent.monitoring` named `_runtime_experimental` - identical to how you can currently switch inputs to the Otel runtime.

In terms of implementation, the changes are very straightforward. In the monitoring injection manager, we set the runtime manager for inputs we add, if it's set in the monitoring configuration.

Most of this PR's code changes lie in tests, and more specifically in the `TestAgentMonitoring` E2E test. This test compares the data collected by agent self-monitoring using beats processes to an equivalent Otel configuration of beats receivers in Hybrid mode. Instead of doing that, we can now just change `agent.monitoring._runtime_experimental`, so the test becomes much simpler conceptually.

I have simplified some of the test logic, but I haven't yet made it compare metrics. This should be doable now, but we have another PR (https://github.com/elastic/elastic-agent/pull/8009 ) in-flight doing it, so I held off.

## Why is it important?

We want to be able to use beats receivers for agent self-monitoring.

## Checklist

Build the agent locally and use the following configuration:

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [...]
    username: elastic
    password: "..."

inputs: []

agent:
  monitoring:
    metrics: true
    logs: true
    _runtime_experimental: otel
```

Looking at Kibana dashboards for the agent integration can prove the data is actually being ingested. You can verify that beats receivers are being used for self-monitoring by looking at their CPU usage - it should be 0.

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/3517

### DIFF
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -415,19 +415,19 @@ func (b *BeatsMonitor) getComponentInfos(components []component.Component, compo
 			componentInfo{
 				ID:             fmt.Sprintf("beat/%s", monitoringMetricsUnitID),
 				BinaryName:     metricBeatName,
-				RuntimeManager: component.DefaultRuntimeManager,
+				RuntimeManager: component.RuntimeManager(b.config.C.RuntimeManager),
 			},
 			componentInfo{
 				ID:             fmt.Sprintf("http/%s", monitoringMetricsUnitID),
 				BinaryName:     metricBeatName,
-				RuntimeManager: component.DefaultRuntimeManager,
+				RuntimeManager: component.RuntimeManager(b.config.C.RuntimeManager),
 			})
 	}
 	if b.config.C.MonitorLogs {
 		componentInfos = append(componentInfos, componentInfo{
 			ID:             monitoringFilesUnitsID,
 			BinaryName:     fileBeatName,
-			RuntimeManager: component.DefaultRuntimeManager,
+			RuntimeManager: component.RuntimeManager(b.config.C.RuntimeManager),
 		})
 	}
 	// sort the components to ensure a consistent order of inputs in the configuration
@@ -445,15 +445,19 @@ func (b *BeatsMonitor) injectLogsInput(cfg map[string]interface{}, componentInfo
 
 	streams = append(streams, b.getServiceComponentFilestreamStreams(componentInfos)...)
 
-	inputs := []interface{}{
-		map[string]interface{}{
-			idKey:        fmt.Sprintf("%s-agent", monitoringFilesUnitsID),
-			"name":       fmt.Sprintf("%s-agent", monitoringFilesUnitsID),
-			"type":       "filestream",
-			useOutputKey: monitoringOutput,
-			"streams":    streams,
-		},
+	input := map[string]interface{}{
+		idKey:        fmt.Sprintf("%s-agent", monitoringFilesUnitsID),
+		"name":       fmt.Sprintf("%s-agent", monitoringFilesUnitsID),
+		"type":       "filestream",
+		useOutputKey: monitoringOutput,
+		"streams":    streams,
 	}
+	// Make sure we don't set anything until the configuration is stable if the otel manager isn't enabled
+	if b.config.C.RuntimeManager != monitoringCfg.DefaultRuntimeManager {
+		input["_runtime_experimental"] = b.config.C.RuntimeManager
+	}
+
+	inputs := []any{input}
 	inputsNode, found := cfg[inputsKey]
 	if !found {
 		return fmt.Errorf("no inputs in config")
@@ -517,6 +521,14 @@ func (b *BeatsMonitor) injectMetricsInput(
 			},
 			"streams": httpStreams,
 		},
+	}
+
+	// Make sure we don't set anything until the configuration is stable if the otel manager isn't enabled
+	if b.config.C.RuntimeManager != monitoringCfg.DefaultRuntimeManager {
+		for _, input := range inputs {
+			inputMap := input.(map[string]interface{})
+			inputMap["_runtime_experimental"] = b.config.C.RuntimeManager
+		}
 	}
 
 	// add system/process metrics for services that can't be monitored via json/beats metrics
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -203,7 +203,6 @@ func (b *BeatsMonitor) MonitoringConfig(
 							}
 							uintPolicyValue := uint(parsedPolicyValue)
 							failureThreshold = &uintPolicyValue
-						case nil: // no change
 						default:
 							return nil, fmt.Errorf("unsupported type for policy failure threshold: %T", policyFailureThresholdRaw)
 						}
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -203,6 +203,7 @@ func (b *BeatsMonitor) MonitoringConfig(
 							}
 							uintPolicyValue := uint(parsedPolicyValue)
 							failureThreshold = &uintPolicyValue
+						case nil: // no change
 						default:
 							return nil, fmt.Errorf("unsupported type for policy failure threshold: %T", policyFailureThresholdRaw)
 						}
--- a/internal/pkg/core/monitoring/config/config.go
+++ b/internal/pkg/core/monitoring/config/config.go
@@ -16,7 +16,10 @@ const (
 	defaultNamespace = "default"
 
 	// DefaultHost is used when host is not defined or empty
-	DefaultHost = "localhost"
+	DefaultHost           = "localhost"
+	ProcessRuntimeManager = "process"
+	OtelRuntimeManager    = "otel"
+	DefaultRuntimeManager = ProcessRuntimeManager
 )
 
 // MonitoringConfig describes a configuration of a monitoring
@@ -33,6 +36,7 @@ type MonitoringConfig struct {
 	MonitorTraces    bool                  `yaml:"traces" config:"traces"`
 	APM              APMConfig             `yaml:"apm,omitempty" config:"apm,omitempty" json:"apm,omitempty"`
 	Diagnostics      Diagnostics           `yaml:"diagnostics,omitempty" json:"diagnostics,omitempty"`
+	RuntimeManager   string                `yaml:"_runtime_experimental,omitempty" config:"_runtime_experimental,omitempty"`
 }
 
 // MonitoringHTTPConfig is a config defining HTTP endpoint published by agent
@@ -118,9 +122,10 @@ func DefaultConfig() *MonitoringConfig {
 			Host:    DefaultHost,
 			Port:    defaultPort,
 		},
-		Namespace:   defaultNamespace,
-		APM:         defaultAPMConfig(),
-		Diagnostics: defaultDiagnostics(),
+		Namespace:      defaultNamespace,
+		APM:            defaultAPMConfig(),
+		Diagnostics:    defaultDiagnostics(),
+		RuntimeManager: DefaultRuntimeManager,
 	}
 }
 
--- a/testing/integration/beat_receivers_test.go
+++ b/testing/integration/beat_receivers_test.go
@@ -17,13 +17,14 @@ import (
 
 	"gopkg.in/yaml.v2"
 
+	"github.com/gofrs/uuid/v5"
+	"github.com/stretchr/testify/require"
+
 	"github.com/elastic/elastic-agent-libs/kibana"
 	"github.com/elastic/elastic-agent-libs/testing/estools"
 	atesting "github.com/elastic/elastic-agent/pkg/testing"
 	"github.com/elastic/elastic-agent/pkg/testing/define"
 	"github.com/elastic/elastic-agent/pkg/testing/tools/testcontext"
-	"github.com/gofrs/uuid/v5"
-	"github.com/stretchr/testify/require"
 )
 
 var (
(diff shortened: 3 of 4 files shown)


testing/integration/beat_receivers_test.go
```diff
@@ -79,73 +75,77 @@ func TestAgentMonitoring(t *testing.T) {
...
+	policyResponse, err := info.KibanaClient.CreatePolicy(policyCtx, createPolicyReq)
+	require.NoError(t, err, "error creating policy")
+
+	// 2. Download the policy, add the API key
+	downloadURL := fmt.Sprintf("/api/fleet/agent_policies/%s/download", policyResponse.ID)
+	resp, err := info.KibanaClient.Connection.SendWithContext(policyCtx, http.MethodGet, downloadURL, nil, nil, nil)
+	require.NoError(t, err, "error downloading policy")
+	policyBytes, err := io.ReadAll(resp.Body)
+	require.NoError(t, err, "error reading policy response")
+	defer resp.Body.Close()
```
ycombinator, line 101: Could we extract this + the `PolicyStruct` type + the parsing of `policyBytes` into a `PolicyStruct` below into https://github.com/elastic/elastic-agent-libs/tree/main/kibana so the code here is simply making a call to something like `info.KibanaClient.GetPolicyDownload(policyResponse.ID)`?

ycombinator: Never mind, while the above would be desirable, I see now that you are just moving this code around from what it was already.

internal/pkg/core/monitoring/config/config.go
```diff
@@ -33,6 +36,7 @@ type MonitoringConfig struct {
 	MonitorTraces    bool                  `yaml:"traces" config:"traces"`
 	APM              APMConfig             `yaml:"apm,omitempty" config:"apm,omitempty" json:"apm,omitempty"`
 	Diagnostics      Diagnostics           `yaml:"diagnostics,omitempty" json:"diagnostics,omitempty"`
+	RuntimeManager   string                `yaml:"_runtime_experimental,omitempty" config:"_runtime_experimental,omitempty"`
```
ycombinator, line 39: Since this is more of an enumeration that supports only a finite set of values (currently two), I'd prefer if we used a custom type here instead of `string`.  Then we could add validation on this type when a string is being unmarshalled to it, and tests around the unmarshalling.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -203,6 +203,7 @@ func (b *BeatsMonitor) MonitoringConfig(
 							}
 							uintPolicyValue := uint(parsedPolicyValue)
 							failureThreshold = &uintPolicyValue
+						case nil: // no change
```
ycombinator, line 206: I'm not following why this case is necessary.  Perhaps adding a test case for it to `TestMonitoringConfigMetricsFailureThreshold` might help me understand when the failureThreshold's type can be `nil`?

internal/pkg/core/monitoring/config/config.go
ycombinator: Could we perhaps use the `component.RuntimeManager` type here?

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -444,15 +445,19 @@ func (b *BeatsMonitor) injectLogsInput(cfg map[string]interface{}, componentInfo
...
-		},
+	input := map[string]interface{}{
+		idKey:        fmt.Sprintf("%s-agent", monitoringFilesUnitsID),
+		"name":       fmt.Sprintf("%s-agent", monitoringFilesUnitsID),
+		"type":       "filestream",
+		useOutputKey: monitoringOutput,
+		"streams":    streams,
+	}
+	// Make sure we don't set anything until the configuration is stable if the otel manager isn't enabled
+	if b.config.C.RuntimeManager != monitoringCfg.DefaultRuntimeManager {
```
ycombinator, line 456: Nit: if we go with https://github.com/elastic/elastic-agent/pull/8031/files#r2070697456, we could abstract this logic into something like:
```
if b.config.C.RuntimeManager.IsDefault() {
...
}
```

internal/pkg/agent/application/monitoring/v1_monitor.go
cmacknz: This PR didn't create all this, but overall the amount of code here required to read a number from a YAML configuration seems excessive overall.

testing/integration/beat_receivers_test.go
```diff
@@ -225,198 +225,50 @@ func TestAgentMonitoring(t *testing.T) {
...
 
 		fmt.Println(time.Now())
-		// Restart elastic-agent
-		output, err = fixture.Exec(ctx, []string{"restart"})
-		require.NoErrorf(t, err, "error restarting agent: %s\ncombinedoutput:\n%s", err, string(output))
 
+		output, err := fixture.InstallWithoutEnroll(ctx, &installOpts)
+		require.NoErrorf(t, err, "error install withouth enroll: %s\ncombinedoutput:\n%s", err, string(output))
+
+		// Ensure elastic-agent is healthy, otherwise we cannot perform restart operation
```
cmacknz, line 263: You should probably ensure that only the monitoring receivers are running, and not both.

Also, the implementation of `fixture.IsHealthy` doesn't understand the OTel status yet, so if the monitoring receivers were failing would this check even catch that?

testing/integration/beat_receivers_test.go
```diff
@@ -457,22 +309,19 @@ service:
 			// Expected to change between agentDocs and OtelDocs
 			"@timestamp",
 			"agent.ephemeral_id",
+			// agent.id is different because it's the id of the underlying beat
```
cmacknz, line 312: We need to look for something in the data to be able to tell these events actually came from a receiver. We will want that when we push this out into the world as well so we can tell which thing wrote the data.

internal/pkg/core/monitoring/config/config.go
swiatekm: I originally intended to, but https://github.com/elastic/go-ucfg/issues/169 makes doing to impossible.

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: I can remove this change. I ran into a problem trying to manipulate this configuration in a E2E test, where defaulting from unmarshalling would set this value to `nil`. Since that just means the value is unchanged from the base config, I added this case.

Relatedly, I agree with Craig that we should just unmarshal this map into a typed struct, either via go-ucfg or mapstructure. It's on my list as a follow-up to the existing monitoring injection changes.

testing/integration/beat_receivers_test.go
khushijain21: > You should probably ensure that only the monitoring receivers are running, and not both.

This is ensured. Under this sub-test, we start elastic-agent with an empty configuration first - to retrieve the path where logs will be stored - and then write the config, and restart the agent. 

We start the agent with monitoring receivers only. 

> Also, the implementation of fixture.IsHealthy doesn't understand the OTel status yet, so if the monitoring receivers were failing would this check even catch that?

This check is intended to check the health status of  EA started with an empty config. This check catches that

testing/integration/beat_receivers_test.go
swiatekm: Honestly, I'm not convinced any of the Kibana policy stuff in this test is even necessary vs just explicitly writing the standalone agent config we want. But like you said, I'm just moving things around here.

testing/integration/beat_receivers_test.go
swiatekm: Hm, good point, that's something I can fix relatively easily.

khushijain21: Also, the comment is stale due to the new changes. 
We never perform the restart operation after this.

testing/integration/beat_receivers_test.go
swiatekm: > events actually came from a receiver.

You mean, were collected by a receiver, right? Distinguishing between telemetry emitted by a receiver vs a beats process should be done in beats receivers themselves imo. It's not really an elastic-agent concern.

I think we'd want to add a new attribute, given that we want to ensure maximum compatibility with existing content. `agent.runtime`, maybe? `agent.*` is already used for attributes related to the agent collecting the data. And we can add it easily to the existing self-monitoring machinery.

cmacknz: You are correct, adding metadata for the source of the data should in the end be done in the receiver. 

My worry is that this test could still pass if you didn't implement your change or there was a bug, I think there is a simpler change you could make in the test itself to solve this.

When you switch agent to use beat receivers, if you also change the monitoring data stream namespace, then the new data stream should never contain data from the monitoring Beat sub-processes and this concern goes away. 

https://github.com/elastic/elastic-agent/blob/815c264cddc9ed27743b526661df9f51aeb88cbd/internal/pkg/core/monitoring/config/config.go#L31

You could set these tests up to write to `metrics-elastic_agent.*-default` (or maybe instead of `default` use `processes`) and `metrics-elastic_agent.*-receivers` for the two different configurations. It might be easier to do this if you stopped agent to reconfigure it instead of doing it as a config reload, so that you can control the exact order (e.g. there is no time window where the sub-processes have the configuration to write to the `receivers` namespace).

cmacknz: That and a check that the Beat sub-processes are no longer running after changing the runtime mode as suggested in https://github.com/elastic/elastic-agent/pull/8031#discussion_r2070721230 should give us confidence this test can't pass because the query then can't pick up documents from sub-processes. We guarantee that when receivers are running, they write to a data stream that is exclusive of the sub-processes.

testing/integration/beat_receivers_test.go
swiatekm: Added the status checks in https://github.com/elastic/elastic-agent/pull/8031/commits/183ff063be1f7216f3be75c89de842cfbaee796f.

testing/integration/beat_receivers_test.go
swiatekm: I added the status checks in https://github.com/elastic/elastic-agent/commit/183ff063be1f7216f3be75c89de842cfbaee796f and separated the namespaces in https://github.com/elastic/elastic-agent/pull/8031/commits/3b2fe0010f4075f5dd47fad96a4b2c1dc5a97f52. The whole test is somewhat messy right now, but I don't want to make this PR too onerous to review by doing it here.

testing/integration/beat_receivers_test.go
```diff
@@ -225,198 +238,68 @@ func TestAgentMonitoring(t *testing.T) {
...
+			httpMetricsPipeline := "pipeline:logs/_agent-component/http/metrics-monitoring"
+			beatsMetricsPipeline := "pipeline:logs/_agent-component/beat/metrics-monitoring"
+			assert.Contains(collect, pipelineStatusMap, fileStreamPipeline)
+			assert.Contains(collect, pipelineStatusMap, httpMetricsPipeline)
+			assert.Contains(collect, pipelineStatusMap, beatsMetricsPipeline)
+
+			// and they should be healthy
+			assert.Equal(collect, int(cproto.CollectorComponentStatus_StatusOK), pipelineStatusMap[fileStreamPipeline].Status)
+			assert.Equal(collect, int(cproto.CollectorComponentStatus_StatusOK), pipelineStatusMap[httpMetricsPipeline].Status)
+			assert.Equal(collect, int(cproto.CollectorComponentStatus_StatusOK), pipelineStatusMap[beatsMetricsPipeline].Status)
```
cmacknz, line 299: Once we've made it to this point, can we also make sure that none of the sub-process monitoring components are there as well?

cmacknz:
This LGTM and testing locally I see the monitoring receivers running. It would be nice if we can make sure that only the beat receivers are used for monitoring, right now we have at least the beat receivers are used for monitoring.

```sh
sudo elastic-development-agent status --output=full
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: 9293b312-f874-4866-bb1c-ebc21244c75c
   │  ├─ version: 9.1.0
   │  └─ commit: 3b2fe0010f4075f5dd47fad96a4b2c1dc5a97f52
   ├─ filestream-default
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '82070'
   │  ├─ filestream-default
   │  │  └─ type: OUTPUT
   │  └─ filestream-default-your-input-id
   │     └─ type: INPUT
   ├─ system/metrics-default
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '82069'
   │  ├─ system/metrics-default
   │  │  └─ type: OUTPUT
   │     └─ type: INPUT
   ├─ pipeline:logs/_agent-component/beat/metrics-monitoring
   │  ├─ status: StatusOK
   │  ├─ exporter:elasticsearch/_agent-component/monitoring
   │  │  └─ status: StatusOK
   │  └─ receiver:metricbeatreceiver/_agent-component/beat/metrics-monitoring
   │     └─ status: StatusOK
   ├─ pipeline:logs/_agent-component/filestream-monitoring
   │  ├─ status: StatusOK
   │  ├─ exporter:elasticsearch/_agent-component/monitoring
   │  │  └─ status: StatusOK
   │  └─ receiver:filebeatreceiver/_agent-component/filestream-monitoring
   │     └─ status: StatusOK
   └─ pipeline:logs/_agent-component/http/metrics-monitoring
      ├─ status: StatusOK
      ├─ exporter:elasticsearch/_agent-component/monitoring
      │  └─ status: StatusOK
      └─ receiver:metricbeatreceiver/_agent-component/http/metrics-monitoring
         └─ status: StatusOK
```

testing/integration/beat_receivers_test.go
swiatekm: Probably easier to just check the whole map recursively.

testing/integration/beat_receivers_test.go
```diff
@@ -225,198 +238,66 @@ func TestAgentMonitoring(t *testing.T) {
...
-		require.Eventually(t, func() bool {
-			err = fixture.IsHealthy(ctx)
-			if err != nil {
-				t.Logf("waiting for agent healthy: %s", err.Error())
-				return false
-			}
-			return true
-		}, 30*time.Second, 1*time.Second)
+		output, err := fixture.InstallWithoutEnroll(ctx, &installOpts)
+		require.NoErrorf(t, err, "error install withouth enroll: %s\ncombinedoutput:\n%s", err, string(output))
```
khushijain21, line 271: [nit]
```suggestion
		require.NoErrorf(t, err, "error install without enroll: %s\ncombinedoutput:\n%s", err, string(output))
```

ycombinator (approved):

khushijain21 (approved):

PR #8035, opened 2025-04-29
Add Pipeline to deploy custom agent image for FIPS testing

michel-laterman:
## What does this PR do?

Add a new buildkite pipeline to build a custom agent image and use it in an ECH deployment for testing.

## Why is it important?

FIPS integration tests will require a custom agent running in the CFT region.

## Checklist

## Disruptive User Impact

N/A


michel-laterman:
buildkite test this

michel-laterman:
@v1v @pazone, can you take a look at this, we need permissions in order to push a custom image to use in the CFT region 
```
denied: requested access to the resource is denied
--
  | Error: Failed pushing docker image: running "docker image push docker.elastic.co/observability-ci/elastic-agent-fips:git-b84b80343415" failed with exit code 1
```

v1v:
> | Error: Failed pushing docker image: running "docker image push docker.elastic.co/observability-ci/elastic-agent-fips:git-b84b80343415" failed with exit code 1

I'm not much familiar with the current user and namespace, as far as I see https://github.com/elastic/elastic-agent/blob/aa224536eadf49f8b9b962df240c0caa4861970e/.buildkite/hooks/pre-command#L17 is where those details are stored. 

However, I think you need to configure the `pre-command` hook to run for the new BK pipelines:
- https://github.com/elastic/elastic-agent/blob/aa224536eadf49f8b9b962df240c0caa4861970e/.buildkite/hooks/pre-command#L39-L47 is the settings for the `elastic-agent-package` BK pipeline.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+          #- fleet
+          #- fleet-endpoint-security
+          #- fleet-airgapped
+          #- fleet-airgapped-privileged
+          #- fleet-privileged
+          #- fleet-upgrade-to-pr-build
+          #- install-uninstall
+          #- fqdn
+          #- deb
+          #- container
```
ycombinator, line 101: Should these be uncommented?

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+  #          variants:
+  #          - "basic,slim,complete,service,elastic-otel-collector"
+  #          - "wolfi,slim-wolfi,complete-wolfi,elastic-otel-collector-wolfi"
+  #          version:
+  #          - v1.27.16
+  #          - v1.28.9
+  #          - v1.29.8
+  #          - v1.30.8
+  #          - v1.31.0
+  #          - v1.32.0
```
ycombinator, line 142: Should this commented out block be removed? If not, maybe add a comment on top of the block explaining why it's commented out for now?

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+      #!/usr/bin/env bash
+      set -euo pipefail
+      mage cloud:image
+      mage cloud:push
+    agents:
+      provider: "gcp"
+      machineType: "n1-standard-8"
+      image: "${IMAGE_UBUNTU_2404_X86_64}"
+
+  - label: Start ESS stack for integration tests
```
ycombinator, line 29: Let's mention FIPS or FRH in here.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+  #          - "wolfi,slim-wolfi,complete-wolfi,elastic-otel-collector-wolfi"
+  #          version:
+  #          - v1.27.16
+  #          - v1.28.9
+  #          - v1.29.8
+  #          - v1.30.8
+  #          - v1.31.0
+  #          - v1.32.0
+
+  - label: ESS stack cleanup
```
ycombinator, line 144: Let's mention FIPS or FRH in here.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+    depends_on:
+      - integration-tests-ubuntu
```
ycombinator, line 146: Should this be `integration-tests-ubuntu-fips`?

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+    agents:
+      image: "docker.elastic.co/ci-agent-images/platform-ingest/buildkite-agent-beats-ci-with-hooks:0.5"
+      useCustomGlobalHooks: true
+
+  - label: Aggregate test reports
+    # Warning: The key has a hook in pre-command
+    key: aggregate-reports-fips
+    depends_on:
+      - integration-tests-ubuntu-fips
+      #- integration-tests-kubernetes
```
ycombinator, line 164: Should this be removed?

.buildkite/integration.pipeline.yml
```diff
@@ -201,3 +201,8 @@ steps:
     depends_on:
       - int-packaging
     command: "buildkite-agent pipeline upload .buildkite/bk.integration.pipeline.yml"
+
+  - label: "Triggering custom ECH integration tests"
```
ycombinator, line 205: Instead of ECH let's use FIPS or FRH to be consistent?  Maybe we should rename the pipeline file to be consistent as well?

magefile.go
```diff
@@ -1001,7 +1001,19 @@ func (Cloud) Push() error {
...
 	}
 
+	fips := os.Getenv(fipsEnv)
+	defer os.Setenv(fipsEnv, fips)
+	fipsVal, err := strconv.ParseBool(fips)
+	if err != nil {
+		fipsVal = false
+	}
+	os.Setenv(fipsEnv, strconv.FormatBool(fipsVal))
+	devtools.FIPSBuild = fipsVal
```
ycombinator, line 1011: I think it would be helpful to add a comment here what this section is doing and why.

ycombinator (commented):
Left a few suggestions but overall this looks good.

.buildkite/bk.fips-integration.pipeline.yml
michel-laterman: I haven't looked into our k8s tests, we would need to pass the FIPS image we create as the `DOCKER_VARIANTS` arg

magefile.go
ycombinator: I see you added a comment. Thanks.  Could you explain in there why you have the `defer os.Setenv(fipsEnv, fips)`?  That bit in particular wasn't obvious to me.

michel-laterman: It's copied from how we handle the env var in other parts of the magefile

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,217 @@
...
+          image: "${IMAGE_UBUNTU_2404_ARM_64}"
+          instanceType: "m6g.2xlarge"
+        matrix:
+          - default
+          - upgrade
+          - upgrade-flavor
+          - standalone-upgrade
+          - fleet
+
+  - group: "fips:Kubernetes"
```
pkoutsovasilis, line 147: fips kubernetes??? are we sure???

michel-laterman: not sure, added for completeness (let's see if it works, or if all tests will skip)

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,186 @@
...
+      useCustomGlobalHooks: true
+
+  - group: "fips:Stateful:Ubuntu"
+    key: integration-tests-ubuntu-fips
+    depends_on:
+      - integration-fips-ess
+    steps:
+      - label: "fips:non-sudo:{{matrix}}"
+        depends_on:
+          - packaging-ubuntu-x86-64-fips
```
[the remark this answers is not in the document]
michel-laterman: Added arm

.buildkite/bk.fips-integration.pipeline.yml
ycombinator: @pkoutsovasilis  I think it will make it easier to distinguish these tests from the regular K8s tests in Agent CI.  We can be more descriptive if the name is confusing.

pkoutsovasilis: the name is not confusing, the actual testcase is what confuses me a bit. However, since this is deliberate then no objection gents, go on 🙂

ycombinator:
CI, specifically the `fips:Kubernetes` step is failing like so:

```
.buildkite/scripts/steps/integration_tests_tf.sh: line 10: asdf: command not found
```

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,217 @@
...
+        matrix:
+          - default
+          - upgrade
+          - upgrade-flavor
+          - standalone-upgrade
+          - fleet
+
+      - label: "fips:arm64:non-sudo:{{matrix}}"
+        depends_on:
+          - packaging-ubuntu-arm64-fips
```
ycombinator, line 103: I don't think this step actually exists yet?

ycombinator:
CI, specifically the `fips:Stateful:Ubuntu` (only the `x86_64` ones) and `fips:Kubernetes` steps are failing like so:

```
.buildkite/scripts/steps/integration_tests_tf.sh: line 10: asdf: command not found
```

Looks like the `platform-ingest-fleet-server-ubuntu-2204-fips` VM image is missing `asdf`.  Here is the PR to add it: https://github.com/elastic/ci-agent-images/pull/1431.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,217 @@
...
+  VAULT_PATH: "kv/ci-shared/observability-ingest/cloud/gcp"
+  ASDF_MAGE_VERSION: 1.14.0
+  FIPS: "true"
+  CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+  CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
+
+  IMAGE_UBUNTU_2404_ARM_64: "platform-ingest-elastic-agent-ubuntu-2404-aarch64-1744855248"
+  IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1744855248"
+  IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-fleet-server-ubuntu-2204-fips"
+  IMAGE_UBUNTU_ARM64_FIPS: "" # TODO We need an ARM image as well
```
ycombinator, line 14: https://github.com/elastic/ingest-dev/issues/5264#issuecomment-2868162957

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,217 @@
...
+      set -euo pipefail
+      source .buildkite/scripts/steps/ess_start.sh
+    artifact_paths:
+      - test_infra/ess/*.tfstate
+      - test_infra/ess/*.lock.hcl
+    agents:
+      image: "docker.elastic.co/ci-agent-images/platform-ingest/buildkite-agent-beats-ci-with-hooks:0.5"
+      useCustomGlobalHooks: true
+
+  - group: "fips:Stateful:Ubuntu"
```
pazone, line 49: Do I understand correctly that we run a set integration test groups in the same way and the only difference is in the VM image?

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,217 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  DOCKER_REGISTRY: "docker.elastic.co"
+  VAULT_PATH: "kv/ci-shared/observability-ingest/cloud/gcp"
+  ASDF_MAGE_VERSION: 1.14.0
+  FIPS: "true"
```
pazone, line 7: I'd recommend reducing the scope of this variable at least to the step level. Currently, it's hard to understand what step is affected by this variable

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,217 @@
...
+  CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+  CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
+
+  IMAGE_UBUNTU_2404_ARM_64: "platform-ingest-elastic-agent-ubuntu-2404-aarch64-1744855248"
+  IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1744855248"
+  IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-fleet-server-ubuntu-2204-fips"
```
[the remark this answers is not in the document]
michel-laterman: I think that the images were created as a result of some experimentation that was being done with the fleet-server repo

.buildkite/bk.fips-integration.pipeline.yml
michel-laterman: VM image, and i think that `FIPS=true` results in `-integration.fips=true` being sent

.buildkite/bk.fips-integration.pipeline.yml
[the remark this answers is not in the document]
pazone: I will create FIPS-compliant images for elastic-agent to avoid possible problems.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,225 @@
...
+steps:
+  - label: Build and push custom elastic-agent image
+    key: integration-fips-cloud-image
+    env:
+      ASDF_TERRAFORM_VERSION: 1.9.2
+      FIPS: "true"
+    command: |
+      #!/usr/bin/env bash
+      set -euo pipefail
+      mage cloud:image
```
pazone, line 23: `cloud:image` invokes the [Package()](https://github.com/elastic/elastic-agent/blob/main/magefile.go#L982) function that packages the agent again. This step takes a considerable amount of time. Can we download the artifacts produced by the packaging step and reuse them?

michel-laterman: We can try, but I'm not sure if the Image step will use artifacts that are already present, @pkoutsovasilis do you know if it will?

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,187 @@
...
+      machineType: "n1-standard-8"
+      image: "${IMAGE_UBUNTU_2404_X86_64}"
+
+  - label: Start ESS stack for FIPS integration tests
+    key: integration-fips-ess
+    depends_on:
+      - integration-fips-cloud-image
+    env:
+      ASDF_TERRAFORM_VERSION: 1.9.2
+      TF_VAR_integration_server_docker_image: "${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}"
```
pazone: Important: We need to add this variable to all test steps in the pipeline. When a step is retried, it triggers the creation of a new, dedicated deployment. The `TF_VAR_integration_server_docker_image` variable must be present in every step to ensure consistency.

michel-laterman: Moved it to the pipeline var declarations.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,186 @@
...
+          .buildkite/scripts/steps/integration_tests_tf.sh {{matrix}} true
+        artifact_paths:
+          - build/**
+          - build/diagnostics/**
+        retry:
+          automatic:
+            limit: 1
+        agents:
+          provider: "aws"
+          imageName: "${IMAGE_UBUNTU_X86_64_FIPS}"
```
pazone: `image` is the correct parameter name

pazone:
* The correct integrations-server image is passed to the initial deployment, and for step retries
* msft-go version is verified at the CI runtime
* Packaging artifacts are now reused from the packaging step group, and integrations server image build takes acceptable 4-5 minutes
What is left: Let's fix the test and leave only the required test groups in the new pipeline. The FIPS images versions need to be updated when promoted

testing/integration/fleetserver_fips_test.go
```diff
@@ -73,14 +73,16 @@ func TestFIPSAgentConnectingToFIPSFleetServerInECHFRH(t *testing.T) {
...
+			break
 		}
 	}
 
 	// Check that this Agent is online (i.e. healthy) and is FIPS-capable. This
 	// will prove that a FIPS-capable Agent is able to connect to a FIPS-capable
 	// Fleet Server, with both running in ECH.
-	require.Equal(t, "online", agentStatus)
 	require.Equal(t, true, agentIsFIPS)
+	require.Equal(t, "online", agentStatus)
```
michel-laterman: @ycombinator this test is currently failing; the instance deployed in ECH does not state it's a FIPs instance; can we confirm that the kibana API returns this information?

Also @kruskall, an earlier test where the status check was first reported a degraded status, i think this is due to the apm-server but have not yet recreated to test.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+steps:
+  - label: Build and push custom elastic-agent image
+    depends_on:
+      - 'packaging-containers-x86-64-fips'
+    key: integration-fips-cloud-image
+    env:
+      FIPS: "true"
+    command: |
+      #!/usr/bin/env bash
+      set -euo pipefail
```
v1v: **nit**: I think we don't need this shebang or pipeline error validation

```suggestion
```

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+  TF_VAR_integration_server_docker_image: "${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}"
+
+  IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1744855248"
+  IMAGE_UBUNTU_X86_64_FIPS: "test-platform-ingest-elastic-agent-ubuntu-2204-fips-1748876280"
+  IMAGE_UBUNTU_ARM64_FIPS: "test-platform-ingest-elastic-agent-ubuntu-2204-fips-aarch64-1748876280"
+
+steps:
+  - label: Build and push custom elastic-agent image
+    depends_on:
+      - 'packaging-containers-x86-64-fips'
```
v1v: I cannot find where this dependency comes from. Is it expected in another BK pipeline? If so, I wonder whether we could add some context here about where that pipeline step is actually defined and the other way around, create a comment in the pipeline step `packaging-containers-x86-64-fips` to link to this BK pipeline.

Otherwise, things can be a bit complicated if no clear indication about those dependencies.

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+
+  - label: Start ESS stack for FIPS integration tests
+    key: integration-fips-ess
+    depends_on:
+      - integration-fips-cloud-image
+    env:
+      ASDF_TERRAFORM_VERSION: 1.9.2
+    command: |
+      #!/usr/bin/env bash
+      set -euo pipefail
```
v1v: Unrequired, see https://github.com/elastic/elastic-agent/blob/1b3801803eb9253999e42c9e92246bdfd3e2d371/.buildkite/scripts/steps/ess_start.sh#L1-L2

```suggestion
```

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+      useCustomGlobalHooks: true
+
+  - group: "fips:Stateful:Ubuntu"
+    key: integration-tests-ubuntu-fips
+    depends_on:
+      - integration-fips-ess
+    steps:
+      - label: "fips:x86_64:non-sudo:{{matrix}}"
+        depends_on:
+          - packaging-ubuntu-x86-64-fips
```
v1v: ditto about the dependency

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+          - fleet
+
+      - label: "fips:arm64:sudo:{{matrix}}"
+        depends_on:
+          - packaging-ubuntu-arm64-fips
+        env:
+          FIPS: "true"
+        command: |
+          buildkite-agent artifact download build/distributions/** . --step packaging-ubuntu-arm64-fips
+          .buildkite/scripts/steps/integration_tests_tf.sh {{matrix}} true
```
v1v: What's the reason we don't use a matrix entry for the `true` and `false` values?, as far as I see `"fips:arm64:non-sudo:{{matrix}}"` and `"fips:arm64:sudo:{{matrix}}"` only differ in the second argument?

The same for the other one

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+  DOCKER_REGISTRY: "docker.elastic.co"
+  VAULT_PATH: "kv/ci-shared/observability-ingest/cloud/gcp"
+  ASDF_MAGE_VERSION: 1.14.0
+  CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+  CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
+  TF_VAR_integration_server_docker_image: "${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}"
+
+  IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1744855248"
+  IMAGE_UBUNTU_X86_64_FIPS: "test-platform-ingest-elastic-agent-ubuntu-2204-fips-1748876280"
+  IMAGE_UBUNTU_ARM64_FIPS: "test-platform-ingest-elastic-agent-ubuntu-2204-fips-aarch64-1748876280"
```
simitt: Are the suffixes `1748876280` a version pinning? If so, how is this kept up-to-date?

.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -48,7 +53,17 @@ outputJSON="build/${fully_qualified_group_name}.integration.out.json"
...
 
 set +e
-TEST_BINARY_NAME="elastic-agent" AGENT_VERSION="${AGENT_VERSION}" SNAPSHOT=true gotestsum --no-color -f standard-quiet --junitfile-hide-skipped-tests --junitfile "${outputXML}" --jsonfile "${outputJSON}" -- -tags integration -test.shuffle on -test.timeout 2h0m0s github.com/elastic/elastic-agent/testing/integration -v -args -integration.groups="${GROUP_NAME}" -integration.sudo="${TEST_SUDO}"
+TEST_BINARY_NAME="elastic-agent" AGENT_VERSION="${AGENT_VERSION}" SNAPSHOT=true \
+  gotestsum --no-color -f standard-quiet \
+  --junitfile-hide-skipped-tests \
+  --junitfile "${outputXML}" \
+  --jsonfile "${outputJSON}" \
+  -- \
+    -tags integration -test.shuffle on -test.timeout 2h0m0s \
```
simitt: Is 2h a realistic timeout for this or was it just set to a really high value?

testing/integration/fleetserver_fips_test.go
simitt: >the instance deployed in ECH does not state it's a FIPs instance;

@michel-laterman  Is this expected?

.buildkite/bk.fips-integration.pipeline.yml
pazone: We have an auto-update mechanism that creates automatic pull requests bumping image versions. https://github.com/elastic/elastic-agent/pull/8211

@v1v What is the preferred way to add these image versions to the autoupdate scope?

v1v: > What is the preferred way to add these image versions to the autoupdate scope?

Let me copy the comment I added in the CI Agents VM repository (since it's a private repository):

> When using the new image in the Elastic Agent pipelines, let's use the same approach of defining an env variable at the top-level, with a similar semantic:
> * IMAGE_UBUNTU_FIPS_2204_X86_64: "platform-ingest-elastic-agent-ubuntu-2204-fips-1748876280"
> * IMAGE_UBUNTU_FIPS_2204_ARM_64: "platform-ingest-elastic-agent-ubuntu-2204-fips-aarch64-1748347205"
> I think the above should work with the sed expression:
> * https://github.com/elastic/elastic-agent/pull/8211#diff-1d4dfe45ed31211e8250fed1dceda1e2cb498d01766f6b7a6e94a90386c06f29R54-R55

v1v: However, since it's a new file it will require to add a new target entry, similar to https://github.com/elastic/elastic-agent/pull/8211/files#diff-1d4dfe45ed31211e8250fed1dceda1e2cb498d01766f6b7a6e94a90386c06f29R46-R55

If this PR gets merged before https://github.com/elastic/elastic-agent/pull/8211, I can update mine, or the other way around, I can contribute here if needed

pazone:
```suggestion
  IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-elastic-agent-ubuntu-2204-fips-1748955449"
  IMAGE_UBUNTU_ARM64_FIPS: "platform-ingest-elastic-agent-ubuntu-2204-fips-aarch64-1748955449"
```

The FIPS images were promoted

.buildkite/scripts/buildkite-integration-tests.sh
pazone: I think the `2h` timeout comes from the OGC framework times. This script is used by different types of integration tests. Test runs usually take less than 1h. So if a test run takes 2h, then something went wrong, and it should be retried. We can adjust this value when all the test types are migrated to buildkite runners

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+          automatic:
+            limit: 1
+        agents:
+          provider: "aws"
+          image: "${IMAGE_UBUNTU_X86_64_FIPS}"
+          instanceType: "m5.2xlarge"
+        matrix:
+          - fleet # currently there is only a single test in the fleet group, add more tests once they have been defined
+
+      - label: "fips:x86_64:sudo:{{matrix}}"
```
pazone: Important: This step skips all tests

.buildkite/bk.fips-integration.pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+          automatic:
+            limit: 1
+        agents:
+          provider: "aws"
+          image: "${IMAGE_UBUNTU_ARM64_FIPS}"
+          instanceType: "m6g.2xlarge"
+        matrix:
+          - fleet
+
+      - label: "fips:arm64:sudo:{{matrix}}"
```
pazone: Important: This step skips all tests

.buildkite/bk.fips-integration.pipeline.yml
michel-laterman: It's from `.buildkite/integration.pipeline.yml` which calls this pipeline and `bk.integrattion.pipeline.yml`, i'll add a comment to point the source out

.buildkite/bk.fips-integration.pipeline.yml
michel-laterman: correct, currently the only test we have is marked as requiring non-sudo

.buildkite/bk.fips-integration.pipeline.yml
michel-laterman: I copied the approach in `bk.integration.pipeline.yml`; i think it's so that the integration tests have increased parallel execution in that pipeline. I don't know if we need that here though, @ycombinator what do you think? I don't think we will have too many FIPS specific tests does it make sense to use less runners?

.buildkite/scripts/buildkite-integration-tests.sh
michel-laterman: Yes, this value is not changed, I don't want to alter our existing integration test assumptions.

testing/integration/fleetserver_fips_test.go
michel-laterman: Nope, but it looks like it was a oneoff failure, it's no longer in the BK pipeline. we just have the degraded state

.buildkite/bk.fips-integration.pipeline.yml
pazone: I understand. We need to decide whether to remove empty steps or enable some sudo tests for FIPS

.buildkite/bk.fips-integration.pipeline.yml
michel-laterman: Moved it back to every step, for some reason steps in `bk.integration.pipeline.yml` were also getting these values

testing/integration/fleetserver_fips_test.go
```diff
@@ -73,14 +73,16 @@ func TestFIPSAgentConnectingToFIPSFleetServerInECHFRH(t *testing.T) {
...
+			break
 		}
 	}
 
 	// Check that this Agent is online (i.e. healthy) and is FIPS-capable. This
 	// will prove that a FIPS-capable Agent is able to connect to a FIPS-capable
 	// Fleet Server, with both running in ECH.
-	require.Equal(t, "online", agentStatus)
 	require.Equal(t, true, agentIsFIPS)
+	//require.Equal(t, "online", agentStatus) // FIXME: Uncomment after https://github.com/elastic/apm-server/issues/17063 is resolved
```
michel-laterman, line 87: @ycombinator, @kruskall In order to try to get this merged so we can run other fips integration tests i've commented this out

.buildkite/bk.fips-integration.pipeline.yml
ycombinator: Yeah, I think we can reduce some duplication here by adding the `true`/`false` for the sudo argument to the `matrix`.  Since we also have the test groups in the `matrix` (currently just one value: `fleet`), we'll probably need to use [`matrix.setup`](https://buildkite.com/docs/pipelines/configure/workflows/build-matrix) instead of just `matrix` so we can have both dimensions in the matrix — test groups and sudo/non-sudo.

pazone: The initial code duplication is intentional. The multi-axis matrix only simplifies the step representation when the sudo and non-sudo test group sets are identical. Any difference requires a matrix adjustment. 
Example from the doc:
```
matrix:
    setup:
      arch:
        - "amd64"
        - "arm64"
      os:
        - "windows"
        - "linux"
      test:
        - "A"
        - "B"
    adjustments:
      - with:
          os: "windows"
          arch: "arm64"
          test: "B"
        soft_fail: true
      - with:
          os: "linux"
          arch: "arm64"
          test: "B"
        skip: true
```
So we have to exclude every unwanted combination explicitly. It makes it harder to define which test groups are executed. Adding and deleting test groups from the matrix may also become tedious. If we still plan to eventually generate these pipelines, a multi-axis matrix may introduce serious complexity to the generator. 
I'd recommend using a multi-axis matrix only if no group changes are expected in the foreseeable future.

michel-laterman: I've changed to use multi-axis matrixes, I think we may only want to add a group or two for FIPS testing

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,142 @@
...
+
+  - label: Start ESS stack for FIPS integration tests
+    key: integration-fips-ess
+    depends_on:
+      - integration-fips-cloud-image
+    env:
+      ASDF_TERRAFORM_VERSION: 1.9.2
+      CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+      CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
+      TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud:git-${BUILDKITE_COMMIT:0:12}"
```
pazone, line 38: Question: For what reason `TF_VAR_integration_server_docker_image` inlined to each step?

ycombinator (approved):
LGTM, great to see this work getting finalized! 

Please get a second LGTM from @pazone or @v1v as well.  Thanks!

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,142 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  DOCKER_REGISTRY: "docker.elastic.co"
+  VAULT_PATH: "kv/ci-shared/observability-ingest/cloud/gcp"
```
v1v: I’d like to clarify the usage of this secret, as it appears to be exclusively associated with the cleanup step and pipeline:

- https://github.com/search?q=repo%3Aelastic%2Felastic-agent%20VAULT_PATH&type=code
- https://github.com/search?q=repo%3Aelastic%2Felastic-agent+gce-cleanup.sh&type=code

I'm not familiar with the background here, @pazone, could you please double-check this? Thanks!

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,142 @@
...
+    env:
+      ASDF_TERRAFORM_VERSION: 1.9.2
+      CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+      CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
+      TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud:git-${BUILDKITE_COMMIT:0:12}"
+    command: |
+      source .buildkite/scripts/steps/ess_start.sh
+    artifact_paths:
+      - test_infra/ess/*.tfstate
+      - test_infra/ess/*.lock.hcl
```
v1v, line 43: I have some concerns about storing Terraform state since it contains sensitive information. Would it be possible to explore alternatives, such as using an S3 bucket for storage? This could help enhance our security.

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,142 @@
...
+    allow_dependency_failure: true
+    command: |
+      buildkite-agent artifact download "build/*.xml" .
+    agents:
+      image: "docker.elastic.co/ci-agent-images/platform-ingest/buildkite-agent-beats-ci-with-hooks:0.5"
+      useCustomGlobalHooks: true
+    soft_fail:
+      - exit_status: "*"
+    plugins:
+      - test-collector#v1.10.1:
```
v1v: Let's use the new pattern from https://github.com/elastic/elastic-agent/pull/8311

```suggestion
      - elastic/vault-secrets#v0.1.0:
          path: "kv/ci-shared/platform-ingest/buildkite_analytics_token"
          field: "token"
          env_var: "BUILDKITE_ANALYTICS_TOKEN"
      - test-collector#v1.11.0:
```

v1v (commented):
Overall, it looks good to me; however, there are some concerns regarding the ergonomics of the security artifacts in the pipeline step and some minor nits with the copy and paste.

pazone:
Generally, it looks good.
Warning:
The new FIPS sudo groups run no tests([example](https://buildkite.com/elastic/elastic-agent/builds/21885/summary/annotations?jid=01974102-7197-4c63-aff8-ce1ff0d6faab)). It creates a VM but runs no tests

.buildkite/bk.integration-fips.pipeline.yml
michel-laterman: I tried putting it in the top level `env`, but it was being used in steps from the `bk.integration.pipeline.yml` file

ycombinator (approved):

.buildkite/bk.integration-fips.pipeline.yml
v1v: > it was being used in steps from the bk.integration.pipeline.yml file

We have reported that issue to Buildkite, so when using `buildkite-agent upload` top-level env variables are copied between pipelines even though they are actually created in individual pipelines. That behaviour is actually not happening when using the `trigger` step.

I'll try to ping the BK support again and see if they actually understand this problem. in my view, it's not a desire behaviour but it's actually happening though

.buildkite/bk.integration-fips.pipeline.yml
pazone: That's right. `VAULT_PATH` is not needed here and in a few other pipelines. I see other minor issues in this and other pipelines. I assigned a cleanup issue to myself

.buildkite/bk.integration-fips.pipeline.yml
pazone: 👍 It deserves a dedicated issue(raised) or even a GH discussion.

pazone (approved):

.buildkite/bk.integration-fips.pipeline.yml
v1v: Well... if it's not needed, I'd suggest to clean this up here instead adding it

```suggestion
```

ycombinator (dismissed):

pazone:
FYI: I raised a [PR](https://github.com/elastic/elastic-agent/pull/8377/) that requires to add buildkite-vault-plugin to each new step.
```
plugins:
          - elastic/vault-secrets#v0.1.0:
              path: "kv/ci-shared/platform-ingest/platform-ingest-ec-prod"
              field: "apiKey"
              env_var: "EC_API_KEY"
```
It fixes an issue with retries on Windows

.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -15,6 +15,11 @@ if [ -z "$TEST_SUDO" ]; then
   exit 1
 fi
 
+if [ "${FIPS:-false}" == "true" ]; then
```
pchila, line 18: Why is this needed to just run go test ? Test code should also be FIPS-capable ?
Not a blocking issue, just a curiosity: is it required by the buildkite agent image we are using for FIPS integration tests ?

pchila (approved):

.buildkite/scripts/buildkite-integration-tests.sh
michel-laterman: Using `GOEXPERIMENT=systemcrypto` forces a check for microsoft/go

michel-laterman:
buildkite test this

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -4,7 +4,7 @@ env:
   DOCKER_REGISTRY: "docker.elastic.co"
   ASDF_MAGE_VERSION: 1.14.0
 
-  IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1744855248"
+  IMAGE_UBUNTU_2404_X86_64: "family/platform-ingest-elastic-agent-ubuntu-2404"
```
pchila: I believe this will undo the buildkite image pinning we introduced recently (@pkoutsovasilis keep me honest here).
@michel-laterman why is this change needed ?

pkoutsovasilis: nice catch @pchila , we switched to VM images pinning gent so this change shouldn't be like it 🙂

v1v: The file `.buildkite/bk.integration-fips.pipeline.yml` needs to be added to the list for the fixed VM images automation. Here are the steps to follow:

1. Duplicate https://github.com/elastic/elastic-agent/blob/738cc402fbab5ac3f205a104d866d6f652cb906e/.ci/updatecli/updatecli-bump-vm-images.yml#L57-L65 for `bk.integration-fips.pipeline.yml` in the file `.ci/updatecli/updatecli-bump-vm-images.yml`.
2. Modify the line `IMAGE_UBUNTU_2404_X86_64: "family/platform-ingest-elastic-agent-ubuntu-2404"` to `IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1749258065"`.
3. Use the latest version from https://github.com/elastic/elastic-agent/pull/8390 and replace `1748955449` with `1749258065`.

Let me know if you need further assistance. 🙇

v1v:
### Updatecli tests

```bash
$ gh pr checkout 8035
$ git checkout -b test/fips-ech-updatecli
$ git push origin test/fips-ech-updatecli
$ # modify .ci/updatecli/values.d/scm.yml to use my fork
  GITHUB_ACTOR=v1v \
  BRANCH_NAME=test/fips-ech-updatecli updatecli apply \ 
    --config .ci/updatecli/updatecli-bump-vm-images.yml \
```

```diff
diff --git a/.ci/updatecli/values.d/scm.yml b/.ci/updatecli/values.d/scm.yml
index 2708262728..c01cb89d50 100644
--- a/.ci/updatecli/values.d/scm.yml
+++ b/.ci/updatecli/values.d/scm.yml
@@ -1,8 +1,8 @@
 scm:
   enabled: true
-  owner: elastic
+  owner: v1v
```

Produced https://github.com/v1v/elastic-agent/pull/9

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,145 @@
...
+      buildkite-agent artifact download "test_infra/ess/**" . --step "integration-fips-ess"
+      ls -lah test_infra/ess
+      .buildkite/scripts/steps/ess_down.sh
+    agents:
+      image: "docker.elastic.co/ci-agent-images/platform-ingest/buildkite-agent-beats-ci-with-hooks:0.5"
+      useCustomGlobalHooks: true
+
+  - label: Aggregate test reports
+    # Warning: The key has a hook in pre-command
+    key: aggregate-reports-fips
```
v1v: I'm not sure whether this is correct, I cannot find any references. Maybe it's related to the recent change using the `plugin` section later on, so it's not needed anymore

.buildkite/hooks/pre-command
```diff
@@ -72,3 +72,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent-binary-dra" ]]; then
     release_manager_login
   fi
 fi
+
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent" && "$BUILDKITE_STEP_KEY" == "integration-fips-cloud-image" ]]; then
```
v1v, line 76: For the future me, I think we can do something about a BK plugin, so we can be explicit about the pipeline when the  docker login is needed, so we avoid using `pre-command` hooks with the combination of `BUILDKITE_STEP_KEY` and `BUILDKITE_PIPELINE_SLUG` for the sake of simplicity

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,145 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  DOCKER_REGISTRY: "docker.elastic.co"
```
v1v, line 4: Do we know whether we need this env variable? I cannot find any references https://github.com/search?q=repo%3Aelastic%2Felastic-agent%20DOCKER_REGISTRY&type=code?

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -0,0 +1,145 @@
...
+  ASDF_MAGE_VERSION: 1.14.0
+
+  IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1749258065"
+  IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-elastic-agent-ubuntu-2204-fips-1748955449"
+  IMAGE_UBUNTU_ARM64_FIPS: "platform-ingest-elastic-agent-ubuntu-2204-fips-aarch64-1748955449"
```
v1v, line 9: Just a question, shall we use `-fips` after `-aarch64` and follow the below naming convention?

- `platform-ingest-elastic-agent-<os-name>-<os-version>(-fips)?(-<arch>)?-<timestamp>`

where:
- `<os-name>` is `ubuntu`, `windows`, `debian`, ...
- `<os-version>` is `2204`, `11`, `10`, `2025`, ...
- `<arch>` is empty for `amd64` or `aarch64` ...
- `<timestamp>` is current timemilliseconds.

I somehow think `-fips` should be at the end similarly done with the name of the Env variable `IMAGE_UBUNTU_ARM64_FIPS`. What do you think?

v1v (commented):
LGTM, with some minor comments

.buildkite/bk.integration-fips.pipeline.yml
michel-laterman: I think this is needed by the docker login

.buildkite/bk.integration-fips.pipeline.yml
michel-laterman: @pchila, I know we recently had discussions for renaming `elastic-agent-fips-cloud -> elastic-agent-cloud-fips`; do we have an option for VM image names?

.buildkite/bk.integration-fips.pipeline.yml
michel-laterman: Removed it

.ci/updatecli/updatecli-bump-vm-images.yml
```diff
@@ -63,3 +63,13 @@ targets:
...
+
+  update-buildkite-bk.integration-fips.pipeline:
+    name: "Update .buildkite/bk.integration-fips.pipeline.yml"
+    sourceid: latestVersion
+    scmid: githubConfig
+    kind: file
+    spec:
+      file: .buildkite/bk.integration-fips.pipeline.yml
+      matchpattern: '(IMAGE_.+): "platform-ingest-elastic-agent-(.+)-(.+)"'
+      replacepattern: '$1: "platform-ingest-elastic-agent-$2-{{ source "latestVersion" }}"'
```
ycombinator, line 75: Since this target is identical to the `update-buildkite-bk.integration.pipeline` above it, except for the `spec.file` value, perhaps we could use a single target and `spec.files` to specific both target files instead?

.buildkite/bk.integration-fips.pipeline.yml
v1v: I'll keep a note here to revisit this, thanks Michel!

v1v (approved):
LGTM, the questions I asked earlier are not blockers, but likely something to be done in  follow-up

.ci/updatecli/updatecli-bump-vm-images.yml
michel-laterman: I think that (and targetting the `fips` images should be done as follow up work

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -121,8 +121,6 @@ steps:
       useCustomGlobalHooks: true
 
   - label: Aggregate test reports
-    # Warning: The key has a hook in pre-command
-    key: aggregate-reports-fips
```
ycombinator: Not opposed to this change but I'm curious why it was made.  Especially since the analogous step in the regular integration tests BK pipeline does have a key (`aggregate-reports`).

.ci/updatecli/updatecli-bump-vm-images.yml
ycombinator: Sounds good.  Could you please make an issue for the follow up work so we don't lose track of it? Thanks.

.buildkite/bk.integration-fips.pipeline.yml
michel-laterman: The aggregate-reports key is not refereced in our pre-command hooks

.ci/updatecli/updatecli-bump-vm-images.yml
michel-laterman: https://github.com/elastic/elastic-agent/issues/8408

PR #8042, opened 2025-04-30
[9.0] Use beats dependency from 9.0 branch

swiatekm:
## What does this PR do?

We depend on beats as a library for beats receivers. Up until now, the agent 9.0 branch depended on the beats main branch in this respect. This PR changes that, and fixes any test failures due to the effective update. These changes are the same as https://github.com/elastic/elastic-agent/pull/8041.

## Why is it important?

Agent branches should depend on the appropriate beats branch, rather than `main`.


swiatekm:
Unfortunately, this update is problematic on 9.0 due to the fact that the Prometheus dependency update in beats was not backported to 9.0. Closing for now.

PR #8056, opened 2025-05-01
Retry enrollment requests when an error is returned, add enrollment timeout

michel-laterman:
## What does this PR do?

Retry enrollment requests when an error is returned until a timeout is reached.
Add `--enroll-timeout` and `FLEET_ENROLL_TIMEOUT` to control how long the timeout is; default 10m.

## Why is it important?

Increase reliability of enrollments.

## Checklist

- ~~I have commented my code, particularly in hard-to-understand areas~~

## Disruptive User Impact

Agents running via container, or command line **without** the delay enrollment option will retry for 10m instead of indefinitely.

### DIFF
--- a/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
+++ b/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: bug-fix
+kind: enhancement
 
 # Change summary; a 80ish characters long description of the change.
 summary: Retry enrollment requests on any error
--- a/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
+++ b/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
@@ -11,14 +11,14 @@
 kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
-summary: Retry enrollment requests on any error
+summary: Retry enrollment requests on any network error
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
 description: |
-  If any error is encountered during an attempted enrollment, the elastic-agent
-  will backoff and retry.
+  If any network error is encountered during an attempted enrollment, the
+  elastic-agent will backoff and retry.
 
 # Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
 component: elastic-agent
--- a/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
+++ b/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
@@ -11,12 +11,14 @@
 kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
-summary: Retry enrollment for all errors
+summary: Retry enrollment requests on any error
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
-#description:
+description: |
+  If any error is encountered during an attempted enrollment, the elastic-agent
+  will backoff and retry.
 
 # Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
 component: elastic-agent
@@ -25,7 +27,7 @@ component: elastic-agent
 # If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
 # NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
 # Please provide it if you are adding a fragment for a different PR.
-#pr: https://github.com/owner/repo/1234
+pr: https://github.com/elastic/elastic-agent/pull/8056
 
 # Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
 # If not present is automatically filled by the tooling with the issue linked to the PR number.
--- a/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
+++ b/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
@@ -19,7 +19,7 @@ summary: Retry enrollment requests on any error
 description: |
   If any error is encountered during an attempted enrollment, the elastic-agent
   will backoff and retry. Add a new --enrollement-timeout flag and
-  FLEET_ENROLLMENT_TIMEOUT env var to set how long it tries for, default 10m.
+  FLEET_ENROLL_TIMEOUT env var to set how long it tries for, default 10m.
 
 # Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
 component: elastic-agent
--- a/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
+++ b/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
@@ -18,7 +18,7 @@ summary: Retry enrollment requests on any error
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
 description: |
   If any error is encountered during an attempted enrollment, the elastic-agent
-  will backoff and retry. Add a new --enrollement-timeout flag and
+  will backoff and retry. Add a new --enroll-timeout flag and
   FLEET_ENROLL_TIMEOUT env var to set how long it tries for, default 10m.
 
 # Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
--- a/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
+++ b/changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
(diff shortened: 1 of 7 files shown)


internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -557,6 +557,7 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
 			retry = true
 		case err != nil:
 			c.log.Warnf("Enrollment failed: %s", err.Error())
+			retry = true
```
blakerouse, line 560: In all cases it seems that `retry` is `true`. We should update the code to just remove the `retry` variable.

changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Retry enrollment for all errors
```
blakerouse, line 14: Can you provide more context to this change from the standpoint of a user reading the changelog? Provide more information on how it helps the user that they upgrade to get this improvement.

blakerouse (changes requested):

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -544,21 +544,16 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 			c.log.Warn("Remote server is not ready to accept connections(Connection Refused), will retry in a moment.")
-			retry = true
 		case errors.Is(err, fleetapi.ErrTemporaryServerError):
 			c.log.Warnf("Remote server failed to handle the request(%s), will retry in a moment.", err.Error())
-			retry = true
 		case err != nil:
 			c.log.Warnf("Enrollment failed: %s", err.Error())
-		}
-		if !retry {
+		case err == nil:
```
blakerouse, line 556: I think this also needs to check for errors from context cancellation. That would be better.

```suggestion
		case errors.Is(err, context.Canceled) || errors.Is(err, context. DeadlineExceeded) || err == nil:
			return err
```

blakerouse (changes requested):

internal/pkg/agent/cmd/enroll.go
```diff
@@ -473,7 +481,8 @@ func enroll(streams *cli.IOStreams, cmd *cobra.Command) error {
 	key, _ := cmd.Flags().GetString("elastic-agent-cert-key")
 	keyPassphrase, _ := cmd.Flags().GetString("elastic-agent-cert-key-passphrase")
 
-	ctx := handleSignal(context.Background())
+	ctx, cancel := context.WithTimeout(handleSignal(context.Background()), enrollTimeout)
```
blakerouse, line 484: What happens here if `--enroll-timeout 0` or `--enroll-timeout -1` is passed? Does `0` result in this completing instantly? I think `-1` would make it go to never timing out, I don't think that is what would happen here.

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -543,25 +543,23 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 			c.log.Warnf("Remote server failed to handle the request(%s), will retry in a moment.", err.Error())
-			retry = true
+		case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded), err == nil:
+			break RETRYLOOP
 		case err != nil:
-			c.log.Warnf("Enrollment failed: %s", err.Error())
+			c.log.Warnf("Error detected: %s, will retry in a moment.", err.Error())
 		}
-		if !retry {
+		if !backExp.Wait() {
```
blakerouse, line 560: I believe this returns false when the context is cancelled, should this not `break RETRYLOOP`?

blakerouse (commented):
I cannot see how this PR ensures that delay enroll doesn't timeout? Seems like to me with the timeout being set to `10m` in the default case, which is correct for a default. Only in the case of delay enrollment, there should be no timeout.

internal/pkg/agent/cmd/enroll.go
michel-laterman: I think a negative value would time out instantly, but I make it so a zero/negative value disables the timeout

changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
```diff
@@ -0,0 +1,36 @@
...
+summary: Retry enrollment requests on any error
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: |
+  If any error is encountered during an attempted enrollment, the elastic-agent
+  will backoff and retry. Add a new --enroll-timeout flag and
+  FLEET_ENROLL_TIMEOUT env var to set how long it tries for, default 10m. A
+  negative value disables the timeout.
```
ycombinator, line 23:
```suggestion
  zero or negative value disables the timeout.
```

internal/pkg/agent/cmd/enroll.go
```diff
@@ -95,6 +96,7 @@ func addEnrollFlags(cmd *cobra.Command) {
 	cmd.Flags().StringSliceP("proxy-header", "", []string{}, "Proxy headers used with CONNECT request: when bootstrapping Fleet Server, it's the proxy used by Fleet Server to connect to Elasticsearch; when enrolling the Elastic Agent to Fleet Server, it's the proxy used by the Elastic Agent to connect to Fleet Server")
 	cmd.Flags().BoolP("delay-enroll", "", false, "Delays enrollment to occur on first start of the Elastic Agent service")
 	cmd.Flags().DurationP("daemon-timeout", "", 0, "Timeout waiting for Elastic Agent daemon")
+	cmd.Flags().DurationP("enroll-timeout", "", 10*time.Minute, "Timeout waiting for Elastic Agent enroll command. A negative value disables the timeout.")
```
ycombinator, line 99:
```suggestion
	cmd.Flags().DurationP("enroll-timeout", "", 10*time.Minute, "Timeout waiting for Elastic Agent enroll command. A zero or negative value disables the timeout.")
```

internal/pkg/agent/cmd/enroll.go
```diff
@@ -285,6 +288,10 @@ func buildEnrollmentFlags(cmd *cobra.Command, url string, token string) []string
 		args = append(args, "--daemon-timeout")
 		args = append(args, daemonTimeout.String())
 	}
+	if enrollTimeout != 0 {
```
ycombinator, line 291: Should this be `> 0` for consistency?

michel-laterman: Nope, we want to pass a negative here so the `install` command can disable the timeout as well

changelog/fragments/1746113477-Retry-enrollment-for-all-errors.yaml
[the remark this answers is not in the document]
ycombinator: Right, that makes sense.

internal/pkg/agent/cmd/container.go
```diff
@@ -517,6 +518,10 @@ func buildEnrollArgs(cfg setupConfig, token string, policyID string) ([]string,
 		args = append(args, "--daemon-timeout")
 		args = append(args, cfg.Fleet.DaemonTimeout.String())
 	}
+	if cfg.Fleet.EnrollTimeout != 0 {
```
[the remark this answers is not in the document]
michel-laterman: same as the other command, passing a negative is needed to disable

internal/pkg/agent/cmd/enroll_cmd.go
swiatekm: The label on break isn't necessary here, but I think it might be better for readability to have it anyway. Does the linter complain about it?

As for the context, we don't stop the wait on cancellation, but I think we should? We don't do anything useful with the signalling channel, so may as well use `ctx.Done()` here I think.

michel-laterman: I've added the label for readability.
Context cancellation is checked is a case above (the  `c.enroll()` method will return a context related error if there is cancellation/timeout).

swiatekm: If the context is cancelled while we're in `backexp.Wait()`, we won't actually exit from it, and I think we should. Is that not right?

michel-laterman: correct, but that's a limitation with the backoff; I think it should be changed in a follow-up PR

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -543,25 +543,23 @@ func (c *enrollCmd) enrollWithBackoff(ctx context.Context, persistentConfig map[
...
 			c.log.Warn("Too many requests on the remote server, will retry in a moment.")
-			retry = true
 		case errors.Is(err, fleetapi.ErrConnRefused):
 			c.log.Warn("Remote server is not ready to accept connections(Connection Refused), will retry in a moment.")
-			retry = true
 		case errors.Is(err, fleetapi.ErrTemporaryServerError):
 			c.log.Warnf("Remote server failed to handle the request(%s), will retry in a moment.", err.Error())
-			retry = true
+		case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded), err == nil:
+			break RETRYLOOP
```
ycombinator, line 556: I understand that this relies on `c.enroll()` returning one of these errors if the context is cancelled.  Wouldn't it be more explicit to directly check for context cancellation here instead?

```suggestion
		case <-ctx.Done():
			break RETRYLOOP
```

blakerouse (dismissed):
This change itself now looks good. I am happy with how this turned out.

swiatekm (dismissed):
Looks good. We should address cancellation for the backoff in a follow-up though.

internal/pkg/agent/cmd/enroll_cmd.go
michel-laterman: Request is done with the fleetapi, it handles context:
https://github.com/elastic/elastic-agent/blob/6b16e9cf85723c8e77ec110af0d9717b1a13dda0/internal/pkg/fleetapi/enroll_cmd.go#L210

having a `ctx.Err()` check here now means that the switch statement checks two things: `err` and `ctx.Err()`; the function eventually returns `err`

PR #8092, opened 2025-05-06
Add ms_tls13kdf buildtag for FIPS builds

michel-laterman:
## What does this PR do?

Add the `ms_tls13kdf` build tag to FIPS builds.

## Why is it important?

microsoft/go TLSv1.3 kdf support (https://github.com/microsoft/go/pull/1662) requires the `ms_tls13kdf` build tag.


pchila:
Shouldn't the tag `ms_tls13kdf` be used for building the `elastic-agent` binary ? I don't think I saw that in the diff

nicholasberlin (dismissed):
LGTM FWIW

ycombinator:
Should we also check for `ms_tls13kdf ` around here? https://github.com/elastic/elastic-agent/blob/9548b96ebcdab303aa8add76447489875e77bd59/dev-tools/packaging/testing/package_test.go#L759

ycombinator (approved):

cmacknz:
This new tag has no effect on the generation of the FIPS notice file because there are no packages whose inclusion are conditional on the use of this tag? Correct?

https://github.com/elastic/elastic-agent/blob/804f1ee872b58c10ede9391204078598f6c8a0f4/dev-tools/mage/target/common/notice.go#L38-L39

We are only filtering on requirefips, just double checking adding another tag to the FIPS build doesn't affect the FIPS notice file.

PR #8104, opened 2025-05-07
[k8s tests] Add versions + comments explaining why

ycombinator:
## What does this PR do?

This PR updates the list of Kubernetes versions we run our k8s tests against.  It also adds comments explaining why each version exists in the list.

## Why is it important?

So our K8s tests run against K8s versions that are supported by various cloud service providers and/or being maintained by the K8s community itself.

### DIFF
--- a/.buildkite/bk.integration.pipeline.yml
+++ b/.buildkite/bk.integration.pipeline.yml
@@ -354,7 +354,7 @@ steps:
             - v1.31.0 # [2][4]
             - v1.32.0 # [2][4]
             - v1.33.0 # [4]
-            # [1] version is in Amazon EKS extended support list (https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions-extended
+            # [1] version is in Amazon EKS extended support list (https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions-extended)
             # [2] version is in Amazon EKS standard support list (https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions)
             # [3] version is in Azure AKS LTS versions list (https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#aks-kubernetes-release-calendar)
             # [4] version is in Kubernetes active branches list (https://kubernetes.io/releases/patch-releases/#detailed-release-history-for-active-branches)
--- a/.buildkite/bk.integration.pipeline.yml
+++ b/.buildkite/bk.integration.pipeline.yml
@@ -345,7 +345,6 @@ steps:
             - "basic,slim,complete,service,elastic-otel-collector"
             - "wolfi,slim-wolfi,complete-wolfi,elastic-otel-collector-wolfi"
             version:
-            - v1.25.16 # [1]
             - v1.26.15 # [1]
             - v1.27.16 # [1][3]
             - v1.28.15 # [1][3]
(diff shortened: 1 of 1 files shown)


swiatekm (dismissed):
Looks good to me, but instead of putting this justification in a buildkite pipeline definition, it should live in our support matrix. Does that already exist, or is it going to be the next step?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -345,12 +345,19 @@ steps:
             - "basic,slim,complete,service,elastic-otel-collector"
             - "wolfi,slim-wolfi,complete-wolfi,elastic-otel-collector-wolfi"
             version:
-            - v1.27.16
-            - v1.28.9
-            - v1.29.8
-            - v1.30.8
-            - v1.31.0
-            - v1.32.0
+            - v1.25.16 # [1]
```
pkoutsovasilis, line 348: actually is this still true for v1.25.16? the link says EoL 1st of May 2025 🙂 maybe we can remove it?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -345,12 +345,19 @@ steps:
...
+            - v1.26.15 # [1]
+            - v1.27.16 # [1][3]
+            - v1.28.15 # [1][3]
+            - v1.29.14 # [1]
+            - v1.30.0 # [2][3][4]
+            - v1.31.0 # [2][4]
+            - v1.32.0 # [2][4]
+            - v1.33.0 # [4]
+            # [1] version is in Amazon EKS extended support list (https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions-extended
```
pkoutsovasilis, line 357: nit
```suggestion
            # [1] version is in Amazon EKS extended support list (https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions-extended)
```

.buildkite/bk.integration.pipeline.yml
ycombinator: Good catch, removed in 7343c07fdb.  Thanks.

ycombinator:
@pkoutsovasilis as you mentioned off-PR today, `v1.26.15` is not supported by the Elastic Agent Helm Chart (it requires a minimum of `v1.27.0-0`).  So, for now, I will remove `v1.26.15` from this PR and not include it in the support matrix either.  

It was being included because EKS offers extended support for it until June 11, 2025.  Do you think it's worth trying to get our Helm Chart to support this version?  My instinct says no, since that support will only be needed for another 5 weeks or so. But I'm also not sure of how much effort it would be to add support for this version to the Helm Chart — maybe it's pretty quick 🤷. WDYT?

pkoutsovasilis:
> @pkoutsovasilis as you mentioned off-PR today, `v1.26.15` is not supported by the Elastic Agent Helm Chart (it requires a minimum of `v1.27.0-0`). So, for now, I have [removed it](4f5e81204e7dbb32d57ae0ef26b526e7898991bf) from this PR and will not include it in the support matrix either.
> 
> It was being included because EKS offers extended support for it until June 11, 2025. Do you think it's worth trying to get our Helm Chart to support this version? My instinct says no, since that support will only be needed for another 5 weeks or so. But I'm also not sure of how much effort it would be to add support for this version to the Helm Chart — maybe it's pretty quick 🤷. WDYT?

I would say we start from 1.27. Don't get me wrong technically speaking there is nothing AFAICT that prevents us from supporting 1.26; but since this is everywhere EoL in almost 1 month from now and all our testing if for 1.27+ I would say to leave it as it 🙂

v1v (approved):
LGTM

PR #8121, opened 2025-05-08
Remove 9.0.0 version pinning from tests

ycombinator:
## What does this PR do?

This PR un-pins the `AGENT_VERSION` from `9.0.0` in various tests, now that `9.0.0` has been released. 

## Why is it important?

The latest version of the Agent, as defined in the `version/version.go` file, will be tested.


pkoutsovasilis (commented):
This looks good to me, I couldn't find anywhere that the pinning is still applied. Waiting for the CI to be green for my approval 🙂

pkoutsovasilis (approved):

PR #8143, opened 2025-05-12
[CI] [Integration tests] .stable_snapshot_version file representing the latest stable snapshot version for integration tests

pazone:
…pshot version

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


ycombinator (commented):
I like the idea of reading the stable snapshot version from a file and auto-updating that file periodically.  We do this in the `fleet-server` repo; see https://github.com/elastic/fleet-server/blob/main/dev-tools/integration/.env and https://github.com/elastic/fleet-server/blob/main/.ci/bump-elastic-stack-snapshot.yml.  It would be good to make the implementations consistent between the repos, since it's the same teams maintaining them.

pazone:
> I like the idea of reading the stable snapshot version from a file and auto-updating that file periodically. We do this in the `fleet-server` repo; see https://github.com/elastic/fleet-server/blob/main/dev-tools/integration/.env and https://github.com/elastic/fleet-server/blob/main/.ci/bump-elastic-stack-snapshot.yml. It would be good to make the implementations consistent between the repos, since it's the same teams maintaining them.

Agreed. The `.env` file approach looks clearer. We currently have the `.package-version` file which is updated by [bump-agent-versions.sh](https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-agent-versions.sh).
IIUC, `.env` file implies the removal of [bump-agent-versions.sh](https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-agent-versions.sh) in favour of UpdateCLI and significant changes in [bump-agent-versions.yml](https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-agent-versions.yml). Which is IMHO also a good idea.

CC @pkoutsovasilis

test_infra/ess/deployment.tf
```diff
@@ -1,7 +1,13 @@
 variable "stack_version" {
   type        = string
   default     = "latest"
-  description = "the stack version to use"
+  description = "the stack deployment version"
+}
+
+variable "stable_snapshot_version" {
```
pkoutsovasilis, line 7: why we need this new var? pinning the images at the a "stable" tag is already supported by setting them inside `pkg/testing/ess/create_deployment_csp_configuration.yaml`, right? 🤔

pazone: Do you mean we can update the `pkg/testing/ess/create_deployment_csp_configuration.yaml` versions instead of extracting the version to another file?

pazone: `pkg/testing/ess/create_deployment_csp_configuration.yaml` doesn't contain the `docker` section anymore.

PR #8146, opened 2025-05-13
[CI] Autoupdate .stable-snapshot-version file

pazone:
## What does this PR do?

Auto-updates `.stable-snapshot-version` when a new tested snapshot is available for integration tests stack.
A follow-up PR for https://github.com/elastic/elastic-agent/pull/8143. 
The [initial PR](https://github.com/elastic/elastic-agent/pull/8143) must be merged and backported to make the changes take effect.

Works within the pre-existed `update-agent-versions` action that updates `.package-version` file. After merging, it will update `.stable-snapshot-version` too.

The stable versions are being taken from a GCP bucket according to the selected release branch by the workflow. 

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


.github/workflows/bump-agent-versions.sh
```diff
@@ -1,15 +1,16 @@
 #!/bin/bash
 set -e
 
+# because this script is running on the matrix, we need to fetch the current branch on every run
+current_ref=$(git rev-parse --abbrev-ref HEAD)
 package_version=$(mage integration:updatePackageVersion)
 version_requirements=$(mage integration:updateVersions)
-changes=$(git status -s -uno testing/integration/testdata/.upgrade-test-agent-versions.yml .package-version)
+stable_snapshot_version=$(mage integration:UpdateStableSnapshot $current_ref)
```
v1v, line 8: what's the reason to capture the mage command in the variable stable_snapshot_version? I cannot find any references

.github/workflows/bump-agent-versions.sh
```diff
@@ -1,15 +1,16 @@
 #!/bin/bash
 set -e
 
+# because this script is running on the matrix, we need to fetch the current branch on every run
```
v1v, line 4: I'm not sure if I understand this comment

.github/workflows/bump-agent-versions.sh
```diff
@@ -1,15 +1,16 @@
 #!/bin/bash
```
v1v, line 1: I'm not sure whether we want to mix up the agent versions with the pinned versions.

For instance, see https://github.com/elastic/elastic-agent/pull/8140

It contains versions in the format major.minor.patch.

While the `.stable-snapshot-version` will contains versions in the format `major.minor.patch-hash`. 

I'd expect  `.stable-snapshot-version` to be updated daily while `testing/integration/testdata/.upgrade-test-agent-versions.yml` not that often, maybe a couple of times monthly.

v1v (commented):
I added a few comments

.github/workflows/bump-agent-versions.sh
pchila: I agree with @v1v on this: we should avoid mixing the 2 version bump flows.
The agent versions play a role in our upgrade tests and should change only when a version bump is performed because of a new release.
ECH version could (and in most cases should) change at each unified release run.

PR #8008, opened 2025-04-28
[testing/integration][otel] move configurations to different files for better readability

VihasMakwana:
## What does this PR do?

The files `otel_test.go` and `beat_receiver_test.go` are quit big. This PR moves templates to different files for better readability. Over the longer term, we would like to group related tests (I'll work on follow-up).

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test

PR #8148, opened 2025-05-14
otel: support system/metrics input with experimental otel runtime

mauri870:
## What does this PR do?

This PR enables the system/metrics input to work with the otel runtime. It needs to be toggled explicitly by setting `_runtime_experimental: otel` in the input configuration.

## Checklist

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/4886
- Relates
https://github.com/elastic/ingest-dev/issues/4882
- Relates
https://github.com/elastic/ingest-dev/issues/4884
- Relates
https://github.com/elastic/ingest-dev/issues/4885


testing/integration/otel_test.go
```diff
@@ -1282,6 +1282,144 @@ agent.monitoring:
...
+	metricsets := []string{"cpu", "memory", "network", "filesystem"}
+	for _, mset := range metricsets {
+		index := fmt.Sprintf(".ds-metrics-system.%s-%s*", mset, info.Namespace)
+		require.Eventuallyf(t,
+			func() bool {
+				findCtx, findCancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer findCancel()
+
+				docs, err := estools.PerformQueryForRawQuery(findCtx, rawQuery, index, info.ESClient)
+				require.NoError(t, err)
```
pkoutsovasilis, line 1412: please correct me if I am wrong, but I believe that this require if we get an err will fail the test no matter if eventually this func returns true. Maybe we should go with `EventuallyWithT` here? 🙂

testing/integration/otel_test.go
```diff
@@ -1282,6 +1282,144 @@ agent.monitoring:
...
+		Namespace  string
+	}
+	esEndpoint, err := getESHost()
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
+	decodedApiKey, err := getDecodedApiKey(esApiKey)
+	require.NoError(t, err)
+	configTemplate := `agent.logging.level: info
```
swiatekm, line 1315: Could we put this into a file? I think that's what we want to do for all of these tests. The code is verbose enough without embedding large strings in it.

mauri870: I suggested this and @VihasMakwana created a PR for it https://github.com/elastic/elastic-agent/pull/8008. Vihas [mentioned](https://github.com/elastic/elastic-agent/pull/8008#discussion_r2075653113) that you were working on a similar task, so it was put on hold.

swiatekm: I think there might be some kind of misunderstanding there. My point was simply that for all of these comparisons, we can just use the agent policy config with a templated `_runtime_experimental` setting. But in any case, if we do want to make this improvement, we can start by doing so in new tests.

mauri870: Ah, got it. Thanks for the clarification. This is just a basic test to verify that the translation works, but it doesn’t compare the actual documents between beats receivers and metricbeat. That part is handled in https://github.com/elastic/elastic-agent/pull/7622. I'm strongly leaning toward closing that PR and adding the proper test here instead—it’ll save us from a lot of back and forth.

pkoutsovasilis (commented):
TestAgentMetricsInput had a flaky run here @mauri870 , so please have a look. The only this that comes to mind is that when this was retried it provisioned an isolated ESS stack so please do make sure that there is no interference when TestAgentMetricsInput is running in parallel from multiple runners

swiatekm:
> TestAgentMetricsInput had a flaky run [here](https://buildkite.com/elastic/elastic-agent/builds/21337#0196eeec-0f1c-4522-9ea2-a45e9dfa9f74/279-937) @mauri870 , so please have a look. The only this that comes to mind is that when this was retried it provisioned an isolated ESS stack so please do make sure that there is no interference when TestAgentMetricsInput is running in parallel from multiple runners

That looks like the `system.network.in.errors` metric doesn't always appear in the output.

@mauri870 I think what we should actually check is that the documents have the same fields, but not that the values are the same. Just ignoring all the fields containing metrics makes this test not very useful.

mauri870:
Found during testing that `event.duration` is sometimes missing from the ingested document with beats receivers. Pushed a fix in https://github.com/elastic/beats/pull/44442.

testing/integration/beat_receivers_test.go
```diff
@@ -377,6 +383,296 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
 	}
 }
 
+// TestAgentMetricsInput is a test that compares documents ingested by
+// agent metrics input and otel metrics input and asserts that they are
```
swiatekm:
```suggestion
// agent system/metrics input in process and otel modes and asserts that they are
```

swiatekm (dismissed):
LGTM

testing/integration/beat_receivers_test.go
```diff
@@ -377,6 +383,296 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
+					if t.Failed() {
+						t.Logf("agent document for metricset %s:\n%s", tt.metricset, agentDoc.StringToPrint())
+						t.Logf("otel document for metricset %s:\n%s", tt.metricset, otelDoc.StringToPrint())
+					}
+				})
+
+				stripNondeterminism(agentDoc, tt.metricset)
+				stripNondeterminism(otelDoc, tt.metricset)
+
+				agentKeys := *agentDoc.FlattenKeys()
```
khushijain21: you maybe be able to use this function https://github.com/elastic/elastic-agent/blob/main/testing/integration/otel_test.go#L1664
to compare if both docs have same keys

mauri870: Thanks, it fits like a glove.

khushijain21 (approved):

PR #8149, opened 2025-05-14
Test qa images with asdf golang plugin pinned

dliappis:


dliappis:
Test was successful, we can proceed with using the fresh images without pinning in a follow up PR.

PR #8174, opened 2025-05-19
Automatically update beats module versions

### LINKED ISSUE #8034: [beats receivers] Create automation to keep beats dependency in go.mod up to date in each branch
https://github.com/elastic/elastic-agent/blob/2e1d7a764888a3d433b6eace8e2a9540d605f76f/go.mod#L16

Elastic Agent now depends on the Beats repository to include Beat receivers in the OTel collector. This dependency is currently updated manually and is set to main in each branch.

Instead, each Elastic Agent release branch must depend on the matching branch in Beats:
* elastic-agent@main depends on beats@main
* elastic-agent@9.0 depends on beats@9.0
* elastic-agent@8.19 depends on beats@8.19
* etc

Creating a PR for each commit to Beats will likely be too much noise, instead:
* Create a PR on a schedule, perhaps once per day.
* Allow triggering the dependency update manually

Creating a GitHub action is likely the easiest solution to for accomplishing those goals.

swiatekm:
## What does this PR do?

Adds automation for updating the beats dependency. This cannot be done using dependabot because:

* Beats isn't properly versioned. We use pseudoversions pointing at commits, which dependabot doesn't support.
* Each elastic-agent branch needs to point at the respective beats branch.

Instead, we take an approach similar to [apm-server](https://github.com/elastic/apm-server/blob/main/.github/workflows/update-beats.yml). We use a scheduled github actions job, which triggers an update using updatecli, which in turn runs the `update:beats` mage target and submits a PR with the changes.

See an example execution in a fork here: https://github.com/swiatekm/elastic-agent/actions/runs/15137576163, and the PRs it created here: https://github.com/swiatekm/elastic-agent/pull/7 https://github.com/swiatekm/elastic-agent/pull/8.

## Why is it important?

In the absence of real beats versioning, we should strive to use latest beats receivers for the given release branch.

## Checklist

Run `mage update:beats main`.

## Related issues

- Closes #8034

### DIFF
--- a/.ci/scripts/update-beats.sh
+++ b/.ci/scripts/update-beats.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+#
+# This script is executed by the automation we are putting in place
+# and it requires the git add/commit commands.
+#
+set -euo pipefail
+
+mage update:beats "${1}"
--- a/.ci/scripts/update-beats.sh
+++ b/.ci/scripts/update-beats.sh
@@ -4,5 +4,7 @@
 # and it requires the git add/commit commands.
 #
 set -euo pipefail
+TARGET_VERSION="${1:?Error: Please provide the target version to update to}"
 
-mage update:beats "${1}"
+echo "~~~ Updating to elastic/beats@${TARGET_VERSION}"
+mage update:beats "${TARGET_VERSION}"
--- a/.github/workflows/bump-beats-version.yml
+++ b/.github/workflows/bump-beats-version.yml
@@ -47,7 +47,7 @@ jobs:
         go-version-file: 'go.mod'
 
     - name: Install mage
-      uses: magefile/mage-action@v3
+      uses: magefile/mage-action@6f50bbb8ea47d56e62dee92392788acbc8192d0b # v3.1.0
       with:
         version: v1.13.0
         install-only: true
--- a/.github/workflows/bump-beats-version.yml
+++ b/.github/workflows/bump-beats-version.yml
@@ -46,11 +46,6 @@ jobs:
       with:
         go-version-file: 'go.mod'
 
-    - name: Set git config
-      run: |
-        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
-        git config --global user.name "github-actions[bot]"
-
     - name: Install mage
       uses: magefile/mage-action@v3
       with:
--- a/.github/workflows/bump-beats-version.yml
+++ b/.github/workflows/bump-beats-version.yml
@@ -1,0 +1,85 @@
+name: update-beats
+
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: '0 0 * * 1-5'
+
+permissions:
+  contents: read
+
+jobs:
+  filter:
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    outputs:
+      matrix: ${{ steps.generator.outputs.matrix }}
+    permissions:
+      contents: read
+    steps:
+    - id: generator
+      uses: elastic/oblt-actions/elastic/active-branches@v1
+      with:
+        exclude-branches: "7.17,8.17,8.18,9.0"
+        filter-branches: true
+
+  update-beats:
+    permissions:
+      contents: write
+      pull-requests: write
+    needs:
+    - filter
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.filter.outputs.matrix) }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ matrix.branch }}
+
+    - name: Install Updatecli in the runner
+      uses: updatecli/updatecli-action@ae3030ce1710c6496214fb1f8fd3bd9437b2a69d # 2.82.0
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: 'go.mod'
+
+    - name: Set git config
+      run: |
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --global user.name "github-actions[bot]"
+
+    - name: Install mage
+      uses: magefile/mage-action@v3
+      with:
+        version: v1.13.0
+        install-only: true
+
+    - name: Run Updatecli in Apply mode
+      run: updatecli apply --config .ci/updatecli/update-beats.yml --values .ci/updatecli/values.d/scm.yml
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - if: ${{ failure()  }}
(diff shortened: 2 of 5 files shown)


dev-tools/mage/target/update/update.go
```diff
@@ -4,9 +4,28 @@
 
 package update
 
-import "github.com/magefile/mage/sh"
+import (
+	"fmt"
 
-// Update updates the generated files (aka make update).
-func Update() error {
```
swiatekm, line 10: This target wasn't used, and the make target it calls doesn't exist, so I deleted it.

swiatekm:
@andrzej-stencel I think we could automate otel updates in a similar way. We'd need some logic to decide whether to update on a given branch, but this infrastructure should otherwise handle it well.

.github/workflows/bump-beats-version.yml
```diff
@@ -0,0 +1,86 @@
...
+    timeout-minutes: 1
+    outputs:
+      matrix: ${{ steps.generator.outputs.matrix }}
+    permissions:
+      contents: read
+    steps:
+    - id: generator
+      uses: elastic/oblt-actions/elastic/active-branches@v1
+      with:
+        exclude-branches: "7.17,8.17,8.18,9.0"
```
v1v, line 23: What's the reason for using `active-branches` if we use only `main`? Maybe we only need `update-beats` with the `matrix.branch == main`?

.github/workflows/bump-beats-version.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: 'go.mod'
+
+    - name: Set git config
+      run: |
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --global user.name "github-actions[bot]"
+
```
v1v, line 53: I don't think we need this, updatecli should be doing that

.github/workflows/bump-beats-version.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Install mage
+      uses: magefile/mage-action@v3
```
v1v, line 55: we prefer to use pinned versions for those github actions that are not coming from the `actions/*` namespace.

.github/workflows/bump-beats-version.yml
```diff
@@ -0,0 +1,86 @@
...
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.filter.outputs.matrix) }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ matrix.branch }}
+
+    - name: Install Updatecli in the runner
+      uses: updatecli/updatecli-action@ae3030ce1710c6496214fb1f8fd3bd9437b2a69d # 2.82.0
```
v1v, line 42: If it helps, you can use https://github.com/elastic/oblt-actions/tree/main/updatecli/run-and-notify

.github/workflows/bump-beats-version.yml
swiatekm: I copied it from here: https://github.com/elastic/elastic-agent/blob/22f9fd2f23b3273eefae8b8f7659ab2c0890e69a/.github/workflows/bump-agent-versions.yml#L47. I'll test removing it in my fork.

swiatekm: Removed successfully.

.github/workflows/bump-beats-version.yml
swiatekm: Pinned it.

.github/workflows/bump-beats-version.yml
swiatekm: That one is deprecated, but I did use the ones it suggested instead.

.ci/scripts/update-beats.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+#
+# This script is executed by the automation we are putting in place
+# and it requires the git add/commit commands.
+#
+set -euo pipefail
+
+mage update:beats "${1}"
```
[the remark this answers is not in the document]
swiatekm: Done.

v1v (approved):

.github/workflows/bump-beats-version.yml
```diff
@@ -0,0 +1,65 @@
...
+    needs:
+    - filter
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.filter.outputs.matrix) }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ matrix.branch }}
```
v1v, line 39: > It's main and 8.19. We will probably enable this for 9.0 at some point in the future as well.

If so, you need to backport to the relevant branches, otherwise `ref` will try to run the updatecli pipeline and scripts from those branches but won't be available.

leehinman (approved):
LGTM  Thank You for automating this.

oakrizan (approved):

.github/workflows/bump-beats-version.yml
v1v: I've just added those labels:

- https://github.com/elastic/elastic-agent/pull/8174#event-17749902545

andrzej-stencel (approved):

.github/workflows/bump-beats-version.yml
andrzej-stencel: My understanding is that the updatecli script needs to be backported, but the GitHub action definition should not be backported. Is this correct?

swiatekm: Scheduled actions only run from the default branch, so this doesn't make any difference. I'd backport everything just to keep things consistent.

.github/workflows/bump-beats-version.yml
```diff
@@ -0,0 +1,65 @@
+name: update-beats
+
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: '0 0 * * 1-5'
```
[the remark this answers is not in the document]
swiatekm: Does that make things better in any way? I'd prefer if it ran on a consistent schedule personally, if it runs daily.

.github/workflows/bump-beats-version.yml
v1v: > but the GitHub action definition should not be backported. Is this correct?

Yes

cmacknz (approved):
Thanks! LGTM.

There is a chance we just leave 9.0 as is, since I suspect most users already on 9.0 will move on to 9.1.0 relatively quickly once it exists.

.github/workflows/bump-beats-version.yml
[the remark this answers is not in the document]
andrzej-stencel: You could just set it to a specific random time, like 6:43 or something. Surely not crucial anyway!

Issue #8176, opened 2025-05-19
Move `agent.download` under `agent.upgrade`


ycombinator:
Currently, the Agent configuration has an `agent.download` namespace like so:

```yml
# agent.download:
#   # source of the artifacts, requires elastic like structure and naming of the binaries
#   # e.g /windows-x86.zip
#   sourceURI: "https://artifacts.elastic.co/downloads/beats/"
#   # path to the directory containing downloaded packages
#   target_directory: "${path.data}/downloads"
#   # timeout for downloading package
#   timeout: 120s
#   # install_path describes the location of installed packages/programs. It is also used
#   # for reading program specifications.
#   install_path: "${path.data}/install"
#   # retry_sleep_init_duration is the duration to sleep for before the first retry attempt. This
#   # duration will increase for subsequent retry attempts in a randomized exponential backoff manner.
#   retry_sleep_init_duration: 30s
```

And an `agent.upgrade` namespace like so:
```yml
# agent.upgrade
#   # rollback settings
#   rollback:
#       # duration in which an upgraded Agent may be manually rolled back.
#       window: 168h
```

AFAICT, the settings under the `agent.download` namespace are used as part of the upgrade process.  So we should:
1. Introduce an `agent.upgrade.download` namespace containing the same settings that are under the `agent.download` namespace today.
2. Deprecate the `agent.download` namespace and schedule it for removal in the next major version release.
3. Perhaps rename some of the `camelCase` settings to `snake_case` in the `agent.download` namespace when we move them under the `agent.upgrade.download` namespace, for consistency.

pchila:
@ycombinator I think we could be on the right path here by replicating the useful `agent.download` settings (ignoring unused settings like `install_path` for example) into `agent.upgrade.download`: AFAIK in the current implementation the download settings are only used during upgrades so moving them makes sense.

To ease the migration we could initialize `agent.upgrade.download` settings from `agent.download` if the new config is not found (this may require some additional checks when loading config though).

Issue #6886, opened 2025-02-17
elastic-agent watcher should persist available rollbacks at the end of the grace period


pchila:
Requires #6884 

At the end of the grace period for a successful upgrade the agent must have persisted the available rollbacks outside of the update marker file which will be cleaned up by the watcher.
Any valid rollback_available entry present in the upgrade marker will be preserved by the watcher during the cleanup phase.

Issue #6884, opened 2025-02-17
Add `rollback_available` fields in upgrade marker


pchila:
Requires #6880 

Add the following fields to the upgrade marker definition:
- `rollback_available`: contains the list of versions still present on disk that the agent can rollback to (in the first implementation limited to 1 element max)
- Versions are composed by the following fields:
  -  `version`: version string corresponding to the instance of agent 
  -  `home`: path relative to Elastic Agent top directory pointing to the instance of agent
  - `valid_until`: timestamp in RFC 3339 format indicating until when this instance should be kept on disk 

For example:
```yaml
prev_version: 8.16.0-SNAPSHOT 
prev_hash: b65953 
prev_versioned_home: data/elastic-agent-8.16.0-SNAPSHOT-b65953 
updated_on: 2024-11-14T14:42:21.082601803Z 
version: 9.0.0-SNAPSHOT 
hash: 1d8229 
versioned_home: data/elastic-agent-9.0.0-SNAPSHOT-1d8229 
acked: true 
action: <action id>
details: 
  target_version: 9.0.0 
  state: UPG_WATCHING
# dedicated key for available rollbacks in upgrade marker
rollback_available:
  - version: 8.16.0-SNAPSHOT,
    home: data/elastic-agent-8.16.0-SNAPSHOT-b65953
    valid_until: "2024-11-21T14:42:21Z"
```

Issue #6882, opened 2025-02-17
Wait for elastic-agent watcher process to complete at startup and schedule cleanup


pchila:

ebeahan:
PR to implement functionality: https://github.com/elastic/elastic-agent/pull/11562

PR #8177, opened 2025-05-19
Pass rollback window duration to upgrade watcher command

### LINKED ISSUE #6880: Add 'rollback_available' parameter when launching elastic-agent watcher before starting new agent version

ycombinator:
## What does this PR do?

This PR enhances the upgrade process to invoke the Upgrade Watcher with an additional CLI option, `--rollback-window`.  This option accepts any value that can be parsed by [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration).  However, the code that invokes the Upgrade Watcher passes this value in seconds, e.g. `180s`.

## Why is it important?

The Upgrade Watcher will use the value of the `--rollback-window` CLI option to help ensure that the upgraded Agent can be rolled back manually within the specified rollback window duration.  Tjhis 

## Checklist

## Disruptive User Impact

None; the Upgrade Watcher is invoked internally as part of the Agent upgrade process. 

## Related issues

- Closes #6880 

## Questions to ask yourself

- ...


internal/pkg/agent/application/application.go
```diff
@@ -118,7 +118,7 @@ func New(
 
 	// monitoring is not supported in bootstrap mode https://github.com/elastic/elastic-agent/issues/1761
 	isMonitoringSupported := !disableMonitoring && cfg.Settings.V1MonitoringEnabled
-	upgrader, err := upgrade.NewUpgrader(log, cfg.Settings.DownloadConfig, agentInfo)
+	upgrader, err := upgrade.NewUpgrader(log, cfg.Settings.DownloadConfig, cfg.Settings.Upgrade, agentInfo)
```
ycombinator, line 121: Not for this PR, but perhaps we should consider making the download settings part of upgrade settings; created https://github.com/elastic/elastic-agent/issues/8176 for discussion.

ycombinator:
Some upgrade integration tests in CI are failing with errors that look related to the changes in this PR:

```
        	            	failed to start agent upgrade to version "9.0.1": exit status 1
        	            	Error initializing version information: reading package version from file "/opt/Elastic/Agent/data/elastic-agent-9.1.0-SNAPSHOT-fb8f43/package.version": open /opt/Elastic/Agent/data/elastic-agent-9.1.0-SNAPSHOT-fb8f43/package.version: no such file or directory
        	            	Error: Failed trigger upgrade of daemon: watcher did not start in time
        	            	context deadline exceeded
        	            	os: process already finished
```

Moving PR back to draft while I investigate and fix.

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -238,3 +242,22 @@ func restartAgent(ctx context.Context, log *logger.Logger, c client.Client) erro
...
+
+	if rollbackWindow > 0 {
+		cmdArgs = append(cmdArgs, "--rollback-window", fmt.Sprintf("%.fs", rollbackWindow.Seconds()))
+	}
+
+	// #nosec G204 -- user cannot inject any parameters to this command
+	return exec.Command(agentExecutable, cmdArgs...)
+}
+
+func isRollbackWindowFeatureAvailable(watcherVersion *version.ParsedSemVer) bool {
```
kaanyalti, line 261: Minor suggestion, "rollbackWindowFeatureFlag" may be a better name for this function.

kaanyalti, line 261: I assume after 9.1.0 is the oldest version we support we won't need this feature flag. I know it is quite a bit down the line, but is there any way to track this so we can remove it when the time comes.

internal/pkg/agent/application/upgrade/rollback_test.go
```diff
@@ -328,6 +329,40 @@ func TestRollback(t *testing.T) {
...
+		require.Equal(t, agentExecutable, cmd.Args[0])
+		require.Equal(t, "watch", cmd.Args[1])
+		require.Equal(t, "--path.config", cmd.Args[2])
+		require.NotEmpty(t, cmd.Args[3])
+		require.Equal(t, "--path.home", cmd.Args[4])
+		require.NotEmpty(t, cmd.Args[5])
+	})
+
+	t.Run("with_rollback_window", func(t *testing.T) {
+		cmd := makeBaseWatchCmd(agentExecutable, 2*time.Hour+15*time.Minute)
```
kaanyalti, line 349: It would be more readable or clearer if you can use a variable for the time and call the string function of the time var when asserting. It is not hard to calculate the seconds, but a variable would make things even easier.

internal/pkg/agent/cmd/watch.go
```diff
@@ -93,6 +96,12 @@ func watchCmd(log *logp.Logger, cfg *configuration.Configuration) error {
 		_ = locker.Unlock()
 	}()
 
+	rollbackWindow, err := cmd.Flags().GetDuration(flagRollbackWindow)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve %s flag value while watching the agent upgrade: %w", flagRollbackWindow, err)
+	}
+	_ = rollbackWindow // TODO: use rollbackWindow when https://github.com/elastic/elastic-agent/issues/6884 is implemented
```
kaanyalti, line 103: [6884 says that ](https://github.com/elastic/elastic-agent/issues/6884) description says that it is dependent on https://github.com/elastic/elastic-agent/issues/6880 which is the issue for this PR. 

The description of https://github.com/elastic/elastic-agent/issues/6880 says it requires https://github.com/elastic/elastic-agent/issues/6881 which seems to be completed. 

Could you please verify that the TODO is correct here?

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -238,3 +242,22 @@ func restartAgent(ctx context.Context, log *logger.Logger, c client.Client) erro
...
+
+func makeBaseWatchCmd(agentExecutable string, rollbackWindow time.Duration) *exec.Cmd {
+	cmdArgs := []string{
+		watcherSubcommand,
+		"--path.config", paths.Config(),
+		"--path.home", paths.Top(),
+	}
+
+	if rollbackWindow > 0 {
+		cmdArgs = append(cmdArgs, "--rollback-window", fmt.Sprintf("%.fs", rollbackWindow.Seconds()))
```
pchila, line 254: Is there a specific reason why this needs to be printed as seconds instead of just using [Duration.String()](https://pkg.go.dev/time#Duration.String) ?
I think that is a more readable/compact format than a float for seconds :thinking:

internal/pkg/agent/application/upgrade/rollback_test.go
```diff
@@ -328,6 +329,40 @@ func TestRollback(t *testing.T) {
...
+
+		// Expected command:
+		// elastic-agent watch --path.config /some/path --path.home /some/path
+		require.Len(t, cmd.Args, 6)
+		require.Equal(t, agentExecutable, cmd.Args[0])
+		require.Equal(t, "watch", cmd.Args[1])
+		require.Equal(t, "--path.config", cmd.Args[2])
+		require.NotEmpty(t, cmd.Args[3])
+		require.Equal(t, "--path.home", cmd.Args[4])
+		require.NotEmpty(t, cmd.Args[5])
```
pchila, line 345: Nit: wouldn't this check be easier with a simple assert.Equal on the `cmd.Args` slice (same for the next test as well) ?
```suggestion
		expectedArgs := []string{agentExecutable, "watch", "--path.config", paths.Config(), "--path.home", paths.Home()}
		assert.Equal(t, expectedArgs, cmd.Args)
```

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -342,8 +344,13 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 
 	watcherExecutable := selectWatcherExecutable(paths.Top(), previous, current)
 
+	rollbackWindow := u.upgradeSettings.Rollback.Window
+	if !isRollbackWindowFeatureAvailable(parsedVersion) {
```
pchila, line 348: Maybe add a small comment here as to why we check the new agent version before passing a rollback window argument

internal/pkg/agent/application/upgrade/rollback.go
pchila: Suggestion on the suggestion: `isRollbackWindowSupported` ;)

ycombinator: We have a concept of feature flags in Agent (see `agent.features.*` settings in Agent configuration) and the rollback window isn't a feature flag in that sense so I don't want to name this function `rollbackWindowFeatureFlag` in case that would cause confusion.  I'll go with `isRollbackWindowSupported`.

internal/pkg/agent/application/application.go
ycombinator: Good question. I don't know the history here but I agree it would be more convenient to pass these settings more dynamically, i.e. right as the upgrade is about to happen, so we don't need an Agent restart if they are changed.  Let me look into how complicated it would be to make such a change; if it's not too bad, I'll include it in this PR, otherwise I'll make a second PR for it that would become a pre-requisite to this PR here.

internal/pkg/agent/application/upgrade/rollback.go
ycombinator: Thinking about this some more, I'm not sure we even need this function and the check that calls it.  

The code that calls the Upgrade Watcher, and therefore the code that could pass `--rollback-window` to the `elastic-agent watch` command, is always run from the Agent that's being upgraded (let's call this the _source_ Agent, as opposed to the Agent that we're upgrading to, the _target_ Agent).

Further, starting from Elastic Agent version 8.14, Agent will always run the Upgrade Watcher from the more recent Agent version of the two versions involved in the upgrade process.

Putting those two facts together, I can't see a scenario where the source Agent has the capability to pass `--rollback-window` to the `elastic-agent watch` command but the Upgrade Watcher wouldn't be guaranteed to support the rollback window feature.

To clarify, here are all possible upgrade scenarios.  Assumption: the entire rollback window feature is implemented and released in 9.1.0:

| Source agent version | Target agent version | Upgrade Watcher selected | Source agent knows how to pass `--rollback-window`? | Upgrade Watcher supports rollback window feature? |
| ------ | ----- | ----- | ----- | ----- | 
| < 8.14 | < 9.1.0 | Source Agent | No | No |
| < 8.14 | >= 9.1.0 | Source Agent | No | No |
| >= 8.14 but < 9.1.0 | < 9.1.0 | Newer Agent (>= 8.14 but < 9.1.0) | No | No |
| >= 8.14 but < 9.1.0 | >= 9.1.0 | Newer Agent  (>= 9.1.0) | No | Yes |
| >= 9.1.0 | < 9.1.0 | Newer Agent (>= 9.1.0) | Yes | Yes |
| >= 9.1.0 | >= 9.1.0 | Newer Agent  (>= 9.1.0) | Yes | Yes |

Looking at all the scenarios, one of these two statements is always true:
* when the source Agent version < 9.1.0, the source Agent doesn't know how to pass `--rollback-window` to the `elastic-agent watch` command.  So it doesn't matter if the Upgrade Watcher supports the rollback window feature.
* when the source Agent version >= 9.1.0, the Upgrade Watcher that's selected is guaranteed to support the rollback window feature.

WDYT @kaanyalti @pchila — do you agree it's OK to remove this function and the associated check that calls it?

internal/pkg/agent/application/upgrade/rollback_test.go
ycombinator: Simplified in 50c2675057.

internal/pkg/agent/application/upgrade/rollback.go
ycombinator: I can't recall why I did this but I can't think of any reasons seconds is necessarily better.  Changed to use Duration.String() in 9f87950cdaf4702a6e1f8ec51a77e1a427dc8fc7.

internal/pkg/agent/application/upgrade/rollback_test.go
ycombinator: Done in 9f87950cdaf4702a6e1f8ec51a77e1a427dc8fc7.

internal/pkg/agent/application/upgrade/upgrade.go
ycombinator: This may no longer be necessary; see https://github.com/elastic/elastic-agent/pull/8177#discussion_r2133989584

internal/pkg/agent/cmd/watch.go
ycombinator: Sorry, I worded my TODO comment poorly.  What I meant is that the `rollbackWindow` variable here will be used as part of implementing https://github.com/elastic/elastic-agent/issues/6884.  I've clarified in 8c943cb932.

internal/pkg/agent/application/application.go
ycombinator: > instead of passing the configuration when the upgrade is happening ?

Just to make sure I'm understanding you correctly here: you are suggesting we should _re-read the Agent configuration file from disk_ and use it when the upgrade is happening?

internal/pkg/agent/application/upgrade/rollback.go
pchila: @ycombinator there's also the cleanup part performed by the agent main process at the end of the rollback window. 
If we take for example an explicit downgrade from 9.1.0 to 8.18.0 the watcher would support the rollback window but the main agent process will not: in such a case the old install cleanup must happen as soon as the watcher is done at the end of the grace period.

I still think this function is needed as it will act as a check whether the elastic-agent main process is able to trigger a delayed cleanup or not.

ycombinator: @pchila Thanks, yes, I forgot this new feature also involves a delayed cleanup by the main elastic-agent process.  So, yes, this function is still needed for that.  

However, I'm not sure this file, `internal/pkg/agent/application/upgrade/rollback.go` is the best place for it anymore.  Also, I'm not sure it makes sense to introduce this function in this PR here as it doesn't have to do with the delayed cleanup.  Are you okay with introducing this function in the delayed cleanup PR?

internal/pkg/agent/application/application.go
pchila: Not necessarily read it from the configuration file on disk, rather lookup the current agent configuration *loaded in memory* and use the values we see at the start of the upgrade process

internal/pkg/agent/application/upgrade/rollback.go
pchila: This is needed to decide whether to pass the rollback window parameter to the watch process depending on the target version of the upgrade. I would rather have that parameter implementation and tests here and those tests should include the downgrade case I described above. If it's too big a change we can separate it but I would still have it refer issue https://github.com/elastic/elastic-agent/issues/6880

If we move it to another issue whoever is gonna work on https://github.com/elastic/elastic-agent/issues/6882 or https://github.com/elastic/elastic-agent/issues/6886 or https://github.com/elastic/elastic-agent/issues/6884 has to come back here and reintroduce this code.

ycombinator: Chatted with @pchila off-PR.  We need to keep this function but it should be called with the target version, not the watcher version.  This is to handle the downgrade scenario, e.g. from 9.1.0 to 8.18.0.  In this scenario, the watcher version that is selected will be 9.1.0.  If the `--rollback-window` parameter were to be passed to this watcher, it would write the rollback window value to the ugprade marker and exit without cleaning up.  However, then the 8.18.0 Agent would start up and, since it is unaware of the rollback window feature, it would not schedule cleanup either!

internal/pkg/agent/application/upgrade/upgrade.go
ycombinator: Added in 39ac19916c.

internal/pkg/agent/application/application.go
ycombinator: See bd961a3a95.  It's not the same as passing the current configuration to the `Upgrade` method each time it's called, but I believe it has the same effect — the configuration in the `Upgrader` is `Reload`ed whenever the configuration changes.

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -238,3 +242,24 @@ func restartAgent(ctx context.Context, log *logger.Logger, c client.Client) erro
...
+
+func makeBaseWatchCmd(agentExecutable string, rollbackWindow time.Duration) *exec.Cmd {
+	cmdArgs := []string{
+		watcherSubcommand,
+		"--path.config", paths.Config(),
+		"--path.home", paths.Top(),
+	}
+
+	if rollbackWindow > 0 {
+		cmdArgs = append(cmdArgs, "--rollback-window", fmt.Sprintf("%s", rollbackWindow.String()))
```
pchila, line 254:
```suggestion
		cmdArgs = append(cmdArgs, "--rollback-window", rollbackWindow.String())
```

ycombinator:
@pchila @kaanyalti Thank you for your patience with this PR as I shifted my focus to other, more time-critical work (FIPS).  This PR is ready for re-review now, when you get a chance.

pchila:
> @pchila @kaanyalti Thank you for your patience with this PR as I shifted my focus to other, more time-critical work (FIPS). This PR is ready for re-review now, when you get a chance.

@ycombinator can we put this one on hold for the moment? The discussion around locking for the upgrade marker may have impacts on the distributions of responsibilities (we may even want the elastic-agent main process to handle the writing of the rollback window, in which case we don't need to pass an extra parameter to the watcher)

pchila:
Superseded by #8767

PR #8199, opened 2025-05-21
Create gateway.yml sample

theletterf:
This adds a sample configuration file for a standalone (nonK8s) Collector in Gateway mode. Configuration is based on the K8s one from https://github.com/elastic/elastic-agent/blob/e15fb713bde0254af089eccc09c01d9e00fa6f68/deploy/helm/edot-collector/kube-stack/values.yaml#L557


internal/pkg/otel/samples/linux/gateway.yml
```diff
@@ -0,0 +1,101 @@
...
+    table:
+      - context: metric
+        statement: route() where instrumentation_scope.name == "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver" or IsMatch(instrumentation_scope.name, "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/*")
+        pipelines: [metrics/infra/ecs, metrics/otel]
+  elasticapm: {} # Elastic APM Connector
+
+processors:
+  elasticinframetrics:
+    add_system_metrics: true
+    add_k8s_metrics: true
```
LikeTheSalad, line 25: Do we need these lines for this use case?

LikeTheSalad: Apart from these lines, everything else looks good to me btw 👍

theletterf: Probably not, but then why were they in the Gateway config for k8s? 🤔 @AlexanderWert do you know?

internal/pkg/otel/samples/linux/gateway.yml
```diff
@@ -0,0 +1,101 @@
...
+      receivers: [routing]
+      processors:
+        - elasticinframetrics
+        - attributes/dataset
+        - resource/process
+        - batch/metrics
+      exporters: [debug, elasticsearch/ecs]
+    metrics/otel:
+      receivers: [routing]
+      processors: [batch]
```
carsonip, line 85:
```suggestion
      processors: [batch/metrics]
```
- any metrics pipeline to elasticsearch exporter with send_batch_max_size is not going to end well.
- Did you get this file / line from anywhere? It is going to be a problem (read: bug) over there as well.

internal/pkg/otel/samples/linux/gateway.yml
```diff
@@ -0,0 +1,101 @@
...
+  elasticsearch/otel:
+    endpoints:
+      - ${ELASTIC_ENDPOINT} # Will be populated from environment variable
+    api_key: ${ELASTIC_API_KEY} # Will be populated from environment variable
+    logs_dynamic_index:
+      enabled: true
+    metrics_dynamic_index:
+      enabled: true
+    traces_dynamic_index:
+      enabled: true
```
carsonip, line 58: what version is this sample for? From 0.122.0 these configs are deprecated and become no-op. https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.122.0

theletterf: Not sure.

theletterf:
@carsonip To provide some context: I've adapted this from the K8s file, as we lacked a standalone Gateway config. None of the values in here are opinionated from my side: We have the opportunity of modifying everything as you see fit here. As soon as we have a good base config, we can post it to the docs.

internal/pkg/otel/samples/linux/gateway.yml
carsonip: as this is going to be backported to all active versions, let's keep this first.

AlexanderWert (approved):

internal/pkg/otel/samples/linux/gateway.yml
```diff
@@ -0,0 +1,100 @@
...
+
+connectors:
+  # Routing connector splits infra metrics from other metrics and routes them to ECS-based pipeline
+  # Other metrics are exported in OTel-native format to Elasticsearch
+  routing:
+    default_pipelines: [metrics/otel]
+    error_mode: ignore
+    table:
+      - context: metric
+        statement: IsMatch(instrumentation_scope.name, "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/*")
```
rogercoll, line 18: Looks like the collector does not like this syntax:

```
Error: failed to build pipelines: editor names must start with a lowercase letter but got 'IsMatch'
2025/05/27 10:13:29 collector server run finished with error: failed to build pipelines: editor names must start with a lowercase letter but got 'IsMatch'
```

Checking how can we fix it

carsonip (approved):

internal/pkg/otel/samples/linux/gateway.yml
rogercoll: > Will this put a higher load on the daemonset because of the overhead of shipping small metrics?

That's right, some benchmarking would help. Another benefit would be to make the daemonset configs more Elastic agnostic, `send_batch_max_size > 0` option looks tied to the Elasticsearch exporter. A better option could be switching the batchers, so the gateway prevents splitting the metrics: (otlp metrics -> daemon (batch) -> gateway (batch/metrics) -> elasticsearch)
 

Created this issue to fix the batching metrics config https://github.com/elastic/opentelemetry-dev/issues/922

carsonip: it does not.

carsonip: > A better option could be switching the batchers

I'm against any sort of send_batch_max_size > 0 for metrics.

carsonip: > I'm against any sort of send_batch_max_size > 0 for metrics.

Reason being there's no guarantee that batch A split into batch [A1, A2] are going to be merged back as A in another collector. A1 could be merged with something else, and A2 would then be sent separate to A1, which will be detected as a duplicate.

rogercoll: My bad, I meant `send_batch_max_size: 0`.

carsonip (commented):
why does it have `backport-skip` and `backport-active-all` at the same time? It should be either or.

PR #8212, opened 2025-05-22
[Test PR]Test fips arm

pazone:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...


.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -24,6 +24,15 @@ fi
...
 # Make sure that all tools are installed
 asdf install
+if [ "$FIPS" == "true" ]; then
+  echo "~~~FIPS: Checking msft-go is installed"
+  if GOEXPERIMENT=systemcrypto go version &>/dev/null; then
+    echo "msft-go is used"
+  else
+    echo "ERROR: Wrong Golang distro is detected"
+    exit 1
+  fi
```
v1v, line 39: Do you think running without the if statements is enought? IIUC, `set -e` is available hence if it failed it will report the error.
```suggestion
  GOEXPERIMENT=systemcrypto go version
```

.tool-versions
```diff
@@ -1,3 +1,2 @@
 mage 1.14.0
-golang 1.24.0
```
v1v, line 2: Will this have any impact?

pazone: We use `.go-version` file to get the golang version. (the pre-command hooks again)

.buildkite/scripts/buildkite-integration-tests.sh
pazone: I'd prefer to keep a literal message to prevent possible uncertainty regarding the error message. Is it a problem if we keep the error message?

v1v: We can enhance clarity and simplify our process:

1. The line `echo "~~~FIPS: Checking msft-go is installed"` effectively serves as a self-contained log group in Buildkite. 
2. Therefore, it may be more efficient to run the command directly rather than adding an if/else structure for echoing and then exiting with an error code. The command `OEXPERIMENT=systemcrypto go version` could yield a different error level, which might provide additional insights.
3. Instead of using `&> /dev/null`, we should consider leaving the output visible, as it may contain important details that could be relevant for our analysis.
4. If a failure occurs, it's related to some misconfiguration anyway.

.tool-versions
v1v: I see, so you mean that `golang 1.24.0` should never have been added to `.tool-versions`? I see it was introduced at https://github.com/elastic/elastic-agent/pull/5777, Do we know the reason?

PR #8248, opened 2025-05-27
feat: hybrid elastic agent invoke collector as subprocess

pkoutsovasilis:
## What does this PR do?

This PR introduces an architectural enhancement to how the Elastic Agent launches and supervises its internal OpenTelemetry (OTEL) collector:

- **Refactors the OTEL manager** to run the collector as a supervised subprocess rather than embedding it directly.
~~- Adds a new `edot-supervised` sub-command to the Elastic Agent, enabling delegated startup of the OTEL collector with direct integration to the supervisor.~~ After review comments: consolidated with the existing `otel` command
- Enables support for the `healthcheckv2` extension in the OTEL configuration to enhance lifecycle and health signal handling from collector components.
- **Improves test coverage and reliability** by refactoring the `otelmanager` test logic to use time-based event checks, ensuring better guarantees.

This work sets the foundation for better fault isolation, process recovery, and alignment with existing process invocation model of elastic agent.

## Why is it important?

Running the collector as a supervised subprocess provides improved fault isolation; a crash in the collector won't directly affect the main agent process.

## Checklist

## Disruptive User Impact

No immediate user impact is expected. The functionality remains the same as when running the collector directly inside the agent process. The new supervision model is internal and gated behind the `edot-supervised` command.

```
mage unitTest
```

## Related issues

N/A

### DIFF
--- a/internal/pkg/otel/manager/manager.go
+++ b/internal/pkg/otel/manager/manager.go
@@ -44,8 +44,8 @@ type collectorRecoveryTimer interface {
 type OTelManager struct {
 	// baseLogger is the base logger for the otel collector, and doesn't include any agent-specific fields.
 	baseLogger *logger.Logger
-	logger *logger.Logger
-	errCh  chan error
+	logger     *logger.Logger
+	errCh      chan error
 
 	// The current configuration that the OTel collector is using. In the case that
 	// the cfg is nil then the collector is not running.
@@ -98,7 +98,7 @@ func NewOTelManager(logger *logger.Logger, logLevel logp.Level, baseLogger *logg
 
 	return &OTelManager{
 		logger:        logger,
-		baseLogger: baseLogger,
+		baseLogger:    baseLogger,
 		errCh:         make(chan error, 1), // holds at most one error
 		cfgCh:         make(chan *confmap.Conf),
 		statusCh:      make(chan *status.AggregateStatus),
--- a/internal/pkg/otel/manager/manager_test.go
+++ b/internal/pkg/otel/manager/manager_test.go
@@ -469,7 +469,7 @@ func TestOTelManager_Run(t *testing.T) {
 			base, _ := loggertest.New("otel")
 			m := &OTelManager{
 				logger:        l,
-				baseLogger: base,
+				baseLogger:    base,
 				errCh:         make(chan error, 1), // holds at most one error
 				cfgCh:         make(chan *confmap.Conf),
 				statusCh:      make(chan *status.AggregateStatus),
--- a/internal/pkg/agent/application/coordinator/coordinator_test.go
+++ b/internal/pkg/agent/application/coordinator/coordinator_test.go
@@ -1078,7 +1078,7 @@ func createCoordinator(t testing.TB, ctx context.Context, opts ...CoordinatorOpt
 	cfg.Port = 0
 	rm, err := runtime.NewManager(l, l, ai, apmtest.DiscardTracer, monitoringMgr, cfg)
 	require.NoError(t, err)
-	otelMgr, err := otelmanager.NewOTelManager(l, logp.InfoLevel, l,otelmanager.EmbeddedExecutionMode)
+	otelMgr, err := otelmanager.NewOTelManager(l, logp.InfoLevel, l, otelmanager.EmbeddedExecutionMode)
 	require.NoError(t, err)
 
 	caps, err := capabilities.LoadFile(paths.AgentCapabilitiesPath(), l)


internal/pkg/otel/manager/healthcheck.go
```diff
@@ -0,0 +1,225 @@
...
+		},
+		ComponentStatusMap: make(map[string]*status.AggregateStatus),
+	}
+}
+
+// injectHeathCheckV2Extension injects the healthcheckv2 extension into the provided configuration.
+func injectHeathCheckV2Extension(conf *confmap.Conf, httpHealthCheckPort int) error {
+	err := conf.Merge(confmap.NewFromStringMap(map[string]interface{}{
+		"extensions": map[string]interface{}{
+			healthCheckExtensionName: map[string]interface{}{
```
swiatekm: We should probably add a uuid as the extension id here. It's possible to get a conflict with user-defined extensions, and I don't think we actually care about the id.

internal/pkg/agent/cmd/otel_supervised.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/elastic-agent/internal/pkg/cli"
+	"github.com/elastic/elastic-agent/internal/pkg/otel"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+)
+
+func newOtelSupervisedCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Command {
```
swiatekm: Is this command necessary? It looks like the only meaningful change between it and the normal `otel` command is the config provider. And that can be specified via a command-line argument. Do we want to avoid having this config provider in the normal `otel` command?

internal/pkg/otel/agentprovider/provider.go
```diff
@@ -7,106 +7,80 @@ package agentprovider
...
+
+	retrieved, err := confmap.NewRetrievedFromYAML(configBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config from buffer: %w", err)
+	}
+
+	conf, err := retrieved.AsConf()
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert config to confmap: %w", err)
 	}
```
swiatekm: Is there a reason not to put this logic in `Retrieve`? That's idiomatic for config providers.

internal/pkg/agent/cmd/otel_supervised.go
pkoutsovasilis: I see your point and to me it makes total sense to keep this as one through the `otel` command. I have to sync with certain folks and check the feasibility of doing so, but good point 🙂

internal/pkg/otel/agentprovider/provider.go
pkoutsovasilis: My understanding is that `Retrieve` can be invoked at any time, especially from the watcher callback (e.g., on SIGHUP or other reload triggers). If we were to move the reading/parsing logic into `Retrieve`, we’d need to re-read from the io.Reader - which in our case is stdin. Typicall reading from stdin is a one-shot stream and on subsequent calls, it would likely return EOF or empty data, resulting in an empty config being returned. This means that a SIGHUP (or similar reload signal) could effectively wipe the config from the collector, which we want to avoid?! By loading and parsing once in NewProvider, we ensure the configuration is fixed and reliably available for the lifetime of the provider. wdyt?

swiatekm: Allright, makes sense. I was thinking about this from the perspective of turning this provider into a generic "stdin" provider which we could upstream, but that can wait until later.

internal/pkg/agent/cmd/otel_supervised.go
pkoutsovasilis: ok the two commands are now consolidated under the `otel` command (commit 14c87c3e805295850b0d4a19a442e07b8a0edbed) let me know your thoughts 🙂

internal/pkg/otel/manager/healthcheck.go
pkoutsovasilis: ok added the a uuid as the name of the extension (commit a8f0d2c9d4d89a34d32af7ad352b48f2379cfa6e) thanks for the assist here

internal/pkg/otel/manager/common_test.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, errors.New("some error")
+	}
+	_, err = findRandomTCPPort()
+	require.Error(t, err)
+
+	netListen = func(string, string) (net.Listener, error) {
+		return &mockListener{0}, nil
+	}
+	_, err = findRandomTCPPort()
+	require.Error(t, err)
```
ycombinator: Maybe also assert that the error in this case is `failed to find random port`?

internal/pkg/agent/cmd/otel_flags.go
```diff
@@ -19,6 +19,7 @@ import (
 const (
 	otelConfigFlagName = "config"
 	otelSetFlagName    = "set"
+	otelSetSupervised  = "supervised"
```
ycombinator: I think this should probably be renamed to follow the other flags' naming convention?
```suggestion
	otelSupervisedFlagName  = "supervised"
```

internal/pkg/agent/cmd/otel_flags.go
```diff
@@ -28,6 +29,11 @@ func setupOtelFlags(flags *pflag.FlagSet) {
 	flags.StringArray(otelSetFlagName, []string{}, "Set arbitrary component config property. The component has to be defined in the config file and the flag"+
 		" has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s")
 
+	flags.Bool(otelSetSupervised, false, "Set the that this collector is running as supervised.")
```
ycombinator:
```suggestion
	flags.Bool(otelSetSupervised, false, "Set that this collector is running as supervised.")
```

internal/pkg/agent/cmd/otel.go
```diff
@@ -57,7 +69,33 @@ func hideInheritedFlags(c *cobra.Command) {
 	})
 }
 
-func runCollector(cmdCtx context.Context, configFiles []string) error {
+func RunCollector(cmdCtx context.Context, configFiles []string, supervised bool) error {
+	// NewForceExtensionConverterFactory is used to ensure that the agent_status extension is always enabled.
```
ycombinator: Could you clarify this comment?  I didn't find any references to `NewForceExtensionConverterFactory` in the code so I'm trying to understand this comment.  Thanks.

internal/pkg/agent/cmd/otel.go
```diff
@@ -57,7 +69,33 @@ func hideInheritedFlags(c *cobra.Command) {
 	})
 }
 
-func runCollector(cmdCtx context.Context, configFiles []string) error {
+func RunCollector(cmdCtx context.Context, configFiles []string, supervised bool) error {
```
ycombinator: IIUC, with the introduction of `supervised` here, we are deciding what the `settings` should be depending on whether the collector is run as a standalone process or as a process that's supervised by Elastic Agent.  I wonder if it might be cleaner to pass `settings` to `RunCollector` instead of `configFiles` and `supervised` and extract the creation of `settings` into it's own function as well, so it could be unit tested.  WDYT?

internal/pkg/otel/agentprovider/provider.go
```diff
@@ -7,106 +7,80 @@ package agentprovider
...
 const schemeName = "elasticagent"
 
-// Provider is a fixed provider that has a factory but only returns the same provider.
-type Provider struct {
-	uri string
-
-	cfg     *confmap.Conf
-	cfgMu   sync.RWMutex
-	updated chan struct{}
+var _ confmap.Provider = (*BufferProvider)(nil)
```
ycombinator: Could you add a comment here explaining why this is necessary?

internal/pkg/otel/manager/common.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"net"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/status"
+)
+
+// for testing purposes
+var netListen = net.Listen
+
+// reportErr reports an error to the error channel, after draining it.
```
ycombinator: Can you explain here why draining `errCh` first is needed?

internal/pkg/otel/manager/healthcheck.go
```diff
@@ -0,0 +1,231 @@
...
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("http://localhost:%d/%s?verbose",
+		httpHealthCheckPort, healthCheckHealthStatusPath), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create http request: %w", err)
+	}
+
+	client := http.DefaultClient
+	res, err := client.Do(req)
```
ycombinator: Nit: only use of `client`:
```suggestion
	res, err := http.DefaultClient.Do(req)
```

internal/pkg/otel/manager/healthcheck.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+}
+
+// injectHeathCheckV2Extension injects the healthcheckv2 extension into the provided configuration.
+func injectHeathCheckV2Extension(conf *confmap.Conf, httpHealthCheckPort int) error {
+	nsUUID, err := uuid.NewV4()
+	if err != nil {
+		return fmt.Errorf("cannot generate UUID V4: %w", err)
+	}
+	healthCheckExtensionID := fmt.Sprintf("healthcheckv2/%s", nsUUID.String())
```
ycombinator: Not sure what the first part of the ID (the part to the left of the `/`) is supposed to mean but would it make sense to use the `healthCheckExtensionName` constant here?

ycombinator (commented):
I left a few comments/questions on the code but I have a much broader question about the implementation:

We already have a way to run Agent-supervised components (e.g. Beats) as sub-processes using the [`commandRuntime`](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L75-L76).  With the implementation here, we are introducing a second way for Agent to run a supervised sub-process.  Could we reimplement this using the `commandRuntime` or is that something you already looked into and decided not to go that way for some reason?

internal/pkg/otel/manager/runtime.go
```diff
@@ -0,0 +1,198 @@
...
+		// collector path is required
+		return nil, errors.New("no collector path provided")
+	}
+
+	if _, err := os.Stat(collectorPath); err != nil {
+		// we cannot access the collector path
+		return nil, fmt.Errorf("cannot access collector path: %w", err)
+	}
+
+	httpHealthCheckPort, err := findRandomTCPPort()
```
blakerouse: We have been wanting to switch the GRPC protocol between its subprocess from a local net port to a unix socket or npipe (on Windows). We have not done that change only because of endpoint. There is a host of issues that come with using a net port, like IP tables rules and things that could cause the Elastic Agent not be able to hit this endpoint.

If it is possible (might not be at the moment), but we should try to use a unix socket or npipe. That would provide the best case and remove this host of issues.

See: https://github.com/elastic/elastic-agent/issues/3998

I just wanted to make a note of it here. Not something that needs to be done in the PR, but I do think its something that we should get right now instead of having to fix it later. A follow up issue once this is merged to track it might be best.

pkoutsovasilis:
Hey @ycombinator 👋  First thanks for the review. Now, the main reason behind this choice, is that [commandRuntime](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L75-L76) is "tightly" coupled with a component.Component, e.g. look at [newCommandRuntime](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L117). That said, OtelManager, at least as of now, knows nothing about [component.Component](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/cmd/logs.go#L53) and in reality the config passed to it derives from multiple component.Component. On top of that, [func (c *commandRuntime) Run(ctx context.Context, comm Communicator)](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L148C1-L149C1) is let's say architectured around Unit states which is something again irrelevant to the Otel-side of things. All of these points made me decide to follow the initial design and try to stay confined inside the `internal/pkg/otel/manager` package. So, all that said, do you still think there is a point trying to leverage and adjust accordingly the `commandRuntime` for this? 🙂

internal/pkg/otel/manager/runtime.go
pkoutsovasilis: Hey @blakerouse 👋 thanks for the comment, I agree 100% that sounds like a wise improvement to make. Currently, with the upstream code of healthcheckv2 extension, I believe this isn't possible because the httpServer drills down to `tcp` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/confighttp.go#L401C2-L401C46). Please do have a look and tell me if I miss something. But sure a follow-up issue for further investigation/upstream implementation feels right 🙂

internal/pkg/otel/manager/healthcheck.go
pkoutsovasilis: this is deliberate more context [here](https://github.com/elastic/elastic-agent/pull/8248#discussion_r2121076122) 🙂

internal/pkg/otel/manager/healthcheck.go
pkoutsovasilis: pushed this commit https://github.com/elastic/elastic-agent/pull/8248/commits/19cf3c38b035dc14476fa0b360f75544583e370a 🙂

internal/pkg/otel/manager/common.go
pkoutsovasilis: added in this commit https://github.com/elastic/elastic-agent/pull/8248/commits/aba027ec5b28c8e7a2f46c31b60633be33a60b80

internal/pkg/otel/agentprovider/provider.go
pkoutsovasilis: added in this commit https://github.com/elastic/elastic-agent/pull/8248/commits/aba027ec5b28c8e7a2f46c31b60633be33a60b80

internal/pkg/agent/cmd/otel.go
pkoutsovasilis: nice catch, that was a leftover command from the code restructuring I did, removed in this commit https://github.com/elastic/elastic-agent/pull/8248/commits/aba027ec5b28c8e7a2f46c31b60633be33a60b80

internal/pkg/otel/manager/common_test.go
pkoutsovasilis: added in this commit https://github.com/elastic/elastic-agent/pull/8248/commits/5dadcbc9df365b45b6f2160b62439d4f46c9e7be

internal/pkg/agent/cmd/otel_flags.go
pkoutsovasilis: reworded in https://github.com/elastic/elastic-agent/pull/8248/commits/9887d603c22242d295618b90a5cca3e8b8199168

internal/pkg/agent/cmd/otel_flags.go
pkoutsovasilis: reworded in https://github.com/elastic/elastic-agent/pull/8248/commits/9887d603c22242d295618b90a5cca3e8b8199168

internal/pkg/agent/cmd/otel.go
pkoutsovasilis: IMO, `RunCollector` is part of the cmd and thus having it as a func that we pass essentially the parameters that derive from the cli args, makes sense. The actual otel collector invoking func is `otel.Run` where we do pass settings directly. That said, I do see the point for unit-testing, thus I did split out to a separate func the generation of setting and added some unit-testing for it here https://github.com/elastic/elastic-agent/pull/8248/commits/a9d34362a15978d1c8436b557045fc8f3a1d8655

internal/pkg/otel/manager/healthcheck.go
swiatekm: The idiomatic way to create this id would be:

```go
	componentType := component.MustNewType(healthCheckExtensionName)
	healthCheckExtensionID := component.NewIDWithName(componentType, nsUUID.String())
	err = conf.Merge(confmap.NewFromStringMap(map[string]interface{}{
		"extensions": map[string]interface{}{
			healthCheckExtensionID.String(): map[string]interface{}{
```

But this is a small nit.

pkoutsovasilis: added the idiomatic way in this 6b3fd98a3ef14c1ad9fb2e60ceaf5d597a263aa0

pkoutsovasilis:
ok something has changed in the status reporting with all the changes that I merged from main, investigating 👀

ycombinator:
> Hey @ycombinator 👋 First thanks for the review. Now, the main reason behind this choice, is that [commandRuntime](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L75-L76) is "tightly" coupled with a component.Component, e.g. look at [newCommandRuntime](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L117). That said, OtelManager, at least as of now, knows nothing about [component.Component](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/cmd/logs.go#L53) and in reality the config passed to it derives from multiple component.Component. On top of that, [func (c *commandRuntime) Run(ctx context.Context, comm Communicator)](https://github.com/elastic/elastic-agent/blob/f9b9359c5c2a671ff57a36d9ed93c2584c3d671b/pkg/component/runtime/command.go#L148C1-L149C1) is let's say architectured around Unit states which is something again irrelevant to the Otel-side of things. All of these points made me decide to follow the initial design and try to stay confined inside the `internal/pkg/otel/manager` package. So, all that said, do you still think there is a point trying to leverage and adjust accordingly the `commandRuntime` for this? 🙂

Thanks for the explanation.  Sounds like it would be significant effort to refactor the commandRuntime interface to make it work for the collector sub-process as well.  Okay, let's not do this just yet.  At the end of the day, it's an implementation detail, so we can always revisit the implementation at a later time if there's a need to do it.

internal/pkg/otel/manager/runtime.go
blakerouse: I do look like the GRPC endpoint of the healthcheckv2extension has an endpoint setting https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/configgrpc.go#L175

It references https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confignet/confignet.go#L73 which does support `unix` socket. Does look like `npipe` is missing and something that otel probably needs to add.

Using GRPC is probably better actually because it has a `Watch` endpoint removing the need to poll the HTTP endpoint. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckv2extension/internal/grpc/grpc.go#L50

pkoutsovasilis: @blakerouse yes we could use the `Watch` endpoint as the "trigger" to get the HTTP status endpoint as the latter is the only one providing us with a `verbose` look since GRPC [link](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/healthcheckv2extension#grpc-service)
> does not provide the additional detail available with the HTTP service reports through `Watch` reports 

On top of that there is another reason which I have to create an issue about upstream for not using the GRPC `Watch` endpoint just yet 🙂

internal/pkg/otel/manager/healthcheck.go
```diff
@@ -0,0 +1,232 @@
...
+func (e *healthCheckEvent) Status() componentstatus.Status { return e.status }
+func (e *healthCheckEvent) Timestamp() time.Time           { return e.timestamp }
+func (e *healthCheckEvent) Err() error                     { return e.err }
+
+// AllComponentsStatuses retrieves the status of all components from the health check endpoint.
+func AllComponentsStatuses(ctx context.Context, httpHealthCheckPort int) (*status.AggregateStatus, error) {
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("http://localhost:%d/%s?verbose",
```
cmacknz, line 80: Any way to use a unix socket or named pipe for this? We occasionally see problems with local firewall rules with the control protocol. You have already dealt with the port conflict problem.

internal/pkg/otel/manager/manager.go
```diff
@@ -39,54 +38,85 @@ type OTelManager struct {
 	// doneChan is closed when Run is stopped to signal that any
 	// pending update calls should be ignored.
 	doneChan chan struct{}
+
+	// collectorBinaryPath is the path to the collector executable
+	collectorBinaryPath string
```
cmacknz: Should this come from a spec file, or the code representation of one?

cmacknz: Maybe this comes in when it becomes a separate binary

internal/pkg/otel/manager/healthcheck.go
cmacknz: Covered here already https://github.com/elastic/elastic-agent/pull/8248/files#r2130174025

cmacknz:
Kind of a similar theme to Shaunak's ask about the commandRuntime, which I understand why we don't use. 

Something else I've noticed is the collector itself isn't modelled as a component, regardless of how we supervise it. This is most obvious in diagnostics where when trying to run this (with status broken for other reasons):

```yaml
collector:
    error: 'failed to get status: Get "http://localhost:63400//health/status?verbose": dial tcp 127.0.0.1:63400: connect: connection refused'
    status: 5
    timestamp: "2025-06-06T16:06:06.566225-04:00"
components: []
fleet_message: Not enrolled into Fleet
fleet_state: 6
log_level: info
message: 1 or more components/units in a failed state
state: 3
```

This enshrines the collector as a separate thing from everything else, and it won't naturally fit into the status reporting back to Fleet. I think in general this raises the question of how we should report the status of the collector back to Fleet so that we don't break the input health reporting UI unnecessarily. With this error right now, I think Fleet would show the agent as failed but it wouldn't be able to display why.

internal/pkg/otel/manager/runtime.go
blakerouse: I think this all just highlights improvements that should be done as each part of the code is ready for it. I just wanted to highlight that using a unix socket and GRPC would be a better long term path. Not a blocker from me on using HTTP at the moment.

blakerouse:
@cmacknz @ycombinator Using the `commandRuntime` requires that the OTEL configuration is transformed into the component model and then the OTEL runtime runs under the RuntimeManager. @pkoutsovasilis is following the same path that was done before with my hybrid works and keeping it completely separate from the core runtime of how Elastic Agent runs beats.

This was done with hybrid because the original goal of the project was that in the future the Elastic Agent would become the OTEL collector. Not converting the configuration into the component model had the benefit of it being a separate thing that in the future will completely replace the Elastic Agent. It actually did very little, parse the configuration, extract only the OTEL parts and send it to the collector.

I think the overall architecture of how this is going to work has now changed. I don't believe the Elastic Agent as a supervisor will ever go away now and it will always be running and supervising the collector. If that is the case then using the component model and running the OTEL collector as a specification and running under its own commandRuntime, might be the best path forward.

To me this all comes down to how we see this working in the future, if its going to be completely replace the Elastic Agent then it should be done as this PR is doing it. If its going to always be ran as a subprocess of the Elastic Agent (feels that way now), then it should be possible be ran as a `commandRuntime`.

cmacknz:
> I think the overall architecture of how this is going to work has now changed. I don't believe the Elastic Agent as a supervisor will ever go away now and it will always be running and supervising the collector. If that is the case then using the component model and running the OTEL collector as a specification and running under its own commandRuntime, might be the best path forward.

Yes we are aligning more towards elastic-agent being an OpAMP supervisor equivalent than a collector instance.

> To me this all comes down to how we see this working in the future, if its going to be completely replace the Elastic Agent then it should be done as this PR is doing it. If its going to always be ran as a subprocess of the Elastic Agent (feels that way now), then it should be possible be ran as a commandRuntime.

I am less worried about how well this aligns with the future, and more worried about how much change we take on at once, specifically with consideration for status reporting back to Fleet about how the collector we are supervising is performing. We don't have a great way to model the single collector process in the UI regardless, but if it is a component we can at least report it's status and it will end up in the right place in the .fleet-agents document.

pkoutsovasilis:
Gents, I really appreciate the thoughtful input.

I understand the concerns around long-term alignment — IMO if and when we fully split EDOT from the Elastic Agent codebase, moving to a commandRuntime-based component would be the natural and cleanest path forward. At that point, integrating with the component model and the RuntimeManager makes perfect sense.

For this PR, though, the primary objective remains scoped to improving resilience by isolating the collector from the Elastic Agent's control plane. The subprocess approach taken here mirrors the intent of the OpAMP supervisor pattern, as originally discussed.

Regarding status reporting, I hear your point @cmacknz , especially around UI modelling and Fleet visibility. But AFAICT, this PR doesn't change the current status reporting mechanics — it simply preserves existing behaviour while ensuring faults in the collector don't bring down the agent. If there's any particular place where this PR unintentionally affects status propagation, please do point it out and I'll make sure to address it. 🙂

blakerouse:
@cmacknz This really is more about internal architecture, because reporting in the component model format to Fleet can be accomplished inside or outside of the runtime manager. I honestly don't see a danger in taking either path I just seem them as different paths. Each have pros/cons and neither to me has more pros than cons over the other.

The biggest issue I see is that the component model doesn't really map 1-1 with the OTEL model of pipelines, receivers, outputs, extensions, etc. Keeping it separate as this PR does make it easier to understand as we are not mapping between formats to make it work (other than the reporting to Fleet part).

The downside to that is that all other components will be in the component model and are not going away (most likely ever, e.g. endpoint). That is what makes me lean towards placing it into the component model, but again it doesn't map well.

@pkoutsovasilis I do agree this is more about the overall path forward and not about this PR specifically. I think the changes in this PR are a good step in the right direction, keep it working the same as the OTELManager does today but just keeps it as a subprocess. I think the points that I highlighted about using GRPC, unix/npipe sockets should be filed as issues once this is merged and be taken in steps to get to a really good stable supervisor of the OTEL collector.

internal/pkg/otel/manager/runtime.go
```diff
@@ -0,0 +1,198 @@
...
+				case currentFailedAttempts > maxFailedAttempts:
+					reportStatus(procCtx, statusCh, aggregateStatus(componentstatus.StatusFatalError, err))
+					return
+				}
+				currentFailedAttempts++
+			} else {
+				if reportedStatus := statuses.Status(); currentStatus != reportedStatus {
+					currentStatus = reportedStatus
+					reportStatus(procCtx, statusCh, statuses)
+				}
```
swiatekm: It looks like this will only emit the statuses when the overall collector status changes, which I don't think is correct. We want to emit it every time any of the substatuses changes as well, so I think we need to retain the whole struct and do the equivalent of a reflect.DeepEqual on it.

Sorry for bringing this up so late in the review process, it came to mind recently when I was working on the status translation.

pkoutsovasilis: ok my original thinking was that this aggregate status will change if an extension/pipeline transitions from OK, to any Error statuses, and vice-versa, which I still believe is correct. That said, I am not quite sure if you meant the following case with your comment, but we might get twice an aggregate status of `StatusRecoverableError` as an example but due to a different components. So yes I will make this always emitting thanks for the catch

pkoutsovasilis: pushed always emitting statuses here bc6cc28d2d60407429ac6e4d5dea7da9273aebed 🙂

pkoutsovasilis:
The fact that `TestAgentMetricsInput` fails only on Windows 2025; it doesn't fail in any other OS, Ubuntu, Debian, Windows 2022, confuses me

the first time with this weird error [[BK link](https://buildkite.com/elastic/elastic-agent/builds/22060#019755f2-d803-4c3f-a692-03b2ac84e5c7/735-1021)]
```
beat_receivers_test.go:561:
--
  | Error Trace:	C:/buildkite-agent/builds/bk-agent-prod-gcp-1749494129705210264/elastic/elastic-agent/testing/integration/beat_receivers_test.go:567
  | C:/Users/Buildkite/.go/go-1.24.3/src/runtime/asm_amd64.s:1700
  | Error:      	Received unexpected error:
  | error performing ES search: stream error: stream ID 269; PROTOCOL_ERROR; invalid header field name "g\x00e\x00oud-request-id"
```

and the second time with the one below [[BK link](https://buildkite.com/elastic/elastic-agent/builds/22060#0197560f-512e-4bf2-a4f5-fa9e4d669339/940-1226)]
```
Error Trace:	C:/buildkite-agent/builds/bk-agent-prod-gcp-1749495548527856165/elastic/elastic-agent/testing/integration/otel_test.go:1666
--
  | C:/buildkite-agent/builds/bk-agent-prod-gcp-1749495548527856165/elastic/elastic-agent/testing/integration/beat_receivers_test.go:684
  | Error:      	expected documents to be equal for metricset cpu
  | Test:       	TestAgentMetricsInput/compare_documents/cpu
  | Messages:   	ignored field "event.ingested" does not exist in either map, please remove it from the ignored fields
```

@mauri870 could you please take a look and tell me if my PR in some way, that I fail to see, causes this?

internal/pkg/otel/manager/runtime.go
swiatekm: Always emitting them is also not ideal, because it forces the coordinator loop to process them every time. What I was worried about were less changes from OK to an error status, but rather changes where a component hasn't even started. I'm not sure how healthcheckv2 calculates status in that case, and I'd feel more confident if we assumed as little as possible here.

pkoutsovasilis: What you say makes sense, but it feels that it should be part of the OTel status manipulation/transformation which IMO should live "closer" to the manager and not in the coordinator. Either way here is the emitting on an actual change aaad4f26ade15ed415c4c85f5c03b115c42e2ec4

swiatekm:
This change will also break process metric reporting. Right now, for components running as beats receivers, we drop all process metrics - after all, they run in the elastic-agent itself, and we already have those metrics for the agent process. Here, however, we move this work to a new process which is not monitored at all. So the self-monitoring configuration will need to be modified to account for this, and we'll need to make sure it looks right in the integration dashboard.

internal/pkg/otel/manager/runtime.go
swiatekm: I think it should live in the otel manager. The fact that the otel manager fetches the status periodically vs subscribing to a GRPC stream of changes is an implementation detail that its callers should not care about.

internal/pkg/otel/manager/healthcheck.go
```diff
@@ -0,0 +1,277 @@
...
+		// status doesn't match
+		return false
+	}
+
+	// NOTE: we don't check the timestamp
+	// as we care only about the event and component statuses/error differences
+	//if !s1.Timestamp().Equal(s2.Timestamp()) {
+	//	// timestamp doesn't match
+	//	return false
+	//}
```
swiatekm: This should be removed.

pkoutsovasilis: what's the issue for having this commented out?! 🙂

swiatekm: What's the purpose of keeping it? It's just noise without any value. Do you expect to uncomment it in the future?

internal/pkg/otel/manager/runtime.go
pkoutsovasilis: @swiatekm sure thing, is this now resolved? 🙂

internal/pkg/otel/manager/healthcheck.go
pkoutsovasilis: Ok let me de-noise stuff 🙂 I value keeping noise at a minimum

pkoutsovasilis: noise cancelation on here 06bb59829f12c02ed9ba53fab70f2408cd839cf1

pkoutsovasilis:
> This change will also break process metric reporting. Right now, for components running as beats receivers, we drop all process metrics - after all, they run in the elastic-agent itself, and we already have those metrics for the agent process. Here, however, we move this work to a new process which is not monitored at all. So the self-monitoring configuration will need to be modified to account for this, and we'll need to make sure it looks right in the integration dashboard.

I tried to proactively investigate and capture any issues with self-monitoring before starting with this PR and in general with the otel supervised approach; evidently neither I, nor the stakeholders I spoke with, spotted the above. This is a valid point and it might make this whole approach harder to implement in a single shot, or at least without refactoring other parts of the code.

cc @cmacknz @ebeahan

cmacknz:
> Regarding status reporting, I hear your point @cmacknz , especially around UI modelling and Fleet visibility. But AFAICT, this PR doesn't change the current status reporting mechanics — it simply preserves existing behaviour while ensuring faults in the collector don't bring down the agent. If there's any particular place where this PR unintentionally affects status propagation, please do point it out and I'll make sure to address it. 🙂

The biggest point is that if the collector process itself has an error (invalid configuration it can't read so it doesn't start or any error that wouldn't come through a pipeline component) that we have a way for that to be visible in Fleet, local agent status, and diagnostics. If this just marks the top level agent status as degraded or failed that works for now.

The other point about should the collector appear in the components list or not probably needs to be forked out of this PR (maybe control plane weekly meeting?) to avoid lots of back and forth. I am aligned with the collector sub-process being its own sub-process runtime because it doesn't use the control protocol. The topic I'm interested in is how or if we unify status reporting between the things we do know need to keep using the control protocol and the collector which doesn't. I think the most accurate thing to do is to report the collector sub-process status separate from the elastic-agent supervisor status for example. I don't think we need to finalize this as part of this PR though as long as errors are detectable.

> This is a valid point and it might make this whole approach harder to implement in a single shot, or at least without refactoring other parts of the code.

If we have to have a temporary config flag (or even just hard coded bool like the otel runtime originally was) to let us split this across PRs I think that's fine and makes sense. There isn't a hard requirement to make this change in one PR though it is easier if that is the case.

pkoutsovasilis:
Thanks everyone for the valuable input and discussion. I’ve pushed a change that gates the invocation mode of the EDOT collector behind an internal runtime variable. By default, the agent will continue to run the collector embedded, ensuring current behaviour and avoiding regressions (e.g., broken process metrics, incomplete diagnostics, etc.). We can switch to subprocess mode once it reaches feature parity with the embedded path. This should allow us to decouple the architecture change from downstream side-effects, and incrementally improve subprocess support in follow-up PRs. 🙂 

cc @cmacknz @blakerouse @ycombinator @ebeahan @swiatekm

changelog/fragments/1748852862-otel-collector-subprocess.yaml
```diff
@@ -0,0 +1,4 @@
+kind: feature
+summary: Add support for running OTEL collector as a supervised subprocess
+component: elastic-agent
+pr: https://github.com/elastic/elastic-agent/pull/8248
```
swiatekm: I don't think we want a changelog entry here. We haven't actually changed anything from the perspective of a user of elastic agent, so it hurts more than it helps to add it imo.

internal/pkg/otel/manager/healthcheck_test.go
```diff
@@ -0,0 +1,272 @@
...
+		//	},
+		//	s2: &status.AggregateStatus{
+		//		Event: &healthCheckEvent{
+		//			status:    componentstatus.StatusOK,
+		//			timestamp: timestamp.Add(time.Hour),
+		//			err:       nil,
+		//		},
+		//	},
+		//	expected: false,
+		//},
```
swiatekm: Is this intentionally left here?

internal/pkg/otel/manager/manager.go
```diff
@@ -7,19 +7,24 @@ package manager
...
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 
-	"github.com/elastic/elastic-agent/internal/pkg/otel"
-	"github.com/elastic/elastic-agent/internal/pkg/otel/agentprovider"
-	"github.com/elastic/elastic-agent/internal/pkg/release"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
+type CollectorRuntime string
```
swiatekm: This shouldn't block this PR, but I'd prefer if we could avoid using the word "runtime" here. :sweat_smile: . It's already overloaded too much in the agent codebase. Maybe something like "ExecutionMode" would be a decent synonym? Or just "Runner"?

pkoutsovasilis: You are more than right, I was too zoomed in the `otel/...` package but `runtime` is certainly used too much. Also, I get the same vibes for "Runner" so I went with "ExecutionMode"-style approach here 4200c8083436b856f7b09d253c8de6ac879b1ccc 🙂

internal/pkg/otel/manager/healthcheck_test.go
pkoutsovasilis: oh that is a noise leftover that the previous de-noising effort missed. Removed here  8e6a2afb2caa7a3db355b6e25f348bd923a7115e 🙂

changelog/fragments/1748852862-otel-collector-subprocess.yaml
pkoutsovasilis: yep not enabled by default, so good catch no changelog needed; removed here fc9ee1cf9ec72635c35fed0d59552b9f6a504e5e 🙂

internal/pkg/otel/manager/common.go
```diff
@@ -0,0 +1,62 @@
...
+	case <-ctx.Done():
+	}
+}
+
+// reportStatus sends the new status to the status channel.
+func reportStatus(ctx context.Context, statusCh chan *status.AggregateStatus, statuses *status.AggregateStatus) {
+	select {
+	case <-ctx.Done():
+		return
+	case statusCh <- statuses:
```
cmacknz, line 41: Should we do the same thing for status that we do with errors above, in only sending the latest status?

Beyond that stale status is not as useful as current status, this pattern eliminates a source of deadlocks.

internal/pkg/otel/manager/execution_embedded.go
```diff
@@ -0,0 +1,80 @@
...
+func newExecutionEmbedded() *embeddedExecution {
+	return &embeddedExecution{}
+}
+
+type embeddedExecution struct {
+}
+
+// startCollector starts the collector in a new goroutine.
+func (r *embeddedExecution) startCollector(ctx context.Context, logger *logger.Logger, cfg *confmap.Conf, errCh chan error, statusCh chan *status.AggregateStatus) (collectorHandle, error) {
+	collectorCtx, collectorCancel := context.WithCancel(context.Background())
```
cmacknz: The sub-process is wired into the parent context but this one isn't, why not? It makes the `Stop` call redundant if the context is already cancelled but that is true of the sub-process case too.

internal/pkg/otel/manager/common.go
cmacknz:
```go
		for {
			statuses, err := AllComponentsStatuses(procCtx, httpHealthCheckPort)
			if err != nil {
				switch {
				case errors.Is(err, context.Canceled):
					reportStatus(ctx, statusCh, aggregateStatus(componentstatus.StatusStopped, nil))
					return
```

This is from execution_subprocess.go and is the kind of situation where a deadlock can show up unexpectedly, if the reading of the status channel is also cancelled by context and has stopped listening. It is hard to reason that all the shutdown orderings are safe so better to just protect against it.

cmacknz: We have had at least one shutdown related deadlock in the past for where this line of thinking is coming from: https://github.com/elastic/elastic-agent-autodiscover/issues/41

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -0,0 +1,210 @@
...
+		defer healthCheckTicker.Stop()
+		for {
+			statuses, err := AllComponentsStatuses(procCtx, httpHealthCheckPort)
+			if err != nil {
+				switch {
+				case errors.Is(err, context.Canceled):
+					reportStatus(ctx, statusCh, aggregateStatus(componentstatus.StatusStopped, nil))
+					return
+				case currentFailedAttempts > maxFailedAttempts:
+					reportStatus(procCtx, statusCh, aggregateStatus(componentstatus.StatusFatalError, err))
```
cmacknz: Technically this error could be transient and not fatal, we have an SDH right now where sub-process missed checkins due to cgroup CPU throttling on k8s.

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -0,0 +1,210 @@
...
+				currentFailedAttempts++
+			} else {
+				if !compareStatuses(currentStatus, statuses) {
+					currentStatus = statuses
+					reportStatus(procCtx, statusCh, statuses)
+				}
+				// after successfully getting the status, reset the failed attempts,
+				// set the timer duration to 30 seconds and max failed attempts to 3
+				currentFailedAttempts = 0
+				timerDuration = 30 * time.Second
```
cmacknz: 30s is quite slow, with the control protocol we need "at least" 30s between check-ins and the "at least" is important. The CheckinV2 stream is bi-directional so the sub-processes can push updates back to agent as soon as they happen, so they reflect in agent immediately.

You don't have a way to do the same thing except for polling faster, just keeping this as 1s seems reasonable to me unless the increased duration is mitigating some performance impact or something.

The max failed attempts can still be set to the equivalent of 1m30s.

cmacknz (commented):
Overall this looks good, I'm generally fine with merging and then fixing the rest in follow ups.

I built this and ran it on my Mac and it didn't work however. This is the patch I used to turn the sub-process mode on:

```diff
diff --git a/internal/pkg/agent/application/application.go b/internal/pkg/agent/application/application.go
index f398f8bd76..22118b54a9 100644
--- a/internal/pkg/agent/application/application.go
+++ b/internal/pkg/agent/application/application.go
@@ -224,7 +224,7 @@ func New(
                return nil, nil, nil, errors.New(err, "failed to initialize composable controller")
        }

-       otelManager, err := otelmanager.NewOTelManager(log.Named("otel_manager"), otelmanager.EmbeddedExecutionMode)
+       otelManager, err := otelmanager.NewOTelManager(log.Named("otel_manager"), otelmanager.SubprocessExecutionMode)
        if err != nil {
                return nil, nil, nil, fmt.Errorf("failed to create otel manager: %w", err)
        }
```


I used the default configuration with monitoring enabled and `agent.monitoring._runtime_experimental: "otel"`.

The collector sub-process is omitted from the output of `elastic-agent status` and it only shows the agent is degraded without showing why.

```sh
❯ sudo elastic-development-agent status --output=full
┌─ fleet
└─ elastic-agent
   ├─ status: (DEGRADED) 1 or more components/units in a failed state
   ├─ info
   │  ├─ id: 7be9cd84-d99d-47f9-b2a0-dae8708ea180
   │  ├─ version: 9.1.0
   │  └─ commit: 99db105ea9f6d1411889c2977cf183b10d94ec00
   └─ system/metrics-default
      ├─ status: (HEALTHY) Healthy: communicating with pid '36690'
      ├─ system/metrics-default
      │  ├─ status: (HEALTHY) Healthy
      │  └─ type: OUTPUT
      └─ system/metrics-default-unique-system-metrics-input
         ├─ status: (HEALTHY) Healthy
         └─ type: INPUT
```

I have to look at diagnostics to see what is wrong in state.yaml:

```sh
collector:
    error: 'failed to get status: Get "http://localhost:49793//health/status?verbose": dial tcp 127.0.0.1:49793: connect: connection refused'
    status: 5
    timestamp: "2025-06-24T15:34:01.920287-04:00"
```

Looking at the logs I don't see anything about the sub-process starting or any mention of the randomly selected port in question. 

I do see there is a collector sub-process running though:

```
❯ ps aux | grep supervised
cmackenzie       38748   0.0  0.0 410208944   1216 s015  S+    3:43PM   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn --exclude-dir=.idea --exclude-dir=.tox supervised
root             36689   0.0  0.3 412406608 111904   ??  S     3:33PM   0:01.14 /usr/local/bin/elastic-development-agent otel --supervised
```

So the list of follow ups once I can see this working is at least:
1. Having a test that puts the agent in sub-process collector mode and shows it works so we can avoid testing this manually separately as we go through iterations.
2. Adding some more detailed logging so we can tell what is going wrong and what mode the collector is in. I don't see the sub-process PID being logged anywhere like it is for regular components for example.
3. Fixing the status reporting so it isn't hidden.

internal/pkg/otel/manager/common.go
swiatekm: Isn't it simpler to have a buffered channel for statuses? The process runtime manager delivers each status in a separate goroutine, which has the same effect as an unbuffered channel with unlimited size. Imo it's more performant and more predictable to have a buffered channel, write to it without blocking, and log an error if we can't.

pkoutsovasilis: > Should we do the same thing for status that we do with errors above, in only sending the latest status?
> Beyond that stale status is not as useful as current status, this pattern eliminates a source of deadlocks.

IMO yes this should be the same, but [the original implementation of that isn't](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/manager/manager.go#L50). For that reason having a follow-up issue about it and not including such a change in this PR is my proposal, wdyt?

> This is from execution_subprocess.go and is the kind of situation where a deadlock can show up unexpectedly, if the reading of the status channel is also cancelled by context and has stopped listening. It is hard to reason that all the shutdown orderings are safe so better to just protect against it.

buffered channel with draining ftw

> So the list of follow ups once I can see this working is at least:
> Having a test that puts the agent in sub-process collector mode and shows it works so we can avoid testing this manually separately as we go through iterations.

We do test this separately https://github.com/elastic/elastic-agent/pull/8248/files#diff-85f1e7c106b97a34527ec1327445133f3e018be75e940bb813f23f385d5547e8R205

> Adding some more detailed logging so we can tell what is going wrong and what mode the collector is in. I don't see the sub-process PID being logged anywhere like it is for regular components for example.
> Fixing the status reporting so it isn't hidden.


Now about your running issues, as you saw  today on my laptop it runs 😆 However I like your case because it allows to deal with an edge-case that appears only on your machine, so please don't change anything yet. I have added here 31892ec8cf7459e8c22b621d3599d7411d343feb some debug messages and I also pass down the logging level to the supervised collector. Thus please to compile it once more and install it with the following in your elastic-agent.yml 

```
agent.logging.level: debug
agent.monitoring._runtime_experimental: "otel"
```

and give me back the diagnostics (hopefully the issue won't be gone) 🙂

internal/pkg/otel/manager/execution_embedded.go
pkoutsovasilis: [The original implementation](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/manager/manager.go#L188) didn't wire it but since it make sense to you as well to have it wired here you are fcce06a37eadcc40e303433f9dc2f20617111392

internal/pkg/otel/manager/execution_subprocess.go
pkoutsovasilis: Switched to a `RecoverableError` here c3630f6edd5a8bc817232e86ddb4e031a8928aa8 and removed the giving up

internal/pkg/otel/manager/execution_subprocess.go
pkoutsovasilis: Ok until we have some equivalent to the control protocol for the supervised collector, what you say makes sense, the respective changes are here 8e5632bced158883c08df0de0d92b976fa9ca7be

internal/pkg/otel/manager/common.go
cmacknz: Bad news, it worked when I retried it.

pkoutsovasilis: noooooooooo 😢

internal/pkg/otel/manager/execution_subprocess.go
cmacknz: Is this actually true that it's 1.5 minutes of failed attempts?

If it times out, that's 5 seconds per timeout 130 times:

```go
func AllComponentsStatuses(ctx context.Context, httpHealthCheckPort int) (*status.AggregateStatus, error) {
	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
	defer cancel()
```

If it fails immediately, it's 130*1s which is 130s or 2m10s.

pkoutsovasilis: that why I added the comment `this should roughly match 1.5 minutes` but would you prefer this to be duration-based? 🙂

internal/pkg/otel/manager/manager_test.go
```diff
@@ -64,182 +67,375 @@ var (
...
+				e.EnsureHealthy(t, updateTime)
+
+				// no configuration should stop the runner
+				updateTime = time.Now()
+				m.Update(nil)
+				e.EnsureOffWithoutError(t, updateTime)
+			},
+		},
+		{
+			name: "subprocess collector killed outside manager",
```
cmacknz: This test should ensure the collector gets restarted when this happens. I am specifically thinking of the case where the collector sub-process is OOMKilled but the parent elastic-agent process is not.

internal/pkg/otel/manager/manager.go
```diff
@@ -95,66 +139,54 @@ func (m *OTelManager) Run(ctx context.Context) error {
 				//
 				// in the case that the configuration is invalid there is no reason to
 				// try again as it will keep failing so we do not trigger a restart
-				if cancel != nil {
-					cancel()
-					cancel = nil
-					provider = nil
-					// don't wait here for <-runErrCh, already occurred
+				if proc != nil {
```
cmacknz, line 198: > // in the case that the configuration is invalid there is no reason to
> // try again as it will keep failing so we do not trigger a restart

I see this comment was pre-existing but I there is a dangerous assumption here that the only reason a collector would exit with an error is an invalid configuration and that we shouldn't restart it because a configuration change will come in to fix it. I'm not even sure upstream can practically guarantee this assumption is true across all possible ways for a program to exit. What happens on panic for example?

It would be much safer here to restart the collector periodically (1s? maybe even shorter but with exponential backoff up to several seconds?) to confirm the error persists than to assume the error is always going to be corrected by a configuration change in the future.

Assuming I'm following this right, this could be handled in a follow up as this PR is large enough and hard to maintain.

internal/pkg/otel/manager/manager_test.go
```diff
@@ -64,182 +67,375 @@ var (
...
+						}
+					case <-timeoutCh:
+						break outer
+					}
+				}
+				assert.Error(t, err, "otel manager should have returned an error")
+			},
+		},
+		{
+			name:                "subprocess collector invalid config",
```
cmacknz, line 425: There should be a test where the collector sub-process panics, and we confirm the sub-process is automatically restarted.

cmacknz:
Thanks, the first time I read this I didn't catch manager_test.go was collapsed so I skipped reviewing all the tests 😅

internal/pkg/otel/manager/execution_subprocess.go
pkoutsovasilis: @cmacknz reworked this a bit to be way closer to 1.5 minute here a91bf7e7686826fd019ce1a1921ec643ea1acb8e 🙂

internal/pkg/otel/manager/manager_test.go
pkoutsovasilis: pushed a backoff recovery mechanism for subprocess mode here 402a98c0a6bb5c8863bd72c43e765c71e50e2a4c with tests 🙂

internal/pkg/otel/manager/manager.go
pkoutsovasilis: The backoff recovery mechanism for subprocess mode here 402a98c0a6bb5c8863bd72c43e765c71e50e2a4c now takes care of that 🙂

internal/pkg/otel/manager/manager_test.go
pkoutsovasilis: introduced a backoff recovery mechanism for subprocess mode here 402a98c0a6bb5c8863bd72c43e765c71e50e2a4c with tests where the collector panics 🙂

pkoutsovasilis:
@swiatekm I believe that there is an issue with the otel status propagation. Specifically I introduced a recovery mechanism for the subprocess execution mode here 402a98c0a6bb5c8863bd72c43e765c71e50e2a4c and then I started validating that the subprocess is indeed restarting when it gets suddenly killed. Although the recovery works as expected when I do an `elastic-agent status` afterwards I see the following:

```
Password:
┌─ fleet
└─ elastic-agent
   ├─ beat/metrics-monitoring
   │  └─ beat/metrics-monitoring
   ├─ filestream-monitoring
   │  └─ filestream-monitoring
   ├─ http/metrics-monitoring
   │  └─ http/metrics-monitoring
   └─ extensions
      ├─ status: StatusOK
      ├─ extension:healthcheckv2/305ace73-df16-44e5-b555-a0ff850bf250
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/337d233e-06da-46f5-a103-700b0ec264c0
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/43d6d95d-5319-4af3-a436-ff561eeddbab
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/7a73daa0-6651-4cfd-a7ab-a4ce30e4d367
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/c76b1d03-8d1a-484d-99b5-8d73c4468767
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/d1cce0b9-fad4-4ed9-b830-38ec8ff55ce7
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/e7324598-80fa-4e96-a01b-f5caa47c9462
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/ee08c63e-c85d-4f60-bc95-fca93ec8ba45
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/fcb98c22-9bb8-468a-b715-3873c149ea12
      │  └─ status: StatusOK
      └─ extension:healthcheckv2/fed3ce22-aabb-4893-a9c7-4b7ad1346b82
         └─ status: StatusOK
```

It seems that there are stale extension entries there. I did check my code and in every status I push out the manager I include only one extension, which hints that this is outside the manager code. Please do have a look and let me know if this is something that my code is causing

swiatekm:
@pkoutsovasilis there's definitely a bug in the status translation code related to extensions, but I can't think of how it would give you that result. You should see an otel manager error in the top-level agent status instead. :confused: 

Filed an issue for the bug here: https://github.com/elastic/ingest-dev/issues/5751.

pkoutsovasilis:
> @pkoutsovasilis there's definitely a bug in the status translation code related to extensions, but I can't think of how it would give you that result. You should see an otel manager error in the top-level agent status instead. 😕
> 
> Filed an issue for the bug here: [elastic/ingest-dev#5751](https://github.com/elastic/ingest-dev/issues/5751).

Take a look [here](https://github.com/elastic/elastic-agent/pull/8248/files#diff-a284df1af9320b1f20551cbf3c1d2772992f2dab31bccddd72d887b4b068519cR66-R69) and [here](https://github.com/elastic/elastic-agent/pull/8248/files#diff-a284df1af9320b1f20551cbf3c1d2772992f2dab31bccddd72d887b4b068519cR97-R100) I think that I filter these out, so I am a little bit surprised that they still appear in the status which is at least surprising 🙂

swiatekm:
> > @pkoutsovasilis there's definitely a bug in the status translation code related to extensions, but I can't think of how it would give you that result. You should see an otel manager error in the top-level agent status instead. 😕
> > Filed an issue for the bug here: [elastic/ingest-dev#5751](https://github.com/elastic/ingest-dev/issues/5751).
> 
> Take a look [here](https://github.com/elastic/elastic-agent/pull/8248/files#diff-a284df1af9320b1f20551cbf3c1d2772992f2dab31bccddd72d887b4b068519cR66-R69) and [here](https://github.com/elastic/elastic-agent/pull/8248/files#diff-a284df1af9320b1f20551cbf3c1d2772992f2dab31bccddd72d887b4b068519cR97-R100) I think that I filter these out, so I am a little bit surprised that they still appear in the status which is at least surprising 🙂

What you're doing there ensures the extension status stays in the collector status, which I think is correct in general. We may want to filter out the status of the healthcheckv2 extension we inject, but I'm not sure that's the right choice.

Still, I have no clue why the extension statuses of previous extensions would show up. We don't merge these values at any point. :confused:

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -128,38 +127,45 @@ func (r *subprocessExecution) startCollector(ctx context.Context, logger *logger
 		reportStatus(ctx, statusCh, currentStatus)
 		// we will check the health of the collector every 1 second for the first 10 attempts
 		// until we get a successful response.
-		maxFailedAttempts := 130 // this should roughly match 1.5 minutes of continuous failed attempts to poll healthcheck v2
-		timerDuration := 1 * time.Second
-		healthCheckTimer := time.NewTimer(timerDuration)
-		currentFailedAttempts := 0
-		defer healthCheckTimer.Stop()
+		const maxFailuresDuration = 130 * time.Second
```
cmacknz, line 130: nit: the comment above this is out of date

internal/pkg/otel/manager/manager.go
```diff
@@ -61,9 +83,10 @@ func NewOTelManager(logger *logger.Logger, logLevel logp.Level, mode ExecutionMo
 		if err != nil {
 			return nil, fmt.Errorf("failed to get the path to the collector executable: %w", err)
 		}
-
+		recoveryTimer = newRecoveryBackoff(time.Second, 10*time.Second)
```
cmacknz: 1s as the initial delay could be shorter, that's possibly enough time for data loss to occur at high data rates. 1s is a long time for a computer.

Since we have exponential backoff we should be able to go even smaller. There's no reason the first retry can't happen basically immediately.

cmacknz: If we aren't confident what this value should be we could consider making it configurable.

internal/pkg/otel/manager/manager.go
```diff
@@ -179,11 +238,15 @@ func (m *OTelManager) Run(ctx context.Context) error {
...
 				if err != nil {
+					m.logger.Errorf("collector exited with error: %v", err)
 					// failed to create the collector (this is different then
 					// it's failing to run). we do not retry creation on failure
 					// as it will always fail. A new configuration is required for
 					// it not to fail (a new configuration will result in the retry)
 					reportErr(ctx, m.errCh, err)
+					// since this is a new configuration we want to start the timer
+					// from the initial delay
+					m.recoveryTimer.ResetInitial()
```
cmacknz: Only doing `ResetInitial` on configuration change seems to mean that for rare failures at steady state (think a panic that happens every 1h-1 day) the restart duration will lengthen for each retry. I don't think this is right.

The reason to backoff is to avoid spamming the machine with a process that just exits immediately.  We probably want to reset the backoff after the collector has been healthy enough.

In standalone agents, the configuration may never change after the initial start. In Fleet managed agents, it can be days to weeks between changes. They are not a constant thing that we can rely on as the "clock" of sub-process management.

cmacknz (approved):
Thank you for the backoff mechanism, I think it could be improved a little bit. This can happen here or in a follow up so I am going to approve, and if the changes happen here I'll re-review.

internal/pkg/otel/manager/manager.go
pkoutsovasilis: nice point, I did extend the recoveryBackoff to reset to the initial backoff after some time here https://github.com/elastic/elastic-agent/pull/8248/commits/704e25e66b08edfb92646076fe62354730d6ed04. Also added a tests to validate that it actually resets to initial interval 🙂

internal/pkg/otel/manager/manager.go
pkoutsovasilis: ok I did go with an initial value of 100 nanoseconds https://github.com/elastic/elastic-agent/pull/8248/commits/704e25e66b08edfb92646076fe62354730d6ed04, at the moment these are configurable at the code level, but sure these settings can be configurable through yaml in the future 🙂

internal/pkg/otel/manager/manager.go
pkoutsovasilis: wait a bit the windows runner are doing their thing.... 🙁

cmacknz (dismissed):
🚢

PR #8249, opened 2025-05-27
[Chore] Update elastic_agent.reference

khushijain21:
## What does this PR do?
`agent.monitoring` section is enabled by default. This PR updates `elastic-agent.reference` file accordingly.

## Checklist

### DIFF
--- a/_meta/config/common.reference.p2.yml.tmpl
+++ b/_meta/config/common.reference.p2.yml.tmpl
@@ -165,18 +165,18 @@ inputs:
 
 # agent.monitoring:
 #   # enabled turns on monitoring of running processes
-#   enabled: false
+#   enabled: true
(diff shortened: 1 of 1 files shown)


internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -486,14 +486,6 @@ func (b *BeatsMonitor) injectMetricsInput(
 	metricsCollectionIntervalString string,
 	failureThreshold *uint,
 ) error {
-	if metricsCollectionIntervalString == "" {
```
swiatekm, line 489: It looks like it's fine to remove this, but it could also be removed from other methods on this struct as well. You don't need to do that in this PR, but I thought it was worth pointing out in case you missed it.

Issue #6889, opened 2025-02-17
Handle rollback command/action in upgrade flow


pchila:

ebeahan:
Current status - pending the work @pchila iterating on in https://github.com/elastic/elastic-agent/pull/8767.

Issue #6890, opened 2025-02-17
Allow elastic-agent rollback in `UPG_WATCHING` state


pchila:
Requires #6878 
Relates #6889

ebeahan:
Current status - pending the work @pchila iterating on in https://github.com/elastic/elastic-agent/pull/8767.

PR #8254, opened 2025-05-27
Lock upgrade marker

### LINKED ISSUE #6878: Lock upgrade marker before writing both in agent and agent watcher

pchila:
## What does this PR do?

Introduce file locking for modifying and reading elastic-agent update marker consistently.
FileLocker is a wrapper around github.com/gofrs/flock that exposes both blocking and non-blocking file locking.
A lock on an update marker lockfile has been introduced at creation, reading and writing: this should ensure that no concurrent read/writes happen between elastic-agent main process and watcher.
This is a prerequisite for manual rollback feature: with such a feature both process will read/write the update marker concurrently.

## Why is it important?

With the manual rollback feature that will be introduced by #6889, both elastic-agent main process and watcher need to modify the update marker. Locking before modification should avoid inconsistent writes.

## Checklist

Aside from unit tests that have been introduced under `internal/pkg/agent/application/filelock` and `internal/pkg/agent/application/upgrade/step_mark.go`, the only use case here is to perform an upgrade and verify that it succeeds.
If we want to test the file locking we can use the `testlocker` binary in `internal/pkg/agent/application/filelock/testlocker/main.go`  and lock the update marker lock file in `<elastic agent root>/data/.update-marker.lock` (holding this lock for too long may generate some failures as there's a timeout of 30s for locking operations.

## Related issues

- Closes #6878
## Questions to ask yourself

- ...


pchila:
buildkite test this

internal/pkg/agent/application/filelock/file_locker.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (fl *FileLocker) Unlock() error {
+	return fl.fileLock.Unlock()
+}
+
+type FileLockerOption func(locker *FileLocker) error
+
+// WithCustomLockedError will set a custom error to be returned
+func WithCustomNotLockedError(customError error) FileLockerOption {
```
ycombinator: I believe this option is primarily provided to make testing easier?  If so, maybe make a note here about that?

ycombinator: Never mind, I see the option is being used to provide `ErrAppAlreadyRunning`.  Ignore my comment 😄

internal/pkg/agent/application/filelock/file_locker.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+	return flocker, nil
+}
+
+func (fl *FileLocker) Lock() error {
+	return fl.LockContext(context.Background())
+}
+
+func (fl *FileLocker) LockContext(ctx context.Context) error {
```
ycombinator, line 55: Please add a godoc comment here explaining the default behavior w.r.t. blocking.

internal/pkg/agent/application/upgrade/marker_access_windows.go
```diff
@@ -52,6 +62,16 @@ func readMarkerFile(markerFile string) ([]byte, error) {
...
+		fileLock, err := lockMarkerFile(markerFile)
+		if err != nil {
+			return fmt.Errorf("locking update marker file %q for writing: %w", markerFile, err)
+		}
+		defer func(fileLock Locker) {
+			errUnlock := fileLock.Unlock()
+			if errUnlock != nil {
+				err = errors.Join(err, fmt.Errorf("unlocking marker file after writing: %w", errUnlock))
+			}
+		}(fileLock)
```
ycombinator: Maybe this could be moved into `writeMarkerFileCommon` so it's not repeated in `marker_access_windows.go` and `marker_access_other.go`?

internal/pkg/agent/application/filelock/file_locker.go
pchila: This is mostly for being able to reimplement AppLocker with the new FileLocker (we rely on a custom error when not being able to lock, so I thought I would make it generic)

internal/pkg/agent/application/filelock/app_locker.go
```diff
@@ -8,41 +8,28 @@ import (
...
 // NewAppLocker creates an AppLocker that locks the agent.lock file inside the provided directory.
 func NewAppLocker(dir, lockFileName string) *AppLocker {
 	if _, err := os.Stat(dir); os.IsNotExist(err) {
 		_ = os.Mkdir(dir, 0755)
 	}
-	return &AppLocker{
-		lock: flock.New(filepath.Join(dir, lockFileName)),
-	}
+
+	lock, _ := NewFileLocker(filepath.Join(dir, lockFileName), WithCustomNotLockedError(ErrAppAlreadyRunning))
```
michalpristas: we don't check error or nil, we can panic later

michalpristas: timeout should be used by default,  we dont want to hang forever

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -324,3 +337,25 @@ func saveMarkerToPath(marker *UpdateMarker, markerFile string, shouldFsync bool)
 func markerFilePath(dataDirPath string) string {
 	return filepath.Join(dataDirPath, markerFilename)
 }
+
+func fileLockName(markerFileName string) string {
+	return markerFileName + ".lock"
+}
+
+func lockMarkerFile(markerFileName string) (Locker, error) {
```
michalpristas: nit: i don't like it's performing an action - locking and constructing an entity(returning instance) at the same time.

internal/pkg/agent/application/filelock/app_locker.go
pchila: Creation of a FileLocker is never blocking (it doesn't try to lock at creation). Moreover, AppLocker was non-blocking in its previous version (using only `flock.TryLock()`) and the impl with FileLocker maintains that behavior.

As for the error handling, our options here are limited since we have no errors to return or logger to write out the error, I will add a panic() which should never happen (the only case where we can get an error at FileLocker creation is by specifiying a zero timeout, which AppLocker doesn't do).
Another option is to change the signature of NewAppLocker to return an error and handle the error in the caller's code

internal/pkg/agent/application/upgrade/step_mark.go
pchila: ~~An alternative is to provide a decorator accepting a `func() error` argument and locking and unlocking around that. The disadvantage with that option is the type constraints on the inner callable object (sadly we cannot declare varargs with Go generics).~~

~~The tradeoff of creating the object and returning it is to abstract the concrete type from the caller and ensure consistent instantiation as long as all the locks are acquired using this function.~~

Edit: tried to split creation and locking and the result is better than I expected, see f381841e4e8506b2c8cdfed4e2f38c9f5ad206b2
Thank you @michalpristas  :heart:

internal/pkg/agent/application/filelock/file_locker.go
pchila: see 33d8228e8ae1e255d2a9b3b35171b65f764bbda8

internal/pkg/agent/application/upgrade/marker_access_windows.go
pchila: It would still be asymmetric, one version for writing but still 2 separate ones for reading :thinking: 
Maybe a little refactor on marker_access_*.go files could be helpful but I think we can do it in a follow-up PR

internal/pkg/agent/application/filelock/app_locker.go
pchila: see 33b7791648fa7b809adc93f9e07ec92f34ed962d

ycombinator (dismissed):

swiatekm:
From what I can see, all the locking when reading or writing to the marker file is non-blocking. What will actually happen if the agent and watcher try to access the file concurrently? One will get an error, and then what?

pchila:
> From what I can see, all the locking when reading or writing to the marker file is non-blocking. What will actually happen if the agent and watcher try to access the file concurrently? One will get an error, and then what?

@swiatekm 
We don't use [`flock's Lock()`](https://pkg.go.dev/github.com/gofrs/flock#Flock.Lock) API as in the documentation we read `It's recommended that TryLock() be used over this function.`.

As an alternative we are using a [`TryLockContext()`](https://pkg.go.dev/github.com/gofrs/flock#Flock.TryLockContext) with a timeout

https://github.com/elastic/elastic-agent/pull/8254/files#diff-a21d895563344387fd97f5e4b5f5769a266362240b0b6ac420113eabfa1301a6R99

https://github.com/elastic/elastic-agent/pull/8254/files#diff-a21d895563344387fd97f5e4b5f5769a266362240b0b6ac420113eabfa1301a6R59-R62

so the FileLocker will keep trying to acquire a lock periodically for the timeout duration and if it fails it will error out.
Right now such an error is fatal when creating the update marker (but under normal condition there should be no contention at that point).

Reading and writing the update marker right now consists of a single writer (main process or watcher) in every phase.
As https://github.com/elastic/elastic-agent/issues/6890 and https://github.com/elastic/elastic-agent/issues/6889 get implemented we are gonna have multiple concurrent writers on update marker where the lock needs to be in place to avoid overwriting information. If a writer gets a locking error it will need to reload the update marker data and retry the (partial) modification

swiatekm (dismissed):

changelog/fragments/1749470526-lock-update-marker.yaml
```diff
@@ -0,0 +1,32 @@
+# Kind can be one of:
```
ycombinator, line 1: Do we need a changelog entry for this change?  It doesn't seem like a user-facing change, more an internal detail?

ycombinator (approved):
Left a question about whether we need the CHANGELOG entry for this change but the implementation itself + tests LGTM!

changelog/fragments/1749470526-lock-update-marker.yaml
pchila: As it introduces a locking mechanism for the update marker I prefer to have it on the release notes, it could give a clue in case some deadlock or some other issue crop up.  
If there was a flag to disable this and I could activate it only when the feature is complete I probably wouldn't add a changelog entry.

internal/pkg/agent/application/upgrade/step_mark.go
michalpristas: it usually is and it optimizes for readability. thank you so much for adding it

michalpristas (approved):
i'm not fun of panics but the alternative is not good as well. this will send us to startup spiral loop but at least we know something's wrong right away - should not happen in very most of the cases.
LGTM otherwise.

PR #8273, opened 2025-05-29
wip: Translate beat receiver status to component state

swiatekm:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -13,7 +13,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/elastic/elastic-agent/internal/pkg/otel/configtranslate"
+	"github.com/elastic/elastic-agent/internal/pkg/otel/translate"
 
 	"go.opentelemetry.io/collector/component/componentstatus"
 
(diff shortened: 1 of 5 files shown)


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -656,77 +658,146 @@ func (c *Coordinator) watchRuntimeComponents(ctx context.Context) {
...
 			select {
-			case c.managerChans.runtimeManagerUpdate <- s:
+			case c.managerChans.runtimeManagerUpdate <- componentState:
 			case <-ctx.Done():
 				return
 			}
+		case otelStatus := <-otelChan:
+			componentStates, err := translate.GetAllComponentStates(otelStatus, c.componentModel)
+			if err != nil {
+				c.setOTelError(err)
```
blakerouse, line 689: This feels like a bug reading the code. I read the code to understand below that `componentStates` can be nil and it will not cause an issue. I think it would be best to add a comment to explain this. That makes it clearer when reading the code that continuing here is okay.

blakerouse (approved):
This change looks good to me. See my one inline comment about adding a code comment.

pkoutsovasilis (dismissed):
LGTM, I do hear the comment of Blake above 🙂

internal/pkg/agent/application/coordinator/coordinator.go
swiatekm: It does sound like a comment is needed. The additional subtlety here is that even if we get an error from extracting component states (which right now can only happen due to a bug), we still want to pass on the full otel status - if only because it will make debugging much easier.

PR #8278, opened 2025-05-30
[ci] fix update versions

pkoutsovasilis:
## What does this PR do?

This PR updates the upgradable version selection logic in `FetchUpgradableVersions` to explicitly exclude previous patch versions when the major and minor versions match the `UpgradeToVersion` version.

Specifically:
- It introduces a check to skip any versions with the same major and minor as the upgrade target if their patch is less than or equal to the `UpgradeToVersion` patch part.
- It rewrites `TestFetchUpgradableVersionsAfterFeatureFreeze` to use table-driven tests with self-contained version data.

## Why is it important?

- Prevents redundant patch upgrades from being included as upgrade candidates when not applicable.
- Fixes test flakiness issue in `PreviousMinor()` where versions like `9.0.1` failed to resolve a previous minor due to ambiguous or stateful data dependencies.

## Checklist

## Disruptive User Impact

None. This change only affects internal version filtering logic used in upgrade tests. There is no impact on users or runtime behaviour of Elastic Agent.

```
mage unitTest
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/6667

### DIFF
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -199,6 +199,10 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 		case !version.Less(*parsedUpgradeToVersion):
 			continue
 
+		// same major, same minor, less or equal patch
+		case version.Major() == parsedUpgradeToVersion.Major() && version.Minor() == parsedUpgradeToVersion.Minor() && version.Patch() <= parsedUpgradeToVersion.Patch():
+			continue
+
 		case recentSnapshotsToFind > 0 && version.IsSnapshot():
 			upgradableVersions = append(upgradableVersions, version.String())
 			recentSnapshotsToFind--
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -199,8 +199,8 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 		case !version.Less(*parsedUpgradeToVersion):
 			continue
 
-		// same major, same minor, less or equal patch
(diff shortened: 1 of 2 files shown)


testing/upgradetest/versions.go
```diff
@@ -199,6 +199,10 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 		case !version.Less(*parsedUpgradeToVersion):
 			continue
 
+		// same major, same minor, less or equal patch
+		case version.Major() == parsedUpgradeToVersion.Major() && version.Minor() == parsedUpgradeToVersion.Minor() && version.Patch() <= parsedUpgradeToVersion.Patch():
+			continue
+
```
swiatekm: Is the `version.Patch() <= parsedUpgradeToVersion.Patch()` check necessary? We already do `case !version.Less(*parsedUpgradeToVersion):` above. If the source version is less than the targeted version, and the major and minor are the same, then patch has to be less.

swiatekm (commented):
LGTM, thank you for fixing this! I had a question about the added condition. I want to make sure the code is as clear as possible, as it has been a source of subtle bugs in the past.

testing/upgradetest/versions.go
pkoutsovasilis: AFAICT, `case !version.Less(*parsedUpgradeToVersion):` will make us discard versions higher, either in minor, major, and patch, than the `parsedUpgradeToVersion`. The newly introduced `case version.Major() == parsedUpgradeToVersion.Major() && version.Minor() == parsedUpgradeToVersion.Minor() && version.Patch() <= parsedUpgradeToVersion.Patch():` will make us discard only previous patch versions than `parsedUpgradeToVersion`. What do I miss?

swiatekm: FYI, I meant only removing the patch comparison while keeping the major and minor ones here. Sorry if that was unclear.

Issue #6132, opened 2024-11-23
Update all OTel upstream dependencies regularly and atomically


ycombinator:
**Describe the enhancement:**
Automation (likely a custom script) that will regularly update upstream OTel dependencies 

**Describe a specific use case for the enhancement or feature:**
We currently use Dependabot for managing updates of upstream OTel dependencies.  Unfortunately, we've observed at times that not all such dependencies get updated in the same PR.  A custom script would allows us to solve this problem.

**What is the definition of done?**
* [ ] A script that will run daily to check if a new version of upstream OTel dependencies has been released.  If it has, the script will put up a PR containing the updates.  
* [ ] All upstream OTel dependencies must be updated in a single PR.
* [ ] The PR should also include an update to the [EDOT Collector README](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/README.md) with new versions of the upstream OTel dependencies.
* [x] Remove the [Dependabot configuration](https://github.com/elastic/elastic-agent/blob/main/.github/dependabot.yml) that currently performs the (incomplete) updates.

cmacknz:
We are also able to use Renovate if needed and it would handle this better.

ebeahan:
> A script that will run daily to check if a new version of upstream OTel dependencies has been released. If it has, the script will put up a PR containing the updates.

@andrzej-stencel does the script you added in https://github.com/elastic/elastic-agent/pull/7159 satisfy the first item from the description? What else do we need to track to complete this one?

andrzej-stencel:
Created PR to automate the OTel bumps: https://github.com/elastic/elastic-agent/pull/8288.

PR #8288, opened 2025-06-02
ci: automatically update OTel components

andrzej-stencel:
## What does this PR do?

Creates a new Updatecli configuration to check if a new version of OTel Collector components is available. If it is, Updatecli runs the `update-otel.sh` script and creates a new PR with the changes.

Also creates a new GitHub Actions workflow to trigger the Updatecli configuration daily. The workflow is configured torun for the `main` branch and for the `8.19` branch.

## Why is it important?

We want to be on the latest version of OTel Collector components on `main` and `8.19` branches. We already had a script to update OTel dependencies that took as parameters the versions to update to. What this PR adds is checking for latest OTel versions online and triggering the script if an update is required.

## Checklist

```bash
BRANCH_NAME=main updatecli diff --config ./.ci/updatecli/update-otel.yml
```

This will succeed locally if you remove the references to SCM from the file `.ci/updatecli/update-otel.yml`: `scms`, `conditions.is-new-version.scmid`, `actions`.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/6132


v1v:
To test locally against the `main` branch, you can use the below snippet (as long as you have `gh` locally):


```bash
$ gh pr checkout 8288
  GITHUB_ACTOR=v1v \
  BRANCH_NAME=main updatecli diff \
    --config ./.ci/updatecli/update-otel.yml 


+++++++++++
+ PREPARE +
+++++++++++
[..]


condition: condition#is-new-version
------------------------
The shell 🐚 command "/bin/sh /var/folders/t7/ghqdh8cx2g12pwb_w0ncmw900000gn/T/updatecli/bin/b8e0f34aa6b0f3e4759c4ad975a6de8175acea171186f323dc1c3c345e68a176.sh" exited on error (exit code 1) with the following output:
----
----

command stderr output was:
----
----
shell command failed. Expected exit code 0 but got 1
✗ shell condition of type "console/output" not passing

target: target#update-otel
------------------


ACTIONS
========

No target found for action "default"

=============================

SUMMARY:



- Update OTel:
	Source:
		✔ [current_contrib] Get current OTel Collector contrib version in go.mod
		✔ [current_core_beta] Get current OTel Collector core beta version in go.mod
		✔ [latest_contrib] Get latest OTel Collector contrib version
		✔ [latest_core_beta] Get latest OTel Collector core beta version
		✔ [latest_core_stable] Get latest OTel Collector core stable version
	Condition:
		✗ [is-new-version] Succeeds if the latest core or contrib OTel Collector version is different than the one in go.mod
	Target:
		- [update-otel] 


Run Summary
===========
Pipeline(s) run:
  * Changed:	0
  * Failed:	0
  * Skipped:	1
  * Succeeded:	0
  * Total:	1
```

.ci/updatecli/update-otel.yml
```diff
@@ -0,0 +1,75 @@
...
+    spec:
+      command: '[ {{ source "current_core_beta" }} != {{ source "latest_core_beta" }} ] || [ {{ source "current_contrib" }} != {{ source "latest_contrib" }} ]'
+
+targets:
+  update-otel:
+    name: Update OTel version
+    kind: shell
+    disablesourceinput: true
+    spec:
+      command: .ci/scripts/update-otel.sh {{ source "latest_core_beta" }} {{ source "latest_core_stable" }} {{ source "latest_contrib" }}
```
v1v, line 59: Otherwise, if no PATH and using ` eval "$(gvm use 1.24.2)"` or similar dynamic tools to configure golang then

```
The shell 🐚 command "/bin/sh /var/folders/t7/ghqdh8cx2g12pwb_w0ncmw900000gn/T/updatecli/bin/171383672ec9263a96ebb1f512e111207e3f71812c7fc93b5a15196b2a0c4104.sh" exited on error (exit code 127) with the following output:
----
=> Updating core from v0.127.0/v1.33.0 to v0.127.0/v1.33.0
=> Updating contrib from v0.127.0 to v0.127.0
=> Running go mod tidy
----

command stderr output was:
----
.ci/scripts/update-otel.sh: line 39: go: command not found


```


```suggestion
      command: .ci/scripts/update-otel.sh {{ source "latest_core_beta" }} {{ source "latest_core_stable" }} {{ source "latest_contrib" }}
      environments:
      - name: PATH
      - name: HOME
```

v1v: OTOH, I see some errors when running it locally:

```
.ci/scripts/update-otel.sh: line 47: elastic-agent-changelog-tool: command not found

----

source: source#current_core_beta
------------------------
✔ version v0.127.0 found for GO module "go.opentelemetry.io/collector/receiver/otlpreceiver"

ERROR: something went wrong in "target#update-otel" : something went wrong in target "update-otel" : "Shell command exited on error."

Pipeline "Update OTel" failed
Skipping due to:
	something went wrong during target execution

```

that I actually fixed by running:

```bash
$ eval "$(gvm use 1.24.2)"
$ go install github.com/elastic/elastic-agent-changelog-tool@latest
$ export PATH=$PATH:~/go/bin
$ updatecli...
```


```
=> Creating changelog fragment
----
⚠ - ran shell command ".ci/scripts/update-otel.sh v0.127.0 v1.33.0 v0.127.0"


ACTIONS
========

No target found for action "default"

=============================

SUMMARY:



⚠ Update OTel:
	Source:
		✔ [current_contrib] Get current OTel Collector contrib version in go.mod
		✔ [current_core_beta] Get current OTel Collector core beta version in go.mod
		✔ [latest_contrib] Get latest OTel Collector contrib version
		✔ [latest_core_beta] Get latest OTel Collector core beta version
		✔ [latest_core_stable] Get latest OTel Collector core stable version
	Target:
		⚠ [update-otel] Update OTel version


Run Summary
===========
Pipeline(s) run:
  * Changed:	1
  * Failed:	0
  * Skipped:	0
  * Succeeded:	0
  * Total:	1
```

Can you confirm whether `elastic-agent-changelog-tool` is expected to be in the path? If so, who is the one actually installing it? I cannot see any references in the .github/workflows/bump-otel-version.yml script

andrzej-stencel: Thanks @v1v , I added the env vars for the Updatecli target https://github.com/elastic/elastic-agent/pull/8288/commits/bc9e62cac22925f9a9fc9f5b4faa90e7f4b67f32 and added code to install changelog tool if missing https://github.com/elastic/elastic-agent/pull/8288/commits/8dfeccde2d5c4e14ada145a1c2705a4770172af2. Let me know if that works for you.

v1v (approved):
Thanks

.github/workflows/bump-otel-version.yml
```diff
@@ -0,0 +1,63 @@
...
+    outputs:
+      matrix: ${{ steps.generator.outputs.matrix }}
+    permissions:
+      contents: read
+    steps:
+    - id: generator
+      uses: elastic/oblt-actions/elastic/active-branches@v1
+      with:
+        # Only run on `main` and `8.19` branches
+        exclude-branches: "7.17,8.17,8.18,9.0"
```
ycombinator: Is the idea here to backport the PR from `main` to active `9.*` branches?  If so, then maybe we should add the `backport-active-9` label to https://github.com/elastic/elastic-agent/pull/8288/files#diff-c48073d6283a20fd780d47db88cc59497ee4f449f67f86b20e4ef86fad4f7dcaR71-R74?

v1v: IIUC, the labels are less efficient when those branches are not in sync. Hence, I think this implementation creates specific PRs per active release branch (excluding the above list), which will avoid situations with merge conflicts.

`backport-active-*` uses the `git cherry-pick` approach and causes conflicts, while the approach in this PR uses `updatecli` targeting specific branches.

swiatekm: I don't think we want to update otel components in patch releases unless we have a good reason for it. They can often contain breaking changes or enhancements which really shouldn't go into a patch release.

Ultimately, I expect this to only happen on `main`, albeit `8.19` might get a few updates given that it's the last minor in the 8.x series.

andrzej-stencel: Agree with what Victor [said](https://github.com/elastic/elastic-agent/pull/8288#discussion_r2127290847), backporting is prone to conflicts, so we're taking a different approach here with creating separate PRs for different branches.

Agree with Mikołaj, the idea is to only automatically update OTel components on "active" branches (which is currently `main` and `8.19`) and not on "maintenance" branches like `8.18` or `9.0`.

As for updating OTel in maintenance branches, I don't think it's worth automating this, because:
1. On the maintenance branches, we don't want to update minor versions of OTel Collector, as the OTel Collector being not stable (v0.x.y), minor changes can contain breaking changes and generally carry a high risk of introducing bugs. We might want to update maintenance branches to patch releases of OTel Collector, but patch releases are [rare by design](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.127.0/docs/release.md#bugfix-release-criteria).
1. Given the two-week release cadence of upstream OTel Collector, the maintenance branches are quickly on a non-latest version of OTel Collector, and patch releases of non-latest versions don't happen at all.

ycombinator: Thanks folks, everything said in this thread makes sense.  

> the idea is to only automatically update OTel components on "active" branches (which is currently main and 8.19) and not on "maintenance" branches like 8.18 or 9.0.

My only concern now is that this configuration becomes one more place to keep updated whenever we cut a new branch, as the definition of active branches will change at that point.  I believe this will first become an issue when `main` starts to point to `9.4.0` and we have `9.3` and `9.2` branches for the previous two minors.  At that point, we'll want to come back here and exclude the `9.1` branch, correct?

swiatekm: Yeah, maybe we just want to hardcode it to `main` and `8.19`, and eventually remove the latter?

andrzej-stencel: Yes, I'm not a fan of this approach used in this PR: "get all the active branches, and exclude ... everything apart from two". I think it would be better to explicitly state the branches that we want this workflow to be run for (currently `main` and `8.19`). I used this solution for consistency with approach used in https://github.com/elastic/elastic-agent/pull/8174/files#diff-e2c8c610c077a247bb8aa660cbffc7a1e2bc16e5dc4e61069087cfe03e3293faR23.

I'll happily change using the `elastic/oblt-actions/elastic/active-branches` action to specifying a static matrix with `main` and `8.19`. Would that work?

andrzej-stencel: Changed in https://github.com/elastic/elastic-agent/pull/8288/commits/ed86c71ee1b818686ed7fc962b54428b02da5521, please take a look @v1v @swiatekm @ycombinator.

.github/workflows/bump-otel-version.yml
```diff
@@ -0,0 +1,47 @@
...
+      pull-requests: write
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: ["main", "8.19"]
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ matrix.branch }}
```
v1v: This will enforce to have `.ci/updatecli/update-otel.yml` up to date, is that the expected behaviour? Or do you prefer maintaining the updatecli pipeline in `main` only? If so, then `ref...` is not needed

v1v (approved):
LGTM, just a question about where the updatecli pipeline needs to be defined

.github/workflows/bump-otel-version.yml
andrzej-stencel: Thanks, I think we should only use the Updatecli config from the `main` branch. Changed in https://github.com/elastic/elastic-agent/pull/8288/commits/4853c2d6c48758d5d9861fc16973e9b7e9346e57.

v1v (approved):

PR #6243, opened 2024-12-09
[CI] Added test report aggregation for test-collector plugin

pazone:
## What does this PR do?
Introduces an Aggregator step that downloads all JUnit reports and sends them Buildkite test analytics.

All test reports now have a fully-qualified name with OS information and sudo/non-sudo flag. 

[Build example](https://buildkite.com/elastic/elastic-agent-extended-testing-bk/builds/294#0193abd8-36ff-4d88-bac2-07c268d5d2a1) for Windows, RPM and Ubuntu with the aggregator step usage. 

[Buildkite test analytics report](https://buildkite.com/organizations/elastic/analytics/suites/elastic-agent-ci/runs/ed88cd5c-ae39-81d4-94d2-3d9b4a5abd96)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/4516

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #6148, opened 2024-11-26
Run BK integration tests on PRs and release branches

pazone:
## What does this PR do?
Runs the new integration testing infrastructure provisioned by buildkite and terraform (ESS)

Integration tests take 1 h 10 minutes instead of ~1:30. 

Independent test groups execution allows to retry tests independently with a dedicated ESS stack:
Screenshot 2024-11-27 at 16 43 16

[Build example 1](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4907#01936da4-9be4-446b-a01c-d1721b3c21c6)
[Build example 2](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4908)
[Build example 3](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4920)
[Build example 4](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/4939)

#### What groups do we run
- [x] default
- [x] container
- [x] fleet-upgrade-to-pr-build
- [x] upgrade
- [x] fleet
- [x] fqdn
- [x] deb
- [x] fleet-airgapped
- [x] fleet-privileged
- [x] fleet-airgapped-privileged
- [x] rpm

#### What OS do we run
- [x] Windows 2022
- [x] Ubuntu 24.04 x86_64
- [x] Ubuntu 24.04 ARM (temporarily disabled due to a [known issue](https://github.com/elastic/ingest-dev/issues/4610))
- [x] RHEL 8 (rpm)

By comparing test reports, I verified that we run the same test sets for groups. Placed the tests in groups in this [spreadsheet](https://docs.google.com/spreadsheets/d/1kuExKnVyjiwpGSl_YUn_enVVwg-XuMnu2wWMMud-sn4/edit?gid=0#gid=0)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #4739, opened 2024-05-13
[CI] Separated extended testing pipeline

pazone:
## What does this PR do?
Depends on https://github.com/elastic/elastic-agent/pull/4738
Extracting long running integration tests to a separate pipeline. 
Added changeset filters to the main the new pipeline
Added notification context for integration tests

Screenshot 2024-05-15 at 17 36 31

**Important:** Introduces a concurrency group for each type of exteneded tests. The concurrency groups are limited to `8`. That means we can run only 8 jobs of a type simultaneously 

For example: 
```
steps:
  - label: "Serverless integration test"
    key: "serverless-integration-tests"
    concurrency_group: elastic-agent-extended-testing/serverless-integration # <- concurrency group
    concurrency: 8 # <- concurrency limit
    env:
      BEAT_NAME: "elastic-agent"
    command: ...
```

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/3674
- Relates https://github.com/elastic/elastic-agent/issues/4710

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #8311, opened 2025-06-03
bk: use in the pre-command when needed only

v1v:
## What does this PR do?

* Use https://github.com/elastic/vault-secrets-buildkite-plugin to gather the BK Test Token
* Bump version https://github.com/buildkite-plugins/test-collector-buildkite-plugin/releases/tag/v1.11.0

## Why is it important?

Avoid exposing secrets in the logs by using only the token when it's needed using the BK plugins approach.

IMPORTANT: I'm not sure if the docs are up-to-date as I cannot see the BK Test Collector being enabled in the ITs.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -58,15 +58,9 @@ if [[ "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
   export API_KEY_TOKEN=$(vault kv get -field apiKey ${CI_ESS_PATH})
   echo ${API_KEY_TOKEN} > ./apiKey
   export TEST_INTEG_AUTH_ESS_APIKEY_FILE=$(realpath ./apiKey)
-
-  # BK analytics
-  echo "--- Prepare BK test analytics token :vault:"
-  BUILDKITE_ANALYTICS_TOKEN=$(vault kv get -field token kv/ci-shared/platform-ingest/buildkite_analytics_token)
-  export BUILDKITE_ANALYTICS_TOKEN
 fi
 
-if [[ "$BUILDKITE_STEP_KEY" == *"aggregate-reports"* ]]; then
-  # BK analytics
+if [[ "$BUILDKITE_STEP_KEY" == *"aggregate-reports"* || "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
   echo "--- Prepare BK test analytics token :vault:"
   BUILDKITE_ANALYTICS_TOKEN=$(vault kv get -field token kv/ci-shared/platform-ingest/buildkite_analytics_token)
   export BUILDKITE_ANALYTICS_TOKEN
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -60,12 +60,6 @@ if [[ "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
   export TEST_INTEG_AUTH_ESS_APIKEY_FILE=$(realpath ./apiKey)
 fi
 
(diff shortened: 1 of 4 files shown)


.buildkite/hooks/pre-command
```diff
@@ -58,18 +58,6 @@ if [[ "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
...
-  echo "--- Prepare BK test analytics token :vault:"
-  BUILDKITE_ANALYTICS_TOKEN=$(vault kv get -field token kv/ci-shared/platform-ingest/buildkite_analytics_token)
-  export BUILDKITE_ANALYTICS_TOKEN
-fi
-
-if [[ "$BUILDKITE_STEP_KEY" == *"aggregate-reports"* ]]; then
-  # BK analytics
-  echo "--- Prepare BK test analytics token :vault:"
-  BUILDKITE_ANALYTICS_TOKEN=$(vault kv get -field token kv/ci-shared/platform-ingest/buildkite_analytics_token)
-  export BUILDKITE_ANALYTICS_TOKEN
```
v1v, line 72: I don't think we need this approach, it's hard to know when it's needed, but using the explicit BK plugin can help us to know when that's needed, as they need each other (`elastic/vault-secrets#v0.1.0` and `test-collector#v1.11.0`

docs/test-framework-dev-guide.md
```diff
@@ -356,7 +356,11 @@ out weight the benefits of creating another group.
           machineType: "n1-standard-8"
           image: "family/platform-ingest-elastic-agent-ubuntu-2404"
         plugins:
```
v1v, line 358: I think these docs are not reflecting what it's actually done in the CI, the `test-collector` is not used in the IT groups, no clue why. @pazone , do you recall what's the reason `test-collector` is not used in `label: "x86_64:sudo: {{matrix}}"` or similar ITs?

pazone: AFAIK, It should be used in all integration test steps.

v1v: https://github.com/elastic/elastic-agent/pull/6148 is the one that added the documentation, however, I don't see those changes reflected in the ITs.

Unfortunately, I do not have enough context about the IT's test reporting (that was something started long time ago https://github.com/elastic/elastic-agent/pull/5113 and https://github.com/elastic/elastic-agent/pull/4739)/

I've found https://github.com/elastic/elastic-agent/pull/6243 and that seems to actually centralise 'all JUnit reports and sends them Buildkite test analytics.'

As far as I see @pazone , you raised those PRs, your insights are valuable, either there is a missing configuration and we are not reporting all the Test results or the documentation is not up-to-date.

v1v:
> Let's make sure that we use test collector in all integration test steps that produce junit.xml

`BUILDKITE_ANALYTICS_TOKEN` is only honoured by the `test-collector` plugin and only the step called `Aggregate test reports` uses it.

Can you elaborate on what you meant here?

docs/test-framework-dev-guide.md
pazone: Sorry for the confusion. Instead of adding this plugin to each test, we have the `Aggregate test reports` step. It downloads all the `*junit*.xml` files and sends them to BK test analytics. I assume the `Aggregate test reports` was introduced without updating the docs.
So it's better to remove the `test-collector` from the example

v1v: Thanks Pavel, in a follow-up, I guess `docs` don't trigger the whole build, but since it has been already approved, I prefer to avoid going through a review process for this PR

PR #8337, opened 2025-06-04
Add custom ECH deployment + smoke test

michel-laterman:
## What does this PR do?

Use the image produced in the packaging step as the integrations server image for an ECH deployment and run a basic smoke test against that deployment. This allows for the changes made in a PR to be tested as a part of the cloud image in addition to the current test coverage.

## Why is it important?

Removes the 1 day delay from producing a SNAPSHOT image to testing the image.

## Checklist

## Disruptive User Impact

N/A


michel-laterman:
This PR copies the approach taken by https://github.com/elastic/elastic-agent/pull/8035, but applies it to the normal integration test pipeline

magefile.go
```diff
@@ -984,6 +984,37 @@ func (Cloud) Image(ctx context.Context) {
...
+
+	// Need to get the FIPS env var flag to see if we are using the normal source cloud image name, or the FIPS variant
+	fips := os.Getenv(fipsEnv)
+	defer os.Setenv(fipsEnv, fips)
+	fipsVal, err := strconv.ParseBool(fips)
+	if err != nil {
+		fipsVal = false
+	}
+	os.Setenv(fipsEnv, strconv.FormatBool(fipsVal))
+	devtools.FIPSBuild = fipsVal
```
swiatekm: Why do we need to modify the environment here? It doesn't look like anything we're changing actually affects loading the image.

v1v (approved):
LGTM. Thanks Michel.

It’s concerning how much duplicated configuration is needed here. We plan to revisit this in the future; however, for now, this is our reality.

magefile.go
michel-laterman: I originally wrote this for the FIPS variation in https://github.com/elastic/elastic-agent/pull/8035 and copied the entire function

magefile.go
```diff
@@ -984,6 +984,35 @@ func (Cloud) Image(ctx context.Context) {
 	Package(ctx)
 }
 
+// Load loads an artifact as a docker image.
+// Looks in build/distributions for an elastic-agent-cloud*.docker.tar.gz artifact and imports it as docker.elastic.co/beats-ci/elastic-agent-cloud:$VERSION
+// DOCKER_IMPORT_SOURCE - override source for import
+func (Cloud) Load() error {
+	snapshot := os.Getenv(snapshotEnv)
+	defer os.Setenv(snapshotEnv, snapshot)
+	os.Setenv(snapshotEnv, "true")
```
swiatekm: This also looks completely unnecessary to me.

swiatekm (approved):
Looks good enough to me, though I think some of the env variable manipulation in cloud:load is unnecessary.

magefile.go
michel-laterman: We want to force SNAPSHOT to be true for this step and restore it afterwards

swiatekm: From what I can see, the only thing we do in this step is run `docker image load ...`, which doesn't care about the snapshot variable as far as I know. Am I wrong about that?

michel-laterman: It's used as part of the version resolution

magefile.go
```diff
@@ -984,6 +984,35 @@ func (Cloud) Image(ctx context.Context) {
 	Package(ctx)
 }
 
+// Load loads an artifact as a docker image.
+// Looks in build/distributions for an elastic-agent-cloud*.docker.tar.gz artifact and imports it as docker.elastic.co/beats-ci/elastic-agent-cloud:$VERSION
+// DOCKER_IMPORT_SOURCE - override source for import
+func (Cloud) Load() error {
```
pchila: Why do we need this implemented as a mage target?
If I understand the flow correctly:
1. image is built and saved in the packaging step
2. image is downloaded in buildkite `Build and push custom elastic-agent image` (spoiler alert: we are not building anything here!)
3.  `mage cloud:load` and `mage cloud:push` are called but they simply perform a `docker load` and a `docker push` from shell
4. control go back to buildkite that passes the pushed image tag (which is present as an env var) to terraform to create the ECH stack.

Why is this back and forth needed between buildkite and mage? Wouldn't this be simpler if the docker load and docker push was called directly from buildkite ?

michel-laterman: I added `load` to be consistent with our other mage targets
Why do we have `cloud:push` as a mage target?

As for simplicity, having `load` and `push` as mage targets allows it to set defaults that work for cloud deployments, `mage cloud:push` by default is easier to use then manually retagging and pushing

.buildkite/hooks/pre-command
```diff
@@ -53,3 +53,8 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent-binary-dra" ]]; then
     release_manager_login
   fi
 fi
+
+# BUILDKITE_PIPELINE_SLUG should match elastic-agent for PRs, and elastic-agent-extended-tests once it has merged to main
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent"* && ("$BUILDKITE_STEP_KEY" == "integration-cloud-image" || "$BUILDKITE_STEP_KEY" == "integration-fips-cloud-image") ]]; then
+    docker_login
+fi
```
michel-laterman: @v1v, Do I still need this for the PR?

michel-laterman:
buildkite test this

.buildkite/hooks/pre-command
v1v: You can now enhance your Buildkite pipeline by explicitly configuring the `integration-cloud-image` step to utilise the plugin for preparing the docker environment. This adjustment will streamline your setup and reduce the black box magic with the `pre-command` hook.

 https://github.com/elastic/elastic-agent/pull/8444 introduced that change. Let me know if you need further assistance

magefile.go
```diff
@@ -997,12 +997,10 @@ func (Cloud) Load() error {
 
 	// Need to get the FIPS env var flag to see if we are using the normal source cloud image name, or the FIPS variant
 	fips := os.Getenv(fipsEnv)
-	defer os.Setenv(fipsEnv, fips)
 	fipsVal, err := strconv.ParseBool(fips)
 	if err != nil {
 		fipsVal = false
 	}
-	os.Setenv(fipsEnv, strconv.FormatBool(fipsVal))
```
simitt: Why is this removed when it was necessary before (I do not see this logic being moved to a different place in this PR)?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -383,6 +461,9 @@ steps:
           ASDF_KIND_VERSION: "0.27.0"
           DOCKER_VARIANTS: "{{matrix.variants}}"
           TARGET_ARCH: "amd64"
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud"
+          TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud:fips-${BUILDKITE_COMMIT:0:12}"
```
simitt: Why is this specific to FIPS where everything else just seems to change the logic to allow pushing a locally built image; generic and unrelated to fips.

magefile.go
michel-laterman: It's not needed

.buildkite/bk.integration.pipeline.yml
michel-laterman: good catch

.buildkite/bk.integration.pipeline.yml
```diff
@@ -38,6 +63,9 @@ steps:
         depends_on:
           - packaging-windows
         env:
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
```
v1v: I might have missed the rationale for using those environment variables per step instead of globally. Could you please explain the reason for avoiding the [top-level env section](https://buildkite.com/docs/pipelines/configure/environment-variables#environment-variable-precedence)?

michel-laterman: When I used top-level env sections when developing this for the FIPS tests I observed that FIPS images were being deployed in the non-FIPS pipeline tests

michel-laterman:
buildkite test this

.buildkite/bk.integration.pipeline.yml
swiatekm: I don't understand why that would be the case. I agree with @v1v that it looks like we're adding the same values to every step. If there's a step which needs fips-specific values, we should set those in it instead.

michel-laterman: I can try with top-level env vars and run it a couple times to see if it was just a bk issue that as resolved itself. However if that's not the case we'll need to have repetition until the issue is fixed.

michel-laterman:
@swiatekm It looks like this is still an issue with BK; The top level environment vars in `bk.integration-fips.yml` are applying to `bk.integration.yml`.
See https://buildkite.com/elastic/elastic-agent/builds/22825#01978ede-7a93-4204-be10-eb80a12ccd5e

Screenshot 2025-06-20 at 2 04 48 PM

I'll revert the change so top level env vars are not used.

cc @v1v

.buildkite/bk.integration.pipeline.yml
michel-laterman: Attempt to move to top-level has failed, Top-level env vars from `bk.integration-fips.pipeline.yml` were used when executing steps in this pipeline

.buildkite/bk.integration.pipeline.yml
```diff
@@ -39,13 +39,38 @@ common:
...
+      buildkite-agent artifact download build/distributions/elastic-agent-cloud-*-linux-amd64.docker.tar.gz . --step 'packaging-containers-x86-64'
+      mage cloud:load
+      mage cloud:push
+    agents:
+      provider: "gcp"
+      machineType: "n1-standard-8"
+      image: "${IMAGE_UBUNTU_2404_X86_64}"
+    plugins:
+      - elastic/vault-docker-login#v0.5.2:
+          secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
```
pkoutsovasilis: this step is redundant to me, there is no need to spawn a new runner, fetch a rather hefty artifact just to push it somewhere. I would propose to utilise the existing packaging step, that has every requirement you need and push it there

.buildkite/bk.integration.pipeline.yml
```diff
@@ -39,13 +39,38 @@ common:
...
+    depends_on:
+      - integration-cloud-image
     notify:
       - github_commit_status:
           context: "buildkite/elastic-agent-extended-testing - ESS stack provision"
     env:
       ASDF_TERRAFORM_VERSION: 1.9.2
+      CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+      CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud"
+      TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud:git-${BUILDKITE_COMMIT:0:12}"
```
pkoutsovasilis: I am a bit skeptical in making the main ESS for the client-centric tests, running an integration server of this PR. I think that this defeats the retry logic with a dedicated ESS we have, because of the following; let's assume that we have an issue in bootstraping fleet-server, etc., etc. and the test fails what's the point in wasting resources by provisioning another ESS, running again all client-tests just to fail because the integration server logic has the issue. Also it will make it harder to track down the reason of the failure as now we have two potential sources at the same time. So instead of introducing this everywhere, I would strongly advise to take advantage of this separation, client-oriented tests run with a stable ESS so that we are sure that every failure originates from a client code (we have stable ESS stack pinning nowadays) and write a dedicated test that evaluates only the integration server which has its own value for sure. wdyt?

.buildkite/bk.integration.pipeline.yml
pkoutsovasilis: I believe that this is happening because instead of triggering the `bk.integration-fips.pipeline.yml` we upload it (for context that is required to report the flaky CI steps) and apparently BK is doing some merging to the existing pipeline, so not that straight forward. And now that we speak about `bk.integration-fips.pipeline.yml` any particular reason that the FIPS-oriented CI steps are not part of a new FIPS named group? 🤔

.buildkite/bk.integration.pipeline.yml
michel-laterman: You mean move this to the packaging step?

.buildkite/bk.integration.pipeline.yml
michel-laterman: I'm not sure what you are proposing here; How would we catch cloud bootstrapping issues without running our own integrations server? Or do you mean that we should only run a custom integration server on certain tests?

If there is an issue with bootstrapping, it would be related to the PR and should fail instead of pass and become a snapshot that fails tomorrow

pkoutsovasilis: Michel if you have two sources of possible failure, would troubleshooting a test failure become more challenging? So yes I am proposing to have another ESS stack that runs the custom integration server from the respective PR and run only certain tests against it, wdyt?

.buildkite/bk.integration.pipeline.yml
pkoutsovasilis: yes, wdyt?

.buildkite/bk.integration.pipeline.yml
michel-laterman: What's a named group, and how do I use it? I'm not familiar enough with bk to know all the mechanisms

.buildkite/bk.integration.pipeline.yml
michel-laterman: I'll make the change

.buildkite/bk.integration.pipeline.yml
michel-laterman: What groups would we run against the custom ECH deployment?

pkoutsovasilis: IMO, none of the existing groups Michel; this should be a specific test that tests at least the following:
* enroll an agent with a package policy
* check that agent reports healthy
* un-enroll agent and check that agent is gone from fleet

anything more than that feels more like fleet-server testing than testing if the bootstrapping part has any issues, but more than happy to include more if you see value in them. Also, I do believe that we should also discuss this in our team meeting, so please bring it up there 🙂

.buildkite/bk.integration.pipeline.yml
pkoutsovasilis: maybe what I have in mind isn't feasible but ideally IMO FIPS would appear in a single square in the CI where every FIPS-related step would be part of it. I don't see the reason being expanded but this is more aesthetic and definitely non-blocking this PR

.buildkite/bk.integration.pipeline.yml
pchila: I agree with @pkoutsovasilis about having a separate group for testing fleet-server bootstrap in cloud deployments (`fleet-bootstrap` would also be good name for it) where the ability to spin up an integration server with the image build from the PR is tested.

This would mean that we have to test just a few things from (client) agent side like being able to create policy, enroll, see that all of this is correctly handled and reported by fleet and cleanup.

.buildkite/bk.integration.pipeline.yml
```diff
@@ -70,6 +73,9 @@ steps:
           - packaging-windows
         env:
           TEST_PACKAGE: "github.com/elastic/elastic-agent/testing/integration/leak"
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud"
```
pchila: Why is CI_ELASTIC_AGENT_DOCKER_IMAGE in so many steps? IIUC this is only used by the mage load/push (and we could get rid of those as well in favor of docker load/tag/push)

I am pretty sure we don't package and push after provisioning ESS. Is this because of retries ? Even if it is we don't have to re-upload the same image, right ?

.buildkite/bk.integration.pipeline.yml
michel-laterman: added a new `ech` group for this

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -9,31 +9,8 @@ env:
   IMAGE_UBUNTU_ARM64_FIPS: "platform-ingest-elastic-agent-ubuntu-2204-fips-aarch64-1749862860"
 
 steps:
-  - label: Build and push custom elastic-agent image
-    depends_on:
-      - 'packaging-containers-x86-64-fips' # Reuse artifacts produced in .buildkite/integration.pipeline.yml
-    key: integration-fips-cloud-image
-    env:
-      FIPS: "true"
```
simitt, line 26: Why is the FIPS: true env not needed anymore?

michel-laterman: this step has mooved to be part of the `integration.pipeline.yml`

simitt: I see that `CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips"` is added in https://github.com/elastic/elastic-agent/pull/8337/files#diff-4a03dfb29d67e2bb6540e5445eccd433f03e7d8e32000bca48367284d8438f18R140; this might be what you are referring to?
Why is this not needed for [`label: "Packaging: Containers linux/arm64 FIPS"`](https://github.com/elastic/elastic-agent/pull/8337/files#diff-4a03dfb29d67e2bb6540e5445eccd433f03e7d8e32000bca48367284d8438f18R152)?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -42,6 +42,54 @@ common:
...
+          useCustomGlobalHooks: true
+      - label: "x86_64:sudo: ech"
+        key: integration-tests-ech
+        depends_on:
+          - integration-ess-custom
+        env:
+          TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud:git-${BUILDKITE_COMMIT:0:12}"
+        command: |
+          buildkite-agent artifact download build/distributions/** . --step 'packaging-ubuntu-x86-64'
+          .buildkite/scripts/steps/integration_tests_tf.sh ech true
```
michel-laterman: The script this calls into finds the deployment by exporting variables with:
```
export ELASTICSEARCH_HOST=$(buildkite-agent meta-data get "es.host")
```
This PR adds two deployments (the one in this group, and one in the step below the group) do we need to add anything to make sure that this test uses the custom ECH deploment, but the other tests use the regular deployment?

cc @v1v @pkoutsovasilis

pkoutsovasilis: I think that this splitting up in multiple steps is redundant @michel-laterman, any reason that this can't be a single step that you provision the ESS in and test the integration server?

.buildkite/scripts/steps/integration-package.sh
```diff
@@ -7,3 +7,8 @@ export SNAPSHOT="true"
 export EXTERNAL="true"
 
 mage package
+
+if [[ ${PUSH_IMAGE:-false} == "true" ]]; then
```
pkoutsovasilis: IMO this script is only for packaging, and it should stay this way 🙂 Either introduce a new script that pushes or directly in the step command after the packaging script invocation push it. That said, why you need first to load it and then push it?! this is already in the docker daemon right?

.buildkite/bk.integration.pipeline.yml
michel-laterman: good idea

.buildkite/scripts/steps/integration-package.sh
michel-laterman: just double checked, yes i've moved this to the bk pipeline commands and removed calling load

.buildkite/bk.integration-fips.pipeline.yml
michel-laterman: added

.buildkite/scripts/steps/integration_tests_custom_ech_tf.sh
```diff
@@ -0,0 +1,26 @@
...
+mage build:testBinaries
+
+# Remove deployment on exit
+trap 'ess_down' EXIT
+
+preinstall_fleet_packages
+
+# Run integration tests
+echo "~~~ Running sudo ECH integration tests"
+./buildkite/scripts/buildkite-integration-tests.sh ech true
```
pkoutsovasilis: let's keep new shell scripts at a minimum please 🙏  IIUC you can do what you need to do, just by introducing a new env var here and make this script [here](https://github.com/elastic/elastic-agent/blob/14ebe7e89eef53240f215a166e90eb5ac6486403/.buildkite/scripts/steps/integration_tests_tf.sh#L35) always provision an ECH stack e.g. `ESS_CREATE`, @v1v such env var makes sense to me, wdyt? 🙂

.buildkite/integration.pipeline.yml
```diff
@@ -97,58 +97,88 @@ steps:
...
+              secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
 
       - label: "Packaging: Containers linux/arm64 FIPS"
         key: packaging-containers-arm64-fips
         env:
           PACKAGES: "docker"
           PLATFORMS: "linux/arm64"
           FIPS: "true"
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips"
```
pkoutsovasilis: are you gonna test with arm64? 🙂 I feel that the answer is no

.buildkite/bk.integration.pipeline.yml
```diff
@@ -44,6 +44,30 @@ common:
         env_var: "EC_API_KEY"  
 
 steps:
+  - label: "Custom ECH Testing"
+    key: integration-tests-ech
+    depends_on:
+      - packaging-ubuntu-x86-64
```
pkoutsovasilis, line 50: since you effectively for now at least test x86-64 integration server I would expect to push only these artifacts 🙂

.buildkite/bk.integration.pipeline.yml
```diff
@@ -44,6 +44,30 @@ common:
...
+    env:
+      ASDF_TERRAFORM_VERSION: 1.9.3
```
pkoutsovasilis: other terraform versions are `1.9.2` why here this needs to be `1.9.3`?

testing/integration/ess/ech_test.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		statusBuffer := new(strings.Builder)
+		err = json.NewEncoder(statusBuffer).Encode(status)
+		if err != nil {
+			t.Logf("error marshaling agent status: %v", err)
+		} else {
+			t.Logf("agent status: %v", statusBuffer.String())
+		}
+
+		return status.State == int(cproto.State_HEALTHY) && status.FleetState == int(cproto.State_HEALTHY)
```
pkoutsovasilis: agent un-enrollement should be part of the smoke testing?

.buildkite/bk.integration.pipeline.yml
michel-laterman: .tool-versions contains 1.9.3, i had earlier failures i was trying to resolve (but this is unreleated)

.buildkite/integration.pipeline.yml
```diff
@@ -97,13 +97,21 @@ steps:
         env:
           PACKAGES: "docker"
           PLATFORMS: "linux/amd64"
-        command: ".buildkite/scripts/steps/integration-package.sh"
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud"
+        command: |
+          source .buildkite/scripts/common.sh
+          .buildkite/scripts/steps/integration-package.sh
+          mage cloud:push
```
pkoutsovasilis:
```suggestion
          source .buildkite/scripts/common.sh
          .buildkite/scripts/steps/integration-package.sh
          echo "~~~ Pushing cloud image"
          mage cloud:push
```

@michel-laterman nit `echo "~~~ Pushing cloud image"` because now the `mage cloud:push`  is appearing under Test packages in the logs

@v1v do you know if we need the `set -euo pipefail` here?

testing/integration/ess/ech_test.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+
+	// Check that the Fleet Server in the deployment is healthy
+	fleetServerHost, err := fleettools.DefaultURL(t.Context(), info.KibanaClient)
+	statusUrl, err := url.JoinPath(fleetServerHost, "/api/status")
+	require.NoError(t, err)
+
+	require.Eventually(t, func() bool {
+		resp, err := http.Get(statusUrl)
+		require.NoError(t, err)
```
pchila: Using require.NoError like this it will immediately terminate the test execution in case of error making the `require.Eventually` redundant. [`require.EventuallyWithT()`](https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithT) is what we want to use here.

.buildkite/integration.pipeline.yml
v1v: No, it's not needed.

However, I'll recommend using `command: <my-script>` and then move the content to the `<my-script>` file.

testing/integration/ess/ech_test.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		statusBuffer := new(strings.Builder)
+		err = json.NewEncoder(statusBuffer).Encode(status)
+		if err != nil {
+			t.Logf("error marshaling agent status: %v", err)
+		} else {
+			t.Logf("agent status: %v", statusBuffer.String())
+		}
+
+		return status.State == int(cproto.State_HEALTHY) && status.FleetState == int(cproto.State_HEALTHY)
```
pchila: It would be nice if we checked the agent state as reported by fleet server to be healthy/online

.buildkite/integration.pipeline.yml
pkoutsovasilis: I personally requested not to introduce new scripts, we already have enough customised scripts. The solution is to have the integration-package.sh be a golang based but we are far from this reality, aren't we @v1v ?! So since `set -euo pipefail` is not needed I will remove it from my proposed changes

v1v: Gotcha

testing/integration/ess/ech_test.go
michel-laterman: This might be something we need to scan for in other tests

.buildkite/bk.integration.pipeline.yml
```diff
@@ -41,9 +41,35 @@ common:
...
+    key: integration-tests-ech
+    depends_on:
+      - packaging-ubuntu-x86-64
+    env:
+      TEST_PACKAGE: "github.com/elastic/elastic-agent/testing/integration/ess"
+      TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud:git-${BUILDKITE_COMMIT:0:12}"
+      FORCE_ESS_CREATE: "true"
+    command: |
+      #!/usr/bin/env bash
+      set -euo pipefail
```
v1v: unrequired

```suggestion
```

.buildkite/integration.pipeline.yml
```diff
@@ -97,13 +97,22 @@ steps:
...
         artifact_paths:
           - build/distributions/**
         agents:
           provider: "gcp"
           machineType: "n2-standard-8"
           diskSizeGb: 200
+        plugins:
+          - elastic/vault-docker-login#v0.5.2:
+              secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
```
v1v: Since you use this twice, I recommend using the anchor approach:
- https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins

You can see how it looks also in the other pipelines.

.buildkite/integration.pipeline.yml
```diff
@@ -126,13 +135,22 @@ steps:
...
+          .buildkite/scripts/steps/integration-package.sh
+          echo "~~~ Pushing cloud image"
+          mage cloud:push
         artifact_paths:
           - build/distributions/**
         agents:
           provider: "gcp"
           machineType: "n2-standard-8"
           diskSizeGb: 200
+        plugins:
```
v1v, line 155: ditto

.buildkite/integration.pipeline.yml
```diff
@@ -126,13 +135,22 @@ steps:
...
           PLATFORMS: "linux/amd64"
           FIPS: "true"
-        command: ".buildkite/scripts/steps/integration-package.sh"
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips"
+        command: |
+          source .buildkite/scripts/common.sh
+          .buildkite/scripts/steps/integration-package.sh
+          echo "~~~ Pushing cloud image"
+          mage cloud:push
```
v1v, line 148: This duplicates the command from L102, I think it can be better to use a common script

v1v (commented):
Thanks for cleaning up some of the cosmetic leftovers. I left a few comments

.buildkite/integration.pipeline.yml
pkoutsovasilis: @v1v let's sync again about the no new scripts policy, and because I requested the no new scripts if you insist on that I will open a follow-up PR having this command as a new script. Resolved?

pkoutsovasilis (approved):
LGTM, @v1v please have another look at your comments, we need this PR as it increases the testing footprint of elastic-agent

.buildkite/integration.pipeline.yml
v1v: (I review about 40 pull requests from 40 different repositories daily, and I've been doing this for the past few weeks. 
I apologise for bringing this up again).

I usually follow these principles when I encounter duplicated content:

- Refactor it in some way
- Use a [matrix](https://buildkite.com/docs/pipelines/configure/workflows/build-matrix)

I recognise that I may be biased, but it's challenging for me to overlook this issue. Having maintained complex pipelines across various CI vendors for several years, I’ve learned the importance of keeping things simple—not only for myself but also for future contributors. I understand the trade-offs and the current implementation’s challenges; I apologise if I seem overly pedantic..

If you feel this should remain as it is, that's perfectly fine. I haven't blocked the issue; I simply wanted to provide some feedback.

Finally, if possible, I believe we should revisit and document some guidelines or principles together to prevent recurring questions in the future.

Thanks as usual 🙇

v1v (approved):
Besides what we discussed in the thread, I approve the CI part

PR #8338, opened 2025-06-04
Add filebeat_input to beat receiver agent monitoring integration test

### LINKED ISSUE #8188: runtime_experimental: otel does not gather elastic_agent.filebeat_input datastream
In the http-metrics-monitoriing component there should be an input for the `elastic-agent.filebeat_input` datastream.  An example from the beat-rendered-config.yml in a diagnostic would be below.  but when `agent.monitoring._runtime_experimental: otel` is used.  The `filebeat_input` does not appear in the `otel-final.yml` from a diagnostic and no data appears in the data_stream.

Relates #8031

```
    - data_stream:
        dataset: elastic_agent.filebeat_input
        namespace: default
        type: metrics
      failure_threshold: 5
      hosts:
        - http+unix:///opt/Elastic/Agent/data/tmp/xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock
      id: metrics-monitoring-filebeat-1
      index: metrics-elastic_agent.filebeat_input-default
      json:
        is_array: true
      metricsets:
        - json
      module: http
      namespace: filebeat_input
      path: /inputs/
      period: 1m0s
      processors:
        - add_fields:
            fields:
                input_id: metrics-monitoring-agent
            target: '@metadata'
        - add_fields:
            fields:
                dataset: elastic_agent.filebeat_input
                namespace: default
                type: metrics
            target: data_stream
        - add_fields:
            fields:
                dataset: elastic_agent.filebeat_input
            target: event
        - add_fields:
            fields:
                stream_id: metrics-monitoring-filebeat-1
            target: '@metadata'
        - add_fields:
            fields:
                id: 6d6c9275-f7d3-4381-bde0-f1ea8cf1bc53
                snapshot: false
                version: 9.1.0
            target: elastic_agent
        - add_fields:
            fields:
                id: 6d6c9275-f7d3-4381-bde0-f1ea8cf1bc53
            target: agent
        - add_fields:
            fields:
                dataset: elastic_agent.filebeat_input
            target: event
        - add_fields:
            fields:
                id: 6d6c9275-f7d3-4381-bde0-f1ea8cf1bc53
                process: filebeat
                snapshot: false
                version: 9.1.0
            target: elastic_agent
        - add_fields:
            fields:
                id: 6d6c9275-f7d3-4381-bde0-f1ea8cf1bc53
            target: agent
        - copy_fields:
            fail_on_error: false
            fields:
                - from: http.agent.beat.cpu
                  to: system.process.cpu
                - from: http.agent.beat.memstats.memory_sys
                  to: system.process.memory.size
                - from: http.agent.beat.handles
                  to: system.process.fd
                - from: http.agent.beat.cgroup
                  to: system.process.cgroup
                - from: http.agent.apm-server
                  to: apm-server
                - from: http.filebeat_input
                  to: filebeat_input
            ignore_missing: true
        - drop_fields:
            fields:
                - http
            ignore_missing: true
        - add_fields:
            fields:
                binary: filebeat
                id: filestream-monitoring
            target: component
```

leehinman:
## What does this PR do?

Adds last part of comparing process based monitoring vs beat receiver based monitoring

- Adds Windows as test case
- adds filebeat_input metrics to config
- adds filebeat_input metrics to integration test

## Why is it important?

We need to make sure that both implementations produce the same documents.

## Checklist

## Disruptive User Impact

None.  Unit test and changes to producing configuration when experimental runtime is used.

```shell
mage integration:single TestClassicAndReceiverAgentMonitoring
```

## Related issues

- Closes #8188
- Closes elastic/ingest-dev#4876

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/monitoring/v1_monitor.go
+++ b/internal/pkg/agent/application/monitoring/v1_monitor.go
@@ -700,8 +700,7 @@ func (b *BeatsMonitor) getHttpStreams(
 			httpStreams = append(httpStreams, httpStream)
 		}
 		// specifically for filebeat, we include input metrics
-		// disabled for filebeat receiver until https://github.com/elastic/beats/issues/43418 is resolved
-		if strings.EqualFold(name, "filebeat") && compInfo.RuntimeManager != component.OtelRuntimeManager {
+		if strings.EqualFold(name, "filebeat") {
 			fbDataStreamName := "filebeat_input"
 			fbDataset := fmt.Sprintf("elastic_agent.%s", fbDataStreamName)
 			fbIndexName := fmt.Sprintf("metrics-elastic_agent.%s-%s", fbDataStreamName, monitoringNamespace)
@@ -842,7 +841,6 @@ func processorsForAgentFilestream() []any {
 		addFormattedIndexProcessor(),
 	)
 	return processors
-
 }
 
 // processorsForServiceComponentFilestream returns processors used for filestream streams for components running as
--- a/testing/integration/beat_receivers_test.go
+++ b/testing/integration/beat_receivers_test.go
@@ -725,6 +725,7 @@ func genIgnoredFields(goos string) []string {
 			"log.file.fingerprint",
 			"log.file.idxhi",
 			"log.file.idxlo",
+			"log.offset",
 		}
 	default:
 		return []string{
@@ -732,6 +733,7 @@ func genIgnoredFields(goos string) []string {
 			"log.file.fingerprint",
 			"log.file.inode",
 			"log.file.path",
+			"log.offset",
 		}
 	}
 }
--- a/testing/integration/beat_receivers_test.go
+++ b/testing/integration/beat_receivers_test.go
@@ -1,7 +1,6 @@
 // Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
 // or more contributor license agreements. Licensed under the Elastic License 2.0;
 // you may not use this file except in compliance with the Elastic License 2.0.
-
 //go:build integration
 
 package integration
@@ -13,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"runtime"
 	"strings"
 	"testing"
 	"text/template"
@@ -42,6 +42,7 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
 		OS: []define.OS{
 			{Type: define.Linux},
 			{Type: define.Darwin},
+			{Type: define.Windows},
 		},
 		Stack: &define.Stack{},
 		Sudo:  true,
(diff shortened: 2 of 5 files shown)


testing/integration/beat_receivers_test.go
```diff
@@ -201,185 +238,225 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
+
+		output, err = receiverFixture.InstallWithoutEnroll(ctx, &installOpts)
+		require.NoErrorf(t, err, "error install with beat receiver monitoring without enroll: %s\ncombinedoutput:\n%s", err, string(output))
+		timestampBeatReceiver := time.Now().UTC().Format("2006-01-02T15:04:05.000Z")
+
+		require.EventuallyWithT(t, func(collect *assert.CollectT) {
+			status, statusErr := receiverFixture.ExecStatus(ctx)
+			assert.NoError(collect, statusErr)
+			// agent should be healthy
+			assert.Equal(collect, int(cproto.State_HEALTHY), status.State)
```
mauri870: You might want to apply these comments as the last argument of the assert or require functions. It will turn failure messages from something like "1 is not equal to 0" to "1 is not equal to 0 [...] agent status should be healthy" which is very helpful for debugging.

testing/integration/beat_receivers_test.go
```diff
@@ -201,185 +238,225 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
+			// we should have no normal components running
+			assert.Zero(collect, len(status.Components))
+
+			// we should have filebeatreceiver and metricbeatreceiver running
+			otelCollectorStatus := status.Collector
+			require.NotNil(collect, otelCollectorStatus)
+			assert.Equal(collect, int(cproto.CollectorComponentStatus_StatusOK), otelCollectorStatus.Status)
+			pipelineStatusMap := otelCollectorStatus.ComponentStatusMap
```
mauri870: Can we use [`assertCollectorComponentsHealthy`](https://github.com/elastic/elastic-agent/blob/5648918/testing/integration/beat_receivers_test.go#L670) here?

testing/integration/beat_receivers_test.go
```diff
@@ -201,185 +238,225 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
+		// Compare process vs beat receiver events
+		for _, tc := range tests {
+			agent := agentDocs[tc.dsType+"-"+tc.dsDataset+"-"+processNamespace].Hits.Hits[0].Source
+			otel := otelDocs[tc.dsType+"-"+tc.dsDataset+"-"+receiverNamespace].Hits.Hits[0].Source
+			ignoredFields := []string{
+				// Expected to change between agentDocs and OtelDocs
+				"@timestamp",
+				"agent.ephemeral_id",
+				"agent.id",
+				"agent.version",
```
mauri870: If we are using the translation from the otel runtime then the versions should match, you could remove this field from the ignored list.

testing/integration/beat_receivers_test.go
```diff
@@ -201,185 +238,225 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
+				"data_stream.namespace",
+				"log.file.inode",
+				"log.file.fingerprint",
+				"log.file.path",
+				"log.offset",
+
+				// needs investigation
+				"event.agent_id_status",
+				"event.ingested",
```
mauri870: event.ingested, event.duration is expected to change, so I don't think it needs investigation 🤔

testing/integration/beat_receivers_test.go
```diff
@@ -201,185 +238,225 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
+
+				// elastic_agent * fields are hardcoded in processor list for now which is why they differ
+				"elastic_agent.id",
+				"elastic_agent.snapshot",
+				"elastic_agent.version",
```
mauri870: This field should be the same I believe

testing/integration/beat_receivers_test.go
mauri870: Ah nvm, I see you used it right below.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -701,7 +701,7 @@ func (b *BeatsMonitor) getHttpStreams(
 		}
 		// specifically for filebeat, we include input metrics
 		// disabled for filebeat receiver until https://github.com/elastic/beats/issues/43418 is resolved
```
khushijain21, line 703: we may want to remove this comment as well

khushijain21:
```suggestion
		
```
we can remove this as well

testing/integration/beat_receivers_test.go
```diff
@@ -54,69 +54,121 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
...
 			ignoreFields: []string{
-				// all process related metrics are dropped for beatreceivers
-				"beat.stats.cgroup",
-				"beat.stats.cpu",
-				"beat.stats.handles",
-				"beat.stats.memstats",
-				"beat.stats.runtime",
 				"beat.elasticsearch.cluster.id",
-				"beat.stats.libbeat.config",
+				"beat.stats.cgroup.cpu.id",
```
khushijain21: we can ignore all `cgroup` metrics by just passing `"beat.stats.cgroup"`, wdyt

same for other process metrics

testing/integration/otel_test.go
```diff
@@ -1648,21 +1648,13 @@ func AssertMapsEqual(t *testing.T, m1, m2 mapstr.M, ignoredFields []string, msg
 	flatM1 := m1.Flatten()
 	flatM2 := m2.Flatten()
 	for _, f := range ignoredFields {
-		hasKeyM1, _ := flatM1.HasKey(f)
-		hasKeyM2, _ := flatM2.HasKey(f)
-
-		if !hasKeyM1 && !hasKeyM2 {
-			assert.Failf(t, msg, "ignored field %q does not exist in either map, please remove it from the ignored fields", f)
-		}
-
```
swiatekm: Why are we removing this? I thought it was here intentionally to ensure the ignored lists are trimmed whenever an element is not necessary,

swiatekm (commented):
I don't get why changes to the existing test cases are necessary in this PR. Why does uncommenting the filebeat input test case not suffice?

testing/integration/beat_receivers_test.go
mauri870: FYI: This field can differ. The `agent.version` will always report the version of the Beats receiver. As of today, this will likely show 9.1.0. However, if you try to backport this to 9.0, the main Agent document will report 9.0, while the Beats receiver will still report 9.1.0.

testing/integration/beat_receivers_test.go
mauri870: The `elastic_agent.version` field reports the version of Elastic Agent, which is the same in both tests. It should be safe to remove this field here so it gets compared properly. 

`agent.version` is another story, it reports the version of the beat, in case of standard beats this will likely match the packaged beat. In case of beats receivers, it can report a different version because we consistently bump the beats dependency.

testing/integration/beat_receivers_test.go
leehinman: :-) no it doesn't.  This section was moved, so it isn't "new", I'll see if we can trim down this list.

leehinman:
> I don't get why changes to the existing test cases are necessary in this PR. Why does uncommenting the filebeat input test case not suffice?

2 things.

1) Because I found that the single query case wasn't sufficient, so I changed it to a slice

```
query           map[string]any
to
query           []map[string]any
```

2) Added Windows

testing/integration/otel_test.go
leehinman: I'm not in favor of coupling this test to beats implementation details.  If beats stops producing a field that we ignore, that isn't a failure for this test.  This test exists to help us make sure the documents are the same.  Both documents not having a field is the same.

That being said, I'll put it back, and we can reevaluate if it becomes a burden.

leehinman:
@mauri870 @swiatekm @khushijain21 please take another look, should be clearer now.

swiatekm:
@leehinman you need to update the monitoring config generation unit tests.

leehinman:
> @leehinman you need to update the monitoring config generation unit tests.

🤦 done.

swiatekm:
FYI, the serverless tests are failing due to an unrelated infrastructure problem right now.

swiatekm (dismissed):

mauri870 (dismissed):

swiatekm:
Looks like the modified test fails right now.

khushijain21 (approved):
Looks good to me!

Issue #8268, opened 2025-05-29
Use config replace logic instead of reinstalling service in switch


michalpristas:
**Describe the enhancement:**

At the moment we do following flow when switching priviledged mode of agent:
- change permissions and other related tasks
- uninstall service
- remove service
- install service
- start service

DoD for this issue is:
- [ ] Following flow
   - change permissions and other related tasks
   - replace config
   - restart service
- [ ] Following mechanisms used for config change
   - Windows: calling [windowsService.UpdateConfig](https://pkg.go.dev/golang.org/x/sys/windows/svc/mgr#Service.UpdateConfig)
   - Systemd/Darwin: modification of a service file

PR #8347, opened 2025-06-05
Refactor user change on service

michalpristas:
This PR refactor the way how we do User change during `unprivileged` and `privileged` actions.

Pre-PR we removed and reinstalled service again with new user setup.

Post-PR we rewrite service file directly (linux and darwin) or use syscall to change config (windows). Then we restart service.

This reduces window for interrupt.
While pre-PR could leave us without a service, in this case we haven't won entirely. We can end up with broken service when service manager decides config/user is not valid/enabled. This at least leaves a trace in system logs. 

Tests covering switch between privilege modes are already present

Closes: https://github.com/elastic/elastic-agent/issues/8268


internal/pkg/agent/install/svc.go
```diff
@@ -111,6 +138,167 @@ func newService(topPath string, opt ...serviceOpt) (service.Service, error) {
 	return service.New(nil, cfg)
 }
 
+func changeSystemdServiceFile(serviceName string, serviceFilePath string, username string, groupName string) error {
```
pchila, line 148: Nit: I understand that `changeSystemdServiceFile` and `changeLaunchdServiceFile` are being unit-tested on all platforms but maybe we can move them to dedicated files in the same package like `systemd.go` and `launchd.go` for example ?

internal/pkg/agent/install/svc.go
```diff
@@ -111,6 +138,167 @@ func newService(topPath string, opt ...serviceOpt) (service.Service, error) {
...
+		return fmt.Errorf("failed to read service file %s: %w", serviceFilePath, err)
+	}
+
+	lines := strings.Split(string(content), "\n")
+	var modifiedLines []string
+	userSet := false
+	groupSet := false
+	inServiceSection := false
+
+	for _, line := range lines {
```
pchila: Is there an alternative to reading the systemd definition file line by line and modifying/adding lines under the [Service] section ?
Maybe `systemctl edit` to create an override file can be used? Or a dedicated parser ?

internal/pkg/agent/install/svc.go
```diff
@@ -111,6 +138,167 @@ func newService(topPath string, opt ...serviceOpt) (service.Service, error) {
...
+	content, err := os.ReadFile(plistPath)
+	if err != nil {
+		return fmt.Errorf("failed to read plist file %s: %w", plistPath, err)
+	}
+
+	contentStr := string(content)
+
+	if username != "" {
+		// Update or add UserName
+		userNameRegex := regexp.MustCompile(`(?s)<key>UserName</key>\s*<string>[^<]*</string>`)
```
pchila: Same considerations as for the `changeSystemd*` function above about using system commands or a dedicated parser instead of matching lines with regular expressions.
If we really need to substitute parts of xml with regular expressions couldn't we specify groups in the regular expression and replace only those instead of reprinting the whole line with an Sprintf ?

internal/pkg/agent/install/svc.go
```diff
@@ -111,6 +138,167 @@ func newService(topPath string, opt ...serviceOpt) (service.Service, error) {
...
+	} else {
+		// Remove GroupName if it exists and no group is specified
+		groupNameRegex := regexp.MustCompile(`(?s)\s*<key>GroupName</key>\s*<string>[^<]*</string>`)
+		contentStr = groupNameRegex.ReplaceAllString(contentStr, "")
+	}
+
+	// Stop the service before modifying
+	if stopFn != nil {
+		if err := stopFn(serviceName); err != nil {
+			fmt.Printf("Warning: failed to stop service %s: %v\n", serviceName, err)
```
pchila: Why is fmt.Printf used here instead of a proper log ? Is it because this function is called from a  command ?

internal/pkg/agent/install/svc_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to reload systemd daemon: %w", err)
+	}
+
+	return nil
+}
+
+// isSystemdRunning checks if systemd is the init system and is running
+func isSystemdRunning() bool {
+	// Check if systemd is PID 1
```
pchila, line 42: Please update comment since what we are doing here is implement the same check as [sd_booted()](https://www.freedesktop.org/software/systemd/man/latest/sd_booted.html)

internal/pkg/agent/install/svc_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	// Alternative check: see if systemctl command exists and works
+	cmd := exec.Command("systemctl", "is-system-running")
+	if err := cmd.Run(); err == nil {
+		return true
+	}
+
+	// Check if /proc/1/comm contains systemd
+	if data, err := os.ReadFile("/proc/1/comm"); err == nil {
+		return string(data) == "systemd\n"
+	}
```
pchila, line 56: If the check above is the documented way of checking for a systemd-booted system why do we have these 2 fallbacks ?

internal/pkg/agent/install/svc_test.go
```diff
@@ -0,0 +1,490 @@
...
+				// Count occurrences to ensure only one GroupName entry
+				groupKeyCount := strings.Count(modifiedStr, "<key>GroupName</key>")
+				require.Equal(t, 1, groupKeyCount, "Should have exactly one GroupName key")
+			} else {
+				require.NotContains(t, modifiedStr, "<key>GroupName</key>", "GroupName key should not be present when no group specified")
+			}
+
+			// Clean up
+			err = os.Remove(plistPath)
+			require.NoError(t, err, "Failed to clean up plist file")
```
pchila: Why is removing a file in a t.TempDir() executed and tested here instead of leaving it to the test cleanup ?
This check has some functional significance ?

internal/pkg/agent/install/svc_test.go
```diff
@@ -0,0 +1,490 @@
...
+			modifiedStr := string(modifiedContent)
+
+			// Verify UserName is set correctly
+			if tc.expectedUser != "" {
+				require.Contains(t, modifiedStr, "<key>UserName</key>", "UserName key should be present")
+				require.Contains(t, modifiedStr, fmt.Sprintf("<string>%s</string>", tc.expectedUser), "UserName value should be correct")
+
+				// Count occurrences to ensure only one UserName entry
+				userKeyCount := strings.Count(modifiedStr, "<key>UserName</key>")
+				require.Equal(t, 1, userKeyCount, "Should have exactly one UserName key")
```
pchila: Maybe all these requires can be changed into asserts so that multiple checks (the group check below for example) can be run on the same modified service file? Something like (warning: code is not properly formatted):
```suggestion
				if assert.Contains(t, modifiedStr, "<key>UserName</key>", "UserName key should be present") {
				assert.Contains(t, modifiedStr, fmt.Sprintf("<string>%s</string>", tc.expectedUser), "UserName value should be correct")

				// Count occurrences to ensure only one UserName entry
				userKeyCount := strings.Count(modifiedStr, "<key>UserName</key>")
				assert.Equal(t, 1, userKeyCount, "Should have exactly one UserName key")
			}
```

internal/pkg/agent/install/svc_linux.go
michalpristas: this is copy of code used already, i need to push this ability to change config upstream and will get rid of this whole section

internal/pkg/agent/install/svc.go
michalpristas: we use it already, it's just you don't see it as it's in third party lib

internal/pkg/agent/install/svc.go
michalpristas: this code is no longer there

internal/pkg/agent/install/svc_linux.go
michalpristas: i don't follow, please be explicit

internal/pkg/agent/install/svc.go
blakerouse: A systemd file is basically an ini file without the nested sections, so an ini file parser will work here.

https://github.com/go-ini/ini

internal/pkg/agent/install/svc.go
blakerouse: I would recommend using a plist parser as that is the file format.

https://github.com/DHowett/go-plist

michalpristas:
need to figure out some Server vs Home behavior differences

blakerouse (approved):
Looks good. Thanks for working through the code review.

PR #8211, opened 2025-05-22
ci: reproducible builds using pinned VM images

v1v:
## What does this PR do?

Pin the VM images and automate the autobump.

The autobump will run weekly on Sundays and will raise a PR with the required changes.

## Why is it important?

This will help with a stable CI.

## Follow-ups

- Simplify the backport strategy and use specific PRs creation per supported branch
- Enable auto-merge

## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## How to test this PR locally

```bash
$ GITHUB_TOKEN=$(gh auth token) \
  GITHUB_ACTOR=v1v \
  updatecli diff --config .ci/updatecli/updatecli-bump-vm-images.yml --values .ci/updatecli/values.d/scm.yml
```

produced

```diff
--- .buildkite/pipeline.yml
+++ .buildkite/pipeline.yml
@@ -6,12 +6,12 @@
 
   # The following images are defined here and their values will be updated by updatecli
   # Please do not change them manually.
-  IMAGE_UBUNTU_2204_X86_64: "platform-ingest-elastic-agent-ubuntu-2204-1744855248"
-  IMAGE_UBUNTU_2204_ARM_64: "platform-ingest-elastic-agent-ubuntu-2204-aarch64-1744855248"
-  IMAGE_WIN_2016: "platform-ingest-elastic-agent-windows-2016-1744855248"
-  IMAGE_WIN_2022: "platform-ingest-elastic-agent-windows-2022-1744855248"
-  IMAGE_WIN_10: "platform-ingest-elastic-agent-windows-10-1744855248"
-  IMAGE_WIN_11: "platform-ingest-elastic-agent-windows-11-1744855248"
+  IMAGE_UBUNTU_2204_X86_64: "platform-ingest-elastic-agent-ubuntu-2204-1754885210"
[...]

PR #8352, opened 2025-06-05
[CI] Stable stack version override

pazone:
## What does this PR do?
Pin ESS images versions in `create_deployment_csp_configuration.yaml`. 
Creates a new UpdateCLI pipeline that updates the `.create_deployment_csp_configuration.yaml` every day for every active branch if there's a new stable snapshot available.

Tested it on my fork for main and 9.0
Example PRs for [main](https://github.com/pazone/elastic-agent/pull/17/files) and [9.0](https://github.com/pazone/elastic-agent/pull/16/files)

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.ci/updatecli/updatecli-bump-stack-version.yml
+++ b/.ci/updatecli/updatecli-bump-stack-version.yml
@@ -40,7 +40,7 @@ actions:
     scmid: githubConfig
     sourceid: lastStableStackSnapshot
     spec:
-      automerge: false
+      automerge: true
       labels:
         - dependencies
         - backport-skip


test_infra/ess/.stable-snapshot-version
```diff
@@ -0,0 +1 @@
+9.1.0-5affded5-SNAPSHOT
```
pazone: The JSONs in the buckets contain versions like `9.1.0-5affded5` (without `-SNAPSHOT`). The deployment requires it to end with `-SNAPSHOT`. So, something must attach the `-SNAPSHOT` suffix: either the TF deployment or the UpdateCLI. 
I decided it will be UpdateCLI. The only reason: we keep the version in `.stable-snapshot-version` in the required format

.github/workflows/bump-stable-snapshot.yml
```diff
@@ -0,0 +1,57 @@
...
+    timeout-minutes: 1
+    outputs:
+      matrix: ${{ steps.generator.outputs.matrix }}
+    permissions:
+      contents: read
+    steps:
+      - id: generator
+        uses: elastic/oblt-actions/elastic/active-branches@v1
+        with:
+          exclude-branches: "7.17"
```
pazone: @v1v I'd highly appreciate a quick pre(re)vew to verify the concept.

.github/workflows/bump-stable-snapshot.yml
```diff
@@ -0,0 +1,57 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.filter.outputs.matrix) }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ matrix.branch }}
```
v1v: I think we can skip this, so we can always use the same `updatecli` pipeline from main, aka `.ci/updatecli/updatecli-bump-stack-version.yml`.

Or do you see a need to have different implementations per pipeline?

test_infra/ess/deployment.tf
```diff
@@ -61,7 +61,13 @@ locals {
 
   ess_region             = coalesce(var.ess_region, "gcp-us-east1")
   deployment_template_id = coalesce(var.deployment_template_id, "gcp-storage-optimized")
-
+  
+  stable_snapshot_version = trim(file("${path.module}/.stable-snapshot-version"), "\n")
```
v1v: **question:** what's the value of `path.module}`?

test_infra/ess/readme.md
```diff
@@ -0,0 +1,24 @@
...
+### Required variables
+`stack_version` - The version of the Elastic Stack to deploy, e.g., `9.1.0-SNAPSHOT`.
+
+### Stable snapshots and images override
+
+Variables can be overridden to use stable snapshots or images instead of the latest snapshot. This is useful for testing against a specific version of the Elastic Stack.
+
+* `integration_server_docker_image` - The Docker image for the integration server
+* `elasticsearch_docker_image` - The Docker image for Elasticsearch
+* `kibana_docker_image` - The Docker image for Kibana, e.g.
```
v1v:
```suggestion
* `kibana_docker_image` - The Docker image for Kibana
```

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,32 @@
...
+      branch: '{{ requiredEnv "BRANCH_NAME" }}'
+      commitusingapi: true
+
+sources:
+  lastStableStackSnapshot:
+    name: "Latest Stable ESS snapshot for current branch"
+    kind: json
+    spec:
+      file: 'https://storage.googleapis.com/artifacts-api/channels/{{ source "branch" }}.json'
+      key: "build"
```
v1v: I've been using `.build` in some other places instead of `build` but I think they are compatible, according to https://www.updatecli.io/docs/plugins/resource/json/#_parameters

test_infra/ess/deployment.tf
```diff
@@ -74,9 +80,23 @@ locals {
...
 
 
-  integration_server_docker_image = coalesce(var.integration_server_docker_image, local.ess_properties.docker.integration_server_image, "docker.elastic.co/cloud-release/elastic-agent-cloud:${var.stack_version}")
-  elasticsearch_docker_image = coalesce(var.elasticsearch_docker_image, local.ess_properties.docker.elasticsearch_image, "docker.elastic.co/cloud-release/elasticsearch-cloud-ess:${var.stack_version}")
-  kibana_docker_image = coalesce(var.kibana_docker_image, local.ess_properties.docker.kibana_image, "docker.elastic.co/cloud-release/kibana-cloud:${var.stack_version}")
+  integration_server_docker_image = coalesce(
+    var.integration_server_docker_image, 
+    local.ess_properties.docker.integration_server_image, 
+    local.default_integration_server_image
+  )
```
v1v: **medium**: I think the `docker_image` suffix could be normalised to `image`, for instance

```suggestion
  integration_server_image = coalesce(
    var.integration_server_image, 
    local.ess_properties.docker.integration_server_image, 
    local.default_integration_server_image
  )
```

so `integration_server_image` is actually the suffix for either the var, the local or the default.

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,32 @@
+# update-cli configuration for automated go updates
```
v1v, line 1: I miss the actions section, see https://www.updatecli.io/docs/core/action/

Have you tried to run the updatecli pipeline locally? What do i normally do to test it out is:

1. Create a test branch in my fork based on this change and push it.
2. Modify `.ci/updatecli/values.d/scm.yml` to use my fork and branch.
3. Run updatecli using the command from [L47](https://github.com/elastic/elastic-agent/pull/8352/files#diff-353a56e0e6144ea8174f2039582945e8c50c74b7e6a323caed2b10e60b9e57dcR47)
4. Before running you need to setup some env variables, the ones you set with `requiredEnv` in the updateclip pipeline.


For instance, see the description at https://github.com/elastic/elastic-agent/pull/8211


Once you run it successfully you can those details in this PR.

Thanks

v1v (commented):
I've reviewed, and it looks promising, there are some warnings and nits.

test_infra/ess/deployment.tf
pazone: 👍 `test_infra/ess`. I decided, since this file is for ess only, let's keep it in this directory

.ci/updatecli/updatecli-bump-stack-version.yml
pazone: Many thanks! 👍

test_infra/ess/deployment.tf
pazone: Good point. That will reduce the variable length. However, these variables are being overwritten in numerous places, like the FIPS pipeline, etc. So, renaming this variable implies various additional changes in pipelines and docs. Shouldn't we postpone this optimization in favour of the PR scope? I'd prefer to change the local variable name, not the public, or leave the variable name as it is for consistency

.github/workflows/bump-stable-snapshot.yml
pazone: The only difference I see is the branch name and the `.stable-snapshot-version` contents. The behaviour should be the same across all active branches

pazone: 👍

swiatekm (dismissed):

test_infra/ess/.stable-snapshot-version
pkoutsovasilis: again, and as discussed in the past, I believe that we should try and utilise this file `create_deployment_csp_configuration.yaml` which will make sure that both the stack that we provision through terraform and mage (our devs should be able to provision a stable ECH stack) are the same stable version. Also another point that I prefer going with `create_deployment_csp_configuration.yaml` is that it operates at a finer granularity and thus we can more easily push custom images, where now with this PR we introduce yet another global that overrides everything 🙂

v1v: I was actually looking for the exact discussion:
- https://github.com/elastic/elastic-agent/pull/8146/files#r2091781751

pazone: I'll try to explain this approach. Perhaps, I need to improve the provided doc. 
* `TF_VAR_integration_server_docker_image` -  has the top priority. It overwrites the image, whatever else we specify
* `create_deployment_csp_configuration.yaml` - is the second priority. If the image version specified here, the `.stable-snapshot-version` value is ignored
* `.stable-snapshot-version` - the default images version. 

`create_deployment_csp_configuration.yaml` comes from OGC-based framework and is partially used in the current ITs. The current implementation allows pinning image versions using this file. If you are absolutely against `.stable-snapshot-version`, I can remove it. But the UpdateCLI logic will become more complex and will require `integration_server_image`, `elasticsearch_image`, `kibana_image` ti be always present

pkoutsovasilis: I feel that we kinda have different expectations here; For me the expected outcome is to pin the ESS stack version to a stable one and not always follow blindly the latest available online. In principle, this should help us maintain our CI reliability (Buildkite) but also the development velocity (mage) because there have been times where the latest available stack version on cloud isn't stable. The fact that mage provisions VMs through OGC, which isn't the only provisioner tbh, shouldn't regard this PR at all IMO. This PR again should and it can target both the CI and the developers, thus, it needs to be based on `create_deployment_csp_configuration.yaml` which is already a common point. If you have concerns with it, please do share them but still the final solution should work holistically; no yaml-strings attached on my side 😆

pazone: Removed `.stable-snapshot-version` as we discussed

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,47 @@
...
+      replacepattern: '$1:{{ source "lastStableStackSnapshot" }}-SNAPSHOT"'
+actions:
+  elastic-agent:
+    kind: github/pullrequest
+    scmid: githubConfig
+    sourceid: lastStableStackSnapshot
+    spec:
+      automerge: false
+      labels:
+        - dependencies
```
pkoutsovasilis, line 45: team label, as well here? which one should go in?

pazone: `"Team:Elastic-Agent-Control-Plane"` ?

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,47 @@
...
+targets:
+  updateStackImages:
+    name: "Update create_deployment_csp_configuration.yaml"
+    sourceid: lastStableStackSnapshot
+    scmid: githubConfig
+    kind: file
+    spec:
+      file: "pkg/testing/ess/create_deployment_csp_configuration.yaml"
+      matchpattern: '(\S*_image: ".*docker.*:)(.+)"'
+      replacepattern: '$1:{{ source "lastStableStackSnapshot" }}-SNAPSHOT"'
```
v1v, line 36: IIUC, the PR with the tests contain an extra `:`:
- https://github.com/pazone/elastic-agent/pull/15/files#r2140548524

test_infra/ess/readme.md
```diff
@@ -0,0 +1,22 @@
+## Terraform-managed Elastic Stack deployment for integration tests 
+
+### Required variables
+`stack_version` - The version of the Elastic Stack to deploy, e.g., `9.1.0-SNAPSHOT`.
```
v1v, line 4: Shall we consider a fixed release version for the example?

test_infra/ess/readme.md
```diff
@@ -0,0 +1,22 @@
...
+
+Variables can be overridden to use stable snapshots or images instead of the latest snapshot. This is useful for testing against a specific version of the Elastic Stack.
+* `integration_server_docker_image` - The Docker image for the integration server
+* `elasticsearch_docker_image` - The Docker image for Elasticsearch
+* `kibana_docker_image` - The Docker image for Kibana
+
+#### The priority of the docker images is as follows:
+example for `integration_server_docker_image`:
+1) Variable override: `TF_VAR_integration_server_docker_image...` takes precedence over
+2) `pkg/testing/ess/create_deployment_csp_configuration.yaml` - `docker` section pins the images versions , e.g.:
```
v1v: I don't fully understand this section.

test_infra/ess/deployment.tf
```diff
@@ -61,7 +61,7 @@ locals {
 
   ess_region             = coalesce(var.ess_region, "gcp-us-east1")
   deployment_template_id = coalesce(var.deployment_template_id, "gcp-storage-optimized")
-
+
```
v1v: `ni`: unrequired

```suggestion
```

v1v (commented):
LGTM, I added a few notes, regarding some extra `:` in the docker image names and a bit more of context in the example section.

.ci/updatecli/updatecli-bump-stack-version.yml
pkoutsovasilis: I would rename the `lastStableStackSnapshot` to `latestStableStackBuild` since the build is what you extract from there

test_infra/ess/deployment.tf
```diff
@@ -73,10 +73,20 @@ locals {
     },
     yamldecode(file("${path.module}/../../pkg/testing/ess/create_deployment_csp_configuration.yaml")))
 
-
-  integration_server_docker_image = coalesce(var.integration_server_docker_image, local.ess_properties.docker.integration_server_image, "docker.elastic.co/cloud-release/elastic-agent-cloud:${var.stack_version}")
-  elasticsearch_docker_image = coalesce(var.elasticsearch_docker_image, local.ess_properties.docker.elasticsearch_image, "docker.elastic.co/cloud-release/elasticsearch-cloud-ess:${var.stack_version}")
-  kibana_docker_image = coalesce(var.kibana_docker_image, local.ess_properties.docker.kibana_image, "docker.elastic.co/cloud-release/kibana-cloud:${var.stack_version}")
+  integration_server_docker_image = coalesce(
```
pkoutsovasilis, line 75: ok why you need to remove the fallback part here? IMO this should be left in, thus, when somebody removes the versions from the yaml it can fallback to the stack_version

.ci/updatecli/updatecli-bump-stack-version.yml
pkoutsovasilis: from the documentation 
>“key” defines the Jsonpath key to manipulate.

so I would feel safer if that was indeed `.build` 🙂

test_infra/ess/readme.md
```diff
@@ -0,0 +1,22 @@
...
+```yaml
+docker:
+  integration_server_docker_image: "docker.elastic.co/cloud-release/elastic-agent-cloud:9.1.0-SNAPSHOT"
+  elasticsearch_docker_image: "..."
+  kibana_docker_image: "..."
```
pkoutsovasilis: and there should be a 3) here which is the stack version in case no `integration_server_docker_image` and no explicit versions are specified in the yaml 🙂

.github/workflows/bump-stable-snapshot.yml
```diff
@@ -0,0 +1,55 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ matrix.branch }}
+
+      - if: ${{ failure() }}
+        uses: elastic/oblt-actions/slack/send@v1
+        with:
+          bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
+          channel-id: "#ingest-notifications"
+          message: ":traffic_cone: Update Stable snapshot: updatecli failed for `${{ github.repository }}@${{ github.ref_name }}`, `@agent-team` and @pavel.zorin please look what's going on <${{ env.JOB_URL }}|here>"
```
pkoutsovasilis: please go by team tags here. so either remove pavel.zorin and add the eng-prod team (please decide among you @dliappis @v1v ) or remove it completely

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,48 @@
...
+actions:
+  elastic-agent:
+    kind: github/pullrequest
+    scmid: githubConfig
+    sourceid: lastStableStackSnapshot
+    spec:
+      automerge: false
```
pkoutsovasilis: this could/should be true? 🙂

.ci/updatecli/updatecli-bump-stack-version.yml
pazone: Good catch. I couldn't see it :( 
The new diff : 
```
 docker:
-  integration_server_image: "docker.elastic.co/cloud-release/elastic-agent-cloud:9.1.0-5affded5-SNAPSHOT"
-  elasticsearch_image: "docker.elastic.co/cloud-release/elasticsearch-cloud-ess:9.1.0-5affded5-SNAPSHOT"
-  kibana_image: "docker.elastic.co/cloud-release/kibana-cloud:9.1.0-5affded5-SNAPSHOT"
+  integration_server_image: "docker.elastic.co/cloud-release/elastic-agent-cloud:9.1.0-48398db3-SNAPSHOT"
+  elasticsearch_image: "docker.elastic.co/cloud-release/elasticsearch-cloud-ess:9.1.0-48398db3-SNAPSHOT"
+  kibana_image: "docker.elastic.co/cloud-release/kibana-cloud:9.1.0-48398db3-SNAPSHOT"
```

New Test PRs: https://github.com/pazone/elastic-agent/pull/17/files, https://github.com/pazone/elastic-agent/pull/16/files

test_infra/ess/readme.md
pazone: We need to discuss it with the team. The `.package-version` file that keeps this version was created a while ago.

.ci/updatecli/updatecli-bump-stack-version.yml
v1v: We don't use that feature at the moment - we have not explore that feature, however it's about the GitHub native support.

https://www.updatecli.io/docs/plugins/actions/github/ is the official docs.

We probably can talk about this for this updatecli pipeline and the other ones in a follow-up, WDYT?

.github/workflows/bump-stable-snapshot.yml
v1v: IMO, anything regarding any failures with the automation should report 'us' - so we know we can act here.

test_infra/ess/deployment.tf
pazone: If someone removes, for example, `docker.integration_server_image` from `create_deployment_csp_configuration.yaml`, the build will expectedly fail. 

I referred to this comment 
> This PR again should and it can target both the CI and the developers, thus, it needs to be based on create_deployment_csp_configuration.yaml, which is already a common point.

Do you want to fall back to `STACK_VERSION`(the latest untested snapshot)? It may be not safe

pkoutsovasilis: yes this shouldn't fail, this should fallback to `var.stack_version` as it was. Because if in the future for any reason we need to unpin this, we can just remove the entries from `create_deployment_csp_configuration.yaml` and just disable the github action, simple and straight forward without surprises 🙂

test_infra/ess/.stable-snapshot-version
pazone: After removing `.stable-snapshot-version`, I still see some strong advantages in the previous implementation:

Pros of the previous approach:

* The `create_deployment_csp_configuration.yaml` file was actively used. If an image was specified there, it would take precedence over `.stable-snapshot-version`.
* `.stable-snapshot-version` acted as a fallback. If no image was defined in `create_deployment_csp_configuration.yaml`, it ensured we still used a tested and automatically updated image. This provided a safety net in case someone removed a key by mistake.

Concerns with the current implementation:

* The previous straightforward UpdateCLI configuration has been replaced with a [regex-based setup](https://github.com/elastic/elastic-agent/pull/8352/files#diff-ba40e457a2ef19bc89b09158258879386bb237b85310d52f7913f7f2f202f9faR32). While I understand regex is already used elsewhere in UpdateCLI configs, I see this as a code smell due to its implicit coupling to YAML structure and key names.
* There's now no fallback mechanism. If someone accidentally removes an image from `create_deployment_csp_configuration.yaml`, we won't fall back to a safe, tested version.

I'm okay with moving forward with the current solution, but I wanted to highlight these potential pitfalls. Let's ensure we're aligned and mindful of these trade-offs

.ci/updatecli/updatecli-bump-stack-version.yml
pazone: Shouldn't we run it less frequently? For example, change the schedule from `daily` to `weekly` to reduce the number of bump PRs? 🤔

test_infra/ess/readme.md
pazone: I updated the doc. Now it explains all the cases in details

test_infra/ess/deployment.tf
pazone: Ok. Added the fallback to `stack_version`. Also added a comment to the `create_deployment_csp_configuration.yaml` that warns about any big changes

test_infra/ess/readme.md
pazone: The doc is updated

.github/workflows/bump-stable-snapshot.yml
pazone: Added robots-ci instead

.ci/updatecli/updatecli-bump-stack-version.yml
pazone: fixed 😄

.ci/updatecli/updatecli-bump-stack-version.yml
v1v: I think that's a question for @pkoutsovasilis , IIUC, they will actually review those PRs. IIUC, the idea afterwards is to enable the `auto-merge` capabilities, I think that's something we can take from https://github.com/elastic/elastic-agent/pull/8359 (Once I confirm it works as expected the mergify(approval/merge))

pkoutsovasilis: ok lemme check with the team please, I am inclined to say that when auto-merge true is applied then we would like to bring this in as fast as possible, but now this means a team member action is needed

pazone: @v1v @pkoutsovasilis Let's decide on it. 
If there are no caveats, I'd make it twice a week and `automerge: true`.

pkoutsovasilis: elastic-agent team decided this to be run daily as we wanna stay as close as possible to the latest stable version. But we would like to get auto-merge for such type of PRs ASAP 🙂

pazone: We will follow up on this when automerge is tested with VM images auto-update. 
CC @v1v

oakrizan (approved):

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,48 @@
+# update-cli configuration for automated go updates
+---
+name: Bump ESS snapshots
```
v1v, line 3: `question`:

you can force the branch name for the updatecli:
- https://github.com/elastic/beats/blob/00dcc0c033415906a461a794916c805caa064174/.github/workflows/updatecli.d/bump-elastic-stack-snapshot.yml#L3

That can be helpful when a new version is available, but an existing Pull Request for that branch is still open. In that case, instead of creating a new PR with a new Branch, it updates the existing open PR with a new commit.

Is that the desired behaviour?

For instance, https://github.com/elastic/beats/pull/44223 was initially updated for [elastic/beats@`b59ae67` (#44223)](https://github.com/elastic/beats/pull/44223/commits/b59ae67120b47dc162fdd68ebd3877608b394e00) but then a day after it was updated to [elastic/beats@`9fa2721` (#44223)](https://github.com/elastic/beats/pull/44223/commits/9fa27216e836d313b3f45a25df4ea3a2201e871a)


https://github.com/elastic/beats/pull/34571 explains a bit how we solved for `elastic/beats`

.github/workflows/bump-stable-snapshot.yml
```diff
@@ -0,0 +1,55 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ matrix.branch }}
+
+      - if: ${{ failure() }}
+        uses: elastic/oblt-actions/slack/send@v1
+        with:
+          bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
+          channel-id: "#ingest-notifications"
+          message: ":traffic_cone: Update Stable snapshot: updatecli failed for `${{ github.repository }}@${{ github.ref_name }}`, `@agent-team` and @robots-ci please look what's going on <${{ env.JOB_URL }}|here>"
```
v1v, line 55: `question`: do we want to notify the EA team if the GH action or updatecli failed?

v1v (commented):
LGTM. I have a question about the expected behavior if PRs get stalled due to failures. Should we reuse the same updatecli 'branch' so that it gets updated if new versions become available, even though the PRs have not been merged yet?

.github/workflows/bump-stable-snapshot.yml
pazone: I think we need to ask @pkoutsovasilis

pkoutsovasilis: yes please 🙂

.ci/updatecli/updatecli-bump-stack-version.yml
pazone: I would prefer updating existing branches. But I can't say for sure that it's the best approach. I'd like to rely on your expertise and advice with UpdateCLI @v1v

v1v: > I would prefer updating existing branches

You can apply what I mentioned above, use the `pipelineid` field with the name of the branch.

I guess you might need to use `force: true` according to https://github.com/updatecli/updatecli/issues/2211#issuecomment-2147430601 to avoid the diverging resets

.github/workflows/bump-stable-snapshot.yml
pazone: agent team is already there

.ci/updatecli/updatecli-bump-stack-version.yml
pazone: Added pipelineid and force

v1v: Thanks Pavel, however, I wanted to say `force: false` rather than `force: true` 🙇 

> if you set force to false, it won't try to update the base branch

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,51 @@
...
+    kind: github
+    spec:
+      user: '{{ requiredEnv "GITHUB_ACTOR" }}'
+      username: '{{ requiredEnv "GITHUB_ACTOR" }}'
+      owner: "{{ .scm.owner }}"
+      repository: "{{ .scm.repository }}"
+      token: '{{ requiredEnv "GITHUB_TOKEN" }}'
+      branch: '{{ requiredEnv "BRANCH_NAME" }}'
+      commitusingapi: true
+      force: true
```
v1v: Sorry mate

```suggestion
      force: false
```

v1v (approved):
LGTM, I approve the changes for the GH actions and Updatecli, there is only one piece that I accidentally mentioned wrongly `force: false` is actually the correct value. Please change that.

Let's wait for the review from others and maybe merge early next week, WDYT?

v1v (approved):
LGTM, thanks!

pazone:
@pkoutsovasilis WDYT?

.ci/updatecli/updatecli-bump-stack-version.yml
```diff
@@ -0,0 +1,51 @@
...
+      file: "pkg/testing/ess/create_deployment_csp_configuration.yaml"
+      matchpattern: '(\S*_image: ".*docker.*):(.+)"'
+      replacepattern: '$1:{{ source "lastStableStackSnapshot" }}-SNAPSHOT"'
+actions:
+  elastic-agent:
+    kind: github/pullrequest
+    scmid: githubConfig
+    sourceid: lastStableStackSnapshot
+    spec:
+      automerge: true
```
pkoutsovasilis, line 43: is the automerge enabled with this PR or is it gonna be a follow-up? I see this issue https://github.com/elastic/observability-robots/issues/2853 and I am not sure I follow

pkoutsovasilis: ok it's only to allow for somebody to enable auto-merge, thanks for tip @v1v

pkoutsovasilis (commented):
Since this PR may result in many PRs (1 for each active branch) daily that the product team will have to review and merge, I would ideally like to have the follow-up PR that auto-approves and merges all of the former in case of a green CI ([relevant issue](https://github.com/elastic/observability-robots/issues/2853)), merged in a timely fashion. Given certain priorities and scheduled time-off I would recommend to leave this PR open and merge it when the follow-up issue is already assigned and being worked on. I believe it's more efficient to invest the product team's time if needed and pin the ESS stack version manually rather than introducing more PRs for review.

pkoutsovasilis (approved):
ok I see that the follow-up PR is here https://github.com/elastic/elastic-agent/pull/8528 and already approved so no reason to delay this one 🙂

PR #8377, opened 2025-06-06
[CI] BK Vault plugin for EC access

pazone:
## What does this PR do?
Introduces BK Vault plugin. Simplifies EC access and prevents possible issues
Windows retry tested in [this build](https://buildkite.com/elastic/elastic-agent/builds/21968/summary/annotations?jid=01974577-82a6-4b48-9d43-c8e49b6addef#01974577-82a6-4b48-9d43-c8e49b6addef/1725-1754)

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


.buildkite/bk.integration.pipeline.yml
```diff
@@ -29,6 +29,11 @@ steps:
     agents:
       image: "docker.elastic.co/ci-agent-images/platform-ingest/buildkite-agent-beats-ci-with-hooks:0.5"
       useCustomGlobalHooks: true
+    plugins:
```
v1v, line 64: We previously discussed this in another thread, and we reached an agreement to use YAML anchors. You can refer to the following link for more details: https://github.com/elastic/fleet-server/pull/4895#discussion_r2096145576.

Since this is an official approach supported by Buildkite and can significantly enhance maintainability, I recommend adopting this pattern whenever the plugin is used more than once in the same Buildkite pipeline.

You can find an example here: https://github.com/elastic/elastic-agent/pull/8244/files#diff-f309b0b791582d35ad84f2c1ef44d26c489f926497a44a3f8e6d72bf3f680572.

.buildkite/scripts/steps/ess.sh
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+function check_ec_access() {
+  # EC_API_KEY provided by buildkite-vault plugin. See the pipeline step definition.
+  if [[ -z "${EC_API_KEY}" ]]; then
+    echo "Error: EC_API_KEY is empty" >&2
+    exit 1
+  fi
+}
```
v1v: I like the idea of failing fast, but I wonder whether if adding an explicit validation is certainly needed here. I'd expect `EC_API_KEY` to be available always when using the plugin, and if not, the consumer will fail drastically.

If it's finally needed, this won't work with `set -euo pipefail`, in fact, it will fail with `line 6: EC_API_KEY: unbound variable`

I think you can use a more bash-idiomatic approach with

```suggestion
function check_ec_access() {
  # EC_API_KEY provided by buildkite-vault plugin. See the pipeline step definition.
  variable=${EC_API_KEY:?Error: EC_API_KEY is not set or is empty. Please ensure it is provided by the buildkite-vault plugin.}
}
```

if will return

```
line 6: EC_API_KEY: Error: EC_API_KEY is not set or is empty. Please ensure it is provided by the buildkite-vault plugin.
```

However, I think `set -euo pipefail` might be the way to go.

v1v (commented):
I really like this approach, so it helps to document the consumers and reduce the overhead in the scripts, those scripts should run in a configured environment for simplicity.

I left a few comments

.buildkite/scripts/steps/ess.sh
pazone: Now I think we don't need this check at all. TF will fail with the corresponding error message. WDYT?

v1v: Yes, I like the simplicity here, if for any reason, we need to add more control, we can revisit it for sure.

.buildkite/bk.integration.pipeline.yml
pazone: Is it safe to use with multiple `buildkite-agent pipeline upload`? What would happen if we upload 2 pipelines with 2 `common` anchors?

v1v: AFAIK, anchors work fine - each anchor represents a plugin, and I envision no issues as it's actually supported by Buildkite, see https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins.

However, I don't fully understand your question. Can you elaborate it a bit more?

pazone: ### pipeline1.yml 
```
common:
   - plugin1: &plugin1
 ```
### pipeline2.yml
```
common:
   - plugin2: &plugin2
```

```
buildkite-agent pipeline upload pipeline1.yml
buildkite-agent pipeline upload pipeline2.yml
```
May this cause the anchors to overwrite like it happened to `env`? I applied the anchors. It's not our case yet

File: .buildkite/scripts/steps/ess.sh (hunk 2 above)
Comment (pazone) [reply .buildkite/scripts/steps/ess.sh]: Removed the checks

File: .buildkite/bk.integration.pipeline.yml [hunk 3]
```diff
@@ -14,6 +14,15 @@ env:
   IMAGE_WIN_2022: "platform-ingest-elastic-agent-windows-2022-1749131198"
   IMAGE_WIN_2025: "platform-ingest-elastic-agent-windows-2025-1749131198"
 
+# Anchors with repeatedly used plugin configurations
+# See https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins
+common:
+  - vault_ec_key: &vault_ec_key
```
v1v: WDYT to use the key with the `pro` suffix so we know it's for accessing `kv/ci-shared/platform-ingest/platform-ingest-ec-prod`? So if for any reason, we use `qa` then we could use the suffix `qa` and so on.

```suggestion
  - vault_ec_key_pro: &vault_ec_key_pro
```

.buildkite/bk.integration.pipeline.yml
```diff
@@ -451,10 +495,7 @@ steps:
     soft_fail:
       - exit_status: "*"
     plugins:
-      - elastic/vault-secrets#v0.1.0:
-          path: "kv/ci-shared/platform-ingest/buildkite_analytics_token"
-          field: "token"
-          env_var: "BUILDKITE_ANALYTICS_TOKEN"
```
v1v: What's the reason for deleting this plugin?

.buildkite/hooks/pre-command
```diff
@@ -17,6 +17,7 @@ DOCKER_REGISTRY="docker.elastic.co"
 DOCKER_REGISTRY_SECRET_PATH="kv/ci-shared/platform-ingest/docker_registry_prod"
 CI_DRA_ROLE_PATH="kv/ci-shared/release/dra-role"
 CI_GCP_OBS_PATH="kv/ci-shared/observability-ingest/cloud/gcp"
+# This key exists for backward compatibility with OGC framework
 CI_ESS_PATH="kv/ci-shared/platform-ingest/platform-ingest-ec-prod"
```
v1v: I’ve noticed that I cannot find any references for `CI_ESS_PATH`, except for [this one](https://github.com/search?q=repo%3Aelastic%2Felastic-agent%20CI_ESS_PATH&type=code) in the `pre-command` hook. Could you clarify whether we are still using `TEST_INTEG_AUTH_ESS_APIKEY_FILE`? If that’s the case, it might be a good idea to raise a GitHub issue to help us track the effort to replace `OGC`. This would allow us to link the GitHub issue in this comment as well. I personally find it helpful to track task references within comments for better organization. Thank you!

.buildkite/bk.integration.pipeline.yml
v1v: Gotcha, thanks for double-checking.

v1v (commented):
Overall, it looks good to me. However, I have a few comments:

- There is a question regarding the inclusion of the environment in the name of the anchor plugin ID.
- Please utilize the GitHub issue related to the OGC replacement to provide additional context.
- Make sure to include the removed plugin in one of the steps.

.buildkite/bk.integration.pipeline.yml
pazone: I'd prefer to follow the `Abstraction over concretion` principle. Consider this as an interface(abstraction) that represents an EC key. And the implementation specifies whether it's `prod` or `qa` key.  
Quick example, if we want to switch to `qa` key, we don't need to rename the variable

.buildkite/hooks/pre-command
pazone: Unfortunately, we can't delete `CI_ESS_PATH` because, as you've noticed as too, it is used in the OGC based framework. We haven't discussed whether we should completely delete the OGC-based pipelines and the framework or not. I can raise an issue but I'm not sure if I completely understand how to describe it

.buildkite/bk.integration.pipeline.yml
v1v: I disagree; the more concrete the name is, the easier it is for anyone to understand its use without needing to see the implementation.

For instance, we [simplified](https://github.com/elastic/elastic-agent/pull/8444) the usage of the `pre-command` because the abstraction led to issues for our users, as detailed in https://github.com/elastic/elastic-agent/pull/8418. Using declarative and explicit naming is beneficial since it makes everything self-documenting.

v1v:
I think there are missing changes in some other pipelines:

```
Start ESS stack for FIPS integration tests
source .buildkite/scripts/steps/ess_start.sh
```

See https://github.com/elastic/elastic-agent/blob/main/.buildkite/bk.integration-fips.pipeline.yml

pazone:
Updated the FIPS pipeline

.buildkite/scripts/steps/ess.ps1
```diff
@@ -13,16 +13,7 @@ function ess_up {
       Write-Error "Error: Specify stack version: ess_up [stack_version]"
       return 1
   }
-
-  $Env:EC_API_KEY = Retry-Command -ScriptBlock {  
-    vault kv get -field=apiKey kv/ci-shared/platform-ingest/platform-ingest-ec-prod
-  }
-
-  if (-not $Env:EC_API_KEY) {
```
pkoutsovasilis, line 21: this check I should remain? maybe change the error message through. @v1v wdyt?

.buildkite/scripts/steps/ess.sh
```diff
@@ -13,13 +13,6 @@ function ess_up() {
     return 1
   fi
 
-  export EC_API_KEY=$(retry -t 5 -- vault kv get -field=apiKey kv/ci-shared/platform-ingest/platform-ingest-ec-prod)
-  
-  if [[ -z "${EC_API_KEY}" ]]; then
```
pkoutsovasilis, line 18: this check I should remain? maybe change the error message through. @v1v wdyt?

.buildkite/scripts/steps/ess.ps1
v1v: I added a comment regarding the bash script:
- https://github.com/elastic/elastic-agent/pull/8377#discussion_r2132172576

I don't know the alternative for `set -u` in powershell. IIUC, terraform will fail if `EC_API_KEY` is not available, but `EC_API_KEY` should be always available when using the BK plugin, as if not, it will fail when accessing it by the plugin.

Otherwise, what's your use case?

.buildkite/scripts/steps/ess.sh
v1v: There is no need, see https://github.com/elastic/elastic-agent/pull/8377#discussion_r2132172576

.buildkite/scripts/steps/ess.ps1
pkoutsovasilis: @v1v I value explicitness, and what you say in the comment indeed may be enforced but it is implicit as it comes from terraform provider, right?

v1v: Yes, I am skeptical about validating something the Terraform provider would validate regardless.

It's important to note that the BK plugin for preparing the necessary secrets is a key requirement. 

In other words, the BK VM runner should have the essential tools available, like `terraform`, `gcloud`, or `vault`. With this setup, we can simplify our process and skip extra validation checks. For instance, we don’t need to run `command -v terraform || (echo "terraform is not installed"; exit 1) && echo "installed, move forward"` since the required tools should already be there. I believe the same logic applies to the environment variable with the secret.

I'm not against, just found the explicit validation a bit redundant, but if you think that's better, then let's add it :)

pazone: What is the status here? We agreed to remove this check here: https://github.com/elastic/elastic-agent/pull/8377#discussion_r2132172576. If the `EC_API_KEY` is missing, Terraform raises an explicit error.

Issue #8087, opened 2025-05-06
Split out integration tests into different packages


ycombinator:
In https://github.com/elastic/elastic-agent/pull/8083#issuecomment-2854281808, @pchila asked:

> Is there a specific reason for the FIPS integration test not to be defined in their own package  and run independently from the melting pot that is the `integration` package right now ?

It's a fair question and can probably be generalized to wondering if we should split out our ever-growing set of tests under the [`integration` package](https://github.com/elastic/elastic-agent/tree/main/testing/integration) into separate packages. 

@pchila since this is your idea, I'll let you expand on any pros/cons so we can then have a discussion about it with the rest of the team.  Thanks.

pchila:
cc @pkoutsovasilis

pchila:
Splitting the `integration` package would allow for fewer filters needed to select the e2e tests we need to run, lowering the number of env vars / cli args and limiting the complexity of [test selection logic](https://github.com/elastic/elastic-agent/blob/83ece2c9450fe9993f880387d59d4053a3e3cbb0/pkg/testing/define/define.go#L142-L231)

Furthermore there are multiple type of tests all in the same package with different purposes:
- [Serverless Integration tests](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/8746#0196f312-a4ae-40bd-9c34-54b10eb43bb8): runs an e2e test as a smoke test against a serverless project 
- [Extended runtime leak tests](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/8746#0196f312-a4ae-4edb-8e4b-41eb3b013853): long-running test to detect aggressive memory leaks while parsing sample logs [controlled by `TEST_LONG_RUNNING` env var](https://github.com/elastic/elastic-agent/blob/83ece2c9450fe9993f880387d59d4053a3e3cbb0/magefile.go#L2784)
- [Beats serverless tests](): currently disabled but tests, [mage targets](https://github.com/elastic/elastic-agent/blob/83ece2c9450fe9993f880387d59d4053a3e3cbb0/magefile.go#L2757-L2781) and [scripts](https://github.com/elastic/elastic-agent/blob/83ece2c9450fe9993f880387d59d4053a3e3cbb0/.buildkite/scripts/steps/beats_tests.sh#L31) are still in the codebase
[...]

PR #8263, opened 2025-05-28
ci(serverless): use OIDC for running the serverless ITs

v1v:
## What does this PR do?

Enable `Serverless` IT step using:
- `oblt-cli`
- OIDC

Use `production` to run the serverless ITs.

Enable automated provisioning of new serverless projects using GitHub actions.

## Why is it important?

Delegate the provision to the existing curated tool and service we provide with `oblt-cli`, and use OIDC to access `ephemeral` secrets only to the serverless projects created automatically.

Delegate the clean-up to the `oblt-test-environment` framework.

## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

### CI

See https://github.com/elastic/elastic-agent/actions/workflows/serverless-project.yml for the project creation

See for [buildkite](https://buildkite.com/elastic/elastic-agent/builds/21657#0197309c-32d0-4c74-add3-b16a9f86d035)

image

## Related issues

- Requires https://github.com/elastic/elastic-agent/pull/8358

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #8396, opened 2025-06-09
[tests] split up serverless and resource leaks integration tests

pkoutsovasilis:
## What does this PR do?

This PR refactors the Elastic Agent integration test structure by introducing dedicated sub-packages for specific test categories. The changes include:

- New `serverless` and `leak` test packages under `testing/integration/`, each with their own `TestMain`.
- Moved long-running resource leak tests to the `leak` package and removed the dependency on the `TEST_LONG_RUNNING` env var.
- Moved `TestLogIngestionFleetManaged` into a reusable function and invoked it from both `integration` and `serverless` packages.
- Normalized test entrypoints across Windows (PowerShell) and Unix (Bash) runners.
- Made `PreinstalledPackages` public and importable by other test packages.
- Renamed `groups_test.go` to `groups.go` to support shared usage.
- Enhanced `mage integration:check` to validate all test packages (`integration/`, `integration/serverless`, `integration/leak`).
- Updated Buildkite test scripts and pipeline definitions to accommodate the split.

## Why is it important?

The `testing/integration` package had grown into a monolithic and hard-to-maintain suite containing various unrelated test types. This structural refactor:

- Improves clarity and maintainability by grouping tests by setup type (e.g., serverless, extended runtime).
- Reduces reliance on complex filtering logic using env vars or CLI args.
- Enables tailored `TestMain` implementations per package for better control of setup/teardown behavior.
- Builds a solid baseline for better alignment of Buildkite pipelines and runtime configurations.

## Checklist

## Disruptive User Impact

None expected for end-users. CI and internal contributors may need to adjust test invocation to use the correct package path or mage targets (e.g., `mage integration:testServerless`, `integration:testForResourceLeaks`).

```
# Run standard integration tests
mage integration:test

# Run serverless smoke tests
mage integration:testServerless

# Run extended leak tests
mage integration:testForResourceLeaks
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/8087
- Closes https://github.com/elastic/elastic-agent/issues/8409
- Closes https://github.com/elastic/elastic-agent/issues/8410

### DIFF
--- a/.buildkite/scripts/steps/integration_tests_tf.sh
+++ b/.buildkite/scripts/steps/integration_tests_tf.sh
@@ -63,12 +63,3 @@ else
   fi
 fi
 
-TESTS_EXIT_STATUS=$?
-
-if [[ $TESTS_EXIT_STATUS -ne 0 ]]; then
-   echo "^^^ +++"
-   echo "Integration tests failed"
-fi
-
-exit $TESTS_EXIT_STATUS
-


pkoutsovasilis:
@pchila @ycombinator please have a look and let's sync 🙂

pchila (dismissed):
LGTM
Thank you for splitting out runtime leak and serverless tests into dedicated packages.
The work on the serverless test could be used to simplify #8263 change to a more manageable size.

.buildkite/scripts/steps/integration_tests_tf.ps1
```diff
@@ -0,0 +1,38 @@
...
+Write-Output "~~~ Building test binaries"
+& mage build:testBinaries
+if ($LASTEXITCODE -ne 0) {
+    Write-Error "Failed to build test binaries"
+    exit 1
+}
+
+$TestsExitCode = 0
+try {
+    # Get-Ess-Stack will start the ESS stack if it a BK retry, otherwise it will retrieve ESS stack metadata
```
v1v: `nit`:

```suggestion
    # Get-Ess-Stack will start the ESS stack if it is a BK retry, otherwise it will retrieve ESS stack metadata
```

.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -47,12 +49,7 @@ outputXML="build/${fully_qualified_group_name}.integration.xml"
 outputJSON="build/${fully_qualified_group_name}.integration.out.json"
 
 echo "~~~ Integration tests: ${GROUP_NAME}"
-GOTEST_ARGS=(-tags integration -test.shuffle on -test.timeout 2h0m0s)
-if [ -n "$TEST_NAME_PATTERN" ]; then
-  GOTEST_ARGS+=(-run="${TEST_NAME_PATTERN}")
-fi
-GOTEST_ARGS+=("github.com/elastic/elastic-agent/testing/integration" -v -args "-integration.groups=${GROUP_NAME}" "-integration.sudo=${TEST_SUDO}")
-
+GOTEST_ARGS=(-tags integration -test.shuffle on -test.timeout 2h0m0s "${TEST_PACKAGE}" -v -args "-integration.groups=${GROUP_NAME}" "-integration.sudo=${TEST_SUDO}")
```
v1v: `nit`: shall we note the rationale for using `-test.timeout 2h0m0s` with a comment?

.buildkite/scripts/buildkite-integration-tests.ps1
```diff
@@ -1,70 +1,73 @@
...
-    $gotestArgs = @("-tags=integration", "-shuffle=on", "-timeout=2h0m0s")
-    if ($TEST_NAME_PATTERN -ne "") {
-        $gotestArgs += "-run=${TEST_NAME_PATTERN}"
-    }
-    $gotestArgs += @("github.com/elastic/elastic-agent/testing/integration", "-v", "-args", "-integration.groups=$GROUP_NAME", "-integration.sudo=$TEST_SUDO")
+
+try
+{
+    Write-Output "~~~ Integration tests: $GROUP_NAME as user: $env:USERNAME"
+    $gotestArgs = @("-tags=integration", "-shuffle=on", "-timeout=2h0m0s", "$env:TEST_PACKAGE", "-v", "-args", "-integration.groups=$GROUP_NAME", "-integration.sudo=$TEST_SUDO")
```
v1v: ditto regarding the timeout, and also what's the reason `-test.timeout` versus `-timeout`

v1v (dismissed):
LGTM, just added a very tiny nits

.buildkite/scripts/buildkite-integration-tests.sh
pkoutsovasilis: @v1v I have seen this comment somewhere else but I can't remember where exactly... Now by default the go test will have a maximum duration of 10 minutes ([link](https://pkg.go.dev/cmd/go#hdr-Testing_flags))
> -timeout d
    If a test binary runs longer than duration d, panic.
    If d is 0, the timeout is disabled.
    The default is 10 minutes (10m).

So this is a built-in mechanism and it acts as an upper execution threshold to fail and exit any tests that haven't completed before the designated time. In our context, some of our tests on normal conditions do last 45m and thus 2 hours feels spacious enough to accommodate for new tests to come but also maintain this upper execution threshold. Does this serve as a rationale, wdyt? 🙂

.buildkite/scripts/buildkite-integration-tests.ps1
pkoutsovasilis: read above for the timeout and for the `-test.` prefix nice catch, I value consistency so I am gonna change this

.buildkite/scripts/buildkite-integration-tests.sh
v1v: Yes, it does, thanks, Panos. Do you think those comments could be added to the script itself? maybe a comment to your link https://github.com/elastic/elastic-agent/pull/8396#discussion_r2135433867

pkoutsovasilis: pushed here a0ed6f1fe759d7360ede7b319f19791a88739aa8, tell me what you think of it 🙂

.buildkite/scripts/buildkite-integration-tests.ps1
pkoutsovasilis: the `-test.` consistency issue pushed here 11f6cf1e2b9951a04b48f555e921019860917c2d 🙂

.buildkite/scripts/steps/integration_tests_tf.ps1
pkoutsovasilis: addressed here 4dbfd5fa7f33da249bc7b1b737cc52f2d9ea27b1 🙂

.buildkite/scripts/steps/integration_tests_tf.ps1
```diff
@@ -0,0 +1,38 @@
...
+    & "$PWD\.buildkite\scripts\buildkite-integration-tests.ps1" $GROUP_NAME $TEST_SUDO
+    $TestsExitCode = $LASTEXITCODE
+} finally {
+    # ess_down will destroy the ESS stack if tf state file is found, aka if this is a BK retry
+    ess_down
+}
+
+if ($TestsExitCode -ne 0) {
+    exit 1
+}
```
v1v: `super-nit`: what do you think with exit the `TestsExitCode` value instead?

```suggestion
exit $TestsExitCode
```

.buildkite/scripts/buildkite-integration-tests.ps1
```diff
@@ -0,0 +1,77 @@
...
+    else
+    {
+        Write-Output "Cannot generate HTML test report: $outputXML not found"
+    }
+}
+
+if ($TestsExitCode -ne 0)
+{
+    exit 1
+}
```
v1v: `super-nit`: ditto about returning the `TestsExitCode` variable instead of using the conditional

```suggestion
exit $TestsExitCode
```

v1v (approved):
LGTM, I reviewed the `.buildkite` folder, again my nits are super-nit.

.buildkite/scripts/steps/integration_tests_tf.ps1
pkoutsovasilis: I followed the original ps1 scripts style but your proposals feel more natural to me, so I am gonna incorporate them, ty 🙂

pkoutsovasilis: pushed here a2a072d9cf90efd96ffc634d3bb7a7d3e7a39004 🙂

.buildkite/scripts/buildkite-integration-tests.ps1
pkoutsovasilis: pushed here a2a072d9cf90efd96ffc634d3bb7a7d3e7a39004 🙂

pchila (dismissed):

pkoutsovasilis:
so the serverless step is deemed flaky because at the first run OGC gave us this error

```
libcloud.common.google.GoogleBaseError: {'message': "Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute' of service 'compute.googleapis.com' for consumer 'project_number:1024180032005'.", 'domain': 'usageLimits', 'reason': 'rateLimitExceeded'}
--
  | 2025-06-09T11:19:54Z <Greenlet at 0x7f98653fef20: _up_async(<LayoutModel: 1>)> failed with GoogleBaseError
```

which again is another hint that we should migrate to native Buildkite runners 🙂

pkoutsovasilis:
ty both @v1v @pchila for your reviews, I am still gonna hold off and wait for one from @ycombinator , having 2 reviewers from the product side it will be better for broadcasting the changes 🙂

magefile.go
```diff
@@ -2269,17 +2273,27 @@ func (Integration) Auth(ctx context.Context) error {
...
 // Matrix runs integration tests on a matrix of all supported remote hosts
 func (Integration) Matrix(ctx context.Context) error {
-	return integRunner(ctx, true, "")
+	return integRunner(ctx, "testing/integration", true, "")
 }
 
 // Single runs single integration test on remote host
 func (Integration) Single(ctx context.Context, testName string) error {
-	return integRunner(ctx, false, testName)
+	return integRunner(ctx, "testing/integration", false, testName)
```
ycombinator, line 2330: Before the split in this PR, I believe it was possible to run a single integration test.  That applied to the serverless or long-running integration tests as well.  

With the change on this line, I believe it will only be possible to run a single integration test in the `testing/integration` folder (as opposed to the `testing/integration/leak` or `testing/integration/serverless` folders)?  This is not a showstopper because there is only one leak test and only one serverless test anyway 🙂 but we should probably have a way to run a single leak or serverless test in case we add more tests to those packages in the future.

testing/integration/leak/leak_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
ycombinator: Nit: if we rename this file to something like `main_test.go` (and the same in the `serverless` package), it'll be easier to identify where the `TestMain` is defined in any integration test package because of the consistency in naming.

testing/integration/serverless/serverless_test.go
```diff
@@ -0,0 +1,46 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
```
ycombinator: Should this have `//go:build integration`?

testing/integration/leak/leak_test.go
```diff
@@ -0,0 +1,49 @@
...
+	if define.AutoDiscover {
+		define.InitAutodiscovery(nil)
+	}
+
+	runExitCode := m.Run()
+
+	if define.AutoDiscover {
+		discoveredTests, err := define.DumpAutodiscoveryYAML()
+		if err != nil {
+			log.Fatalf("Error dumping autodiscovery YAML: %v\n", err)
```
ycombinator: Should we exit here?
```suggestion
			log.Fatalf("Error dumping autodiscovery YAML: %v\n", err)
			os.Exit(1)
```

testing/integration/leak/leak_test.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		err = os.WriteFile(define.AutoDiscoveryOutput, discoveredTests, 0644)
+		if err != nil {
+			log.Fatalf("Error writing autodiscovery data in %q: %v\n", define.AutoDiscoveryOutput, err)
```
ycombinator: Should we exit here?

```suggestion
			log.Fatalf("Error writing autodiscovery data in %q: %v\n", define.AutoDiscoveryOutput, err)
			os.Exit(1)
```

testing/integration/serverless/serverless_test.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		err = os.WriteFile(define.AutoDiscoveryOutput, discoveredTests, 0644)
+		if err != nil {
+			log.Printf("Error writing autodiscovery data in %q: %s", define.AutoDiscoveryOutput, err)
+			os.Exit(1)
+		}
+	}
+
+	os.Exit(runExitCode)
```
ycombinator: At the moment, the implementation `TestMain` is identical between `leak` and `serverless` packages.  If possible, it would be good to share the implementation code to prevent unintentional divergence (e.g. https://github.com/elastic/elastic-agent/pull/8396/files#r2136172728 and https://github.com/elastic/elastic-agent/pull/8396/files#r2136171866).

testing/integration/leak/leak_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/elastic/elastic-agent/pkg/testing/define"
+
+	// import order matters, keep it here
+	_ "github.com/elastic/elastic-agent/testing/integration"
+)
+
+var flagSet = flag.CommandLine
+
+func init() {
+	define.RegisterFlags("integration.", flagSet)
```
ycombinator: Thinking out loud here.  Does it make sense to namespace the prefix further to include the package name, so in this case it would be `integration.leak`? 🤔 

On one hand, I don't think it's necessary because all the flags being registered in `define.RegisterFlags` are common to all integration test types.  

On the other hand, it would give us a way to introduce more type-specific flags later.  

Perhaps we could do a hybrid approach where the flags that apply to all test types stay under the `integration` prefix but if/when we need flags that are specific to a test type, they are registered with the `integration.<test type>` prefix?  With this approach, no changes are needed to this PR right now.

ycombinator (commented):
Thanks for taking these first steps towards splitting up our integration tests package, @pkoutsovasilis!  I left a couple of questions/comments but they are minor.

testing/integration/serverless/serverless_test.go
pkoutsovasilis: IMO we want to allow divergence, as the k8s integration tests will have to diverge in the `TestMain` 🙂 How about we apply DRY, when all remaining integration tests are split and we can make a clearer decision then, wdyt?

testing/integration/leak/leak_test.go
pkoutsovasilis: if we do that then we lose the power of invoking each integration tests package the same way, e.g. look [here](https://github.com/elastic/elastic-agent/blob/a2a072d9cf90efd96ffc634d3bb7a7d3e7a39004/.buildkite/scripts/buildkite-integration-tests.ps1#L56) this will have to be specialised in each integration tests package if we do namespacing. So I would say here we don't want to diverge 😆  wdyt?

testing/integration/leak/leak_test.go
pkoutsovasilis: actuall log.Fatalf is like that 

```
// Fatalf is equivalent to [Printf] followed by a call to [os.Exit](1).
func Fatalf(format string, v ...any) {
	std.Output(2, fmt.Sprintf(format, v...))
	os.Exit(1)
}
``` 

so no need to exit 🙂

testing/integration/leak/leak_test.go
pkoutsovasilis: look below to the comment about exiting 🙂

magefile.go
pkoutsovasilis: yes we can make dedicated mage target to allow single test invocation, wdyt?

ycombinator: Sounds good.  I'm also OK doing that in a follow up PR as we don't currently have a need for it, as long as we don't forget about it.

testing/integration/serverless/serverless_test.go
ycombinator: Sounds good to me; I see the value in waiting until we have made more splits to see what remains as common and what diverges in the `TestMain`s. As long as we don't forget to do it, though. 😄

testing/integration/leak/leak_test.go
ycombinator: Good point.  Agreed on leaving the common options under the top-level `integration.` namespace.

WDYT (for a future PR, if/when we need it) about this suggestion?

> Perhaps we could do a hybrid approach where the flags that apply to all test types stay under the integration prefix but if/when we need flags that are specific to a test type, they are registered with the integration.<test type> prefix? With this approach, no changes are needed to this PR right now.

pkoutsovasilis: yy I think it's more than sensible to add specific namespaced flags to each package according to the needs. So it sounds great 🙂

testing/integration/serverless/serverless_test.go
pkoutsovasilis: pushed here 7f12158139edf49257579eeba6793640f8359e4e 🙂

testing/integration/leak/leak_test.go
pkoutsovasilis: pushed here 919d25aa31959b0dda01571202bedda836862c50, feel free to un-resolve if you spot anything else 🙂

magefile.go
pkoutsovasilis: created this one https://github.com/elastic/elastic-agent/issues/8414 so we won't forget 😉

testing/integration/serverless/serverless_test.go
pkoutsovasilis: created this one https://github.com/elastic/elastic-agent/issues/8413 so we won't forget to do it 😉

ycombinator (dismissed):

v1v (dismissed):
I reviewed the recent changes at https://github.com/elastic/elastic-agent/pull/8396/commits, I didn't see any further changes in the filers under the `.buildkite` folder - hence LGTM :)

testing/integration/leak/leak_test.go
ycombinator: In that case, we should use `log.FatalF` and remove the `os.Exit(1)`s from the `serverless/main_test.go` as well to make them consistent.

pkoutsovasilis: one delivery with extra `log.FatalF` is here 2df4ac3507caf8ac707e34a7f99639d2bbb481b9 🙂

.buildkite/scripts/buildkite-integration-tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+try
+{
+    Write-Output "~~~ Integration tests: $GROUP_NAME as user: $env:USERNAME"
+    # -test.timeout=2h0m0s is set because some tests normally take up to 45 minutes.
+    # This 2-hour timeout provides enough room for future, potentially longer tests,
+    # while still enforcing a reasonable upper limit on total execution time.
+    # See: https://pkg.go.dev/cmd/go#hdr-Testing_flags
+    $gotestArgs = @("-tags=integration", "-test.shuffle=on", "-test.timeout=2h0m0s", "$env:TEST_PACKAGE", "-v", "-args", "-integration.groups=$GROUP_NAME", "-integration.sudo=$TEST_SUDO")
+    & gotestsum --no-color -f standard-quiet --junitfile-hide-skipped-tests --junitfile "${outputXML}" --jsonfile "${outputJSON}" -- @gotestArgs
+    $TestsExitCode = $LASTEXITCODE
```
dliappis, line 58: today we discussed a problematic pattern related to the use of finally with ps1 and buildkite with @v1v and @pazone ; if there is a test failure above, only the last command -- in this case probably something from lines 65 onwards -- executed will be surfaced and expanded in the log output) and will make it hard to understand what was the actual failure in the test.

So I suggest that if `$TestsExitCode != 0` we expand this section using `^^^+++` as documented in https://buildkite.com/docs/pipelines/configure/managing-log-output#collapsing-output

.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -52,12 +54,11 @@ outputXML="build/${fully_qualified_group_name}.integration.xml"
...
-GOTEST_ARGS+=("github.com/elastic/elastic-agent/testing/integration" -v -args "-integration.groups=${GROUP_NAME}" "-integration.sudo=${TEST_SUDO}" "-integration.fips=${FIPS:-false}")
-
+# -test.timeout=2h0m0s is set because some tests normally take up to 45 minutes.
+# This 2-hour timeout provides enough room for future, potentially longer tests,
+# while still enforcing a reasonable upper limit on total execution time.
+# See: https://pkg.go.dev/cmd/go#hdr-Testing_flags
+GOTEST_ARGS=(-tags integration -test.shuffle on -test.timeout 2h0m0s "${TEST_PACKAGE}" -v -args "-integration.groups=${GROUP_NAME}" "-integration.sudo=${TEST_SUDO}" "-integration.fips=${FIPS:-false}")
 set +e
 TEST_BINARY_NAME="elastic-agent" AGENT_VERSION="${AGENT_VERSION}" SNAPSHOT=true \
   gotestsum --no-color -f standard-quiet --junitfile-hide-skipped-tests --junitfile "${outputXML}" --jsonfile "${outputJSON}" -- "${GOTEST_ARGS[@]}"
```
dliappis, line 64: same comment for the next line as for windows

.buildkite/scripts/steps/integration_tests_tf.ps1
```diff
@@ -0,0 +1,36 @@
...
+    Write-Error "Failed to build test binaries"
+    exit 1
+}
+
+$TestsExitCode = 0
+try {
+    # Get-Ess-Stack will start the ESS stack if it is a BK retry, otherwise it will retrieve ESS stack metadata
+    Get-Ess-Stack -StackVersion $PACKAGE_VERSION
+    & "$PWD\.buildkite\scripts\buildkite-integration-tests.ps1" $GROUP_NAME $TEST_SUDO
+    $TestsExitCode = $LASTEXITCODE
```
dliappis, line 32: same comment as for the previous ps1 script -- if e.g. both a test fails and ess_down too, the latter doesn't fail the build but it'll be expanded masking the real test issue.

.buildkite/scripts/buildkite-integration-tests.ps1
dliappis, line 58:
```suggestion
    $TestsExitCode = $LASTEXITCODE
    if ($TestsExitCode -ne 0) {
        Write-Host "^^^ +++"
        Write-Host "Test failures occurred"
    }
```

.buildkite/scripts/buildkite-integration-tests.sh
dliappis: I can't leave a comment for the next line (after `TESTS_EXIT_STATUS=$?`) but we could do the same as above in bash i.e.

```shell
TESTS_EXIT_STATUS=$?
if [[ $TESTS_EXIT_STATUS -ne 0 ]]; then
    echo "^^^ +++"
    echo "Test failures occurred"
fi
```

.buildkite/scripts/steps/integration_tests_tf.ps1
dliappis, line 32:
```suggestion
    $TestsExitCode = $LASTEXITCODE
    if ($TestsExitCode -ne 0) {
        Write-Host "^^^ +++"
        Write-Host "Test failures occurred"
    }
```

pkoutsovasilis: Have a look here fa9657c139769167b8ee3ba2a9abae621eb7e0e9 🙂

.buildkite/scripts/buildkite-integration-tests.sh
pkoutsovasilis: Have a look here fa9657c139769167b8ee3ba2a9abae621eb7e0e9 🙂

.buildkite/scripts/buildkite-integration-tests.ps1
pkoutsovasilis: Have a look here fa9657c139769167b8ee3ba2a9abae621eb7e0e9 🙂

pkoutsovasilis:
@dliappis, as we discussed during our offline sync, I tried to simulate the failure log group expansion with an artifical failure and I think it works as expected. Go to the BK build https://buildkite.com/elastic/elastic-agent/builds/22202, click on any of the failed steps and the respective log group should be auto-expanded already for you. wdyt? 🙂

dliappis:
> @dliappis, as we discussed during our offline sync, I tried to simulate the failure log group expansion with an artifical failure and I think it works as expected. Go to the BK build https://buildkite.com/elastic/elastic-agent/builds/22202, click on any of the failed steps and the respective log group should be auto-expanded already for you. wdyt? 🙂

Indeed this is really great. The testing section that failed got expanded as we want:

image

dliappis (approved):
LGTM from CI PoV (all the stuff under `.buildkite` looks good and is tested)
In general, this is a great improvement which will allow a ton of simplification in how we invoke tests (currently with custom scripts). Thanks!

pchila (approved):
LGTM :rocket:

.buildkite/scripts/buildkite-integration-tests.ps1
```diff
@@ -0,0 +1,80 @@
+param(
+    [string]$GROUP_NAME,
+    [string]$TEST_SUDO
+)
+
+if (-not $GROUP_NAME)
+{
+    Write-Error "Error: Specify the group name: sudo-integration-tests.ps1 <group_name> <true|false>"
```
v1v: `very nit`:

```suggestion
    Write-Error "Error: Specify the group name: buildkite-integration-tests.ps1 <group_name> <true|false>"
```

.buildkite/scripts/buildkite-integration-tests.ps1
```diff
@@ -0,0 +1,80 @@
...
+    exit 1
+}
+
+if (-not $TEST_SUDO)
+{
+    Write-Error "Error: Specify the test sudo: sudo-integration-tests.ps1 <group_name> <true|false>"
```
v1v: `very nit`:

```suggestion
    Write-Error "Error: Specify the test sudo: buildkite-integration-tests.ps1 <group_name> <true|false>"
```

.buildkite/scripts/steps/integration_tests_tf.sh
```diff
@@ -65,3 +63,12 @@ else
   fi
 fi
 
+TESTS_EXIT_STATUS=$?
```
v1v: `question`:  does `set -euo pipefail` have a higher precedence and `ESTS_EXIT_STATUS=$?` won't be honoured if any errors? Maybe a trap could be the way to go here?

dliappis: That's a good comment. It's probably my fault suggesting that we do what we did on another shell script, however, there [set +e](https://github.com/elastic/elastic-agent/pull/8396/files#diff-95168e530be96c94bafabea8a7f0bb1781509bae27e8975009c8775a3748df89L61) is defined before the test execution and thus allows the teardown to run.

In this case, there is no teardown and the script will fail and expand once anything here fails. So we don't need this section (lines 66->74)

.buildkite/scripts/buildkite-integration-tests.ps1
pkoutsovasilis: pushed here 0b85587a079f375221017a9b4029de5ef9b76080 🙂

.buildkite/scripts/buildkite-integration-tests.ps1
pkoutsovasilis: pushed here 0b85587a079f375221017a9b4029de5ef9b76080 🙂

.buildkite/scripts/steps/integration_tests_tf.sh
pkoutsovasilis: nice catch, this is redundant, removed here f088a5f585cb20299268fd4ad3def9b11535cb56 🙂

v1v (approved):
LGTM, I reviewed the recent changes in the `.buldkite` folder after reviewing them earlier too :)
Thank you for your collaboration on the various review comments 🙇

PR #8417, opened 2025-06-09
Integration Test: upgrade Integrations Server in ECH

ycombinator:
## What does this PR do?
This PR adds an integration test, `TestUpgradeIntegrationsServer`,  that spins up an ECH deployment that's at least of version `8.19.0` and upgrades it to a later, randomly-chosen version available in that deployment's ECH region. 

The test will be run with FIPS-capable artifacts against the FRH ECH environment (when it's ready).  The test will not run on every commit but only on PRs that have a special label on them,  `Testing:run:TestUpgradeIntegrationsServer`.  This label will be automatically added to PRs that bump the https://github.com/elastic/elastic-agent/blob/main/testing/integration/testdata/.upgrade-test-agent-versions.yml file, thereby ensuring the test runs whenever new versions of the stack are available for testing.

## Why is it important?

To ensure that Integrations Server in ECH (which runs Elastic Agent as a Fleet Server) can be upgraded without problems.

## Checklist

## Disruptive User Impact

None; this PR adds a new integration test.


pkg/testing/ess/statful_provisioner.go
```diff
@@ -142,12 +145,72 @@ func (p *statefulProvisioner) Delete(ctx context.Context, stack common.Stack) er
...
+	return versions, nil
+}
+
+func (p *StatefulProvisioner) createDeployment(ctx context.Context, r common.StackRequest, tags map[string]string) (*CreateDeploymentResponse, error) {
 	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
 	defer cancel()
 
 	p.logger.Logf("Creating cloud stack %s [stack_id: %s]", r.Version, r.ID)
-	name := fmt.Sprintf("%s-%s", strings.Replace(p.cfg.Identifier, ".", "-", -1), r.ID)
+	name := fmt.Sprintf("%s-%s", strings.ReplaceAll(p.cfg.Identifier, ".", "-"), r.ID)
```
ycombinator, line 216: This change has nothing to do with the topic of this PR; it is just there to make the linter happy.

cmacknz:
> To ensure that Integrations Server in ECH (which runs Elastic Agent as a Fleet Server) can be upgraded without problems.

What changes in a PR of ours would cause an integrations server upgrade to fail? We have broken integrations server starting up plenty of times but I've never seen us break upgrading it, which I think is just replacing the container version.

Something should test this, but we may want it to happen on a schedule. This should probably just be driven by snapshots?

ycombinator:
> > To ensure that Integrations Server in ECH (which runs Elastic Agent as a Fleet Server) can be upgraded without problems.
> 
> What changes in a PR of ours would cause an integrations server upgrade to fail? We have broken integrations server starting up plenty of times but I've never seen us break upgrading it, which I think is just replacing the container version.
> 
> Something should test this, but we may want it to happen on a schedule. This should probably just be driven by snapshots?

Yeah, good point, there's no need to run this test on every PR or even every commit to `main`.  Perhaps we could run it as part of the same automation that updates https://github.com/elastic/elastic-agent/blob/main/testing/integration/testdata/.upgrade-test-agent-versions.yml?

cmacknz:
> Yeah, good point, there's no need to run this test on every PR or even every commit to main. Perhaps we could run it as part of the same automation that updates https://github.com/elastic/elastic-agent/blob/main/testing/integration/testdata/.upgrade-test-agent-versions.yml?

This makes a lot of sense if we can set it up.

I also suspect that there might not be much of a difference between Integrations server starting up for the first time and Integrations server upgrading in terms of what needs to work for it to succeed. TBH I don't know enough about how it works.

Elastic Agent is stateless with no data dependencies, Fleet Server does depend on some Elasticsearch indices but it doesn't own the index templates or mappings for them (Fleet does). If we somehow broke the .fleet-* system indices I think Integrations Server would still upgrade just fine and Fleet wouldn't work.

pkoutsovasilis:
Hello folks, sorry for chiming in but I like what you discuss here 🙂 So, the PRs that bump the [.upgrade-test-agent-versions.yml](https://github.com/elastic/elastic-agent/blob/main/testing/integration/testdata/.upgrade-test-agent-versions.yml) run daily and an automated PR opens against each active branch, given that there are actual changes. Maybe we could introduce a PR label that such PRs will have?! In this way, if we wanna test this under a normal PR we could just add the label and have it tested, wdyt? 🙂

oakrizan (approved):
reviewed only CI part, lgtm

ycombinator:
Thanks for chiming in, @pkoutsovasilis 

> Maybe we could introduce a PR label that such PRs will have?! In this way, if we wanna test this under a normal PR we could just add the label and have it tested, wdyt? 🙂

I'm not sure we will need to run this test on normal PRs (see https://github.com/elastic/elastic-agent/pull/8417#issuecomment-2964075432 and https://github.com/elastic/elastic-agent/pull/8417#issuecomment-2967967263) but, sure, your label idea would give us the option to do that if/when we wanted.  Adding that special label on the PRs that bump the [.upgrade-test-agent-versions.yml](https://github.com/elastic/elastic-agent/blob/main/testing/integration/testdata/.upgrade-test-agent-versions.yml) so that our BK pipeline could check for it's existence (how is TBD), and then, if it exists, run this test, makes sense to me.  So now the next question is: how can our BK pipeline get access to labels for the PR it's running CI for?  Is this something we already know how to do?  Maybe @v1v @oakrizan @pazone could shed some light here?

oakrizan:
>how can our BK pipeline get access to labels for the PR it's running CI for

We can access the labels in the pipelines, eg. https://github.com/elastic/beats/blob/main/.buildkite%2Fauditbeat%2Fauditbeat-pipeline.yml#L266

ycombinator:
Thanks @oakrizan for the quick response and link!  I'll update this PR with changes to the BK pipeline to conditionally run `TestUpgradeIntegrationsServer` only when the special label is present and to also add the special label automatically on PRs that update `.upgrade-test-agent-versions.yml`.

Now to decide what to name our special label. 🙂 

We have a `Testing` label in this repo already.  How about "extending" it to indicate we want CI to run the test in this PR, i.e. we introduce a new label named `Testing:run:TestUpgradeIntegrationsServer`? Too verbose? I'm open to other suggestions.

kaanyalti (commented):
The changes look good to me, will approve once ci is green

kaanyalti (approved):

oakrizan (dismissed):
lgtm, reviewed CI related part

testing/integration/upgrade_integrations_server_test.go
```diff
@@ -0,0 +1,158 @@
...
+		APIKey:     echApiKey,
+		Region:     echRegion,
+	})
+	require.NoError(t, err)
+	prov.SetLogger(t)
+	statefulProv, ok := prov.(*ess.StatefulProvisioner)
+	require.True(t, ok)
+
+	minStartVersion := version.NewParsedSemVer(8, 19, 0, "", "")
+	startVersion, endVersion := getRandomStackVersionsPair(t, statefulProv, minStartVersion, nil)
```
michel-laterman: Why use a random pair? Shouldn't we test a couple common scenarios? i.e., `LAST MINOR -> LATEST`, `LAST PATCH -> LATEST`, or `PREV MAJOR -> LATEST`?
Sourced by a file similar to `testing/integration/testdata/.upgrade-test-agent-versions.yml`?

ycombinator: The random pair was to make sure that, over time, we end up testing several different version pairs but any single CI run didn't take took long. But I'll switch it to your suggestion.  It'll have pretty much the same effect; the only starting version we'll miss is `9.0.1` as it's neither the last minor or last patch or previous major that's been released.  I think that's OK as we don't have a test for covering the integrations server upgrade from that version anyway.

ycombinator: Changed in 36ae69c176.

testing/integration/upgrade_integrations_server_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+func getUpgradeStartVersions(t *testing.T) version.SortableParsedVersions {
+	versions, err := upgradetest.GetUpgradableVersions()
+	require.NoError(t, err, "could not get upgradable versions")
+
+	filteredVersions := make([]*version.ParsedSemVer, 0)
+	for _, ver := range versions {
+		if ver.Less(*upgradetest.Version_8_19_0_SNAPSHOT) {
+			continue // Skip versions older than 8.19.0
+		}
```
michel-laterman: We should also skip all 9.0.x versions as we are not producing a FIPS build for that release

ycombinator: Done in 9bd1130e59.

testing/integration/upgrade_integrations_server_test.go
```diff
@@ -109,19 +105,37 @@ func TestUpgradeIntegrationsServer(t *testing.T) {
...
 }
+
+func isFIPSCapableVersion(ver *version.ParsedSemVer) bool {
+	// Versions < 8.19.0 are not FIPS-capable
+	if ver.Less(version.NewParsedSemVer(8, 19, 0, "", "")) {
+		return false
+	}
+
+	// Version 9.0.0 is not FIPS-capable
+	if ver.Equal(version.NewParsedSemVer(9, 0, 0, "", "")) ||
```
michel-laterman: It's all 9.0 patches, not just 9.0.0

ycombinator: Good catch; fixed in 0e90ad82f2.

michel-laterman (dismissed):
LGTM 🚀

oakrizan (approved):
lgtm, reviewed CI related part

pkg/testing/ess/deployment.go
```diff
@@ -324,23 +399,15 @@ func overallStatus(statuses ...DeploymentStatus) DeploymentStatus {
 //go:embed create_deployment_request.tmpl.json
 var createDeploymentRequestTemplate string
 
-//go:embed create_deployment_csp_configuration.yaml
+//go:embed deployment_csp_configuration.yaml
```
pkoutsovasilis, line 402: I am pretty sure that this file being renamed breaks the following:
* https://github.com/elastic/elastic-agent/blob/73737c9a3c70d9bdbeade078097dcfae1bb61719/.ci/updatecli/updatecli-bump-stack-version.yml#L29
* https://github.com/elastic/elastic-agent/blob/73737c9a3c70d9bdbeade078097dcfae1bb61719/.mergify.yml#L39
* https://github.com/elastic/elastic-agent/blob/73737c9a3c70d9bdbeade078097dcfae1bb61719/test_infra/ess/readme.md#L9

testing/integration/upgrade_integrations_server_test.go
```diff
@@ -0,0 +1,140 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+//go:build integration
```
pkoutsovasilis: this no longer belongs in the main `testing/integration` package. Ideally there should be a dedicated `fips` one but until that day this should live inside `testing/integration/ess`

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -81,6 +81,7 @@ steps:
               - "true"
             groups:
               - fleet # currently there is only a single test in the fleet group, add more tests once they have been defined
+              - ech-deployment # TODO: remove once tested in PR
```
pkoutsovasilis: @ycombinator please remember to remove this when you are done with your testing 🙂

pkoutsovasilis (changes requested):

pkg/testing/ess/deployment.go
ycombinator: Good catch, @pkoutsovasilis – thanks! Will make sure to update all references.

.buildkite/bk.integration-fips.pipeline.yml
ycombinator: Thanks for catching; removed in b5196b3917.

testing/integration/upgrade_integrations_server_test.go
ycombinator: Relocated in 601a601835.

pkg/testing/ess/deployment.go
ycombinator: Fixed in a9f5e880c12d7da7ee62defb9a536bdb3a4d25e7.

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -113,6 +113,26 @@ steps:
             groups:
               - fleet
 
+      - label: "fips:upgrade-ech-deployment"
+        if: build.env("BUILDKITE_PULL_REQUEST") == "true" &&  build.env("GITHUB_PR_LABELS") =~ /.*(Testing:run:TestUpgradeIntegrationsServer).*/
+        env:
+          FIPS: "true"
+          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
+          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
+          TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud:git-${BUILDKITE_COMMIT:0:12}"
```
pkoutsovasilis:
```suggestion
      - label: "fips:upgrade-ech-deployment"
        if: build.env("BUILDKITE_PULL_REQUEST") == "true" &&  build.env("GITHUB_PR_LABELS") =~ /.*(Testing:run:TestUpgradeIntegrationsServer).*/
        env:
          FIPS: "true"
          CUSTOM_IMAGE_TAG: "git-${BUILDKITE_COMMIT:0:12}"
          CI_ELASTIC_AGENT_DOCKER_IMAGE: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud"
          TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-fips-cloud:git-${BUILDKITE_COMMIT:0:12}"
          TEST_PACKAGE: "github.com/elastic/elastic-agent/testing/integration/ess"
```

ycombinator:
Adding `Testing:run:TestUpgradeIntegrationsServer` to see if it triggers the test in this PR.

ycombinator:
buildkite test this

ycombinator:
I added the `Testing:run:TestUpgradeIntegrationsServer` label on this PR and then [manually triggered a build](https://github.com/elastic/elastic-agent/pull/8417#issuecomment-3006456704).  I was expecting the build to run the `fips:upgrade-ech-deployment` buildkite step (introduced in this PR) but it [doesn't look like it did](https://buildkite.com/elastic/elastic-agent/builds/23082). 😞 

I'm going to try and temporarily remove the conditional check for the presence of the label in the Buildkite pipeline to try and narrow down the problem.

ycombinator:
> I'm going to try and temporarily remove the conditional check for the presence of the label in the Buildkite pipeline to try and narrow down the problem.

Removing the check seems to have done the trick:

Screenshot 2025-06-25 at 18 17 11

So it looks like something is wrong with the check.  I've added it back via b541ff01247ce462276c1e02b1400bcb6a606f6a.  @oakrizan can you spot what might be wrong with the check?  Thanks.

.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -113,6 +113,27 @@ steps:
             groups:
               - fleet
 
+      - label: "fips:upgrade-ech-deployment"
+        if: build.env("BUILDKITE_PULL_REQUEST") == "true" &&  build.env("GITHUB_PR_LABELS") =~ /.*(Testing:run:TestUpgradeIntegrationsServer).*/
```
oakrizan: Please use `build.env("BUILDKITE_PULL_REQUEST") != "false"`.
Here's the described behaviour of the env var: https://buildkite.com/docs/pipelines/configure/environment-variables#:~:text=BUILDKITE_PULL_REQUEST%20%23,a%20pull%20request.

ycombinator: Fixed in f3927c205f.

ycombinator:
Thanks @oakrizan for your input! 

I can see in the [BK UI](https://buildkite.com/elastic/elastic-agent/builds/23103#0197abf6-a190-4ec9-8453-92277f793337) that the `fips:upgrade-ech-deployment` step was run with the label present 🎉.  Let me try to remove the label from this PR now and trigger a new BK build.  I would expect that step not to be run in this case.

ycombinator:
buildkite test this

ycombinator:
I can see in the [BK UI](https://buildkite.com/elastic/elastic-agent/builds/23113) that the latest build did not run the `fips:upgrade-ech-deployment` step, as expected, since the PR no longer had the `Testing:run:TestUpgradeIntegrationsServer` label on it.  So the label check is now working correctly.

This PR is ready for review again.

ycombinator:
buildkite test this

.buildkite/scripts/steps/ess_start.sh
```diff
@@ -19,3 +19,4 @@ buildkite-agent meta-data set "kibana.host" $KIBANA_HOST
 buildkite-agent meta-data set "kibana.username" $KIBANA_USERNAME
 buildkite-agent meta-data set "kibana.pwd" $KIBANA_PASSWORD
 buildkite-agent meta-data set "integrations_server.host" $INTEGRATIONS_SERVER_HOST
+buildkite-agent meta-data set "ec_api_key" $EC_API_KEY
```
pkoutsovasilis: nit: this is not required

.buildkite/scripts/steps/integration_tests_tf.sh
```diff
@@ -46,6 +46,7 @@ else
   export KIBANA_USERNAME=$(buildkite-agent meta-data get "kibana.username")
   export KIBANA_PASSWORD=$(buildkite-agent meta-data get "kibana.pwd")
   export INTEGRATIONS_SERVER_HOST=$(buildkite-agent meta-data get "integrations_server.host")
+  export EC_API_KEY=$(buildkite-agent meta-data get "ec_api_key")
```
pkoutsovasilis: nit: this is not required

pkoutsovasilis (dismissed):
except two nits, which we can deal in a follow-up PR since we have to have one either way to remove redundant env vars from FIPS-related CI steps, this LGTM. Thx for addressing all of my comments @ycombinator 🙂

pkoutsovasilis:
I had another look in the new CI step [here](https://buildkite.com/elastic/elastic-agent/builds/23139#0197ae71-33af-49e7-84f4-677e8ef28ba1/190-1111), Does it actually run anything @ycombinator ?

ycombinator:
> I had another look in the new CI step [here](https://buildkite.com/elastic/elastic-agent/builds/23139#0197ae71-33af-49e7-84f4-677e8ef28ba1/190-1111), Does it actually run anything @ycombinator ?

Yes, it runs one test, `TestUpgradeIntegrationsServer`.  You can see it in the Artifacts tab, `build/TEST-report.html`.

.buildkite/scripts/steps/integration_tests_tf.sh
ycombinator: Removed in 4c363e332f.

.buildkite/scripts/steps/ess_start.sh
ycombinator: Removed in 4c363e332f.

ycombinator:
> ... since we have to have one either way to remove redundant env vars from FIPS-related CI steps

PR for cleaning up these env vars: https://github.com/elastic/elastic-agent/pull/8710

oakrizan (dismissed):
lgtm for ci part

pkoutsovasilis (approved):
LGTM 🙂

oakrizan (approved):
lgtm for ci part

Issue #8419, opened 2025-06-09
[Flaky Test]: <test_case> – <short_error_message>


michel-laterman:
### Failing test case

TestFIPSAgentConnectingToFIPSFleetServerInECHFRH

### Error message

expcted: true actual: false

### Build

https://buildkite.com/elastic/elastic-agent/builds/22071#01975659-8dc0-4db0-b000-7b4b009a5626

### OS

Linux

### Stacktrace and notes

```markdown
The FIPS ECH test is flakey.

In the example build it failed to detect that the intgrations server (fleet-server) instance associated with `95e4a28646a0400fb9df756e6fa2c958` was running FIPS compatible distribution.

Logging onto the ECH deployment we were able to check the doc related to the fleet-server and found it was flagged as a FIPS compatible distribution:

GET .fleet-agents/_doc/dd468dd6-94d9-4bbb-8623-358e99c014f1
#! this request accesses system indices: [.fleet-agents-7], but in a future major version, direct access to system indices will be prevented by default
{
  "_index": ".fleet-agents-7",
  "_id": "dd468dd6-94d9-4bbb-8623-358e99c014f1",
  "_version": 90,
  "_seq_no": 89,
  "_primary_term": 1,
  "found": true,
  "_source": {
    "access_api_key_id": "JWhkVpcBtgv6KyRZJyQy",
    "action_seq_no": [
      -1
    ],
    "active": true,
    "agent": {
      "id": "dd468dd6-94d9-4bbb-8623-358e99c014f1",
      "version": "9.1.0"
    },
    "enrolled_at": "2025-06-09T20:31:33Z",
    "local_metadata": {
      "elastic": {
        "agent": {
          "build.original": "9.1.0-SNAPSHOT (build: 2e7352b337e693be46b404cf0847a8acff9a6f2a fips-distribution: true at 2025-06-09 19:54:53 +0000 UTC)",
          "complete": false,
          "fips": true,
          "id": "dd468dd6-94d9-4bbb-8623-358e99c014f1",
          "log_level": "info",
          "snapshot": true,
          "unprivileged": true,
          "upgradeable": false,
          "version": "9.1.0"
        }
      },
      "host": {
        "architecture": "x86_64",
        "hostname": "ff634161e9c0",
        "id": "",
        "ip": [
          "127.0.0.1/8",
          "172.17.0.37/16"
        ],
        "mac": [
          "02:42:ac:11:00:25"
        ],
        "name": "ff634161e9c0"
      },
[...]

PR #8421, opened 2025-06-09
`TestFIPSAgentConnectingToFIPSFleetServerInECHFRH `: Use `require.Eventually` to try and address flakiness

ycombinator:
## What does this PR do?

This PR amends the `TestFIPSAgentConnectingToFIPSFleetServerInECHFRH` integration test to wrap the flaky assertion and it's pre-requisite code in a `require.Eventually` block.

## Why is it important?

To avoid flakiness in the test.

## Related issues

Closes #8419

PR #8422, opened 2025-06-09
Attempt to make test more reliable by querying ES directly

michel-laterman:
## What does this PR do?

Attempt to make the TestFIPSAgentConnectingToFIPSFleetServerInECHFRH test more reliable by querying ES for agents running the cloud policy instead of gathering agents from Kibana.

## Related issues

- Relates #8419


ycombinator:
I don't love that we are reaching into ES docs to determine whether the Agent is FIPS-capable. I would prefer to instead use the Kibana Fleet API and check its response to determine whether the Agent is FIPS-capable. I'm trying this out in https://github.com/elastic/elastic-agent/pull/8421 with a `require.Eventually` wrapped around the Kibana Fleet API call and test assertion.  That said, I'm happy to abandon my PR if the flakiness persists with it and go with this PR here if it's reliably not flaky.

swiatekm:
I like the solution in Shaunak's PR more as well. The FIPS agent should eventually be accessible via Kibana's API. Going around it to ES itself feels unnecessarily hacky.

michel-laterman:
Closing as the other solution seems to work well enough for the moment

michel-laterman:
As discussed in the FIPS sync, i've fixed this PR to work with the eventually loop so that we can eliminate calls to Kibana as it is a component that is only introduced for testing.

ycombinator (dismissed):
LGTM.  Let's run this in CI at least 10 times to make sure there's no flakiness.

michel-laterman:
buildkite test this

v1v:
I have unassigned my teams from the review. It seems that the issue is not related to the CI itself. Please let me know if we can assist with anything else. Thanka

michel-laterman:
buildkite test this

swiatekm (approved):
If this actually fixes the flakiness, I'd be interested to find out why going through Kibana introduces it in the first place.

michel-laterman:
buildkite test this

michel-laterman:
buildkite test this

michel-laterman:
buildkite test this

ycombinator (approved):

Issue #8381, opened 2025-06-06
When using the OTel runtime, status of non-otel inputs is hidden


cmacknz:
I think this likely a bug in https://github.com/elastic/elastic-agent/pull/8273

If you have an agent configuration where self-monitoring is configured to use the experimental otel runtime but contains other non-otel inputs, the status of the non-otel inputs is no longer reported.

```sh
❯ sudo elastic-development-agent version
Binary: 9.1.0-SNAPSHOT (build: 14a599750d1e8cccfd1666d34e6862a147ff7e59 at 2025-06-06 19:05:32 +0000 UTC)
Daemon: 9.1.0-SNAPSHOT (build: 14a599750d1e8cccfd1666d34e6862a147ff7e59 at 2025-06-06 19:05:32 +0000 UTC)

❯ sudo elastic-development-agent inspect
agent:
  logging:
    to_stderr: true
  monitoring:
    _runtime_experimental: otel
    enabled: true
    logs: true
    metrics: true
inputs:
- data_stream:
    namespace: default
  id: unique-system-metrics-input
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  - data_stream:
      dataset: system.memory
    metricsets:
    - memory
  - data_stream:
      dataset: system.network
    metricsets:
    - network
  - data_stream:
      dataset: system.filesystem
    metricsets:
    - filesystem
  type: system/metrics
  use_output: default
outputs:
  default:
    api_key: <REDACTED>
    hosts:
    - 127.0.0.1:9200
    preset: balanced
    type: elasticsearch
```

When running the status command with this configuration, the status of the `system/metrics` input is no longer reported:

```sh
❯ sudo elastic-development-agent status --output=full
┌─ fleet
│  └─ status: (STOPPED) Not enrolled into Fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: fd0baf0c-e7dd-479b-ae1b-057e92270e1c
   │  ├─ version: 9.1.0
   │  └─ commit: 14a599750d1e8cccfd1666d34e6862a147ff7e59
   ├─ beat/metrics-monitoring
   │  ├─ status: (HEALTHY) HEALTHY
   │  ├─ beat/metrics-monitoring
   │  │  ├─ status: (HEALTHY) Healthy
   │  │  └─ type: OUTPUT
   │  └─ beat/metrics-monitoring-metrics-monitoring-beats
[...]

PR #8435, opened 2025-06-10
Fix non-otel component status being incorrectly deleted

swiatekm:
## What does this PR do?

Otel component status processing needs to emit fake STOPPED events for components which don't exist anymore. This was incorrectly taking place for components running as beats processes, resulting in them incorrectly reporting a STOPPED state.

I've also made two minor refactoring changes:

* The coordinator's `watchRuntimeComponents` method now takes two channels as parameters instead of obtaining them from component fields. This makes it easier to test.
* The `translate.DropComponentStateFromOtelStatus` now no longer modifies its input, but instead returns a modified copy.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/8381

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator_unit_test.go
+++ b/internal/pkg/agent/application/coordinator/coordinator_unit_test.go
@@ -1478,7 +1478,9 @@ func TestCoordinatorTranslatesOtelStatusToComponentState(t *testing.T) {
 
 	select {
 	case finalOtelStatus := <-coord.managerChans.otelManagerUpdate:
-		// we shouldn't have any status remaining for the otel collector
+		// we shouldn't have any status remaining for the otel collector, as the status we've pushed earlier only
+		// contains beats receiver status for the "filestream-default" component
+		// this status is removed from the otel collector status, because it's reported as component state instead
 		assert.Empty(t, finalOtelStatus.ComponentStatusMap)
 	case <-ctx.Done():
 		t.Fatal("timeout waiting for coordinator to receive status")
--- a/internal/pkg/agent/application/coordinator/coordinator_unit_test.go
+++ b/internal/pkg/agent/application/coordinator/coordinator_unit_test.go
@@ -1336,47 +1336,48 @@ func TestCoordinatorTranslatesOtelStatusToComponentState(t *testing.T) {
 
 	// Set a one-second timeout -- nothing here should block, but if it
 	// does let's report a failure instead of timing out the test runner.
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
 	defer cancel()
 	logger := logp.NewLogger("testing")
 
 	statusChan := make(chan *status.AggregateStatus)
-	otelManager := &fakeOTelManager{
-		statusChan: statusChan,
-	}
 
-	componentModel := []component.Component{
-		{
-			ID:             "filestream-default",
-			InputType:      "filestream",
-			OutputType:     "elasticsearch",
-			RuntimeManager: component.OtelRuntimeManager,
-			InputSpec: &component.InputRuntimeSpec{
-				BinaryName: "agentbeat",
-				Spec: component.InputSpec{
-					Command: &component.CommandSpec{
-						Args: []string{"filebeat"},
-					},
-				},
-			},
-			Units: []component.Unit{
-				{
-					ID:   "filestream-unit",
-					Type: client.UnitTypeInput,
-					Config: &proto.UnitExpectedConfig{
-						Streams: []*proto.Stream{
-							{Id: "test-1"},
-							{Id: "test-2"},
-						},
-					},
-				},
-				{
-					ID:   "filestream-default",
-					Type: client.UnitTypeOutput,
+	runtimeStateChan := make(chan runtime.ComponentComponentState)
+
+	otelComponent := component.Component{
+		ID:             "filestream-default",
+		InputType:      "filestream",
+		OutputType:     "elasticsearch",
+		RuntimeManager: component.OtelRuntimeManager,
+		InputSpec: &component.InputRuntimeSpec{
+			BinaryName: "agentbeat",
+			Spec: component.InputSpec{
+				Command: &component.CommandSpec{
+					Args: []string{"filebeat"},
 				},
 			},
 		},
+		Units: []component.Unit{
+			{
+				ID:   "filestream-unit",
+				Type: client.UnitTypeInput,
+				Config: &proto.UnitExpectedConfig{
+					Streams: []*proto.Stream{
+						{Id: "test-1"},
+						{Id: "test-2"},
+					},
+				},
+			},
+			{
+				ID:   "filestream-default",
+				Type: client.UnitTypeOutput,
+			},
+		},
 	}
+	processComponent := otelComponent
+	processComponent.RuntimeManager = component.ProcessRuntimeManager
+	processComponent.ID = "filestream-process"
+
 	otelStatus := &status.AggregateStatus{
 		Event: componentstatus.NewEvent(componentstatus.StatusOK),
 		ComponentStatusMap: map[string]*status.AggregateStatus{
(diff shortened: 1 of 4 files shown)


internal/pkg/agent/application/coordinator/coordinator_unit_test.go
```diff
@@ -1336,47 +1336,48 @@ func TestCoordinatorTranslatesOtelStatusToComponentState(t *testing.T) {
 
 	// Set a one-second timeout -- nothing here should block, but if it
 	// does let's report a failure instead of timing out the test runner.
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
```
pchila: Why did this timeout change from 1 second to 1 hour ?

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
```diff
@@ -1446,8 +1447,55 @@ func TestCoordinatorTranslatesOtelStatusToComponentState(t *testing.T) {
...
+	// push the otel status into the coordinator
+	select {
+	case statusChan <- otelStatus:
+	case <-ctx.Done():
+		t.Fatal("timeout waiting for coordinator to receive status")
+	}
+
+	select {
+	case finalOtelStatus := <-coord.managerChans.otelManagerUpdate:
+		// we shouldn't have any status remaining for the otel collector
```
pchila: could you please expand about why we should not have any status remaining here ?

internal/pkg/otel/translate/status.go
```diff
@@ -56,26 +56,28 @@ func GetAllComponentStates(otelStatus *status.AggregateStatus, components []comp
...
 // given status.
-func DropComponentStateFromOtelStatus(otelStatus *status.AggregateStatus) error {
+func DropComponentStateFromOtelStatus(otelStatus *status.AggregateStatus) (*status.AggregateStatus, error) {
 	if otelStatus == nil {
-		return nil
+		return nil, nil
 	}
-	for pipelineStatusId := range otelStatus.ComponentStatusMap {
+
+	newStatus := deepCopyStatus(otelStatus)
```
pchila, line 64: what's the reason for deep copying the otel status ?

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
swiatekm: We're not adding any plain otel pipelines in this test, so we shouldn't have any status for them. We should only have states for components running either as beats receivers in the otel collector, or as beats processes.

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
swiatekm: Good catch, this is a debugging leftover. Will remove.

internal/pkg/otel/translate/status.go
swiatekm: It's safer for this function to be pure and not mutate its inputs. I had to debug a confusing test failure because of the previous behavior, and decided to fix it at the root. The status is relatively small so copying it won't have any noticable performance ramifications.

pkoutsovasilis (commented):
one thing that I am not super excited about this approach, and I have expressed it some time ago, is that it lives inside the coordinator 🙂 So given that @swiatekm I know you are already working on an approach that migrates this inside the otel manager would it make sense instead of reviewing and merging this one, to go directly with another PR of the other approach?

swiatekm:
> one thing that I am not super excited about this approach, and I have expressed it some time ago, is that it lives inside the coordinator 🙂 So given that @swiatekm I know you are already working on an approach that migrates this inside the otel manager would it make sense instead of reviewing and merging this one, to go directly with another PR of the other approach?

We could do that. I have a draft of that approach [here](https://github.com/elastic/elastic-agent/pull/8529) and it does fix this problem simply by moving the logic to its own component. But it's a pretty big PR so I'd prefer to keep the logical changes it does to a minimum. So I'd prefer to merge this one first, and then have the other one be a pure refactor. Does that make sense?

internal/pkg/otel/translate/status.go
pchila: I agree that modifying input parameters is bad practice and probable cause of strange errors.
However if there was a test failure related to modifying `otelStatus` it means that it's shared either in test or production code. Did you find the root cause of the test failure ?

swiatekm: In `TestCoordinatorTranslatesOtelStatusToComponentState`, the test case I added to in this PR, we repeatedly send the same status to the coordinator while modifying other parameters. I could have copied the status in the test itself, but I decided to make the function itself safer instead. As far as I'm aware this does not happen in production scenarios - we always get a new status from the otel collector.

pkoutsovasilis: @swiatekm I hear you about purity and mutation, and please do correct me if I am wrong, `logComponentStateChange` is doing just that right? it mutates the passed state and actually that was the reason for the bug which you fix in this PR by having two states, one for processes and for otel, is my understanding correct?

swiatekm: That's not correct. `logComponentStateChange` is its own separate world with its own state tracking for the sole purpose of emitting nicer log messages about state changes. The root cause of the bug was an additional bit of logic for otel component statuses, where we'd emit synthetic STOPPED states if we previously had a state for a component, but it later disappeared. The bug caused this bit of logic to also emit these states for components running as beats processes. It had nothing to do with modifying the plain otel pipeline status.

pkoutsovasilis: ok then I need to understand better; AFAICT `logComponentStateChange` seems to mutate the coordinatorState [here](https://github.com/elastic/elastic-agent/blob/8eef818c8e766cc7242167afbf606d78dfc272f5/internal/pkg/agent/application/coordinator/coordinator.go#L798) so there goes the pure and mutating principle. Now about the splitting up of the states, if you didn't split them up [here](https://github.com/elastic/elastic-agent/blob/8eef818c8e766cc7242167afbf606d78dfc272f5/internal/pkg/agent/application/coordinator/coordinator.go#L708) then I expect to happen what you describe as the bug in the PR description. What am I missing?

internal/pkg/agent/application/coordinator/coordinator_unit_test.go
pchila: IIUC beats receivers running in OTel collector pipelines are not reported back in the otelStatus but are only present in the "normal" processStatus ? If that's the case could you improve the comment above for non beats receivers specialists ?

swiatekm: Indeed, we try to make the beats receiver status as close as possible to beats processes. As we don't want to double-report the status, this means we remove them from the otel collector status. I'll add a comment here to make that clearer.

internal/pkg/otel/translate/status.go
swiatekm: The `watchRuntimeComponents` function has its own internal state, which is *not* actually the coordinator state in spite of poor parameter naming. Only this state is modified in it. Component state updates from it are then forwarded to the coordinator main loop, which updates the *actual* Coordinator state.

Before the otel runtime, the only reason for this internal state to exist was to provide nicer component state change logs - thanks to it, it can log what the previous state was. When I added support for statuses coming from the otel manager, I also used this internal state for a different purpose - emitting fake STOPPED states for components which disappear from the Otel status. The reason the bug exists, and why splitting the state fixes it, is that that new code would assume that any components missing from the Otel status are actually stopped - which is obviously not true for components running as beats processes!

Does that help?

pkoutsovasilis: Not fully but at this point I am leaning towards saying that I am covered by the following

> The reason the bug exists, and why splitting the state fixes it, is that that new code would assume that any components missing from the Otel status are actually stopped - which is obviously not true for components running as beats processes!

and I will try to raise any further questions I have in the draft PR you posted below. @pchila for me this is resolved, wdyt?

pchila: Sure, I was waiting for the explanation as well

pkoutsovasilis (approved):

PR #8449, opened 2025-06-11
Expose UpdateMarkerFile func with transactional lock

michalpristas:
Complementary function to work introduced in #8254 

Allows read-modify-write withing single transational lock


internal/pkg/agent/application/upgrade/marker_access_test.go
```diff
@@ -95,6 +100,37 @@ func TestWriteMarkerFileWithTruncation(t *testing.T) {
...
+		DesiredOutcome:    OUTCOME_UPGRADE,
+	}
+	tmp := t.TempDir()
+	markerFile := filepath.Join(tmp, "marker")
+	require.NoError(t, saveMarkerToPath(marker, markerFile, true, true))
+
+	// update marker
+	err := UpdateMarkerFile(markerFile, func(m *UpdateMarker) {
+		m.Version = "1.2.3-up"
+	})
```
pchila: Could we have 2 concurrent (i.e. different goroutines) partial updates on 2 different fields and verify that both updates apply cleanly?

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -277,11 +278,50 @@ func CleanMarker(log *logger.Logger, dataDirPath string) error {
...
+		return fmt.Errorf("locking update marker file %q for reading: %w", markerFile, err)
+	}
+	defer func(fileLock Locker) {
+		errUnlock := fileLock.Unlock()
+		if errUnlock != nil {
+			err = goerrors.Join(err, fmt.Errorf("unlocking marker file after reading: %w", errUnlock))
+		}
+	}(fileLock)
+
+	marker, err := loadMarker(markerFile, false)
```
pchila: Are we shadowing err here ? (thinking of the `errors.Join()` in line 296 mostly)

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -277,11 +278,50 @@ func CleanMarker(log *logger.Logger, dataDirPath string) error {
 // LoadMarker loads the update marker. If the file does not exist it returns nil
 // and no error.
 func LoadMarker(dataDirPath string) (*UpdateMarker, error) {
-	return loadMarker(markerFilePath(dataDirPath))
+	return loadMarker(markerFilePath(dataDirPath), true)
 }
 
-func loadMarker(markerFile string) (*UpdateMarker, error) {
-	markerBytes, err := readMarkerFile(markerFile)
+func UpdateMarkerFile(markerFile string, opts ...func(*UpdateMarker)) error {
```
pchila: More than options I think we are dealing with updater functions here
```suggestion
type markerUpdater func(*UpdateMarker)

func UpdateMarkerFile(markerFile string, updates ...markerUpdater) error {
```

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -277,11 +278,50 @@ func CleanMarker(log *logger.Logger, dataDirPath string) error {
...
+	}
+
+	if err := saveMarkerToPath(marker, markerFile, true, false); err != nil {
+		return fmt.Errorf("failed updating marker file: %w", err)
+	}
+
+	return nil
+}
+
+func loadMarker(markerFile string, isSafe bool) (*UpdateMarker, error) {
```
pchila: Is there a specific reason why somebody would want to load the update marker with an isSafe=false ? I am not seeing the usecase here..

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -335,8 +375,11 @@ func saveMarkerToPath(marker *UpdateMarker, markerFile string, shouldFsync bool)
 	if err != nil {
 		return err
 	}
+	if isSafe {
```
pchila: Same isSafe question as for the load. If we want to allow reading/writing the file concurrently with other updates, we can abstract the `Locker` and pass a different implementation that won't lock the lockfile

pchila (commented):
I think we are missing a testcase.
This change is also opening the possibility for caller of `(load|save)MarkerToPath` to read/write the update marker in an unsafe way with the `isSafe` parameters.

michalpristas:
@pchila  i resolved everything because i redid how this is executed

.gitignore
```diff
@@ -10,6 +10,8 @@
 **/html_docs
 data
 internal/pkg/agent/operation/tests/downloads
+pkg/core/process/testsignal/testsignal
+pkg/core/process/testsignal/testsignal.exe
```
pkoutsovasilis, line 14: nice 👍

internal/pkg/agent/application/upgrade/step_mark_test.go
```diff
@@ -25,6 +25,8 @@ import (
 	agtversion "github.com/elastic/elastic-agent/pkg/version"
 )
 
+var noopLocker = &dummyLocker{}
```
pkoutsovasilis: maybe I am missing something but this doesn't have to be a `dummyLocker` but it can be an actual `noopLocker`? 🙂

michalpristas: yes race condition, dummy was first

pkoutsovasilis (commented):
The code changes LGTM, @pchila please do have a look 🙂

pchila:
I would put this on hold until we have a clearer idea if we want to maintain locking for the upgrade marker file.
Should we decide to change the solution to a lock-free implementation this should not be merged and #8254 should be reverted

pchila:
Closing this as PR #8254  on which this was based has been reverted.

PR #8463, opened 2025-06-12
Short-circuit readMarkerFile if upgrade marker does not exist

pchila:
Add a quick check with os.Stat() on upgrade marker file before attempting to lock and read the marker file data. This would also prevent errors while trying to lock the marker file when the data directory does not exist yet.

## What does this PR do?

Avoids locking upgrade marker when the file doesn't exist.
Locking has been introduced by https://github.com/elastic/elastic-agent/pull/8254 but it stops agent from starting up if the `data` directory does not exist.

Benchbuilder run showing the problem: https://buildkite.com/elastic/elastic-agent-benchmark/builds/1901#019761fe-716d-44ea-87c3-b665b0162601

## Why is it important?

Trying to read a non-existing file should not stop agent from running.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/marker_access_test.go
+++ b/internal/pkg/agent/application/upgrade/marker_access_test.go
@@ -15,8 +15,57 @@ import (
 	"testing"
 
 	"github.com/fsnotify/fsnotify"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func TestReadNotExistingMarkerFile(t *testing.T) {
+	testcases := []struct {
+		name                string
+		setup               func(t *testing.T, tmpDir string) string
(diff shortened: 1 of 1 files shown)


alexsapran (commented):
LGTM from my side, doing follow up benchmarks using this change show it worked.
But I would expect someone from the team to chime in here as well.

pkoutsovasilis:
@pchila , to me this doesn't require a changelog, but maybe for you it does. Either way please handle that by either skipping the changelog or pushing one 🙂

pchila:
> @pchila , to me this doesn't require a changelog, but maybe for you it does. Either way please handle that by either skipping the changelog or pushing one 🙂

I feel the same, this change does not require an entry in changelog as it's completely transparent to users

PR #8481, opened 2025-06-12
Disable microsoft/go build telemetry

michel-laterman:
## What does this PR do?

Disable microsoft/go build telemetry in the spec and magefile.

### DIFF
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -2,6 +2,7 @@
 
 env:
   ASDF_MAGE_VERSION: 1.14.0
+  MS_GOTOOLCHAIN_TELEMETRY_ENABLED: "0"
 
   IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1749862860"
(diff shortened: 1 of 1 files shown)


simitt:
@pchila could you give this a review also please, as I believe you are the most familiar with the Elastic Agent build parts and it is important to always opt out.

ycombinator:
Shouldn't we also check for `MS_GOTOOLCHAIN_TELEMETRY_ENABLED` in here?

https://github.com/elastic/elastic-agent/blob/ac9ee9a34c541b68f3609195f34563042abfb3f4/dev-tools/packaging/testing/package_test.go#L804-L821

michel-laterman:
I don't think we need to check for the MS toolchain var in the binary; the var controls what the go binary does, not what we produce

ycombinator (approved):

ycombinator (approved):

michel-laterman:
I think that having it in the fips bk pipeline is much clearer then using a hook

v1v (dismissed):
LGTM, Could the environment variable be defined at the beginning of the `env` section? Instead of the step itself?

ycombinator (approved):

michel-laterman:
buildkite test this

michel-laterman:
buildkite test this

Issue #8486, opened 2025-06-13
[Flaky Test]: TestAgentMetricsInput - ignored field "event.ingested" does not exist in either map


swiatekm:
### Failing test case

TestAgentMetricsInput 

### Error message

ignored field "event.ingested" does not exist in either map, please remove it from the ignored fields

### Build

https://buildkite.com/elastic/elastic-agent/builds/22376#019768c6-ad95-41f5-92e0-db3d3f531a31

### OS

Linux

### Stacktrace and notes

We can probably just remove this check, it exists more to try to keep the test code clean rather than to verify any agent behavior. That said, it may be worth investigating why this attribute is sometimes missing.

cmacknz:
event.ingested is added by the Fleet final ingest pipeline https://github.com/elastic/kibana/blob/58b764eb4bc9554db1018cc5d9ba11cf46df2257/x-pack/platform/plugins/shared/fleet/server/constants/fleet_es_assets.ts#L202

If we aren't querying the event out of ES for comparison it won't be there. It would be very odd to me if it was inconsistently populated. I would expect it to be always present or always missing.

leehinman:
> If we aren't querying the event out of ES for comparison it won't be there. It would be very odd to me if it was inconsistently populated. I would expect it to be always present or always messing.

We pull the events from ES, so `event.ingested` should be there on every single document.

swiatekm:
So it really shouldn't be in the ignored list at all?

leehinman:
> So it really shouldn't be in the ignored list at all?

It's value should be ignored.  With 2 runs, we would expect the value to be different.  But we shouldn't be ignoring the presence of the field.

So I'm starting to think we have 2 types of fields to ignore.  One where we expect the value to be different, but the field should exist (event.ingested, \@timestamp, etc).  And another type where the presence or absence of the field isn't relevant (presence of metric that is time dependent).

mauri870:
> So I'm starting to think we have 2 types of fields to ignore. One where we expect the value to be different, but the field should exist. And another type where the presence or absence of the field isn't relevant (presence of metric that is time dependent).

That is true, I think E2E tests written more recently take this into account. 

https://github.com/elastic/elastic-agent/blob/b8882aa/testing/integration/otel_test.go#L1645-L1683

https://github.com/elastic/elastic-agent/blob/511197098b1a8da5d6610beef8ea88a7158b831f/testing/integration/beat_receivers_test.go#L591-L600

[...]

PR #8487, opened 2025-06-13
Don't fail otel tests if both documents don't contain an ignored field

swiatekm:
## What does this PR do?

Changes the logic of comparing documents produced by agent in non-otel and otel modes. Up until now, if neither document contained a field that should be ignored, the test would fail. After the change, the test proceeds with comparing the remaining fields.

## Why is it important?

There's some flakiness related to the ignored fields, which may sometimes not appear. This is largely unrelated to what these tests are actually checking, so disabling it temporarily shouldn't impact us negatively.

This should be re-enabled once we figure out the root cause of https://github.com/elastic/elastic-agent/issues/8486.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/8486

### DIFF
--- a/testing/integration/ess/otel_test.go
+++ b/testing/integration/ess/otel_test.go
@@ -1648,13 +1648,15 @@ func AssertMapsEqual(t *testing.T, m1, m2 mapstr.M, ignoredFields []string, msg 
 	flatM1 := m1.Flatten()
 	flatM2 := m2.Flatten()
 	for _, f := range ignoredFields {
-		hasKeyM1, _ := flatM1.HasKey(f)
-		hasKeyM2, _ := flatM2.HasKey(f)
-
-		if !hasKeyM1 && !hasKeyM2 {
-			assert.Failf(t, msg, "ignored field %q does not exist in either map, please remove it from the ignored fields", f)
-		}
-
+		// Checking ignored fields is disabled until we resolve an issue with event.ingested not being set
+		// in some cases.
+		// See https://github.com/elastic/elastic-agent/issues/8486 for details.
+		//hasKeyM1, _ := flatM1.HasKey(f)
+		//hasKeyM2, _ := flatM2.HasKey(f)
+		//
+		//if !hasKeyM1 && !hasKeyM2 {
+		//	assert.Failf(t, msg, "ignored field %q does not exist in either map, please remove it from the ignored fields", f)
+		//}
 		flatM1.Delete(f)
 		flatM2.Delete(f)
 	}


leehinman:
I wonder if we need 2 different kinds of ignored fields.  One where we expect the value to change, but the field should be present on both, and another where the field may or may not be present.

leehinman (approved):

leehinman:
Approved.  But lets open a new issue to try and figure out why the `TestAgentMetricsInput` runs into this but the `TestClassicAndReceiverAgentMonitoring` is not.

swiatekm:
I'm going to keep my original flaky test issue open, and track this problem there.

testing/integration/otel_test.go
```diff
@@ -1648,13 +1648,6 @@ func AssertMapsEqual(t *testing.T, m1, m2 mapstr.M, ignoredFields []string, msg
 	flatM1 := m1.Flatten()
 	flatM2 := m2.Flatten()
 	for _, f := range ignoredFields {
-		hasKeyM1, _ := flatM1.HasKey(f)
-		hasKeyM2, _ := flatM2.HasKey(f)
-
-		if !hasKeyM1 && !hasKeyM2 {
-			assert.Failf(t, msg, "ignored field %q does not exist in either map, please remove it from the ignored fields", f)
-		}
-
```
ycombinator: Sorry to add delay, but I don't have enough context on this topic so I have a couple of questions first: 1. What was the intent of this check? Was it to help keep the ignored fields list as small as possible?  2. What is the purpose of keeping an ignored fields list?

swiatekm: The intent was indeed to keep the list as small as possible. The list exists so we can meaningfully compare data produced by beats processes to data produced by beats receivers. We know, for example, that timestamps or metric values will be different.

testing/integration/ess/otel_test.go
```diff
@@ -1648,13 +1648,6 @@ func AssertMapsEqual(t *testing.T, m1, m2 mapstr.M, ignoredFields []string, msg
 	flatM1 := m1.Flatten()
 	flatM2 := m2.Flatten()
 	for _, f := range ignoredFields {
-		hasKeyM1, _ := flatM1.HasKey(f)
-		hasKeyM2, _ := flatM2.HasKey(f)
-
-		if !hasKeyM1 && !hasKeyM2 {
-			assert.Failf(t, msg, "ignored field %q does not exist in either map, please remove it from the ignored fields", f)
-		}
-
```
pkoutsovasilis, line 1657: I feel that this just disappearing might lead us to losing track of having to do that and the reasons behind it, so wdyt if we instead comment these out and put a link to the Flaky github issue that has the necessary context of why or how this is happening? 🙂

testing/integration/otel_test.go
swiatekm: Done.

pkoutsovasilis (approved):

pkoutsovasilis:
@swiatekm is this test also part of 9.1 branch? please do a check 🙂

swiatekm:
> @swiatekm is this test also part of 9.1 branch? please do a check 🙂

It's a helper function used in many tests, including some on 9.1.

PR #6144, opened 2024-11-25
Return actionable error message when enrolling

kaanyalti:
- Enhancement

## What does this PR do?

This PR adds checks to the enroll command to respond with an error message in case the user executing the command and the user that's the owner of the elastic program files don't match. Replaces #6038 based on the [following comment](https://github.com/elastic/elastic-agent/pull/6038#discussion_r1850253347) 

## Why is it important?

Currently there are no checks in place to prevent or correct the situation where a user who is not the owner of the program files executes enroll. This leads to a broken state.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test

## How to test this PR locally

**The testing steps are only for linux and mac**
- Deploy ess, install agent in unprivileged mode and enroll into fleet
- Unenroll the agent from the fleet ui
- Run the enroll command as root user and validate the error message
## Related issues

- Closes #4889

PR #8503, opened 2025-06-14
fix(7794): update Windows re-enrollment logic and rename test

kaanyalti:
Type of change:  
- Bug

## What does this PR do?

This PR updates the Windows enrollment logic in Elastic Agent, specifically addressing the overly strict file ownership checks that were introduced in a previous fix. Instead of removing the Windows-specific logic, this change implements the file ownership check as a no-op for Windows. The platform-specific files remain, but the check is now always allowed on Windows. This is a partial revert of the earlier logic, with the intent to restore expected behavior for privileged users on Windows.

## Why is it important?

The original bug (see [issue #4889](https://github.com/elastic/elastic-agent/issues/4889), fixed in [PR #6144](https://github.com/elastic/elastic-agent/pull/6144)) was that if the agent was installed unprivileged, running `enroll` as a privileged user would break the agent. The fix at the time was to disallow re-enrollment as root if the agent was installed unprivileged, using a strict file owner check. This worked for Unix, where impersonation is easy, but on Windows, due to how the `elastic-agent-user` is set up, impersonation is not feasible.

The strict SID check on Windows turned out to be too restrictive: even if the agent was installed as an admin, other admin users could not re-enroll, leading to the current issue ([#7794](https://github.com/elastic/elastic-agent/issues/7794)). The workaround (changing file ownership to SYSTEM) was not reliable.

The changes in this PR implement the ownership check as a no-op for Windows, allowing any admin to re-enroll. The platform-specific files are retained for clarity and maintainability. This change also reintroduces the original risk: if the agent is installed unprivileged, a privileged user can still break the agent by re-enrolling. The platform-specific files are retained for clarity and maintainability.

## Related issues

- Fixes [#7794: Unable to re-enrolled installed Elastic Agent on Windows](https://github.com/elastic/elastic-agent/issues/7794)
- Original bug and fix: [#4889](https://github.com/elastic/elastic-agent/issues/4889), [PR #6144](https://github.com/elastic/elastic-agent/pull/6144)

## Checklist

## Disruptive User Impact

- Windows users who previously could not re-enroll the agent as an admin should now be able to do so.
- The original risk (privileged re-enrollment breaking unprivileged installs) is reintroduced for Windows.

1. Build the agent on Windows.
2. Attempt to enroll, unenroll, and re-enroll the agent as both privileged and unprivileged users.
3. Privileged re-enroll should work. Unprivileged re-enroll should fail.
3. Run the renamed integration test (`re-enroll_test.go`) to verify correct behavior.

## Related issues

- Closes [#7794: Unable to re-enrolled installed Elastic Agent on Windows](https://github.com/elastic/elastic-agent/issues/7794) 
- Relates #8544

### DIFF
--- a/testing/integration/re-enroll_test.go
+++ b/testing/integration/re-enroll_test.go
@@ -32,6 +32,11 @@ type testCase struct {
 	assertion   AssertFunc
 }
 
+// TestReEnrollUnprivileged verifies that re-enrollment as a privileged user fails
+// when the agent was installed unprivileged. This enforces the file ownership check
+// on Unix platforms. On Windows, this check is a no-op as of PR #8503, so this test
+// is not run for windows. See the discussion in PR #8503 (https://github.com/elastic/elastic-agent/pull/8503)
+// and comment (https://github.com/elastic/elastic-agent/pull/8503#discussion_r2152603141) for context.
 func TestReEnrollUnprivileged(t *testing.T) {
 	info := define.Require(t, define.Requirements{
 		Group: Default,


testing/integration/re-enroll_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	out, err := fixture.Exec(ctx, enrollArgs)
+	require.Error(t, err)
+	require.Contains(t, string(out), cmd.UserOwnerMismatchError.Error())
+}
+
+func TestReEnrollUnprivileged(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Skipping test on Windows")
+	}
```
swiatekm: Shouldn't we set this in the requirements instead?

swiatekm (commented):
The fix and added test look good to me. I'd like some clarity on what the follow-up will be. Are we planning to address the Windows case, or are we resigned to leaving it as-is?

kaanyalti:
> The fix and added test look good to me. I'd like some clarity on what the follow-up will be. Are we planning to address the Windows case, or are we resigned to leaving it as-is?

In short yes, we are going to address the issue. The actual fix would be to enable root users to execute a re-enroll without breaking the agent. I'll create an issue and link it here.

kaanyalti:
Added the follow up bug ticket in the description https://github.com/elastic/elastic-agent/issues/8544

testing/integration/re-enroll_test.go
```diff
@@ -0,0 +1,97 @@
...
+	os          []define.OS
+	assertion   func(*testing.T, string, error)
+}
+
+func TestReEnrollUnprivileged(t *testing.T) {
+	info := define.Require(t, define.Requirements{
+		Group: Default,
+		Stack: &define.Stack{},
+		Sudo:  true,
+		OS: []define.OS{
```
cmacknz, line 45: Why can't we run this test on Windows?

testing/integration/re-enroll_test.go
```diff
@@ -0,0 +1,97 @@
...
+	policy, err := info.KibanaClient.CreatePolicy(ctx, policyReq)
+	require.NoError(t, err)
+
+	enrollmentApiKey, err := tools.CreateEnrollmentToken(t, ctx, info.KibanaClient, policy.ID)
+	require.NoError(t, err)
+
+	agentID, err := tools.InstallAgentForPolicyWithToken(ctx, t, installOpts, fixture, info.KibanaClient, enrollmentApiKey)
+	require.NoError(t, err)
+
+	_, err = info.KibanaClient.UnEnrollAgent(ctx, kibana.UnEnrollAgentRequest{ID: agentID})
```
cmacknz: I don't think this is strictly necessary, I think you can just run enroll twice (with the first time being at installation time).

testing/integration/re-enroll_test.go
```diff
@@ -0,0 +1,97 @@
...
+	require.NoError(t, err)
+
+	enrollUrl, err := fleettools.DefaultURL(ctx, info.KibanaClient)
+	require.NoError(t, err)
+
+	enrollArgs := []string{"enroll", "--url", enrollUrl, "--enrollment-token", enrollmentApiKey.APIKey, "--force"}
+
+	out, err := fixture.Exec(ctx, enrollArgs)
```
cmacknz: We should probably check that agent commands other than enroll can execute after this. Just ensuring agent is health via status command should be enough.

This will avoid the test allowing re-enrolling but the agent directory permissions still being broken in some way as part of the re-enroll actually happening.

testing/integration/re-enroll_test.go
cmacknz: It looks like it originally ran on Windows without issue in https://github.com/elastic/elastic-agent/pull/6144

cmacknz:
I tested this manually on a GCP VM and confirmed that install+enroll followed by enroll does not work for 9.0.2 but does with this PR.

internal/pkg/agent/cmd/enroll_match_fileowner_windows.go
```diff
@@ -6,87 +6,11 @@
 
 package cmd
 
-import (
-	"fmt"
+import "github.com/elastic/elastic-agent/internal/pkg/agent/errors"
 
-	"golang.org/x/sys/windows"
+var UserOwnerMismatchError = errors.New("the command is executed as root but the program files are not owned by the root user. execute the command as the user that owns the program files")
```
cmacknz: It seems odd that this error is in the windows specific file.

testing/integration/re-enroll_test.go
kaanyalti: The previous fix (PR #6144) enforced this test case on Windows, but it was too restrictive and blocked legitimate admin re-enrollment scenarios, which led to the current issue. This PR reverts that logic for Windows, so the unprivileged re-enroll test is no longer valid for that platform. Once there is a fix to enable root/admin to re-enroll, we can enable this for windows as well, though at that time the assertion would be different.

internal/pkg/agent/cmd/enroll_match_fileowner_windows.go
kaanyalti: Does not have to be, will move it

testing/integration/re-enroll_test.go
cmacknz: So the specific problem is that for this test, agent is the unprivileged elastic-agent-user but the test SSH user running the commands is the Buildkite user which has SYSTEM permissions?

Let's add this as a comment, and then file a docs issue in https://github.com/elastic/ingest-docs/issues to document this for the affected versions.

It sounds like to test this properly we need to be able to login in to the Windows test VM as the elastic-agent-user and we can't.

testing/integration/re-enroll_test.go
kaanyalti: Added status check

testing/integration/re-enroll_test.go
kaanyalti: When I check the user when triggering tests manually, I see that it is `OGC-WINDOWS-AMD\windows (SID: S-1-5-21-4096518103-1923715678-2754154236-1001)`, so not system. I don't know if it makes any difference in this case whether the user is system or an admin. I'll create an issue in docs for this.

testing/integration/re-enroll_test.go
```diff
@@ -0,0 +1,114 @@
...
+	_, err = tools.InstallAgentForPolicyWithToken(ctx, t, installOpts, fixture, info.KibanaClient, enrollmentApiKey)
+	require.NoError(t, err)
+
+	enrollUrl, err := fleettools.DefaultURL(ctx, info.KibanaClient)
+	require.NoError(t, err)
+
+	enrollArgs := []string{"enroll", "--url", enrollUrl, "--enrollment-token", enrollmentApiKey.APIKey, "--force"}
+
+	out, err := fixture.Exec(ctx, enrollArgs)
+	assertFunc(t, fixture, string(out), err)
```
cmacknz: `assertFunc` is the same for both cases, I think it would be clearer if you just had the status check in the test logic here instead of duplicating it (unless I'm missing some difference between the two calls they look very similar).

testing/integration/re-enroll_test.go
cmacknz: Interesting, trying to chase through the code which user executes the tests on windows is not simple.

I think you should comment explicitly why this test does not run on Windows. Something like "Skipped on Windows because you currently cannot execute the necessary commands as the elastic-agent-user directly". 

It does not jump out immediately that this test doesn't run on Windows and the comment will make it more obvious.

testing/integration/re-enroll_test.go
kaanyalti: Refactored test helper function to address this

testing/integration/re-enroll_test.go
kaanyalti: Created this docs issue https://github.com/elastic/ingest-docs/issues/1824

Issue #8208, opened 2025-05-21
Ensure that diagnostics contain all information necessary to debug beat receivers when using the OTel runtime


cmacknz:
- [x] The Beat logs should be included in the logs/elastic-agent-* log files, they will just be emitted by the collector instead. The component metadata should be the same as it was before. (https://github.com/elastic/elastic-agent/pull/8606)
- [x] The Beat state should appear in state.yaml, and degraded or failed Beat inputs must appear here. A system/metrics input in a Metricbeat receiver with a permissions error must continue to be visible.
- [x] The generated OTel collector configuration file must be present, likely in a new file. Existing files containing computed, expected, and actual agent configuration should account for the otel runtime mode.
  - Failure to generate the OTel collector configuration should be obvious and include the error.
- [x] Beat input metrics should continue to be captured and included.
- [x] The Filebeat registry and state should continue to be captured and included.
- [x] CPU and heap profiles should continue to be captured, just for fewer processes.
- [x] Consider including snapshots of OTel collector specific debug zpages https://github.com/open-telemetry/opentelemetry-specification/blob/main/development/trace/zpages.md

Changes to diagnostics should be reflected in our documentation and communicated to the agent-analyzer team:
- https://github.com/elastic/agent-analyzer
- https://github.com/elastic/ingest-docs/blob/main/docs/en/ingest-management/commands.asciidoc#elastic-agent-diagnostics-command

lucabelluccini:
> Changes to diagnostics should be reflected in our documentation and communicated to the agent-analyzer team

\+ Support please (keep us in the loop)

swiatekm:
Let's see where we are and what we need to do here.

> * [ ]  The Beat logs should be included in the logs/elastic-agent-* log files, they will just be emitted by the collector instead. The component metadata should be the same as it was before.

This is already the case, although it looks like we have some repeated fields. I'll see if I can normalize it. It would also be nice to have a test comparing these.

> * [x]  The Beat state should appear in state.yaml, and degraded or failed Beat inputs must appear here. A system/metrics input in a Metricbeat receiver with a permissions error must continue to be visible.

This is already the case insofar as beats receiver status appears in that file. The status does not include per-stream status at this time, but we should track that in a separate issue imo.

> * [ ]  The generated OTel collector configuration file must be present, likely in a new file. Existing files containing computed, expected, and actual agent configuration should account for the otel runtime mode.
>         * Failure to generate the OTel collector configuration should be obvious and include the error.

[...]

PR #8529, opened 2025-06-16
Add otel component manager to the coordinator

swiatekm:
## What does this PR do?

Adds a new manager object for running components (in the sense of the agent's Component Model) in an otel collector.

Doing this involves two basic activities:
* Translating agent configurations into beats receiver configurations for the otel collector.
* Translating otel collector statuses into component states.

Up until now, the logic for these was haphazardly spread across the agent coordinator. This PR moves all of it into a new object - the `OtelComponentManager` - which can run both raw otel collector configurations and agent components in a single otel collector instance.

This new manager encapsulates all the logic involved in interfacing between the agent coordinator and the otel collector. In the near future, it will also take on additional responsibilities, like generating diagnostics for components it runs.

The only new logic this PR introduces lives in the new manager's main loop, and has to do with how updates and configurations are moved around. The rest is either existing logic moved to a new location, and new tests for that old logic.

## Checklist

## Related issues

- Relates #8208

### DIFF
--- a/internal/pkg/agent/application/application.go
+++ b/internal/pkg/agent/application/application.go
@@ -227,6 +227,9 @@ func New(
 	}
 
 	otelManager, err := otelmanager.NewOTelManager(log.Named("otel_manager"), logLevel, baseLogger, otelmanager.EmbeddedExecutionMode)
+	if err != nil {
+		return nil, nil, nil, err
+	}
 	otelComponentManager := componentmanager.NewOtelComponentManager(log.Named("otel_component_manager"), otelManager, agentInfo, monitor.ComponentMonitoringConfig)
 	coord := coordinator.New(log, cfg, logLevel, agentInfo, specs, reexec, upgrader, runtime, configMgr, varsManager, caps, monitor, isManaged, otelComponentManager, actionAcker, compModifiers...)
 	if managed != nil {


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -134,15 +132,25 @@ type RuntimeManager interface {
...
+
+	// UpdateComponents updates the current components model.
+	UpdateComponents(model component.Model)
+
+	// WatchComponents returns a channel to watch for component state updates.
+	WatchComponents() <-chan runtime.ComponentComponentState
+
+	// MergedOtelConfig returns the merged Otel collector configuration, containing both the plain config and the
+	// component config.
+	MergedOtelConfig() *confmap.Conf
```
swiatekm, line 153: Note: This is temporary, and is only necessary for diagnostics. The plan is to move all the Otel-related diagnostic hooks into the OtelComponentManager. Afterwards, there won't be any reason for the coordinator to track this value.

internal/pkg/otel/componentmanager/manager.go
```diff
@@ -0,0 +1,367 @@
...
+
+		case otelStatus := <-m.otelManager.Watch():
+			componentUpdates, err := m.handleOtelStatusUpdate(otelStatus)
+			if err != nil {
+				m.reportError(ctx, err)
+			}
+			m.sendCollectorStatusUpdate(ctx)
+			m.sendComponentStateUpdates(ctx, componentUpdates)
+		}
+	}
```
pkoutsovasilis, line 147: the more I look at the newly introduced otel component manager the more I think that this shouldn't exist and every extra bit of logic here should become part of the existing otel manager. AFAICT every channel or other input you have here is already existing there and the transition of the logic seems straight-forward. I think that by migrating the logic there it will be easier for us to maintain/debug/extend the otel manager as we won't have another concurrent layer of channels to pass through, wdyt?

swiatekm: I think we could do that, yeah. The current implementation is nice in that it's only concerned with translating to and from the component model, and then the OtelManager itself is only concerned with managing the otel collector, with no dependencies on the component model. But you're right that this separation may not be worth the mental overhead of the additional asynchronous behavior it introduces.

To be honest, one of the main reasons I did it this way was because it let me not interfered with #8248. Would you be ok with merging this PR as-is, and moving all of the logic into the otel manager after the dust settles?

pkoutsovasilis: https://github.com/elastic/elastic-agent/pull/8248 is ready to be merged; there was a hiccup with unit-tests on windows runners yesterday but @cmacknz had approved it, so this is close 🙂 That said, how about avoiding this back and forth and starting with the transition of this code there?

swiatekm: I'll see how much of a lift it is and get back to you. If it's simple enough, I don't mind doing it now rather than later.

pkoutsovasilis: I believe that this has to go inside the otel manager either way. So yes do have a look but if it is too much to lift with https://github.com/elastic/elastic-agent/pull/8248 then this should go inside otel manager as it is now on `main` and https://github.com/elastic/elastic-agent/pull/8248 will be adjusted to accommodate for your changes.

swiatekm: I did what you suggested in https://github.com/elastic/elastic-agent/pull/8737. I'm not sure I feel better about it than this PR. OtelManager feels like it does too much with those changes imo. I don't have very strong feelings either way though, so please have a look and let me know what you think.

swiatekm:
> This change looks great to me code-wise, though I think I agree with @pkoutsovasilis that it would be nicer to add the component management to the existing otel manager as in #8737. (But control plane consensus should take precedence.)

That's a second vote for the approach in #8737, so I'm going to move this one into draft and put the other one up for review instead.

swiatekm:
#8737 was merged, closing this.

PR #8530, opened 2025-06-16
Add components YAML file with Core and Extended info

theletterf:
This adds a manual components.yaml file to map the support status of each EDOT Collector components: Core or Extended. This is necessary to properly render the automated table in the docs: https://www.elastic.co/docs/reference/opentelemetry/edot-collector/components

See: https://www.elastic.co/docs/reference/opentelemetry/compatibility/nomenclature#core-components

More context:

> The intent here is that this list ONLY contains OTel components that we explicitly categorized as core (i.e. with full support etc.). All other OTel components must not be in this file and are then assumed to be extended.
> 
> Promoting a component to core MUST be a very explicit decision that requires PM involvement. So, the process of this promotion step basically will be by updating this file.
> 
> Any other components that we add to EDOT (as extended components won't and shouldn't be added automatically to this file).
> 
> So, if we have a test that ensures that all components listed in this file are actually dependencies, I don't see a big risk of this file to get out of sync.

### DIFF
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -2,6 +2,7 @@
 
 env:
   ASDF_MAGE_VERSION: 1.14.0
+  MS_GOTOOLCHAIN_TELEMETRY_ENABLED: "0"
 
   IMAGE_UBUNTU_2404_X86_64: "platform-ingest-elastic-agent-ubuntu-2404-1749862860"
   IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-elastic-agent-ubuntu-2204-fips-1749862860"
--- a/dev-tools/packaging/packages.yml
+++ b/dev-tools/packaging/packages.yml
@@ -40,6 +40,7 @@ settings:
       cgo: true
       env:
         GOEXPERIMENT: systemcrypto
+        MS_GOTOOLCHAIN_TELEMETRY_ENABLED: "0"
       tags:
         - requirefips
         - ms_tls13kdf
--- a/magefile.go
+++ b/magefile.go
@@ -342,6 +342,7 @@ func (Build) WindowsArchiveRootBinary() error {
 		// there is no actual FIPS relevance for this particular binary
 		// but better safe than sorry
 		args.ExtraFlags = append(args.ExtraFlags, "-tags=requirefips,ms_tls13kdf")
+		args.Env["MS_GOTOOLCHAIN_TELEMETRY_ENABLED"] = "0"
 		args.CGO = true
 	}
 


internal/pkg/otel/components.yaml
```diff
@@ -0,0 +1,114 @@
...
+    class: Core
+  routingconnector:
+    class: Core
+  spanmetricsconnector:
+    class: Extended
+
+# Extensions
+
+  filestorage:
+    class: Unknown
```
AlexanderWert:
```suggestion
    class: Core
```

internal/pkg/otel/components.yaml
```diff
@@ -0,0 +1,114 @@
...
+  pprofextension:
+    class: Unknown
+
+# Providers
+
+  envprovider:
+    class: Unknown
+  fileprovider:
+    class: Unknown
```
AlexanderWert:
```suggestion
    class: Core
```

internal/pkg/otel/components.yaml
```diff
@@ -0,0 +1,114 @@
...
+    class: Core
+  spanmetricsconnector:
+    class: Extended
+
+# Extensions
+
+  filestorage:
+    class: Unknown
+  healthcheckextension:
+    class: Unknown
```
AlexanderWert:
```suggestion
    class: Extended
```

internal/pkg/otel/components.yaml
```diff
@@ -0,0 +1,114 @@
...
+  memorylimiterextension:
+    class: Unknown
+  pprofextension:
+    class: Unknown
+
+# Providers
+
+  envprovider:
+    class: Unknown
```
AlexanderWert:
```suggestion
    class: Core
```

swiatekm:
It'd be best if we could add a test that verifies all the components we ship in EDOT are present in this file. That doesn't need to happen in this PR if you're not up to it @theletterf, but then we should file a follow-up issue.

internal/pkg/otel/core-components.yaml
pchila, line 1: What is the usage for this file? Is it here only for documentation purposes?
How do we ensure that the components in this file and what is actually used in the binary are aligned ?

theletterf: It's only for docs purposes. Testing could be done through what @swiatekm suggests. The idea of using this and updating it manually (as it's always a curated list) comes from @AlexanderWert.

pchila: Manual updates is not always foolproof so I agree with @swiatekm that this needs testing. 
Who is supposed to check and keep the list manually up-to-date ?

theletterf: @AlexanderWert WDYT?

theletterf:
@swiatekm @AlexanderWert I've created a simple test that uses the same strategy and the current docs automation, that is, reading go.mod. Probably not as elegant as using Factories, but it works.

internal/pkg/otel/core-components.yaml
AlexanderWert: > Who is supposed to check and keep the list manually up-to-date ?

The intent here is that this list ONLY contains OTel components that we explicitly categorized as `core` (i.e. with full support etc.). All other OTel components must not be in this file and are then assumed to be `extended`. 

Promoting a component to core MUST be a very explicit decision that requires PM involvement. So, the process of this promotion step basically will be by updating this file.

Any other components that we add to EDOT (as `extended` components won't and shouldn't be added automatically to this file).

So, if we have a test that ensures that all components listed in this file are actually dependencies, I don't see a big risk of this file to get out of sync.

internal/pkg/otel/core_components_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+// contains checks if a string is present in a slice
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
```
pchila: This can be removed if  [`assert.Contains`](https://pkg.go.dev/github.com/stretchr/testify/assert#Contains) is used

internal/pkg/otel/core_components_test.go
```diff
@@ -0,0 +1,168 @@
...
+	var missingComponents []string
+	for _, component := range yamlComponents {
+		if !contains(goModComponents, component) {
+			missingComponents = append(missingComponents, component)
+		}
+	}
+
+	if len(missingComponents) > 0 {
+		t.Errorf("The following components from core-components.yaml are missing in go.mod: %v", missingComponents)
+	}
```
pchila: This can be simplified by using the correct assertions.
```suggestion
	for _, component := range yamlComponents {
		assert.Contains(t, goModComponents, component)
	}
```

internal/pkg/otel/core_components_test.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return components, nil
+}
```
pchila: This code can be simplified by a `go list -json github.com/elastic/elastic-agent` (you should be able to omit the package if you want) call, parsing the json and then accessing the `Deps` attribute to filter the list of dependencies.
Moreover if the purpose of this test is only to check that all the entries in core component yaml are listed as dependencies containing `processor|receiver|extension|exporter|connector|confmap/provider|storage` is it really needed to filter go.mod dependencies ?

theletterf: I've simplified the test, PTAL.

internal/pkg/otel/core_components_test.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// TestCoreComponentsInGoMod verifies that all components listed in core-components.yaml
+// are present in the project dependencies.
+func TestCoreComponentsInGoMod(t *testing.T) {
+	// Extract components from project dependencies
+	moduleComponents, err := extractComponentsFromDeps()
+	if err != nil {
+		t.Fatalf("Failed to extract components from dependencies: %v", err)
+	}
```
pchila:
```suggestion
	require.NoError(t, err, "Failed to extract components from dependencies")
```

internal/pkg/otel/core_components_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Print found components for debugging
+	t.Logf("Found components in dependencies: %v", moduleComponents)
+
+	// Load components from core-components.yaml
+	yamlComponents, err := loadCoreComponentsYAML()
+	if err != nil {
+		t.Fatalf("Failed to load core-components.yaml: %v", err)
+	}
```
pchila:
```suggestion
	require.NoError(t, err, "Failed to load core-components.yaml)
```

internal/pkg/otel/core_components_test.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		// Special case for filestorage which is under extension/storage/
+		if strings.Contains(dep, "/extension/storage/") {
+			parts := strings.Split(dep, "/")
+			if len(parts) > 0 {
+				componentName := parts[len(parts)-1]
+				components = append(components, componentName)
+			}
+		}
```
pchila, line 96: Not sure why `/extension/storage/` needs a specific treatment here while `confmap/provider` doesn't :thinking:

pchila (commented):
Just a couple of suggestions for asserting the absence of errors (non-blocking)
As a more general question: @swiatekm was talking about a test that would ensure that all the components we ship in EDOT would appear in the file, while the test checks that go.mod contains the component listed in the file (the inclusion is reversed). Which condition we want to maintain here ?

theletterf:
@pchila See https://github.com/elastic/elastic-agent/pull/8530#discussion_r2151456054 for the rationale on why we're testing things this way.

internal/pkg/otel/core_components_test.go
theletterf: Providers (there are severals) are already handled in line 70. This extension, for some reason, is the only one carrying an extra path segment. 🤷🏻

theletterf:
@pchila Fine if we skip CI / tests for this one?

pchila:
> @pchila Fine if we skip CI / tests for this one?

I am afraid we have to wait for the CI to finish after the last merge :sweat:

Issue #8319, opened 2025-06-04
[Flaky Test]: TestKubernetesAgentStandaloneKustomize – failed to render kustomize


pkoutsovasilis:
### Failing test case

TestKubernetesAgentStandaloneKustomize

### Error message

failed to render kustomize

### Build

https://buildkite.com/elastic/elastic-agent/builds/21747#019736a5-8d7a-4be2-a9bf-b8520c731474/503-564

### OS

Linux

### Stacktrace and notes

```markdown
=== Failed
=== FAIL: testing/integration TestKubernetesAgentStandaloneKustomize/drop_ALL_capabilities_-_rootful_agent (32.84s)
    kubernetes_agent_standalone_test.go:1114:
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1748968918258658670/elastic/elastic-agent/testing/integration/kubernetes_agent_standalone_test.go:1114
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1748968918258658670/elastic/elastic-agent/testing/integration/kubernetes_agent_standalone_test.go:167
        	Error:      	Received unexpected error:
        	            	accumulating resources: accumulating resources from 'https://github.com/kubernetes/kube-state-metrics/': MalformedYAMLError: yaml: line 98: mapping values are not allowed in this context in File: https://github.com/kubernetes/kube-state-metrics/
        	Test:       	TestKubernetesAgentStandaloneKustomize/drop_ALL_capabilities_-_rootful_agent
        	Messages:   	failed to render kustomize
    --- FAIL: TestKubernetesAgentStandaloneKustomize/drop_ALL_capabilities_-_rootful_agent (32.84s)
```

pkoutsovasilis:
another occurrence [here](https://buildkite.com/elastic/elastic-agent/builds/21737#0197363e-af5e-4b62-9d22-3b7cd5b108ea/240-301)

pkoutsovasilis:
another occurence [here](https://buildkite.com/elastic/elastic-agent/builds/21738#01973644-8b0f-4280-b6cc-8e58da050f21/469-514)

pkoutsovasilis:
another occurrence [here](https://buildkite.com/elastic/elastic-agent/builds/21729#01973601-abce-45cc-90ae-bf7c5f8ff3f2/389-437)

pkoutsovasilis:
[...]

PR #8575, opened 2025-06-18
[ci] fix k8s integration tests flakiness

pkoutsovasilis:
## What does this PR do?

This PR introduces the following changes:
1. **Vendor external Kubernetes artifacts for testing**:
   - Kube-state-metrics (KSM) Helm subchart is now vendored to eliminate remote fetches during k8s integration tests.
   - Rendered manifests from the `kustomize` configuration are vendored to remove the runtime dependency on `kustomize`.
   - The kube-stack Helm chart is also vendored to avoid network pulls.

   To support this, a new `mage` target called `integration:buildKubernetesTestData` was added. This target is now invoked as a prerequisite by:
   - `integration:testKubernetes`
   - `integration:testKubernetesMatrix`
   - `integration:testKubernetesSingle`

   This aims to prevent CI failures from GitHub/network issues and addresses [#8319](https://github.com/elastic/elastic-agent/issues/8319).

2. **Fix decoding of Beats-style API Keys in K8s tests**:
   - [PR #7754](https://github.com/elastic/elastic-agent/pull/7754) introduced a regression that broke Beats-style API key generation required by the Helm and Kustomize k8s integration tests.
   - This PR patches the decoding logic, and highlights the need to validate event ingestion in all integration tests.

3. **Fix `%CA_TRUSTED%` environment variable injection in Kustomize tests**:
   - Recent failures revealed malformed `ca_trusted_fingerprint` values caused by `%CA_TRUSTED%` placeholders not being overridden.
   - This PR sets `CA_TRUSTED` to an empty value in the relevant test environments to ensure expected TLS behavior.
   
    ```
    {"log.level":"info","@timestamp":"2025-06-18T02:40:53.068Z","message":"'ca_trusted_fingerprint' set, looking for matching fingerprints","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"filestream-default","type":"filestream"},"log":{"source":"filestream-default"},"log.logger":"tls","log.origin":{"file.line":180,"file.name":"tlscommon/tls_config.go","function":"github.com/elastic/elastic-agent-libs/transport/tlscommon.trustRootCA"},"service.name":"filebeat","ecs.version":"1.6.0","ecs.version":"1.6.0"}
    {"log.level":"error","@timestamp":"2025-06-18T02:40:53.068Z","message":"Error dialing decode 'ca_trusted_fingerprint': encoding/hex: invalid byte: U+0025 '%'","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"filestream-default","type":"filestream"},"log":{"source":"filestream-default"},"ecs.version":"1.6.0","log.logger":"esclientleg","log.origin":{"file.line":39,"file.name":"transport/logging.go","function":"github.com/elastic/elastic-agent-libs/transport/httpcommon.(*HTTPTransportSettings).RoundTripper.LoggingDialer.func2"},"service.name":"filebeat","network.transport":"tcp","server.address":"767b2b9229dd4bd098dab11b95e74c64.us-west2.gcp.elastic-cloud.com:443","ecs.version":"1.6.0"}
    ```

4. **Increase memory limits for Elastic Agent in Kustomize-based tests**:
   - Multiple `OOMKilled` errors were observed ([example 1](https://buildkite.com/elastic/elastic-agent/builds/22661#019780c5-1c7f-4c8b-b03d-9e25a64836f4/526-572), [example 2](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/9040#01977fdf-ffc0-46e9-ac71-869a0b8c6fc7/479-529)).
   - A bump in memory limits is applied specifically for the Kustomize scenario, where all inputs run on a single DaemonSet pod, potentially spawning more Beat subprocesses.

## Why is it important?

- Ensures Kubernetes integration tests are resilient to GitHub outages and network instability by removing external dependencies.
- Fixes broken functionality introduced in previous PRs related to API key handling.
- Prevents TLS misconfiguration caused by unintended environment variable injections.
- Addresses out-of-memory crashes that reduce test reliability and increase CI flakiness.

## Checklist

## Disruptive User Impact

None expected. All changes are isolated to the Kubernetes integration test setup and do not affect runtime or user configurations.

```
mage integration:buildKubernetesTestData

EXTERNAL=true SNAPSHOT=true PACKAGES=docker DOCKER_VARIANTS=basic PLATFORMS=linux/arm64 mage package

INSTANCE_PROVISIONER="kind" TEST_PLATFORMS="kubernetes/arm64/1.33.0/basic" mage integration:TestKubernetes
```

## Related issues

- Closes [#8319](https://github.com/elastic/elastic-agent/issues/8319)


magefile.go
```diff
@@ -2372,6 +2376,32 @@ func (Integration) TestKubernetesMatrix(ctx context.Context) error {
...
+func (Integration) BuildKubernetesTestData(ctx context.Context) error {
+	// build the dependencies for the elastic-agent helm chart
+	mg.Deps(Helm.BuildDependencies)
+
+	// download opentelemetry-kube-stack helm chart
+	kubeStackHelmChartPath, err := devtools.DownloadFile(k8s.KubeStackChartURL, filepath.Join("testing", "integration", "k8s", "testdata"))
+	if err != nil {
+		return fmt.Errorf("failed to download opentelemetry-kube-stack helm chart: %w", err)
+	}
+	if err := os.Rename(kubeStackHelmChartPath, filepath.Join("testing", "integration", "k8s", k8s.KubeStackChartPath)); err != nil {
```
pchila: Why is this os.Rename needed? From the code it seems to me like the `DownloadFile` function above should create the file with the correct name in the correct position :thinking:

testing/integration/k8s/k8s.go
```diff
@@ -0,0 +1,20 @@
...
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package k8s
+
+import "path/filepath"
+
+const (
+	AgentKustomizePath = "./testdata/elastic-agent-kustomize.yaml"
+	AgentHelmChartPath = "../../../deploy/helm/elastic-agent"
```
pchila: maybe we can try to be consistent in using `filepath` package with relative paths like it's done for `KubeStackChartPath` ?

pchila (approved):
One question and a small nitpick on some test path (both are non-blocking)
LGTM otherwise

testing/integration/k8s/k8s.go
pkoutsovasilis: nice catch pushed 99cdb1a97c16a90603ca79bacead694a970ac618 🙂

magefile.go
pkoutsovasilis: yep you are right, I guess it's just me, but I would like the `devtools.DownloadFile` to accept a targetPath and not a targetDir to feel "safer" 😄  pushed this here 1196468261caa53561a80512368327bd58e3753d

pchila (approved):

pchila:
After some testing with @pkoutsovasilis , it seems that we can vendor our helm dependencies in uncompressed form: this means that instead of including [deploy/helm/elastic-agent/charts/kube-state-metrics-5.30.1.tgz](https://github.com/elastic/elastic-agent/pull/8575/files#diff-f814a6171a6c21f95d36a4160cd3c2668fc580aa27186526a26bb3aedcfac8cb) we can include the uncompressed `deploy/helm/elastic-agent/charts/kube-state-metrics` directory (the same is valid for the [testing/integration/k8s/testdata/opentelemetry-kube-stack-0.3.9.tgz helm chart](https://github.com/elastic/elastic-agent/pull/8575/files#diff-2b98de4430ceeb2b8e3349f6d9db7172ece8dd6bad9069c1631ac842e742c5f6) in `testdata`.

This has the benefit of not having to include a binary file in our git changes and we can more clearly see what changes when chart version gets bumped.

@pkoutsovasilis could you please add a commit with such a change ?

pchila (approved):

pchila (approved):
Even if it increases the number of new files, I prefer the exploded charts to having a .tgz committed in git.
It's a shame for the lint GH action not to support diffs over 20k lines for PRs so it cannot filter the linter violations only to the modified files but this should not be a recurring problem.

magefile.go
```diff
@@ -3829,9 +3872,147 @@ func updateYamlFile(path string, keyVal ...struct {
...
+
 // BuildDependencies builds the dependencies for the Elastic-Agent Helm chart.
-func (Helm) BuildDependencies() error {
-	return helm.BuildChartDependencies(helmChartPath)
+func (h Helm) BuildDependencies() error {
+	settings := cli.New()
+	settings.SetNamespace("")
+	actionConfig := &action.Configuration{}
+
+	chartFile, err := os.ReadFile(fmt.Sprintf("%s/Chart.yaml", helmChartPath))
```
swiatekm: nit:

```suggestion
	chartFile, err := os.ReadFile(filepath.Join(helmChartPath, "Chart.yaml"))
```

magefile.go
```diff
@@ -3829,9 +3872,147 @@ func updateYamlFile(path string, keyVal ...struct {
...
+		return fmt.Errorf("could not write Helm repository config: %w", err)
+	}
+
+	return nil
+}
+
 // BuildDependencies builds the dependencies for the Elastic-Agent Helm chart.
-func (Helm) BuildDependencies() error {
-	return helm.BuildChartDependencies(helmChartPath)
+func (h Helm) BuildDependencies() error {
```
swiatekm, line 3930: How is the implementation here different from calling `helm dependency update`? Worth adding a comment explaining why we're doing something custom.

swiatekm (approved):
LGTM. It's unfortunate that we need to vendor all this code, and in an ideal world I'd prefer if these artifacts were cached locally on our CI runners instead. But this should work and probably won't be too burdensome to maintain.

Left some questions and nitpicks that shouldn't block merging.

magefile.go
pkoutsovasilis: Great question. Initially we had to write our own logic of `helm dependency update` since the latter would fail if you didn't had the dependency repos inside your helm by running beforehand `helm repo add ...`. We went for more convenience here and the `mage helm:buildDependencies` adds all the repos before essentially performing the dependency update. Now, in the context of this PR, there is another inconvenience,  `helm dependency update` will always download the subcharts as `.tgz` archives and doesn't decompress them. Thus, for this PR I extended the `mage helm:buildDependencies` to do that as well. I will add a comment to make this more clear 🙂

magefile.go
pkoutsovasilis: pushed here https://github.com/elastic/elastic-agent/pull/8575/commits/4c59037b9feeb633aa06fad8b02ead675ab08190 🙂

magefile.go
pkoutsovasilis: updated godoc pushed here https://github.com/elastic/elastic-agent/pull/8575/commits/963742ee38fd304b213e65db6c8db2cf7ab6eff5 🙂

swiatekm (approved):

PR #8594, opened 2025-06-19
tools: use docker compose and quiet for pulling

v1v:
## What does this PR do?

Use `docker compose`
Use `--pull --quiet`

## Why is it important?

Docker-compose reached EoL last year or the year before.
Avoid too much verbose output in the logs when pulling docker images.

## Checklist

## Related issues

- Relates to https://github.com/elastic/beats/pull/44324

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/common.go
+++ b/dev-tools/mage/common.go
@@ -209,16 +209,6 @@ func dockerInfo() (*DockerInfo, error) {
 	return &info, nil
 }
 
-// HaveDockerCompose returns an error if docker-compose is not found on the
-// PATH.
-func HaveDockerCompose() error {
-	_, err := exec.LookPath("docker-compose")
-	if err != nil {
-		return fmt.Errorf("docker-compose is not available")
-	}
-	return nil
-}
-
 // HaveKubectl returns an error if kind is not found on the PATH.
 func HaveKubectl() error {
 	_, err := exec.LookPath("kubectl")
--- a/dev-tools/mage/integtest_docker.go
+++ b/dev-tools/mage/integtest_docker.go
@@ -50,12 +50,9 @@ func (d *DockerIntegrationTester) Use(dir string) (bool, error) {
 	return false, nil
 }
 
-// HasRequirements ensures that the required docker and docker-compose are installed.
+// HasRequirements ensures that the required docker and docker compose are installed.
 func (d *DockerIntegrationTester) HasRequirements() error {
 	if err := HaveDocker(); err != nil {
-		return err
-	}
-	if err := HaveDockerCompose(); err != nil {
 		return err
 	}
 	return nil
@@ -66,7 +63,7 @@ func (d *DockerIntegrationTester) StepRequirements() IntegrationTestSteps {
 	return IntegrationTestSteps{&IntegrationTestStep{}}
 }
 
-// Test performs the tests with docker-compose.
+// Test performs the tests with docker compose.
 func (d *DockerIntegrationTester) Test(dir string, mageTarget string, env map[string]string) error {
 	var err error
 	d.buildImagesOnce.Do(func() { err = dockerComposeBuildImages() })
@@ -85,8 +82,8 @@ func (d *DockerIntegrationTester) Test(dir string, mageTarget string, env map[st
 	goPkgCache := filepath.Join(filepath.SplitList(build.Default.GOPATH)[0], "pkg/mod/cache/download")
 	dockerGoPkgCache := "/gocache"
 
-	// Execute the inside of docker-compose.
-	args := []string{"-p", dockerComposeProjectName(), "run",
+	// Execute the inside of docker compose.
+	args := []string{"compose", "-p", dockerComposeProjectName(), "run",
 		"-e", "DOCKER_COMPOSE_PROJECT_NAME=" + dockerComposeProjectName(),
 		// Disable strict.perms because we mount host dirs inside containers
 		// and the UID/GID won't meet the strict requirements.
@@ -125,13 +122,13 @@ func (d *DockerIntegrationTester) Test(dir string, mageTarget string, env map[st
 		composeEnv,
 		os.Stdout,
 		os.Stderr,
-		"docker-compose",
+		"docker",
 		args...,
 	)
 
 	err = saveDockerComposeLogs(dir, mageTarget, composeEnv)
 	if err != nil && testErr == nil {
-		// saving docker-compose logs failed but the test didn't.
+		// saving docker compose logs failed but the test didn't.
 		return err
 	}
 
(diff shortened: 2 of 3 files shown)


dev-tools/mage/common.go
```diff
@@ -209,16 +209,6 @@ func dockerInfo() (*DockerInfo, error) {
 	return &info, nil
 }
 
-// HaveDockerCompose returns an error if docker-compose is not found on the
```
v1v, line 212: Assuming `docker compose` is available, so by validating `docker` I think we can remove this condition. It's not available, it will fail regardless.

dev-tools/mage/integtest_docker.go
```diff
@@ -50,14 +50,11 @@ func (d *DockerIntegrationTester) Use(dir string) (bool, error) {
 	return false, nil
 }
 
-// HasRequirements ensures that the required docker and docker-compose are installed.
+// HasRequirements ensures that the required docker and docker compose are installed.
```
v1v:
```suggestion
// HasRequirements ensures that the required docker is installed.
```

pchila:
@v1v 
`docker-compose` is not really used in CI and we can use `docker compose` when we need it so thank you for removing references to obsolete tools.

I am not so sure about reducing the docker logs though, since those have been useful to debug build/packaging issues in the past; removing part of those logs means that in order to investigate issues, those must be reproduced and analyzed off buildkite or with a dedicated PR. 
What is the impact of those docker logs that we are trying to mitigate/remove?

dliappis (commented):
I am skeptical about the use of `--quiet`; while I understand the motivation to reduce verbosity related to the individual layers, by making all stdout quiet we are losing previous information about the docker image (e.g. golang crossbuild) that got pulled, thus making troubleshooting unnecessarily harder.

v1v:
My concern is that those details are too verbose and unstructured; however, I think using `--progress=plain` mitigates the former.

I understand your view, probably, we could say if a retry then enable the prior approach, otherwise use `--quiet` by default.  Then, we know it's easy to debug.

However, maybe I'm overthinking, I feel you prefer to do a step back and only make the changes for:
- `docker compose --progress=plain`
- `docker build --progres=plain`

That's something I took from https://github.com/elastic/golang-crossbuild/pull/528

If no objections, I'll remove the `--quiet`

dev-tools/mage/crossbuild.go
```diff
@@ -294,7 +294,7 @@ func (b GolangCrossBuilder) Build() error {
 	// container. So we ensure that all path separators are `/`.
 	buildCmd = filepath.ToSlash(buildCmd)
 
-	dockerRun := sh.RunCmd("docker", "run")
+	dockerRun := sh.RunCmd("docker", "run", "--quiet")
```
v1v, line 297: According to the docs:

```bash
$ docker run --help
...
      --pull string                      Pull image before running ("always", "missing", "never") (default "missing")
      -q, --quiet                            Suppress the pull output
```

See https://docs.docker.com/reference/cli/docker/container/run/

image

dev-tools/mage/integtest_docker.go
```diff
@@ -255,13 +254,13 @@ func dockerComposeBuildImages() error {
...
 
-	args := []string{"-p", dockerComposeProjectName(), "build", "--force-rm"}
+	args := []string{"compose", "-p", dockerComposeProjectName(), "--progress=plain", "build", "--force-rm"}
 	if _, noCache := os.LookupEnv("DOCKER_NOCACHE"); noCache {
 		args = append(args, "--no-cache")
 	}
 
 	if _, forcePull := os.LookupEnv("DOCKER_PULL"); forcePull {
-		args = append(args, "--pull")
+		args = append(args, "--pull", "--quiet")
```
v1v:
```suggestion
		args = append(args, "--pull")
```

dev-tools/mage/pkgtypes.go
```diff
@@ -785,7 +785,7 @@ func runFPM(spec PackageSpec, packageType PackageType) error {
 	}
 	spec.OutputFile = packageType.AddFileExtension(filepath.Join(distributionsDir, outputFile))
 
-	dockerRun := sh.RunCmd("docker", "run")
+	dockerRun := sh.RunCmd("docker", "run", "--quiet")
```
v1v, line 788: See https://github.com/elastic/elastic-agent/pull/8594/files#r2158465644

dev-tools/mage/integtest_docker.go
```diff
@@ -85,8 +82,8 @@ func (d *DockerIntegrationTester) Test(dir string, mageTarget string, env map[st
 	goPkgCache := filepath.Join(filepath.SplitList(build.Default.GOPATH)[0], "pkg/mod/cache/download")
 	dockerGoPkgCache := "/gocache"
 
-	// Execute the inside of docker-compose.
-	args := []string{"-p", dockerComposeProjectName(), "run",
+	// Execute the inside of docker compose.
+	args := []string{"compose", "-p", dockerComposeProjectName(), "run",
```
v1v:
```suggestion
	args := []string{"compose", "-p", dockerComposeProjectName(), "--progress=plain", "run",
```

pkoutsovasilis:
> My concern is that those details are too verbose and unstructured; however, I think using `--progress=plain` mitigates the former.
> 

`--progress=plain` makes sense and most probably makes the output more readable which is something good IMO

> I understand your view, probably, we could say if a retry then enable the prior approach, otherwise use `--quiet` by default. Then, we know it's easy to debug.

Again sorry but I don't agree, here you seem to care only for failures, but we would also like to be in the position of looking the docker build for green steps. 

> However, maybe I'm overthinking, I feel you prefer to do a step back and only make the changes for:
> 
> * `docker compose --progress=plain`
> * `docker build --progres=plain`
> 
> That's something I took from [elastic/golang-crossbuild#528](https://github.com/elastic/golang-crossbuild/pull/528)

Again to understand the purpose of this PR, I am gonna echo the same question of @pchila; What is the impact of those docker logs that we are trying to mitigate/remove? 🙂

v1v:
> What is the impact of those docker logs that we are trying to mitigate/remove? 

It's about ergonomics and usability. Seeing logs regarding an interactive shell is certainly misleading and therefore makes the UI experience really bad - people need to download the log as it's massive.

For instance, see https://buildkite.com/elastic/elastic-agent-extended-testing/builds/9082#01978c8c-60aa-4e4a-a0a9-1c61f59e7329/6-7

Then, if you click on open https://buildkite.com/organizations/elastic/pipelines/elastic-agent-extended-testing/builds/9082/jobs/01978c8c-60aa-4e4a-a0a9-1c61f59e7329/log

image


Maybe, it's me, but having a UI that cannot show relevant details because logs are massive sounds like something to improve. Hence this PR. However, I don't use this CI massively. Probably I'm biased here.

If you think the changes are not relevant, I'm happy to change the scope and focus only on the `docker
 compose` if agreed, or close this one.

michel-laterman:
I think using `--progress plain` will help with readability on buildkite, even though it may not dramatically reduce log volume

ebeahan:
@v1v checking in if we should leave this one open or close until you have time to come back. Thanks!

v1v:
I think the team can consider what to do in the future, I prefer to close this now

PR #8606, opened 2025-06-20
Ensure beats receivers log the same metadata as beats processes

swiatekm:
## What does this PR do?

It ensures logs from beats receivers forwarded by elastic-agent have the same metadata as ones from beats processes. It's worth noting they aren't identical, as beats receiver logs currently have the following additional attributes:

- `otelcol.component.id`
- `'otelcol.component.kind`
- `otelcol.signal`
- `resource`

Whether these should exist as they are is a separate issue. This PR is only concerned with the backwards compatibility of beats receiver logs relative to beats process logs.

The only semantic change in this PR is passing the base logger to the Otel Collector instead of the elastic-agent logger. This is consistent with what we do for beats processes.

I've also added a test that compares a specifically chosen log line between beats receivers and processes without sending it to Elasticsearch. My intent was to verify the metadata before it goes through any processing pipelines, be they local or remote.

## Why is it important?

Beats receivers should be as similar to beats processes as possible in all of their external outputs. Some of the log messages are different during startup, as it uses different code paths, but most of them are identical. More importantly, the metadata used to query these logs in ES must be identical.

## Checklist

Build agent locally, run it with two identical inputs using different runtimes. Observe that the emitted logs are compatible.

## Related issues


leehinman (dismissed):
nice.  Thank You for the integration test.

testing/integration/beat_receivers_test.go
```diff
@@ -710,6 +685,176 @@ outputs:
...
+			configOptions{
+				RuntimeExperimental: "process",
+			}))
+	processConfig := configBuffer.Bytes()
+	t.Cleanup(func() {
+		if t.Failed() {
+			t.Log("Contents of agent config file:\n")
+			println(string(processConfig))
+		}
+	})
```
pchila: `t.Log` will print only if the test fails or the `-test.v` flag is specified, so this could simply be a `t.Logf()` call.
If it's a simple t.Logf there is no reason to have it in a cleanup.
```suggestion
	t.Logf("Contents of agent config file:\n%s", string(processConfig))
```

Same for the other logs that have the if `t.Failed()` below

testing/integration/beat_receivers_test.go
```diff
@@ -710,6 +685,176 @@ outputs:
...
+
+	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+		var statusErr error
+		status, statusErr := fixture.ExecStatus(ctx)
+		assert.NoError(collect, statusErr)
+		assertBeatsHealthy(collect, &status, component.ProcessRuntimeManager, 1)
+		return
+	}, 1*time.Minute, 1*time.Second)
+
+	cancel()
```
pchila: If the context is cancelled at this point after at most 1 minute spent waiting for the beats to become healthy, why is the context deadline set to 5 minutes ? (same as the run with the beats receivers below)

testing/integration/beat_receivers_test.go
```diff
@@ -710,6 +685,176 @@ outputs:
...
+			}
+
+			if message, ok := logRecord["message"].(string); !ok || !strings.HasPrefix(message, "Beat name:") {
+				continue
+			}
+
+			return logRecord, nil
+		}
+		return nil, nil
+	}
```
pchila: this function could be extracted out of this (already quite long) test

internal/pkg/otel/manager/manager.go
```diff
@@ -197,7 +200,7 @@ func (m *OTelManager) startCollector(cfg *confmap.Conf, errCh chan error) (conte
 		otel.WithExtensionFactory(NewAgentStatusFactory(m)))
 	settings.DisableGracefulShutdown = true // managed by this manager
 	settings.LoggingOptions = []zap.Option{zap.WrapCore(func(zapcore.Core) zapcore.Core {
-		return m.logger.Core() // use same zap as agent
+		return m.baseLogger.Core() // use same zap as agent
```
pchila: Is the comment still valid ? Are we still using the same zap as agent?

testing/integration/beat_receivers_test.go
swiatekm: We always run our integration tests with `-v`, so this would always print this config file. Honestly though, I'm not sure it really helps much in this test, so I'll just delete it.

testing/integration/beat_receivers_test.go
swiatekm: The deadline is for the whole subtest and is more of a general timeout, whereas the 1 minute wait is just for the agent to become healthy. Maybe the general timeout isn't necessary if the setup steps are guaranteed not to block for extended periods of time, but I don't know the test framework well enough to be confident of that.

internal/pkg/otel/manager/manager.go
swiatekm: It's not false per se, but it's misleading. I'll amend it.

swiatekm:
> I suppose I have one big question left: how much of this must change again after PR https://github.com/elastic/elastic-agent/pull/8248 is merged ?

I don't think it'll change much, we just pass in a different logger when starting the otel collector.

testing/integration/beat_receivers_test.go
pchila: The general deadline is a good thing to have to avoid having tests that are stuck or abnormally slow delay the whole integration test run, my observation was that we run elastic-agent processes for 1 minute for each mode without much processing or blocking in between.
Also,  the fact that they are always called `ctx, cancel` while being 3 different contexts with different scopes and duration doesn't help.

pchila (dismissed):

testing/integration/beat_receivers_test.go
swiatekm: I see what you mean. I've created separate subcontexts for both agent processes I'm starting in this test, which have no deadline and only exist to stop the agent by cancelling the context. The parent context with the deadline remains, but it's the only one with a deferred cancel. See https://github.com/elastic/elastic-agent/pull/8606/commits/653f226dab9fbd46b71f5d9f6c5141798f13f236. 

Does that make things better?

leehinman (approved):

khushijain21 (approved):

testing/integration/beat_receivers_test.go
pchila: Yes, thank you

pchila:
@swiatekm It seems that  when `TestBeatsReceiverLogs` fails, it leaves running elastic-agent processes behind (I had a look at `build/default_sudo_Linux_6.11.0-1015-gcp_x86_64.integration.out.json` in [this BK step](https://buildkite.com/elastic/elastic-agent/builds/23063#0197a827-f73d-474c-9b53-afed88ab09e3) and after failure an extra elastic-agent process is detected by all following tests).

Could you please have a look ?

swiatekm:
> @swiatekm It seems that when `TestBeatsReceiverLogs` fails, it leaves running elastic-agent processes behind (I had a look at `build/default_sudo_Linux_6.11.0-1015-gcp_x86_64.integration.out.json` in [this BK step](https://buildkite.com/elastic/elastic-agent/builds/23063#0197a827-f73d-474c-9b53-afed88ab09e3) and after failure an extra elastic-agent process is detected by all following tests).
> 
> Could you please have a look ?

I had some issues with context creation that I fixed, but even with those issues, I'm not sure how the agent could keep running after the parent context was cancelled. I also found a bug in the `testcontext.WithDeadline` function, which I [fixed](https://github.com/elastic/elastic-agent/pull/8606/commits/3459113f50008c1f4b3ce6de2b631e1919be2543), but this also shouldn't have prevented agent from being stopped by the context cancellation.

pchila (approved):

Issue #8634, opened 2025-06-23
Upgrade test artifact fetcher does not validate checksums or retry corrupted downloads.


cmacknz:
For an example where this would help see, https://buildkite.com/elastic/elastic-agent/builds/22893

```
=== RUN   TestStandaloneUpgrade/Upgrade_8.16.6_to_8.17.8-SNAPSHOT_(privileged)
    fetcher_artifact.go:213: Downloading artifact from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.16.6-linux-x86_64.tar.gz
    fetcher_artifact.go:292: Downloading artifact progress 49.38%
    fetcher_artifact.go:292: Downloading artifact progress 100.00%
    fetcher_artifact.go:259: Completed downloading artifact from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.16.6-linux-x86_64.tar.gz
    fetcher_artifact.go:213: Downloading artifact from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.16.6-linux-x86_64.tar.gz.sha512
    fetcher_artifact.go:259: Completed downloading artifact from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.16.6-linux-x86_64.tar.gz.sha512
    fetcher_artifact.go:213: Downloading artifact from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.16.6-linux-x86_64.tar.gz.asc
    fetcher_artifact.go:259: Completed downloading artifact from https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.16.6-linux-x86_64.tar.gz.asc
    fixture.go:293: Extracting artifact elastic-agent-8.16.6-linux-x86_64.tar.gz to /tmp/TestStandaloneUpgrade-Upgrade_8.16.6_to_8.17.8-SNAPSHOT_(privileged)125321857
    upgrade_standalone_test.go:68: 
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1750687179594797174/elastic/elastic-agent-extended-testing/testing/integration/upgrade_standalone_test.go:68
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1750687179594797174/elastic/elastic-agent-extended-testing/testing/integration/upgrade_standalone_test.go:43
        	Error:      	Received unexpected error:
[...]

Issue #7853, opened 2025-04-15
[Flaky Test]: TestFleetManagedUpgradeUnprivileged – failed to find artifact URI for version 9.0.0-SNAPSHOT: failed to find snapshot information for version "9.0.0-SNAPSHOT"


pchila:
### Failing test case

TestFleetManagedUpgradeUnprivileged

### Error message

failed to find artifact URI for version 9.0.0-SNAPSHOT: failed to find snapshot information for version "9.0.0-SNAPSHOT": unexpected status code 502 from https://snapshots.elastic.co/latest/9.0.0-SNAPSHOT.json

### Build

https://buildkite.com/elastic/elastic-agent/builds/19990#01963868-e50a-4dbe-8513-8a03dc136813/194-481

### OS

Linux

### Stacktrace and notes

```markdown
=== FAIL: testing/integration TestFleetManagedUpgradeUnprivileged (11.76s)
    fetcher.go:93: Using existing artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz
    fixture.go:295: Extracting artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestFleetManagedUpgradeUnprivileged279240741
    fixture.go:313: Completed extraction of artifact elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz to /tmp/TestFleetManagedUpgradeUnprivileged279240741
    fixture.go:997: Components were not modified from the fetched artifact
    fixture.go:670: >> running binary with: [/tmp/TestFleetManagedUpgradeUnprivileged279240741/elastic-agent-9.0.0-SNAPSHOT-linux-x86_64/elastic-agent version --binary-only --yaml]
    upgrade_fleet_test.go:95:
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1744703397443442954/elastic/elastic-agent/testing/integration/upgrade_fleet_test.go:95
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1744703397443442954/elastic/elastic-agent/testing/integration/upgrade_fleet_test.go:54
        	Error:      	Received unexpected error:
        	            	failed to find artifact URI for version 9.0.0-SNAPSHOT: failed to find snapshot information for version "9.0.0-SNAPSHOT": unexpected status code 502 from https://snapshots.elastic.co/latest/9.0.0-SNAPSHOT.json
[...]

PR #8670, opened 2025-06-25
[testing] validate artifact hashes in artifact fetcher

pkoutsovasilis:
## What does this PR do?

This PR improves the robustness of the test artifact fetcher by validating the integrity of downloaded packages and introducing retry logic for both corrupted downloads and snapshot metadata fetch failures. Specifically, it:

- Validates the `.sha512` checksum of each downloaded artifact to detect corruption.
  - Automatically retries corrupted downloads, up to 3 attempts,  using a constant backoff of 3 seconds.
- Applies retry logic to snapshot metadata resolution (e.g., `.json` file for snapshot build IDs), which previously failed with transient errors.

## Why is it important?

This PR addresses two known failure cases in upgrade test pipelines:

- **Corrupted downloads**: Previously, corrupted `.tar.gz` artifacts would not be detected, leading to tar extraction errors such as `flate: corrupt input before offset...`. This caused test flakiness and debugging overhead.
- **Transient 502 errors during metadata fetch**: Snapshot builds occasionally return 502 errors when fetching the latest snapshot manifest. Without retry logic, these tests fail unnecessarily.

By introducing checksum validation and retry logic, this PR makes upgrade tests more reliable and deterministic.

## Checklist

## Disruptive User Impact

None. This change only affects internal test infrastructure and does not impact users of the Elastic Agent or its public APIs.

```
mage unitTest
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/8634
- Closes to https://github.com/elastic/elastic-agent/issues/7853

### DIFF
--- a/pkg/testing/fetcher_artifact.go
+++ b/pkg/testing/fetcher_artifact.go
@@ -6,8 +6,6 @@ package testing
 
 import (
 	"context"
-	"crypto/sha512"
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -21,6 +19,7 @@ import (
 
 	"github.com/cenkalti/backoff/v5"
 
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/upgrade/artifact/download"
 	semver "github.com/elastic/elastic-agent/pkg/version"
 )
 
@@ -135,7 +134,7 @@ func (r *artifactResult) Fetch(ctx context.Context, l Logger, dir string) error 
 		}
 
 		// check package hash
-		if err = checkPackageSHA512Hash(dst, dst+extHash); err != nil {
+		if err = download.VerifySHA512Hash(dst); err != nil {
 			l.Logf("inconsistent package hash detected: %s", err)
 			return nil, fmt.Errorf("inconsistent package hash: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/testing/fetcher_artifact.go
```diff
@@ -203,6 +234,32 @@ func findLatestSnapshot(ctx context.Context, doer httpDoer, version string) (bui
...
+
+	hasher := sha512.New()
+	pkgData, err := os.ReadFile(pkgPath)
+	if err != nil {
+		return fmt.Errorf("failed to read package data: %w", err)
+	}
+	_, err = hasher.Write(pkgData)
+	if err != nil {
+		return fmt.Errorf("failed to calculate package hash: %w", err)
+	}
```
pchila: Packages can be big, it would be preferable to use an io.Reader here, something like
```suggestion
	pkgFile, err :=  os.Open(pkgPath)
	if err != nil {
		return fmt.Errorf("failed to read package data: %w", err)
	}
	defer pkgFile.Close()
	
	_, err = io.Copy(hasher, pkgFile)
	if err != nil {
		return fmt.Errorf("failed to calculate package hash: %w", err)
	}
```

pkoutsovasilis: good catch, pushed here https://github.com/elastic/elastic-agent/pull/8670/commits/de859d60ff4043b3bc41bc8a8dc229c302c9d9d4 🙂

pchila (approved):
LGTM

pkg/testing/fetcher_artifact.go
```diff
@@ -203,6 +234,43 @@ func findLatestSnapshot(ctx context.Context, doer httpDoer, version string) (bui
 	}
 }
 
+func checkPackageSHA512Hash(pkgPath, pkgHashPath string) error {
```
swiatekm: Could we use the same function as agent itself uses for verifying artifacts during upgrades here?

pkoutsovasilis: good point, we should definitely reuse whatever already exists, 2a9b1a70e78b8eb616f541aca8e5441e95104515 here you are

pchila (approved):

Issue #6698, opened 2025-02-03
Revert elastic-agent 9.0.0 pinning


pchila:
In PR https://github.com/elastic/elastic-agent/pull/6682 we needed to pin elastic agent package version to 9.0.0 due to issues with calculating previous minor version after bumping agent version to 9.1.0 (see CI runs for PR https://github.com/elastic/elastic-agent/pull/6649).

The problem comes from the fact that 9.0.0 version is not yet released and that there are not 2 minor versions in the 9.x major version.

The goal of this issue is to revert the hack introduced in PR https://github.com/elastic/elastic-agent/pull/6682:
- either waiting for 9.0.0 to be released (more realistically we would need also 9.1.0 released)
- or changing the way previous minor versions are retrieved allowing for some slack when missing versions.

ebeahan:
> or changing the way previous minor versions are retrieved allowing for some slack when missing versions.

I believe we're tracking this in [#6667](https://github.com/elastic/elastic-agent/issues/6667)? Anything beyond the scope of #6667 we should also track here, or can this be closed?

pchila:
#6667 has been created to track the issues with the `TestPreviousMinor()` unit test specifically that are listed [here](https://github.com/elastic/elastic-agent/issues/6667#issuecomment-2812421348)

We have a bigger problem with the logic of `PreviousMinor()` that we encountered with this [set of versions ](https://github.com/elastic/elastic-agent/blob/fc0939e051cb8b7429532783103cc4c7dff18446/testing/integration/testdata/.upgrade-test-agent-versions.yml) while having a current version set to `9.1.0-SNAPSHOT`.
If the logic issues of PreviousMinor are to be solved through #6667 as well, then the related PR #8673 needs to include a test with the same data and some logic changes to make it work correctly.

/cc @kaanyalti

PR #8673, opened 2025-06-25
Fix/6667 fix previous minor unit test

### LINKED ISSUE #6667: Review TestPreviousMinor() (and version in general) unit tests

kaanyalti:
- Bug

## What does this PR do?

Updated PreviousMinor function to receive version and upgradeable version list to make the function testable. Document current contract with unit tests.

## Why is it important?

The unit test for PreviousMinor depended on global vars where the value of these vars were set via init. The unit test did not formalize any expected behaviour

## Checklist

## Disruptive User Impact

None

Running the unit test for previous minor and running the integration tests that use previous minor is sufficient.

**Effected integration tests:**
`TestUpgradeAgentWithTamperProtectedEndpoint_RPM`
`TestUpgradeAgentWithTamperProtectedEndpoint_DEB`
`TestUpgradeBrokenPackageVersion`
`TestStandaloneUpgradeWithGPGFallbackOneRemoteFailing`
`TestStandaloneUpgradeRollback`
`TestStandaloneUpgradeRollbackOnRestarts`
`TestStandaloneUpgradeFailsWhenUpgradeIsInProgress`
`TestStandaloneUpgradeRetryDownload`
`TestStandaloneUpgradeUninstallKillWatcher`

## Related issues

- Closes #6667 
- Creates #8910

### DIFF
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -85,9 +85,7 @@ type AgentVersions struct {
 	TestVersions []string `yaml:"testVersions"`
 }
 
-var (
-	agentVersions *AgentVersions
-)
+var agentVersions *AgentVersions
 
 func init() {
 	AgentVersionsFilename = filepath.Join("testing", "integration", "testdata", ".upgrade-test-agent-versions.yml")
@@ -249,14 +247,10 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 }
 
 // PreviousMinor returns the previous minor version available for upgrade.
-func PreviousMinor() (*version.ParsedSemVer, error) {
-	versions, err := GetUpgradableVersions()
+func PreviousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer) (*version.ParsedSemVer, error) {
+	current, err := version.ParseVersion(currentVersion)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get upgradable versions: %w", err)
-	}
-	current, err := version.ParseVersion(define.Version())
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse the current version %s: %w", define.Version(), err)
+		return nil, fmt.Errorf("failed to parse the current version %s: %w", currentVersion, err)
 	}
 
 	// Special case: if we are in the first release (or a patch release of the first release) of a new major (so vX.0.x), we should
@@ -267,7 +261,7 @@ func PreviousMinor() (*version.ParsedSemVer, error) {
 		// will only contain minors from the previous major (vX-1). Further, since the
 		// version list is sorted in descending order (newer versions first), we can return the
 		// first item from the list as it will be the newest minor of the previous major.
-		for _, v := range versions {
+		for _, v := range upgradeableVersions {
 			if v.Less(*current) {
 				return v, nil
 			}
@@ -276,7 +270,7 @@ func PreviousMinor() (*version.ParsedSemVer, error) {
 		return nil, ErrNoPreviousMinor
 	}
 
-	for _, v := range versions {
+	for _, v := range upgradeableVersions {
 		if v.Prerelease() != "" || v.BuildMetadata() != "" {
 			continue
 		}
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -246,31 +246,31 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 	return upgradableVersions, nil
 }
 
-// PreviousMinor returns the previous minor version available for upgrade.
+// PreviousMinor returns the previous minor version available for upgrade from the given list of upgradeable versions.
+//
+// The function follows these logic rules:
+//  1. If the current version is a first minor release (minor = 0) or a prerelease of the first minor (minor = 1 with prerelease),
+//     it returns the most recent version from the previous major that is less than the current version.
+//  2. For all other cases, it returns the most recent previous minor version within the same major version.
+//  3. The function skips versions with prerelease tags or build metadata when looking for previous minors.
+//  4. If no suitable previous minor version is found, it returns ErrNoPreviousMinor.
 func PreviousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer) (*version.ParsedSemVer, error) {
 	current, err := version.ParseVersion(currentVersion)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse the current version %s: %w", currentVersion, err)
 	}
 
-	// Special case: if we are in the first release (or a patch release of the first release) of a new major (so vX.0.x), we should
-	// return the latest release from the previous major.
-	if current.Minor() == 0 {
-		// Since the current version is the first release of a new major (vX.0.0), there
-		// will be no minor versions in the versions list from the same major (vX). The list
-		// will only contain minors from the previous major (vX-1). Further, since the
-		// version list is sorted in descending order (newer versions first), we can return the
-		// first item from the list as it will be the newest minor of the previous major.
-		for _, v := range upgradeableVersions {
-			if v.Less(*current) {
-				return v, nil
+	var mostRecentVersion *version.ParsedSemVer
+
+	for _, v := range upgradeableVersions {
+		if (current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")) && v.Less(*current) {
+			if mostRecentVersion == nil {
+				mostRecentVersion = v
+			} else if mostRecentVersion.Less(*v) {
+				mostRecentVersion = v
 			}
 		}
 
-		return nil, ErrNoPreviousMinor
-	}
-
-	for _, v := range upgradeableVersions {
 		if v.Prerelease() != "" || v.BuildMetadata() != "" {
 			continue
 		}
@@ -278,6 +278,11 @@ func PreviousMinor(currentVersion string, upgradeableVersions []*version.ParsedS
 			return v, nil
 		}
 	}
+
+	if mostRecentVersion != nil {
+		return mostRecentVersion, nil
+	}
+
 	return nil, ErrNoPreviousMinor
 }
 
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -272,8 +272,9 @@ func previousMinor(currentVersion string, upgradeableVersions []*version.ParsedS
 
 	var mostRecentVersion *version.ParsedSemVer
 
+	isFirstMajorOrFirstMinorPrerelease := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
+
 	for _, v := range upgradeableVersions {
-		isFirstMajorOrFirstMinorPrerelease := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
 		if isFirstMajorOrFirstMinorPrerelease && v.Less(*current) {
 			if mostRecentVersion == nil || mostRecentVersion.Less(*v) {
 				mostRecentVersion = v
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -257,40 +257,27 @@ func PreviousMinor() (*version.ParsedSemVer, error) {
 }
 
 // previousMinor returns the previous minor version available for upgrade from the given list of upgradeable versions.
-//
-// The function follows these logic rules:
-//  1. If the current version is a first minor release (minor = 0) or a prerelease of the first minor (minor = 1 with prerelease),
-//     it returns the most recent version from the previous major that is less than the current version.
-//  2. For all other cases, it returns the most recent previous minor version within the same major version.
-//  3. The function skips versions with prerelease tags or build metadata when looking for previous minors.
-//  4. If no suitable previous minor version is found, it returns ErrNoPreviousMinor.
 func previousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer) (*version.ParsedSemVer, error) {
 	current, err := version.ParseVersion(currentVersion)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse the current version %s: %w", currentVersion, err)
 	}
 
-	var mostRecentVersion *version.ParsedSemVer
-
-	isFirstMajorOrFirstMinorPrerelease := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
-
-	for _, v := range upgradeableVersions {
-		if isFirstMajorOrFirstMinorPrerelease && v.Less(*current) {
-			if mostRecentVersion == nil || mostRecentVersion.Less(*v) {
-				mostRecentVersion = v
+	if current.Minor() == 0 {
+		for _, v := range upgradeableVersions {
+			if v.Less(*current) {
+				return v, nil
 			}
 		}
+	}
 
+	for _, v := range upgradeableVersions {
 		if v.Prerelease() != "" || v.BuildMetadata() != "" {
 			continue
 		}
 		if v.Major() == current.Major() && v.Minor() < current.Minor() {
 			return v, nil
 		}
-	}
-
-	if mostRecentVersion != nil {
-		return mostRecentVersion, nil
 	}
 
 	return nil, ErrNoPreviousMinor
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -263,12 +263,21 @@ func previousMinor(currentVersion string, upgradeableVersions []*version.ParsedS
 		return nil, fmt.Errorf("failed to parse the current version %s: %w", currentVersion, err)
 	}
 
+	// Special case: if we are in the first release (or a patch release of the first release) of a new major (so vX.0.x), we should
+	// return the latest release from the previous major.
 	if current.Minor() == 0 {
+		// Since the current version is the first release of a new major (vX.0.0), there
+		// will be no minor versions in the versions list from the same major (vX). The list
+		// will only contain minors from the previous major (vX-1). Further, since the
+		// version list is sorted in descending order (newer versions first), we can return the
+		// first item from the list as it will be the newest minor of the previous major.
 		for _, v := range upgradeableVersions {
 			if v.Less(*current) {
 				return v, nil
 			}
 		}
+
+		return nil, ErrNoPreviousMinor
 	}
 
 	for _, v := range upgradeableVersions {
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -263,10 +263,9 @@ func PreviousMinor(currentVersion string, upgradeableVersions []*version.ParsedS
 	var mostRecentVersion *version.ParsedSemVer
 
 	for _, v := range upgradeableVersions {
-		if (current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")) && v.Less(*current) {
-			if mostRecentVersion == nil {
-				mostRecentVersion = v
-			} else if mostRecentVersion.Less(*v) {
+		isFirstMajorOrFirstMinorPrerelease := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
+		if isFirstMajorOrFirstMinorPrerelease && v.Less(*current) {
+			if mostRecentVersion == nil || mostRecentVersion.Less(*v) {
 				mostRecentVersion = v
 			}
 		}
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -253,59 +253,39 @@ func PreviousMinor() (*version.ParsedSemVer, error) {
 		return nil, err
 	}
 
-	return previousMinor(currentVersion, upgradeableVersions, false)
-}
-
-func PreviousMinorReleaseOnly() (*version.ParsedSemVer, error) {
-	currentVersion := define.Version()
-	upgradeableVersions, err := GetUpgradableVersions()
-	if err != nil {
-		return nil, err
-	}
-
-	return previousMinor(currentVersion, upgradeableVersions, true)
+	return previousMinor(currentVersion, upgradeableVersions)
 }
 
 // previousMinor returns the previous minor version available for upgrade from the given list of upgradeable versions.
 //
-// The function attempts to find a suitable previous version using the following strategy:
-//
-//  1. First, it searches for a released previous minor version within the same major version.
-//     This search skips any versions with prerelease tags or build metadata.
-//
-//  2. If no previous minor is found in the same major, and the current version is either:
-//     - A first major release (minor = 0), or
-//     - A prerelease of the first minor (minor = 1 with prerelease)
-//     Then it falls back to searching the previous major version.
-//
-//  3. When falling back to the previous major:
-//     - If fallbackReleaseVersion is true, only released versions are considered
-//     - If fallbackReleaseVersion is false, any version (including prereleases) is considered
-//     - The most recent qualifying version from the previous major is returned
-func previousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer, fallbackReleaseVersion bool) (*version.ParsedSemVer, error) {
+// The function follows these logic rules:
+//  1. If the current version is a first minor release (minor = 0) or a prerelease of the first minor (minor = 1 with prerelease),
+//     it returns the most recent version from the previous major that is less than the current version.
+//  2. For all other cases, it returns the most recent previous minor version within the same major version.
+//  3. The function skips versions with prerelease tags or build metadata when looking for previous minors.
+//  4. If no suitable previous minor version is found, it returns ErrNoPreviousMinor.
+func previousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer) (*version.ParsedSemVer, error) {
 	current, err := version.ParseVersion(currentVersion)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse the current version %s: %w", currentVersion, err)
 	}
 
 	var mostRecentVersion *version.ParsedSemVer
-	needsFallback := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
+
+	isFirstMajorOrFirstMinorPrerelease := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
 
 	for _, v := range upgradeableVersions {
-		isPrerelease := v.Prerelease() != "" || v.BuildMetadata() != ""
-
-		if !isPrerelease && v.Major() == current.Major() && v.Minor() < current.Minor() {
-			return v, nil
-		}
-
-		if needsFallback && v.Less(*current) {
-			if fallbackReleaseVersion && isPrerelease {
-				continue
-			}
-
+		if isFirstMajorOrFirstMinorPrerelease && v.Less(*current) {
 			if mostRecentVersion == nil || mostRecentVersion.Less(*v) {
 				mostRecentVersion = v
 			}
+		}
+
+		if v.Prerelease() != "" || v.BuildMetadata() != "" {
+			continue
+		}
+		if v.Major() == current.Major() && v.Minor() < current.Minor() {
+			return v, nil
 		}
 	}
 
--- a/testing/upgradetest/versions.go
+++ b/testing/upgradetest/versions.go
@@ -253,39 +253,59 @@ func PreviousMinor() (*version.ParsedSemVer, error) {
 		return nil, err
 	}
 
-	return previousMinor(currentVersion, upgradeableVersions)
+	return previousMinor(currentVersion, upgradeableVersions, false)
+}
+
+func PreviousMinorReleaseOnly() (*version.ParsedSemVer, error) {
+	currentVersion := define.Version()
+	upgradeableVersions, err := GetUpgradableVersions()
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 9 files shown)


pchila:
@kaanyalti, as I understand it, this PR refactors PreviousMinor unit tests to be able to pass arbitrary sets of version in various testcases correct? This is a prerequisite for fixing the issues we had with PreviousMinor erroring out when bumping to version 9.1.0 (see https://github.com/elastic/elastic-agent/issues/6698#issuecomment-3003961359), so there should be some follow-up to this.

Would it also be possible to limit the changes to the `upgradetest` package for the moment?

testing/upgradetest/versions.go
```diff
@@ -248,14 +246,10 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 }
 
 // PreviousMinor returns the previous minor version available for upgrade.
-func PreviousMinor() (*version.ParsedSemVer, error) {
-	versions, err := GetUpgradableVersions()
-	if err != nil {
-		return nil, fmt.Errorf("failed to get upgradable versions: %w", err)
-	}
-	current, err := version.ParseVersion(define.Version())
+func PreviousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer) (*version.ParsedSemVer, error) {
```
ycombinator: How about leaving the signature of `PreviousMinor()` as is and introducing a new function, say `previousMinor(currentVersion string, upgradeableVersions []*version.ParsedSemVer) (*version.ParsedSemVer, error)`, that `PreviousMinor()` calls?  This way you don't need to change all the callers of `PreviousMinor()` but you can still have its logic be testable.

testing/upgradetest/versions_test.go
```diff
@@ -249,33 +248,118 @@ func TestGetUpgradableVersions(t *testing.T) {
 }
 
 func TestPreviousMinor(t *testing.T) {
-	currentParsed, err := version.ParseVersion(bversion.Agent)
-	require.NoError(t, err)
+	testCases := []struct {
+		name                string
```
ycombinator: Nit: use a `map[string]struct{` instead, where the keys of the map are what you are putting in the `name` field of the struct.  This way you don't need a separate `name` field but, more importantly, the test cases get run in a random order each time, which prevents any test case from having unintended side-effects that could affect another test case.

testing/upgradetest/versions_test.go
```diff
@@ -249,33 +248,118 @@ func TestGetUpgradableVersions(t *testing.T) {
...
+			name:           "should return the previous minor from the same major and skip prerelease versions and versions with metadata",
+			currentVersion: "9.1.0",
+			upgradeableVersions: []string{
+				"9.0.3-SNAPSHOT",
+				"9.0.2+metadata",
+				"9.0.1",
+				"8.19.0-SNAPSHOT",
+				"8.18.2",
+			},
+			expectedVersion: "9.0.1",
```
ycombinator: It's been a while and I don't see comments about it, so I'll defer to @pchila on this.  Should the expected version here be `9.0.1` or `9.0.3-SNAPSHOT`?  Intuitively, I would expect the latter but maybe there is some reason we don't want to include snapshot versions or versions with metadata?  If so, we should add this reason to the godoc for the `TestPreviousMinor` function.

kaanyalti: I do not know the reason either, but this is how the function works at the moment. @pchila could you please clarify this?

kaanyalti:
> @kaanyalti, as I understand it, this PR refactors PreviousMinor unit tests to be able to pass arbitrary sets of version in various testcases correct? This is a prerequisite for fixing the issues we had with PreviousMinor erroring out when bumping to version 9.1.0 (see [#6698 (comment)](https://github.com/elastic/elastic-agent/issues/6698#issuecomment-3003961359)), so there should be some follow-up to this.
> 

Yes, that’s correct, this PR refactors PreviousMinor so that it can be properly unit tested with arbitrary sets of versions. I’m also open to expanding the scope to address the issue with version 9.1.0 that you mentioned.

 @ebeahan would love your thoughts on how you’d prefer to proceed. In your earlier [comment](https://github.com/elastic/elastic-agent/issues/6698#issuecomment-2936355604), you mentioned that #6667 might already address the logic issue. If that’s the intended direction, I’m happy to update #6667's description and adjust the scope accordingly.

kaanyalti:
> Would it also be possible to limit the changes to the `upgradetest` package for the moment?

@pchila Just to make sure I understand correctly, are you referring to changes like [this one](https://github.com/elastic/elastic-agent/pull/8673/files#diff-cdbdb54b4ffd330d24eabac043b62424594d2441b46c430a98748ff85598496cL87-L89) or more broadly to what @ycombinator mentioned [here](https://github.com/elastic/elastic-agent/pull/8673#discussion_r2168882944) which includes changes in the upgrade test files outside the upgradetest package?

testing/upgradetest/versions.go
kaanyalti: Thanks for the suggestion! I actually considered the approach you’re describing as I was working on this. Since the total number of call sites is relatively small (about 9), I opted to update them directly rather than introduce a wrapper just for compatibility. That said, I’m happy to refactor it to keep the original PreviousMinor() signature if you think it’s better for consistency or long-term maintenance, just let me know what you prefer!

kaanyalti:
Hi @ebeahan and @pchila, I went ahead and updated the function to address the scenario mentioned in [this comment](https://github.com/elastic/elastic-agent/issues/6698#issuecomment-3003961359). With this change, I believe the clarifications I asked earlier may no longer be necessary, but please feel free to weigh in if you still see anything worth adjusting.

testing/upgradetest/versions.go
ycombinator: It's not so much about backwards compatibility as it's about ergonomics and ensuring correctness.  `PreviousMinor()`'s callers shouldn't really need to care about _where_ the versions for determining the previous minors are being sourced from.  So I think giving them a narrower interface is actually better developer experience in this case. 

Another perspective is that the only reason to introduce the new signature is for testability — and that's a very good reason! But if that's really the _only_ reason, then creating a private helper function that's testable while having a public wrapper function that's ergonomic achieves both goals — testability and ergonomic developer experience.

kaanyalti: Thanks for the explanation, that makes a lot of sense. I’ll go ahead and refactor the code to keep the original PreviousMinor() signature and move the logic into a private helper for testability.

kaanyalti:
ci is failing at the moment because 9.2.0 snapshot is not available yet

kaanyalti:
buildkite test this

testing/upgradetest/versions_test.go
pchila: It depends who you ask, that is the problem with having code that is changed without having clear unit tests defining the expectations.
`PreviousMinor()` has been changed so many times to fit the moment's needs that its semantics have been forgotten even by the authors.
As a general rule, I think there should be some configurable filter if a given test can use a SNAPSHOT or it needs an officially released version for some reason.

As such I would leave the unit test as it is to document the current behavior and "fix" PreviousMinor in a separate PR

testing/upgradetest/versions.go
```diff
@@ -248,42 +246,52 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
...
-		for _, v := range versions {
-			if v.Less(*current) {
-				return v, nil
+	var mostRecentVersion *version.ParsedSemVer
+
+	for _, v := range upgradeableVersions {
+		isFirstMajorOrFirstMinorPrerelease := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
+		if isFirstMajorOrFirstMinorPrerelease && v.Less(*current) {
+			if mostRecentVersion == nil || mostRecentVersion.Less(*v) {
+				mostRecentVersion = v
```
pchila: What is the general idea behind the checks on `current.Minor()` and `current.Prerelease()` and why are those evaluated at every iteration ?

testing/upgradetest/versions_test.go
```diff
@@ -249,33 +248,124 @@ func TestGetUpgradableVersions(t *testing.T) {
...
+				"8.17.6",
+				"7.17.29-SNAPSHOT",
+			},
+			expectedVersion: "8.19.0-SNAPSHOT+metadata",
+			expectError:     false,
+		},
+		"should return the most recent version from previous major when the current version is the first major release": {
+			currentVersion: "9.0.0",
+			upgradeableVersions: []string{
+				"8.19.0-SNAPSHOT+metadata",
```
pchila: Is `8.19.0-SNAPSHOT+metadata` selected only because of the build metadata fragment? Or is it for the `0` patch version ? This does not seem very consistent with the behavior of the testcase in line 257 just above where `9.0.3-SNAPSHOT` and `9.0.2+metadata` are ignored as previous minor versions for `9.1.0`

testing/upgradetest/versions_test.go
```diff
@@ -249,33 +248,124 @@ func TestGetUpgradableVersions(t *testing.T) {
...
+			},
+			expectedVersion: "8.19.0-SNAPSHOT+metadata",
+			expectError:     false,
+		},
+		"should return error when no previous minor is found": {
+			currentVersion: "9.1.0",
+			upgradeableVersions: []string{
+				"9.2.0",
+				"9.1.1",
+				"8.19.0-SNAPSHOT+metadata",
```
pchila: So the difference between this and the previous testcase is that we have some 9.x versions but none that is lower than `9.1.0` ? And this changes the output from an `8.19.0...` fallback to an error ?

pchila (commented):
Thank you @kaanyalti for this refactor and the addition of proper unit tests.
The tests show the inconsistencies of PreviousMinor() that derive from trying to adapt the function at every version bump to fit the needs of the moment.
@ebeahan Do you think we should fix the problems in this same PR or maybe it should be done in a follow-up (my preferred option) ?

testing/upgradetest/versions.go
kaanyalti: You’re right that isFirstMajorOrFirstMinorPrerelease doesn’t need to be evaluated in every iteration, I’ve moved it outside the loop.

The purpose of that check is to relax version selection when the current version is either the first minor prerelease or when no first major release is defined. It essentially addresses your earlier comment: [“so there should be some follow-up to this.”](https://github.com/elastic/elastic-agent/pull/8673#issuecomment-3007247371).

I added it here since it felt like a small, relevant improvement to the function, and based on [this comment](https://github.com/elastic/elastic-agent/issues/6698#issuecomment-2936355604), it seemed that #6667 was expected to resolve the core issue. I had asked earlier about whether this PR should be limited strictly to test fixes but wasn’t sure how to proceed, so I went ahead with this approach.

That said, I’m happy to split it out if keeping the scope narrower is preferred.

testing/upgradetest/versions_test.go
kaanyalti: 8.19.0-SNAPSHOT+metadata is selected because the current version is the first patch release, that’s the core logic behind the PreviousMinor function, and it’s consistent with behaviour that existed prior to my changes. This was originally introduced in [this PR](https://github.com/elastic/elastic-agent/pull/7888).

What I’ve done here is extend that logic slightly to also account for the case where the current version is the first minor prerelease and there’s no released first major. In both cases, the idea is to return the most recent available version from the previous minor, ensuring consistency in version selection around release boundaries.

As for the test on line 257, the key difference is that 9.1.0 is a released minor version, not a prerelease, so the additional logic I introduced doesn’t apply in that case. The behaviour remains consistent with the original implementation, where PreviousMinor skips over prereleases and versions with build metadata. This aligns with @ycombinator’s comment [here](https://github.com/elastic/elastic-agent/pull/link).

testing/upgradetest/versions_test.go
kaanyalti: In this case, since 9.1.0 is a released version (not a prerelease), the function expects a previous minor release, 9.0.x. Since there’s none in the list, it returns an error instead of falling back to an 8.x version

testing/upgradetest/versions_test.go
kaanyalti: The unit test changes here are meant to document the current behavior of the function. Previously, PreviousMinor returned the latest released version from the previous minor if one was available, unless the current version was the first major release, in which case it would fall back to the most recent version, including prereleases (introduced in https://github.com/elastic/elastic-agent/pull/7888).

What I added builds on that by handling the case described [here](https://github.com/elastic/elastic-agent/issues/6698#issuecomment-3003961359). Whether that particular addition belongs in this PR is a fair question. I’ve addressed that consideration in [this comment](https://github.com/elastic/elastic-agent/pull/8673#discussion_r2176942393).

testing/upgradetest/versions_test.go
pchila: Correct me if I am wrong, but in our testing we will always have a currentVersion that is a `SNAPSHOT ` so this means that the Previous minor will always return snapshots if it finds them?

I am becoming more and more convinced that we need to let the test choose if a snapshot is acceptable or not, for example:
- tests using PGP key will work with the last snapshot we can find from unified release
- upgrade tests most likely would need a released version to start from

For both cases `currentVersion` will be the same.

ycombinator: Yeah, I'm ++ to having a default behavior for this function re: snapshots but letting a test override it.

kaanyalti: @pchila You’re right, that holds when the current version is either 9.0.0-SNAPSHOT or 9.1.0-SNAPSHOT and there’s no 9.0.x release available. For these specific cases, we could make the function configurable to prefer a released version over a SNAPSHOT.

kaanyalti: I added an option to override the previous minor to return only release versions for the edge cases that are mentioned in this thread (9.0.0, 9.1.0-SNAPSHOT). I kept the function signature the same for `PreviousMinor` and added a new function `PreviousMInorReleaseOnly` to call `previousMinor` with override.

testing/upgradetest/versions_test.go
```diff
@@ -249,33 +248,177 @@ func TestGetUpgradableVersions(t *testing.T) {
 }
 
 func TestPreviousMinor(t *testing.T) {
-	currentParsed, err := version.ParseVersion(bversion.Agent)
-	require.NoError(t, err)
+	testCases := map[string]struct {
```
michalpristas: can you sort test based on current version?
i'd expect us to test
major release - 9.0.0
first patch of new major - 9.0.1
first minor release 9.1.0
first patch of first minor 9.1.1
xth patch 9.1.15
xth major 9.2.0
first patch 9.2.1 
xth patch 9.2.15

plus same for snapshots

with having 8.x in history of releases.
this should put clear contract and expectations for desired behavior, then once we agree on this we can tweak implemenation to fit

testing/upgradetest/versions.go
```diff
@@ -248,42 +246,73 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
 	return upgradableVersions, nil
 }
 
-// PreviousMinor returns the previous minor version available for upgrade.
 func PreviousMinor() (*version.ParsedSemVer, error) {
-	versions, err := GetUpgradableVersions()
+	currentVersion := define.Version()
```
pchila, line 250: I this we should be adding documentation to `PreviousMinor()`, not removing it.

testing/upgradetest/versions.go
```diff
@@ -248,42 +246,73 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
...
+	upgradeableVersions, err := GetUpgradableVersions()
```
pchila, line 251: Maybe we can keep it simpler? (cannot do a single suggestion block because of deleted lines but this should render the idea)
```suggestion
func PreviousMinor() (*version.ParsedSemVer, error) {
	return PreviousMinorVersion(false)
}

func PreviousMinorVersion(includeSnapshots bool) (*version.ParsedSemVer, error) {
	currentVersion := define.Version()
	upgradeableVersions, err := GetUpgradableVersions()
	if err != nil {
		return nil, err
	}

	return previousMinor(currentVersion, upgradeableVersions, includeSnapshots)
}
```

testing/upgradetest/versions.go
```diff
@@ -248,42 +246,73 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
...
+//
+//  2. If no previous minor is found in the same major, and the current version is either:
+//     - A first major release (minor = 0), or
+//     - A prerelease of the first minor (minor = 1 with prerelease)
+//     Then it falls back to searching the previous major version.
+//
+//  3. When falling back to the previous major:
+//     - If fallbackReleaseVersion is true, only released versions are considered
+//     - If fallbackReleaseVersion is false, any version (including prereleases) is considered
+//     - The most recent qualifying version from the previous major is returned
```
pchila: This (or maybe part of this) should be the documentation of the public functions.

testing/upgradetest/versions.go
```diff
@@ -248,42 +246,73 @@ func findRequiredVersions(sortedParsedVersions []*version.ParsedSemVer, reqs Ver
...
-	for _, v := range versions {
-		if v.Prerelease() != "" || v.BuildMetadata() != "" {
-			continue
-		}
-		if v.Major() == current.Major() && v.Minor() < current.Minor() {
+	var mostRecentVersion *version.ParsedSemVer
+	needsFallback := current.Minor() == 0 || (current.Minor() == 1 && current.Prerelease() != "")
+
+	for _, v := range upgradeableVersions {
+		isPrerelease := v.Prerelease() != "" || v.BuildMetadata() != ""
```
pchila: I am not sure that `v.BuildMetadata() != ""` would indicate a prerelease (Independent agent releases have build metadata but those are not prereleases)
Moreover, I was expecting that the boolean added in the signature would control whether we get SNAPSHOTs or not and a SNAPSHOT version is a prerelease version :thinking: 

From the code it seems that SNAPSHOTs would only be considered when `needsFallback` is true and that in only true for `x.0.y` or `x.1.y` versions: I am not sure that is correct.

testing/upgradetest/versions_test.go
pchila: I meant that we should have the option to get SNAPSHOTs regardless of the `(0|1) minor version` fallback (which I would love to get rid of as that looks like a hotfix introduced when we got into trouble with the new 9.x versions)

testing/upgradetest/versions_test.go
kaanyalti: I updated the tests quite a bit and addressed this

testing/upgradetest/versions.go
kaanyalti: The idea behind my implementation was to cover the boundary scenarios so that we won't have to pin any versions or require any other manual intervention to unblock ci  while keeping the overall behaviour of the function the same. Based on the discussions on this PR I decided to stay confined to the original scope of this issue and address only the testability of the function. The requirements around how the function should behave and generally how we should be handling versions in integrations tests is unclear. I will shortly post an update in the pr thread with a follow up issue for clarifying what exactly needs to be done.

kaanyalti:
Based on all the discussion on this PR what I realized is that the requirements for what the PreviousMinor should be doing or how we should be handling versions and what the version api for integration tests should be is nebulous. I removed any logical changes I have made to `previousMinor` due to this. Having a back and forth discussion in this PR thread and trying to reveal what needs to be done bit by bit is slow and inefficient.

I created this issue https://github.com/elastic/elastic-agent/issues/8910 to clarify the requirements for version handling in integration tests and confined my implementation to the initial scope of the parent issue which was to address the testability of the function and to document the current contract via unit tests. 

cc: @ebeahan @pchila @ycombinator

testing/upgradetest/versions.go
kaanyalti: The comments and relevant implementation is removed as explained here https://github.com/elastic/elastic-agent/pull/8673#issuecomment-3052740573

testing/upgradetest/versions.go
kaanyalti: The implementation here is not relevant as explained in this comment https://github.com/elastic/elastic-agent/pull/8673#issuecomment-3052740573

testing/upgradetest/versions.go
kaanyalti: The changes here are removed as explained in this comment https://github.com/elastic/elastic-agent/pull/8673#issuecomment-3052740573

testing/upgradetest/versions_test.go
kaanyalti: I have just added this comment https://github.com/elastic/elastic-agent/pull/8673#issuecomment-3052740573 addressing the changes here.

testing/upgradetest/versions.go
kaanyalti: Based on the discussion in this PR thread I have removed any logical changes and opted for splitting the implementation changes. Here is the relevant comment https://github.com/elastic/elastic-agent/pull/8673#issuecomment-3052740573. The follow up issue is this https://github.com/elastic/elastic-agent/issues/8910

testing/upgradetest/versions_test.go
```diff
@@ -14,6 +16,167 @@ import (
 	"github.com/elastic/elastic-agent/pkg/version"
 )
 
+func generateTestVersions(startVersion, endVersion string) ([]*version.ParsedSemVer, error) {
```
pchila: I understand this is what @michalpristas asked but I wonder if we can cover arbitrary sets of versions with this mechanism. 
@michalpristas What was the issue with the `[]string` containing the explicit versions of each testcase again ?

testing/upgradetest/versions_test.go
```diff
@@ -14,6 +16,167 @@ import (
...
+		if !versions[i].Equal(*versions[j]) {
+			return versions[j].Less(*versions[i])
+		}
+		return versions[i].Original() < versions[j].Original()
+	})
+
+	return versions, nil
+}
+
+func TestGenerateTestVersions(t *testing.T) {
```
pchila: This is the test for the test utility function above ?

kaanyalti: Yes this is the test for the utility function

testing/upgradetest/versions_test.go
kaanyalti: Based on @michalpristas’s earlier [comment](https://github.com/elastic/elastic-agent/pull/8673#discussion_r2189565792), my understanding was that the goal was to better organize the test cases and expand coverage. The utility function was something I introduced after running into limitations with manually listing versions for each test case.

At first, I was writing explicit []string version slices for every test. This quickly made the test definitions large, harder to follow, and error-prone. I also tried using several predefined slices that could be reused across tests, but I couldn’t find a practical way to make that work, each test required slightly different sets of versions. I then considered slicing from a shared list (e.g. allVersions[i:j]), but that approach made the tests even harder to understand and less expressive. Slicing by index made it unclear what specific versions were actually being tested, which hurt readability.

The utility function solved these issues by allowing concise, declarative version ranges directly in each test case. It also enforces the correct ordering internally, returning versions from newest to oldest, consistent with how GetUpgradableVersions is documented to behave.

This matters because the previousMinor function doesn’t inherently understand version ordering, and technically it shouldn’t, but due to the way it’s used (and how GetUpgradableVersions is expected to behave), the current implementation relies on the input being in descending order. Without the utility function, it would be easy to accidentally write test inputs in the wrong order, especially for someone unfamiliar with version semantics.

Side note: 
While GetUpgradableVersions does have test coverage, the existing test is similar to how previousMinor was tested before my changes, it relies on global state and doesn’t clearly document expected behaviour. This makes the test harder to reason about and less effective as a specification or safety net when refactoring.

testing/upgradetest/versions_test.go
```diff
@@ -14,6 +16,167 @@ import (
...
+	}
+
+	if !start.Less(*end) && !start.Equal(*end) {
+		return nil, fmt.Errorf("start version %s is newer than end version %s", startVersion, endVersion)
+	}
+
+	for major := start.Major(); major <= end.Major(); major++ {
+		// Arbitrarily chosen well defined range of minor versions
+		minorStart := 0
+		minorEnd := 19
```
michalpristas: nit: i'd make this configurable, for now it will probably be the same in every call

testing/upgradetest/versions_test.go
michalpristas: > several predefined slices that could be reused across tests, 

that was my idea how this could work

i prefer simpler solution tbh,but i see what you did with this and have nothing against it functionally, it is tested, we should probably test with some kind of expectations full or random list to make sure versions are generated correctly.

pchila: @kaanyalti At this point I would prefer to have versions being specified as `[]string` across all testcases.
Feel free to define variables to reduce repetition if needed, but I would stay with the simple solution instead of having a version generator

kaanyalti: @pchila @michalpristas I removed the utility function and started using []string for test versions as both of you suggested

michalpristas (approved):
approving again, less moving parts and better readability compared to previous state

testing/upgradetest/versions_test.go
```diff
@@ -411,6 +249,113 @@ func TestGetUpgradableVersions(t *testing.T) {
...
 
 func TestPreviousMinor(t *testing.T) {
+	combineSlices := func(slices ...[]string) []string {
+		var result []string
+		for _, s := range slices {
+			result = append(result, s...)
+		}
+		return result
+	}
+	previousMinorVersions := []string{
```
pchila, line 259: Nit: this does not look like a typical content of `testing/integration/testdata/.upgrade-test-agent-versions.yml` (not saying that we want to use the file, but the content should be kinda similar to that).
What is the advantage in having each of:
- release
- IAR
- snapshot
- SNAPSHOT + IAR (which does not really exist)
for 8.19, 8.17, 8.17 (times 2 because of patch versions) ?

Does it add coverage having more versions to select from ?

kaanyalti: I intentionally included all these version permutations to ensure that the behavior of previousMinor is fully and explicitly documented. While the test data may not mirror what’s typically found in testdata/.upgrade-test-agent-versions.yml, this file serves a different purpose: to exercise and validate all combinations that are syntactically valid and parseable by the version parser.

Even though PreviousMinor is currently the only caller, we’re actively discussing changes to these functions, and I want the test coverage to anticipate and support that evolution. Including cases like SNAPSHOT + IAR, even if not currently used in practice, ensures that the function’s behaviour is clear and consistent for all valid inputs.

This level of coverage avoids ambiguity and makes the logic easier to reason about in the future. I believe it’s important to keep the test cases comprehensive for this reason.

pchila: > I intentionally included all these version permutations to ensure that the behavior of previousMinor is fully and explicitly documented. While the test data may not mirror what’s typically found in testdata/.upgrade-test-agent-versions.yml, this file serves a different purpose: to exercise and validate all combinations that are syntactically valid and parseable by the version parser.

Could you point where having the same 4 version variations (1 of which is not currently produced by any CI/release process)  for  8.19.15, 8.19.1, 8.19.0, 8.18.15, 8.18.1, 8.18.0, 8.17.15, 8.17.1, 8.17.0  validates PreviousMinor() logic ? 
In all the testcases I didn't see anything lower than `8.19.15` as an expected result, in which testcases the previous versions play a role exactly ?

> 
> Even though PreviousMinor is currently the only caller, we’re actively discussing changes to these functions, and I want the test coverage to anticipate and support that evolution. Including cases like SNAPSHOT + IAR, even if not currently used in practice, ensures that the function’s behaviour is clear and consistent for all valid inputs.
> 
SNAPSHOT + IAR is *not* supported (because it doesn't make sense to make an Independent Agent Release of something that is not released already) so it doesn't add any useful testcase.

> This level of coverage avoids ambiguity and makes the logic easier to reason about in the future. I believe it’s important to keep the test cases comprehensive for this reason.

The level of coverage would not change if you removed all the 8.17 and 8.18 versions from the array given the testcases you have. Coverage is not directly proportional to the number of inputs differing by a minor/patch version, it's about which usecase is being tested and having so much noise in input does not help things

kaanyalti: My intent here is not to mirror production or CI data, but to clearly document and make explicit the current behaviour of the PreviousMinor function across all syntactically valid version formats, including edge cases.

Since this function operates purely on version strings and does not incorporate any filtering based on release realism or existence, I believe the tests should reflect a broad and systematic coverage. This helps surface edge-case behaviours and ensures future readers or contributors can reason about the logic without ambiguity or needing to infer implicit constraints.

While it’s true that current test assertions don’t rely on all older versions in the list, the structure and variety of the test data itself play a role in documenting the contract. This is particularly important as we’re discussing expanding the use of this logic, and I want the groundwork to be as robust and transparent as possible.

I agree that coverage isn’t simply about quantity, but in this case, the variety does help illustrate subtle distinctions, including some unintuitive behaviours that only show up with uncommon combinations. I believe this level of input data is appropriate given the goal of documenting behaviour with clarity and a margin of safety, even if it goes slightly beyond the minimum required to validate the current implementation.

michalpristas: This test will likely evolve based on the behavior we align on in the linked issue. That said, I believe the current coverage is sufficient—it sets clear expectations and helps move the discussion forward.

Let’s resolve this thread and get it shipped.

pchila: @michalpristas thank you for chiming in.
If we are in a hurry to ship this we can always discuss what the significant/useful set of versions for this unit test should be in a follow-up issue/PR.

Consider this thread resolved, for the sake of speed.

PR #8696, opened 2025-06-26
Fix otel extension status reporting

swiatekm:
## What does this PR do?

Fix a bug where Elastic Agent would enter a failed state if components were running as beats receivers and the otel collector also had an extension defined via hybrid mode.

The result would be the following agent status:

```
┌─ fleet
└─ elastic-agent
   ├─ status: (FAILED) OTel manager failed: pipeline status id extensions is not a pipeline
   ├─ extensions
   │  ├─ status: StatusOK
   │  └─ extension:health_check/v2
   │     └─ status: StatusOK
```

## Why is it important?

We should report status for otel extensions correctly.

## Checklist

Build agent locally and run it with the following configuration:

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    username: "elastic"
    password: ""

agent:
  monitoring:
    enabled: true
    _runtime_experimental: otel

inputs: []

# Embedded Otel configuration
receivers:
  nop:
exporters:
  nop:
extensions:
  health_check/v2:
service:
  extensions: [health_check/v2]
  pipelines:
    logs:
      receivers: [nop]
      exporters: [nop]
```

Then run `elastic-agent status`. You should see the extension status and the statuses of monitoring components:

```
┌─ fleet
└─ elastic-agent
   ├─ beat/metrics-monitoring
   │  └─ beat/metrics-monitoring
   ├─ filestream-monitoring
   │  └─ filestream-monitoring
   ├─ http/metrics-monitoring
   │  └─ http/metrics-monitoring
   ├─ extensions
   │  ├─ status: StatusOK
   │  └─ extension:health_check/v2
   │     └─ status: StatusOK
   └─ pipeline:logs
      ├─ status: StatusOK
      ├─ exporter:nop
      │  └─ status: StatusOK
      └─ receiver:nop
         └─ status: StatusOK

```

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/5751

### DIFF
--- a/internal/pkg/otel/translate/status.go
+++ b/internal/pkg/otel/translate/status.go
@@ -64,7 +64,10 @@ func DropComponentStateFromOtelStatus(otelStatus *status.AggregateStatus) (*stat
 	newStatus := deepCopyStatus(otelStatus)
 	for pipelineStatusId := range newStatus.ComponentStatusMap {
 		pipelineId := &pipeline.ID{}
-		componentKind, pipelineIdStr := parseEntityStatusId(pipelineStatusId)
+		componentKind, pipelineIdStr, parseErr := parseEntityStatusId(pipelineStatusId)
+		if parseErr != nil {
+			return nil, parseErr
+		}
 		if componentKind == "extensions" {
 			continue
 		}
@@ -94,7 +97,10 @@ func getOtelRuntimePipelineStatuses(otelStatus *status.AggregateStatus) (map[str
 
 	for pipelineStatusId, pipelineStatus := range otelStatus.ComponentStatusMap {
 		pipelineId := &pipeline.ID{}
-		componentKind, pipelineIdStr := parseEntityStatusId(pipelineStatusId)
+		componentKind, pipelineIdStr, parseErr := parseEntityStatusId(pipelineStatusId)
+		if parseErr != nil {
+			return nil, parseErr
+		}
 		if componentKind == "extensions" {
 			continue
 		}
@@ -182,8 +188,11 @@ func getUnitOtelStatuses(pipelineStatus *status.AggregateStatus) (
 
 	for otelCompStatusId, otelCompStatus := range pipelineStatus.ComponentStatusMap {
 		var otelComponentID otelcomponent.ID
-		componentKind, otelComponentIDStr := parseEntityStatusId(otelCompStatusId)
-		err := otelComponentID.UnmarshalText([]byte(otelComponentIDStr))
+		componentKind, otelComponentIDStr, parseErr := parseEntityStatusId(otelCompStatusId)
+		if parseErr != nil {
+			return nil, nil, parseErr
+		}
+		err = otelComponentID.UnmarshalText([]byte(otelComponentIDStr))
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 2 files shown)


internal/pkg/otel/translate/status.go
```diff
@@ -267,8 +273,12 @@ func otelStatusToUnitState(status componentstatus.Status) client.UnitState {
 
 // parseEntityStatusId parses an entity status ID into its kind and entity ID. An entity can be a pipeline or otel component.
 // The ID is expected to be in the format "kind:entityId", where kind is either "pipeline" or the otel component type (e.g., "receiver", "exporter").
+// The returned entityId may be empty - this is true for the top-level "extensions" key.
 // This format is used by the healthcheckv2 extension.
 func parseEntityStatusId(id string) (kind string, entityId string) {
+	if id == "extensions" {
+		return "extensions", ""
+	}
```
pkoutsovasilis: this seems a little bit too specific to `extensions`. I would generalise that in the lines if `id` doesn't contain a `:` then return the `id, ""`, wdyt?

swiatekm: As far as I know this is the only situation where this id is valid without a colon in it. So I'd prefer to make it an explicit exception rather than potentially return incorrect values.

pkoutsovasilis: great we should always prefer explicitness, I am with you 🙂  So, since you ultimately want to avoid incorrect values, return an error if there is not a `:` in the `id` this is explicit, wdyt?

swiatekm: Sure, done.

pkoutsovasilis (commented):
I did merge main and run this locally with subprocess mode and I still see this behaviour everytime I "kill" the collector subprocess

```
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   └─ extensions
      ├─ status: StatusOK
      ├─ extension:healthcheckv2/0018a7f6-cbea-4a2f-b314-9e75eb9aa55f
      │  └─ status: StatusOK
      ├─ extension:healthcheckv2/5c5f54ad-54f4-4ed9-bf9d-a7e8a8b3a161
      │  └─ status: StatusOK
      └─ extension:healthcheckv2/6c2bbefd-f0cf-4a76-aba5-5df3412d0185
         └─ status: StatusOK
```

Is the above something that this PR should handle? I am still not quite sure why I even see `extensions` in the elastic-agent status 😄

swiatekm:
> I did merge main and run this locally with subprocess mode and I still see this behaviour everytime I "kill" the collector subprocess
> 
> ```
> ┌─ fleet
> └─ elastic-agent
>    └─ extensions
>       ├─ status: StatusOK
>       ├─ extension:healthcheckv2/0018a7f6-cbea-4a2f-b314-9e75eb9aa55f
>       │  └─ status: StatusOK
>       ├─ extension:healthcheckv2/5c5f54ad-54f4-4ed9-bf9d-a7e8a8b3a161
>       │  └─ status: StatusOK
>       └─ extension:healthcheckv2/6c2bbefd-f0cf-4a76-aba5-5df3412d0185
>          └─ status: StatusOK
> ```
> 
> Is the above something that this PR should handle? I am still not quite sure why I even see `extensions` in the elastic-agent status 😄

And this only happens in subprocess mode?

pkoutsovasilis:
> And this only happens in subprocess mode?

Is there any other case where we actively have an extension loaded at the moment?! IIUC, as this regards only hybrid elastic agent, for the embedded mode no extensions are loaded at the moment, right?

swiatekm:
@pkoutsovasilis I found the root cause of that bug. It has nothing to do with status reporting, we're actually running multiple extensions in the subprocess collector. The reason is that we don't make a copy of the collector config at any point, so we continue adding new healthcheckv2 extensions every time the collector is restarted.

#8529 fixes this by accident because it always creates its own config file. Up to you if we want to fix it separately first.

In any case, this PR fixes a different unrelated issue, and we shouldn't block it on that one.

pkoutsovasilis:
> @pkoutsovasilis I found the root cause of that bug. It has nothing to do with status reporting, we're actually running multiple extensions in the subprocess collector. The reason is that we don't make a copy of the collector config at any point, so we continue adding new healthcheckv2 extensions every time the collector is restarted.
> 
> #8529 fixes this by accident because it always creates its own config file. Up to you if we want to fix it separately first.
> 
> In any case, this PR fixes a different unrelated issue, and we shouldn't block it on that one.

okkk now I see, under the subprocess mode we mutate the cfg but the same cfg is always mutated. If we didn't generate a random UUID for the healthcheck extension that would be fine as well I guess. Agreed let's deal with that separately

pkoutsovasilis (approved):

PR #8728, opened 2025-06-28
[AutoOps] Add `autoops-es.yml` to Packages

pickypg:
This adds the `autoops-es.yml` to enable simpler configuration for users that would like to make use of AutoOps to ship data.

## What does this PR do?

Adds a configuration to all non-FIPS releases for AutoOps (AutoOps is untested in FIPS environments) so that users can run (directly or indirectly) `elastic-agent otel autoops-es.yml` with the supplied environment variables.

## Why is it important?

Rather than forcing users to provide their own configuration for what is a standard one, we can help them all by providing one source of truth.

## Checklist

## Disruptive User Impact

This only adds a configuration file. There is no disruption beyond an extra file for those that do not want it.

Run the OTel configuration:

```
./elastic-agent otel --config autoops-es.yml
```

## Related issues

- Relates https://github.com/elastic/opex-product/issues/475

### DIFF
--- a/autoops-es.yml
+++ b/autoops-es.yml
@@ -1,0 +1,43 @@
+receivers:
+  metricbeatreceiver:
+    metricbeat:
+      modules:
+        # Metrics
+        - module: autoops_es
+          hosts: ${env:AUTOOPS_ES_URL}
+          period: 10s
+          metricsets:
+            - cat_shards
+            - cluster_health
+            - cluster_settings
+            - license
+            - node_stats
+            - tasks_management
+        # Templates
+        - module: autoops_es
+          hosts: ${env:AUTOOPS_ES_URL}
+          period: 24h
+          metricsets:
+            - cat_template
+            - component_template
+            - index_template
+    processors:
+      - add_fields:
+          target: autoops_es
+          fields:
+            token2: ${env:AUTOOPS_TOKEN}
+    output:
+      otelconsumer:
+    telemetry_types: ["logs"]
+
+exporters:
+  otlphttp:
+    headers:
+      Authorization: "AutoOpsToken ${env:AUTOOPS_TOKEN}"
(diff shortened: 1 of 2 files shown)


ycombinator:
Hi @pickypg, it makes sense to include the OTel configuration for AutoOps in the Agent package. However, I'm wondering if we should place it under the `otel_samples/` folder instead of at the top-level.  We currently have a single "getting started" OTel configuration file at the top-level, `otel.yml` and other useful OTel configuration files under the `otel_samples/` folder.

@swiatekm do you have any opinions here?

swiatekm:
> Hi @pickypg, it makes sense to include the OTel configuration for AutoOps in the Agent package. However, I'm wondering if we should place it under the `otel_samples/` folder instead of at the top-level. We currently have a single "getting started" OTel configuration file at the top-level, `otel.yml` and other useful OTel configuration files under the `otel_samples/` folder.
> 
> @swiatekm do you have any opinions here?

`otel_samples/` sounds better to me as well.

@pickypg is this being done with the understanding that metricbeatreceiver is not even in technical preview, much less GA?

pickypg:
> `otel_samples/` sounds better to me as well.

I'm fine moving it there. This is meant to actually be used (as-is without manipulation outside of environment injection) rather than a "sample", but that's just a name.

> @pickypg is this being done with the understanding that metricbeatreceiver is not even in technical preview, much less GA?

Yep!

ycombinator (approved):
Packaging changes LGTM.  I built a Windows artifact and extracted it and the `autoops_es.yml` file showed up under `otel_samples/autoops_es.yml`.  

I did not run the pipeline though.  Either you could tell me how to set up AutoOps so I could set the `AUTOOPS_*` env vars for the pipeline to run OR you could test the pipeline by itself on Windows (since we know the packaging bit is working correctly).

PR #8737, opened 2025-06-30
Move beat receiver component logic to the otel manager

swiatekm:
NOTE: This is an alternative implementation of #8529, which adds the new functionality to the existing Otel manager, instead of creating a new object for it.

## What does this PR do?

Move beat receiver component logic to the otel manager.

This logic consists of two tasks conceptually:

* Translating agent configurations into beats receiver configurations for the otel collector.
* Translating otel collector statuses into component states.

Up until now, the logic for these was haphazardly spread across the agent coordinator. This PR moves all of it into the OtelManager, which can now run both raw otel collector configurations and agent components in a single otel collector instance.

The OtelManager now encapsulates all the logic involved in interfacing between the agent coordinator and the otel collector. In the near future, it will also take on additional responsibilities, like generating diagnostics for components it runs.

The only new logic this PR introduces lives in the new manager's main loop, and has to do with how updates and configurations are moved around. The rest is either existing logic moved to a new location, and new tests for that old logic.

## Checklist

## Related issues

- Relates #8208


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -149,15 +148,25 @@ type RuntimeManager interface {
...
 
-	// Update updates the current configuration for OTel.
-	Update(cfg *confmap.Conf)
+	// UpdateCollector updates the current plain configuration for the otel collector.
+	UpdateCollector(cfg *confmap.Conf)
 
-	// Watch returns the chanel to watch for configuration changes.
-	Watch() <-chan *status.AggregateStatus
+	// WatchCollector returns a channel to watch for collector status updates.
+	WatchCollector() <-chan *status.AggregateStatus
```
pkoutsovasilis, line 159: thinking out loud here, by renaming and making some of these funcs Collector explicit feels a little bit off from the end-goal we are after in the long run, which is to abstract out any otel collector specific logic and converge our codebase to support both the former and the pre-existing other runtimes. Thus if we keep the funcs of this interface as generic as possible we could always try and migrate this manager next to the other that is for "traditional" beats, etc. Again this comment is not a blocking one and I write it in a thinking out loud fashion 🙂

internal/pkg/otel/manager/manager.go
```diff
@@ -211,50 +255,41 @@ func (m *OTelManager) Run(ctx context.Context) error {
...
-					recoveryDelay := m.recoveryTimer.ResetInitial()
-					m.logger.Errorf("collector exited with error (will try to recover in %s): %v", recoveryDelay.String(), err)
-				} else {
-					// all good at the moment (possible that it will fail)
-					reportErr(ctx, m.errCh, nil)
-				}
+			err = m.applyMergedConfig(ctx, collectorStatusCh, collectorRunErr)
+			// report the error unconditionally to indicate that the config was applied
+			reportErr(ctx, m.errCh, err)
+		case componentModel := <-m.componentUpdateCh:
```
pkoutsovasilis: Why we need to channels that essentially affect the configuration of the edot collector? If I am tracing that correctly there is one call here

https://github.com/elastic/elastic-agent/blob/5da6654014cc87f74c7e4037288e4752468f74ba/internal/pkg/agent/application/coordinator/coordinator.go#L1442-L1445

and then there is this other
https://github.com/elastic/elastic-agent/blob/5da6654014cc87f74c7e4037288e4752468f74ba/internal/pkg/agent/application/coordinator/coordinator.go#L1652-L1655

which then we merge inside the manager by calling `m.updateMergedConfig()`. This seems a bit inefficient to me as we are gonna effectively restart the collector. Instead why don't we merge everything before calling `c.otelMgr.UpdateComponents` this could even be a helper function that lives inside the manager. This will simplify the config change logic of the manager. And if now you are thinking how to emit the stopped status for components that are not part of the new config, if this is something that we have to do, I would propose before applying the new config and restarting the otel collector see which components are not part of the new config and emit the appropriate status by calling `m.reportComponentStateUpdates`. wdyt?

internal/pkg/otel/manager/manager.go
```diff
@@ -211,50 +255,41 @@ func (m *OTelManager) Run(ctx context.Context) error {
...
+			err = m.applyMergedConfig(ctx, collectorStatusCh, collectorRunErr)
+			// report the error unconditionally to indicate that the config was applied
+			reportErr(ctx, m.errCh, err)
+		case otelStatus := <-collectorStatusCh:
+			componentUpdates, err := m.handleOtelStatusUpdate(otelStatus)
+			if err != nil {
+				reportErr(ctx, m.errCh, err)
+			}
+			reportCollectorStatus(ctx, m.collectorStatusCh, m.currentCollectorStatus)
+			m.reportComponentStateUpdates(ctx, componentUpdates)
```
pkoutsovasilis: I think that we shouldn't have two separate status groups here, one for the collector and another for components. IMO, the collector statuses are kinda hierarchical higher than the components, in simple words, if a collector is deemed unhealthy then the components cannot/shouldn't be reported healthy, does that make sense? In general I would expect the manager to report only component statuses, `in` goes otel config `out` comes component statuses. One way to approach this, which will also help us in not using buffered channels like  `statusUpdateChannelSize`, is to have a status status equivalent of `collectorRunErr`, eg. `collectorStatus` scoped solely within the `Run` func (buffered channel of 1). This can be used exactly in the same fashion as `collectorRunErr` we do care only about the latest so drain before writing, and inside the manager run loop we get one more case reading from this channel where you do the transformation from `status.AggregateStatus` to component statuses. Now if you are thinking how to emit stopped statuses, when you get a `collectorRunErr` this means that the collector is no longer running and thus nothing should write to `collectorStatus`, so you can easily emit from there the component statuses

internal/pkg/otel/manager/manager.go
```diff
@@ -264,20 +299,221 @@ func (m *OTelManager) Errors() <-chan error {
...
+
+// updateMergedConfig builds the merged configuration for the otel manager by merging the base collector configuration
+// with the component configuration, and updates the otel manager with the merged configuration.
+func (m *OTelManager) updateMergedConfig() error {
+	mergedCfg, err := m.buildMergedConfig()
+	if err != nil {
+		return err
+	}
+
+	m.collectorConfigMutex.Lock()
```
pkoutsovasilis: no need for a mutex here, this can be `atomic.Pointer`?

internal/pkg/otel/manager/manager.go
```diff
@@ -264,20 +299,221 @@ func (m *OTelManager) Errors() <-chan error {
...
+	defer m.collectorConfigMutex.Unlock()
+	m.mergedCollectorCfg = mergedCfg
+	return nil
+}
+
+func (m *OTelManager) applyMergedConfig(ctx context.Context, collectorStatusCh chan *status.AggregateStatus, collectorRunErr chan error) error {
+	if m.proc != nil {
```
pkoutsovasilis, line 355: I am not a fun of proc getting migrated to the OtelManager struct, since this is purely scoped to the Run function. So how about making this a private function where you pass the manager and proc as parameters? 🙂

pkoutsovasilis (commented):
I did left some comments, @swiatekm take a look and tell me what you think. PS: I am more than happy to have a dedicated sync about that 🙂

internal/pkg/agent/application/coordinator/coordinator.go
swiatekm: Hm, I could make this function return both statuses. Realistically, we always have both of them, because they're sourced from the same underlying collector status. Would that help?

internal/pkg/otel/manager/manager.go
swiatekm: The real reason this merge needs to happen in the manager is that the manager needs to know which components it's supposed to be running, not just to emit correct statuses, but also diagnostics. It's possible to infer the components from the otel configuration to an extent, but I feel like that would be unnecessarily error-prone.

How about we have a single update function that takes both the collector config and components? The caller would need to be careful to always include both, but I don't think that's a big deal. WDYT?

```
Update(collectorCfg *confmap.Conf, components []component.Component)
```

internal/pkg/otel/manager/manager.go
swiatekm: I tried doing it that way and it was very awkward. There's a lot of return values that have different meaning depending on what the function ended up doing. I'll make a change in a dedicated commit so you can see for yourself.

internal/pkg/otel/manager/manager.go
swiatekm: But I need to emit the collector status, because I might have otel components from hybrid mode, and not from the component model. Unless I'm misunderstanding your point here.

internal/pkg/agent/application/coordinator/coordinator.go
pkoutsovasilis: After our sync, I do see that keeping this otel manager as close as possible the other runtime manager (too many managers 😄 ) is something that requires more thinking and most probably will be a non-negligible effort to happen. Thus no need to think that in this PR 🙂

internal/pkg/otel/manager/manager.go
swiatekm: This mutex will be used to protect other values in the future. I'll change the name to be more generic.

internal/pkg/otel/manager/manager.go
swiatekm: Here's how the change looks: https://github.com/elastic/elastic-agent/pull/8737/commits/2edcb08e76276162a0bb8d4aad7df36f1adf71b1. If you're happy with that, we can leave it as-is. I'd like to revisit this in the future though, I don't like how the requirement for the last status to be `nil` is making the logic more complicated than it needs to be.

internal/pkg/otel/manager/manager.go
swiatekm: Changed to use a single function here: https://github.com/elastic/elastic-agent/pull/8737/commits/9b4d73e1155e8fe0d305c5499d32d46ea12cb3d3

internal/pkg/otel/manager/manager.go
swiatekm: Renamed it and described its purpose in a comment: https://github.com/elastic/elastic-agent/pull/8737/commits/086588b45558a587a9b366f5271f70c22e8e41d3

internal/pkg/otel/manager/manager.go
swiatekm: And a small follow-up: https://github.com/elastic/elastic-agent/pull/8737/commits/f76030a1cf849485e129e3662acbf24fe22de4d6

internal/pkg/otel/manager/manager_test.go
```diff
@@ -404,7 +470,7 @@ func TestOTelManager_Run(t *testing.T) {
 				//
 				// a nil error just means that the collector is trying to restart
 				// which clears the error on the restart loop
-				timeoutCh := time.After(time.Second * 5)
+				timeoutCh := time.After(time.Minute * 5)
```
pkoutsovasilis: sorry I just spotted this one; do we actually need to bump that time.After to the scale of minutes or is this something that slipped while you were developing/updating the tests here? 🙂

swiatekm: Good catch, definitely a debugging change.

swiatekm: Reverted.

pkoutsovasilis (commented):
thanks for addressing my comments @swiatekm. As you have already mentioned, there are points that we should revisit in the future but this PR feels to me a step towards the right direction:
* the final otel config is produced inside the otel manager
* statuses are fabricated and emitted from the manager. A future improvement here is to investigate if we could emit only one status and satisfy both the collector and components status needs, which probably will help in simplifying the code
* sure the manager loop got a little bit "heavier" but we can try in the future to improve code readability

So code changes wise this counts as an improvement. Now that said, I think we have an issue 😄 

I did compile and run this on my machine and for both execution modes I see only the following reported through `elastic-agent status --output full`

```
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: c3189bbf-0e33-4b6b-aab7-80a4a3a0647b
   │  ├─ version: 9.2.0
   │  └─ commit: 86fc4fcea1209c63a3b810f941449b743bc0b440
   └─ system/metrics-default
      ├─ status: (STARTING) Starting: spawned pid '1284'
      ├─ system/metrics-default
      │  ├─ status: (STARTING) Starting: spawned pid '1284'
      │  └─ type: OUTPUT
      └─ system/metrics-default-unique-system-metrics-input
         ├─ status: (STARTING) Starting: spawned pid '1284'
         └─ type: INPUT
```

which makes me believe that there is a tiny bit somewhere in the status channels that is missed!? but is this reproducible also on your end?

swiatekm:
> thanks for addressing my comments @swiatekm. As you have already mentioned, there are points that we should revisit in the future but this PR feels to me a step towards the right direction:
> 
>     * the final otel config is produced inside the otel manager
> 
>     * statuses are fabricated and emitted from the manager. A future improvement here is to investigate if we could emit only one status and satisfy both the collector and components status needs, which probably will help in simplifying the code
> 
>     * sure the manager loop got a little bit "heavier" but we can try in the future to improve code readability
> 
> 
> So code changes wise this counts as an improvement. Now that said, I think we have an issue 😄
> 
> I did compile and run this on my machine and for both execution modes I see only the following reported through `elastic-agent status --output full`
> 
> ```
> ┌─ fleet
> └─ elastic-agent
>    ├─ info
>    │  ├─ id: c3189bbf-0e33-4b6b-aab7-80a4a3a0647b
>    │  ├─ version: 9.2.0
>    │  └─ commit: 86fc4fcea1209c63a3b810f941449b743bc0b440
>    └─ system/metrics-default
>       ├─ status: (STARTING) Starting: spawned pid '1284'
>       ├─ system/metrics-default
>       │  ├─ status: (STARTING) Starting: spawned pid '1284'
>       │  └─ type: OUTPUT
>       └─ system/metrics-default-unique-system-metrics-input
>          ├─ status: (STARTING) Starting: spawned pid '1284'
>          └─ type: INPUT
> ```
> 
> which makes me believe that there is a tiny bit somewhere in the status channels that is missed!? but is this reproducible also on your end?

Can you post your exact configuration? I tested 

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    username: "elastic"
    password: "..."

agent:
  logging:
    to_stderr: true
  monitoring:
    enabled: false

inputs:
- data_stream:
    namespace: default
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  - data_stream:
      dataset: system.memory
    metricsets:
    - memory
  - data_stream:
      dataset: system.network
    metricsets:
    - network
  - data_stream:
      dataset: system.filesystem
    metricsets:
    - filesystem
  type: system/metrics
  use_output: default
  _runtime_experimental: otel
```

and got:

```
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: fd4b48da-1a8a-42f3-aca2-321df32b45d7
   │  ├─ version: 9.2.0
   │  └─ commit: 86fc4fcea1209c63a3b810f941449b743bc0b440
   └─ system/metrics-default
      ├─ status: (HEALTHY) HEALTHY
      ├─ system/metrics-default
      │  ├─ status: (HEALTHY) Healthy
      │  └─ type: OUTPUT
      └─ system/metrics-default-unique-system-metrics-input
         ├─ status: (HEALTHY) Healthy
         └─ type: INPUT
```

and if I switch back to process mode for the system/metrics input:

```
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: fd4b48da-1a8a-42f3-aca2-321df32b45d7
   │  ├─ version: 9.2.0
   │  └─ commit: 86fc4fcea1209c63a3b810f941449b743bc0b440
   └─ system/metrics-default
      ├─ status: (HEALTHY) Healthy: communicating with pid '517716'
      ├─ system/metrics-default
      │  ├─ status: (HEALTHY) Healthy
      │  └─ type: OUTPUT
      └─ system/metrics-default-unique-system-metrics-input
         ├─ status: (HEALTHY) Healthy
         └─ type: INPUT
```

pkoutsovasilis:
@swiatekm 👋  this is my config as reported from the inspect sub-command

```
agent:
  logging:
    to_stderr: true
  monitoring:
    _runtime_experimental: otel
    enabled: true
inputs:
- data_stream:
    namespace: default
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  - data_stream:
      dataset: system.memory
    metricsets:
    - memory
  - data_stream:
      dataset: system.network
    metricsets:
    - network
  - data_stream:
      dataset: system.filesystem
    metricsets:
    - filesystem
  type: system/metrics
  use_output: default
outputs:
  default:
    api_key: <REDACTED>
    hosts:
    - 127.0.0.1:9200
    preset: balanced
    type: elasticsearch
```

for the subprocess approach I would expect to see something like this (PS: I get the same for the in-process approach without the `extensions` section)

```
┌─ fleet
└─ elastic-agent
   ├─ info
   │  ├─ id: f24c5572-89c7-450e-87b8-14cfd8741a83
   │  ├─ version: 9.2.0
   │  └─ commit: 86fc4fcea1209c63a3b810f941449b743bc0b440
   ├─ beat/metrics-monitoring
   │  ├─ beat/metrics-monitoring
   │  │  ├─ status: (DEGRADED) Elasticsearch request failed: dial tcp 127.0.0.1:9200: connect: connection refused
   │  │  └─ type: OUTPUT
   │     └─ type: INPUT
   ├─ filestream-monitoring
   │  ├─ filestream-monitoring
   │  │  ├─ status: (DEGRADED) Elasticsearch request failed: dial tcp 127.0.0.1:9200: connect: connection refused
   │  │  └─ type: OUTPUT
   │  └─ filestream-monitoring-filestream-monitoring-agent
   │     └─ type: INPUT
   ├─ http/metrics-monitoring
   │  ├─ http/metrics-monitoring
   │  │  ├─ status: (DEGRADED) Elasticsearch request failed: dial tcp 127.0.0.1:9200: connect: connection refused
   │  │  └─ type: OUTPUT
   │  └─ http/metrics-monitoring-metrics-monitoring-agent
   │     └─ type: INPUT
   ├─ system/metrics-default
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '67276'
   │  ├─ system/metrics-default
   │  │  └─ type: OUTPUT
   │     └─ type: INPUT
   └─ extensions
      ├─ status: StatusOK
      ├─ extension:healthcheckv2/a8aa1840-f3cd-4464-9b0b-68251f8ce626
      │  └─ status: StatusOK
```

The interesting thing here is that after some minutes (not sure how many) these do appear in the output even with the code from this PR. I need to re-run elastic-agent from `main` I have the impression that these appear in the output way faster

pkoutsovasilis (approved):
I did run again the code of this PR again with both execution modes of the collector (in-process and sub-process) and now all according statuses appear in-time through `elastic-agent status --output full` so I am gonna choose to unravel this mystery by saying that my previous attempt was on Friday and probably I missed something 🙂 

LGTM

PR #8753, opened 2025-06-30
EDOT collector: include the forward connector.

cmacknz:
https://github.com/open-telemetry/opentelemetry-collector/tree/main/connector/forwardconnector

Include the forwardconnector, which we are highly likely to need for OTel native integrations to allow for unique processors per receiver without having to duplicate pipelines with exporters.

To test this PR I used the following configuration showing how to have 2 filelog receivers that could have separate processors applied but export via the same exporter.

```yaml
receivers:
  filelog:
    include: [ /var/log/system.log ]
    start_at: beginning
  filelog/2:
    include: [ /var/log/wifi.log ]
    start_at: beginning

processors:
  resource:
    attributes:
    - key: service.name
      action: insert
      value: elastic-otel-test

exporters:
  debug:
    verbosity: detailed
    sampling_initial: 10000
    sampling_thereafter: 10000

connectors:
  forward:

extensions:
  memory_limiter:
    check_interval: 1s
    limit_mib: 700
    spike_limit_mib: 180
  health_check:
    endpoint: "localhost:13133"
    path: "/health/status"

service:
  pipelines:
    logs/system:
      receivers: [filelog]
      processors: [resource]
      exporters: [forward]
    logs/wifi:
      receivers: [filelog/2]
      processors: [resource]
      exporters: [forward]
    logs/export:
      receivers: [forward]
      exporters: [debug]
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,6 +79,7 @@ require (
 	go.elastic.co/ecszap v1.0.3
 	go.elastic.co/go-licence-detector v0.7.0
 	go.opentelemetry.io/collector/component/componentstatus v0.127.0
+	go.opentelemetry.io/collector/connector/forwardconnector v0.127.0
 	go.opentelemetry.io/collector/pipeline v0.127.0
 	go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.127.0
 	go.opentelemetry.io/collector/receiver/nopreceiver v0.127.0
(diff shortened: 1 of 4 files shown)


pkoutsovasilis (commented):
Indeed forwardconnector seems that it does the trick, so LGTM. Only one minor thingie @cmacknz please run `mage fmt` and repush 😄

pchila (commented):
Code looks ok, I wonder if the new component should be listed in `internal/pkg/otel/core-components.yaml` as introduced by PR #8530

Holding off approval till CI is green

cmacknz:
Core components means users can open SDHs against them and we'll try to fix them. I don't think we we are at the point where we need that for the foward connector as we don't depend on it for anything yet.

swiatekm:
@cmacknz could you add this component to this [file](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/testdata/all-components.yml)? That way we can be confident it's actually possible to start a collector with it included in a pipeline.

cmacknz:
> @cmacknz could you add this component to this [file](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/testdata/all-components.yml)? That way we can be confident it's actually possible to start a collector with it included in a pipeline.

Done in https://github.com/elastic/elastic-agent/pull/8753/commits/1d856a2d51bc35440ac71bfa7e322fcca67bae7c

PR #8766, opened 2025-07-01
Add ability for local output configuration to add to policy configuration

blakerouse:
## What does this PR do?

Adds the ability for the local output configuration to add extra options to policy configuration.

## Why is it important?

This allows some local private settings to be provided by the local configuration and not be viewable in the policy in Fleet.

## Checklist

## Disruptive User Impact

None

1. Enroll the Elastic Agent into Fleet
2. Adjust the `elastic-agent.yml` to have an `outputs` with a matching output name and setting.
3. Observe that the setting is applied on that Elastic Agent for that output with the exact same name.


internal/pkg/agent/application/application.go
```diff
@@ -305,3 +307,70 @@ func mergeFleetConfig(ctx context.Context, rawConfig *config.Config) (storage.St
...
+	if len(parsed.Outputs) == 0 {
+		return noop
+	}
+
+	return func(change coordinator.ConfigChange) coordinator.ConfigChange {
+		cfg := change.Config()
+		outputs, err := cfg.Agent.Child("outputs", -1)
+		if err != nil {
+			// no outputs in configuration; do nothing
+			return change
```
pkoutsovasilis, line 342: this err can essentially be `ErrMissing` or a conversion one, so how about maintain the return for the former, but for the latter logging it before returning?

internal/pkg/agent/application/application.go
```diff
@@ -305,3 +307,70 @@ func mergeFleetConfig(ctx context.Context, rawConfig *config.Config) (storage.St
 
 	return store, cfg, nil
 }
+
+// injectOutputOverrides takes local configuration for specific outputs and applies them to the configuration.
+//
+// The name of the output must match or no options will be overwritten.
+func injectOutputOverrides(log *logger.Logger, rawConfig *config.Config) func(change coordinator.ConfigChange) coordinator.ConfigChange {
```
pkoutsovasilis, line 314: if we indeed patch an output, could we also log it with info level just to make it easier to capture this event from agent-logs? Also, although I think this is happening already just to be safe I am gonna ask; we are gonna get the patched output in the agent diagnostics right? 🙂

internal/pkg/agent/application/application.go
blakerouse: I added the log message for non-`ErrMissing`. Sadly it requires a specific function to check as the `go-ucfg` doesn't support the `Unwrap` interface.

Filed an issue here so we can look at getting this fixed: https://github.com/elastic/go-ucfg/issues/212

internal/pkg/agent/application/application.go
blakerouse: I added the info log.

Yes the patched output will be included in the diagnostics because the patch happens before the configuration gets into the coordinator and the diagnostics information about configuration comes from the coordinator.

blakerouse:
@pkoutsovasilis Thanks for the review. I have updated the code with your suggestions and responded inline.

pkoutsovasilis:
> @pkoutsovasilis Thanks for the review. I have updated the code with your suggestions and responded inline.

Thank you for addressing my comments @blakerouse and for filing an issue on ucfg. Should we deal with the linter errors, I think they are valid, wdyt? 🙂

blakerouse:
@pkoutsovasilis thanks!

PR #8767, opened 2025-07-01
Elastic Agent upgrade: lock-free manual rollback

### LINKED ISSUE #6880: Add 'rollback_available' parameter when launching elastic-agent watcher before starting new agent version

### LINKED ISSUE #6887: Add `--rollback` option to `elastic-agent` upgrade subcommand
`elastic-agent upgrade` [command](https://github.com/elastic/elastic-agent/blob/293092ca9b14a349874166e39baecbf99aaa2f15/internal/pkg/agent/cmd/upgrade.go#L85-L113) will need to have a `--rollback` option to initiate rollbacks of standalone agents.

This new option needs to be translated in a new parameter in the upgrade request sent to the GRPC server.
(hint: a boolean similar to the one added to the upgrade action described in elastic/fleet-server#4838 would simplify things and keep a consistent interface)

### LINKED ISSUE #6890: Allow elastic-agent rollback in `UPG_WATCHING` state

pchila:
## What does this PR do?
This PR introduces a new `--rollback` flag to `elastic-agent upgrade` command that will switch back to a previous elastic-agent installation, effectively rolling back an upgrade.
This PR makes the elastic agent main process "take over" the watcher applocker to write the rollback request before running the watcher again, which will perform the rollback.
The Upgrader now receives a `rollback` boolean which will trigger:
1. a takeover of the watcher applocker
2. it will extract the versionedHome of the version we want to rollback to
3. launch a new `elastic-agent watch --rollback `command which will take care of the actual rollback, restart and cleanup of the agent

On the watcher side, the watch loop is now interruptible  to be able to have the watcher exit gracefully.

This PR also changes the way the watcher process is launched and signaled on Windows, needed to implement any sort of graceful shutdown: the gist of it is that now the watcher processes are always launched while holding a Windows console (even when the elastic-agent service does not have one). The watcher processes can be terminated gracefully by using `elastic-agent watch --takedown` that will connect to the watcher console and send a `Ctrl+Break` event, cutting short a watch operation without side effect.

## Why is it important?

This is the initial implementation for a manual rollback flow.
In this first iteration, the rollback only works during the grace period (that is while the watcher is still running), in follow-up PRs this functionality will be extended for the duration of `agent.upgrade.rollback.window`.

## Checklist

1. Package elastic agent twice from this PR:
  ```shell
  SNAPSHOT=true EXTERNAL=true PACKAGES=tar.gz  PLATFORMS="linux/amd64" mage -v package
  ```
  ```shell
  AGENT_PACKAGE_VERSION="9.2.0+20250701000000" SNAPSHOT=true EXTERNAL=true PACKAGES=tar.gz  PLATFORMS="linux/amd64" mage -v package
  ```
2. Install the version `9.2.0-SNAPSHOT` as usual
3. Set a rollback window duration > 0, for example 2 hours:
  ```yaml
    agent.upgrade:
      rollback:
        window: 2h
  ```
4. Trigger an update to the other package (saved on disk):
  ```
  elastic-agent upgrade --skip-verify --source-uri=file:///vagrant/build/distributions 9.2.0+20250701000000-SNAPSHOT
  ```
5. Wait for the new agent to come online and the upgrade details to signal UPG_WATCHING state
6. Manually rollback to the previous version: 
  ```shell
  elastic-agent upgrade --rollback 9.2.0-SNAPSHOT
  ```

Notes:
- Trying to rollback after the grace period does not work and may break the agent install (this is because the watcher is still cleaning up the upgrade marker and the previous install at the end of the grace period)

## Related issues

- Closes #6880
- Closes #6887 
- Closes #6890

## Questions to ask yourself

- ...


internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -377,6 +384,108 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
+
+				// this should be run continuously and concurrently to attempting to get the app locker
+				for _, pid := range pids {
+					u.log.Debugf("attempting to kill watcher process with PID: %d", pid)
+					process, findProcErr := os.FindProcess(pid)
+					if findProcErr != nil {
+						u.log.Errorf("error finding process with PID: %d: %s", pid, findProcErr)
+						continue
+					}
+					killProcErr := process.Kill()
```
cmacknz: I think this is dangerous without coordination between the agent and the watcher to make sure we don't kill the watcher while it is in the process of rolling back the currently running agent. 

I still think if the watcher is running, we could communicate the need to rollback via the `StateWatch` it already has on this agent instead of having to kill it:

https://github.com/elastic/elastic-agent/blob/676e18de4b410a752cb9cef0de44918c3624ac74/internal/pkg/agent/application/upgrade/watcher.go#L157

Is there a reason that wouldn't work? It seems like it is a safer way to trigger this to me.

cmacknz: The edge case to this approach is if the grace period expires as you indicate a rollback should happen, the watcher might exit before it can see the need to rollback. 

That is a nicer edge case than the agent installation being broken, which is my main worry with unconditionally killing the watcher.

pchila: The `Kill()` is the PoC part here, the point to bring home is that the elastic agent main process should take over *any* watcher process that is competing over the appLocker.
This is needed to remove concurrent writes to the upgrade marker: the actual rollback will start later when a new watcher is launched.
The `StateWatch` usage that you are proposing would just make the watcher exit or it would trigger an immediate rollback (driven by the watcher) ?

cmacknz: > The StateWatch usage that you are proposing would just make the watcher exit or it would trigger an immediate rollback (driven by the watcher) ?

It could be either of these (probably immediate rollback is simpler?), the main point is to avoid a situation where we kill the watcher while it is rolling back we need a graceful shutdown mechanism or a way to avoid having to shut down the watcher to trigger a rollback.

cmacknz: One thing I remember about file locks on Windows is that they are not always released immediately.

[On windows the lock is implemented with](https://github.com/gofrs/flock/blob/12753ea298e1aeb97f0881ff3fc07eabcb2b86e5/flock_windows.go#L68C17-L68C27) [LockFileEx](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex)

> If a process terminates with a portion of a file locked or closes a file that has outstanding locks, the locks are unlocked by the operating system. However, the time it takes for the operating system to unlock these locks depends upon available system resources. Therefore, it is recommended that your process explicitly unlock all files it has locked when it terminates. If this is not done, access to these files may be denied if the operating system has not yet unlocked them.

So we need to be careful with locks in the case where the program can exit unexpectedly (e.g. panic). The watcher is simple enough that this is pretty unlikely and I don't think we've seen it before. We did observe the Beats having problems with this though in the past as they are more likely to panic or be OOMKilled.

pchila: @cmacknz  Tried using SIGTERM and SIGINT (for the windows part) in 18c1804c2818efd8fcd180e2f617bda125536b83 but we receive those signals when the agent starting the upgrade rexecs, which then brings the watcher to terminate the watch immediately and cleanup :disappointed: 

We need less commonly used signals or an entirely different way of making the watcher shutdown gracefully (without writing concurrently files)

cmacknz: CTRL_BREAK might do it https://github.com/elastic/elastic-agent/pull/7738, though in the receiving process Go will translate it into SIGINT so maybe not.

I think my main worry about the use of locks is one of the processes being killed, not properly releasing the lock, and dooming future upgrades.

Do we need to kill the watcher? We could use the state watch to communicate the intent to rollback, and then have the agent ensure the watcher is always running when a rolback is request but hasn't been performed yet (e.g. if the watcher exits after the grace period because of perfect timeing it just gets started again). We could also look at using a separate socket/named pipe to have agent tell the watcher to rollback.

pchila: @cmacknz, yesterday looking at the PR with @pkoutsovasilis we found out the source of the SIGINT signal when re-executing: it was explicitly set as `Pdeathsig` signal when launching watcher
https://github.com/elastic/elastic-agent/pull/8767/commits/4f73814fc1ee402021dfb4c3025e2d25a440082e#diff-168080314caf1d3868d593889dc36edec0ec2f12fd50e37d5a3d57e05274a10aL41

Removing that restored watching functionality on agent restart

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -377,50 +411,132 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
...
 
-func selectWatcherExecutable(topDir string, previous agentInstall, current agentInstall) string {
-	// check if the upgraded version is less than the previous (currently installed) version
-	if current.parsedVersion.Less(*previous.parsedVersion) {
-		// use the current agent executable for watch, if downgrading the old agent doesn't understand the current agent's path structure.
-		return paths.BinaryPath(filepath.Join(topDir, previous.versionedHome), agentName)
-	} else {
-		// use the new agent executable as it should be able to parse the new update marker
-		return paths.BinaryPath(filepath.Join(topDir, current.versionedHome), agentName)
+func (u *Upgrader) rollbackToPreviousVersion(ctx context.Context, topDir string, now time.Time, version string, action *fleetapi.ActionUpgrade) (reexec.ShutdownCallbackFn, error) {
```
ycombinator, line 474: Just want to clarify one thing: is `version` here the version the user was upgrading to before deciding to rollback? Or is it the version that the user wants to rollback to?  Reading the code in the function, I believe it's the latter but can you confirm?

ycombinator: Also, I believe the `action` parameter is not being used in this function?

pchila: The version is the version the user wants to rollback to. The code checks that the specific version is available as a valid rollback.

The action is unused at the moment because rolling back from Fleet is not yet implemented in this PR, it will be used to ack the Fleet action when the manual rollback for managed agent will be implemented.

internal/pkg/agent/application/upgrade/rollback_windows.go
```diff
@@ -19,11 +26,90 @@ const (
...
+	}
 	return cmd
 }
+
+func StartWatcherCmd(log *logger.Logger, createCmd cmdFactory, opts ...WatcherInvocationOpt) (*exec.Cmd, error) {
+
+	invocationOpts := applyWatcherInvocationOpts(opts...)
+
+	// allocConsole
+	r1, _, consoleErr := allocConsoleProc.Call()
```
cmacknz, line 59: Is there any need to cleanup after this call to avoid resource leaks?

https://learn.microsoft.com/en-us/windows/console/freeconsole has a name that indicates it does this, but that's not really what it seems to do.

I can't find documentation suggesting we need to, but this has been a source of memory leaks in the past so I wanted to double check.

ycombinator:
```
$ sudo elastic-agent upgrade --rollback 9.2.0-SNAPSHOT
Upgrade triggered to version 9.2.0-SNAPSHOT, Elastic Agent is currently restarting

$ sudo elastic-agent status
┌─ fleet
├─ elastic-agent
└─ upgrade_details
   ├─ target_version: 9.2.0+20250701000000
   ├─ state: UPG_ROLLBACK
   └─ metadata
      └─ reason: manual rollback requested
```

Two UX suggestions:
1. Could we make the first command's output say "Rollback triggered to version ..."?
2. Could we include the version we're rolling back to in the `upgrade_details.metadata.reason`?

internal/pkg/agent/cmd/watch_notwindows.go
```diff
@@ -0,0 +1,62 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
ycombinator: Nit: I think the convention for naming such files is `*_other.go`.

ycombinator:
Looking at the `rollbackToPreviousVersion` method of the `Upgrader`, it eventually ends up calling the `TakeDownWatcher` method of `commandWatcherGrappler`.  This ends up spawning a new `elastic-agent watch ... --takedown` process.  This process finds Watcher processes (that are not itself) and kills them.  I'm trying to understand why this second process was necessary to find and kill the Watcher processes?  Couldn't the main Agent process — the one where `rollbackToPreviousVersion` was called — itself find and kill Watcher processes? Does this have something to do with needing a Windows console?

internal/pkg/agent/application/upgrade/rollback_windows.go
pchila: We don't want to free this here, the allocated console will be used by the subprocess.
Consoles are shared by multiple processes so freeing it only signals that a process wants to detach from it, not that it will be necessarily freed.
https://learn.microsoft.com/en-us/windows/console/allocconsole#remarks
https://learn.microsoft.com/en-us/windows/console/freeconsole#remarks

In this specific case we are creating a new console just before launching an `elastic-agent watch` subprocess so that the watcher subprocess will inherit a console on which it will be able to receive the CTRL+BREAK signals.
The parent process detaches from the console in line [79](https://github.com/elastic/elastic-agent/pull/8767/files/5d6d7f47ed86ab75edc962d0453ae1e3057f93a6#diff-8c3df0ba5c4d84f3a44c643191470c63350a5d477abed36a8ece1f9dbaab4af7R78-R82) (this is done to be able to return the error in case detaching fails) but it could as easily be transformed into a `defer` statement and just log the error if detaching fails.

pchila:
> Looking at the `rollbackToPreviousVersion` method of the `Upgrader`, it eventually ends up calling the `TakeDownWatcher` method of `commandWatcherGrappler`. This ends up spawning a new `elastic-agent watch ... --takedown` process. This process finds Watcher processes (that are not itself) and kills them. I'm trying to understand why this second process was necessary to find and kill the Watcher processes? Couldn't the main Agent process — the one where `rollbackToPreviousVersion` was called — itself find and kill Watcher processes? Does this have something to do with needing a Windows console?

@ycombinator 
yes, signaling the watcher process needs a console on windows which the agent main process does not receive when running as a service (it can be allocated using https://learn.microsoft.com/en-us/windows/console/allocconsole though).

I experimented a bit with a `watcherGrappler` implementation that would work in-process  https://github.com/elastic/elastic-agent/pull/8767/commits/8e401cd3167ed5faac1134ac8510d095dbb1799c#diff-00f6b46ac72c0cdfa5f216d9b81dbfab8e70afb0eff043f96719244e9ca3e19fR354

In-process implementation is less isolated than having a separate watch command  and requires sharing a "contract" between the main process and the watch process regarding [how to take down](https://github.com/elastic/elastic-agent/pull/8767/commits/8e401cd3167ed5faac1134ac8510d095dbb1799c#diff-54f33d64f964a5e5228506fe1464465b9a74b2557165afa60291122847c05ff9R95-R133) a watcher process while having everything in the `watch` subcommand allows the main process to abstract away from the details on how to stop a watcher process.

Both solutions are feasible, I tried to go for the one that required fewer assumptions but we can discuss pros and cons of each approach

pchila:
> ```
> $ sudo elastic-agent upgrade --rollback 9.2.0-SNAPSHOT
> Upgrade triggered to version 9.2.0-SNAPSHOT, Elastic Agent is currently restarting
> 
> $ sudo elastic-agent status
> ┌─ fleet
> ├─ elastic-agent
> │  └─ status: (HEALTHY) Running
> └─ upgrade_details
>    ├─ target_version: 9.2.0+20250701000000
>    ├─ state: UPG_ROLLBACK
>    └─ metadata
>       └─ reason: manual rollback requested
> ```
> 
> Two UX suggestions:
> 
>     1. Could we make the first command's output say "Rollback triggered to version ..."?
> 
We could, it would mean changing this [fmt.Fprintf](https://github.com/elastic/elastic-agent/blob/623b65bbe21ea0d3eeabd0e22717428f6e6b930e/internal/pkg/agent/cmd/upgrade.go#L236) in the `upgrade` command depending on the `rollback` flag. Maybe we should change also other prints ?

>     2. Could we include the version we're rolling back to in the `upgrade_details.metadata.reason`?

The reason string is there also to allow Fleet to distinguish an automatic rollback (which may be retried) from a manual one. It's not a big change to add the version in the string but I don't know if that would make Fleet UI detection more difficult. @juliaElastic any preference/thoughts ?

internal/pkg/agent/cmd/watch_notwindows.go
pchila: The [go cmd doc](https://pkg.go.dev/cmd/go#hdr-Build_constraints) doesn't give any indications for that case and I came up with that name to match the `//go:build !windows` expression.
It can be renamed if it clashes with the name conventions already in the repo.

ycombinator: Yeah, it's just convention I've seen used in our repos.  I'd prefer if we went with `_other.go` so we're consistent, that's all.  Not a big deal either way, though.

ycombinator:
> We could, it would mean changing this [fmt.Fprintf](https://github.com/elastic/elastic-agent/blob/623b65bbe21ea0d3eeabd0e22717428f6e6b930e/internal/pkg/agent/cmd/upgrade.go#L236) in the upgrade command depending on the rollback flag. Maybe we should change also other prints ?

I think now that _manual, i.e. user-initiated_ rollbacks are a thing, the UX messaging becomes more important.  So I would be in favor of clearly calling out upgrades vs. rollbacks in any user-facing messaging.

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -35,6 +35,43 @@ const (
 
 // Rollback rollbacks to previous version which was functioning before upgrade.
 func Rollback(ctx context.Context, log *logger.Logger, c client.Client, topDirPath, prevVersionedHome, prevHash string) error {
+	return RollbackWithOpts(ctx, log, c, topDirPath, prevVersionedHome, prevHash)
+}
+
+var FatalRollbackError = errors.New("Fatal rollback error")
```
michel-laterman: (nitpick) go errors should have lower cases

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -56,12 +93,34 @@ func Rollback(ctx context.Context, log *logger.Logger, c client.Client, topDirPa
...
 
+	// Hook
+	if settings.preRestartHook != nil {
+		hookErr := settings.preRestartHook(ctx, log, topDirPath)
+		if hookErr != nil {
+			if errors.Is(hookErr, FatalRollbackError) {
+				return fmt.Errorf("pre-restart hook failed: %w", hookErr)
+			} else {
+				log.Warnf("pre-restart hook failed: %v", hookErr)
+			}
```
michel-laterman:
```suggestion
			}
			log.Warnf("pre-restart hook failed: %v", hookErr)
```
the if condition returns, we can get rid of the else

internal/pkg/agent/application/upgrade/rollback_windows.go
```diff
@@ -19,11 +26,90 @@ const (
...
+		// process group.
+		//
+		// Creating a child with CREATE_NEW_PROCESS_GROUP disables CTLR_C_EVENT
+		// handling for the child, so the only way to gracefully stop it is with
+		// a CTRL_BREAK_EVENT signal.
+		// https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags
+		//
+		// Watcher process will also need a console in order to receive CTRL_BREAK_EVENT on windows.
+		// Elastic Agent main process running as a service does not have a console allocated and the watcher process will also
+		// outlive its parent during an upgrade operation so we add the CREATE_NEW_CONSOLE flag.
```
michel-laterman: I really like that this comment block is here and explains why we need to pass these flags. But I don't see a `CREATE_NEW_CONSOLE` flag, are we using the flag? or is the console allocated through `allocConsoleProc.Call()` in `StartWatcherCmd`?

internal/pkg/agent/application/upgrade/rollback_windows.go
```diff
@@ -19,11 +26,90 @@ const (
...
+	invocationOpts := applyWatcherInvocationOpts(opts...)
+
+	// allocConsole
+	r1, _, consoleErr := allocConsoleProc.Call()
+	if r1 == 0 {
+		if !errors.Is(consoleErr, windows.ERROR_ACCESS_DENIED) {
+			return nil, fmt.Errorf("error allocating console: %w", consoleErr)
+		} else {
+			log.Warnf("Already possessing a console")
+		}
```
michel-laterman:
```suggestion
		}
		log.Warnf("Already possessing a console")
```

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -298,37 +300,31 @@ func TestUpgraderAckAction(t *testing.T) {
...
 
 	t.Run("AckAction with acker - failed ack", func(t *testing.T) {
-		acker := &fakeAcker{}
+		mockAcker := ackermocks.NewAcker(t)
 
 		errAck := errors.New("ack error")
-		acker.On("Ack", mock.Anything, action).Return(errAck)
-		acker.On("Commit", mock.Anything).Return(nil)
+		mockAcker.EXPECT().Ack(mock.Anything, action).Return(errAck)
+		// no expectation on Commit() since it shouldn't be called after an error during Ack()
```
michel-laterman, line 326: Do you want to keep the `AssertNotCalled` check for this as part of the test?

internal/pkg/agent/application/upgrade/rollback_windows.go
pchila: That comment is outdated, coming from a previous iteration where I was experimenting with CREATE_NEW_CONSOLE but there are some restrictions with other flags, especially CREATE_NEW_PROCESS_GROUP (see https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags).
The current solution is to use CREATE_NEW_PROCESS_GROUP while allocating a new console in `StartWatcherCmd`. Will fix the comment :+1:

internal/pkg/agent/application/upgrade/upgrade_test.go
pchila: When using an [expecter struct](https://vektra.github.io/mockery/v2.53/features/#expecter-structs), any call that is not expected will error out and the expected calls are asserted using [AssertExpectations](https://pkg.go.dev/github.com/stretchr/testify/mock#Mock.AssertExpectations), so not having an expectation on `Commit()` will make the test fail if such a call is made.

The previous code was instrumenting the mock to return `nil` for calls on `Commit` just to assert at the end that this should not have been called. I think it's cleaner to setup only the expected calls as above.

internal/pkg/agent/application/upgrade/rollback.go
pchila: Fixed in 67615510687b543b2925e4e03cfe9b61d12735b8

internal/pkg/agent/application/upgrade/rollback.go
pchila: Fixed in 67615510687b543b2925e4e03cfe9b61d12735b8

internal/pkg/agent/application/upgrade/rollback_windows.go
pchila: Fixed in 67615510687b543b2925e4e03cfe9b61d12735b8

internal/pkg/agent/application/upgrade/rollback_windows.go
pchila: Fixed in 67615510687b543b2925e4e03cfe9b61d12735b8

internal/pkg/agent/cmd/watch_notwindows.go
pchila: Addressed in 9b63febdfe73111bd4d36cf2aabddd53aa7962ee

juliaElastic:
> The reason string is there also to allow Fleet to distinguish an automatic rollback (which may be retried) from a manual one. It's not a big change to add the version in the string but I don't know if that would make Fleet UI detection more difficult. @juliaElastic any preference/thoughts ?

I think it's okay to add a version, we can do a partial text matching on Fleet side, something like `manual rollback requested to version X`

ycombinator:
Hey @pchila, this PR mostly LGTM.  Could you just address this one bit:

> I think it's okay to add a version, we can do a partial text matching on Fleet side, something like `manual rollback requested to version X`

Because I'm still seeing:

```
$ sudo elastic-agent status
┌─ fleet
├─ elastic-agent
└─ upgrade_details
   ├─ target_version: 9.2.0+20250715000000
   ├─ state: UPG_ROLLBACK
   └─ metadata
      └─ reason: manual rollback requested
```

Thanks!

pkoutsovasilis:
thanks for addressing my initial concerns about the new field in the upgrade details @pchila (after the sync we had offline), [your PR here](https://github.com/elastic/elasticsearch/pull/134711) did uncover another bigger issue. That said, I am starting to have a look here at the code in the meantime

PS: every-time I looked at this PR there are conflicts, what other features/bug-fixes we have that overlap and most importantly can we give some time here for this big PR to go in?

internal/pkg/agent/application/upgrade/rollback_test.go
```diff
@@ -504,10 +716,6 @@ func createUpdateMarker(t *testing.T, log *logger.Logger, topDir, newAgentVersio
 		hash:          oldAgentHash,
 		versionedHome: oldAgentVersionedHome,
 	}
-	err := markUpgrade(log,
-		paths.DataFrom(topDir),
-		newAgentInstall,
-		oldAgentInstall,
-		nil, nil, OUTCOME_UPGRADE)
+	err := markUpgrade(log, paths.DataFrom(topDir), time.Now(), newAgentInstall, oldAgentInstall, nil, nil, 0)
```
pkoutsovasilis: nit; add a comment here to capture what's the expectation when `rollbackWindow` is 0.

internal/pkg/agent/application/upgrade/rollback_windows.go
```diff
@@ -19,11 +26,89 @@ const (
...
+		log.Warnf("Already possessing a console")
+	}
+	cmd := createCmd()
+	log.Infow("Starting upgrade watcher", "path", cmd.Path, "args", cmd.Args, "env", cmd.Env, "dir", cmd.Dir)
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start Upgrade Watcher: %w", err)
+	}
+	list, consoleErr := getConsoleProcessList()
+	if consoleErr != nil {
+		log.Errorf("failed to get console process list: %v", consoleErr)
```
pkoutsovasilis: `getConsoleProcessList()` is good for debugging purposes so failing at it maybe should be a warning?
```suggestion
		log.Warnf("failed to get console process list: %v", consoleErr)
```

internal/pkg/agent/application/upgrade/rollback_windows.go
```diff
@@ -19,11 +26,89 @@ const (
...
+	} else {
+		log.Infof("Found console processes %v", list)
```
pkoutsovasilis: `getConsoleProcessList()` is good for debugging purposes so let's make that a `Debugf`
```suggestion
		log.Debugf("Found console processes %v", list)
```

internal/pkg/agent/application/upgrade/watcher.go
```diff
@@ -257,3 +263,131 @@ func (ch *AgentWatcher) checkFailures() bool {
...
+func (a AgentWatcherHelper) SelectWatcherExecutable(topDir string, previous agentInstall, current agentInstall) string {
+	return selectWatcherExecutable(topDir, previous, current)
+}
+
+func (a AgentWatcherHelper) WaitForWatcher(ctx context.Context, log *logger.Logger, markerFilePath string, waitTime time.Duration) error {
+	return waitForWatcher(ctx, log, markerFilePath, waitTime)
+}
+
+func (a AgentWatcherHelper) TakeOverWatcher(ctx context.Context, log *logger.Logger, topDir string) (*filelock.AppLocker, error) {
+	return takeOverWatcher(ctx, log, new(commandWatcherGrappler), topDir, 30*time.Second, 500*time.Millisecond, 100*time.Millisecond)
```
pkoutsovasilis: do we want this 30*time.Second be configurable? if not let's make it a constant on top of the file so we can easily track it and change it in the future if we have to

internal/pkg/agent/cmd/upgrade.go
```diff
@@ -162,6 +167,15 @@ func upgradeCmdWithClient(input *upgradeInput) error {
 		return fmt.Errorf("failed to retrieve command flag information while trying to upgrade the agent: %w", err)
 	}
 
+	rollback, err := cmd.Flags().GetBool(flagRollback)
```
pkoutsovasilis: `rollback` here shadows the `func rollback` maybe change to rollbackFlag?!

pkoutsovasilis (commented):
from a quick look, I am continuing

internal/pkg/agent/application/upgrade/rollback_windows.go
pchila: Fixed in b8de122b4b9f8edea9291f8380fe39718ef7c76f

internal/pkg/agent/application/upgrade/rollback_windows.go
pchila: Fixed in b8de122b4b9f8edea9291f8380fe39718ef7c76f

internal/pkg/agent/application/upgrade/watcher.go
pchila: Used more descriptive constants in 80f65af24c2bc930fe32af523eeda7ac2fbef4d9
Making it configurable could be useful but it would probably come in a follow-up PR

internal/pkg/agent/cmd/upgrade.go
pchila: addressed in 1a376f15f7e145c2c6f73f68580a7de113e9ad99

pkoutsovasilis (commented):
@pchila I did some manual testing of this, and I noticed that if I hit the rollback cmd during either `UPG_DOWNLOADING` or `UPG_EXTRACTING ` I get an error that it cannot stat the upgrade marker. However, this recovers after a while but I feel that  maybe the rollback should be allowed only in `UPG_WATCHING`?

PS: If it makes any difference I did find another bug for my merged PR https://github.com/elastic/elastic-agent/pull/9634 which I am gonna open a fix about 😢

pchila:
> @pchila I did some manual testing of this, and I noticed that if I hit the rollback cmd during either `UPG_DOWNLOADING` or `UPG_EXTRACTING ` I get an error that it cannot stat the upgrade marker. However, this recovers after a while but I feel that maybe the rollback should be allowed only in `UPG_WATCHING`?

@pkoutsovasilis 
Good catch! That is an actual race condition we should protect ourselves from.
Added a state check in f05a783e904b72bd6e281a068e65c2268a4eac93

pkoutsovasilis (commented):
ok after the last commit I see that `sudo elastic-agent upgrade --rollback 9.2.0-SNAPSHOT` doesn't execute on states UPG_EXTRACTING and UPG_DOWNLOADING. However, if I snipe the command when the state is at `UPG_RESTARTING` this executes and prints this message `Rollback triggered to version , Elastic Agent is currently restarting`. Then when agent actually restarts and I re-issue the rollback cmd I get `Error: Failed trigger upgrade of daemon: no rollbacks available`

internal/pkg/agent/cmd/watch.go
```diff
@@ -53,24 +70,97 @@ func newWatchCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Command
...
+				fmt.Fprintf(streams.Err, "Error setting parent process signals: %v\n", err)
+				os.Exit(errorSettingParentSignalsExitCode)
+			}
+
+			takedown, _ := c.Flags().GetBool(takedownFlagName)
+			if takedown {
+				err = takedownWatcher(context.Background(), log, utils.GetWatcherPIDs)
+				if err != nil {
+					log.Errorf("error taking down watcher: %v", err)
+					os.Exit(5)
```
michel-laterman, line 84: What does 5 mean? Why isn't it a const defined with the other exit codes?

internal/pkg/agent/cmd/watch.go
```diff
@@ -35,14 +34,32 @@ import (
...
+	takedownFlagName      = "takedown"
+	takedownFlagShorthand = "t"
+
+	rollbackFlagName      = "rollback"
+	rollbackFlagShorthand = "r"
+
+	// error exit codes
+	errorSettingParentSignalsExitCode = 6
+	errorRollbackToValue              = 7
+	errorRollbackFailed               = 8
```
michel-laterman, line 48: Do we want to define exit codes 1-5 here as well?

michel-laterman (approved):

internal/pkg/agent/cmd/watch.go
pchila: No exit codes were defined as constants before this change and this is the first new exit code added which followed the style of the existing code (when I had to add more exit codes after this one I started to use constant to make the meaning explicit). We could define a constant for it if it's a major issue.

pchila:
> ok after the last commit I see that `sudo elastic-agent upgrade --rollback 9.2.0-SNAPSHOT` doesn't execute on states UPG_EXTRACTING and UPG_DOWNLOADING. However, if I snipe the command when the state is at `UPG_RESTARTING` this executes and prints this message `Rollback triggered to version , Elastic Agent is currently restarting`. Then when agent actually restarts and I re-issue the rollback cmd I get `Error: Failed trigger upgrade of daemon: no rollbacks available`

@pkoutsovasilis 
The update/rollbacks triggered from command line and executed by the GRPC server do not have any syncronization/queuing but execute concurrently.
What is happening is that as soon as a watcher starts watching because of an upgrade being executed, it writes the state in the update marker which in turn causes an update in the upgrade details to `details.StateWatching` and that allows a rollback request to execute even if the agent is about to restart.
The rollback code is not able to distinguish if a restart has already happened, so it tries to execute a rollback when things are still in flux, ending up with weird behaviors like the one you described.

Will have a look and check what options are there to close this loophole without rewriting the whole thing :wink:

pchila:
@pkoutsovasilis 
I tried to reproduce the issue and I noticed that the error message is coming from https://github.com/elastic/elastic-agent/blob/a77d4cdf7a4104eccf23d59743e51adbb4891803/internal/pkg/agent/cmd/upgrade.go#L229 and the command interprets the agent's GRPC server shutting down as operation completed correctly (this happens already in main for upgrade commands due to https://github.com/elastic/elastic-agent/pull/4519).

In my tests I can reproduce the error message but I checked and the rollback is not actually performed: instead the agent restart after the upgrade and then I can still rollback correctly to the previous version by reissuing the command (the install is not corrupted or broken from what I can see).

```shell
root@2225b2f69a3c:/# elastic-agent upgrade --rollback 9.2.0-SNAPSHOT
Rollback triggered to version , Elastic Agent is currently restarting
root@2225b2f69a3c:/# elastic-agent version
Binary: 9.2.0+20250917-SNAPSHOT (build: f05a783e904b72bd6e281a068e65c2268a4eac93 at 2025-09-17 09:39:37 +0000 UTC)
Daemon: 9.2.0+20250917-SNAPSHOT (build: f05a783e904b72bd6e281a068e65c2268a4eac93 at 2025-09-17 09:39:37 +0000 UTC)
root@2225b2f69a3c:/# elastic-agent upgrade --rollback 9.2.0-SNAPSHOT
Rollback triggered to version 9.2.0-SNAPSHOT, Elastic Agent is currently restarting
root@2225b2f69a3c:/#
```

internal/pkg/agent/cmd/watch.go
pchila: We only react to these specific errors in this code path, so defining exit codes 1–5 here isn't necessary. Predefining exit codes that we don't handle would add noise without providing value.
If it's needed we can always fix this in a follow-up PR.

internal/pkg/agent/cmd/watch_notwindows.go
pchila: @ycombinator 
On the basis of a later [approval](https://github.com/elastic/elastic-agent/pull/8767#pullrequestreview-3226429368) I am assuming that this conversation can be resolved. If there's still something missing here that should be addressed in a follow-up PR please open an issue or unresolve the conversation and ping me.

cmacknz:
Thanks all for the extra attention to manual testing this one got before merge :)

ebeahan:
🚀

Issue #8137, opened 2025-05-09
Some config values stale after updating standalone config map


henrikno:
For confirmed bugs, please report:
- Version: 8.17.2
- Operating System: AKS, k8s, docker image
- Steps to Reproduce:

I'm using the standalone helm chart with the state stored in a host directory.
https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-standalone/elastic-agent-standalone-daemonset.yaml

We have a copy if the chart to be able to change things, so there might be discrepancies, but the setup is very similar.
I updated the log level in the config map via a helm value:
```
    agent:
      logging:
        level: {{ .Values.global.log_level | default "info" }}
```
Then I killed the pod, it started again, but still with the old log level.

Then I execed into the pod, `kubectl exec --stdin --tty -n elastic-agent pod/elastic-agent-xyx -- /bin/bash`
elastic-agent inspect returns the old value of log level.

I cleaned up the state directory `/var/lib/elastic-agent-standalone/kube-system/state` and restarted and it got the new value. But that also leads to duplicate logs so I want to avoid that.

I think this happens with some other values than log level but haven't reproduced it.

henrikno:
Sometimes this also happens with config values.
I updated the configmap, I can see the new value in the configmap in k8s. I killed the old pod, and it started a new one, but it logged that it used the old value, and elastic-agent inspect shows the old value. The config in this case was a change in the inputs, changing the bucket name of an S3 input.
elastic-agent is running this command: `elastic-agent container -c /etc/elastic-agent/agent.yml -e`

And if I check /etc/elastic-agent/agent.yml it contains the _new_ value. So somewhere it has "cached" the old value.

cmacknz:
@pkoutsovasilis any ideas what could be going on here?

swiatekm:
Killing the Pod is essentially starting agent in a new environment, so in principle the old value could've only been saved in the agent's persistent state somehow. I'm not aware of any mechanism that could've made it happen though.

pkoutsovasilis:
@cmacknz I believe that there is something happening with the state (which is persisted at the state directory /var/lib/elastic-agent-standalone/kube-system/state) and the following code parts seem to cause some overrides to always apply but and a new config not being saved 

https://github.com/elastic/elastic-agent/blob/d00a1b8c1249e296efba11d678f9a24d54b8b1e3/internal/pkg/agent/cmd/run.go#L445-L456

[...]

PR #8784, opened 2025-07-02
fix: don't override log level from state for standalone agents

pkoutsovasilis:
## What does this PR do?

This PR fixes a bug where the log level for standalone Elastic Agent installations was incorrectly restored from the persisted state directory (`/var/lib/elastic-agent-standalone/...`) instead of being sourced from the current configuration file. This resulted in stale configuration values after updates to the ConfigMap, particularly when the pod was restarted without clearing the state directory.

The change ensures that for standalone mode, the `log.level` value is always taken from the configuration, not restored from previously persisted agent state. For managed (Fleet) mode, the previous behavior is preserved.

## Why is it important?

This fix addresses an issue where changes to logging configuration were not taking effect unless the agent's state directory was manually deleted. This led to operational confusion and could cause logs to be emitted at incorrect verbosity levels.

Ensuring that the standalone agent reliably reflects the current configuration helps users maintain observability and avoid side effects like duplicated logs when restarting pods.

## Checklist

## Disruptive User Impact

This change should not cause disruption. It fixes the logic so that standalone agents correctly reflect updated configuration values after pod restarts, without requiring manual state cleanup. Users relying on previously persisted log levels (instead of ConfigMap changes) may see a difference in behavior, but this aligns with the intended functionality.

```
helm install elastic-agent ./deploy/helm/elastic-agent -n kube-system  --set outputs.default.type=ESPlainAuthBasic --set outputs.default.url='...' --set outputs.default.username='...' --set outputs.default.password='...' --set agent.presets.perNode.agent.logging.level=info

helm upgrade elastic-agent ./deploy/helm/elastic-agent -n kube-system  --set outputs.default.type=ESPlainAuthBasic --set outputs.default.url='...' --set outputs.default.username='...' --set outputs.default.password='...' --set agent.presets.perNode.agent.logging.level=debug
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/8137

### DIFF
--- a/internal/pkg/agent/application/info/agent_info.go
+++ b/internal/pkg/agent/application/info/agent_info.go
@@ -60,6 +60,9 @@ type AgentInfo struct {
 	esHeaders map[string]string
 }
 
+// for unit testing
+var doLoadAgentInfoWithBackoff = loadAgentInfoWithBackoff
+
 // NewAgentInfoWithLog creates a new agent information.
 // In case when agent ID was already created it returns,
 // this created ID otherwise it generates
@@ -67,7 +70,7 @@ type AgentInfo struct {
 // If agent config file does not exist it gets created.
 // Initiates log level to predefined value.
 func NewAgentInfoWithLog(ctx context.Context, level string, createAgentID bool) (*AgentInfo, error) {
-	agentInfo, isStandalone, err := loadAgentInfoWithBackoff(ctx, false, level, createAgentID)
+	agentInfo, isStandalone, err := doLoadAgentInfoWithBackoff(ctx, false, level, createAgentID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


swiatekm (approved):
LGTM, though I'd love to have a simple unit test for this function.

pkoutsovasilis:
> LGTM, though I'd love to have a simple unit test for this function.

more than a fair point, lemme check what I can cook for it

internal/pkg/agent/application/info/agent_info.go
```diff
@@ -76,9 +76,13 @@ func NewAgentInfoWithLog(ctx context.Context, level string, createAgentID bool)
 		return nil, fmt.Errorf("failed to determine root/Administrator: %w", err)
 	}
 
+	if !isStandalone {
```
pkoutsovasilis, line 82: making a comment to block merging, I will investigate if this can be unit-tested

pchila (approved):
LGTM

michel-laterman (approved):

PR #8786, opened 2025-07-02
Handle rollback in upgrade flow

michalpristas:
Added handling for rollback in upgrade flow

This path is not invokable so far, CLI and Action handling will be added later. 

Flow
- when flag is set to true we enter handleRollback
- handleRollback loads and store marker with updated desired outcome in case
  - marker file is present and there are available rollbacks in marker file
  - we are in a valid state (watching, rollback available)
- then it starts watcher if necessary that handles rollback itself 

Unit tests added to make sure rollback flow is triggered when flagged 
Fixes: #6889


pchila (changes requested):
Please hold off any PR that would impact manual rollback for the moment.
The solution is being revisited so no need to add more code that may need changes

michalpristas:
closing PR as per @pchila comment. can reopen if this turns up viable

PR #8837, opened 2025-07-03
[CODEOWNERS] Set `otel_samples/autoops_es.yml` ownership to OpEx team

pickypg:
The team will need to be given write access to the repo (like the `beats` repo) to stop `CODEOWNERS` from complaining about errors in the file.

It locally complained when I tried to wildcard the OS directory, so I just listed all three variants (which are identical) explicitly.

I'm unsure if this should be backported to 9.1 where that file also lives or not.


ycombinator (approved):

ycombinator:
> I'm unsure if this should be backported to 9.1 where that file also lives or not

I think it's safer to backport since the file exists on that branch.

cmacknz:
> The team will need to be given write access to the repo (like the beats repo) to stop CODEOWNERS from complaining about errors in the file.

✅

cmacknz (approved):

Issue #7944, opened 2025-04-23
Elastic Agent not serving policy to Endpoint


gabriellandau:
Elastic Agent 8.17.4
Windows 10 22H2 19045.5737

Elastic Agent 8.17.4 was installed over 8.18.0, potentially with `install -f`.  After the downgrade, the Endpoint process starts but remains blocked waiting for policy from Agent.  

Endpoint has three threads blocked on Agent:
 - CheckInLoop blocked in `grpc::ClientReaderWriter<proto::ActionResponse,proto::ActionRequest>::Read`
 - ActionLoop blocked in `grpc::ClientReaderWriter<proto::ActionResponse,proto::ActionRequest>::Read`
 - StateReportLoop blocked in `grpc::internal::WriterInterface<proto::CheckinObserved>::Write`

Status:
```
C:\>"C:\Program Files\Elastic\Agent\elastic-agent.exe" status
┌─ fleet
│  └─ status: (HEALTHY) Connected
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   └─ endpoint-default
      ├─ status: (STARTING) Starting: endpoint service runtime
      ├─ endpoint-default
      │  └─ status: (STARTING) Starting: endpoint service runtime
      └─ endpoint-default-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
         └─ status: (STARTING) Starting: endpoint service runtime

C:\>"C:\Program Files\Elastic\Endpoint\elastic-endpoint.exe" status
- elastic-agent
  - status: (HEALTHY) Connected
- elastic-endpoint
  - status: (HEALTHY) Running (no policy)
```

The process tree:
![Image](https://github.com/user-attachments/assets/9eff42c9-d63b-4365-ab24-1511b5e27d92)

This issue is intentionally sparse.  I shared additional information internally, including memory dumps, logs, and pprofs.

cc @blakerouse @cmacknz

cmacknz:
On the agent side of this, we observed the agent service runtime appear to be stuck at:

```
                                                 1   100% |   github.com/elastic/elastic-agent/pkg/component/runtime.(*serviceRuntime).processCheckin github.com/elastic/elastic-agent/pkg/component/runtime/service.go:490
         0     0%   100%          1  0.71%                | github.com/elastic/elastic-agent/pkg/component/runtime.(*runtimeComm).CheckinExpected github.com/elastic/elastic-agent/pkg/component/runtime/runtime_comm.go:185
[...]

PR #8881, opened 2025-07-08
fix: blocking issues of runtime communicator

pkoutsovasilis:
## What does this PR do?

This PR fixes multiple deadlock conditions in the runtime communicator logic and improves overall lifecycle handling. Specifically, it:
- Handles potential deadlock when sending the initial observed message to the runtime if the communicator is destroyed.
- Handles deadlock when the runtime calls `CheckinExpected` (init expected check-in) and the communicator is already destroyed.
- Handles deadlock when the runtime calls `CheckinExpected` (init expected check-in) and the server has been disconnected.
- Although highly unlikely due to existing synchronisation primitives, it doesn't block against multiple init checkin messages that arrive after the first that completed the init checkin process.
- Removes a redundant goroutine in the `checkin` method, which also allows returning accurate gRPC status codes back to the client. (PS: this optimisation can be also applied for the actions handling)
- Adds comprehensive unit tests covering all the above scenarios to prevent regressions.

## Why is it important?

These fixes ensure the runtime communicator behaves predictably and safely across lifecycle boundaries like shutdown, reconnection, and concurrent access. Without these changes, users could experience hangs or lost check-in signals in rare but critical failure modes.

You can see that the previous implementation fails to complete these scenarios under test in CI [here](https://buildkite.com/elastic/elastic-agent/builds/23728#0197e8c9-0c0d-4d32-b97e-cdb5db6e6fb9/135-642)

## Checklist

## Disruptive User Impact

This is a bug-fix for internal lifecycle logic and is not expected to introduce changes in behaviour or configuration requirements for end users.

```
mage unitTest
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/7944

### DIFF
--- a/pkg/component/runtime/runtime_comm.go
+++ b/pkg/component/runtime/runtime_comm.go
@@ -442,7 +442,7 @@ func genServerName() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.Replace(u.String(), "-", "", -1), nil
+	return strings.ReplaceAll(u.String(), "-", ""), nil
 }
 
 func sendExpectedChunked(server proto.ElasticAgent_CheckinV2Server, msg *proto.CheckinExpected, chunkingAllowed bool, maxSize int) error {
--- a/pkg/component/runtime/runtime_comm.go
+++ b/pkg/component/runtime/runtime_comm.go
@@ -27,7 +27,7 @@ import (
 
 // Communicator provides an interface for a runtime to communicate with its running component.
 type Communicator interface {
-	// WriteConnInfo writes the connection information to the writer, informing the component it has access
+	// WriteStartUpInfo writes the connection information to the writer, informing the component it has access
 	// to the provided services.
 	WriteStartUpInfo(w io.Writer, services ...client.Service) error
 	// CheckinExpected sends the expected state to the component.
@@ -63,6 +63,7 @@ type runtimeComm struct {
 	initCheckinObserved   *proto.CheckinObserved
 	initCheckinExpectedCh chan *proto.CheckinExpected
 	initCheckinObservedMx sync.Mutex
+	runtimeCheckinDone    chan struct{}
 
 	actionsConn     bool
 	actionsDone     chan bool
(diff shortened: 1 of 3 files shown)


pkg/component/runtime/runtime_comm.go
```diff
@@ -225,85 +240,93 @@ func (c *runtimeComm) checkin(server proto.ElasticAgent_CheckinV2Server, init *p
...
 	select {
 	case <-c.checkinExpected:
 	default:
 	}
+	c.initCheckinObserved = init
+	c.runtimeCheckinDone = make(chan struct{})
 	c.initCheckinObservedMx.Unlock()
+	defer func() {
+		close(c.runtimeCheckinDone)
+	}()
```
blakerouse: I think you should assign this to a local variable as well that way the `defer func()` is always that channel. Otherwise it could happen (maybe not in current state of code, but in a future change) that `close` is not called on that channel if `c.runtimeCheckinDone` is assigned to a different channel.

```suggestion
	c.initCheckinObserved = init
	checkinDone := make(chan struct{})
	c.runtimeCheckinDone = checkinDone
	c.initCheckinObservedMx.Unlock()
	defer func() {
		close(checkinDone)
	}()
```

blakerouse (commented):
Looks really good. I like all the additional testing. That will hopefully keep this from dead locks in the future as well.

I do have one comment that is inline. See if it make sense.

pkg/component/runtime/runtime_comm.go
pkoutsovasilis: Thanks, I see the point you’re making. I wasn’t initially sure what the best approach would be: leaving it as-is to let future test failures catch unintended changes, or guarding against them proactively. Since I couldn’t decisively lean one way, I went with your suggestion and assigned the channel to a local variable. I also passed it explicitly to the defer function to ensure it’s always the same channel being closed 🙂 respective commit b211ef16279f0835d21fbc655937f9451260a015

PR #8921, opened 2025-07-09
Add `backport-active-all` label to Golang bump PRs

ycombinator:
## What does this PR do?

We have [automation](https://github.com/elastic/elastic-agent/blob/main/.ci/updatecli/updatecli-bump-golang.yml) to bump Golang versions in this repository.  This PR updates that automation to add the `backport-active-all` label on generated PRs so that, when those PRs are merged into `main`, Mergify automatically creates their backports to all active branches.

## Why is it important?

To keep the Golang version in sync across all active branches.

(Also to get @pkoutsovasilis [off my back](https://github.com/elastic/elastic-agent/pull/8919#issuecomment-3054217371) 😉)

### DIFF
--- a/.ci/updatecli/updatecli-bump-golang.yml
+++ b/.ci/updatecli/updatecli-bump-golang.yml
@@ -24,7 +24,7 @@ actions:
       automerge: false
       labels:
         - dependencies
-        - backport-skip
+        - backport-active-all
       title: '[Automation] Bump Golang version to {{ source "latestGoVersion" }}'
       description: |
         It requires the version to be bumped first in golang-crossbuild project, then a new release will be added to:


pkoutsovasilis (approved):

v1v (approved):
LGTM. Just a tiny question, I'd expect no merge conflicts when doing backports, or will that be happening often, if so, then I'd say we can change the strategy and use the matrix automation, see https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-agent-versions.yml#L9-L24 for instance

ycombinator:
> LGTM. Just a tiny question, I'd expect no merge conflicts when doing backports, or will that be happening often, if so, then I'd say we can change the strategy and use the matrix automation, see https://github.com/elastic/elastic-agent/blob/main/.github/workflows/bump-agent-versions.yml#L9-L24 for instance

Yeah, generally for these Golang version bump PRs, I wouldn't expect merge conflicts either.  They will happen occasionally, but I think it should be rare.  

That being said, I think the matrix automation approach is even safer and I don't see a downside to it — in terms of outcome, it's no different than adding the `backport-active-all` label to such PRs as the source of active branches is the same in both cases.  I also recall a similar/related discussion around dependency bumps at EAH (IIRC @swiatekm brought it up) and it was decided that we preferred to have dependency version bump PRs go directly against target branches instead of going against `main` and then being backported.  So, all in all, I'd be in favor of the matrix automation approach instead of the label approach.

Issue #8846, opened 2025-06-02
[beat_receivers] Redact or drop logs that could contain event data in the monitoring filestream instance


leehinman:
- Relates https://github.com/elastic/elastic-agent/issues/8845

Modified based on https://github.com/elastic/elastic-agent/issues/8846#issuecomment-3036700167 and https://github.com/elastic/elastic-agent/issues/8846#issuecomment-3036700195.

The scope of this issue is now to ensure we drop or redact any logs that could contain event data in the monitoring filestream instance, as explained in https://github.com/elastic/elastic-agent/issues/8845 we can now longer rely on the event logs to help us do this moving forward.

Original Description:

> Beat receivers need to support the `logging.event_data` functionality.  Because beat receivers use an already configured zap.Core, we need to make sure that the beat receiver logger can be configured to add the same fields as [LoggingWithTypedOutputsLocal](https://github.com/elastic/elastic-agent-libs/blob/5e7f706d9239dfe530085a514107531b0136df3e/logp/configure/logging.go#L118) that is used in [stand alone beats](https://github.com/elastic/beats/blob/1d00d2040cf368e60293919f25436d4da8f471ef/libbeat/cmd/instance/beat.go#L791) and that `elastic-agent` can use these fields to separate out the event data the same way it does today.

cmacknz:
Elasticsearch added a query parameter we can use with _bulk to prevent it from giving us back field values on error: 

- https://github.com/elastic/elasticsearch/pull/120725

This would solve the problem in the ES exporter which is the most likely place to have this problem. It wouldn't do anything for cases where a component we do not control logs OTLP events on error or at the debug error, for example a processor logs field values on error. It's less clear how to solve that part of this.

The event logging strategy in Beats and Elastic Agent today worked because we had total control of all the log statements and could instrument the ones that wanted to log event content appropriately.

cmacknz:
There was a lot of discussion about this today in the beats receivers meeting. A big point was that our goal here is not actually to maintain the event log, it is to prevent shipping sensitive data or credentials to Elasticsearch clusters accidentally while maintaining our and our users ability to debug problems. We don't need the event log files for this, any solution that prevents our monitoring Filebeat from shipping event data to the monitoring output will accomplish this.

[...]

PR #8988, opened 2025-07-14
[beatreceiver] Add test to ensure sensitive logs from ES exporter are not indexed by filestream-monitoring

khushijain21:
## What does this PR do?
This PR adds a test to ensure sensitive logs containing original input, from ES exporter, are not indexed to fleet by `filestream-monitoring`.

If a document is failed to index for any reason, ES exporter logs the document at debug level.

The format looks something like this 

```json
 {"log.level":"debug",
 "@timestamp":"2025-07-14T06:35:24.235Z",
 "message":"failed to index document; input may contain sensitive data",
 "input":"{\"create\":{\"_index\":\"logs-test-default\"}}\n{\"host\":{\"hostname\":\"Khushis-MacBook-Pro.local\",\...}\n", 
 "index":"logs-test-default",
 "error.type":"strict_dynamic_mapping_exception",
 "error.reason":"",
 "ecs.version":"1.6.0",
}
```

Since the `input` field above contains a json document and `filestream-monitoring`'s [ndjson parser](https://github.com/khushijain21/elastic-agent/blob/eventLogTest/internal/pkg/agent/application/monitoring/v1_monitor.go#L579-L588) does not recurseively parse JSON - we can be sure this field in never indexed. 

This test explicity ensure this behavior is always true

This test is a follow up as part of the discussion here https://github.com/elastic/elastic-agent/pull/8738#discussion_r2190263752

## Checklist

```
source ./env.sh; AGENT_VERSION=9.2.0 go test -tags integration -run TestSensitiveLogsESExporter ./testing/integration/ess -count=1 -v
```

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/8846


internal/pkg/otel/translate/otelconfig.go
```diff
@@ -406,7 +406,10 @@ func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
 	// we also want to use dynamic log ids
 	esConfig["logs_dynamic_id"] = map[string]any{"enabled": true}
 
-	// for compatibility with beats, we want bodymap mapping
```
khushijain21, line 424: this is already set by `ToOTelConfig` method

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,223 @@ func genIgnoredFields(goos string) []string {
...
+	require.NoError(t,
+		template.Must(template.New("config").Parse(configTemplate)).Execute(&configBuffer,
+			otelConfigOptions{
+				InputPath:  inputFilePath,
+				ESEndpoint: esEndpoint,
+				ESApiKey:   decodedApiKey,
+				Namespace:  info.Namespace,
+			}))
+
+	ctx, cancel := testcontext.WithDeadline(t, context.Background(), time.Now().Add(5*time.Minute))
```
pchila: Please use t.Context() (when testcontext.WithDeadline()  was written this was not available)

```suggestion
	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Minute)
```

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,223 @@ func genIgnoredFields(goos string) []string {
...
+	defer cancel()
+	err = fixture.Prepare(ctx)
+	require.NoError(t, err)
+
+	err = fixture.Configure(ctx, configBuffer.Bytes())
```
pchila, line 976: No checking error here?
```suggestion
	err = fixture.Configure(ctx, configBuffer.Bytes())
	require.NoError(t, err)
```

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,223 @@ func genIgnoredFields(goos string) []string {
...
+			t.Log(output.String())
+		}
+	})
+
+	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+		var statusErr error
+		status, statusErr := fixture.ExecStatus(ctx)
+		assert.NoError(collect, statusErr)
+		assertBeatsHealthy(collect, &status, component.OtelRuntimeManager, 2)
+		return
```
pchila: I don't think this is needed here
```suggestion
```

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,223 @@ func genIgnoredFields(goos string) []string {
...
+			},
+		},
+	}
+
+	// Make sure find the logs
+	var monitoringDoc estools.Documents
+
+	assert.EventuallyWithT(t,
+		func(ct *assert.CollectT) {
+			findCtx, findCancel := context.WithTimeout(context.Background(), 10*time.Second)
```
pchila:
```suggestion
			findCtx, findCancel := context.WithTimeout(t.Context(), 10*time.Second)
```

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,223 @@ func genIgnoredFields(goos string) []string {
...
+			require.NoError(ct, err)
+
+			assert.GreaterOrEqual(ct, monitoringDoc.Hits.Total.Value, 1)
+		},
+		2*time.Minute, 5*time.Second,
+		"Expected atleast %d log, got %d", 1, monitoringDoc.Hits.Total.Value)
+
+	inputField := monitoringDoc.Hits.Hits[0].Source["input"]
+	_, ok := inputField.(string)
+	assert.False(t, ok, "ES logs contains sensitive data that includes the original input")
```
pchila: What type should the `input` field be? 
If this is testing that we don't ingest the original event shouldn't we check that it doesn't contain some "secret" value ?

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,223 @@ func genIgnoredFields(goos string) []string {
...
+		Stack: &define.Stack{},
+	})
+	tmpDir := t.TempDir()
+	numEvents := 50
+	// Create the data file to ingest
+	inputFile, err := os.CreateTemp(tmpDir, "input.txt")
+	require.NoError(t, err, "failed to create temp file to hold data to ingest")
+	inputFilePath := inputFile.Name()
+	for i := 0; i < numEvents; i++ {
+		_, err = inputFile.Write([]byte(fmt.Sprintf("Line %d\n", i)))
```
pchila, line 907: Could you add a comment explaining why these will fail to index?

testing/integration/ess/beat_receivers_test.go
khushijain21: yes, I added a more explicit statement now that tests for exact message field

cmacknz:
> Since the input field above contains a json document and filestream-monitoring's [ndjson parser](https://github.com/khushijain21/elastic-agent/blob/eventLogTest/internal/pkg/agent/application/monitoring/v1_monitor.go#L579-L588) does not recurseively parse JSON - we can be sure this field in never indexed.

What do the indexed documents look like in Elasticsearch?

khushijain21:
> What do the indexed documents look like in Elasticsearch?

It omits the "input" field and keeps everything else.

This is an example of it

<details>

<summary> sample-json </summary>

```json
{
  "_index": ".ds-logs-elastic_agent-default-2025.07.14-000001",
  "_id": "AZgIrQI6w-s6Q9UpdaX0",
  "_version": 1,
  "_source": {
    "container": {
      "id": "_38x0gd965v0156jwblb9x6h0000gn"
    },
    "otelcol.signal": "logs",
    "agent": {
      "name": "Khushis-MacBook-Pro.local",
      "id": "38359b6f-01b5-4e09-9f46-cf252eb4ad98",
      "type": "filebeat",
      "ephemeral_id": "eed651d5-dfac-48e7-ab10-5f7fd3209816",
      "version": "9.2.0"
    },
    "resource": {
      "service.version": "9.2.0",
      "service.instance.id": "21adeddd-8edb-448d-92f2-1abce8f4c300",
      "service.name": "/var/folders/cv/_38x0gd965v0156jwblb9x6h0000gn/T/TestSensitiveLogsESExporter604144974/elastic-agent-9.2.0-darwin-aarch64/elastic-agent"
    },
    "log": {
      "file": {
        "inode": "69166795",
        "path": "/private/var/folders/cv/_38x0gd965v0156jwblb9x6h0000gn/T/TestSensitiveLogsESExporter604144974/elastic-agent-9.2.0-darwin-aarch64/data/elastic-agent-7db5c8/logs/elastic-agent-20250714.ndjson",
        "device_id": "16777229",
        "fingerprint": "93156b077776fd0e326494f227871a836ef42717b65b11008377810b13330a6e"
      },
      "offset": 1254135
    },
    "elastic_agent": {
      "id": "38359b6f-01b5-4e09-9f46-cf252eb4ad98",
      "version": "9.2.0",
      "snapshot": false
    },
    "index": "logs-sensitive-wfp06i2spajjnq1dhf049pl3fuo2kccvpnsyhnvw",
    "message": "failed to index document; input may contain sensitive data",
    "log.origin": {
      "file.line": 384,
      "function": "github.com/open-telemetry/opentelemetry-collector-contrib/exporter/elasticsearchexporter.flushBulkIndexer",
      "file.name": "elasticsearchexporter@v0.129.0/bulkindexer.go"
    },
    "input": {
      "type": "filestream"
    },
    "error.reason": "",
    "error.type": "document_parsing_exception",
    "@timestamp": "2025-07-14T11:23:32.724Z",
    "ecs": {
      "version": "8.0.0"
    },
    "data_stream": {
      "namespace": "default",
      "type": "logs",
      "dataset": "elastic_agent"
    },
    "host": {
      "hostname": "Khushis-MacBook-Pro.local",
      "os": {
        "build": "24F74",
        "kernel": "24.5.0",
        "name": "macOS",
        "family": "darwin",
        "type": "macos",
        "version": "15.5",
        "platform": "darwin"
      },
      "ip": [
        "fe80::8:efd2:c3b2:2502",
        "2405:201:d00a:e8b1:4b:c2c3:30e0:8045",
        "2405:201:d00a:e8b1:3521:c813:b6c9:6a87",
        "192.168.29.114",
        "fe80::5418:f1ff:fe20:ce1c",
        "fe80::5418:f1ff:fe20:ce1c",
        "fe80::4977:48e6:7d97:df8f",
        "192.168.64.1",
        "fe80::7cf3:4dff:feae:ef64",
        "fd63:b2ca:f9c5:50e6:14:bb00:f72a:a174",
        "fe80::21d6:b5bb:76e9:b890",
        "fe80::8cd3:24f3:733e:e444",
        "fe80::ce81:b1c:bd2c:69e"
      ],
      "name": "khushis-macbook-pro.local",
      "id": "194CE554-9A32-579D-B00E-029E6412C7D4",
      "mac": [
        "1E-2B-07-D9-EC-0B",
        "1E-2B-07-D9-EC-0C",
        "1E-2B-07-D9-EC-0D",
        "1E-2B-07-D9-EC-EB",
        "1E-2B-07-D9-EC-EC",
        "1E-2B-07-D9-EC-ED",
        "36-B4-B1-E5-79-80",
        "36-B4-B1-E5-79-84",
        "36-B4-B1-E5-79-88",
        "56-18-F1-20-CE-1C",
        "7E-F3-4D-AE-EF-64",
        "96-3B-EB-76-C1-59",
        "B2-41-D2-A9-2F-4E",
        "DE-ED-2B-E5-DA-82"
      ],
      "architecture": "arm64"
    },
    "log.level": "debug",
    "otelcol.component.id": "elasticsearch/_agent-component/default",
    "otelcol.component.kind": "exporter",
    "event": {
      "dataset": "elastic_agent"
    }
  },
  }
```

</details>

env.sh-4927487305822283922
pchila, line 1: @khushijain21 I am pretty sure that this file does not belong in git, especially with the credentials in it

khushijain21: you are right, that was a mistake

khushijain21:
> What setting in the elastic-agent monitoring config will instruct elasticsearch-exporter to not log the entire document in case of a mapping error?

We do not make any changes on agent monitoring that will instruct es-exporter to *not* log failed documents.
Instead, ES exporter *will* log entire document on failure - we only ensure those logs are not picked by self-monitoring component.

> we only ensure those logs are not picked by `filestream-monitoring`

And this again is already supported because of how we configure parsing json messages on `filestream-monitoring. (More in the issue description itself)
This test only ensures that this behavior is always true because it is a security concern.

> I feel like this test should be a unit test verifying that monitoring configuration (or any beat receiver configuration for that matter) contains the correct configuration for the elasticsearch explorer.

I am not sure this is possible, because we require atleast one working input sending incorrect data while `filestream-monitoring` monitors the logs file, indexes all the events. And we retrieve these events to ensure they don't have sensitive data. But I'd appreciate any thoughts/possible ways to convert this to UT.

v1v:
I'm not sure why I've been included in the pool of reviewers. Please let me know if I can help and how. For now, I removed us from it. 🙏

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,233 @@ func genIgnoredFields(goos string) []string {
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut,
+		esEndpoint+"/_index_template/no-dynamic-template",
+		bytes.NewReader(jsonData))
+	if err != nil {
+		return fmt.Errorf("could not create http request to ES server: %v", err)
+	}
+
+	// Set content type header
+	req.Header.Set("Content-Type", "application/json")
+	req.SetBasicAuth(user, pass)
```
swiatekm: Doesn't the client already do this for us?

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,233 @@ func genIgnoredFields(goos string) []string {
...
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
+	decodedApiKey, err := getDecodedApiKey(esApiKey)
+	require.NoError(t, err)
+
+	configTemplate := `
+agent.grpc:
+  port: 6799
```
swiatekm: I don't think this is necessary.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,233 @@ func genIgnoredFields(goos string) []string {
...
+        paths:
+          - {{.InputPath}}
+        prospector.scanner.fingerprint.enabled: false
+        file_identity.native: ~
+outputs:
+  default:
+    type: elasticsearch
+    hosts: [{{.ESEndpoint}}]
+    api_key: "{{.ESApiKey}}"
+    preset: "balanced"
```
swiatekm: I think this is the default, so no need to set it.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,233 @@ func genIgnoredFields(goos string) []string {
...
+agent:
+  monitoring:
+    enabled: true
+    metrics: false
+    logs: true
+    _runtime_experimental: otel
+agent.logging.level: debug
+agent.logging.to_stderr: true
+`
```
swiatekm: Are these necessary?

swiatekm (commented):
The logic of the test looks correct to me, left some nitpicks about the specifics.

testing/integration/ess/beat_receivers_test.go
khushijain21: yes, because we see sensitive logs at debug level only. And having logs on stderr is required so that we can see agent logs on failure here https://github.com/khushijain21/elastic-agent/blob/eventLogTest/testing/integration/ess/beat_receivers_test.go#L1005

testing/integration/ess/beat_receivers_test.go
khushijain21: I rechecked - so it sets basic auth. Removed that part.

But we still need to set the headers

swiatekm (approved):

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,226 @@ func genIgnoredFields(goos string) []string {
...
+			findCtx, findCancel := context.WithTimeout(t.Context(), 10*time.Second)
+			defer findCancel()
+
+			monitoringDoc, err = estools.PerformQueryForRawQuery(findCtx, rawQuery, "logs-elastic_agent-default*", info.ESClient)
+			require.NoError(ct, err)
+
+			assert.GreaterOrEqual(ct, monitoringDoc.Hits.Total.Value, 1)
+		},
+		2*time.Minute, 5*time.Second,
+		"Expected atleast %d log, got %d", 1, monitoringDoc.Hits.Total.Value)
```
mauri870:
```suggestion
		"Expected at least %d log, got %d", 1, monitoringDoc.Hits.Total.Value)
```

mauri870:
I found a suspicious error in the output for the Windows test: `Could not init registrar: failed to open store 'filebeat'`. That sounds like the store is corrupted somehow. This is the same kind of error we got when running multiple receivers: https://github.com/elastic/beats/issues/43832#issuecomment-3029331439.

The beat receiver starts asynchronously, that means receiver.Start() returns success to the collector, but the internal Beat startup might still fail — and you’d only see that in the logs.

Looking at your test config, you have agent monitoring and a filebeat receiver, that's two beats receivers running on the same process. I'm pretty sure that is the first time that we are able to reproduce this problem with a real collector instead of unit tests.

cmacknz:
> I feel like this test should be a unit test verifying that monitoring configuration (or any beat receiver configuration for that matter) contains the correct configuration for the elasticsearch explorer.

> I am not sure this is possible, because we require atleast one working input sending incorrect data while filestream-monitoring monitors the logs file, indexes all the events. And we retrieve these events to ensure they don't have sensitive data. But I'd appreciate any thoughts/possible ways to convert this to UT.

We could look at testing this by relying on the individual pieces to hold up their part of the contract. 1. There is a test in Beats proving we don't recursively parse ndjson using exactly these error logs as a test case and 2. a test showing agent generates a Beat configuration with the ndjson parser.

If this was the only thing we needed to do to keep event data out of monitoring I could buy this, but it isn't. We also did https://github.com/elastic/elastic-agent/issues/9014 and I think this test should be updated to verify that drop processor as well.

I think it is valuable to have a a single end to end test proving that all the configuration we need to prevent sensitive data leaking into monitoring clusters actually works correctly together. Leaking sensitive data is very expensive from a process perspective and causes trust in our product to go down so I think it is worth testing.

If we don't have an end to end test of this here, it will be tested nowhere.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,226 @@ func genIgnoredFields(goos string) []string {
...
+		assert.NoError(collect, statusErr)
+		assertBeatsHealthy(collect, &status, component.OtelRuntimeManager, 2)
+	}, 1*time.Minute, 1*time.Second)
+
+	rawQuery := map[string]any{
+		"query": map[string]any{
+			"bool": map[string]any{
+				"must": map[string]any{
+					"match_phrase": map[string]any{
+						"message": "failed to index document; input may contain sensitive data",
```
cmacknz, line 1021: It would be a better test if we put some known sensitive patterns into the event like variations of api_key, password, an HTTP auth header, etc and then checked to make sure none of those fields are present.

I think this works for now, but this won't catch if the error message changes or the format of the log changes to have a sensitive value somewhere else.

khushijain21:
> Looking at your test config, you have agent monitoring and a filebeat receiver, that's two beats receivers running on the same process. I'm pretty sure that is the first time that we are able to reproduce this problem with a real collector instead of unit tests.

In that case, we might have to fix our global states before we can proceed with this. Thanks for the investigation @mauri870

khushijain21:
Blocked by https://github.com/elastic/beats/issues/44903

for reasons mentioned in this comment https://github.com/elastic/elastic-agent/pull/8988#issuecomment-3084369780

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,258 @@ func genIgnoredFields(goos string) []string {
...
+	require.NoError(t, err, "error getting elasticsearch endpoint")
+	esApiKey, err := createESApiKey(info.ESClient)
+	require.NoError(t, err, "error creating API key")
+	require.True(t, len(esApiKey.Encoded) > 1, "api key is invalid %q", esApiKey)
+	decodedApiKey, err := getDecodedApiKey(esApiKey)
+	require.NoError(t, err)
+
+	configTemplate := `
+agent.grpc:
+  port: 6790
```
swiatekm: This isn't necessary in integration tests.

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -421,8 +421,11 @@ func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
...
-	esConfig["mapping"] = map[string]any{"mode": "bodymap"}
+	// logs failed documents at debug level
+	esConfig["telemetry"] = map[string]any{
+		"log_failed_docs_input": true,
+	}
```
swiatekm, line 427: I don't mind this change, but it is a change. So you're not just adding a test. I'd edit the PR title and mention this in the description, for future generations.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,258 @@ func genIgnoredFields(goos string) []string {
 		}
 	}
 }
+
+// TestSensitiveLogsESExporter tests sensitive logs from ex-exporter are not sent to fleet
```
swiatekm, line 884:
```suggestion
// TestSensitiveLogsESExporter tests sensitive logs from es-exporter are not sent to fleet
```

swiatekm (commented):
A few minor comments, but looks good otherwise.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -877,3 +880,258 @@ func genIgnoredFields(goos string) []string {
...
+func TestSensitiveLogsESExporter(t *testing.T) {
+	// ES exporter logs the original document if they are failed to index only when
+	// "telemetry::log_failed_docs_input" setting is true and debug level is set
```
mauri870:
```suggestion
	// The ES exporter logs the original document on indexing failure only if
	// the "telemetry::log_failed_docs_input" setting is enabled and the log level is set to debug.
```

khushijain21:
This PR is going to be flaky because of global paths in filebeat. Keeping this in draft mode until https://github.com/elastic/beats/issues/44903 is resolved

PR #8991, opened 2025-07-14
Add component and unit diagnostics for beats receivers

swiatekm:
## What does this PR do?

Add the ability for the otel manager to output diagnostics for components and units it runs. This PR doesn't add any actual diagnostic content, just the necessary scaffolding and tests.

The filebeat registry and beat metrics will be added in follow-up PRs.

## Why is it important?

Beat receivers need to be covered by diagnostics to ensure issues with them can be effectively diagnosed.

## Checklist

Build the agent, run it with self-monitoring using beats receivers enabled, and collect diagnostics. You should see the expected directories without any files.

## Related issues


pierrehilbert:
@swiatekm should we also backport this to 9.1?

swiatekm:
> @swiatekm should we also backport this to 9.1?

We'd also need to backport https://github.com/elastic/elastic-agent/pull/8737. I think it might be easier to hold off for now and then make backport decisions once the full milestone 1 of beats receivers is done.

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -767,12 +775,28 @@ func (c *Coordinator) PerformAction(ctx context.Context, comp component.Componen
...
+	return diags
 }
 
 // PerformComponentDiagnostics executes the diagnostic action for the provided components.
 func (c *Coordinator) PerformComponentDiagnostics(ctx context.Context, additionalMetrics []cproto.AdditionalDiagnosticRequest, req ...component.Component) ([]runtime.ComponentDiagnostic, error) {
-	return c.runtimeMgr.PerformComponentDiagnostics(ctx, additionalMetrics, req...)
+	var diags []runtime.ComponentDiagnostic
+	runtimeDiags, err := c.runtimeMgr.PerformComponentDiagnostics(ctx, additionalMetrics, req...)
+	if err != nil {
+		return nil, err
```
pkoutsovasilis: let's wrap this error capturing that this was the traditional runtime that returned this error. And the one below that is from otel manager, wdyt? 🙂

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -0,0 +1,101 @@
...
+	// if no request is provided, then perform diagnostics for all units
+	if len(req) == 0 {
+		for _, comp := range currentComponents {
+			for _, unit := range comp.Units {
+				req = append(req, runtime.ComponentUnitDiagnosticRequest{
+					Component: comp,
+					Unit:      unit,
+				})
+			}
+		}
```
pkoutsovasilis: no need to populate req with all the components and then iterate through all of them just to return them all right?! return directly diagnostics?!

```suggestion
		for _, comp := range currentComponents {
			for _, unit := range comp.Units {
				diagnostics = append(diagnostics, runtime.ComponentUnitDiagnostic{
					Component: comp,
					Unit:      unit,
				})
			}
		}
		return diagnostics
```

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -0,0 +1,101 @@
...
+	ctx context.Context, additionalMetrics []cproto.AdditionalDiagnosticRequest, req ...component.Component,
+) ([]runtime.ComponentDiagnostic, error) {
+	var diagnostics []runtime.ComponentDiagnostic
+	m.mx.RLock()
+	currentComponents := m.components
+	m.mx.RUnlock()
+
+	// if no request is provided, then perform diagnostics for all components
+	if len(req) == 0 {
+		req = currentComponents
```
pkoutsovasilis, line 81: I would propose the same as above, no need to create a map and re-iterate here, you know it's gonna be all components, right?

internal/pkg/otel/manager/diagnostics_test.go
```diff
@@ -0,0 +1,79 @@
...
+	otherComp.InputSpec.Spec.Command.Args = []string{"metricbeat"}
+
+	m := &OTelManager{
+		logger:     logger,
+		components: []component.Component{filebeatComp, otherComp},
+	}
+
+	diags, err := m.PerformComponentDiagnostics(context.Background(), nil)
+	require.NoError(t, err)
+	require.Len(t, diags, 2)
```
pkoutsovasilis: it would be less error prone to future changes to define a map [string componentID] { value the args } and then in the require, you expect the `diags` Len to be equal of the map. At least make a const `expectedDiagsLen` declared with a comment on it why it is 2 🙂

internal/pkg/otel/manager/diagnostics_test.go
```diff
@@ -0,0 +1,79 @@
...
+	otherComp.InputSpec.Spec.Command.Args = []string{"metricbeat"}
+
+	m := &OTelManager{
+		logger:     logger,
+		components: []component.Component{filebeatComp, otherComp},
+	}
+
+	t.Run("diagnose all units when no request is provided", func(t *testing.T) {
+		diags := m.PerformDiagnostics(t.Context())
+		require.Len(t, diags, 4) // two components, two units per component
```
pkoutsovasilis: same as above

internal/pkg/otel/manager/manager.go
```diff
@@ -293,8 +293,10 @@ func (m *OTelManager) Errors() <-chan error {
 // This method updates the internal collector and component configurations and triggers a rebuild of the merged
 // configuration that combines them.
 func (m *OTelManager) handleConfigUpdate(cfg *confmap.Conf, components []component.Component) error {
+	m.mx.Lock()
```
pkoutsovasilis: hmmm tricky locking here, you lock here then you release. Then inside `handleConfigUpdate` you call `updateMergedConfig`  which calls `buildMergedConfig` that reads `m.collectorCfg` without any lock. have a look, ideally we should try to guard as high as possible in the frames stack

internal/pkg/otel/manager/diagnostics.go
swiatekm: That works if we're only returning empty diagnostics, as we are right now, but this will change in the future. Then I'd need to duplicate that logic here before the early return.

I could stick everything into a condition, and do the component filtering in an `else` clause as well, if you think that's better. Honestly, I don't think making these copies is expensive, and the logic is more compartmentalized the way it is right now, but I don't mind changing it.

internal/pkg/otel/manager/diagnostics.go
swiatekm: You end up with the following:

```go
	// if no request is provided, then perform diagnostics for all components
	if len(req) == 0 {
		for _, comp := range currentComponents {
			diagnostics = append(diagnostics, runtime.ComponentDiagnostic{
				Component: comp,
			})
		}
	} else {
		// create a map of component by id, this is used to filter out components that do not exist in the manager
		compByID := make(map[string]component.Component)
		for _, comp := range req {
			compByID[comp.ID] = comp
		}

		// create empty diagnostics for components that exist in the manager
		for _, existingComp := range currentComponents {
			if inputComp, ok := compByID[existingComp.ID]; ok {
				diagnostics = append(diagnostics, runtime.ComponentDiagnostic{
					Component: inputComp,
				})
			} else {
				m.logger.Warnf("requested diagnostics for component %s, but it does not exist in the manager", existingComp.ID)
			}
		}
	}

	// actually generate the diagnostics here
```

Do you think that's more readable?

internal/pkg/agent/application/coordinator/coordinator.go
swiatekm: Honestly, this is a bit of new logic. Maybe we should always do both and combine the errors instead?

swiatekm: How about this: https://github.com/elastic/elastic-agent/pull/8991/commits/9f805e764863b38d40730627bc3f754aa09d5bac?

internal/pkg/otel/manager/manager.go
swiatekm: Hm, good point. I was very careful to lock for as short a time as possible in the diagnostics function, as that can potentially take multiple seconds of real time. Here, we may also need to wait for the collector subprocess to stop. I think this being able to delay diagnostics is fine though.

internal/pkg/otel/manager/diagnostics.go
pkoutsovasilis: yes if you return the diagnostics directly from the first if

internal/pkg/otel/manager/diagnostics.go
pkoutsovasilis: @swiatekm I understand you have a chain of PRs upcoming, but it is kinda hard to not comment on the code I see now because something else is following. For the next PRs let's try to maintain the feature implementation as a whole, so I can properly review it 🙂

internal/pkg/otel/manager/manager.go
swiatekm: See https://github.com/elastic/elastic-agent/pull/8991/commits/798afa2909be54362e26cddedf4412f08cfe6bcc. I'm not very happy with this, and may consider using channels after all, but it should be correct.

internal/pkg/otel/manager/diagnostics_test.go
swiatekm: I've just made the expected value explicit: https://github.com/elastic/elastic-agent/pull/8991/commits/89079b57cb7fd871fc46b8eb7d116e8b3d5a4d3d. I don't see any value in converting it to a map.

internal/pkg/otel/manager/diagnostics_test.go
swiatekm: See https://github.com/elastic/elastic-agent/pull/8991/commits/89079b57cb7fd871fc46b8eb7d116e8b3d5a4d3d.

internal/pkg/otel/manager/diagnostics.go
swiatekm: Here's a draft of my immediate follow-up, if you'd like to see how this function evolves: https://github.com/elastic/elastic-agent/pull/9029.

internal/pkg/otel/manager/diagnostics.go
swiatekm: Like so: https://github.com/elastic/elastic-agent/pull/8991/commits/b73c34619b0f338a35e342659b28ae80761d8226?

internal/pkg/otel/manager/manager.go
pkoutsovasilis: ok bear with me here, but how about the following to be more targeted with our locking

you remove `handleConfigUpdate` and `buildMergedConfig` from manager.go and instead you introduce a new package private function with the following signature

```
buildMergedConfig(cfgUpdate configUpdate, agentInfo info.Agent, beatMonitoringConfigGetter translate.BeatMonitoringConfigGetter) (*confmap.Conf, error) {
// you have everything you need to merge the config
...
}
```

now you can do the following inside the `case cfgUpdate := <-m.updateCh:`

```
....
mergedCfg, err = buildMergedConfig(cfgUpdate, m.agentInfo, m.beatMonitoringConfigGetter)
if err != nil {
    ....
}

m.mx.Lock()
m.collectorCfg = cfgUpdate.collectorCfg
m.components = cfgUpdate.components
m.mergedCollectorCfg = mergedCfg
m.mx.Unlock()
```

wdyt? 🙂

swiatekm: Sure, that's definitely an improvement, will do. What I'm not especially happy about is that only the reads in methods which don't use the main loop are covered by the lock. This is *correct* because the main loop is completely serial, but it's unpleasantly subtle.

swiatekm: See https://github.com/elastic/elastic-agent/pull/8991/commits/0a8c7b14869ec9244a74a057fd7025cc8857c022.

Issue #8207, opened 2025-05-21
Ensure switching to the otel runtime does not result in input data duplication.


cmacknz:
When changing the state of the beat receiver feature flag for the first time, ensure there is no data duplication.

Most users will experience the switch to beat receivers after an agent upgrade, as we change the default value of the feature flags in consecutive agent versions. Ensure that in this case previously collected data is not re-ingested.

AlexanderWert:
remaining things are tracked in https://github.com/elastic/elastic-agent/issues/9058

PR #9029, opened 2025-07-16
Add filebeat registry to beat receiver diagnostics

swiatekm:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...


swiatekm:
@pkoutsovasilis I suspect some funny business with filebeat receiver not writing the registry when it's supposed to in some circumstances. @leehinman any clue what could be causing this? I'm fairly confident my PR is correct in that it looks for the registry in the right place and includes it in diagnostics if it exists.

leehinman:
> @pkoutsovasilis I suspect some funny business with filebeat receiver not writing the registry when it's supposed to in some circumstances. @leehinman any clue what could be causing this? I'm fairly confident my PR is correct in that it looks for the registry in the right place and includes it in diagnostics if it exists.

It is very likely that the filebeat receiver registry was written to one of the metricbeat receiver's "data" directory.  This is because of https://github.com/elastic/beats/issues/44903.   To verify just look in the "run" directory if you see something like:

```
# find .
./beat/metrics-monitoring/registry
./beat/metrics-monitoring/registry/filebeat
./beat/metrics-monitoring/registry/filebeat/meta.json
./beat/metrics-monitoring/registry/filebeat/active.dat
./beat/metrics-monitoring/registry/filebeat/2.json
./beat/metrics-monitoring/registry/filebeat/log.json
...
```

that is exactly what is going on.  only the `filestream-monitoring` directory should have a `registry/filebeat` directory.

If that is happening, this PR is blameless.

swiatekm:
I think the integration test is lying, and that I possibly tricked myself every time I tested this locally. The reason is that if you first run filebeat in process mode and don't clear the data afterwards, you'll be left with a state directory the diagnostics code will happily pick up, even though it wasn't actually written by the beats receiver. This does mean the PR is correct and would work if the path handling in libbeat were correct, but the feature as a whole doesn't work.

For now, let's hold off on merging it.

leehinman:
> For now, let's hold off on merging it.

just for tracking.  blocked by https://github.com/elastic/elastic-agent/issues/8207

swiatekm:
The libbeat bug is triggered by running more than one beat receiver. I've modified my integration test to do that and to ensure it doesn't reuse the data from the beat process. I've also filed an issue to track the libbeat problem as related to diagnostics: https://github.com/elastic/ingest-dev/issues/5892. Do you think we can merge this change under these circumstances @pkoutsovasilis ?

testing/integration/ess/diagnostics_test.go
```diff
@@ -324,29 +324,27 @@ agent.monitoring.enabled: false
...
 	t.Run("filebeat process", func(t *testing.T) {
+		f, err := define.NewFixtureFromLocalBuild(t, define.Version(), integrationtest.WithAllowErrors())
+		require.NoError(t, err)
+		err = f.Prepare(ctx)
+		require.NoError(t, err)
+		// Create the data file to ingest
+		inputFile, err := os.CreateTemp(t.TempDir(), "input.txt")
+		require.NoError(t, err, "failed to create temp file to hold data to ingest")
+		err = os.WriteFile(inputFile.Name(), []byte("hello world\n"), 0644)
+		require.NoError(t, err, "failed to write data to temp file")
```
pkoutsovasilis: this is failing for windows because the t.Cleanup inside the t.TempDir is getting called before the cleanup of the `define.NewFixtureFromLocalBuild`, so this should change. Also I would propose to make these two tests follow  the slice of test cases, as the only diff between them is the `expectedCompDiagnosticsFiles`, template `Runtime`, and template `InputFile`

pkoutsovasilis:
> The libbeat bug is triggered by running more than one beat receiver. I've modified my integration test to do that and to ensure it doesn't reuse the data from the beat process. I've also filed an issue to track the libbeat problem as related to diagnostics: [elastic/ingest-dev#5892](https://github.com/elastic/ingest-dev/issues/5892). Do you think we can merge this change under these circumstances @pkoutsovasilis ?

In general, I prefer addressing bugs that affect new code or features before merging. That said, I think we can make an exception in this case. The code changes LGTM, I tested them on both Linux and Mac OS; let's address the CI on Windows 🙂

testing/integration/ess/diagnostics_test.go
pkoutsovasilis: I still see the same error [here](https://buildkite.com/elastic/elastic-agent/builds/24376#01983bc7-dc9e-4215-a03b-06e46327ff25/765-1246). I can only think of something "interesting" happening during fixture cleanup that leads to having agent/filebeat still running 🤔

pkoutsovasilis (approved):

leehinman (approved):

testing/integration/ess/diagnostics_test.go
swiatekm: That was me just not closing the file, it turns out. :sweat:

PR #9037, opened 2025-07-16
Enable beats receivers for monitoring by default.

cmacknz:


cmacknz:
I don't think we can enable this automatically until we can conditionally do it based on the ES output -> exporter translations that are confirmed to work.

PR #9048, opened 2025-07-17
ci: build agent from snapshot DRA

pkoutsovasilis:
## What does this PR do?

WIP do not merge yet

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


pkoutsovasilis:
I am gonna add also some extra reviewers 🙂 
- @pchila ; I used parts of your code changes, from when you did the packaging refactor, so please do have a look here
- @blakerouse ; As we have already discussed this PR paves the way to build elastic-agent with a "fixed" version of beats which is required for agentless, thus please have a look here and check for any gaps
- @v1v this PR also resolves our discussion [here](https://github.com/elastic/elastic-agent/pull/8978#issuecomment-3088559810) so please do have a look

.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -37,14 +37,16 @@ echo "~~~ Running integration tests as $USER"
 
 make install-gotestsum
 
-if [[ -z "${AGENT_VERSION:-}" ]]; then
-  # Parsing version.go. Will be simplified here: https://github.com/elastic/ingest-dev/issues/4925
+if [[ -f "${WORKSPACE}/.package-version" ]]; then
```
pchila: I think we still need to check that `AGENT_VERSION` env var is empty before setting it from the `.package-version` file.
This can be used as an escape hatch if for some reason there is a need to override AGENT_VERSION from outside regardless of the contents of `.package-version` or `version.go` (It is going to be a much rarer case thanks to the improvements contained in this PR, though)

this applies to `.buildkite/scripts/buildkite-integration-tests.ps1` and `.buildkite/scripts/buildkite-k8s-integration-tests.sh` too

.buildkite/scripts/steps/integration_tests_tf.ps1
```diff
@@ -27,7 +32,7 @@ $TestsExitCode = 0
 try {
     Write-Output "~~~ Running integration tests"
     # Get-Ess-Stack will start the ESS stack if it is a BK retry, otherwise it will retrieve ESS stack metadata
-    Get-Ess-Stack -StackVersion $PACKAGE_VERSION
+    Get-Ess-Stack -StackVersion $STACK_VERSION $STACK_BUILD_ID
```
pchila: Let's try to be consistent using named parameters for the `Get-Ess-Stack` function call
```suggestion
    Get-Ess-Stack -StackVersion $STACK_VERSION -StackBuildId $STACK_BUILD_ID
```
or we could remove the named parameters altogether
```suggestion
    Get-Ess-Stack $STACK_VERSION $STACK_BUILD_ID
```

dev-tools/mage/packageversion.go
```diff
@@ -0,0 +1,103 @@
...
+func initPackageVersion() {
+	_, err := os.Stat(PackageVersionFilename)
+	if err != nil {
+		return
+	}
+
+	if os.Getenv("USE_PACKAGE_VERSION") != "true" {
+		return
+	}
+
```
pchila: I think these checks should be reversed and an error should be returned.
```suggestion
func initPackageVersion() error {
	if os.Getenv("USE_PACKAGE_VERSION") != "true" {
		return nil
	}
	_, err := os.Stat(PackageVersionFilename)
	if err != nil {
		return fmt.Errorf("package version file %s not found: %w", PackageVersionFilename, err)
	}
```

I see that this is used in the settings.go for global initialization which is probably the reason for it not to return anything but we can always panic if asked to use the package version and that resulted in an error.

dev-tools/mage/settings.go
```diff
@@ -166,6 +166,10 @@ func initGlobals() {
 
 	ManifestURL = EnvOr(ManifestUrlEnvVar, "")
 	PackagingFromManifest = ManifestURL != ""
+
+	// order matters this must be called last as it will override some of the
+	// values above
+	initPackageVersion()
```
pchila: We can panic here in case of problems to bail out immediately
```suggestion
	err = initPackageVersion()
	if err != nil {
	  panic(err)
	}
```

pkoutsovasilis:
[elastic-agent-package](https://buildkite.com/elastic/elastic-agent-package) test pipeline: https://buildkite.com/elastic/elastic-agent-package/builds/7181 ✅

testing/integration/ess/upgrade_broken_package_test.go
```diff
@@ -32,6 +32,7 @@ func TestUpgradeBrokenPackageVersion(t *testing.T) {
 		Local: false, // requires Agent installation
 		Sudo:  true,  // requires Agent installation
 	})
+	t.Skip("Skip this test because it is not compatible with .package-version pinning")
```
pchila, line 35: This test cannot work when package version and the string contained in `version.go` diverge because of this [`snippet`](https://github.com/elastic/elastic-agent/blob/6c005a48168b9c3c15cde01bb5dc98396eec0fe5/internal/pkg/agent/application/paths/common.go#L218-L227) which will stop returning the correct versioned home.

As already discussed, maybe we can find an automated way to disable the test when package version and the version string in code diverge (we can file a follow-up issue, we don't need to solve this in this PR)

pchila (commented):
Thank you for this PR!
I left a couple of comments, nothing major ;)

.buildkite/scripts/steps/integration_tests_tf.ps1
pkoutsovasilis: ohhh nice catch! I didn't realise that I passed one parameter by name and another by position, I need to remove Powershell from my CV 🥲

.buildkite/scripts/buildkite-integration-tests.sh
pkoutsovasilis: ok fair point, better safe than sorry is the approach. lemme see what I can do

dev-tools/mage/packageversion.go
pkoutsovasilis: in other words, let's panic! ok in this context it makes sense 🙂

.buildkite/scripts/buildkite-integration-tests.sh
pkoutsovasilis: take a look here c1f5ed68f6041666f10d23a8b5a75f2251051fc3

.buildkite/scripts/steps/integration_tests_tf.ps1
pkoutsovasilis: went with named parameters here 312485bd0111ba90cba5c163cc6f93878de4decd

dev-tools/mage/packageversion.go
pkoutsovasilis: here we run into a bit of quirk. So the devtools package is being imported in multiple places which essentially breaks the absolute path of `PackageVersionFilename`. Initially I tried to put all the utilised funcs in other places under a sub-package of devtools but the blast radius skyrocketed and tbh I wasn't comfortable which such radical changes. Thus I fallback to logging if os.stat returns a not exist err and not panicking which made the CI happy again. Take a look here 4c3ffa94cd44f57738e0767eb27e0cf143fa003f and here b097bd0dc29df91164eaa9fde9bf298b2fbe0754 and lemme know what you think

dev-tools/mage/settings.go
pkoutsovasilis: since this regards the comment above, I will resolve it just to have the discussion under one place

dev-tools/mage/packageversion.go
pchila: I am fine with the solution in https://github.com/elastic/elastic-agent/commit/b097bd0dc29df91164eaa9fde9bf298b2fbe0754 , having a proper log is enough to debug

.buildkite/scripts/buildkite-integration-tests.sh
pchila: Looks good :+1:

testing/integration/ess/upgrade_broken_package_test.go
pchila: Created https://github.com/elastic/elastic-agent/issues/9188 to follow-up

pchila (approved):
LGTM

blakerouse:
@pkoutsovasilis Why not make `USE_PACKAGE_VERION=true` the default? I would prefer that. I think getting a consistent build is much better than it depends on when you ran it.

pkoutsovasilis:
@blakerouse I went with the opt-in approach (USE_PACKAGE_VERSION=true) rather than making it the default (opt-out) because we currently have a lot of mage targets and scripts that rely on the agent version, and I can’t guarantee nothing will break if we switch everything over immediately (aka there are not part of our normal CI flow).

For example, `packageAgentCore` and `packageUsingDRA` are used outside our integration test CI flow and for these we should not apply USE_PACKAGE_VERSION, they should always go by the version in version.go and in the supplied manifest, respectively. This part of the code is highly critical — it’s how we release our software — so I’d rather start conservatively. Does that make sense to you?

.buildkite/scripts/steps/integration-cloud-image-push.sh
```diff
@@ -0,0 +1,20 @@
...
+
+source .buildkite/scripts/common.sh
+
+echo "~~~ Pushing cloud image"
+
+if [ "${FIPS:-false}" == "true" ]; then
+  CI_ELASTIC_AGENT_DOCKER_IMAGE="docker.elastic.co/beats-ci/elastic-agent-cloud-fips"
+else
+  CI_ELASTIC_AGENT_DOCKER_IMAGE="docker.elastic.co/beats-ci/elastic-agent-cloud"
+fi
```
v1v: very `nit`

```suggestion
suffix=""
if [ "${FIPS:-false}" == "true" ]; then
  suffix="-fips"
fi
CI_ELASTIC_AGENT_DOCKER_IMAGE="docker.elastic.co/beats-ci/elastic-agent-cloud${suffix}"
```

v1v (approved):
LGTM, thanks Panos!

dev-tools/mage/packageversion.go
```diff
@@ -0,0 +1,109 @@
...
+	pvBytes, err := json.MarshalIndent(pv, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal package version: %w", err)
+	}
+
+	_, err = f.Write(pvBytes)
+	if err != nil {
+		return fmt.Errorf("failed to write package version: %w", err)
+	}
```
swiatekm: We could reduce the number of lines here with `os.WriteFile`.

swiatekm (approved):
LGTM, had one nitpick you can address if you want to.

dev-tools/mage/packageversion.go
pkoutsovasilis: valid point, I don't know why I went the long path, but this proposal makes sense

.buildkite/scripts/steps/integration-cloud-image-push.sh
pkoutsovasilis: take a look here a1a493722319191daed231c2f27e64fd41a70469 🙂

dev-tools/mage/packageversion.go
pkoutsovasilis: take a look here d6a460c0ddd9a15a3fd2ed3ce543fa036ef322d3 🙂

blakerouse:
> @blakerouse I went with the opt-in approach (USE_PACKAGE_VERSION=true) rather than making it the default (opt-out) because we currently have a lot of mage targets and scripts that rely on the agent version, and I can’t guarantee nothing will break if we switch everything over immediately (aka there are not part of our normal CI flow).
> 
> For example, `packageAgentCore` and `packageUsingDRA` are used outside our integration test CI flow and for these we should not apply USE_PACKAGE_VERSION, they should always go by the version in version.go and in the supplied manifest, respectively. This part of the code is highly critical — it’s how we release our software — so I’d rather start conservatively. Does that make sense to you?

Those targets are critical I get that, but a simple `mage package` should produce the same result that the CI is going to produce for a PR locally. Otherwise testing as a developer could be confusing. What is ran in CI is a different build then what you ran locally with `mage package`. Unless you know to use `USE_PACKAGE_VERSION=true`. To me it just makes it much better to make `USE_PACKAGE_VERSION=true` for all build cases expect in the specific cases where we don't want that like in `packageAgentCore` and `packageUsingDRA`.

I have another question that I am struggling to understand the complete flow. The agentless publish job pulls the elastic-agent build from the snapshots DRA and mirrors it so it can be used in serverless. With this change it means that is published to the snapshot DRA is always what comes from `.package-version`, correct?

pkoutsovasilis:
@pchila @v1v @swiatekm I did find a bug where I don't actually build from manifest (fun times). Please hold off with your reviews, I need to investigate this

pkoutsovasilis:
@blakerouse you are thinking only about the package, but the .package-version has to affect also the testing part. With the current code structure this has to live and called by an `init` func. So how are you thinking of telling apart, with the previous in mind, about which mage target is this about and how we should apply the .package-version overrides?

blakerouse:
@pkoutsovasilis I am not thinking about implementation details. I am just thinking about the developer experience and the consistency we should provide when it comes to building a package locally and how it should match what the CI will be using. I think it would be best to make this behavior the default behavior, that way we can be sure that through out all of CI, local, agentless, containers, etc all come with the binaries that are committed into `.package-version`. There are only the two cases that you mentioned that need the other behavior. I am sure we can come up with a way to make that happen. If you need help brain storming or if you want me to take on that part I can.

pkoutsovasilis:
@blakerouse following your logic and caring about developer experience, `mage package` should enable by default EXTERNAL="true" and SNAPSHOT="true" (because this is how it is in our CI) which we don't, should we also change that?

blakerouse:
> @blakerouse following your logic and caring about developer experience, `mage package` should enable by default EXTERNAL="true" and SNAPSHOT="true" (because this is how it is in our CI) which we don't, should we also change that?

I would expect that if `.package-version` is always being used that `EXTERNAL=true` would have to be true. How could it work if `EXTERNAL=false`? 

Yes, I believe they should be default. They are almost always set, easier that way. If you need the other way then you can set them explicitly to false.

pkoutsovasilis:
@blakerouse I see the point you're making, and I’m not saying there isn’t merit there.  
However, for a change this significant in our packaging flow - especially with all the unknowns around how various mage targets are used - I feel much more comfortable sticking with the current approach, where we explicitly opt in via environment variables for specific scenarios.  

This doesn’t mean it has to stay this way forever. It just allows us to validate this approach with the smallest possible blast radius first. I’m happy to explore the defaults you suggest in a follow‑up PR, but for this one I’d prefer to keep it targeted and minimal - its goal is to ensure that the CI remains reliable when releasing a new agent version before a DRA has been produced.

blakerouse:
@pkoutsovasilis Taking this in steps works for me, but I do believe it should be the default almost always and in the near future. 

You did skip this one question of mine, this is just more for my knowledge:

> I have another question that I am struggling to understand the complete flow. The agentless publish job pulls the elastic-agent build from the snapshots DRA and mirrors it so it can be used in serverless. With this change it means that is published to the snapshot DRA is always what comes from .package-version, correct?

swiatekm (approved):
Did the same spot check as Paolo and everything looked correct.

.buildkite/scripts/buildkite-integration-tests.sh
```diff
@@ -38,14 +38,20 @@ echo "~~~ Running integration tests as $USER"
...
+    echo "~~~ Agent version: ${AGENT_VERSION} (from .package-version)"
+  else
+    AGENT_VERSION=$(grep "const defaultBeatVersion =" version/version.go | cut -d\" -f2)
+    AGENT_VERSION="${AGENT_VERSION}-SNAPSHOT"
+    echo "~~~ Agent version: ${AGENT_VERSION} (from version/version.go)"
+  fi
+
+  export AGENT_VERSION
+else
+  echo "~~~ Agent version: ${AGENT_VERSION} (specified by env var)"
```
fr4nc1sc0-r4m0n, line 52: Just a QQ: I see this code is repeated [here](https://github.com/elastic/elastic-agent/pull/9048/files#diff-59087217e78ee67e56ab89fae2f04a36447bcb4fa7753ad4491f8521901eaadbR12-R23). Do you think could be extracted in a separate script just to reuse code?

pkoutsovasilis: thanks for raising that point @fr4nc1sc0-r4m0n ; the truth is that we are already at a maximum "capacity" of scripts  in the CI, and some of them are not even used. Thus, the plan forward (@v1v is aware of that) is to start consolidating them into fewer scripts and ideally transitioning them to golang code. That said, introducing another script just to dedup some variable handling feels like against this cause, but if we decide that we should do that I will open a follow-up PR, wdyt?

fr4nc1sc0-r4m0n: Makes sense for me. Thank you @pkoutsovasilis

blakerouse (approved):
lgtm

PR #9094, opened 2025-07-21
Mount Go build cache into crossbuild container

swiatekm:
## What does this PR do?

Mounts the Go build cache into the golang-crossbuild container. To facilitate this and allow the container to generate the build cache, it also changes the crossbuild container to run as the host user.

This has been tested locally on both Mac and Linux, and the CI passes for this PR. It probably requires somewhat more careful testing, as permission changes in the host build cache could potentially result in breakage for unified releases.

## Why is it important?

This reduces the build time of agent during packaging by around 75%. On my machine, the same-architecture build goes from 2 minutes to 25 seconds.

## Checklist

Run `mage crossbuild` and time it.

## Related issues

- Relates https://github.com/elastic/observability-dev/issues/4702

### DIFF
--- a/dev-tools/mage/build.go
+++ b/dev-tools/mage/build.go
@@ -166,6 +166,17 @@ func GolangCrossBuild(params BuildArgs) error {
 			"only be executed within the golang-crossbuild docker environment")
 	}
 
+	defer DockerChown(filepath.Join(params.OutputDir, params.Name+binaryExtension(GOOS)))
+	defer DockerChown(filepath.Join(params.OutputDir))
+
+	mountPoint, err := ElasticBeatsDir()
+	if err != nil {
+		return err
+	}
+	if err := sh.Run("git", "config", "--global", "--add", "safe.directory", mountPoint); err != nil {
+		return err
+	}
+
 	return Build(params)
 }
 
--- a/dev-tools/mage/crossbuild.go
+++ b/dev-tools/mage/crossbuild.go
@@ -9,7 +9,6 @@ import (
 	"go/build"
 	"log"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -334,14 +333,9 @@ func (b GolangCrossBuilder) Build() error {
 
 	buildCacheLocation := "/tmp/.cache/go-build"
 	if CrossBuildMountBuildCache {
-		// Mount the go build cache into the container.
-		out, err := exec.Command("go", "env", "GOCACHE").Output()
-		if err != nil {
-			return fmt.Errorf("failed to get GOCACHE: %w", err)
-		}
-		cacheDir := strings.TrimSpace(string(out))
+		// Mount the go build cache volume into the container.
 		args = append(args,
-			"-v", fmt.Sprintf("%s:%s", cacheDir, buildCacheLocation),
+			"-v", fmt.Sprintf("%s:%s", CrossBuildBuildCacheVolumeName, buildCacheLocation),
 		)
 	}
 
@@ -368,7 +362,6 @@ func (b GolangCrossBuilder) Build() error {
 		"--env", fmt.Sprintf("DEV=%v", DevBuild),
 		"--env", fmt.Sprintf("EXTERNAL=%v", ExternalBuild),
 		"--env", fmt.Sprintf("FIPS=%v", FIPSBuild),
-		"--user", fmt.Sprintf("%d:%d", uid, gid),
 		"-v", repoInfo.RootDir+":"+mountPoint,
 		"-w", workDir,
 		image,
--- a/dev-tools/mage/settings.go
+++ b/dev-tools/mage/settings.go
@@ -75,7 +75,8 @@ var (
 	CrossBuildMountModcache = EnvOr("CROSSBUILD_MOUNT_MODCACHE", "true") == "true"
 
 	// CrossBuildMountBuildCache mounts the Go build cache into golang-crossbuild containers
-	CrossBuildMountBuildCache = EnvOr("CROSSBUILD_MOUNT_GOCACHE", "true") == "true"
+	CrossBuildMountBuildCache      = EnvOr("CROSSBUILD_MOUNT_GOCACHE", "true") == "true"
+	CrossBuildBuildCacheVolumeName = "elastic-agent-crossbuild-build-cache"
 
 	BeatName        = EnvOr("BEAT_NAME", defaultName)
 	BeatServiceName = EnvOr("BEAT_SERVICE_NAME", BeatName)
--- a/dev-tools/mage/clean.go
+++ b/dev-tools/mage/clean.go
@@ -25,7 +25,7 @@ var DefaultCleanPaths = []string{
 	"_meta/kibana/7/index-pattern/{{.BeatName}}.json",
 }
 
-// Clean clean generated build artifacts.
+// Clean clean generated build artifacts and caches.
 func Clean(pathLists ...[]string) error {
 	if len(pathLists) == 0 {
 		pathLists = [][]string{DefaultCleanPaths}
@@ -37,6 +37,9 @@ func Clean(pathLists ...[]string) error {
 				return err
 			}
 		}
+	}
+	if CrossBuildMountBuildCache {
+		return sh.Run("docker", "volume", "rm", "-f", CrossBuildBuildCacheVolumeName)
 	}
 	return nil
 }


dev-tools/mage/build.go
```diff
@@ -166,17 +166,6 @@ func GolangCrossBuild(params BuildArgs) error {
...
-	defer DockerChown(filepath.Join(params.OutputDir))
-
-	mountPoint, err := ElasticBeatsDir()
-	if err != nil {
-		return err
-	}
-	if err := sh.Run("git", "config", "--global", "--add", "safe.directory", mountPoint); err != nil {
-		return err
-	}
-
```
swiatekm: All of this was necessary because files written in the container were owned by root. They're now owned by the host user to begin with.

dev-tools/mage/crossbuild.go
```diff
@@ -169,11 +170,9 @@ func CrossBuild(options ...CrossBuildOption) error {
 		return err
 	}
 
-	if CrossBuildMountModcache {
-		// Make sure the module dependencies are downloaded on the host,
-		// as they will be mounted into the container read-only.
-		mg.Deps(func() error { return gotool.Mod.Download() })
-	}
```
swiatekm: The `CrossBuildMountModCache` variable is always true, so I simply removed it.

swiatekm:
@v1v both of them succeeded at the build steps, but failed at the publishing step, even though I set them to dry-run. Not sure if that's in some way an effect of this change or if I made a mistake.

See:
* https://buildkite.com/elastic/elastic-agent-package/builds/7168#01985144-a389-4509-af44-f1717edd59b2
* https://buildkite.com/elastic/elastic-agent-binary-dra/builds/4211#0198518a-dc7c-4bb0-ad62-f7d65c48a0b9

v1v:
That's an expected error, I think it does not support feature branches:

```
The specified project directory '/release/project-configs/chore/crossbuild-build-cache' does not exist.
```

https://buildkite.com/elastic/elastic-agent-package/builds/7168#01985144-a389-4509-af44-f1717edd59b2/282-310
https://buildkite.com/elastic/elastic-agent-binary-dra/builds/4211#0198518a-dc7c-4bb0-ad62-f7d65c48a0b9/145-173


As far as I see, you can override them with `DRA_BRANCH=main`:

- https://github.com/elastic/elastic-agent/blob/8fbdaaa979db0976e33612be647bb69ffaac70f8/.buildkite/scripts/steps/dra-publish.sh#L8

---

I'm gonna run them again:

- https://buildkite.com/elastic/elastic-agent-package/builds/7187
- https://buildkite.com/elastic/elastic-agent-binary-dra/builds/4232

using the env variables:

```
DRA_BRANCH="main"
DRA_DRY_RUN="--dry-run"
DRA_PROJECT_ARTIFACT_ID="agent-core"
DRA_PROJECT_ID="elastic-agent-core"
DRA_VERSION=9.2.0
DRA_WORKFLOW="snapshot"
```

pchila:
> both of them succeeded at the build steps, but failed at the publishing step, even though I set them to dry-run. Not sure if that's in some way an effect of this change or if I made a mistake.

You would need to define `DRA_BRANCH="main"` (or another release branch) and `DRA_DRY_RUN="--dry-run"` like in this build https://buildkite.com/elastic/elastic-agent-package/builds/7181

Just be careful because without the appropriate `DRA_DRY_RUN` value you will end up overwriting DRAs built from the release branches with one from your PR

dev-tools/mage/crossbuild.go
```diff
@@ -315,18 +317,26 @@ func (b GolangCrossBuilder) Build() error {
...
 		args = append(args, "--env", "VERSION_QUALIFIER="+versionQualifier)
 	}
-	if CrossBuildMountModcache {
-		// Mount $GOPATH/pkg/mod into the container, read-only.
-		hostDir := filepath.Join(build.Default.GOPATH, "pkg", "mod")
-		args = append(args, "-v", hostDir+":/go/pkg/mod:ro")
+
+	// Mount $GOPATH/pkg/mod into the container, read-only.
+	hostDir := filepath.Join(build.Default.GOPATH, "pkg", "mod")
+	args = append(args, "-v", hostDir+":/go/pkg/mod:ro")
```
v1v: `nit`: use fmt.Sprintf as used in `L320`?
```suggestion
	args = append(args, "-v",  fmt.Sprintf("%s:/go/pkg/mod:ro", hostDir))
```

v1v (approved):
LGTM. I've just left a comment; however, I'm not an expert in the build system. Please take my `+1` as only an acknowledgment.

dev-tools/mage/build.go
pkoutsovasilis: I think that this relates to https://github.com/elastic/elastic-agent/issues/9433 cc @dliappis @fr4nc1sc0-r4m0n

dev-tools/mage/crossbuild.go
pkoutsovasilis: or even better make that overridable by an env var

dev-tools/mage/crossbuild.go
```diff
@@ -315,18 +317,26 @@ func (b GolangCrossBuilder) Build() error {
...
+	args = append(args, "-v", fmt.Sprintf("%s:/go/pkg/mod:ro", hostDir))
+
+	// Mount the go build cache into the container.
+	out, err := exec.Command("go", "env", "GOCACHE").Output()
+	if err != nil {
+		return fmt.Errorf("failed to get GOCACHE: %w", err)
 	}
+	cacheDir := strings.TrimSpace(string(out))
+	buildCacheMountPoint := "/tmp/.cache/go-build"
+	args = append(args, "-v", fmt.Sprintf("%s:%s", cacheDir, buildCacheMountPoint))
```
pkoutsovasilis: I would strongly recommend an escape hatch here, so if this ever creates issues we can easily disable it. E.g. you could easily have that under the umbrella of `CrossBuildMountModcache` and allow the latter to be set to false by an env var

pchila: +1

dev-tools/mage/crossbuild.go
pchila: +1

dev-tools/mage/crossbuild.go
```diff
@@ -344,12 +361,14 @@ func (b GolangCrossBuilder) Build() error {
...
 		"--rm",
 		"--env", "GOFLAGS=-mod=readonly",
+		"--env", fmt.Sprintf("GOCACHE=%s", buildCacheLocation), // ensure this is writable by the user
 		"--env", "MAGEFILE_VERBOSE="+verbose,
 		"--env", "MAGEFILE_TIMEOUT="+EnvOr("MAGEFILE_TIMEOUT", ""),
 		"--env", fmt.Sprintf("SNAPSHOT=%v", Snapshot),
 		"--env", fmt.Sprintf("DEV=%v", DevBuild),
 		"--env", fmt.Sprintf("EXTERNAL=%v", ExternalBuild),
 		"--env", fmt.Sprintf("FIPS=%v", FIPSBuild),
+		"--user", fmt.Sprintf("%d:%d", uid, gid),
```
pkoutsovasilis: after working on https://github.com/elastic/elastic-agent/pull/9812, I realized I don’t really like the approach of running the container with whichever uid:gid invoked it.

Even though it’s not the case yet, golang-crossbuild is fundamentally a build container and it’s reasonable to assume a consistent $HOME so that all required tools are properly resolved. Changing the uid:gid makes this setup more fragile. For example, when running with a different uid, the container already prints the message "I have no name".

To avoid that, I’d suggest two possible alternatives:
	1.	Don’t change the uid:gid – instead, docker chown the mounted Go module build cache path to EXEC_UID and EXEC_GID (as is already being done [here](https://github.com/elastic/elastic-agent/blob/main/dev-tools/mage/build.go#L169)).
	2.	Don’t mount the host Go build cache – mount a dedicated folder used only by the golang-crossbuild invocation (maybe in build?!). The performance gains remain essentially the same, but without the fragility. (I would vote for this one)

dev-tools/mage/build.go
pkoutsovasilis: this has to stay, look at my other comment about uid:gid container invocation

dev-tools/mage/crossbuild.go
swiatekm: Fair. I don't like 1 because it feels similarly hacky, and the consequences of temporarily having files owned by root in the build cache are unknown. The main problem with 2 is that it would likely be quite large, and potentially not easy to clean up, though we could manually do so in `mage clean`. I'll see how big it gets in practice.

swiatekm:
Looks like the cache is around 5 GB per platform, which is a fair amount.

dev-tools/mage/settings.go
```diff
@@ -72,7 +72,11 @@ var (
 
 	// CrossBuildMountModcache mounts $GOPATH/pkg/mod into
 	// the crossbuild images at /go/pkg/mod, read-only,  when set to true.
-	CrossBuildMountModcache = true
+	CrossBuildMountModcache = EnvOr("CROSSBUILD_MOUNT_MODCACHE", "true") == "true"
+
+	// CrossBuildMountBuildCache mounts the Go build cache into golang-crossbuild containers
+	CrossBuildMountBuildCache      = EnvOr("CROSSBUILD_MOUNT_GOCACHE", "true") == "true"
+	CrossBuildBuildCacheVolumeName = "elastic-agent-crossbuild-build-cache"
```
pkoutsovasilis, line 79: ok docker named volume, I like it 👍

dev-tools/mage/crossbuild.go
swiatekm: Done, now controlled via `CROSSBUILD_MOUNT_MODCACHE`, true by default.

swiatekm:
I tested the dra publishing pipelines again and didn't see any issues, merging.

PR #9095, opened 2025-07-21
Always enable beat receiver http monitoring endpoints

swiatekm:
## What does this PR do?

When generating diagnostics for beat receivers, we can't get metrics over the control protocol like we do for beat processes. Instead, we need to use the http monitoring endpoint. However, these endpoints are disabled if agent self-monitoring is disabled.

Always enable them for beats receivers.

This PR also adds some missing unit tests for the otel config translation package.

## Why is it important?

We want to have beat metrics in diagnostics even if self-monitoring is disabled.

## Checklist

Run agent with an input in otel mode, collect diagnostics, and confirm that the final otel config contains the http monitoring section.

## Related issues


internal/pkg/agent/application/coordinator/coordinator_test.go
```diff
@@ -1078,8 +1076,7 @@ func createCoordinator(t testing.TB, ctx context.Context, opts ...CoordinatorOpt
 	cfg.Port = 0
 	rm, err := runtime.NewManager(l, l, ai, apmtest.DiscardTracer, monitoringMgr, cfg)
 	require.NoError(t, err)
-	otelMgr, err := otelmanager.NewOTelManager(l, logp.InfoLevel, l, otelmanager.EmbeddedExecutionMode, ai, monitoringMgr.ComponentMonitoringConfig)
-	require.NoError(t, err)
+	otelMgr := &fakeOTelManager{}
```
swiatekm, line 1079: This change is necessary because otherwise we get a circular dependency for test code. None of the coordinator tests in this file use the real otel manager for anything.

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -234,7 +236,18 @@ func getReceiversConfigForComponent(
 	}
 
 	// add monitoring config if necessary
```
khushijain21, line 238: I think we can remove this comment then

PR #9097, opened 2025-07-22
Add beat metrics to beat receiver diagnostics

swiatekm:
## What does this PR do?

It adds beat stats and input metrics to beat receiver diagnostics.

Note that this currently targets #9095, and will switch to main once that is merged. 

## Why is it important?

Diagnostics for beat receivers should contain the same information as they do for beat processes.

## Checklist

Build agent from this PR, then start it with self-monitoring in otel mode and generate diagnostics. You should see `beats_metrics.json` and `input_metrics.json` in the components' respective subdirectories in the diagnostics archive.

## Related issues

### DIFF
--- a/internal/pkg/otel/manager/diagnostics.go
+++ b/internal/pkg/otel/manager/diagnostics.go
@@ -135,7 +135,7 @@ func (m *OTelManager) PerformComponentDiagnostics(
 		if translate.GetBeatNameForComponent(&diag.Component) == "filebeat" {
 			// include filebeat registry, reimplementation of a filebeat diagnostic hook
 			registryTarGzBytes, err := FileBeatRegistryTarGz(m.logger, diag.Component.ID)
-			errs = append(errs, err)
+			errs = append(errs, fmt.Errorf("failed to create filebeat registry archive: %w", err))
 			if registryTarGzBytes != nil {
 				m.logger.Debugf("created registry tar.gz, size %d", len(registryTarGzBytes))
 				results = append(results, &proto.ActionDiagnosticUnitResult{
--- a/internal/pkg/otel/manager/diagnostics.go
+++ b/internal/pkg/otel/manager/diagnostics.go
@@ -9,6 +9,8 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
@@ -16,6 +18,8 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/monitoring"
 
 	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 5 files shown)


v1v:
I have removed it from the code review. The changes appear unfamiliar, but please let me know if you think we should assist. Thank you.

swiatekm:
> I have removed it from the code review. The changes appear unfamiliar, but please let me know if you think we should assist. Thank you.

Do you know why you're being added as reviewers here? It doesn't look like I'm modifying any files you own according to CODEOWNERS.

v1v:
> Do you know why you're being added as reviewers here? It doesn't look like I'm modifying any files you own according to CODEOWNERS.

No idea, it seems some force-pushed caused it:
- https://github.com/elastic/elastic-agent/pull/9097#event-18752616642

image

One last thing, to make the review process smoother, it would be great if we could avoid force-pushing when pull requests are ready. This way, we can keep the reviews more manageable and focused. 🙇

leehinman (dismissed):

leehinman (dismissed):

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -119,29 +123,114 @@ func (m *OTelManager) PerformComponentDiagnostics(
 
 	for idx, diag := range diagnostics {
 		var results []*proto.ActionDiagnosticUnitResult
+		var errs []error
+		jsonMetricDiagnostics, err := GetBeatJsonMetricsDiagnostics(ctx, diag.Component.ID)
+		errs = append(errs, err)
```
pkoutsovasilis, line 128: I find it weird that we always append an err (no nil checking) but more importantly even if err != nil we append the results. Looking at `GetBeatJsonMetricsDiagnostics` and `GetBeatInputMetricsDiagnostics` these do return a `nil []*proto.ActionDiagnosticUnitResult` when `err != nil`. Btw do these really need to return `[]*proto.ActionDiagnosticUnitResult`? I suspect that the answer is there will be another PR that utilises this func signature but I need to check with you 🙂

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -119,29 +123,114 @@ func (m *OTelManager) PerformComponentDiagnostics(
 
 	for idx, diag := range diagnostics {
 		var results []*proto.ActionDiagnosticUnitResult
+		var errs []error
```
pkoutsovasilis, line 126: since you do concatenate all of these errs with errors.Join, let's use it directly here and errors.Join to a single error that you assign to diagnostics[idx].Err

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -119,29 +123,114 @@ func (m *OTelManager) PerformComponentDiagnostics(
...
 	}
 
 	return diagnostics, nil
 }
 
+func GetBeatJsonMetricsDiagnostics(ctx context.Context, componentID string) ([]*proto.ActionDiagnosticUnitResult, error) {
+	var results []*proto.ActionDiagnosticUnitResult
+	beatMetrics, err := GetBeatMetricsPayload(ctx, componentID, "/stats")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get beat metrics: %w", err)
```
pkoutsovasilis:
```suggestion
		return nil, fmt.Errorf("failed to get stats beat metrics: %w", err)
```

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -119,29 +123,114 @@ func (m *OTelManager) PerformComponentDiagnostics(
...
+	}
+
+	beatMetrics, err = formatJSON(beatMetrics)
+	if err != nil {
+		return nil, fmt.Errorf("failed to format beat metrics: %w", err)
```
pkoutsovasilis:
```suggestion
		return nil, fmt.Errorf("failed to format stats beat metrics: %w", err)
```

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -119,29 +123,114 @@ func (m *OTelManager) PerformComponentDiagnostics(
...
+		Generated:   timestamppb.Now(),
+	})
+	return results, nil
+}
+
+func GetBeatInputMetricsDiagnostics(ctx context.Context, componentID string) ([]*proto.ActionDiagnosticUnitResult, error) {
+	var results []*proto.ActionDiagnosticUnitResult
+	inputMetrics, err := GetBeatMetricsPayload(ctx, componentID, "/inputs/")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get input metrics: %w", err)
```
pkoutsovasilis:
```suggestion
		return nil, fmt.Errorf("failed to get input beat metrics: %w", err)
```

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -119,29 +123,114 @@ func (m *OTelManager) PerformComponentDiagnostics(
...
+	}
+
+	inputMetrics, err = formatJSON(inputMetrics)
+	if err != nil {
+		return nil, fmt.Errorf("failed to format input metrics: %w", err)
```
pkoutsovasilis:
```suggestion
		return nil, fmt.Errorf("failed to format input beat metrics: %w", err)
```

internal/pkg/otel/manager/diagnostics_test.go
```diff
@@ -51,9 +61,18 @@ func TestPerformComponentDiagnostics(t *testing.T) {
...
-	diags, err := m.PerformComponentDiagnostics(context.Background(), nil)
+	diags, err := m.PerformComponentDiagnostics(t.Context(), nil)
 	require.NoError(t, err)
-	assert.Equal(t, expectedDiags, diags)
+	for i, d := range diags {
+		assert.Equal(t, expectedDiags[i].Component.ID, d.Component.ID)
+		// we should have errors set about not being able to connect to monitoring endpoints
+		require.NotNil(t, d.Err)
+		assert.ErrorContains(t, d.Err, "failed to get beat metrics")
+		assert.ErrorContains(t, d.Err, "failed to get input metrics")
```
pkoutsovasilis:
```suggestion
		assert.ErrorContains(t, d.Err, "failed to get stats beat metrics")
		assert.ErrorContains(t, d.Err, "failed to get input beat metrics")
```

internal/pkg/otel/manager/diagnostics_test.go
```diff
@@ -110,6 +129,74 @@ func TestPerformDiagnostics(t *testing.T) {
 	})
 }
 
+func TestBeatMetrics(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Skip test on Windows, it's annoying to set up an npipe http server.")
```
pkoutsovasilis: let's rephrase that to technically cumbersome than being annoying, I am not sure that latter counts as a valid reason 🙂

internal/pkg/otel/manager/diagnostics_test.go
```diff
@@ -110,6 +129,74 @@ func TestPerformDiagnostics(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.Len(t, obs.All(), 0)
+	require.Len(t, diags, 1)
+
+	diag := diags[0]
+	assert.Equal(t, filebeatComp, diag.Component)
+	// two metrics diagnostics and one filebeat registry
+	require.Len(t, diag.Results, 2, "expected 2 diagnostics, got error: %w", diag.Err)
+
+	t.Run("beat metrics", func(t *testing.T) {
```
pkoutsovasilis:
```suggestion
	t.Run("stats beat metrics", func(t *testing.T) {
```

internal/pkg/otel/manager/diagnostics_test.go
```diff
@@ -110,6 +129,74 @@ func TestPerformDiagnostics(t *testing.T) {
...
+		beatMetrics := diag.Results[0]
+		assert.Equal(t, "beat_metrics", beatMetrics.Name)
+		assert.Equal(t, "Metrics from the default monitoring namespace and expvar.", beatMetrics.Description)
+		assert.Equal(t, "beat_metrics.json", beatMetrics.Filename)
+		assert.Equal(t, "application/json", beatMetrics.ContentType)
+		assert.Equal(t, expectedMetricData, beatMetrics.Content)
+	})
+
+	t.Run("input metrics", func(t *testing.T) {
```
pkoutsovasilis:
```suggestion
	t.Run("input beat metrics", func(t *testing.T) {
```

internal/pkg/otel/manager/diagnostics.go
swiatekm: I'm not sure I understand. I add all the errors into the slice and then join them at the end. Would you prefer me to do `multiErr = errors.Join(multiErr, err)`? That'd result in unnecessary nesting.

internal/pkg/otel/manager/diagnostics.go
swiatekm: Those functions return slices because they were originally a single function returning multiple diagnostics, and it survived the refactor. But, don't you think the code looks nicer this way, without any conditionals? Both `append` and `errors.Join` handle nil values correctly.

internal/pkg/otel/manager/diagnostics_test.go
swiatekm: Sure. But I think it's a valid reason regardless. The http server dialer is an implementation detail of the test setup and has nothing to do with the code paths being tested here. So making it work on Windows would be annoying for no benefit.

internal/pkg/otel/manager/diagnostics.go
pkoutsovasilis: what's the added value between you controlling a slice of errors and not letting errors.Join manage it's internal slice of errors, especially since you do use errors.Join at the end so the same unnecessary nesting will occur?!

internal/pkg/otel/manager/diagnostics.go
pkoutsovasilis: hmmm actually no 🙂 I do value explicitness over appending unconditionally, thus if you don't have use-case for this being a slice I would highly recommend to switch the signature to return `*proto.ActionDiagnosticUnitResult, error`

internal/pkg/otel/manager/diagnostics_test.go
pkoutsovasilis: sure thing if it annoys you that much leave the skip reason as is 😄

internal/pkg/otel/manager/diagnostics.go
swiatekm: Sure, can do, it's no big deal.

internal/pkg/otel/manager/diagnostics.go
pkoutsovasilis: ok I did find some nesting happening inside errors.Join so even the outcome is the same let's avoid doing that

internal/pkg/otel/manager/diagnostics_test.go
swiatekm: I added my reasoning to the message.

internal/pkg/otel/manager/diagnostics.go
swiatekm: Done.

pkoutsovasilis (approved):

Issue #5235, opened 2024-07-31
Agent reports if there is insufficient disk space during upgrade, cleans up, and prevents upgrade


ycombinator:
When Agent tries to perform an upgrade, the local disk may not have sufficient space to a) download the new Agent artifact or b) to extract the downloaded Agent artifact.  When this happens, Agent should:
* reports this situation to the user (by reporting it to Fleet for Fleet-managed Agents), 
* cleans up any partially downloaded or extracted files, and
* prevents the upgrade from proceeding further

Acceptance critieria includes an E2E test proving the above steps are being taken.

nimarezainia:
@cmacknz Is this complicated to achieve? I think everything is there we just need the message sent when this situation arises.
I believe the download clean up is also happening today.

cc: @ebeahan

cmacknz:
It should not be complicated no. We will need to make sure we have a cross platform way to detect disk space on the correct drive (and permissions to do so) which will take some experimenting, but the upgrade status reporting to mark the upgrade as failed already exists.

michalpristas:
my take is that we should still treat runtime failures as primary source of truth and doing some pre-emptive checks (optionally in one take or 2 PRs) based on disk size and content-length.
there are few issues with pre-checks. first free disk space can change, we can have process poluting disk space that would make upgrade fail. we can check content-length if repository reports it and fail if we don't have sufficient space for download, but we actually don't know what is the size of archive + extracted agent upfront. 
we need to figure out on which volume target path is located and support detection of free space across different OSes and filesystems (or use some 3rd party library). Statfs used in [go library](https://pkg.go.dev/golang.org/x/sys/unix) is using `statfs` syscall that was deprecated by LSB

kaanyalti:
> my take is that we should still treat runtime failures as primary source of truth and doing some pre-emptive checks (optionally in one take or 2 PRs) based on disk size and content-length. there are few issues with pre-checks. first free disk space can change, we can have process poluting disk space that would make upgrade fail. we can check content-length if repository reports it and fail if we don't have sufficient space for download, but we actually don't know what is the size of archive + extracted agent upfront. we need to figure out on which volume target path is located and support detection of free space across different OSes and filesystems (or use some 3rd party library). Statfs used in [go library](https://pkg.go.dev/golang.org/x/sys/unix) is using `statfs` syscall that was deprecated by LSB

I agree with @Michal’s comment and wanted to share some thoughts and clarification questions.

[...]

PR #9122, opened 2025-07-24
Enhancement/5235 insufficient disk handling retry shows underlying error

kaanyalti:
**PR 1/6**

- Enhancement

## What does this PR do?

- Updates the upgrade downloaders to return insufficient disk error
- Introduces custom error for insufficient disk space
- Insufficient disk errors stops download retries

## Why is it important?

- The current error is less user friendly and includes unnecessary information
- Disk errors get replaced by context deadline error when retries time out, this pr fixes this for disk space errors.

## Checklist

## Disruptive User Impact

None, unless user relies on specific error messages in upgrade details in the status output.

- Build for windows, mac, linux
- Install agent (managed and standalone)
- Fill up disk until there is  approximately 200mb left
- Trigger upgrade
  - From fleet
  - Cli with remote url and file
- Validate the upgrade detail error message shows insufficient disk error message both in the status output and on fleet ui.

## Related issues

- Relates #5235 
- Prerequisite for #9322

### DIFF
--- a/internal/pkg/agent/application/upgrade/step_download_test.go
+++ b/internal/pkg/agent/application/upgrade/step_download_test.go
@@ -275,9 +275,10 @@ func TestDownloadWithRetries(t *testing.T) {
 
 	t.Run("insufficient disk space stops retries", func(t *testing.T) {
 		numberOfAttempts := 0
+		diskSpaceError := downloadErrors.OS_DiskSpaceErrors[0]
 		mockDownloaderCtor := func(version *agtversion.ParsedSemVer, log *logger.Logger, settings *artifact.Config, upgradeDetails *details.Details) (download.Downloader, error) {
 			numberOfAttempts++
-			return &mockDownloader{"", downloadErrors.ErrInsufficientDiskSpace}, nil
+			return &mockDownloader{"", diskSpaceError}, nil
 		}
 
 		a := newArtifactDownloader(&settings, testLogger)
@@ -293,7 +294,7 @@ func TestDownloadWithRetries(t *testing.T) {
 		require.Equal(t, "", path)
 
 		require.Equal(t, 1, numberOfAttempts)
-		require.ErrorIs(t, err, downloadErrors.ErrInsufficientDiskSpace)
+		require.ErrorIs(t, err, diskSpaceError)
 
 		require.NotZero(t, *upgradeDetailsRetryUntil)
 		require.False(t, *upgradeDetailsRetryUntilWasUnset)


internal/pkg/agent/application/upgrade/insufficient_disk_space_err_windows.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"errors"
+
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+
+	winSys "golang.org/x/sys/windows"
+)
+
+// ToDiskSpaceError returns a generic disk space error if the error is a disk space error
+func ToDiskSpaceErrorFunc(log *logger.Logger) func(error) error {
```
michalpristas: if you write this as a function in a similar manner than go provides error classification checks e.g [IsNotExist](https://cs.opensource.google/go/go/+/go1.24.5:src/os/error.go;l=80) you can save yourself good bunch of argument passing, making this PR way easier to review.

michalpristas: you can end up with following only on affected places
```go
if upgrade.IsInssuficientDisk(err){
 return upgrade.ErrInsufficientDisk
}
```

you can log more details optionally

kaanyalti: I think I can actually just use errors.As

kaanyalti: I thought of this again. In order to get the same functionality as I do now with go's api I would have to make the code less testable. I pretty much would have to do what I'm doing in the ToDiskSpaceError where I am calling it. So I would have to get rid of the abstraction I have. The only variable I'm passing is the logger, which I don't necessarily have to, it is just for better logging.

kaanyalti:
failing due to known flaky test

internal/pkg/agent/application/upgrade/errors/disk_space.go
```diff
@@ -0,0 +1,30 @@
...
+
+package errors
+
+import (
+	"errors"
+)
+
+const insufficientDiskSpaceErrorStr = "insufficient disk space"
+
+var ErrInsufficientDiskSpace = &InsufficientDiskSpaceError{Err: errors.New(insufficientDiskSpaceErrorStr)}
```
ycombinator: Could you get away with just the following (I think [you could](https://go.dev/play/p/IVXd5waMcCp))?
```suggestion
var ErrInsufficientDiskSpace = errors.New(insufficientDiskSpaceErrorStr)
```

If so, you wouldn't need the type below.

ycombinator:
Hey @kaanyalti, I'm having a bit of a hard time reviewing this PR because of the amount of changes in it.  I understand most of them are related to testing — making the code more testable and/or adding more test coverage — and that's a good thing.   But I do think these changes have made the code more indirect and harder to review.  

Would it be possible to break this PR up into multiple PRs, each one only containing a set of related changes?  For example, you're introducing a `ProgressReporter` type in this PR and tests + code related to it.  Would it be possible to extract these changes into their own separate PR, then have a follow up PR that builds upon them with another set of related changes, and so on? Ultimately, it would be ideal if there was a final PR in such a series that introduced `ErrInsufficientDiskSpace` and only any changes strictly related to that.

kaanyalti:
> Hey @kaanyalti, I'm having a bit of a hard time reviewing this PR because of the amount of changes in it. I understand most of them are related to testing — making the code more testable and/or adding more test coverage — and that's a good thing. But I do think these changes have made the code more indirect and harder to review.
> 
> Would it be possible to break this PR up into multiple PRs, each one only containing a set of related changes? For example, you're introducing a `ProgressReporter` type in this PR and tests + code related to it. Would it be possible to extract these changes into their own separate PR, then have a follow up PR that builds upon them with another set of related changes, and so on? Ultimately, it would be ideal if there was a final PR in such a series that introduced `ErrInsufficientDiskSpace` and only any changes strictly related to that.

Yes I can split the PR, will start posting the smaller PRs right away

kaanyalti:
After having a discussion with @ycombinator  I switched my approach to mocking stdlib functions and simplified the tests. This way I was able to remove much of the complexity from before and reduced the amount of changes. I won't be splitting the PR, however I'd still be happy to do so if it'll help reviewers.

internal/pkg/agent/application/upgrade/errors/disk_space.go
kaanyalti: I updated my implementation to use this approach

internal/pkg/agent/application/upgrade/insufficient_disk_space_err_windows.go
kaanyalti: I updated my implementation to use this approach

internal/pkg/agent/application/upgrade/common/stdlib_wrappers.go
```diff
@@ -0,0 +1,14 @@
...
+package common
+
+import (
+	"io"
+	"os"
+)
+
+var Copy = io.Copy
+var OpenFile = os.OpenFile
+var MkdirAll = os.MkdirAll
```
ycombinator: As I've mentioned a few times now off-PR, please add comments explaining the rationale for any unusual code you're introducing.  In this particular case, the normal thing for a developer to do would be to use `io.Copy`, etc. directly in their code.  So if they encounter these lines here, they're going to wonder why they exist.

internal/pkg/agent/application/upgrade/artifact/download/errors/disk_space_errors_unix.go
```diff
@@ -0,0 +1,14 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
ycombinator: Nit: It is traditional to name files that cover all OSes except one with the `_other` suffix (instead of the `_unix` suffix here).

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
```diff
@@ -61,6 +66,7 @@ func (e *Downloader) Download(ctx context.Context, a artifact.Artifact, version
 
 	hashPath, err := e.download(e.config.OS(), a, *version, ".sha512")
 	downloadedFiles = append(downloadedFiles, hashPath)
+
```
ycombinator: Nit: I think this was introduced unintentionally?

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
```diff
@@ -53,7 +55,7 @@ type Downloader struct {
 
 // NewDownloader creates and configures Elastic Downloader
 func NewDownloader(log *logger.Logger, config *artifact.Config, upgradeDetails *details.Details) (*Downloader, error) {
-	client, err := config.HTTPTransportSettings.Client(
+	client, err := config.Client(
```
ycombinator: Was this change necessary for this PR?  If not, please revert it.  If it's just cleanup/refactoring, please do that in a separate PR.  We have this documented in our contributing guidelines as well: https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#as-a-reviewee-ie-author-of-a-pr.

Same for the similar change below as well.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
```diff
@@ -108,18 +114,18 @@ func (e *Downloader) downloadFile(filename, fullPath string) (string, error) {
...
+		if err := common.MkdirAll(destinationDir, 0755); err != nil {
 			return "", err
 		}
 	}
 
-	destinationFile, err := os.OpenFile(fullPath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, packagePermissions)
+	destinationFile, err := common.OpenFile(fullPath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, packagePermissions)
 	if err != nil {
-		return "", errors.New(err, "creating package file failed", errors.TypeFilesystem, errors.M(errors.MetaKeyPath, fullPath))
+		return "", fmt.Errorf("%w: %w", errors.New("creating package file failed", errors.TypeFilesystem, errors.M(errors.MetaKeyPath, fullPath)), err)
```
ycombinator: I'm pretty sure you want the first verb / format-specifier to be `%v` here.  `%w` is meant for wrapping errors (see https://go.dev/blog/go1.13-errors#wrapping-errors-with-w).  In this case I see the `errors.New` error as the primary error and `err` as the error being wrapped.

Same goes for the other places in this PR that use the same construction.

internal/pkg/agent/application/upgrade/step_unpack_test.go
```diff
@@ -481,8 +481,7 @@ func checkExtractedFilesOutOfOrder(t *testing.T, versionedHome string) {
 	}
```
ycombinator: I thought you were going to handle unpacking in a follow up PR?  If so, please move the changes in this file to that PR. Again, as per our [contributing guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#as-a-reviewee-ie-author-of-a-pr), we should strive to keep our PRs atomic so they're easy to review.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
```diff
@@ -45,6 +46,10 @@ func (e *Downloader) Download(ctx context.Context, a artifact.Artifact, version
 	downloadedFiles := make([]string, 0, 2)
 	defer func() {
 		if err != nil {
+			if downloadErrors.IsDiskSpaceError(err) {
+				err = downloadErrors.ErrInsufficientDiskSpace
```
ycombinator: Why not just return `err` as-is here?  Why are you setting it to `downloadErrors.ErrInsufficientDiskSpace`?  That seems like atypical behavior so if there's a good reason for it, please add a comment here so it makes sense to other developers (as well as future you) looking at this code outside the context of this PR.

Same goes for other places in this PR where this assignment is being done.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader_test.go
```diff
@@ -291,3 +296,100 @@ func TestDownloader_DownloadAsc(t *testing.T) {
...
+					}
+				case "mkdirAll":
+					tmpMkdirAll := common.MkdirAll
+					t.Cleanup(func() {
+						common.MkdirAll = tmpMkdirAll
+					})
+					common.MkdirAll = func(path string, perm os.FileMode) error {
+						return tc.mockReturnedError
+					}
+				}
```
ycombinator: Consider extracting this into a helper function like `mockStdlibFunctions` or similar.  Not only does that make the actual test more readable but it also explains what this block of code is doing and why it is needed.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader_test.go
```diff
@@ -291,3 +296,100 @@ func TestDownloader_DownloadAsc(t *testing.T) {
...
+				require.NoError(t, err, "failed to create source artifact hash file")
+
+				downloader := NewDownloader(config)
+				targetArtifactPath, err := downloader.Download(context.Background(), agentSpec, parsedVersion)
+
+				require.ErrorIs(t, err, tc.expectedError)
+
+				if errors.Is(err, downloadErrors.ErrInsufficientDiskSpace) {
+					require.Equal(t, downloadErrors.ErrInsufficientDiskSpace, err)
+				}
```
ycombinator: I'm a bit confused by this block of code.  Isn't it sufficient to just check that `err` is equal to `tc.expectedError`?

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader_test.go
ycombinator: In fact, I see you are dong a very similar setup in `http.TestDownloadDiskSpaceError` as well.  You could probably parameterize your helper function such that you can call it from both places.  Of course, the tests being in different Go packages means you'll need to define this helper function in a non-test file so it can be shared.

Perhaps it should go in `common/stdlib_wrappers.go`, since that file exists solely for the purposes of testing anyway?

ycombinator:
> [kaanyalti](https://github.com/kaanyalti) requested a review from [ycombinator](https://github.com/ycombinator) [1 minute ago](https://github.com/elastic/elastic-agent/pull/9122#event-19118002678)

@kaanyalti There are still unaddressed comments on this PR from my initial review.  Please address all of them before re-requesting a review.  Thanks.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader_test.go
kaanyalti: I'm checking the error strings match as well so that the error displayed to the user does not contain unnecessary information that clutters the ui.

Mentioned in this comment https://github.com/elastic/elastic-agent/issues/5235#issuecomment-3113951311

ycombinator: Okay, again, this pattern of assertion is not typical, right?  So please comment here above the code what your _intent_ / _rationale_ is.  

In general, other developers can figure out _what_ the code is doing by reading the code itself.  But it's much harder, and sometimes impossible, to figure _why_ the code is doing what it is.  So whenever we introduce unusual patterns in the code (including test code), we should be in the habit of explaining our intent / rationale via inline comments.  We shouldn't expect other developers to do a git blame to figure out which PR a change was introduced in, then look at comments on that PR, follow links to issues, etc. to understand the reason behind a piece of code.  And, BTW, sometimes the "other developer" will be you a year from now, having forgotten all the context by then 😄.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader_test.go
kaanyalti: Added a mocker utility in the common dir and updated the tests to use the utility function.

internal/pkg/agent/application/upgrade/common/stdlib_wrappers.go
kaanyalti: Added comments

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
kaanyalti: Updated this to join the error and return. My goal was to be able to call errors.Is with ErrInsufficientDiskSpace. Now we're not losing the context of other errors.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader_test.go
kaanyalti: Removed this assertion. I changed how I propagate the errors, so this is assertion is not relevant anymore

internal/pkg/agent/application/upgrade/step_unpack_test.go
kaanyalti: This was unnecessarily committed, removed it

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
kaanyalti: Added comments explaining why this is added

kaanyalti:
Created this minor cleanup pr https://github.com/elastic/elastic-agent/pull/9346

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
kaanyalti: Using `%v` resulted in linting errors. Replaced this with errors.Join

ycombinator (approved):
LGTM!  Thanks for working through the review.

kaanyalti:
@ycombinator @michalpristas the pr is ready for review

changelog/fragments/1754914220-return-disk-space-error-message-when-agent-runs-out-of-space-while-upgrading.yaml
```diff
@@ -0,0 +1,32 @@
+# Kind can be one of:
```
ycombinator: I would actually remove this CHANGELOG entry and add it to the final PR of this collection that gets merged.  That way, if for some reason only some of these PRs make it into the next release, we don't have a CHANGELOG entry that's promising more than what we're actually delivering.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -197,6 +198,16 @@ func checkUpgrade(log *logger.Logger, currentVersion, newVersion agentVersion, m
 func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string, action *fleetapi.ActionUpgrade, det *details.Details, skipVerifyOverride bool, skipDefaultPgp bool, pgpBytes ...string) (_ reexec.ShutdownCallbackFn, err error) {
 	u.log.Infow("Upgrading agent", "version", version, "source_uri", sourceURI)
 
+	defer func() {
+		if err != nil {
+			// Add the disk space error to the error chain if it is a disk space error
+			// so that we can use errors.Is to check for it
+			if upgradeErrors.IsDiskSpaceError(err) {
```
michalpristas: it may already contain ErrInsufficientDiskSpace right?
just check if IsDiskSpaceError and is not ErrInsufficientDiskSpace

ycombinator:
> Validate the upgrade detail error message shows insufficient disk error message both in the status output and on fleet ui.

Here is the output of `sudo elastic-agent status` when trying to upgrade Elastic Agent (built with this PR) on a VM with limited disk space:
```
$ sudo elastic-agent status
┌─ fleet
├─ elastic-agent
└─ upgrade_details
   ├─ target_version: 9.1.2
   ├─ state: UPG_FAILED
   ├─ action_id: 34ae475d-4e54-4912-98b5-9df297190f0d
   └─ metadata
      ├─ failed_state: UPG_DOWNLOADING
      ├─ error_msg: failed download of agent binary: unable to download package: package '/opt/Elastic/Agent/data/elastic-agent-9.1.0-SNAPSHOT-419720/downloads/elastic-agent-9.1.2-linux-arm64.tar.gz' not found: open /opt/Elastic/Agent/data/elastic-agent-9.1.0-SNAPSHOT-419720/downloads/elastic-agent-9.1.2-linux-arm64.tar.gz: no such file or directory
      │  call to 'https://snapshots.elastic.co/9.1.0-4bd45409/downloads/beats/elastic-agent/elastic-agent-9.1.2-linux-arm64.tar.gz' returned unsuccessful status code: 404
      │  copying fetched package failed
      │  write /opt/Elastic/Agent/data/elastic-agent-9.1.0-SNAPSHOT-419720/downloads/elastic-agent-9.1.2-linux-arm64.tar.gz: no space left on device
      │  insufficient disk space
      └─ retry_until: 1h59m35.500609792s
```

@kaanyalti Is this what is expected, particularly in the `upgrade_details.metadata.error_msg` field?   Or should it only show the `insufficient disk space` error?  Personally, I'm good with the entire error chain that is shown here because it has lots of useful details, e.g. the path where the downloaded file is being written to.  But I'm wondering if showing the entire chain is in line with the requirements for this issue — I seem to recall somewhere (but can't find it now) that we wanted the `insufficient disk space` error to bubble up to the very top of the error message (or maybe I'm misremembering)?

kaanyalti:
@ycombinator That is exactly the error that is expected here. This PR https://github.com/elastic/elastic-agent/pull/9392 updates the coordinator to provide only "insufficient disk space" error message. I have also mentioned this in the comment here https://github.com/elastic/elastic-agent/issues/5235#issuecomment-3197547324

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: The downloaders do not add `ErrInsufficientDiskSpace` in the error chain. Both the fs and http downloader test assert that the errors propagate out of the downloaders as they are.

One minor exception to this is the httpy downloader. The http downloader checks for disk space error and replaces the error with `ErrInsufficientDiskSpace` only when calling the reporter so that the status reported shows the concise "insufficient disk space error" message. The original error is propagated up the call stack.

changelog/fragments/1754914220-return-disk-space-error-message-when-agent-runs-out-of-space-while-upgrading.yaml
kaanyalti: Removed the changelog, will add it in a separate PR

kaanyalti:
@ycombinator just removed the changelog fragment, will add it in a separate PR

ycombinator (approved):
LGTM pending @michalpristas's feedback.

kaanyalti:
@ycombinator and @michalpristas I will make changes to the stdlib wrappers based on the comments here 
https://github.com/elastic/elastic-agent/pull/9349#discussion_r2285041353. Once the stdlib wrapper changes in #9349 are finalized, I will update this PR accordingly.

blakerouse (dismissed):
This change overall looks good and has great coverage. The mocking of the standard library functions are strange, but I understand why they needed to be done this way. Provides the ability for well written tests and coverage.

michalpristas (approved):

pkoutsovasilis:
> This change overall looks good and has great coverage. The mocking of the standard library functions are strange, but I understand why they needed to be done this way. Provides the ability for well written tests and coverage.

Sorry gents, but I still don’t see the benefit of exposing stdlib functions and their mocks in the way this PR currently does.  

As an example [here](https://github.com/elastic/elastic-agent/pull/9122/files#diff-70970162bac440b48d75bb2a8ece0708b101d9b401e55873a23c7056e30b984aR1447-R1473) and [here](https://github.com/elastic/elastic-agent/pull/9122/files#diff-c2ce251f8a805bfd66ce61b76a34439ebd0e7f9cf4708928b35f6394e1db03c6R46-R56), we have to define the body of each mocked stdlib func and then call another func that overrides the public vars in the `common` package. I don’t see how this is better than simply having private package-scoped vars for the stdlib funcs we want to override. Tests in the same package could then replace them directly according to their needs, without importing a package that exposes globals intended for overrides. That feels safer, simpler, and more consistent.

That said, I understand opinions may differ. Please just note that since such a change is accepted I can’t move forward with my reviews, so I’d appreciate if the rest of the PRs assigned to me could also be reviewed by either you @ycombinator, @blakerouse, @michalpristas.

blakerouse:
@pkoutsovasilis As stated I believe the was the mocking was done is rather weird exposing them that way, but I understood the reason. I also agree with you that it could be done privately, and that would be a bit cleaner.

@kaanyalti Could you make these private to each module that uses them and just do a simple mock that way? Now that I read this again I do believe @pkoutsovasilis is correct and placing these in each module with a private var would be cleaner and safer.

kaanyalti:
@blakerouse I do agree with @pkoutsovasilis as well, and will update the stdlib abstractions to be private package scoped vars. 
I have mentioned in this comment thread https://github.com/elastic/elastic-agent/pull/9349#discussion_r2277581965 that I will keep the mock utility as it does clean up the tests quite a bit. This mock utility will have to be in a separate package to avoid circular references.

kaanyalti:
> @blakerouse I do agree with @pkoutsovasilis as well, and will update the stdlib abstractions to be private package scoped vars. I have mentioned in this comment thread [#9349 (comment)](https://github.com/elastic/elastic-agent/pull/9349#discussion_r2277581965) that I will keep the mock utility as it does clean up the tests quite a bit. This mock utility will have to be in a separate package to avoid circular references.

Revisiting my approach here, I cannot simply make stdlib vars private. I am using these vars to mock file system interactions in upgrade_test.go. Testing the upgrade function behavior without these mocks is not straightforward because downloadArtifacts cannot be mocked. 

I'm going to address stdlib use case by case. For downloaders I will introduce an interface encapsulating MkdirAll, Copy, and OpenFile. For some of the cases in the upgrade package I will have to use the approach I followed here because not all of the functions that call stdlib belong to the upgrader struct. Properly injecting these functions would require changes in function signatures and in the upgrader struct itself. I had tried doing this before going with variable abstraction as it had introduced a lot of changes which made the prs hard to review.

kaanyalti:
@ycombinator @michalpristas @blakerouse @pkoutsovasilis the PR is ready for review

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -101,10 +113,12 @@ func (u *Upgrader) SetClient(c fleetclient.Sender) {
 	if c == nil {
 		u.log.Debug("client nil, resetting Fleet Server URI")
 		u.fleetServerURI = ""
+		u.artifactDownloader.withFleetServerURI("")
```
kaanyalti, line 116: downloadArtifacts was using the server uri from upgrader, now that it is separated from the upgrader the server uri needs to be set explicitly

kaanyalti:
Quick update, caught a logic error. IsDiskSpaceError should not be returning true for ErrInsufficientDiskSpace. With the current implementation we can end up chaining multiple ErrInsufficientDiskSpace which is not what we want to do. I will address this tomorrow in the morning in a new commit, but the rest of the PR is still valid and ready for review.

@ycombinator @pkoutsovasilis @blakerouse @michalpristas

kaanyalti:
> Quick update, caught a logic error. IsDiskSpaceError should not be returning true for ErrInsufficientDiskSpace. With the current implementation we can end up chaining multiple ErrInsufficientDiskSpace which is not what we want to do. I will address this tomorrow in the morning in a new commit, but the rest of the PR is still valid and ready for review.
> 
Fixed this logic error in my last commit

@ycombinator @pkoutsovasilis @blakerouse @michalpristas

kaanyalti:
Adressing ci failure

kaanyalti:
Addressed the ci failure with my last commit

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
```diff
@@ -6,6 +6,7 @@ package fs
 
 import (
 	"context"
+	goerrors "errors"
```
ycombinator, line 9: Not for this PR, but for a follow up PR: we should be able to use this `errors` package from stdlib for all error handling in this file, instead of using the `github.com/elastic/elastic-agent/internal/pkg/agent/errors` package in some places.

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
```diff
@@ -49,6 +51,12 @@ type Downloader struct {
 	config         *artifact.Config
 	client         http.Client
 	upgradeDetails *details.Details
+	// The following are abstractions for stdlib functions so that we can mock them in tests.
+	copy     func(dst io.Writer, src io.Reader) (int64, error)
+	mkdirAll func(name string, perm os.FileMode) error
+	openFile func(name string, flag int, perm os.FileMode) (*os.File, error)
+	// Abstraction for the disk space error check function so that we can mock it in tests.
+	isDiskSpaceErrorFunc func(err error) bool
```
ycombinator, line 59: Nit: I don't think you need the `Func` at the end.  The type and usage of this field shows that it's a function.

```suggestion
	isDiskSpaceError func(err error) bool
```

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -1292,3 +1297,79 @@ func (f *fakeAcker) Commit(ctx context.Context) error {
...
+	fleetServerURI string
+}
+
+func (m *mockArtifactDownloader) downloadArtifact(ctx context.Context, parsedVersion *agtversion.ParsedSemVer, sourceURI string, upgradeDetails *details.Details, skipVerifyOverride, skipDefaultPgp bool, pgpBytes ...string) (_ string, err error) {
+	return "", m.returnError
+}
+
+func (m *mockArtifactDownloader) withFleetServerURI(fleetServerURI string) {
+	m.fleetServerURI = fleetServerURI
+}
```
ycombinator, line 1312: Simplification: Since all the invocations of this method are from the same package, `upgrade`, I think you don't need the method.  You could simply set the `fleetServerURI` field on the downloader object directly.  We can introduce a method if/when code outside the `upgrade` package ever needs to set the `fleetServerURI`.

kaanyalti: The withFleetServerURI function is there because I'm using `artifactDownloadHandler` interface in the upgrader instead of the `artifactDownloader` struct so that I can mock the downloadArtifact function. I can get the underlying type and set it that way but defeats the purpose of using the interface.

ycombinator: Okay, I see what you mean.  Makes sense the way you have it.

internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go
kaanyalti: Noted, thank you.

internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go
kaanyalti: I will address this and the comment above in a separate PR as they are nits.

PR #9133, opened 2025-07-25
Cache binaries downloaded for packaging locally

swiatekm:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,11 @@ To build a local version of the agent for development, run the command below. Th
 DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64 PACKAGES=tar.gz mage -v package
 ```
 
+If you build the same agent package often (when running integration tests, for example),
+you can also set KEEP_ARCHIVE=true in your environment. The packaging step will then
+avoid deleting the binary archive after the package is generated, removing the need to
+re-download binaries on every invocaion.
+
 The resulting package will be produced in the build/distributions directory. The version is controlled by the value in [version.go](version/version.go).
 To install the agent extract the package and run the install command:
 
--- a/magefile.go
+++ b/magefile.go
@@ -1196,11 +1196,15 @@ func packageAgent(ctx context.Context, platforms []string, dependenciesVersion s
 		log.Printf("dependencies extracted from package specs: %v", dependencies)
 	}
 
+	keepArchive := os.Getenv("KEEP_ARCHIVE") != ""
+
 	// download/copy all the necessary dependencies for packaging elastic-agent
 	archivePath, dropPath, dependencies := collectPackageDependencies(platforms, dependenciesVersion, packageTypes, dependencies)
 
 	// cleanup after build
-	defer os.RemoveAll(archivePath)
+	if !keepArchive {
+		defer os.RemoveAll(archivePath)
+	}
 	defer os.RemoveAll(dropPath)
 	defer os.Unsetenv(agentDropPath)
 
--- a/dev-tools/mage/downloads/utils.go
+++ b/dev-tools/mage/downloads/utils.go
@@ -7,70 +7,46 @@ package downloads
 import (
 	"fmt"
 	"io"
+	"net/http"
 	"os"
-	"path/filepath"
-	"strings"
 
 	"github.com/cenkalti/backoff/v4"
-	"github.com/gofrs/uuid/v5"
 )
 
 // downloadRequest struct contains download details ad path and URL
 type downloadRequest struct {
-	URL                 string
-	DownloadPath        string
-	UnsanitizedFilePath string
+	URL        string
+	TargetPath string
 }
 
 // downloadFile will download a url and store it in a temporary path.
 // It writes to the destination file as it downloads it, without
 // loading the entire file into memory.
 func downloadFile(downloadRequest *downloadRequest) error {
-	var filePath string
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -108,6 +108,11 @@ To build a local version of the agent for development, run the command below. Th
 DEV=true EXTERNAL=true SNAPSHOT=true PLATFORMS=linux/amd64 PACKAGES=tar.gz mage -v package
 ```
 
+If you build the same agent package often (when running integration tests, for example),
+you can also set KEEP_ARCHIVE=true in your environment. The packaging step will then
+avoid deleting the binary archive after the package is generated, removing the need to
+re-download binaries on every invocaion.
```
pkoutsovasilis:
```suggestion
re-download binaries on every invocation.
```

magefile.go
```diff
@@ -1196,12 +1196,16 @@ func packageAgent(ctx context.Context, platforms []string, dependenciesVersion s
 		log.Printf("dependencies extracted from package specs: %v", dependencies)
 	}
 
+	keepArchive := os.Getenv("KEEP_ARCHIVE") != ""
```
pkoutsovasilis, line 1202: should `KEEP_ARCHIVE` env var also affect the `mage clean` target?

swiatekm: I don't think so. `mage clean` is a clear signal that the user wants all the build-related data gone. Although it's true that we're not very consistent about what stays and what is dropped during the build itself.

pkoutsovasilis: that was exactly my point; `mage clean` as of today cleans many stuff, e.g. `.agent-testing` folder, which more test-related than build-related data. Thus I would find it useful to be able to tell clean everything else except the downloaded archives, but if you feel that is gonna hinder the developer experience, sure thing

swiatekm:
> I did look at the code and the endgoal makes total sense to me and it is welcome improvement. There is only this part that we rely on the server telling us if something has changed or not (by using "If-Modified-Since" http header) that I am not sure how I feel about it. By leveraging the latter we kinda "expose" our packaging to any bugs of the artifacts server (has caused us issues in the past). Maybe a more controlled by us solution would be to download first the sha512 file compare it with what we have locally and if different download. wdyt?

For what it's worth, I think our artifacts API just passes that header to the underlying object storage, which should have a correct implementation. I originally thought we always check the downloaded file against the hash, but it seems like we might not? If we did, that would solve the problem, because even if the artifacts api was wrong, we'd get a hash mismatch and at least see the problem.

For the record, I think actual packaging steps in the CI should always redownload the artifacts, ideally by calling `mage clean`.

pkoutsovasilis:
> For what it's worth, I think our artifacts API just passes that header to the underlying object storage, which should have a correct implementation. I originally thought we always check the downloaded file against the hash, but it seems like we might not?

and for this exact reason of having to speculate when we rely on an external server, I would prefer utilising the `sha512` files. (PS: if what you say is true and the underlying storage relies on the modified time of the file, is not the most reliable thing out there). Also, as you do mention above, you would solve another problem we currently have of not validating the components we download by hash, win win scenario 🙂

swiatekm:
> > For what it's worth, I think our artifacts API just passes that header to the underlying object storage, which should have a correct implementation. I originally thought we always check the downloaded file against the hash, but it seems like we might not?
> 
> and for this exact reason of having to speculate when we rely on an external server, I would prefer utilising the `sha512` files. (PS: if what you say is true and the underlying storage relies on the modified time of the file, is not the most reliable thing out there). Also, as you do mention above, you would solve another problem we currently have of not validating the components we download by hash, win win scenario 🙂

I would consider `If-Changed-Since` headers on say, S3 or GCS, to be reliable. Surely there's tons of code out there that relies on this working correctly. But the integrity check is useful regardless, because the mod time on the client isn't necessarily reliable, as many processes can easily mess with it, even by accident.

In the meantime, how about I submit a PR with just the download refactor? It has benefits on its own, like not passing the content of downloaded files as strings. :sweat_smile:

michalpristas:
i suspect this is primarily for local usage. 
also i would prefer to have other option as well meaning cleanup before build in case i had last build with KEEP and i changed my mind

swiatekm:
> i suspect this is primarily for local usage. also i would prefer to have other option as well meaning cleanup before build in case i had last build with KEEP and i changed my mind

Isn't `mage clean` good enough for that? I'd prefer not to add even more environment variables. We already have a ton.

swiatekm:
I've rebased this change on #9218, making it much simpler. With https://github.com/elastic/elastic-agent/pull/9226, we always verify checksums, so we'll immediately know if the artifacts API served us a stale download, or if the file is corrupted for whatever reason. So this change should be quite safe now. It's only really intended for local use, although we can later look into keeping the downloaded binaries on CI runners, if we want.

@pkoutsovasilis @michalpristas can you review again? :pray:

pkoutsovasilis (approved):

pkoutsovasilis:
@swiatekm please merge the latest main in here; the failure you see is because I merged this one https://github.com/elastic/elastic-agent/pull/9832 and I managed to hit your CI run just right in the middle; so the clean up step run with the latest code in main but the pipeline was loaded before the merge so any changes are lost. 

PS: we are working on a solution to get away from such failures

pkoutsovasilis (approved):

PR #9148, opened 2025-07-28
[Migrate agent to different cluster] Accept signed action 

michalpristas:
This PR adds support for signed actions. When an action is signed, we use the content embedded in the signature.

Validation behavior:
- If a validation key is configured:
  - The action must be signed.
  - If it’s unsigned or the signature is malformed, execution fails.
- If no validation key is configured:
  - Unsigned actions are allowed.


michalpristas:
Need to figure out why endpoint is failing

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -570,20 +570,20 @@ func (c *Coordinator) StateSubscribe(ctx context.Context, bufferLen int) chan St
...
+	c.mx.RLock()
+	defer c.mx.RUnlock()
+	return c.protection
+}
+
+// setProtection sets protection configuration
+func (c *Coordinator) setProtection(protectionConfig protection.Config) {
+	c.mx.Lock()
+	c.protection = protectionConfig
+	c.mx.Unlock()
```
ycombinator: Nit: I know you're just uncommenting existing code here but is there some reason not to use `defer` here?

```suggestion
	defer c.mx.Unlock()
	c.protection = protectionConfig
```

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1490,10 +1496,11 @@ func (c *Coordinator) processConfigAgent(ctx context.Context, cfg *config.Config
...
 
-	// Disabled for 8.8.0 release in order to limit the surface
-	// https://github.com/elastic/security-team/issues/6501
-
-	// c.setProtection(protectionConfig)
+	protectionConfig, err := protection.GetAgentProtectionConfig(m)
+	if err != nil && !errors.Is(err, protection.ErrNotFound) {
+		return fmt.Errorf("could not read the agent protection configuration: %w", err)
+	}
+	c.setProtection(protectionConfig)
```
ycombinator, line 1501: Since you are reintroducing this code (by uncommenting it), should the following section also be uncommented/reintroduced? https://github.com/elastic/elastic-agent/blob/64e5457f653c1f11a7fc60d07b5adeaf6092a2f3/internal/pkg/agent/application/actions/handlers/handler_action_application.go#L53-L66

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -341,11 +343,9 @@ type Coordinator struct {
 	// value that is sent to the runtime manager).
 	componentModel []component.Component
 
-	// Disabled for 8.8.0 release in order to limit the surface
-	// https://github.com/elastic/security-team/issues/6501
-
-	// mx         sync.RWMutex
-	// protection protection.Config
+	// Protection section
+	mx         sync.RWMutex
```
ycombinator: Looking at the call paths to read and write the `protection` field below, are we sure we need a mutex?  I'm not seeing any concurrent access (but I might've missed it).

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: no this is different action that we should not touch at this step

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: yes atm this should be ok, it probably can be different once we introduce other actions. you think this is too much of a toll?

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: i know it's a micro-optimization but I also prefer no defer in simple cases as this one to save some cycles

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: Yeah, it's a different action but does it need to also start doing the signature validation?  Just clarifying if this is something we should a) do in a follow up PR or b) not touch at all?

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: It's not so much about saving cycles as forgetting to unlock in case this code grows in complexity and we have some early returns, etc.

internal/pkg/agent/application/coordinator/coordinator.go
ycombinator: I might be misremembering but didn't @faec go through the coordinator code at one point and make it so it didn't have any mutexes because they were causing some flaky behavior?  That's the main reason I don't want to reintroduce them unless we actually need them.

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: we should do this in a followup
linked issue says:
> The infrastructure for action signing was implemented as part of the tamper protection implementation and should be enabled as part of this work, initially scoped only to the migrate action to reduce risk.

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: yes indeed, i also went through it once again to make sure it's safe to remove. it's gone now

PR #9198, opened 2025-07-30
[CI][FIPS] Reconfigure pipeline to use Staging GovCloud/FRH ESS environment

ycombinator:
## What does this PR do?

This PR reconfigures the FIPS Buildkite pipeline to a) point it to the Staging GovCloud/FRH ESS environment and b) use an ESS API key from that environment for spinning up deployments.

## Why is it important?

To run FIPS-related tests against the officially-configured FedRamp High (FRH) ESS environment.

### DIFF
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -24,6 +24,7 @@ steps:
       ASDF_TERRAFORM_VERSION: 1.9.2
       FIPS: "true"
       TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips:git-${BUILDKITE_COMMIT:0:12}"
+      EC_ENDPOINT: "https://api.staging.elastic-gov.com"
     command: |
       source .buildkite/scripts/steps/ess_start.sh
     artifact_paths:
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -24,7 +24,7 @@ steps:
       ASDF_TERRAFORM_VERSION: 1.9.2
       FIPS: "true"
       EC_ENDPOINT: "https://api.staging.elastic-gov.com"
-      TF_VAR_ess_region: "us-gov-east-1"
+      ESS_REGION: "us-gov-east-1"
       TF_VAR_deployment_template_id: "aws-general-purpose"
       TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips:git-${BUILDKITE_COMMIT:0:12}"
       TF_VAR_docker_images_name_suffix: "-fips"
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -25,7 +25,6 @@ steps:
       FIPS: "true"
       EC_ENDPOINT: "https://api.staging.elastic-gov.com"
       ESS_REGION: "us-gov-east-1"
-      STACK_BUILD_ID: ""
       TF_VAR_deployment_template_id: "aws-general-purpose"
       TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips:git-${BUILDKITE_COMMIT:0:12}"
       TF_VAR_docker_images_name_suffix: "-fips"
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -24,9 +24,11 @@ steps:
       ASDF_TERRAFORM_VERSION: 1.9.2
       FIPS: "true"
       EC_ENDPOINT: "https://api.staging.elastic-gov.com"
-      TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips:git-${BUILDKITE_COMMIT:0:12}"
       TF_VAR_ess_region: "us-gov-east-1"
       TF_VAR_deployment_template_id: "aws-general-purpose"
+      TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips:git-${BUILDKITE_COMMIT:0:12}"
+      TF_VAR_elasticsearch_docker_image: "docker.elastic.co/cloud-release/elasticsearch-cloud-ess-fips:latest"
+      TF_VAR_kibana_docker_image: "docker.elastic.co/cloud-release/kibana-cloud-fips:latest"
     command: |
       source .buildkite/scripts/steps/ess_start.sh
     artifact_paths:
--- a/.buildkite/bk.integration-fips.pipeline.yml
+++ b/.buildkite/bk.integration-fips.pipeline.yml
@@ -28,8 +28,6 @@ steps:
       TF_VAR_deployment_template_id: "aws-general-purpose"
       TF_VAR_integration_server_docker_image: "docker.elastic.co/beats-ci/elastic-agent-cloud-fips:git-${BUILDKITE_COMMIT:0:12}"
       TF_VAR_docker_images_name_suffix: "-fips"
-#      TF_VAR_elasticsearch_docker_image: "docker.elastic.co/cloud-release/elasticsearch-cloud-ess-fips:latest"
-#      TF_VAR_kibana_docker_image: "docker.elastic.co/cloud-release/kibana-cloud-fips:latest"
     command: |
       source .buildkite/scripts/steps/ess_start.sh
     artifact_paths:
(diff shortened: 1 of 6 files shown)


.buildkite/bk.integration-fips.pipeline.yml
```diff
@@ -11,9 +11,9 @@ env:
 # This section is used to define the plugins that will be used in the pipeline.
 # See https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins
 common:
-  - vault_ec_key_prod: &vault_ec_key_prod
+  - vault_ec_key_staging_frh_gov: &vault_ec_key_staging_frh_gov
       elastic/vault-secrets#v0.1.0:
-        path: "kv/ci-shared/platform-ingest/platform-ingest-ec-prod"
+        path: "kv/ci-shared/platform-ingest/platform-ingest-ec-staging-frh-gov"
```
ycombinator: This is a made-up path for now.  Will update once I have the actual path from the Observability Eng Prod team (issue in private repo so not linking here).

.buildkite/scripts/steps/ess.sh
```diff
@@ -17,14 +16,14 @@ function ess_up() {
   BUILDKITE_BUILD_CREATOR="${BUILDKITE_BUILD_CREATOR:-"$(get_git_user_email)"}"
   BUILDKITE_BUILD_NUMBER="${BUILDKITE_BUILD_NUMBER:-"0"}"
   BUILDKITE_PIPELINE_SLUG="${BUILDKITE_PIPELINE_SLUG:-"elastic-agent-integration-tests"}"
-  
-  pushd "${TF_DIR}"    
+
+  export TF_LOG=DEBUG
```
ycombinator: This is intended for debugging and should be removed before this PR is approved/merged.

ycombinator:
@pkoutsovasilis I could use your input here on how best to resolve the CI failure on this PR, since it pertains to the contents of the `.package-version` file.

The "Start ESS stack for FIPS integration tests" step is [failing](https://buildkite.com/elastic/elastic-agent/builds/25257#0198a656-0c89-4133-85c0-a47731bc1281/146-306) in CI like so:

```
│ Error: failed creating deployment
--
  | │
  | │   with ec_deployment.integration-testing,
  | │   on deployment.tf line 100, in resource "ec_deployment" "integration-testing":
  | │  100: resource "ec_deployment" "integration-testing" {
  | │
  | │ api error: 1 error occurred:
  | │ 	* clusters.cluster_plan_change_prohibited: The requested cluster configuration change is not permitted. Value of field 'elasticsearch.docker_image' cannot
  | │ be changed to [docker.elastic.co/cloud-release/elasticsearch-cloud-ess-fips:9.2.0-0c2a1cac-SNAPSHOT]. (resources.elasticsearch[0].elasticsearch.docker_image)
  | │
```

This is happening because there is no Docker image with this tag (`9.2.0-0c2a1cac-SNAPSHOT`) in the **Staging GovCloud** ESS environment, which is the environment that we run our FIPS tests against.  The tag comes from https://github.com/elastic/elastic-agent/blob/c1de40fe692fcc8288e85a9fbfbd839d4a7b7e1d/.package-version#L7

That field is being read in the terraform file used for creating the ESS deployment over here: https://github.com/elastic/elastic-agent/blob/c1de40fe692fcc8288e85a9fbfbd839d4a7b7e1d/test_infra/ess/deployment.tf#L82

Just for testing, I temporarily hardcoded the tag as `9.2.0-SNAPSHOT` in the terraform file and it worked.  So I know the issue is isolated to the tag not existing in the environment, and everything else in our configuration is OK.

Do you have any thoughts on how we could accommodate the Staging GovCloud environment in our testing configuration?  

Here is one idea I have; let me know if you have a better suggestion:

1. In https://github.com/elastic/elastic-agent/blob/main/.package-version, create a new field, `stack_version`, which will have the same value as `stack_build_id` but without the SHA.  So, today, the value of this new field would be `9.2.0-SNAPSHOT`.   The automation that keeps that file updated will set this new field as well.
2.  We pass this new field value to the Terraform file in addition to passing `stack_build_id` today. 
4. In the Terraform file, we introduce a new boolean variable `fips`.  The FIPS Buildkite pipeline will pass this value as `true` to the Terraform file.
5. In the Terraform file, if `fips` is set to `true`, we use the value of `stack_version` otherwise we use the value of `stack_build_id`, to create the image tag. 

Another solution, similar to the above, but needing more automation would be to store something like a `frh_stack_build_id` field in the `.package-version` file and have automation set that based on trying to find the stack build ID actually available in the ESS Staging GovCloud environment that matches `stack_build_id` the closest.  

Keep in mind that we have a deadline to finish FIPS testing by the end of this week so we may need to go with a less-ideal solution for now and then revisit it to make it better.

pkoutsovasilis:
Hey @ycombinator 👋 I do like the following proposal
> In https://github.com/elastic/elastic-agent/blob/main/.package-version, create a new field, stack_version, which will have the same value as stack_build_id but without the SHA. So, today, the value of this new field would be 9.2.0-SNAPSHOT. The automation that keeps that file updated will set this new field as well.
We pass this new field value to the Terraform file in addition to passing stack_build_id today.
In the Terraform file, we introduce a new boolean variable fips. The FIPS Buildkite pipeline will pass this value as true to the Terraform file.
In the Terraform file, if fips is set to true, we use the value of stack_version otherwise we use the value of stack_build_id, to create the image tag.

but let's move fast to get this up and running and we figure the details of how `.package-version` can be utilised here afterwards. What do you think of the following?

1. [here](https://github.com/elastic/elastic-agent/blob/c1de40fe692fcc8288e85a9fbfbd839d4a7b7e1d/.buildkite/scripts/steps/integration_tests_tf.sh#L27) do a change in the fashion of `if FIPS=TRUE then STACK_BUILD_ID=""`
2. similar to the above [here](https://github.com/elastic/elastic-agent/blob/c1de40fe692fcc8288e85a9fbfbd839d4a7b7e1d/.buildkite/scripts/steps/ess_start.sh#L8)
3. windows you don't care at the moment, but it's similar to the above [here](https://github.com/elastic/elastic-agent/blob/c1de40fe692fcc8288e85a9fbfbd839d4a7b7e1d/.buildkite/scripts/steps/integration_tests_tf.ps1#L19)

this will result using only the STACK_VERSION and not the STACK_BUILD_ID

ycombinator:
The "Start ESS stack for FIPS integration tests" step in CI is [failing](https://buildkite.com/elastic/elastic-agent/builds/25278#0198a99d-2353-4523-8e24-77765c9f786e/146-304) with the error:

```
ec_deployment.integration-testing: Creating...
--
  | ╷
  | │ Error: failed creating deployment
  | │
  | │   with ec_deployment.integration-testing,
  | │   on deployment.tf line 100, in resource "ec_deployment" "integration-testing":
  | │  100: resource "ec_deployment" "integration-testing" {
  | │
  | │ api error: 1 error occurred:
  | │ 	* clusters.cluster_plan_change_prohibited: The requested cluster configuration change is not permitted. Value of field 'elasticsearch.docker_image' cannot
  | │ be changed to [docker.elastic.co/cloud-release/elasticsearch-cloud-ess-fips:9.2.0-SNAPSHOT]. (resources.elasticsearch[0].elasticsearch.docker_image)
  | │
  | │
  | ╵
  | ╷
  | │ Error: failed creating deployment
  | │
  | │   with ec_deployment.integration-testing,
  | │   on deployment.tf line 100, in resource "ec_deployment" "integration-testing":
  | │  100: resource "ec_deployment" "integration-testing" {
```

I'm confused because the same image tag succeeded in the previous CI build: https://buildkite.com/elastic/elastic-agent/builds/25276#0198a930-20fb-4c9e-93c3-80392012a236/136-180

[EDIT] I've compared the Terraform execution plans from both CI builds ([previous - successful](https://buildkite.com/elastic/elastic-agent/builds/25276#0198a930-20fb-4c9e-93c3-80392012a236/136-166) vs. [current - failed](https://buildkite.com/elastic/elastic-agent/builds/25278#0198a99d-2353-4523-8e24-77765c9f786e/146-176)) and they are identical except for the `docker.elastic.co/beats-ci/elastic-agent-cloud-fips` image's SHA (which is expected to be different since we're building different commits) and the `buildkite_id` tag (which is also expected to be different as it's different builds).

[EDIT] I've also run `terraform apply` manually with the [same terraform configuration](https://github.com/elastic/elastic-agent/tree/main/test_infra/ess) that's being used in CI like so:
```
TF_VAR_docker_images_name_suffix=-fips \
TF_VAR_stack_version=9.2.0-SNAPSHOT \
TF_VAR_integration_server_docker_image="docker.elastic.co/cloud-release/elastic-agent-cloud-fips:9.2.0-SNAPSHOT" \
TF_VAR_deployment_template_id="aws-general-purpose" \
TF_VAR_ess_region="us-gov-east-1" \
EC_API_KEY="XXXXXXX" \
EC_ENDPOINT="https://api.staging.elastic-gov.com" \
terraform apply
```

I've run this three times so far, one after the other (with a `terraform destroy` in between).  The first attempt succeeded, the second attempt failed with the same error that we're seeing in CI, and the third attempt succeeded as well.  So I'm starting to think something about the Staging GovCloud environment might be flaky?

michel-laterman (approved):

oakrizan (approved):
lgtm for CI part

PR #9199, opened 2025-07-30
Don't overwrite existing headers when mapping agentInfo headers into elasticsearch output

### LINKED ISSUE #9197: Enrollment with --headers override headers with the same name in the output configuration
At the moment if you enroll the Elastic Agent with `--headers` then those headers are then placed into the agent information file. Those headers are sent both to Fleet Server and Elasticsearch. There is a case (agentless) where those need to be different.

The configuration includes a different header value in the configuration but the Elastic Agent overwrites it with the value from the agent information. I believe this should only place the header value there if there is not already a header present. If the header is present it should take priority over the header in agent information.

blakerouse:
## What does this PR do?

Changes the behavior of `toIntermediate` to only add headers to an output if the output doesn't already have that header defined.

## Why is it important?

This allows the use-case where the same header name is used for Fleet and Elasticsearch, but the value of each is different.

## Checklist

## Disruptive User Impact

None

## Related issues

- Closes #9197

### DIFF
--- a/pkg/component/component_test.go
+++ b/pkg/component/component_test.go
@@ -2166,7 +2166,8 @@ func TestToComponents(t *testing.T) {
 						"type":    "elasticsearch",
 						"enabled": true,
 						"headers": map[string]interface{}{
-							"header-two": "val-2",
+							"header-two":   "val-2",
+							"header-three": "val-3",
 						},
 					},
 				},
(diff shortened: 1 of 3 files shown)


pkg/component/component_test.go
```diff
@@ -2214,7 +2216,8 @@ func TestToComponents(t *testing.T) {
 				},
 			},
 			headers: &testHeadersProvider{headers: map[string]string{
-				"header-one": "val-1",
+				"header-one":   "val-1",
+				"header-three": "val-3-diff",
```
ycombinator, line 2220: Just making sure I understand the precedence order here correctly: this value of `header-three`, `val-3-diff` gets evaluated second, and therefore it should **not** be used since `header-three` was already set to `val-3` first, correct?

blakerouse: Correct.

ycombinator (approved):

PR #9218, opened 2025-07-31
Refactor binary downloads in agent packaging targets

swiatekm:
## What does this PR do?

Refactors how the agent packaging target downloads binaries. The intent is to make it possible to cache the binaries locally and avoid having to redownload them. See https://github.com/elastic/elastic-agent/pull/9133 for the full context.

The PR consists of the following changes:

* The download function now always gets a target path as an input, so it doesn't need to create it on its own conditionally.
* We use standard library functions for the download, so we don't need to pass the content as a string, and can instead write it directly to the target file, saving memory.
* Remove some unused functions.

## Why is it important?

This makes the download code more maintainable and more performant. It also unlocks further packaging improvements.

## Checklist

## Related issues

- Relates https://github.com/elastic/observability-dev/issues/4702

### DIFF
--- a/dev-tools/mage/downloads/utils.go
+++ b/dev-tools/mage/downloads/utils.go
@@ -7,6 +7,7 @@ package downloads
 import (
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


dev-tools/mage/downloads/utils.go
```diff
@@ -79,12 +56,16 @@ func downloadFile(downloadRequest *downloadRequest) error {
 		return err
 	}
 
-	_, err = io.Copy(tempFile, fileReader)
+	_, err = io.Copy(targetFile, fileReader)
+	if err != nil {
+		return fmt.Errorf("writing file %s: %w", targetFile.Name(), err)
+	}
+	err = fileReader.Close()
```
pkoutsovasilis: let's close the fileReader (under the hood resp.Body) on every return path after this point, I don't have an issue not propagating the err from the Close
```suggestion
	defer fileReader.Close()
	_, err = io.Copy(targetFile, fileReader)
	if err != nil {
		return fmt.Errorf("writing file %s: %w", targetFile.Name(), err)
	}
```

pkoutsovasilis (commented):
Overall looks good to me only one comment about closing the resp.body

dev-tools/mage/downloads/utils.go
michalpristas: we should also drain the body in case of copy error to prevent the leak in case it was not fully read

swiatekm: I don't think not draining the response body causes a leak. From what I understand, it can cause problems with HTTP connection reuse, which may or may not affect us here. But it's just one line, so I can easily add it in the defer Panos is asking for.

swiatekm: I decided to move the copy to the inner download function instead: https://github.com/elastic/elastic-agent/pull/9218/commits/9bf418f943f78555cf5f995ff2a2a3bec0dcb1a3. It's possible to get an error reading from the response body, and in that case we should retry, so this works correctly. An unfortunate side effect is that we also retry on write errors to disk, but that doesn't really have any meaningful side effects. And distinguishing between these with Go's syscall error types is annoying, so I'd rather leave the code simpler.

michalpristas (approved):

pkoutsovasilis (approved):

Issue #3342, opened 2023-09-01
Windows uninstall can fail with `Error: failed to remove installation directory` or `Access is denied`


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.10.0 BC3
BUILD: 66261
COMMIT: 56348fa0ed0719679e24d6c58dc3dbee03928c4e
Artifact Link: https://staging.elastic.co/8.10.0-091ff659/downloads/beats/elastic-agent/elastic-agent-8.10.0-windows-x86_64.zip
```

**Host OS and Browser version:** Windows, All

**Preconditions:**
1. 8.10.0 BC3 Kibana cloud environment should be available.
2. Windows agent should be installed.

**Steps to reproduce:**
1. Navigate to CLI.
2. Under Elastic\Agent run `.\elastic-agent.exe uninstall`.
3. `Error: failed to remove installation directory` on running uninstall command.
4. Observe a log file gets retained in the installation directory.

**Expected Result:**
No errors should be observed on running agent uninstall command.

**Screen Recording:**
![32](https://github.com/elastic/elastic-agent/assets/77374876/013f9ac7-7423-42a9-b0f6-84c70f7b98aa)
![33](https://github.com/elastic/elastic-agent/assets/77374876/41822217-547c-463c-b06c-378f4f1435cf)

amolnater-qasource:
@manishgupta-qasource Please review.

amolnater-qasource:
Uninstall command issue for BC2 reported at https://github.com/elastic/fleet-server/issues/2935

manishgupta-qasource:
Secondary review for this ticket is Done

jlind23:
@pierrehilbert @cmacknz worth looking at this.

pierrehilbert:
@AndersonQ you worked on two PR related to this problem in the past days. Can you see where the problem could come from?

harshitgupta-qasource:
Hi Team

While testing on 8.9.2 BC1 build, we observed this issue reproducible there too.

**Observations:**

- Error `failed to remove installation directory` is observed on running agent uninstall command.

**Build details:**
VERSION: 8.9.2 BC1
BUILD: 64883
COMMIT: 21f3ebd6e951d102f41b0299c35e030c3c9e8eb6
Artifacts: https://staging.elastic.co/8.9.2-e4235bb7/summary-8.9.2.html

**Screenshot**
[...]

PR #9224, opened 2025-08-01
Retry saving Agent information file (`fleet.enc`) on Windows

### LINKED ISSUE #5862: [Windows] Access is denied error when rotating fleet.enc file
Filing a bug from an internal report of this error.

`C:\Program Files\Elastic\Agent\fleet.enc: rename C:\ProgramFiles\Elastic\Agent\fleet.enc.tmp C:\ProgramFiles\Elastic\Agent\fleet.enc: Access is denied`

https://github.com/elastic/elastic-agent-libs/blob/4babafd5ed1e5079acf74212ed3da01740b22de7/file/helper_windows.go#L26-L49

The SafeFileRotate implementation on Windows makes a call to `os.Remove` that can fail with an Access is Denied error if a process still has the file open when the removal is attempted. It is unclear what process has the file open, potentially there is AV software performing a periodic scan of the file system.

We could add retries like we did for the uninstall command that used to hit this error more frequently: https://github.com/elastic/elastic-agent/blob/30688d4eb86315906ccebb21ff10c945e813ffca/internal/pkg/agent/install/uninstall.go#L246-L253

We could also switch to using Win32 calls directly.  Then we could use the [MoveFileExA](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexa) directly with the MoveOnReboot flag.  We would have to check what happens if you have multiple moves scheduled on reboot.

ycombinator:
## What does this PR do?

This PR retries saving the Agent information file, `fleet.enc`, on Windows.  The retries are done every 50ms for up to 2 seconds.

## Why is it important?

Saving the Agent information file involves renaming/moving a file to its final destination. However, on Windows, it is sometimes not possible to rename/move a file to its destination file because the destination file is locked by another process (e.g. antivirus software). For such situations, we now retry the save operation on Windows.

## Checklist

## Disruptive User Impact

None.

## Related issues

- Closes #5862

### DIFF
--- a/internal/pkg/agent/application/actions/handlers/handler_helpers_other.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_helpers_other.go
@@ -1,0 +1,18 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+//go:build !windows
+
+package handlers
+
+import (
+	"io"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
+)
+
+// saveConfigToStore saves the given configuration (reader) to the given store
+func saveConfigToStore(store storage.Store, reader io.Reader) error {
+	return store.Save(reader)
+}
--- a/internal/pkg/agent/application/actions/handlers/handler_helpers_other.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_helpers_other.go
@@ -9,10 +9,12 @@ package handlers
 import (
 	"io"
 
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+
 	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
 )
 
 // saveConfigToStore saves the given configuration (reader) to the given store
-func saveConfigToStore(store storage.Store, reader io.Reader) error {
+func saveConfigToStore(store storage.Store, reader io.Reader, _ *logger.Logger) error {
 	return store.Save(reader)
 }
--- a/internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
@@ -13,7 +13,7 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/elastic/elastic-agent-libs/file"
+	"github.com/cenkalti/backoff/v4"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
--- a/internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
@@ -7,12 +7,16 @@
 package handlers
 
 import (
+	"context"
+	"errors"
 	"io"
+	"syscall"
 	"time"
 
 	"github.com/elastic/elastic-agent-libs/file"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
 const saveRetryInterval = 50 * time.Millisecond
@@ -21,6 +25,32 @@ const saveRetryDuration = 2 * time.Second
 // saveConfigToStore saves the given configuration (reader) to the given store.
 // On Windows platforms, the save operation is retried if the error is an
 // ACCESS_DENIED error, which can happen if the file is locked by another process.
-func saveConfigToStore(store storage.Store, reader io.Reader) error {
-	return store.Save(reader, file.WithRenameRetries(saveRetryDuration, saveRetryInterval))
+func saveConfigToStore(store storage.Store, reader io.ReadSeeker, log *logger.Logger) error {
+	ctx, cancel := context.WithTimeout(context.Background(), saveRetryDuration)
+	defer cancel()
+
+	bo := backoff.WithContext(backoff.NewConstantBackOff(saveRetryInterval), ctx)
+
+	return backoff.Retry(func() error {
+		err := store.Save(reader)
+		if err == nil {
+			// Save succeeded
+			return nil
+		}
+
+		if !errors.Is(err, syscall.ERROR_ACCESS_DENIED) {
+			// Save failed due to an error that is not ACCESS_DENIED. Immediately
+			// indicate failure without retrying further.
+			log.Debugf("Saving configuration to store failed: %v. Not retrying.", err)
+			return backoff.Permanent(err)
+		}
+
+		if _, seekErr := reader.Seek(0, io.SeekStart); seekErr != nil {
+			log.Debugf("Saving configuration to store failed: %v. Failed to reset reader: %v. Not retrying.", err, seekErr)
+			return backoff.Permanent(errors.Join(err, seekErr))
+		}
+
+		log.Debugf("Saving configuration to store failed: %v. Retrying...", err)
+		return err
+	}, bo)
 }
--- a/internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
@@ -1,0 +1,26 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+//go:build windows
+
+package handlers
+
+import (
+	"io"
+	"time"
+
+	"github.com/elastic/elastic-agent-libs/file"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
+)
+
+const saveRetryInterval = 50 * time.Millisecond
+const saveRetryDuration = 2 * time.Second
+
+// saveConfigToStore saves the given configuration (reader) to the given store.
+// On Windows platforms, the save operation is retried if the error is an
+// ACCESS_DENIED error, which can happen if the file is locked by another process.
+func saveConfigToStore(store storage.Store, reader io.Reader) error {
+	return store.Save(reader, file.WithRenameRetries(saveRetryDuration, saveRetryInterval))
+}
--- a/changelog/fragments/1754346816-fleet-enc-save-retries.yaml
+++ b/changelog/fragments/1754346816-fleet-enc-save-retries.yaml
@@ -1,0 +1,36 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: On Windows, retry saving the Agent information file to disk
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: >
+  Saving the Agent information file involves renaming/moving a file to its final
+  destination. However, on Windows, it is sometimes not possible to rename/move a
+  file to its destination file because the destination file is locked by another
+  process (e.g. antivirus software). For such situations, we now retry the save operation
+  on Windows.
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: elastic-agent
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/elastic-agent/pull/9224
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: https://github.com/elastic/elastic-agent/issues/5862
--- a/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
@@ -391,8 +391,7 @@ func saveConfig(agentInfo info.Agent, validatedConfig *configuration.Configurati
 			errors.TypeUnexpected, errors.M("hosts", validatedConfig.Fleet.Client.Hosts))
 	}
 
-	err = store.Save(reader)
-	if err != nil {
+	if err := saveConfigToStore(store, reader); err != nil {
 		return errors.New(
 			err, "fail to persist new Fleet Server API client hosts",
 			errors.TypeFilesystem, errors.M("hosts", validatedConfig.Fleet.Client.Hosts))
--- a/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
@@ -291,7 +291,7 @@ func (h *PolicyChangeHandler) handlePolicyChange(ctx context.Context, c *config.
 	}
 
 	// persist configuration
-	err = saveConfig(h.agentInfo, h.config, h.store)
+	err = saveConfig(h.agentInfo, h.config, h.store, h.log)
 	if err != nil {
 		return fmt.Errorf("saving config: %w", err)
 	}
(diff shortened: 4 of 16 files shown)


ycombinator:
@pchila @swiatekm Following the feedback on https://github.com/elastic/elastic-agent-libs/pull/333, I'm replacing that PR with this one.  At the moment, I've sketched out two alternate implementations in this PR — one that specifically retries saving the `fleet.enc` file when the policy is being handled from Fleet and the other that's a bit more generic / at a lower-level where it retries saving any file using the `storage` package.  In both implementations, the retries are only done if the OS is Windows and error from the save/rotate operation is `ACCESS_DENIED`. 

Let me know which of these approaches you prefer and I'll get rid of the other one + add tests.  Or if you prefer a totally different approach to solving this [bug](https://github.com/elastic/elastic-agent/issues/5862), that's fine too — just let me know.  Thanks!

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -0,0 +1,46 @@
...
+			// Save failed due to an error that is not ACCESS_DENIED. Immediately
+			// indicate failure without retrying further.
+			return false
+		}
+
+		// Keep retrying...
+	}
+
+	// We exhausted retries without succeeding, so indicate failure.
+	return false
```
pchila: What about using a dependency we already have like https://github.com/cenkalti/backoff/ to implement a Retry() with a constant backoff ?
Something similar to this example https://github.com/cenkalti/backoff/blob/v5/example_test.go

This would also avoid some pitfalls like using [time.Sleep](https://pkg.go.dev/time#Sleep) that only gives guarantees about the **minimum** sleep duration (probably not ideal with a tight 50 ms interval between attempts.

ycombinator: Done in 6a0f3e91266fb5087279d462476859c233281290.

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -0,0 +1,53 @@
...
+// ACCESS_DENIED error, which can happen if the file is locked by another process.
+func saveConfigToStore(store storage.Store, reader io.Reader) error {
+	var saveErr error
+	retryableSaveFn := func() error {
+		saveErr = store.Save(reader)
+		if errors.Is(saveErr, syscall.ERROR_ACCESS_DENIED) {
+			// Retryable error, so return it
+			return saveErr
+		}
+
```
pchila: Any error that is not  retryable should be returned with `backoff.Permanent` to signal that we need to break the loop.
```suggestion
		if errors.Is(saveErr, syscall.ERROR_ACCESS_DENIED) {
			// Retryable error, so return it
			return saveErr
		}
		if err != nil {
			// Non-retryable error, so mark it as permanent
			return backoff.Permanent(saveErr)
		}
```

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -0,0 +1,53 @@
...
+		// saveErr is an error that should not be retried. Return nil to
+		// signal to the retrier that it should not retry.
```
pchila: Swallowing the error (not returning it and not logging it) does not seem like a good idea

ycombinator:
> I think we can start with retrying only at the handler_policy_change level (without the additional retry loop at EncryptedStore

Can you clarify what you mean by this? In other words, what would you like to see done differently than the current implementation in this PR?

pchila:
> > I think we can start with retrying only at the handler_policy_change level (without the additional retry loop at EncryptedStore
> 
> Can you clarify what you mean by this? In other words, what would you like to see done differently than the current implementation in this PR?

I would do without the encrypted store retry (I am referring to [this part](https://github.com/elastic/elastic-agent/pull/9224/files#diff-590f1282056bf97bebc69f203019c99730d59369d0a7fc192da8223d658aa07c) that duplicate the original implementation of the handler retry).

The bug we are trying to fix is relative to not being able to write a new `fleet.enc` on windows because of `Access is denied` error so I would keep the fix focused on that operation (unless there is concrete evidence that the bug originates in the *diskstore implementation).

ycombinator:
> I would do without the encrypted store retry (I am referring to [this part](https://github.com/elastic/elastic-agent/pull/9224/files#diff-590f1282056bf97bebc69f203019c99730d59369d0a7fc192da8223d658aa07c) that duplicate the original implementation of the handler retry).

> The bug we are trying to fix is relative to not being able to write a new fleet.enc on windows because of Access is denied error so I would keep the fix focused on that operation (unless there is concrete evidence that the bug originates in the *diskstore implementation).

@pchila My bad, I just forgot to remove the previous implementation (the one you linked to) which is why I was unsure as to what I was missing 🤦.  Thanks for the clarification.  Removed the previous implementation in 866d457bca37f0be15760f837db2c945d19b2f49.

ycombinator:
The `TestSaveConfigToStore` unit test introduced in this PR is failing.

```
PS C:\Users\Shaunak Kashyap\development\github\elastic-agent> go test -v ./internal/pkg/agent/application/actions/handlers/... -run TestSaveConfigToStore -count 10
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.55s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.55s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.51s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.51s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.51s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.55s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.56s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.52s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.55s)
=== RUN   TestSaveConfigToStore
--- PASS: TestSaveConfigToStore (1.51s)
PASS
ok      github.com/elastic/elastic-agent/internal/pkg/agent/application/actions/handlers        15.461s
PS C:\Users\Shaunak Kashyap\development\github\elastic-agent>
```

Moving PR to draft while I investigate this issue.

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -0,0 +1,57 @@
...
+func saveConfigToStore(store storage.Store, reader io.ReadSeeker) error {
+	retryableSaveFn := func() error {
+		err := store.Save(reader)
+		if errors.Is(err, syscall.ERROR_ACCESS_DENIED) {
+			// Retryable error, so reset reader position to start and return the error
+			if _, seekErr := reader.Seek(0, io.SeekStart); seekErr != nil {
+				// Could not reset reader position; we can no longer retry
+				return backoff.Permanent(errors.Join(err, seekErr))
+			}
+
```
pchila: Nitpick: could we add a debug log here to track the retryable errors encountered and the retry attempts ?
Ideally it would also be useful to log which processes have the file open but in Windows it doesn't seem a straightforward task.
It seems like we should use [NtQueryInformationFile](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntqueryinformationfile?redirectedfrom=MSDN) with a [FileProcessIdsUsingFileInformation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ne-wdm-_file_information_class) although it seems to be available only from Windows Vista onwards and system reserved :/

internal/pkg/agent/application/actions/handlers/handler_helpers_other.go
```diff
@@ -0,0 +1,18 @@
...
+package handlers
+
+import (
+	"io"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/storage"
+)
+
+// saveConfigToStore saves the given configuration (reader) to the given store
+func saveConfigToStore(store storage.Store, reader io.Reader) error {
```
pchila: Since we have no retries on non-windows systems the function can accept a simple `io.Reader` instead of an `io.ReadSeeker` required for the windows version ? Maybe we should keep the signature consistent ?

pchila (dismissed):
Just a couple of nitpicks, nothing blocking. Please have a look.

internal/pkg/agent/application/actions/handlers/handler_helpers_other.go
ycombinator: Ha, I was debating this one too.  I can justify it either way.  Sure, let's go with consistency!

ycombinator: Done in e7302426d9.

ycombinator: Done in e7302426d9.

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
ycombinator: Added debug logging in 6d70a49e64.

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -24,21 +25,30 @@ const saveRetryDuration = 2 * time.Second
...
-func saveConfigToStore(store storage.Store, reader io.ReadSeeker) error {
+func saveConfigToStore(store storage.Store, reader io.ReadSeeker, log *logger.Logger) error {
+	log = log.Named("saveConfigToStore_windows")
+	numSaveAttempts := 0
+
 	retryableSaveFn := func() error {
 		err := store.Save(reader)
+		numSaveAttempts++
+		log.Debugf("save attempt %d", numSaveAttempts)
+
```
pchila:
```suggestion
		log.Debugf("save config attempt failed: %q.", err.Error())
```

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -24,21 +25,30 @@ const saveRetryDuration = 2 * time.Second
...
 		if errors.Is(err, syscall.ERROR_ACCESS_DENIED) {
 			// Retryable error, so reset reader position to start and return the error
 			if _, seekErr := reader.Seek(0, io.SeekStart); seekErr != nil {
 				// Could not reset reader position; we can no longer retry
+				log.Debugf("got retryable error: %v, but no longer retrying save due to error resetting reader: %v", err.Error(), seekErr.Error())
 				return backoff.Permanent(errors.Join(err, seekErr))
 			}
 
+			log.Debugf("retrying save due to retryable error: %v", err.Error())
```
pchila:
```suggestion
			log.Debugf("retrying save due to retryable error")
```

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -24,21 +25,30 @@ const saveRetryDuration = 2 * time.Second
...
 			return err
 		}
 
 		if err != nil {
 			// Non-retryable error, so mark it as permanent
+			log.Debugf("no longer retrying save due to non-retryable error: %v", err.Error())
```
pchila:
```suggestion
			log.Debugf("no longer retrying save due to non-retryable error")
```

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
ycombinator: Wouldn't we want to know what the retryable error is?

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
ycombinator: Never mind, I see you are logging the error in the common log line above.  This is what I get for going through review comments in reverse 🙂.

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
ycombinator: At this point we don't know if the attempt failed or not.  Also, `err` might be `nil` at this point so `err.Error()` is not safe.  That's why I was logging the errors inside the error-checking blocks below.

pchila: the logs that are below are missing the error if another error occurs (like resetting the buffer).
If we want to keep them within the "if" statements we should probably log them at the very beginning of the blocks.
Having an attempt counter is not extremely useful, we can count the occurrences directly in the logs.

ycombinator: Take a look at ccade717de and let me know what you think.

internal/pkg/agent/application/actions/handlers/handler_helpers_windows.go
```diff
@@ -0,0 +1,63 @@
...
+const saveRetryDuration = 2 * time.Second
+
+// saveConfigToStore saves the given configuration (reader) to the given store.
+// On Windows platforms, the save operation is retried if the error is an
+// ACCESS_DENIED error, which can happen if the file is locked by another process.
+func saveConfigToStore(store storage.Store, reader io.ReadSeeker, log *logger.Logger) error {
+	log = log.Named("saveConfigToStore_windows")
+
+	retryableSaveFn := func() error {
+		err := store.Save(reader)
```
blakerouse, line 35: What is the actual cause of the error? To me it would be the rename that is in error. It was able to write the temporary file, but then fails to rename it because something is now accessing it.

What is the error case that it fails on the actual write of the temporary file? That seems like an error there would be a permissions issue.

The worry I have with placing this change at this level is that we are setting ourselves up for another failure right behind the other one. The only reason this is working is because we just happen to be fast enough the second time.

I feel like it would be better to place the retry logic at the rename action level. That way the code can ensure that the rename can happen. If we place it here it will write another temporary file and it might be accessed by something else on the system.

What also occurs with the temporary file that could not be renamed on retry? It couldn't rename it, I would expect that it would not be able to delete it. Does that mean its just littering the filesystem with temporary files?

ycombinator: > What is the actual cause of the error? To me it would be the rename that is in error. It was able to write the temporary file, but then fails to rename it because something is now accessing it.

That's correct.  The error happens when trying to rename the temporary file to the final one. This is an example of the error:

```
C:\Program Files\Elastic\Agent\fleet.enc: rename C:\ProgramFiles\Elastic\Agent\fleet.enc.tmp C:\ProgramFiles\Elastic\Agent\fleet.enc: Access is denied
```

> I feel like it would be better to place the retry logic at the rename action level. That way the code can ensure that the rename can happen. If we place it here it will write another temporary file and it might be accessed by something else on the system.

Could you clarify where you mean by the "rename action level"? Did you mean something like https://github.com/elastic/elastic-agent-libs/pull/333 or do you mean something else?

blakerouse: The implementation in https://github.com/elastic/elastic-agent-libs/pull/333 is more correct to me. We do not want this all over the code-base. Any time we do a SafeFileRotate this will be an issue on Windows and the developer of the SafeFileRotate should not need to worry about this issue on Windows. I believe there are a few places that SafeFileRotate is used and we should not need this retry logic all over the place. The implementation here also has all the issues that I stated above:

- Littering the filesystem with temporary files (won't be able to remove if cannot rename)
- Creating more files that required for the retry
- The fact that it could just fail again due to timing as the retry is to high in the call stack

The comment about taking longer to write a file on Windows because of this should not be an issue. On any OS the developer never really knows how long a file write will take. It could be on a spinning disk, an NFS, or an some fuse based mount that is very slow. I don't think blocking for the retry is wrong in the original implementation that @ycombinator proposed.

I am open to having the ability to pass options to the `SafeFileRotate` as `...opts` that doesn't change the signature of the function but adds the ability to turn off the retry logic. I don't think we would want to ever turn that off in the Elastic Agent use-case, so that seems like adding something before it is really even needed.

cmacknz: Yeah I agree, a big part of the problem is developers not knowing they need retries for these operations in the first place, so it seems better to me to encapsulate them where they can't be forgotten. We have some history with this specific `Access is Denied` error (see https://github.com/elastic/elastic-agent/issues/3342 for example though in that case the solution was able to avoid the potential for conflict entirely).

I also like the suggestion to accept `...opts` as a way to turn them off in case someone wants a different retry strategy, in that case the developer knows what they are doing.

This seems aligned with the spirit of https://github.com/elastic/elastic-agent-libs/pull/333#issuecomment-3135824288 in the original PR as well.

ycombinator: There seems to be enough consensus now in favor of the implementation in https://github.com/elastic/elastic-agent-libs/pull/333 with variadic options.  So I'm bringing back that PR in favor of this one here.

ycombinator:
With the revised implementation of https://github.com/elastic/elastic-agent-libs/pull/333, callers of `SafeFileRotate` get retries on Windows and no retries on other OSes by default.  However, callers can pass an optional third argument to `SafeFileRotate` to configure the retry behavior.  With this implementation, this PR here is no longer needed so I'm closing it unmerged now.

ycombinator:
While we're going to keep the changes in https://github.com/elastic/elastic-agent-libs/pull/333, we are having some trouble bringing them into Elastic Agent (see https://github.com/elastic/elastic-agent/pull/9499#issuecomment-3207496169).  And we're pressed for time as we're trying to get the bug behind all this work to be fixed in a release that's coming up very shortly.  

So, as a temporary measure, we are going to revive this PR here and merge it.  

Once we get the changes from https://github.com/elastic/elastic-agent-libs/pull/333 into Elastic Agent (by eventually merging https://github.com/elastic/elastic-agent/pull/9499), we'll revert this PR here.

ycombinator:
Not sure why/how this got closed again.  Reopening...

cmacknz (approved):

PR #9213, opened 2025-07-31
adjust endpoint-security check timeout

intxgo:
## What does this PR do?

Adjust Elastic Endpoint check command timeout to accommodate for planned implementation changes on endpoint side.

## Why is it important?

When timeout is reached Agent will proceed to reinstall Endpoint which might not be necessary otherwise.

## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #9227, opened 2025-08-01
Increase endpoint verify command timeout to 60s.

cmacknz:
Increase the timeout of the endpoint-security verify command to 60s.

This implements the `check` function used to determine if we need to call `endpoint-security install`. 

https://github.com/elastic/elastic-agent/blob/1aa5e6d5494ae72381e579eaa1f50b4c7dede39a/pkg/component/runtime/service.go#L353-L364

The `verify` command is responsible for upgrades, and is gaining some self-healing capabilities that could take additional time like (re)starting the endpoint OS service. Increasing to 60s as suggested by @bjmcnic


cmacknz:
Backporting to all active branches to match the corresponding endpoint PR

cmacknz:
Apparently I was beaten to this: https://github.com/elastic/elastic-agent/pull/9213

Closing.

PR #9257, opened 2025-08-06
[beatreceivers] Integrate beatsauthextension

khushijain21:
- Relates https://github.com/elastic/opentelemetry-collector-components/pull/722

## What does this PR do?
This PR adds `beatsauthextension`  to elasticsearch config translation layer and also propagates local logger to the translation logic

## Why is it important?
It replaces es-exporter's `Rountripper` with beat's implementation of it. This is important for 1:1 behavioral compatibility between beats and beatreceivers 

## Checklist

## Disruptive User Impact

None

Check E2E test


internal/pkg/otel/translate/otelconfig.go
```diff
@@ -154,6 +164,25 @@ func getCollectorConfigForComponent(
...
+		// this is required because otel's merging logic expects []interface
+		serviceInterface := make([]interface{}, len(maps.Keys(extensionConfig)))
+		for i, v := range maps.Keys(extensionConfig) {
+			serviceInterface[i] = v
+		}
+
+		serviceMap["extensions"] = serviceInterface
+	}
+
+	fmt.Println(confmap.NewFromStringMap(fullConfig))
```
leehinman: Do we want to print out the config?

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -429,3 +487,34 @@ func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
...
+// ca_sha256
+func getBeatsAuthExtensionConfig(cfg *config.C) (map[string]any, error) {
+	var sslConfig httpcommon.HTTPTransportSettings
+	err := cfg.Unpack(&sslConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// if ssl is not enabled or if verification mode is none, then we do not need beatsauth extension
+	if !sslConfig.TLS.IsEnabled() || sslConfig.TLS.VerificationMode.String() == "none" {
```
cmacknz: Is there a reason to avoid include the auth extension all the time? That will be much easier to reason about, rather than having it conditional on the inclusion of the configuration.

Why can't we just inject beats auth extension all the time and then pass it the entire ssl block so it can figure out what to do?

We are going to need to handle the ssl parameters the same way for Logstash and Kafka once we support those as outputs as well.

internal/pkg/otel/translate/otelconfig.go
khushijain21: no, that was a left over. Removed it

internal/pkg/otel/translate/otelconfig.go
khushijain21: we would need this condition atleast - `if !sslConfig.TLS.IsEnabled()` 

This is because if we enable auth extension on ES exporter at all times and if a user has not enabled ssl explicitly under outputs config in elastic-agent - then instead of going through the default TLS (implemented by GO), all HTTP requests will be through our custom verification code.

khushijain21: we can add it all times - that shouldn't be a problem. 

> We are going to need to handle the ssl parameters the same way for Logstash and Kafka once we support those as outputs as wel

We would have to test this extension with other outputs before that. Added a comment in the code as well

testing/integration/ess/otel_test.go
```diff
@@ -1059,6 +1060,8 @@ outputs:
     hosts: [{{.ESEndpoint}}]
     api_key: "{{.ESApiKey}}"
     preset: "balanced"
+    ssl.enabled: true
+    ssl.verification_mode: full
```
cmacknz, line 1066: What was the failure mode for these before? Would they just be ignored?

I am just double checking that this test actually tests the auth extension, we want a test that fails if it isn't used correctly. 

If these parameters were able to be ignored previously, and the test doesn't only work with SSL configuration enabled by the extension, then this test doesn't test the extension as it would path without it. 

That the Elasticsearch configuration hasn't changed as well here makes me think this is the case.

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -119,15 +121,23 @@ func getExporterID(exporterType otelcomponent.Type, outputName string) otelcompo
 	return otelcomponent.NewIDWithName(exporterType, exporterName)
 }
 
+// getExtensionID returns the id for beatsauth extension
+// outputName here is name of the output defined in elastic-agent.yml. For ex: default, monitoring
+func getExtensionID(outputName string) otelcomponent.ID {
```
cmacknz:
```suggestion
func getBeatsAuthExtensionID(outputName string) otelcomponent.ID {
```

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -429,3 +474,42 @@ func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
 func BeatDataPath(componentId string) string {
 	return filepath.Join(paths.Run(), componentId)
 }
+
+// getBeatsAuthExtensionConfig sets following ssl parameters on beatsauth
+// verification_mode
+// ca_trusted_fingerprint
+// ca_sha256
+func getBeatsAuthExtensionConfig(cfg *config.C) (map[string]any, error) {
```
cmacknz, line 492: Why do we need this? Why can't we just pass the `ssl` configuration into the auth extension? If there is no `ssl` section the default configuration of the extension should match what we want.

It looks like the auth extension wants `tls` instead of `ssl`? IMO the auth extension exists purely for compatibility with Beats and should accept the same config format and be able to deal with the parameters being named `ssl`.

In the case where a user migrates from standalone beats to beats receivers, they will either need to do this translation manually or we can make the extension accept the beat ssl config exactly and have it do the right thing. This is a much nicer experience.

cmacknz: I think what we might have done is manually configure the `tls` options in the exporter except when they aren't supported, they go in the auth extension.

I see how we got there, but it also opens us to potential conflicts between settings that won't be resolved properly because neither the auth extension or the exporter configuration knows about all of them.

cmacknz: Right now our instructions for somebody manually trying to write a collector configuration from their existing agent or beats configuration is "manually translate to the similar tls options except if it's one of these unsupported options then use the auth extension but only for the unsupported options".

It is way easier if we can say "copy paste the ssl block into the auth extension" and then it just works.

cmacknz: If we agree on this, this can be done as a follow up improvement it doesn't need to block this PR.

testing/integration/ess/otel_test.go
cmacknz: It looks like the auth extension itself does a lot of testing against a test HTTP server which is great, https://github.com/elastic/opentelemetry-collector-components/blob/main/extension/beatsauthextension/authenticator_test.go

So there's less need to verify in depth here, but having a single test to prove it gets wired in correctly makes sense and that test shouldn't pass regardless of if the extension is used.

khushijain21: >  am just double checking that this test actually tests the auth extension, we want a test that fails if it isn't used correctly.

Yes `verification_mode` was previously ignored.

>  I am just double checking that this test actually tests the auth extension, we want a test that fails if it isn't used correctly.

Since we are enabling `beatsauthextension` unconditionally for components with otel runtime. Basically all our current tests are now routed to use ssl verification via this authenticator.

internal/pkg/otel/translate/otelconfig.go
khushijain21: so my understanding was we wanted `beatsauthextension` to support ssl parameter currently not supported upstream and this extension is a temporary middleware until we get there

We only have three of such unsupported ssl parameters. 

> In the case where a user migrates from standalone beats to beats receivers, they will either need to do this translation manually or we can make the extension accept the beat ssl config exactly

Our translation logic available in beats should take care of this - any use case where we want users to do this manually?

The only reason to not redirect complete ssl support to `beatsauthextension` is because we would be discarding the upstream tls workings entirely. If our plan is to merge all our beat settings with upstream - this path seems less ideal.

kaanyalti:
Aside from the minor naming nit I don't have much else to add to what's already been pointed out or asked in the comments. The changes look good to me, waiting for ci to pass.

internal/pkg/otel/translate/otelconfig.go
cmacknz: >  If our plan is to merge all our beat settings with upstream - this path seems less ideal.

Our plan right now is to have a collector configuration that behaves _exactly_ the same as Elastic Agent does now. Any difference in TLS setting and behaviour is an immediate data collection outage for whoever is using this so it's really important we get it right. It is a lot easier for me to verify the settings have the same behaviour if we use exactly the TLS logic we do today.

> Our translation logic available in beats should take care of this - any use case where we want users to do this manually?

People are absolutely going to write beats receivers configurations by hand, but I think you are right that if we give people tooling to translate their beats configs they will probably use that from the start instead of writing them from scratch.

cmacknz: The most important thing I want is tests verifying that TLS is behaviourally the same between Beats and our generated exporter configuration after the result of config translation.

It looks to me like the config translation and auth extension interact but we haven't fully tested them thoroughly together:

* We handle verification_mode none in the config translation in https://github.com/elastic/beats/blob/b93bb0c7abc3ceeeca86aa8bde7dce4a08d85849/libbeat/otelbeat/oteltranslate/tls_otel.go#L99-L103
* We handle it again in the auth extension https://github.com/elastic/opentelemetry-collector-components/blob/892a5dfec61c419400c63d7e9b4317745b2f7072/extension/beatsauthextension/authenticator.go#L56

cmacknz: There are some potentially problematic combinations I don't see tests for. For example if using `verify_certificate` in the beats auth extension, but not relaxing the exporter to tls.InsecureSkipVerify, there is the possibility the collector TLS configuration we end up with is more strict than the checks for `verify_certificate`.

cmacknz: The safest way I see to be _exactly_ compatible with what we had before is to have the auth extension completely replace the transport's TLS config with one generated based on the exact same set of `ssl` settings in the output. 

Right now we are replacing the TLS config based only on some of the SSL parameters (the ones the beats auth extension accepts): https://github.com/elastic/opentelemetry-collector-components/blob/892a5dfec61c419400c63d7e9b4317745b2f7072/extension/beatsauthextension/authenticator.go#L55-L64

It's really hard for me to verify that this is exactly the same behaviourally just from looking at it.

To me, it would be safer to start using the exact same TLS config we would have generated in Beats, and then over time shift things to the collector config where we can, than to do the opposite which is what I think we are doing now (trying to use the collector `tls` options and augment them with beats options).

cmacknz: Thinking more about this (and discussing briefly in the most recent beats receivers meeting) we should go farther than just having the auth extension handle the SSL settings, it should handle all of the httpcommon.Transport settings so that the transport we use is as close as possible to what Beats would use.

- https://github.com/elastic/elastic-agent/issues/9633

I think it is much safer for us to start from a place of exact compatibility and gradually port things over to the collector's equivalent support, instead of trying to get the collector to support our entire historical array of SSL and transport settings.

testing/integration/ess/otel_test.go
khushijain21: Added an E2E test case that shows beatsauthextension is working correctly. in [cb2beac](https://github.com/elastic/elastic-agent/pull/9257/commits/cb2beac6981202f3076367c91214f122c6ba2ed5)

go.mod
```diff
@@ -768,6 +769,7 @@ replace (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/consumption/armconsumption => github.com/elastic/azure-sdk-for-go/sdk/resourcemanager/consumption/armconsumption v1.1.0-elastic
 	github.com/apoydence/eachers => github.com/poy/eachers v0.0.0-20181020210610-23942921fe77 //indirect, see https://github.com/elastic/beats/pull/29780 for details.
 	github.com/dop251/goja => github.com/elastic/goja v0.0.0-20190128172624-dd2ac4456e20
+	github.com/elastic/opentelemetry-collector-components/extension/beatsauthextension => github.com/khushijain21/opentelemetry-collector-components/extension/beatsauthextension v0.0.0-20250902040834-316a18b416a5
```
khushijain21: This will be replaced with a real version when this PR is merged https://github.com/elastic/elastic-agent/pull/9257

khushijain21:
Can the reviewers take a look again? This PR has been modified since the last review
@cmacknz @swiatekm

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/elastic/cloud-on-k8s/v2 v2.0.0-20250327073047-b624240832ae
 	github.com/elastic/elastic-agent-autodiscover v0.10.0
 	github.com/elastic/elastic-agent-client/v7 v7.17.2
-	github.com/elastic/elastic-agent-libs v0.23.0
+	github.com/elastic/elastic-agent-libs v0.23.2
```
cmacknz: Can you pull this update and related fixes out of this PR? I want to avoid us having to rush through this one and any related testing because we need the elastic-agent-libs update it contains.

cmacknz: (it'll also make the PR easier to review)

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -429,3 +473,26 @@ func translateEsOutputToExporter(cfg *config.C) (map[string]any, error) {
 func BeatDataPath(componentId string) string {
 	return filepath.Join(paths.Run(), componentId)
 }
+
+// getBeatsAuthExtensionConfig sets following http parameters on beatsauth
+// currently this is only supported for elasticsearch output
+func getBeatsAuthExtensionConfig(cfg *config.C) (map[string]any, error) {
+	sslConfig := elasticsearch.ESDefaultTransportSettings()
```
cmacknz: This doesn't seem right to me, what I would expect is we can take the ssl*, proxy*, and the two timeout settings from the output and just pass them to the auth extension. This shouldn't require us to set any defaults ahead of time (unless I'm missing that this is what the Beat output does internally).

khushijain21: The ES output sets these defaults internally https://github.com/khushijain21/beats/blob/removetls/libbeat/outputs/elasticsearch/config.go#L86-L92

go.mod
khushijain21: This version has been updated in another PR

internal/pkg/otel/translate/otelconfig.go
cmacknz:
```suggestion
	defaultTransportSettings := elasticsearch.ESDefaultTransportSettings()
```

internal/pkg/otel/translate/otelconfig_test.go
```diff
@@ -246,6 +271,13 @@ func TestGetOtelConfig(t *testing.T) {
 			},
 			"timeout":           90 * time.Second,
 			"idle_conn_timeout": 3 * time.Second,
+			"auth": map[string]any{
+				"authenticator": "beatsauth/_agent-component/default",
+			},
+			"tls": map[string]any{
```
cmacknz, line 310: Why do we still have a `tls` block here? This would conflict with https://www.elastic.co/docs/reference/beats/filebeat/configuration-ssl#supported-protocols if set in the ssl.* settings.

internal/pkg/otel/translate/otelconfig_test.go
```diff
@@ -246,6 +271,13 @@ func TestGetOtelConfig(t *testing.T) {
 			},
 			"timeout":           90 * time.Second,
 			"idle_conn_timeout": 3 * time.Second,
+			"auth": map[string]any{
```
cmacknz, line 307: The timeout and idle_conn_timeout above should be in the auth extension shouldn't they?

cmacknz:
Can we have a test that can catch https://github.com/elastic/elastic-agent/issues/9803 to prove it's fixed?

I think in general my main concern is making sure we test this thoroughly, because anywhere this doesn't work the way it should will likely break connectivity to Elasticsearch.

https://github.com/elastic/beats/issues/45491 has the main list of cases to consider. For some of the most important ones we already have tests we can use for inspiration, like https://github.com/elastic/elastic-agent/blob/615190be3f6a28403883f6b4983764558173e0e7/testing/proxytest/proxytest_test.go#L25

cmacknz:
For automated testing, we need to make sure that each top level configuration in HTTPTransportSettings applies to make sure we are capturing them correctly at least, and automate test important cases (self signed certs, ca fingerprint, mtls, proxies) in one big test or separately.

We can generally take that the actual features are implemented by elastic-agent-libs and we have to prove that we are able to use them all and there are no unexpected conflicts.

The connection timeout and idle connection timeout are two that probably need dedicated tests where we hopefully don't need automated testing for every combination of SSL and proxy settings.

khushijain21:
> The connection timeout and idle connection timeout are two that probably need dedicated tests where we hopefully don't need automated testing for every combination of SSL and proxy settings.

All our current otel tests already use these two parameters (since these are set by default).

> and automate test important cases (self signed certs, ca fingerprint, mtls, proxies) in one big test or separately.
Should this be done in a separate PR (tracked by this issue https://github.com/elastic/beats/issues/45491)

> For automated testing, we need to make sure that each top level configuration in HTTPTransportSettings
Each top level configuration including the ssl block is explicitly set on `TestOTelFilestreamInput` in this PR - so I think that should be good. I'll add a comment there too

khushijain21:
> Can we have a test that can catch https://github.com/elastic/elastic-agent/issues/9803 to prove it's fixed?

This wasn't exactly a bug. It was known that `ca_trusted_fingerprint` is not supported by ES exporter yet. An upcoming test for self signed certs, ca_trusted_fingerprint etc should prove this works

internal/pkg/otel/translate/otelconfig_test.go
khushijain21: Should be cleaned by this PR in beats 
https://github.com/elastic/beats/pull/46428

internal/pkg/otel/translate/otelconfig_test.go
khushijain21: Should be cleaned by this PR in beats 
https://github.com/elastic/beats/pull/46428

khushijain21:
>  and automate test important cases (self signed certs, ca fingerprint, mtls, proxies) in one big test or separately.

Working on these tests in beats repo in this PR https://github.com/elastic/beats/pull/46428

khushijain21:
> The connection timeout and idle connection timeout are two that probably need dedicated tests where we hopefully don't need automated testing for every combination of SSL and proxy settings.

so every otel test currently is setting timeout and idle connection timeout on beatsauth. So I think our current tests should do

cmacknz:
> so every otel test currently is setting timeout and idle connection timeout on beatsauth. So I think our current tests should do

Are they verifying the timeouts have the correct duration? I don't think using the parameters in a configuration is enough, we have to verify the behavior of the transport is what we told it to do. We need to verify the behavior not that the configuration looks like we think it should. This is how we missed correctly limiting the number of connections leading to https://github.com/elastic/beats/pull/46111.

khushijain21:
All explicit tests will be done in a follow-up PR

cmacknz:
One test I do think should be added here is a test that verifies we get an auth extension per exporter when there are multiple exporters with the correct IDs. Ideally with a new test or an existing test updated that shows the auth extensions are correctly independent with conflicting configurations.

I think the code for this to work is here but I don't see a test for this quickly skimming through again.

khushijain21:
>One test I do think should be added here is a test that verifies we get an auth extension per exporter when there are multiple exporters with the correct IDs. Ideally with a new test or an existing test updated that shows the auth extensions are correctly independent with conflicting configurations.

Added in https://github.com/elastic/elastic-agent/pull/9257/commits/b7cb3957e1dd16ebf6b7a2dce8e6b768b014f384

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -147,13 +170,22 @@ func getCollectorConfigForComponent(
 		},
 	}
```
swiatekm, line 171: What happens if the user sets their own extensions in hybrid mode configuration? We currently merge these naively, so we'd override either ours or theirs. We need to account for this in the otel manager. This isn't really an issue with this PR, but adding extensions here will actually make triggerring it possible.

internal/pkg/otel/translate/otelconfig_test.go
```diff
@@ -366,76 +463,105 @@ func TestGetOtelConfig(t *testing.T) {
...
 				},
+			},
+			expectedConfig: confmap.NewFromStringMap(map[string]any{
+				"exporters": map[string]any{
+					"elasticsearch/_agent-component/primaryOutput":   expectedESConfig("primaryOutput"),
+					"elasticsearch/_agent-component/secondaryOutput": expectedESConfig("secondaryOutput"),
+				},
+				"extensions": map[string]any{
+					"beatsauth/_agent-component/primaryOutput":   expectedExtensionConfig,
+					"beatsauth/_agent-component/secondaryOutput": expectedExtensionConfig,
```
cmacknz: This test would be stronger if the output configurations and therefore generated extension configurations were different.

cmacknz:
A couple of observations testing this manually:

1. The extension is present in the status output unconditionally in the collector format. This should get fixed in the yet to be created issue for tying auth extension status to output unit status. We can address this separately as a sub-issue in https://github.com/elastic/elastic-agent/issues/9771.

```
┌─ fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   └─ extensions
      ├─ status: StatusOK
      └─ extension:beatsauth/_agent-component/monitoring
         └─ status: StatusOK
```

2. If I use the configuration below, looking in diagnostics then the auth extension configuration is missing the `ssl` blocks which surprised me.

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    api_key: "example-key"
    preset: balanced
    ssl.key: /tmp/key.pem
    ssl.certificate: /tmp/cert.pem

inputs:
  - type: system/metrics
    data_stream.namespace: default
    use_output: default
    streams:
      - metricsets:
        - cpu
        data_stream.dataset: system.cpu

agent.monitoring:
  enabled: true
  logs: true
  metrics: true
  _runtime_experimental: otel
```

In otel-merged.yml from generated from that configuration I see the following with the ssl parameters missing:

```yaml
exporters:
    elasticsearch/_agent-component/monitoring:
        api_key: <REDACTED>
        auth:
            authenticator: <REDACTED>
        batcher:
            enabled: true
            max_size: 1600
            min_size: 0
        compression: gzip
        compression_params:
            level: 1
        endpoints:
            - http://127.0.0.1:9200
        idle_conn_timeout: 3s
        logs_dynamic_id:
            enabled: true
        mapping:
            mode: bodymap
        retry:
            enabled: true
            initial_interval: 1s
            max_interval: 1m0s
            max_retries: 3
        timeout: 1m30s
extensions:
    beatsauth/_agent-component/monitoring:
        idle_connection_timeout: 3s
        proxy_disable: false
        timeout: 1m30s
```

The full agent status from this config shows that the ssl parameters are respected by the standard beat system/metrics input but appear to be ignored by the monitoring exporter since it's trying to connect to ES instead of failing to read the non-existent .pem files I told it to look for:

```
sudo elastic-development-agent status
┌─ fleet
└─ elastic-agent
   ├─ status: (FAILED) OTel manager failed: failed to generate otel config: error translating config for output: monitoring, unit: http/metrics-monitoring, error: failed unpacking config. open /tmp/cert.pem: no such file or directory /tmp/cert.pem accessing config
   ├─ beat/metrics-monitoring
   │  └─ beat/metrics-monitoring
   ├─ filestream-monitoring
   │  └─ filestream-monitoring
   ├─ http/metrics-monitoring
   │  └─ http/metrics-monitoring
   ├─ system/metrics-default
   │  ├─ status: (HEALTHY) Healthy: communicating with pid '64594'
   │  ├─ system/metrics-default
   │  │  └─ status: (FAILED) could not start output: failed to reload output: open /tmp/cert.pem: no such file or directory /tmp/cert.pem accessing 'elasticsearch'
   │     └─ status: (STARTING) Starting
   └─ extensions
      ├─ status: StatusOK
      └─ extension:beatsauth/_agent-component/monitoring
         └─ status: StatusOK
```

This one needs to be investigated and fixed before we can merge this.

3. Mikolaj's point about extension conflicts is a good one and can also be tackled as a separate PR and issue. I think this is something we need more testing for in general with hybrid agent.

cmacknz:
For testing we already have https://github.com/elastic/beats/issues/45491 for tracking that separately beyond what is already done in here.

internal/pkg/otel/translate/otelconfig.go
cmacknz: https://github.com/elastic/opentelemetry-dev/issues/756 this is a good point we should add a test for this, we can do it separately so Khushi can get this in before her time off next week.

khushijain21:
> A couple of observations testing this manually:

For your second point above.

> The full agent status from this config shows that the ssl parameters are respected by the standard beat system/metrics input but appear to be ignored by the monitoring exporter since it's trying to connect to ES instead of failing to read the non-existent .pem files I told it to look for:

I believe you started `elastic-agent` without the ssl block first first and then introduced 
```
    ssl.key: /tmp/key.pem
    ssl.certificate: /tmp/cert.pem
``` 

so what happens is elastic-agent tries to create a new otel config from this, fails because it cannot read the non-existentent pem file and otel-manager continues with the previous `elastic-agent.yml` config without the ssl block. This is why you it continues to show connection refused.

Although, it does report the error here. "Failed to translate the config"
```
└─ elastic-agent
   ├─ status: (FAILED) OTel manager failed: failed to generate otel config: error translating config for output: monitoring, unit: http/metrics-monitoring, error: failed unpacking config. open /tmp/cert.pem: no such file or directory /tmp/cert.pem accessing config
```

> In otel-merged.yml from generated from that configuration I see the following with the ssl parameters missing:

This is also why the diagnostics have missing ssl config. It never built the new config with incorrect ssl file path in the first place


The standard beat system/metrics throws an explicit error that it failed to read the pem file because we haven't set the `_runtime_experimental:otel` and it starts the standard beat

internal/pkg/otel/translate/otelconfig_test.go
khushijain21: Added in [f61a6e8](https://github.com/elastic/elastic-agent/pull/9257/commits/f61a6e8be508724e7bff3de7e5c29c11131c4f6f)

cmacknz:
> Although, it does report the error here. "Failed to translate the config"

Thanks I totally missed agent was reporting the failure at the top level, we can iterate on this separately as part of https://github.com/elastic/elastic-agent/issues/9771. Your explanation for why the ssl blocks are missing from the otel-merged.yml make sense, I would prefer we show what we are trying and failing to run there but I don't think it needs to block this PR.

internal/pkg/otel/translate/otelconfig_test.go
cmacknz: Thanks!

khushijain21:
@cmacknz can you approve this PR? so we can merge this

PR #9266, opened 2025-08-06
[FIPS] Only consider versions available in ECH region for `TestUpgradeIntegrationsServer`

ycombinator:
## What does this PR do?

This PR restricts the versions used as start versions in `TestUpgradeIntegrationsServer` to those actually available in the ECH region where the integrations server is being run.

## Why is it important?

The test should not attempt to use a start version that doesn't actually exist in the ECH region where the integrations server deployment will be created.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] ~I have made corresponding changes to the documentation~
- [ ] ~I have made corresponding change to the default configuration files~
- [ ] ~I have added tests that prove my fix is effective or that my feature works~
- [ ] ~I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~
- [ ] ~I have added an integration test or an E2E test~

## Disruptive User Impact

None; this PR modifies an integration test.

## Related issues

Follow up to https://github.com/elastic/elastic-agent/pull/9207#issuecomment-3157529153

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

Issue #9286, opened 2025-08-07
[CI] Removing the `Testing:run:TestUpgradeIntegrationsServer` label on PRs does not skip the `TestUpgradeIntegrationsServer` test


ycombinator:
In https://github.com/elastic/elastic-agent/pull/8417, we added a new integration test, `TestUpgradeIntegrationsServer`.  We also added logic to our Buildkite pipeline so that this test would be run only on PRs that had the `Testing:run:TestUpgradeIntegrationsServer` label set on them: https://github.com/elastic/elastic-agent/blob/974a19d278dd8ec29076a113e31d8e7bcf3b4f52/.buildkite/bk.integration-fips.pipeline.yml#L104

As such, we expect that PRs where that label is removed after being set would not run the `TestUpgradeIntegrationsServer` test in subsequent CI runs.  However, this is not what is actually happening: https://github.com/elastic/elastic-agent/pull/9283#issuecomment-3165131918

ycombinator:
@elastic/observablt-robots perhaps someone from your team could take a look and help figure out how to fix this bug?  Thanks!

oakrizan:
Ran several tests and was not able to reproduce the issue:

- https://buildkite.com/elastic/elastic-agent/builds/27249 with label Testing:run:TestUpgradeIntegrationsServer
- https://buildkite.com/elastic/elastic-agent/builds/27250 without label Testing:run:TestUpgradeIntegrationsServer
- https://buildkite.com/elastic/elastic-agent/builds/27251 with label Testing:run:TestUpgradeIntegrationsServer (added 2nd time)
- https://buildkite.com/elastic/elastic-agent/builds/27252 without label Testing:run:TestUpgradeIntegrationsServer (removed 2nd time)

v1v:
> [buildkite.com/elastic/elastic-agent/builds/27250](https://buildkite.com/elastic/elastic-agent/builds/27250) without label Testing:run:TestUpgradeIntegrationsServer

IIUC, it ran, or did I miss something?

v1v:
> As such, we expect that PRs where that label is removed after being set would not run the TestUpgradeIntegrationsServer test in subsequent CI runs. However, this is not what is actually happening: https://github.com/elastic/elastic-agent/pull/9283#issuecomment-3165131918

[...]

PR #9290, opened 2025-08-07
Use stdlib function instead of own to check for slice membership

ycombinator:
Quick cleanup on the heels of https://github.com/elastic/elastic-agent/pull/9266.  Thought of this right after I merged that PR 🤦.

### DIFF
--- a/testing/integration/ess/upgrade_integrations_server_test.go
+++ b/testing/integration/ess/upgrade_integrations_server_test.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"sort"
 	"testing"
 	"time"
@@ -136,19 +137,10 @@ func filterVersionsForECH(t *testing.T, versions []*version.ParsedSemVer, echPro
 
 	filteredVersions := make([]*version.ParsedSemVer, 0)
 	for _, ver := range versions {
-		if isVersionInList(ver, echVersions) {
+		if slices.Contains(echVersions, ver) {
 			filteredVersions = append(filteredVersions, ver)
 		}
 	}
 
 	return filteredVersions
-}
-
-func isVersionInList(candidateVersion *version.ParsedSemVer, allowedVersions []*version.ParsedSemVer) bool {
-	for _, allowedVersion := range allowedVersions {
-		if allowedVersion.Equal(*candidateVersion) {
-			return true
-		}
-	}
-	return false
 }


pkoutsovasilis (commented):
Tbh I reviewed the original PR and I thought that you went the manual approach becuase slices.Contains does essentially the following

```
// Index returns the index of the first occurrence of v in s,
// or -1 if not present.
func Index[S ~[]E, E comparable](s S, v E) int {
	for i := range s {
		if v == s[i] {
			return i
		}
	}
	return -1
}
```

and because you pass pointers this essentially is not checking the version equality but the ptr ref, what am I missing?

Also I see in the CI that adding the label doesn't invoke the step which is really weird

ycombinator:
> and because you pass pointers this essentially is not checking the version equality but the ptr ref, what am I missing?

Yeah, good point.  I could do something like this:

```go
for _, ver := range versions {
	if slices.ContainsFunc(echVersions, func(candidateVer *version.ParsedSemVer) bool {
		return candidateVer.Equal(*ver)
	}) {
		filteredVersions = append(filteredVersions, ver)
	}
}
```

But I'm not sure that's buying us much over the current manual approach.  WDYT?

> Also I see in the CI that adding the label doesn't invoke the step which is really weird

In fairness, I added the label after the CI run had already started.  Let me rebuild now and see if the label takes effect.

pkoutsovasilis:
> Yeah, good point. I could do something like this:
> 
> ```go
> for _, ver := range versions {
> 	if slices.ContainsFunc(echVersions, func(candidateVer *version.ParsedSemVer) bool {
> 		return candidateVer.Equal(*ver)
> 	}) {
> 		filteredVersions = append(filteredVersions, ver)
> 	}
> }
> ```
>
> But I'm not sure that's buying us much over the current manual approach. WDYT?

It feels like the same approach that is already there 🙂 
 
> > Also I see in the CI that adding the label doesn't invoke the step which is really weird
> 
> In fairness, I added the label after the CI run had already started. Let me rebuild now and see if the label takes effect.

I cannot wrap my head around how BK reacts to the gh labels, because the test is still not invoked

ycombinator:
> It feels like the same approach that is already there 🙂

Agreed.  Let's just stick with what we already have.  Closing this PR unmerged. 

> I cannot wrap my head around how BK reacts to the gh labels, because the test is still not invoked

Same. Let's use https://github.com/elastic/elastic-agent/issues/9286 to fix the automation for this label.

PR #9291, opened 2025-08-07
fix: push the cloud images from packaging steps

pkoutsovasilis:
## What does this PR do?

During the backport of https://github.com/elastic/elastic-agent/pull/9282 I missed the fact that this https://github.com/elastic/elastic-agent/pull/8337 never got backported to 9.1. This leads to certain failures [context here](https://github.com/elastic/elastic-agent/pull/9262#issuecomment-3166096573)

So, this PR ensures that cloud images built as part of the packaging steps for integration tests are properly pushed to the Docker registry. Previously, this push step was part of a separate FIPS-specific step (`integration-fips-cloud-image`), which has now been removed.

The logic to load and push the cloud image has been relocated to the packaging steps themselves (`packaging-containers-x86-64-fips` and others), specifically using the new `integration-cloud-image-push.sh` script. In addition, a shared Vault Docker login plugin has been introduced and reused via anchors across these steps.

## Why is it important?

This fix ensures that the required images for FIPS integration tests are available without relying on a separate Buildkite step and most importantly they do use the `USE_PACKAGE_VERSION` env var

## Checklist

## Disruptive User Impact

None. This is a CI-only change and has no impact on users of the Elastic Agent.

This is a CI-only change thus maybe open a draft PR

## Related issues

- N/A

### DIFF
--- a/.buildkite/integration.pipeline.yml
+++ b/.buildkite/integration.pipeline.yml
@@ -9,6 +9,11 @@ env:
   IMAGE_UBUNTU_2204_X86_64: "platform-ingest-elastic-agent-ubuntu-2204-1753491662"
   IMAGE_UBUNTU_2204_ARM_64: "platform-ingest-elastic-agent-ubuntu-2204-aarch64-1753491662"
 
+common:
+  - vault_docker_login: &vault_docker_login
+      elastic/vault-docker-login#v0.5.2:
+        secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
+
 steps:
   - group: "Integration tests: packaging"
     key: "int-packaging"
@@ -108,6 +113,7 @@ steps:
           PLATFORMS: "linux/amd64"
         command: |
           .buildkite/scripts/steps/integration-package.sh
+          .buildkite/scripts/steps/integration-cloud-image-push.sh
         artifact_paths:
           - build/distributions/**
         agents:
@@ -115,6 +121,8 @@ steps:
           machineType: "n2-standard-8"
           diskSizeGb: 200
           image: "${IMAGE_UBUNTU_2204_X86_64}"
+        plugins:
+          - *vault_docker_login
 
       - label: "Packaging: Containers linux/arm64"
         key: packaging-containers-arm64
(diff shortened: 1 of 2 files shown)


.buildkite/integration.pipeline.yml
```diff
@@ -108,13 +113,16 @@ steps:
           PLATFORMS: "linux/amd64"
         command: |
           .buildkite/scripts/steps/integration-package.sh
+          .buildkite/scripts/steps/integration-cloud-image-push.sh
```
dliappis, line 116: Iirc when having two separate commands like this if the first fails and the next succeeds the step won't fail.

I think we need either `&&` or `set -euo pipefail` as the first statement (the latter will fail the step on any error inside any script not just based on the exit code though)

Same comment for the rest

pkoutsovasilis: thanks for the point @dliappis, I had the impression that this wasn't required https://github.com/elastic/elastic-agent/pull/8337#discussion_r2195300630. But if it is, I need to do a follow-up PR and start from main and then bring over 9.1 and 8.19. wdyt?

dliappis: > had the impression that this wasn't required [#8337 (comment)](https://github.com/elastic/elastic-agent/pull/8337#discussion_r2195300630). But if it is, I need to do a follow-up PR and start from main and then bring over 9.1 and 8.19. wdyt?

I wouldn't do `set -euo pipefail`, but **if memory serves right** we may need to `.buildkite/scripts/steps/integration-package.sh && .buildkite/scripts/steps/integration-cloud-image-push.sh` to ensure if either fails, the step fails too. I'll try a test later today with a PR somewhere to confirm if we need to. For now I am good with things as they are.

dliappis (approved):
LGTM

.buildkite/integration.pipeline.yml
pkoutsovasilis: no worries @dliappis, I see what you mean and let's go with better safe than sorry so I will open a follow up PR to go with && approach

Issue #7260, opened 2025-03-06
Disable permission inheritance on Windows


belimawr:
OSQuery will not load extensions if they can be written or modified by non-privileged accounts, from their [documentation](https://osquery.readthedocs.io/en/stable/deployment/extensions/):
>First, a note: the osquery agent will refuse to load an extension executable from the filesystem if the file's permissions allow write or modify by non-privileged accounts. Before loading an extension, change the owner of the your_extension.ext file to be the root account.
>
>On Windows, because of permission inheritance, just changing the owner of a file is not sufficient. You must also change the owner of the parent directory, remove all inherited DACLs, and disable inheritance. For example, if your osquery extensions are in the .\Extensions directory, the following commands will set permissions that satisfy osquery:
>
>```
>icacls .\Extensions /setowner Administrators /t
>icacls .\Extensions /grant Administrators:f /t
>icacls .\Extensions /inheritance:r /t
>icacls .\Extensions /inheritance:d /t
>```

So we need to ensure that when installing the Elastic-Agent we disable permissions inheritance. This can be done in the [`FixPermissions`](https://github.com/elastic/elastic-agent/blob/535610ab75213ee6e03b0d276e291ba925502749/internal/pkg/agent/perms/windows.go#L25) function.

Acceptance criteria:
- New installations of the Elastic-Agent do not have permissions inheritance enabled for the Agent folder.
- Current installations should disable inheritance on Elastic-Agent start up.

# Related issues
- Closes: https://github.com/elastic/elastic-agent/issues/6792
- Is blocked by: https://github.com/elastic/elastic-agent/pull/7059

@pierrehilbert, @jlind23 can we get some one from the control-plane to work on this.

pierrehilbert:
cc @ebeahan a good candidate for the next iteration content.

cmacknz:
We have a reproduction for the problem this is trying to fix documented in https://github.com/elastic/elastic-agent/issues/6792#issuecomment-2768651823

We have to prevent users from being able to add themselves to the ACLs unexpectedly via the UI.

matter1-git:
any update on this?

cmacknz:
No update but I think we should fix this, let me follow up internally.

In terms of implementation I think we need to look at setting `PROTECTED_DACL_SECURITY_INFORMATION` as described [here](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-information). The package we use for acls has this behind a simple to use `inherit` boolean.

[...]

PR #9294, opened 2025-08-08
Added support for user defined hooks - prerun

michalpristas:
Added support for user defined hooks - prerun

Spec definition sample
```yaml
version: 2
inputs:
  - name: demo-server
    description: "Demo Server"
    platforms:
      - linux/amd64
    outputs:
      - elasticsearch
    command:
      args:
        - "logging.to_stderr=true"
    hooks:
      pre_run:
        - hook_type: fix-permissions
          args:
            path: "fs/temp"
            user: michal
            group: staff
            mask: 0740
```

### DIFF
--- a/internal/pkg/agent/perms/opts.go
+++ b/internal/pkg/agent/perms/opts.go
@@ -18,9 +18,16 @@ const (
 type opts struct {
 	mask      os.FileMode
 	ownership utils.FileOwner
+	inherit   bool
 }
 
 type OptFunc func(o *opts)
+
+func WithInherit(inherit bool) OptFunc {
+	return func(o *opts) {
+		o.inherit = inherit
+	}
+}
 
 // WithMask adjusts the default mask used.
 func WithMask(mask os.FileMode) OptFunc {
--- a/pkg/component/hooks.go
+++ b/pkg/component/hooks.go
@@ -1,0 +1,261 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package component
+
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/pkg/component/hooks"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+)
+
+// Hook execution points
+const (
+	HookPreRun = "pre-run"
+)
+
+// Hook types
+const (
+	HookTypeFixPermissions = "fix-permissions"
+)
+
+var supportedHooks = []string{
+	HookTypeFixPermissions,
+}
+
+// HookDefinition defines a single hook with its type and arguments
+type HookDefinition struct {
+	Type string                 `config:"hook_type" yaml:"type" json:"type"`
+	Args map[string]interface{} `config:"args,omitempty" yaml:"args,omitempty" json:"args,omitempty"`
+}
+
+// ComponentHooks defines all hooks for a component organized by execution point
+type ComponentHooks struct {
+	PreRun  []HookDefinition `config:"pre_run,omitempty" yaml:"pre_run,omitempty" json:"pre_run,omitempty"`
+	PostRun []HookDefinition `config:"post_run,omitempty" yaml:"post_run,omitempty" json:"post_run,omitempty"`
+}
+
+// GetHooks returns hooks for a specific execution point
+func (c *ComponentHooks) GetHooks(point string) []HookDefinition {
+	switch point {
+	case HookPreRun:
+		return c.PreRun
+	default:
+		return nil
+	}
+}
+
+// Validate checks that all hooks in ComponentHooks are of known types
+func (c *ComponentHooks) Validate() error {
+	allHooks := []struct {
+		point string
+		hooks []HookDefinition
+	}{
+		{HookPreRun, c.PreRun},
+	}
+
+	var validationErrors []string
+
+	for _, hookGroup := range allHooks {
+		for i, hook := range hookGroup.hooks {
+			if hook.Type == "" {
+				validationErrors = append(validationErrors,
+					fmt.Sprintf("hook at index %d in %s has empty type", i, hookGroup.point))
+				continue
+			}
+
+			if !isValidHookType(hook.Type) {
+				validationErrors = append(validationErrors,
+					fmt.Sprintf("unknown hook type '%s' at index %d in %s. Supported types: %v",
+						hook.Type, i, hookGroup.point, supportedHooks))
+			}
+		}
+	}
+
+	if len(validationErrors) > 0 {
+		return fmt.Errorf("hook validation failed: %s", strings.Join(validationErrors, "; "))
+	}
+
+	return nil
+}
+
+// Run executes the hook with its arguments
+func (hd *HookDefinition) Run() error {
+	switch hd.Type {
+	case HookTypeFixPermissions:
+		return hd.fixPermissions()
+	default:
+		return fmt.Errorf("unknown hook type: %s", hd.Type)
+	}
+}
+
+// GetStringArg safely extracts a string argument
+func (hd *HookDefinition) GetArg(key string) (interface{}, bool) {
+	if val, exists := hd.Args[key]; exists {
+		return val, true
+	}
+	return "", false
+}
+
+// GetStringArg safely extracts a string argument
+func (hd *HookDefinition) GetStringArg(key string) (string, bool) {
+	if val, exists := hd.Args[key]; exists {
+		if str, ok := val.(string); ok {
+			return str, true
+		}
+	}
+	return "", false
+}
+
+// GetIntArg safely extracts a int argument
+func (hd *HookDefinition) GetIntArg(key string) (int, bool) {
+	if val, exists := hd.Args[key]; exists {
+		switch v := val.(type) {
+		case int:
+			return v, true
+		case int32:
+			return int(v), true
+		case int64:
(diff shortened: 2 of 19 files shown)


pkg/component/hooks/fix_permissions_test.go
```diff
@@ -0,0 +1,5 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package hooks
```
pkoutsovasilis: something is missing from here

internal/pkg/agent/perms/opts.go
```diff
@@ -18,10 +18,17 @@ const (
 type opts struct {
 	mask      os.FileMode
 	ownership utils.FileOwner
+	inherit   bool
 }
 
 type OptFunc func(o *opts)
 
+func WithInherit(inherit bool) OptFunc {
```
pkoutsovasilis, line 27: add a godoc explaining what is enabled with this Option

pkg/component/hooks.go
```diff
@@ -0,0 +1,261 @@
...
+	"strings"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/pkg/component/hooks"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+)
+
+// Hook execution points
+const (
+	HookPreRun = "pre-run"
```
pkoutsovasilis, line 20: I think that the Hook we are after here is `PostLoad` which is gonna execute that after the spec is loaded by elastic-agent. Having that being invoked every time we run the input of the spec seems inefficient and I think that the former covers the needs of the bug you are trying to fix here

pkg/component/hooks.go
```diff
@@ -0,0 +1,261 @@
...
+)
+
+// Hook types
+const (
+	HookTypeFixPermissions = "fix-permissions"
```
pkoutsovasilis: this is `apply-permissions` you don't know if the hook definition is gonna fix or break something

pkg/component/hooks.go
```diff
@@ -0,0 +1,261 @@
...
+// HookDefinition defines a single hook with its type and arguments
+type HookDefinition struct {
+	Type string                 `config:"hook_type" yaml:"type" json:"type"`
+	Args map[string]interface{} `config:"args,omitempty" yaml:"args,omitempty" json:"args,omitempty"`
+}
+
+// ComponentHooks defines all hooks for a component organized by execution point
+type ComponentHooks struct {
+	PreRun  []HookDefinition `config:"pre_run,omitempty" yaml:"pre_run,omitempty" json:"pre_run,omitempty"`
+	PostRun []HookDefinition `config:"post_run,omitempty" yaml:"post_run,omitempty" json:"post_run,omitempty"`
```
pkoutsovasilis: not exactly needed 

```suggestion
	PostRun []HookDefinition `config:"post_run,omitempty" yaml:"post_run,omitempty" json:"post_run,omitempty"`
```

pkg/component/hooks.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+		}
+	}
+
+	if len(validationErrors) > 0 {
+		return fmt.Errorf("hook validation failed: %s", strings.Join(validationErrors, "; "))
+	}
+
+	return nil
+}
```
pkoutsovasilis, line 85: this is untested

pkg/component/hooks.go
```diff
@@ -0,0 +1,261 @@
...
+	if val, exists := hd.Args[key]; exists {
+		switch v := val.(type) {
+		case int:
+			return v, true
+		case int32:
+			return int(v), true
+		case int64:
+			return int(v), true
+		case uint64:
+			return int(v), true //nolint:gosec // G115 Conversion from int to uint32 is safe here.
```
pkoutsovasilis, line 126: why the conversion is safe here? I understand that we do that for PIDs, but this is an arbitrary argument

pkg/component/runtime/command.go
```diff
@@ -360,6 +362,15 @@ func (c *commandRuntime) start(comm Communicator) error {
...
 		return nil
 	}
+
+	// pre run hook
+	if c.current.InputSpec != nil {
+		c.log.Debugf("running PreRun hooks for %q", c.current.ID)
+		if err := c.current.InputSpec.Spec.Hooks.ExecuteHooks(component.HookPreRun, c.log); err != nil {
+			return fmt.Errorf("failed during pre-run hook execution for %q: %w", c.current.ID, err)
+		}
+	}
```
pkoutsovasilis, line 372: as I commented above, adding a pre-run hook seems as not required for the needs of the issue and this I believe could be satisfied by a `PostLoad` hook https://github.com/elastic/elastic-agent/blob/3c453fec11845c09ed1890014b94ea31a9b6b7ef/pkg/component/load.go#L164-L171

pkg/component/hooks.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	path = injectComponentsPath(path)
+	inheritPermissions, _ := hd.GetBoolArg("inherit_permissions")
+	username, _ := hd.GetStringArg("user")
+	groupname, _ := hd.GetStringArg("group")
+	failOnNotExist, _ := hd.GetBoolArg("fail_on_path_not_exist")
+	mask, _ := hd.GetIntArg("mask")
+
+	return hooks.FixPermissions(path, inheritPermissions, username, groupname, failOnNotExist, mask)
+}
```
pkoutsovasilis: it would be nice if you could write an integration test with osquery to validate that this works as expected

pkoutsovasilis (commented):
I did left some comments. In general if I understand correctly the issue this PR is trying to solve https://github.com/elastic/elastic-agent/issues/7260, we want to disable inheritance not only for the top path but also for the path that `osquery-extension.exe`, `osqueryd.exe` reside, aka `data/elastic-agent-${commit_sha}/components`. To this end, I would imagine something "safer" IMO that doesn't involve dynamic hooks and args that derive from yaml. 

Thinking out load, this could be a new field in the spec of an input that defines that the folder that the component resides in should have inheritance disabled e.g.

```
....
spec:
  attributes:
  - FolderInheritanceDisabled
 ...
```
which when we load the spec at elastic-agent runtime we can execute only once and deterministically without any external yaml-based input dependencies. wdyt?

cc @cmacknz 

PS: I also feel that we need an arg to disable executing the hooks as an escape hatch for potential SDHs

cmacknz:
> In general if I understand correctly the issue this PR is trying to solve https://github.com/elastic/elastic-agent/issues/7260, we want to disable inheritance not only for the top path but also for the path that osquery-extension.exe, osqueryd.exe reside, aka data/elastic-agent-${commit_sha}/components.

Yes it seems like we want permission inheritance disabled recursively from the top path, not just on the top path itself. I am rebuilding my mental context on this from my own previous comments like https://github.com/elastic/elastic-agent/issues/7260#issuecomment-3059957499

Thinking more on this, it's not clear why we would want this to be property of a binary declared via the spec file and not a property of the entire agent directory tree. Scoping it to just osquery is less risky and I like that aspect, but I don't know why the choice to disable inheritance would be per binary.

I think in general the biggest thing here for whatever solution we decide on is to have proof that it fixes this problem both for new installs, and for already existing installs so we don't force everyone to reinstall. I see @pkoutsovasilis has already asked for more testing along these lines, thanks for that.

cmacknz:
> Thinking more on this, it's not clear why we would want this to be property of a binary declared via the spec file and not a property of the entire agent directory tree. 

Iterating on this thought a bit more and the need for some proof this truly fixes it: is disabling permission inheritance on just the osquery executable files enough? Do we have to consider the containing directories like components/? If we do that, we can't really have this be a per binary choice because the parent directories are shared amongst all of them.

pkg/component/hooks.go
michalpristas: having this run before every run is kinda required. when you break and apply proper permissions and then administrator opens up location from explorer, windows adds permissions for opened directory causing osquery to fail on next run.

pkg/component/hooks.go
michalpristas: Elastic Agent is not supported on 32-bit operating systems.

pkg/component/hooks.go
pkoutsovasilis: but this happens only because the folder has inheritance enabled? if you disable that this should not happen?! or at least this is my understanding of it

pkg/component/hooks.go
pkoutsovasilis: this is a uint64 -> int @michalpristas

michalpristas:
> > Thinking more on this, it's not clear why we would want this to be property of a binary declared via the spec file and not a property of the entire agent directory tree.
> 
> Iterating on this thought a bit more and the need for some proof this truly fixes it: is disabling permission inheritance on just the osquery executable files enough? Do we have to consider the containing directories like components/? If we do that, we can't really have this be a per binary choice because the parent directories are shared amongst all of them.


as i understood, the problem is osquery performing checks on `./Extensions` directory. so performing inheritance break on each run everywhere seems really inefficient and i'm not sure there is any functional benefits of doing so. also with a problem of visiting admin breaking a check we would need to perform this recursively before any process start(?).

pkg/component/hooks/fix_permissions_test.go
michalpristas: oh yes I split it per arch.

pkg/component/hooks.go
michalpristas: yes once we agree on the outcome i need to do spec change in beats repo, then i can implement integration test. i see this as 3 steps

michalpristas:
closing,

cmacknz:
Is there an issue or PR with the alternative solution yet we can link to?

PR #9308, opened 2025-08-11
feat: separate EDOT from elastic-agent 

pkoutsovasilis:
## What does this PR do?

This PR extracts EDOT (Elastic Distribution for OpenTelemetry) code from the elastic-agent control-plane binary into a standalone binary. The changes include:

- **Code Restructuring**: Moved all EDOT/OTel-related code from `internal/pkg/otel` to `internal/edot/pkg` and created a new standalone module with its own `go.mod`
- **New Binary**: Created a standalone `edot` binary with its own entrypoint at `internal/edot/main.go` that consolidates OpenTelemetry Collector functionality
- **Thin Wrapper**: Replaced the `elastic-agent otel` subcommand with a thin wrapper that executes the standalone `edot` binary as a supervised subprocess
- **Build Configuration**: Updated the build process (`magefile.go`) to generate two separate binaries: `elastic-agent` (control plane) and `edot` (data plane)
- **CI/CD Updates**: Updated scripts (`.ci/scripts/update-otel.sh`), GitHub workflows, and CODEOWNERS to reference the new `internal/edot` structure

The architecture now cleanly separates the control plane (elastic-agent) from the data plane (edot), with elastic-agent managing the edot collector as a supervised subprocess when needed. There are still some remaining libbeat dependencies in the elastic-agent package due to config translation, but with this PR it would be way easier to eliminate them in the future.

## Why is it important?

**Architecture Benefits:**
- Clean separation between control plane and data plane code
- Prevents transitive EDOT/OpenTelemetry dependencies from bloating the core elastic-agent binary
- Allows elastic-agent and EDOT to be updated independently
- Enables better memory sharing when multiple processes of the same binary run

**Runtime Memory Optimization:**
- Reduces runtime memory when EDOT is not actively running
- Previously, importing EDOT code increased binary size and runtime memory even when the collector was not being used
- Large static data segments (.text and .rodata) are no longer loaded into elastic-agent memory unless EDOT is explicitly invoked

## Checklist

## Disruptive User Impact

**Minimal user impact expected:**

The `elastic-agent otel` command continues to work as before from a user perspective. The elastic-agent binary now acts as a thin wrapper that executes the `edot` binary, which should be transparent to users.

**Deployment considerations:**
- The elastic-agent distribution now includes two binaries (`elastic-agent` and `edot`) instead of one
- Security teams may need to add the new `edot` binary to their security detection exception lists (see elastic/global-security-exceptionlist#878)
- Both 8.19 and main branches will receive this change to keep the repository structure consistent and avoid backport challenges

* Build elastic-agent container `USE_PACKAGE_VERSION="true" SNAPSHOT="true" EXTERNAL="true" PLATFOR
MS="linux/arm64" PACKAGES="docker" DOCKER_VARIANTS="basic" mage package` and verify that the new edot binary is located under `/usr/share/elastic-agent/data/elastic-agent-{commit_sha}/`

* Build elastic-agent install archive `EXTERNAL="true" SNAPSHOT="true" PLATFORMS="linux/arm64" PACKAGES="tar.gz" mage package`. Uncompress the install archive, and enable beats receivers for agent self-monitoring. Install the archive in the develop namespace: `sudo ./build/distributions/elastic-agent-9.3.0-SNAPSHOT-linux-arm64/elastic-agent install --develop`. Elastic-agent should work as expected.

* Run any unit or integration tests; they should be successful

## Related issues

N/A


go.mod
```diff
@@ -26,12 +26,6 @@ require (
 	github.com/elastic/go-sysinfo v1.15.3
 	github.com/elastic/go-ucfg v0.8.9-0.20250307075119-2a22403faaea
 	github.com/elastic/mock-es v0.0.0-20250324153755-573fc6c0ac4b
-	github.com/elastic/opentelemetry-collector-components/connector/elasticapmconnector v0.4.0
```
kruskall: Q: do you think it would be possible to also drop the beats dependency in elastic-agent ?

pkoutsovasilis: if we do transfer the config translation from elastic-agent to EDOT (as an extension maybe?!), then yes we can drop all beats dependencies from elastic-agent. However, this is easier said than done, and also we might not even split up elastic-agent and EDOT

theletterf:
We might have to update the docs automation, too.

dev-tools/mage/target/common/notice.go
```diff
@@ -44,6 +44,8 @@ func Notice() (err error) {
 // generateNotice generates a generateNotice file with the name outputFilename.
 // see getDependentModules for use of additionalTags.
 func generateNotice(outputFilename string, additionalTags ...string) error {
+	// TODO(splitting-edot): need to figure out how to make this work with EDOT separated
```
swiatekm, line 47: The outcome should be the same as it is for other binaries shipped in the agent package. To achieve that, I think the following needs to happen:

1. This function is updated to take a go.mod location as an argument.
2. We add mage targets to generate edot-specific notice files in the edot directory.
3. These notice files are added to the package

.ci/updatecli/updatecli-bump-golang.yml
```diff
@@ -122,6 +122,15 @@ targets:
       content: 'go {{ source "latestGoVersion" }}'
       file: wrapper/windows/archive-proxy/go.mod
       matchpattern: 'go \d+.\d+.\d+'
+  update-edot-gomod-full-version:
```
swiatekm, line 125: We need to add another target for bumping the beats dependency in edot too.

swiatekm (commented):
Could we do the package rename and move in a separate PR first? It accounts for a lot of changes in this PR, but seems independent of it. We could even backport it to all active branches to make future backports easier. Then this PR would just be making it a separate module and the tooling adjustment. Do you think that's feasible?

.ci/updatecli/updatecli-bump-golang.yml
pkoutsovasilis: why this change here https://github.com/elastic/elastic-agent/blob/647e17f93bed7f13dd339d48dc42fe11dd0b47b7/dev-tools/mage/target/update/update.go#L28 doesn't satisfy your comment?

dev-tools/mage/target/common/notice.go
pkoutsovasilis: @swiatekm I tend to agree with you but there is already a NOTICE.txt inside `/usr/share/elastic-agent/data/elastic-agent-{commit_sha}/components/NOTICE.txt` so it needs some thinking how this edot-specific NOTICE.txt will appear it's not the code changes that block this from happening. I will discuss this more with the team

pkoutsovasilis:
> Could we do the package rename and move in a separate PR first? It accounts for a lot of changes in this PR, but seems independent of it. We could even backport it to all active branches to make future backports easier. Then this PR would just be making it a separate module and the tooling adjustment. Do you think that's feasible?

This isn't the easiest PR to break into smaller ones, or at least IMO not without duplicating the code, initially at least. However, if we do break it apart and not deliver the full thing under a single PR we are losing the CI actually testing the thing as whole. Honestly looking at the diff, the number of changes is scary but it is mostly NOTICE.txt changes; the actual changes are not that hefty. Either way I am gonna discuss this PR more with @ebeahan and @cmacknz

v1v (dismissed):
LGTM - I only reviewed the changes in the `.ci/updatecli` and `.github/workflows`

.ci/updatecli/updatecli-bump-golang.yml
swiatekm: Ah, so we do it in a single PR. Ok, didn't catch that.

theletterf (dismissed):

ycombinator (commented):
Looks like there are some files in `docs/reference/edot-collector/config/` that are still referencing the old path:

```
docs/reference/edot-collector/config/default-config-standalone.md:365:[Logs - ES]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/platformlogs.yml
docs/reference/edot-collector/config/default-config-standalone.md:366:[Logs - OTLP]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/managed_otlp/platformlogs.yml
docs/reference/edot-collector/config/default-config-standalone.md:367:[Logs &#124; Metrics - ES]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/platformlogs_hostmetrics.yml
docs/reference/edot-collector/config/default-config-standalone.md:368:[Logs &#124; Metrics - OTLP]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/managed_otlp/platformlogs_hostmetrics.yml
docs/reference/edot-collector/config/default-config-standalone.md:369:[Logs &#124; Metrics &#124; App - ES]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/logs_metrics_traces.yml
docs/reference/edot-collector/config/default-config-standalone.md:370:[Logs &#124; Metrics &#124; App - OTLP]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/managed_otlp/logs_metrics_traces.yml
docs/reference/edot-collector/config/default-config-standalone.md:371:[Gateway mode]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/heads/main/internal/pkg/otel/samples/linux/gateway.yml
```

ycombinator:
I see there are still some files/folders left under `github.com/elastic/elastic-agent/internal/pkg/otel`. Are we not moving _all_ files/folders from this location to `github.com/elastic/elastic-agent/internal/edot/pkg`?

magefile.go
```diff
@@ -404,33 +414,14 @@ func (Build) TestBinaries() error {
...
-			return fmt.Errorf("cannot determine darwin OS major version: %w", err)
-		}
-
-		if osMajorVer > 13 {
-			// Workaround for https://github.com/golang/go/issues/67854 until it
-			// is resolved.
-			args = append(args, "-ldflags", "-extldflags='-ld_classic'")
-		}
-	}
-
```
ycombinator, line 420: Was this removal along with the removal of `getMacOSMajorVersion` below intentional?

ycombinator:
With this PR, I'm still seeing some OTel Collector modules referenced in the root `go.mod` file. Is this expected?

magefile.go
pkoutsovasilis: oops nice catch, this wasn't intentional and it probably slipped through the cracks during rebasing. added that back in 2a10385d7cc99f531c500b2fcf79672f758e7f7e

pkoutsovasilis:
> I see there are still some files/folders left under `github.com/elastic/elastic-agent/internal/pkg/otel`. Are we not moving _all_ files/folders from this location to `github.com/elastic/elastic-agent/internal/edot/pkg`?

> With this PR, I'm still seeing some OTel Collector modules referenced in the root go.mod file. Is this expected?

both of these are still the case because we do have components that reside in the elastic-agent package and do have such dependencies, namely the beats-diagnostics extension and the config translation that elastic-agent does to otel.

pkoutsovasilis:
> Looks like there are some files in `docs/reference/edot-collector/config/` that are still referencing the old path:
> 
> ```
> docs/reference/edot-collector/config/default-config-standalone.md:365:[Logs - ES]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/platformlogs.yml
> docs/reference/edot-collector/config/default-config-standalone.md:366:[Logs - OTLP]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/managed_otlp/platformlogs.yml
> docs/reference/edot-collector/config/default-config-standalone.md:367:[Logs &#124; Metrics - ES]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/platformlogs_hostmetrics.yml
> docs/reference/edot-collector/config/default-config-standalone.md:368:[Logs &#124; Metrics - OTLP]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/managed_otlp/platformlogs_hostmetrics.yml
> docs/reference/edot-collector/config/default-config-standalone.md:369:[Logs &#124; Metrics &#124; App - ES]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/logs_metrics_traces.yml
> docs/reference/edot-collector/config/default-config-standalone.md:370:[Logs &#124; Metrics &#124; App - OTLP]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v{{version.edot_collector}}/internal/pkg/otel/samples/linux/managed_otlp/logs_metrics_traces.yml
> docs/reference/edot-collector/config/default-config-standalone.md:371:[Gateway mode]: https://raw.githubusercontent.com/elastic/elastic-agent/refs/heads/main/internal/pkg/otel/samples/linux/gateway.yml
> ```

yeah the issue here that I saw and based on that I did not change these, is that these point to tags and even if this PR goes in for older tags these references are correct. So I am not 100% sure how to correctly handle/change them, @theletterf maybe you could help here? 🙂

dev-tools/mage/crossbuild.go
```diff
@@ -172,7 +173,30 @@ func CrossBuild(options ...CrossBuildOption) error {
...
+			})
+			if err != nil {
+				return err
+			}
+			for _, file := range goModFiles {
+				dir, err := filepath.Abs(filepath.Dir(file))
+				if err != nil {
+					return fmt.Errorf("error getting absolute dir of %q: %w", file, err)
+				}
+				fmt.Printf(">> Running go download tidy inside %q\n", dir)
```
pchila, line 188: Did you mean:
```suggestion
				fmt.Printf(">> Running go mod download inside %q\n", dir)
```

internal/edot/cmd/diagnostics_otel.go
```diff
@@ -111,3 +112,19 @@ func aggregateComponentDiagnostics(diags []client.DiagnosticComponentResult) []c
 	}
 	return result
 }
+
+func createFile(filepath string) (*os.File, error) {
+	// Ensure all the folders on filepath exist as os.Create does not do so.
+	// 0777 is the same permission, before unmask, os.Create uses.
+	dir := path.Dir(filepath)
+	if err := os.MkdirAll(dir, 0777); err != nil {
```
pchila, line 120: Is there a specific reason for `0777` permissions ? Wouldn't a `0755` suffice ?

internal/pkg/agent/cmd/otel.go
```diff
@@ -5,76 +5,84 @@
...
+			elasticAgentExecutableDir, err := paths.RetrieveExecutableDir()
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to get the path of edot executable: %w", err)
 			}
-			supervised, err := cmd.Flags().GetBool(manager.OtelSetSupervisedFlagName)
-			if err != nil {
-				return err
+
+			edotExecutable := filepath.Join(elasticAgentExecutableDir, "edot")
```
pchila, line 35: Nit: maybe create a constant with the EDOT executable name ?

internal/pkg/agent/cmd/otel.go
```diff
@@ -5,76 +5,84 @@
...
-				return err
+
+			// Create the arguments
+			var edotArgs []string
+			if len(os.Args) > 1 {
+				edotArgs = os.Args[1:]
 			}
-			supervisedMonitoringURL, err := cmd.Flags().GetString(manager.OtelSupervisedMonitoringURLFlagName)
+
+			// TODO(splitting-edot): find a way to decouple properly elastic-agent flags from edot flags
```
pchila, line 46: Hint: maybe there's a way to use the elastic-agent[ common flag definition](https://github.com/elastic/elastic-agent/blob/b74942add0a8066bd6939a1a7fe374201f207058/internal/pkg/agent/cmd/common.go#L68-L70)  to filter out the unwanted arguments from the command line ?

magefile.go
```diff
@@ -1806,8 +1811,8 @@ func useDRAAgentBinaryForPackage(ctx context.Context, manifestURL string, versio
 
 		log.Printf("copying %q to %q", srcBinaryPath, dstBinaryPath)
 
-		err = filecopy.Copy(srcBinaryPath, dstBinaryPath, filecopy.Options{
-			PermissionControl: filecopy.PerservePermission,
+		err = copy.Copy(srcBinaryPath, dstBinaryPath, copy.Options{
```
pchila, line 1814: Not relevant to the specific line, placing the comment here because it's related to  `useDRAAgentBinaryForPackage` function. Is `edot` binary pushed as a DRA like `elastic-agent-core` (or maybe it should be bundled in it) ? How would the `edot` binary be packaged during unified release builds where `elastic-agent` binary is pulled from DRA ?

pchila (commented):
Completed a first pass I left a few comments.
The possible blocker is the question about using `edot` binary from `elastic-agent-core` DRA during unified builds, the others are mostly nitpicks

internal/pkg/agent/cmd/otel.go
```diff
@@ -5,76 +5,84 @@
...
+			edotArgs = slices.Filter(nil, edotArgs, func(arg string) bool {
+				switch arg {
+				case "-c", "-v", "-e", "-d":
+					return false
+				default:
+					return true
+				}
+			})
+
+			proc, err := process.Start(
```
swiatekm, line 56: To make this behave the same way as `elastic-agent otel` does currently, shouldn't we reexec here instead?

pkoutsovasilis:
ok after discussing with the team we decided to close this PR and have other smaller follow-up ones that incrementally achieve the same thing

PR #9313, opened 2025-08-11
Retry starting/installing service upon failure

ycombinator:
## What does this PR do?

This PR enhances the service runtime to retry starting a service if it fails to start, up to a maximum number of attempts.  Starting a service also installs it first.

## Why is it important?

To make service startup resilient to transient errors.

## Checklist

- [ ] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues

-

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #9320, opened 2025-08-12
Allow Endpoint service to keep running if its `install` exit code is non-fatal

ycombinator:
## What does this PR do?

This PR allows services such as Endpoint to define a list of non-fatal exit codes for any of their operations.  If the service runtime detects such a non-fatal exit code when it attempts to execute the operation, it will log a warning and continue to run the service.

## Why is it important?

Exit code 28 from Endpoint is non-fatal because the Endpoint service is actually still running and able to check-in with Elastic Agent.  It is emitted when Agent is unable to (re)install Endpoint while being tamper protected.  

The fix in this PR is a temporary fix for a corner case bug that arises when the following scenario unfolds:
- Agent is running with a policy that has Endpoint in it and has tamper protection enabled.
- Agent is upgraded.
- During the upgrade, while Agent is in the `UPG_WATCHING` state and Endpoint has been upgraded, the Upgrade Watcher encounters a situation that causes it to roll back Agent, e.g. the Agent service is killed.
- Agent rolls back successfully but is unable to rollback Endpoint as well.  This is because Agent's Service Runtime tries to invoke Endpoint's `install` operation, which attempts to rollback Endoint but isn't able to due to tamper protection.  The operation exits with a `28` exit code, which Agent interprets as a fatal error and reports the Endpoint component as `FAILED`.  
- Further, Agent shuts down the connection info server so even though Endpoint _is actually still running_ (albeit the wrong version compared to Agent), Endpoint is unable to check-in with Agent and eventually reports itself as being `ORPHANED`.

## Checklist

## Disruptive User Impact

None; in the corner case described above, Agents that were reporting as Orphaned will now instead report as Healthy.  Agent logs will contain a warning about continuing to run the Endpoint service after receiving a `28` non-fatal exit code.

1. Build Agent with this PR and enroll it in Fleet.  You may want to build the Agent with `AGENT_PACKAGE_VERSION` lower than the latest so you can easily upgrade the Agent later.
2. Add the Elastic Defend integration to the Agent's policy.
3. Enable tamper protection for the Agent.
4. Upgrade the Agent.
5. On the Agent's host, run `sudo /opt/Elastic/Endpoint/elastic-endpoint version` until you notice that the Endpoint version has been upgraded.
6. Kill the Elastic Agent service.  On Linux: `sudo systemctl stop elastic-agent`.
7. Wait for Agent to be downgraded.
8. Check the Fleet UI for the next few minutes and ensure that Agent keeps reporting itself as Healthy.

## Related issues

-

## Questions to ask yourself

- ...


cmacknz:
> This exit code is returned by the Endpoint install operation when Agent is unable to (re)install Endpoint while being tamper protected.

It comes from the check command during upgrades as well, since the `endpoint-security verify` command is what upgrades endpoint.

cmacknz:
There is another ongoing support case where we suspect a spurious error caused us to get into this situation, I think because any error causes us to stop the connection info server and not restart it again.

This makes me think we should have us periodically re-attempt starting endpoint on failure at a low rate (could just be a constant 30s). I still think the non fatal error approach is best for the known case of invalid uninstall token on rollback, but there could be other cases where things are left in a broken state otherwise.

ycombinator:
> This makes me think we should have us periodically re-attempt starting endpoint on failure at a low rate (could just be a constant 30s). I still think the non fatal error approach is best for the known case of invalid uninstall token on rollback, but there could be other cases where things are left in a broken state otherwise.

I'll keep this PR here about the non-fatal exit code.  Here's a separate PR for restarting endpoint: https://github.com/elastic/elastic-agent/pull/9313.  It's missing the 30s delay there; I'll add it.

pkoutsovasilis:
@ycombinator 👋  this PR should help with the CI failures in these PRs right?
* https://github.com/elastic/elastic-agent/pull/9339
* https://github.com/elastic/elastic-agent/pull/9335
* https://github.com/elastic/elastic-agent/pull/9334 (the Debian and Ubuntu failures)

ycombinator:
> @ycombinator 👋 this PR should help with the CI failures in these PRs right?
> 
> * [[main][Automation] Update versions #9339](https://github.com/elastic/elastic-agent/pull/9339)
> * [[8.19][Automation] Update versions #9335](https://github.com/elastic/elastic-agent/pull/9335)
> * [[9.1][Automation] Update versions #9334](https://github.com/elastic/elastic-agent/pull/9334) (the Debian and Ubuntu failures)

Ummm, no, because in these cases it's Endpoint's `uninstall` command that's returning the `28` error code.  This PR here only accepts that error as non-fatal when the service runtime is trying to start Endpoint, which calls its `check` and `install` commands. 

Also, this PR here was not made in response to those CI failures.  It was more in response to SDHs.  Something must've changed recently; I don't recall those tests failing consistently in CI like this.

ycombinator:
Unit test introduced in this PR, `TestCISKeepsRunningOnNonFatalExitCodeFromStart` is failing on Windows.  Moving PR back to draft while I investigate.

michalpristas (approved):

michalpristas (approved):

specs/endpoint-security.spec.yml
```diff
@@ -45,6 +45,8 @@ inputs:
             - "--resources"
             - "endpoint-security-resources.zip"
           timeout: 600s
+          non_fatal_exit_codes:
+            - 28 # tamper protection enabled but could not (re)install
```
cmacknz, line 49: This needs to be on the `check` operation as well or this will continue to fail in upgrades.

The `check` command does upgrades, the `install` command is responsible for installing for the first time.

cmacknz: (at least this is what I remember)

cmacknz: Did we confirm that this fix on its own is enough to mitigate the problem of exit code 28 on upgrade rollback by reproducing that known problem scenario?

ycombinator: > This needs to be on the check operation as well or this will continue to fail in upgrades.

Okay, I will make a follow up PR to add this.

> Did we confirm that this fix on its own is enough to mitigate the problem of exit code 28 on upgrade rollback by reproducing that known problem scenario?

Yes, it does.  I just double checked it by building Agent off `main` (with `AGENT_PACKAGE_VERSION=9.1.0` so I could upgrade it) and following the test instructions from this PR.

ycombinator: > > This needs to be on the check operation as well or this will continue to fail in upgrades.
> 
> Okay, I will make a follow up PR to add this.

https://github.com/elastic/elastic-agent/pull/9401

cmacknz: Thanks!

kevinr-security:
@cmacknz which version is this planned to ship in & when, if you don't mind me asking?

cmacknz:
It'll be in the next patch for each version. There will be both a 9.0.6 and a 9.1.3 very soon that will contain this fix. 

8.18.6 and 8.19.3 will come a little bit after that with the fix in that series.

kevinr-security:
image


This issue does not appear to be fixed in 9.1.3 - The Orphaned problem that is.

cmacknz:
On Windows we need https://github.com/elastic/elastic-agent/pull/9445 will be in the upcoming 9.1.4.

cmacknz:
There is also another active support case related to the orphaned status that is possibly a different issue, but there are no conclusions yet. That investigation is with the Elastic Defend engineers right now.

cmacknz:
Coming back with the outcome of that other support case, there is an issue in endpoint as well that will also be fixed in 9.1.4.

Quoting one of the Defend engineers:

> We're already aware of a regression problem introduced in:
v9.1.1, v9.1.2, v9.1.3, v9.0.5, v9.0.6, v8.19.1, v8.19.2, v8.19.3, v8.18.5, v8.18.6

> With the mentioned release we changed behavior of verify command to check if Endpoint service is really running, start if it's not running, otherwise report error to Agent to reinstall it.

> We've pinpointed the problem to a tiny race condition in comms routine, which we fixed and will release with next patch releases: v9.1.4, v9.0.7, v8.19.4, v8.18.7
> 
> The problem happens only on Windows.
> 
> In the log the relevant messages are:
> 
> ```
> error: Main.cpp:652 Failed to start endpoint services: Unreachable: exit status 30, try install"
> error: Util.cpp:525 Setting up minifilter registry keys failed.: exit status 284
> However bear in mind that the customer mentioned problems with v9.0.4 are not related to this.
> ```
> 
> The workaround is to restart the Agent service whilst the Endpoint service is already running.

The orphaned status reporting has let us a detect a lot of previously invisible problems we are working through fixing.

Issue #8864, opened 2025-07-07
retry to checkin are untimely


Jimyhjy:
<img width="1651" height="872" alt="Image" src="https://github.com/user-attachments/assets/e70d18a6-544f-4ce4-a917-d85b8dda8fc7" />

when checkin failed，why need long time for retry next times?

<img width="1439" height="662" alt="Image" src="https://github.com/user-attachments/assets/c5d5c5c7-4c7d-44bd-89ec-d1e6397c62af" />

this unbuffered channel，Is it possible for it to cause a blocking wait?

cmacknz:
The screenshot is cutting off what the retry should be, in the `retry_after_ns` field. Can you include the entire log line?

https://github.com/elastic/elastic-agent/blob/de39caecf4e4f0c7f8ffd839eff0bf85b5376e61/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L199-L201

I can see there was only 1 failure so the delay here seems unusually high, but this will let us confirm if the retry was calculated wrong or if we were stuck somewhere we shouldn't be as you suggest.

Jimyhjy:
> The screenshot is cutting off what the retry should be, in the `retry_after_ns` field. Can you include the entire log line?
> 
> [elastic-agent/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go](https://github.com/elastic/elastic-agent/blob/de39caecf4e4f0c7f8ffd839eff0bf85b5376e61/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L199-L201)
> 
> Lines 199 to 201 in [de39cae](/elastic/elastic-agent/commit/de39caecf4e4f0c7f8ffd839eff0bf85b5376e61)
> 
>  f.log.Warnw("Possible transient error during checkin with fleet-server, retrying", 
>  	"error.message", err, "request_duration_ns", took, "failed_checkins", f.checkinFailCounter, 
>  	"retry_after_ns", bo.NextWait()) 
> I can see there was only 1 failure so the delay here seems unusually high, but this will let us confirm if the retry was calculated wrong or if we were stuck somewhere we shouldn't be as you suggest.

this is the complete log，"retry_after_ns":647791371979，about 10 minute.
[...]

PR #9342, opened 2025-08-13
fix: jitter backoff strategy reset

rubenruizdegauna:
## What does this PR do?

This PR fixes the reset mechanism of the jitter back off strategy. Jitter value calculation range was not reset, so it could lead to bigger duration than expected.

Random generation function has been abstracted to be able to be tested in a deterministic way. 

## Why is it important?

Not resetting the jitter generation value can make it to indefinitely increase and retries would be delayed.

## Checklist

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/8864

-

## Questions to ask yourself

- ...


ebeahan:
@rubenruizdegauna (and for reviewers) - think about if we should backport and what branches to target if we do.

cmacknz:
Since this is a small bug fix, we should be able to backport to all active release branches.

ebeahan:
@michalpristas @michel-laterman can you please review this small change for @rubenruizdegauna?

michalpristas (approved):
lgtm

PR #9343, opened 2025-08-13
[backport][8.19] bump elastic-agent dependency

khushijain21:
## What does this PR do?
This PR updates `beats`, `elastic-agent-autodiscover`, `elastic-agent-libs` and `otel-collector` versions

Relates https://github.com/elastic/elastic-agent/pull/9252

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -578,6 +578,8 @@ outputs:
 		ignoredFields := []string{
 			"@timestamp",
 			"agent.id",
+			// We ignore agent.version because a different version is forced in CI
+			"agent.version",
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (commented):
@khushijain21 I think you need to bump beats dependency at this commit `6f8eb3b26012c61d24bba0c2b048ee925f3eeb97` (derives from the DRA we have in 8.19 atm)

khushijain21:
This commit `6f8eb3b26012c61d24bba0c2b048ee925f3eeb97` points to an older version of collector so I think we cannot.
I have ignored the `agent.version` field that causes conflict.

Other tests also similarly ignore this field

pkoutsovasilis (approved):

PR #9322, opened 2025-08-12
Enhancement/5235 handle insufficient disk space errors in artifact unpack

kaanyalti:
**PR 2/6**

- Enhancement

## What does this PR do?

- This is a follow up on https://github.com/elastic/elastic-agent/pull/9122. 
- Introduces disk space error type and relevant functions
- Modifies the upgrader unpack step to wrap errors so that the errors are propagated correctly.
- Adds relevant tests, asserts that errors, specifically the disk space error, get propagated up the call chain.

## Why is it important?

Non-wrapped errors mask the underlying disk space error when using errors.Is. This is need to address the original issue.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- [x] I have added an integration test or an E2E test

## Disruptive User Impact

If the agent runs into disk space errors when it is unpacking the downloaded agent archive, the status and fleet ui will show "insufficient disk space" as the error message

## How to test this PR locally

- Build for windows, mac, linux
- Install agent (managed and standalone)
- Fill up disk until there is  approximately 500mb left
- Trigger upgrade
  - From fleet
  - Cli with remote url and file
- Validate the upgrade detail error message shows insufficient disk error message both in the status output and on fleet ui.

- Run the upgrade, step_unpack tests

## Related issues

- Relates #5235 
- Requires #9347
- Requires #9122 
- Prerequisite for #9349

PR #9349, opened 2025-08-13
Enhancement/5235 correctly wrap errors from copyActionDir and copyRunDirectory

kaanyalti:
**PR 3/6**

- Enhancement

## What does this PR do?

- Adds error wrapping in copayActionStore and copyRunDirectory functions so that the errors are properly propagated.
- Adds tests for copyActionStore, copyRunDirectory and error handling test for the Upgrade function

## Why is it important?

This pr makes sure that the errors that originate from copyActionStore and copyRunDirectory get propagated correctly up the call stack.

## Checklist

## Disruptive User Impact

none

- Build for windows, mac, linux
- Install agent (managed and standalone)
- Fill up disk until almost full
- Trigger upgrade
  - From fleet
  - Cli with remote url and file
- Validate the upgrade detail error message shows insufficient disk error message both in the status output and on fleet ui.

- Run the upgrade tests in upgrade_test.go

## Related issues

- Relates #5235 
- Requires #9322 
- Prerequisite for #9366

### DIFF
--- a/internal/pkg/agent/application/upgrade/upgrade_test.go
+++ b/internal/pkg/agent/application/upgrade/upgrade_test.go
@@ -1490,7 +1490,7 @@ func TestSetClient(t *testing.T) {
 }
 
 func TestCopyActionStore(t *testing.T) {
-	log, _ := loggertest.New("TestCoyActionStore")
+	log, _ := loggertest.New("TestCopyActionStore")
 
 	actionStoreContent := "initial agent action_store.yml content"
 	actionStateStoreYamlContent := "initial agent state.yml content"


internal/pkg/agent/application/upgrade/common/stdlib_wrappers.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"io"
+	"os"
+)
+
+// The following are wrappers for stdlib functions so that we can mock them in tests.
+var Copy = io.Copy
+var OpenFile = os.OpenFile
+var MkdirAll = os.MkdirAll
```
kaanyalti: These ones are introduced for the other PRs related to the same [parent issue](https://github.com/elastic/elastic-agent/issues/5235)

- [unpack step pr](https://github.com/elastic/elastic-agent/pull/9322/files#diff-796b153d203e88ac77eb6c1df7a908be3588cc30dbc7917b974dc7d)
- [download step pr](https://github.com/elastic/elastic-agent/pull/9122/files#diff-796b153d203e88ac77eb6c1df7a908be3588cc30dbc7917b974dc7dd16609dceR13-R15)

Did not remove these wrapper vars because without these it is hard to convey the need for the mocking utility in stdlib_wrappers_mocker.go

kaanyalti:
~~The changelog relevant to this PR is added in #9122~~
The changelog will be in a separate PR

kaanyalti:
@pkoutsovasilis @michel-laterman the PR is ready for review

internal/pkg/agent/application/upgrade/common/stdlib_wrappers.go
pkoutsovasilis: I am scared of this package and I would vote of it to be re-visited  cc @ycombinator

pkoutsovasilis (commented):
Similar to my other comment @kaanyalti I can't tell which files are regarding this PR https://github.com/elastic/elastic-agent/pull/9322#pullrequestreview-3131239764

internal/pkg/agent/application/upgrade/common/stdlib_wrappers.go
ycombinator: @pkoutsovasilis could you elaborate why this is scary? Is it because these are globals? What if they were package-private instead (so we'd need to move these into packages where they are being used) instead of having a `common` package.

pkoutsovasilis: exactly! IMO these should be private to the package that calls them and thus only be overridden according to the case by the unit tests that belong to that package

michel-laterman: I think we can probably refactor to use https://pkg.go.dev/testing/fstest

kaanyalti: @pkoutsovasilis and @ycombinator I can duplicate stdlib_wrappers into each package where they are needed. I'll of course keep only the necessary functions in place. 
I am still thinking of keeping the mock utility in common as it is used in tests in several different packages.

ycombinator: > I am still thinking of keeping the mock utility in common as it is used in tests in several different packages.

Okay, depending on the implementation, you'll want to make sure the shared package does not expose any globals that could cause interference amongst tests (to @pkoutsovasilis point [earlier](https://github.com/elastic/elastic-agent/pull/9349#discussion_r2285041353)).

kaanyalti:
@pkoutsovasilis I added information on all the PR descriptions addressing the repeated common files. I also updated this comment in the main issue https://github.com/elastic/elastic-agent/issues/5235#issuecomment-3197547324. Please let me know if these updates clarify the ambiguity around the repeated files.

internal/pkg/agent/application/upgrade/common/stdlib_wrappers.go
kaanyalti: The shared package will only have the mock util. I need to put it in a separate package to avoid circular references as it will be used in tests in multiple packages in the upgrade dir (_upgrade, download, fs, http and so on_).

internal/pkg/agent/application/upgrade/step_unpack.go
```diff
@@ -36,7 +36,6 @@ type UnpackResult struct {
 }
 
 type copyFunc func(dst io.Writer, src io.Reader) (written int64, err error)
-type mkdirAllFunc func(name string, perm fs.FileMode) error
```
kaanyalti, line 39: moved this type to upgrade.go

kaanyalti:
@ycombinator @michel-laterman @pkoutsovasilis this pr is ready for review

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -547,49 +561,55 @@ func rollbackInstall(ctx context.Context, log *logger.Logger, topDirPath, versio
...
 
-func copyActionStore(log *logger.Logger, newHome string) error {
-	// copies legacy action_store.yml, state.yml and state.enc encrypted file if exists
-	storePaths := []string{paths.AgentActionStoreFile(), paths.AgentStateStoreYmlFile(), paths.AgentStateStoreFile()}
-	log.Infow("Copying action store", "new_home_path", newHome)
+func copyActionStoreProvider(readFile readFileFunc, writeFile writeFileFunc) copyActionStoreFunc {
+	return func(log *logger.Logger, newHome string) error {
+		// copies legacy action_store.yml, state.yml and state.enc encrypted file if exists
+		storePaths := []string{paths.AgentActionStoreFile(), paths.AgentStateStoreYmlFile(), paths.AgentStateStoreFile()}
+		log.Infow("Copying action store", "new_home_path", newHome)
```
michel-laterman, line 568: How often should this occur? Do we want this as a debug level?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -547,49 +561,55 @@ func rollbackInstall(ctx context.Context, log *logger.Logger, topDirPath, versio
...
+
+		for _, currentActionStorePath := range storePaths {
+			newActionStorePath := filepath.Join(newHome, filepath.Base(currentActionStorePath))
+			log.Infow("Copying action store path", "from", currentActionStorePath, "to", newActionStorePath)
```
michel-laterman, line 572: This should probably be a debug level entry

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -547,49 +561,55 @@ func rollbackInstall(ctx context.Context, log *logger.Logger, topDirPath, versio
...
-		currentActionStore, err := os.ReadFile(currentActionStorePath)
-		if os.IsNotExist(err) {
-			// nothing to copy
-			continue
-		}
-		if err != nil {
-			return err
+			// using writeFile instead of os.WriteFile for testability
+			if err := writeFile(newActionStorePath, currentActionStore, 0o600); err != nil {
+				return err
```
michel-laterman: We should add the paths to the error here

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -663,7 +683,9 @@ func readDirs(dir string) ([]string, error) {
 	return dirs, nil
 }
 
-func copyDir(l *logger.Logger, from, to string, ignoreErrs bool) error {
+// var fileDirCopyFunc = copy.Copy // abstraction for testability
```
michel-laterman: we should remove this comment

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -78,6 +78,16 @@ type unpackHandler interface {
...
 
+// Types used to abstract copyActionStore, copyRunDirectory and github.com/otiai10/copy.Copy
+type copyActionStoreFunc func(log *logger.Logger, newHome string) error
+type copyRunDirectoryFunc func(log *logger.Logger, oldRunPath, newRunPath string) error
+type fileDirCopyFunc func(from, to string, opts ...copy.Options) error
+
+// Types used to abstract stdlib functions
+type mkdirAllFunc func(name string, perm fs.FileMode) error
+type readFileFunc func(name string) ([]byte, error)
+type writeFileFunc func(name string, data []byte, perm fs.FileMode) error
```
michel-laterman, line 89: I'm not a big fan of how many custom types and abstractions that is being introduced here, but i'm not sure of a better way to do it.

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -1428,3 +1488,188 @@ func TestSetClient(t *testing.T) {
 	upgrader.SetClient(&mockSender{})
 	require.Equal(t, "mockURI", upgrader.artifactDownloader.(*mockArtifactDownloader).fleetServerURI)
 }
+
+func TestCopyActionStore(t *testing.T) {
+	log, _ := loggertest.New("TestCoyActionStore")
```
michel-laterman:
```suggestion
	log, _ := loggertest.New("TestCopyActionStore")
```

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -1428,3 +1488,188 @@ func TestSetClient(t *testing.T) {
...
+				newActionStorePaths = append(newActionStorePaths, newActionStorePath)
+				err = os.MkdirAll(filepath.Dir(newActionStorePath), 0o755)
+				require.NoError(t, err, "error creating directory %s", filepath.Dir(newActionStorePath))
+			}
+
+			err := testCase.copyActionStore(log, newHome)
+			if testCase.expectedError != nil {
+				require.Error(t, err, "copyActionStoreFunc should return error")
+				require.ErrorIs(t, err, testCase.expectedError, "copyActionStoreFunc error mismatch")
+				return
```
michel-laterman, line 1596: Should this be a continue?

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -1428,3 +1488,188 @@ func TestSetClient(t *testing.T) {
...
+			newRunPath := filepath.Join(baseDir, "new_dir", "run")
+
+			err = os.MkdirAll(newRunPath, 0o700)
+			require.NoError(t, err, "error creating new run directory")
+
+			err = testCase.copyRunDirectory(log, oldRunPath, newRunPath)
+			if testCase.expectedError != nil {
+				require.Error(t, err, "copyRunDirectoryFunc should return error")
+				require.ErrorIs(t, err, testCase.expectedError, "copyRunDirectoryFunc should return test error")
+				return
```
michel-laterman, line 1662: same here, continue?

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: Thank you for pointing it out

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: This log level did not change. I can update it but it is not really relevant to this PR.

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: Same as the above, the log level did not change, I can address it but I would prefer to do it in a separate PR.

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: Will add paths. Just want to note it here for documentation that we were returning only the error before.

internal/pkg/agent/application/upgrade/upgrade_test.go
kaanyalti: This is in a t.Run("test name", func(t testing.T){}). I'm returning from the func in the Run call. I think return is correct, but now that you're asking I'm going to look at it again.

internal/pkg/agent/application/upgrade/upgrade_test.go
kaanyalti: Same as the above, I'm returning from the func in the Run call. I think return is correct, but now that you're asking I'm going to look at it again.

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: I went through several iterations where in one of them I tried to gather things into logical groups so that I can create interfaces to address exactly what you're pointing out here. That PR ended up being ridiculously large. Having these disparate types really reduced the implementation complexity, albeit not being super clean.

PR #9363, opened 2025-08-14
Add autoops chart

idanl21:
This PR is adding the to elastic-agent the option to support autoOps Otel agent, it adds: 

- option to enable the AutoOps configs
- create an autoOps secret
- mount the AutoOps secret to env variables inside of the elastic-agent deployment\

command to install as an example - 

Pull the chart - 
```
helm repo add elastic https://helm.elastic.co/
```
Install the agent -
```bash
helm install elastic-agent ./deploy/helm/elastic-agent \
  --set autoOps.enabled=true \
  --set kube-state-metrics.enabled=false \
  --set kubernetes.enabled=false \
  --set autoOps.autoops_token="XXXXXX" \
  --set autoOps.autoops_es_url="https://XX-XX.es.us-central1.gcp.cloud.es.io" \
  --set autoOps.temp_resource_id="6XXXXXXX" \
  --set autoOps.otel_url="https://otel-collector.auto-ops.us-east-2.aws.cloud.elastic.co:4318" \
  --set autoOps.cloud_connected_mode_api_key="XXXXXX" \
  --set autoOps.cloud_connected_mode_api_url="XXXXXX" \
  --set autoOps.es_api_key="XXXXXXX:XXXXX" \
  ```

### CODE REVIEW

File: deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+
+  - name: AUTOOPS_ES_URL
+    valueFrom:
+      secretKeyRef:
+        name: {{ $agentName }}-autoops
+        key: autoops-es-url
+
+  # Only if API key provided
+  {{- if $.Values.autoOps.es_api_key }}
+  - name: ELASTICSEARCH_READ_API_KEY
```
pkoutsovasilis, line 94: please prefix all specific to AutoOps env vars with AUTOOPS, this way we make sure this doesn't collide with any other integration that might get added in the chart in the future

deploy/helm/elastic-agent/values.schema.json
```diff
@@ -331,6 +331,48 @@
...
+                },
+                "es_api_key": {
+                    "type": "string",
+                    "description": "Monitored ES cluster api key"
+                },
+                "elastic_cloud_connected_mode_api_key": {
+                    "type": "string",
+                    "description": "Cloud connected api key"
+                },
+                "elastic_cloud_connected_mode_api_url": {
```
pkoutsovasilis: we need also some validations here when the autoOps.enabled is set true I expect several of these properties not be empty

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
```diff
@@ -0,0 +1,125 @@
...
+    endpoint: ${env:AUTOOPS_OTEL_URL}
+
+service:
+  pipelines:
+    logs:
+      receivers: [metricbeatreceiver]
+      exporters: [otlphttp]
+  telemetry:
+    logs:
+      encoding: json
```
pkoutsovasilis, line 58: `ELASTICSEARCH_READ_USERNAME`, `ELASTICSEARCH_READ_PASSWORD` are not used here?

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
```diff
@@ -0,0 +1,125 @@
+{{- define "elasticagent.autoops.init" -}}
```
pkoutsovasilis, line 1: please add the autoops team as the owner of this folder `deploy/helm/elastic-agent/templates/integrations/_auto_ops` inside .github/CODEOWNERS

pkoutsovasilis (commented):
This is shaping really nice, I left some comments

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
idanl21: same answer as here - https://github.com/elastic/elastic-agent/pull/9363#discussion_r2280776770

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
idanl21: The issue is that ELASTICSEARCH_READ_PASSWORD, ELASTICSEARCH_READ_USERNAME, and ELASTICSEARCH_READ_API_KEY are already predefined environment variables. From what I understand, the agent code itself handles all of the authentication configuration, which means we don’t have the flexibility to rename these variables here. I’m not entirely sure why it was designed this way (I believe @pickypg originally set this up), but it seems intentional.

@pickypg, could you help us understand the reasoning behind this approach?

deploy/helm/elastic-agent/values.schema.json
idanl21: Hey, sorry, but I'm not sure how you want it to look. Could you please try to explain it better, or perhaps send an example?

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
pkoutsovasilis: I don't see how or where these env vars are "predefined" but maybe I am missing something, I am gonna wait for @pickypg to chime in

deploy/helm/elastic-agent/values.schema.json
pkoutsovasilis: Here is an example https://github.com/elastic/elastic-agent/blob/5b3506b667c9713dd23a58b4221bfcb9bab047cb/deploy/helm/elastic-agent/values.schema.json#L448-L457 and here is an official link that explains more about conditional schema validation https://json-schema.org/understanding-json-schema/reference/conditionals

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
pickypg: These keys are provided via the installation wizard that sets up the configuration and usage of it, but they are **not** used in YAML because the Elastic Agent running in OTel mode providing environment variables **cannot** provide blank values (the config will error out if the environment variable is not provided if you try to set it via the YAML), it simply provides them to the code. As an example:

```yaml
username: ${env:ELASTICSEARCH_READ_USERNAME}
password: ${env:ELASTICSEARCH_READ_PASSWORD}
api_key: ${env:ELASTICSEARCH_READ_API_KEY}
```

Since you inherently cannot provide both methods, one set or the other will be blank and when it is blank, then it actually causes an error because it's interpreted incorrectly when left blank. (https://github.com/elastic/beats/issues/45102)

We can rename these variables, but that inherently requires a code release. Those keys are used by Stack Monitoring, so they are not specific to AutoOps (we share the client code with Stack Monitoring, so we inherit those variables).

pkoutsovasilis: I am not sure I follow, the only config that will be loaded by elastic-agent through this PR when autoops integration is enabled is this one https://github.com/elastic/elastic-agent/pull/9363/files#diff-aff807a45ee574741f262afd724dcdc5963709cf8a8f3332c836e6daa6fcccebR1. I don't see how these env vars being used anywhere in this config, so maybe we need extra config there? Happy to have a sync about that

pickypg: Helm / K8s (and other supported mechanisms, like Docker) will pass in those fields and the _code_ picks up their values. That's why you don't see it used in the config itself because of the OTel mode bug (https://github.com/elastic/beats/issues/45102), they cannot be specified if they may be end up unused (and since we support two different forms of auth, inherently one or the other won't be specified).

https://github.com/elastic/beats/blob/a683d8bdfb09f3fb9d0b010eb09b077fcf377b2f/metricbeat/module/elasticsearch/metricset.go#L37-L38

pkoutsovasilis: I wasn't aware of that issue in the code. If I follow [your PR](https://github.com/elastic/beats/pull/45092) that introduced support for defining these env vars @pickypg 

> Try setting all three at the same time (Metricbeat will fail to start because that is an invalid setup just like putting all three directly in config)

So here we need to add exclusively either `ELASTICSEARCH_READ_API_KEY` or `ELASTICSEARCH_READ_PASSWORD` and `ELASTICSEARCH_READ_USERNAME` because these ifs here don't do that

deploy/helm/elastic-agent/values.schema.json
idanl21: @pkoutsovasilis 
added, please take a look 

Thanks!

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
idanl21: Added it under values.schema.json

deploy/helm/elastic-agent/values.schema.json
pkoutsovasilis: I think that the validations are not correct @idanl21 . 

1. you check for key existence and not if the value is empty, I think that we care for the latter
2. because you include `es_password`, `es_password` and `es_api_key` inside the `values.yaml` essentially the `oneOf` validation you have constantly fails, so it needs some extra thinking

You can see the above by running the following:
```
helm template --debug elastic-agent ./deploy/helm/elastic-agent -n kube-system --set kube-state-metrics.enabled=false --set kubernetes.enabled=false --set autoOps.enabled=true --set autoOps.token="" --set autoOps.autoops_otel_url="" --set autoOps.autoops_temp_resource_id="" --set autoOps.es_api_key=""
```

To avoid having to manually check this during the PR review, I need you to add a new example inside `examples` folder that contains an `agent-values.yaml` which enables the autoops integration. Then run `mage helm:renderExamples`. If this succeeds at least we can be sure that the chart renders successfully with the values you added, and it also checked in the CI. Also include the rendered manifest in your PR.

PS: please let the resolving of the comments to the reviewer, having them unresolved is helpful in re-gainning context

idanl21: Hey :) 
changed a bit the schema, now if you provide non auth it faild, also if you provide both
you can test it using 
```helm template --debug elastic-agent ./deploy/helm/elastic-agent -n kube-system \
  --set kube-state-metrics.enabled=false \
  --set kubernetes.enabled=false \
  --set autoOps.enabled=true \
  --set-string autoOps.autoops_token="tok-123" \
  --set-string autoOps.autoops_otel_url="https://otel.example.com:4318" \
  --set-string autoOps.autoops_temp_resource_id="res-abc" \
  --set-string autoOps.es_api_key="API_KEY_123" \
  --set autoOps.es_username=null \
  --set autoOps.es_password=null
```
OR
```helm template --debug elastic-agent ./deploy/helm/elastic-agent -n kube-system \
  --set kube-state-metrics.enabled=false \
  --set kubernetes.enabled=false \
  --set autoOps.enabled=true \
  --set-string autoOps.autoops_token="tok-123" \
  --set-string autoOps.autoops_otel_url="https://otel.example.com:4318" \
  --set-string autoOps.autoops_temp_resource_id="res-abc" \
  --set-string autoOps.es_username="elastic" \
  --set-string autoOps.es_password="supersecret" \
  --set autoOps.es_api_key=null
```

deploy/helm/elastic-agent/values.yaml
```diff
@@ -300,6 +300,17 @@ kubernetes:
...
+  enabled: false
+  autoops_token: null
+  autoops_es_url: null
+  es_password: null
+  es_username: null
+  temp_resource_id: null
+  otel_url: null
+  es_api_key: null
+  cloud_connected_mode_api_key: null
+  cloud_connected_mode_api_url: null
```
pkoutsovasilis: do not do `null` values as these should always be a `string` and not a null object, if you want them to be here comment them out

deploy/helm/elastic-agent/values.schema.json
pkoutsovasilis: @idanl21 I have to insist, I need an example that enables this integration so this the rendering can be validated in the CI

idanl21: Added an example :)

deploy/helm/elastic-agent/examples/autoops-agent/README.md
```diff
@@ -0,0 +1,40 @@
+# Example: Managed by Fleet Elastic Agent
+
+In this example we deploy an Elastic Agent that is managed by [Fleet](https://www.elastic.co/guide/en/fleet/current/manage-agents-in-fleet.html).
+
+## Prerequisites:
+1. Follow [this guide](https://www.elastic.co/guide/en/fleet/current/install-fleet-managed-elastic-agent.html#elastic-agent-installation-steps) to set up an agent policy and enroll an agent to it. Do not download any binary, from the proposed enrollment command just extract the Fleet URL (`--url=$FLEET_URL`) and Enrollment token (`--enrollment-token=$FLEET_TOKEN`).
```
pkoutsovasilis: this doesn't seem correct, I am pretty sure that this integration isn't supported for fleet-managed elastic-agents

deploy/helm/elastic-agent/examples/autoops-agent/README.md
```diff
@@ -0,0 +1,40 @@
...
+    ```yaml
+    autoOps:
+    enabled: true
+    autoops_token: "REPLACE_ME_TOKEN" #Replace with your AutoOps Token (Provided by AutoOps installation wizard)
+    autoops_otel_url: "https://otel.example.com:4318" #Repalce with Otel collector URL (Provided by AutoOps installation wizard)
+    autoops_temp_resource_id: "RESOURCE_ID" # (Provided by AutoOps installation wizard)
+    es_api_key: "REPLACE_ME_API_KEY" #Replace with your ElasticSearch cluster's API_KEY
+    es_username: ""
+    es_password: ""
+    ```
```
pkoutsovasilis: why we need that since you have the the `helm install` command above?

deploy/helm/elastic-agent/examples/autoops-agent/autoops-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+agent:
+  presets:
+    autoOps:
+      mode: deployment
+      serviceAccount:
+        create: true
+      clusterRole:
+        create: false
+      nodeSelector:
+        kubernetes.io/os: linux
```
pkoutsovasilis: this is already built-in in the chart values.yaml you don't need it

deploy/helm/elastic-agent/examples/autoops-agent/autoops-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+autoOps:
+  enabled: true
+  autoops_token: "REPLACE_ME_TOKEN"
+  autoops_otel_url: "https://otel.example.com:4318"
+  autoops_temp_resource_id: "REPLACE_ME_RESOURCE_ID"
+  es_api_key: "REPLACE_ME_API_KEY"
+  es_username: ""
+  es_password: ""
```
pkoutsovasilis: it shouldn't be needed to define empty values as well

deploy/helm/elastic-agent/examples/autoops-agent/README.md
idanl21: Changed it

deploy/helm/elastic-agent/examples/autoops-agent/README.md
idanl21: Removed

deploy/helm/elastic-agent/examples/autoops-agent/autoops-values.yaml
idanl21: removed

deploy/helm/elastic-agent/examples/autoops-agent/rendered/manifest.yaml
```diff
@@ -0,0 +1,194 @@
...
+    app.kubernetes.io/instance: example
+    app.kubernetes.io/version: 9.2.0
+stringData:
+  autoops-token: "REPLACE_ME_TOKEN"
+  autoops-es-url: "https://ccm-test-eu-west-1.es.us-east-2.aws.elastic-cloud.com"
+  temp-resource-id: 
+  otel-url: 
+  es-api-key: "REPLACE_ME_API_KEY"
+  cloud-connected-mode-api-key: 
+  cloud-connected-mode-api-url:
```
pkoutsovasilis: I believe that these shouldn't be empty here @idanl21 ? take a look at the secret template

idanl21: Hey
Fixed the naming and it looks better now
@pkoutsovasilis

deploy/helm/elastic-agent/values.yaml
```diff
@@ -300,6 +300,17 @@ kubernetes:
...
+  enabled: false
+  autoops_token: "OEdumJLuS"
+  autoops_otel_url: "https://platform-metrics.es.us-central1.gcp.cloud.es.io"
+  autoops_temp_resource_id: "07f968d2-e75e-48c6-ssss-89f805aaab5cced"
+  es_api_key: ""
+  es_username: ""
+  es_password: ""
+  elastic_cloud_connected_mode_api_key: "essu_V0RoSlEwOUthME1kxbEplRFIxV0RWSk4zVkNkdz09AAAAAFh4v2g"
+  elastic_cloud_connected_mode_api_url: "API_CCM_URL"
+  autoops_es_url: "https://ccm-test-eu-west-1.es.us-east-2.aws.elastic-cloud.com"
```
pkoutsovasilis: these values being here feels wrong; to begin with these look like an real urls and token which is an oopsy, but more importantly having such values in values.yaml it means that these will be the default values for any user. I am fairly confident that this isn't what you want to do here.

deploy/helm/elastic-agent/templates/integrations/_auto_ops/secret.yaml
```diff
@@ -0,0 +1,33 @@
...
+  es-password: {{ $.Values.autoOps.es_password | quote }}
+  {{- end }}
+  autoops-es-url: {{ $.Values.autoOps.autoops_es_url | quote }}
+  temp-resource-id: {{ $.Values.autoOps.autoops_temp_resource_id | quote }}
+  otel-url: {{ $.Values.autoOps.autoops_otel_url | quote }}
+  {{- if $.Values.autoOps.es_api_key }}
+  es-api-key: {{ $.Values.autoOps.es_api_key | quote }}
+  {{- end }}
+  cloud-connected-mode-api-key: {{ $.Values.autoOps.elastic_cloud_connected_mode_api_key | quote }}
+  cloud-connected-mode-api-url: {{ $.Values.autoOps.elastic_cloud_connected_mode_api_url | quote }}
```
pkoutsovasilis, line 37: since you allow empty strings for `autoops_es_url`, `elastic_cloud_connected_mode_api_key` and `elastic_cloud_connected_mode_api_url` these needed to be wrapped in if's

```suggestion
  autoops-token: {{ $.Values.autoOps.autoops_token | quote }}
  {{- if and $.Values.autoOps.es_username $.Values.autoOps.es_password }}
  es-username: {{ $.Values.autoOps.es_username | quote }}
  es-password: {{ $.Values.autoOps.es_password | quote }}
  {{- end }}
  {{- if $.Values.autoOps.autoops_es_url }}
  autoops-es-url: {{ $.Values.autoOps.autoops_es_url | quote }}
  {{- end }}
  temp-resource-id: {{ $.Values.autoOps.autoops_temp_resource_id | quote }}
  otel-url: {{ $.Values.autoOps.autoops_otel_url | quote }}
  {{- if $.Values.autoOps.es_api_key }}
  es-api-key: {{ $.Values.autoOps.es_api_key | quote }}
  {{- end }}
  {{- if $.Values.autoOps.elastic_cloud_connected_mode_api_key }}
  cloud-connected-mode-api-key: {{ $.Values.autoOps.elastic_cloud_connected_mode_api_key | quote }}
  {{- end }}
  {{- if $.Values.autoOps.elastic_cloud_connected_mode_api_url }}
  cloud-connected-mode-api-url: {{ $.Values.autoOps.elastic_cloud_connected_mode_api_url | quote }}
  {{- end }}
```

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
```diff
@@ -0,0 +1,125 @@
...
+    valueFrom:
+      secretKeyRef:
+        name: {{ $agentName }}-autoops
+        key: otel-url
+
+  - name: AUTOOPS_ES_URL
+    valueFrom:
+      secretKeyRef:
+        name: {{ $agentName }}-autoops
+        key: autoops-es-url
```
pkoutsovasilis, line 89: if this is an empty string do not create the env var

```suggestion
  {{- if $.Values.autoOps.autoops_es_url }}
  - name: AUTOOPS_ES_URL
    valueFrom:
      secretKeyRef:
        name: {{ $agentName }}-autoops
        key: autoops-es-url
 {{- end }}
```

deploy/helm/elastic-agent/templates/integrations/_auto_ops/_auto_ops.tpl
```diff
@@ -0,0 +1,125 @@
...
+      secretKeyRef:
+        name: {{ $agentName }}-autoops
+        key: cloud-connected-mode-api-key
+
+  - name: ELASTIC_CLOUD_CONNECTED_MODE_API_URL
+    valueFrom:
+      secretKeyRef:
+        name: {{ $agentName }}-autoops
+        key: cloud-connected-mode-api-url
+        optional: true
```
pkoutsovasilis, line 129: similar to above

```suggestion
{{- if $.Values.autoOps.elastic_cloud_connected_mode_api_key }}
  - name: ELASTIC_CLOUD_CONNECTED_MODE_API_KEY
    valueFrom:
      secretKeyRef:
        name: {{ $agentName }}-autoops
        key: cloud-connected-mode-api-key
{{- end }}

{{- if $.Values.autoOps.elastic_cloud_connected_mode_api_url }}
  - name: ELASTIC_CLOUD_CONNECTED_MODE_API_URL
    valueFrom:
      secretKeyRef:
        name: {{ $agentName }}-autoops
        key: cloud-connected-mode-api-url
        optional: true
{{- end }}
```

deploy/helm/elastic-agent/values.yaml
idanl21: DO you prefer it to be blank ("") or with a meaningless examples ("AutoOps_Token") for example?

pkoutsovasilis: It’s not really a matter of preference — it depends entirely on how these values are consumed. Relaying this back to you: should these values always be set by the user for the chart to work, or are they truly optional?
* If they must be set, then the values.schema.json should be updated to mark them as required and not-empty so Helm can validate user input.
* If they are optional, then they should default to "" (empty string), since they are strings and the Helm templates should filter them out correctly when unset.

This way we avoid shipping meaningless tokens/URLs as defaults and keep the chart behaviour consistent.

idanl21: The AutoOps values are not optional and cannot have real defaults because they’re customer-specific.
To enforce this, I updated `values.schema.json` so that when `autoOps.enabled=true:`

the fields are required (in `then.required`), and

they must be non-empty (`minLength: 1 in then.allOf`).

This works, but it causes `mage helm:updateAgentVersion helm:renderExamples` to fail whenever the example values file keeps those fields as empty strings, because the schema correctly rejects empty values.
What can i do in this case?

pkoutsovasilis: I don't see what you see @idanl21 , at the HEAD of this PR (40fc0a7a88df56955ab7ca1bcad2f1b55e0ddc3a), I did make the values.yaml be like below
```yaml
autoOps:
  enabled: false
  autoops_token: ""
  autoops_otel_url: ""
  autoops_temp_resource_id: ""
  es_api_key: ""
  es_username: ""
  es_password: ""
  elastic_cloud_connected_mode_api_key: ""
  elastic_cloud_connected_mode_api_url: ""
  autoops_es_url: ""
```

and running `mage helm:updateAgentVersion helm:renderExamples` doesn't give me any error. 

There are still three params that I think you should mark as required with `"minLength": 1` in your schema, namely `autoops_es_url`, `elastic_cloud_connected_mode_api_key`, `elastic_cloud_connected_mode_api_url`

idanl21: hey @pkoutsovasilis 
CI is still falling here...
And i still can see the same errors when i run ` mage helm:updateAgentVersion helm:renderExamples`
can we please have a quick call to solve that?

pkoutsovasilis: hey @idanl21 I just pushed the files that made the CI fail and kinda show that `mage helm:updateAgentVersion helm:renderExamples` complete successfully. Now you may focus on the values.yaml changes that I believe you should do. @ebeahan please arrange a quick call between @idanl21 and a team member

ebeahan: @idanl21 did we get this sorted? Looks like @pkoutsovasilis pushed in 1ea7f0373830be52da2836dee2795ddb6aa51eeb passed CI?

If not - reach out and I'll set something up.

idanl21: Thanks for the help folks!
@ebeahan, yeah, it looks like it passed now, if @pkoutsovasilis is approving the Values file, I can test it again and after a successful deploy we can start and deploy

idanl21: Hey @ebeahan 
A kind reminder about that

ebeahan: @idanl21 I've asked other engineers (@ycombinator @swiatekm ) to move the review forward.

However, [this point](https://github.com/elastic/elastic-agent/pull/9363#discussion_r2351172336) made by @pkoutsovasilis still applies and needs to be resolved. 

> The AutoOps values are not optional and cannot have real defaults because they’re customer-specific.

How are users getting these values as part of onboarding AutoOps? As an example, EDOT collector includes a step to create secrets used by Helm in the next step:

```shell
kubectl create namespace opentelemetry-operator-system
kubectl create secret generic elastic-secret-otel \
  --namespace opentelemetry-operator-system \
  --from-literal=elastic_otlp_endpoint='https://some-endpoint:443' \
  --from-literal=elastic_api_key='abcd1234'
helm upgrade --install opentelemetry-kube-stack open-telemetry/opentelemetry-kube-stack \
  --namespace opentelemetry-operator-system \
  --values 'https://raw.githubusercontent.com/elastic/elastic-agent/refs/tags/v9.2.0/deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml' \
  --version '0.10.5'
  ```

Comment (idanl21) [reply deploy/helm/elastic-agent/values.yaml]: @ebeahan, @ycombinator and @swiatekm 
> How are users getting these values as part of onboarding AutoOps? As an example, EDOT collector includes a step to create secrets used by Helm in the next step:

The values are provided by our AutoOps UI installation wizard (I would love to show you that in a short conversation if you like). Basically, the secret includes all the necessary data, and the Agent's deployment consumes the details from the value.

We indeed cannot have default values, cause each customer will need to insert different values here.

Comment (swiatekm) [reply deploy/helm/elastic-agent/values.yaml]: My view of how this should work:
- If the values are required, they should be empty in the Helm Chart's values.yaml and validated appropriately in the schema. From what I can see, the latter is already the case.
- For your example, you can provide realistic-looking placeholders.

If a default value is something we don't want users to actually have in their configuration, it shouldn't be a default. Does that make sense?

Comment (idanl21) [reply deploy/helm/elastic-agent/values.yaml]: @swiatekm 
So i agree, but that's the main issue here
When I keep the values empty on `values.yaml`, the CI and `mage helm:updateAgentVersion helm:renderExamples`
are falling.
the error - 

``` 
Error: failed to lint helm chart for example deploy/helm/elastic-agent/examples/autoops-agent: - autoOps: Must validate "then" as "if" was valid
- autoOps.autoops_es_url: String length must be greater than or equal to 1
- autoOps: Must validate all the schemas (allOf)

values don't meet the specifications of the schema(s) in the following chart(s):
elastic-agent:
- autoOps: Must validate "then" as "if" was valid
- autoOps.autoops_es_url: String length must be greater than or equal to 1
- autoOps: Must validate all the schemas (allOf)
```

Comment (swiatekm) [reply deploy/helm/elastic-agent/values.yaml]: https://github.com/elastic/elastic-agent/pull/9363/commits/e2ebe1fb943c96ee70616db4ddc544f46976e95a passes the checks locally for me, let's see if the CI agrees.

Comment (idanl21) [reply deploy/helm/elastic-agent/values.yaml]: Strange, for me https://github.com/elastic/elastic-agent/commit/e2ebe1fb943c96ee70616db4ddc544f46976e95a still falling locally 
But lets see, thanks!

Comment (swiatekm) [reply deploy/helm/elastic-agent/values.yaml]: @idanl21 can you check what version of helm you have installed?

Comment (idanl21) [reply deploy/helm/elastic-agent/values.yaml]: `version.BuildInfo{Version:"v3.13.3",` GitCommit:"c8b948945e52abba22ff885446a1486cb5fd3474", GitTreeState:"clean", `GoVersion:"go1.21.5"}`

Comment (idanl21) [reply deploy/helm/elastic-agent/values.yaml]: @swiatekm 
I see that it passed now!
What are the new steps needed here to start and test it on QA?

Comment (swiatekm) [reply deploy/helm/elastic-agent/values.yaml]: @idanl21 you can try updating your Helm version to at least 3.15. As for this PR, I'm fine approving it the way it is right now.

**Review by swiatekm — DISMISSED** [review]

File: deploy/helm/elastic-agent/values.yaml (hunk 11 above)
Comment (swiatekm) [reply deploy/helm/elastic-agent/values.yaml]: Once we merge, it will get published to the dev helm repo at https://helm-dev.elastic.co/ as version 9.3.0-beta, and you'll be able to use it that way.

**Review by ycombinator — APPROVED** [review]

**Review by ycombinator — DISMISSED** [review]
Could we please add a CHANGELOG fragment for this change?  See https://github.com/elastic/elastic-agent-changelog-tool/?tab=readme-ov-file#usage.  Thanks!

File: deploy/helm/elastic-agent/values.yaml (hunk 11 above)
Comment (idanl21) [reply deploy/helm/elastic-agent/values.yaml]: Great!
Thanks @swiatekm 
We're supposed to support it from version 7.15 (@pickypg @arnonst - let me know if I'm wrong). Should we merge it into every version-branch to start and support that?

**Comment by idanl21** [comment]
> Could we please add a CHANGELOG fragment for this change? See https://github.com/elastic/elastic-agent-changelog-tool/?tab=readme-ov-file#usage. Thanks!

Hey  @ycombinator :)
Added the CHANGELOG, Please take a look.
Thanks!

File: changelog/fragments/1762856188-added-opex-to-elastic-agent-helm-chart.yaml [hunk 15]
```diff
@@ -0,0 +1,45 @@
...
+
+# REQUIRED for breaking-change, deprecation, known-issue
+# impact:
+
+# REQUIRED for breaking-change, deprecation, known-issue
+# action:
+
+# REQUIRED for all kinds
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component:
```
ebeahan:
```suggestion
component: elastic-agent
```

changelog/fragments/1762856188-added-opex-to-elastic-agent-helm-chart.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+# AUTOMATED
+# OPTIONAL to manually add other PR URLs
+# PR URL: A link the PR that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+# pr: https://github.com/owner/repo/1234
```
ebeahan:
```suggestion
pr: https://github.com/elastic/elastic-agent/pull/9363
```

changelog/0.1.0.yaml
```diff
@@ -0,0 +1,70 @@
+version: 0.1.0
```
ebeahan: Please remove this file. You only need to create the `fragment`; rendering and creating the changelog is handled by maintainers for a release.

docs/release-notes/_snippets/0.1.0.asciidoc
```diff
@@ -0,0 +1,53 @@
+// begin 0.1.0 relnotes
```
ebeahan: Please remove this file. You only need to create the `fragment`; rendering and creating the changelog is handled by maintainers for a release.

ebeahan (changes requested):
Other than the yaml added to `changelog/fragments`. All the other release note files need removed.

You only need to create the fragment for your PR; rendering and creating the changelog is handled by maintainers for a release.

idanl21:
@ebeahan 
Thanks for the review Eric!
Addressed all your comments. Please let me know if there is anything else missing.

ebeahan (approved):
Removing my request for changes.

@pickypg can you make on final review for the AutoOps side?

ebeahan (changes requested):
@idanl21 we still need to clean up all the unnecessary *.md files the changelog tool added.

@pickypg can you also review again from the AutoOps side?

idanl21:
@ebeahan 
do you mean to the files that sits under docs/release-notes/_snippets/* ?

ebeahan:
> do you mean to the files that sits under docs/release-notes/_snippets/* ?

Please remove:

* `docs/release-notes/_snippets`
* `changelog/0.1.0.yaml`

idanl21:
@ebeahan 
Removed the unnecessary files

swiatekm (approved):

idanl21:
@ebeahan @swiatekm 

Any idea why the pipeline failed here?

pierrehilbert:
Hey @idanl21 
Sorry for that, this failure is unrelated to your change and currently investigated by the team.

ycombinator (approved):

theletterf (approved):
Approving to unblock.

pierrehilbert (approved):

PR #9366, opened 2025-08-14
Enhancement/5235 wrap errors when marking upgrade

kaanyalti:
**PR 4/6**

- Enhancement

## What does this PR do?

- Modifies the mark upgrade step to wrap errors so that the errors are propagated correctly.
- Adds relevant tests, asserts that errors get propagated up the call chain.

## Why is it important?

Errors are not propagated up the call chain in some instances, this prevents checking for error types.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~~
- ~~[ ] I have added an integration test or an E2E test~~

## Disruptive User Impact

none

## How to test this PR locally

Run the ugprade_test and step_unpack_test tests

## Related issues

- Relates #5235 
- Requires #9349 
- Prerequisite for #9386

PR #9386, opened 2025-08-15
Enhancement/5235 upgrade cleans up downloads and extracted agent

kaanyalti:
**PR 5/6**

- Enhancement

## What does this PR do?

- Updates the download step to return path after download succeeds even in error cases
- Adds deferred cleanup process in the upgrade function to cleanup download and extracted agent versioned home paths
- Updates error handling sequence in the upgrade function to add download and versioned home paths in the cleanup list
- Adds relevant tests

## Why is it important?

- Currently http and fs downloaders cleanup if they fail however if there is any other error (verifier errors) in the download process, the agent does not cleanup the downloaded artifacts
- The upgrade process reverts the symlink changes and cleans up the new versioned home only after the symlink switch happens. If anything in between the downloaders and symlink switch fails, the agent does not clean up anything.

## Checklist

## Disruptive User Impact

none

- Run the upgrade_test tests
- Build the agent for windows, mac and linux
- Try to upgrade from one version to another without sha512 and force verification failure and validate that the downloads directory is cleaned up.
- The way I triggered the second cleanup scenario was by filling up the disk space just enough so that the agent can partially extract the downloaded archive but fails midway. Then I validated that the versioned home for the new agent is removed along with the downloads in the initial agent's versioned home directory.

## Related issues

- Relates #5235 
- Requires #9366 
- Prerequisite for #9392


kaanyalti:
@pkoutsovasilis @michel-laterman the pr is ready for review

internal/pkg/agent/application/upgrade/step_download.go
```diff
@@ -169,6 +172,8 @@ func newDownloader(version *agtversion.ParsedSemVer, log *logger.Logger, setting
 	return composed.NewDownloader(fs.NewDownloader(settings), snapDownloader, httpDownloader), nil
 }
 
+var newVerifierFunc = newVerifier // abstraction for testability
```
michalpristas: why this way and not as a property of an object, in case you have 2 upgraders (you don't atm) this is shared between them

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -238,6 +252,14 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	}
 
 	archivePath, err := u.downloadArtifact(ctx, parsedVersion, sourceURI, det, skipVerifyOverride, skipDefaultPgp, pgpBytes...)
+
+	// If the artifactPath is not empty, then the artifact was downloaded.
+	// There may still be an error in the download process, so we need to add
+	// downloads path to the cleanup slice.
+	if archivePath != "" {
+		cleanupPaths = append(cleanupPaths, paths.Downloads())
```
michalpristas: ?? why are we checking archive path when appending download path, also when making os.RemoveAll on cleanup path where `paths.Download` is present, we need to make sure that during download missing download path is not problematic

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -273,6 +295,20 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	}
 	u.log.Debugf("detected used flavor: %q", detectedFlavor)
 	unpackRes, err := u.unpack(version, archivePath, paths.Data(), detectedFlavor)
+
+	// If VersionedHome is empty then unpack has not started unpacking the
+	// archive yet. There's nothing to clean up. Return the error.
+	if unpackRes.VersionedHome == "" {
```
michalpristas, line 388: @pchila is still field always present? i recall it can be empty in some cases, but maybe it's just something else i had in mind

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -515,6 +545,8 @@ func rollbackInstall(ctx context.Context, log *logger.Logger, topDirPath, versio
 	return nil
 }
 
+var copyActionStoreFunc = copyActionStore // abstraction for testability
```
michalpristas: same as with previous occurrence
i don't know if this is just a matter of my personal taste, but i don't like this pattern

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: VersionedHome can be empty. The unzip and untar functions can error out before setting VersionedHome

- https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/application/upgrade/step_unpack.go#L81-L107
- https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/application/upgrade/step_unpack.go#L316-L340

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: > "why are we checking archive path when appending download path"

The archive path is empty if either fs or http downloader fails. If these downloaders fail they clean up after themselves, so there is nothing to clean up. I don't have to check for the archive path and add the downloads dir no matter what. I wanted to be more prescriptive by adding this check


> "also when making os.RemoveAll on cleanup path where paths.Download is present, we need to make sure that during download missing download path is not problematic"

I will update this with specific file paths. TargetPath vs Downloads confused me, specifically because the upgrade function removes the downloads path when the upgrade is successful. I will post a question in the team channel to clarify my understanding here. To use specific file paths I have to keep the artifact path check.

internal/pkg/agent/application/upgrade/step_download.go
kaanyalti: After discussing with @ycombinator, I had implemented abstractions for stdlib functions this way. Since this function and copyActionStoreFunc (_and copyRunDirectory, but it is not in this PR_) are called directly from other functions, I decided to approach these the same way I did with stdlib functions. I did think of putting them as struct properties, but then the upgrader struct would get cluttered. The upgrader struct doesn't necessarily need these functions as properties. I also thought of adding receivers to these functions and using interfaces to abstract them, but decided to avoid it for the sake of keeping changes minimal.

That being said I'm open to changing how these functions are abstracted.

ycombinator: Yeah, I was in favor of a minimal abstraction until we had a use case for something more.

kaanyalti:
Last commit has issues, looking into it

internal/pkg/agent/application/upgrade/upgrade.go
pchila: @michalpristas 
If the tar or zip archive is valid VersionedHome is always present (either from `manifest.yaml` or generated as `elastic-agent-<hash>` for backward compatibility in case the package is an older one).

internal/pkg/agent/application/upgrade/upgrade_test.go
```diff
@@ -1748,3 +1809,23 @@ func TestCopyRunDirectory(t *testing.T) {
 		})
 	}
 }
+
+type mockSender struct{}
```
kaanyalti, line 1826: Moved client tests to the bottom of the test file, not a necessary change but it made more sense to me to have this test at the bottom, can move it back up.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -293,6 +302,15 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, sourceURI string
 	}
 
 	archivePath, err := u.artifactDownloader.downloadArtifact(ctx, parsedVersion, sourceURI, det, skipVerifyOverride, skipDefaultPgp, pgpBytes...)
+
+	// If the artifactPath is not empty, then the artifact was downloaded.
+	// There may still be an error in the download process, so we need to add
+	// the archive and hash path to the cleanup slice.
+	if archivePath != "" {
+		archiveHashPath := archivePath + ".sha512"
```
ycombinator: In `internal/pkg/agent/application/upgrade/artifact/download/verifier.go`, there's a function named [`getHashFileName`](https://github.com/elastic/elastic-agent/blob/a77d4cdf7a4104eccf23d59743e51adbb4891803/internal/pkg/agent/application/upgrade/artifact/download/verifier.go#L124).  WDYT about using it here so we're consistent in deciding how the name of the hash file is generated? At minimum, you'll need to export that function so it's public.

michel-laterman: 👍

michel-laterman (approved):
I think it looks good, but I would suggest looking into what @ycombinator suggested

ycombinator (commented):
Just a minor suggestion, then LGTM.

internal/pkg/agent/application/upgrade/upgrade.go
kaanyalti: Very good idea indeed. Thank you for pointing it out. I had just looked at the implementation in the downloaders and had done the same. 

[http downloader example](https://github.com/elastic/elastic-agent/blob/a77d4cdf7a4104eccf23d59743e51adbb4891803/internal/pkg/agent/application/upgrade/artifact/download/http/downloader.go#L165-L180)

[fs downloader example](https://github.com/elastic/elastic-agent/blob/a77d4cdf7a4104eccf23d59743e51adbb4891803/internal/pkg/agent/application/upgrade/artifact/download/fs/downloader.go#L70)

Will update this right away

kaanyalti: updated the name of the function to `AddHashExtension` and added it here instead of hard coded string

internal/pkg/agent/application/upgrade/artifact/download/verifier.go
```diff
@@ -121,20 +121,20 @@ func VerifySHA512HashWithCleanup(log infoWarnLogger, filename string) error {
 	return nil
 }
 
-func getHashFileName(filename string) string {
+func AddHashExtension(file string) string {
```
ycombinator, line 124: Maybe a nit, but I'd prefer if we named this as `GetHashFileName`.  

Rationale: `AddHashExtension` betrays the implementation detail — it exposes what this function is doing, while `GetHashFileName` abstracts away the implementation details.  Tomorrow, if we make the function do more than just add an extension to derive the hash file name, we'll need to change the name if we named it `AddHashExtension` today.  Whereas, naming the function `GetHashFileName` focusses on _what_ the function does, not _how_ it does it and gives us the freedom to change the implementation in the future without having to rename the function.

kaanyalti: I changed the name because I'm not using it to get a file name. I'm using it to add hash extension to a file path. I don't want the code to imply that the string I'm passing into this function is a file name, it would be confusing.

kaanyalti: In order to use the function the way it's name suggests `GetHashFileName`, I would have to get the file name from the path, call it with the function to add the extension and then join it back with the base path, which I think is overkill. If the function ends up doing more than adding a hash extension we can change the name again when it is necessary.

kaanyalti:
@ycombinator @michel-laterman I will merge changes from target to address ci failures, and will request reviews again, sorry for the inconvenience. I put a post about this in the team channel

PR #9462, opened 2025-08-19
fix: endpoint with tamper protection deb upgrade

pkoutsovasilis:
## What does this PR do?

This PR fixes upgrade and reinstall failures from a deb package involving Elastic Agent and Endpoint by ensuring that the `elastic-agent` service is explicitly stopped before we attempt to stop the `endpoint` service or remove its vault directory.  

Specifically:
- Updated the `preinstall.sh` template so that if `elastic-agent` is running, it is stopped before interacting with Endpoint.
- Re-enabled the integration test for installing the same version over an already installed agent, which was previously skipped due to constant failures.  

PS: thanks to @gabriellandau for pointing out the existence of such an interference

## Why is it important?

Without this change, the `elastic-agent` process could continue to invoke Endpoint’s `verify` logic in the background during package upgrades.  
This race condition allowed Endpoint to restart right after being stopped, which recreated the vault directory and led to uninstall/upgrade failures (exit code 28).  

By explicitly stopping `elastic-agent` before managing Endpoint, we eliminate these conflicts and make upgrades deterministic and reliable.  
This restores passing CI for upgrade and reinstall tests across multiple version ranges (e.g. [9.1.2→9.2.0](https://github.com/elastic/elastic-agent/pull/9450), [9.0.5→9.1.2](https://github.com/elastic/elastic-agent/pull/9449), [8.18.5→8.19.1](https://github.com/elastic/elastic-agent/pull/9447)).  

## Checklist

## Disruptive User Impact

No disruptive impact is expected.  
The change only affects package preinstall scripts, ensuring the agent is stopped before managing the Endpoint service.  
Users upgrading Elastic Agent will benefit from more reliable upgrades without needing to take manual action.  

You can either run the respective integration tests or
1. Install `9.1.2` version of Elastic Agent through deb, enroll it to Fleet and install Defend integration.
2. Upgrade Elastic-agent through deb to `9.2.0` version
3. Verify that the upgrade proceeds without Endpoint uninstall errors (exit code 28).

## Related issues

- N/A

### DIFF
--- a/dev-tools/packaging/templates/linux/preinstall.sh.tmpl
+++ b/dev-tools/packaging/templates/linux/preinstall.sh.tmpl
@@ -2,6 +2,14 @@
 
 set -e
 
+# we need to stop the agent to have all components stopped as well
+if systemctl --quiet is-active elastic-agent; then
+  echo "elastic-agent is running, stopping it"
+  systemctl --quiet stop elastic-agent
+else
+  echo "elastic-agent is not running"
+fi
+
 SERVICE_NAME="ElasticEndpoint"
 
 # Check if the endpoint service is installed
@@ -10,13 +18,6 @@ if systemctl list-unit-files --type=service | grep -q "^${SERVICE_NAME}.service"
 
     # Remove the vault directory if it exists
     if [ -d "/opt/Elastic/Endpoint/state/vault" ]; then
-        # we need to stop the agent to prevent it from invoking endpoint
-        if systemctl --quiet is-active elastic-agent; then
-          echo "elastic-agent is running, stopping it"
-          systemctl --quiet stop elastic-agent
-        else
-          echo "elastic-agent is not running"
-        fi
 
         # Check if the endpoint is running
         if systemctl --quiet is-active $SERVICE_NAME; then


dev-tools/packaging/templates/linux/preinstall.sh.tmpl
```diff
@@ -10,6 +10,13 @@ if systemctl list-unit-files --type=service | grep -q "^${SERVICE_NAME}.service"
 
     # Remove the vault directory if it exists
     if [ -d "/opt/Elastic/Endpoint/state/vault" ]; then
+        # we need to stop the agent to prevent it from invoking endpoint
+        if systemctl --quiet is-active elastic-agent; then
+          echo "elastic-agent is running, stopping it"
+          systemctl --quiet stop elastic-agent
+        else
+          echo "elastic-agent is not running"
+        fi
```
blakerouse: I think it would be better to just always have this and not only do it when endpoint vault exists.

blakerouse (commented):
Well sleuthed, glad this is a simple fix after all. See my inline comment about just doing this all the time not only when the vault exists.

dev-tools/packaging/templates/linux/preinstall.sh.tmpl
pkoutsovasilis: I totally agree, changing it

pkoutsovasilis: changed it in https://github.com/elastic/elastic-agent/pull/9462/commits/25d77346c66a90bd4c69e81059fdb82f6f0d1257

blakerouse (approved):
Looks good. Mirror's self-upgrade as well, all components are stopped before the upgrade occurs.

PR #9560, opened 2025-08-25
Fix/5871 redact secrets in diagnostics

kaanyalti:
- Bug

## What does this PR do?

Updates diagnostics redaction process to redact secrets from pre-config, computed-config, components-expected and components-actual files in the diagnostics archive

## Why is it important?

This pr increases the scope of the redaction to all files where they may be exposed

## Checklist

## Disruptive User Impact

none

- Build the agent
- Create policy with integrations that have secrets, I used aws-cloudwatch
- Enroll the agent
- Request diagnostics
- Validate that the secrets are redacted

## Questions to ask yourself

- ...


kaanyalti:
PR is ready for review
@pchila @ycombinator

kaanyalti:
Addressing the ci failures

kaanyalti:
Addressed the ci failures

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -356,22 +357,47 @@ func writeRedacted(errOut, resultWriter io.Writer, fullFilePath string, fileResu
...
+func redactMap[K comparable](errOut io.Writer, inputMap map[K]interface{}, sliceElem bool) map[K]interface{} {
 	if inputMap == nil {
 		return nil
 	}
+
+	redactionMarkers := []string{}
 	for rootKey, rootValue := range inputMap {
+		if keyString, ok := any(rootKey).(string); ok {
+			// Find siblings that have the redaction marker.
+			if strings.Contains(keyString, "mark_redact_") {
```
blakerouse: Please use a constant for the `"mark_redact_"`. Could you also use `"__mark_redact_"`? I just feel like its less likely to cause a conflict, we don't want a `"mark_redact"` key to cause a conflict... never probably going to happen but paranoia is better!

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -383,10 +409,20 @@ func redactMap[K comparable](errOut io.Writer, inputMap map[K]interface{}) map[K
 
 			}
 		}
-
 		inputMap[rootKey] = rootValue
+	}
 
+	for _, redactionMarker := range redactionMarkers {
+		keyToRedact := strings.TrimPrefix(redactionMarker, "mark_redact_")
```
blakerouse: Use the constant here.

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -356,22 +357,47 @@ func writeRedacted(errOut, resultWriter io.Writer, fullFilePath string, fileResu
...
 	if inputMap == nil {
 		return nil
 	}
+
+	redactionMarkers := []string{}
 	for rootKey, rootValue := range inputMap {
+		if keyString, ok := any(rootKey).(string); ok {
+			// Find siblings that have the redaction marker.
+			if strings.Contains(keyString, "mark_redact_") {
+				redactionMarkers = append(redactionMarkers, keyString)
```
blakerouse: Should this call `continue`? Does it need to run the other code, the `rootValue` would always be a boolean correct? Maybe adding that check here, to ensure that its a true marker would be good.

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -383,10 +409,20 @@ func redactMap[K comparable](errOut io.Writer, inputMap map[K]interface{}) map[K
...
+		for rootKey := range inputMap {
+			if keyString, ok := any(rootKey).(string); ok {
+				if keyString == keyToRedact {
+					inputMap[rootKey] = REDACTED
+				}
+			}
+		}
 	}
+
 	return inputMap
```
blakerouse, line 434: I think it would be best to also remove the `mark_redact_` keys from the outputs as well, that way its a clean mapping and doesn't have `mark_redact_` littered about it.

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -588,44 +624,75 @@ func saveLogs(name string, logPath string, zw *zip.Writer) error {
...
+
+	res := []string{}
+	if err := secretPaths.Unpack(&res); err != nil {
+		return nil, fmt.Errorf("failed to unpack secret_paths: %w", err)
+	}
+
+	return res, nil
+}
+
+const redactionMarkerPrefix = "mark_redact_"
```
blakerouse: You have a constant here, you should use it above as well.

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -588,44 +624,75 @@ func saveLogs(name string, logPath string, zw *zip.Writer) error {
...
-		} else {
-			fmt.Fprintf(errOut, "Unable to find secret path %q for redaction.\n", key)
+		lastPathSep := strings.LastIndex(sp, ".")
+		parentPath := sp[:lastPathSep]
+		keyName := sp[lastPathSep+1:]
+
+		secretKeyName := redactionMarkerPrefix + keyName
+		secretKeyPath := parentPath + "." + secretKeyName
+
+		if err := cfg.Agent.SetBool(secretKeyPath, -1, true, ucfg.PathSep(".")); err != nil {
```
blakerouse, line 697: Nice!

blakerouse (changes requested):
Overall this is close, a few comments.

internal/pkg/diagnostics/diagnostics.go
kaanyalti: You're right, thanks for pointing it out, if the key is the marker then we can just add it to the list and continue.

kaanyalti: I can also check for the bool value.

internal/pkg/diagnostics/diagnostics.go
kaanyalti: Done

internal/pkg/diagnostics/diagnostics.go
kaanyalti: Updated this section. Moved the check into the switch block.

kaanyalti:
Addressed the comments @blakerouse, ready for review

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -352,29 +353,58 @@ func writeRedacted(errOut, resultWriter io.Writer, fullFilePath string, fileResu
...
 			switch cast := rootValue.(type) {
 			case map[string]interface{}:
-				rootValue = redactMap(errOut, cast)
+				rootValue = redactMap(errOut, cast, sliceElem)
 			case map[interface{}]interface{}:
-				rootValue = redactMap(errOut, cast)
+				rootValue = redactMap(errOut, cast, sliceElem)
 			case map[int]interface{}:
-				rootValue = redactMap(errOut, cast)
+				rootValue = redactMap(errOut, cast, sliceElem)
```
ycombinator, line 376: Nit: you could collapse these three cases into one since they're doing the same thing.

```
case map[string]interface{}, map[interface{}]interface{}, map[int]interface{}:
```

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -352,29 +353,58 @@ func writeRedacted(errOut, resultWriter io.Writer, fullFilePath string, fileResu
...
+				// reach redaction markers that are inside slice items. Set SliceElem to true
+				// to avoid global redaction of slice elements.
+				for i, value := range cast {
+					switch m := value.(type) {
+					case map[string]interface{}:
+						cast[i] = redactMap(errOut, m, true)
+					case map[interface{}]interface{}:
+						cast[i] = redactMap(errOut, m, true)
+					case map[int]interface{}:
+						cast[i] = redactMap(errOut, m, true)
```
ycombinator, line 389: Nit: you could collapse these three cases into one since they're doing the same thing.

```suggestion
					case map[string]interface{}, map[interface{}]interface{}, map[int]interface{}:
						cast[i] = redactMap(errOut, m, true)
```

internal/pkg/diagnostics/diagnostics.go
ycombinator: Never mind, this doesn't work for the `map[interface{}]interface{}` case.

internal/pkg/diagnostics/diagnostics_test.go
```diff
@@ -155,164 +157,212 @@ mapping:
...
+			err = writeRedacted(&errOut, &out, "testPath", file)
 			require.NoError(t, err)
 
-			t.Logf("Error output: %s", errOut.String())
-			assert.Equal(t, tc.expect, out.String())
+			var expect map[string]any
+			err = yaml.Unmarshal(out.Bytes(), &expect)
+			require.NoError(t, err, "failed to unmarshal output")
+
+			assert.Equal(t, tc.expect, expect, "output does not match expected")
```
ycombinator: Nit: use `actual` for the actual value.
```suggestion
			var actual map[string]any
			err = yaml.Unmarshal(out.Bytes(), &actual)
			require.NoError(t, err, "failed to unmarshal output")

			assert.Equal(t, tc.expect, actual, "output does not match expected")
```

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -357,6 +357,8 @@ type Coordinator struct {
 	// run a ticker that checks to see if we have a new PID.
 	componentPIDTicker         *time.Ticker
 	componentPidRequiresUpdate *atomic.Bool
+
+	secretMarkerFunc func(*logger.Logger, *config.Config) error
```
ycombinator, line 362: Maybe add a comment that this field is being introduced for testability?

internal/pkg/diagnostics/diagnostics_test.go
kaanyalti: Updated the var name

internal/pkg/agent/application/coordinator/coordinator.go
kaanyalti: Added comment

ycombinator (approved):
LGTM as well!

kaanyalti:
Buildkite test this

kaanyalti:
Note: Current ci issues are due to an ongoing buildkite incident

Issue #9434, opened 2025-08-18
Elastic-agent loses UPG_SCHEDULED status when restarted


pkoutsovasilis:
When the elastic-agent is in the UPG_SCHEDULED state and the service is restarted, the state is lost. After the restart, the agent no longer shows UPG_SCHEDULED, even though an upgrade was previously scheduled. This results in inconsistent upgrade workflows.

https://github.com/user-attachments/assets/15916f66-dae4-40f2-b248-5d4bcbef9131

For confirmed bugs, please report:
- Version: All
- Operating System: All

cmacknz:
Quoting Panos' explanation from Slack:

> it happens because we create empty upgrade details when we instantiate the coordinator and we never extract them from the action queue store. Also, a Dispatch won't be helpful here because there we derive the upgrade details only by looking at the actions the fleet gateway sends us (again not the stored ones) . In simple words we have a bug how we handle the state

lucabelluccini:
I've created https://support.elastic.dev/knowledge/view/8ad28e5f and mentioned the fix will be in 8.18.7, 8.19.4, 9.0.7, 9.1.4.
It seems we are not planning 8.17.x at the moment.

Issue #8778, opened 2025-07-02
Agents remains in UPG_SCHEDULED state when upgrade action is cancelled


juliaElastic:
Bug discovered in scale tests: https://github.com/elastic/ingest-dev/issues/5774

There is a testcase that cancels upgrade action, which leaves agents in UPG_SCHEDULED state. 
This causes the next testcase  auto upgrade agents fail as auto upgrade doesn't pick up agents with upgrade details in non-terminal state.

This can be reproduced by starting an upgrade action with a rollout duration (e.g. 10m) and then cancelling the action.
The agents are stuck in scheduled state.

amolnater-qasource:
Hi @juliaElastic 
Just an update that we had a similar issue reported here https://github.com/elastic/elastic-agent/issues/4293 where agent remains stuck on cancelling the schedule upgrade.

Could you please confirm if these are related?
Thanks!!

juliaElastic:
@amolnater-qasource Thanks for the link, it seems to be the same issue. cc @cmacknz @ebeahan

cmacknz:
https://github.com/elastic/elastic-agent/issues/4293#issuecomment-1955023892 was my original assessment of where this is going wrong. Closing that issue as a duplicate and using this one.

mbudge:
We've also got a few agents impacted by this. 

We'll stop using the scheduled upgrade feature until this is fixed.

Is this highlighted in the release notes?

Thanks!

cmacknz:
> Is this highlighted in the release notes?

Following up on the release notes, confirming the affected versions and working to find a work around. We have the internal support case for this as well now.

cmacknz:
The work around is to force the agent to upgrade via the /fleet/agents API:

```
curl --request POST \
  --url https://<KIBANA_HOST>/api/fleet/agents/<AGENT_ID>/upgrade \
  --user "<SUPERUSER_NAME>:<SUPERUSER_PASSWORD>" \
  --header 'Content-Type: application/json' \
  --header 'kbn-xsrf: as' \
  --data '{"version": "<VERSION>","force": true}'
```

mbudge:
> The work around is to force the agent to upgrade via the /fleet/agents API:
> 
> ```
> curl --request POST \
>   --url https://<KIBANA_HOST>/api/fleet/agents/<AGENT_ID>/upgrade \
>   --user "<SUPERUSER_NAME>:<SUPERUSER_PASSWORD>" \
[...]

PR #9562, opened 2025-08-25
fix: scheduled upgrade details state

pkoutsovasilis:
## What does this PR do?

This PR fixes how Elastic Agent persists and reports upgrade details across check-ins, restarts, cancellations, and expired scheduled upgrade actions.

Specifically:
- (https://github.com/elastic/elastic-agent/commit/95ab8574440caafe06d2a1d25a1660fe21db2c35) Ensures the `Coordinator` is initialized with the correct upgrade details, either from the upgrade marker file or queued Fleet actions. 
- (https://github.com/elastic/elastic-agent/commit/95ab8574440caafe06d2a1d25a1660fe21db2c35, https://github.com/elastic/elastic-agent/commit/cabe8c7690a20bf0b6059f3b21df92a5043ba291) Extends the `ActionDispatcher` logic to:
  - Deduplicate upgrade actions from the received fleetgateway actions by keeping only the first encountered upgrade in input order.
  - Correctly track when upgrade details need to be updated (new upgrade actions, cancellations, expirations, etc.).
  - Correctly handle upgrade action cancellations.
  - Correctly handle expired scheduled upgrade actions and keep persisting them until explicitly canceled or replaced by a newer upgrade action.
  - Correctly handle retried upgrade actions.
  - Correctly update upgrade details when an upgrade action is scheduled for a retry.
  - Correctly update the upgrade details to `nil` when an upgrade action is dispatched without an error.
- (https://github.com/elastic/elastic-agent/commit/a16a9f387ea00a98bd83c2506ff3504428cf616b) Refactors and extends related dispatcher tests to include an actual action queue and cover new flows (queueing, canceling, expiring).
- (https://github.com/elastic/elastic-agent/commit/3e7324e974609147223ff73ed3eaa0b905c03695) Adds new **integration tests** under `testing/integration/ess/scheduled_upgrade_details_test.go` to verify that scheduled upgrade details are:
  - Reported correctly when received.
  - Reported correctly when expired.
  - Preserved across Agent restarts.
  - Cleared on cancel.

All the new business logic is captured by https://github.com/elastic/elastic-agent/commit/95ab8574440caafe06d2a1d25a1660fe21db2c35 (+234 -126 lines changed) and https://github.com/elastic/elastic-agent/commit/cabe8c7690a20bf0b6059f3b21df92a5043ba291 (+174 -20 lines changed)

## Why is it important?

Previously, scheduled upgrade details could be lost or incorrectly reported:
- They were not consistently preserved across Agent restarts.
- Cancel actions did not clear the reported upgrade details.
- Expired scheduled upgrade actions were treated as ephemeral and not persisted.

These inconsistencies caused confusion in Fleet UI and sometimes left users unable to perform upgrades.  
With this fix, users now get a consistent and accurate view of upcoming, active, or failed upgrades across the entire Agent lifecycle.

## Checklist

## Disruptive User Impact

None expected. This change only improves correctness of reported upgrade details.  
Users upgrading to this version will see more consistent and reliable upgrade state reporting in Fleet.

1. Run the new integration tests
2. Or build the code of this PR
  * Provision a 9.1.2 cloud stack
  * `AGENT_PACKAGE_VERSION=9.1.1 EXTERNAL=true SNAPSHOT=true PLATFORMS="linux/arm64" PACKAGES="tar.gz" mage package` (ty @ycombinator for the proposal)
  * Install and enroll the Agent to Fleet
  * Schedule an upgrade → check it is reported.
  * Restart the Agent → check details persist.
  * Cancel the upgrade → check details clear.

https://github.com/user-attachments/assets/92545d1e-da61-4131-8760-04311657f47f

Also I tested the approach mentioned [here](https://github.com/elastic/elastic-agent/issues/8778#issuecomment-3133874202) to force update an agent and everything is reflected correctly to the upgrade details

https://github.com/user-attachments/assets/d7f85c50-e451-405a-a14e-fc9feaa3375f

(PS: For the needs of the videos above I compiled this PR with pseudo prior version 9.1.1)

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/8778
- Closes https://github.com/elastic/elastic-agent/issues/9434

### DIFF
--- a/internal/pkg/agent/application/dispatcher/dispatcher.go
+++ b/internal/pkg/agent/application/dispatcher/dispatcher.go
@@ -123,8 +123,8 @@ func (ad *ActionDispatcher) Dispatch(ctx context.Context, detailsSetter details.
 
 	upgradeDetailsNeedUpdate := false
 
-	// remove any upgrade actions from the queue if there is an upgrade action in the actions
-	ad.removeQueuedUpgrades(actions, &upgradeDetailsNeedUpdate)
+	// remove duplicate upgrade actions from fleetgateway actions and remove all upgrade actions in the queue
+	actions = ad.compactAndRemoveQueuedUpgrades(actions, &upgradeDetailsNeedUpdate)
 
 	// add any scheduled actions to the queue (we don't check the start time here, as we will check it later)
 	// and remove them from the passed actions
@@ -137,6 +137,7 @@ func (ad *ActionDispatcher) Dispatch(ctx context.Context, detailsSetter details.
 	actions = ad.mergeWithQueuedActions(now, actions, &upgradeDetailsNeedUpdate)
 
 	if upgradeDetailsNeedUpdate {
+		upgradeDetailsNeedUpdate = false
 		ad.updateUpgradeDetails(now, detailsSetter)
 	}
 
@@ -166,7 +167,7 @@ func (ad *ActionDispatcher) Dispatch(ctx context.Context, detailsSetter details.
 			rAction, ok := action.(fleetapi.RetryableAction)
 			if ok {
 				rAction.SetError(err) // set the retryable action error to what the dispatcher returned
-				ad.scheduleRetry(ctx, rAction, acker)
+				ad.scheduleRetry(ctx, rAction, acker, &upgradeDetailsNeedUpdate)
 				continue
 			}
 			ad.log.Errorf("Failed to dispatch action id %q of type %q, error: %+v", action.ID(), action.Type(), err)
@@ -178,6 +179,10 @@ func (ad *ActionDispatcher) Dispatch(ctx context.Context, detailsSetter details.
 
 	if err = acker.Commit(ctx); err != nil {
 		reportedErr = err
+	}
+
+	if upgradeDetailsNeedUpdate {
+		ad.updateUpgradeDetails(now, detailsSetter)
 	}
 
 	if len(actions) > 0 {
@@ -285,7 +290,17 @@ func (ad *ActionDispatcher) mergeWithQueuedActions(ts time.Time, actions []fleet
 	dequeuedActions := ad.queue.DequeueActions()
 
 	for _, action := range dequeuedActions {
-		exp, _ := action.Expiration()
+		exp, err := action.Expiration()
+		if err != nil {
+			if !errors.Is(err, fleetapi.ErrNoExpiration) {
+				// this is not a non-expiring scheduled action, e.g. there is a malformed expiration time set
+				ad.log.Warnf("failed to get expiration time for scheduled action [id = %s]: %v", action.ID(), err)
+				continue
+			}
+			// this is a non-expiring scheduled action
+			actions = append(actions, action)
+			continue
+		}
 		if ts.After(exp) {
 			if action.Type() == fleetapi.ActionTypeUpgrade {
 				// this is an expired upgrade action thus we need to recalculate the upgrade details
@@ -308,7 +323,7 @@ func (ad *ActionDispatcher) mergeWithQueuedActions(ts time.Time, actions []fleet
 	// Put expired upgrade actions back onto the queue to persist them across restarts.
 	// These are handled the same as non-expired upgrade scheduled actions and are only removed when:
 	// 1) a cancel action is received (look at dispatchCancelActions func) or
-	// 2) a newer upgrade action (scheduled or not) removes them. (look at removeQueuedUpgrades func)
+	// 2) a newer upgrade action (scheduled or not) removes them. (look at compactAndRemoveQueuedUpgrades func)
 	for _, expiredUpgradeAction := range expiredUpgradeActions {
 		startTime, err := expiredUpgradeAction.StartTime()
 		if err != nil {
@@ -320,31 +335,41 @@ func (ad *ActionDispatcher) mergeWithQueuedActions(ts time.Time, actions []fleet
 	}
 
 	// if an upgrade action is included in the immediate dispatchable actions
-	// mark upgradeDetailsNeedUpdate as false since the upgrade details will be set by it
+	// mark upgradeDetailsNeedUpdate as true
 	if slices.ContainsFunc(actions, func(action fleetapi.Action) bool {
 		return action.Type() == fleetapi.ActionTypeUpgrade
 	}) {
-		*upgradeDetailsNeedUpdate = false
+		*upgradeDetailsNeedUpdate = true
 	}
 
 	return actions
 }
 
-// removeQueuedUpgrades will scan the passed actions and if there is an upgrade action it will remove all upgrade actions
-// in the queue but not alter the passed list. This is done to try to only have the most recent upgrade action executed.
-// However, it does not eliminate duplicates in retrieved directly from the gateway. Also, if upgrade actions are removed
-// from the queue, upgradeDetailsNeedUpdate will be set to true.
-func (ad *ActionDispatcher) removeQueuedUpgrades(actions []fleetapi.Action, upgradeDetailsNeedUpdate *bool) {
-	for _, action := range actions {
+// compactAndRemoveQueuedUpgrades deduplicates *upgrade* actions from the given fleetgateway actions by keeping only the
+// first encountered upgrade in input order (dropping any subsequent upgrades from the returned slice). If an upgrade
+// action is found, it also removes any queued upgrade actions from the queue. When queued upgrade actions are removed,
+// upgradeDetailsNeedUpdate is set to true.
+func (ad *ActionDispatcher) compactAndRemoveQueuedUpgrades(input []fleetapi.Action, upgradeDetailsNeedUpdate *bool) []fleetapi.Action {
+	var actions []fleetapi.Action
+	var upgradeAction fleetapi.Action
+	for _, action := range input {
 		if action.Type() == fleetapi.ActionTypeUpgrade {
+			if upgradeAction == nil {
+				upgradeAction = action
+			} else {
+				ad.log.Warnf("Found extra upgrade action in fleetgateway actions [id = %s]", action.ID())
+				continue
+			}
 			if n := ad.queue.CancelType(fleetapi.ActionTypeUpgrade); n > 0 {
 				ad.log.Debugw("New upgrade action retrieved from gateway, removing queued upgrade actions", "actions_found", n)
 				// upgrade action(s) got removed from the queue so upgrade actions changed
 				*upgradeDetailsNeedUpdate = true
 			}
-			return
 		}
+		actions = append(actions, action)
 	}
+
+	return actions
 }
 
 // updateUpgradeDetails will construct the upgrade details based queue actions (assuming expired ones are still in the queue)
@@ -364,9 +389,11 @@ func (ad *ActionDispatcher) updateUpgradeDetails(ts time.Time, detailsSetter det
 }
 
 // scheduleRetry will schedule a retry for the passed action. Note that this adjusts the start time of the action
-// but doesn't affect expiration time.
-func (ad *ActionDispatcher) scheduleRetry(ctx context.Context, action fleetapi.RetryableAction, acker acker.Acker) {
+// but doesn't affect expiration time. If the action is scheduled to be retried and it is an upgrade action,
+// upgradeDetailsNeedUpdate will be set to true.
+func (ad *ActionDispatcher) scheduleRetry(ctx context.Context, action fleetapi.RetryableAction, acker acker.Acker, upgradeDetailsNeedUpdate *bool) {
 	attempt := action.RetryAttempt()
+	ad.log.Warnf("Re-scheduling action id %q of type %q, because it failed to dispatch with error: %+v", action.ID(), action.Type(), action.GetError())
 	d, err := ad.rt.GetWait(attempt)
 	if err != nil {
 		ad.log.Errorf("No more retries for action id %s: %v", action.ID(), err)
@@ -390,6 +417,11 @@ func (ad *ActionDispatcher) scheduleRetry(ctx context.Context, action fleetapi.R
 	if err != nil {
 		ad.log.Errorf("retry action id %s attempt %d failed to persist action_queue: %v", action.ID(), attempt, err)
 	}
+
+	if action.Type() == fleetapi.ActionTypeUpgrade {
+		*upgradeDetailsNeedUpdate = true
+	}
+
 	if err := acker.Ack(ctx, action); err != nil {
 		ad.log.Errorf("Unable to ack action retry (id %s) to fleet-server: %v", action.ID(), err)
 		return
@@ -458,5 +490,9 @@ func GetScheduledUpgradeDetails(log *logger.Logger, actions []fleetapi.Scheduled
 		details.StateScheduled,
 		nextUpgradeActionID)
 	upgradeDetails.Metadata.ScheduledAt = &nextUpgradeStartTime
+	// scheduled upgrade actions can have errors if retried
+	if err := nextUpgradeAction.GetError(); err != nil {
+		upgradeDetails.Metadata.ErrorMsg = fmt.Sprintf("A prior dispatch attempt failed with: %v", err)
+	}
 	return upgradeDetails
 }
--- a/internal/pkg/agent/application/dispatcher/dispatcher_test.go
+++ b/internal/pkg/agent/application/dispatcher/dispatcher_test.go
@@ -297,7 +297,7 @@ func TestActionDispatcher(t *testing.T) {
 
 		action1 := &mockScheduledAction{}
 		action1.On("StartTime").Return(time.Time{}, fleetapi.ErrNoStartTime)
-		action1.On("Expiration").Return(time.Now().Add(time.Hour), fleetapi.ErrNoStartTime)
+		action1.On("Expiration").Return(time.Now().Add(time.Hour), fleetapi.ErrNoExpiration)
 		action1.On("Type").Return(fleetapi.ActionTypeCancel)
 		action1.On("ID").Return("id")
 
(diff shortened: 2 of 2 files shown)


ycombinator:
Tested this PR with the given instructions.  I see the scheduled upgrade being reported both in the Fleet UI and in the output of `sudo elastic-agent status` as part of the `upgrade_details` key 👍.   

I restarted Elastic Agent three different ways after that: by running `sudo elastic-agent restart`, by stopping and starting the elastic-agent service, and by killing the elastic-agent process and letting the service bring it back up. In each case, the `UPG_SCHEDULED` state in `upgrade_details` remained the same as before 👍.

Before the scheduled time arrived, I canceled the upgrade via the Fleet UI. I see that the `UPG_SCHEDULED` state is no longer seen either in the Fleet UI or in the output of `sudo elastic-agent status` as part of the `upgrade_details` key 👍.

Reviewing the code changes now...

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -192,70 +203,165 @@ func detectTypes(actions []fleetapi.Action) []string {
...
+		actions = append(actions, action)
+	}
+
+	ad.log.Debugf("Gathered %d actions from queue, %d non-upgrade actions expired, %d upgrade actions expired",
+		len(dequeuedActions), len(expired), len(expiredUpgradeActions))
+	ad.log.Debugf("Expired non-upgrade actions: %v", expired)
+	ad.log.Debugf("Expired upgrade actions (won't be removed from the queue): %v", expiredUpgradeActions)
+
+	for _, expiredUpgradeAction := range expiredUpgradeActions {
+		startTime, err := expiredUpgradeAction.StartTime()
```
blakerouse, line 328: Can you add a comment here saying that at this point in the code there really should be a `StartTime` on the action, because these are expired upgrade actions. I just want to make warnings in the code clear, because before this change it was not clear.

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -192,70 +203,165 @@ func detectTypes(actions []fleetapi.Action) []string {
...
 }
 
+// updateUpgradeDetails will construct the upgrade details based queue actions (assuming expired ones are still in the queue)
+// and if ad.lastUpgradeDetails is different from the new upgrade details, it will update them.
+func (ad *ActionDispatcher) updateUpgradeDetails(ts time.Time, detailsSetter details.Observer) {
+	// no upgrade details from expired actions check the stored actions in the queue
+	upgradeDetails := GetScheduledUpgradeDetails(ad.log, ad.queue.Actions(), ts)
+
+	if ad.lastUpgradeDetailsIsSet && ad.lastUpgradeDetails.Equals(upgradeDetails) {
+		// calculate
```
blakerouse:
```suggestion
		// already the same details; do nothing
```

blakerouse (commented):
The code looks good, just a two nits that I found.

Overall the testing looks good. The integration testing is good. I do have a request for an extra integration test. I didn't see a test case for scheduling an upgrade action, the action running, it failing, then scheduling a new action not scheduled, that replaces that action with the same version and URI. I believe that could be added with an upgrade image behind a transparent proxy, where the first action always fails to download, and then the following up action it allows it to succeed.

pkoutsovasilis:
Thanks for the review @blakerouse - appreciate the close look!

On the extra integration test: In general I agree more coverage around upgrades is always valuable. That said the flow you’re describing kicks in after an upgrade action is dispatched. At that point the action leaves the dispatcher’s action queue and becomes part of bkgActions slice which is managed by the upgrader. So it isn’t exercising the dispatcher changes in this PR. I’d like to keep this PR scoped to "actions that go through the dispatcher" where the related issues are, and avoid adding an integration test that primarily validates coordinator behavior here.

I can, however, add a unit test in this PR that forces the upgrade action handler to fail initially, then schedules a new upgrade action (same version/URI) and verifies the dispatcher dispatches the new one - matching the spirit of your ask while staying within dispatcher boundaries. WDYT?

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -290,86 +396,64 @@ func (ad *ActionDispatcher) scheduleRetry(ctx context.Context, action fleetapi.R
...
-			}
-			ad.lastUpgradeDetails = details.NewDetails(version, details.StateFailed, e.ID())
-			ad.lastUpgradeDetails.Fail(fmt.Errorf("upgrade action %q expired on %s",
-				e.ID(), expiration))
-
-			upgradeDetailsSetter(ad.lastUpgradeDetails)
-		}
-	}
-}
+// GetScheduledUpgradeDetails returns the upgrade details of the next upgrade action, if any. It also adjusts accordingly
```
ycombinator:
```suggestion
// GetScheduledUpgradeDetails returns the upgrade details of the next scheduled upgrade action, if any. It also adjusts accordingly
```

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -290,86 +396,64 @@ func (ad *ActionDispatcher) scheduleRetry(ctx context.Context, action fleetapi.R
...
-		if nextUpgrade == nil {
-			nextUpgrade = uAction
-			continue
+		// not all scheduled upgrade actions have an expiration, e.g. a scheduled retried upgrade because it failed
+		// does not have an expiration
+		upgradeExpirationTime, err := upgradeAction.Expiration()
+		if err != nil {
+			if !errors.Is(err, fleetapi.ErrNoExpiration) {
+				// this is not a non-expiring upgrade action
+				log.Errorf("failed to get expiration time for scheduled upgrade action [id = %s]: %v", upgradeAction.ID(), err)
```
ycombinator, line 460: Considering not all scheduled upgrade actions are expected to have an expiration, should we make this a warning instead of an error?

ycombinator: Never mind, I got confused by the error message.  I see you are logging an error that occurred when getting the expiration time, but only when the action is _expected_ not to have an expiration.  The error is not that the action has no expiration time.

blakerouse:
@pkoutsovasilis A unit test for that would work.

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -192,70 +203,165 @@ func detectTypes(actions []fleetapi.Action) []string {
...
+
+		ad.log.Debugf("Adding action id: %s to queue.", sAction.ID())
+		ad.queue.Add(sAction, start.Unix())
+
+		if isUpgradeAction {
+			// there is an upgrade action added to the queue so we need to update the upgrade details
+			*upgradeDetailsNeedUpdate = true
+		}
+
+		continue
```
ycombinator: Nit: not sure you need this continue at the end of the loop?

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -192,70 +203,165 @@ func detectTypes(actions []fleetapi.Action) []string {
...
+// cancel actions are dispatched separately as they may remove items from the queue. If cancel actions remove upgrade actions
+// from the queue, upgradeDetailsNeedUpdate will be set to true.
+func (ad *ActionDispatcher) dispatchCancelActions(ctx context.Context, actions []fleetapi.Action, upgradeDetailsNeedUpdate *bool, acker acker.Acker) []fleetapi.Action {
+	queuedUpgradeActions := maps.Collect(func(yield func(K string, V fleetapi.Action) bool) {
+		for _, action := range ad.queue.Actions() {
+			if _, ok := action.(*fleetapi.ActionUpgrade); !ok {
+				continue
+			}
+
+			if !yield(action.ID(), action) {
```
ycombinator: Since you are only using the keys of `queuedUpgradeActions` for an existence check further down, there is no need for the value to the the `action` object.  You could make it a 0-byte value by setting it to `struct{}{}` here.  Of course, the type` of `V` in the `yield` function signature will need to become `struct{}` as well.

internal/pkg/agent/application/dispatcher/dispatcher.go
```diff
@@ -192,70 +203,165 @@ func detectTypes(actions []fleetapi.Action) []string {
...
+		}
+		actions = append(actions, action)
+	}
+
+	ad.log.Debugf("Gathered %d actions from queue, %d non-upgrade actions expired, %d upgrade actions expired",
+		len(dequeuedActions), len(expired), len(expiredUpgradeActions))
+	ad.log.Debugf("Expired non-upgrade actions: %v", expired)
+	ad.log.Debugf("Expired upgrade actions (won't be removed from the queue): %v", expiredUpgradeActions)
+
+	for _, expiredUpgradeAction := range expiredUpgradeActions {
```
ycombinator, line 327: Could you add a comment here explaining why we add back the expired upgrade actions to the queue?

ycombinator (approved):
Code LGTM.  Just left a few minor suggestions.

internal/pkg/agent/application/dispatcher/dispatcher.go
pkoutsovasilis: have a look here 6f75b29b772af12fe38091555902a056c14b9f35

internal/pkg/agent/application/dispatcher/dispatcher.go
pkoutsovasilis: changed in 6f75b29b772af12fe38091555902a056c14b9f35

internal/pkg/agent/application/dispatcher/dispatcher.go
pkoutsovasilis: have a look here 6f75b29b772af12fe38091555902a056c14b9f35

internal/pkg/agent/application/dispatcher/dispatcher.go
pkoutsovasilis: removed here 85f29443742f05a4e1644c793f89891f0232ba93

internal/pkg/agent/application/dispatcher/dispatcher.go
pkoutsovasilis: nice proposal, done in bd24f7929ea2c43230feb1bfc23ab739039c353c

internal/pkg/agent/application/dispatcher/dispatcher.go
pkoutsovasilis: take a look here 6f75b29b772af12fe38091555902a056c14b9f35

pkoutsovasilis:
@blakerouse @ycombinator while working on the unit test we agreed on, I ran into a few more issues (story of my life :bowtie:). That’s why I added a new commit (https://github.com/elastic/elastic-agent/commit/cabe8c7690a20bf0b6059f3b21df92a5043ba291) with additional business logic changes. I’ve updated the PR description, but to call them out specifically:

> Deduplicate upgrade actions from the received fleetgateway actions by keeping only the first encountered upgrade in input order.

We now keep only the first encountered upgrade action in input order. I’m not sure how/why fleetgateway can send multiple upgrade actions, but it causes problems. Example: if we get both an immediate upgrade and a scheduled upgrade, which one should set the upgrade details? And what if the scheduled one gets dispatched in the next window while the immediate one is still running? That feels like something that shouldn’t be allowed, but I’d like your thoughts @blakerouse @ycombinator 

> Correctly handle retried upgrade actions.

Previously ([code link](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/application/dispatcher/dispatcher.go#L236)), retried immediate upgrade actions (no expiration) were incorrectly treated as expired because `exp` was `0,` and `ts.After(exp)` was always true.

> Correctly update upgrade details when an upgrade action is scheduled for a retry.

The dispatcher now updates upgrade details on retry, including the error from the prior attempt. Unfortunately, Fleet UI still shows only the `UPG_SCHEDULED` tag without an error indicator (e.g. a warning/triangle), but the agent state correctly captures this - so I think this is fine.

> Correctly update the upgrade details to `nil` when an upgrade action is dispatched without an error.

Learned here that the coordinator upgrader can skip setting upgrade details (e.g. if the action is ignored because it’s the same version/URI). Without resetting, the last dispatcher-set upgrade details would “stick” to the agent and, since the action gets ack’ed, there’d be no cancellation path. Fix: whenever the dispatcher sees a dispatchable upgrade action, it sets upgrade details to `nil`. That way, if the upgrader doesn’t set new ones, we don’t end up with stale data. IMO this needs to be investigated properly and most probably change

---

Now, all that said, I think there’s another upcoming issue: in [PR #8767](https://github.com/elastic/elastic-agent/pull/8767), available rollback versions are becoming part of the upgrade details. That can create issues since the dispatcher knows nothing about which versions are available for rollback, so it could incorrectly set upgrade details with these missing. From what I understand now about how upgrade details flow through the code, rollback versions don’t really belong in upgrade details - they should probably be part of the check-in request body instead. Fleet UI could then combine those with upgrade details to decide whether to show a rollback option, similar to the existing flow that Fleet UI decides that there is an available upgrade for an agent.

@pchila @ebeahan let's discuss about that - maybe I am missing something

pchila:
> Now, all that said, I think there’s another upcoming issue: in [PR #8767](https://github.com/elastic/elastic-agent/pull/8767), available rollback versions are becoming part of the upgrade details. That can create issues since the dispatcher knows nothing about which versions are available for rollback, so it could incorrectly set upgrade details with these missing. From what I understand now about how upgrade details flow through the code, rollback versions don’t really belong in upgrade details - they should probably be part of the check-in request body instead. Fleet UI could then combine those with upgrade details to decide whether to show a rollback option, similar to the existing flow that Fleet UI decides that there is an available upgrade for an agent.

@pkoutsovasilis agreed, looking at the upgrade details' behavior that this PR is dealing with for scheduled upgrades and retries I feel like the list of available rollbacks should not be in the details metadata (that location was chosen initially as a "convenient" spot since that was already part of agent status and check-in payload).

Let's discuss offline with @ebeahan involvement how to avoid introducing more complexity on the upgrade details functionality.

internal/pkg/agent/application/dispatcher/dispatcher.go
blakerouse: Thanks!

blakerouse:
> @blakerouse @ycombinator while working on the unit test we agreed on, I ran into a few more issues (story of my life :bowtie:). That’s why I added a new commit ([cabe8c7](https://github.com/elastic/elastic-agent/commit/cabe8c7690a20bf0b6059f3b21df92a5043ba291)) with additional business logic changes. I’ve updated the PR description, but to call them out specifically:

@pkoutsovasilis Great that these extra issues where caught now, with added unit tests. The changes adding in that commit look good, very well commented.

> 
> > Deduplicate upgrade actions from the received fleetgateway actions by keeping only the first encountered upgrade in input order.
> 
> We now keep only the first encountered upgrade action in input order. I’m not sure how/why fleetgateway can send multiple upgrade actions, but it causes problems. Example: if we get both an immediate upgrade and a scheduled upgrade, which one should set the upgrade details? And what if the scheduled one gets dispatched in the next window while the immediate one is still running? That feels like something that shouldn’t be allowed, but I’d like your thoughts @blakerouse @ycombinator

I don't think there is really away of having Fleet prevent this. I could see two requests coming in near the same time one scheduled and the other not. To me the non-scheduled wins and should just happen, the scheduled one should just be ignored as the upgrade is now happening.

internal/pkg/agent/application/dispatcher/dispatcher.go
ycombinator: Thanks.  

Now I'm wondering why we only remove expired actions from the queue when we get a cancel or newer upgrade action removes them.  I understand we have to keep them in the queue to survive restarts so we can report that they expired in upgrade details but why not remove them immediately after that reporting? Why wait till we get a cancel or newer upgrade action?

Anyway, I'm not going to block the improvements made by this PR on this question; it's more for my understanding.  Depending on the answer, I think it would be helpful to document it in the code somewhere but, again, that's for a follow up PR.

pkoutsovasilis: no let's discuss this so all of us are in the same page and then we merge the fixes. As you already said, the reason these are added back in the queue and not removed, is to survive elastic-agent restarts; the next time agent start it will load the details from the marker or from the action queue. If expired actions are not there it will emit `nil` upgrade details so the expiration will be lost from the fleet UI (look [here](https://github.com/elastic/elastic-agent/pull/9562/files#diff-f9f401b9d676a38a417d02f0a706e625a7f7b9214865641f5f2c4ed23a2d1f19R226)). Does that explanation help?

ycombinator: Yeah, no, this far I understood as well from the code.  What I'm wondering is, once we set the upgrade details for the expired actions, why don't we remove them from the action queue immediately after that?

pkoutsovasilis: because if we did that and restarted this would result in our next checkin sending nil upgrade details and thus the Fleet UI wouldn't show the expiration anymore? happy to have a sync about

ycombinator: Ah okay, I got it now.  We discussed this scenario in one of the team syncs.  We need the Fleet UI to keep showing the failed state from the upgrade details until the user takes some action.  Okay, makes sense now but maybe add a comment about this in the code so I don't ask this question again in 3 weeks 😅?

pkoutsovasilis: I’ve added already a comment clarifying that these need to be persisted, which should be enough to prevent someone from altering this without checking in with the team first. For now, I’ll go ahead and merge this since it fixes the issue we have at hand. And if in 3 weeks you forget about it 😅 I’ll open a follow-up PR to extend the comment further and explicitly mention that this was a team decision to always show expired ones in Fleet UI until the user takes another upgrade action.

Issue #9574, opened 2025-08-26
Agent fails to migrate on repeated attempts after initial migration from one cluster to another.


amolnater-qasource:
**Kibana version:**
```
VERSION: 9.2.0-SNAPSHOT
BUILD: 89749
COMMIT: 9007297d8280c30040034707054745efaa708958
Artifact Link: https://snapshots.elastic.co/9.2.0-0df011a0/downloads/beats/elastic-agent/elastic-agent-9.2.0-SNAPSHOT-linux-x86_64.tar.gz
```

**Preconditions:**
1. 9.2.0-SNAPSHOT Kibana cloud environment should be available.
2. An agent must be installed and managed by Fleet, with Agent policy having System integration.
3. At least two clusters are available for migration testing.

**Steps to reproduce:**
1. Navigate to the Agents tab in Kibana.
2. Select an agent and initiate migration to a new cluster (Cluster B) using the "Migrate Agent" action, providing the new Fleet Server URL and token for an agent policy.
3. Confirm that the agent successfully migrates to Cluster B and becomes healthy.
4. Attempt to migrate the same agent again, using the "Migrate Agent" action and new Fleet Server URL/token.
5. Observe the agent is renenrolled on the same cluster again and also remains stuck in `Updating` state. 

**Expected behavior:**
Agent should migrate successfully on repeated attempts, maintaining a healthy status after each migration, regardless of the number of migrations or cluster changes.

**Agent Logs:**

[elastic-agent-diagnostics-2025-08-26T09-06-04Z-00.zip](https://github.com/user-attachments/files/21984994/elastic-agent-diagnostics-2025-08-26T09-06-04Z-00.zip)

**Agent JSON:**

[ip-172-31-88-136-agent-details.json.zip](https://github.com/user-attachments/files/21985012/ip-172-31-88-136-agent-details.json.zip)

**Screen Capture:**

<img width="1512" height="778" alt="Image" src="https://github.com/user-attachments/assets/dfb139a4-c6be-44fa-a57e-216a9d1d6fd6" />

<img width="857" height="124" alt="Image" src="https://github.com/user-attachments/assets/18f152c9-b962-49ad-9cfd-f015c5b20ddb" />

**Related Feature:**
https://github.com/elastic/kibana/issues/217619

**Additional Notes:**
- The issue appears specifically after the agent has been migrated once; further migration attempts consistently fail.

amolnater-qasource:
@muskangulati-qasource Please review.

[...]

PR #9594, opened 2025-08-27
Enroll options contains parsable URI after FromFleetConfig call

michalpristas:
when we use enroll options in migrate flow constructed by FromFleetConfig we can fail on url.Parse as URL is preset from Host that omits protocol information. 

Prolematic is  subsequent call of
```
options := FromFleetConfig()
options.RemoteConfig()
```
where RemoteConfig parses URI 

Logic is the same as we use in [client URI function](https://github.com/elastic/elastic-agent/blob/392fc972241caf6043f477f5f4c08377b5588e54/internal/pkg/remote/client.go#L255).

This PR changes construction of EnrollOptions.URL in a way that Protocol information is not lost

Related: https://github.com/elastic/elastic-agent/issues/9574

### DIFF
--- a/internal/pkg/agent/application/enroll/options.go
+++ b/internal/pkg/agent/application/enroll/options.go
@@ -7,6 +7,7 @@ package enroll
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/elastic/elastic-agent-libs/transport/httpcommon"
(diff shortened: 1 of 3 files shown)


blakerouse (approved):
Looks good.

internal/pkg/agent/application/enroll/options.go
```diff
@@ -177,6 +178,12 @@ func FromFleetConfig(cfg *configuration.FleetAgentConfig) EnrollOptions {
 		options.URL = cfg.Client.Hosts[0]
 	}
 
+	// URL at this point may not be parsable by url.Parse (missing protocol)
+	if host := options.URL; len(host) > 0 && !strings.HasPrefix(host, string(remote.ProtocolHTTPS)+"://") &&
+		!strings.HasPrefix(host, string(remote.ProtocolHTTP)+"://") {
+		options.URL = string(cfg.Client.Protocol) + "://" + host
```
pchila, line 184: If I understand correctly, in case we have a fleet URL that does not start with `http://` or `https://` we prepend whatever is in `cfg.Client.Protocol` + `://` to the URL.
 
 I did a quick snippet on [go playground](https://go.dev/play/p/xe_zGJXr3Rd?v=)  just to see what parsing some URLs would return and it would seem to work fine for no protocol, http or https.
 
 Even if we don't want to use [`url.Parse`](https://pkg.go.dev/net/url#Parse) here, wouldn't it be safer to try to match `://` and see if we get a scheme preceding that?

michalpristas: if you check related issue there's uri that if you try you'll end up with a failure
https://go.dev/play/p/wBDvAju5gNm

Issue #8544, opened 2025-06-16
Enable Safe Re-enrollment by Root/Admin for Unprivileged Elastic Agent Installs on All Platforms


kaanyalti:
## Background & Problem Statement
The original issue ([#4889](https://github.com/elastic/elastic-agent/issues/4889)) highlighted that re-enrolling an Elastic Agent installed as an unprivileged user could break the agent if the re-enroll was performed by a privileged user (root/admin).

PR [#6144](https://github.com/elastic/elastic-agent/pull/6144) addressed this by disallowing re-enrollment by privileged users when the agent was installed unprivileged, with actionable error messages. This approach worked for Linux and Darwin, but was too strict for Windows, where impersonation is not feasible.

PR [#8503](https://github.com/elastic/elastic-agent/pull/8503) relaxed the restriction for Windows, allowing any admin to re-enroll, but this reintroduced the risk of breaking the agent.

## Desired Outcome
Root or admin users should be able to safely execute a re-enroll on an agent that was installed unprivileged, on all supported platforms (Linux, Darwin, Windows), **without requiring user impersonation**.

- The underlying cause that leads to agent breakage when a privileged user re-enrolls an unprivileged install should be identified and fixed for all platforms.

## Additional Context
- [Original issue #4889](https://github.com/elastic/elastic-agent/issues/4889)
- [PR #6144 (initial fix, too strict for Windows)](https://github.com/elastic/elastic-agent/pull/6144)
- [PR #8503 (relaxed for Windows, but not a full solution)](https://github.com/elastic/elastic-agent/pull/8503)

kaanyalti:
# Important context

The PR that had introduced the windows bug was a result of my investigation in https://github.com/elastic/elastic-agent/issues/4705. My findings can be found here https://github.com/elastic/elastic-agent/issues/4705#issuecomment-2155164637.

For context this underlying error that is mentioned in the description of the PR
```
Error: failed to communicate with Elastic Agent daemon: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /opt/Elastic/Agent/elastic-agent.sock: connect: connection refused"
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.14/fleet-troubleshooting.html
``` 

Now that I know more about the agent I see that this is an error where there needs to be manual intervention to recover from. The following is what I suspect happens:
- Agent is installed in unprivileged mode
- Root user executes enroll
[...]

PR #9604, opened 2025-08-27
Fix/8544 mac linux unprivileged reenroll

kaanyalti:
- Bug

## What does this PR do?

Updates the enroll command so that the file permissions are fixed when a privileged user executes enroll. Implemented only for mac/linux. Permission fix is executed every time root executes enroll and enroll is not triggered by install. Windows implementation will be in a follow up pr. Split the PRs as windows implementation may end up being more involved.

## Why is it important?

Currently a root user needs to run `sudo -u elastic-agent-user elastic-agent enroll ...` in order to re-enroll an unprivileged agent. 

## Checklist

## Disruptive User Impact

None

- Build the agent
- Install and enroll with `--unprivileged` flag
- Execute enroll as root
- Verify that agent is still healthy

## Related issues

- Relates #8544 
- Prerequisite for #9623 -> follow up windows PR


internal/pkg/agent/cmd/enroll.go
```diff
@@ -504,9 +491,26 @@ func doEnroll(streams *cli.IOStreams, cmd *cobra.Command) error {
...
 		fixPermissions = &perms
-	}
-	if runtime.GOOS == "darwin" {
-		fixPermissions = nil
+
+		if runtime.GOOS == "darwin" {
+			fixPermissions = nil
+		}
+	} else {
+		agentState, err := getDaemonState(ctx)
```
blakerouse: This makes it a requirement that the Elastic Agent daemon is running for it to be re-enrolled. I don't think we want to do that. Just use `getFileOwnerFromPath` to determine if its unprivileged.

kaanyalti: I saw the ci failures and tracked them down to my changes. I was just about to remove the status check when I saw your comment. Thank you for pointing it out as well.

internal/pkg/agent/cmd/enroll.go
```diff
@@ -366,6 +366,41 @@ func buildEnrollmentFlags(cmd *cobra.Command, url string, token string) []string
...
+		fixPermissions = &perms
+
+		if os == "darwin" {
+			fixPermissions = nil
+		}
+	} else {
+		if hasRoot && os != "windows" { // windows is a no-op, will be addressed in a separate PR
+			perms, err := getOwnerFromPath(paths.TopBinaryPath())
+			if err != nil {
+				return nil, fmt.Errorf("failed to get owner from path %s: %w", paths.TopBinaryPath(), err)
```
kaanyalti: Should this be relaxed? If we fail to get owner from path should we block enroll? I am leaning towards relaxing this as any ownership errors can be remedied manually.

@blakerouse

blakerouse: In my opinion I think it is better to error. If we cannot get the permissions then I believe we can be sure that the enrollment will setup the files correctly. The user needs to at least have the permissions to read the permissions of the file.

internal/pkg/agent/cmd/enroll.go
```diff
@@ -366,6 +366,41 @@ func buildEnrollmentFlags(cmd *cobra.Command, url string, token string) []string
...
+	var fixPermissions *utils.FileOwner
+	if fromInstall {
+		perms, err := getFileOwnerFromCmd(cmd)
+		if err != nil {
+			// no context is added because the error is clear and user facing
+			return nil, err
+		}
+		fixPermissions = &perms
+
+		if os == "darwin" {
```
michalpristas: we could do this sooner before we do any work like querying for file owner, right

internal/pkg/agent/cmd/enroll_windows.go
```diff
@@ -53,3 +53,8 @@ func getSIDFromCmd(cmd *cobra.Command, param string) (*windows.SID, error) {
 	}
 	return nil, nil
 }
+
+func getOwnerFromPath(path string) (utils.FileOwner, error) {
+	// TODO: Implement this
+	return utils.FileOwner{}, nil
```
michalpristas, line 59: so this will result in Admnistrator owner while when we do assignement for darwin fixPermission=nil we skip it entirely. 

shouldn't we skip this for windows as well until you have followup PR?

kaanyalti: This function is only called when enroll is not triggered by install. The darwin check on the other hand is only for enroll triggered by an install. I am already skipping this for windows.

```
if hasRoot && os != "windows" { // windows is a no-op, will be addressed in a separate PR
			perms, err := getOwnerFromPath(paths.Top())
			if err != nil {
				return nil, fmt.Errorf("failed to get owner from path %s: %w", paths.Top(), err)
			}
			fixPermissions = &perms
		}
```

and the test for asserting the skip is this 

```
"should skip fixing permissions when not from installer with root on windows": {
			fromInstall:               false,
			hasRoot:                   true,
			goos:                      "windows",
			wantOwner:                 nil,
			expectOwnerFromCmdCalled:  false,
			expectOwnerFromPathCalled: false,
		}
```

kaanyalti: Do you mean that fixing permissions should be skipped for darwin at all times? Can you please clarify what you mean?

michalpristas: i think darwin is ok, what i meant here is that when we are running from not install we could have non-root permissions. 
now in the meantime is it better to reset it to administrator or leave as is? this is what i wanted you to think about. 
at the moment get owner from path returns uid and gid == "" which means that aside from admin nothing is added and could maybe lead to permission loss after enroll.

kaanyalti: The goal of this fix is to keep the permissions as they are through an enroll. The problem right now is caused by enroll changing permissions to root when agent is not privileged. Running fix permissions with each enroll ensures that agent files have the appropriate permissions for the user that owns the agent process.

Otherwise the agent won't be able to start back up as it won't be able to access the files it needs

kaanyalti: The function is not executed for windows in this PR. The functionality for this specific aspect of the enroll command/process is unchanged for windows as pointed out in this comment

```
// windows is a no-op, will be addressed in a separate PR
```

internal/pkg/agent/cmd/enroll.go
```diff
@@ -366,6 +366,39 @@ func buildEnrollmentFlags(cmd *cobra.Command, url string, token string) []string
...
+			return nil, nil
+		}
+		perms, err := getFileOwnerFromCmd(cmd)
+		if err != nil {
+			// no context is added because the error is clear and user facing
+			return nil, err
+		}
+		return &perms, nil
+	} else {
```
michalpristas: readability nit: do we need else? we have another nested if
and previous one returns in each path

kaanyalti: Not necessary, will remove

kaanyalti:
CI is failing due fips stack errors

ebeahan:
Buildkite test this

blakerouse:
Code looks good. I am building it now, and will run it before +1.

michalpristas (approved):

michalpristas:
as blake seems to be testing it as well please wait for him to get back to you

blakerouse:
I was able to test this on Linux arm64 and it is working as expected. After re-enrollment as root on an unprivileged installation all files written to disk have the correct permissions and the service is running successfully.

PR #9612, opened 2025-08-27
Add /readiness and /liveness when enrolling with the container

### LINKED ISSUE #9611: Liveness endpoint is not present when the container is enrolling
When the Elastic Agent container is enrolling (which happens before run), then it cannot check the `/liveness` endpoint. This causes the container to not be marked healthy and be killed by kubernetes in the case that enrollment doesn't succeed in the time window set by the checks.

blakerouse:
## What does this PR do?

Starts the monitoring endpoint when Elastic Agent is running as a container and has enrollment enabled. This allows the healthchecks in Kubernetes to succeed even when it is enrolling into Fleet. If enrollment to Fleet takes longer than the k8s healthchecks then it can result in the pod not being reported as healthy, which can cause it to be killed and then it never enrolls.

This also adds a simple `/readiness` endpoint that provides a different between ready and alive. There is a difference, and this provides the difference between the two.

## Why is it important?

This ensures that healthchecks do not prevent enrollment from actually working. Enrollment will still fail and the container will restart, giving the same behavior.

## Checklist

## Disruptive User Impact

None

```
$ docker run -it --rm -p 5066:5066 -e FLEET_ENROLL=1 -e FLEET_URL=https://invalid-url:443 -e FLEET_ENROLLMENT_TOKEN=invalid-token ${build_image}
$ curl -v http://localhost:5066/readiness
$ curl -v http://localhost:5066/liveness
```

## Related issues

- Closes #9611


internal/pkg/agent/application/monitoring/liveness.go
```diff
@@ -56,6 +56,13 @@ func livenessHandler(coord CoordinatorState) func(http.ResponseWriter, *http.Req
 	return func(w http.ResponseWriter, r *http.Request) error {
 		w.Header().Set("Content-Type", "application/json; charset=utf-8")
 
+		if coord == nil {
```
pchila, line 59: Is not having a coordinator the best way to detect that elastic-agent didn't complete enrollment yet ?
I was expecting something related to `managedConfigManager` and the existence of a fleet url and  token.

internal/pkg/agent/application/monitoring/server_test.go
```diff
@@ -92,32 +91,31 @@ func TestHTTPReloadEnableBehavior(t *testing.T) {
...
 		t.Run(testCase.name, func(t *testing.T) {
-			serverReloader, err := NewServer(logp.L(), api.Config{}, nil, nil, fakeCoordCfg, "linux", testCase.initConfig)
+			serverReloader, err := NewServer(logp.L(), nil, nil, fakeCoordCfg, testCase.initConfig)
 			require.NoError(t, err)
 
 			t.Logf("starting server...")
 			serverReloader.Start()
 			if testCase.httpOnAtInit {
-				waitOnReturnCode(t, http.StatusOK, "liveness", "?failon=failed", serverReloader)
+				waitOnReturnCode(t, http.StatusOK, "processes", "", serverReloader)
```
pchila, line 100: Why did all the `/liveness`  calls become `/processes` in this case ? Was this an inappropriate use of `/liveness` endpoint ?

pchila (commented):
Left a couple of comments, mostly for clarification.
Reading the PR description, I cannot help but wonder if keeping the elastic-agent alive long enough to complete enrollment should be implemented with a [`startup probe`](managedConfigManager) instead since readiness and liveness probes will not start until the startupProbe succeeds.
Maybe @pkoutsovasilis can weigh in on that as well

internal/pkg/agent/application/monitoring/liveness.go
blakerouse: Based on how this code works it is. The only time no coord is provided is during the container enrollment phase.

internal/pkg/agent/application/monitoring/server_test.go
blakerouse: Because of this line moving from here https://github.com/elastic/elastic-agent/pull/9612/files#diff-aa18619f339bb3607a4a14c45c9594b45e90163d18578c394e31528a10c6aea0L77 to here https://github.com/elastic/elastic-agent/pull/9612/files#diff-aa18619f339bb3607a4a14c45c9594b45e90163d18578c394e31528a10c6aea0R53.

The check was `isProcessStatsEnabled(cfg)` but `liveness` endpoint should always be available. This tested the toggle for turning this on and off.

blakerouse:
> Left a couple of comments, mostly for clarification. Reading the PR description, I cannot help but wonder if keeping the elastic-agent alive long enough to complete enrollment should be implemented with a [`startup probe`](managedConfigManager) instead since readiness and liveness probes will not start until the startupProbe succeeds. Maybe @pkoutsovasilis can weigh in on that as well

@pchila This is meant to not fail for agentless. We do not want the container to stop trying to enroll, we handle that with `FLEET_ENROLL_TIMEOUT=-1`. We want the pod to always be considered in good health, if its trying to enroll then its good. The default is not `FLEET_ENROLL_TIMEOUT=-1` so for normal use cases the container will fail to enroll, the container will exit non-zero and that will be reported as a crash of the pod on the kubernetes side.

pchila:
> @pchila This is meant to not fail for agentless. We do not want the container to stop trying to enroll, we handle that with `FLEET_ENROLL_TIMEOUT=-1`. We want the pod to always be considered in good health, if its trying to enroll then its good. The default is not `FLEET_ENROLL_TIMEOUT=-1` so for normal use cases the container will fail to enroll, the container will exit non-zero and that will be reported as a crash of the pod on the kubernetes side.

The startup probe is meant to signal k8s when a container can be considered "started" so that liveness and readiness checks can take over.  If elastic-agent is still enrolling and would normally fail the liveness checks, using the startup probe can be useful to give agent container the time it needs without changing the liveness and readiness checks.

blakerouse:
@pchila I assume you are saying that using a startup probe could remove the need to have the health check endpoints even available during enrollment. That still will not work, because there is a `failureThreshold` on the startup probe. That will result in a point in time where it will still mark the container as failed. We want this to run indefinitely, if its trying to enroll then it is working.

internal/pkg/agent/application/monitoring/readiness.go
```diff
@@ -0,0 +1,33 @@
...
+// This is meant to be a very simple check, just ensure that the service is up and running. For more detail
+// about the liveness of the service use the livenessHandler.
+func readinessHandler(coord CoordinatorState) func(http.ResponseWriter, *http.Request) error {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+
+		if coord != nil {
+			// to be ready the coordinator must be active
+			isUp := coord.IsActive(time.Second * 10)
+			if !isUp {
```
ycombinator: Nit: simplification.
```suggestion
			if !coord.IsActive(time.Second * 10) {
```

internal/pkg/agent/application/monitoring/liveness.go
ycombinator: If enrollment fails, we retry indefinitely.  However, if I'm following this code path correctly, `coord` will always be `nil` here in the container scenario so the liveness check will continue to succeed.  Is this the expected behavior?

blakerouse: Yes we want it to report that it is alive. It is alive and trying to enroll.

ycombinator (approved):

pchila:
> @pchila I assume you are saying that using a startup probe could remove the need to have the health check endpoints even available during enrollment. That still will not work, because there is a `failureThreshold` on the startup probe. That will result in a point in time where it will still mark the container as failed. We want this to run indefinitely, if its trying to enroll then it is working.

I would think that an agent container that didn't managed to enroll after a given period of time (let's say 5 or 10 mins for example) is not starting correctly and it should be terminated (and restarted according to its `restartPolicy`). 
Having an agent that keeps running and trying to enroll indefinitely while reporting itself alive and ready doesn't feel correct, but I may be misunderstanding the usecase.

If you already evaluated the startup probe and discarded it because there's no upper bound to the time it takes to an agent to enroll, I assume that presenting the elastic-agent alive and ready even if it's not able to perform any work until enrollment happens is a deliberate design choice.

blakerouse:
> > @pchila I assume you are saying that using a startup probe could remove the need to have the health check endpoints even available during enrollment. That still will not work, because there is a `failureThreshold` on the startup probe. That will result in a point in time where it will still mark the container as failed. We want this to run indefinitely, if its trying to enroll then it is working.
> 
> I would think that an agent container that didn't managed to enroll after a given period of time (let's say 5 or 10 mins for example) is not starting correctly and it should be terminated (and restarted according to its `restartPolicy`). Having an agent that keeps running and trying to enroll indefinitely while reporting itself alive and ready doesn't feel correct, but I may be misunderstanding the usecase.
> 
> If you already evaluated the startup probe and discarded it because there's no upper bound to the time it takes to an agent to enroll, I assume that presenting the elastic-agent alive and ready even if it's not able to perform any work until enrollment happens is a deliberate design choice.

Yes this is a deliberate design choice. As the PR description explains a normal container will still fail with enrollment because enrollment has a timeout. That timeout is configurable, so users can adjust to what they find appropriate. In this case the timeout is -1, run indefinately.

blakerouse:
@ycombinator I applied your suggestion, but that dismissed your review. If you can look again that would be great.

ycombinator (approved):

Issue #9629, opened 2025-08-28
Upgrade actions can be ignored due to stale entries in bkgActions


pkoutsovasilis:
### Background
Actions dequeued from the dispatcher action store and dispatched to the coordinator upgrader:
  1. are not retried on failure
  2. Are not persisted across an agent restart (although they should be re-sent by fleet-server on the first checkin after restart?!)

**Note:** this is probably worth tracking as a separate investigation issue.

### Issue

If Elastic-Defend and tamper protection of agent is enabled, action can remain stale in the bkgActions. Specifically we add the action invoking `getAsyncContext` but we return an err without emptying `bkgActions` in case Elastic-Defend can't acknowledge the upgrade action

https://github.com/elastic/elastic-agent/blob/3bcb53f13d7b2fc6072a61503fd442afe279c354/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L53-L74

### Impact
* Upgrade actions may remain permanently stuck in bkgActions.
* Subsequent upgrade attempts with the same version and source are ignored.
* Likely the cause of multiple recent internal error reports.

For confirmed bugs, please report:
- Version: All active releases
- Operating System: All

ycombinator:
As part of resolving this issue, it would be good to add better logging to this piece of code, as it is very complex and difficult to debug when we see issues from users: https://github.com/elastic/elastic-agent/issues/9536

PR #9634, opened 2025-08-28
fix: send upgrade action to units before adding it to bkgActions

pkoutsovasilis:
## What does this PR do?

This PR refactors the `Coordinator.Upgrade` API to use functional options (`UpgradeOpt`) instead of long argument lists, and introduces a `preUpgradeCallback`.  

Specifically:
* Moves the action proxying of Endpoint upgrades into a callback passed to the coordinator. This ensures that action dispatch only happens after the initial upgrade checks succeed (already upgrading, agent upgradeable, agent upgrade capability allowed, etc.), which is a more logical sequence.  
* Reorders checks inside `Coordinator.Upgrade` to make the flow more consistent (e.g. first verifying that the coordinator is not already upgrading before continuing).  
* Updates tests and server code to use the new options-based API.

All business logic changes are captured here bcdba7bb76ba18289482582521547fe85486ea79 (+90 -29 lines changed)

## Why is it important?

Currently, upgrade actions can remain stuck in `bkgActions` when tamper protection is enabled (see #9629). This occurs because errors returned from the Endpoint action proxying are not correctly handled, leaving stale entries behind.  

By moving this logic into a `preUpgradeCallback`, any error from proxying is now surfaced through the coordinator’s upgrade flow, which ensures that `bkgActions` are cleaned up correctly.  

I am not fully satisfied with these changes — they are a pragmatic fix until we perform a larger rewrite to centralize upgrade action handling in a single place. Still, they are necessary to resolve the immediate issue of stuck upgrade actions.

## Checklist

## Disruptive User Impact

None expected - this change only affects the internal proxying of upgrade actions to Endpoint and does not alter the upgrade request API.

```
mage unitTest
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/9629


internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
```diff
@@ -26,15 +28,17 @@ type Upgrade struct {
 	bkgCancel  context.CancelFunc
 	bkgMutex   sync.Mutex
 
-	tamperProtectionFn func() bool // allows to inject the flag for tests, defaults to features.TamperProtection
+	tamperProtectionFn         func() bool                                                                                                                            // allows to inject the flag for tests, defaults to features.TamperProtection
+	notifyUnitsOfProxiedAction func(ctx context.Context, log *logp.Logger, action dispatchableAction, ucs []unitWithComponent, performAction performActionFunc) error // allows to inject the function for tests, defaults to notifyUnitsOfProxiedAction
```
cmacknz:
```suggestion
	notifyUnitsOfProxiedActionFn func(ctx context.Context, log *logp.Logger, action dispatchableAction, ucs []unitWithComponent, performAction performActionFunc) error // allows to inject the function for tests, defaults to notifyUnitsOfProxiedAction
```

The type and the actual implementation having the same name confused me briefly until I realized that's what was happening, staring the assignment below I could not figure out where the default value came from until I checked out the code.

internal/pkg/agent/application/actions/handlers/handler_action_upgrade_test.go
```diff
@@ -368,6 +376,144 @@ func TestUpgradeHandlerNewVersion(t *testing.T) {
...
+			require.NoError(t, err, "Handle should not return an error")
+
+			select {
+			case <-upgradeCalledChan:
+				break
+			case <-time.After(10 * time.Second):
+				t.Fatal("mockCoordinator.Upgrade was not called in time")
+			}
+
+			assert.False(t, notifyUnitsCalled.Load(), "notifyUnitsOfProxiedAction should not be called")
```
cmacknz: It's not obvious why this is unconditionally false, instead of being dependent on the value of `tc.shouldProxyToEndpoint`.

I also would expect a test where this function is called but returns an error to handle that case properly (which is also what I thought the point of this fix was but perhaps I am missing something fundamental).

cmacknz (commented):
The approach looks good to me as a way to fix this without totally rewriting everything. A couple of small comments.

internal/pkg/agent/application/actions/handlers/handler_action_upgrade_test.go
pkoutsovasilis: Thanks for the note! Yeah, this is always expected to be false - the idea is to validate that the func is never called directly, but only passed as a `PreUpgradeCallback` to the coordinator upgrader. It’s mocked anyway, so even if I had a test where this exact callback returns an error, it wouldn’t add much value.

That said, I did add a dedicated unit test on the coordinator side with a callback that returns an error, and there I verify both that the error is propagated and that the upgrade details are marked as failed:
https://github.com/elastic/elastic-agent/pull/9634/files#diff-21de8fb82a5ebccaa0ca3afb0cb8bfbe10d5b8f920021785c2d485ee35a3556cR544-R586

IMO what you’re suggesting would essentially be equivalent to that test. Wdyt?

cmacknz: OK yes I see now I was reading this test expecting it to do what the other one did, which was what was misleading me.

This line is verifying that `notifyUnitsOfProxiedAction` is never called directly as you say which is a great check to had. It would be clearer if you added this context either directly to the error message in the assertion or as a comment in the code.

cmacknz: I can now see that the two tests together verify the properties we need:
1. This file verifying that errors from the coordinator upgrade handler result in background actions being updated appropriately
2. The coordinator unit test verifying that a pre upgrade check returning an error returns a coordinator error.

cmacknz:
After your clarifications my only remaining comments are suggestions for readability or documentation, which I think are necessary but if you do these and get approval from someone else go ahead and merge and don't feel the need to wait for me to review again.

internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go
pkoutsovasilis: changed here 71181e28cff3e12d9bf1f3b01c2f9ec39f44b52a

internal/pkg/agent/application/actions/handlers/handler_action_upgrade_test.go
pkoutsovasilis: added an explanation for this assertion here 9c2f633555d38472cee41e0251d87ef190366b47

cmacknz (approved):
Thanks for the follow up!

PR #9643, opened 2025-08-29
Manual rollback after grace period

pchila:
## What does this PR do?

This PR builds upon #8767 by allowing to manually rollback an elastic agent even after the grace period ends.
This PR uses a registry of elastic-agent available rollbacks introduced with PR #10344.

Having a registry of agent versions available for rollback allows for detection of possible manual rollback targets. The same available rollbacks are written in `.update-marker` file so that the watcher will preserve them at the end of the grace period, still allowing for a rollback after the watcher exits.

Installs available for rollbacks are assigned a TTL, so they can be cleaned up after a given time (governed by the `agent.upgrade.rollback.window` setting).

With this PR, **cleanup and normalization of agent installs happens only at startup**. 
In a follow-up PR:
  - the agent will be able to schedule a cleanup to run without needing a restart
  - old agent installs available for rollback will be cleaned up when initiating a new upgrade operation (in order to save disk space)

## Why is it important?

This PR allows to manually rollback an elastic-agent upgrade within a configurable window that extends beyond the grace period (the period during which an automatic rollback may be triggered by the upgraded agent misbehaving).

## Checklist

## Disruptive User Impact
No impact for users as the feature is deactivated when using the default value of `agent.upgrade.rollback.window: 0`

1. Package elastic agent twice from this PR:
  ```shell
  SNAPSHOT=true EXTERNAL=true PACKAGES=tar.gz  PLATFORMS="linux/amd64" mage -v package
  ```
  ```shell
  AGENT_PACKAGE_VERSION="9.3.0+build20251022000000" BEAT_VERSION="9.3.0-SNAPSHOT" EXTERNAL=true PACKAGES=tar.gz  PLATFORMS="linux/amd64" mage -v package
  ```
2. Install the version `9.3.0-SNAPSHOT` as usual
3. Set a rollback window duration > 0, for example 10 minutes and a shorter grace period and check_interval for watcher:
  ```yaml
agent.upgrade:
    watcher:
        grace_period: 1m
        error_check.interval: 5s
    rollback:
        window: 10m
  ```
4. Trigger an update to the other package (saved on disk):
  ```
  elastic-agent upgrade --skip-verify --source-uri=file:///vagrant/build/distributions 9.3.0+build20251022000000
  ```
5. Wait for the new agent to come online and check the upgrade details for UPG_WATCHING state. After the grace period of 1 minute, upgrade details should disappear and the watcher should exit.

6.  Verify that the `data/elastic-agent-9.3.0-SNAPSHOT-<hash>` directory is still present after watcher exited.

7. Manually rollback to the previous version: 
  ```shell
  elastic-agent upgrade --rollback 9.3.0-SNAPSHOT
  ```

8. Check the output of `elastic-agent status` and verify that the agent restarted with version `9.2.0-SNAPSHOT` and that the directory `data/elastic-agent-9.3.0+build20251022000000-<hash>` contains only logs.

## Related issues

- Requires https://github.com/elastic/elastic-agent/pull/10344

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -80,15 +80,15 @@ func RollbackWithOpts(ctx context.Context, log *logger.Logger, c client.Client, 
 
 	symlinkPath := filepath.Join(topDirPath, AgentName)
 
-	var symlinkTarget string
-	if prevVersionedHome != "" {
-		symlinkTarget = paths.BinaryPath(filepath.Join(topDirPath, prevVersionedHome), AgentName)
-	} else {
+	if prevVersionedHome == "" {
 		// fallback for upgrades that didn't use the manifest and path remapping
 		hashedDir := fmt.Sprintf("%s-%s", AgentName, prevHash)
-		// paths.BinaryPath properly derives the binary directory depending on the platform. The path to the binary for macOS is inside of the app bundle.
-		symlinkTarget = paths.BinaryPath(filepath.Join(paths.DataFrom(topDirPath), hashedDir), AgentName)
+		prevVersionedHome = filepath.Join("data", hashedDir)
 	}
+
+	// paths.BinaryPath properly derives the binary directory depending on the platform. The path to the binary for macOS is inside of the app bundle.
+	symlinkTarget := paths.BinaryPath(filepath.Join(topDirPath, prevVersionedHome), AgentName)
+
 	// change symlink
 	if err := changeSymlink(log, topDirPath, symlinkPath, symlinkTarget); err != nil {
 		return err


ycombinator:
@pchila this is a rather large PR, even after excluding the LOC for generated files.  Could you please break it up into two PRs: one that introduces the registry of installs, adapting the existing Agent code to interact with this registry, and a follow up PR that makes the manual rollback code interact with the registry? Thanks.

pchila:
> @pchila this is a rather large PR, even after excluding the LOC for generated files. Could you please break it up into two PRs: one that introduces the registry of installs, adapting the existing Agent code to interact with this registry, and a follow up PR that makes the manual rollback code interact with the registry? Thanks.

https://github.com/elastic/elastic-agent/pull/10344
[Remaining changes exclusive to this PR](https://github.com/elastic/elastic-agent/pull/9643/files/377f80d420dd3af595b41e358e79efd7648225d4..6f4272bc8e9460b6e81302d2793c4cf956707bab)

internal/pkg/agent/application/application.go
```diff
@@ -296,6 +308,86 @@ func New(
...
+// - if we just rolled back: the update marker is checked and in case of rollback we clean up the entry about the failed upgraded install
+// - check all the entries:
+//   - verify that the home directory for that install still exists (remove what does not exist anymore)
+//   - TODO check TTLs of installs to schedule delayed cleanup while the agent is running
+//
+// This function will NOT error out, it will log any errors it encounters as warnings but any error must be treated as non-fatal
+func normalizeInstallDescriptorAtStartup(log *logger.Logger, topDir string, now time.Time, initialUpdateMarker *upgrade.UpdateMarker, rollbackSource rollbacksSource) {
+	// Check if we rolled back and update the install descriptor
+	if initialUpdateMarker != nil && initialUpdateMarker.Details != nil && initialUpdateMarker.Details.State == details.StateRollback {
+		// Reset the TTL for the current version if we are coming off a rollback
```
ycombinator, line 321: I can't find in the code where the reset is being done.

internal/pkg/agent/application/application.go
```diff
@@ -296,6 +308,86 @@ func New(
...
+	}
+
+	var versionedHomesToCleanup []string
+	for versionedHome, ttlMarker := range rollbacks {
+
+		versionedHomeAbsPath := filepath.Join(topDir, versionedHome)
+
+		if versionedHomeAbsPath == paths.HomeFrom(topDir) {
+			// skip the current install
+			log.Warnf("Found a TTL marker for the currently running agent at %s. Skipping cleanup...", versionedHome)
```
ycombinator, line 351: Not sure this should be a warning, as this is almost always expected to happen, right?

internal/pkg/agent/application/application.go
```diff
@@ -296,6 +308,86 @@ func New(
...
+			continue
+		}
+
+		_, err = os.Stat(versionedHomeAbsPath)
+		if errors.Is(err, os.ErrNotExist) {
+			log.Warnf("Versioned home %s corresponding to agent install descriptor %+v  is not found on disk", versionedHomeAbsPath, ttlMarker)
```
ycombinator: Same, should this be a warning? Feels like normal/expected cleanup, no?

internal/pkg/agent/application/upgrade/manual_rollback.go
```diff
@@ -0,0 +1,258 @@
...
+		parsedVersion: currentParsedVersion,
+		version:       updateMarker.Version,
+		hash:          updateMarker.Hash,
+		versionedHome: updateMarker.VersionedHome,
+	}
+
+	return previous, current, nil
+}
+
+func getAvailableRollbacks(rollbackWindow time.Duration, now time.Time, currentVersion string, parsedCurrentVersion *version.ParsedSemVer, currentVersionedHome string, currentHash string) map[string]TTLMarker {
```
ycombinator, line 241: Doesn't look like `currentHash` is being used?

internal/pkg/agent/application/upgrade/manual_rollback.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/elastic/elastic-agent/internal/pkg/fleetapi"
+	"github.com/elastic/elastic-agent/internal/pkg/release"
+	"github.com/elastic/elastic-agent/pkg/core/logger"
+	"github.com/elastic/elastic-agent/pkg/version"
+	agtversion "github.com/elastic/elastic-agent/version"
+)
+
+const disableRollbackWindow = time.Duration(0)
+
+func (u *Upgrader) rollbackToPreviousVersion(ctx context.Context, topDir string, now time.Time, version string, action *fleetapi.ActionUpgrade) (reexec.ShutdownCallbackFn, error) {
```
ycombinator, line 28: I don't think `action` is needed here?

internal/pkg/agent/application/upgrade/manual_rollback.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	prevAgentInstall := agentInstall{
+		parsedVersion: prevAgentParsedVersion,
+		version:       targetTTLMarker.Version,
+		hash:          targetTTLMarker.Hash,
+		versionedHome: targetInstall,
+	}
+
+	upgradeDetails := details.NewDetails(release.VersionWithSnapshot(), details.StateRequested, "" /*action.ID*/)
```
ycombinator, line 130: I'm wondering if we should introduce a new state, something like `UPG_REQUESTED_ROLLBACK` to distinguish it from `UPG_REQUESTED`?  Might make it easier to reason about the state of the upgrade when looking at a diagnostic.  That also makes the upgrade marker created here representative of the state of the upgrade and not just a fake one.

internal/pkg/agent/application/upgrade/manual_rollback.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, fmt.Errorf("stat() on upgrade marker %q failed: %w", updateMarkerPath, err)
+	}
+
+	var watcherExecutable string
+	var versionedHomeToRollbackTo string
+	var updateMarkerExistsBeforeRollback bool
+
+	if errors.Is(err, os.ErrNotExist) {
+		// there is no upgrade marker, we need to extract available rollbacks from agent installs
```
ycombinator: What is the scenario where a manual rollback would be requested but the upgrade marker wouldn't exist?

internal/pkg/agent/application/upgrade/manual_rollback.go
```diff
@@ -0,0 +1,258 @@
...
+	err = withTakeOverWatcher(ctx, log, topDir, watcherHelper, func() error {
+		// read the upgrade marker
+		updateMarker, err = LoadMarker(paths.DataFrom(topDir))
+		if err != nil {
+			return fmt.Errorf("loading marker: %w", err)
+		}
+
+		if updateMarker == nil {
+			return ErrNilUpdateMarker
+		}
```
ycombinator, line 170: Do you need this code? Isn't `updateMarker` already in scope?  Or is it possible the update marker file could disappear between now and when this closure gets invoked?

internal/pkg/agent/application/upgrade/manual_rollback.go
```diff
@@ -0,0 +1,258 @@
...
+	var watcherExecutable string
+	var versionedHomeToRollbackTo string
+	var updateMarkerExistsBeforeRollback bool
+
+	if errors.Is(err, os.ErrNotExist) {
+		// there is no upgrade marker, we need to extract available rollbacks from agent installs
+		watcherExecutable, versionedHomeToRollbackTo, err = rollbackUsingAgentInstalls(u.log, u.watcherHelper, u.availableRollbacksSource, topDir, now, version, u.markUpgrade)
+	} else {
+		updateMarkerExistsBeforeRollback = true
+		watcherExecutable, versionedHomeToRollbackTo, err = rollbackUsingUpgradeMarker(ctx, u.log, u.watcherHelper, topDir, now, version)
```
ycombinator, line 52: Just wondering if we need this case.  Couldn't we always `rollbackUsingAgentInstalls`?

internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -126,17 +127,21 @@ func RollbackWithOpts(ctx context.Context, log *logger.Logger, c client.Client,
 		return nil
 	}
 
+	if prevVersionedHome == "" {
+		prevVersionedHome = filepath.Join("data", fmt.Sprintf("%s-%s", AgentName, prevHash))
+	}
```
ycombinator: Trying to understand: why is this needed? Could this could be set earlier in this function as part of the else clause on line 85? Also, instead of hardcoding "data", probably better to use `paths.DataFrom(topDirPath)`?

internal/pkg/agent/cmd/watch.go
```diff
@@ -198,7 +198,12 @@ func watchCmd(log *logp.Logger, topDir string, cfg *configuration.UpgradeWatcher
 		// if we're not within grace and marker is still there it might mean
 		// that cleanup was not performed ok, cleanup everything except current version
 		// hash is the same as hash of agent which initiated watcher.
-		if err := installModifier.Cleanup(log, paths.Top(), paths.VersionedHome(topDir), release.ShortCommit(), true, false); err != nil {
+		versionedHomesToKeep := make([]string, 0, len(marker.RollbacksAvailable)+1)
```
ycombinator, line 201: Nit: the capacity here, specifically the `len(marker.RollbacksAvailable)` part depends on `appendAvailableRollbacks` using the same length slice.  Might be safer not to make that assumption and leave the capacity as 0 (default)?

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: Manual rollback being requested after grace period expires.
Upgrade marker is cleaned up when the elastic agent watcher determines that the upgraded agent is running without issues but the rollback window should always be longer than the grace period (except for degenerate cases)

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: Give the current state of upgrade details I am hesitant to add another state there. This set here is just to accomodate the initialization code we have in the startup of the rolled back agent.
The `UPG_REQUESTED` is the same state that agent will **unconditionally** enter upon receiving an upgrade request/action of any kind

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: the difference is that at this point I have the guarantee that no other process is manipulating the upgrade marker, so I reload the most recent state.
The previous upgrade marker loading is there to detect the more recent agent version in order to re-launch the appropriate watcher in case of rollback failure.

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: well spotted! it's missing in the TTLMarker struct below

internal/pkg/agent/application/upgrade/rollback.go
pchila: This is needed because it's the expected fallback in case versionedHome is not present.

The cleanup function can be called with data coming from older upgrade marker versions that do not contain an explicit versionedHome (this appeared only in agent >8.13.0) so a fallback is needed (similar to what is happening at line 87/88 just above)
The `path.DataFrom()` can be used but it return an absolute path while here we need a relative one so it would look like:
```
absData := paths.Data(topDir)
prevVersionedHome, err = filepath.Rel(topDir, filepath.Join(absData, fmt.Sprintf("%s-%s", AgentName, prevHash)))
```

internal/pkg/agent/application/application.go
pchila: A few lines below (328)

```golang
		// remove the current versioned home TTL marker
		delete(rollbacks, initialUpdateMarker.PrevVersionedHome)
		err = rollbackSource.Set(rollbacks)
```

internal/pkg/agent/application/application.go
pchila: It should not be a normal case having cleaned up after a rollback in the block above (lines 320-335).
If we still have a TTL for the version of agent that is currently starting it means that something went wrong somewhere. The code will protect the current install by not trying to clean up the current install but I believe that we should at least log a warning as this will not impede the current operation but could be a symptom of something that needs fixing

internal/pkg/agent/application/application.go
pchila: This is even weirder: we got a versionedHome from the rollbackSource but the directory does not exist. Depending on the implementation of the rollbackSource, that may signal some incorrect available rollback data or a corrupted install.
It should be rare but I still think it should be logged if that happens

internal/pkg/agent/cmd/watch.go
pchila: We can leave the capacity at zero and reallocate as necessary, however the resulting slice of versioned homes to keep at the end of the grace period will always be the `<number of available rollbacks > + <the new versioned home of the upgraded elastic-agent>`

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: It's gonna be used in the next PR where rolling back a managed agent will be supported.
The action will need to be acked anyways.
In the meantime I can use the `_` identifier in this PR to mark the parameter as unused.

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: As I was implementing the feature I preferred giving priority to the upgrade marker if present (meaning we are still in the grace period, we need to stop gracefully any watcher that may be running and then perform a rollback).
The rollbackUsingAgentInstalls assumes there's no watcher running, so it can simply select which watcher to launch and which versionedHome to rollback to based on the current install state.

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: please have a look at 1cd2407e1b

testing/integration/ess/upgrade_rollback_test.go
```diff
@@ -390,23 +396,96 @@ func TestStandaloneUpgradeManualRollback(t *testing.T) {
...
+			rollbackTrigger: func(ctx context.Context, t *testing.T, client client.Client, startFixture, endFixture *atesting.Fixture) {
+				// trim -SNAPSHOT at the end of the fixture version as that is reported as a separate flag
+				expectedVersion := endFixture.Version()
+				expectedSnapshot := false
+				if strings.HasSuffix(expectedVersion, "-SNAPSHOT") {
+					expectedVersion = strings.TrimSuffix(endFixture.Version(), "-SNAPSHOT")
+					expectedSnapshot = true
+				}
+
+				// It will take at least 2 minutes before the agent exits the grace period (see fastWatcherCfgWithRollbackWindow)
```
cmacknz, line 453: Is it possible to run with a grace period of 0 to speed this test up?

internal/pkg/agent/cmd/watch.go
```diff
@@ -198,7 +198,12 @@ func watchCmd(log *logp.Logger, topDir string, cfg *configuration.UpgradeWatcher
...
+		// current version needs to be kept
+		versionedHomesToKeep = append(versionedHomesToKeep, paths.VersionedHome(topDir))
+		versionedHomesToKeep = appendAvailableRollbacks(log, marker, versionedHomesToKeep)
+		log.Infof("About to clean up upgrade. Keeping versioned homes: %v", versionedHomesToKeep)
+		if err := installModifier.Cleanup(log, paths.Top(), true, false, paths.VersionedHome(topDir)); err != nil {
```
cmacknz: Is it intentional that `versionedHomesToKeep` is not passed as an argument to `installModified.Cleanup`? Why?

internal/pkg/agent/cmd/watch.go
```diff
@@ -198,7 +198,12 @@ func watchCmd(log *logp.Logger, topDir string, cfg *configuration.UpgradeWatcher
 		// if we're not within grace and marker is still there it might mean
 		// that cleanup was not performed ok, cleanup everything except current version
 		// hash is the same as hash of agent which initiated watcher.
-		if err := installModifier.Cleanup(log, paths.Top(), paths.VersionedHome(topDir), release.ShortCommit(), true, false); err != nil {
+		versionedHomesToKeep := make([]string, 0, len(marker.RollbacksAvailable)+1)
+		// current version needs to be kept
+		versionedHomesToKeep = append(versionedHomesToKeep, paths.VersionedHome(topDir))
```
cmacknz, line 203: 👍 I think just this would have been enough to prevent the problem fixed in https://github.com/elastic/elastic-agent/pull/10579. Something a bit stronger like making sure we never cleaning up the version the elastic-agent symlink points to would also help.

For that matter, we potentially should move cleanup out of the watcher entirely and try to get it's responsibility down to just switching the symlink back on rollback. Thoughts? Beyond the scope of what should happen in this PR but evaluating options to make this process less brittle as I'm looking at the code involved.

internal/pkg/agent/application/application.go
```diff
@@ -296,6 +308,86 @@ func New(
...
 }
 
+// normalizeInstallDescriptorAtStartup will check the install descriptor checking:
+// - if we just rolled back: the update marker is checked and in case of rollback we clean up the entry about the failed upgraded install
+// - check all the entries:
+//   - verify that the home directory for that install still exists (remove what does not exist anymore)
+//   - TODO check TTLs of installs to schedule delayed cleanup while the agent is running
+//
+// This function will NOT error out, it will log any errors it encounters as warnings but any error must be treated as non-fatal
+func normalizeInstallDescriptorAtStartup(log *logger.Logger, topDir string, now time.Time, initialUpdateMarker *upgrade.UpdateMarker, rollbackSource rollbacksSource) {
```
cmacknz: Does `installDescriptor` in the name still make sense? Isn't this naming from prior to the switch to TTL markers?

internal/pkg/agent/application/application.go
```diff
@@ -296,6 +308,86 @@ func New(
...
+	// Check if we rolled back and update the install descriptor
```
cmacknz: For example, this comment references an install descriptor that no longer exists in the code.

internal/pkg/agent/application/application.go
cmacknz: Should this code clean up the TTL marker for the current agent version if we 've gotten into this state? Won't the marker hang around indefinitely otherwise?

internal/pkg/agent/application/application.go
cmacknz: There are a couple more references to install descriptors in logs and such as well.

internal/pkg/agent/application/application.go
```diff
@@ -296,6 +308,86 @@ func New(
...
+
+		if err != nil {
+			log.Warnf("error checking versioned home %s for agent install: %s", versionedHomeAbsPath, err.Error())
+			continue
+		}
+
+		if now.After(ttlMarker.ValidUntil) {
+			// the install directory exists but it's expired. Remove the files.
+			log.Infof("agent install descriptor %+v is expired, removing directory %q", ttlMarker, versionedHomeAbsPath)
+			if cleanupErr := install.RemoveBut(versionedHomeAbsPath, true); cleanupErr != nil {
```
cmacknz, line 370: If someone is waiting to free up disk space from expired rollback versions, and there is some transient failure here, their only option is to go onto the agent machine and run the `elastic-agent restart` command until it works from what I can tell.

Should this cleanup process be on a timer until it succeeds or something similar, instead of a one shot attempt at each start? If we did do this, it could be a follow up improvement.

internal/pkg/agent/application/application.go
```diff
@@ -131,6 +139,10 @@ func New(
 	isMonitoringSupported := !disableMonitoring && cfg.Settings.V1MonitoringEnabled
 
 	availableRollbacksSource := upgrade.NewTTLMarkerRegistry(log, paths.Top())
+	if platform.OS != component.Container {
```
cmacknz: This should exclude RPM and DEB as well shouldn't it? Should this be a call to `upgrade.IsUpgradeable()` instead?

internal/pkg/agent/application/application.go
pchila: The TTLMarker could be removed, I took the no action approach as I thought that should/could/would be investigated by a human so we try to leave everything as-is in the normalization phase.

testing/integration/ess/upgrade_rollback_test.go
pchila: It would be possible but some [common code for the upgrade test](https://github.com/elastic/elastic-agent/blob/1cf28a8f2916316d7d9e4c9cdc30e7d82694121a/testing/upgradetest/upgrader.go#L384-L385) will expect to see a watcher come up before calling a `postUpgradeHook` (which is used to cut short the upgrade code and take back control):
a grace period of 0 would have the watcher not even try to watch and exit immediately and that would need to be handled in `PerformUpgrade` utility function.

In the last build the duration of this test is ~214s (3m34s) and this modification would save 1m overall at the expense of some flakiness due to trying to detect a watcher that is done immediately.

internal/pkg/agent/cmd/watch.go
pchila: Good catch!
No it isn't intentional, this is happening to work ok atm because we only preserve topDir (and that is why all the integration tests work so far) but it seems that some modifications were lost during refactoring/rebasing on top of PR https://github.com/elastic/elastic-agent/pull/10344

Will fix it.

internal/pkg/agent/cmd/watch.go
pchila: This wouldn't have saved us in https://github.com/elastic/elastic-agent/pull/10579 as in that case it was the previous version watcher that was misbehaving. This is the sad reality of our upgrade mechanism we cannot influence what code included in past versions does.

I still believe that upgrade should be entirely handled by the new version that is used to spawn a separate `upgrade` process (similar to watcher) that will take over just after downloading the new archive.

internal/pkg/agent/application/application.go
pchila: It should exclude anything that does not support `elastic-agent upgrade` commands.
`Container` was the most obvious but it's true that RPM and DEB should not mess with `TTLMarker`s since they have a separate upgrade mechanism.
`upgrade.IsUpgradeable()` seems to have a slightly different meaning:

```golang
// IsUpgradeable when agent is installed and running as a service or flag was provided.
func IsUpgradeable() bool {
	// only upgradeable if running from Agent installer and running under the
	// control of the system supervisor (or built specifically with upgrading enabled)
	return release.Upgradeable() || (paths.RunningInstalled() && info.RunningUnderSupervisor())
}
```
- [ `release.Upgradeable()`](https://github.com/elastic/elastic-agent/blob/57bed542f063f6a27b7c6f992a94d19e7c2304e4/internal/pkg/release/upgrade.go#L8-L10) is relying on a [package variable that is documented as a debug flag](https://github.com/elastic/elastic-agent/blob/57bed542f063f6a27b7c6f992a94d19e7c2304e4/internal/pkg/release/version.go#L29-L31) and AFAICT it's set only if `DEV == true` during [packaging](https://github.com/elastic/elastic-agent/blob/57bed542f063f6a27b7c6f992a94d19e7c2304e4/magefile.go#L2053-L2056)
-  [`paths.RunningInstalled()`](https://github.com/elastic/elastic-agent/blob/57bed542f063f6a27b7c6f992a94d19e7c2304e4/internal/pkg/agent/application/paths/common.go#L335-L342) will check for the presence of the `.installed` marker file (which should be absent in agents running from docker images, deb and rpm installs)
- `info.RunningUnderSupervisor()`will check if the current agent process has been launched by the system manager (should still be true for deb, rpm and containers that start through scripts/`tini`)

All in all this will cover the deb,rpm and container case thanks to the `paths.RunningInstalled()` check with some exceptions made for DEV builds (which should not matter for testing anyways), will change the check.

internal/pkg/agent/cmd/watch.go
pchila: fixed in https://github.com/elastic/elastic-agent/pull/9643/commits/4ccc86728006b46b8801271481a018c1a0c10ba2

internal/pkg/agent/application/application.go
pchila: changed it in https://github.com/elastic/elastic-agent/pull/9643/commits/25743948ec8253f9f14b24991f1bbbac0fb2a92d

internal/pkg/agent/application/application.go
pchila: cleaned up in https://github.com/elastic/elastic-agent/pull/9643/commits/dcf0ede92e3ba49bd079e551f66a24f91201fc79

internal/pkg/agent/application/application.go
pchila: cleaned up in https://github.com/elastic/elastic-agent/pull/9643/commits/dcf0ede92e3ba49bd079e551f66a24f91201fc79

internal/pkg/agent/application/application.go
pchila: In a follow-up PR, async cleanup of old agent version will be implemented, based on the first read of TTL markers during agent startup.
In case of problem during the cleanup the interval for the next check can be adjusted to retry the normalization after some backoff.

internal/pkg/agent/cmd/watch.go
cmacknz: Right we can't fix the old watcher, but we can put a guard into the current watcher code so that one day when it is the "old" watcher the problem won't happen, which is what I am looking for.

pchila:
buildkite test this

testing/integration/ess/upgrade_rollback_test.go
cmacknz: I don't want to hold up this PR any longer to save 1m, thanks for including the timing change.

internal/pkg/agent/application/application.go
cmacknz: OK I can buy into that, trying to correct too much just increases the places where things can go wrong

cmacknz:
Code looks good to me, I went through the manual test steps and it worked with the following observations:

- I don't see rollbacks available reported from the status command, just the watching state:
```
┌─ fleet
├─ elastic-agent
└─ upgrade_details
   ├─ target_version: 9.3.0+build20251022000000
   ├─ state: UPG_WATCHING
   └─ metadata
```
- The agent status appears that it will indefinitely contain the rolled back status, this is fine since that's what happened. Someone may ask us to change this later I think it will permanently highlight the agent in the Fleet UI with the rolled back status which is less useful if it was manual and not automatic.
```
┌─ fleet
├─ elastic-agent
└─ upgrade_details
   ├─ target_version: 9.3.0+build20251022000000
   ├─ state: UPG_ROLLBACK
   └─ metadata
      └─ reason: manual rollback requested to version 9.3.0-SNAPSHOT
```

cmacknz:
@ycombinator still has open comments to resolve, @ycombinator can you confirm these are addressed and approve? This LGTM but I want to avoid approving when there are open comments from someone else remaining.

internal/pkg/agent/application/upgrade/manual_rollback.go
ycombinator: Thanks, that makes sense.  I think it would be good add some inline comments explaining why we need the two separate approaches (basically what you explained to me above in the PR comment).

internal/pkg/agent/application/upgrade/manual_rollback.go
ycombinator: Right, of course! I think it would be good to expand on this inline comment a bit explaining why/when the upgrade marker could not exist but a manual rollback could still be requested.  Something like:

```suggestion
		// there is no upgrade marker because the rollback was requested after the watcher grace period had elapsed, so we need to extract available rollbacks from agent installs
```

internal/pkg/agent/application/upgrade/rollback.go
ycombinator: Agreed re: using `path.DataFrom()` being overkill here since we need the relative path anyway.

ycombinator: Regarding setting `prevVersionedHome` here, I was more pointing out that we already check for `prevVersionedHome == ""` on line 86 (`else` clause), so couldn't we just set it inside that clause instead of here?

pchila: I see now: yeah the check already at line 86 makes this redundant. I guess it slipped in due to all the refactoring/rebasing/restructuring that happened. The check at this line can be removed.

pchila: Fixed in 52ab103953

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: Fixed in 52ab103953

internal/pkg/agent/application/upgrade/manual_rollback.go
pchila: Fixed in 52ab103953

Issue #4971, opened 2024-06-20
[windows] move service startup to beginning of run function


leehinman:
Currently `elastic-agent` does several things before it communicates with the Windows Service Manager and registers itself and reports as running.

https://github.com/elastic/elastic-agent/blob/81e112dfb6202d24a0b2a8d73067b091c3647cf0/internal/pkg/agent/cmd/run.go#L143

This can cause problems because the service manager has a very short timeout between when it starts a process and expects that process to communicate that it is in StartPending state.  For Windows we need to move this as close as possible to the start of the run function.

michalpristas:
@cmacknz looks like this one is popping up again in escalations

@ebeahan labeled as a good first issue it may be suitable for some of our new friends in a team.

kevinr-security:
Can confirm we are seeing this issue especially on a couple of large customers. Urgently need this to be looked at because it cascades into multiple issues with Orphaned states, API Key issues etc.

cmacknz:
Thanks, cc @michalpristas who has seen some other cases of slower windows service startup popping up.

I don't think we have a root cause yet but we are aware something has changed here that needs to be looked into.

kevinr-security:
Case #01907529 is one such issue.

kevinr-security:
We have hundreds of affected endpoints (with Defend level logging) if you want to examine. All manner of builds and specs, some high-end.

Only point of note is multiple EDR platforms deployed on most of the hosts that could contribute to startup delays, but that's just a theory.

cmacknz:
Thanks for the context, the few reports we have seen so far are on busier or more resource constrained machines so multiple EDRs would fit into that depending on what exactly is going on.

The change described here in theory can only help but depending on what the root cause is (there are ways for delays to creep in before this code can execute when the Go runtime is initializing) it might not fix it.

cmacknz:
If you've noticed this after an upgrade the version you started at and the version the problem showed up in would help us narrow the search for a code change that could have caused this.

I don't think the case you linked has made its way to engineering yet, once that happens we can ask for more information there.

stevenpyper:
@cmacknz We didn't notice this after a particular update, but we have confirmed the issue across all versions 9.0.0-9.0.5 as well as confirmed it on 8.17 which is the oldest version we have available.

Found the issue after looking at agents marked as orphaned where the defend agent had started but the elastic agent had not. Confirmed that this error was seen across multiple estates at least 3 months ago but has became much more common within the last month.

[...]

PR #9672, opened 2025-09-01
Refactor to optimise for Windows service manager checkin

michalpristas:
Details to be added later

Related #4971


internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+	executeFinished: make(chan struct{}),
+}
+
+func init() {
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		wg.Done()
+		processWindowsControlEvents(stopBeat)
```
bjmcnic, line 42: It seems this will at least guarantee the go routine started before we return from `init`, but would better to wire it up such that `wg.Done()` doesn't get called until `changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}` in `Execute` ?

internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/elastic/elastic-agent-libs/logp"
+
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/debug"
+)
+
+// couldNotConnect is the errno for ERROR_FAILED_SERVICE_CONTROLLER_CONNECT.
+const couldNotConnect syscall.Errno = 1063
+
+type beatService struct {
```
michalpristas, line 24: code change: bringin in dependency

internal/pkg/agent/cmd/agentrun/run.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return cmd
+}
+
+func runService(testingMode bool, fleetInitTimeout time.Duration) error {
+	// Windows: Mark service as stopped.
+	// After this is run, the service is considered by the OS to be stopped.
+	// This must be the first deferred cleanup task (last to execute).
+	defer func() {
+		agentservice.NotifyTermination()
```
michalpristas, line 85: code change: calling internal implementation

internal/pkg/agent/cmd/agentrun/run.go
```diff
@@ -0,0 +1,93 @@
...
+		agentservice.WaitExecutionDone()
+	}()
+
+	service.BeforeRun()
```
michalpristas, line 89: these are original

internal/pkg/agent/cmd/agentrun/run.go
```diff
@@ -0,0 +1,93 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package agentrun
```
michalpristas, line 5: code change: run -> agentrun to help with alphabetical ordering during init

testing/integration/ess/init_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/elastic/elastic-agent/pkg/core/process"
+	atesting "github.com/elastic/elastic-agent/pkg/testing"
+	"github.com/elastic/elastic-agent/pkg/testing/define"
+	"github.com/elastic/elastic-agent/pkg/testing/tools/testcontext"
+	"github.com/elastic/elastic-agent/testing/installtest"
+	"github.com/elastic/elastic-agent/testing/integration"
+)
+
+func TestInitOrderNotDegraded(t *testing.T) {
```
michalpristas, line 28: code change: added test
not entirely happy with testing criteriabut at least it checks it is called during init and not at the end of it

internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+var serviceInstance = &beatService{
+	stopCallback:    nil,
+	done:            make(chan struct{}),
+	executeFinished: make(chan struct{}),
+}
+
+func init() {
+	var wg sync.WaitGroup
```
michalpristas, line 37: code change: added waitgroup to prevent further execution of another init functions up to the point when goroutine is up and runnin

internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+func NotifyTermination() {
+	serviceInstance.stop()
+}
+
+func trySendState(s svc.State, changes chan<- svc.Status) {
+	select {
+	case changes <- svc.Status{State: s}:
+	case <-time.After(500 * time.Millisecond): // should never happen, but don't make this blocking
+	}
+}
```
bjmcnic, line 186: Why is this name prefixed with try and why is there a timeout?  Have we encountered failures or blocking?  Under the covers this should hit Win32's `SetServiceStatus()`.  I haven't seen other services coded to guard against that blocking.  It succeeds or fails in a timely manner.   If there is a case where we need to timeout and not block, 500ms seems short.  Is there a reason that's a good timeout?

internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+
+	trySendState(svc.StopPending, changes)
+	defer trySendState(svc.Stopped, changes)
+
+	logp.L().Info("changed windows service state to svc.StopPending, invoking stopCallback")
+	m.stopCallback()
+
+	// Block until notifyWindowsServiceStopped below is called. This is required
```
bjmcnic: I'm not finding `notifyWindowsServiceStopped` below (like in this file).  Does the comment need to be updated?  Or am I missing it?

internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+	// as the windows/svc package will transition the service to STOPPED state
+	// once this function returns.
+	<-m.done
+	logp.L().Debug("windows service state changed to svc.Stopped")
```
bjmcnic:
```suggestion
	logp.L().Debug("windows service state changing to svc.Stopped")
```

bjmcnic: The call won't occur until after the logged line, right?

internal/pkg/agent/agentservice/agentservice_windows.go
```diff
@@ -0,0 +1,200 @@
...
+// If response is not retrieved within 500 millisecond wait is aborted.
+func WaitExecutionDone() {
+	if isWinService, err := svc.IsWindowsService(); err != nil || !isWinService {
+		// not a service, don't wait
+		return
+	}
+
+	select {
+	case <-serviceInstance.executeFinished:
+	case <-time.After(500 * time.Millisecond):
```
bjmcnic: 500ms seems short if this wait is important.  On the successful path we'd still succeed fast, but is there a reason on the problematic path we want to give up that fast?

internal/pkg/agent/agentservice/agentservice_windows.go
michalpristas: in fact i don't know the reason for this one, this is pulled from external third party library as is. 
in go we consider it a good practice to guard against these issues in case there's a bug i guess

internal/pkg/agent/agentservice/agentservice_windows.go
michalpristas: yes should be NotifyTermination

internal/pkg/agent/agentservice/agentservice_windows.go
michalpristas: that's right

internal/pkg/agent/agentservice/agentservice_windows.go
michalpristas: i think this is only being defensive

internal/pkg/agent/agentservice/agentservice_windows.go
michalpristas: this is interesting idea. i will try to play with that. now that the code is in our repo it should be possible

internal/pkg/agent/agentservice/agentservice_windows.go
bjmcnic: I anticipate this will reliably succeed in a timely fashion, but perhaps we should grow the timeout a bit to ensure we never risk prematurely timing it out while it's on the path to succeeding.  Maybe 5000ms instead of 500ms?  That should still suffice to guard against an unexpected bug that would cause it to block.

internal/pkg/agent/agentservice/agentservice_windows.go
bjmcnic: I again worry 500ms is so short it might interrupt success on a slow or busy host.  Would 2000 or 3000ms still work and provide enough defense?

michalpristas: increased time sounds good to me. not aggressive and provides guarantees. i will update these soon

internal/pkg/agent/agentservice/agentservice_windows.go
michalpristas: tried to play with it, it is somewhat tricky when it comes to non-service agents during enrollment. i'll call current behavior as good enough

ebeahan:
@michalpristas do we still need to pursue the changes here? Or can we close?

michalpristas:
My preference would be to close and go with Blakes proposal as it is less fragile, more deterministic

PR #9716, opened 2025-09-03
[Upgrade] Cleanup: Remove unnecessary type assertions

ycombinator:
## What does this PR do?

This PR removes unnecessary type assertions inside the `Upgrade` struct's `getAsyncContext` method.

## Why is it important?

To simplify the code and make it a bit more readable.

Right before the `getAsyncContext` method is called, we [assert](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L48-L53) that `action` is of type `*fleetapi.ActionUpgrade`.  So we don't need to repeat those assertions inside the `getAsyncContext` method as well.

### DIFF
--- a/internal/pkg/fleetapi/action.go
+++ b/internal/pkg/fleetapi/action.go
@@ -266,7 +266,7 @@ type ActionUpgradeData struct {
 	Retry int `json:"retry_attempt,omitempty" yaml:"retry_attempt,omitempty" mapstructure:"-"`
 }
 
-func (a ActionUpgrade) String() string {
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (approved):
I understand that this PR needs to go in first from another one, thus, to avoid any delays due to timezone diffs, I will approve it now since this LGTM and thus you can merge it when the CI is green @ycombinator 

PS: I don't expect any CI surprises here now that the linters do not complain, this is straight-forward

Issue #9534, opened 2025-08-22
Agent gets unhealthy when migrated with Elastic Defend integration to a  new cluster with policy having Elastic Defend integration.


amolnater-qasource:
**Kibana version:**
```
VERSION: 9.2.0-SNAPSHOT
BUILD: 89688
COMMIT: 6e2386a13132c8860593038b65eb435b0b421c67
Artifact Link: https://snapshots.elastic.co/9.2.0-021aded8/downloads/beats/elastic-agent/elastic-agent-9.2.0-SNAPSHOT-linux-x86_64.tar.gz
```

**Preconditions:**
1. 9.2.0-SNAPSHOT Kibana cloud environment should be available.
2. An agent should be installed with Agent policy having Elastic Defend integration and tamper protection disabled.

**Steps to reproduce:**
1. Navigate to Agents tab
2. Click Actions>Migrate Agent and add new Fleet Server URL and token for the agent policy having Elastic Defend integration.
3. Click Migrate Agent.
4. Observe agent is migrated to the new cluster.

**Expected behavior:**
Agent should get Healthy when migrated with Elastic Defend integration to a  new cluster with policy having Elastic Defend integration.

**Agent Logs:**

[elastic-agent-diagnostics-2025-08-22T07-38-23Z-00.zip](https://github.com/user-attachments/files/21933231/elastic-agent-diagnostics-2025-08-22T07-38-23Z-00.zip)

**Agent JSON:**
[ip-172-31-88-136-agent-details.zip](https://github.com/user-attachments/files/21933230/ip-172-31-88-136-agent-details.zip)

**Screen Capture:**

<img width="974" height="209" alt="Image" src="https://github.com/user-attachments/assets/fac2201a-9a34-47bb-83c2-4c0c5316ac99" />

https://github.com/user-attachments/assets/5e52f844-4bf4-4638-95e1-ed56dd27adf9

**Related Feature:**
https://github.com/elastic/kibana/issues/217619

amolnater-qasource:
@muskangulati-qasource Please review.

muskangulati-qasource:
Secondary review is Done for this ticket!

cmacknz:
Endpoint's API key was invalidated:

```json
[...]

PR #9801, opened 2025-09-08
Forward migrate action to endpoint

michalpristas:
Related issue is fixed on endpoint side and requires us to forward action to them.

Strategy here is to send MIGRATION action to them so they can prepare and update keys when new config arrives (after reenroll). without this they would consider new policy as untrustworthy 
`isMigrationInProgress` bool was added to coordinator for that purpose, we don't clear it to default value, migration is successful therefore we need to reexec very soon. 

Related: https://github.com/elastic/elastic-agent/issues/9534


specs/endpoint-security.spec.yml
```diff
@@ -14,6 +14,7 @@ inputs:
     proxied_actions: &proxied_actions
       - UNENROLL
       - UPGRADE
+      - MIGRATE
```
pkoutsovasilis, line 17: I would expect here to proxy the action to endpoint as we do [here](https://github.com/elastic/elastic-agent/blob/05f62c2a8e9766c0d513c0791c386d38d67276c4/internal/pkg/agent/application/actions/handlers/handler_action_unenroll.go#L80) for `unenroll` and [here](https://github.com/elastic/elastic-agent/blob/05f62c2a8e9766c0d513c0791c386d38d67276c4/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L65) for `upgrade`. Is the `migrate` action being proxied to endpoint somewhere else that I am missing?

michalpristas: you're right, i added notification after successful migration just before we reexec so we don't notify endpoint in case of invalid condition (like us running fleet server).

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -100,11 +102,33 @@ func (h *Migrate) Handle(ctx context.Context, a fleetapi.Action, ack acker.Acker
 		return fmt.Errorf("migration of agent to a new cluster failed: %w", err)
 	}
 
+	// action is all rigth we can notify endpoint
+	if err := h.notifyComponents(ctx, action); err != nil {
+		// config is cleaned up already we cannot revert
+		h.log.Warnf("failed to notify components, aborting: %v", err)
```
pkoutsovasilis: ok so what's the expectation here if this returns an error? Should we maybe notify endpoint before actually calling coord.Migrate? the latter happens with the upgrade action

michalpristas: yes that's also an alternative, endpoint will be longer in a state that allows acceptance of mixed origin policy though. up to the point when we restart so we could reexec for failures as well. 

this is about compromise, i found this one better but i'm up for a discussion if you dont like this

michalpristas: for visibility, waiting on PM to get some feedback about lesser even of these two options

internal/pkg/agent/application/actions/handlers/handler_action_migrate_test.go
```diff
@@ -337,6 +346,57 @@ func TestActionMigratelHandler(t *testing.T) {
...
+		h := NewMigrate(log, mockAgentInfo, coord)
+		h.tamperProtectionFn = func() bool { return false }
+
+		require.NoError(t, h.Handle(t.Context(), action, ack))
+		coord.AssertNumberOfCalls(t, "Migrate", 1)
+
+		// ack not delegated to migrate coordinator, failure is reported
+		ack.AssertNumberOfCalls(t, "Ack", 0)
+		ack.AssertNumberOfCalls(t, "Commit", 0)
+		coord.AssertNumberOfCalls(t, "ReExec", 1)
```
kaanyalti: Should there be an assertion for `PerformAction`? If not I'm not sure how the test is asserting that endpoint is notified.

```suggestion
		coord.AssertNumberOfCalls(t, "ReExec", 1)
		coord.AssertNumberOfCalls(t, "PerformAction", 1)
```

kaanyalti: I ran the the test with a mock performAction func in notifyComponents call as shown here 
```
func (h *Migrate) notifyComponents(ctx context.Context, migrateAction *fleetapi.ActionMigrate) error {
	tempFunc := func(context.Context, component.Component, component.Unit, string, map[string]interface{}) (map[string]interface{}, error) {
		return nil, nil
	}
	state := h.coord.State()
	ucs := findMatchingUnitsByActionType(state, fleetapi.ActionTypeMigrate)
	if len(ucs) > 0 {
		// err := notifyUnitsOfProxiedAction(ctx, h.log, migrateAction, ucs, h.coord.PerformAction)
		err := notifyUnitsOfProxiedAction(ctx, h.log, migrateAction, ucs, tempFunc)
		if err != nil {
			return err
		}
	} else {
		// Log and continue
		h.log.Debugf("No components running for %v action type", fleetapi.ActionTypeMigrate)
	}

	return nil
}
```
and the test passes without the `PerformAction` call count assertion

michalpristas: thanks kaan

michalpristas: i removed this specific test as responsibility shifted in between components

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
michalpristas: we now treat notification as required before marking migration a success. 
in case we fail to notify we restore previous config and stay with current cluster.

michalpristas:
just tested this with endpoint changes in place, we don't see ES failures anymore

PR #9864, opened 2025-09-10
Allow overriding AGENT_PACKAGE_VERSION and MANIFEST_URL when USE_PACKAGE_VERSION=true

pchila:
## What does this PR do?
This PR allows overriding `AGENT_PACKAGE_VERSION` and `MANIFEST_URL` when packaging using the information in `package.version` file. This is achieved by loading the package.version data before looking up the relevant env vars and using the values assigned from package version as fallbacks if the env vars are not defined.

## Why is it important?
This allows for easy repackaging elastic-agent from a working directory, in order to have 2 "different" versions for the elastic-agent package: this is useful mostly when testing some upgrade scenarios.
Most of the time `USER_PACKAGE_VERSION=true` by itself will do a fine job for packaging so the usage of this change should be limited to some very specific testing.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/dev-tools/mage/settings.go
+++ b/dev-tools/mage/settings.go
@@ -162,17 +162,18 @@ func initGlobals() {
 
 	versionQualifier, versionQualified = os.LookupEnv("VERSION_QUALIFIER")
 
-	agentPackageVersion = EnvOr(agentPackageVersionEnvVar, "")
-
-	ManifestURL = EnvOr(ManifestUrlEnvVar, "")
-	PackagingFromManifest = ManifestURL != ""
-
 	// order matters this must be called last as it will override some of the
 	// values above
 	err = initPackageVersion()
 	if err != nil {
 		panic(fmt.Errorf("failed to init package version: %w", err))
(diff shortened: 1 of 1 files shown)


dev-tools/mage/settings.go
```diff
@@ -162,17 +162,18 @@ func initGlobals() {
 
 	versionQualifier, versionQualified = os.LookupEnv("VERSION_QUALIFIER")
 
-	agentPackageVersion = EnvOr(agentPackageVersionEnvVar, "")
-
-	ManifestURL = EnvOr(ManifestUrlEnvVar, "")
-	PackagingFromManifest = ManifestURL != ""
-
 	// order matters this must be called last as it will override some of the
 	// values above
```
pchila: This comment probably needs to be reworded now
```suggestion
	// order matters: this will override some of the values. Those values can be used 
	// as fallback for the variables below (mainly agentPackageVersion and ManifestURL)
```

PR #9865, opened 2025-09-10
Add ingest-docs team as CODEOWNERS for release notes and docset.yml

colleenmcginnis:
Related to #9440

Adds ingest-docs team as CODEOWNERS for `docs/release-notes` and `docset.yml`. 

In my experience, someone with admin rights in this repo will need to add the `@elastic/ingest-docs` team with write permissions even if all of elastic already has write access. 

>When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

From: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners


ebeahan (approved):
LGTM - @elastic/ingest-docs should now have write permissions.

PR #9886, opened 2025-09-11
fix: ensure EDOT subprocess shuts down gracefully on agent termination

pkoutsovasilis:
## What does this PR do?

Ensures Elastic Agent **gracefully** shuts down the EDOT (collector) subprocess when the agent receives a terminating signal (e.g., `SIGTERM`) - i.e., when the `OTelManager` context is cancelled - instead of immediately killing it.

Key changes:
- Replace `collectorHandle.Stop(ctx context.Context)` with `Stop(waitTime time.Duration)` and honor it in both **embedded** and **subprocess** execution modes.
- Set `waitTimeForStop` to **30s** (aligned with Beats subprocess defaults).
- Add warnings when the supervised collector fails to stop gracefully or times out.
- Make `updateCh` channel buffered to a size of **1** and drain before write (pattern used elsewhere) to avoid shutdown/reconfig delays.
- Unit-Tests:
  - Extend test harness to allow configurable shutdown delays via `TEST_SUPERVISED_COLLECTOR_DELAY`.
  - Validate both outcomes:
    - Delay > 30s ⇒ subprocess is force-killed.
    - Delay < 30s ⇒ subprocess exits cleanly within the timeout.

## Why is it important?

Previously, on shutdown the agent could terminate in a way that **kills EDOT immediately**, risking incomplete cleanup of telemetry pipelines and leaving resources in a bad state. Waiting (up to 30s) for a graceful EDOT exit improves cleanup, stability, and hybrid agent resilience.

## Checklist

## Disruptive User Impact

No disruptive impact expected. The only behavior change is during shutdown: the agent now waits up to **30s** for EDOT to exit gracefully before force-killing it. This improves stability and predictability for users..

```
mage unitTest
```

## Related issues

N/A

### DIFF
--- a/internal/pkg/otel/manager/manager.go
+++ b/internal/pkg/otel/manager/manager.go
@@ -34,8 +34,6 @@ type ExecutionMode string
 const (
 	SubprocessExecutionMode ExecutionMode = "subprocess"
 	EmbeddedExecutionMode   ExecutionMode = "embedded"
-	// waitTimeForStop is the time to wait for the collector to stop before killing it.
-	waitTimeForStop = 30 * time.Second
 )
 
 type collectorRecoveryTimer interface {
@@ -106,6 +104,9 @@ type OTelManager struct {
 
 	// collectorRunErr is used to signal that the collector has exited.
 	collectorRunErr chan error
+
+	// stopTimeout is the timeout to wait for the collector to stop.
+	stopTimeout time.Duration
 }
 
 // NewOTelManager returns a OTelManager.
@@ -116,6 +117,7 @@ func NewOTelManager(
 	mode ExecutionMode,
 	agentInfo info.Agent,
 	beatMonitoringConfigGetter translate.BeatMonitoringConfigGetter,
+	stopTimeout time.Duration,
 ) (*OTelManager, error) {
 	var exec collectorExecution
 	var recoveryTimer collectorRecoveryTimer
@@ -154,6 +156,7 @@ func NewOTelManager(
 		execution:                  exec,
 		recoveryTimer:              recoveryTimer,
 		collectorRunErr:            make(chan error),
+		stopTimeout:                stopTimeout,
 	}, nil
 }
 
@@ -176,7 +179,7 @@ func (m *OTelManager) Run(ctx context.Context) error {
 			// our caller context is cancelled so stop the collector and return
 			// has exited.
 			if m.proc != nil {
-				m.proc.Stop(waitTimeForStop)
+				m.proc.Stop(m.stopTimeout)
 			}
 			return ctx.Err()
 		case <-m.recoveryTimer.C():
@@ -205,7 +208,7 @@ func (m *OTelManager) Run(ctx context.Context) error {
 			if err == nil {
 				// err is nil means that the collector has exited cleanly without an error
 				if m.proc != nil {
-					m.proc.Stop(waitTimeForStop)
+					m.proc.Stop(m.stopTimeout)
 					m.proc = nil
 					updateErr := m.reportOtelStatusUpdate(ctx, nil)
 					if updateErr != nil {
@@ -248,7 +251,7 @@ func (m *OTelManager) Run(ctx context.Context) error {
 				// in the case that the configuration is invalid there is no reason to
 				// try again as it will keep failing so we do not trigger a restart
 				if m.proc != nil {
-					m.proc.Stop(waitTimeForStop)
+					m.proc.Stop(m.stopTimeout)
 					m.proc = nil
 					// don't wait here for <-collectorRunErr, already occurred
 					// clear status, no longer running
@@ -343,7 +346,7 @@ func buildMergedConfig(cfgUpdate configUpdate, agentInfo info.Agent, monitoringC
 
 func (m *OTelManager) applyMergedConfig(ctx context.Context, collectorStatusCh chan *status.AggregateStatus, collectorRunErr chan error) error {
 	if m.proc != nil {
-		m.proc.Stop(waitTimeForStop)
+		m.proc.Stop(m.stopTimeout)
 		m.proc = nil
 		select {
 		case <-collectorRunErr:
--- a/internal/pkg/otel/manager/manager_test.go
+++ b/internal/pkg/otel/manager/manager_test.go
@@ -289,6 +289,8 @@ func TestOTelManager_Run(t *testing.T) {
 	testBinary := filepath.Join(wd, "testing", "testing")
 	require.FileExists(t, testBinary, "testing binary not found")
 
+	const waitTimeForStop = 30 * time.Second
+
 	for _, tc := range []struct {
 		name                string
 		execModeFn          func(collectorRunErr chan error) (collectorExecution, error)
@@ -685,6 +687,7 @@ func TestOTelManager_Run(t *testing.T) {
 				doneChan:          make(chan struct{}),
 				collectorRunErr:   make(chan error),
 				recoveryTimer:     tc.restarter,
+				stopTimeout:       waitTimeForStop,
 			}
 
 			executionMode, err := tc.execModeFn(m.collectorRunErr)
@@ -737,9 +740,11 @@ func TestOTelManager_Run(t *testing.T) {
 func TestOTelManager_Logging(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	const waitTimeForStop = 30 * time.Second
+
 	base, obs := loggertest.New("otel")
 	l, _ := loggertest.New("otel-manager")
-	m, err := NewOTelManager(l, logp.DebugLevel, base, EmbeddedExecutionMode, nil, nil)
+	m, err := NewOTelManager(l, logp.DebugLevel, base, EmbeddedExecutionMode, nil, nil, waitTimeForStop)
 	require.NoError(t, err, "could not create otel manager")
 
 	go func() {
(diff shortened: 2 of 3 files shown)


internal/pkg/otel/manager/manager.go
```diff
@@ -131,7 +136,7 @@ func NewOTelManager(
 		recoveryTimer = newRestarterNoop()
 		exec = newExecutionEmbedded()
 	default:
-		return nil, errors.New("unknown otel collector exec")
+		return nil, errors.New("unknown otel collector execModeFn")
```
pchila: Bad rename? Or is this intentional? Moreover, it could be useful to include which execution mode was not supported, what about:
```suggestion
		return nil, fmt.Errorf("unknown otel collector execution mode: %q", mode )
```

internal/pkg/otel/manager/manager.go
```diff
@@ -402,6 +405,15 @@ func (m *OTelManager) Update(cfg *confmap.Conf, components []component.Component
...
 		components:   components,
 	}
+
+	// we care only about the latest config update
+	select {
+	case <-m.updateCh:
+	case <-m.doneChan:
+		return
+	default:
+	}
```
pchila, line 418: I understand that we are trying to drain the `m.updateCh` channel here, but is having `case <- m.doneChan` on the same select correct here? IIUC there's no deterministic way to prioritize one over the other if both channels can be read from.

internal/pkg/otel/manager/testing/testing.go
pchila, line 1: Can we add a brief comment explaining what is the purpose and the behavior of this test binary ?

internal/pkg/otel/manager/manager.go
pkoutsovasilis: Hey Paolo 👋 - good callout! In this spot, prioritisation isn’t the goal. The goal is to make sure the writer (which is not the manager run loop) never blocks, aka prevent a deadlock; either because the manager is already done, or because we were able to drain one pending config from m.updateCh and/or written the new one. If both channels are readable, it doesn’t matter which fires - if we’re done, we return; otherwise we consume one update. We’re not relying on deterministic ordering here, wdyt?

internal/pkg/otel/manager/manager.go
pkoutsovasilis: changed in here 29c9317c0e130799e1d3dcd44e6eadead786f562

internal/pkg/otel/manager/testing/testing.go
pkoutsovasilis: added here cb65d3684f65d2f6d93f236fcf2fe6971ccf2e81

internal/pkg/otel/manager/manager.go
pchila: Makes sense. If it's just to keep the writer from blocking then it makes sense to me.

internal/pkg/otel/manager/testing/testing.go
pchila: Thank you :smile:

pchila (dismissed):
LGTM

internal/pkg/otel/manager/testing/testing.go
```diff
@@ -28,9 +45,15 @@ func main() {
 		})
 	}
 
-	err := cmd.RunCollector(ctx, nil, true, "debug")
+	err = cmd.RunCollector(ctx, nil, true, "debug")
 	if err == nil || errors.Is(err, context.Canceled) {
+		if shutdownDelay > 0 {
+			<-time.After(shutdownDelay)
+		}
```
swiatekm: Can't we move this check above the conditional instruction? It's identical in both branches.

internal/pkg/otel/manager/manager.go
```diff
@@ -402,6 +405,15 @@ func (m *OTelManager) Update(cfg *confmap.Conf, components []component.Component
 		collectorCfg: cfg,
 		components:   components,
 	}
+
+	// we care only about the latest config update
+	select {
```
swiatekm, line 413: Is the caller waiting for 30 seconds, blocked on the update channel, while the collector process exits, what we're trying to avoid here? I don't mind the solution, but whenever I see this pattern, I wonder if there's an architecturally more elegant way of dealing with this. Or, if it isn't simpler to have a larger buffered channel and return a serious error if it's full. WDYT? Mostly thinking out loud, this shouldn't be a blocker for merging.

internal/pkg/otel/manager/manager_test.go
```diff
@@ -479,14 +480,128 @@ func TestOTelManager_Run(t *testing.T) {
...
+				return &testExecution{exec: subprocessExec}, nil
+			},
+			restarter: newRecoveryBackoff(100*time.Nanosecond, 10*time.Second, time.Minute),
+			testFn: func(t *testing.T, m *OTelManager, e *EventListener, exec *testExecution, managerCtxCancel context.CancelFunc, collectorRunErr chan error) {
+				delayDuration := 40 * time.Second // the otel manager stop timeout is waitTimeForStop (30 seconds)
+				err := os.Setenv("TEST_SUPERVISED_COLLECTOR_DELAY", delayDuration.String())
+				require.NoError(t, err, "failed to set TEST_SUPERVISED_COLLECTOR_DELAY env var")
+				t.Cleanup(func() {
+					_ = os.Unsetenv("TEST_SUPERVISED_COLLECTOR_DELAY")
+				})
```
swiatekm: I think using t.SetEnv would give the same effect with less code.

internal/pkg/otel/manager/manager_test.go
```diff
@@ -479,14 +480,128 @@ func TestOTelManager_Run(t *testing.T) {
...
+				return &testExecution{exec: subprocessExec}, nil
+			},
+			restarter: newRecoveryBackoff(100*time.Nanosecond, 10*time.Second, time.Minute),
+			testFn: func(t *testing.T, m *OTelManager, e *EventListener, exec *testExecution, managerCtxCancel context.CancelFunc, collectorRunErr chan error) {
+				delayDuration := 5 * time.Second // the otel manager stop timeout is waitTimeForStop (30 seconds)
+				err := os.Setenv("TEST_SUPERVISED_COLLECTOR_DELAY", delayDuration.String())
+				require.NoError(t, err, "failed to set TEST_SUPERVISED_COLLECTOR_DELAY env var")
+				t.Cleanup(func() {
+					_ = os.Unsetenv("TEST_SUPERVISED_COLLECTOR_DELAY")
+				})
```
swiatekm: See above.

swiatekm (dismissed):
LGTM, some minor nitpicks and a non-blocking question.

internal/pkg/otel/manager/manager.go
```diff
@@ -34,6 +34,8 @@ type ExecutionMode string
 const (
 	SubprocessExecutionMode ExecutionMode = "subprocess"
 	EmbeddedExecutionMode   ExecutionMode = "embedded"
+	// waitTimeForStop is the time to wait for the collector to stop before killing it.
+	waitTimeForStop = 30 * time.Second
```
cmacknz: We probably want this to be configurable in case it turns out to be too long in some circumstances. Not urgent or blocking, we can do this later.

internal/pkg/otel/manager/manager.go
pkoutsovasilis: @swiatekm you have mentioned this pattern a couple of times now, so let me clarify the reasoning again since I think we might be viewing this differently.

The best-effort draining of a channel before writing to it serves perfectly the case where we only care about the latest message (otel configuration in our case). While this pattern might initially resemble Kafka producers and consumers, it's fundamentally different - we have a single writer (coordinator run loop) and a single reader (manager run loop) with the specific guarantee that the writer won't block and the reader only cares about the latest value.

Now, regarding the buffered channel alternative: the moment you make updateCh a larger buffered channel, you open up a scenario where the manager run loop accumulates many config updates and keeps looping through only those without reacting to anything else - which would be a bug. The manager needs to stay responsive to other events, not get stuck processing a backlog of stale updates.

I am not quite sure why this seems inelegant to you, but it serves our specific architectural needs well: non-blocking writes with latest-value-wins semantics. For these reasons, I'll keep the capacity at 1. WDYT?

internal/pkg/otel/manager/testing/testing.go
pkoutsovasilis: moved here b3a06f0bc0ec5848d73ee90def877277c39bc9ef

internal/pkg/otel/manager/manager_test.go
pkoutsovasilis: changed in 1341be9ded782322df10b3fcbc0ea0ddf5850f7b

internal/pkg/otel/manager/manager_test.go
pkoutsovasilis: changed in 1341be9ded782322df10b3fcbc0ea0ddf5850f7b

internal/pkg/otel/manager/manager.go
pkoutsovasilis: how about this c99794996e768f038b485939b1cd02aafd5db724?

cmacknz:
LGTM, @swiatekm can do the approving here since he still has unresolved comments.

internal/pkg/otel/manager/manager.go
swiatekm: I agree that all of this works if the assumptions you list are correct. I don't think they're documented, though, and it would be both reasonable and easy to miss for the coordinator to start relying on the fact that each config update should result in a response over the error channel, for example. But if we can drop some config updates, this isn't actually true, and we should document it and find some way to verify we're not depending on the otel manager having semantics it doesn't actually have.

My lack of comfort here has less to do with the specific behavior you're introducing, and more with the ambiguity of the guarantees the otel manager gives its users, if that makes more sense.

But I don't want to hold this PR up, we can move this discussion to a better venue.

Issue #8032, opened 2025-04-29
[beats receivers] OTel based pipeline must continue to report queue metrics for the elastic_agent integration


cmacknz:
The primary queue when using Beats receivers is the one in the collector. We need to export queue metrics as part of self-monitoring. 

<img width="854" alt="Image" src="https://github.com/user-attachments/assets/dc8ce152-ce65-40ae-b595-28670530752c" />

Ideally they are backwards compatible with the existing metrics for queue depth. If they cannot have the same names then the same semantic meaning.

`libbeat.pipeline.events.active` measures the number of events in the processing pipeline including those in processors, and has a value that can be slightly above the queue size because of this.

https://github.com/elastic/integrations/blob/037cda7c46f84ca7c1d8055105f5d7fab59fda50/packages/elastic_agent/data_stream/filebeat_logs/fields/fields.yml#L48-L57

```yaml
- name: monitoring.metrics.libbeat.pipeline
  type: group
  fields:
    - name: events
      type: group
      fields:
        - name: active
          type: long
          metric_type: gauge
          description: Number of events currently in the pipeline. The maximum size is the configured queue size plus one event per event pipeline client.
```

Direct measurements of the data in the queue were added recently, specifically with `queue.filled.pct.events` giving the queue fill level in percent directly.

https://github.com/elastic/integrations/blob/037cda7c46f84ca7c1d8055105f5d7fab59fda50/packages/elastic_agent/data_stream/filebeat_logs/fields/fields.yml#L66-L77

```yaml
    - name: queue.acked
      type: long
      metric_type: counter
      description: Number of events successfully acknowledged by the output
    - name: queue.max_events
      type: long
      metric_type: counter
      description: Maximum number of events in a queue
    - name: queue.filled.pct.events
      type: float
      metric_type: gauge
      description: Maximum number of events in a queue
```

leehinman:
# Short term
Get the current data at https://opentelemetry.io/docs/collector/internal-telemetry/ and get it displayed in the elastic-agent dashboards

# Longer term
Change what is in https://opentelemetry.io/docs/collector/internal-telemetry/, so we can better provide info like queue pct filled.

cmacknz:
Likely this will require https://github.com/elastic/elastic-agent/issues/8970, the internal telemetry will have the OTLP data format and we'll need to put that in a metrics datastream with the .otel suffix.

cmacknz:
[...]

Issue #8970, opened 2025-07-11
[beats receivers] Introduce new metrics-elastic_agent.collector monitoring data stream


cmacknz:
Introduce a new monitoring data stream for metrics collected from the embedded or supervised OpenTelemetry collector:

* metrics-elastic_agent.collector.otel-* (may need to be elastic_agent/collector.otel) for OTLP formatted metrics collected from the OpenTelemetry collector.
* metrics-elastic_agent.collector-* for ECS formatted metrics collected from the OpenTelemetry collector. Ideally we can avoid needing both of these but if we do, this will be the convention.

In general follow the convention that OTLP formatted documents have the dataset suffixed with `.otel`. There are default mappings for .otel suffixed data streams [here](https://github.com/elastic/elasticsearch/tree/b52c6f7f27a1ed802e99df617b7a6bb505060b58/x-pack/plugin/otel-data/src/main/resources/component-templates) that may cover this index, if not or if we want to map additional information we will need to define it in the [elastic_agent](https://github.com/elastic/integrations/tree/main/packages/elastic_agent) package

Acceptance Criteria
* A test exists confirming that metrics from the collector are ingested into the desired index(es), and that metrics for specific component IDs and names can be queried.

cmacknz:
One thing to note is `metrics-*.otel-*` datastreams will have attributes define TSDB dimensions by default so each unique set of attributes is a separate time series.

cmacknz:
We'll need to add this to the list of known monitoring indices in Fleet: https://github.com/elastic/kibana/blob/eed21dbca87029b245c44f5d8d4d64c9f491c877/x-pack/platform/plugins/shared/fleet/common/constants/agent_policy.ts#L17-L35

This is in addition to defining the data stream(s) in the elastic_agent integration.

cmacknz:
Same comment from https://github.com/elastic/elastic-agent/issues/8969#issuecomment-3348635637 applies here

cmacknz:
Closing, I'm not sure this is exactly what we'll do anymore.

PR #9928, opened 2025-09-12
Ingest internal telemetry from the OTel Collector when it is running

faec:
Add a monitoring component to ingest telemetry values from the collector's prometheus endpoint into the `elastic_agent.collector` dataset. This includes a remapping script processor to convert the collector's metrics to their closest ECS equivalents when possible.

This is a temporary approach that will not scale to full collector metrics in a general configuration, but gives us a backwards-compatible stopgap during the initial monitoring-only rollout.

## Checklist

Enable agent monitoring while at least one component (including monitoring itself) uses the otel runtime. Prometheus metrics from the collector should be ingested to the `elastic_agent.collector` dataset.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/8032

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -319,7 +319,7 @@ func TestClassicAndReceiverAgentMonitoring(t *testing.T) {
 		var statusErr error
 		status, statusErr := beatReceiverFixture.ExecStatus(ctx)
 		assert.NoError(collect, statusErr)
-		assertBeatsHealthy(collect, &status, component.OtelRuntimeManager, 3)
+		assertBeatsHealthy(collect, &status, component.OtelRuntimeManager, 4)
 		return
 	}, 1*time.Minute, 1*time.Second)
 


internal/pkg/agent/application/monitoring/component/v1_monitor.go
```diff
@@ -486,6 +512,47 @@ func (b *BeatsMonitor) monitoringNamespace() string {
 	return defaultMonitoringNamespace
 }
 
+func (b *BeatsMonitor) getCollectorTelemetryEndpoint() string {
+	type metricsReaderConfig struct {
```
faec: This was the least verbose way I could find to extract the prometheus endpoint configuration (the mapstructure annotations don't seem to accept dotted field names), but I'd be happy to learn if there's a better one.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
```diff
@@ -486,6 +512,47 @@ func (b *BeatsMonitor) monitoringNamespace() string {
...
+				} `mapstructure:"metrics"`
+			} `mapstructure:"telemetry"`
+		} `mapstructure:"service"`
+	}
+	if err := b.otelConfig.Unmarshal(&otel, confmap.WithIgnoreUnused()); err == nil {
+		for _, reader := range otel.Service.Telemetry.Metrics.Readers {
+			p := reader.Pull.Exporter.Prometheus
+			if p.Host != "" || p.Port != 0 {
+				host := "localhost"
+				port := 8888
```
cmacknz: Is there any way to get this to use a named pipe/unix socket instead?

If we can't do that, the port will need to be configurable and to keep the `--develop` option working to allow two agents on the same host is has to be a random free port. https://github.com/pkoutsovasilis/elastic-agent/blob/c99794996e768f038b485939b1cd02aafd5db724/internal/pkg/agent/configuration/grpc.go#L41-L46

If we could just default this to a random free port initially it would be easiest, with the option to specify a fixed for cases where users need to work around local firewall rules.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
```diff
@@ -529,6 +596,20 @@ func (b *BeatsMonitor) injectMetricsInput(
 		},
 	}
 
+	if usingOtelRuntime(componentInfos) {
+		prometheusStream := b.getPrometheusStream(failureThreshold, metricsCollectionIntervalString)
+		inputs = append(inputs, map[string]interface{}{
+			idKey:        fmt.Sprintf("%s-collector", monitoringMetricsUnitID),
+			"name":       fmt.Sprintf("%s-collector", monitoringMetricsUnitID),
+			"type":       "prometheus/metrics",
```
cmacknz, line 600: Using the Metricbeat prometheus module I think is faster, but this means we'll be using ECS instead of SemConv and the OTLP schema for data from the collector which is our eventual end goal.

Did you look at using the upstream prometheus receiver to do this instead? The data format will be different (OTLP) and you'll need to use processors to set the target data stream.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -730,6 +811,40 @@ func (b *BeatsMonitor) getHttpStreams(
...
+func (b *BeatsMonitor) getPrometheusStream(
+	failureThreshold *uint,
+	metricsCollectionIntervalString string,
+) any {
+	prometheusHost := b.getCollectorTelemetryEndpoint()
+
+	// want metricset "collector"
+	monitoringNamespace := b.monitoringNamespace()
+	indexName := fmt.Sprintf("metrics-elastic_agent.collector-%s", monitoringNamespace)
+	dataset := "elastic_agent.collector"
```
cmacknz: You need https://github.com/elastic/elastic-agent/issues/8970 (this issue specifies an OTLP formatted data stream as is which would require the prometheus receiver) for this to be writable by Fleet managed agents I think. Right now I would expect the agent API key from Fleet to not have permissions to create or index to this data stream.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
cmacknz: The more I think about this the more I think we only have two ways to do this:

1. We use either of the prometheus/metrics module or the prometheus receiver, but we need to process the data so the field names are the same as what they are today without introducing any new data streams. This is collecting the data from a different place but keeping it compatible.
2. We use only the prometheus receiver and ship the data in OTLP format to a new datastream. This is our end goal and is a breaking change to monitoring so can't go in a patch release.

Given the timeline for this I think we have to go with approach one here and try to get the data to look like it does now for the queue and outputs.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
cmacknz: We might be able to get support for unix sockets by using gRPC instead, see https://github.com/elastic/elastic-agent/pull/9963 for an example which is enabling internal telemetry for the EDOT collector (`elastic-agent otel` mode)

internal/pkg/agent/application/monitoring/component/v1_monitor.go
cmacknz: We need to prototype option 1 ASAP, because if we can transform the internal telemetry pieces we need to keep the elastic_agent dashboard mostly working as is with the existing ECS data we can take the time we need to polish presenting the underlying OTLP data directly to users.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
swiatekm: Any reason not to use the official [config struct ](https://github.com/open-telemetry/opentelemetry-collector/blob/ba0b327d5fe91b230f97e48f56f3c191bd89df2d/service/config.go#L13)instead?

internal/pkg/agent/application/monitoring/otel_remap.js
```diff
@@ -0,0 +1,115 @@
...
+  // when combined with the overall success/failure totals for the exporter.
+  if (received_docs_count != null && total_sent_valid && total_failed_valid) {
+    var active_count = received_docs_count - total_sent - total_failed;
+    event.Put("beat.stats.libbeat.output.events.active", active_count);
+    keep_event = true;
+  }
+
+  if (!keep_event) {
+    event.Put("event.cancelled_flag", 1);
+    //event.Cancel();
```
michalpristas: probably leftover (?)

internal/pkg/agent/application/monitoring/component/v1_monitor.go
```diff
@@ -435,6 +452,16 @@ func (b *BeatsMonitor) getComponentInfos(components []component.Component, compo
 			RuntimeManager: component.RuntimeManager(b.config.C.RuntimeManager),
 		})
 	}
+	// If any other component uses the Otel runtime, also add a component to monitor
+	// its telemetry.
+	if b.config.C.MonitorMetrics && usingOtelRuntime(componentInfos) {
```
michalpristas, line 462: you can detect this in a loop above, no need to loop it again, although cleaner

internal/pkg/agent/application/monitoring/component/v1_monitor.go
michalpristas: i think with 8888 we'll run into a lot of conflicts, i see sockets much safer

faec: Sockets would be nicer but it doesn't seem possible with the current Collector code? Unless I misunderstood the outcome of the beats receiver meetings, that example might use grpc but collector configs don't yet support non-http urls (and in any case that wouldn't work with metricbeat -- though hopefully we'll be dumping the metricbeat approach for 9.3 anyway).

I agree 8888 will run into conflicts, but that's already a problem independent of this PR (which is why this PR checks for alternate port configs).

faec:
This has been heavily edited to prioritize backwards compatibility for 9.2, and I think is now ~as close as we can come to that without an entirely different approach [which I think we should start work on immediately, but won't be ready for 9.2].

internal/pkg/agent/application/monitoring/component/v1_monitor.go
graphaelli: I've been down this rabbit hole a few times - https://github.com/prometheus/prometheus/issues/12024 implies it's open for discussion but not implemented.  https://github.com/prometheus/node_exporter/issues/759 for example uses this as a reason not to implement support

cmacknz: Using a fixed port is basically unreleasable IMO and would be a breaking change for anyone that already has something on this port.

We can use whatever transport we have to but we have to get away from specifying the fixed port here and requiring user intervention to change it. We would have to build a UI component to change the port in Fleet for example for that to even be viable.

cmacknz: > I've been down this rabbit hole a few times - prometheus/prometheus#12024 implies it's open for discussion but not implemented.

This almost looks like it's just waiting for someone to say they are willing implement it.

cmacknz: OK we already have this solved elsewhere since we control both ends of this:

https://github.com/elastic/elastic-agent/blob/cb9c826c8e66e75abdf50cbee635c61538b79f23/internal/pkg/otel/manager/execution_subprocess.go#L90-L93

Find a random free port and then configure the collector and receiver to use it.

We will want the ability to specify a fixed port for anyone that needs to deal with local iptables or firewall rules which is rare but has happened to the gRPC control protocol port which is also still on localhost. Just this time we want a random port to be the default instead.

cmacknz: https://github.com/elastic/elastic-agent/issues/10206 because we didn't allow specifying a fixed port in the healthcheck extension.

internal/pkg/agent/application/monitoring/component/otel_remap.js
```diff
@@ -0,0 +1,139 @@
...
+    event.Cancel();
+    return;
+  }
+
+  // Hack: if the scope is elastic-custom fields, deterministically mangle the
+  // agent.id. Since the label set is different, these are passed through in
+  // different events, and if we don't do this one of the events will be
+  // rejected as a duplicate since they have the same component id, agent id,
+  // and metricset.
+  var id = event.Get("agent.id");
```
swiatekm, line 24: Doesn't this effectively mean `agent.id` is wrong for this data? 

Relatedly, can you elaborate on why exactly this is rejected as a duplicate?

faec: (Discussed in the receiver sync but):

Yes, `agent.id` will be "wrong" but it has no semantic content, it's just a random-but-consistent UUID for the metricbeat that generated the metrics in this event. Here we modify the id deterministically by incrementing the last digit, so this essentially makes it look like there are metrics coming from two different metricbeat instances. (The original unmangled value is also still present in the event metadata in other fields, so it isn't really losing data either.)

If we don't do this, events can be rejected as duplicates by Elasticsearch because the prometheus fields are split into multiple events by the source Metricbeat, but the [dimension fields](https://github.com/elastic/integrations/blob/main/packages/elastic_agent/data_stream/metricbeat_metrics/fields/ecs.yml) for the datastream we're targeting are all the same. `agent.id` is the only dimension field with no semantic content that would affect dashboard visualizations.

(All that said, I agree this is extremely undesirable and we should urgently figure out a better solution for the next release, this is just to buy us time during the initial monitoring rollout.)

internal/pkg/agent/application/monitoring/component/v1_monitor.go
faec: Unless you feel strongly I'd rather leave it separate, there's no harm in looping again and I'd rather keep the logic in a standalone helper with clear semantics than interleave the checks with the other work of this function.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
faec: Context from this morning's meeting: short-term @swiatekm will look into addressing the fixed-port issue as a bug fix independent of this change by having Agent pick a random port for telemetry when generating the collector config, and medium-term we will look for alternatives that don't require an open TCP port at all.

internal/pkg/agent/application/monitoring/component/otel_remap.js
ebeahan: > All that said, I agree this is extremely undesirable and we should urgently figure out a better solution for the next release

If we haven't already, can we file an issue to track?

ebeahan (dismissed):
LGTM with one nit.

Approving to unblock the reviewer gate, but there's still unit test failures in CI which need resolved.

faec:
> If we haven't already, can we file an issue to track?

https://github.com/elastic/elastic-agent/issues/10220

ycombinator (approved):

internal/pkg/agent/application/monitoring/v1_monitor.go
cmacknz: https://github.com/elastic/elastic-agent/issues/10277

PR #9931, opened 2025-09-12
Add agent_policy_id and policy_revision_idx to checkin requests

### LINKED ISSUE #6446: Fleet check-in should send policy_id and revision
## Overview

Currently when the Elastic Agent checks-in with Fleet Server it doesn't send the `policy_id` or `revision` of the policy that it is currently running. The Fleet Server stores this information by the fact that the Elastic Agent ACK'd the policy change notification, but there are many cases where this could be come out of sync.

### VM Snapshot

1. VM is snapshotted
2. new policy revision occurs
3. ACK'd by Elastic Agent (stored new revision in Fleet)
4. VM is rolled back

Now the running Elastic Agent policy is the old version, but to Fleet it is the new version.

### Bad Error Case

This is just a weird case but a coding issue could result in this problem.

1. New revision is sent to Elastic Agent
2. Policy failed to be saved to disk (could be coding issue or just with filesystem problem)
3. policy revision is ACK'd anyway (shouldn't happen, but if it does...)

Elastic Agent is now running old version of policy but Fleet Server believes that its the new revision

### Backup/restore of `fleet.enc`

In the case of backup/restore of `fleet.enc`.

1. `fleet.enc` is backed up
2. new policy revision occurs
3. ACK'd by Elastic Agent (stored new revision in Fleet)
4. `fleet.enc` is replaced with backup from 1
5. Elastic Agent restarted

Elastic Agent is now running old version of policy but Fleet Server believes that its the new revision

## How to solve it?

Upon check-in the Elastic Agent should be sending its current policy ID and revision. That is then compared to what Fleet Server expects and if it is not correct then it sends the correct policy.

michel-laterman:
## What does this PR do?

Add the agent_policy_id and policy_revision_idx attributes to checkin requests.
These attributes are sources from the action stored as a part of the state.
Add a feature flag to disable sending acks for policy change actions; behaviour for policy change acks has not been changed with this addition (they are always sent).

## Why is it important?

The policy information in fleet-server and agent may go out of sync; this may occur in cases where a VM restores from a snapshot.

## Checklist

## Disruptive User Impact

N/A

## Related issues

- Closes #6446

### DIFF
--- a/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
@@ -473,7 +473,6 @@ type policyChange struct {
 	cfg        *config.Config
 	action     fleetapi.Action
 	acker      acker.Acker
-	commit     bool
 	ackWatcher chan struct{}
 }
 
@@ -482,9 +481,9 @@ func newPolicyChange(
 	config *config.Config,
 	action fleetapi.Action,
 	acker acker.Acker,
-	commit bool) *policyChange {
+	makeCh bool) *policyChange {
 	var ackWatcher chan struct{}
-	if commit {
+	if makeCh {
 		// we don't need it otherwise
 		ackWatcher = make(chan struct{})
 	}
@@ -493,7 +492,6 @@ func newPolicyChange(
 		cfg:        config,
 		action:     action,
 		acker:      acker,
-		commit:     true,
 		ackWatcher: ackWatcher,
 	}
 }
@@ -502,22 +500,21 @@ func (l *policyChange) Config() *config.Config {
 	return l.cfg
 }
 
+// Ack sends an ack for the associated action if the results are expected.
+// An ack will not be sent for a POLICY_CHANGE action, but will be when this method is used by UNENROLL actions.
 func (l *policyChange) Ack() error {
-	if l.action == nil {
+	if l.action == nil || l.ackWatcher == nil {
 		return nil
 	}
 	err := l.acker.Ack(l.ctx, l.action)
 	if err != nil {
 		return err
 	}
-	if l.commit {
-		err := l.acker.Commit(l.ctx)
-		if l.ackWatcher != nil && err == nil {
-			close(l.ackWatcher)
-		}
-		return err
+	err = l.acker.Commit(l.ctx)
+	if err == nil {
+		close(l.ackWatcher)
 	}
-	return nil
+	return err
 }
 
 // WaitAck waits for policy change to be acked.
@@ -525,7 +522,7 @@ func (l *policyChange) Ack() error {
 // Caller is responsible to use any reasonable deadline otherwise
 // function call can be endlessly blocking.
 func (l *policyChange) WaitAck(ctx context.Context) {
-	if !l.commit || l.ackWatcher == nil {
+	if l.ackWatcher == nil {
 		return
 	}
 
--- a/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
+++ b/internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
@@ -29,6 +29,7 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/fleetapi/client"
 	"github.com/elastic/elastic-agent/internal/pkg/remote"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
+	"github.com/elastic/elastic-agent/pkg/features"
 )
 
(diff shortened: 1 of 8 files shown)


michel-laterman:
~~Adding integration/e2e tests requires the fleet-server to be implemented first: https://github.com/elastic/fleet-server/pull/5501
I'll keep this as a draft until the above PR is merged.~~

I'm not changing the default behaviour for the agent with regards to acks.

internal/pkg/agent/application/actions/handlers/handler_action_policy_change.go
```diff
@@ -502,30 +500,29 @@ func (l *policyChange) Config() *config.Config {
 	return l.cfg
 }
 
+// Ack sends an ack for the associated action if the results are expected.
+// An ack will not be sent for a POLICY_CHANGE action, but will be when this method is used by UNENROLL actions.
 func (l *policyChange) Ack() error {
```
michel-laterman, line 511: It may be less confusing to make this a nop, and provide something else for unenroll actions to use

pkg/features/features.go
```diff
@@ -21,6 +21,10 @@ import (
 // 8.11+ - default is enabled
 const defaultTamperProtection = true
 
+// The default value of the disable policy change acks flag if the flag is missing.
+// 9.2 - disabled (acks are sent)
+const defaultDisablePolicyChangeAcks = false
```
michel-laterman, line 26: I've added this as a feature toggle, but we will need to expose the setting through the UI before we can disable acks from the agent.

The fleet-server will do a little extra work on checkins (comparing API key hashes), but this will not result in extra requests being made to ES.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -356,15 +357,21 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
 	// Fix loglevel with the current log level used by coordinator
 	ecsMeta.Elastic.Agent.LogLevel = state.LogLevel.String()
 
+	action := f.stateStore.Action()
```
blakerouse, line 360: This action is always a policy change action? If that is the case would it just be better to have this return that struct type always. Why make it possible to be different types, that could cause the following two function calls the possibility to fail?

pkg/features/features.go
blakerouse: What is the user incentive to go turn this feature flag on?

blakerouse (commented):
Overall this looks good, but I have 2 questions.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
michel-laterman: It can be an `UNENROLL` action if the agent was unenrolled, the existing function assumes it's an action of one of these types

pkg/features/features.go
michel-laterman: I was thinking that we would turn it on by default in a later release (9.3 for example), and allow users to toggle it in case they have issues with the new behaviour.

blakerouse: Okay, so not something that a user will really turn on. More something that we would turn on.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
blakerouse: Ah, that makes sense.

michel-laterman:
https://github.com/elastic/elastic-agent/blob/ab68480e77a3aee4098304c7c03101a8bcdb8b2a/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L161-L176
After fleet checkin, the agent sends all actions through a channel to the dispatcher. They are executed concurrently with the checkin loop; the ticker by default has a 1s duration with up to 500ms jitter.
There is no guarantee that the POLICY_CHANGE action is executed before the next checkin. 
cc @blakerouse

blakerouse:
@michel-laterman Thanks for the clarification from the call today. I don't think this should be an issue with this PR, but we might want to make just the policy change blocking, at least until we know its either applied or not applied. That would really reduce the load on Fleet Server, could be a scale improvement really.

We could do something like:

```golang
ctx, cancel := context.WithTimeout(ctx, 5 * time.Second)
defer cancel()
waitForPolicyApply := f.handleActions(actions)
select {
case <-waitForPolicyApply:
case <-ctx.Done():
}
```

michel-laterman:
Created https://github.com/elastic/elastic-agent/issues/10130 to track

blakerouse:
@michel-laterman Thanks!

PR #9982, opened 2025-09-16
Add support for quicker Fleet check-in upon component status

### LINKED ISSUE #9348: Add support for quicker Fleet check-in upon component status
**Describe the enhancement:**

At the moment the Elastic Agent checks into Fleet every 5 minutes as long as no action occurs on the Fleet-side. If an input is having an issue and is in error or degraded state that information is not relayed to Fleet until at most 5 minutes.

This provides a bad user experience when the issue is some configuration or authentication issue with an external service. The long delay of knowing when things are working correctly makes it very hard to watch an integration and check on its health.

The following configuration would allow this behavior change.

```
fleet:
  enabled: true
  checkin:
    min_timeout: 5s
    max_timeout: 5m
```

The default would stay:

```
fleet:
  enabled: true
  checkin:
    min_timeout: 5m
    max_timeout: 5m
```

`min_timeout` is the minimal amount of time to wait for check-in. If a state change occurs in less than `min_timeout` then the Elastic Agent should wait that amount of time to check-in again.

`max_timeout` is the maximum amount of time to wait for check-in. If the state has not changed by the `max_timeout` then it will check-in again.

When a users is adding a cloud integration that is pulling from an external service and they get the credentials wrong they are not information of this fact for up to 5 minutes.

Check-in with Fleet is performed during the `min_timeout` and `max_timeout` when an observed state change occurs.

This relates to a similar issue https://github.com/elastic/elastic-agent/issues/2257, but that issue was talking about making the overall check-in longer. Which this change could also allow by adjusting the `max_timeout` to a higher value.

moukoublen:
## What does this PR do?
At the moment, the fleet server "holds" a long poll of 5 minutes on every elastic agent check-in. As a result, when an input has an issue and is in an error or degraded state, information is not relayed to Fleet until at most 5 minutes.

In this PR, a "state watch" was introduced into the fleet gateway component to cancel ongoing check-in requests if a state change occurs while the elasti-agent waits on check-in.

The flow is this:

   * During checking start, (using a mutex) the current state is fetched and a ctx with cancel is created. The cancel function is stored for later usage.
   * On the happy path (check-in completes successfully), the cancel function is being cleaned up.
   * If a state update is received in the state channel while waiting for a check-in response (long poll), the context is canceled due to a specific cause. This leads to the check-in re-start with the new updated state

**Worst case scenario**: During the agent's start-up, potentially many state changes might occur in the inputs till the "healthy" state is achieved. At that moment, a lot of cancellations might happen. The already implemented backoff mechanism covers this case as well. 

The "state watch and cancel ongoing checkin" functionality works only if the fleet agent config `checkin.mode` is set to `on_state_changed` (will be set by agentless controller on agentless). Otherwise, the flow works without any change. 

New configuration `checkin` added under the `FleetAgentConfig`

```yaml
fleet:
  # ...
  mode: "standard" # choose between `standard` or `on_state_change`
  request_backoff_init: 5s # default 60 seconds
  request_backoff_max: 10m # default 10 mins
```

On agentless we can set (through the agentless controller) the above config to enable fast checkin.

## Why is it important?
On agentless environments where the agent logs are not available to the customer, the only way to communicate operational issues (e.g., misconfigured credentials) is through the degraded state status message of the input/component. But creating an agentless agent with wrong credentials will take up to 5 minutes to display the degraded status, and meanwhile, the input will appear as healthy. This provides a bad experience to the customer. 

## Checklist

## Related issues

- Closes #9348 

## Questions to ask yourself

- ...


internal/pkg/agent/application/managed_mode.go
```diff
@@ -168,13 +170,22 @@ func (m *managedConfigManager) Run(ctx context.Context) error {
 		}
 	}
 
+	agentlessFastCheckin, _ := strconv.ParseBool(os.Getenv("AGENTLESS_FAST_LOGIN"))
```
olegsu: In case of error while parsing the value of `AGENTLESS_FAST_LOGIN` the fallback is to existing behavior, right?
Can you log this so it will be clear that the existing behavior used due to error?

```go
if v, ok := os.LookupEnv("AGENTLESS_FAST_LOGIN"); ok && v != "" {
	if b, err := strconv.ParseBool(v); err != nil {
		m.log.Warnf("invalid AGENTLESS_FAST_LOGIN=%q; fast login is disabled.", v)
	}
}
```

moukoublen: Yes, if anything fails, the normal flow of long polls is used. (Also, the name was a typo, 😅 I will rename the variable to `FAST_CHECKIN`.) 

Sure I will add a warning in such case.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -447,3 +454,114 @@ func RequestBackoff(done <-chan struct{}) backoff.Backoff {
...
+func (s *CheckinStateFetcher) FetchState(ctx context.Context) (coordinator.State, context.Context) {
+	state := s.fetcher()
+	return state, ctx
+}
+
+func (s *CheckinStateFetcher) Done()                                     {}
+func (s *CheckinStateFetcher) StartStateWatch(ctx context.Context) error { return nil }
+
+func FastCheckinEnabled(log *logger.Logger) bool {
+	if v, exists := os.LookupEnv("FAST_CHECKIN"); exists {
```
moukoublen: @cmacknz, as we discussed, I followed the env var approach. Is this the right place to put that?

blakerouse: This is not really the proper way the change the behavior. This should be a configuration setting here - https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/configuration/fleet.go

The setting should be changed with environment here - https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/cmd/setup_config.go#L77

This should be a setting that users can adjust even not running in the container, so it needs to be a Fleet setting.

That also means this should be something that could be controlled in Fleet UI. If it can be controlled there then the gateway code needs to handle it changing why the Elastic Agent is running, without the need to restart. See here for how that is done on a policy change - https://github.com/elastic/elastic-agent/blob/2c23284b0255a811c05ac61377bd27ea746c3528/internal/pkg/agent/application/managed_mode.go#L183-L190

moukoublen: Thanks @blakerouse for pointing out the setup_config, I will take a look and add it to the config. 

Regarding on-the-fly changes: 

With a first thought, I think that the way to support it is:
  * Subscribe to state changes and start the go routine that reads from the channel by default
  * Based on the config (enable or disable fast check-in), react or not-react when a state change is received. 

Which means that we need to use the mutex and the `FastCheckinStateFetcher` in both flows (enabled/disabled) and just not react on state changes when the fast checkin is disabled. 

I am pointing this out because I thought the disabled path should be exactly as it was (so no mutex on the path). (_Even though, to be fair, the mutex will follow the fast path in the disabled scenario because only one routine (the checkin loop) will lock/unlock it, so it will be relatively cheap._)  

Unless there is a way to change the `StateFetcher` on the fly based on config changes and start/stop the go routine accordingly. But I feel this is more complicated to be done. 

WDYT?

blakerouse: I would prefer to see the StateFetcher be changed on the fly as I believe that provides the benefit of turning the feature off and it going back to how it use to work.

I think the on-the-fly changes can be delayed, but updating it to at least use the configuration and env from the container entrypoint would be acceptable.

internal/pkg/agent/application/managed_mode.go
```diff
@@ -211,6 +219,10 @@ func (m *managedConfigManager) Run(ctx context.Context) error {
 	// Run the gateway.
 	gatewayRunner := runner.Start(gatewayCtx, func(ctx context.Context) error {
 		defer gatewayErrorsRunner.Stop()
+		if m.cfg.Fleet.FastCheckin {
+			stateWatch := runner.Start(context.Background(), stateFetcher.StartStateWatch)
```
blakerouse, line 225: Why a different context and not the gatewayCtx?

internal/pkg/agent/application/managed_mode.go
```diff
@@ -168,13 +168,21 @@ func (m *managedConfigManager) Run(ctx context.Context) error {
 		}
 	}
 
+	var stateFetcher fleetgateway.StateFetcher
+	if m.cfg.Fleet.FastCheckin {
+		gatewayStateSub := m.coord.StateSubscribe(ctx, 32)
```
blakerouse, line 173: The subscription is cancelled when the context is cancelled. The `StartStateWatch` uses a background context, should these use the same context?

internal/pkg/agent/configuration/fleet.go
```diff
@@ -19,6 +19,7 @@ type FleetAgentConfig struct {
 	Client              remote.Config      `config:",inline" yaml:",inline"`
 	Info                *AgentInfo         `config:"agent" yaml:"agent"`
 	Server              *FleetServerConfig `config:"server" yaml:"server,omitempty"`
+	FastCheckin         bool               `config:"fast_checkin" yaml:"fast_checkin,omitempty"`
```
blakerouse: I think this needs more structure because this will evolve overtime. It will most likely have more settings and things added that effect checkin.

Could you add a new `FleetCheckin` structure and add:


```suggestion
	Checkin         FleetCheckin               `config:"checkin" yaml:"checkin,omitempty"`
```

Then have:

```golang
type FleetCheckin struct {
    Mode string `config:"mode" yaml:"mode,omitempty"`
}

func (f *FleetCheckin) InitDefaults() {
   f.Mode = "standard"
}

func (f *FleetCheckin) Validate() error {
  if f.Mode != "standard" && f.Mode != "on_state_change" {
    return errors.New("check.mode must be either 'standard' or 'on_state_change'")
  }
  return nil
}
```

blakerouse (changes requested):

internal/pkg/agent/application/managed_mode.go
moukoublen: Indeed, I think it should be the `ctx` of the anonymous function, to get cancelled when that gets cancelled. 

Update: I just remembered, the thought is that this runner should run with background context like the `gatewayErrorsRunner` and be stopped explicitly when the `Stop` function is called (which will happen once the gatewayRunner ends).

internal/pkg/agent/application/managed_mode.go
moukoublen: The StartStateWatch's ctx will get cancelled once the `defer stateWatch.Stop()` is called, which will happen when `gatewayCtx` gets cancelled which will happen when the root ctx of managedConfigManger `ctx` gets cancelled. 


So the flow is -if I am not mistaken- 

root ctx becomes cancelled -> 
subscription stops and channel closes -> 
`(m *managedConfigManager) Run(ctx context.Context) error {` calls the defered cancallation of gatewayCtx ->
gatewayRunner stops -> 
stateWatch.Stop()

I think there is a tiny window of stateWatcher reading from a closed channel (and so perform multiple reads of empty state). 

Perhaps its better to add a check on state watched. 
```go
		case st, isOpen := <-s.stateChan:
			if !isOpen {
				s.log.Info("FleetGateway state watching channel closed, stopping loop.")
				return nil
			}
```

This will ensure that no matter the window, the state watch loop ends when the state subscription channel is closed. 

Update: added.

internal/pkg/agent/configuration/fleet.go
moukoublen: Sure, I added the struct. I also made the default values usable by accepting both the empty string and "standard" as standard.

olegsu (approved):

changelog/fragments/1758032089-quicker-fleet-check-in.yaml
```diff
@@ -0,0 +1,45 @@
...
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# REQUIRED for all kinds
+# Change summary; a 80ish characters long description of the change.
+summary: add support for quicker fleet check-in upon component status
```
cmacknz: Let's remove this changelog, I don't want anyone outside of us ourselves attempting to use this so best they don't know we support it.

We have had occasional requests for this and we haven't qualified it will work beyond a couple of agents under management.

internal/pkg/agent/application/gateway/fleet/fleet_gateway_test.go
```diff
@@ -445,6 +445,62 @@ func TestFleetGateway(t *testing.T) {
 		default:
 		}
 	})
+
+	t.Run("Test cancel checkin on sate update", func(t *testing.T) {
```
cmacknz:
```suggestion
	t.Run("Test cancel checkin on state update", func(t *testing.T) {
```

internal/pkg/agent/application/managed_mode.go
```diff
@@ -168,13 +168,21 @@ func (m *managedConfigManager) Run(ctx context.Context) error {
 		}
 	}
 
+	var stateFetcher fleetgateway.StateFetcher
+	if m.cfg.Fleet.Checkin.IsModeOnStateChanged() {
```
cmacknz, line 172: Let's log which fetcher we've decided to use in here.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -494,3 +499,106 @@ func getPolicyRevisionIDX(action fleetapi.Action) int64 {
...
+		select {
+		case <-ctx.Done():
+			s.log.Info("FleetGateway state watching stopped")
+			return ctx.Err()
+		case st, isOpen := <-s.stateChan:
+			if !isOpen {
+				s.log.Info("FleetGateway state watching channel closed, stopping loop.")
+				return nil
+			}
+			s.log.Infof("FleetGateway state change notification %#v", st)
```
cmacknz: Logging the coordinator state is dangerous because the underlying component model contains all configuration which has in the past contained secrets.

Try to only log the parts you need. Also this has the potential to generate a lot of log spam but I do like the idea of starting to log when we start a checkin and why that was.

There is a debug log at the start of each checkin that could be tidied up to do this https://github.com/elastic/elastic-agent/blob/fcc1752e8749c34a7487f9529d8e0360a2d98fa3/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L194

moukoublen: Good catch, thanks! That logging of state was intended during development and test for verification and escaped cleanup. Will cleanup and add other logs as you suggested.

changelog/fragments/1758032089-quicker-fleet-check-in.yaml
moukoublen: The reason I added it was because of a workflow that was failing, requesting a fragment. Can I overcome that?

moukoublen:
@cmacknz yes, it has been tested end-to-end with a fleet server and cloudbeat going to degraded state after startup (to emulate the cancellations of checking and also the fast appearance of degraded state with status message in kibana which was the initial requirement). 

On Monday I could also post a video/logs 


> We may see some log spam about the client breaking the connection

Due to the usage of the backoff, even if we have a "fast-attack" of state changes at the beginning, it's not translated to 1:1 check-in requests and cancellations. They might be more for sure, but due to the backoff, we will not have an uncontrolled burst of logs and cancellations/retries. 

That is because while waiting for the backoff, any state change that arrives will not cancel any ongoing request since there will not be any. 

Also, I excluded the "on state changed" cancellation from the counter that decides to log an error.

I will provide log files. 



Potentially, we can implement a rate-limiting-like mechanism on cancellation, for example, not canceling if another cancellation occurs within X time period. That way we can ensure that no more than one cancellation every e.g. 5 seconds can occur.

example: 

```golang
func (s *FastCheckinStateFetcher) invalidateState() {
	s.mutex.Lock()
	defer s.mutex.Unlock()

	// the throttle check can be done perhaps without mutex lock.
	now := s.now()
	if now.Sub(s.lastInvalidation) < s.throttle {
		return
	}

	if s.cancel != nil {
		s.cancel(errComponentStateChanged)
		s.cancel = nil
		s.lastInvalidation = s.now()
	}
}
```

(Unfortunately, a throttling mechanism like that leaves a small window open -undermining the initial purpose- of the state change to be propagated to kibana after 5 minutes.)

internal/pkg/agent/application/managed_mode.go
moukoublen: Added the checkin mode in log.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
moukoublen: > but I do like the idea of starting to log when we start a checkin and why that was.

We could add the a check of "last checkin outcome" and log conditionally `"Checking started"` / `"Checking started because of state change"` but since I added an explicit [warining log](https://github.com/elastic/elastic-agent/pull/9982/files#diff-c81fb678489d398d5e6d8686f2a17fb86e890a613276acb1227f1aaab0eb2cf8R204-R207) when a check-in gets cancelled because of state change `"Check in cancelled because of state change, retrying"`, I feel it gets the job done for now. wdyt?

internal/pkg/agent/application/managed_mode.go
blakerouse: Thanks for the detail and updated that to handle the channel being closed.

internal/pkg/agent/application/managed_mode.go
blakerouse: Yeah, that makes sense.

internal/pkg/agent/configuration/fleet.go
blakerouse: Nice, thank you.

changelog/fragments/1758032089-quicker-fleet-check-in.yaml
blakerouse: You can add the `skip-changelog` label on the PR and it will allow you to not have it.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
blakerouse: @moukoublen I believe the log you already added covers it.

blakerouse (commented):
Overall this looks good. I am going to wait for a green CI before approval, but this looks good.

olegsu (approved):

internal/pkg/agent/configuration/fleet.go
```diff
@@ -47,5 +54,40 @@ func DefaultFleetAgentConfig() *FleetAgentConfig {
...
 		Info:    &AgentInfo{},
+		Checkin: DefaultFleetCheckin(),
+	}
+}
+
+func DefaultFleetCheckin() FleetCheckin {
+	return FleetCheckin{
+		Mode:               fleetCheckinModeStandard,
+		RequestBackoffInit: 60 * time.Second,
+		RequestBackoffMax:  10 * time.Minute,
```
cmacknz: This is defined twice, once here and once in fleet_gateway.go and it looks like this one takes precedence because it's non-zero. If someone changes the settings in fleet_gateway.go they'll be ignored as far as I can tell which is confusing. Can you define these in a single place?

https://github.com/elastic/elastic-agent/blob/dc21d56dbf15d948a395f3c2370339450cb2bcb0/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L43-L47

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
cmacknz: This comment is out of date at least, since you set settings.Backoff when creating the gateway.

cmacknz, line 474: This ignores any changes from `settings.Backoff`. This is also used outside of the gateway itself from what I can see.

Maybe it just needs to be renamed to `DefaultRequestBackoff` and use the defaults defined in the configuration/fleet.go so there's a single definition of the defaults.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -348,7 +363,7 @@ func (f *FleetGateway) execute(ctx context.Context) (*fleetapi.CheckinResponse,
 	}
 
 	// get current state
-	state := f.stateFetcher()
+	state, stateCTX := f.stateFetcher.FetchState(ctx)
```
cmacknz:
```suggestion
	state, stateCtx := f.stateFetcher.FetchState(ctx)
```

nit: we don't capitalize CTX anywhere else in the agent code.

internal/pkg/agent/configuration/fleet.go
```diff
@@ -47,5 +54,40 @@ func DefaultFleetAgentConfig() *FleetAgentConfig {
...
+	return f.Mode == fleetCheckinModeOnStateChanged
+}
+
+func (f *FleetCheckin) Validate() error {
+	if f.Mode != "" && f.Mode != fleetCheckinModeStandard && f.Mode != fleetCheckinModeOnStateChanged {
+		return errors.New("checkin.mode must be either 'standard' or 'on_state_change'")
 	}
+
+	if f.RequestBackoffMax < f.RequestBackoffInit {
+		return errors.New("checkin.request_backoff_max must be grater than or equal to checkin.request_backoff_init")
```
cmacknz:
```suggestion
		return errors.New("checkin.request_backoff_max must be greater than or equal to checkin.request_backoff_init")
```

internal/pkg/agent/configuration/fleet.go
moukoublen: @cmacknz, at first, I set the default values exclusevily inside `internal/pkg/agent/configuration/fleet.go`, but they arrived zero in the gateway component (I didn't investigate why, I suspect `mergeFleetConfig` and the merge that is happening). 

I think the best it would be to leave the default values inside the `fleet_gateway` and not in the `fleet.go`.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
moukoublen: To arrange well with my comment [here](https://github.com/elastic/elastic-agent/pull/9982#discussion_r2388724440), I'd suggest keeping the default values inside the `fleet_gateway.go`.

internal/pkg/agent/configuration/fleet.go
moukoublen: I removed the default values from `configuration/fleet.go`.

cmacknz: Yeah sure I don't have a preference which one is the only one, as long as it isn't defined in two places that don't interact in an obvious way.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -494,3 +513,118 @@ func getPolicyRevisionIDX(action fleetapi.Action) int64 {
...
+		stateChan: stateChan,
+		cancel:    nil,
+		mutex:     sync.Mutex{},
+	}
+}
+
+// Fetch wraps the state fetching to send in the check-in request under the checkin state mutex.
+// After the state is fetched the checkin cancellation function has be initialized and the new context
+// is returned.
+func (s *FastCheckinStateFetcher) FetchState(ctx context.Context) (coordinator.State, context.Context) {
```
cmacknz, line 549: I like the approach of returning a context that cancels when the state changes, this is a really nice way to get the request to return without needing to change anything in the lower layers with how the HTTP request happens 👍 

I also think you only need that special context, and the state that gets returned here isn't guaranteed to be the one returned from `stateChan` below. The attempt to invalidate the state on change isn't really doing anything to provide strong guarantees over just calling `m.coord.State()` like we were before. 

This makes me think you could simplify this into an API that takes only the context and returns another context, which depending on the fleet settings is either augmented to cancel on state change or you return the input context as a no-op.

You wouldn't be touching the `StateFetcher` interface in that case because you don't actually use the state from the state watch here.

The logic you have is "break long poll on state change, then grab the current state when checking in again (whatever that is not necessarily the one that caused the checkin to be broken originally)".

cmacknz:
```go
		case _, isOpen := <-s.stateChan:
			if !isOpen {
				s.log.Info("FleetGateway state watching channel closed, stopping loop.")
				return nil
			}
			// TODO: consider check for specific changes e.g. degraded?
			s.invalidateState()
		}
```

This block can't ever hold `s.mutex` for it's entire duration, which is what you'd need to make sure the state you check in with was the one that caused the change. 

If there were two rapid state changes, e.g. one in the channel again immediately after reading from it but before you had a chance to call `invalidateState` the second time then you can trigger a new checkin from the first state change but the actual state value will be from the second change (because fleet_gateway samples the immediately value of the coordinator state).

I don't think this is worth optimizing for, just that this code is trying to provide a guarantee around what state changes are observed that it doesn't actually accomplish. Better to just not try to do it at all and simplify. State changes that happen that fast probably aren't worth reporting to users anyway.

Otherwise, I think you'd need a different approach to feed state changes into the fleet gateway as they happen.

moukoublen: > and the state that gets returned here isn't guaranteed to be the one returned from stateChan below. 

Yes, and that's on purpose in a way. 

So what I wanted to achieve is: `FetchState` returns a state and a context that is valid as long as this state is valid. 

The moment we receive a state (sync/atomicly), we receive a context that dictates if this current state is valid. 

If the state gets invalidated, I don't want to keep the specific state that arrived in the channel because we might check in again in `X` seconds (due to the backoff) and by then the state might be different. I want to receive  the latest state just before the checkin request. 


> This makes me think you could simplify this into an API that takes only the context and returns another context, which depending on the fleet settings is either augmented to cancel on state change or you return the input context as a no-op.

I am afraid that this would potentially leave a window open for unnecessary cancellations. If I just receive a cancellable state without (with mutex) receiving the state (so not in a single atomic step), and receive the state in the next step, then from the moment I get the cancallable context till the moment I receive the state, a state change might arrive and cancel a context that should not get cancelled because the state I am about to recieve is valid. 

Like this:

```plain
-- [FleetGateway] get cancellable CTX

---- (in parallel) [StateWatcher] meanwhile a state arrives in the channel and CTX gets invaldated

-- [FleetGateway] calls .GetState() (totally updated state)

-- [FleetGateway] CTX is invalid 
```

That's why the flow has atomic steps of 

```plain
-- [FleetGateway] Get a State with a CTX that is valid as long as the state is valid
-- [FleetGateway] Checkin with that state and that CTX
```

cmacknz: This is what I see as a possible execution path happening right now:

In fleet_gateway.go `state, stateCTX := f.stateFetcher.FetchState(ctx)` happens and sets the state and context.

Concurrently in the state fetcher, there is a state change and the following block executes and invalidates the context:

```go
		case _, isOpen := <-s.stateChan:
			if !isOpen {
				s.log.Info("FleetGateway state watching channel closed, stopping loop.")
				return nil
			}
			// TODO: consider check for specific changes e.g. degraded?
			s.invalidateState()
		}
```

Again in fleet_gateway.go `resp, took, err := cmd.Execute(stateCTX, req)` executes but returns immediately because the state was invalidated.

In this case the only thing doing any work towards the checkin behaviour is the context cancellation that happens in `invalidateState`. The `state` returned from `FetchState` could have used the previous implementation and you'd get the same result, so IMO returning a state there is unnecessary.

I think you'd end up with a separate context fetcher interface that does basically the same thing behaviourally, it just makes it clearer the state you fetch is always just a sample of the current state.

moukoublen: > If there were two rapid state changes, e.g. one in the channel again immediately after reading from it but before you had a chance to call invalidateState the second time then you can trigger a new checkin from the first state change but the actual state value will be from the second change (because fleet_gateway samples the immediately value of the coordinator state).


Due to the backoff this will not happen. 

when the first state gets cancelled, the checkin inner loop will wait for at least a couple of seconds. Meanwhile, any other state change will try to run `invalidateState`, but it will be a noop.

After the backoff is done, the latest state will be fetched and attempt again.


> just that this code is trying to provide a guarantee around what state changes are observed that it doesn't actually accomplish.

I think that precisely because of the backoff + atomic steps it does accomplish that.   

Even if a cancellation happens, that it shouldn't (which will not), it will re-attempt in a while (after the bo) to check in with the latest state.

cmacknz: I think the behaviour doesn't actually change with what I've proposed, so now that I've thought about it a bit more it's just a different way to structure things.

moukoublen:
@blakerouse CI seems ok now

blakerouse (approved):
Looks good. Thanks for working through this with me and getting it to be mergable.

PR #9985, opened 2025-09-16
Don't apply unnecessary otel config updates

swiatekm:
## What does this PR do?

We do a basic check to verify if an otel config update actually changes anything, and skip applying it if it doesn't. Otel configuration reloads are relatively expensive, and we want to avoid them whenever possible. This is especially impactful for self-monitoring configuration, which rarely changes even in environments with lots of config churn.

This is accomplished by marshalling the config to json and hashing it using xxhash. The particular method was chosen because it's fast, and config comparisons are known to be expensive elsewhere in the codebase for beats processes.

There's some subtlety involved in the change, as it's not clear if the manager should emit a nil error for configurations it doesn't apply. I decided that it shouldn't, because it can override a real error if the configuration is faulty.

## Why is it important?

Otel configuration reloading is expensive and we want to avoid it if possible.

## Checklist

### DIFF
--- a/internal/pkg/otel/manager/manager.go
+++ b/internal/pkg/otel/manager/manager.go
@@ -543,7 +543,7 @@ func (m *OTelManager) maybeUpdateMergedConfig(mergedCfg *confmap.Conf) (updated 
 
 	m.mergedCollectorCfg = mergedCfg
 	m.mergedCollectorCfgHash = mergedCfgHash
-	return bytes.Compare(mergedCfgHash, previousConfigHash) != 0 || err != nil, err
+	return !bytes.Equal(mergedCfgHash, previousConfigHash) || err != nil, err
 }
 
 // reportComponentStateUpdates sends component state updates to the component watch channel. It first drains


internal/pkg/otel/manager/manager.go
```diff
@@ -532,3 +556,25 @@ func (m *OTelManager) reportComponentStateUpdates(ctx context.Context, component
...
+
+// calculateConfmapHash calculates a hash of a given configuration. It's optimized for speed, which is why it
+// json encodes the values directly into a xxhash instance, instead of converting to a map[string]any first.
+func calculateConfmapHash(conf *confmap.Conf) (uint64, error) {
+	if conf == nil {
+		return 0, nil
+	}
+
+	h := xxhash.New()
+	encoder := json.NewEncoder(h)
```
cmacknz, line 580: Out of curiosity, why endcode to JSON when the input configuration was YAML originally? Any chance of incompatibility biting us?

The way this is written it would just result in an extra configuration application and not be fatal which is 👍

swiatekm: No major reason. JSON is simpler so encoding tends to be faster, and there's a package for it in the standard library. Generally speaking, otel config is serializable to any format supported by koanf, it's just not exposed in the API, resulting in needing to either convert it to a Go map or do a slight workaround like what I did here.

cmacknz: Thanks, can you add this explanation directly as a comment in the code here?

internal/pkg/otel/manager/manager.go
```diff
@@ -282,14 +285,22 @@ func (m *OTelManager) Run(ctx context.Context) error {
...
 			m.mx.Unlock()
 
-			err = m.applyMergedConfig(ctx, collectorStatusCh, m.collectorRunErr)
-			// report the error unconditionally to indicate that the config was applied
-			reportErr(ctx, m.errCh, err)
+			if configUpdateErr != nil {
+				m.logger.Warn("failed to calculate hash of merged config, proceeding with update", zap.Error(configUpdateErr))
+			}
+
+			if configChanged {
```
cmacknz, line 299: The chance of a hash collision is tiny but it's not impossible, if that were to happen, we'd have no way to know. The collector would just silently use a stale configuration.

Pointless reloading the collector right now seems worth this trade off, but we should put a debug log in here with the result of the comparison and the calculated hash values so we have some chance to diagnose this if it ever did happen.

internal/pkg/otel/manager/manager_test.go
```diff
@@ -1361,3 +1368,162 @@ func assertOtelStatusesEqualIgnoringTimestamps(t require.TestingT, a, b *status.
...
+				},
+			})
+			hash3, err := calculateConfmapHash(conf3)
+			require.NoError(t, err)
+			assert.NotEqual(t, hash1, hash3)
+		})
+	})
+}
+
+func TestOTelManager_maybeUpdateMergedConfig(t *testing.T) {
```
cmacknz, line 1443: The test cases seem reasonable, is it worth having one with a complete and representative collector configuration just to make sure nothing strange happens?

The integration tests would probably catch if it were totally broken regardless.

internal/pkg/otel/manager/manager.go
swiatekm: You mean put a debug log in the other branch of this conditional, right? Sure, I can do that.

internal/pkg/otel/manager/manager_test.go
swiatekm: Sure, I can use the same config we already do for manager tests, not a problem.

internal/pkg/otel/manager/manager.go
swiatekm: Probably best to log it unconditionally, actually.

internal/pkg/otel/manager/manager.go
```diff
@@ -282,14 +285,30 @@ func (m *OTelManager) Run(ctx context.Context) error {
 
 			// this is the only place where we mutate the internal config attributes, take a write lock for the duration
 			m.mx.Lock()
-			m.mergedCollectorCfg = mergedCfg
+			previousConfigHash := m.mergedCollectorCfgHash
+			configChanged, configUpdateErr := m.maybeUpdateMergedConfig(mergedCfg)
```
cmacknz, line 290: There is an edge case in here, maybe two.

1. If the calculated hash of the first config were ever legitimately zero (assuming that's an option), this wouldn't apply it. I think you can handle this with the fact that the `mergedCfg` will be `nil`.
2. Related to handling of previous config and `nil` mergedCfg, what is the behaviour when the otel manager stops and starts? For example because an OTel integration was added, removed, and then the exact same config added again? Would this cause us not to apply the configuration? Similar question around what happens when the otel runtime gets turned on and off, though in that case maybe the entire state of the otel manager gets cleared?

swiatekm: The function calculating the hash returns 0 for a `nil` config, and this is also the state the manager starts in (`nil` config and 0 hash), so 1. is really just another hash collision. Not sure how much effort we want to put into detecting this, to be honest. If we don't trust the hash function, we can instead compare the configurations with something like `reflect.DeepEqual`, but that will be significantly slower.

For 2, there isn't really a notion of starting or stopping the manager itself. We create it at startup and then it always runs. There is a notion of starting and stopping the otel collector, but that's simple enough - we start it when we get a non-nil config and stop it when we get a nil config. It's not any different from handling any other config value - it works as long as there's no collision.

Would it help if I switched this to, say, FNV128, for a larger hash value? Intuitively, the set of valid otel configs is relatively small, so the chance of getting a collision even at 64 bits seems miniscule, but I don't think performance is so critical here that using a slower hash function will make any real difference.

cmacknz: > The function calculating the hash returns 0 for a nil config, and this is also the state the manager starts in (nil config and 0 hash), so 1. is really just another hash collision

Yeah agreed, special casing one collision doesn't really change anything for the other collision cases. Those still exist at a similarly low probability.

> Would it help if I switched this to, say, FNV128, for a larger hash value? Intuitively, the set of valid otel configs is relatively small, so the chance of getting a collision even at 64 bits seems miniscule, but I don't think performance is so critical here that using a slower hash function will make any real difference.

Yeah I think using more bits is helpful here, I am a bit torn because I see the value of preventing pointless restarts (that could lead to data loss) but it also introduces a tiny chance of a configuration silently failing to apply. 

This is a case where an optimization (the hash) could make the system fail to do one of its core jobs - apply configuration updates to a running collector. Also not having the optimization can make us worse at another core job - data collection.

swiatekm: Switched to a 128 bit hash here: https://github.com/elastic/elastic-agent/pull/9985/commits/a3e5ef964e025979d81a41ef6509f9a8f5c48d9f.

cmacknz (approved):
Thanks, I'm good with the trade off that we potentially have hash collisions preventing config updates vs us losing data more frequently because the collector restarts pointlessly.

PR #9987, opened 2025-09-16
Set Default Telemetry Collection Interval in EDOT to 60s

girodav:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
+++ b/deploy/helm/edot-collector/kube-stack/managed_otlp/values.yaml
@@ -139,6 +139,7 @@ collectors:
                 - "DELETED"
         # [K8s Cluster Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sclusterreceiver)
         k8s_cluster:
+          collection_interval: 60s
           auth_type: serviceAccount # Determines how to authenticate to the K8s API server. This can be one of none (for no auth), serviceAccount (to use the standard service account token provided to the agent pod), or kubeConfig to use credentials from ~/.kube/config.
           node_conditions_to_report:
             - Ready
@@ -375,7 +376,7 @@ collectors:
               type: container
         # [Hostmetrics Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver)
         hostmetrics:
-          collection_interval: 10s
+          collection_interval: 60s
           root_path: /hostfs # Mounted node's root file system
           scrapers:
             cpu:
@@ -446,7 +447,7 @@ collectors:
         # [Kubelet Stats Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/kubeletstatsreceiver)
         kubeletstats:
           auth_type: serviceAccount # Authentication mechanism with the Kubelet endpoint, refer to: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/kubeletstatsreceiver#configuration
-          collection_interval: 20s
+          collection_interval: 60s
           endpoint: ${env:OTEL_K8S_NODE_NAME}:10250
           node: "${env:OTEL_K8S_NODE_NAME}"
           # Required to work for all CSPs without an issue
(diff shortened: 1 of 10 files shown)


changelog/fragments/1758039757-reduce-default-telemetry-frequency.yaml
```diff
@@ -0,0 +1,45 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: breaking-change
```
cmacknz:
```suggestion
kind: bug-fix
```

I don't think this is really a breaking change in the sense that users don't have to do anything specific when upgrading to a version with these configurations.

cmacknz (approved):
nit on the changelog otherwise LGTM

lucabelluccini:
May I ask if there are prebuilt dashboards or Kibana O11y apps which expect to have more than 1 sample per minute? Did we validate the correct rendering of the metrics if sampled only once per minute?

rogercoll (approved):

girodav:
> May I ask if there are prebuilt dashboards or Kibana O11y apps which expect to have more than 1 sample per minute? Did we validate the correct rendering of the metrics if sampled only once per minute?

As far as I know, only the Otel K8s / system metrics content packs would be affected by this change. We do not expect any breakage, but after merging this, we are going to have to double-check each backported version.

lucabelluccini:
- Hello @smith / @roshan-elastic Do we have tests which validate the correct behavior of the Hosts UI and associated tabs if metrics are collected once per minute instead of every 10s?
- @girodav do we have a follow up issue to track Hosts UI and Otel K8s / system metrics content packs keep working?

roshan-elastic:
Hi @lucabelluccini,

> Do we have tests which validate the correct behavior of the Hosts UI and associated tabs if metrics are collected once per minute instead of every 10s?

I'm not sure if we have/need specific tests as I'd imagine that any default interval periods should work nicely in Dashboards, alerts etc as well as the Hosts UI (and we don't want to build in specific exceptions for one UI).

We respect the default interval periods that Lens provides so I don't think we need any specific exception.

@smith - Do you have any thoughts?

cc @mlunadia in case this is relevant

smith:
> Do we have tests which validate the correct behavior of the Hosts UI and associated tabs if metrics are collected once per minute instead of every 10s?

I don't know how this could be relevant to the UI. It will display the data that's in Elasticsearch.

lucabelluccini:
What I mean is... is there any visualization in our curated UIs which expect 1 or more samples per minute?

I just got today a user reporting that switching the Elastic Agent System integration to 5m interval for the process metrics (instead of 1m default) breaks the Hosts / Processes tab. This example is **unrelated** to this PR, but made me think about this change on EDOT as it's changing the amount of metrics per minutes available. (The investigation of this specific issue is ongoing and it is not in scope of this PR)

image

roshan-elastic:
Ah - yes, I see @lucabelluccini.

@smith - I'm not sure if the processes tab would be affected but can we check?

Also, I'm wondering how we might catch potential issues like this so we can catch them before our customers do - do you have any thoughts?

roshan-elastic:
FYI @lucabelluccini - we've created an issue to look at how we can make the processes tab more resilient:

https://github.com/elastic/kibana/issues/237094

It hasn't been prioritised yet but FYI

lucabelluccini:
Hello @roshan-elastic 
https://github.com/elastic/kibana/issues/237094 is associated to the System integration though (it has been opened by @iblancof as a consequence of the user I've mentioned in https://github.com/elastic/elastic-agent/pull/9987#issuecomment-3338919197 which I cannot share here).

My question would be about validating that our curated uis and OTel assets are not breaking following the collection interval from 10s to 60s in EDOT Collector Agent mode when used for K8s and VM Observability.

My "recommendation" is that every time something is changed in our EDOT configs for o11y use-cases, we also validate our O11y curated UIs and Otel Integration packages are not going to break with those new values. Please reach me out via Slack if further clarifications are needed.

roshan-elastic:
Thanks @lucabelluccini - understood!

herrBez:
I second what Luca said. I think we should also consider adjusting the k8s_otel dashboard to keep into account this new time interval. Here https://github.com/elastic/integrations/blob/main/packages/kubernetes_otel/kibana/dashboard/kubernetes_otel-cluster-overview.json#L1087 we are using 1m and if you are "unlucky" you may end up having an "empty" panel

PR #9992, opened 2025-09-16
fix: always clear the coordinator overridden state on err inside upgrade of coordinator

pkoutsovasilis:
## What does this PR do?

This PR fixes a regression introduced by [#9634](https://github.com/elastic/elastic-agent/pull/9634) where the coordinator’s `overrideState` could remain set if an upgrade attempt failed early (e.g. agent not upgradeable, capability check denied, or pre-upgrade callback returned an error).  

Specifically, this PR:
- Adds calls to `ClearOverrideState()` before returning from all early failure branches inside `Coordinator.Upgrade`.
- Extends the coordinator test suite to assert that `overrideState` is reset to `nil` after a failing `preUpgradeCallback`, preventing stale state from leaking into subsequent upgrade attempts.

## Why is it important?

Without this change, failed upgrades could leave the coordinator in a state that incorrectly reflects an ongoing upgrade. This blocks future upgrade attempts until the Elastic Agent is restarted, which is disruptive and operationally undesirable.  

By clearing the override state on failure, we ensure the coordinator always returns to a clean state, enabling subsequent upgrade attempts to proceed without requiring a restart.

## Checklist

## Disruptive User Impact

Previously, users would need to manually restart the Elastic Agent after a failed upgrade attempt in order to retry an upgrade.  
With this fix, the agent automatically clears the override state, removing the need for manual intervention.

 Run `mage unitTest` and confirm that all tests pass, including the updated coordinator tests.

## Related issues

- Relates https://github.com/elastic/elastic-agent/pull/9634

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -765,6 +765,7 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 
 	// early check outside of upgrader before overriding the state
 	if !c.upgradeMgr.Upgradeable() {
+		c.ClearOverrideState()
 		det.Fail(ErrNotUpgradable)
 		return ErrNotUpgradable
 	}
@@ -772,6 +773,7 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 	// early check capabilities to ensure this upgrade actions is allowed
 	if c.caps != nil {
 		if !c.caps.AllowUpgrade(version, sourceURI) {
+			c.ClearOverrideState()
 			det.Fail(ErrNotUpgradable)
 			return ErrNotUpgradable
 		}
@@ -780,6 +782,7 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 	// run any pre upgrade callback
 	if uOpts.preUpgradeCallback != nil {
 		if err := uOpts.preUpgradeCallback(ctx, c.logger, action); err != nil {
+			c.ClearOverrideState()
 			det.Fail(err)
 			return err
 		}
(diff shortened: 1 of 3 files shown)


cmacknz:
> Without this change, failed upgrades could leave the coordinator in a state that incorrectly reflects an ongoing upgrade. This blocks future upgrade attempts until the Elastic Agent is restarted, which is disruptive and operationally undesirable.

This is potentially very very disruptive and needs to be in a known issue asap. Upgrades fail all the time (e.g. download failures) so this is going to cause a lot of people to manually restart agents.

This bug was released already in v8.18.7 and v9.0.7. We should raise this as a blocker for the upcoming 9.1.4 and 8.19.4 releases ASAP as well.

* v8.18.7 https://github.com/elastic/elastic-agent/commit/1b51b95fb565607b3926c34736a0e34822baabe8
* v9.0.7 https://github.com/elastic/elastic-agent/commit/08341bbff784026382f8760a3133acc6db586b1c

cmacknz:
Reading the PR, the description is maybe overselling the impact here.

What specific conditions lead you into this state?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -780,6 +782,7 @@ func (c *Coordinator) Upgrade(ctx context.Context, version string, sourceURI str
 	// run any pre upgrade callback
 	if uOpts.preUpgradeCallback != nil {
 		if err := uOpts.preUpgradeCallback(ctx, c.logger, action); err != nil {
+			c.ClearOverrideState()
```
cmacknz, line 785: This seems like the most impactful part of this where agent will become unupgradeable anytime we fail to notify endpoint of an upgrade action?

https://github.com/elastic/elastic-agent/blob/b5e1dadabaeab82386c8f0e702b7e17bac794f84/internal/pkg/agent/application/actions/handlers/handler_action_upgrade.go#L70-L78

PR #10002, opened 2025-09-17
Cleanup EC2 VMs created by FIPS integration tests

ycombinator:
## What does this PR do?

This PR adds a new Buildkite pipeline and associated script to clean up any lingering EC2 VMs created by FIPS integration tests.  The pipeline is executed every 4 hours.

## Why is it important?

Elastic Agent CI runs FIPS integration tests.  These integration tests create FIPS-compliant VMs in EC2 where the tests are actually executed.  Sometimes these CI builds fail without cleaning up said VMs. 

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/misc/ec2-cleanup.yml
+++ b/.buildkite/misc/ec2-cleanup.yml
@@ -7,14 +7,13 @@ accounts:
     options:
       key: "${ACCOUNT_KEY}"
       secret: "${ACCOUNT_SECRET}"
-      project: "${ACCOUNT_PROJECT}"
 
 scanners:
   - account_name: "${ACCOUNT_PROJECT}"
     resources:
       - type: "node"
         regions:
-          - "us-central1"
+          - "us-west-2"
         filters:
           - type: "<"
             pointer: "/extra/creationTimestamp"


.buildkite/scripts/steps/ec2-cleanup.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+export ACCOUNT_KEY_SECRET=$(vault kv get -field=client_email $VAULT_PATH)
+export ACCOUNT_SECRET=$(vault kv get -field=private_key $VAULT_PATH)
+export ACCOUNT_PROJECT_SECRET=$(vault kv get -field=project_id $VAULT_PATH)
```
ycombinator, line 7: Not sure if these are the correct fields in Vault for EC2? I copied these from the analogous `gce-cleanup.sh` file.  @v1v?

v1v: I think we need to use a different approach here, and avoid accessing vault secrets at runtime - they are not masked.

The recommended approach is to use https://github.com/elastic/vault-secrets-buildkite-plugin/

However, I think we could use OIDC and avoid the situation of using static secrets:
- https://github.com/elastic/oblt-aws-auth-buildkite-plugin
- plus the OIDC configuration in the specific AWS account

catalog-info.yaml
```diff
@@ -440,6 +440,53 @@ spec:
         everyone:
           access_level: BUILD_AND_READ
 
+---
```
v1v, line 443: I normally suggest to create this catalog-info change in a first PR. Why? Buildkite integration with Terrazzo cannot be tested unless is merged.

So you could create a PR afterwards and point to your branch for testing purposes using the BK UI and validate changes work as expected during your PR development.

catalog-info.yaml
```diff
@@ -440,6 +440,53 @@ spec:
...
+      provider_settings:
+        build_pull_request_forks: false
+        build_pull_requests: false # requires filter_enabled and filter_condition settings as below when used with buildkite-pr-bot
+        publish_commit_status: false # do not update status of commits for this pipeline
+        build_tags: false
+        build_branches: false
+        filter_enabled: true
+        # required by "build_pull_requests: true" when used with buildkite-pr-bot
+        filter_condition: >-
+          build.pull_request.id == null || (build.creator.name == 'elasticmachine' && build.pull_request.id != null)
```
v1v, line 475: If no BK PR bot, I think we can remove this section

catalog-info.yaml
```diff
@@ -440,6 +440,53 @@ spec:
...
+---
+# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/e57ee3bed7a6f73077a3f55a38e76e40ec87a7cf/rre.schema.json
+apiVersion: backstage.io/v1alpha1
+kind: Resource
+metadata:
+  name: buildkite-elastic-agent-ec2-cleanup
+  description: Clean up stale EC2 instances
+  links:
+    - title: Pipeline
+      url: https://buildkite.com/elastic/elastic-agent
```
v1v, line 452:
```suggestion
      url: https://buildkite.com/elastic/elastic-agent-ec2-cleanup
```

ycombinator:
> I think the current implementation in the `ingest-dev` project could be used too. So we won't need this PR but only the changes from `.buildkite/misc/ec2-cleanup.yml`.

Did you mean the `ingest-dev` repo?  If so, what implementation are you referring to?  I'm not finding anything that uses `cloud-reaper` under https://github.com/elastic/ingest-dev/tree/main/.buildkite.

v1v:
I actually got confused with the ESS cleanup and https://github.com/elastic/elastic-agent/blob/main/.buildkite/pipeline.elastic-agent-gce-cleanup.yml. Somehow, I thought it was in the private repository.

I think using pipeline.elastic-agent-gce-cleanup.yml could be enough, it also runs every 4 hours, what do you think?

ycombinator:
> I think using pipeline.elastic-agent-gce-cleanup.yml could be enough, it also runs every 4 hours, what do you think?

The FIPS VMs are [spun up in EC2](https://github.com/elastic/elastic-agent/blob/710c49f45433e2f136a6e41cae980c1aa37dabdd/.buildkite/bk.integration-fips.pipeline.yml#L72), so how could we use the GCE cleanup script?

v1v:
Good point, the naming is not meaningful in that case - I'm certainly getting blind after the long day, hence my bad with my suggestions 🙇 


-----

Although, can you elaborate on what you mean by 'spun up in EC2'? IIUC, those VMs are managed by the BK provisioner Gobld, hence we don't need to clean them up - they are destroyed automatically after the step finished (regardless of the success)

IIRC, the GCE cleanup was for the `ogc` when running ITs, but IIRC, that's not the case anymore

ycombinator:
> Although can you elaborate what you mean with spun up in EC2? IIUC, those VMs are managed by the BK provisioner Gobld

I linked to the line in my comment but basically the buildkite step has `agents.provider: "aws"`.  Doesn't this mean the VM used that step is provisioned in EC2?  

> IIRC, the GCE cleanup was for the ogc when running ITs, but IIRC, that's not the case anymore

Yeah, this was my initial thought too.  The VMs should now be managed by Buildkite.  However, I found that GCE VM cleanup pipeline is finding and cleaning up VMs some times ([example](https://buildkite.com/elastic/elastic-agent-gce-cleanup/builds/3254#01990c04-c2b1-46ba-88e2-d325f0c102f8)).  So it would seem something is still creating the OGC VMs and leaving them orphaned?

ycombinator:
> Although, can you elaborate on what you mean by 'spun up in EC2'? IIUC, those VMs are managed by the BK provisioner Gobld, hence we don't need to clean them up - they are destroyed automatically after the step finished (regardless of the success)

Okay, I see you edited [your comment](https://github.com/elastic/elastic-agent/pull/10002#issuecomment-3321232244) to add the note about automatic cleanup regardless of success of the step.  In that case, it seems like we don't need this PR after all, which is great!

I do think we need to look into why the [GCE cleanup pipeline is finding VMs sometimes](https://github.com/elastic/elastic-agent/pull/10002#issuecomment-3321279676) but we can do that separately from this PR.  

Given all of the above, I will close this PR without merging, unless there are any objections.

v1v:
> Given all of the above, I will close this PR without merging, unless there are any objections.

Thanks Shaunak for walking me through the specifics here 🙇 

> it would seem something is still creating the OGC VMs and leaving them orphaned?

I see names that are not `elastic-agent` but other things, `name=ogc-linux-amd64-ubuntu-2404-fleet-endpoint-security-d2ce state=running ):`

Maybe, developers run those ogc commands locally. I have not investigated much the state of the art for `ogc` in the ingest projects though, but I agree we can do it in another PR or GH issues discussion.

Thanks!

PR #10003, opened 2025-09-17
feat: emit system resource metrics for EDOT subprocess

pkoutsovasilis:
## What does this PR do?

This PR introduces support for emitting **system resource metrics for the EDOT (Elastic Distribution of OpenTelemetry) collector** when it runs as a subprocess of the Elastic Agent.  

Key changes:
- Added a new `features.agent.otel.subprocess_execution` feature flag to control whether the OTel collector runs in **subprocess execution mode**.  
  - This flag **defaults to `false`** for now (maintaining existing behaviour), but is expected to default to `true` in the imminent future.  
- Extended the monitoring configuration generation logic to create a **dedicated HTTP metrics stream** for the EDOT subprocess, going by the name `elastic-agent/collector`, capturing only its system resources usage.  
- Updated `OTelManager` construction to honor the execution mode parsed from the feature flags, rather than always running in embedded mode.  

## Why is it important?

Running the collector as a subprocess improves **resilience** by isolating the control plane from the data plane. Emitting metrics for the EDOT process ensures **operational visibility**, allowing users to observe and troubleshoot its resource usage independently from the main Elastic Agent process.

## Checklist

## Disruptive User Impact

No disruptive impact expected.  
- When `features.agent.otel.subprocess_execution` is `false` (default), behavior is unchanged.  
- When the flag is enabled, users will see an additional monitoring stream for `elastic-agent/collector`.

1. Build Elastic Agent from this branch.
2. Enable the subprocess execution mode in your `elastic-agent.yml`:
   ```yaml
   agent:
     features:
       otel:
         subprocess_execution: true
   ```
3. Install elastic-agent.
4. Verify in Kibana’s Agent Metrics dashboard that a separate metrics stream for elastic-agent/collector appears. PS: you might have to install the elastic-agent integration if it's not already installed

Screenshot 2025-09-17 at 2 39 39 PM

## Related issues

N/A

### DIFF
--- a/internal/pkg/otel/config/config.go
+++ b/internal/pkg/otel/config/config.go
@@ -10,39 +10,37 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/otel/manager"
 )
 
-var executionMode = manager.EmbeddedExecutionMode
+const defaultExecMode = manager.EmbeddedExecutionMode
 
 type execModeConfig struct {
 	Agent struct {
 		Features struct {
 			Otel *struct {
-				SubprocessExecution bool `json:"subprocess_execution" yaml:"subprocess_execution" config:"subprocess_execution"`
+				SubprocessExecution *bool `json:"subprocess_execution,omitempty" yaml:"subprocess_execution,omitempty" config:"subprocess_execution,omitempty"`
 			} `json:"otel,omitempty" yaml:"otel,omitempty" config:"otel,omitempty"`
 		} `json:"features" yaml:"features" config:"features"`
 	} `json:"agent" yaml:"agent" config:"agent"`
 }
 
-// SetExecutionModeFromConfig sets the execution mode of the OTel runtime based on the config.
-func SetExecutionModeFromConfig(log *logp.Logger, conf *config.Config) {
+// GetExecutionModeFromConfig returns the execution mode of the OTel runtime based on the config.
+func GetExecutionModeFromConfig(log *logp.Logger, conf *config.Config) manager.ExecutionMode {
 	var c execModeConfig
 	if err := conf.UnpackTo(&c); err != nil {
-		log.Warnf("failed to unpack config in otel init execution mode: %v", err)
-		return
+		log.Warnf("failed to unpack config when getting otel runtime execution mode: %v", err)
+		return defaultExecMode
 	}
 
-	if c.Agent.Features.Otel != nil && c.Agent.Features.Otel.SubprocessExecution {
-		executionMode = manager.SubprocessExecutionMode
+	if c.Agent.Features.Otel == nil {
+		return defaultExecMode
+	}
+
+	if c.Agent.Features.Otel.SubprocessExecution == nil {
+		return defaultExecMode
+	}
+
+	if *c.Agent.Features.Otel.SubprocessExecution {
+		return manager.SubprocessExecutionMode
 	} else {
-		executionMode = manager.EmbeddedExecutionMode
+		return manager.EmbeddedExecutionMode
 	}
-}
-
-// GetExecutionMode returns the execution mode of the OTel runtime.
-func GetExecutionMode() manager.ExecutionMode {
-	return executionMode
-}
-
-// IsSubprocessExecution returns true if the OTel runtime is running in subprocess mode.
-func IsSubprocessExecution() bool {
-	return executionMode == manager.SubprocessExecutionMode
 }
--- a/internal/pkg/agent/application/application.go
+++ b/internal/pkg/agent/application/application.go
@@ -123,9 +123,8 @@ func New(
 		override(cfg)
 	}
 
-	otelconfig.SetExecutionModeFromConfig(log, rawConfig)
-	otelExecMode := otelconfig.GetExecutionMode()
-	isOtelSubprocessExecution := otelExecMode == otelmanager.SubprocessExecutionMode
+	otelExecMode := otelconfig.GetExecutionModeFromConfig(log, rawConfig)
+	isOtelExecModeSubprocess := otelExecMode == otelmanager.SubprocessExecutionMode
 
 	// monitoring is not supported in bootstrap mode https://github.com/elastic/elastic-agent/issues/1761
 	isMonitoringSupported := !disableMonitoring && cfg.Settings.V1MonitoringEnabled
@@ -133,7 +132,7 @@ func New(
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("failed to create upgrader: %w", err)
 	}
-	monitor := monitoring.New(isMonitoringSupported, cfg.Settings.DownloadConfig.OS(), cfg.Settings.MonitoringConfig, agentInfo, isOtelSubprocessExecution)
+	monitor := monitoring.New(isMonitoringSupported, cfg.Settings.DownloadConfig.OS(), cfg.Settings.MonitoringConfig, agentInfo, isOtelExecModeSubprocess)
 
 	runtime, err := runtime.NewManager(
 		log,
--- a/internal/pkg/agent/cmd/inspect.go
+++ b/internal/pkg/agent/cmd/inspect.go
@@ -30,6 +30,7 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/config/operations"
 	"github.com/elastic/elastic-agent/internal/pkg/diagnostics"
 	otelconfig "github.com/elastic/elastic-agent/internal/pkg/otel/config"
+	"github.com/elastic/elastic-agent/internal/pkg/otel/manager"
 	"github.com/elastic/elastic-agent/pkg/component"
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 	"github.com/elastic/elastic-agent/pkg/utils"
@@ -409,8 +410,9 @@ func getMonitoringFn(ctx context.Context, logger *logger.Logger, cfg map[string]
 	if err != nil {
 		return nil, fmt.Errorf("could not load agent info: %w", err)
 	}
-	otelconfig.SetExecutionModeFromConfig(logger, config)
-	monitor := monitoring.New(agentCfg.Settings.V1MonitoringEnabled, agentCfg.Settings.DownloadConfig.OS(), agentCfg.Settings.MonitoringConfig, agentInfo, otelconfig.IsSubprocessExecution())
+	otelExecMode := otelconfig.GetExecutionModeFromConfig(logger, config)
+	isOtelExecModeSubprocess := otelExecMode == manager.SubprocessExecutionMode
+	monitor := monitoring.New(agentCfg.Settings.V1MonitoringEnabled, agentCfg.Settings.DownloadConfig.OS(), agentCfg.Settings.MonitoringConfig, agentInfo, isOtelExecModeSubprocess)
 	return monitor.MonitoringConfig, nil
 }
 


internal/pkg/otel/manager/manager.go
```diff
@@ -109,6 +111,25 @@ type OTelManager struct {
...
+		return EmbeddedExecutionMode
+	}
+
+	if flags.OtelSubprocessExecution() {
+		return SubprocessExecutionMode
+	} else {
+		return EmbeddedExecutionMode
+	}
+}
+
```
swiatekm: These functions look like they should live in the features package instead. Are they here because the mode enum is defined here? I think I'd prefer to define the mode enum elsewhere (do we have a dedicated constants package?), and import it from there. Then we avoid the need to import the otel manager code in places that don't really need it.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -661,20 +667,28 @@ func (b *BeatsMonitor) getHttpStreams(
...
 		if !isSupportedMetricsBinary(binaryName) {
 			continue
 		}
 
 		endpoints := []interface{}{PrefixedEndpoint(BeatsMonitoringEndpoint(compInfo.ID))}
+		if compInfo.RuntimeManager == component.OtelRuntimeManager && compInfo.ID == httpMonitoringComponentInfoID && b.isOtelRuntimeSubprocess {
+			// when Otel runtime is running as subprocess, we need to monitor it as a separate stream, thus we utilise the already
+			// exposed endpoint of httpMonitoringComponentInfoID.
+			edotSubprocessEndpoints = endpoints
+		}
```
swiatekm: Does this mean we're actually collecting the same data twice, just processing it differently? Once as beat receiver metrics, and once as otel subprocess metrics? I think that's fine to do for now, but it might be worth an extended comment.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -661,20 +667,28 @@ func (b *BeatsMonitor) getHttpStreams(
...
+	var edotSubprocessEndpoints []interface{}
+
 	for _, compInfo := range componentInfos {
 		binaryName := compInfo.BinaryName
 		if !isSupportedMetricsBinary(binaryName) {
 			continue
 		}
 
 		endpoints := []interface{}{PrefixedEndpoint(BeatsMonitoringEndpoint(compInfo.ID))}
+		if compInfo.RuntimeManager == component.OtelRuntimeManager && compInfo.ID == httpMonitoringComponentInfoID && b.isOtelRuntimeSubprocess {
```
swiatekm: I don't think this condition is completely right. We can have self-monitoring running in beats processes, but other inputs running as beats receivers. With this condition, we wouldn't monitor the otel subprocess, even though we can and should. I think it should be logically `If self-monitoring metrics are enabled, otel is running as a subprocess, and any component uses the otel runtime`. A check similar to what @faec added in https://github.com/elastic/elastic-agent/pull/9928.

pkoutsovasilis: Valid point, for some reason I thought that we are switching to otel runtime by default for the monitoring needs without supporting beats subprocess, but this was never agreed. And that means that I am gonna have to go down the path of injecting my own monitoring server in edot collector no matter what. thx

internal/pkg/agent/application/monitoring/v1_monitor.go
pkoutsovasilis: yes but the processors need to be different and I am not sure if these sets of processors are compatible to be merged in one. E.g. in the ones regarding resource metrics the component.id has to be `elastic-agent/collector` and in the other metrics ones it derives from the component ID. With that said, what comment would you like to see here?

internal/pkg/otel/manager/manager.go
pkoutsovasilis: continue with your proposal and please say in which package specific you want the mode constants to live in

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: I don't think you need to do anything special here. If there's a component running in the otel collector, then the beat receiver for that component should have a monitoring endpoint open, and you can just use that, the same way you're using the monitoring component right now. Is there a problem with doing that?

pkoutsovasilis: what if we run edot collector from the pure otel config without any beat receiver? we should monitor that as well right?

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: The comment should acknowledge that we're ingesting the same data twice with different metadata - once for the receiver and once for the otel subprocess. I'd like this to be as clear as possible from the code itself, because it's unintuitive. I had to stare at these lines for a minute before I was confident this was correct, and I spent a lot of time working with this part of the codebase. I want to spare our descendants the confusion.

internal/pkg/agent/application/monitoring/v1_monitor.go
swiatekm: I suppose we could have an otel collector running only hybrid-mode configuration, in which case you do need the monitoring server after all.

pkoutsovasilis: I am thinking about hybrid elastic-agent, and I believe since we can have otel-specific config in elastic-agent.yml this means that there is a possibility of edot collector running without any beat receiver, right?

swiatekm: Yes, that can happen. And it would be a regression, because with embedded mode, we get these metrics as part of the elastic-agent process.

cmacknz:
> Extended the monitoring configuration generation logic to create a dedicated HTTP metrics stream for the EDOT subprocess, going by the name elastic-agent/collector, capturing only its system resources usage.

From what I know/remember, I think this metrics.elastic-agent/collector-* index will have to be in the list of well known monitoring indices in Fleet defined [here](https://github.com/elastic/kibana/blob/eed21dbca87029b245c44f5d8d4d64c9f491c877/x-pack/platform/plugins/shared/fleet/common/constants/agent_policy.ts#L17-L35) in Kibana, and it'll also have to be defined as a data stream in the [elastic_agent package](https://github.com/elastic/integrations/tree/main/packages/elastic_agent/data_stream).

I do not have the entire history on why the monitoring indices need to be defined in Kibana, but I don't think we'll have Fleet managed output API keys generated with permissions to write to the new data stream without that.

Edit: The test that would prove the need for this is to make sure this data stream can be written to for a Fleet managed agent.

internal/pkg/otel/manager/manager.go
cmacknz: Keep in mind is that the intent is that the sub-process feature flag will be very short lived, it's existence is to allow us to quickly turn it off if we discover there's a problem as the change is deployed without needing to wait for the revert PR + snapshot build time.

Once we are convinced there are no bugs, we can remove this flag as sub-process will be the default execution mode.

cmacknz:
There is an existing test that I think could be easily updated to prove this works e2e for Fleet managed agents, by expecting the elastic-agent/collector component to have system metrics populated.

https://github.com/elastic/elastic-agent/blob/58460e78a12469bfe86ca0a01a74ba81afe59402/testing/integration/ess/metrics_monitoring_test.go#L95-L103

internal/pkg/otel/manager/manager.go
cmacknz: What that means IMO is if we restructure things it should be to make this easier to remove later.

internal/pkg/agent/application/monitoring/v1_monitor.go
```diff
@@ -106,14 +108,15 @@ type monitoringConfig struct {
...
 	return &BeatsMonitor{
 		enabled: enabled,
 		config: &monitoringConfig{
 			C: cfg,
 		},
-		operatingSystem: operatingSystem,
-		agentInfo:       agentInfo,
+		operatingSystem:         operatingSystem,
+		agentInfo:               agentInfo,
+		isOtelRuntimeSubprocess: isOtelRuntimeSubprocess,
```
swiatekm, line 119: I think you need to set this in the `Reload` method as well. Otherwise a change to the feature flag will not be visible after a config reload.

internal/pkg/agent/application/monitoring/v1_monitor.go
pkoutsovasilis: ok with the changes here https://github.com/elastic/elastic-agent/pull/10003/commits/cc1dab8275bb6b58e3c7b36614499d3844497d43 this is no longer the case and EDOT subprocess got a dedicated metrics server to extract system metrics for the process

internal/pkg/otel/manager/manager.go
pkoutsovasilis: so I extracted everything in a separate `config` package that is easier to remove when the subprocess mode becomes the default one without being able to disable it (soonish), take a look here https://github.com/elastic/elastic-agent/pull/10003/commits/4cd10aa641a5e298dabed246f1f55b98fbcd6afa

internal/pkg/otel/monitoring/monitoring.go
```diff
@@ -0,0 +1,134 @@
...
+		Logger:         log,
+		EphemeralID:    ephemeralID,
+		SystemMetrics:  monitoring.Default.GetOrCreateRegistry("system"),
+		ProcessMetrics: monitoring.Default.GetOrCreateRegistry("beat"),
+	}); err != nil {
+		return nil, fmt.Errorf("failed to setup metrics: %w", err)
+	}
+
+	r := mux.NewRouter()
+	r.HandleFunc("/stats", func(w http.ResponseWriter, r *http.Request) {
```
swiatekm, line 56: Why can't we use the same stats handler as the agent's monitoring server?

pkoutsovasilis:
> From what I know/remember, I think this metrics.elastic-agent/collector-* index will have to be in the list of well known monitoring indices in Fleet defined [here](https://github.com/elastic/kibana/blob/eed21dbca87029b245c44f5d8d4d64c9f491c877/x-pack/platform/plugins/shared/fleet/common/constants/agent_policy.ts#L17-L35) in Kibana, and it'll also have to be defined as a data stream in the [elastic_agent package](https://github.com/elastic/integrations/tree/main/packages/elastic_agent/data_stream).
> 
> I do not have the entire history on why the monitoring indices need to be defined in Kibana, but I don't think we'll have Fleet managed output API keys generated with permissions to write to the new data stream without that.
> 
> Edit: The test that would prove the need for this is to make sure this data stream can be written to for a Fleet managed agent.

Look, the resource metrics for the EDOT subprocess are being written to the existing elastic-agent.elastic-agent dataset, so we don’t need to make any changes on the Kibana side. I didn’t do anything special, and the metrics show up correctly in the Agent Metrics dashboard.
Screenshot 2025-09-24 at 6 00 01 PM

> There is an existing test that I think could be easily updated to prove this works e2e for Fleet managed agents, by expecting the elastic-agent/collector component to have system metrics populated.

As discussed offline with @cmacknz, since this setting is short-lived, we’ll handle the end-to-end testing in a follow-up PR when subprocess execution mode becomes the default.

internal/pkg/agent/application/monitoring/v1_monitor.go
pkoutsovasilis: this is the thing though, otel execution mode is not meant to be changed at runtime, since this is defined at the instantiation time of the former [here](https://github.com/elastic/elastic-agent/blob/4cd10aa641a5e298dabed246f1f55b98fbcd6afa/internal/pkg/agent/application/application.go#L248). Thus the only way to change the mode effectively is to update the config and then restart elastic-agent. Does this makes sense?

internal/pkg/otel/monitoring/monitoring.go
pkoutsovasilis: My initial effort was actually to utilise the same code as found [here](https://github.com/elastic/elastic-agent/blob/4cd10aa641a5e298dabed246f1f55b98fbcd6afa/internal/pkg/agent/application/monitoring/server.go#L30). However this code is not modular so we would get way more than just `/stats` but more importantly it doesn't allow me to control the socket/npipe path. Now instead of making the existing code modular and, thus, risking of changing something that might lead to issues with code that currently works, I chose to extract what was needed from there and invoke a monitoring server with just the `/stats` path exposed. Also keeping in mind that this endpoint might become part of EDOT diagnostics extension, this seemed like the less radical way to proceed. WDYT?

internal/pkg/agent/application/monitoring/v1_monitor.go
cmacknz: The intent of being able to change the otel execution mode is to mitigate the chance we have a problem while deploying it and have to waste 12 hours doing a revert + waiting for a new snapshot. 

People having problems with the Otel collector in general will just toggle the _runtime_experimental flags which do work at runtime.

I'm fine with not reloading this flag and we don't intend for it to be a permanent flag, it's just there as a quick fallback in case of problems during initial deployment internally. I am much more motivated to get this deployed than to try to make the configuration flag perfect.

pkoutsovasilis:
> Changes look good to me, waiting for ci to be green

@kaanyalti the CI won't get green as the `Merge coverage reports` step is failing with this error 

```
2025/09/23 17:44:57 OVERLAP MERGE: github.com/elastic/elastic-agent/internal/pkg/otel/components.go {88 43 118 17 6 68} {88 43 122 17 7 5}
```

which I am not quite sure what causes that but I am fairly confident that isn't an error due to the functionality introduced by this PR

internal/pkg/otel/monitoring/monitoring.go
swiatekm: Allright, that works.

swiatekm (dismissed):

kaanyalti (approved):

internal/pkg/otel/config/config.go
```diff
@@ -0,0 +1,48 @@
...
+
+package config
+
+import (
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/elastic-agent/internal/pkg/config"
+	"github.com/elastic/elastic-agent/internal/pkg/otel/manager"
+)
+
+var executionMode = manager.EmbeddedExecutionMode
```
blakerouse: Does this really need to be a global? Could we find a way of not making this a global and instead pass this information around.

blakerouse (commented):
Not going to block this PR from merging as it looks good, other than my 1 inline comment.

cmacknz:
Built locally and can see it's shipping metrics 👍 

I did notice when inspecting diagnostics that we don't get profiles from the collector sub-process, so I created https://github.com/elastic/elastic-agent/issues/10135 to track including these separately from this PR and added it to the parent issue for the sub-process conversion.

internal/pkg/otel/config/config.go
pkoutsovasilis: ok valid point and I am not in favour of globals, so lemme check if this can be altered (I think it does)

pkoutsovasilis:
> Built locally and can see it's shipping metrics 👍
> 
> I did notice when inspecting diagnostics that we don't get profiles from the collector sub-process, so I created #10135 to track including these separately from this PR and added it to the parent issue for the sub-process conversion.

Thanks for creating the tracking issue in my mind, this could/should be covered by the EDOT diagnostics extension [PR here](https://github.com/elastic/elastic-agent/pull/10052)

cmacknz:
OK I'm not as concerned as which piece of work it is part of, more with ensuring it gets implemented.

internal/pkg/otel/config/config.go
pkoutsovasilis: changed in here https://github.com/elastic/elastic-agent/pull/10003/commits/c181a17403c5d536ece6c2e020f0bcd71bf38ce2

cmacknz:
Out of curiosity I tried to hit the unix socket myself, and I'm getting a 404 on `/stats` but not on `/debug/stats`. The metricbeat modules hit `/stats` and seem to generate data, what am I missing here?

```
❯ ps aux | rg otel
root             96349   0.0  0.3 412492384 154320   ??  S     4:46PM   0:07.76 /Library/Elastic/Agent-Development/elastic-agent otel --supervised --supervised.logging.level=info --supervised.monitoring.url=unix:///Library/Elastic/Agent-Development/data/tmp/jTgjFSdPDnYoazUu541NazzsJSv6daDn.sock

❯ sudo curl --unix-socket /Library/Elastic/Agent-Development/data/tmp/jTgjFSdPDnYoazUu541NazzsJSv6daDn.sock http:/sta
ts
404 page not found

❯ sudo curl --unix-socket /Library/Elastic/Agent-Development/data/tmp/jTgjFSdPDnYoazUu541NazzsJSv6daDn.sock http:/debug/stats
{"beat":{"cpu":{"system":{"ticks":3405,"time":{"ms":3405}},"total":{"ticks":7885,"time":{"ms":7885},"value":7885.641791666666},"user":{"ticks":4480,"time":{"ms":4480}}},"info":{"ephemeral_id":"f42f9139-32de-4226-b9cd-b4b214fa053f","name":"elastic-agent/c","uptime":{"ms":3267401},"version":"9.2.0"},"memstats":{"gc_next":52060306,"memory_alloc":26579376,"memory_sys":67912968,"memory_total":462099568,"rss":158072832},"runtime":{"goroutines":173}},"filebeat":{"harvester":{"closed":1,"gzip_closed":0,"gzip_open_files":0,"gzip_running":0,"gzip_started":0,"open_files":1,"running":1,"skipped":0,"started":2},"input":{"log":{"files":{"renamed":0,"truncated":0}}}},"libbeat":{"config":{"module":{"running":0,"starts":0,"stops":0},"reloads":0,"scans":0}},"processor":{"add_host_metadata":{"fqdn_lookup_failed":0}},"registrar":{"states":{"cleanup":0,"current":0,"update":0},"writes":{"fail":0,"success":0,"total":0}},"system":{"cpu":{"cores":14},"load":{"1":2.3857,"15":2.1997,"5":2.2944,"norm":{"1":0.1704,"15":0.1571,"5":0.1639}}}}%
```

internal/pkg/agent/cmd/otel.go
```diff
@@ -102,18 +118,23 @@ func RunCollector(cmdCtx context.Context, configFiles []string, supervised bool,
 	defer cancel()
 	go service.ProcessWindowsControlEvents(stopCollector)
 
-	return otel.Run(ctx, stop, settings)
+	return otel.Run(ctx, stop, settings.otelSettings)
```
cmacknz, line 121: Unrelated to this, but `go service.ProcessWindowsControlEvents(stopCollector)` above I'm not sure is actually correct when this is running as a sub-process where it isn't able to be a Windows service itself

pkoutsovasilis:
@cmacknz I think that this is quirk of `curl` and unix-sockets. Invoking it like that `curl --unix-socket /usr/share/elastic-agent/data/tmp/jTgjFSdPDnYoazUu541NazzsJSv6daDn.sock http://localhost/stats` works as expected

```bash
# curl --unix-socket /usr/share/elastic-agent/data/tmp/jTgjFSdPDnYoazUu541NazzsJSv6daDn.sock http://localhost/stats
{"beat":{"cgroup":{"cpu":{"id":"/","stats":{"periods":0,"throttled":{"ns":0,"periods":0}}},"memory":{"id":"/","mem":{"usage":{"bytes":4232740864}}}},"cpu":{"system":{"ticks":6780,"time":{"ms":6780}},"total":{"ticks":34140,"time":{"ms":34140},"value":34140},"user":{"ticks":27360,"time":{"ms":27360}}},"handles":{"limit":{"hard":1048576,"soft":1048576},"open":33},"info":{"ephemeral_id":"893ffa32-6758-4393-9e32-4f627378b4a1","name":"elastic-agent/c","uptime":{"ms":3293315},"version":"9.2.0"},"memstats":{"gc_next":57802386,"memory_alloc":32886168,"memory_sys":84828440,"memory_total":826911448,"rss":285278208},"runtime":{"goroutines":265}},"filebeat":{"harvester":{"closed":0,"gzip_closed":0,"gzip_open_files":0,"gzip_running":0,"gzip_started":0,"open_files":1,"running":1,"skipped":0,"started":1},"input":{"log":{"files":{"renamed":0,"truncated":0}}}},"libbeat":{"config":{"module":{"running":0,"starts":0,"stops":0},"reloads":0,"scans":0}},"processor":{"add_host_metadata":{"fqdn_lookup_failed":0}},"registrar":{"states":{"cleanup":0,"current":0,"update":0},"writes":{"fail":0,"success":0,"total":0}},"system":{"cpu":{"cores":12},"load":{"1":0.19,"15":0.29,"5":0.19,"norm":{"1":0.0158,"15":0.0242,"5":0.0158}}}}
```

so it seems that the domain part is extracted even if it is a random value as debug in your case or localhost in mine

cmacknz (approved):
Besides my inability to hit /stats locally despite agent generate data this LGTM, I did notice we are probably unnecessarily registering to process Windows service control events as a sub-process but I'm not sure if that is harmful or not.

cmacknz:
> curl --unix-socket /usr/share/elastic-agent/data/tmp/jTgjFSdPDnYoazUu541NazzsJSv6daDn.sock http://localhost/stats

OK yes I remember this now, thanks that does what I expected.

internal/pkg/agent/cmd/otel.go
pkoutsovasilis: good point and indeed this seems redundant in the subprocess execution; I will investigate a little bit more just to make sure that we are not disabling something critical and I will move that out when EDOT is running as supervised process in the follow up PR that sets subprocess as default (it's kinda straight forward)

internal/pkg/otel/config/config.go
blakerouse: Thanks for changing it.

blakerouse (approved):
Looks good.

Issue #9890, opened 2025-09-11
[Flaky Test]: TestBeatsReceiverLogs – Error:  Condition never satisfied


pchila:
### Failing test case

TestBeatsReceiverLogs

### Error message

A series of asserts is failing regarding beats receivers logs is failing, leaving an elastic agent process running

### Build

https://buildkite.com/elastic/elastic-agent/builds/26642#01993909-3d3d-42d4-a960-c6046ee71593/319-691

### OS

Linux

### Stacktrace and notes

```markdown
2025-09-11 14:26:32 UTC
	    beat_receivers_test.go:734:
2025-09-11 14:26:32 UTC
	        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-aws-1757598785638537982/elastic/elastic-agent/testing/integration/ess/beat_receivers_test.go:810
2025-09-11 14:26:32 UTC
	        	            				/opt/buildkite-agent/builds/bk-agent-prod-aws-1757598785638537982/elastic/elastic-agent/testing/integration/ess/beat_receivers_test.go:738
2025-09-11 14:26:32 UTC
	        	            				/opt/buildkite-agent/.asdf/installs/golang/1.24.7/go/src/runtime/asm_arm64.s:1223
2025-09-11 14:26:32 UTC
	        	Error:      	Not equal:
2025-09-11 14:26:32 UTC
	        	            	expected: 2
2025-09-11 14:26:32 UTC
	        	            	actual  : 0
2025-09-11 14:26:32 UTC
	    beat_receivers_test.go:734:
2025-09-11 14:26:32 UTC
	        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-aws-1757598785638537982/elastic/elastic-agent/testing/integration/ess/beat_receivers_test.go:811
2025-09-11 14:26:32 UTC
	        	            				/opt/buildkite-agent/builds/bk-agent-prod-aws-1757598785638537982/elastic/elastic-agent/testing/integration/ess/beat_receivers_test.go:738
2025-09-11 14:26:32 UTC
	        	            				/opt/buildkite-agent/.asdf/installs/golang/1.24.7/go/src/runtime/asm_arm64.s:1223
2025-09-11 14:26:32 UTC
	        	Error:      	Not equal:
2025-09-11 14:26:32 UTC
	        	            	expected: "beat-v2-client"
2025-09-11 14:26:32 UTC
	        	            	actual  : ""
[...]

PR #10030, opened 2025-09-18
Improve observability for TestBeatsReceiverLogs

swiatekm:
## What does this PR do?

Improves the observability and resource cleanup in the TestBeatsReceiverLogs integration test. This test is somewhat unusual in that it wants to run agent twice with different configurations, verifying that the emitted logs are the same. Our integration test fixtures don't support that very well, so we instead prepare two separate fixtures for each execution. This, however, prevented us from calling `fixture.Install`, which does a lot of nice things, including cleanup and collection of diagnostics if the test fails. This change ensures that the same cleanup is applied to TestBeatsReceiverLogs.

## Why is it important?

This test can be flaky sometimes, and it's difficult to understand the reason without diagnostics.

## Checklist

## Related issues

- Resolves https://github.com/elastic/elastic-agent/issues/9890

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -738,6 +738,7 @@ agent.monitoring.enabled: false
 	fixture, cmd, processOutput := prepareAgentCmd(t, agentProcessCtx, processConfig)
 
 	require.NoError(t, cmd.Start())
+	fixture.RegisterCleanup()
 	t.Cleanup(func() {
 		agentProcessCancel()
 		_ = cmd.Wait() // second wait returns an error, which we ignore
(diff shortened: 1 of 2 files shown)


testing/integration/ess/beat_receivers_test.go
```diff
@@ -737,9 +735,18 @@ agent.monitoring.enabled: false
 
 	// use a subcontext for the agent
 	agentProcessCtx, agentProcessCancel := context.WithCancel(ctx)
-	fixture, cmd, output := prepareAgentCmd(t, agentProcessCtx, processConfig)
+	fixture, cmd, processOutput := prepareAgentCmd(t, agentProcessCtx, processConfig)
 
 	require.NoError(t, cmd.Start())
+	fixture.RegisterCleanup()
```
pchila, line 741: Instead of registering only the cleanup functions here,  is it not possibleto use the normal Install() and  run with/without beats receivers as subtests and then compare in the parent test ? Are install logs compared as well ?

swiatekm: Installing makes agent run as a service, which makes fetching logs locally more involved, especially given the test runs on all three operating systems. I think the right thing to do here is probably to just spin up a stack and query the logs from ES based on agent id. That feels like overkill, but it would make the test much simpler.

swiatekm: Hm, I wasn't aware we have a `logs` command :sweat_smile:. If that works as expected, I should be able to do what you propose.

swiatekm: I rewrote this test using standard integration test fixtures in https://github.com/elastic/elastic-agent/pull/10153. I'm going to close this PR. Thanks for the hint, Paolo!

swiatekm:
Closing in favor of https://github.com/elastic/elastic-agent/pull/10153.

PR #10052, opened 2025-09-19
[otel] Implement EDOT diagnostics extension

VihasMakwana:
## What does this PR do?

This PR add a new elasticdiagnsotics extension. This new extension will be used in hybrid mode and will perform diagnostics for running otel collector. It will:
- Acts as a registrar and keeps track of common diagnostic hooks.
- Collects profiles using `runtime/pprof`.
- Collects internal telemetry exposed by the OTeL Collector.
- Implements the `extensioncapabilities.ConfigWatcher` interface and stores the latest configuration of the running collector.
- Listens for diagnostic requests and provides diagnostic data. 

## Checklist

## Disruptive User Impact

None that i can think of. I have to do some final testing on windows to verify this.

## Related issues

- Relates https://github.com/elastic/beats/issues/44604


internal/pkg/otel/translate/otelconfig.go
```diff
@@ -279,6 +279,8 @@ func getReceiversConfigForComponent(
 			},
 		}
 	}
+	// indicate that beat receivers are managed by the elastic-agent
+	receiverConfig["management.otel.enabled"] = true
```
VihasMakwana, line 313: @cmacknz We'll set the `management.otel.enabled` here. Related to our convo: https://github.com/elastic/beats/pull/46539/#pullrequestreview-3241511599

internal/pkg/otel/elasticdiagnosticsextension/config.go
```diff
@@ -0,0 +1,25 @@
+package elasticdiagnosticsextension
```
cmacknz: The extension should have a readme document its design, ideally with a simple diagram showing the socket communication paths

internal/pkg/otel/elasticdiagnosticsextension/config.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type Config struct {
+	Host    string `mapstructure:"host"`
+	Network string `mapstructure:"network"`
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{
+		Network: "unix",
```
cmacknz: I don't think this is correct for Windows where the protocol scheme would be `npipe` for example https://github.com/elastic/elastic-agent/blob/351576b2dd3b2c6cdc25b391ef253bcc292c74e6/internal/pkg/agent/application/paths/common.go#L42-L43

internal/pkg/otel/elasticdiagnosticsextension/factory.go
```diff
@@ -0,0 +1,26 @@
...
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/extension"
+)
+
+var (
+	DiagnosticsExtensionID component.Type = metadata.Type
+)
+
+func NewFactory() extension.Factory {
+	return extension.NewFactory(metadata.Type, createDefaultConfig, newExtension, component.StabilityLevelDeprecated)
```
cmacknz: Stability level probably shouldn't be deprecated to start :)

internal/pkg/otel/elasticdiagnosticsextension/config.go
cmacknz: Ah, TIL https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

Windows has supported AF_UNIX directly since Windows 10.

https://pkg.go.dev/internal/syscall/windows defines `SupportUnixSocket` and says:

> SupportUnixSocket indicates whether the current Windows version supports Unix Domain Sockets. The minimal requirement is Windows 10.0.17063.

We do have `Windows Server 2012/R2` in the Elastic Agent support matrix along with unconditional support of all Windows 10 versions https://www.elastic.co/support/matrix

There is a chance this will not work for some set of the user base, we have continual asks to keep supporting older windows versions so I think we probably have to use a named pipe here instead like we do everywhere else to be safe.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,207 @@
...
+		w.WriteHeader(503)
+		w.Header().Add("content-type", "application/json")
+		fmt.Fprintf(w, "{'error':'%v'}", err)
+		return
+	}
+	w.Header().Add("content-type", "application/json")
+	w.Write(b)
+}
+
+func extractMetricAddress(readers []otelconf.MetricReader) string {
```
cmacknz: If we switched internal telemetry to use the OTLP exporter instead we'd lose this?

internal/pkg/otel/elasticdiagnosticsextension/config.go
cmacknz: Our other control socket is gRPC which will do the right thing with named pipes so that could push us to gRPC instead of HTTP, but I think HTTP is nicer better because it's easier to interact with manually on the command line.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (d *diagnosticsExtension) Shutdown(ctx context.Context) error {
+	if err := d.server.Shutdown(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *diagnosticsExtension) NotifyConfig(ctx context.Context, conf *confmap.Conf) error {
+	d.collectorConfig = conf
```
cmacknz, line 150: This update can happen concurrently with attempts to read it, you should mutex protect it.

internal/pkg/otel/elasticdiagnosticsextension/extension.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	go func() {
+		if err := d.server.Serve(d.listener); err != nil && err != http.ErrServerClosed {
+			d.logger.Error("HTTP server error", zap.Error(err))
+		}
+	}()
+
+	d.globalHooks["collector_config"] = &diagHook{
+		description: "full collector configuration",
+		filename:    "otel-merged.yaml",
```
cmacknz: Moving this in here means we can't get it if the collector is failing to start, because the extension isn't running which is concerning if a configuration error caused the problem.

It was safer to have this where it was before I think.

We may actually want both versions, the one agent wants to give to the collector and this one here which is the last config the collector picked up itself.

@swiatekm do you have opinions here?

internal/pkg/otel/translate/otelconfig.go
cmacknz: I'll mark https://github.com/elastic/elastic-agent/issues/10069 as being closed by this PR then.

internal/pkg/otel/server.go
```diff
@@ -0,0 +1,133 @@
...
+	listener net.Listener
+	server   *grpc.Server
+	logger   *logp.Logger
+}
+
+func NewServer() *server {
+	return &server{logger: logp.L()}
+}
+
+func (s *server) Start() error {
```
cmacknz: Why do we need this server? 

Why isn't the flow `elastic-agent command` --existing gRPC diagnostics request--> `elastic-agent service` --http--> `collector sub-process`?

Can't the elastic-agent process supervising the collector know to hit the collector's diagnostics extension without needing a separate gRPC server for it?

And on the command line without installing the EDOT collector via the `elastic-agent install` command, you'd just use curl or whatever to get diagnostics manually?

internal/pkg/otel/extension/elasticdiagnostics/extension.go
cmacknz: Reference: https://github.com/elastic/elastic-agent/pull/9928#discussion_r2349876089

internal/pkg/otel/elasticdiagnosticsextension/factory.go
VihasMakwana: 🥲 fixed it now.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: We can have a prometheus based exporter and an OTLP based exporter simultaneously, to export the internal telemetry. https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics.

**_If_** we decide to keep **_only_** OTLP, then yes, we would lose this. But I don't see a reason why we should keep only OTLP.

internal/pkg/otel/server.go
VihasMakwana: This server is necessary for `elastic-agent otel diagnostics` -> `elastic-agent otel` process communication. 

If the user is running a standalone `elastic-agent otel` process in one terminal and if they want to diagnostic the EDOT, they can run `elastic-agent otel diagnostics` to get only EDOT specific diagnostics. 

This server will not be created/used if we're operating elastic-agent in hybrid mode, as a service.

internal/pkg/otel/elasticdiagnosticsextension/extension.go
VihasMakwana: I see. I think we can have both.

internal/pkg/otel/elasticdiagnosticsextension/config.go
VihasMakwana: Done. Let me know if you would like any adjustments in README.

internal/pkg/agent/cmd/diagnostics_otel.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/elastic/elastic-agent/internal/pkg/cli"
+	"github.com/elastic/elastic-agent/internal/pkg/diagnostics"
+	"github.com/elastic/elastic-agent/pkg/control"
+	"github.com/elastic/elastic-agent/pkg/control/v2/client"
+)
+
+func newOtelDiagnosticsCommand(streams *cli.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "diagnostics",
+		Short: "Start the Elastic Agent in otel mode",
```
blakerouse: This says it starts the elastic agent in otel mode? That is not correct.

internal/pkg/agent/cmd/diagnostics_otel.go
```diff
@@ -0,0 +1,62 @@
...
+		Long:  "This command starts the Elastic Agent in otel mode.",
```
blakerouse: I believe this is incorrect.

internal/pkg/agent/cmd/otel.go
```diff
@@ -81,6 +84,16 @@ func RunCollector(cmdCtx context.Context, configFiles []string, supervised bool,
 	if err != nil {
 		return fmt.Errorf("failed to prepare collector settings: %w", err)
 	}
+
+	// create and start a new server that will listen for the diagnostics requests.
+	server := otel.NewServer()
```
blakerouse: Should this really be started here or should it be encapsulated into an extension that is started by the collector runtime?

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -305,6 +305,60 @@ func ZipArchive(
 	return zipLogs(zw, ts, topPath, excludeEvents)
 }
 
+func ZipArchiveEDOT(
```
blakerouse: This is almost the same function signature as `ZipArchive`. Why do you need this specific one? Being that the types are the same, why not just make it re-use the same function.

internal/pkg/otel/elasticdiagnosticsextension/config.go
blakerouse: This should definitely use npipe on Windows, even if its just to ensure consistency.

internal/pkg/otel/elasticdiagnosticsextension/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"errors"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+type Config struct {
+	Host    string `mapstructure:"host"`
+	Network string `mapstructure:"network"`
```
blakerouse: Is this standard otel structure for a configuration of an endpoint? Just curious, I generally do not know.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,252 @@
...
+			err := d.collectorConfig.Unmarshal(&serviceCfg)
+			if err != nil {
+				return fmt.Appendf(nil, "error: failed to get internal telemetry: %v", err)
+			}
+			if serviceCfg.Service.Telemetry.Metrics.Level == configtelemetry.LevelNone {
+				return []byte("internal telemetry is disabled")
+			}
+			addr := extractMetricAddress(serviceCfg.Service.Telemetry.Metrics.Readers)
+
+			resp, err := http.Get(fmt.Sprintf("http://%s/metrics", addr))
```
blakerouse: Is this the only way to get this? Basically this is performing a GRPC call to get diagnostics that internally will call a HTTP endpoint? That seems incorrect.

There is no timeout defined here or context handling. This would need to be improved to at least do those things.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (d *diagnosticsExtension) Shutdown(ctx context.Context) error {
+	if err := d.server.Shutdown(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *diagnosticsExtension) NotifyConfig(ctx context.Context, conf *confmap.Conf) error {
```
blakerouse, line 147: This means that the extension can handle a new configuration? What if the `network` or `host` is different? Doesn't that need to be handled here?

blakerouse (changes requested):

internal/pkg/agent/cmd/otel.go
VihasMakwana: I'm giving this a second thought and we might not need this. 
We can directly call the extension's endpoint when we run `elastic-agent otel diagnostics` command. Wdyt?

internal/pkg/otel/elasticdiagnosticsextension/config.go
VihasMakwana: Kind of, yeah. Here's a complete struct of otel equivalent. https://github.com/open-telemetry/opentelemetry-collector/blob/a56655d1b5a074f9cf01ca727e5db51da3e9ac8a/config/confighttp/server.go#L34-L95

It doesn't support unix sockets, so I didn't reuse that.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: > Is this the only way to get this? Basically this is performing a GRPC call to get diagnostics that internally will call a HTTP endpoint? That seems incorrect.

Yeah, this is what I meant in my earlier [comment](https://github.com/elastic/elastic-agent/pull/10052#discussion_r2369925801). We can directly perform the HTTP call in the new subcommand. I think that would be better. What do you think?

> There is no timeout defined here or context handling. This would need to be improved to at least do those things.

Yes, that and there's also some work required to fix socket permissions. I'll do that.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: > This means that the extension can handle a new configuration?

No. By adding this method, we implement the [ConfigWatcher](https://pkg.go.dev/go.opentelemetry.io/collector/extension/extensioncapabilities#ConfigWatcher) interface and the collector will basically give us the latest running configuration of the collector. We, as an extension, store the config and return it when a diagnostic request is made.

blakerouse: Ah, thanks for the clarity.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
blakerouse: I would prefer to see the diagnostic command directly perform the HTTP request.

internal/pkg/otel/elasticdiagnosticsextension/config.go
blakerouse: That doesn't look like `network` is really used. It would probably be better to use `http://` or `unix://` or `npipe://` to define the protocol to use. I also prefer `endpoint` over host, similar to the link you provided.

internal/pkg/agent/cmd/otel.go
blakerouse: I think that would be better.

cmacknz: +1, similar comment thread in https://github.com/elastic/elastic-agent/pull/10052#discussion_r2364425869

Start simpler and we can add on a control socket style implementation if there's a need for it later.

internal/pkg/agent/cmd/diagnostics_otel.go
VihasMakwana: Done

internal/pkg/otel/elasticdiagnosticsextension/config.go
VihasMakwana: @blakerouse @cmacknz I've reused the existing `github.com/elastic/elastic-agent/pkg/ipc` package to get the listener. That package takes care of npipe/unix nuances and it is already used in v2 protocol.

internal/pkg/otel/server.go
VihasMakwana: Not applicable anymore as we don't create this server.

internal/pkg/otel/extension/elasticdiagnostics/README.md
```diff
@@ -0,0 +1,55 @@
+# elasticdiagnosticsextension
```
cmacknz, line 1: I think this whole extension should be in an extension sub-directory and we should mirror the directory structure of contrib like we did in https://github.com/elastic/beats/tree/d3be9bf1567bebaf08361c194d325c4038b43618/x-pack/otel (you can copy over the top level readme there).

This should become internal/pkg/otel/extension/elasticdiagnostics

internal/pkg/diagnostics/diagnostics.go
VihasMakwana: Hi!

The `ZipArchive` deals with three layers: Agent level diagnostics, component level diagnostics and unit level diagnostics. 

In EDOT, we only deal with Agent level diagnostics and component level diagnostics. 

I've made a few changes to `ZipArchive` for it to work with EDOT.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
swiatekm: I checked if there was a more direct way of getting the metrics, but it looks like there isn't a way to pass a custom metrics reader into the telemetry settings. So we either need to hit the Prometheus endpoint, or start an OTLP server in the extension and have the collector write to it over. Probably simplest to do as Blake says and make the request from the command.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,248 @@
...
+	d.confgMtx.Lock()
+	defer d.confgMtx.Unlock()
+	d.collectorConfig = conf
+	return nil
+}
+
+func (d *diagnosticsExtension) RegisterDiagnosticHook(componentName string, description string, filename string, contentType string, hook func() []byte) {
```
swiatekm, line 156: This the API we expose for beat receivers to register their hooks, correct? We should add a comment noting this, so it doesn't get haphazardly changed in the future.

blakerouse:
@VihasMakwana  I agree with @swiatekm that splitting the extension into its own PR and then a followup to use it would make this easier to review. At the moment its a very large PR.

internal/pkg/otel/elasticdiagnosticsextension/extension.go
swiatekm: Both sounds good to me as long as we're clear what they each represent.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
blakerouse: I assume this means this interface is searched for and if it implements it then it can be performed. If that is the case then something like this should be added to the file, to ensure that if that interface changes that it is caught at compile time and not just silently removed.

`var _ {interfaceName} = (*diagnosticsExtension)(nil)`

blakerouse (commented):
Overall this is looking better. I do have one comment about the interface.

internal/pkg/diagnostics/diagnostics.go
```diff
@@ -226,8 +226,9 @@ func ZipArchive(
 	if err != nil {
 		return fmt.Errorf("error creating .zip header for components/ directory: %w", err)
 	}
+
 	// iterate over components
-	for dirName, units := range compDirs {
+	for dirName, comp := range componentResults {
```
VihasMakwana, line 231: I am essentially reversing the order in which we loop through units/components. Previously, we looped through units and then we find the component for that unit. Now, we loop through components first and then go through units for a given component.

As EDOT (or OTel) doesn't have a concept of units, we need to update the `ZipArchive` a little bit for it to work properly.

VihasMakwana:
@blakerouse @cmacknz @swiatekm This up for review now!

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: Don't apply for now as we removed the telemetry.

go.mod
```diff
@@ -386,7 +386,7 @@ require (
 	github.com/fearful-symmetry/gomsr v0.0.1 // indirect
 	github.com/fearful-symmetry/gorapl v0.0.4 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
-	github.com/foxboron/go-tpm-keyfiles v0.0.0-20250323135004-b31fac66206e // indirect
+	github.com/foxboron/go-tpm-keyfiles v0.0.0-20250903184740-5d135037bd4d // indirect
```
swiatekm: What are these dependency changes for?

testing/integration/ess/diagnostics_otel_test.go
```diff
@@ -0,0 +1,122 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
swiatekm: Why not put this test in diagnostics_test.go?

internal/pkg/otel/diagnostics.go
```diff
@@ -0,0 +1,52 @@
...
+	"io"
+	"net"
+	"net/http"
+
+	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
+	"github.com/elastic/elastic-agent/internal/pkg/otel/extension/elasticdiagnostics"
+	"github.com/elastic/elastic-agent/pkg/control/v2/client"
+)
+
+func PerformDiagnosticsExt(ctx context.Context, cpu bool) (*elasticdiagnostics.Response, error) {
```
swiatekm:
```suggestion
func PerformDiagnosticsExt(ctx context.Context, includeCPUProfile bool) (*elasticdiagnostics.Response, error) {
```

I like the parameter name to be more descriptive. Can you also add a docstring to this function?

internal/pkg/otel/manager/manager.go
```diff
@@ -361,9 +362,35 @@ func buildMergedConfig(cfgUpdate configUpdate, agentInfo info.Agent, monitoringC
...
+	}
+
 	return mergedOtelCfg, nil
 }
 
+func (m *OTelManager) injectDiagnosticsExtension(config *confmap.Conf) error {
+	if _, err := os.Stat(paths.DiagnosticsExtensionSocket()); !os.IsNotExist(err) {
+		// socket already exists. User is most likely using "development" namespace.
+		// generate a unique socket name by fetching current timestamp
+		paths.SetDiagnosticsExtensionSocket(fmt.Sprintf("%d-edot-diagnostics-extension.sock", time.Now().UnixMilli()))
```
swiatekm: I don't think this works the way you intend it to. If we generate a new path here, there's no way for the command to know about it, as it runs in its own process.

VihasMakwana: yeah, i see what you mean. If we're under development, we'll be under `/Library/Elastic/Agent-Development/elastic-agent`, which is different than default namespace.  
I think we'll be fine by removing this.

internal/pkg/otel/diagnostics.go
VihasMakwana: DOne

testing/integration/ess/diagnostics_otel_test.go
VihasMakwana: Yeah, that's better. I moved it to `diagnostics_test.go`.

go.mod
VihasMakwana: I think these were leftovers from cleanup. I'll update them.

VihasMakwana: Done

internal/pkg/otel/manager/manager.go
VihasMakwana: I've removed this section.

internal/pkg/otel/manager/manager.go
```diff
@@ -361,9 +362,30 @@ func buildMergedConfig(cfgUpdate configUpdate, agentInfo info.Agent, monitoringC
...
 	}
 
+	if err := m.injectDiagnosticsExtension(mergedOtelCfg); err != nil {
+		return nil, fmt.Errorf("failed to inject diagnostics: %w", err)
+	}
+
 	return mergedOtelCfg, nil
 }
 
+func (m *OTelManager) injectDiagnosticsExtension(config *confmap.Conf) error {
```
swiatekm: This doesn't need to be a manager method, as it doesn't use the manager at all. Nor does `buildMergedConfig` above need to be one. Imo it's better to use pure functions where feasible, as it's easier to understand what data they're actually operating on.

testing/integration/ess/diagnostics_test.go
```diff
@@ -437,6 +437,102 @@ agent.monitoring.enabled: false
...
+	require.NoError(t, err)
+	cmd, err := f.PrepareAgentCommand(ctx, []string{"-e"})
+	require.NoError(t, err)
+
+	output := strings.Builder{}
+	cmd.Stderr = &output
+	cmd.Stdout = &output
+
+	err = cmd.Start()
+	require.NoError(t, err)
```
swiatekm, line 514: It's better to call `fixture.Install` and then use the `logs` command to fetch the logs as in https://github.com/elastic/elastic-agent/pull/10153. Then you get all the nice perks of the fixture, including diagnostics if the test fails.

testing/integration/ess/diagnostics_test.go
```diff
@@ -437,6 +437,102 @@ agent.monitoring.enabled: false
...
+	extractionDir := t.TempDir()
+
+	stat, err := os.Stat(diagZip)
+	require.NoErrorf(t, err, "stat file %q failed", diagZip)
+	require.Greaterf(t, stat.Size(), int64(0), "file %s has incorrect size", diagZip)
+
+	extractZipArchive(t, diagZip, extractionDir)
+
+	expectedFiles := []string{
+		"edot/otel-merged-actual.yaml",
```
swiatekm, line 532: Are we not checking for component diagnostics here because they don't exist yet?

internal/pkg/otel/manager/manager.go
```diff
@@ -361,9 +362,30 @@ func buildMergedConfig(cfgUpdate configUpdate, agentInfo info.Agent, monitoringC
...
+	extensionCfg := map[string]any{
+		"extensions": map[string]any{
+			"elastic_diagnostics": map[string]any{
+				"endpoint": paths.DiagnosticsExtensionSocket(),
+			},
+		},
+	}
+	if config.IsSet("service::extensions") {
+		extensionList := config.Get("service::extensions").([]interface{})
+		extensionList = append(extensionList, "elastic_diagnostics")
```
swiatekm, line 387: It'd be safer to check if the extension is already on the list. We've had issues in the past related to extensions being added more than once.

swiatekm (dismissed):
LGTM overall, had some more comments but they aren't blocking.

testing/integration/ess/diagnostics_test.go
VihasMakwana: I can work on this as a follow-up. I'm manually preparing and running the fixture because for some reason, it starts with the port `6789` port and fails instantly because there's default elastic agent always running on my Mac.

I tried to make it like `TestBeatDiagnostics` but as I said, it fails instantly with `cannot bind to port`. I can get a linux VM and try it out later.

wdyt?

internal/pkg/otel/manager/manager.go
VihasMakwana: Done!

testing/integration/ess/diagnostics_test.go
VihasMakwana: I think I missed this. I've added it.

internal/pkg/otel/manager/manager.go
VihasMakwana: Done!

internal/pkg/diagnostics/diagnostics.go
blakerouse: Thanks for explaining that here. Makes sense.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
blakerouse: Thanks for adding.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -22,14 +22,16 @@ import (
...
 	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
 	"github.com/elastic/elastic-agent-libs/logp"
 	"github.com/elastic/elastic-agent/internal/pkg/diagnostics"
 	"github.com/elastic/elastic-agent/pkg/ipc"
 )
 
 var (
-	_ component.Component = (*diagnosticsExtension)(nil)
+	_ component.Component             = (*diagnosticsExtension)(nil)
+	_ otelmanager.DiagnosticExtension = (*diagnosticsExtension)(nil)
```
blakerouse, line 37: A comment over these lines to explain there present and why they should not be removed would be good. Future engineers might not realize there purpose and can easily remove it without it causing any issues.

blakerouse (approved):
This looks good and is ready to merge. I do have one request for an added comment.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: Done!

blakerouse (approved):
Thanks for adding the comment. Looks good.

ebeahan:
[v1.33.0:amd64:basic,slim,complete,service,elastic-otel-collector](https://buildkite.com/elastic/elastic-agent/builds/27812#01999ae1-d1dd-44a4-b469-ddd64f1b8dc1) is timing out in CI to obtain a BK agent, but we don't see this as a blocker. Force merging.

Issue #8047, opened 2025-04-30
[monitoring] /liveness?failon=degraded not returning HTTP 500


andrewkroh:
According to the text in the elastic-agent.yml, the liveness endpoint should return HTTP 500 when a component is in the DEGRADED state. But I am seeing it return HTTP 200.

https://github.com/elastic/elastic-agent/blob/f807fccca92cb143705c94e94e3f477ef83b794f/_meta/elastic-agent.yml#L126

```
$ curl -w 'HTTP %{http_code}' 'http://localhost:6791/liveness?failon=degraded'
HTTP 200
```

```
$ elastic-agent status
┌─ fleet
│  └─ status: (HEALTHY) Connected
└─ elastic-agent
   ├─ status: (DEGRADED) 1 or more components/units in a degraded state
   └─ cel-default
      ├─ status: (HEALTHY) Healthy: communicating with pid '26'
      └─ cel-default-cel-ti_abusech-7d5f8550-1b5b-11ee-8ec2-15184cec64a8
         └─ status: (DEGRADED) single event error object returned by evaluation: {"error":{"code":"402","id":"402 Payment Required","message":"POST:{\n    \"query_status\": \"ratelimited\",\n    \"msg\": \"Your request has been rate-limited. Please visit https:\\/\\/abuse.ch\\/rate-limit\\/ for more information.\"\n}"}}
```

```
elastic-agent version
Binary: 8.17.4 (build: 378c3e87685deb271aef538b708167fb458b51ba at 2025-03-20 12:08:33 +0000 UTC)
Daemon: 8.17.4 (build: 378c3e87685deb271aef538b708167fb458b51ba at 2025-03-20 12:08:33 +0000 UTC)
```

cmacknz:
We don't iterate over the units, only the components list when considering state.

https://github.com/elastic/elastic-agent/blob/44d9e12400ce3a8d94540c0c75d4c0e5c8a079f6/internal/pkg/agent/application/monitoring/liveness.go#L80-L85

cmacknz:
That'll miss every input or output specific error

rubenruizdegauna:
This is already [fixed](https://github.com/elastic/elastic-agent/pull/9955) on latest 8.x (8.19.4) 

```shell
elastic-agent status
┌─ fleet
│  └─ status: (STARTING) 
└─ elastic-agent
   ├─ status: (DEGRADED) 1 or more components/units in a degraded state
   ├─ mysql/metrics-default
[...]

PR #10060, opened 2025-09-19
fix: take into account units state on liveness

rubenruizdegauna:
## What does this PR do?

This PR adds the check of the component's unit's state on liveness endpoint. If a component state is healthy, but a unit state is degraded or failed, the liveness endpoint will return a 500.

## Checklist

## Disruptive User Impact
Liveness probes will now fail if a component state is healthy but any of the units is failed or degraded, likely causing the container to be restarted (see https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#liveness-probe).

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/8047

### DIFF
--- a/internal/pkg/agent/application/monitoring/liveness.go
+++ b/internal/pkg/agent/application/monitoring/liveness.go
@@ -84,9 +84,17 @@ func livenessHandler(coord CoordinatorState) func(http.ResponseWriter, *http.Req
 		}
 
 		unhealthyComponent := false
+	componentsLoop:
 		for _, comp := range state.Components {
 			if (failConfig.Failed && comp.State.State == client.UnitStateFailed) || (failConfig.Degraded && comp.State.State == client.UnitStateDegraded) {
 				unhealthyComponent = true
+				break componentsLoop
+			}
+			for _, unit := range comp.State.Units {
+				if (failConfig.Failed && unit.State == client.UnitStateFailed) || (failConfig.Degraded && unit.State == client.UnitStateDegraded) {
+					unhealthyComponent = true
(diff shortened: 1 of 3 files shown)


internal/pkg/agent/application/monitoring/liveness.go
```diff
@@ -84,9 +84,17 @@ func livenessHandler(coord CoordinatorState) func(http.ResponseWriter, *http.Req
...
+	componentsLoop:
 		for _, comp := range state.Components {
 			if (failConfig.Failed && comp.State.State == client.UnitStateFailed) || (failConfig.Degraded && comp.State.State == client.UnitStateDegraded) {
 				unhealthyComponent = true
+				break componentsLoop
+			}
+			for _, unit := range comp.State.Units {
+				if (failConfig.Failed && unit.State == client.UnitStateFailed) || (failConfig.Degraded && unit.State == client.UnitStateDegraded) {
+					unhealthyComponent = true
+					break componentsLoop
```
blakerouse: Instead of using a break to a label, this seems like it would be better to just be a small helper function that will return if the component is healthy, checking the overall component state first returning early and then checking the units.

rubenruizdegauna: Good point! I followed the same approach as with `otelhelpers`. Let me know if this is what you had in mind.

d7cf4b859ec32d6bae71ac1f6c2fd4031b3eab11

blakerouse (approved):
Thanks for cleaning up the loop. Looks good.

PR #10087, opened 2025-09-22
Fall back to process runtime if otel runtime is unsupported

### LINKED ISSUE #9098: [beats receivers] Conditionally enable Beats receivers only if configuration translation succeeds
As we roll out the architecture change to beats receivers there will be cases where configuration translation is only possible for a subset of all configurations. A current example would be Kerberos authentication in the Elasticsearch output being an unsupported translation.

In these cases we expect a standard error to be returned from the translation methods that allow agent to automatically fall back to the current runtime implementation.

The primary use case for this is eagerly migrating use case to beats receivers by defaulting the feature flag to true, in which we will want to fall back in some known cases with limitations.

As an optimization for user experience, when the feature flag defaults off but is set to true manually we should not silently fall back for either fail immediately or ensure that the affected components are marked as failed in the agent status report.

### LINKED ISSUE #9696: [beats receivers] Only run components with supported outputs as Beats receivers
Right now the only supported output type is Elasticsearch: https://github.com/elastic/elastic-agent/blob/889c6d3aa1db72d9aab970a0d7a79b80c956869b/internal/pkg/otel/translate/otelconfig.go#L45

Components are however split between the sub-process and otel runtime based solely on the feature flag state right now: https://github.com/elastic/elastic-agent/blob/889c6d3aa1db72d9aab970a0d7a79b80c956869b/internal/pkg/agent/application/coordinator/coordinator.go#L1683-L1687

When an unsupported output is used with Beats receivers, it is sent to the OTel configuration manager which will detect it can't correctly translate configurations with the unsupported output type and simply not run it. This will break users using non-Elasticsearch outputs (Kafka, Logstash) today when we default agent monitoring to use beats receivers by default.

https://github.com/elastic/elastic-agent/blob/889c6d3aa1db72d9aab970a0d7a79b80c956869b/internal/pkg/otel/manager/manager.go#L305-L312

https://github.com/elastic/elastic-agent/blob/889c6d3aa1db72d9aab970a0d7a79b80c956869b/internal/pkg/otel/translate/otelconfig.go#L87-L95

https://github.com/elastic/elastic-agent/blob/889c6d3aa1db72d9aab970a0d7a79b80c956869b/internal/pkg/otel/translate/otelconfig.go#L81-L86

## Acceptance Criteria

- A test exists proving that a configuration using either of the unsupported Logstash or Kafka outputs as the monitoring output when `agent.monitoring._runtime_experimental: otel` is set continues to run the monitoring beats as sub-processes.
  - A log line and the agent diagnostics clearly indicate the monitoring beats continue to be run as sub-processes even though it was requested to run them as receivers.

swiatekm:
## What does this PR do?

Makes inputs use the process runtime if they're configured to use the otel runtime, and the latter cannot support them. Currently there are two possible reasons for this:

* The output is not suppored at all - logstash is an example.
* The output configuration uses unsupported config options - `allow_older_versions: false` for the elasticsearch output for example.

A log line is also emitted if this happens.

I've also moved the component monitoring code into its own package to avoid dependency cycles.

## Why is it important?

We want to enable the otel runtime progressively without ever breaking a user's working setup. This means that a fallback is necessary, even if it may involve going against the explicit configuration. Later, we may add a flag which causes this to be an error instead.

## Checklist

Build the agent package and run it locally with either an elasticsearch output using an unsupported option (this PR uses `allow_older_versions: false`), or a kafka or logstash output. Then look at the status and logs. You should see a log line warning that your input was switched to the process runtime.

## Related issues

- Closes #9696 
- Closes #9098

## Questions to ask yourself

- ...


internal/pkg/otel/translate/otelconfig.go
```diff
@@ -92,18 +92,35 @@ func GetOtelConfig(
...
+	return nil
 }
 
 // getSupportedComponents returns components from the given model that can be run in an Otel Collector.
 func getSupportedComponents(model *component.Model) []*component.Component {
 	var supportedComponents []*component.Component
 
 	for _, comp := range model.Components {
-		if IsComponentOtelSupported(&comp) {
+		if VerifyComponentIsOtelSupported(&comp) == nil {
```
cmacknz: We should be logging that we are ignoring the request to run a component with the otel runtime because of an unsupported configuration here instead of silently falling back.

I think we need to report what is using beats receivers vs what is requested back to fleet for telemetry purposes as well. I will create a separate issue for that.

internal/pkg/otel/translate/otelconfig_test.go
```diff
@@ -1040,3 +1040,118 @@ func TestGetReceiversConfigForComponent(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		{
+			name: "unsupported output type",
+			component: &component.Component{
+				ID:         "unsupported-output",
+				InputType:  "filestream",
+				OutputType: "kafka", // unsupported
```
cmacknz, line 1092: Let's cover logstash as well just for completeness.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -793,6 +793,98 @@ agent.monitoring.enabled: false
...
+    _runtime_experimental: otel
+    streams:
+      - metricsets:
+        - cpu
+outputs:
+  default:
+    type: elasticsearch
+    hosts: [http://localhost:9200]
+    api_key: placeholder
+    allow_older_versions: false # not supported by the elasticsearch exporter
```
cmacknz: Pick something other than `allow_older_versions`, which we are likely to fix in the near future, just to avoid churn on the tests.

Something like `indices` which we are very unlikely to support would be better in these tests.

internal/pkg/otel/translate/otelconfig.go
cmacknz: I can see this function gets called again separately where this happens which is a bit confusing.

internal/pkg/otel/manager/diagnostics.go
```diff
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/monitoring"
+	component2 "github.com/elastic/elastic-agent/internal/pkg/agent/application/monitoring/component"
```
cmacknz:
```suggestion
	componentmonitoring "github.com/elastic/elastic-agent/internal/pkg/agent/application/monitoring/component"
```

`component2` doesn't really communicate anything about what the package does but I understand naming things is hard

cmacknz: I see `componentmonitoring` in the other file let's use that

internal/pkg/otel/translate/otelconfig.go
swiatekm: I can log an error here. This is a defensive check that shouldn't really be triggered anymore with the other changes in this PR, unless we have a bug.

testing/integration/ess/beat_receivers_test.go
swiatekm: Done, used `indices` everywhere.

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -92,19 +92,38 @@ func GetOtelConfig(
...
+	if errors.Is(compErr, errors.ErrUnsupported) {
+		return fmt.Errorf("unsupported configuration for %s: %w", comp.ID, compErr)
+	}
+
+	return nil
 }
 
 // getSupportedComponents returns components from the given model that can be run in an Otel Collector.
-func getSupportedComponents(model *component.Model) []*component.Component {
+func getSupportedComponents(model *component.Model, logger *logp.Logger) []*component.Component {
```
blakerouse: Very much a nit, but most function signatures take the logger first. I don't think this is applied across the code-base as much as I would like, but think its would provide more consistency if we strive for that.

```suggestion
func getSupportedComponents(logger *logp.Logger, model *component.Model) []*component.Component {
```

blakerouse (dismissed):
This change looks good and well tested. I added 1 nit, but it is very much a nit. Feel free to ignore, if you like. ;-)

cmacknz:
LGTM too, thanks for addressing my comments.

cmacknz (approved):

PR #9344, opened 2025-08-13
[otel] add receivers: apache, iis, mysql, postgresql v0.130.0

mykola-elastic:
## What does this PR do?

Receivers added:
- mysqlreceiver
- apachereceiver
- iisreceiver
- postgresqlreceiver

## Why is it important?

Needed in order to add OTEL content packs to elastic/integrations which use the data from the receivers listed

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [ ] ~I have commented my code, particularly in hard-to-understand areas~
- [x] I have made corresponding changes to the documentation
- [ ] ~I have made corresponding change to the default configuration files~
- [ ] ~I have added tests that prove my fix is effective or that my feature works~
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [x] I have added an integration test or an E2E test

## Related issues

- Closes https://github.com/elastic/obs-integration-team/issues/439

PR #10103, opened 2025-09-23
Update code ownership of core-components.yml

theletterf:
With increased components' movement, the role of `core-components.yml` is becoming critical. The ingest docs team is not qualified to decide on the content of that file.

This PR suggests adding the PM and EM for these file decisions as exclusive owners of the file. That also means they would be gating components updates and changes in each Collector release.

Case in point: the status of some new receivers in https://github.com/elastic/elastic-agent/pull/9344

+CC @mlunadia @AlexanderWert @colleenmcginnis


theletterf:
Can be merged after https://github.com/elastic/catalog-info/pull/2463 is merged.

AlexanderWert (approved):

Issue #10080, opened 2025-09-22
User is unable to migrate agent due to "ErrInvalidToken" error.


amolnater-qasource:
**Kibana version:**
VERSION: 9.2.0-SNAPSHOT
BUILD: 90821
COMMIT: 579ec67503caf5cd5d146d92c12543ada6ceba84
Artifact Link: https://snapshots.elastic.co/9.2.0-6c62938e/downloads/beats/elastic-agent/elastic-agent-9.2.0-SNAPSHOT-linux-x86_64.tar.gz

**Preconditions:**
1. An agent should be installed.
2. There are clusters available for migration testing.

**Steps to reproduce:**
1. Navigate to the Agents tab in Kibana.
2. Select an agent for migration.
3. Attempt to trigger the "Migrate Agent" action by providing the Fleet Server URL and enrollment token for an agent policy.
4. Observe that the migration process fails with an "ErrInvalidToken" error.

**Expected behavior:**
Agent migration should succeed if a valid enrollment token and Fleet Server URL are provided.

**Agent Logs:**

[elastic-agent-diagnostics-2025-09-21T17-08-00Z-00.zip](https://github.com/user-attachments/files/22458697/elastic-agent-diagnostics-2025-09-21T17-08-00Z-00.zip)

**Agent JSON:**
[ip-172-31-30-101-agent-details (1).json](https://github.com/user-attachments/files/22458700/ip-172-31-30-101-agent-details.1.json)

**Screen Capture:**

<img width="1512" height="798" alt="Image" src="https://github.com/user-attachments/assets/e8a52029-95ac-40c7-9156-7f72dada9861" />

**Related Feature:**
https://github.com/elastic/kibana/issues/217619

**Note:**
We have cross checked that we are using correct enrollment token and Fleet server URL.

amolnater-qasource:
@muskangulati-qasource Please review.

muskangulati-qasource:
Secondary review is Done for this ticket!

michalpristas:
if you use values using `elastic-agent enroll` command does it work?

amolnater-qasource:
Hi @michalpristas 

Thank you for looking into this issue.
Yes, we are successfully able to install the Agent on new Kibana with same values.

**Screenshot:**
<img width="857" height="244" alt="Image" src="https://github.com/user-attachments/assets/f78ae62f-4241-4140-8f01-d725ec5f0aad" />

The same values fail if we use it for Migration action.

[...]

PR #10115, opened 2025-09-23
keep enrollment token when replacing data with signed

michalpristas:
Need to replace enrollment token with the one from data itself, signed data contains secret reference that is replaced by fleet server on action push and signed is not recomputed

Fixes: https://github.com/elastic/elastic-agent/issues/10080


internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
```diff
@@ -77,12 +77,17 @@ func (h *Migrate) Handle(ctx context.Context, a fleetapi.Action, ack acker.Acker
 		return err
 	}
 
+	// signed data contains secret reference
```
ycombinator: This comment isn't super clear (to me).  Specifically, what is the secret reference here?  It sounds like it has something to do with the enrollment token which is why you need to overwrite it with the value of `action.Data.EnrollmentToken` after `action.Data` is unmarshalled from `signedData`.  Could you elaborate a bit more in the inline comment please?

ycombinator (dismissed):
Built Agent from this PR and was able to successfully migrate it to another cluster.  Just left some minor feedback about clarifying a comment in the code but otherwise LGTM.

internal/pkg/agent/application/actions/handlers/handler_action_migrate.go
michalpristas: will update it but i did not want to include too much details

ycombinator (approved):

PR #10083, opened 2025-09-22
Add elasticapmprocessor to core components metadata

theletterf:
This adds `elasticapmprocessor` to the Core components metadata file used to generate the docs.

The change will happen in 9.2

PR #10116, opened 2025-09-23
Replace elastic trace processor

gregkalapos:
In 9.2 we want to replace the `elastictrace` processor with the `elasticapm` processor in EDOT. This PR prepares these changes.

## Why is it important?

The `elastictrace` processor only enriched traces, the `elasticapm` processor does the same (it's the same code) plus it also enriches logs and metrics. 

## Checklist

- [x] I have made corresponding changes to the documentation - ~docs PR here: https://github.com/elastic/elastic-agent/pull/10083~ sorry, that's not the doc, I know @theletterf has this on his radar. 


internal/pkg/otel/README.md
```diff
@@ -75,8 +75,8 @@ This section provides a summary of components included in the Elastic Distributi
 | [attributesprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/processor/attributesprocessor/v0.135.0/processor/attributesprocessor/README.md) | v0.135.0 |
 | [batchprocessor](https://github.com/open-telemetry/opentelemetry-collector/blob/processor/batchprocessor/v0.135.0/processor/batchprocessor/README.md) | v0.135.0 |
 | [cumulativetodeltaprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/processor/cumulativetodeltaprocessor/v0.135.0/processor/cumulativetodeltaprocessor/README.md) | v0.135.0 |
+| [elasticapmprocessor](https://github.com/elastic/opentelemetry-collector-components/blob/processor/elasticapmprocessor/v0.2.0/processor/elasticapmprocessor/README.md) | v0.2.0 |
 | [elasticinframetricsprocessor](https://github.com/elastic/opentelemetry-collector-components/blob/processor/elasticinframetricsprocessor/v0.17.0/processor/elasticinframetricsprocessor/README.md) | v0.17.0 |
-| [elastictraceprocessor](https://github.com/elastic/opentelemetry-collector-components/blob/processor/elastictraceprocessor/v0.10.0/processor/elastictraceprocessor/README.md) | v0.10.0 |
```
ChrsMark, line 84: I wonder if `elastictraceprocessor` should get deprecated and remain in EDOT for a couple of releases before getting removed completely.

gregkalapos: My only concern with this is that adding both at the same time in a pipeline isn't tested and it may cause issues. In that case trace enrichment would happen twice - which may work out well, other than performance hit, but we haven't tested it.

Of course for that to happen, both need to be added to a pipeline, so it won't happen out-of-the-box.

So for a user, it should be either one or the other processor, but never use both. If a specific version of EDOT has both, we'd run this risk of maybe having both in a pipeline.

theletterf (dismissed):

internal/pkg/otel/README.md
swiatekm: I'd be inclined to leave both in for one release, but clearly mark it the elastic trace processor as deprecated.  In an ideal world, this processor would've had its stability changed to deprecated in the final release, but as the code was removed already from opentelemetry-collector-components, this is the next best thing.

gregkalapos: What's the best way to mark it as deprecated? It seems CI is not happy to add something to this readme file: https://buildkite.com/elastic/elastic-agent/builds/27508#019980ac-2efd-43f9-b1d1-9a6440c1f4c9/162-609

swiatekm: @theletterf can you think of the right place to put that kind of notice? It should exist somewhere in the public docs rather than a hidden README in this repository, imo.

theletterf: This should at least go in the release notes. Has the relevant fragment been added? I haven't checked.

swiatekm: Yeah, it's already in the changelog fragment. I was wondering if it should appear anywhere else.

theletterf: Potentially in the generated components table or a callout. I'll add it next week.

theletterf: @gregkalapos @swiatekm Since we currently don't track deprecated components using metadata, let's communicate this in the release notes only. We can figure out a mechanism later on as we deprecate more components. +CC @AlexanderWert

gregkalapos: Thanks @theletterf - I also think calling this out in the release notes is a good idea - closing this thread.

gregkalapos:
Could someone with write access please take a look at this before FF? 

I see these fails in CI:

> The job, [Win2025:sudo:fleet-endpoint-security](https://buildkite.com/elastic/elastic-agent/builds/27756#01999729-581e-407f-bf70-b374667ce44d), has been canceled as it failed to get an agent after 5 tries.

and some other similar ones.

I don't see any corresponding logs for those - it's a bit hard to figure out what's wrong there.

gregkalapos:
This still fails, but I see unrelated logs, e.g.:

```
<head></head>
=== Failed
--
  | === FAIL: testing/integration/ess TestECH (0.00s)
  | panic: failed to create elasticsearch client: cannot create client: cannot parse url: parse "\x1b[33m╷\x1b[0m\x1b[0m\n\x1b[33m│\x1b[0m \x1b[0m\x1b[1m\x1b[33mWarning: \x1b[0m\x1b[0m\x1b[1mNo outputs found\x1b[0m\n\x1b[33m│\x1b[0m \x1b[0m\n\x1b[33m│\x1b[0m \x1b[0m\x1b[0mThe state file either has no outputs defined, or all the defined outputs are empty. Please define an output in your configuration with the `output` keyword\n\x1b[33m│\x1b[0m \x1b[0mand run `terraform refresh` for it to become available. If you are using interpolation, please verify the interpolated value is not empty. You can use the\n\x1b[33m│\x1b[0m \x1b[0m`terraform console` command to assist.\n\x1b[33m╵\x1b[0m\x1b[0m": net/url: invalid control character in URL [recovered]
```

Could someone please help out? We'd also need an approve (assuming the PR is ok, of course) from someone with write access.

swiatekm (approved):

gregkalapos:
Thanks @swiatekm!

PR #10118, opened 2025-09-23
Fix multiarch service image push from main to serverless

blakerouse:
## What does this PR do?

Fixes the building and publishing of the service container to serverless from main.

## Why is it important?

Without this change the image is only arm64 and not amd64.


blakerouse:
Closing in favor of https://github.com/elastic/elastic-agent/pull/10129

PR #5129, opened 2024-07-12
[CI] Separated packaging step

pazone:
## What does this PR do?
 Separated packaging step

## Why is it important?
Now we don't need to re-package agent when restarting tests from BK UI

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/4142

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #10129, opened 2025-09-24
Fix multiarch service image push from main to serverless

blakerouse:
## What does this PR do?

Fixes the building and publishing of the service container to serverless from main.

## Why is it important?

Without this change the image is only arm64 and not amd64.


.buildkite/pipeline.agentless-app-release.yaml
```diff
@@ -1,13 +1,56 @@
...
-    key: "mirror-elastic-agent"
-    command: ".buildkite/scripts/steps/ecp-internal-release.sh"
+  - label: "Packaging: Service Container linux/amd64"
+    key: packaging-service-container-amd64
+    env:
+      PACKAGES: "docker"
+      PLATFORMS: "linux/amd64"
+      DOCKER_VARIANTS: "service"
+    command: |
+      .buildkite/scripts/steps/integration-package.sh
```
v1v, line 19: I'm not sure if the name of the script is correct here, IIUC, that script is used by the integrations pipeline, hence the reason for that name.

It was introduced at https://github.com/elastic/elastic-agent/pull/5129/files

I wonder if the name should be different. I'm not very familiar with the specifics of agentless, but it might cause some issues if it's changed for other reasons and there are multiple consumers.

.buildkite/pipeline.agentless-app-release.yaml
```diff
@@ -1,13 +1,56 @@
...
+      diskSizeGb: 400
 
+  - label: "Packaging: Service Container linux/arm64"
+    key: packaging-service-container-arm64
+    env:
+      PACKAGES: "docker"
+      PLATFORMS: "linux/arm64"
+      DOCKER_VARIANTS: "service"
+    command: |
+       .buildkite/scripts/steps/integration-package.sh
```
v1v, line 34: ditto

.buildkite/scripts/steps/ecp-internal-release.sh
```diff
@@ -15,12 +15,11 @@
 # this material is strictly forbidden unless prior written
 # permission is obtained from Elasticsearch B.V.
 
-set -eu
+set -eux
```
v1v: let's avoid `-x` and print verbose output in the logs by default

.buildkite/scripts/steps/ecp-internal-release.sh
```diff
@@ -15,12 +15,11 @@
 # this material is strictly forbidden unless prior written
```
v1v, line 15: wonder if this script should be owned by a specific team in the CODEOWNERS, so any changes need their approval and we, robots, don't need to be in the review process.

.buildkite/scripts/steps/ecp-internal-release.sh
```diff
@@ -33,37 +32,39 @@ write_annotation() {
...
 
-DOCKER_TAG="git-${GIT_SHORT_COMMIT}"
-PRIVATE_IMAGE="${PRIVATE_REPO}:${DOCKER_TAG}"
+# ARM64 (overwrites AMD64 tags)
+docker load -i ./build/distributions/elastic-agent-service-$DOCKER_TAG-$BUILD_VERSION-linux-arm64.docker.tar.gz
+docker image tag "elastic-agent-service:$DOCKER_TAG" "$PRIVATE_IMAGE"
+docker push "$PRIVATE_IMAGE"
+ARM64_DIGEST=$(docker image inspect --format "{{index .RepoDigests 0}}" "$PRIVATE_IMAGE")
 
-# TODO: let's avoid accessing vault directly but use the vault plugin itself
```
v1v, line 51: Thanks! 🙏

nkvoll (dismissed):
Separate from @v1v's comments, the resulting image from https://buildkite.com/elastic/agentless-serverless-release/builds/127/steps/canvas?sid=01997d89-cb01-4e1b-989b-b085a8458de9 / `ecp-internal-release.sh` seems correct for me, expected size, containing both architectures, starts up elastic-agent successfully for me on both architectures. 👍

.buildkite/pipeline.agentless-app-release.yaml
blakerouse: @v1v I use that script because we want the image built the same way that the integration steps verified the image. I can duplicate it and name it differently, but re-using that script was kinda a point of saying we CI validate the same build, so we will use the same build.

I can add a comment in the file to ensure that its clear that its used in both, would that work? Or do you think duplicating the file with a different name is better?

.buildkite/scripts/steps/ecp-internal-release.sh
blakerouse: I can update CODEOWNERS to only be the control plane team.

.buildkite/pipeline.agentless-app-release.yaml
v1v: No strong opinion, a comment sounds a good apporach here

blakerouse:
@v1v I have updated the areas requested.

@nkvoll If you could give it another review as the change has dismissed your previous review.

.buildkite/pipeline.agentless-app-release.yaml
blakerouse: UPdated.

.buildkite/pipeline.agentless-app-release.yaml
blakerouse: Updated

nkvoll (approved):

.buildkite/scripts/steps/integration-package.sh
```diff
@@ -3,6 +3,12 @@ set -euo pipefail
 
 source .buildkite/scripts/common.sh
 
+# This script is not only used by integration CI, but also the agentless release pipeline to build the docker images
+# that get released to serverless. USE_PACKAGE_VERSION is important here for both CI and agentless, as it validates
+# that the components bundled to serverless are the same components that the CI validated.
+#
+# Making a change here can affect the released images to agentless, so be cautious.
```
v1v, line 10: `nit`: add the docs to the header instead

```suggestion
# This script is not only used by integration CI, but also the agentless release pipeline to build the docker images
# that get released to serverless. USE_PACKAGE_VERSION is important here for both CI and agentless, as it validates
# that the components bundled to serverless are the same components that the CI validated.
#
# Making a change here can affect the released images to agentless, so be cautious.

source .buildkite/scripts/common.sh
```

.buildkite/pipeline.agentless-app-release.yaml
```diff
@@ -1,13 +1,56 @@
...
+    artifact_paths:
+      - "build/distributions/elastic-agent-service-git-*.docker.tar.gz"
```
v1v, line 36: shall we use a more generic expression? It uses packages and platforms:


```suggestion
      - "build/distributions/elastic-agent-service-*"
```

.buildkite/pipeline.agentless-app-release.yaml
```diff
@@ -1,13 +1,56 @@
...
+  - label: "Packaging: Service Container linux/amd64"
+    key: packaging-service-container-amd64
+    env:
+      PACKAGES: "docker"
+      PLATFORMS: "linux/amd64"
+      DOCKER_VARIANTS: "service"
+    command: |
+      .buildkite/scripts/steps/integration-package.sh
+    artifact_paths:
+      - "build/distributions/elastic-agent-service-git-*.docker.tar.gz"
```
v1v, line 21:
```suggestion
      - "build/distributions/elastic-agent-service-*"
```

.buildkite/pipeline.agentless-app-release.yaml
blakerouse: This specifically only wants the docker artifacts, no need for anything else.

.buildkite/pipeline.agentless-app-release.yaml
blakerouse: Same as above.

.buildkite/pipeline.agentless-app-release.yaml
v1v: I mean that the PLATFORMS and PACKAGES env variables already limit the generated artifacts to docker artifacts for the specific architecture, hence `docker.tar.gz` is not needed. It was just a very `nit` :)

PR #10131, opened 2025-09-24
Update journald tests now that Filebeat supports watching folders

belimawr:
## What does this PR do?

It updates the Journald tests so they're use a folder instead of a glob pattern, which fixes some errors that started appearing after https://github.com/elastic/beats/pull/46682 was merged.

It also updates the Beats dependency version so it includes the fix mentioned above.

## Why is it important?

CI was constantly failing, some examples:
 - https://github.com/elastic/elastic-agent/pull/10092
 - https://github.com/elastic/elastic-agent/pull/10122
 - https://github.com/elastic/elastic-agent/pull/10094

## Checklist

~~## Disruptive User Impact~~

```
DEV=true SNAPSHOT=true EXTERNAL=true PACKAGES="docker" DOCKER_VARIANTS="basic,complete,elastic-otel-collector" PLATFORMS=linux/amd64 mage -v package
SNAPSHOT=true INSTANCE_PROVISIONER=kind mage -v integration:testKubernetesSingle TestKubernetesJournaldInputOtel
SNAPSHOT=true INSTANCE_PROVISIONER=kind mage -v integration:testKubernetesSingle TestKubernetesJournaldInput
```

~~## Related issues~~

## Questions to ask yourself

- ...


testing/integration/k8s/testdata/journald-input.yml
```diff
@@ -19,4 +19,5 @@ inputs:
     streams:
       - id: journald-input-id
         paths:
-          - "/opt/journald/*/*"
+          - "/opt/journald/"
```
cmacknz, line 22: @belimawr doesn't this mean the change in https://github.com/elastic/beats/pull/46682 was a breaking change to configuration?

Wouldn't external users have to do the same thing?

belimawr: It's actually the opposite :disappointed:. We broke it when migrating to Journald, specifying a folder directly should have always worked.

Our [docs](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-journald#filebeat-input-journald-paths) always stated:
> A list of paths that will be crawled and fetched. Each path can be a directory path (to collect events from all journals in a directory), or a file path. If you specify a directory, Filebeat merges all journals under the directory into a single journal and reads them.
>
> If no paths are specified, Filebeat reads from the default journal.

It's on my mental list to update the release notes/docs today to be clear about this change.

belimawr: Another option is to try detecting glob patterns and passing them as paths and only simple folders as folders. That actually might be a better solution.

belimawr: A PR restoring the previous behaviour wile keeping the feature/fix working: https://github.com/elastic/beats/pull/46792

Regardless of the change in Filebeat, the correct way to ingest a journal via path is by using `"/opt/journald/"`. So the tests here must be updated.

pkoutsovasilis:
@belimawr should we backport this PR to resolve the issues to all active branches?

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/docker/docker v28.3.3+incompatible
 	github.com/docker/go-units v0.5.0
 	github.com/dolmen-go/contextio v0.0.0-20200217195037-68fc5150bcd5
-	github.com/elastic/beats/v7 v7.0.0-alpha2.0.20250918130837-16c4d9a1aa8c
+	github.com/elastic/beats/v7 v7.0.0-alpha2.0.20250924190604-3dc72030429f
```
swiatekm, line 17: Can you avoid making this change in this PR? This will make it impossible to backport cleanly, as the beats dependency needs to mirror its respective branch in each agent branch.

belimawr: I could, It's a :chicken: :egg: problem.

TL;DR: a change (fix) in Filebeat broke the workaround we are using in the Journald input configuration here, I updated the Beats dependency to they're in sync with the Agentbeat version shipped (both having the same fix). Even though the change in Beats is a fix, it breaks the buggy behaviour, so https://github.com/elastic/beats/pull/46792 will keep the buggy behaviour while still supporting the correct behaviour for ingesting Journal folders.

If we wait for https://github.com/elastic/beats/pull/46792 to be merged and released as snapshot build, this whole PR becomes irrelevant and can be closed without merging.

Today is Friday, we can merge https://github.com/elastic/beats/pull/46792 today (and hopefully all the backports), if we miss the snapshot generation, it will be done by Monday, and then we can re-create all the "Update Versions" PR that are kinda blocked right now.

belimawr: At some point, sooner rather than later, we should have the journald configuration updated from
```yaml
        paths:
          - "/opt/journald/*/*"
```
to
```yaml
        paths:
          - "/opt/journald/"
```

For the sake of correctness in our tests. This is also testing a common use case of ingesting Kubernetes node logs via Journald.

belimawr: Edit, I fixed the paths in my previous comment.

swiatekm: Allright, I'm fine merging it as-is. But we'll have to be careful with backports. In fact, you should set `backport-active-all` on this PR and then resolve dependencies in each backport, imo.

belimawr: That makes sense.

Issue #2530, opened 2023-04-20
Heap profiles in diagnostics are unusable as of 8.6


cmacknz:
The heap profiles captured in the agent diagnostics starting in 8.6 are not usable, they contain no meaningful information. Prior to 8.6 these profiles used to contain the complete allocation graph. For example here is a heap profile captured from fleet-server:

![image](https://user-images.githubusercontent.com/3466215/233443636-5ae98fb1-68bf-4017-8141-dba9f0bf5135.png)

In 8.6 there was a change to how diagnostics were collected for Go programs. Diagnostics are now collected using diagnostic hooks in the Elastic Agent client package:

https://github.com/cmacknz/elastic-agent-client/blob/141cf7b58d33c0ad8b3fa427bbd3af9c32773437/pkg/client/client_v2.go#L783-L802

```go
func (c *clientV2) registerPprofDiagnostics(name string, description string) {
	c.RegisterDiagnosticHook(name, description, fmt.Sprintf("%s.txt", name), "plain/text", func() []byte {
		var w bytes.Buffer
		err := pprof.Lookup(name).WriteTo(&w, 1)
		if err != nil {
			// error is returned as the content
			return []byte(fmt.Sprintf("failed to write pprof to bytes buffer: %s", err))
		}
		return w.Bytes()
	})
}
```

This appears to have unintentionally broken heap diagnostics collection. Simpler diagnostics like the goroutine profile do appear to work correctly with this implementation.

cmacknz:
In 8.5 and earlier we relied on being able to reach the /debug/pprof endpoint to collect profiles instead:

https://github.com/elastic/elastic-agent/blob/0e1a7396f356f8af5ad28c3c8e941256f50c86a2/internal/pkg/agent/control/server/server.go#L500-L518

andrewkroh:
While trying to troubleshoot a suspected leak in 8.7.0 I just realized that the collected profiles are in the "legacy text format" instead of the gzip compressed protobuf format. In addition to fixing this problem, can I also request that we enable the `/debug/pprof` endpoints on the monitoring socket. For example, this is how Filebeat is run:

[...]

PR #10141, opened 2025-09-25
[otel] - Add Diagnostics Extension 

VihasMakwana:
## What does this PR do?

Split from https://github.com/elastic/elastic-agent/pull/10052

This PR add a new elasticdiagnsotics extension. This new extension will be used in hybrid mode and will perform diagnostics for running otel collector. It will:

- Acts as a registrar and keeps track of common diagnostic hooks.
- Collects profiles using runtime/pprof.
- Collects internal telemetry exposed by the OTeL Collector.
- Implements the extensioncapabilities.ConfigWatcher interface and stores the latest configuration of the running collector.
- Listens for diagnostic requests and provides diagnostic data.

## Checklist

See https://github.com/elastic/elastic-agent/pull/10052

## Related issues
- Relates https://github.com/elastic/beats/issues/44604

### DIFF
--- a/internal/pkg/otel/config/config.go
+++ b/internal/pkg/otel/config/config.go
@@ -10,7 +10,7 @@ import (
 	"github.com/elastic/elastic-agent/internal/pkg/otel/manager"
 )
 
-const defaultExecMode = manager.EmbeddedExecutionMode
+const defaultExecMode = manager.SubprocessExecutionMode
 
 type execModeConfig struct {
 	Agent struct {
--- a/internal/pkg/otel/manager/manager_test.go
+++ b/internal/pkg/otel/manager/manager_test.go
@@ -746,41 +746,74 @@ func TestOTelManager_Run(t *testing.T) {
 }
 
 func TestOTelManager_Logging(t *testing.T) {
+	wd, erWd := os.Getwd()
+	require.NoError(t, erWd, "cannot get working directory")
+
+	testBinary := filepath.Join(wd, "testing", "testing")
+	require.FileExists(t, testBinary, "testing binary not found")
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	const waitTimeForStop = 30 * time.Second
 
 	base, obs := loggertest.New("otel")
 	l, _ := loggertest.New("otel-manager")
-	m, err := NewOTelManager(l, logp.DebugLevel, base, EmbeddedExecutionMode, nil, nil, waitTimeForStop)
-	require.NoError(t, err, "could not create otel manager")
 
-	go func() {
-		err := m.Run(ctx)
-		assert.ErrorIs(t, err, context.Canceled, "otel manager should be cancelled")
-	}()
+	for _, tc := range []struct {
+		name       string
+		execModeFn func(collectorRunErr chan error) (collectorExecution, error)
+	}{
+		{
+			name: "in-process execution",
+			execModeFn: func(collectorRunErr chan error) (collectorExecution, error) {
+				return newExecutionEmbedded(), nil
+			},
+		},
+		{
+			name: "subprocess execution",
+			execModeFn: func(collectorRunErr chan error) (collectorExecution, error) {
+				return newSubprocessExecution(logp.DebugLevel, testBinary)
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			// the execution mode passed here is overridden below so it is irrelevant
+			m, err := NewOTelManager(l, logp.DebugLevel, base, EmbeddedExecutionMode, nil, nil, waitTimeForStop)
+			require.NoError(t, err, "could not create otel manager")
 
-	// watch is synchronous, so we need to read from it to avoid blocking the manager
-	go func() {
-		for {
-			select {
-			case <-m.WatchCollector():
-			case <-ctx.Done():
-				return
-			}
-		}
-	}()
+			executionMode, err := tc.execModeFn(m.collectorRunErr)
+			require.NoError(t, err, "failed to create execution mode")
+			testExecutionMode := &testExecution{exec: executionMode}
+			m.execution = testExecutionMode
 
-	cfg := confmap.NewFromStringMap(testConfig)
-	m.Update(cfg, nil)
+			go func() {
+				err := m.Run(ctx)
+				assert.ErrorIs(t, err, context.Canceled, "otel manager should be cancelled")
+			}()
 
-	// the collector should log to the base logger
-	assert.EventuallyWithT(t, func(collect *assert.CollectT) {
-		logs := obs.All()
-		require.NotEmpty(collect, logs, "Logs should not be empty")
-		firstMessage := logs[0].Message
-		assert.Equal(collect, "Internal metrics telemetry disabled", firstMessage)
-	}, time.Second*10, time.Second)
+			// watch is synchronous, so we need to read from it to avoid blocking the manager
+			go func() {
+				for {
+					select {
+					case <-m.WatchCollector():
+					case <-ctx.Done():
+						return
+					}
+				}
+			}()
+
+			cfg := confmap.NewFromStringMap(testConfig)
+			m.Update(cfg, nil)
+
+			// the collector should log to the base logger
+			assert.EventuallyWithT(t, func(collect *assert.CollectT) {
+				logs := obs.All()
+				require.NotEmpty(collect, logs, "Logs should not be empty")
+				firstMessage := logs[0].Message
+				assert.Equal(collect, "Internal metrics telemetry disabled", firstMessage)
+			}, time.Second*10, time.Second)
+		})
+	}
 }
 
 // statusToYaml converts the status.AggregateStatus to a YAML string representation.
--- a/pkg/component/platforms.go
+++ b/pkg/component/platforms.go
@@ -81,6 +81,11 @@ var GlobalPlatforms = Platforms{
 		Arch: AMD64,
 		GOOS: Windows,
 	},
+	{
+		OS:   Windows,
+		Arch: ARM64,
+		GOOS: Windows,
+	},
 }
 
 // String returns the platform string identifier.
--- a/specs/testbeat.spec.yml
+++ b/specs/testbeat.spec.yml
@@ -11,6 +11,7 @@ inputs:
       - windows/amd64
       - container/amd64
       - container/arm64
+      - windows/arm64
     outputs: &outputs
       - elasticsearch
       - kafka
--- a/testing/integration/ess/metrics_monitoring_test.go
+++ b/testing/integration/ess/metrics_monitoring_test.go
@@ -7,9 +7,12 @@
 package ess
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/http/httputil"
 	"testing"
 	"time"
 
@@ -19,6 +22,7 @@ import (
 
 	"github.com/elastic/elastic-agent-libs/kibana"
 	"github.com/elastic/elastic-agent-libs/testing/estools"
+	otelMonitoring "github.com/elastic/elastic-agent/internal/pkg/otel/monitoring"
 	atesting "github.com/elastic/elastic-agent/pkg/testing"
 	"github.com/elastic/elastic-agent/pkg/testing/define"
 	"github.com/elastic/elastic-agent/pkg/testing/tools"
@@ -29,8 +33,9 @@ type MetricsRunner struct {
 	suite.Suite
 	info         *define.Info
 	agentFixture *atesting.Fixture
-
-	ESHost string
+	policyID     string
+	policyName   string
+	ESHost       string
 }
 
 func TestMetricsMonitoringCorrectBinaries(t *testing.T) {
@@ -76,10 +81,54 @@ func (runner *MetricsRunner) SetupSuite() {
 	policyResp, _, err := tools.InstallAgentWithPolicy(ctx, runner.T(), installOpts, runner.agentFixture, runner.info.KibanaClient, basePolicy)
 	require.NoError(runner.T(), err)
 
+	runner.policyName = policyResp.Name
+	runner.policyID = policyResp.ID
+
 	_, err = tools.InstallPackageFromDefaultFile(ctx, runner.info.KibanaClient, "system",
 		integration.PreinstalledPackages["system"], "testdata/system_integration_setup.json", uuid.Must(uuid.NewV4()).String(), policyResp.ID)
 	require.NoError(runner.T(), err)
 
+}
+
+func (runner *MetricsRunner) addMonitoringToOtelRuntimeOverwrite() {
+	addMonitoringOverwriteBody := fmt.Sprintf(`
+{
+  "name": "%s",
+  "namespace": "default",
+  "overrides": {
+    "agent": {
+      "monitoring": {
+        "_runtime_experimental": "otel"
+      }
+    }
+  }
+}
+`, runner.policyName)
+	resp, err := runner.info.KibanaClient.Send(
+		http.MethodPut,
+		fmt.Sprintf("/api/fleet/agent_policies/%s", runner.policyID),
+		nil,
+		nil,
+		bytes.NewBufferString(addMonitoringOverwriteBody),
+	)
+	if err != nil {
+		runner.T().Fatalf("could not execute request to Kibana/Fleet: %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		// On error dump the whole request response so we can easily spot
+		// what went wrong.
+		runner.T().Errorf("received a non 200-OK when adding overwrite to policy. "+
+			"Status code: %d", resp.StatusCode)
+		respDump, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			runner.T().Fatalf("could not dump error response from Kibana: %s", err)
+		}
+		// Make debugging as easy as possible
+		runner.T().Log("================================================================================")
+		runner.T().Log("Kibana error response:")
+		runner.T().Log(string(respDump))
+		runner.T().FailNow()
+	}
 }
 
 func (runner *MetricsRunner) TestBeatsMetrics() {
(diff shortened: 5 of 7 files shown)


internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,263 @@
...
+			err := d.collectorConfig.Unmarshal(&serviceCfg)
+			if err != nil {
+				return fmt.Appendf(nil, "error: failed to get internal telemetry: %v", err)
+			}
+			if serviceCfg.Service.Telemetry.Metrics.Level == configtelemetry.LevelNone {
+				return []byte("internal telemetry is disabled")
+			}
+			addr := extractMetricAddress(serviceCfg.Service.Telemetry.Metrics.Readers)
+
+			req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s/metrics", addr), nil)
```
VihasMakwana: @blakerouse @swiatekm Continuing from our discussion https://github.com/elastic/elastic-agent/pull/10052#discussion_r2369801285:

We can't call this endpoint from the new `otel diagnostics` subcommand because we don't know which port the metrics exporter is using. To get the port, we need access to the running config, which isn't available in `otel diagnostics`. 

We can take a guess and assume it's running on default port i.e. `8888`, but what if user has changed that?

I think this this is the simplest way to get the internal telemetry for both commands i.e. `elastic-agent diagnostics` and `elastic-agent otel diagnostics`.

VihasMakwana:
Most of the changes in this PR are from `NOTICE.txt`

pkg/control/v2/client/dial.go
```diff
@@ -18,13 +18,13 @@ import (
...
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
-		grpc.WithContextDialer(dialer),
+		grpc.WithContextDialer(Dialer),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMsgSize)),
 	)
 	return grpc.DialContext(ctx, address, opts...) //nolint:staticcheck // Only the deprecated version allows this call to be blocking
 }
 
-func dialer(ctx context.Context, addr string) (net.Conn, error) {
+func Dialer(ctx context.Context, addr string) (net.Conn, error) {
```
VihasMakwana, line 27: I'm making this public as we'll reuse this `Dialer` in https://github.com/elastic/elastic-agent/pull/10052/ to deal with npipe/unix dialing for different OS'.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
swiatekm: This works fine for `elastic-agent diagnostics` because we can infer where the unix domain socket file is based on the location of the agent binary. We can probably achieve the same thing by making agent write a predetermined file containing the port (and maybe other information?) when it starts the otel collector. Then the diagnostic command can read that file and know what to do.

Not sure if this is worth the hassle, though.

VihasMakwana: > This works fine for elastic-agent diagnostics because we can infer where the unix domain socket file is based on the location of the agent binary. We can probably achieve the same thing by making agent write a predetermined file containing the port (and maybe other information?) when it starts the otel collector. Then the diagnostic command can read that file and know what to do.

Yeah, I did a similar thing in my first PR but it was too much for now. So, we decided to start simpler and later add a socket/file if needed. 

> Not sure if this is worth the hassle, though.

With 9.2 FF, it's most likely not 😅

VihasMakwana: @swiatekm @blakerouse @cmacknz As per our discussion, I've removed the internal telemetry for now. 

I've created a follow-up issue to work later: https://github.com/elastic/elastic-agent/issues/10152

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,239 @@
...
+		return err
+	}
+	ipc.CleanupListener(d.logp, d.diagnosticsConfig.Endpoint)
+	return nil
+}
+
+func (d *diagnosticsExtension) registerGlobalDiagnostics() {
+	d.globalHooks["collector_config"] = &diagHook{
+		description: "full collector configuration",
+		filename:    "edot/otel-merged.yaml",
```
cmacknz:
```suggestion
		filename:    "edot/otel-merged-actual.yaml",
```

We use the expected vs actual naming elsewhere and I think it would apply here too, the one at the root is the expected from agent and the one here is what the collector was last (actually) run.

internal/pkg/otel/extension/elasticdiagnostics/README.md
```diff
@@ -0,0 +1,62 @@
+# elasticdiagnosticsextension
+
+`elasticdiagnosticsextension` is an internal package for peforming diagnostics and is used in conjunction with EDOT.
```
cmacknz, line 3: Mention that it's designed to return diagnostics compatible with the control protocols diagnostics action result type https://github.com/elastic/elastic-agent-client/blob/888026ef85e1c9190fe76eb158cf21d9c9c02920/elastic-agent-client.proto#L424-L437

VihasMakwana: Done

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,239 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/elastic/elastic-agent/internal/pkg/diagnostics"
+	"github.com/elastic/elastic-agent/pkg/ipc"
+)
+
+var (
+	_ component.Component = &diagnosticsExtension{}
```
pkoutsovasilis: nit: no need to allocate also the value, just the type is enough to provide this build time guarantee you are after
```suggestion
	_ component.Component = (*diagnosticsExtension)(nil)
```

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (d *diagnosticsExtension) Shutdown(ctx context.Context) error {
+	if d.server == nil {
+		return nil
+	}
+	if err := d.server.Shutdown(ctx); err != nil {
+		return err
+	}
+	ipc.CleanupListener(d.logp, d.diagnosticsConfig.Endpoint)
+	return nil
```
pkoutsovasilis, line 102: if shutdown can/is-allowed to be called from a separate goroutine here we have an issue

internal/pkg/otel/extension/elasticdiagnostics/factory.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/extension"
+
+	"github.com/elastic/elastic-agent/internal/pkg/otel/extension/elasticdiagnostics/internal/metadata"
+)
+
+var (
+	DiagnosticsExtensionID component.Type = metadata.Type
```
pkoutsovasilis, line 17: where is this used exactly?

VihasMakwana: This will be used in https://github.com/elastic/elastic-agent/pull/10052/, to inject the extension.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
swiatekm: It can be called concurrently, and might also be called without `Start` having been called first, as per https://github.com/open-telemetry/opentelemetry-collector/blob/6ccdc890b16fc5b148dadfb8c7622da655823f21/component/component.go#L39. Probably safest to wrap it in a mutex.

VihasMakwana: Thanks for your review! 😄

internal/pkg/otel/extension/elasticdiagnostics/generated_component_test.go
```diff
@@ -0,0 +1,76 @@
...
+		// Use a different endpoint on Windows as /tmp/test.sock is not valid
+		cfg.Endpoint = fmt.Sprintf("npipe://%s", cfg.Endpoint)
+	}
+	t.Run("shutdown", func(t *testing.T) {
+		e, err := factory.Create(context.Background(), extensiontest.NewNopSettings(typ), cfg)
+		require.NoError(t, err)
+		err = e.Shutdown(context.Background())
+		require.NoError(t, err)
+	})
+	t.Run("lifecycle", func(t *testing.T) {
```
swiatekm, line 51: We should have tests here checking if calling Shutdown twice is valid, if calling Shutdown without Start doesn't panic, and so on.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating listener: %w", err)
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/diagnostics", d)
+
+	d.server = &http.Server{
+		Handler:           mux,
+		ReadHeaderTimeout: 30 * time.Second,
```
swiatekm, line 81: Is this the same value we use in the control protocol?

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,244 @@
...
+		})
+	}
+
+	b, err := json.Marshal(Response{
+		GlobalDiagnostics:    globalResults,
+		ComponentDiagnostics: componentResults,
+	})
+	if err != nil {
+		d.logger.Error("Failed marshaling response", zap.Error(err))
+		w.WriteHeader(503)
```
swiatekm: Shouldn't this be a 500 instead? It should only happen due to bugs in either agent or beats.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,244 @@
...
+		w.Header().Add("content-type", "application/json")
+		if _, err := fmt.Fprintf(w, "{'error':'%v'}", err); err != nil {
+			d.logger.Error("Failed writing response to client.", zap.Error(err))
+		}
+		return
+	}
+	w.Header().Add("content-type", "application/json")
```
swiatekm: Looks like we do this both here and in the error branch, so how about moving it in front of the error check?

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: Control protocol uses grpc, so nope. The closest value is grpc's `ConnectionTimeout`, which by default is 120 seconds. We don't override this in control protocol.

VihasMakwana: I think 30 seconds is okay. Many places in collector/contrib uses various values from 20 seconds to 1 minute. (default is 1 minute).
what do you say?

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: Done!

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: Done!

internal/pkg/otel/extension/elasticdiagnostics/generated_component_test.go
VihasMakwana: Done.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
swiatekm: I think it's fine, just wanted to know where the value originated.

swiatekm (dismissed):

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (d *diagnosticsExtension) registerGlobalDiagnostics() {
+	d.globalHooks["collector_config"] = &diagHook{
+		description: "full collector configuration",
+		filename:    "edot/otel-merged-actual.yaml",
+		contentType: "application/yaml",
+		hook: func() []byte {
+			if d.collectorConfig == nil {
+				return []byte("no active OTeL Configuration")
```
cmacknz:
```suggestion
				return []byte("no active OTel Configuration")
```

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,243 @@
...
+	logger   *zap.Logger
+	logp     *logp.Logger
+
+	diagnosticsConfig *Config
+	collectorConfig   *confmap.Conf
+	componentHooks    map[string][]*diagHook
+	globalHooks       map[string]*diagHook
+
+	hooksMtx sync.Mutex
+	confgMtx sync.Mutex
```
cmacknz:
```suggestion
	configMtx sync.Mutex
```

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,243 @@
...
+	return nil
+}
+
+func (d *diagnosticsExtension) registerGlobalDiagnostics() {
+	d.globalHooks["collector_config"] = &diagHook{
+		description: "full collector configuration",
+		filename:    "edot/otel-merged-actual.yaml",
+		contentType: "application/yaml",
+		hook: func() []byte {
+			if d.collectorConfig == nil {
```
cmacknz, line 113: This hook needs to hold `configMtx` until it returns

cmacknz: This means `configMtx` will be held while `hooksMtx` is also held in `ServeHTTP` which I think is fine but wanted to highlight this in case I missed something about the interactions.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
```diff
@@ -0,0 +1,243 @@
...
+	return nil
+}
+
+// RegisterDiagnosticHook API exposes the ability for beat receivers to register their hooks.
+// NOTE: Changing the function signature will require changes to libbeat and beatreceivers. Proceed with caution.
+func (d *diagnosticsExtension) RegisterDiagnosticHook(componentName string, description string, filename string, contentType string, hook func() []byte) {
+	d.hooksMtx.Lock()
+	defer d.hooksMtx.Unlock()
+	if _, ok := d.componentHooks[componentName]; ok {
+		d.componentHooks[componentName] = append(d.componentHooks[componentName], &diagHook{
```
cmacknz, line 155: The type here being `map[string][]*diagHook` means the `[]*diagHook` can end up having multiple of the exact same hook registered and nothing would ever clean them up or prevent that except a collector restart.

Multiple calls to `Start` over in Beats would cause this for example https://github.com/elastic/beats/blob/b6c79e4437dc98e5d038af7ad40a57c01c4b6318/x-pack/libbeat/cmd/instance/receiver.go#L104-L114

One mitigation would be to have the Beat unregister the hook in `Stop` but you may be better off using the arguments to construct a key and just having a plan `map[string]*diagHook` as the type.

I think this type maybe was carried over from the original design but then the components were sub-processes and having this be called multiple times was impossible because the process was always restarted.

cmacknz: We are using the array here to allow multiple hooks for the same component which would still work, and could even use a multi-level map if that turned out to be easier (it would make deleting all hooks for a component easier if we needed to do that when the component goes away).

internal/pkg/otel/extension/elasticdiagnostics/extension_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+		b, err := io.ReadAll(resp.Body)
+		require.NoError(collect, err)
+		res := Response{}
+
+		// test that the response can be unmarshalled
+		require.NoErrorf(collect, json.Unmarshal(b, &res), "failed to unmarshal response: %s", string(b))
+		require.NotEmpty(collect, res.GlobalDiagnostics)
+		foundCPU := false
+		for _, global := range res.GlobalDiagnostics {
```
cmacknz, line 92: You should validate the complete structure of the response is what you need it to be, and also validate that the profiles types are not empty and can be uncompressed.

"Surely if it exists with the right name it is usable" you may think, but https://github.com/elastic/elastic-agent/issues/2530 taught us the answer to that is "no" the hard way and you can find the validation of the profiles you should do in https://github.com/elastic/elastic-agent-client/pull/69/files

You should also manually verify `go tool pprof` does what you expect with what the extension outputs.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: > The type here being map[string][]*diagHook means the []*diagHook can end up having multiple of the exact same hook registered and nothing would ever clean them up or prevent that except a collector restart.

We do a collector restart because otel doesn't support runtime config update. Ref https://github.com/elastic/elastic-agent/blob/fb467a707c478961804d229e5781548d79298624/internal/pkg/otel/manager/manager.go#L367-L369

Every time we get a config update, we stop the old process and start a new process with latest config 

https://github.com/elastic/elastic-agent/blob/fb467a707c478961804d229e5781548d79298624/internal/pkg/otel/manager/manager.go#L404-L406

For now, I don't think this is a problem but we can definitely work on it as a follow-up. I'll do some manual testing for this and get back to you.

internal/pkg/otel/extension/elasticdiagnostics/extension_test.go
VihasMakwana: > You should validate the complete structure of the response is what you need it to be, and also validate that the profiles types are not empty and can be uncompressed.

Agree. I've done that. I plan to do it again before merging this. I'll update the integration test in https://github.com/elastic/elastic-agent/pull/10052 to verify profiles.

internal/pkg/otel/extension/elasticdiagnostics/extension.go
VihasMakwana: nice catch. thankss!

internal/pkg/otel/extension/elasticdiagnostics/generated_component_test.go
```diff
@@ -0,0 +1,97 @@
...
+		require.EventuallyWithT(t, func(collect *assert.CollectT) {
+			firstExt, err := factory.Create(context.Background(), extensiontest.NewNopSettings(typ), cfg)
+			require.NoError(collect, err)
+			require.NoError(collect, firstExt.Start(context.Background(), newMdatagenNopHost()))
+			require.NoError(collect, firstExt.Shutdown(context.Background()))
+		}, 5*time.Second, 100*time.Millisecond)
+
+		require.EventuallyWithT(t, func(collect *assert.CollectT) {
```
VihasMakwana, line 59: This is needed because on windows, the named pipe is only deleted after last open handle is closed. 
For some reason, if we immediately try to create a new extension after shutting down previous extension, we might encounter `access is denied`/`pipe already exists` error.

Unfortunately, we don't have control over pipe deletion [from user space](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-createpipe#:~:text=An%20instance%20of%20a%20pipe%20is%20always%20deleted%20when%20the%20last%20handle%20to%20the%20instance%20of%20the%20named%20pipe%20is%20closed.). The best we can do is retry the test for few more times. 

It passes consistently on my windows with 1000 tries with retries.

internal/pkg/otel/extension/elasticdiagnostics/extension_test.go
```diff
@@ -65,14 +87,39 @@ func TestExtension(t *testing.T) {
...
+				require.Equal(collect, "Test Hook2", comp.Description)
+				require.Equal(collect, []byte("diagnostic data 2"), comp.Content)
+			}
+		}
 		require.NoError(collect, resp.Body.Close())
 	}, 10*time.Second, 1*time.Millisecond, "extension did not start in time")
 }
+
+func verifyPprof(t *testing.T, content []byte) {
+	prof, err := profile.ParseData(content)
```
cmacknz, line 122: TIL this exists, though I think it might accept the non-gz compressed variants that `go tool pprof` doesn't like (just quickly skimming the docs)

cmacknz (approved):

PR #10143, opened 2025-09-25
dev-tools: add profiling special case for edot

florianl:
## What does this PR do?

https://github.com/elastic/elastic-agent/pull/9888 added profiling as an receiver to EDOT. Fix tooling to add this receiver to autogenerated list of components for EDOT.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

PR #10144, opened 2025-09-25
Add comments on docs generation

theletterf:
This adds comments on docs generation to two pages of reference docs.

+CC @florianl @AlexanderWert


docs/reference/edot-collector/components.md
```diff
@@ -20,6 +20,11 @@ The {{edot}} (EDOT) Collector includes embedded Collector components from the [O
 
 The components included in the EDOT Collector are categorized into **[Core]** and **[Extended]** components. The following table describes the current components included in the EDOT Collector, their source, and support status.
 
+% The following table is automatically generated from the EDOT Collector source code.
+% Automation is handled by /docs/scripts/update-docs/update-components-docs.py, which
```
florianl, line 24: Why is this using a different approach, than it is used for generating https://github.com/elastic/elastic-agent/tree/main/internal/pkg/otel#components?

Does  /docs/scripts/update-docs/update-components-docs.py and https://github.com/elastic/elastic-agent/blob/main/dev-tools/mage/otel/deps.go generate the same output? Does special case handling (like https://github.com/elastic/elastic-agent/pull/10143) be updated in multiple places?
I'm mostly concerned about the differences reading and interpreting go.mod.

theletterf: That's a valid concern. I wasn't aware of the README automation when I ported the docs over from the `opentelemetry` repo. For the time being, I would just document current status.

@AlexanderWert Would it make sense to review the existing automation for the README and make it common across both concerns?

PR #10153, opened 2025-09-25
Rewrite the beats receiver log test using common tooling

swiatekm:
## What does this PR do?

Re-enables the integration test for comparing logs between beats processes and beats receivers and rewrites it to use normal integration test tooling, making it simpler and more observable:

- Instead of uninstalling and reinstalling agent with new configuration, we install it once and just let it reload the configuration, verifying that this happened via status.
- Use the normal install process and let agent run as a service instead of running it manually.
- Fetches the logs by running `elastic-agent logs` instead of capturing agent output directly.
- Relax the status check to only verify what we care about. This test sets an ES host that doesn't exist, so the status will eventually be degraded, but the test doesn't really care about this.

## Why is it important?

This test should be enabled and reliable.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/9890

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -722,13 +722,13 @@ agent.monitoring.enabled: false
 	require.NoError(t,
 		template.Must(template.New("config").Parse(configTemplate)).Execute(&configBuffer,
 			configOptions{
-				RuntimeExperimental: component.ProcessRuntimeManager.String(),
+				RuntimeExperimental: string(component.ProcessRuntimeManager),
 			}))
 	processConfig := configBuffer.Bytes()
 	require.NoError(t,
 		template.Must(template.New("config").Parse(configTemplate)).Execute(&configBuffer,
 			configOptions{
-				RuntimeExperimental: component.OtelRuntimeManager.String(),
+				RuntimeExperimental: string(component.OtelRuntimeManager),
 			}))
 	receiverConfig := configBuffer.Bytes()
 	// this is the context for the whole test, with a global timeout defined


testing/integration/ess/beat_receivers_test.go
```diff
@@ -735,53 +733,74 @@ agent.monitoring.enabled: false
...
-	})
+	// change configuration and wait until the beats receiver is healthy
+	err = fixture.Configure(ctx, receiverConfig)
+	require.NoError(t, err)
 
 	require.EventuallyWithT(t, func(collect *assert.CollectT) {
 		var statusErr error
-		status, statusErr := fixture.ExecStatus(agentReceiverCtx)
+		status, statusErr := fixture.ExecStatus(ctx)
 		assert.NoError(collect, statusErr)
```
ycombinator: I know this line wasn't changed by this PR but should this be a `require.NoError`? IOW, is there any benefit to checking `status` if we have a non-nil `statusErr` at this point?

testing/integration/ess/beat_receivers_test.go
```diff
@@ -735,53 +733,74 @@ agent.monitoring.enabled: false
...
-	agentProcessCtx, agentProcessCancel := context.WithCancel(ctx)
-	fixture, cmd, output := prepareAgentCmd(t, agentProcessCtx, processConfig)
+	// since we set the output to a nonexistent ES endpoint, we expect it to be degraded, but the input to be healthy
+	assertBeatsReady := func(t *assert.CollectT, status *atesting.AgentStatusOutput, runtime component.RuntimeManager) {
+		t.Helper()
 
-	require.NoError(t, cmd.Start())
+		var componentVersionInfoName string
+		switch runtime {
+		case "otel":
```
ycombinator:
```suggestion
		case component.OtelRuntimeManager:
```

testing/integration/ess/beat_receivers_test.go
ycombinator:
```suggestion
				RuntimeExperimental: component.ProcessRuntimeManager.String(),
```

ycombinator:
```suggestion
				RuntimeExperimental: component.OtelRuntimeManager.String(),
```

testing/integration/ess/beat_receivers_test.go
swiatekm: It should be, yeah. I'll fix it.

PR #10158, opened 2025-09-25
Update run command to have a critical section to ensure errors are always logged

### LINKED ISSUE #9099: Unable to recover from corrupted upgrades when the host system reboots
We recently had a support case where the host Windows system appeared to crash or shutdown shortly after we entered the "Upgrade Replacing" state. This is the state where the the top level `elastic-agent` symlink has been pointed at the newly downloaded and extracted version of the agent. This was the state of the upgrade marker at the time of the problem:

```yaml
version: 9.0.3+build202507110136
hash: 5f1479
versioned_home: data\elastic-agent-9.0.3+build202507110136-5f1479
updated_on: 2025-07-15T18:21:24.0025395-07:00
prev_version: 9.0.3
prev_hash: 5f1479
prev_versioned_home: data\elastic-agent-9.0.3-5f1479
acked: false
action:
    id: 1072a966-3972-4af0-a7e4-bdfd6fc90f5c
    type: UPGRADE
    version: 9.0.3+build202507110136
details:
    target_version: 9.0.3+build202507110136
    state: UPG_REPLACING
    action_id: 1072a966-3972-4af0-a7e4-bdfd6fc90f5c
    metadata:
        download_percent: 1
        retry_until: null
desired_outcome: UPGRADE
```

In the specific case the contents of the `.elastic-agent.active.commit` and the `.package-version` files had the commit hash corrupted into the bytes `0000000000000a` instead of the expected `5f147906dab8702d8738a71f627a07a0725b5abc` for the `9.0.3+build202507110136` hotfix release. 

All `elastic-agent` commands were failing with the following error:
```
Error initializing version information: parsing version "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00": version string does not match expected format
```

In this situation, the upgrade watcher will not have been running because the host system rebooted, and while the OS will attempt to start the `elastic-agent` service because the path to the executable is still valid the service likely exited because
of the dependency on valid version files. This likely makes this a situation where the existing upgrade design could have recovered if we started the upgrade watcher as early as possible with no dependency on the version (perhaps having to discover the path to the watcher differently if the path files are corrupted). 

We should setup tests that stage an agent in various representative versions of this situation and make changes so that it can recover. The conditions would be:

* The agent service is not yet started (or agent is not installed to simplify control)
* Two versions of the agent on disk in the data sub-directory
* The upgrade watcher is not yet running and/or has been stopped
* The active version of the agent has a runtime problem but is able to start
  * The upgrade marker could be correct
  * The version files could be corrupt
  * The path layout could not be what is expected
  * Etc
* The agent watcher should be able to start and switch the symlink to the working version of the agent.

If the top level `elastic-agent` symlink were to be corrupted, or the active version of the agent executable were to be corrupted when the pervious one was not, we would likely need another design change to address this situation. We can discuss doing this depending on complexity but prioritize fixing the known and observed bug above in the current design first.

blakerouse:
## What does this PR do?

Updates the `run` subcommand to always log any errors to the logger.

## Why is it important?

There where still some cases where the Elastic Agent can fail to run but the reason is not written to the log.

## Checklist

- ~~[ ] I have added an integration test or an E2E test`~

## Disruptive User Impact

None.

## Related issues

- Closes #9099

### DIFF
--- a/internal/pkg/agent/cmd/run.go
+++ b/internal/pkg/agent/cmd/run.go
@@ -6,6 +6,7 @@ package cmd
 
 import (
 	"context"
+	goerrors "errors"
 	"fmt"
 	"net/url"
 	"os"
@@ -88,7 +89,6 @@ func newRunCommandWithArgs(_ []string, streams *cli.IOStreams) *cobra.Command {
 			testingMode, _ := cmd.Flags().GetBool("testing-mode")
 			if err := run(nil, testingMode, fleetInitTimeout); err != nil && !errors.Is(err, context.Canceled) {
 				fmt.Fprintf(streams.Err, "Error: %v\n%s\n", err, troubleshootMessage())
-				logExternal(fmt.Sprintf("%s run failed: %s", paths.BinaryName, err))
 				return err
 			}
 			return nil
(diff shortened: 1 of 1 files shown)


internal/pkg/agent/cmd/run.go
```diff
@@ -141,56 +141,81 @@ func run(override application.CfgOverrider, testingMode bool, fleetInitTimeout t
...
+			errs = append(errs, err)
+		}
+	}
+
+	// try load config, but don't error yet
 	cfg, err := loadConfig(ctx, override)
 	if err != nil {
-		return err
+		// failed to load configuration, just load the default to create the logger
+		cfg = configuration.DefaultConfiguration()
```
ycombinator, line 197: Should we also collect this `err` in `errs` so we know that the configuration couldn't be loaded and the default configuration is being used?

internal/pkg/agent/cmd/run.go
```diff
@@ -141,56 +141,81 @@ func run(override application.CfgOverrider, testingMode bool, fleetInitTimeout t
...
 	modifiers ...component.PlatformModifier,
 ) error {
-	err := coordinator.RestoreConfig()
+	var errs []error
+
+	// early handleUpgrade, but don't error yet
+	upgradeDetailsFromMarker, err := handleUpgrade()
 	if err != nil {
-		return err
+		errs = append(errs, err)
```
ycombinator: Might be good to add some context to these errors as they'll all be logged at once?

```suggestion
		errs = append(errs, fmt.Errorf("failed to handle upgrade: %w", err))
```

internal/pkg/agent/cmd/run.go
```diff
@@ -200,10 +225,42 @@ func runElasticAgent(
...
+	if len(errs) > 0 {
+		return logReturn(l, goerrors.Join(errs...))
+	}
+
+	// actually run the agent now
+	err = runElasticAgent(ctx, cancel, baseLogger, l, cfg, override, stop, testingMode, fleetInitTimeout, upgradeDetailsFromMarker, modifiers...)
+	if err != nil {
+		return logReturn(l, err)
+	}
+	return nil
```
ycombinator: Nit.
```suggestion
	return logReturn(l, err)
```

internal/pkg/agent/cmd/run.go
blakerouse: Yes I should have. Thanks for catching that. Fixed here https://github.com/elastic/elastic-agent/pull/10158/commits/159d2613f3ec2a2cc9ca4e0b596e552f7b86e85f

internal/pkg/agent/cmd/run.go
blakerouse: Yeah that is a good idea. Fixed here https://github.com/elastic/elastic-agent/pull/10158/commits/159d2613f3ec2a2cc9ca4e0b596e552f7b86e85f

internal/pkg/agent/cmd/run.go
blakerouse: Fixed in https://github.com/elastic/elastic-agent/pull/10158/commits/159d2613f3ec2a2cc9ca4e0b596e552f7b86e85f

blakerouse:
@ycombinator Thanks for the review. I implemented the requested fixes, if you can give this another review.

Issue #10154, opened 2025-09-25
[beats receivers] Beatsauth extension is not removed or reused when configuration is reloaded


cmacknz:
During internal testing we observe multiple instances of the beatsauth extension being present in the otel-merged.yml file after several configuration reloads have happened.

```
    extensions:
        - beatsauth/_agent-component/monitoring
        - beatsauth/_agent-component/monitoring
        - beatsauth/_agent-component/monitoring
```

swiatekm:
For the record, I think this currently works fine anyway due to the otel collector implementation effectively deduplicating the list. Extensions can optionally depend on each other, so the otel collector does a [topological sort ](https://github.com/open-telemetry/opentelemetry-collector/blob/6ccdc890b16fc5b148dadfb8c7622da655823f21/service/extensions/graph.go#L28)on the dependency graph, which has the side effect of collapsing all extensions with the same id into one.

PR #10172, opened 2025-09-26
Deduplicate extensions when generating otel config

swiatekm:
## What does this PR do?

Ensures the extension list in the otel config we generate for components always has only one instance of each extension. Currently, we get the same beatsauth extension multiple times, which only works due to a quirk of otel's config processing.

## Why is it important?

The config we currently generate for the otel collector is invalid, and is only accepted and works correctly due to implementation details.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10154

### DIFF
--- a/internal/pkg/otel/translate/otelconfig.go
+++ b/internal/pkg/otel/translate/otelconfig.go
@@ -63,8 +63,8 @@ func GetOtelConfig(
 	if len(components) == 0 {
 		return nil, nil
 	}
-	otelConfig := confmap.New()      // base config, nothing here for now
-	extensionList := []interface{}{} // we have to maintain a list because otel does not merge lists, it overrides them. This is a known issue: see https://github.com/open-telemetry/opentelemetry-collector/issues/8754
+	otelConfig := confmap.New()     // base config, nothing here for now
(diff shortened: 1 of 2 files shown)


cmacknz (approved):
Replaced the content of otelconfig.go with the content from one commit before this change but kept the new test, and the new test successfully reproduces the problem and confirms this is the fix.

```sh
❯ gotest ./internal/pkg/otel/translate/... -run TestGetOtelConfig
--- FAIL: TestGetOtelConfig (0.01s)
    --- FAIL: TestGetOtelConfig/multiple_filestream_inputs_and_one_output (0.00s)
        otelconfig_test.go:904:
                Error Trace:    /Users/cmackenzie/go/src/github.com/elastic/elastic-agent/internal/pkg/otel/translate/otelconfig_test.go:904
                Error:          Not equal:
                                expected: map[string]interface {}
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -417,3 +417,4 @@
                                  (string) (len=7) "service": (map[string]interface {}) (len=2) {
                                -  (string) (len=10) "extensions": ([]interface {}) (len=1) {
                                +  (string) (len=10) "extensions": ([]interface {}) (len=2) {
                                +   (string) (len=34) "beatsauth/_agent-component/default",
                                    (string) (len=34) "beatsauth/_agent-component/default"
                Test:           TestGetOtelConfig/multiple_filestream_inputs_and_one_output
```

👍

PR #10202, opened 2025-09-29
[ipc/;listener][windows] Implement retry mechanism for listener creation

VihasMakwana:
## What does this PR do?

Adds a retry mechanism when creating a named pipe listener on Windows.
Instead of failing immediately if listener creation fails, it now retries for up to 5 seconds with a short delay between attempts.

## Why is it important?

Sometimes the named pipe might not be immediately available. This was observed when using monitoring with beat receivers.

## Checklist

### DIFF
--- a/pkg/ipc/listener_windows.go
+++ b/pkg/ipc/listener_windows.go
@@ -9,8 +9,10 @@ package ipc
 import (
 	"fmt"
 	"net"
+	"os"
 	"os/user"
 	"strings"
+	"time"
 
 	"golang.org/x/sys/windows"
 
@@ -43,7 +45,13 @@ func CreateListener(log *logger.Logger, address string) (net.Listener, error) {
 }
 
 func CleanupListener(log *logger.Logger, address string) {
-	// nothing to do on windows
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		_, err := os.Stat(npipe.TransformString(address))
+		if os.IsNotExist(err) {
+			break
+		}
+	}
 }
 
 func securityDescriptor(log *logger.Logger) (string, error) {
--- a/pkg/ipc/listener_windows.go
+++ b/pkg/ipc/listener_windows.go
@@ -9,7 +9,6 @@ package ipc
 import (
 	"fmt"
 	"net"
-	"os"
 	"os/user"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


leehinman:
I don't think this will fix https://github.com/elastic/ingest-dev/issues/6133

That error is:
```
filebeatreceiver:could not start the HTTP server for the API: failed to listen on the named pipe \\.\pipe\xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock: open \\.\pipe\xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock: Access is denied.
```

That error is coming from filebeatreceiver, not elastic-agent.  I think we need to do the retry in the libbeat api server:

https://github.com/elastic/beats/blob/e55e0bedd6176c7c6c88dfdf012c14620cb85a29/libbeat/api/server.go#L73-L79

VihasMakwana:
> I don't think this will fix [elastic/ingest-dev#6133](https://github.com/elastic/ingest-dev/issues/6133)
> 
> That error is:
> 
> ```
> filebeatreceiver:could not start the HTTP server for the API: failed to listen on the named pipe \\.\pipe\xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock: open \\.\pipe\xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock: Access is denied.
> ```
> 
> That error is coming from filebeatreceiver, not elastic-agent. I think we need to do the retry in the libbeat api server:
> 
> https://github.com/elastic/beats/blob/e55e0bedd6176c7c6c88dfdf012c14620cb85a29/libbeat/api/server.go#L73-L79

Oh!
I faced the same problem with elastic-agent and beatreceivers and this is what fixed for me.
Anyways, I'll apply the same fix in libbeat.

VihasMakwana:
> I don't think this will fix [elastic/ingest-dev#6133](https://github.com/elastic/ingest-dev/issues/6133)
> 
> That error is:
> 
> ```
> filebeatreceiver:could not start the HTTP server for the API: failed to listen on the named pipe \\.\pipe\xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock: open \\.\pipe\xTEtpJ7117ppc6OYvJCaYHbDW8mLjXGe.sock: Access is denied.
> ```
> 
> That error is coming from filebeatreceiver, not elastic-agent. I think we need to do the retry in the libbeat api server:
> 
> https://github.com/elastic/beats/blob/e55e0bedd6176c7c6c88dfdf012c14620cb85a29/libbeat/api/server.go#L73-L79

Thank you for letting me know, Lee!

pkg/ipc/listener_windows.go
blakerouse, line 34: To make this change this is really going to have to take a context. Which I know is not great because changing the function signature means it needs to be updated everywhere. But if this is going to take up to 5 seconds, context cancelling will need to be handled. The user might Ctrl-C the elastic-agent it should not be stuck in this section and wait up to 5 seconds before it returns.

VihasMakwana: @blakerouse what if we create a `signals.Notify(..)` here and exit on `sigint` or `sigterm`, instead of passing context all the way from command?

cmacknz: https://github.com/search?q=repo%3Aelastic%2Felastic-agent%20CreateListener&type=code

It looks to me like there are only 3 calls to this, two of which have a context already. The one that doesn't could just be given `context.TODO()`  without making anything worse. Listening to signals in more than one place would be way messier IMO and possibly complicate the problems we are already having with the Windows service lifetime being managed properly.

https://github.com/elastic/elastic-agent/blob/278440a1d520eeeffe713b730416e0487fc9f64a/pkg/component/runtime/conn_info_server.go#L31-L38

ebeahan:
@VihasMakwana do you plan on coming back to this PR in the near future? Or can we close until you're ready to resume?

VihasMakwana:
I'll close this for now 👍

PR #10217, opened 2025-09-30
Fix error logging for provider failures

### LINKED ISSUE #9284: Elastic Agent: env provider fails with unhelpful log message %!s(<nil>)
Elastic Agent Version:
8.18.4 (running as Fleet-managed agent via Helm Chart)

Elastic Stack Version (Cloud):
8.18.4 (Elastic Cloud deployment)

Deployment Context:
Agent deployed via Helm (elastic-agent chart v8.18.4)
Mode: daemonset

All required ${env.*} variables are explicitly defined using extraEnvs in the Helm chart
Variables include:
NODE_NAME
NODE_IP
POD_NAME
KUBERNETES_SERVICE_HOST
KUBERNETES_SERVICE_PORT
etc.

Observed Behavior:
In the Elastic Agent logs, the following error message appears repeatedly every 30 seconds:

provider "env" failed to run (will retry in 30s): %!s(<nil>)

Expected Behavior:
Either no error (if all required env values are present), or
A meaningful error message indicating which variable is missing or causing the failure

Actual Behavior:
The log contains a Go formatting fallback string %!s(<nil>), which strongly suggests that the error value passed to the logger is nil, and the Stringer is not handled correctly.

Code Reference:
The problematic log line seems to originate from controller.go line 471, inside the composable.providers.env provider controller:

logger.Errorf("provider %q failed to run (will retry in %s): %s", provider.Type(), retryDur, err)

If err is nil, this results in the log message:
provider "env" failed to run (will retry in 30s): %!s(<nil>)

Root Cause Hypothesis:
err may be nil, and the retry mechanism is triggered regardless

Notes:
Agent behavior is otherwise functional — all Kubernetes metrics and logs are collected correctly and delivered to Elastic
This issue affects log clarity and may mislead users during debugging

straistaru:
## What does this PR do?

Avoid logging a nil error for provider exits.

## Why is it important?

Some providers intentionally stop and return a nil error; they already log the exit reason internally. The previous log formatted a nil error into "%!s()", which can be confusing (e.g. `provider "env" failed to run (will retry in 30s): %!s()`). This change removes the incorrect formatting and only includes the error text when non-nil.

## Checklist

## Disruptive User Impact

None.

## Related issues

- Closes #9284

### DIFF
--- a/changelog/fragments/1759334752-fix-error-logging-for-provider-failures.yaml
+++ b/changelog/fragments/1759334752-fix-error-logging-for-provider-failures.yaml
@@ -1,0 +1,45 @@
+# REQUIRED
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# REQUIRED for all kinds
+# Change summary; a 80ish characters long description of the change.
+summary: Fix an incorrectly formatted log message when a provider fails
+
+# REQUIRED for breaking-change, deprecation, known-issue
+# Long description; in case the summary is not enough to describe the change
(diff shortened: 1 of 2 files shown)


swiatekm:
If the provider already logged the reason internally, do we need to emit any more log lines? The line you're adding seems confusing as well. If we want to log something here, I'd drop the `with a nil error`, as that's an implementation detail.

ycombinator (approved):

PR #10219, opened 2025-09-30
Fix Windows service registration (C entrypoint)

### LINKED ISSUE #4971: [windows] move service startup to beginning of run function

### LINKED ISSUE #10183: Extend use of the Windows archive proxy to elastic-agent.exe to runtime to register with the Windows service manager as fast as possible
- Relates https://github.com/elastic/elastic-agent/issues/4971
- Relates https://github.com/elastic/elastic-agent/pull/9672

We continue to have reports that the Elastic Agent Windows service is too slow to register with the Windows service manager at boot time, especially on machines that are resource constrained. In these cases the Elastic Defend service has no such problems, indicating the problem is specific to Elastic Agent.

Current theories are that we register with the Windows service manager too late, or that the complex dependency graph causes too much work to happen at init slowing down initial registration. However, there is no concrete root cause yet.

Possible approaches so far have been refactoring the import order to try to speed up initialization in https://github.com/elastic/elastic-agent/pull/9672, and a planned refactor of the dependency graph to split the data collection and OpenTelemetry collector dependencies into a separate binary. These could work but are complex.

Instead, we could use a small Windows binary as the entrypoint whose job is to register with the Windows service manager as fast as possible and then execute the main elastic-agent.exe binary. Functionally this would be very similar to the existing [archive-proxy](https://github.com/elastic/elastic-agent/blob/main/wrapper/windows/archive-proxy/main.go), which is used at installation time to avoid shipping two copies of elastic-agent.exe in the .zip archive since .zip does not support symlinks.

If we extended use of the [archive-proxy](https://github.com/elastic/elastic-agent/blob/main/wrapper/windows/archive-proxy/main.go) executable to this purpose it would:
- Remain after installation as part of normal execution.
- Register with the windows service manager ASAP and maintain the smallest module dependency graph possible.
- Handle Service Manager requests for Status, Stop & Shutdown.
- Ideally be in the versioned data path, so that it can be introduced via upgrade (and similarly rolled back if the introduction goes poorly)
- Exit when the underlying elastic-agent.exe exits so that a misbehaving elastic-agent.exe is not hidden from the Windows service manager.
  - Alternatively the agent could be restarted instantly by the archive-proxy but this potentially adds complexity and we would need a backoff algorithm to avoid eating all available CPU in a restart loop.

blakerouse:
## What does this PR do?

This works by providing an entrypoint for the elastic-agent.exe to not be the go runtime, but instead be a C function.
This `int main()` in `main.c` being the entrypoint allows registering with Windows service manager as soon as it starts.
This is before the go runtime is even initialized or any of the `func init()` for all dependencies are called. Once
the application has registered with the Windows service manager (if its running as a service) then it calls `GoRun`
from `main.go`. At this point now the go runtime will start and all the `func init()` will be performed.

## Why is it important?

This is required because of the size of the dependency graph with Elastic Agent the amount of time that it
can take to start when a system is under heavy load (especially during startup) can cause the Elastic Agent to
not respond to the Windows service manager in-time. This causes the Elastic Agent to not enter the correct state
and it fails to work properly.

## Checklist

- [x] I have added an integration test or an E2E test (current integration tests will validate it)

## Disruptive User Impact

None

Run `mage package` and test it on Windows by installing it and controlling the service. You can also use `procmon.exe` if you want to watch when and how early that it registers with the Windows service manager.

## Related issues

- Closes #4971 
- Closes #10183

### DIFF
--- a/internal/pkg/agent/cmd/status.go
+++ b/internal/pkg/agent/cmd/status.go
@@ -47,7 +47,8 @@ func newStatusCommand(_ []string, streams *cli.IOStreams) *cobra.Command {
 				// the elastic-agent is not healthy. we don't want to write it again as it will dirty
 				// the output. os.Exit cannot be used os we need this to come all the way back up the
 				// call stack.
-				if errors.As(err, &ExitCodeError{}) {
+				var exitCodeErr *ExitCodeError
+				if errors.As(err, &exitCodeErr) {
 					return err
 				}
 				fmt.Fprintf(streams.Err, "Error: %v\n%s\n", err, troubleshootMessage)


windows/main.c
```diff
@@ -0,0 +1,89 @@
...
+void ServiceCleanup(void) {
+    if (serviceStatusHandle) {
+        serviceStatus.dwCurrentState = SERVICE_STOPPED;
+        SetServiceStatus(serviceStatusHandle, &serviceStatus);
+    }
+}
+
+// ServiceCtrlHandler is called when the service is being controlled by Windows
+void WINAPI ServiceCtrlHandler(DWORD ctrlCode) {
+    switch (ctrlCode) {
```
michalpristas, line 24: we should also handle SERVICE_CONTROL_INTERROGATE

michalpristas: https://learn.microsoft.com/en-us/windows/win32/services/service-control-handler-function

> All services must accept and process the SERVICE_CONTROL_INTERROGATE control code.

windows/main.c
```diff
@@ -0,0 +1,89 @@
...
+        case SERVICE_CONTROL_STOP:
+        case SERVICE_CONTROL_SHUTDOWN:
+            serviceStatus.dwCurrentState = SERVICE_STOP_PENDING;
+            serviceStatus.dwWin32ExitCode = 0;
+            serviceStatus.dwCheckPoint = 0;
+            serviceStatus.dwWaitHint = 0;
+            SetServiceStatus(serviceStatusHandle, &serviceStatus);
+            GoStop();
```
michalpristas, line 35: my worry is how do we wait for processes to finish work, flush things
thing to think about before we move this out of draft

windows/main.c
```diff
@@ -0,0 +1,89 @@
...
+
+// ServiceCtrlHandler is called when the service is being controlled by Windows
+void WINAPI ServiceCtrlHandler(DWORD ctrlCode) {
+    switch (ctrlCode) {
+        case SERVICE_CONTROL_STOP:
+        case SERVICE_CONTROL_SHUTDOWN:
+            serviceStatus.dwCurrentState = SERVICE_STOP_PENDING;
+            serviceStatus.dwWin32ExitCode = 0;
+            serviceStatus.dwCheckPoint = 0;
+            serviceStatus.dwWaitHint = 0;
```
bjmcnic, line 33: If we know that `GoStop()` may take a while, perhaps we should explicitly set these instead of relying on OS defaults (I believe 30s, but can be configured differently by Admins).

windows/main.c
```diff
@@ -0,0 +1,89 @@
...
+    SetServiceStatus(serviceStatusHandle, &serviceStatus);
+
+    // register the cleanup function to inform that it is stopped now
+    //
+    // atexit is used so no matter what happens in the goruntime the service is
+    // marked as stopped
+    atexit(ServiceCleanup);
+
+    // run the golang runtime
+    GoRun();
```
bjmcnic: Are we gaining anything with atexit() here?  If we just put a call to ServiceCleanup() after the call to GoRun() won't those be effectively the same?  If the OS terminates us because the GoRun() doesn't return promptly after the stop, or if GoRun() throws an unhandled exception, I don't believe the atexit() queue is going to execute.

windows/main.c
blakerouse: I will look at handling this.

windows/main.c
blakerouse: This flow is just setting it to stop pending and calling `GoStop`. `GoStop` is non-blocking and will return instantly. Its job is to just cancel the context and return.

I have actually updated this flow in my latest commit and I am seeing integration tests success with the latest changes.

windows/main.c
blakerouse: `GoStop` will return instantly. It is non-blocking. It just signals for it to stop, not actually wait for it to stop.

windows/main.c
blakerouse: I have updated this flow to remove the atexit and handle this differently. My believe at the time that the `atexit` would allow things like `os.Exit` to be using in go code or a panic and it would be called. That is not the case, I have updated the code to handle `panics` and no longer using `os.Exit` to ensure that we get a clean boundary between C and go code.

windows/main.c
bjmcnic: I understand that `GoStop()` will not block, but after setting the service to `STOP_PENDING` with the SCM we need to worry about how quickly we get to setting `STOPPED` in `ServiceMain()` when `GoRun()` returns.

windows/main.c
blakerouse: Added here fafd92a1186230cf97146f673732a36d4a19caf8

nfritts:
I assume this doesn't happen because it's using c-archive instead of c-shared, but it's worth confirming that this approach isn't impacted by: https://github.com/golang/go/issues/11100

If the service can non-deterministically exit with a non-0 exitcode, it has the potential to annoy windows and result in complicated SDHs

ebeahan:
Discussed internally that we're not going to continue pursuing this approach at the current time. Closing - will re-open if we decide to revisit.

achiskon:
Is there anyone attending this? Problem persists, no solution provided

cmacknz:
We are trying a different approach first, we believe that the problem is caused by dependency initialization happening before we can register with the windows service manager. We split a large chunk of these dependencies out into a separate binary in https://github.com/elastic/elastic-agent/pull/11821 which we believe will help this. 

We have not managed to reproduce the startup delay internally yet, so we cannot 100% confirm this fixes it, but we are still aware this problem exists and that we have to fix it and will continue to iterate on it.

nimarezainia:
@achiskon if you are experiencing this issue currently, would you be able to verify this potential fix addresses your specific issue? I believe these changes will be available in 9.3

achiskon:
> @achiskon if you are experiencing this issue currently, would you be able to verify this potential fix addresses your specific issue? I believe these changes will be available in 9.3

Sure once 9.3 is final, we can revert to this thread with our results

achiskon:
Can anyone confirm that the fix will be part of 9.3?

At a small customer with 7 endpoints, 6 of them had the elastic agent die after a reboot... This is very inconvenient

ebeahan:
I can confirm the [changes](https://github.com/elastic/elastic-agent/pull/10219#issuecomment-3729424433) @cmacknz mentioned from https://github.com/elastic/elastic-agent/pull/11821 will be shipped as part of 9.3.

Issue #4973, opened 2024-06-21
Handle new action for switching Agent from privileged to unprivileged mode


ycombinator:
**Describe the enhancement:**

Elastic Agents have the ability to run in either privileged mode, i.e. with a privileged user like `root` on Linux systems, or unprivileged mode, i.e. with an unprivileged user.

Moreover, Agents running in privileged mode have the ability to switch themselves to unprivileged mode.  This ability is [being exposed](https://github.com/elastic/elastic-agent/pull/4621) via the Agent CLI.  We now wish to expose this same ability using the Fleet UI.  

⚠️  **Important note:** Fleet UI users should _only_ be switch Agents from privileged to unprivileged mode, **not the other way around**.

For this, Agent will need to handle a new action from Fleet.

**Describe a specific use case for the feature:**

Security: Allowing users to reduce the privileges required to run Elastic Agent.

**What is the definition of done?**

* [ ] A new action is defined for switching Agent from privileged to unprivileged mode
* [ ] When agent receives said action, it switches from privileged to unprivileged mode
* [ ] When the switch succeeds, Agent communicates this success to Fleet
* [ ] If the switch fails, Agent communicates this failure to Fleet
* [ ] Test exercising the success scenario
* [ ] Test exercising the failure scenario

pierrehilbert:
> ⚠️ **Important note:** Fleet UI users should _only_ be switch Agents from privileged to unprivileged mode, **not the other way around**.

Switching the other way will technically be impossible but I agree we should make it clear from the UI when we will add this feature there.

blakerouse:
To add some details on the technical implementation for this work. The flow of this should work as the following:

1. Upon receiving the action to switch to unprivileged mode the Elastic Agent should store that action into the state store, but **NOT** ACK it.
2. Then it should perform the `elastic-agent unprivileged` as a sub-process, ensuring to create the process in a way where when the daemon process is stopped that it will not stop or kill the spawned `elastic-agent unprivileged` process.
3. `elastic-agent unprivileged` should then perform the work (understand here there is a chance that if something goes wrong the process is not coming back without manual intervention) **this needs to be made clear in the UI**
4. Elastic Agent should then restart, read the state store, determine that it has a unprivileged mode action and ACK the action if it is now unprivileged.

ycombinator:
[...]

PR #10231, opened 2025-10-01
Handle `PRIVILEGE_LEVEL_CHANGE` action

michalpristas:
to test this you need a 9.2 kibana with FF config:
```yml
xpack.fleet.enableExperimental: ["enableAgentPrivilegeLevelChange"]
```

This PR adds a handler for PRIVILEGE_LEVEL_CHANGE action.
Once action comes and permissions are fixed we call the code identical to `privilege` or `unprivilege` CLI commands. 

Service manager handles user changes differently Windows being least problematic.
Systemd could do `daemon-reload` but for consistency MacOS and Linux are handled in a same way:
- change to service file is made
- agent is rebooted
- desired user is detecting by reading a service definition file
- permissions are fixed if needed
- if it is non-root, setgid,setuid calls are made to drop privileges
- we continue running.
This is not performed if we are running using proper user

Fixes: #4973


internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+	ch chan coordinator.ConfigChange,
+) *PrivilegeLevelChange {
+	return &PrivilegeLevelChange{
+		log:   log,
+		coord: coord,
+		ch:    ch,
+	}
+}
+
+// Handle handles UNENROLL action.
```
pchila: Did you mean PRIVILEGE_LEVEL_CHANGE ?
```suggestion
// Handle handles PRIVILEGE_LEVEL_CHANGE action.
```

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+		username = action.Data.UserInfo.Username
+		groupname = action.Data.UserInfo.Groupname
+		password = action.Data.UserInfo.Password
+	}
+	username, password = install.UnprivilegedUser(username, password)
+	groupname = install.UnprivilegedGroup(groupname)
+
+	// apply empty config to stop processing
+	unenrollPolicy := newPolicyChange(ctx, config.New(), a, acker, true, false)
+	h.ch <- unenrollPolicy
```
pchila: Is this needed to stop components while applying the change?
Can the change be applied as part of the callback passed to the ReExec without "faking" an empty policy?

internal/pkg/agent/cmd/run.go
```diff
@@ -660,12 +692,11 @@ func initTracer(agentName, version string, mcfg *monitoringCfg.MonitoringConfig)
 
 func setupMetrics(
 	logger *logger.Logger,
-	operatingSystem string,
 	cfg *monitoringCfg.MonitoringConfig,
 	tracer *apm.Tracer,
 	coord *coordinator.Coordinator,
 ) (*reload.ServerReloader, error) {
-	if err := report.SetupMetrics(logger, agentName, version.GetDefaultVersion()); err != nil {
+	if err := report.SetupMetrics(logger, agentName, version.GetDefaultVersion()); err != nil { //nolint:staticcheck // will be fixed in subsequent PR
```
pchila: This is for the report.SetupMetrics deprecation correct?
I am having the same lint violation in another PR :sweat_smile:

internal/pkg/agent/cmd/run_darwin.go
```diff
@@ -0,0 +1,83 @@
...
+// logExternal logs the error to an external log.  On non-windows systems this is a no-op.
+func logExternal(msg string) {
+}
+
+func getDesiredUser() (string, string, error) {
+	serviceName := paths.ServiceName()
+	plistPath := fmt.Sprintf("/Library/LaunchDaemons/%s.plist", serviceName)
+
+	content, err := os.ReadFile(plistPath)
+	if os.IsNotExist(err) {
```
pchila: we can use  `errors.Is` here (see https://pkg.go.dev/os#IsNotExist)
```suggestion
	if errors.Is(err, fs.ErrNotExist) {
```

internal/pkg/agent/cmd/run_darwin.go
```diff
@@ -0,0 +1,83 @@
...
+		// not running as a service
+		return "", "", nil
+	}
+	if err != nil {
+		return "", "", fmt.Errorf("failed to read plist file %s: %w", plistPath, err)
+	}
+
+	dec := plist.NewDecoder(bytes.NewReader(content))
```
pchila: Nit: since we are using a Decoder, which requires a reader, why not opening the file directly with [`os.Open`](fs.ErrNotExist) and using the `*os.File` as a reader ?

internal/pkg/agent/cmd/run_darwin.go
```diff
@@ -0,0 +1,83 @@
...
+	// change group first, setuid will drop permission to change group
+	if ownership.GID > 0 {
+		// not necessary, just in case.
+		if err := syscall.Setegid(ownership.GID); err != nil {
+			return fmt.Errorf("failed to set eGID: %w", err)
+		}
+
+		if err := syscall.Setgid(ownership.GID); err != nil {
+			return fmt.Errorf("failed to set GID: %w", err)
+		}
```
pchila, line 31: Could you elaborate a bit on these 2 system calls?
Why would be the first one "not necessary, just in case" ? What's the difference with the second one ?

internal/pkg/agent/install/componentvalidation/validation.go
pchila, line 1: Naming: why is this new subpackage inside `install ` called `service` ?
Given the hierarchy I was expecting something to do with elastic-agent service setup but if I understand correctly this deals with policy, configuration and state of (service) components

internal/pkg/agent/install/svc.go
```diff
@@ -225,7 +225,10 @@ func changeLaunchdServiceFile(serviceName string, plistPath string, username str
 	if err != nil {
 		return fmt.Errorf("failed to access service file for write at %q: %w", plistPath, err)
 	}
-	defer func() { _ = fileWriter.Close() }()
+	defer func() {
+		_ = fileWriter.Sync()
```
pchila, line 229: any specific reason why the Sync is needed here? (it's not bad in general, just curious if there's a specific scenario that prevents us from relying only on OS buffer flushing on disk)

internal/pkg/agent/install/switch.go
```diff
@@ -119,3 +92,47 @@ func SwitchExecutingMode(topPath string, pt *progressbar.ProgressBar, username s
...
+		var err error
+		ownership, err = EnsureUserAndGroup(username, groupName, pt, username == ElasticUsername)
+		if err != nil {
+			// context for the error already provided in the EnsureUserAndGroup function
+			return ownership, err
+		}
+	}
+
+	// **start critical section**
+	// after this point changes will be made that can leave the installed Elastic Agent broken if they do not
```
pchila, line 109: Would it be worth to log entering/leaving this section? I think it could be useful to know if agent was interrupted in the middle of this block in case of bug investigations

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
michalpristas: we want to stop processing so we can change permissions without components failing.

internal/pkg/agent/cmd/run.go
michalpristas: yes i've seen this addressed in some PR so i'm not addressing it here

internal/pkg/agent/install/componentvalidation/validation.go
michalpristas: it handles service installation

internal/pkg/agent/install/svc.go
michalpristas: because we may reexec before change is written to disk. i've run into this

internal/pkg/agent/cmd/run_darwin.go
michalpristas: we could do that, service files are not that large so i went with this one. better debugging as you can see the content right away

internal/pkg/agent/cmd/run_darwin.go
michalpristas: we could be just with setgid but i was advised by endpoint folks it may be worth to apply seteuid. maybe in case when you don't have privileges to call setgid?

internal/pkg/agent/install/switch.go
michalpristas: we don't have a logger we could use describer. we inject debugDescriber that outputs to log, but you need to have debug enabled. this was called solely from CLI so no log needed.

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
pchila: if we use the re-exec callback, that is run when the components are already stopped, I think.
I am just a little concerned of handling a fake empty policy just to stop components. Is this fake policy saved? Do we uninstall components like endpoint as a consequence ?

internal/pkg/agent/install/componentvalidation/validation.go
pchila: Just for clarification: are we talking about service components or elastic-agent service installation?

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
michalpristas: fake empty input is used during unenroll, it's not a new concept, it is not persisted.
we also don't use reexec callback, we pass nil
endpoint might be a problem but it requires root so if you're running endpoint this action is not even allowed and reinstalling endpoint is still a valid scenario for some cases

internal/pkg/agent/install/componentvalidation/validation.go
michalpristas: i'm up for better naming, first only `EnsureNoServiceComponentIssues` was exported so service felt good.
the thing is it needs to be separate package due to cycles

michalpristas: does `componentvalidation` feel more descriptive?

internal/pkg/agent/cmd/run_darwin.go
pchila: Could you add a comment explaining that in the code ?

michalpristas:
added testing steps to description

ycombinator:
> Test steps:
>  ...
> check file system permissions are correct

```
$ find /opt/Elastic/Agent -type f -user root | wc -l
0
$ find /opt/Elastic/Agent -type f -group root | wc -l
0
$ find /opt/Elastic/Agent -type f -group elastic-agent | wc -l
364
$ find /opt/Elastic/Agent -type f -user elastic-agent-user | wc -l
364
```

> check process is using proper user/group

```
$ ps -axfo user:20,group:20,pid,comm | grep elastic-agent
elastic-agent-user   elastic-agent           1522 elastic-agent
elastic-agent-user   elastic-agent           2809  \_ agentbeat
elastic-agent-user   elastic-agent           2815  \_ agentbeat
elastic-agent-user   elastic-agent           2836  \_ agentbeat
```

> check service file is updated

I see the service file on Linux (`/etc/systemd/system/elastic-agent.service`) updated to include:
```
User=elastic-agent-user
Group=elastic-agent
```

> restart machine recheck everything

Checked after restarting and all of the above is still the same.

internal/pkg/fleetapi/action.go
```diff
@@ -644,6 +648,61 @@ func (a *ActionApp) MarshalMap() (map[string]interface{}, error) {
...
+	// UserInfo Optional user info data.
+	UserInfo *UserInfo `json,yaml:"user_info,omitempty"`
+}
+
+// UserInfo Optional user info data.
+type UserInfo struct {
+	// Groupname Custom group used to access Elastic Agent files.
+	Groupname string `json,yaml:"groupname,omitempty"`
+
+	// Password Password for user specified by username.
```
ycombinator:
```suggestion
	// Password for user specified by username.
```

internal/pkg/fleetapi/action.go
```diff
@@ -644,6 +648,61 @@ func (a *ActionApp) MarshalMap() (map[string]interface{}, error) {
...
+	Password string `json,yaml:"password,omitempty"`
+
+	// Username Username of custom user used to run Elastic Agent.
```
ycombinator:
```suggestion
	// Username of custom user used to run Elastic Agent.
```

internal/pkg/fleetapi/action.go
```diff
@@ -644,6 +648,61 @@ func (a *ActionApp) MarshalMap() (map[string]interface{}, error) {
...
+	return a.ActionType
+}
+
+func (a *ActionPrivilegeLevelChange) String() string {
+	var s strings.Builder
+	s.WriteString("id: ")
+	s.WriteString(a.ActionID)
+	s.WriteString(", type: ")
+	s.WriteString(a.ActionType)
+	return s.String()
```
ycombinator, line 675: Would it be useful to include `a.Data.String()` (after defining that method) as well?

internal/pkg/agent/install/switch.go
```diff
@@ -10,18 +10,22 @@ import (
...
 
 	"github.com/kardianos/service"
-	"github.com/schollz/progressbar/v3"
 
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/paths"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/perms"
 	"github.com/elastic/elastic-agent/pkg/utils"
 )
 
+// Describer is used to describe execution progress
```
ycombinator: This is a far too generic name, IMO.  Could we call it `ProgressDescriber`?

changelog/fragments/1759318728-support-for-privilege_level_change.yaml
```diff
@@ -0,0 +1,45 @@
...
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# REQUIRED for all kinds
+# Change summary; a 80ish characters long description of the change.
+summary: support for privilege_level_change
```
ycombinator: More end-user friendly?
```suggestion
summary: Add support for downgrading a running Agent's privileges from Fleet
```

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+
+	unenrollCtx, cancel := context.WithTimeout(ctx, unenrollTimeout)
+	defer cancel()
+
+	unenrollPolicy.WaitAck(unenrollCtx)
```
ycombinator: This pattern is seen in `handler_action_unenroll.go` as well.  Perhaps we should abstract it out to a helper function in `handler_helpers.go`, something like `removeComponents` or `stopProcessing`?

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+	topPath := paths.Top()
+	_, err = install.SwitchServiceUser(topPath, &debugDescriber{h.log}, username, groupname, password)
+	if err != nil {
+		// error already adds context
+		return err
+	}
+
+	// ack
+	if err := acker.Ack(ctx, a); err != nil {
+		h.log.Errorw("failed to ACK an action",
```
ycombinator, line 101: Would it be helpful to include `a.String()` in the log message here (and below for the commit)?

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+			"error.message", err)
```
ycombinator: Would it be useful to log the action as well here?
```suggestion
		h.log.Errorw("failed to ACK an action",
			"error.message", err,
			"action", a)
```

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if err := acker.Commit(ctx); err != nil {
+		h.log.Errorw("failed to commit ACK of an action",
+			"error.message", err)
```
ycombinator: Would it be useful to log the action as well here?
```suggestion
		h.log.Errorw("failed to commit ACK of an action",
			"error.message", err,
			"action", a)
```

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		// error already adds context
+		return err
+	}
+
+	// ack
+	if err := acker.Ack(ctx, a); err != nil {
+		h.log.Errorw("failed to ACK an action",
+			"error.message", err)
+	}
```
ycombinator, line 97: Should we be returning here so we don't try to commit the failed ack below?

internal/pkg/agent/cmd/run.go
```diff
@@ -206,6 +207,37 @@ func runElasticAgent(
 		return logReturn(l, fmt.Errorf("failed to check for root/Administrator privileges: %w", err))
 	}
 
+	// fix permissions
```
ycombinator, line 265: I'm trying to understand why this section was needed in this PR to handle the `privilege_level_change` action.  Seems like something we probably were missing anyway?

internal/pkg/fleetapi/action.go
michalpristas: it could include username and password right

internal/pkg/agent/cmd/run.go
michalpristas: this handles inproperly applied change. 
if we boot as root and we fix permissions just to be sure we have them right, we are capable of fixing it with the current permission set, so i find it better than have broken agent
compared to CLI change, when you do switch from CLI you have access to host. with fleet initiated action you don't and I find this as a nice way how to limit admins need to ssh into a machine

internal/pkg/agent/application/actions/handlers/handle_action_privilege_level_change.go
michalpristas: no i dont see this as a hard failure preventing us to boot to new permission set. 
commit should be a noop in this case unless there's something else in the queue that we rather commit before rebooting

internal/pkg/agent/cmd/run.go
```diff
@@ -261,6 +262,37 @@ func runElasticAgent(
...
+			return fmt.Errorf("failed to determine target user: %w", err)
+		}
+
+		var ownership utils.FileOwner
+		if userName != "" || groupName != "" {
+			ownership, err = install.EnsureUserAndGroup(userName, groupName, &debugDescriber{l}, true)
+			if err != nil {
+				return fmt.Errorf("failed to setup user: %w", err)
+			}
+		}
```
blakerouse, line 287: I think this should be done in the action. We need to limit the chances of errors for this action. By doing this in the action, we can ensure that any issues creating the user or group happen in the action before the restart even occurs.

internal/pkg/agent/cmd/run.go
```diff
@@ -261,6 +262,37 @@ func runElasticAgent(
 		return fmt.Errorf("failed to check for root/Administrator privileges: %w", err)
 	}
 
+	// fix permissions
+	if isRoot {
+		userName, groupName, err := getDesiredUser()
```
blakerouse: Please call this in the action just as a check. I think it would be great to know before the restart happens that this call will work and do the correct thing. This is just being defensive, which with this type of action we need to be as defensive as possible.

internal/pkg/agent/cmd/run.go
```diff
@@ -261,6 +262,37 @@ func runElasticAgent(
...
+		var ownership utils.FileOwner
+		if userName != "" || groupName != "" {
+			ownership, err = install.EnsureUserAndGroup(userName, groupName, &debugDescriber{l}, true)
+			if err != nil {
+				return fmt.Errorf("failed to setup user: %w", err)
+			}
+		}
+
+		topPath := paths.Top()
+		err = perms.FixPermissions(topPath, perms.WithOwnership(ownership))
```
blakerouse, line 290: If you do the two changes above this leaves this failing as the worse case scenario. If this fails, the agent is dead.

Because of that would it be better to just log this as an error and not hard fail? Maybe trying to run is better than hard failing.

internal/pkg/agent/cmd/run_darwin.go
blakerouse: This is hard failing on Setegid, so if they are saying it should do both. I expect that it would not error if it failed for `Setegid`.

internal/pkg/agent/cmd/run_linux.go
```diff
@@ -0,0 +1,72 @@
...
+	return username, groupname, nil
+}
+
+func dropRootPrivileges(ownership utils.FileOwner) error {
+	// change group first, setuid will drop permission to change group
+	if ownership.GID > 0 {
+		// Omitting setegid and using only setgid sets both real and effective group IDs,
+		// but explicit control over the effective group ID, which is used for access checks, is lost.
+		if err := syscall.Setegid(ownership.GID); err != nil {
+			return fmt.Errorf("failed to set eGID: %w", err)
```
blakerouse: Same comment I made in `run_darwin.go`.

internal/pkg/agent/cmd/run_windows.go
```diff
@@ -22,3 +23,7 @@ func logExternal(msg string) {
 	}
 	_ = eLog.Error(1, msg)
 }
+
+func getDesiredUser() (string, string, error) { return "", "", nil }
+
+func dropRootPrivileges(ownership utils.FileOwner) error { return nil }
```
blakerouse: What is the flow here for Windows? Because you are not calling `FixPermissions` before updating the service on Windows (unless I am reading this wrong). Doesn't that mean this will just completely fail to run on Windows.

blakerouse (changes requested):

internal/pkg/agent/cmd/run_windows.go
michalpristas: fixpermissions is part of `SwitchServiceUser` flow, drop root is a fallback for not cooperative service managers

blakerouse: Ah, okay. Thanks for the clarity.

PR #10234, opened 2025-10-01
Enable macOS unit tests for macOS 15

oakrizan:
## What does this PR do?
Created new GitHub workflow to run macOS unit tests on macOS 15 (arm64):
 - on push to main and release branches
 - for PRs

## Why is it important?

Required by the support matrix.

## Checklist

- [X] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- ~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)~

NB! On macOS 14 unit tests fail when using `RACE_DETECTOR=true TEST_COVERAGE=true`, that will be solved separately. Related issue: https://github.com/elastic/elastic-agent/issues/10572

Test run: https://github.com/elastic/elastic-agent/actions/runs/18524318026

### DIFF
--- a/.github/workflows/macos-unit-tests.yml
+++ b/.github/workflows/macos-unit-tests.yml
@@ -21,14 +21,7 @@ concurrency:
 
 jobs:
   unit-tests:
-    name: Unit tests ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - os: macos-15
-            flags: 'RACE_DETECTOR=true TEST_COVERAGE=true '
-    runs-on: ${{ matrix.os }}
+    runs-on: macos-15
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v5
@@ -48,5 +41,5 @@ jobs:
       - name: Install mage
         run: make mage
 
-      - run: |
-          ${{ matrix.flags }} mage unitTest
+      - name: mage unitTest
+        run: RACE_DETECTOR=true TEST_COVERAGE=true mage unitTest


pierrehilbert (dismissed):
I love that PR!

ebeahan (dismissed):

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
+name: macos-unit-tests
+
+on:
+  workflow_dispatch:
+  pull_request:
```
v1v: Let's disable on PRs for now, we will enable it after reviewing this works as expected for merge commits in the active branches

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+  push:
+    branches:
+      - main
+      - 8.*
```
v1v:
```suggestion
      - 8.19
```

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+  contents: read
+
+## Concurrency only allowed in the main branch.
+## So old builds running for old commits within the same Pull Request are cancelled
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  unit-tests-push:
```
v1v: Can you refactor this using the matrix exclusion instead of using the if condition? See https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ macos-14 ]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
```
v1v, line 30: why?

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: .go-version
+
+      - name: Install make
+        run: |
+          brew install make
+          export PATH="/usr/local/opt/make/libexec/gnubin:$PATH"
+        shell: bash
```
v1v: this is not a composite action then we don't need this

```suggestion
```

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Install mage
+        run: make mage
+        shell: bash
```
v1v: ditto

```suggestion
```

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Beat version
+        run: |
+          beatVersion=$(make get-version)
+          echo "BEAT_VERSION=$beatVersion" >> $GITHUB_ENV
+        shell: bash
```
v1v:
```suggestion
```

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      # On macOS 14 running unit tests with `RACE_DETECTOR=true TEST_COVERAGE=true` causes failure:
+      # `clang: error: linker command failed with exit code 1`
+      # Related issue: https://github.com/elastic/elastic-agent/issues/10572
+      - run: |
+          mage unitTest
+        shell: bash
```
v1v:
```suggestion
```

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Install mage
+        run: make mage
+        shell: bash
+
+      - name: Beat version
+        run: |
+          beatVersion=$(make get-version)
+          echo "BEAT_VERSION=$beatVersion" >> $GITHUB_ENV
```
v1v: Couple of things:

1) Why is this needed? As far as I see it no env variable it will default to something, see https://github.com/elastic/elastic-agent/blob/0e533344c831c769d4ecf845978d7074d6149d43/magefile.go#L543-L547

If it's needed, I wonder if this could be merged in the next step for clarity?

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,99 @@
...
+        shell: bash
+
+      # On macOS 14 running unit tests with `RACE_DETECTOR=true TEST_COVERAGE=true` causes failure:
+      # `clang: error: linker command failed with exit code 1`
+      # Related issue: https://github.com/elastic/elastic-agent/issues/10572
```
v1v: If it fails, what's the reason to enable it?

v1v (commented):
LGTM, I left a few comments:
- Enable for now on merge commits (no PRs - until we get a sense these steps work as expected)
- Refactor common jobs and use the native matrix support if possible. Maybe we can enable both MacOS runners always for simplicity.
- Nits checkout depth 0 and shell: bash
- `BEAT_VERSION` env variable

.github/workflows/macos-unit-tests.yml
oakrizan: Because [Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags](https://github.com/actions/checkout#:~:text=Only%20a%20single%20commit%20is%20fetched%20by%20default%2C%20for%20the%20ref/SHA%20that%20triggered%20the%20workflow.%20Set%20fetch%2Ddepth%3A%200%20to%20fetch%20all%20history%20for%20all%20branches%20and%20tags).
Some tests require full history to work with release branches, here's the example of the failure: https://github.com/elastic/elastic-agent/actions/runs/18191935376/job/51788733598#step:7:1082

v1v:
```
=== Failed
=== FAIL: pkg/testing/tools/git TestGetReleaseBranches (0.04s)
    git_test.go:22: []
    git_test.go:23: 
        	Error Trace:	/Users/runner/work/elastic-agent/elastic-agent/pkg/testing/tools/git/git_test.go:23
        	Error:      	Should NOT be empty, but was []
        	Test:       	TestGetReleaseBranches
```

IIUC, that's certainly I'd say it shall be skipped for running the UTs for MacOS.

Regardless, can you add a comment explaining that `pkg/testing/tools/git/git_test.go:23` requires a full history?

.github/workflows/macos-unit-tests.yml
oakrizan: Will update this, thanks! Need to obtain the bets version from the versio.go file as done in `.buildkite/scripts/common.sh`

.github/workflows/macos-unit-tests.yml
oakrizan: It fails only if `RACE_DETECTOR=true TEST_COVERAGE=true`  are set. Running tests without them works as expected.
The issue to adjust the tests to run mage target with those parameters is created.

My assumption was is better to run tests rather than no to run them at all.

.github/workflows/macos-unit-tests.yml
oakrizan: You are right it's not needed.

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  unit-tests:
+    name: Unit tests ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-14
+            flags: ''
```
v1v: Please add a link with the reference to the GH issue, so it's clear why this flag is unset, and we can track the work after it's fixed.

OTOH, do you know if the BK Orka for EA actually runs the UTs with those flags? Just wonder if the issue is in the GH runners or also in Orka

.github/workflows/macos-unit-tests.yml
v1v: I meant if you can a comment in the Github workflow explaining that pkg/testing/tools/git/git_test.go:23 requires a full history, no the other way around, sorry for the misleading

ycombinator:
The `ld` failure we're seeing looks like https://github.com/golang/go/issues/67854.

ycombinator:
> The `ld` failure we're seeing looks like [golang/go#67854](https://github.com/golang/go/issues/67854).

Testing the workaround mentioned in that issue.

ycombinator:
> > The `ld` failure we're seeing looks like [golang/go#67854](https://github.com/golang/go/issues/67854).
> 
> Testing the workaround mentioned in that issue.

Apologies for all the commits as I was trying out various things related to the workaround.  I finally got it working and have cleaned things up and squashed all my commits into 1f9987a63b to keep the history clean. 

I'm seeing a lot of warnings from `ld` in the MacOS 15 unit tests that look like this:

```
ld: warning: '/private/var/folders/q0/wmf37v850txck86cpnvwm_zw0000gn/T/go-link-4029072216/000019.o' has malformed LC_DYSYMTAB, expected 98 undefined symbols to start at index 1626, found 95 undefined symbols starting at index 1626
```

Looks like [this problem will be fixed in Go 1.26](https://github.com/golang/go/issues/61229#issuecomment-1988965927) so we should have nothing to worry about.

.github/workflows/macos-unit-tests.yml
v1v: [`1f9987a` (#10234)](https://github.com/elastic/elastic-agent/pull/10234/commits/1f9987a63b3c81bff9b85b851d111ecc84d304a9) has been merged here, shall we enable back the env variables?

oakrizan: Yes, the existing macos15 step runs unit tests with those flags - there's a single unit test script executed for all unix-based OS versions.

oakrizan:
>[1f9987a (#10234)](https://github.com/elastic/elastic-agent/pull/10234/commits/1f9987a63b3c81bff9b85b851d111ecc84d304a9) has been merged here, shall we enable back the env variables?

Unit tests still fail on macOS 14 when running them with RACE_DETECTOR and TEST_COVERAGE: https://github.com/elastic/elastic-agent/actions/runs/18589240641/job/53000225318?pr=10234

oakrizan:
I have reverted back arguments for macOS 14: https://github.com/elastic/elastic-agent/actions/runs/18592315002

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  unit-tests:
+    name: Unit tests ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-14
```
cmacknz: I don't see value in testing on 14 if we have 15 in unit tests, so we could just not test this here.

Covering more versions is more valuable in integration tests (and in that case, probably not every version in PRs).

v1v: We have detected some issues when running `macos-14`:
- https://github.com/elastic/elastic-agent/issues/10572

I don't know whether those are important or unrelated. 

If `macos-14` should not be triggered regardless, I'll update this PR accordingly, let me know what to do here.

cmacknz: https://github.com/elastic/elastic-agent/pull/10234#issuecomment-3413562642 suggests this is a Go problem not an agent problem.

I would remove MacOS 14 here and let's just test the latest version.

.github/workflows/macos-unit-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+            # On macOS 14 running unit tests with `RACE_DETECTOR=true TEST_COVERAGE=true` causes failure:
+            # `clang: error: linker command failed with exit code 1`
+            # Related issue: https://github.com/elastic/elastic-agent/issues/10572
+            flags: ''
```
v1v:
```suggestion
```

.github/workflows/macos-unit-tests.yml
v1v: [`1c9e243` (#10234)](https://github.com/elastic/elastic-agent/pull/10234/commits/1c9e243a76baa09c7b37d8263f703cf4f696521b)

ycombinator (approved):

Issue #10206, opened 2025-09-29
[beats receivers] Add ability to specify fixed port for the healthcheck extension and Prometheus metrics


cmacknz:
https://github.com/elastic/elastic-agent/blob/cb9c826c8e66e75abdf50cbee635c61538b79f23/internal/pkg/otel/manager/execution_subprocess.go#L90-L93

We find a random TCP port for the health check extension, this is and should continue to be the default behavior.

We should allow specifying a fixed port however for cases where users may need to whitelist the new port in local iptables or firewall rules.

Probably this belongs in the `agent.monitoring` section alongside the same port specification used for scrape prometheus metrics from collector internal telemetry that is being introduced in. https://github.com/elastic/elastic-agent/pull/9928

blakerouse:
We should strive for this to really be a unix socket or a named pipe, removing the whole class of issues related to it being a local port.

Default really should be the socket or named pipe with the option with it being a fixed local port.

cmacknz:
Agree, I remember discussing this but can't remember why it was localhost, perhaps another case of the collector not supporting it yet.

pkoutsovasilis:
@blakerouse @cmacknz this exact point came up again in [#8248 (discussion)](https://github.com/elastic/elastic-agent/pull/8248#discussion_r2130229714). Given the current limitation, I believe the following is a good-for-now plan:
* Keep TCP on 127.0.0.1 with random port as the default.
* [captured by this issue] Add a fixed-port override for environments that need firewall/iptables whitelisting.
  -  @cmacknz what the path under `agent.monitoring`? I don't see any new sections there in https://github.com/elastic/elastic-agent/pull/9928

Of course, if upstream adds UDS/named pipe support in the future, we can revisit and flip the implementation.

cmacknz:
> @cmacknz what the path under agent.monitoring? I don't see any new sections there

Right we haven't specified this and https://github.com/elastic/elastic-agent/pull/10240 will have the same need. CC @swiatekm 

1. We introduce a new `agent.collector` settings section and put the healthcheck port and the internal telemetry port. This is maybe leaking implementation details out with the naming but it's the most straight forward and gives us another place to stick collector settings if we ever need them.
```yaml
agent.collector:
  healthcheck_port: 1234 # Defaults to a random free port
  telemetry_port: 4567 # Defaults to a random free port
```

[...]

PR #10240, opened 2025-10-01
Use a random port for otel collector monitoring endpoint

swiatekm:
## What does this PR do?

Uses a random port for the otel collector's Prometheus endpoint. Right now this is hardcoded at the default.

Normally, this port needs to be known during config generation (so we can add the right self-monitoring configuration), so it needs to be determined before the coordinator starts. In this PR, we instead put it into an environment variable and use the otel collector's ability to load configuration values from the variable. As a result, we can defer determining the actual port as late as possible, and even use a different port on each configuration reload, allowing us to recover from port binding conflicts.

This becomes a bit awkward with the embedded collector, where we need to call `SetEnv`, always an anti-pattern. But we eventually expect to retire that mode of execution, and it's not the default anymore, so it should be fine.

In the process, I'm also allowing both the metrics port and the healthcheck port to be passed into the otel manager as parameters. This is preparation for making them configurable in a follow-up.

## Why is it important?

The port shouldn't be hardcoded. In the event of a conflict, the otel collector can't start.

## Checklist

Build the agent locally, run it with otel self-monitoring, generate diagnostics, then check the configuration.

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/6175


internal/pkg/agent/application/monitoring/component/v1_monitor.go
```diff
@@ -115,16 +112,23 @@ type monitoringConfig struct {
...
 			C: cfg,
 		},
-		otelConfig:              otelCfg,
-		operatingSystem:         operatingSystem,
-		agentInfo:               agentInfo,
-		isOtelRuntimeSubprocess: isOtelRuntimeSubprocess,
+		operatingSystem:             operatingSystem,
+		agentInfo:                   agentInfo,
+		isOtelRuntimeSubprocess:     isOtelRuntimeSubprocess,
+		otelCollectorMonitoringPort: otelCollectorMonitoringPort,
```
cmacknz: We are going to want the ability to specific a fixed port for this just like in https://github.com/elastic/elastic-agent/issues/10206. We'll figure out what section it should be in that issue since Panos has this same problem.

swiatekm: Makes sense. Once we know where we want to put it, it should be easy to add with the way this PR is currently structured.

Worth noting that this random port selection is inherently racy. Not sure how much we can do about some other process claiming the port after we've found it but before the otel collector actually binds to it. I suppose we could detect the failure and reroll, but it's going to be unpleasantly complex.

cmacknz: The healthcheck and the monitoring selection both picking the same random port would be unfortunate.

swiatekm: That we can actually guard against by selecting both ports in the same place. There isn't any particular reason to choose a different extension port every collector restart, like we do right now.

internal/pkg/agent/application/application.go
```diff
@@ -126,14 +128,25 @@ func New(
 
 	otelExecMode := otelconfig.GetExecutionModeFromConfig(log, rawConfig)
 	isOtelExecModeSubprocess := otelExecMode == otelmanager.SubprocessExecutionMode
+	otelCollectorMetricsPort, err := utils.FindRandomTCPPort()
```
pkoutsovasilis: Why do we need to find a random port here, and only once? Since we already plan to support this, it seems simpler to derive the port from config (defaulting to `0` for this PR as well as when the user doesn't specify a port in the config), letting the manager allocate a random port when it’s `0`. This also reduces the “racy” window you mention [here](https://github.com/elastic/elastic-agent/pull/10240/files#r2395462741).  

The main reason to re-pick a random port on every restart is exactly what you noted earlier: we can’t guarantee the initially chosen port remains free. So, again if this is not set by the user in the config, the best-effort is to find a free random port and spawn the collector immediately. WDYT?

swiatekm: We need to know what the port is ahead of time so we can include it in the self-monitoring config.

pkoutsovasilis: we already mutate the config at the manager [here](https://github.com/elastic/elastic-agent/blob/b1a7991cfd641b2ba5d108f18dea19dcaba5dacd/internal/pkg/otel/manager/manager.go#L375) can't we add the port here, instead of pre-defining it in the monitor?

swiatekm: We technically could, but it would either require us to redo how monitoring config injection works (right now it operates on the component model), or do an awful hack where we introspect the config in the manager and replace a placeholder port with the real one. And even if we do all this, we'd just shorten the window for the race, not actually eliminate it.

I think it's actually more reliable to keep the port the same between restarts. If we pick a new one every time, that's a chance of conflict on each restart, whereas if we keep it the same, it's less likely another application will steal it, since we've already been holding it for a while.

leehinman:
I'm probably paranoid from the npipe permission issue, but do we have a test that starts and stops the collector multiple times in rapid succession?  Since we are re-using the same port number it might be worth double checking.

swiatekm:
I'm going to try an approach using an environment variable, so I can inject this much later in the pipeline. Will undraft once I know if that's feasible.

internal/pkg/agent/application/application.go
swiatekm: Went with an approach using an environment variable, where the port is added by the execution before collector startup, same as the healthcheck port.

internal/pkg/otel/manager/common.go
```diff
@@ -52,21 +52,30 @@ func reportCollectorStatus(ctx context.Context, statusCh chan *status.AggregateS
...
+		}
+		listeners = append(listeners, l)
 
-	port := l.Addr().(*net.TCPAddr).Port
-	err = l.Close()
-	if err != nil {
-		return 0, err
+		port := l.Addr().(*net.TCPAddr).Port
+		if port == 0 {
+			return nil, fmt.Errorf("failed to find random port")
```
pkoutsovasilis, line 76: you need to close any already listening listeners in the case of error

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -212,6 +221,41 @@ func (r *subprocessExecution) startCollector(ctx context.Context, logger *logger
...
+	portIndex := 0
+	if healthCheckPort == 0 {
+		healthCheckPort = ports[portIndex]
+		portIndex++
+	}
+	if metricsPort == 0 {
+		metricsPort = ports[portIndex]
+	}
+	return healthCheckPort, metricsPort, nil
+}
```
pkoutsovasilis, line 255: not blocking; this feels too iffy, it would be way simpler to change the `findRandomTCPPorts` to be something like the following

```
func findRandomTCPPort(ports ...*int) (err error) {
	count := len(ports)
	if len(ports) == 0 {
		return nil
	}

	listeners := make([]net.Listener, 0, count)
	defer func() {
		for _, listener := range listeners {
			if closeErr := listener.Close(); closeErr != nil {
				err = errors.Join(err, fmt.Errorf("error closing listener: %w", closeErr))
			}
		}
	}()

	for _, port := range ports {
		if port == nil {
			continue
		}

		l, err := netListen("tcp", "localhost:0")
		if err != nil {
			return err
		}
		listeners = append(listeners, l)

		*port = l.Addr().(*net.TCPAddr).Port
		if *port == 0 {
			return fmt.Errorf("failed to find random port")
		}
	}

	return nil
}
```

swiatekm: I don't like putting this logic in `findRandomTcpPorts`, but using a slice of pointers in `getCollectorPorts` would make it cleaner.

swiatekm: How about https://github.com/elastic/elastic-agent/pull/10240/commits/ff5b1c23013ba8c9dc993670ae6097970fcd2749?

internal/pkg/otel/manager/common.go
```diff
@@ -52,21 +53,30 @@ func reportCollectorStatus(ctx context.Context, statusCh chan *status.AggregateS
...
-		return 0, err
-	}
+// findRandomTCPPort finds count random available TCP ports on the localhost interface.
+func findRandomTCPPorts(count int) (ports []int, err error) {
+	ports = make([]int, 0, count)
+	listeners := make([]net.Listener, 0, count)
+	defer func() {
+		for _, listener := range listeners {
+			if closeErr := listener.Close(); closeErr != nil {
+				err = errors.Join(err, fmt.Errorf("error closing listener: %w", closeErr))
```
cmacknz, line 63: Does this need to be a fatal error? It is right now, granted we should retry on spurious ones.

internal/pkg/otel/manager/common_test.go
```diff
@@ -18,9 +18,13 @@ import (
...
 
 func TestFindRandomPort(t *testing.T) {
-	port, err := findRandomTCPPort()
+	portCount := 2
+	ports, err := findRandomTCPPorts(portCount)
 	require.NoError(t, err)
-	require.NotEqual(t, 0, port)
+	require.Len(t, ports, portCount)
+	for _, port := range ports {
+		require.NotEqual(t, 0, port)
```
cmacknz: You should also assert the two ports aren't the same. They shouldn't be, you are holding the listeners open while allocating the following ports, but it's an important part of the implementation.

internal/pkg/otel/manager/manager_test.go
```diff
@@ -816,6 +819,125 @@ func TestOTelManager_Logging(t *testing.T) {
...
+			})
+
+			// the execution mode passed here is overridden below so it is irrelevant
+			m, err := NewOTelManager(
+				l,
+				logp.DebugLevel,
+				base, EmbeddedExecutionMode,
+				nil,
+				metricsPort,
+				healthCheckPort,
```
cmacknz, line 876: Is there a way to have this test start with a port conflict and then automatically recover? That is an important part of this implementation that I'd like to see verified as working somewhere.

internal/pkg/otel/manager/common.go
swiatekm: It's unclear, but it's safer to do so in my opinion. I don't think it's possible for the port to remain bound in any of the circumstances that would lead to an error when closing the socket, but in principle it's possible.

internal/pkg/otel/manager/manager_test.go
swiatekm: I can write a separate test where I bind to a port, then supply my own implementation of net.Listen that returns that port on the first call and a random port on the second.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
```diff
@@ -519,34 +514,7 @@ func (b *BeatsMonitor) monitoringNamespace() string {
...
-						return host + ":" + strconv.Itoa(port)
-					}
-				}
-			}
-		}
-	}
-
-	// If there is no explicit configuration, the collector publishes its telemetry on port 8888.
-	return "localhost:8888"
+	return fmt.Sprintf("localhost:${env:%s}", OtelCollectorMetricsPortEnvVarName)
```
ycombinator, line 524: In the previous implementation (fixed port), this function was guaranteed to return a port number, defaulting to 8888.  In the new implementation (env var), how would the collector behave if the `EDOT_COLLECTOR_METRICS_PORT` env var was not set?  This function would return `localhost:${env:EDOT_COLLECTOR_METRICS_PORT}` but what if that env var is never set — would this cause any problems when the collector tries to start up?

internal/pkg/otel/manager/common.go
```diff
@@ -52,21 +53,30 @@ func reportCollectorStatus(ctx context.Context, statusCh chan *status.AggregateS
 	}
 }
 
-// findRandomTCPPort finds a random available TCP port on the localhost interface.
-func findRandomTCPPort() (int, error) {
-	l, err := netListen("tcp", "localhost:0")
-	if err != nil {
-		return 0, err
-	}
+// findRandomTCPPort finds count random available TCP ports on the localhost interface.
```
ycombinator:
```suggestion
// findRandomTCPPorts finds count random available TCP ports on the localhost interface.
```

internal/pkg/agent/application/monitoring/component/v1_monitor.go
swiatekm: It wouldn't successfully start up. But that would be a bug in the otel manager. Our current contract is that it has to be set, so it's better if it fails if it's not, rather than continue running with a default.

internal/pkg/otel/manager/execution_embedded.go
```diff
@@ -22,11 +26,12 @@ import (
 	"github.com/elastic/elastic-agent/pkg/core/logger"
 )
 
-func newExecutionEmbedded() *embeddedExecution {
-	return &embeddedExecution{}
+func newExecutionEmbedded(metricsPort int) *embeddedExecution {
```
ycombinator, line 31: Might be good to mention in the godoc comment here that `metricsPort` = 0 means a random port will be used?

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -212,6 +221,35 @@ func (r *subprocessExecution) startCollector(ctx context.Context, logger *logger
 	return ctl, nil
 }
 
+func (r *subprocessExecution) getCollectorPorts() (healthCheckPort int, metricsPort int, err error) {
```
ycombinator, line 224: Nit: would this be a simpler / easier to follow implementation?


```suggestion
func (r *subprocessExecution) getCollectorPorts() (healthCheckPort int, metricsPort int, err error) {
	if r.collectorHealthCheckPort > 0 {
		healthCheckPort = r.collectorHealthCheckPort
	} else {
		ports, findPortsErr := findRandomTCPPorts(1)
		if findPortsErr != nil {
			return 0, 0, err
		}
		healthCheckPort = ports[0]
	}

	if r.collectorMetricsPort > 0 {
		metricsPort = r.collectorMetricsPort
	} else {
		ports, findPortsErr := findRandomTCPPorts(1)
		if findPortsErr != nil {
			return 0, 0, err
		}
		metricsPort = ports[0]
	}

	return healthCheckPort, metricsPort, nil
}
```

It does require calling `findRandomTCPPorts` twice.  Maybe you were trying to avoid that for some reason?

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -212,6 +221,35 @@ func (r *subprocessExecution) startCollector(ctx context.Context, logger *logger
...
+	// we need at least one random port, create it
+	ports, err := findRandomTCPPorts(len(randomPorts))
+	if err != nil {
+		return 0, 0, err
+	}
+	for i, port := range ports {
+		*randomPorts[i] = port
+	}
+	return healthCheckPort, metricsPort, nil
+}
```
ycombinator, line 255: Nit: would this be a simpler / easier to follow implementation?


```suggestion
func (r *subprocessExecution) getCollectorPorts() (healthCheckPort int, metricsPort int, err error) {
	if r.collectorHealthCheckPort > 0 {
		healthCheckPort = r.collectorHealthCheckPort
	} else {
		ports, findPortsErr := findRandomTCPPorts(1)
		if findPortsErr != nil {
			return 0, 0, err
		}
		healthCheckPort = ports[0]
	}

	if r.collectorMetricsPort > 0 {
		metricsPort = r.collectorMetricsPort
	} else {
		ports, findPortsErr := findRandomTCPPorts(1)
		if findPortsErr != nil {
			return 0, 0, err
		}
		metricsPort = ports[0]
	}

	return healthCheckPort, metricsPort, nil
}
```

It does require calling `findRandomTCPPorts` twice.  Maybe you were trying to avoid that for some reason?

internal/pkg/agent/application/monitoring/component/v1_monitor.go
ycombinator: Thanks.  Is this expectation (contract) mentioned somewhere, like in an inline comment or something?

internal/pkg/otel/manager/execution_subprocess.go
swiatekm: Yes, the reason is that we can be sure not to have a conflict between the two. The logic unfortunately has four different cases that all need to fit under one `findRandomTCPPorts` call. I'll be very happy to change it if you can show me a clearer implementation.

internal/pkg/agent/application/monitoring/component/v1_monitor.go
swiatekm: I thought I added one, but apparently not. I'll add a short comment here and a longer one where the constant is defined.

swiatekm: Do the comments added [here](https://github.com/elastic/elastic-agent/pull/10240/commits/f53b8e8988a7ab1cf9cbae72742db368224dfff1) make things clearer?

internal/pkg/otel/manager/manager_test.go
swiatekm: I added this test, but its setup is pretty convoluted. It essentially causes `net.Listen` to return a listener which ignores calls to `Close()`, keeping the ports bound until the test finishes. But it does work and shows us restarting the collector once before it gets to a healthy state.

internal/pkg/otel/manager/execution_subprocess.go
cmacknz: Yes calling `findRandomTCPPorts` twice with it's current implementation introduces the low but not impossible risk it allocates the same port twice. 

In the current implementation, it holds the listener bound to each successively allocated port until it has allocated the requested amount which prevents this collision from happening

internal/pkg/otel/manager/manager_test.go
cmacknz: Interestingly, when I ran this test locally to double check my understanding of it, it failed

```
--- FAIL: TestOTelManager_PortConflict (10.03s)
    manager_test.go:1022: 
        	Error Trace:	/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:1025
        	            				/opt/homebrew/Cellar/go/1.25.0/libexec/src/runtime/asm_arm64.s:1268
        	Error:      	Expected value not to be nil.
        	Messages:   	collector status should not be nil
    manager_test.go:1022: 
        	Error Trace:	/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/internal/pkg/otel/manager/manager_test.go:1022
        	Test:       	TestOTelManager_PortConflict
```

Let's see what CI thinks, otherwise the test LGTM

swiatekm: Hm, hadn't had it fail locally, and it succeeded in CI as well. Maybe the time limit is too small?

ycombinator (dismissed):
LGTM pending resolution of https://github.com/elastic/elastic-agent/pull/10240#discussion_r2408252737

internal/pkg/otel/manager/manager_test.go
swiatekm: @cmacknz are you sure you ran the version on main? It should show otel logs if the test fails.

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -36,7 +38,9 @@ const (
 	OtelSupervisedMonitoringURLFlagName = "supervised.monitoring.url"
 )
 
-func newSubprocessExecution(logLevel logp.Level, collectorPath string) (*subprocessExecution, error) {
+// newExecutionSubProcess creates a new execution which runs the otel collector in a subprocess. A metricsPort or
```
blakerouse:
```suggestion
// newSubprocessExecution creates a new execution which runs the otel collector in a subprocess. A metricsPort or
```

blakerouse (approved):
Looks good, glad the environment variable substitution worked out.

pkoutsovasilis (dismissed):

internal/pkg/otel/manager/manager_test.go
cmacknz: Yep it is failing for possibly unrelated reasons I think it can't start the diagnostics extension `listen unix /tmp/elastic-agent/VQh97r-KzxDzS6OwRBDAAK4aSN8DXg3r.sock: bind: permission denied]`

```
    manager_test.go:987: {Entry:{Level:error Time:2025-10-07 13:26:37.591 -0400 EDT LoggerName: Message:Failed to start extension Caller:undefined Stack:} Context:[{Key:log.origin Type:23 Integer:0 String: Interface:map[file.line:58 file.name:extensions/extensions.go function:go.opentelemetry.io/collector/service/extensions.(*Extensions).Start]} {Key:resource Type:23 Integer:0 String: Interface:map[service.instance.id:32691dc5-46ee-4dcb-85ab-e5fd2cce348c service.name:/Users/cmackenzie/go/src/github.com/elastic/elastic-agent/internal/pkg/otel/manager/testing/testing service.version:9.3.0]} {Key:otelcol.component.id Type:15 Integer:0 String:elastic_diagnostics Interface:<nil>} {Key:otelcol.component.kind Type:15 Integer:0 String:extension Interface:<nil>} {Key:error Type:23 Integer:0 String: Interface:map[message:error creating listener: listen unix /tmp/elastic-agent/VQh97r-KzxDzS6OwRBDAAK4aSN8DXg3r.sock: bind: permission denied]} {Key:ecs.version Type:15 Integer:0 String:1.6.0 Interface:<nil>}]}
```

cmacknz: My `/tmp/elastic-agent` directory exists because I'm running an installed agent that uses beats receivers as root and the test isn't root so can't create a socket in that directory.

Running the test as root it passes I will file a separate bug for this.

cmacknz (approved):

ebeahan:
@swiatekm there's a question about backporting this to `8.19`: https://github.com/elastic/elastic-agent/pull/10449#issuecomment-3389448195. Please double check what branches this change should land in.

PR #10259, opened 2025-10-01
[OTel] Fix CLI example to use proper syntax

pickypg:
This uses the expected `::` syntax instead of `.` for the `--set` argument.

## What does this PR do?

Fixes the example text for showing users how to use the `--set` argument.

## Why is it important?

The documentation shows the correct format.

### DIFF
--- a/internal/pkg/agent/cmd/otel_flags.go
+++ b/internal/pkg/agent/cmd/otel_flags.go
@@ -27,7 +27,7 @@ func setupOtelFlags(flags *pflag.FlagSet) {
 		" single location can be set per flag entry e.g. `--config=file:/path/to/first --config=file:path/to/second`.")
 
 	flags.StringArray(otelSetFlagName, []string{}, "Set arbitrary component config property. The component has to be defined in the config file and the flag"+
-		" has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s")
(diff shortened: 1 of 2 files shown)


internal/pkg/agent/cmd/otel_flags.go
```diff
@@ -27,7 +27,7 @@ func setupOtelFlags(flags *pflag.FlagSet) {
 		" single location can be set per flag entry e.g. `--config=file:/path/to/first --config=file:path/to/second`.")
 
 	flags.StringArray(otelSetFlagName, []string{}, "Set arbitrary component config property. The component has to be defined in the config file and the flag"+
-		" has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s")
+		" has a higher precedence. Array config properties are overridden and maps are joined. Example --set \"processors::batch::timeout=2s\"")
```
pickypg, line 30: Down below, on line 96, we will actually replace `.` with `::`, which seems wrong given that we cannot use YAML syntax _and_ that prevents using `.` in field names.

Because we are also doing the helper behavior, we _prevent_ allowing `yaml:key: value` syntax because we explicitly require `=` to be present for the `key=value` assignment.

internal/pkg/agent/cmd/otel_flags.go
pickypg, line 96: This is where we replace `.`, probably improperly.

pickypg, line 86: This is where we require `=`.

Issue #10267, opened 2025-10-01
[OTel] `--set` should merge with existing config


pickypg:
I added a few tests to exemplify this scenario:

https://github.com/elastic/elastic-agent/pull/10259/commits/c652afc0ed01da66e3c916ca1f721ed03520da37

But, imagine that you have a configuration that looks like:

https://github.com/elastic/elastic-agent/blob/278440a1d520eeeffe713b730416e0487fc9f64a/internal/pkg/otel/samples/linux/autoops_es.yml#L1-L23

Ideally, we should be able to add `ssl.verification_mode` to "both" modules by a simply:

```
--set "receivers::metricbeatreceiver::metricbeat::modules=[{ ssl: { verification_mode: none } }, { ssl: { verification_mode: none } }]"
```

This is parsed, but it ultimately fails because it overwrites the existing `modules`. Separately, it seems like we should be able to manipulate the array directly:

```
--set "receivers::metricbeatreceiver::metricbeat::modules::0::ssl::verification_mode=none" \
--set "receivers::metricbeatreceiver::metricbeat::modules::1::ssl::verification_mode=none"
```

As a result, the only way to add or change settings within that array is to provide all settings at the same time:

```
--set 'receivers::metricbeatreceiver::metricbeat::modules=[{module: autoops_es, hosts: "${env:AUTOOPS_ES_URL}", period: 10s, metricsets: [cat_shards, cluster_health, cluster_settings, license, node_stats, tasks_management], ssl: { verification_mode: none } }, {module: autoops_es, hosts: "${env:AUTOOPS_ES_URL}", period: 24h, metricsets: [cat_template, component_template, index_template], ssl: { verification_mode: none } }]'
```

The way that `--set` is described, it seems like merging is the desired behavior rather than replacement. And, the way that I would also like to use it from the outside is to append. If I want to overwrite, I can change the relevant key(s).

cmacknz:
I think this is probably an upstream collector problem, @VihasMakwana recently had an upstream RFC accepted to properly support merging list like you are describing and he's probably the best person to comment on the state of this.

- https://github.com/open-telemetry/opentelemetry-collector/pull/13256

VihasMakwana:
@pickypg I have not personally tested my approach with the `--set` flag, but looking at the code of config merging, it should work.

VihasMakwana:
here's the implementation PR: https://github.com/open-telemetry/opentelemetry-collector/pull/13551

[...]

PR #10268, opened 2025-10-01
[AutoOps] Provide Sample Configuration to Debug Data

pickypg:
This provides a separate OTel configuration to allow users to better understand what data they are shipping with the AutoOps ES module.

## What does this PR do?

This adds a new OTel configuration available for every OS that helps users to understand what data the AutoOps ES module exports.

## Why is it important?

Many users have asked what data is collected by AutoOps and this helps them to explore it locally before shipping it anywhree.

## Checklist

## Disruptive User Impact

None. This adds a new, optional configuration that the user can run to better understand AutoOps.

Run the Elastic Agent against an Elasticsearch cluster.

```bash
./elastic-agent --config otel_samples/autoops_es_debug.yml
```

You will want to supply these environment variables (with values filled in relevant to your environment):

```bash
# Values are examples and have no effect other than to appear in output unchanged
AUTOOPS_TEMP_RESOURCE_ID=example-id
AUTOOPS_TOKEN=example_token

# Values must work:
AUTOOPS_ES_URL=https://localhost:9200

# Username / Password _OR_ API Key
AUTOOPS_ES_USERNAME=my-username
AUTOOPS_ES_PASSWORD=my-secure-password

AUTOOPS_ES_API_KEY=myapikey==
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -23,6 +23,6 @@
 /internal/pkg/otel/samples @elastic/ingest-otel-data @elastic/ingest-docs
 /internal/pkg/otel/core-components.yaml @elastic/ingest-otel-leads
 /internal/pkg/composable/providers/kubernetes @elastic/elastic-agent-control-plane
-/internal/pkg/otel/samples/darwin/autoops_es.yml @elastic/opex
-/internal/pkg/otel/samples/linux/autoops_es.yml @elastic/opex
-/internal/pkg/otel/samples/windows/autoops_es.yml @elastic/opex
+/internal/pkg/otel/samples/darwin/autoops_es*.yml @elastic/opex
+/internal/pkg/otel/samples/linux/autoops_es*.yml @elastic/opex
+/internal/pkg/otel/samples/windows/autoops_es*.yml @elastic/opex
--- a/changelog/fragments/1759359160-autoops-debug-output.yaml
+++ b/changelog/fragments/1759359160-autoops-debug-output.yaml
@@ -1,0 +1,45 @@
+# REQUIRED
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# REQUIRED for all kinds
+# Change summary; a 80ish characters long description of the change.
+summary: Add debug exporter to AutoOps OTel config sample
+
+# REQUIRED for breaking-change, deprecation, known-issue
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# description:
+
+# REQUIRED for breaking-change, deprecation, known-issue
+# impact:
+
+# REQUIRED for breaking-change, deprecation, known-issue
+# action:
+
+# REQUIRED for all kinds
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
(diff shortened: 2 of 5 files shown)


ycombinator (commented):
Seeing as these new pipelines with the `debug` exporters are identical to their `otlphttp` exporter counterparts except, of course, for the exporter used, I wonder if you could consolidate these into one set of pipelines that defined both exporters and used a [`routingconnector`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/routingconnector) based on an environment variable to route the data to one or the other exporter?

That would save a lot of duplication and the need for ensuring that the two sets of pipelines are kept in sync whenever one changes.

pickypg:
> the need for ensuring that the two sets of pipelines are kept in sync whenever one changes.

I see ~this~ your potential change as a major upside. However, wouldn't using a connector add a potential performance barrier? This is my own lack of knowledge, but is that a one-time check or would that be a per-telemetry event check? If it's a one-time setup cost, then I think it's the way to go. If it's a cost for every piece of data flowing through the agent, then I'd be happy to pay the overhead cost of an extra config file versus that downstream cost because large clusters will be sending a lot of data through the pipeline and the debug output is mostly for customers to feel comfortable using it, rather than a regular-use.

ycombinator:
> wouldn't using a connector add a potential performance barrier? This is my own lack of knowledge, but is that a one-time check or would that be a per-telemetry event check? 

I'm not familiar with how the `routingconnector` works either (pinging @swiatekm or @andrzej-stencel in case they have some insights) but I suspect it inspects every event since the conditions are based on OTTL.  I'm not sure if there's any optimization if the condition doesn't reference any event fields but just refers to "static" environment variables instead.  

---

If we decide to go with two sets of nearly-identical configuration files, we might want to consider introducing a `_templates` or similar folder that's not included in the Agent distribution but is used at build time to generate the two near-identical configuration files from the same template file via something like a `mage otel:generateSamples` target that's also a dependency of the `mage check` target, similar to the `mage otel:readme` target.

pickypg:
> If we decide to go with two sets of nearly-identical configuration files, we might want to consider introducing a _templates or similar folder

I like this. I have been actively thinking about how to provide a set of configuration that enables providing TLS / SSL configuration without being a total duplication. If I had different ways to piece together the receivers, I could have the exporters and pipeline provided by a different file, then allow the "user" to merge them together as needed (even better with https://github.com/elastic/elastic-agent/issues/10267).

ycombinator:
> If I had different ways to piece together the receivers, I could have the exporters and pipeline provided by a different file, then allow the "user" to merge them together as needed

I'm far from an OTel configuration expert and I'm starting to think there's probably already mechanisms or tricks to merge OTel configuration files defining different components and/or ideas for generating configuration from templates that someone has already come up with.  So I'd really like to hear from folks like @andrzej-stencel and @swiatekm who are active in the OTel community before we head down any concrete paths.

pickypg:
Happy to hear their insights, but unfortuantely I don't think "merging" really works in any functional way yet without

- https://github.com/open-telemetry/opentelemetry-collector/pull/13256
- https://github.com/open-telemetry/opentelemetry-collector/pull/13551

I tried locally with the latest Elastic Agent to use `?merge_paths` and it failed with

> invalid uri: "otel_samples/autoops_es_ssl.yml?merge_paths=receivers::metricbeatreceiver::metricbeat::modules"

andrzej-stencel:
The Routing connector approach will surely have a performance penalty. Whether it's big or small or negligible would need to be tested. I think however that using the Routing connector in this case will harm configuration readability and comprehensibility, which is the opposite that we seem to want here - the example configurations for debugging should be simple, to enhance user understanding.

I'm in favor of generating the similar configurations from a template, as mentioned [above](https://github.com/elastic/elastic-agent/pull/10268#issuecomment-3362588276).

andrzej-stencel (approved):
It's good as it is now. Would be good to prevent duplication by e.g. templating.

ycombinator (approved):
LGTM for now.  Would be good to reduce duplication via templating as discussed in this PR but that can be done via a follow up PR.

cmacknz (approved):

colleenmcginnis (approved):

Issue #10262, opened 2025-10-01
[beats receivers] EDOT diagnostics are not collected when diagnostics are requested through Fleet


cmacknz:
When using the request diagnostics action in Fleet, the edot/ diagnostics sub-directory is missing from the diagnostics. They are present when collected using `elastic-agent diagnostics` locally.

<img width="1923" height="332" alt="Image" src="https://github.com/user-attachments/assets/5909c41b-d144-4fa7-84ec-e6c91e74d237" />

We are likely missing something in the diagnostics action handler to trigger them.

https://github.com/elastic/elastic-agent/blob/18beeba11ea6812bcb799b8f0ae76e55676f67e9/internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go#L90-L100

PR #10291, opened 2025-10-03
[edot][diagnostics] - add global diagnostics to handler

VihasMakwana:
## What does this PR do?

This PR updates diagnostics handler to also fetch EDOT diagnostics. This is needed for it to work from fleet.

## Checklist

https://github.com/elastic/elastic-agent/issues/10262z
## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10262


VihasMakwana:
I'll add a test in this PR.

internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
```diff
@@ -247,6 +250,26 @@ func (h *Diagnostics) runHooks(ctx context.Context, action *fleetapi.ActionDiagn
 			Generated:   time.Now().UTC(),
 		})
 	}
+	resp, err := otel.PerformDiagnosticsExt(ctx, collectCPU)
+	if errors.Is(err, syscall.ENOENT) || errors.Is(err, syscall.ECONNREFUSED) {
```
cmacknz, line 254: We should have a better way to detect when the EDOT sub-process should be running than this, because the way this is done would cover up actual bugs if we expected EDOT to be running but we got one of these errors.

VihasMakwana: This is something that's been on my mind as well. 
So far, I think we can follow libbeat's implementation, where it detects if beats is running under agent or not. Ref: https://github.com/elastic/beats/blob/ce5b866a48f1afb86077a0664dfedd05f42f9100/libbeat/management/agent.go#L27-L35

cmacknz: We are possibly handling this at the wrong level here, because you would like need the detection to be a method on the coordinator.

However, the coordinator's PerformDiagnostics has a direct call into the OTel manager already and it knows whether it exists and how EDOT is running.

https://github.com/elastic/elastic-agent/blob/605864f454ddc2d97f550012feab8a4361453ba5/internal/pkg/agent/application/coordinator/coordinator.go#L883-L891

This appears to get called for unit diagnostics in the `elastic-agent diagnostics` command handle on the server side of the gRPC call.

https://github.com/elastic/elastic-agent/blob/605864f454ddc2d97f550012feab8a4361453ba5/pkg/control/v2/server/server.go#L317

cmacknz: The `otel.PerformDiagnosticsExt` should probably be inside the `otelMgr.PerformDiagnostics` call and then it will know if EDOT should exist.

VihasMakwana: The `otelMgr` is started regardless if we have any otel components or not. 

https://github.com/elastic/elastic-agent/blob/92d5c1daa23ee0e01f7fe68fe1f47b51ebdd79a3/internal/pkg/agent/application/application.go#L248-L252

Hence, the `otelMgr.PerformDiagnostics` is called regardless of EDOT's existence. 

https://github.com/elastic/elastic-agent/blob/92d5c1daa23ee0e01f7fe68fe1f47b51ebdd79a3/internal/pkg/agent/application/coordinator/coordinator.go#L900-L905

It simply returns empty diagnostics if no otel managed components are running.

VihasMakwana: > We should have a better way to detect when the EDOT sub-process should be running than this, because the way this is done would cover up actual bugs if we expected EDOT to be running but we got one of these errors.

In your opinion, can this work be done in a follow-up? We have few more places where we detect if extension is running on the basis of error code.

cmacknz: Yes sure put it into an issue.

The reason I suggest putting this in `otelMgr` is because it is the one manager that knows if the collector should be running or not, otherwise you'll be doing things like looking at the current runtime of every component in the component model.

VihasMakwana: opened https://github.com/elastic/elastic-agent/issues/10353 to track this.

internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
```diff
@@ -247,6 +250,26 @@ func (h *Diagnostics) runHooks(ctx context.Context, action *fleetapi.ActionDiagn
...
+		// We're not running the EDOT if:
+		//  1. Either the socket doesn't exist
+		//	2. It is refusing the connections.
+		h.log.Debugf("Couldn't fetch diagnostics from EDOT: %v", err)
+		return diags, nil
+	}
+	if err != nil {
+		return diags, err
+	}
+	for _, r := range resp.GlobalDiagnostics {
```
cmacknz, line 264: Why do we omit `ComponentDiagnostics` here?

cmacknz (commented):
I manually tested this and it works as expected with the process and otel runtime. One question about component diagnostics before approving.

internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go
VihasMakwana: @cmacknz Component diagnostics is collected via the coordinator's `PerformComponentDiagnostics` method. 

https://github.com/elastic/elastic-agent/blob/8a5a163254cda7498d20f07815a6a242e3bb4a96/internal/pkg/agent/application/actions/handlers/handler_action_diagnostics.go#L327-L331

Here's the implementation:

https://github.com/elastic/elastic-agent/blob/8a5a163254cda7498d20f07815a6a242e3bb4a96/internal/pkg/agent/application/coordinator/coordinator.go#L893-L907

As you can see, it calls the otel manager's `ComponentDiagnostics`.

PR #10343, opened 2025-10-06
feat: rework elasticsearch output translation to otel config to exclu…

pkoutsovasilis:
…de validation errors

## Checklist

## Related issues

-

## Questions to ask yourself

- ...


blakerouse (approved):
This change looks good to me. In call it was decided to just merge this with the otel bump. CI is green!

swiatekm:
> This change looks good to me. In call it was decided to just merge this with the otel bump. CI is green!

We should also bump the version in beats, though.

pkoutsovasilis:
> We should also bump the version in beats, though.

@swiatekm does this need to happen before this PR, or after? How these get bumped today in beats?

swiatekm:
> > We should also bump the version in beats, though.
> 
> @swiatekm does this need to happen before this PR, or after? How these get bumped today in beats?

You bump in opentelemetry-collector-components, then in beats, and then in agent. I think it's fine to merge this PR, but if you want to be thorough, you should also bump component versions to ones which match the otel core version. Does that make sense?

swiatekm:
The changes look good to me, but I'd really like to have a test verifying that the basic function of this PR works - that is, if we provide invalid TLS configuration to an elasticsearch output used by a beats receiver, the collector still starts and we get the right status. I think an integration test is the right way to do this, but I'd also accept a unit test for the otel manager.

pkoutsovasilis:
> You bump in opentelemetry-collector-components, then in beats, and then in agent. I think it's fine to merge this PR, but if you want to be thorough, you should also bump component versions to ones which match the otel core version. Does that make sense?

I can’t confidently own that bump chain right now. I’ll move this PR to Draft and keep it scoped to the minimal change, letting the existing version-bump mechanisms land first. Once those are in, I’ll rebase and flip back to Ready for Review.

> The changes look good to me, but I'd really like to have a test verifying that the basic function of this PR works - that is, if we provide invalid TLS configuration to an elasticsearch output used by a beats receiver, the collector still starts and we get the right status. I think an integration

Agreed. I’ll add a test for that

cc @ebeahan @cmacknz

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -524,13 +524,28 @@ func BeatDataPath(componentId string) string {
 
 // getBeatsAuthExtensionConfig sets http transport settings on beatsauth
 // currently this is only supported for elasticsearch output
-func getBeatsAuthExtensionConfig(cfg *config.C) (map[string]any, error) {
+func getBeatsAuthExtensionConfig(outputCfg *config.C) (map[string]any, error) {
 	defaultTransportSettings := elasticsearch.ESDefaultTransportSettings()
-	err := cfg.Unpack(&defaultTransportSettings)
+
```
khushijain21, line 529: there is caveat in handling `proxy_url`. I have created a PR to accommodate this. PTAL https://github.com/pkoutsovasilis/elastic-agent/pull/1

pkoutsovasilis: thanks for handling also that, it is now merged here

pkoutsovasilis: @khushijain21 after looking over some discussions on slack about some CI test failures, I realised that the code we merged from your PR highlights an issue. Specifically, if we define `proxy_url` with the value of `http://${env:ES_ENDPOINT}:9200` which is valid I think, the translation produces `http://ES_ENDPOINT:9200` which isn't correct. I believe that you should have another look on that but in the meantime I am dropping the code for `proxy_url` from this PR

khushijain21: What exactly should be the right value? Because I see [here](https://github.com/elastic/elastic-agent/blob/mergify/bp/9.1/pr-9341/elastic-agent.docker.yml#L7) that if we had `hosts: '${ELASTICSEARCH_HOSTS:http://elasticsearch:9200}'` then it is resolved as `hosts: http://elasticsearch:9200'` - similarly to `proxy_url` above. What is incorrect here?

swiatekm: I don't think that's a bug, exactly, but it's a bit confusing. It's consistent for agent to use its own configuration templating syntax in output config, rather than otel's. Maybe we should detect the latter and error, but this shouldn't block this PR.

PR #10344, opened 2025-10-06
Introduce install TTL markers

pchila:
## What does this PR do?
~~This PR introduces a registry of elastic-agent installs by maintaining a YAML structure in .installed installation marker.
As soon as a new version of agent is unpacked on disk during an upgrade, the installation is added to the registry, updated when the symlinks are rotated to such installation and removed after the installation is deleted from disk.~~
At @cmacknz 's request the install registry has been scrapped substituted by a more limited but simpler mechanism.

This PR introduces TTL marker files in versioned homes of Elastic Agent versions to keep track of available manual rollback targets and the time until those installs should remain on disk.
When upgrading, existing installs are marked with a `.ttl` YAML file, for example:
```yaml
version: 9.3.0-SNAPSHOT
hash: abcdef
valid_until: 2025-10-17T14:01:13+02:00
```

This allows to determine which available rollbacks are still on disk even after the upgrade marker is removed at the end of a successful upgrade.

## Why is it important?
This PR is a prerequisite for introducing manual rollbacks beyond grace period with PR #9643

## Checklist

## Related issues

- Relates #9643

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/upgrade_test.go
+++ b/internal/pkg/agent/application/upgrade/upgrade_test.go
@@ -1604,6 +1604,7 @@ func TestUpgradeErrorHandling(t *testing.T) {
 			checkVersionedHomeCleanup: true,
 			setupMocks: func(t *testing.T, mockAgentInfo *info.MockAgent, mockRollbackSrc *mockAvailableRollbacksSource, mockWatcherHelper *MockWatcherHelper) {
 				mockAgentInfo.EXPECT().Version().Return("9.0.0")
+				mockRollbackSrc.EXPECT().Set(map[string]TTLMarker(nil)).Return(nil)
 			},
 		},
 		"should add disk space error to the error chain if downloadArtifact fails with disk space error": {


pchila:
Linter errors are already addressed in #9643

pkoutsovasilis (approved):
LGTM, I manually tested that:
* .installed keeps track of installed versions when I upgrade
* .installed keeps track of installed version when I manually rollback

It doesn't yet clean up entries in .installed - I did manually edit the ttl and restarted elastic-agent - but this I can see from the code that this is in https://github.com/elastic/elastic-agent/pull/9643

pchila:
> LGTM, I manually tested that:
> 
>     * .installed keeps track of installed versions when I upgrade
> 
>     * .installed keeps track of installed version when I manually rollback
> 
> 
> It doesn't yet clean up entries in .installed - I did manually edit the ttl and restarted elastic-agent - but this I can see from the code that this is in #9643; @pchila keep me honest here

Correct. The [handling for the expired installs](https://github.com/elastic/elastic-agent/pull/9643/files#diff-f9f401b9d676a38a417d02f0a706e625a7f7b9214865641f5f2c4ed23a2d1f19R370-R380) is in the follow-up [PR](https://github.com/elastic/elastic-agent/pull/9643) which allows for manual rollbacks while not in the grace period.
(Manually rolling back during the grace period does not need to rely on the install registry so the cleanup is still missing at this stage)

ycombinator (commented):
Starting to review this PR now but one general comment: could you please add godoc comments for any exported types, capturing what they represent / their purpose?  Thanks.

pkg/utils/install/file_source.go
```diff
@@ -0,0 +1,123 @@
...
+		installDescriptor = v1.NewInstallDescriptor()
+	}
+
+	existingInstalls := installDescriptor.AgentInstalls
+	installDescriptor.AgentInstalls = make([]v1.AgentInstallDesc, len(existingInstalls)+1)
+	installDescriptor.AgentInstalls[0] = desc
+	copied := copy(installDescriptor.AgentInstalls[1:], existingInstalls)
+	if copied != len(existingInstalls) {
+		return nil, fmt.Errorf("error adding new install %v to existing installs %v", desc, existingInstalls)
+	}
```
ycombinator: WDYT about encapsulating this into a method on `InstallDescriptor`, something like `AddInstall(AgentInstallDesc)`?

pkg/utils/install/file_source.go
```diff
@@ -0,0 +1,123 @@
...
+	if installDescriptor == nil {
+		return nil, fmt.Errorf("no install descriptor found at %q", dp.descriptorFile)
+	}
+
+	for i := range installDescriptor.AgentInstalls {
+		err = modifierFunc(&installDescriptor.AgentInstalls[i])
+		if err != nil {
+			return nil, fmt.Errorf("modifying agent install %s: %w", installDescriptor.AgentInstalls[i].VersionedHome, err)
+		}
+	}
```
ycombinator: This makes me a bit nervous, calling `modifierFunc` on every agent install in `installDescriptor.AgentInstalls`.  

From what I can tell looking at the two (non-test) calls of the `ModifyInstallDesc` method, in both cases the intent is to modify a specific agent install based on the `versionedHome` value.  And the responsibility of choosing that specific agent install is currently on the callers end, inside their definition of the `modifierFunc` function they supply.  

Could we instead move that responsibility into the `ModifyInstallDesc` method, by having it take `agentInstall AgentInstallDesc` as an argument and then only calling `modifierFunc` if `agentInstall.VersionedHome == installDescriptor.AgentInstalls[i].VersionedHome` in this loop here?  I know it's not as generic as the current implementation but it seems safer and would satisfy the current use cases?

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -160,16 +161,21 @@ func markUpgradeProvider(updateActiveCommit updateActiveCommitFunc, writeFile wr
 			Details:           upgradeDetails,
 		}
 
-		if rollbackWindow > disableRollbackWindow && agent.parsedVersion != nil && !agent.parsedVersion.Less(*Version_9_2_0_SNAPSHOT) {
+		if agent.parsedVersion != nil && !agent.parsedVersion.Less(*Version_9_2_0_SNAPSHOT) {
```
ycombinator: I think we can leave this line as-is for this PR?  I know it will change in the follow up PR.  It would also clear the linter warning for `const disableRollbackWindow` above.

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -160,16 +161,21 @@ func markUpgradeProvider(updateActiveCommit updateActiveCommitFunc, writeFile wr
...
 			// if we have a not empty rollback window, write the prev version in the rollbacks_available field
 			// we also need to check the destination version because the manual rollback and delayed cleanup will be
 			// handled by that version of agent, so it needs to be recent enough
-			marker.RollbacksAvailable = []RollbackAvailable{
-				{
-					Version:    previousAgent.version,
-					Home:       previousAgent.versionedHome,
-					ValidUntil: updatedOn.Add(rollbackWindow),
-				},
+			for _, rollback := range availableRollbacks {
```
ycombinator: Does the inline comment above need to change now with this code change?

ycombinator (commented):
@pchila Thanks for breaking out this PR from https://github.com/elastic/elastic-agent/pull/9643 — I appreciate it as it makes each PR more focussed on a single logical change and easier to review (at least for me).  Left some minor comments but this is looking good!

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -400,6 +409,24 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, rollback bool, s
...
+	currentVersionedHome, err := filepath.Rel(paths.Top(), paths.Home())
+	if err != nil {
+		return nil, fmt.Errorf("calculating home path relative to top, home: %q top: %q : %w", paths.Home(), paths.Top(), err)
+	}
+
+	_, err = u.installDescriptorSource.AddInstallDesc(
+		v1.AgentInstallDesc{Version: version, VersionedHome: unpackRes.VersionedHome, Hash: unpackRes.Hash, Flavor: detectedFlavor, Active: false},
+	)
+	if err != nil {
+		err = fmt.Errorf("error encountered when adding install description: %w", err)
```
cmacknz: Does this need to be fatal to the entire upgrade process? If it's something like not having enough disk space it is, if it's some unexpected bug in `AddInstallDesc` purely related to rollbacks, failure should just make future rollback unavailable and not prevent upgrade entirely.

Persistent errors here that are not truly fatal basically doom us, there's no way past it. This is maybe a place where if we had a force upgrade for the agent itself we'd use it, so that we see the error when it happens but the user has a way to try to ignore it.

cmacknz: I would much rather "rollback feature unexpectedly didn't work" vs "all future upgrades blocked because of error from new feature"

cmacknz: I can see why we might consider this fatal based on the `error encountered when setting new install description as active` path below, but I am still worried that any unexpected problems in here could potentially doom users to not being able to remotely upgrade their agents anymore.

Is there any way to compartmentalize the errors to only prevent the new rollback feature from working instead of taking the entire upgrade process with  it?

pkg/utils/install/file_source.go
pchila: InstallDescriptor is a data structure defined in `api/v1` package. I'd rather not add methods to that and keep it as a plain data struct

pkg/utils/install/file_source.go
pchila: This is done as a precise design choice: the modifier function will be able to modify the install descriptor contents in a single operation (not exactly transactional), operating on multiple installations.

This is needed when "moving" the active flag from one install to another or making other modification that require touching multiple installs.

I am not sure where the nervousness is coming from, but how would you modify only one attribute  if you pass in an entire `AgentInstallDesc` ? You would substitute the entire struct ?

internal/pkg/agent/application/upgrade/step_mark.go
pchila: Fixed in the follow-up PR [sneak peek](https://github.com/elastic/elastic-agent/pull/9643/files#diff-2b13dda38618eb37f9e83e90746670287f667ce38d493673210e59abff042431R163-R173)

The whole for loop will change and the available rollbacks will be injected as a parameter to the function call

internal/pkg/agent/application/upgrade/upgrade.go
pchila: The point of the install registry is to keep track which version of the agent is located where on disk.
That's the reason why as soon as a new version is unpacked we start tracking it with this call here, then set it to active once the symlinks are rotated further down.

We can make the error non-fatal but I am not sure what chances an upgrade process will have if agent is not able to write a small YAML fragment on disk, compared to copying the agent state further into the upgrade process.

There is no way to compartmentalize the errors as the install registry will be used by the rollback feature but  it's not conditioned by activating such feature, it is always on and capturing the new agent installs.

If we want to transform the install registry into an ancillary feature of manual rollbacks we can, but that would make it pretty limited.

internal/pkg/agent/application/upgrade/step_mark.go
pchila: I am not sure I understand the reason here. The linter error will be taken care of in the follow-up PR, so I wouldn't bother changing it at this point.
The reason why this has been removed from here is to concentrate the logic for available rollbacks in `upgrade.go` rather than letting `step_mark.go` decide the contents of the upgrade marker.
It is only done halfway here is because I had to split the original PR commits at some point.

internal/pkg/agent/application/upgrade/upgrade.go
pchila: @cmacknz treated install registry operation errors as non fatal in cd18e71ec54a64fbd3a26ea69ba399c6b1a07a96

pkg/utils/install/file_source.go
ycombinator: No, the nervousness is more coming from the fact that the `modifierFunc` gets called on _all_ entries and it's up to the implementation of that function to choose/filter the correct one.  I was thinking the choosing/filtering logic could go in the loop here instead, with the function only being called on the entry that passes the filter.

pchila: This would mean that switching an active install from a version to another would require two separate calls to Modify(). One think to keep in mind is that in this PR the install registry is not yet used concurrently but that will become necessary as asynchronous cleanup is implemented.
The idea behind the `ModifyInstallDesc` is that it allows for modification of the agent installs contained in the `v1.InstallDescriptor` without being able to add or remove any of those.

pchila:
CI failures in build https://buildkite.com/elastic/elastic-agent/builds/28194 are due to `TestSensitiveLogsESExporter` integration test failing (untouched by this PR but recently introduced with https://github.com/elastic/elastic-agent/pull/9341 )
Edit: fixed by a rebase on latest `main`

pkg/utils/install/file_source.go
ycombinator: Understood now.  There is a use case where you need to modify two entries at once.  In that case, this is good as-is.  Thanks for the discussion.

pkg/api/v1/install.go
```diff
@@ -0,0 +1,53 @@
...
+	TTL *time.Time `yaml:"ttl,omitempty" json:"ttl,omitempty"`
+}
+
+type AgentInstallDesc struct {
+	OptionalTTLItem `yaml:",inline" json:",inline"`
+	Version         string `yaml:"version,omitempty" json:"version,omitempty"`
+	Hash            string `yaml:"hash,omitempty" json:"hash,omitempty"`
+	VersionedHome   string `yaml:"versionedHome,omitempty" json:"versionedHome,omitempty"`
+	Flavor          string `yaml:"flavor,omitempty" json:"flavor,omitempty"`
+	Active          bool   `yaml:"active,omitempty" json:"active,omitempty"`
```
cmacknz: Why do we need to model the active installation, instead of just the set of available versions for rollback?

internal/pkg/agent/install/install.go
```diff
@@ -516,13 +521,15 @@ func hasAllSSDs(block ghw.BlockInfo) bool {
 
 // CreateInstallMarker creates a `.installed` file at the given install path,
 // and then calls fixInstallMarkerPermissions to set the ownership provided by `ownership`
-func CreateInstallMarker(topPath string, ownership utils.FileOwner) error {
+func CreateInstallMarker(topPath string, ownership utils.FileOwner, home string, version string, flavor string) error {
 	markerFilePath := filepath.Join(topPath, paths.MarkerFileName)
-	handle, err := os.Create(markerFilePath)
+	installDescProvider := install.NewFileDescriptorSource(markerFilePath)
+	installDesc := v1.AgentInstallDesc{Version: version, VersionedHome: home, Flavor: flavor, Active: true}
+	_, err := installDescProvider.AddInstallDesc(installDesc)
```
cmacknz: If I install agent from this branch I see `.installed` is empty initially which is not what I expect from reading this block.

```
❯ sudo elastic-development-agent version
Binary: 9.3.0-SNAPSHOT (build: c2223f287c7bb1d69524a2c6c261670d681de168 at 2025-10-07 21:00:49 +0000 UTC)
Daemon: 9.3.0-SNAPSHOT (build: c2223f287c7bb1d69524a2c6c261670d681de168 at 2025-10-07 21:00:49 +0000 UTC)
~/Downloads/builds/elastic-agent-9.3.0-SNAPSHOT-darwin-aarch64
❯ sudo bat /Library/Elastic/Agent/.installed
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /Library/Elastic/Agent/.installed   <EMPTY>
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────
~/Downloads/builds/elastic-agent-9.3.0-SNAPSHOT-darwin-aarch64
❯
```

pkg/api/v1/install.go
cmacknz: If we got rid the active install, would we also be able to remove the logic around `activateInstallInRegistry` in upgrade.go and just create entries with their TTL instead?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -498,6 +517,65 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, rollback bool, s
...
+
+func getAvailableRollbacks(rollbackWindow time.Duration, now time.Time, newVersionedHome string, descriptor *v1.InstallDescriptor) []v1.AgentInstallDesc {
+	if rollbackWindow == 0 {
+		// if there's no rollback window it means that no rollback should survive the watcher cleanup at the end of the grace period.
+		return nil
+	}
+
+	res := make([]v1.AgentInstallDesc, 0, len(descriptor.AgentInstalls))
+	for _, installDesc := range descriptor.AgentInstalls {
+		if installDesc.VersionedHome != newVersionedHome && (installDesc.TTL == nil || now.Before(*installDesc.TTL)) {
```
cmacknz: - https://github.com/elastic/ingest-dev/issues/6061

I think there is some danger here, if the system went to sleep long enough (consider people turn off their laptops on vacation) the previous install could be detected as expired before launching the watcher for an upgrade. We also want to make sure people can't brick their agents during upgrade if they set a rollback TTL of a very small value accidentally and remove the ability to do an  automated rollback.

Are we able to keep the previous versions around and be able to roll back to it  regardless of what the rollback configuration or TTL might be if there is an active upgrade in progress or something along those lines?

pkg/api/v1/install.go
pchila: The install registry has the responsibility of keeping track of agent installation and which is the active install at any moment.

The active version should not be a valid rollback target (there is a window between the link rotation and restart where the active version and the version that is currently running do not match).
The active flag is also used to avoid removing it during cleanup of old installs (there's no case where you would want that to happen). 

Again the install registry is not there only for the available rollbacks, if we want it to become that I will remove this and change the implementation once more (maybe by adding another file  under the `data` directory which will have a lifecycle separate from the upgrade marker). Let me know if you would prefer that implementation @cmacknz

internal/pkg/agent/install/install.go
pchila: You expectation is correct, I am not sure why you don't have the entry as in my test (and in [pkoutsovasilis' ones](https://github.com/elastic/elastic-agent/pull/10344#pullrequestreview-3305409984)) the current install is there, although I see that you are using `elastic-agent-development` for checking the version
`sudo elastic-development-agent version` 
but then you look in the non-development topDir for checking the `.installed`
`sudo bat /Library/Elastic/Agent/.installed`

Maybe you could take another look ?

internal/pkg/agent/application/upgrade/upgrade.go
pchila: If I understand you correctly, a version that is still involved in an **active** upgrade should not be cleaned up even if the TTL expires, right?

internal/pkg/agent/install/install.go
cmacknz: Behold! If I actually look the right place it is as expected:

```
❯ sudo bat /Library/Elastic/Agent-Development/.installed
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /Library/Elastic/Agent-Development/.installed
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ version: co.elastic.agent/v1
   2   │ kind: InstallDescriptor
   3   │ agentInstalls:
   4   │ - version: 9.3.0-SNAPSHOT
   5   │   versionedHome: data/elastic-agent-9.3.0-SNAPSHOT-c2223f
   6   │   flavor: basic
   7   │   active: true
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

cmacknz:
@pchila and I met today and came to the following conclusions:

* We should have a way to feature flag the rollback functionality, so that we can surface errors related to new code and fail the upgrade but also have a way to bypass unexpected bugs.
  * The simplest path to do this is to use the rollback window parameter we already support and will expose in the Fleet UI. 
  * We can interpret having a rollback_window of 0 as a request to disable the rollback logic.
  * In concrete terms of the discussion of this PR, this would mean we treat updating the registry of available rollback versions as fatal errors unless rollback window is zero where we either ignore or simply don't update the registry at all.
* We should move the installation registry out of the `.installed` file.
  * The largest reason to do this is that it makes it much easier for us to change or relocate the list of registry versions by deleting or ignoring the new file.
    * For example if we had a local database available, we would prefer to track this there.
  * As part of this Paolo suggested we should only track versions available for rollback through the upgrade process, and remove the updating the registry at install time to further simplify things.

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz: Yes to mitigate the chance that the agent machine was asleep and the wall clock advanced but the upgrade itself has not e.g. the watcher should still be running (once the bug I linked above is fixed...)

internal/pkg/agent/application/upgrade/upgrade.go
pchila: This is no longer relevant as the install registry has been scrapped (see PR description)

pkg/api/v1/install.go
pchila: This is no longer relevant as the install registry has been scrapped (see PR description)

pchila:
> @pchila and I met today and came to the following conclusions:
> 
>     * We should have a way to feature flag the rollback functionality, so that we can surface errors related to new code and fail the upgrade but also have a way to bypass unexpected bugs.
>       
>       * The simplest path to do this is to use the rollback window parameter we already support and will expose in the Fleet UI.
>       * We can interpret having a rollback_window of 0 as a request to disable the rollback logic.
>       * In concrete terms of the discussion of this PR, this would mean we treat updating the registry of available rollback versions as fatal errors unless rollback window is zero where we either ignore or simply don't update the registry at all.
> 
>     * We should move the installation registry out of the `.installed` file.
>       
>       * The largest reason to do this is that it makes it much easier for us to change or relocate the list of registry versions by deleting or ignoring the new file.
>         
>         * For example if we had a local database available, we would prefer to track this there.
>       * As part of this Paolo suggested we should only track versions available for rollback through the upgrade process, and remove the updating the registry at install time to further simplify things.

@cmacknz see d4cc617fceca97994a1f51e6d32bf24606a795f0

internal/pkg/agent/application/upgrade/ttl_marker_source.go
```diff
@@ -0,0 +1,118 @@
...
+const ttlMarkerName = ".ttl"
+
+var defaultMarkerGlobPattern = filepath.Join("data", "elastic-agent-*", ttlMarkerName)
+
+type TTLMarkerRegistry struct {
+	baseDir               string
+	markerFileGlobPattern string
+}
+
+func (T TTLMarkerRegistry) AddOrReplace(m map[string]TTLMarker) error {
```
ycombinator: Could we make the visibility of this method private to the struct?   Looks like the interface to the TTLMarkerRegistry is just `Get()` and `Set()`.

internal/pkg/agent/application/upgrade/ttl_marker_source.go
```diff
@@ -0,0 +1,118 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+const ttlMarkerName = ".ttl"
+
+var defaultMarkerGlobPattern = filepath.Join("data", "elastic-agent-*", ttlMarkerName)
+
+type TTLMarkerRegistry struct {
+	baseDir               string
+	markerFileGlobPattern string
```
ycombinator, line 23: The operations on this struct loops over the installation folders and performs deletions or writes on the TTL marker within each.  If something goes wrong, we error out early.  This might leave the installation state in a partial state which could be hard to debug.  Could we add a logger to this struct and log the deletion and write operations at a debug level?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -400,6 +407,11 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, rollback bool, s
 		return nil, err
 	}
 
+	currentVersionedHome, err := filepath.Rel(paths.Top(), paths.Home())
```
ycombinator: Any particular reason this block got moved up?

internal/pkg/agent/cmd/run.go
```diff
@@ -753,7 +753,7 @@ func handleUpgrade() (*details.Details, error) {
 		return nil, err
 	}
 
-	return upgradeMarker.Details, nil
+	return upgradeMarker, nil
```
ycombinator, line 756: I'm curious why this needed to be changed?  Maybe I missed something but from what I can tell, we are now passing around the upgrade marker instead of upgrade details but in `application.go`, we pull out the upgrade details from the upgrade marker anyway.

pkg/utils/manifest/version.go
```diff
@@ -0,0 +1,15 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package manifest
+
+import v1 "github.com/elastic/elastic-agent/pkg/api/v1"
+
+func GetFullVersion(manifest *v1.PackageManifest) string {
```
ycombinator: Is this function being used anywhere?

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -66,9 +66,9 @@ var (
 	ErrNilUpdateMarker      = errors.New("loaded a nil update marker")
 	ErrEmptyRollbackVersion = errors.New("rollback version is empty")
 	ErrNoRollbacksAvailable = errors.New("no rollbacks available")
-
-	// Version_9_2_0_SNAPSHOT is the minimum version for manual rollback and rollback reason
-	Version_9_2_0_SNAPSHOT = agtversion.NewParsedSemVer(9, 2, 0, "SNAPSHOT", "")
+	ErrAgentInstallNotFound = errors.New("agent install descriptor not found")
+	// Version_9_3_0_SNAPSHOT is the minimum version for manual rollback and rollback reason
```
cmacknz, line 70: > Version_9_3_0_SNAPSHOT is the minimum version for manual rollback

A bit beyond the scope of this PR, but one of the advantages of rollback being implemented as a "switch from current to previous agent saved on disk" is that it would work for any previous agent version from the point where the agent was introduced.

Have we now lost that with the marker getting removed and this change? Or are we just losing the rollback reason?

What we had before was a 9.3.0 agent could rollback to any version, do we now only support a 9.4.0 agent rolling back to a 9.3.0 agent as the first supported version combination?

cmacknz: Are we mixing two different things together?

Any version < 9.2.0 has logic that doesn't look inside the upgrade marker and only looks for its presence. This would mean that upgrade details can't report the rollback state after a rollback happens on these versions, which is OK because some versions are old enough to not understand upgrade details at all if you go back far enough. In this case the design should allow the rollback and we document the limitation that you'll only see the agent version change and might lose other details reporting.

Any version < 9.3.0 doesn't understand the install registry (in the case where we do an upgrade to a lower version like our tests currently do).

pkg/utils/manifest/path_mapper.go
```diff
@@ -0,0 +1,30 @@
...
+	"path"
+	"strings"
+)
+
+// PathMapper is a utility object that will help with File mappings specified in a v1/Manifest
+type PathMapper struct {
+	mappings []map[string]string
+}
+
+func (pm PathMapper) Map(packagePath string) string {
```
cmacknz, line 17: This could be unit tested, I see it used in step_unpack but don't see unit tests that exercise all the cases for this.

internal/pkg/agent/application/upgrade/upgrade.go
```diff
@@ -455,21 +467,24 @@ func (u *Upgrader) Upgrade(ctx context.Context, version string, rollback bool, s
...
 		current,  // new agent version data
 		previous, // old agent version data
-		action, det, rollbackWindow); err != nil {
+		action, det, availableRollbacks); err != nil {
 		u.log.Errorw("Rolling back: marking upgrade failed", "error.message", err)
 		rollbackErr := u.rollbackInstall(ctx, u.log, paths.Top(), hashedDir, currentVersionedHome)
 		return nil, goerrors.Join(err, rollbackErr)
 	}
 
+	if err = u.availableRollbacksSource.Set(availableRollbacks); err != nil {
```
cmacknz: It's possible for this to fail when `markUpgrade` above fails and vice versa.

This leaves agent in an inconsistent state doesn't it?

Is there a way to have a single source of truth, like "joining" the .ttl file(s) into the upgrade marker by first reading the .ttl file when going to write the marker?

I see normalizeInstallDescriptorAtStartup in https://github.com/elastic/elastic-agent/pull/9643 is maybe intended to do some of this reconciliation.

cmacknz (commented):
Thanks for simplifying, a couple of comments but direction overall LGTM

internal/pkg/agent/application/upgrade/upgrade.go
pchila: @cmacknz I guess you are mixing 2 things together:
- the rollback reason is not gated by any settings and since https://github.com/elastic/elastic-agent/pull/10579 does some version checks to detect if the version we are **rolling back to** can handle correctly an upgrade marker with a UPG_ROLLBACK state (>= 9.2.0-SNAPSHOT).
 This piece of code was modified before PR https://github.com/elastic/elastic-agent/pull/10579 was created and now will need to be adapted.
 
- The manual rollback does a version check to verify that the version we are **upgrading to**  is able to correctly handle the delayed cleanup of an install that was left on disk as a possible rollback target. This needs to be checked to avoid leaving an install that will not be cleaned at the end of the rollback window (>= 9.3.0-SNAPSHOT).

internal/pkg/agent/application/upgrade/ttl_marker_source.go
pchila: The Get and Set operation will return an error related to the first install where there's been a problem so that should be handled by the callers (which will most likely log it).
I can add a logger to the struct and trace each install operation, although it won't add more information for the error cases, it will only allow to follow the execution of the TTLMarkerRegistry

internal/pkg/agent/application/upgrade/upgrade.go
pchila: leftover from the install registry

internal/pkg/agent/cmd/run.go
pchila: leftover from the install registry

pkg/utils/manifest/version.go
pchila: leftover from the install registry

internal/pkg/agent/application/upgrade/upgrade.go
pchila: In the previous iteration you were worried that errors tracking available rollbacks could interrupt the upgrade process and asked for such errors to be treated as non-fatal.
Now you are concerned about availableRollbacksSource failing and leaving the marker files in an inconsistent state ?
The recovery/cleanup logic for a failing upgrade is in `u.rollbackInstall()` so if we want to abort and maintain things consistent we should stop the upgrade and cleanup any ttl marker in `u.rollbackInstall()`

The reconciliation you see in `normalizeInstallDescriptorAtStartup` is intended to cleanup expired installs at startup (a second, async cleanup while the agent is running is in the works)

pkg/utils/manifest/path_mapper.go
pchila: This **is** unit-tested. It's been moved to its own file for the install registry (which is now scrapped).
This code is tested in both the step_unpack and step_mark tests since 8.13.0.
If it makes it easier to see I will move it back to where it was.

internal/pkg/agent/application/upgrade/ttl_marker_source.go
pchila: Done in 508ecc0e1575987a41bd66ad32d24ef6a56f29c6

pkg/utils/manifest/version.go
pchila: removed in 11a5a70bc9bd48cb9f8a94fa88bd543525111475

pkg/utils/manifest/path_mapper.go
pchila: d8ff91f366f54c1ce3a87d09837caaa0aecda5cf

internal/pkg/agent/application/upgrade/ttl_marker_source.go
pchila: eaa7582807d2534879032323322481c35a1b78dc

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz: Yep for some reason I thought this was the version used in https://github.com/elastic/elastic-agent/pull/10579 but clearly it is not.

internal/pkg/agent/application/upgrade/upgrade.go
cmacknz: My concern is that we store `availableRollbacks` in two places that I can see here:

1. Once in the upgrade marker via the `u.markUpgrade`. I missed this was a fatal error above, and I think it has to be because the upgrade process can't function if the upgrade marker isn't created.
2. Again in `u.availableRollbacksSource.Set`, which can now fail even though `u.markUpgrade` succeeded (where yes - this being non-fatal was my request).

If `u.availableRollbacksSource.Set` were a fatal error, the upgrade marker is still kept around in versions after 9.2.0 so I don't think making this fatal fixes the potential inconsistency unless I'm missing something. We would report available rollbacks via the marker that potentially (though highly unlikely) don't have .ttl markers.

pchila: The reporting will be implemented using the ttl marker files. The reason why the available rollbacks are in the upgrade marker as well is to let the watcher know which installs it needs to preserve at the end of the grace period without the need to read extra file(s) (keep in mind that this solution was implemented when the install registry was in place).
The `rollback*` function in this and the next PR will give precedence to the upgrade marker if it exists, then fallback to the available rollbacks extracted from the related source once the upgrade marker is gone.
In case agent manages to write the marker but not the TTL files, we can rollback during the grace period but not after that and the install will be cleaned up only when performing another install.

We can avoid writing the available rollbacks in the upgrade marker, although it becomes imperative that the watcher uses the **same available rollback source implementation** as the agent starting the upgrade (keeping in mind that the watcher is likely a different elastic-agent version).

cmacknz: Thanks. Now that we have the rollback window acting as a feature flag for the new logic:

```go
func getAvailableRollbacks(rollbackWindow time.Duration, now time.Time, currentVersion string, parsedCurrentVersion *agtversion.ParsedSemVer, currentVersionedHome string) map[string]TTLMarker {
	if rollbackWindow == 0 {
		// if there's no rollback window it means that no rollback should survive the watcher cleanup at the end of the grace period.
		return nil
	}
```

What do you think of making `u.availableRollbacksSource.Set` a fatal error and moving it before the `markUpgrade` call?

When I first asked about making this a fatal error, we didn't have this bypass. Now that we do have a way to make this call a no-op, it's safe to make it fatal because we have a way get past it if there is some unexpected error. I think this would get rid of the inconsistency problem here safely.

pchila: Yes it would, along with clearing any ttl markers in `u.rollbackInstall()` when things go wrong ;)

internal/pkg/agent/application/upgrade/step_mark.go
```diff
@@ -23,10 +23,9 @@ import (
 const markerFilename = ".update-marker"
 const disableRollbackWindow = time.Duration(0)
 
-// RollbackAvailable identifies an elastic-agent install available for rollback
-type RollbackAvailable struct {
+// TTLMarker marks an elastic-agent install available for rollback
+type TTLMarker struct {
```
cmacknz, line 27: Did you consider storing and reporting the creation date along with valid until? We could perhaps get that directly from the file system too.

Something I am thinking about is giving users a way to purge available rollback versions on disk before valid until expires. 

In the current implementation this gets set at the time of the upgrade action introducing the rollback version with no way to accelerate or immediately delete that can I see. In the theme of  handling the unexpected, if a user needs to free up disk space taken up by a rollback version I don't see a way to do that right now besides waiting unless I am missing it.

I don't expect this to be handled completely this PR just raising the idea, I can fork out a separate GH issue if this isn't already on your radar or handled in some other way.

pchila: It was considered in the early stages of the related RFC:

> A manual cleanup operation may be added in future enhancements, cutting short the rollback window in case users need to regain disk space but it’s out of scope of this RFC.

But yeah the design is such that it allows cleanup of previous installs while the agent is running, either automatically (when the rollback window expires) or manually (we need to decide what kind of command should be issued via CLI, Fleet has not really been considered for this case)

internal/pkg/agent/application/upgrade/upgrade.go
pchila: @cmacknz done in 478435f348743667d1ef8185663c57f7b1c174cf

internal/pkg/agent/application/upgrade/step_mark.go
cmacknz: Thanks, created https://github.com/elastic/elastic-agent/issues/10695

PR #10374, opened 2025-10-07
Prevent migration of agents on Docker 

michalpristas:
Just exposing platform used in component specs for check to happen, other option would be to inject it from the top

this seems less disruptive. 

Related https://github.com/elastic/kibana/issues/231151


blakerouse (approved):
Nice and simple. I like it.

PR #10378, opened 2025-10-07
feat!: remove Elastic Trace processor from EDOT Collector

andrzej-stencel:
## What does this PR do?

Removes the [Elastic Trace processor](https://github.com/elastic/opentelemetry-collector-components/blob/processor/elastictraceprocessor/v0.10.0/processor/elastictraceprocessor/README.md) from EDOT Collector, replacing all usages with its direct drop-in replacement, the [Elastic APM processor](https://github.com/elastic/opentelemetry-collector-components/blob/processor/elasticapmprocessor/v0.2.0/processor/elasticapmprocessor/README.md).

## Why is it important?

The Elastic Trace processor was deprecated and its code was removed in https://github.com/elastic/opentelemetry-collector-components/pull/784.

## Checklist

## Disruptive User Impact

This is a breaking change for users if they use the processor in their EDOT Collector configurations.

The fix is simple: replace `elastictrace` with `elasticapm` in their configurations.

```console
go build .
sudo ./elastic-aget otel components
```

### DIFF
--- a/docs/reference/edot-collector/custom-collector.md
+++ b/docs/reference/edot-collector/custom-collector.md
@@ -113,10 +113,10 @@ processors:
     gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.130.0
   cumulativetodeltaprocessor :
     gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/cumulativetodeltaprocessor v0.130.0
-  elastictraceprocessor :
-    gomod: github.com/elastic/opentelemetry-collector-components/processor/elastictraceprocessor v0.9.0
   elasticinframetricsprocessor :
     gomod: github.com/elastic/opentelemetry-collector-components/processor/elasticinframetricsprocessor v0.16.0
+  elastictraceprocessor :
+    gomod: github.com/elastic/opentelemetry-collector-components/processor/elastictraceprocessor v0.9.0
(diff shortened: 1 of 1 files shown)


andrzej-stencel:
> Please revert the changes in `components.md` and `custom-collector.md`, since those docs are meant to show the latest stable release (9.1.5) and will update automatically upon 9.2 release.

Right, fixed in https://github.com/elastic/elastic-agent/pull/10378/commits/56127abc17d61a02d8c19afbb5faed6dfbcaff6c.

ycombinator (approved):

ChrsMark (approved):
@gregkalapos mind taking a look too?

gregkalapos:
We just had a zoom call with @andrzej-stencel and @AlexanderWert and concluded that removing this processor would be a breaking change, which we'd like to avoid.

On the other hand, the code of the elastic trace processor is already removed from the collector components repo, so the last version depends on older upstream components, which makes bumping versions harder - so this would have not worked probably anyways as it was in this repo.

Plan is that I'll try to create a small processor with the same name `elastictraceprocessor` in the component repo, which just forwards to the new `elasticapmprocessor`. But if we want to avoid breaking changes, we'll need a follow up PR and add back `elastictraceprocessor` with the new version.

PR #10385, opened 2025-10-07
Update .dockerignore to include .cache directory

florianl:
## What does this PR do?

When inspecting Elastic Agent docker images[^1] for symbolization, I noticed executables in the following path: `/usr/share/elastic-agent/.cache/` that increase significantly the size of the image.

[^1]: docker.elastic.co/elastic-agent/elastic-agent-complete:8.19.4 in particular

## Why is it important?

Reduce the size of docker images by not including `.cache`.

The current[^1] size of `.cache` seems to be quite large:
```
$ du -h -d 1
914M	./.cache
```

## Checklist

### DIFF
--- a/changelog/fragments/1759851450-docker-ignore-cache.yaml
+++ b/changelog/fragments/1759851450-docker-ignore-cache.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: other
+
+# Change summary; a 80ish characters long description of the change.
+summary: do not include .cache directory in docker images
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: "all"
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234


straistaru:
First, I don't think that adding `.cache` in the `.dockerignore` does anything regarding `/usr/share/elastic-agent/.cache/` inside the Docker image.
I built the elastic-agent with these changes and the contents of the `.cache` directory are the same with the ones in the `docker.elastic.co/elastic-agent/elastic-agent-complete:8.19.4` image you mention.
```
$ cd /usr/share/elastic-agent/.cache
$ ls -l 
total 4
drwxr-xr-x 6 elastic-agent elastic-agent 4096 Sep 17 10:00 ms-playwright
$ du -h -d 1
888M    ./ms-playwright
888M    .
```
And that is because the contents of this directory are created directly inside the image, when installing `@elastic/synthetics` during this step [here](https://github.com/elastic/elastic-agent/blob/8087ea35213ca311309bdf357ae203a3811f0d06/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl#L248-L249) or [here](https://github.com/elastic/elastic-agent/blob/8087ea35213ca311309bdf357ae203a3811f0d06/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl#L305-L306), for the `complete` variant and `complete-wolfi` variant respectively.

So, to remove these files, we should add a new step: `RUN rm -rf {{ $beatHome }}/.cache/`

But second, are we sure we actually want to remove these files? The name `.cache` implies this, but I am not sure.

florianl:
@straistaru the **compressed** size of [docker.elastic.co/elastic-agent/elastic-agent-complete:8.19.4](https://www.docker.elastic.co/r/elastic-agent/elastic-agent-complete?limit=2000&offset=0)  is 1336 MB. I didn't check elastic-agent/elastic-agent-complete-wolfi:8.19.4, as I don't see this image being deployed in the environment I have access to.

With 914M the initally provided point by `du -h -d 1` shows the uncompressed size. And there are more opportunities, reducing the image even further.

> First, I don't think that adding .cache in the .dockerignore does anything regarding /usr/share/elastic-agent/.cache/ inside the Docker image.
I built the elastic-agent with these changes and the contents of the .cache directory are the same with the ones in the docker.elastic.co/elastic-agent/elastic-agent-complete:8.19.4 image you mention.

I think, it is up to the team to investigate why docker templates in https://github.com/elastic/elastic-agent/tree/main/dev-tools/packaging/templates/docker ignore the existence and instructions in .dockerignore.

straistaru:
> I think, it is up to the team to investigate why docker templates in https://github.com/elastic/elastic-agent/tree/main/dev-tools/packaging/templates/docker ignore the existence and instructions in .dockerignore.

@florianl, I don't think the dockerfile template ignores the `.dockerignore`.

Please correct me if I am wrong, but the `.dockerignore` only controls the context that is sent to the docker build command and doesn't influence what happens inside the container during RUN steps.

kruskall:
According to the doc 

> A [.dockerignore file](https://docs.docker.com/build/concepts/context/#dockerignore-files) lets you specify which local files should not be included in the build contex

however the `.cache` directory is not a local file but it's created in the docker image during a `RUN` statement: the one at https://github.com/elastic/elastic-agent/blob/492e2d73231c24c33235630cd1683ca5944debca/dev-tools/packaging/templates/docker/Dockerfile.elastic-agent.tmpl#L248

This can be observed with `dive docker.elastic.co/elastic-agent/elastic-agent-complete:8.19.4`, turning off "unmodified files` and scrolling layer by layer until the RUN statement mentioned above.

IIUC adding `.cache` to the `.dockerignore` won't prevent the layer and the .cache directory from being created.

pchila:
I agree with what @straistaru and @kruskall said above. Adding `.cache` to `.dockerignore` has no effect if there's no `.cache` in the elastic-agent repo when triggering the docker build.

PR #10412, opened 2025-10-08
fix: do not report agent managed otel extensions statuses

pkoutsovasilis:
## Checklist

## Related issues

-

## Questions to ask yourself

- ...


internal/pkg/otel/manager/manager_test.go
```diff
@@ -600,6 +603,41 @@ func TestOTelManager_Run(t *testing.T) {
...
+
+				componentType, err := otelComponent.NewType(healthCheckExtensionName)
+				require.NoError(t, err, "failed to create component type")
+
+				healthCheckExtensionID := otelComponent.NewIDWithName(componentType, nsUUID.String()).String()
+
+				ports, err := findRandomTCPPorts(3)
+				require.NoError(t, err, "failed to find random tcp ports")
+				subprocessExec.collectorHealthCheckPort = ports[0]
+				subprocessExec.collectorHealthCheckPort = ports[1]
```
cmacknz: You are assigning the same port twice which seems a little weird.

pkoutsovasilis: it's 100% weird because it's a typo, on the good news we allocate a random tcp port for it 😆 thanks for catching that fixed in 9b8535d0cbeceaaf14d689e93b3d3b6ed534c61a

internal/pkg/otel/manager/manager_test.go
```diff
@@ -1198,6 +1236,17 @@ func TestOTelManager_handleOtelStatusUpdate(t *testing.T) {
 					"pipeline:logs": {
 						Event: componentstatus.NewEvent(componentstatus.StatusOK),
 					},
+					"extensions": {
```
cmacknz, line 1239: If I take this out of this test it has no effect, what is this testing?

internal/pkg/otel/manager/manager.go
```diff
@@ -505,6 +506,24 @@ func (m *OTelManager) MergedOtelConfig() *confmap.Conf {
 // and prepares component state updates for distribution to watchers.
 // Returns component state updates and any error encountered during processing.
 func (m *OTelManager) handleOtelStatusUpdate(otelStatus *status.AggregateStatus) ([]runtime.ComponentComponentState, error) {
+	// Remove agent managed extensions from the status report
+	if otelStatus != nil {
+		if extensionsMap, exists := otelStatus.ComponentStatusMap["extensions"]; exists {
```
cmacknz, line 511: I don't see this tested anywhere explicitly unless I am missing it.

Trying the PR locally it does work to hide the internal statuses, but I didn't mess around with adding in extensions in manually configured collector pipelines or anything.

internal/pkg/otel/manager/manager_test.go
pkoutsovasilis: it does have an effect and actually validates the code in handleOtelStatusUpdate. 
* Here we define the expected collector status and I deliberately don't specify the statuses of the extensions I introduced above https://github.com/elastic/elastic-agent/blob/9b8535d0cbeceaaf14d689e93b3d3b6ed534c61a/internal/pkg/otel/manager/manager_test.go#L1252-L1260
* Here we call `handleOtelStatusUpdate` https://github.com/elastic/elastic-agent/blob/9b8535d0cbeceaaf14d689e93b3d3b6ed534c61a/internal/pkg/otel/manager/manager_test.go#L1335-L1337
* Here we check that the returned status matches the expected one https://github.com/elastic/elastic-agent/blob/9b8535d0cbeceaaf14d689e93b3d3b6ed534c61a/internal/pkg/otel/manager/manager_test.go#L1350-L1352

If you still don't see it comment out these lines https://github.com/elastic/elastic-agent/blob/9b8535d0cbeceaaf14d689e93b3d3b6ed534c61a/internal/pkg/otel/manager/manager.go#L510-L525 and invoke the unit-test it should fail. 

What am I missing here?

internal/pkg/otel/manager/manager.go
pkoutsovasilis: I answered in the comment above. PS I have added a dedicated unit-test here https://github.com/elastic/elastic-agent/blob/9b8535d0cbeceaaf14d689e93b3d3b6ed534c61a/internal/pkg/otel/manager/manager_test.go#L607-L639 that checks that a user-defined healthcheck extension status is reported

internal/pkg/otel/manager/manager_test.go
cmacknz: OK yeah I got it and confirm it does what you describe, should have been removing the additional code to check the test failed and not modifying the test. Thanks.

PR #10418, opened 2025-10-08
Migrate to Mockery v3

pchila:
## What does this PR do?
Migrate mockery to [v3](https://vektra.github.io/mockery/v3.5/v3/) and adapt existing tests.
All mocks are now in `mocks.go` in the same package as the interfaces they are mocking.
Default `mocks_test.go` filename does not work as interfaces are often defined on the implementer side rather than the consumer side.

## Why is it important?
To keep our mock generation tool up to date.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- [ ] I have added an integration test or an E2E test

## How to test this PR locally
Install mockery v3 (for example version [v3.5.5](https://github.com/vektra/mockery/releases/tag/v3.5.5) ) and run `mage dev:regenerateMocks`

## Related issues

- Closes #9277

## Questions to ask yourself

- How are we going to support this in production?
- How are we going to measure its adoption?
- How are we going to debug this?
- What are the metrics I should take care of?
- ...

PR #10436, opened 2025-10-09
[tests][integration]: move EDOT mock hanlder to mock_server_test.go

VihasMakwana:
After https://github.com/elastic/elastic-agent/pull/10418, all the mock handlers live in `mocks.go`. This PR fixes the tests and moves `handlers/diagnostics_extension_mock.go` code to `mock_server_test.go` to unblock main.

### DIFF
--- a/internal/pkg/agent/application/actions/handlers/mock_server_test.go
+++ b/internal/pkg/agent/application/actions/handlers/mock_server_test.go
@@ -1,0 +1,48 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package handlers
+
+import (
+	"encoding/json"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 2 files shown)


internal/pkg/agent/application/actions/handlers/mock_server_test.go
pchila, line 1: please do not use http.Server in tests: httptest.Server is more lightweight while still implementing all the relevant methods.
This can be fixed in a follow-up PR but please have a look at https://pkg.go.dev/net/http/httptest

pchila:
> After #10418, all the mock handlers live in `mocks.go`. This PR fixes the tests and moves `handlers/diagnostics_extension_mock.go` code to `mock_server_test.go` to unblock main.

After #10418 all the stretchr/testify `mock.Mock`s generated by mockery have been moved in-package with the migration to mockery v3.
The hand-written MockServer should not have been placed among the generated mocks since the code would be lost at the first mock regeneration.

internal/pkg/agent/application/actions/handlers/mock_server_test.go
VihasMakwana: I had a quick look at it. Looking at its internals, it is hardcoded for `tls` only, whereas we rely on sockets and npipes for our integration tests.

I'll take a look at this in depth and confirm if we can use this in a follow-up. Thanks for mentioning this package!!

PR #10446, opened 2025-10-09
mage: Improve error messages and resource handling

orestisfl:
## What does this PR do?
This commit includes minor developer tooling improvements:
- Improves the checksum mismatch error message for better diagnostics.
- Fix spurious "error closing" logs when double-closing files explicitly and with `defer`

## Checklist


dev-tools/mage/downloads/utils_test.go
```diff
@@ -86,6 +86,7 @@ func TestVerifyChecksum(t *testing.T) {
 		require.NoError(t, os.WriteFile(checksumPath, []byte(checksumContent), 0644))
 
 		err := verifyChecksum(checksumPath)
-		assert.ErrorContains(t, err, "does not match expected checksum")
+		contains := fmt.Sprintf("%s checksum mismatch: expected=%s", fileName, hashHex)
+		assert.ErrorContains(t, err, contains)
```
pchila, line 90: You could use [`assert.ErrorContainsf()`](https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.ErrorContainsf)
```suggestion
		assert.ErrorContainsf(t, err, "%s checksum mismatch: expected=%s", fileName, hashHex)
```

dev-tools/mage/pkgtypes.go
```diff
@@ -759,6 +748,14 @@ func PackageTarGz(spec PackageSpec) error {
 	return nil
 }
 
+func closeOrLog(closer io.Closer, what string) {
+	err := closer.Close()
+	if err == nil || errors.Is(err, os.ErrClosed) {
```
pchila, line 756: This is a change in behavior: you are explicitly ignoring `os.ErrClosed`. Wouldn't it be better to figure out where we are closing the same `io.Closer` twice ?

dev-tools/mage/downloads/utils_test.go
orestisfl: The `f` part of `ErrorContainsf` is for the message printed on failure
See example:
```go
actualObj, err := SomeFunction()
a.ErrorContainsf(err,  expectedErrorSubString, "error message %s", "formatted")
```

```
=== RUN   TestVerifyChecksum/checksum_mismatch
    utils_test.go:90: 
        	Error Trace:	/home/orestis/src/elastic-agent/dev-tools/mage/downloads/utils_test.go:90
        	Error:      	Error "file.txt checksum mismatch: expected=309ecc489c12d6eb4cc40f50c902f2b4d0ed77ee511a7c7a9bcd3ca86d4cd86f989dd35bc5ff499670da34255b45b0cfd830e81f605dcf7dc5542e93ae9cd76f actual=3fa064fffac7ab8dd1921733c492ef202be66dd39f408e7e168c25b26fb0f59e8c2e44882ccd31a3b5816d2dc2351254afd34fc793950edd1fbb73b1435235cd" does not contain "%s checksum mismatch: expected=%s"
        	Test:       	TestVerifyChecksum/checksum_mismatch
        	Messages:   	file.txt%!(EXTRA string=309ecc489c12d6eb4cc40f50c902f2b4d0ed77ee511a7c7a9bcd3ca86d4cd86f989dd35bc5ff499670da34255b45b0cfd830e81f605dcf7dc5542e93ae9cd76f)
--- FAIL: TestVerifyChecksum/checksum_mismatch (0.00s)
```

dev-tools/mage/pkgtypes.go
orestisfl: That can be done as well but I figured cleanliness is slightly preferred over having to set nil to the file pointers, at least for this part of the code. Other parts don't even log the `Close` errors.

pierrehilbert:
@pchila / @rubenruizdegauna is any change expected here or are we good to go?

dev-tools/mage/pkgtypes.go
pchila: I would still prefer knowing where the close errors come from.
Hiding an error like this means that it's unlikely to be investigated as not visible on logs.
I will leave it up to @elastic/elastic-agent-control-plane team to decide if this should go in

orestisfl: @pchila take a look at the full function. In the happy path, if Close returns an error, that error is propagated to the caller meaning it will be noticed as the operation fails. Only in the case that PackageTarGz exits for any other error, after the files were closed, the double-close error will be ignored. That can only happen here: https://github.com/orestisfl/elastic-agent/blob/62f163431a1a8988ca9a6821a557061ed6506688/dev-tools/mage/pkgtypes.go#L746

swiatekm: I agree with @orestisfl that this is correct. This function is a bit unusual in that it needs to close its writers explicitly, because it then proceeds to do other operations on the file it writes. However, it still needs to close the writers when it exists due to an error, hence the defers. There's no way to check if the writer is already closed other than trying to close it and catching the error. The change in this PR simply makes it so the happy path (close normally, and then close in a defer again) doesn't produce spurious error logs.

dev-tools/mage/pkgtypes.go
```diff
@@ -669,27 +670,15 @@ func PackageTarGz(spec PackageSpec) error {
...
+	defer closeOrLog(gzWriter, "gzip writer")
 
 	// Create a new tar archive.
 	w := tar.NewWriter(gzWriter)
-	defer func() {
-		if err := w.Close(); err != nil {
-			log.Printf("failed to close tar writer: %v", err)
-		}
-	}()
+	defer closeOrLog(gzWriter, "tar writer")
```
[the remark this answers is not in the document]
orestisfl: Pre-existing but valid. Fixing.

PR #10492, opened 2025-10-10
[AutoOps] Improve Batching Configuration

pickypg:
This brings in recommendations from external testing on OTel data shipping to better understand solid batching.

Note: `block_on_overflow: true` is going to become the default behavior in a future version of OTel.

## What does this PR do?

Updates the batching behavior of the OTLP exporter to be more performant.

## Why is it important?

This should improve performance end-to-end.

## Checklist

- [x] ~~I have made corresponding changes to the documentation~~
- [x] ~~I have added tests that prove my fix is effective or that my feature works~~
- [x] ~~I have added an integration test or an E2E test~~

## Disruptive User Impact

None. This should actually improve performance in practically all scenarios.

Run the Elastic Agent using the configurations provided and ensure it still ships data.

## Related issues

Internal discussions around better queueing.

### DIFF
--- a/changelog/fragments/1760341169-add-logs_metrics_traces.yml-sample-in-EDOT-for-Windows.yaml
+++ b/changelog/fragments/1760341169-add-logs_metrics_traces.yml-sample-in-EDOT-for-Windows.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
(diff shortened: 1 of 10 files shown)


gizas:
@pickypg are we going to also update the https://github.com/elastic/elastic-agent/blob/main/deploy/helm/edot-collector/kube-stack/values.yaml? 

If yes, in this are we going to split the otlphttp exporter one for logs and one for metrics/traces?

pickypg:
> are we going to also update the https://github.com/elastic/elastic-agent/blob/main/deploy/helm/edot-collector/kube-stack/values.yaml?

This PR is explicitly for AutoOps for Self-managed (via Cloud Connected) usage, which currently only uses the `logs` telemetry type because that's all that Elastic Agent supports.

There is a wider effort that is standardizing the batching behavior for EDOT, but it's _not_ intended to be this PR.

PR #10514, opened 2025-10-13
[edot] add sample for otel windows

VihasMakwana:
## What does this PR do?

The sample file `logs_metrics_traces.yml` is shipped in linux and darwin distributions, but is missing from windows.
This PR adds a sample for windows.


VihasMakwana:
@theletterf I'd say all active 9 branches and 8.19 because it doesn't work as per the docs.

theletterf:
Feel free to merge when ready.

Issue #9058, opened 2025-07-17
[beats receivers] Enable beats receivers for internal monitoring data collection by default


cmacknz:
Enable beats receivers for internal monitoring data collection by default on main and in integration tests so we can start soak testing it in development and test environments.

Draft PR showing some unit test failures to address:
- https://github.com/elastic/elastic-agent/pull/9037

That PR also need to change the existing monitoring data comparison test to start with beats receivers off (or just reverse the test order where they start on and are then turned off): https://github.com/cmacknz/elastic-agent/blob/184b0584efc99fd8f3ad9dc1a33018a44a8f726a/testing/integration/ess/beat_receivers_test.go#L43

swiatekm:
I've started checking how our integration tests react to the change in https://github.com/elastic/elastic-agent/pull/10204. Thus far, I've had to:

* Modify the golden file for generated self-monitoring config. We should probably have two of these, separately for running in process and otel mode.
* The runtime leak test needs to detect running agent subprocesses differently. It should probably just look at child processes instead of using status, because right now it wouldn't detect us leaking an otel process. See https://github.com/elastic/elastic-agent/pull/10204/commits/0f3b193221cc62b3c4e6fda95e75d8aa4886d263.
* Our fixtures check if agent was successfully uninstalled by looking at running process commands. As we start the otel collector by calling `elastic-agent otel`, it gets detected as another agent. Fix [here](https://github.com/elastic/elastic-agent/pull/10204/commits/f4f4aee2fdc6f65826463e2defb5fd4e502c1567) only counts agents who aren't children of other agents, although this should eventually become irrelevant when we split the binaries.
* TBD, some tests are still failing.

Overall, I don't see any semantic failures yet, just assertions with implicit assumptions that aren't true anymore.

swiatekm:
Currently blocked by https://github.com/elastic/ingest-dev/issues/6250.

swiatekm:
Blocked by https://github.com/elastic/elastic-agent/issues/10645.

cmacknz:
Moved https://github.com/elastic/elastic-agent/issues/10645 into the parent issue and out of this one, let's get beats receivers for monitoring turned on by default ahead of the fix for this coming in to get a head start on any additional problems that might pop out.

cmacknz:
Marked as unblocked.

PR #10537, opened 2025-10-14
Fix tests in preparation for making otel self monitoring default

swiatekm:
## What does this PR do?

Fixes tests which implicitly assume beats are running as processes as opposed to otel receivers. More specifically:

- https://github.com/elastic/elastic-agent/pull/10537/commits/2952a127abb716dc2b0a4b952fe4689acdf87c77 makes the runtime leak test only look at beats running as processes.
- https://github.com/elastic/elastic-agent/pull/10537/commits/c0d83858f4c58e5f30d607c1a8e0de635ffd6c67 makes integration tests check for running agent processes by only looking at processes without parents which are also elastic agents.
- https://github.com/elastic/elastic-agent/pull/10537/commits/72855e55785d4dfe8a18c9fcbceadd517326aec2 explicitly sets the runtime in tests where it matters.
- https://github.com/elastic/elastic-agent/pull/10537/commits/9d8cc7a3c6a6309dd2421e958cf5fd3339532d55 checks for beats metrics in a more generic way.
- https://github.com/elastic/elastic-agent/pull/10537/commits/1c1a0c3fc66d176b73aedb86828844e164962d2f checks for the component build hash in a way that covers beats receivers as well.

## Why is it important?

We want to enable beats receivers for self-monitoring by default. This PR makes the test adjustments separately to show they all pass for beats processes as well.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/9058

### DIFF
--- a/testing/integration/ess/package_version_test.go
+++ b/testing/integration/ess/package_version_test.go
@@ -194,7 +194,12 @@ func TestComponentBuildHashInDiagnostics(t *testing.T) {
 	require.Lenf(t, splits, 2,
 		"expected split of beats output version to be split into 2, it was split into %q",
 		strings.Join(splits, "|"))
-	wantBuildHash := splits[0]
+	filebeatBuildHash := splits[0]
+
+	// get the elastic-agent build hash
+	agentVersion, err := f.ExecVersion(ctx)
+	require.NoError(t, err, "failed to get agent version")
+	agentCommit := agentVersion.Binary.Commit
 
 	diagZip, err := f.ExecDiagnostics(ctx)
 	require.NoError(t, err, "failed collecting diagnostics")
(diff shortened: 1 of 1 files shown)


testing/integration/ess/metrics_monitoring_test.go
```diff
@@ -172,8 +172,7 @@ func (runner *MetricsRunner) TestBeatsMetrics() {
 			query = genESQuery(agentStatus.Info.ID,
 				[][]string{
 					{"match", "component.id", cid},
-					{"exists", "field", "system.process.cpu.total.value"},
-					{"exists", "field", "system.process.memory.size"},
```
cmacknz, line 176: Why did these have to change to a less specific field? I wouldn't have expected anything related to `system.process` to change with the switch to beats receivers?

swiatekm: We don't send these metrics for beats receivers, we specifically drop them because they're misleading and lead to problems in the integration. This was always the case ever since we added monitoring for them.

cmacknz: OK yes we were adding these via renaming I thought these were actually coming from a system/metrics input.

PR #10538, opened 2025-10-14
[agentprovider] Used passed context to detect collector shutdown

VihasMakwana:
## What does this PR do?

This PR fixes a potential race condition where we **might** call `watcher` even after collector has shutdown. Use `passedCtx` to detect the shutdown.

In the long term, we don't need this because we'll default to subprocess execution mode. But it's still blocking 9.1 PRs  :(

## Why is this important?

The `TestOTelManager_Run` is consistently failing on my 9.1 backport due to this.


VihasMakwana:
I can confirm that the fix works on [backport](https://github.com/elastic/elastic-agent/pull/10201): 

https://buildkite.com/elastic/elastic-agent/builds/28620

pkoutsovasilis (approved):

PR #10539, opened 2025-10-14
Improve docs automation for Gateway configs and deprecated components

theletterf:
This improves the existing docs automation to:

- Update the Gateway Collector configs tables automatically with new feature versions rows when detected, and updating the patch versions for existing feature versions.
- It adds a "deprecated" metadata file, owned by Ingest leads, to mark deprecated Collector components and update the table.

Fixes https://github.com/elastic/opentelemetry-dev/issues/1102

### DIFF
--- a/docs/scripts/update-docs/update-components-docs.py
+++ b/docs/scripts/update-docs/update-components-docs.py
@@ -106,7 +106,9 @@ def get_deprecated_components(version='main'):
         
     try:
         data = yaml.safe_load(content)
-        return data.get('deprecated', [])
+        deprecated = data.get('deprecated', [])
+        # Handle case where 'deprecated:' exists but has no items (returns None)
+        return deprecated if deprecated is not None else []
     except yaml.YAMLError as e:
         print(f"Warning: Error parsing deprecated-components.yaml from tag {version_tag}: {e}")
         return []
--- a/docs/scripts/update-docs/update-components-docs.py
+++ b/docs/scripts/update-docs/update-components-docs.py
@@ -28,6 +28,7 @@ GATEWAY_8X_TAG = 'edot-gateway-8x-table'
 EDOT_COLLECTOR_DIR = '../../../docs/reference/edot-collector'
 TEMPLATE_COLLECTOR_COMPONENTS_TABLE = 'templates/components-table.jinja2'
 TEMPLATE_COLLECTOR_OCB_FILE = 'templates/ocb.jinja2'
+TEMPLATE_GATEWAY_TABLE = 'templates/gateway-table.jinja2'
 COMPONENT_DOCS_YAML = '../../../docs/reference/edot-collector/component-docs.yml'
 DEFAULT_CONFIG_FILE = '../../../docs/reference/edot-collector/config/default-config-standalone.md'
 DEPRECATED_COMPONENTS_YAML = '../../../internal/pkg/otel/deprecated-components.yaml'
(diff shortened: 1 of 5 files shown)


theletterf:
@colleenmcginnis This adds one more file to the scope of automation in this repo. We can update the registry now or later.

docs/scripts/update-docs/update-components-docs.py
```diff
@@ -269,6 +304,194 @@ def get_component_docs_mapping(source_file):
...
+                })
+                print(f"  New version found: {version_str} (tag: {version_info['tag']})")
+            else:
+                print(f"  Skipping {version_str}: gateway.yml not found at {version_info['tag']}")
+    
+    if not new_rows and not updated_rows:
+        print(f"  No updates needed for {major_version}.x")
+        return content
+    
+    # Update the table content
```
AlexanderWert: WDYT about using a jinja template here, instead of replacing the table in code here?

Once we retrieved all the versions to be generated (i.e. old ones and new ones), wouldn't it be easier to maintain the generation logic in a template?

docs/scripts/update-docs/update-components-docs.py
```diff
@@ -269,6 +304,194 @@ def get_component_docs_mapping(source_file):
...
+    Args:
+        content: The file content to update
+        tag: The marker tag for the table section
+        major_version: The major version number (8 or 9)
+    
+    Returns:
+        Updated file content
+    """
+    # Parse existing rows in the table (with their current tags)
+    existing_rows = parse_gateway_table(content, tag)
```
AlexanderWert: why do we need to parse the existing table?

Who about having a function `get_all_minor_versions_above` (similar to `get_latest_minor_versions`) that would return all minor versions for a major above a specified (hardcoded) minor version. That would allow to consistently generate all of the table in the same way and not requiring to parse the existing table. Would maybe simplify the code here?

docs/scripts/update-docs/update-components-docs.py
```diff
@@ -161,11 +189,18 @@ def get_otel_components(version='main', component_docs_mapping=None):
     for comp in otel_components:
         # Extract the component name without the suffix (e.g., 'filelogreceiver' from 'filelogreceiver ')
         comp_name = comp['name'].strip()
+        
+        # Check if this component is deprecated (takes precedence)
+        if comp_name in deprecated_components:
+            comp['support_status'] = 'Deprecated'
+            comp['is_deprecated'] = True
```
AlexanderWert: this is not being used anywhere, yet, right?
i.e. would need to update https://github.com/elastic/elastic-agent/blob/main/docs/scripts/update-docs/templates/components-table.jinja2 as well to mark the deprecated components in the rendered table

docs/scripts/update-docs/update-components-docs.py
theletterf: Good idea!

docs/scripts/update-docs/update-components-docs.py
theletterf: +1

theletterf:
@AlexanderWert Refactored to use Jinja2 templates!

alexandra5000 (approved):

AlexanderWert (approved):

pchila (approved):
It's an OTel docs-only PR.
Approving only to unblock the PR but this should have no impact on elastic-agent control plane files

PR #10544, opened 2025-10-14
Integration test to check data re-ingest when switching runtimes

### LINKED ISSUE #8207: Ensure switching to the otel runtime does not result in input data duplication.

leehinman:
## What does this PR do?

Adds integration test that checks if switching to otel runtime for
monitoring causes the elastic-agent log files to be re-ingested.  Also
check to make sure that restarting the agent with otel runtime does
not result in re-ingesting the log files.

## Why is it important?

re-ingestion is wasteful and if restarts caused state to be lost we
could loose data.

## Checklist

## Disruptive User Impact

None.  Integration test only

```shell
mage -v integration:single TestMonitoringNoDuplicates
```

## Related issues

- Closes #8207

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -1258,13 +1258,14 @@ func setStrictMapping(client *elasticsearch.Client, index string) error {
 // re-ingest logs.
 //
 // Flow
-// 1) Create policy in Kibana with just monitoring and "process" runtime
-// 2) Install and Enroll
-// 3) restart agent, to roll log files
-// 4) Switch to monitoring "otel" runtime
-// 5) restart agent 3 times, making sure healthy between restarts
-// 6) query ES for monitoring logs with aggregation on fingerprint and line number, ideally 0 duplicates but possible to have a small number
-// 7) uninstall
+//  1. Create policy in Kibana with just monitoring and "process" runtime
+//  2. Install and Enroll
+//  3. Switch to monitoring "otel" runtime
+//  4. restart agent 3 times, making sure healthy between restarts
+//  5. switch back to "process" runtime
+//  6. query ES for monitoring logs with aggregation on fingerprint and line number,
+//     ideally 0 duplicates but possible to have a small number
+//  7. uninstall
 func TestMonitoringNoDuplicates(t *testing.T) {
 	info := define.Require(t, define.Requirements{
 		Group: integration.Default,
@@ -1278,7 +1279,9 @@ func TestMonitoringNoDuplicates(t *testing.T) {
 		Sudo:  true,
 	})
 
-	ctx, cancel := testcontext.WithDeadline(t, context.Background(), time.Now().Add(5*time.Minute))
+	ctx, cancel := testcontext.WithDeadline(t,
+		context.Background(),
+		time.Now().Add(5*time.Minute))
 	t.Cleanup(cancel)
 
 	policyName := fmt.Sprintf("%s-%s", t.Name(), uuid.Must(uuid.NewV4()).String())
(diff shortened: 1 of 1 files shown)


testing/integration/ess/beat_receivers_test.go
```diff
@@ -1254,3 +1250,280 @@ func setStrictMapping(client *elasticsearch.Client, index string) error {
...
+	err = json.Unmarshal(resultBuf, &aggResults)
+	aggs, ok := aggResults["aggregations"].(map[string]any)
+	require.Truef(t, ok, "'aggregations' wasn't a map[string]any, result was %s", string(resultBuf))
+	dups, ok := aggs["duplicates"].(map[string]any)
+	require.Truef(t, ok, "'duplicates' wasn't a map[string]any, result was %s", string(resultBuf))
+	buckets, ok := dups["buckets"].([]any)
+	require.Truef(t, ok, "'buckets' wasn't a []any, result was %s", string(resultBuf))
+
+	// It is possible to have a batch in flight not get acked
+	// before elastic-agent shutsdown. That can lead to
```
ycombinator, line 1520: Is there no way to wait (for a reasonable amount of time, maybe a few seconds?) on that last batch to be indexed before restarting Agent in the test?  That would make the test more deterministic IMO.

leehinman: The test is checking to see if the "last" log message from startup has been indexed.  That is part of what what the `healthcheck` function is doing.  This does make the test much more consistent.  This is also why the test disables monitoring and then restarts before enabling it with otel mode.   But I haven't found a way to make this totally deterministic.

There is always a window where elastic-agent can produce a log, elastic-agent or agentbeat "sees" the log and tries to send it, then we kill elastic-agent.  In that scenario the registry on disk will always be behind what elasticsearch has indexed and we will produce a duplicate on the next restart.  But this should be a small number of duplicates and should not be the "entire file" which is what we are trying to make sure isn't happening.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -1254,3 +1250,280 @@ func setStrictMapping(client *elasticsearch.Client, index string) error {
...
+	// turn off monitoring
+	policy.Agent.Monitoring["enabled"] = false
+	updatedPolicyBytes, err = yaml.Marshal(policy)
+	require.NoErrorf(t, err, "error marshalling policy, struct was %v", policy)
+	err = fut.Configure(ctx, updatedPolicyBytes)
+	require.NoError(t, err)
+
+	// restart to make sure beat processes have exited
+	restartBytes, err = fut.Exec(ctx, []string{"restart"})
+	require.NoErrorf(t, err, "Restart error: %s, output was: %s", err, string(restartBytes))
```
swiatekm: This is probably fine, but the more idiomatic way would be to update the policy in Kibana and wait until agent reports it as applied. Any specific reason to reload the configuration locally instead?

testing/integration/ess/beat_receivers_test.go
swiatekm: We don't kill elastic-agent though, we restart it in an orderly way. In principle, there shouldn't be any duplicates in this case, right?

leehinman: > In principle, there shouldn't be any duplicates in this case, right?

I started out with this assumption too.  But, we don't have a "only once" delivery guarantee, we have a "at least once" delivery guarantee.   Because of this having occasional duplicates is "normal behavior", but re-ingesting the whole file would not be normal behavior.  Which is why I switched to accepting a "small percentage" of duplicates.  Even if we shut down "in an orderly way", there are still timeouts that could cause us to exit without receiving a reply from elasticsearch even though the elasticsearch bulk request eventually succeeds.  And these timeouts are necessary because we don't want to prevent shutting down due to network issues etc.

testing/integration/ess/beat_receivers_test.go
leehinman: That's a good idea.  Do we have an example of using the API to change the runtime?

leehinman: @swiatekm updated to using API to make changes, plus enrolled in Fleet.

leehinman:
Putting back in draft until https://github.com/elastic/elastic-agent/issues/10645 is fixed.

swiatekm:
@leehinman I think we can close this. I added this test in #10857.

PR #10564, opened 2025-10-15
[9.1][tests] - Fix flaky TestClientWithCertificate

VihasMakwana:
Fix TestClientWithCertificate, which is flaky on 9.1. 
This PR updates the test case to match main

### DIFF
--- a/internal/pkg/remote/client_fips_test.go
+++ b/internal/pkg/remote/client_fips_test.go
@@ -189,7 +189,7 @@ func TestClientWithCertificate(t *testing.T) {
 			require.EventuallyWithT(
 				t,
 				func(collect *assert.CollectT) {
-					return assert.Contains(collect, serverLog.String(), test.expectedServerLog)
+					assert.Contains(collect, serverLog.String(), test.expectedServerLog)
 				},
 				100*time.Millisecond, 10*time.Millisecond,
 			)


VihasMakwana:
We'll probably need this for 8.19 as well.

pchila (commented):
IIUC this is a manual backport of https://github.com/elastic/elastic-agent/pull/10019/files#diff-5be2ee845b3042500dbb180c173cd24f63ba730b6be3c7736d988ffe24ca975b
only for the test modifications, correct ?

VihasMakwana:
> IIUC this is a manual backport of https://github.com/elastic/elastic-agent/pull/10019/files#diff-5be2ee845b3042500dbb180c173cd24f63ba730b6be3c7736d988ffe24ca975b only for the test modifications, correct ?

Yes Paolo. Correct.

pchila (commented):
Not sure how this passed the first round of code reviews: the `assert.Contains(t...)` would fail the test (not immediately) even with a single assertion failure, which is the opposite of what `require.Eventually` is used for.
The fact that the fix for this on main has been embedded in a bump [PR](https://github.com/elastic/elastic-agent/pull/10019) made it so that it wouldn't be automatically backported.

Changes look ok, waiting for CI for approval.
@VihasMakwana could you check if this change needs to be backported also on other active branches ?

VihasMakwana:
> @VihasMakwana could you check if this change needs to be backported also on other active branches ?

Out of active branches, this change exists on 9.2.
We need to backport it to 9.1 and 8.19.

We don't have any more releases for 9.0, so we can skip that.

VihasMakwana:
@pchila Could you please approve?

ycombinator (approved):

PR #10631, opened 2025-10-16
Return an error when a variable doesn't exist. Add new `|?` syntax.

### LINKED ISSUE #2261: Only running one instance of httpjson when there should be two
[discarded-input.yml.txt](https://github.com/elastic/elastic-agent/files/10725753/discarded-input.yml.txt)

The attached agent policy should start two instances of the httpjson input, but only ends up starting the input with ID `httpjson-ti_abusech-5c2d6965-8136-4237-8c08-20a54da01a13`.

Looking the diagnostics the second input with ID `httpjson-ti_misp-911740ef-8690-424f-8e0c-e7a540a178d6` is present in the pre-config.yaml but removed in the `computed-config.yaml` suggesting that the agent has decided it does not need to run it without explaining why.

blakerouse:
## What does this PR do?

Now when a variable cannot be substituted it will return an error instead of silently removing the input. Adds new syntax for `|?` at the end of a variable chain to allow the previous behavior.

## Why is it important?

Previously, when a variable didn't exist, the entire input would be silently removed without any error or log message. This change fixes that behavior by throwing an error when a variable cannot be substituted in an input.

## Checklist

## Disruptive User Impact

This will require users to update the policy or configuration if they want the previous behavior.

1. Add a missing variable into your policy and configuration and then run the Elastic Agent. You will see the following error:

```
{"log.level":"error","@timestamp":"2025-10-16T10:14:27.653-0400","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator.(*Coordinator).runLoopIteration","file.name":"coordinator/coordinator.go","file.line":1526},"message":"applying new policy: generating component model: rendering inputs failed: no matching vars: ${env.MISSING_VAR}","log":{"source":"elastic-agent"},"ecs.version":"1.6.0"}
```

2. Add the `|?` at the end and run it again, see that no error is logged and that the input is missing.

## Related issues

- Closes #2261


blakerouse:
This has an issue with kubernetes and hints. I am going to have to look into that more and see if this approach will even work with that, might not be possible.

PR #10633, opened 2025-10-16
Reduce PR CI runtime

pchila:
## What does this PR do?

This PR tries to bring down the total run time for elastic-agent CI on buildkite when running on PRs.
The changes contained here are mostly low-risk:
- Right-size BK agent VMs for the workload type (see unit tests and packaging steps machine changes
- Remove unnecessary wait statement so steps start as soon as possible
- Pass `test.short` flag when running integration tests on PR, allowing test code to skip longer tests
- Reduce the number of k8s version we test and parallelize k8s tests by image variant

## Why is it important?
This is important to reduce the turnaround time for PRs.

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/testing/integration/ess/upgrade_standalone_flavors_test.go
+++ b/testing/integration/ess/upgrade_standalone_flavors_test.go
@@ -9,6 +9,7 @@ package ess
 import (
 	"context"
 	"fmt"
+	"slices"
 	"testing"
 	"time"
 
@@ -30,10 +31,6 @@ func TestStandaloneUpgrade_Flavor_Basic(t *testing.T) {
 		Sudo:  true,  // requires Agent installation
 	})
 
-	if testing.Short() {
-		t.Skip("skipping test in short mode")
-	}
-
 	minVersion := upgradetest.Version_9_0_0_SNAPSHOT
 	currentVersion, err := version.ParseVersion(define.Version())
 	require.NoError(t, err)
@@ -44,6 +41,10 @@ func TestStandaloneUpgrade_Flavor_Basic(t *testing.T) {
 
 	versionList, err := upgradetest.GetUpgradableVersions()
 	require.NoError(t, err)
+
+	// sort versionList in ascending order
+	slices.SortFunc(versionList, version.CompareNillableParsedSemVer)
+
 	endVersion, err := version.ParseVersion(define.Version())
 	require.NoError(t, err)
 
@@ -82,6 +83,11 @@ func TestStandaloneUpgrade_Flavor_Basic(t *testing.T) {
 				testStandaloneUpgradeFlavorCheck(t, startVersion, define.Version(), true, false, checkFn)
 			})
 		}
+
+		if testing.Short() {
+			t.Log("skipping further tests in short mode")
+			break
+		}
 	}
 }
 
@@ -91,10 +97,6 @@ func TestStandaloneUpgrade_Flavor_Servers(t *testing.T) {
 		Local: false, // requires Agent installation
 		Sudo:  true,  // requires Agent installation
 	})
-
-	if testing.Short() {
-		t.Skip("skipping test in short mode")
-	}
 
 	minVersion := upgradetest.Version_9_0_0_SNAPSHOT
 	currentVersion, err := version.ParseVersion(define.Version())
@@ -106,6 +108,10 @@ func TestStandaloneUpgrade_Flavor_Servers(t *testing.T) {
 
 	versionList, err := upgradetest.GetUpgradableVersions()
 	require.NoError(t, err)
+
+	// sort versionList in ascending order
+	slices.SortFunc(versionList, version.CompareNillableParsedSemVer)
+
 	endVersion, err := version.ParseVersion(define.Version())
 	require.NoError(t, err)
 
@@ -143,6 +149,11 @@ func TestStandaloneUpgrade_Flavor_Servers(t *testing.T) {
 				testStandaloneUpgradeFlavorCheck(t, startVersion, define.Version(), true, true, checkFn)
 			})
 		}
+
+		if testing.Short() {
+			t.Log("skipping further tests in short mode")
+			break
+		}
 	}
 }
 
@@ -152,10 +163,6 @@ func TestStandaloneUpgrade_Flavor_UpgradeFromUnflavored(t *testing.T) {
 		Local: false, // requires Agent installation
 		Sudo:  true,  // requires Agent installation
 	})
-
-	if testing.Short() {
-		t.Skip("skipping test in short mode")
-	}
 
 	minVersion := upgradetest.Version_9_0_0_SNAPSHOT
 	currentVersion, err := version.ParseVersion(define.Version())
@@ -167,6 +174,10 @@ func TestStandaloneUpgrade_Flavor_UpgradeFromUnflavored(t *testing.T) {
 
 	versionList, err := upgradetest.GetUpgradableVersions()
 	require.NoError(t, err)
+
+	// sort versionList in ascending order
+	slices.SortFunc(versionList, version.CompareNillableParsedSemVer)
+
 	endVersion, err := version.ParseVersion(define.Version())
 	require.NoError(t, err)
 
@@ -203,6 +214,11 @@ func TestStandaloneUpgrade_Flavor_UpgradeFromUnflavored(t *testing.T) {
 			t.Run(fmt.Sprintf("Upgrade %s to %s (unprivileged)", startVersion, define.Version()), func(t *testing.T) {
 				testStandaloneUpgradeFlavorCheck(t, startVersion, define.Version(), true, false, checkFn)
 			})
+		}
+
+		if testing.Short() {
+			t.Log("skipping further tests in short mode")
+			break
 		}
 	}
 }
--- a/pkg/version/version_parser.go
+++ b/pkg/version/version_parser.go
@@ -320,3 +320,28 @@ type SortableParsedVersions []*ParsedSemVer
 func (spv SortableParsedVersions) Len() int           { return len(spv) }
 func (spv SortableParsedVersions) Swap(i, j int)      { spv[i], spv[j] = spv[j], spv[i] }
 func (spv SortableParsedVersions) Less(i, j int) bool { return spv[i].Less(*spv[j]) }
+
+// CompareNillableParsedSemVer is a comparison function for sorting *ParsedSemVer in ascending order
+func CompareNillableParsedSemVer(a, b *ParsedSemVer) int {
+	// nils are always smaller
+	if a == nil {
+		return -1
+	}
+	if b == nil {
+		return 1
+	}
+	return CompareParsedSemVer(*a, *b)
+}
+
+// CompareParsedSemVer is a comparison function for sorting ParsedSemVer in ascending order
+func CompareParsedSemVer(a, b ParsedSemVer) int {
+	if a.Less(b) {
+		return -1
+	}
+
+	if b.Less(a) {
+		return 1
+	}
+
+	return 0
+}


pchila:
buildkite test this

pchila:
buildkite test this

pchila:
buildkite test this

pchila:
buildkite test this

.buildkite/integration.pipeline.yml
```diff
@@ -57,11 +57,11 @@ steps:
           image: "${IMAGE_UBUNTU_2204_X86_64}"
 
       # Packaging linux/arm64
```
v1v: and windows too?

.buildkite/pipeline.yml
```diff
@@ -37,6 +37,7 @@ steps:
         agents:
           provider: "gcp"
           image: "${IMAGE_UBUNTU_2204_X86_64}"
+          machineType: "n2-highcpu-8"
```
v1v, line 40: do we know what's the % of improvement when using this machine type?

.buildkite/integration.pipeline.yml
```diff
@@ -186,16 +154,12 @@ steps:
...
         agents:
           provider: "aws"
-          instanceType: "c6g.4xlarge"
+          instanceType: "c6g.2xlarge"
           diskSizeGb: 200
           image: "${IMAGE_UBUNTU_2204_ARM_64}"
 
   - label: "Triggering Integration tests"
-    depends_on:
-      - int-packaging
```
v1v, line 195: neat! 🙇

.buildkite/scripts/steps/integration_tests_tf.sh
```diff
@@ -66,6 +63,9 @@ if [[ "${GROUP_NAME}" == "kubernetes" ]]; then
   source .buildkite/scripts/install-kubectl.sh
   .buildkite/scripts/buildkite-k8s-integration-tests.sh $@
 else
+  echo "~~~ Building test binaries"
+  mage build:testBinaries
+
```
v1v, line 69: IIUC, k8s does not need the binaries at all, it uses a different type of artefacts instead, shall we add a comment about it to help us in the future, or is it trivial?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -524,28 +524,35 @@ steps:
...
-              - v1.30.0
-              - v1.31.0
+#              - v1.27.16
+#              - v1.28.15
+#              - v1.29.14
+#              - v1.30.0
+#              - v1.31.0
               - v1.32.0
               - v1.33.0
+              - v1.34.0
```
swiatekm: I think it might be more practical to only run on the first and last supported versions, so 1.27 and 1.34 in our case. This is what a lot of Kubernetes operators do for their tests, as it basically guarantees compatibility for all the in-between versions outside of bugs in Kubernetes itself. Then we can run on all supported versions on `main`. WDYT?

.buildkite/scripts/steps/integration_tests_tf.sh
pchila: Correct. This saves a couple of minutes on the k8s runners. I will add a comment for that
Fixed in 87c228d8bb

.buildkite/pipeline.yml
pchila: Comparing [main](https://buildkite.com/elastic/elastic-agent/builds/29488#019a29fb-781a-4193-8d7e-79769f71ebd5) (which runs on `n1-standard-4`) and [this PR](https://buildkite.com/elastic/elastic-agent/builds/29491#019a2a32-8f38-4193-9029-cfe1d4594224) we passed from ~43 minutes to ~25 minutes so ~40% improvement for ubuntu 22.04.
The `highcpu` variant is due to the fact that the extra RAM coming with doubling the cores was mostly unused.

For windows 2022 on the same builds we went from 1h34m to 49m so ~ 48% improvement. In this case we need the `n2-standard-8` to avoid ending up out of memory when parallelizing unit test binary compiling and linking  on windows machines.

.buildkite/integration.pipeline.yml
pchila: yes, forgot to change the comments.
Addressed in 19475e242d

.buildkite/integration.pipeline.yml
pchila: This was an easy one :wink:

.buildkite/bk.integration.pipeline.yml
pchila: The reason for removing anything < 1.32 is that those versions already [reached EOL](https://kubernetes.io/releases/)

The number of k8s versions we test does not really influence the overall duration of CI runs, the real gain here is the parallelization over the docker variant.

As for which versions to select for PRs the problem is that the pipeline is in control of that, so excluding some version from the matrix is a little trickier than a simple filter on a list -> https://buildkite.com/docs/pipelines/configure/workflows/build-matrix#excluding-combinations-from-the-build-matrix

A solution would be to duplicate the step and add a condition to run one or the other in case of PR or not but I don't know if it's worth the extra effort

.buildkite/pipeline.yml
v1v: That's fantastic news @pchila , thanks so much for providing those details and making this change!

.buildkite/bk.integration.pipeline.yml
pchila: Separated the PR and release branch k8s integration test steps in fa5fef12576094080951e7441e04774bde9ace17 : in PR runs only the min and max k8s version is tested while the full set is used for branch builds.

For the full set of k8s versions, a parallelization of docker variants is not possible as the resulting number of job would be greater than 70 (current maximum supported by `matrix`)

.buildkite/bk.integration.pipeline.yml
```diff
@@ -1,5 +1,16 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
+# bogus key to hold the list of k8s version to test against
+k8s_test_versions: &k8s_test_versions
```
v1v: I didn't know anchors were available also for other pieces; somehow, I thought they were only available for BK plugins:
- https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins

That's good ot know

pchila:
Failure in build https://buildkite.com/elastic/elastic-agent/builds/29527 seems unrelated to the changes in the pipeline, all non FIPS-related steps are successful

testing/integration/ess/upgrade_standalone_flavors_test.go
```diff
@@ -30,6 +30,10 @@ func TestStandaloneUpgrade_Flavor_Basic(t *testing.T) {
 		Sudo:  true,  // requires Agent installation
 	})
 
+	if testing.Short() {
```
swiatekm: What's the reason we're marking these particular tests as long? Do they take a lot of time?

pchila: About 1h30m overall for the group

blakerouse (approved):
Thanks for this change. Looks good.

swiatekm (dismissed):

pchila:
Last run https://buildkite.com/elastic/elastic-agent/builds/29601 took 1h54m because of `Win2022:sudo:fleet` failure due to `TestFleetManagedUpgradeRollbackOnRestarts/downgrade_from_current_version_to_previous_minor` :disappointed:

.buildkite/bk.integration.pipeline.yml
```diff
@@ -43,19 +43,30 @@ common:
...
+    - v1.32.0
+    - v1.33.0
+    - &k8s_max_test_version v1.34.0
 
 steps:
   - label: "Custom ECH Testing"
     key: integration-tests-ech
     depends_on:
-      - packaging-ubuntu-x86-64
+      - packaging-containers-x86-64
```
ycombinator: Since you've done similar naming cleanup for `packaging-ubuntu-x86-64-fips` -> `packaging-amd64-fips` elsewhere in this PR, could we also clean this up (and the corresponding `key`, of course) to be `packaging-containers-amd64`?

.buildkite/bk.integration.pipeline.yml
```diff
@@ -534,18 +547,54 @@ steps:
...
             version:
-              - v1.27.16
-              - v1.28.15
-              - v1.29.14
-              - v1.30.0
-              - v1.31.0
-              - v1.32.0
-              - v1.33.0
+              *k8s_test_versions
+      - label: ":open-pull-request: :kubernetes: {{matrix.version}}:amd64:{{matrix.variants}}"
```
ycombinator, line 564: The `env`, `command`, `artifact_paths`, `retry`, `agents`, `plugins`, and `matrix.setup.variants` blocks are identical to the previous non-PR step.  Could we use YAML references to reduce the duplication to avoid the risk of drift?

.buildkite/bk.integration.pipeline.yml
pchila: IIUC you want me to change `x86-64` ->`amd64` in the step keys ?

.buildkite/bk.integration.pipeline.yml
pchila: the `matrix.setup.variants` are not the same as we cannot parallelize in the same way when testing all k8s versions.

I am not sure that YAML anchors are supported for all the keys you listed, BK docs are very [plugin-focused](https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins)

.buildkite/bk.integration.pipeline.yml
pchila: fb9f3c1c956cfb59bce1237b044508aa4fdaa6fc

.buildkite/bk.integration.pipeline.yml
pchila: Not sure that the DRY version is more ergonomic than a straight duplication of the entire step :thinking: 
c6f1ba495b3334c72bd07ced026ee9d74e22276b

testing/integration/ess/upgrade_standalone_flavors_test.go
cmacknz: Every 9.x agent has a flavor so I am a bit hesitant to say this should be skipped completely.

Looking at an example test report [here](https://web.buildkiteartifacts.com/e0f3970e-3a75-4621-919f-e6c773e2bb12/018f7c56-3a0b-46d9-a3d0-bd009d6414b7/019a281e-7a7d-437e-9d13-a0c87dbc7ecc/019a2840-4dfb-4146-b587-774e770d9cdb/build/TEST-report.html?jwt=eyJhbGciOiJIUzM4NCJ9.eyJzdWIiOiI2MjM2NzoxNDUxNTMiLCJhdWQiOiJ3ZWIuYnVpbGRraXRlYXJ0aWZhY3RzLmNvbSIsInBhdGgiOiJlMGYzOTcwZS0zYTc1LTQ2MjEtOTE5Zi1lNmM3NzNlMmJiMTIvMDE4ZjdjNTYtM2EwYi00NmQ5LWEzZDAtYmQwMDlkNjQxNGI3LzAxOWEyODFlLTdhN2QtNDM3ZS05ZDEzLWEwYzg3ZGJjN2VjYy8wMTlhMjg0MC00ZGZiLTQxNDYtYjU4Ny03NzRlNzcwZDljZGIiLCJpc3MiOiJidWlsZGtpdGUuY29tIiwiZXhwIjoxNzYxNzcxNzAwLCJpYXQiOjE3NjE3NzExMDB9.t-MTaLvzKwCwMIdAH_FKpZdnt5DE13Sg12X8nD5aJFOD-q5wXoop65sZMPU3DOQ4) this is doing an exhaustive test of version compatibility.

In PRs could we have the `short` flag only test the oldest compatible version and the current build instead?

We perhaps need to use that strategy for all upgrade tests since most problems are oldest to newest and not the versions in between.

pchila: dec760a583e802717e49c8c0118240d9d0eea7c1

Issue #10593, opened 2025-10-15
[Helm] Support all options for Elastic Agent Standalone Outputs


btrieger:
**Describe the enhancement:**
Currently the Elastic Agent Helm chart only supports Elasticsearch type outputs. Ideally it should support all 3 output types(elasticsearch, kafka, and logstash).  Additionally it does not support all of the possible configuration options for Elasticsearch such as presets to allow tuning of the output. 

**Describe a specific use case for the enhancement or feature:**
Elastic Agent needs to send data to kafka or logstash. Additionally Elastic Agent has performance issues and needs further tuning to optimize EPS.

**What is the definition of done?**
Support for all elastic agent output options with example values.yaml files for all output options. This should also include the rendered output for these examples.

PR #10644, opened 2025-10-16
[Helm] Add Logstash Output to Elastic Agent Standalone

btrieger:
## What does this PR do?

This PR adds in the Logstash output for elastic agents to the standalone helm chart. 

## Why is it important?

Without this, standalone agents cannot connect to Logstash clusters. This makes the chart unusable for end users who have logstash in between agent and elasticsearch. 

## Checklist

- [X] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [X] I have made corresponding changes to the documentation
- [X] I have made corresponding change to the default configuration files

## Disruptive User Impact

This should have no user impact and be backward compatable.

Follow steps in examples/kubernetes-logstash-output

## Related issues

- Relates #10593


pierrehilbert:
Hey @btrieger 
Thanks for your contribution.
Could you please add a changelog fragment?

btrieger:
@pierrehilbert added the changelog fragment

btrieger:
I added the remaining ssl flags and they should work with any of the outputs. I didn't provide an example of using them as I don't have a functioning logstash cluster that needs them but I did validate that they are being set as expected with a template command. Was thinking maybe I could document all the possible input variables outside the values schema json for every day users. Would it be best to put this in the examples readme for logstash? 

On a side note the rest of the outputs should remove the fromYaml and toYaml around the the call to the SSLConfig rendering as it removes the quotes and in the case of ca fingerprints this could prove problematic.

deploy/helm/elastic-agent/templates/agent/k8s/_pod_template.yaml
```diff
@@ -149,6 +149,10 @@ template:
           {{- with ($presetVal).outputs }}
           {{- range $outputName, $outputVal := . -}}
           {{- (include (printf "elasticagent.output.%s.preset.envvars" ($outputVal).type) (list $ $outputName $outputVal)) | nindent 10 }}
+          {{- $outputSSLConfig := dig "ssl" dict $outputVal -}}
+          {{- if not (empty $outputSSLConfig) }}
+          {{- (include "elasticagent.output.render.sslconfig.envvars" (list $outputSSLConfig $outputName $agentName) | nindent 10) }}
+          {{- end }}
```
pkoutsovasilis, line 155: if already checks for empty values so this can be simplified like
```suggestion
          {{- $outputSSLConfig := dig "ssl" dict $outputVal -}}
          {{- if $outputSSLConfig }}
          {{- (include "elasticagent.output.render.sslconfig.envvars" (list $outputSSLConfig $outputName $agentName) | nindent 10) }}
          {{- end }}
```

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
```diff
@@ -76,6 +76,12 @@ Validate fleet configuration
 {{- range $idx, $certificateAuthorityConfig := $sslCertificateAuthoritiesConfig -}}
 {{- include "elasticagent.init.valueFrom" (list $ $certificateAuthorityConfig (printf "%s.ssl.certificate_authorities.%d" $outputName $idx)) -}}
 {{- end -}}
+{{- if not (  $outputVal | dig "ssl" "certificate" dict | empty )  }}
+{{- include "elasticagent.init.valueFrom" (list $ $outputVal.ssl.certificate (printf "%s.ssl.certificate" $outputName)) -}}
+{{- end -}}
```
pkoutsovasilis:
```suggestion
{{- if (dig "ssl" "certificate" dict $outputVal) }}
{{- include "elasticagent.init.valueFrom" (list $ $outputVal.ssl.certificate (printf "%s.ssl.certificate" $outputName)) -}}
{{- end -}}
```

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
```diff
@@ -76,6 +76,12 @@ Validate fleet configuration
...
+{{- if not (  $outputVal | dig "ssl" "certificate" dict | empty )  }}
+{{- include "elasticagent.init.valueFrom" (list $ $outputVal.ssl.key (printf "%s.ssl.key" $outputName)) -}}
+{{- end -}}
```
pkoutsovasilis: Something is off here
```suggestion
{{- if (dig "ssl" "key" dict $outputVal)  }}
{{- include "elasticagent.init.valueFrom" (list $ $outputVal.ssl.key (printf "%s.ssl.key" $outputName)) -}}
Comment view{{- end -}}
```

deploy/helm/elastic-agent/templates/agent/k8s/_secret.tpl
```diff
@@ -67,6 +67,18 @@
...
+    {{- .ssl.certificate.value | nindent 4 }}
+  {{- end }}
+  {{- if ne ( . | dig "ssl" "key" "value" "not_found") "not_found"  }}
+  {{ printf "%s%s" $idx ".ssl.key" }} : |-
+    {{- .ssl.key.value | nindent 4 }}
+  {{- end }}
+  {{- if ne ( . | dig "ssl" "key_passphrase" "value" "not_found") "not_found"  }}
+  {{ printf "%s%s" $idx ".ssl.key_passphrase" }} : |-
+    {{- .ssl.key_passphrase.value | nindent 4 }}
+  {{- end }}
```
pkoutsovasilis, line 81: since .ssl.{certificate, key, key_passphrase} are deriving from ValueFrom schema I would expect here to not use use printf to construct the key but utilise the `._key` as it is done above https://github.com/elastic/elastic-agent/pull/10644/files#diff-426bb707731f4bf25cef0bb9038b329d0c29324c31fac48ce8a99b978c8939c5R65

deploy/helm/elastic-agent/templates/agent/_outputs.tpl
```diff
@@ -127,21 +127,141 @@ namespace: {{.}}
...
+
+  {{- with index $outputVal "queue.mem.flush.timeout" }}
+  queue.mem.flush.timeout: {{. | quote}}
+  {{- end }}
+  {{- with index $outputVal "queue.mem.flush.min_events" }}
+  queue.mem.flush.min_events: {{.}}
+  {{- end }}
+  {{- with index $outputVal "queue.mem.events" }}
+  queue.mem.events: {{.}}
+  {{- end }}
```
pkoutsovasilis: I believe that there shouldn't be any fields with dot notation in them here. If a user sets through cli such values they will appear as proper yaml and then this here won't work

deploy/helm/elastic-agent/templates/agent/_outputs.tpl
```diff
@@ -127,21 +127,141 @@ namespace: {{.}}
...
+  {{- with $outputVal.bulk_max_size}}
+  bulk_max_size: {{.}}
+  {{- end }}
+  {{- with index $outputVal "backoff.max" }}
+  backoff.max: {{. | quote }}
+  {{- end }}
+  {{- with index $outputVal "backoff.init" }}
+  backoff.init: {{. | quote }}
+  {{- end }}
```
pkoutsovasilis: ditto

btrieger:
@pkoutsovasilis  will do I was tyring to keep it small but it seemed kind of unavoidable if I wanted to add all the variables available for logstash.

deploy/helm/elastic-agent/templates/agent/k8s/_secret.tpl
btrieger: I will try this again. For some reason ._key wasn't populating not sure if it was because I wasn't using a loop or because key, key_passphrase, and certificate are not arrays or lists.

deploy/helm/elastic-agent/templates/agent/_outputs.tpl
btrieger: Would it be best to build out the schema to make them all objects? I think it is best to keep variable names in line with what they are in our docs for clarity? You can still set these via the cli if you use quotes around the variable name. I have done it with other charts in the past. 

Speaking of keeping variable names in line. What are your thoughts on making a 2nd copy of certificate_authorities and verification_mode and deprecating the others. I would make a clean switch but don't want to break anyones chart that is using this one.

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
btrieger: is something off with the suggested change I see Comment view in clear test not sure if you meant to paste something?

deploy/helm/elastic-agent/templates/agent/_outputs.tpl
pkoutsovasilis: > Would it be best to build out the schema to make them all objects? I think it is best to keep variable names in line with what they are in our docs for clarity?

yes they are like that in our docs but elastic-agent does under the hood the splitting up of dot notation, I do believe that is more error prone to say to the user that a key has to wrapped across some quotes for it to find it's appropriate place in the values, thus I am inclining of saying to expand all of them as objects

> Speaking of keeping variable names in line. What are your thoughts on making a 2nd copy of certificate_authorities and verification_mode and deprecating the others. I would make a clean switch but don't want to break anyones chart that is using this one.

sure mark the existing ones as deprecated but they still have to render properly and after some versions these can go away

deploy/helm/elastic-agent/templates/agent/_helpers.tpl
pkoutsovasilis: something is off in the sense of here you were checking originally `"ssl" "certificate"` where IIUC you should be checking for `"ssl" "key"`

deploy/helm/elastic-agent/templates/agent/k8s/_secret.tpl
btrieger: Got this working not sure why my first try using _key was just not printing anything for the key.

deploy/helm/elastic-agent/templates/agent/_outputs.tpl
btrieger: I think doing the deprecation may be best in a separate PR otherwise this one will get event longer. I have made the other changes suggested.

swiatekm:
@btrieger is it possible to do a basic integration test in a Kubernetes cluster? We have tooling for setting up a cluster and a ESS stack, so we'd need to start logstash with the right configuration, have agent ship to logstash, and verify if we see the data in ES. Would that be a lot of work, in your view?

btrieger:
@swiatekm it shouldn't be to bad. Where is the tooling located? I have run it all locally and confirmed it is working so I have the k8s manifests

swiatekm:
@btrieger have a look at https://github.com/elastic/elastic-agent/blob/fd86b05cba86f33286e678937a4797bc3f5083c0/testing/integration/k8s/kubernetes_agent_standalone_test.go#L208. You can even skip sending the data and just demonstrate that agent is healthy with a logstash output.

btrieger:
@swiatekm I added a logstash test. Let me know your thoughts.

testing/integration/k8s/kubernetes_agent_standalone_test.go
```diff
@@ -1178,6 +1242,164 @@ func k8sStepHelmDeploy(chartPath string, releaseName string, values map[string]a
...
+						}
+					}
+
+					// Alternative check: just verify the API is responding and has valid structure
+					if _, hasJVM := result["jvm"]; hasJVM {
+						if _, hasPipelines := result["pipelines"]; hasPipelines {
+							t.Logf("Logstash API is responding with pipeline data")
+							return true
+						}
+					}
```
swiatekm: Do we really need to do all this just to check Logstash health?

swiatekm (commented):
The test looks good to me, minus the rather involved health check. Does it really need to be so complicated?

testing/integration/k8s/kubernetes_agent_standalone_test.go
btrieger: Updated it to use the _health_report api which I was struggling to get working initially. Should be a lot cleaner.

btrieger:
> I still think there's something a bit funky with the logstash health check, but I don't want to hold this PR back, as the test does work and checks what it's meant to check. Everything else looks good to me. 👍

I initially just tried to look at the pod health but it was returning ready before the pipeline actually started. So I switched to an exec in and hit the health endpoint. We can certainly revisit this.

swiatekm:
@btrieger FYI, here's the actual test failure:

```
=== Failed
=== FAIL: testing/integration/k8s TestKubernetesAgentHelm/helm_standalone_agent_default_kubernetes_unprivileged_with_logstash_output (35.60s)
    kubernetes_agent_standalone_test.go:1242:
        	Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1765816838289595169/elastic/elastic-agent/testing/integration/k8s/kubernetes_agent_standalone_test.go:1242
        	            				/opt/buildkite-agent/builds/bk-agent-prod-gcp-1765816838289595169/elastic/elastic-agent/testing/integration/k8s/kubernetes_agent_standalone_test.go:823
        	            	values don't meet the specifications of the schema(s) in the following chart(s):
        	            	elastic-agent:
        	            	- at '/outputs/default': validation failed
        	            	  - at '/outputs/default/hosts': invalid jsonType [1]string
        	            	  - at '/outputs/default/ssl/certificateAuthorities': invalid jsonType []map[string]interface {}
        	            	  - at '/outputs/default': 'allOf' failed
        	            	    - at '/outputs/default': validation failed
        	            	      - at '/outputs/default/ssl/certificateAuthorities': invalid jsonType []map[string]interface {}
        	            	      - at '/outputs/default/hosts': invalid jsonType [1]string
        	Test:       	TestKubernetesAgentHelm/helm_standalone_agent_default_kubernetes_unprivileged_with_logstash_output
        	Messages:   	failed to install helm chart
```

swiatekm:
Still getting 

```
        	            	values don't meet the specifications of the schema(s) in the following chart(s):
        	            	elastic-agent:
        	            	- at '/outputs/default': validation failed
        	            	  - at '/outputs/default/ssl/certificateAuthorities': invalid jsonType []map[string]interface {}
        	            	  - at '/outputs/default/hosts': invalid jsonType []string
        	            	  - at '/outputs/default': 'allOf' failed
        	            	    - at '/outputs/default': validation failed
        	            	      - at '/outputs/default/hosts': invalid jsonType []string
        	            	      - at '/outputs/default/ssl/certificateAuthorities': invalid jsonType []map[string]interface {}
```

btrieger:
Yeah. I am confused what caused it. It was just a merge with main that
suddenly caused it and the code it is referencing all still looks correct.
Will take a look at what all the recent changes in main were.

swiatekm (approved):

Issue #7756, opened 2025-04-08
Elastic-Agent does not allow child process to clean up during shutdown


belimawr:
When the Elastic-Agent is shutting down it does not wait for its child process to gracefully exit, thus not allowing the to perform the necessary clean up, which can leave active resources after all process have exited (see related issues). This happens in most, if not all platforms.

## TL;DR
When starting child process, the Elastic-Agent configures them to be forcefully killed once it's own process exits, alongside with that the shutdown process for child process is fully asynchronous, thus even if the Elastic-Agent correctly signals child process to exit, they do not have enough time to shutdown because the Elastic-Agent process exits, which makes the OS to kill all of its child process.

## The long version
When creating child process the Elastic-Agent configures them to be killed once it exits, this is handled differently on each OS, then when exiting the Elastic-Agent asynchronously signals the child process to exit and do not wait for their process to return.

When a shutdown signal is received by the Elastic-Agent, it signals the components/child process to exit:

https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/manager.go#L784-L799

The `stoppedWg` created only waits for the components to be in a stopped state, not for their actual process to exit.
https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/manager.go#L845-L854

And the component gets removed from `m.current` when it starts its shutdown
https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/manager.go#L906-L916

Following the code we can see `s.shuttingDown` being set to `true` and afterwards the shutdown signal being sent to the process

https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/runtime.go#L179-L188
https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/command.go#L307-L315

The channel is read and the stop method called
https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/command.go#L167-L170

https://github.com/elastic/elastic-agent/blob/1e2a3872afd01de88cf1a29e9fdca3d170048246/pkg/component/runtime/command.go#L412-L440

[...]

PR #10650, opened 2025-10-17
fix: zombie processes during restart

pkoutsovasilis:
## What does this PR do?

This PR fixes zombie/defunct processes that are left behind when Elastic Agent re-executes itself during restart. The fix involves:

1. **Decreasing the EDOT collector shutdown timeout** from 30 seconds to 3 seconds to accommodate the default 5-second timeout of the coordinator shutdown timeout
   * **Adding a safety net** that waits an additional second after killing a process to ensure `Wait()` is called
3. **Improving graceful shutdown handling** in the EDOT collector subprocess manager to ensure proper process cleanup
4. **Adding debug logging** throughout the shutdown process to better trace subprocess termination
5. **Adding an integration test** that verifies no zombie processes are left behind after agent restart

### Root Cause
When the Elastic Agent re-executes itself during restart, the following sequence occurs:

1. If a subprocess (particularly the EDOT collector or command components) takes longer than the coordinator's 5-second shutdown timeout, the agent proceeds to `execve` itself
3. During `execve`, [all threads other than the calling thread are destroyed](https://man7.org/linux/man-pages/man2/execve.2.html)
4. This triggers the `PDeathSig` mechanism we enable for subprocesses
5. **However**, the parent process (pre-execve Elastic Agent) never reaps (waits for) the exit status of spawned subprocesses
6. Result: these subprocesses end up as **defunct/zombie processes**

### Why This Affects EDOT More Than Beats
Beats subprocesses typically terminate almost immediately (within the 5-second window), so they don't become zombies. However, the EDOT collector's shutdown time seemed to be affected by:
- Number of pipeline workers
- Elasticsearch exporter configuration

### Impact
- **Resource leaks**: Zombie processes consume PIDs and kernel memory
- **Operational issues**: Accumulation of zombies over multiple restarts
- **Config update delays**: EDOT subprocess restarts on every config change, and 20+ second shutdowns create significant latency

This fix ensures proper process cleanup regardless of shutdown duration while maintaining graceful termination when possible.

## Checklist

## Disruptive User Impact

Users may notice:
- Agent restarts take slightly longer (up to 35 seconds instead of 5 seconds in worst case)
- However, this ensures clean shutdowns and prevents zombie accumulation
- The tradeoff is worthwhile as zombie processes can cause operational issues over time

Run `TestMetricsMonitoringCorrectBinaries` integration test

## Related issues

- https://github.com/elastic/elastic-agent/issues/7756

### DIFF
--- a/internal/pkg/agent/application/coordinator/coordinator.go
+++ b/internal/pkg/agent/application/coordinator/coordinator.go
@@ -229,12 +229,12 @@ type VarsManager interface {
 type ComponentsModifier func(comps []component.Component, cfg map[string]interface{}) ([]component.Component, error)
 
 // managerShutdownTimeout is how long the coordinator will wait during shutdown
-// to receive termination states from its managers. As the default stop timeout
-// for components spawned as subprocesses is 30 seconds (process.DefaultConfig()),
-// we need to wait a bit longer to ensure that all managers have had time to terminate.
-// Note: if this timeout doesn't accommodate for the subprocess stop timeout,
-// the Wait of subprocess might never be called, and we may end up with Zombie processes
-// during restart in Unix systemss.
+// to receive termination states from its managers.
+// Note: The current timeout (5s) is shorter than the default stop timeout for
+// subprocess components (30s from process.DefaultConfig()). This means the
+// coordinator may not wait for the subprocesses to finish terminating, preventing
+// Wait() from being called on them. This will result in zombie processes
+// during restart on Unix systems.
 const managerShutdownTimeout = time.Second * 5
 
 type configReloader interface {


pkg/component/runtime/command.go
```diff
@@ -433,17 +434,20 @@ func (c *commandRuntime) stop(ctx context.Context) error {
...
+	c.log.Debugf("gracefully stopping pid %d", c.proc.PID)
 	return c.proc.Stop()
 }
 
 func (c *commandRuntime) startWatcher(info *process.Info, comm Communicator) {
 	go func() {
 		err := comm.WriteStartUpInfo(info.Stdin)
 		if err != nil {
-			_, _ = c.logErr.Write([]byte(fmt.Sprintf("Failed: failed to provide connection information to spawned pid '%d': %s", info.PID, err)))
+			_, _ = fmt.Fprintf(c.logErr, "Failed: failed to provide connection information to spawned pid '%d': %s", info.PID, err)
```
swiatekm, line 450: I think using `logErr` here might be wrong. That's the logger for messages written to standard output by the beat we're running. Agent should log to `c.log` always.

testing/integration/ess/metrics_monitoring_test.go
```diff
@@ -207,7 +211,39 @@ func (runner *MetricsRunner) TestBeatsMetrics() {
...
 		}
 		return true
-	}, time.Minute*10, time.Second*10, "could not fetch metrics for all known components in default install: %v", componentIds)
+	}, time.Minute*10, time.Second*10, "could not fetch metrics for edot collector")
+
+	if runtime.GOOS == "windows" {
+		return
+	}
+
+	// restart the agent to try and cause
```
swiatekm:
```suggestion
	// restart the agent to try and cause the otel collector subprocess to become defunct
```

?

testing/integration/ess/metrics_monitoring_test.go
```diff
@@ -207,7 +211,39 @@ func (runner *MetricsRunner) TestBeatsMetrics() {
...
+		if err != nil {
+			t.Logf("waiting for agent healthy: %s", err.Error())
+			return false
+		}
+		return true
+	}, 1*time.Minute, 1*time.Second)
+
+	procStats := process.Stats{
+		// filtering with '.*elastic-agent' or '^.*elastic-agent$' doesn't
+		// seem to work as expected, filtering is done in the for loop below
```
swiatekm: We're not actually filtering anything in the loop below.

pkg/component/runtime/command.go
pkoutsovasilis: @swiatekm this was under a commit to address linter issues, and clearly on the left it is `c.logErr` so I am not confident changing that here

testing/integration/ess/metrics_monitoring_test.go
pkoutsovasilis: changed in e2f8cbdef30fa6ad1a9ff1351a878fe33134f152

testing/integration/ess/metrics_monitoring_test.go
pkoutsovasilis: changed in e2f8cbdef30fa6ad1a9ff1351a878fe33134f152

pkg/component/runtime/command.go
swiatekm: Sure, but I think we should address it in a follow-up of some sort. Or at least figure out if it's supposed to be this way for some reason.

internal/pkg/agent/application/application.go
```diff
@@ -266,7 +266,7 @@ func New(
 		agentInfo,
 		cfg.Settings.Collector,
 		monitor.ComponentMonitoringConfig,
-		cfg.Settings.ProcessConfig.StopTimeout,
+		3*time.Second, // this needs to be shorter than 5 * time.Seconds (coordinator.managerShutdownTimeout) otherwise we might end up with defunct processes
```
michalpristas: worth to make it a const with a comment

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -309,19 +310,26 @@ func (s *procHandle) Stop(waitTime time.Duration) {
...
 
+	// since we are here this means that the process either got an error at stop or did not stop within the timeout,
+	// kill it and give one more mere second for the process wait to be called
+	_ = s.processInfo.Kill()
 	select {
-	case <-time.After(waitTime):
-		s.log.Warnf("timeout waiting (%s) for the supervised collector to stop, killing it", waitTime.String())
-		// our caller ctx is Done; kill the process just in case
-		_ = s.processInfo.Kill()
+	case <-time.After(1 * time.Second):
```
michalpristas, line 331: so worst case is waitTime + 1s. please update func docs

internal/pkg/agent/application/application.go
pkoutsovasilis: changed in f66af911dbef379e9f010cbabec81e16868ab943

internal/pkg/otel/manager/execution_subprocess.go
pkoutsovasilis: changed in f66af911dbef379e9f010cbabec81e16868ab943

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -229,7 +229,12 @@ type VarsManager interface {
 type ComponentsModifier func(comps []component.Component, cfg map[string]interface{}) ([]component.Component, error)
 
 // managerShutdownTimeout is how long the coordinator will wait during shutdown
-// to receive termination states from its managers.
+// to receive termination states from its managers. As the default stop timeout
+// for components spawned as subprocesses is 30 seconds (process.DefaultConfig()),
+// we need to wait a bit longer to ensure that all managers have had time to terminate.
+// Note: if this timeout doesn't accommodate for the subprocess stop timeout,
+// the Wait of subprocess might never be called, and we may end up with Zombie processes
+// during restart in Unix systemss.
```
swiatekm: This is out of date now, right? We're not waiting more than 30s anymore.

pkoutsovasilis: we do wait for 30s for the beats subprocesses, and this comment here holds the truth

swiatekm (approved):
LGTM, one nit about a comment that isn't blocking.

internal/pkg/agent/application/coordinator/coordinator.go
pkoutsovasilis: ok re-reading this, it can definitely improve in the phrasing sense

pkoutsovasilis: re-worded in c8c8830bd41b820b7fe28ceadb642ce24387a567

Issue #10634, opened 2025-10-16
[beats receivers] Monitoring receiver status persists after an unenroll action when they are not actually running


cmacknz:
- Relates https://github.com/elastic/elastic-agent/pull/10204#issuecomment-3408125884

The reproduction is what the `testInstallAndUnenrollWithEndpointSecurity` does in the failing tests in the PR above:

1. Enroll agent in a policy with defend that uses beats receivers
2. Unenroll the agent via the Fleet UI
3. Wait for Defend to be removed
4. Observe the state of the beats receivers for monitoring continue to be reported
5. Also observe the collector sub-process is not running. This seems to be a bug where we are not clearing the state on unenroll. Unenroll through fleet sends agent and empty agent policy to execute to remove all inputs. We seem to stop the collector but not clear its status.

```sh
ubuntu@ubuntu24:~/elastic-agent-9.3.0-SNAPSHOT-linux-arm64$ systemctl status elastic-agent
● elastic-agent.service - Elastic Agent is a unified agent to observe, monitor and protect your system.
     Loaded: loaded (/etc/systemd/system/elastic-agent.service; enabled; preset: enabled)
     Active: active (running) since Wed 2025-10-15 16:15:04 EDT; 5min ago
   Main PID: 7497 (elastic-agent)
      Tasks: 12 (limit: 1059)
     Memory: 135.7M (peak: 637.1M)
        CPU: 5.586s
     CGroup: /system.slice/elastic-agent.service
             └─7497 elastic-agent

ubuntu@ubuntu24:~/elastic-agent-9.3.0-SNAPSHOT-linux-arm64$ sudo elastic-agent status --output=full
┌─ fleet
│  └─ status: (HEALTHY) Connected
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: 5bcd2335-398b-4a39-aa67-9185ccb0538f
   │  ├─ version: 9.3.0
   │  └─ commit: c26dee602037f4808433a7a40d9fcb7097555267
   ├─ beat/metrics-monitoring
   │  ├─ status: (HEALTHY) HEALTHY
   │  ├─ beat/metrics-monitoring
   │  │  ├─ status: (HEALTHY) Healthy
   │  │  └─ type: OUTPUT
   │  └─ beat/metrics-monitoring-metrics-monitoring-beats
   │     ├─ status: (HEALTHY) Healthy
   │     └─ type: INPUT
   ├─ filestream-monitoring
[...]

PR #10594, opened 2025-10-15
Switch to otel self-monitoring by default

swiatekm:
## What does this PR do?

Switches Elastic Agent self-monitoring to use the otel runtime by default. The self-monitoring components will run as receivers within a managed otel collector, rather than individual processes.

## Why is it important?

This is the first step in an effort to give a similar treatment to all inputs. The benefit of this will be improved resource efficiency and compatibility with the otel universe.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/elastic-agent#changelog)
- ~~[ ] I have added an integration test or an E2E test~~

## Disruptive User Impact

There should be no user-facing impact, other than reduced resource consumption. If there is one, this change can be reversed by setting `agent.monitoring._runtime_experimental: process`.

## How to test this PR locally

Build agent and run it with the default configuration. You should see an otel collector process, and the self-monitoring inputs should be listed in status as `beats-receiver`.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/9058

PR #10675, opened 2025-10-20
Make otel runtime component updates synchronous

swiatekm:
## What does this PR do?

It makes the component status updates emitted by the otel manager fully synchronous. The coordinator is required to consume the updates, and not doing so will block the otel manager. This brings the otel manager in line with the runtime manager, which behaves the same way.

At the same time, the way status updates are handled internally by the otel manager is simplified. The manager used to emit synthetic `nil` updates in various circumstances to indicate the otel collector is not running. This does not happen anymore, and is now the responsibility of the collector execution. The otel manager's involvement in updates is limited to reading them from an internal channel, doing some processing, and writing them to external channels for consumption by the coordinator. This makes it much clearer how and when the manager can be blocked.

It's worth noting that the internal channel is still buffered and can drop updates. This isn't an issue, as it's the otel manager who holds the state information necessary to know which components have stopped running. As long as updates emitted by the otel manager itself are not dropped, the coordinator will eventually see the correct status information.

## Why is it important?

Currently, the otel manager doesn't act like the coordinator implicitly expects. The coordinator does its own component state accounting, and only removes components after it receives an explicit STOPPED state for them. However, the otel manager would emit component state updates as point-in-time snapshots. Because the external update channel was buffered, and it was drained on a new update emission, the update containing the STOPPED state could be dropped in favor of one where the component didn't exist at all. This caused bugs like https://github.com/elastic/elastic-agent/issues/10634, where components would persist in the agent status despite the otel collector not running at all.

This change also fixes endpoint test failures in https://github.com/elastic/elastic-agent/pull/10594, where we enable otel self-monitoring by default.

## Checklist

Build agent locally and try the reproduction from https://github.com/elastic/elastic-agent/issues/10634.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10634

### DIFF
--- a/internal/pkg/otel/manager/manager.go
+++ b/internal/pkg/otel/manager/manager.go
@@ -419,6 +419,13 @@ func (m *OTelManager) applyMergedConfig(ctx context.Context, collectorStatusCh c
 	if m.proc != nil {
 		m.proc.Stop(m.stopTimeout)
 		m.proc = nil
+		// We wait here for the collector to exit before possibly starting a new one. The execution indicates this
+		// by sending an error over the appropriate channel. It will also send a nil status that we'll either process
+		// after exiting from this function and going back to the main loop, or it will be overridden by the status
+		// from the newly started collector.
+		// This is the only blocking wait inside the main loop involving channels, so we need to be extra careful not to
+		// deadlock.
+		// TODO: Verify if we need to wait for the error at all. Stop() is already blocking.
 		select {
 		case <-collectorRunErr:
 		case <-ctx.Done():
--- a/internal/pkg/otel/manager/execution.go
+++ b/internal/pkg/otel/manager/execution.go
@@ -15,6 +15,11 @@ import (
 )
 
 type collectorExecution interface {
+	// startCollector starts the otel collector with the given arguments, returning a handle allowing it to be stopped.
+	// Cancelling the context will stop all goroutines involved in the execution.
+	// The collector will report status events in the statusCh channel and errors on errCh in a non-blocking fashion,
+	// draining the channel before writing to it.
+	// After the collector exits, it will emit an error describing the exit status (nil if successful) and a nil status.
 	startCollector(ctx context.Context, logger *logger.Logger, cfg *confmap.Conf, errCh chan error, statusCh chan *status.AggregateStatus) (collectorHandle, error)
 }
 


internal/pkg/otel/manager/manager_test.go
```diff
@@ -1641,6 +1654,107 @@ func TestOTelManagerEndToEnd(t *testing.T) {
 	})
 }
 
+// TestManagerAlwaysEmitsStoppedStatesForComponents checks that the manager always emits a STOPPED state for a component
+// at least once, even if we're slow to retrieve the state. This is part of the contract with the coordinator.
+func TestManagerAlwaysEmitsStoppedStatesForComponents(t *testing.T) {
```
cmacknz, line 1659: If I copy paste this test exactly as is into manager_test.go on main it still passes.

```sh
~/go/src/github.com/elastic/elastic-agent main ⇡235 !1 ?2 ·········································· 16s 03:33:11 PM
❯ while gotest ./internal/pkg/otel/manager/... -run TestManagerAlwaysEmitsStoppedStatesForComponents --count 1; do :;
 done
ok      github.com/elastic/elastic-agent/internal/pkg/otel/manager      2.418s
?       github.com/elastic/elastic-agent/internal/pkg/otel/manager/testing      [no test files]
ok      github.com/elastic/elastic-agent/internal/pkg/otel/manager      2.382s
?       github.com/elastic/elastic-agent/internal/pkg/otel/manager/testing      [no test files]
ok      github.com/elastic/elastic-agent/internal/pkg/otel/manager      2.410s
?       github.com/elastic/elastic-agent/internal/pkg/otel/manager/testing      [no test files]
ok      github.com/elastic/elastic-agent/internal/pkg/otel/manager      2.443s
?       github.com/elastic/elastic-agent/internal/pkg/otel/manager/testing      [no test files]
ok      github.com/elastic/elastic-agent/internal/pkg/otel/manager      2.478s
?       github.com/elastic/elastic-agent/internal/pkg/otel/manager/testing      [no test files]
```

It's kind of hard for me to see exactly what fixed the problem, but I did try to reproduce manually with a Linux agent and it unenrolled without stale status and previously I hit the bug this wants to fix every time.

swiatekm: It passes because the test setup already contains half the fix by changing the channel from buffered to unbuffered in the manager initialization. I've changed it to use the manager constructor directly, and now it fails on main as it should.

internal/pkg/otel/manager/manager.go
```diff
@@ -625,18 +597,10 @@ func (m *OTelManager) maybeUpdateMergedConfig(mergedCfg *confmap.Conf) (updated
...
+// in favor of newer ones here, as the coordinator expected incremental updates.
 func (m *OTelManager) reportComponentStateUpdates(ctx context.Context, componentUpdates []runtime.ComponentComponentState) {
-	select {
-	case <-ctx.Done():
-		// context is already done
-		return
-	case <-m.componentStateCh:
-	// drain the channel first
-	default:
-	}
```
swiatekm, line 639: @cmacknz For reference, this is the actual fix. Other non-test changes in this PR happen so the fix doesn't cause deadlocks in the manager, and more broadly, to make the flow easier to understand.

swiatekm:
> I also think some comments could be added, the changes in the executions from aggregateStatus(componentstatus.StatusStopped, nil) to nil in particular feels like a change that could come back accidentally if someone else touches these files.

Changing those would make tests fail, so it's not that easy. I've added comments to indicate what the status should be, but otherwise I think I'd need to write a short readme on how these components interact.

PR #10681, opened 2025-10-20
[Helm charts] Properly quote boolean values

ycombinator:
## What does this PR do?

This PR fixes the quoting of boolean values `agent.fleet.insecure` and `agent.fleet.force`.

## Why is it important?

Prior to the fix, the rendered values for these fields would appear like so, with an extra pair of quotes, which was causing the rendered value to be ineffective:

```
- name: FLEET_INSECURE
   value: '"true"'
- name: FLEET_FORCE
   value: '"true"'
```

## Checklist

## Disruptive User Impact

None; fixes a bug.

1. Add `insecure: true` [here](https://github.com/elastic/elastic-agent/blob/6186951dfbad2a7e0e1a37c26097d5b4d9d38dba/deploy/helm/elastic-agent/examples/fleet-managed/fleet-values.yaml#L10).
2.  Run `mage helm:renderExamples`
3.  Verify that the env var in the rendered [manifest](https://github.com/elastic/elastic-agent/blob/6186951dfbad2a7e0e1a37c26097d5b4d9d38dba/deploy/helm/elastic-agent/examples/fleet-managed/rendered/manifest.yaml) only has a single pair of double quotes around the value.

Thanks @pkoutsovasilis for suggesting the fix and testing steps.

### DIFF
--- a/changelog/fragments/1760996949-helm-bool.yaml
+++ b/changelog/fragments/1760996949-helm-bool.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Fix quoting of boolean values in Helm charts.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (commented):
hey @ycombinator, thanks for addressing this one. One proposal; go ahead add the `insecure: true` and `force: true` in the example you mention in the local testing. Then run `mage helm:renderExamples` and push all affected files in this PR. This way this will be always checked in the CI and hopefully will prevent any future issues with that

ycombinator:
> hey @ycombinator, thanks for addressing this one. One proposal; go ahead add the `insecure: true` and `force: true` in the example you mention in the local testing. Then run `mage helm:renderExamples` and push all affected files in this PR. This way this will be always checked in the CI and hopefully will prevent any future issues with that

With `insecure: true`, the rendered env var is `FLEET_INSECURE: "true"`. Are we sure we want to encourage this, by making it part of our examples?

Similarly, with `force: true`, the rendered env var is `FLEET_FORCE: "true"`.  This would cause the Agent container to re-enroll any time the container is created, regardless of whether a `fleet.enc` (or previously `fleet.yml`) file is present or not.  Are we sure we want to encourage this, by making it part of our examples?

swiatekm (commented):
Can you set these in one of the example configs, so we can see that the values are now rendered correctly?

ycombinator:
> Can you set these in one of the example configs, so we can see that the values are now rendered correctly?

Shoot, I missed this comment before hitting merge.  Sorry about that!

FWIW, I did test this locally; the steps are mentioned in the PR's description under "How to test this PR locally".

PR #10734, opened 2025-10-23
Reduce the scope of the legacy gcp secrets

fr4nc1sc0-r4m0n:
## What does this PR do?

This PR aims to reduce the scope of the legacy `kv/ci-shared/observability-ingest/cloud/gcp` secret only for the `integration-test-matrix` pipeline.
This is part of the global process about removing this secret usage.

## Why is it important?

It allows us to simplify the pre-command script and to limit the usage of a legacy secret which will be removed in the future.

## Checklist

## Related issues

- This is part of: https://github.com/elastic/observability-robots/issues/2800

## Questions to ask yourself

- ...

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/docker/docker v28.4.0+incompatible
 	github.com/docker/go-units v0.5.0
 	github.com/dolmen-go/contextio v0.0.0-20200217195037-68fc5150bcd5
-	github.com/elastic/beats/v7 v7.0.0-alpha2.0.20251025075340-e216ddecd7cf
+	github.com/elastic/beats/v7 v7.0.0-alpha2.0.20251027221544-36cfb330ca66
 	github.com/elastic/cloud-on-k8s/v2 v2.0.0-20250327073047-b624240832ae
 	github.com/elastic/elastic-agent-autodiscover v0.10.0
 	github.com/elastic/elastic-agent-client/v7 v7.17.2
--- a/go.sum
+++ b/go.sum
@@ -509,8 +509,8 @@ github.com/elastic/azure-sdk-for-go/sdk/resourcemanager/consumption/armconsumpti
 github.com/elastic/azure-sdk-for-go/sdk/resourcemanager/consumption/armconsumption v1.1.0-elastic/go.mod h1:0vCBR1wgGwZeGmloJ+eCWIZF2S47grTXRzj2mftg2Nk=
 github.com/elastic/bayeux v1.0.5 h1:UceFq01ipmT3S8DzFK+uVAkbCdiPR0Bqei8qIGmUeY0=
 github.com/elastic/bayeux v1.0.5/go.mod h1:CSI4iP7qeo5MMlkznGvYKftp8M7qqP/3nzmVZoXHY68=
-github.com/elastic/beats/v7 v7.0.0-alpha2.0.20251025075340-e216ddecd7cf h1:lmn+A1n8mTJvLrphU3Vi6JxZUBsD3YWgrjo7SgxgrFc=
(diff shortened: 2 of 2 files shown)


.buildkite/hooks/pre-command
```diff
@@ -34,7 +34,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent-package" ]]; then
   fi
 fi
 
-if [[ "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "integration-test-matrix" ]]; then
```
v1v: Can you confirm if the name of the pipeline is correct? See https://github.com/elastic/elastic-agent/blob/5546bd7cb69e89db33beb28069c40fbb2a9f1357/catalog-info.yaml#L360-L371

I can see `BUILDKITE_PIPELINE_SLUG="buildkite-elastic-agent-integration-matrix"` in the builds:
- https://buildkite.com/elastic/buildkite-elastic-agent-integration-matrix/builds/76

fr4nc1sc0-r4m0n: You are absolutely right! Thanks

.buildkite/hooks/pre-command
```diff
@@ -34,7 +34,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-agent-package" ]]; then
   fi
 fi
 
-if [[ "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "buildkite-elastic-agent-integration-matrix" ]]; then
```
v1v, line 37: IIUC, `API_KEY_TOKEN` is used in the its, so we will need to create a new conditional for the matrix and the GCP

.buildkite/hooks/pre-exit
```diff
@@ -14,11 +14,13 @@ if [[ "$BUILDKITE_STEP_KEY" == *"integration-tests"* ]]; then
   SNAPSHOT=true mage integration:clean
 fi
 
-if [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
-  if test -f "$GOOGLE_APPLICATION_CREDENTIALS"; then
-    rm $GOOGLE_APPLICATION_CREDENTIALS
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "buildkite-elastic-agent-integration-matrix" ]]; then
```
v1v: we don't need this conditional here, I think the conditionals are safe here:
- if variable exists
- if file exists

Regarldess of the BK pipeline, let's keep it simple

fr4nc1sc0-r4m0n: I thought the same but found this error:

- https://buildkite.com/elastic/elastic-agent/builds/29152#019a10f5-1f13-4e1d-80ef-b1edac2f88e4

The thing is that I don't see the `set -u`  option set which is the reason for this error so I preferred to check the pipeline slug as we did in the pre-command.

v1v: Change:

- `"$GOOGLE_APPLICATION_CREDENTIALS"` → `"${GOOGLE_APPLICATION_CREDENTIALS:-}"`
- `"$TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE"` → `"${TEST_INTEG_AUTH_GCP_SERVICE_TOKEN_FILE:-}"`


This ensures the `pre-exit` won't fail when these environment variables are not set and `set -u` is set.

v1v (approved):

fr4nc1sc0-r4m0n:
Some tests are failing but they do not appear to be related with this change

v1v:
@fr4nc1sc0-r4m0n , forgot to mention that anytime we touch any CI related things, we need to backport to all the active branches - that's the default behaviour.

For such I added the backport-active-all label,

In some cases we need to be a bit picky and only backport to some branches, btut default it's always to all the active branches

PR #10851, opened 2025-10-27
Allow different beats versions in integration tests

swiatekm:
## What does this PR do?

Makes tests which compare data emitted by beats processes and beats receivers to ignore differences in the agent version.

## Why is it important?

These values can be different when the version has been bumped in beats, but the respective commit hasn't been pulled in as a dependency. We shouldn't block CI in that case.

## Checklist


swiatekm:
> There should be a way to make the version equal, beats should take on the agent version which should match the beats receiver version, but this isn't an important enough problem that I want to block us on sorting out why it doens't work. So let's go with good enough for now.

In an ideal world, it should be the same commit, but getting there is unfortunately not that simple.

Issue #10826, opened 2025-10-24
Test EDOT collector onboarding reference configurations can be used to start a collector successfully


cmacknz:
- Relates https://github.com/elastic/elastic-agent/pull/10822

There are per OS getting started configuration files for the EDOT collector in this repository that apparently are not tested here, allowing simple things like typos and configuration name changes from dependency updates to break them.

The simplest tests that could happen would be unit tests to ensure the collector starts if given these as the starting configuration. The reference for testing this can be taken from https://github.com/elastic/elastic-agent/blob/d6c60645a0e65b2a1457c51c95504f915b2d83f7/internal/pkg/otel/run_nofips_test.go#L19-L55 where the modification would be to conditionally use the configuration based on the current GOOS. Our unit tests run on each of Linux, Windows, and Darwin already.

Further validation could be done using an integration test to make sure they ship the expected data, but this only supports Linux and Windows right now.

cmacknz:
@AlexanderWert can you direct this to the correct owning team, since the Elastic Agent teams don't own these configurations and we should make sure they continue to stay functional?

blakerouse:
We should definitely make it so the tests run based on the sample file being present. We shouldn't make it where a test must be written, the unit tests should just auto run because the file exists. Or make a test that ensures that all files are tested, and if one is not tested it fails.

cmacknz:
Closed by https://github.com/elastic/elastic-agent/pull/10856/files#diff-a53f8414104a3336273da9456b3d6d2d8b6356ed2569f82db58a192fd4b7efac

PR #10856, opened 2025-10-27
[samples] remove unneeded file_storage and fix a bug

VihasMakwana:
## What does this PR do?

Remove the `file_storage` as it's not used anywhere and use `flush_timeout`.

Also added unit test cases to validate the samples using `DryRun`.

Screenshot 2025-10-27 at 7 35 58 PM


cmacknz:
I think https://github.com/elastic/elastic-agent/issues/10826 should be addressed here, it doesn't seem like adding a basic unit test that the collector starts is that much work. If it is a lot of work let me know and we can discuss a better strategy to address this.

internal/pkg/otel/samples_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/collector/otelcol"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func TestSamples(t *testing.T) {
+	// Set environment variables required by the sample configurations.
+	// These are just sample values to allow the collector to start.
+	os.Setenv("STORAGE_DIR", t.TempDir())
```
pchila: In `internal/pkg/otel/samples/windows/managed_otlp/logs_metrics_traces.yml` above the usage of `STORAGE_DIR` has been removed, so I am assuming this will be used by something else ?

VihasMakwana: Yes.
It will be used by linux and darwin configs.
They still need this param

internal/pkg/otel/samples_test.go
```diff
@@ -0,0 +1,74 @@
...
+	os.Setenv("ELASTIC_ENDPOINT", "http://localhost:9200")
```
swiatekm: All this should just be [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv).

internal/pkg/otel/samples_test.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		return nil
+	})
+	assert.NoError(t, err)
+}
+
+func testSample(t *testing.T, configFile string) {
+	settings := NewSettings("test", []string{configFile})
+	settings.LoggingOptions = []zap.Option{zap.WrapCore(func(zapcore.Core) zapcore.Core {
+		return zapcore.NewNopCore()
```
swiatekm, line 56: It'd be better to make this an instance of https://pkg.go.dev/go.uber.org/zap/zaptest/observer and print the logs if the test fails, making troubleshooting much easier.

VihasMakwana: @swiatekm Enabling this results in a race condition. I'll work on this as a follow-up once I find complete RCA of the race condition.

Issue #10645, opened 2025-10-16
[beats receivers] switch monitoring runtime from process to otel deletes run directories


leehinman:
If you switch the monitoring "_runtime_experimental" from "process" to "otel" the run directories are all deleted.

If you use inotifywatch you can observe this
``` shell
inotifywait -mr /opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/filestream-monitoring/registry/filebeat/ MODIFY log.json
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/filestream-monitoring/registry/filebeat/ MODIFY log.json
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/filestream-monitoring/registry/filebeat/ MODIFY log.json
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/filestream-monitoring/registry/filebeat/ CLOSE_WRITE,CLOSE log.json
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/ OPEN,ISDIR beat
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/beat/ OPEN,ISDIR
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/beat/ OPEN,ISDIR metrics-monitoring
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/beat/metrics-monitoring/ OPEN,ISDIR
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/beat/ ACCESS,ISDIR metrics-monitoring
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/beat/metrics-monitoring/ ACCESS,ISDIR
/opt/Elastic/Agent/data/elastic-agent-9.3.0-SNAPSHOT-3c3de2/run/beat/ ACCESS,ISDIR metrics-monitoring
[...]

PR #10857, opened 2025-10-27
Move component working directory management to coordinator

swiatekm:
## What does this PR do?

It makes the coordinator responsible for creating and removing the working directories of components. Until now, CommandRuntime and ServiceRuntime did this on their own, whereas the OtelManager didn't do it at all. We move the logic for creating and removing the directories into the component module itself, and call it from the Coordinator. The new logic is as follows.

- When we generate a new component model, we ensure each component has a working directory.
- When we get a STOPPED state for a component *AND* that component is not present in the component model, we delete its working directory.

The condition that the component must not be present in the current model fixes an issue where the working directory would be deleted when the component was being moved between runtimes.

## Why is it important?

1. Currently, the Otel runtime does not remove the working directories at all. The behavior should be the same no matter where the component runs.
2. If a component is moved from the process runtime to the otel runtime, its working directory is incorrectly deleted. This causes filestream to re-ingest all logs, for example.

Note that I've included an integration test from https://github.com/elastic/elastic-agent/pull/10544, which specifically checks point 2 from above.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10645
- Relates https://github.com/elastic/elastic-agent/issues/8207

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/elastic/elastic-agent/pkg/component"
+	"github.com/elastic/elastic-agent/pkg/testing/tools/fleettools"
 	"github.com/elastic/go-elasticsearch/v8"
 
 	"github.com/gofrs/uuid/v5"
@@ -1355,13 +1356,14 @@ func setStrictMapping(client *elasticsearch.Client, index string) error {
 // re-ingest logs.
 //
 // Flow
-// 1) Create policy in Kibana with just monitoring and download it
-// 2) Install with monitoring with "process" runtime
-// 3) restart agent, to roll log files
-// 4) Switch to monitoring "otel" runtime
-// 5) restart agent 3 times, making sure healthy between restarts
-// 6) query ES for monitoring logs with aggregation on fingerprint and line number, should be 0 duplicates
-// 7) uninstall
+//  1. Create policy in Kibana with just monitoring and "process" runtime
+//  2. Install and Enroll
+//  3. Switch to monitoring "otel" runtime
+//  4. restart agent 3 times, making sure healthy between restarts
+//  5. switch back to "process" runtime
+//  6. query ES for monitoring logs with aggregation on fingerprint and line number,
+//     ideally 0 duplicates but possible to have a small number
+//  7. uninstall
 func TestMonitoringNoDuplicates(t *testing.T) {
 	info := define.Require(t, define.Requirements{
 		Group: integration.Default,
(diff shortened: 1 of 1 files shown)


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -958,21 +963,23 @@ func (c *Coordinator) watchRuntimeComponents(
...
-			for _, componentState := range componentStates {
-				logComponentStateChange(c.logger, state, &componentState)
-				// Forward the final changes back to Coordinator, unless our context
-				// has ended.
-				select {
-				case c.managerChans.runtimeManagerUpdate <- componentState:
-				case <-ctx.Done():
-					return
-				}
-			}
```
swiatekm, line 971: This is just an incidental cleanup. The two branches of this select do the same thing.

pkg/component/runtime/manager_fake_input_test.go
```diff
@@ -792,9 +791,10 @@ func (suite *FakeInputSuite) TestManager_BadUnitToGood() {
 				} else {
 					unit, ok := state.Units[ComponentUnitKey{UnitType: client.UnitTypeInput, UnitID: "fake-input"}]
 					if ok {
-						if unit.State == client.UnitStateFailed {
+						switch unit.State {
```
swiatekm, line 816: All of these changes were demanded by the linter.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -911,6 +908,138 @@ agent.monitoring.enabled: false
 	assert.Equal(t, expectedMessage, message)
 }
 
+// TestComponentWorkDir verifies that the component working directory is not deleted when moving the component from
+// the process runtime to the otel runtime.
+func TestComponentWorkDir(t *testing.T) {
```
cmacknz, line 914: This is very similar to what TestCoordinatorManagesComponentWorkDirs in the unit test covers isn't it? 

TestMonitoringNoDuplicates later on basically also requires this to work to pass? Is this test giving us anything that isn't already covered elsewhere?

cmacknz:
Did a first pass and this looks good, want to see CI pass before approving and take a second look with fresher eyes tomorrow (some else approving also covers this).

There is also the requirement that we don't change the run directory names from what they were before this change, which I double checked manually to confirm it is the same.

testing/integration/ess/beat_receivers_test.go
swiatekm: True. I wanted to have an integration test that explicitly tests this behavior for agent as a black box, but I can see the argument that it's redundant. If and when we decide to change this behavior, this test could also be expanded to cover that.

testing/integration/ess/endpoint_security_test.go
```diff
@@ -714,10 +714,11 @@ func testInstallWithEndpointSecurityAndRemoveEndpointIntegration(t *testing.T, i
 	// Verify that the Endpoint directory was correctly removed.
 	// Regression test for https://github.com/elastic/elastic-agent/issues/3077
 	agentInstallPath := fixture.WorkDir()
-	files, err := os.ReadDir(filepath.Clean(filepath.Join(agentInstallPath, "..")))
+	elasticInstallPath := filepath.Clean(filepath.Join(agentInstallPath, ".."))
+	files, err := os.ReadDir(elasticInstallPath)
 	require.NoError(t, err)
 
-	t.Logf("Checking directories at install path %s", agentInstallPath)
+	t.Logf("Checking directories at install path %s", elasticInstallPath)
```
swiatekm, line 721: This is just a small cleanup I did while troubleshooting failures of this test.

changelog/fragments/1761679059-move-component-workdir-coordinator.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Move component working directory management to coordinator
```
blakerouse:
```suggestion
summary: Fix issues where switching to OTEL runtime would cause data to be re-ingested
```

Having the changelog more using facing is always better and easier for our users to understand what they are getting when the upgrade. A user really wouldn't understand the description you provided, that is more of a description for us.

blakerouse (approved):
This change looks good to me. My only comment is really on the changelog.

internal/pkg/agent/application/coordinator/coordinator.go
blakerouse: Thanks for the note.

pkg/component/runtime/manager_fake_input_test.go
blakerouse: Thanks for the note.

testing/integration/ess/beat_receivers_test.go
cmacknz: Thanks, I was mostly approaching this from the angle of not over testing to try to keep CI executing time under control. The test itself is fine.

cmacknz (approved):

testing/integration/ess/beat_receivers_test.go
```diff
@@ -1219,3 +1348,267 @@ func setStrictMapping(client *elasticsearch.Client, index string) error {
...
+//
+// Flow
+// 1) Create policy in Kibana with just monitoring and download it
+// 2) Install with monitoring with "process" runtime
+// 3) restart agent, to roll log files
+// 4) Switch to monitoring "otel" runtime
+// 5) restart agent 3 times, making sure healthy between restarts
+// 6) query ES for monitoring logs with aggregation on fingerprint and line number, should be 0 duplicates
+// 7) uninstall
+func TestMonitoringNoDuplicates(t *testing.T) {
```
leehinman, line 1367: Up to you.  But I did rework the test to use Fleet instead of stand alone.   Plus starting at process, switching to otel, 3 restarts, then switching back to process.

https://github.com/leehinman/elastic-agent/blob/beatreceiver_no_dups/testing/integration/ess/beat_receivers_test.go

I think you would just remove https://github.com/leehinman/elastic-agent/blob/3ad4cb9d503b9229d440f1689b03c405aff1705c/testing/integration/ess/beat_receivers_test.go#L1531 and remove 
https://github.com/leehinman/elastic-agent/blob/3ad4cb9d503b9229d440f1689b03c405aff1705c/testing/integration/ess/beat_receivers_test.go#L1519-L1522

changelog/fragments/1761679059-move-component-workdir-coordinator.yaml
swiatekm: My thinking was that if this change breaks something, it's going to be easier to figure out my summary might have something to do with it, vs "we fixed things". But it's a small thing, I don't mind using yours.

Issue #10643, opened 2025-10-16
[beats receivers] Add ability to mute Elasticsearch exporter health status


cmacknz:
- Relates https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39562

The Elasticsearch exporter introduced connection health reporting, which will make it's way into Elastic Agent as Beats begin running as receivers starting with the monitoring Beats.

This is the initial introduction of output health and we may find that it has bugs or is too sensitive, so add the ability to mute the problems it detects in cases where a bug fix is needed. It is also the case that the Fleet UI does not visualize output health as well as input health (yet) so this also gives us the option to hide the feature until we are ready to have it work properly end to end.

Ideally, this would be configured per output and can be a toggle in the advanced YAML settings box. This avoids the need to build any additional UI around this toggle. This would have Elastic Agent receiver configuration like below but introduce the requirement to track the state of the toggle per input.

```yaml
outputs:
  default:
    type: elasticearch:
    status_reporting: 
      enabled: false
```

We could also potentially add this at the agent policy level, perhaps grouping it with the collector settings. This would likely be simpler to implement but require a new setting in the agent policy settings UI or a call to the policy override API to set until a toggle can be added directly in the UI.

```yaml
agent:
  collector:
    status_reporting: 
      enabled: false
```

VihasMakwana:
@cmacknz Just to confirm, this is what we want:

1. If status reporting for output is disabled, the `elastic-agent status` command should report:
```
┌─ fleet
│  └─ status: (STOPPED) Not enrolled into Fleet
└─ elastic-agent
   ├─ status: (HEALTHY) Running
   ├─ info
   │  ├─ id: c910a574-5d3a-423d-bec2-1b16df69b922
   │  ├─ version: 9.3.0
   │  └─ commit: 03a81c005d38f8190053c73e9033c7a94cefb2ac
   ├─ beat/metrics-monitoring
   │  ├─ status: (HEALTHY) HEALTHY
   │  └─ beat/metrics-monitoring-metrics-monitoring-beats
   │     ├─ status: (HEALTHY) Healthy
   │     └─ type: INPUT
   ├─ filestream-monitoring
   │  ├─ status: (HEALTHY) HEALTHY
   │  └─ filestream-monitoring-filestream-monitoring-agent
   │     ├─ status: (HEALTHY) Healthy
   │     └─ type: INPUT
   ├─ http/metrics-monitoring
[...]

PR #10890, opened 2025-10-29
[beatsreceivers] add option to mute exporter status

VihasMakwana:
## What does this PR do?

Add ability to mute exporter status for otel-managed components. With the recent introduction of connection health reporting in ES exporter, we may find it to be too sensitive. 
This PR exposes an option under `outputs` section, to disable status reporting for the output.

### Usage

Running agent with following config will result in "always healthy" status for the output.

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [...]
    username: "elastic"
    password: "..."
    preset: balanced
    status_reporting:
      enabled: false                 # <------ IMPORTANT

inputs:
  - type: system/metrics
    _runtime_experimental: otel
    data_stream.namespace: default
    use_output: default
    streams:
      - metricsets:
        - cpu
        data_stream.dataset: system.cpu
      - metricsets:
        - memory
        data_stream.dataset: system.memory
      - metricsets:
        - network
        data_stream.dataset: system.network
      - metricsets:
        - filesystem
        data_stream.dataset: system.filesystem
```

### Testing

Added a UT

Closes https://github.com/elastic/elastic-agent/issues/10643

### DIFF
--- a/internal/pkg/otel/manager/manager.go
+++ b/internal/pkg/otel/manager/manager.go
@@ -327,15 +327,17 @@ func (m *OTelManager) Run(ctx context.Context) error {
 				m.logger.Debugf(
 					"new config hash (%d) is identical to the old config hash (%d), skipping update",
 					m.mergedCollectorCfgHash, previousConfigHash)
-			}
 
-			// there was a config update. Force fetch the latest collector status.
-			// drain the channel first
-			select {
-			case <-forceFetchStatusCh:
-			default:
+				// there was a config update, but the hash hasn't changed.
+				// Force fetch the latest collector status in case the user modified the output.status_reporting flag.
+				//
+				// drain the channel first
+				select {
+				case <-forceFetchStatusCh:
+				default:
+				}
+				forceFetchStatusCh <- struct{}{}
 			}
-			forceFetchStatusCh <- struct{}{}
 
 		case otelStatus := <-collectorStatusCh:
 			err = m.reportOtelStatusUpdate(ctx, otelStatus)
--- a/testing/integration/ess/otel_test.go
+++ b/testing/integration/ess/otel_test.go
@@ -2221,6 +2221,7 @@ service:
 
 func TestOutputStatusReporting(t *testing.T) {
 	define.Require(t, define.Requirements{
+		Sudo:  true,
 		Group: integration.Default,
 		Local: false,
 		Stack: nil,


cmacknz:
I tried this quickly locally, and it seems like it isn't working.

I am running an agent from this branch, I have monitoring set to use beats receivers but pointed at localhost where there is no ES running, and I have status_reporting disabled, but the output for each monitoring Beat is still degraded.

```sh
❯ sudo elastic-development-agent version
Binary: 9.3.0-SNAPSHOT (build: c066018d6229f66f95214778e96bf4e289efa682 at 2025-10-30 19:41:12 +0000 UTC)
Daemon: 9.3.0-SNAPSHOT (build: c066018d6229f66f95214778e96bf4e289efa682 at 2025-10-30 19:41:12 +0000 UTC)

❯ sudo elastic-development-agent inspect
agent:
  logging:
    to_stderr: true
  monitoring:
    _runtime_experiemental: otel
    enabled: true
    logs: true
    metrics: true
inputs:
- data_stream:
    namespace: default
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  - data_stream:
      dataset: system.memory
    metricsets:
    - memory
  - data_stream:
      dataset: system.network
    metricsets:
    - network
  - data_stream:
      dataset: system.filesystem
    metricsets:
    - filesystem
  type: system/metrics
  use_output: default
outputs:
  default:
    api_key: <REDACTED>
    hosts:
    - 127.0.0.1:9200
    preset: balanced
    status_reporting:
      enabled: false
    type: elasticsearch

┌─ fleet
└─ elastic-agent
   ├─ beat/metrics-monitoring
   │  └─ beat/metrics-monitoring
   ├─ filestream-monitoring
   │  └─ filestream-monitoring
   ├─ http/metrics-monitoring
   │  └─ http/metrics-monitoring
   └─ prometheus/metrics-monitoring
      ├─ status: (DEGRADED) DEGRADED
      └─ prometheus/metrics-monitoring
         └─ status: (DEGRADED) Elasticsearch request failed: dial tcp 127.0.0.1:9200: connect: connection refused
```

cmacknz:
I see the problem and my reproduction notes above aren't complete:
1. I started agent with this config but with status_reporting unset (so enabled)
2. I edited the config to specify `status_reporting.enabled: false` and waited for agent to reload the config.

The problem is this was displaying the last state, and this doesn't emit a state change back to OK when the status_reporting value is changed. If I run `elastic-agent restart` then it works which isn't enough, because there is not yet a way to restart Fleet managed agents remotely. 

If someone makes a policy change to mute status reporting, this has to take effect just from the policy change.

swiatekm:
> I see the problem and my reproduction notes above aren't complete:
> 
>     1. I started agent with this config but with status_reporting unset (so enabled)
> 
>     2. I edited the config to specify `status_reporting.enabled: false` and waited for agent to reload the config.
> 
> 
> The problem is this was displaying the last state, and this doesn't emit a state change back to OK when the status_reporting value is changed. If I run `elastic-agent restart` then it works which isn't enough, because there is not yet a way to restart Fleet managed agents remotely.
> 
> If someone makes a policy change to mute status reporting, this has to take effect just from the policy change.

That's an annoying problem. The agent status (really the coordinator status) will not change until the otel collector emits another status, and that only happens if something changes. Just changing status_reporting doesn't affect the exporter configuration, so it doesn't actually change anything as far the otel manager is concerned.

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -175,10 +175,7 @@ func (r *subprocessExecution) startCollector(ctx context.Context, logger *logger
 			} else {
 				maxFailuresTimer.Reset(maxFailuresDuration)
 				removeManagedHealthCheckExtensionStatus(statuses, r.healthCheckExtensionID)
-				if !compareStatuses(currentStatus, statuses) {
-					currentStatus = statuses
-					r.reportSubprocessCollectorStatus(procCtx, statusCh, statuses)
-				}
+				r.reportSubprocessCollectorStatus(procCtx, statusCh, statuses)
```
VihasMakwana: @swiatekm @cmacknz I've updated the health check implementation so that it now reports status, even if the status hasn’t changed, to address the issue Craig identified.
I've also increased the health check period to `10s`.

pierrehilbert: Does this change introduce overhead for the Agent by requiring it to always update the status, even when nothing has changed? 
I assume there was a reason we didn’t implement it this way initially.

Would it be possible instead to “force” a status refresh only when `status_reporting.enabled` is changed?

VihasMakwana: I'm wondering the same. I'm currently trying another option out which is more desirable for us in the long term.

VihasMakwana: Addressed in latest commit. We don't need to always push status.

VihasMakwana:
> I see the problem and my reproduction notes above aren't complete:
> 
> 1. I started agent with this config but with status_reporting unset (so enabled)
> 2. I edited the config to specify `status_reporting.enabled: false` and waited for agent to reload the config.
> 
> The problem is this was displaying the last state, and this doesn't emit a state change back to OK when the status_reporting value is changed. If I run `elastic-agent restart` then it works which isn't enough, because there is not yet a way to restart Fleet managed agents remotely.
> 
> If someone makes a policy change to mute status reporting, this has to take effect just from the policy change.

@cmacknz this should be fixed in latest commit. I've tested various other scenarios as well.

pierrehilbert:
Windows test failure is a known unrelated issue: https://github.com/elastic/elastic-agent/issues/10944
@VihasMakwana could you please fix the check-ci failure?

internal/pkg/otel/manager/execution.go
```diff
@@ -20,7 +20,7 @@ type collectorExecution interface {
 	// The collector will report status events in the statusCh channel and errors on errCh in a non-blocking fashion,
 	// draining the channel before writing to it.
 	// After the collector exits, it will emit an error describing the exit status (nil if successful) and a nil status.
-	startCollector(ctx context.Context, baseLogger *logger.Logger, logger *logger.Logger, cfg *confmap.Conf, errCh chan error, statusCh chan *status.AggregateStatus) (collectorHandle, error)
+	startCollector(ctx context.Context, baseLogger *logger.Logger, logger *logger.Logger, cfg *confmap.Conf, errCh chan error, statusCh chan *status.AggregateStatus, forceFetchStatusCh chan struct{}) (collectorHandle, error)
```
swiatekm, line 28: You should update the docstring with this new parameter, ideally explaining why it exists.

internal/pkg/otel/translate/status.go
```diff
@@ -90,6 +90,110 @@ func DropComponentStateFromOtelStatus(otelStatus *status.AggregateStatus) (*stat
 	return newStatus, nil
 }
 
+// AlterPipelineStatus modifies the given otel status by muting exporter statuses for muted components.
+// It also updates parent pipeline statuses based on child components.
+func AlterPipelineStatus(
```
swiatekm: This function really just does one thing - it mutes exporter status and updates the pipeline status accordingly. So I think a name like `maybeMuteExporterStatuses` would be more descriptive.

testing/integration/ess/otel_test.go
```diff
@@ -2218,3 +2218,190 @@ service:
...
+			StatusReportingEnabled: true,
+		})
+	ctx, cancel := testcontext.WithDeadline(t, context.Background(), time.Now().Add(5*time.Minute))
+	defer cancel()
+	err = fixture.Prepare(ctx)
+	require.NoError(t, err)
+
+	err = fixture.Configure(ctx, configBuffer.Bytes())
+
+	cmd, err := fixture.PrepareAgentCommand(ctx, []string{"-e"})
```
swiatekm: Instead of doing this, please install the agent using `fixture.Install` and change the configuration by calling `fixture.Configure`. Running the command manually is error-prone and doesn't leave diagnostics for troubleshooting. Check the tests in `beats_receivers_test.go` for examples.

testing/integration/ess/otel_test.go
```diff
@@ -2218,3 +2218,190 @@ service:
...
+		err = fixture.IsHealthy(ctx)
+		if err != nil {
+			t.Logf("waiting for agent healthy: %s", err.Error())
+			return false
+		}
+		return true
+	}, 30*time.Second, 1*time.Second)
+
+	actualHits := &struct{ Hits int }{}
+	assert.Eventually(t,
```
swiatekm: Do we actually need to check this for the purposes of this test?

testing/integration/ess/otel_test.go
```diff
@@ -2218,3 +2218,190 @@ service:
...
+func TestOutputStatusReporting(t *testing.T) {
+	info := define.Require(t, define.Requirements{
+		Group: integration.Default,
+		Local: true,
+		OS: []define.OS{
+			{Type: define.Windows},
+			{Type: define.Linux},
+			{Type: define.Darwin},
+		},
+		Stack: &define.Stack{},
```
swiatekm: Does this test really need to send data to ES? It's a lot of code and potential complexity. Why not just start with localhost and then disable status reporting?

testing/integration/ess/otel_test.go
VihasMakwana: Not really. I'll remove this section.

testing/integration/ess/otel_test.go
VihasMakwana: Yeah you're right. I've fixed it.

swiatekm (approved):

internal/pkg/otel/manager/manager.go
```diff
@@ -328,6 +329,14 @@ func (m *OTelManager) Run(ctx context.Context) error {
 					m.mergedCollectorCfgHash, previousConfigHash)
 			}
 
+			// there was a config update. Force fetch the latest collector status.
+			// drain the channel first
+			select {
```
cmacknz: Shouldn't this be in the `configChanged` block above?

testing/integration/ess/otel_test.go
```diff
@@ -2218,3 +2218,127 @@ service:
...
+		status, err := fixture.ExecStatus(ctx)
+		if err != nil {
+			t.Logf("waiting for agent degraded: %s", err.Error())
+			return false
+		}
+		return status.State == int(cproto.State_DEGRADED)
+	}, 30*time.Second, 1*time.Second)
+
+	// Disable status reporting.
+	// This should result in HEALTHY state
```
cmacknz, line 2308: The order of setting the status reporting flags should be reversed. This test can pass if reporting degraded is simply delayed before you can observe it.

You want to confirm degraded first, then mute the status, then confirm you are healthy.

internal/pkg/otel/manager/manager.go
swiatekm: You're right, it should be in the else branch.

VihasMakwana: yeah, adding it to else branch would do the work. 
We will be starting a new process if config has changed and that would trigger a new status, so that's good.

testing/integration/ess/otel_test.go
VihasMakwana: We check for DEGRADED first, right before this section. Ref https://github.com/elastic/elastic-agent/blob/fff23b6a78ee9547e23ec1080ccddd1a0aa7db95/testing/integration/ess/otel_test.go#L2297-L2304.

We then disable the status reporting and confirm that we're healthy.

cmacknz: Ah I misled because there wasn't a comment for this and I skimmed through the test too quickly :)

cmacknz:
Tested again locally and it worked this time, thanks!

PR #10897, opened 2025-10-29
feat: remove in-process execution mode of edot collector

pkoutsovasilis:
## What does this PR do?

This PR removes the in-process (embedded) execution mode for the OTel collector, leaving only the subprocess execution mode. Specifically:

- **Removes embedded execution implementation**: Deletes `internal/pkg/otel/manager/execution_embedded.go` and removes the `EmbeddedExecutionMode` constant
- **Removes execution mode configuration**: Deletes `internal/pkg/otel/config/config.go` which parsed the `agent.features.otel.subprocess_execution` configuration flag
- **Hardcodes subprocess mode**: Updates `application.go` and `inspect.go` to always use subprocess execution mode instead of reading it from configuration

## Why is it important?

This change is **required** to eventually split EDOT (Elastic Distribution for OpenTelemetry) from elastic-agent into a separate binary and module (as was done in [this PR](https://github.com/elastic/elastic-agent/pull/9308)) as the in-process execution mode cannot be supported in such scenario.

## Checklist

## Disruptive User Impact

**None**. The OTel collector running in subprocess mode is already the default behavior, and the `agent.features.otel.subprocess_execution` configuration option was never intended for public usage.

Run any/all integration or unit test, they should run successfully 

## Related issues

N/A

### DIFF
--- a/internal/pkg/agent/cmd/inspect.go
+++ b/internal/pkg/agent/cmd/inspect.go
@@ -416,7 +416,7 @@ func getMonitoringFn(ctx context.Context, logger *logger.Logger, cfg map[string]
 	if err != nil {
 		return nil, fmt.Errorf("could not load agent info: %w", err)
 	}
-	monitor := componentmonitoring.New(agentCfg.Settings.V1MonitoringEnabled, agentCfg.Settings.DownloadConfig.OS(), agentCfg.Settings.MonitoringConfig, agentInfo, true)
+	monitor := componentmonitoring.New(agentCfg.Settings.V1MonitoringEnabled, agentCfg.Settings.DownloadConfig.OS(), agentCfg.Settings.MonitoringConfig, agentInfo)
 	return monitor.MonitoringConfig, nil
 }
 
--- a/internal/pkg/agent/application/application.go
+++ b/internal/pkg/agent/application/application.go
@@ -136,7 +136,6 @@ func New(
 		cfg.Settings.DownloadConfig.OS(),
 		cfg.Settings.MonitoringConfig,
 		agentInfo,
-		true,
 	)
 
 	runtime, err := runtime.NewManager(
--- a/internal/pkg/agent/application/monitoring/component/testdata/monitoring_config_full_otel.yaml
+++ b/internal/pkg/agent/application/monitoring/component/testdata/monitoring_config_full_otel.yaml
@@ -461,6 +461,67 @@ inputs:
     failure_threshold: 5
     hosts:
     - placeholder
+    id: metrics-monitoring-edot-collector
+    index: metrics-elastic_agent.elastic_agent-default
+    metricsets:
+    - json
+    namespace: agent
+    path: /stats
+    period: 1m0s
+    processors:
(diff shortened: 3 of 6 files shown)


internal/pkg/agent/cmd/inspect.go
```diff
@@ -418,9 +416,7 @@ func getMonitoringFn(ctx context.Context, logger *logger.Logger, cfg map[string]
 	if err != nil {
 		return nil, fmt.Errorf("could not load agent info: %w", err)
 	}
-	otelExecMode := otelconfig.GetExecutionModeFromConfig(logger, config)
-	isOtelExecModeSubprocess := otelExecMode == manager.SubprocessExecutionMode
-	monitor := componentmonitoring.New(agentCfg.Settings.V1MonitoringEnabled, agentCfg.Settings.DownloadConfig.OS(), agentCfg.Settings.MonitoringConfig, agentInfo, isOtelExecModeSubprocess)
+	monitor := componentmonitoring.New(agentCfg.Settings.V1MonitoringEnabled, agentCfg.Settings.DownloadConfig.OS(), agentCfg.Settings.MonitoringConfig, agentInfo, true)
```
swiatekm: Can we drop the attribute from monitoring config as well?

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -110,9 +110,9 @@ func (r *subprocessExecution) startCollector(ctx context.Context, logger *logger
 		return nil, fmt.Errorf("failed to marshal config to yaml: %w", err)
 	}
 
-	stdOut := runtimeLogger.NewLogWriterWithDefaults(logger.Core(), zapcore.Level(r.logLevel))
+	stdOut := runtimeLogger.NewLogWriterWithDefaults(baseLogger.Core(), zapcore.Level(r.logLevel))
 	// info level for stdErr because by default collector writes to stderr
-	stdErr := runtimeLogger.NewLogWriterWithDefaults(logger.Core(), zapcore.Level(r.logLevel))
+	stdErr := runtimeLogger.NewLogWriterWithDefaults(baseLogger.Core(), zapcore.Level(r.logLevel))
```
swiatekm, line 115: This is a bug fix, right? We should've always been logging this using the agent's logger. Maybe there should be a test that checks this somewhere.

pkoutsovasilis: there is already a test that checks [this here](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/manager/manager_test.go#L814) but as you can tell it was sharing previously the logger between the in-process and sub-process test cases. As a result this was always coming out successful as the in-process case had shipped the expected logs before the subprocess mode

internal/pkg/agent/cmd/inspect.go
pkoutsovasilis: dropped https://github.com/elastic/elastic-agent/pull/10897/commits/7faee28f6856e3f8e3e59b08b56035f65a91f052

PR #10906, opened 2025-10-29
fix(bk): serverless-beats-tests start failing after merging #10734

v1v:
## What does this PR do?

Restore the previous environmental context for the `serverless-beats-tests`
Run `serverless-beats-tests` if touching some of the core files.

## Why is it important?

Regression, and help with early detection
## Checklist

NA

Only in the CI

This PR triggered:
- https://buildkite.com/elastic/beats-agent-serverless-tests/builds/1293

image

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -358,7 +358,7 @@ steps:
                 - .buildkite/scripts/steps/beats_tests.sh
                 - .buildkite/hooks/pre-command
               config:
-                trigger: "elastic-agent-extended-testing"
+                trigger: "beats-agent-serverless-tests"
                 build:
                   commit: "${BUILDKITE_COMMIT}"
                   branch: "${BUILDKITE_BRANCH}"


ycombinator (dismissed):
LGTM but I'd also like @pchila to review as he's currently working on trying to improve our CI times.

.buildkite/pipeline.yml
```diff
@@ -344,6 +344,25 @@ steps:
...
+      - monorepo-diff#v1.2.0:
+          diff: "git diff --name-only origin/${GITHUB_PR_TARGET_BRANCH}...HEAD"
+          interpolation: false
+          watch:
+            - path:
+                - .buildkite/serverless.beats.tests.yml
+                - .buildkite/scripts/steps/beats_tests.sh
+                - .buildkite/hooks/pre-command
+              config:
+                trigger: "elastic-agent-extended-testing"
```
v1v: I use `trigger`; otherwise, the if condition in the `pre-command` will require granting access to the main pipeline or include more sophisticated conditionals.

The trigger step will make the GitHub build comment not work nicely for this step; however, for simplicity, I think it's worth keeping it simple for now, even if the github build comment won't contain those details clearly - this pipeline is not intended to change often.

I decided not to use the BK PR bot, because it does not support skipping the very first build for a PR, see https://github.com/elastic/logstash/pull/18327 (that introduced an artificial pipeline to manage those cases)

.buildkite/pipeline.yml
```diff
@@ -344,6 +344,25 @@ steps:
       commit: "${BUILDKITE_COMMIT}"
       branch: "${BUILDKITE_BRANCH}"
 
+  # Trigger for pull requests - Serverless Beats Tests
+  # this should help detecting issues earlier in the development cycle
+  - label: "Trigger Serverless Beats Tests"
+    if: build.pull_request.id != null
```
pchila, line 350: Does this mean that this would run on elastic-agent PRs ? I don't think we want that, just that's possible to trigger from beats, right @cmacknz @ebeahan  ?

v1v: No, it won't run on PRs unless it matches any changes in the list of files, see `L357-L360`:

https://github.com/elastic/elastic-agent/pull/10906/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R352-R359

That's intended to help detect breaking changes in the BK pipeline for `Serverless Beats Tests` before merging; otherwise, we can end up with some regressions, as happened recently.

----------

In my view, any changes in any pipeline should be triggered automatically on PR basis by default, this is one way to reduce the burden of PR/merging/breaking by PR/breaking/iterating/merging

.buildkite/pipeline.yml
v1v:
```suggestion
                trigger: "beats-agent-serverless-tests"
```

.buildkite/pipeline.yml
cmacknz: This makes sense to me, that we run this on elastic-agent PRs that change the pipeline definition exclusively but it otherwise remains triggered by Beats.

fr4nc1sc0-r4m0n:
We should backport this change to all active branches, right?

.buildkite/pipeline.yml
ebeahan: ++ not running in PRs changing the pipeline definition is how this wasn't surfaced in https://github.com/elastic/elastic-agent/pull/10734

PR #10931, opened 2025-10-30
Add workflow for generating release notes

colleenmcginnis:
## What does this PR do?

Adds a GitHub workflow that can be manually triggered to generate release notes and open a PR.

As currently written, the workflow needs to be manually triggered by the control plane release manager (@ebeahan or @pierrehilbert). They need to provide the SHA of the build candidate for the release and the version number. The team could choose to trigger it another way in the future as long as they can pass it those two pieces of information.

image

It automatically opens a PR adding the generated `changelog.yml` file, cleaning up fragments, and adding the relevant Markdown release note files. Here's the outcome from an earlier test: https://github.com/colleenmcginnis/elastic-agent/pull/2.

A couple notes:
* The `@colleenmcginnis` in the [test PR](https://github.com/colleenmcginnis/elastic-agent/pull/2) will be replaced with the username of whoever triggered the job.
* Currently the action won't automatically run the preview build because (I _think_) using the default GitHub action token restricts you from using that action to trigger another action. I'm not sure if there's an alternative we could use in this repo. 

## Why is it important?

This will ensure that release notes are always generated using the latest version of elastic-agent-changelog-tool.

## Checklist

## Disruptive User Impact

N/A

N/A

## Related issues

N/A


colleenmcginnis:
@ebeahan @pierrehilbert once I get a ✅  that this looks good, I'll open similar PRs in fleet-server and beats.

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+          ./bin/elastic-agent-changelog-tool cleanup
+          ./bin/elastic-agent-changelog-tool render --version ${{ inputs.version }}
+      - name: Get minor
+        id: get-minor
+        uses: actions/github-script@v8
+        with:
+          script: return "${{ inputs.version }}".replace(/^(\d+\.\d+).+$/m, "$1")
+          result-encoding: string
+      - name: Open PR
+        uses: peter-evans/create-pull-request@v7
```
v1v: Pin actions to a full-length commit SHA:
- https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

```suggestion
        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e #v7.0.8
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          commit-message: add the ${{ inputs.version }} ${{ env.product }} release notes
+          branch: ${{ inputs.version }}-release-notes
+          base: ${{ steps.get-minor.outputs.result }}
+          add-paths: |
+            changelog/
+            docs/
+          labels: ${{ env.labels }}
```
v1v: `nit`: since `labels` env variable is not used anywhere else, wonder if it could be added here?

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+  build-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.bc_commit_sha }}
+      - name: Generate release notes
+        run: |
```
v1v: `nit`: i'm a big fan of using make for containing all the steps rused by the CI, it helps users to run the same as the CI without the need to copy and paste the snippet from the GH action

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+    inputs:
+      bc_commit_sha:
+        description: 'Commit SHA for the BC'
+        type: string
+        required: true
+      version:
+        description: 'Version to be released (for example, `9.2.1`)'
+        type: string
+        required: true
+
```
v1v, line 14: Add default top-level permission, to help to follow the least-permissive access

```suggestion
permissions:
  contents: read
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+env:
+  product: Elastic Agent
+  labels: docs,forwardport-main,release,skip-changelog,Team:Docs
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
v1v: why is this needed? IIRC, that's when using the `gh` cli, but in general, I think it should not be needed, but I might be wrong here

pierrehilbert (dismissed):

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Checkout
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.bc_commit_sha }}
+      - name: Generate release notes
+        run: |
+          GOBIN=$PWD/bin go install github.com/elastic/elastic-agent-changelog-tool@latest
+          ./bin/elastic-agent-changelog-tool build --version ${{ inputs.version }}
+          ./bin/elastic-agent-changelog-tool cleanup
+          ./bin/elastic-agent-changelog-tool render --version ${{ inputs.version }}
```
v1v: [Use an intermediate environment variable](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable) for `inputs.version`



```suggestion
          ./bin/elastic-agent-changelog-tool build --version "${VERSION}"
          ./bin/elastic-agent-changelog-tool cleanup
          ./bin/elastic-agent-changelog-tool render --version "${VERSION}"
        env:
          VERSION: ${{ inputs.version }}
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+
+env:
+  product: Elastic Agent
+  labels: docs,forwardport-main,release,skip-changelog,Team:Docs
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  build-changelog:
+    runs-on: ubuntu-latest
+    steps:
```
v1v: Add the required permissions for the GH_TOKEN at the job level to ensure the PR can be created regardless with the least-permissive access.

```suggestion
    permissions:
      contents: write
      pull-requests: write
    steps:
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
+name: Generate release notes
```
v1v, line 1: Wonder whether the ownership for this GH Action could be added to the CODEOWNERS? Or are we good if we, robots, are the ones?

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,58 @@
...
+            docs/
+          labels: ${{ env.labels }}
+          title: "Add ${{ env.product }} ${{ inputs.version }} release notes"
+          body: |
+            Adds ${{ env.product }} ${{ inputs.version }} release notes.
+
+            ## Checklist
+
+            - [ ] @${{ github.triggering_actor }} review this PR to make sure the correct items are included
+            - [ ] @elastic/ingest-docs-team review the language
```
v1v: wonder if CODEOWNERS could help here as well so those files are owned by the ingest-docs-team?

v1v (commented):
This is a great feature. I provided comments on our current security posture and suggestions for improvements when running this locally.

.github/workflows/release-notes.yml
colleenmcginnis: It could be! I put it in an env variable because I'm going add the same action to two other repos and the labels will be different in those repos so I was trying to isolate all the information that needs to be customized per repo.

v1v: `nit`: maybe we can create a composite GitHub action instead of mutliple workflows with the same implementation?

.github/workflows/release-notes.yml
colleenmcginnis: I'm not sure what you mean, but I'm interested in learning! Can you expand on what would need to be done or point me to an example?

.github/workflows/release-notes.yml
colleenmcginnis: I added it with the intention that it would be used by elastic-agent-changelog-tool ([docs](https://github.com/elastic/elastic-agent-changelog-tool/blob/main/docs/github-authentication.md)).

v1v:
> Currently the action won't automatically run the preview build because (I think) using the default GitHub action token restricts you from using that action to trigger another action. I'm not sure if there's an alternative we could use in this repo.


There are two alternatives here:

a) using the agnostic GitHub token created by the Platform recently
b) using https://github.com/actions/create-github-app-token and installing one of our GitHub apps


the current one, wil need to close/reopen the PR to trigger a build

.github/workflows/release-notes.yml
colleenmcginnis: Well you definitely had good feedback in this review. 🙂  I'm not sure if @ebeahan or @pierrehilbert have an opinion here.

.github/workflows/release-notes.yml
colleenmcginnis: They should already be owned by the ingest docs team (that's why I didn't auto-add any reviewers). I added the checklist to make sure it's clear to the dev team and the docs team who is responsible for what and in what order.

.github/workflows/release-notes.yml
colleenmcginnis: That would definitely be preferable! This started as a proof of concept built by me (someone with limited experience with GitHub actions) to get something out there before the next scheduled release, but we _could_ move the logic to elastic/elastic-agent-changelog-tool and reuse it in this repo, fleet-server, and beats. @v1v if I gave that a try, would you be up for reviewing it?

karenzone (dismissed):
LGTM from my PoV. Looking forward to getting this implemented. 
Thanks, @colleenmcginnis

.github/workflows/release-notes.yml
v1v: I'd suggest moving the env variable closer to the consumer, in other words, instead of being available globally only in the step where it runs the `elastic-agent-changelog-tool`

.github/workflows/release-notes.yml
v1v: Absolutely!

I suggest transferring the logic from the workflow step labelled "Generate release notes" into a [Makefile goal](https://www.gnu.org/software/make/manual/html_node/Goals.html). 

This approach will centralise the steps in one location, making it easier to execute both locally and in the CI without needing to type multiple commands in the terminal.

In this case, there is already an existing Makefile in the repository.
- https://github.com/elastic/elastic-agent/blob/main/Makefile

Therefore, you could create the make goal and the env variable storing the release notes version.

```bash
# Support the release notes version with the following environment variable.
RELEASE_NOTES_VERSION?=

## generate-release-notes : Generates the release notes for the given release version
.PHONY: generate-release-notes
generate-release-notes:
	@GOBIN=$PWD/bin go install github.com/elastic/elastic-agent-changelog-tool@latest
	./bin/elastic-agent-changelog-tool build --version $(RELEASE_NOTES_VERSION)
	./bin/elastic-agent-changelog-tool cleanup
	./bin/elastic-agent-changelog-tool render --version $(RELEASE_NOTES_VERSION)
```

Finally you can run it locally or in the CI by typing:

```bash
$ make generate-release-notes RELEASE_NOTES_VERSION=1.2.3
```

.github/workflows/release-notes.yml
v1v: Sure thing!

.github/workflows/release-notes.yml
v1v: However if you use a GitHub composite action as mentioned at https://github.com/elastic/elastic-agent/pull/10931/files#r2481513774 then you can skip this comment :)

.github/workflows/release-notes.yml
ebeahan: I think it makes sense to set `ingest-docs` as owners and approvers.

@pierrehilbert wdyt?

ebeahan: We can also add `ingest-docs` as a co-codeowner so we're all added to review but no one's blocking.

pierrehilbert: Agree on that!

.github/workflows/release-notes.yml
colleenmcginnis: Here's my attempt 🙂  https://github.com/elastic/elastic-agent-changelog-tool/pull/223

v1v (approved):

v1v (dismissed):
Thanks @colleenmcginnis for attending my suggestions, this is brilliant!! 👏

ebeahan (dismissed):

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,26 @@
...
+    uses: elastic/elastic-agent-changelog-tool/.github/workflows/generate-release-notes.yml@main
+    with:
+      product: Elastic Agent
+      labels: docs,forwardport-main,release,skip-changelog,Team:Docs
+      bc_commit_sha: ${{ inputs.bc_commit_sha }}
+      version: ${{ inputs.version }}
+      release_manager: ${{ github.triggering_actor }}
+    permissions:
+      contents: write
+      pull-requests: write
```
v1v: Requires: https://github.com/elastic/elastic-agent-changelog-tool/pull/225

```suggestion
      contents: read
      id-token: write
```

v1v (approved):

karenzone (approved):
Thank you!

pierrehilbert (approved):

PR #10945, opened 2025-10-31
Revert unit test runners VMs to n2-standard-8

pchila:
## What does this PR do?
Change Unit tests buildkite runner for `Unit tests - Ubuntu 22.04` pipeline step to `n2-standard-8` to avoid buildkite agents being lost because of memory spikes.

## Why is it important?
To reduce CI flakiness introduced by https://github.com/elastic/elastic-agent/pull/10633

## Checklist

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,7 +37,7 @@ steps:
         agents:
           provider: "gcp"
           image: "${IMAGE_UBUNTU_2204_X86_64}"
-          machineType: "n2-highcpu-8"
+          machineType: "n2-standard-8"
         retry:
           automatic:
             limit: 1


v1v (approved):

.buildkite/pipeline.yml
```diff
@@ -37,7 +37,7 @@ steps:
         agents:
           provider: "gcp"
           image: "${IMAGE_UBUNTU_2204_X86_64}"
-          machineType: "n2-highcpu-8"
+          machineType: "n2-standard-8"
```
dliappis, line 40: Drive-by comment:

If the intention of was to benefit from more vCPUs but keeping the same amount of RAM, it'd be better served by https://gcloud-compute.com/n2-highcpu-32.html.

To keep the same cpu/mem ratio, but benefit from much more powerful CPUs, you could instead go a couple of generations newer i.e. https://gcloud-compute.com/n4-standard-8.html -- **caveats**: the newer a generation, the higher the chance of hitting out of capacity issues when launching the VMs/workers. It'll also be slightly more expensive `$363` vs `$442` per month with sustained user discount.

It's all about tradeoffs.

pchila: The intent was the opposite: use the `highcpu` variant to reduce the amount of RAM per CPU core as RAM was less that 50% used.
Here we are reverting to the `standard` variant as it seems that there are some RAM consumption spikes while building go binaries in parallel (go test has a default parallelism equal to the numbers of cores) that seem to cause the loss of the agent by buildkite

dliappis: > The intent was the opposite: use the `highcpu` variant to reduce the amount of RAM per CPU core as RAM was less that 50% used.

Thanks for the clarification! However I wonder reading https://github.com/elastic/elastic-agent/pull/10633

> This PR tries to bring down the total run time for elastic-agent CI on buildkite when running on PRs.

How would the intention to bring down the total time be achieved by just reducing RAM given that n2-highcpu-8 and n2-standard-8 have the same amount of CPUs (`8`)?

Issue #10909, opened 2025-10-28
[Fleet] Invalid token causes agent migration to hang and block new attempts


rseldner:
**Kibana version:**
9.2.0.

**Elasticsearch version:**
9.2.0

**Original install method (e.g. download page, yum, from source, etc.):**
ESS/ECH

**Describe the bug:**
The agent migration feature sort of breaks if you specify an invalid/revoked token.
It gets stuck on that action.
Subsequent migration attempts do nothing.

No impact if you:
- restart the agent itself
- restart Kibana
- restart fleet
- cancel the migration agent action

However, I can get it working again if I switch the agent to a different policy, restart the agent, and try the migration again.

I reproduced this on two separate clusters where I was testing the feature and migrating the agent back and forth.

This does not seem to occur on all types of failures.  I forced a connectivity problem (disabled routing to the new fleet server) and the migration task failed and completed, as one would expect, and allowed me to successfully retry.

Sidenote:
Sometimes the agent would stop checking into the original fleet and go offline (though ingest continued to work) in which case an agent restart on the host was needed to bring it back online.

**Steps to reproduce:**
1. Install and Enroll elastic agent into fleet
2. Attempt to migrate the agent but specify an known invalid string for the token (any arbitrary text will do)
    - This failed migration will indefinitely remain in an IN_PROGRESS state
3. Attempt to migrate the agent again using a valid token
    - this action will never run and also remain IN_PROGRESS indefinitely
4. Attach the agent to any other agent policy
5. retry the migration (this succeeds)

**Expected behavior:**
Failed agent tasks should not run indefinitely
Subsequent migration attempts should be able to run without any extra shenanigans

**Screenshots (if relevant):**
Sorry for the size of the screenshot.  I had to zoom out a lot to capture the info from the screenshot
<img width="1729" height="978" alt="Image" src="https://github.com/user-attachments/assets/a412b72a-19c2-4f37-a4d7-7d2db8cb7a2c" />

Agent Action cancellation request succeeded...

<img width="950" height="125" alt="Image" src="https://github.com/user-attachments/assets/a9588521-3799-48c2-87a5-5dafef4d00c7" />

but continues `IN_PROGRESS` as do the subsequent corrected attempts

<img width="932" height="685" alt="Image" src="https://github.com/user-attachments/assets/1d5c4a2d-2116-4347-8388-938bfde01bd8" />

**Provide logs and/or server output (if relevant):**
[...]

PR #10946, opened 2025-10-31
Enrollment should not be infinite when triggered from action

michalpristas:
When enrollment fails it retries indefinitely. This is fine when triggered from CLI but when running from Action it can lead to confusing outcomes (e.g Migration never finishes)

This PR addresses this in 2 ways
First it sets 3 retries for action triggered enrollments
Second it handles 401 as something that we should not retry on.

Previously indefinitely in agent activity
image

With fix:
image

Fixes: https://github.com/elastic/elastic-agent/issues/10909

### DIFF
--- a/internal/pkg/agent/application/enroll/enroll.go
+++ b/internal/pkg/agent/application/enroll/enroll.go
@@ -17,8 +17,6 @@ import (
 
 	"gopkg.in/yaml.v2"
 
-	"github.com/DataDog/datadog-agent/pkg/util/log"
-
 	"github.com/elastic/elastic-agent-libs/transport/tlscommon"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/filelock"
 	"github.com/elastic/elastic-agent/internal/pkg/agent/application/info"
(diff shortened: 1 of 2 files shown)


internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -682,6 +682,7 @@ func (c *Coordinator) Migrate(
 		options,
 		store,
 		backoffFactory,
+		3, // try enrollment for 5 times and fail
```
ycombinator: One of these is not like the other 😄 😉

internal/pkg/agent/cmd/enroll_cmd.go
```diff
@@ -176,7 +176,9 @@ func (c *enrollCmd) Execute(ctx context.Context, streams *cli.IOStreams) error {
 		enrollDelay,
 		*c.options,
 		c.configStore,
-		c.backoffFactory)
+		c.backoffFactory,
+		-1, // try indefinitely, let user cancel the action
```
ycombinator, line 180: Let's make the `-1` a constant like `InfiniteRetries` or `InfiniteAttempts` or something.  Will help readability in the `RETRYLOOP` code in `enroll.go` as well.

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -57,6 +57,7 @@ func EnrollWithBackoff(
 	options EnrollOptions,
 	configStore saver,
 	backoffFactory func(done <-chan struct{}) backoff.Backoff,
+	maxRetries int,
```
ycombinator: Prefer `maxAttempts` as retries could be misinterpreted as always making one attempt first and then retrying N times.

internal/pkg/agent/application/enroll/enroll.go
```diff
@@ -92,17 +93,19 @@ func EnrollWithBackoff(
 	signal := make(chan struct{})
 	defer close(signal)
 	backExp := backoffFactory(signal)
+	retryNo := 0
 
 RETRYLOOP:
```
ycombinator, line 108: Just noticed there are no unit tests on this file.  Would be an improvement to maybe extract the `RETRYLOOP` block into its own helper function and unit test that for the various cases in the switch.

internal/pkg/agent/application/coordinator/coordinator.go
michalpristas: yeah i reduced, 5 took a lot of time, thanks

PR #10959, opened 2025-10-31
Updating otel gateway config sample

LikeTheSalad:
## What does this PR do?

Adding the new `elasticapm` processor to the rest of signals in the gateway configuration sample.

## Why is it important?

The new processor, `elasticapm`, was created to support all signals, not only traces. However, the outcome is the same as if it only supported traces if we don't encourage users to add it to other signals as well.

## Checklist

- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~

## Related issues

-

## Questions to ask yourself

- ...

### DIFF
--- a/changelog/fragments/1761921215-otel-gateway-config-update.yaml
+++ b/changelog/fragments/1761921215-otel-gateway-config-update.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: other
+
+# Change summary; a 80ish characters long description of the change.
+summary: Adding the new "elasticapm" processor to the rest of signals in the gateway configuration sample.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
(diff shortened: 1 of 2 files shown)


theletterf (approved):

theletterf:
@LikeTheSalad I think we should at least backport to 9.2. To which versions can this apply?

LikeTheSalad:
> @LikeTheSalad I think we should at least backport to 9.2. To which versions can this apply?

Good point, I've added the label. Thanks!

theletterf:
Would be nice to roll this out in 9.2.1 I guess, and first get a review from the Collector team.

gregkalapos (commented):
After seeing the other discussion from the internally linked repo, I'm not sure we want to do this either.

This would mean that every single log and metric will be enriched with basically apm related fields. So let's say if a service only sends otel logs, and don't really use apm, then we'd still add a bunch of fields, which may not be needed. 

@LikeTheSalad what's the exact use-case here? 

I know mobile agents rely on logs for lots of things... is that maybe the reason?

Could we maybe handle this by providing a dedicated pipeline for those agents where we enrich logs and metrics, but then keep the general pipeline without this enrichment?

Also.. should we add to both, or would only log enrichment be enough?

LikeTheSalad:
> I know mobile agents rely on logs for lots of things... is that maybe the reason?

Yes. The mobile UI dashboards rely on enriched logs.

> Could we maybe handle this by providing a dedicated pipeline for those agents where we enrich logs and metrics, but then keep the general pipeline without this enrichment?

I'm not sure about the deeper details of the EDOT Collector, but I guess as long as there's a way for users to send their mobile telemetry data in a way that logs are also enriched, it works for me.

> Also.. should we add to both, or would only log enrichment be enough?

For mobile use cases, logs are enough. And I can't foresee a use case that would need enriched metrics. The reason I added them is because the `elasticapm` processor supports metrics too, so it seems likely that since we added that support, we should advertise it in the docs.

ebeahan:
Cleaning up the repo's open PR list by closing stale items.

Feel free to reopen when work is ready to continue.

Issue #10423, opened 2025-10-08
[beats receivers] Allow specifying and preserving exporter configuration in Elastic Agent outputs that is preserved through configuration translation


cmacknz:
- Relates https://github.com/elastic/elastic-agent/issues/10364

There are several cases, mostly around debugging and troubleshooting, where it would valuable for us to be able to specify exporter configuration entries directly in an Elastic Agent output configuration block that are preserved through the translation process or potentially even take precedence over the translated values.

We can imagine something like the following where the `exporter` sub-section is ignored when beats receivers is not in use:

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    api_key: "example-key"
    preset: balanced
    otel:
      beatsauth:
        enabled: false
      exporter:
        include_source_on_error: true # Preserved during config translation
        log_failed_docs_input: false # Override the translated log_failed_docs_input: true added in https://github.com/elastic/elastic-agent/pull/9341
```
```

**Comment by cmacknz** [comment]
We may want a way to bypass the internally generated beatsauth extension as well, so we'd need a syntax for that, maybe something like:

```
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    api_key: "example-key"
    preset: balanced
    otel:
      beatsauth:
        enabled: false
      exporter:
        include_source_on_error: true # Preserved during config translation
        log_failed_docs_input: false # Override the translated log_failed_docs_input: true added in https://github.com/elastic/elastic-agent/pull/9341
```

leehinman:
One possibility to "bypass" the beats auth extension might be to have the extension not replace the `exporters`  section if the `exporters` section already exists.  That way you can pass in any valid exporter config and the extension doesn't have to parse or know anything about it.

blakerouse:
@cmacknz I see that you defined `beatsauth` and `exported` as sub-keys of `otel`. Are you expecting all fields here to be defined or would you expect anything to just be translated? Basically everything that is under `beatsauth` would go to the `beatsauth` extension, and everything under `exporter` would go to the exporter. Or do you believe that every field should be defined.

cmacknz:
> Basically everything that is under beatsauth would go to the beatsauth extension, and everything under exporter would go to the exporter. 

[...]

PR #10992, opened 2025-11-03
Allow overriding otel exporter and extension configuration for outputs

swiatekm:
## What does this PR do?

Allows otel component configurations generated from output configuration to be overridden. In the output configuration, we have a new optional `otel` section:

```yaml
outputs:
  default:
    type: elasticsearch
    hosts: [127.0.0.1:9200]
    api_key: "example-key"
    otel:
      extensions:
        beatsauth:
          timeout: "60s"
      exporter:
        include_source_on_error: true
```

`otel.extensions.<extension_name>` contains configuration for the generated extension, while `otel.exporter`, for the exporter.

These configurations are merged into the generated configuration and take precedence over it. There aren't any validity checks here, so the feature should be used with care. It's only intended as a stopgap measure to mitigate unforeseen problems.

## Why is it important?

If it turns out there's an issue with the configuration translation, we want to have an escape hatch to mitigate it for affected customers. 

## Checklist

Build the agent package, then run it with the following configuration:

```yaml
agent:
  logging:
    to_stderr: true
    level: debug
  monitoring:
    enabled: false
inputs:
- data_stream:
    namespace: default
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  type: system/metrics
  use_output: default
  _runtime_experimental: otel
outputs:
  default:
    username: elastic
    password: placeholder
    hosts:
    - 127.0.0.1:9200
    type: elasticsearch
    otel:
      exporter:
        sending_queue:
          queue_size: 50
      extensions:
        beatsauth:
          timeout: "60s"
```

Then collect diagnostics and look at the final otel configuration. It should contain the output overrides.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10423

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -1130,7 +1130,6 @@ func TestSensitiveLogsESExporter(t *testing.T) {
 		},
 		Stack: &define.Stack{},
 	})
-	t.Skip("Enable this test after https://github.com/elastic/elastic-agent/issues/10423 is implemented with appropriate config")
 	tmpDir := t.TempDir()
 	numEvents := 50
 	// Create the data file to ingest
@@ -1184,6 +1183,10 @@ outputs:
     type: elasticsearch
     hosts: [{{.ESEndpoint}}]
     api_key: "{{.ESApiKey}}"
+    otel:
+      exporter:
+        telemetry:
+          log_failed_docs_input: true
 agent:
   monitoring:
     enabled: true


cmacknz:
You can Unskip the test below now after enabling `telemetry::log_failed_doc_input` using this override feature and it should pass.

https://github.com/elastic/elastic-agent/blob/a24b6fa061b24cd7be826bec6eae7bdfa4ec0590/testing/integration/ess/beat_receivers_test.go#L1117-L1133

cmacknz (approved):
Tested locally, works as expected. Thanks!

PR #11016, opened 2025-11-04
Fix assertion in otel shutdown test

swiatekm:
## What does this PR do?

Fixes an assertion in an integration test.

## Why is it important?

This makes the test less flaky. There's been some problems with it, see https://buildkite.com/elastic/elastic-agent/builds/29862#019a4b3e-776e-4c68-a2b6-c67253f68fc2 for example.

### DIFF
--- a/testing/integration/ess/otel_test.go
+++ b/testing/integration/ess/otel_test.go
@@ -140,7 +140,9 @@ service:
 
 	// stop the collector and check that it emitted logs indicating a graceful shutdown
 	require.NoError(t, cmd.Process.Signal(os.Interrupt))
-	require.NoError(t, cmd.Wait())
+	if waitErr := cmd.Wait(); waitErr != nil {
+		assert.ErrorContains(t, waitErr, "signal: interrupt")
+	}
 	assert.Contains(t, output.String(), "Shutdown complete")
 }
 


swiatekm:
In the interest of going faster, I'm going to manually backport this change to https://github.com/elastic/elastic-agent/pull/10996 and https://github.com/elastic/elastic-agent/pull/10997.

testing/integration/ess/otel_test.go
```diff
@@ -140,7 +140,9 @@ service:
 
 	// stop the collector and check that it emitted logs indicating a graceful shutdown
 	require.NoError(t, cmd.Process.Signal(os.Interrupt))
-	require.NoError(t, cmd.Wait())
+	if waitErr := cmd.Wait(); waitErr != nil {
+		assert.ErrorContains(t, waitErr, "signal: interrupt")
+	}
```
pchila, line 145: I am not sure I understand: we are changing the assertion from "it should not exit with an error code" to "it's ok to receive an error as long as it contains `signal: interrupt`" ?

Why is that? Is a "signal: interrupt" a graceful shutdown like the log above implies ?

swiatekm: Whether the shutdown is graceful or not is checked by looking at logs. The original test had the incorrect assumption that sending a SIGINT to the process will cause `Wait()` to return without error, but it can in fact return an `exec.ExitError` containing the process state and signal. I'm not sure why the test passed to begin with, to be honest - maybe there's some kind of race condition involved.

But in any case, this isn't important to what the test is actually checking, so I'd like to fix it before investigating why it's flaky in the first place. I could also just skip the check, too.

pchila: [Wait() documentation](https://pkg.go.dev/os/exec#Cmd.Wait) looks pretty clear-cut: assuming that the command is running, we get an error when the exit code of the process is a failing one.
Circling back to the original question: should the process that handles SIGINT gracefully exit with a failure exit code ?

swiatekm: Ah, I see what you mean. Yeah, it should exit with success, so the fact that it doesn't sometimes is evidence that something's not entirely right in there.

swiatekm:
This test failure hasn't come up again, so I'm going to close this PR. Truthfully, I have no idea how it could've happened, so I'm willing to chalk it up to some kind of weird test environment problem.

Issue #10821, opened 2025-10-24
[beats receivers] Add environment variables to the elastic-agent container command to enable/disable beats receivers


cmacknz:
Add two new environment variables to the `elastic-agent container` command:

1. `ENABLE_BEATS_RECEIVERS` as a boolean that when true uses Beats receivers for everything in the policy that can support Beats receivers. This should functionally address https://github.com/elastic/elastic-package/issues/2975 once elastic-package can support it.
2. `ENABLE_BEATS_RECEIVERS_MONITORING` as a boolean that is a short cut for setting the `_monitoring_experiemental: otel` setting in the `agent.monitoring` configuration, controlling whether beats receivers are used for self-monitoring or not.
3. Update https://github.com/elastic/elastic-agent/blob/main/docs/hybrid-agent-beats-receivers.md to document the two new variables for the container command.

mauri870:
Posting a short update. I tried to get my bearing on this task, but so far I haven't been able to reach a satisfactory implementation. Discussed this a bit with @swiatekm today and we reached similar conclusions.

I've opened a PR https://github.com/elastic/elastic-agent/pull/11018 , but the implementation raises more questions than it answers. Testing this is quite time-consuming, so I’ve spent most of the time preparing integration tests, at least it's now straightforward to test. 

The main open questions I have are:

1. Should these envs be a force setting or more like a default setting? 
The policy ends up replacing them if we set them when the agent starts up. If these need to apply after the loaded policy, the logic might have to be handled deeper in the codebase, which adds complexity.

2. Should `ENABLE_BEATS_RECEIVERS` also set `agent.monitoring` to otel runtime?
Since the monitoring components are in the components list, looping over the components and setting each one to otel eventually updates monitoring too.

3. The idea of reading environment variables inside of EA control plane code does not seem right
Today most envs for the container subcommand are read once in the container subcommand and then translated to config options or command line flags which are then passed to EA. Another downside is that we cannot differentiate if the env came from `elastic-agent container` or not looking at the env alone.

4. Is `ENABLE_BEATS_RECEIVERS_MONITORING` useful at all? Nowadays monitoring runtime is always otel.

---

[...]

PR #11018, opened 2025-11-04
otel: add env variable to set agent.monitoring runtime

mauri870:
## What does this PR do?

This PR adds a new experimental environment variable, AGENT_MONITORING_RUNTIME_EXPERIMENTAL, which overrides the runtime used by `agent.monitoring` components. This env takes precedence over the value set in configuration files, but has lower precedence than the policy when the agent is enrolled in Fleet.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10821


internal/pkg/core/monitoring/config/config.go
```diff
@@ -111,6 +112,13 @@ type BufferConfig struct {
 
 // DefaultConfig creates a config with pre-set default values.
 func DefaultConfig() *MonitoringConfig {
+	monRuntimeManager := DefaultRuntimeManager
+	monRuntimeEnv := os.Getenv("AGENT_MONITORING_RUNTIME_EXPERIMENTAL")
+	switch monRuntimeEnv {
+	case ProcessRuntimeManager, OtelRuntimeManager:
+		monRuntimeManager = monRuntimeEnv
+	}
+
```
mauri870, line 121: For reviewers: The right place for this logic should be [`containerCfgOverrides`](https://github.com/elastic/elastic-agent/blob/b74942a/internal/pkg/agent/cmd/container.go#L807), but unfortunately due to config reloading the value is discarded. This is an experimental environment variable that will eventually be removed, so hopefully it is fine to keep it here.

mauri870:
@swiatekm Assigned you as a reviewer since @blakerouse and @michalpristas are on rotation this week. The PR is quite simple and we already agreed on the approach, so the review should be quick. Thanks!

Issue #10965, opened 2025-10-31
archive-proxy is compiled for the host architecture and not the target architecture


cmacknz:
The binary at https://github.com/elastic/elastic-agent/tree/main/wrapper/windows/archive-proxy is compiled for the wrong architecture and ends up not being executable when trying to install Elastic Agent on a Windows X86 machine when cross compiling from an ARM Mac.

```sh
# On an M4 Mac
EXTERNAL=true SNAPSHOT=true PACKAGES=zip PLATFORMS=windows/amd64 mage package
~/Downloads/rdp_share/elastic-agent-9.2.1-SNAPSHOT-windows-x86_64 ······································ 03:49:40 PM
❯ file elastic-agent.exe
elastic-agent.exe: PE32+ executable (console) Aarch64, for MS Windows
~/Downloads/rdp_share/elastic-agent-9.2.1-SNAPSHOT-windows-x86_64 ······································ 03:49:45 PM
❯ file data/elastic-agent-c2c359/elastic-agent.exe
data/elastic-agent-c2c359/elastic-agent.exe: PE32+ executable (console) x86-64 (stripped to external PDB), for MS Windows
```

cmacknz:
https://github.com/elastic/elastic-agent/blob/c2c3599f474e335f08521f156941526539adcbc0/magefile.go#L294

cmacknz:
Changed in https://github.com/elastic/elastic-agent/pull/9751

PR #9751, opened 2025-09-04
Add Windows/arm64 support for Agent

dwhyrock:
## What does this PR do?
The goal of this PR is to get Elastic Agent building and packaging a Windows ARM artifact.

Both Beats and Endpoint are currently producing Windows/arm64 artifacts that can be packaged.

This is producing a valid Elastic Agent build that has Agentbeat, OsQuerybeat, and Endpoint components packaged with it: https://buildkite.com/elastic/elastic-agent/builds/27828#01999b48-3d7d-424d-b2b1-aa1c97d29b09

**NOTE**: There is currently no CI support for Windows/arm64, so testing in CI has not been added.  Also, this support for Windows/arm64 is not fully supported for customer use, so there will be no changelog added since it won't be published on the download pages until it's fully vetted/tested/etc.

This PR does the following:
 * Updates packaging step names in `.buildkite/bk.integration.pipeline.yml` to reflect windows architecture
 * Adding a `packaging-windows-arm64` step to `.buildkite/integration.pipeline.yml`
 * Added `windows/arm64` as a `PLATFORMS` list item for `build-dra-snapshot-arm` step in `.buildkite/pipeline.elastic-agent-binary-dra.yml`
 * Added `windows/arm64` as a `PLATFORMS` list item to `package-elastic-agent-arm` step in `.buildkite/pipeline.elastic-agent-package.yml`
 * Added `windows/arm64` case for crossbuilding in `dev-tool/mage/crossbuild.go`
 * Added `windows/arm64` entries to various structs in `dev-tools/mage` and `dev-tools/packaging`
 * Added `windows-arm64` as valid `platforms` item in `dev-tools/packaging/packages.yml`
 * Updated `fleet-server` to support all the valid platforms, except the new `windows/arm64` one
 * Updated files in `pkg/` to support `windows/arm64` and not error out when encountering it
 * Updated `specs/endpoint-security.spec.yml` to not error out if `windows/arm64` is the OS/arch
 * Updated `magefile.go` to support `windows/arm64` in building and packaging logic

## Why is it important?
Getting lots of demand for the ability to run on Windows OS w/ arm64 architecture.

## Checklist

- [x] I have read and understood the [pull request guidelines](https://github.com/elastic/elastic-agent/blob/main/CONTRIBUTING.md#pull-request-guidelines) of this project.
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
[...]

PR #11054, opened 2025-11-05
Compile the Windows archive proxy for the target GOARCH instead of the host GOARCH

cmacknz:
- Closes https://github.com/elastic/elastic-agent/issues/10965

When support for windows/arm builds was added in https://github.com/elastic/elastic-agent/pull/9751 the Go archive proxy was switched from a fixed `amd64` GOARCH to the host GOARCH. This works in CI which doesn't cross-compile for Windows but broke cross-compiling from darwin/arm64 to windows/amd64 for anyone on a Mac. This PR fixes that.

To test:

```sh
❯ EXTERNAL=true SNAPSHOT=true PACKAGES=zip PLATFORMS=windows/amd64,windows/arm64 mage package

# ...

❯ file build/distributions/elastic-agent-9.3.0-SNAPSHOT-windows-x86_64/elastic-agent.exe
build/distributions/elastic-agent-9.3.0-SNAPSHOT-windows-x86_64/elastic-agent.exe: PE32+ executable (console) x86-64, for MS Windows

❯ file build/distributions/elastic-agent-9.3.0-SNAPSHOT-windows-arm64/elastic-agent.exe
build/distributions/elastic-agent-9.3.0-SNAPSHOT-windows-arm64/elastic-agent.exe: PE32+ executable (console) Aarch64, for MS Windows

```

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -260,7 +260,7 @@ func (Build) GenerateConfig() error {
 	return sh.Copy(filepath.Join(buildDir, configFile), filepath.Join(metaDir, configFile))
 }
 
-// WindowsArchiveRootBinary compiles a binary to be placed at the root of the windows elastic-agent archive. This binary
+// windowsArchiveRootBinaryForGoArch compiles a binary to be placed at the root of the windows elastic-agent archive. This binary
 // is a thin proxy to the actual elastic-agent binary that resides in the data/elastic-agent-{commit-short-sha}
 // directory of the archive.
 func (Build) windowsArchiveRootBinaryForGoArch(goarch string) error {


magefile.go
```diff
@@ -263,8 +263,8 @@ func (Build) GenerateConfig() error {
 // WindowsArchiveRootBinary compiles a binary to be placed at the root of the windows elastic-agent archive. This binary
```
ycombinator:
```suggestion
// windowsArchiveRootBinaryForGoArch compiles a binary to be placed at the root of the windows elastic-agent archive. This binary
```

ycombinator (dismissed):
Just a comment update, otherwise LGTM

magefile.go
cmacknz: Thanks, needs re-review :)

michel-laterman (approved):
lgtm

PR #11155, opened 2025-11-12
Add test for running Log input as Filestream in Beatsreceivers

belimawr:
## What does this PR do?

This PR adds a test that ensures Beatsreceivers ingesting logs can:
- Start running with a Log input configuration and ingest logs
- Stop, then re-start with the feature flag to run Filestream input instead of the Log input
- The Filestream input running from the Log input configuration continues ingesting the file from where the Log input had stopped.

Currently there is no way to run a Beatreceiver directly from a standalone Beats binary, hence the test in this repo.

## Why is it important?
It allows us to run the Filestream input instead of the Log input for a Log input configuration

## Checklist

## Author's Checklist
 - [x] Wait for https://github.com/elastic/beats/pull/46587 to be merged
 - [x] Remove overwrites
   - [x] `dev-tools/notice/overrides.json`
   - [x] `go.mod`
 - [x] Update comments on testing/integration/ess/logfile.go 
 - [x] Replace `testing/integration/ess/logfile.go` with the functionality provided by https://github.com/elastic/elastic-agent-libs/pull/369

~~## Disruptive User Impact~~

### Running integration tests
Run the test:
1. Package the Elastic Agent for your OS
    ```
    DEV=true SNAPSHOT=true EXTERNAL=true PACKAGES="tar.gz" PLATFORMS=linux/amd64 mage -v package
    ```
2. Run `TestFilebeatReceiverLogAsFilestream` from `testing/integration/ess/otel_log_as_filestream_test.go` 
    ```
    AGENT_VERSION="9.4.0-SNAPSHOT" SNAPSHOT=true TEST_PLATFORMS="linux/amd64" mage -v integration:single TestFilebeatReceiverLogAsFilestream
    ```

### Manual test
Run the Elastic Agent in Otel mode setting `features.log_input_run_as_filestream.enabled: true` for the filebeatreceiver you want to test.

1. Package the Elastic Agent for your OS
    ```
    DEV=true SNAPSHOT=true EXTERNAL=true PACKAGES="tar.gz" PLATFORMS=linux/amd64 mage -v package
    ```
2. Create a log file
    ```
    docker run --rm mingrammer/flog -n 50 > /tmp/flog.log
    ```
3. Extract and run Elastic Agent in otel mode using a configuration like the one below
    ```
    ./elastic-agent otel --config otel.yml
    ```

    <details><summary>otel.yml</summary>
    <p>

    ```yaml
    receivers:
      filebeatreceiver:
        filebeat:
          inputs:
            - type: log
              id: a-unique-id
              allow_deprecated_use: true
              paths:
                - /tmp/flog.log
        output:
          otelconsumer:
        logging:
          level: debug
          selectors:
            - '*'
        features.log_input_run_as_filestream.enabled: true

    exporters:
      elasticsearch:
        user: elastic
        password: changeme
        endpoints:
          - https://localhost:9200
        tls:
          insecure_skip_verify: true
        logs_index: my-index
        sending_queue:
          enabled: true
          wait_for_result: true # Avoid losing data on shutdown
          block_on_overflow: true
        mapping:
          mode: none

    service:
      pipelines:
        logs:
          receivers: [filebeatreceiver]
          exporters: [elasticsearch]
      telemetry:
        logs:
          level: DEBUG
          encoding: json
          disable_stacktrace: true
    ```

    </p>
    </details> 

## Related issues

- Blocked by https://github.com/elastic/beats/pull/46587

## Questions to ask yourself

- ...


belimawr:
`TestOtelLogsIngestion` is failing, there is already a [PR](https://github.com/elastic/elastic-agent/pull/11254) open to fix it.

belimawr:
Keeping CI green is a moving target with all the recent test flakiness and conflicts in `go.mod` every time some change in the Beats import happen.

I'll keep this PR as is until https://github.com/elastic/beats/pull/46587 gets merged and I can put this PR in its final form.

belimawr:
CI is failing because the FIPS stack is not getting ready :/

belimawr:
The last failure was due to BK issues: `Exited with status -1 (agent lost)`. I'm re-running the tests/updating the branch to see if they pass this time.

swiatekm:
Why does this test need to live in the agent repo? At a glance, it doesn't use any agent-specific functionality.

belimawr:
> Why does this test need to live in the agent repo? At a glance, it doesn't use any agent-specific functionality.

Because this is where all the infrastructure to run Elastic Agent, start stacks, VMs, manage multiple OSes, and test them is.

The whole point of the test is to test the beatsreceivers part of the Elastic Agent with the functionality of running Filestream instead of the Log input.

swiatekm:
> > Why does this test need to live in the agent repo? At a glance, it doesn't use any agent-specific functionality.
> 
> Because this is where all the infrastructure to run Elastic Agent, start stacks, VMs, manage multiple OSes, and test them is.
> 
> The whole point of the test is to test the beatsreceivers part of the Elastic Agent with the functionality of running Filestream instead of the Log input.

Doesn't beats have the necessary infrastructure? If I recall us moving a bunch of similar tests to the beats repository recently. @khushijain21 @mauri870 can you chime in on this?

I'm pushing back against this because you're using practically no agent code in this integration test, and the code you are using is exercises by other tests already. The agent integration test suite is already unwieldy, I'd rather not add to it without good reason.

khushijain21:
@belimawr the test you have written is similar to what we have in beats' repo for beatreceivers. [code](https://github.com/elastic/beats/blob/mappingmode2/x-pack/filebeat/tests/integration/otel_test.go#L45)

The necessary testing framework is already available in beats.

belimawr:
If you folks prefer, I can move it to the Beats repo. I don't remember exactly why I wrote this test in the Elastic Agent repo, I recall someone requesting it, but I cannot find the thread right now :man_shrugging: . I'll move it to the Beats repo.

belimawr:
Reopening it after some discussion with @swiatekm. The whole point of this test is to execute the Beatsreceivers the same way an user would and ensure that when we enable the feature flag to run Filestream from a Log input configuration, everything works as expected.

I'll also update the PR description to clarify it.

testing/integration/ess/otel_log_as_filestream_test.go
```diff
@@ -0,0 +1,351 @@
...
+		"Filestream did not reach EOF")
+
+	// Stop Elastic Agent
+	StopElasticAgentOtel(t, cmd, agentLogFile)
+
+	// Ensure there was no data duplication
+	waitEventsInES(100)
+}
+
+func StartElasticAgentOtel(
```
swiatekm: Do you need these functions? If you look at the other pure otel tests  in otel_test.go, they accomplish the same thing by using the integration test framework's fixtures.

swiatekm (commented):
The test itself looks correct to me, but I'd prefer if it used fixtures from our integration test framework instead of adding new ones. This PR makes a lot of changes it doesn't need to make to accomplish what it wants to do, imo. The ES key cleanup is good, but I don't see why you need to touch files outside testing/integration, for example.

testing/integration/ess/otel_log_as_filestream_test.go
belimawr: > Do you need these functions?

Yes, I do. So my test needs to stop the Elastic Agent, change configuration and then start it back again. The problem is that on Windows when I send the signal to stop Elastic Agent, the signal gets propagated to all parent process, which includes the test itself and the buildkite agent, they all die, which leads to odd CI failures.

This function abstracts this part of the process, look at https://github.com/belimawr/elastic-agent/blob/3d67ae8f2929570906657a8bc0f3c878573956e3/testing/integration/ess/otel_log_as_filestream_test.go#L268. It uses the [process](https://github.com/elastic/elastic-agent/tree/main/pkg/core/process) to correctly configure the `exec.Cmd` so stopping it won't kill the parent process.

I looked at `otel_test.go` and other methods in the fixture and client, however I did not find any graceful way to stop the Elastic Agent process.

In this PR I call [`process.Terminate`](https://github.com/elastic/elastic-agent/blob/a54b5b7eda88f90e8364633a0e1e54776e9ac1a3/pkg/core/process/process.go#L179C6-L179C15) that uses the OS specific version to gracefully shutdown Elastic Agent whenever possible.

belimawr:
> The test itself looks correct to me, but I'd prefer if it used fixtures from our integration test framework instead of adding new ones.

Which fixtures exactly are you talking about? Are they `StartElasticAgentOtel` and `StopElasticAgentOtel`? I explained the reasoning form them at https://github.com/elastic/elastic-agent/pull/11155#discussion_r2737240701

> This PR makes a lot of changes it doesn't need to make to accomplish what it wants to do, imo. The ES key cleanup is good, but I don't see why you need to touch files outside testing/integration, for example.

I'm a bit confused here, are you referring to the changes on `pkg/core/process`? I needed the to gracefully stop the Elastic Agent process (which includes Windows with all of its process management particularities). I just exposed the package's functionality in a way I could use instead of re-implementing it locally in the tests package.

I did not find any other way within out integration tests framework to gracefully stop the Elastic Agent. Maybe I missed something? If you know a better way, I can give it a try.

testing/integration/ess/otel_log_as_filestream_test.go
swiatekm: Is there a problem with doing the same thing as [TestOtelFileProcessing](https://github.com/elastic/elastic-agent/blob/3a56b4f11c4196b4625fa12979ecbb95280a3417/testing/integration/ess/otel_test.go#L150)?

```go

        fixture, err := define.NewFixtureFromLocalBuild(t, define.Version(), aTesting.WithAdditionalArgs([]string{"--config", otelConfigPath}))
	require.NoError(t, err)

	ctx, cancel := testcontext.WithDeadline(t, context.Background(), time.Now().Add(10*time.Minute))
	defer cancel()
	err = fixture.Prepare(ctx, fakeComponent)
	require.NoError(t, err)

	// remove elastic-agent.yml, otel should be independent
	require.NoError(t, os.Remove(filepath.Join(fixture.WorkDir(), "elastic-agent.yml")))

	var fixtureWg sync.WaitGroup
	fixtureWg.Add(1)
	go func() {
		defer fixtureWg.Done()
		err = fixture.RunOtelWithClient(ctx)
	}()
```

If you want to stop the collector, you simply cancel the context and wait on the waitgroup. There's no need to work with processes directly.

belimawr: Yes, as I mentioned before, using using the fixture directly does not allow to gracefully shutdown Elastic Agent. When the context passed to `fixture.RunOtelWithClient` is cancelled the process is forcefully killed instead of gracefully shutdown. That causes issues to the test, because we need Filebeat to correctly shutdown.

I even tried the approach you suggested, just to be on the safe side. The whole Elastic Agent process gets forcefully terminated.

swiatekm: That's the behavior we should fix, then. That fixture should try to exit gracefully, wait a configurable amount of time, and only then kill the process.

belimawr: Something like a `Shutdown` method in the fixture? Yeah, I can implement something like that.

belimawr:
The tests are failing because in my last [commit](https://github.com/elastic/elastic-agent/pull/11155/commits/300c7f10b63258187251e142d0c7717b57e0e3c8) I modified `fixture.PrepareAgentCommand` to use `process.Cmd` which allows to gracefully shutdown the child process on Windows, however this is having unexpected effects in other tests, hence the failures.

pkg/testing/fixture.go
```diff
@@ -556,9 +561,17 @@ func (f *Fixture) RunOtelWithClient(ctx context.Context, states ...State) error
 	return f.executeWithClient(ctx, "otel", false, false, false, states...)
 }
 
+// Stop gracefully stops the Elastic Agent process
+func (f *Fixture) Stop() error {
+	f.procMutex.Lock()
+	defer f.procMutex.Unlock()
+	f.stopping = true
+	return f.proc.Stop()
```
swiatekm: I think you need to check whether `f.proc` is set here, and also return an informative error message if `f.installed` is set, similar to what `executeWithClient` does. We don't want to give the impression that you can stop an installed agent this way.

swiatekm (commented):
Other than needing more guards for the fixture addition, this looks good to me.

pkg/testing/fixture.go
belimawr: Thanks for catching that! The installed case didn't cross my mind! I'll fix those soon.

swiatekm (dismissed):

pkg/testing/fixture.go
```diff
@@ -556,9 +561,33 @@ func (f *Fixture) RunOtelWithClient(ctx context.Context, states ...State) error
 	return f.executeWithClient(ctx, "otel", false, false, false, states...)
 }
 
+// Stop gracefully stops the Elastic Agent process that has been started
+// by [RunOtelWithCliet] or [Run].
+// If the Elastic Agent has been installed, or the process
+// has not been started by [RunOtelWithCliet] or [Run],
```
mauri870:
```suggestion
// has not been started by [RunOtelWithClient] or [Run],
```

swiatekm (approved):

mauri870 (approved):
LGTM, and thanks for adjusting the other tests!

v1v:
There are some conflicts with backporting:
- https://github.com/elastic/elastic-agent/pull/13967
- https://github.com/elastic/elastic-agent/pull/13966

@belimawr, I wonder whether this PR could be backported to `9.3` and `8.19`, so we can backport https://github.com/elastic/elastic-agent/pull/13642 without any conflicts.

belimawr:
> There are some conflicts with backporting:
> 
> * [[9.3] (backport #13642) Replace Terraform ESS provisioning with oblt-cli for Windows and Linux but no FIPS #13967](https://github.com/elastic/elastic-agent/pull/13967)
> * [[8.19] (backport #13642) Replace Terraform ESS provisioning with oblt-cli for Windows and Linux but no FIPS #13966](https://github.com/elastic/elastic-agent/pull/13966)
> 
> @belimawr, I wonder whether this PR could be backported to `9.3` and `8.19`, so we can backport #13642 without any conflicts.

On a quick glance, 9.3 contains the feature this PR is testing, so backporting to 9.3 is possible. However, 8.19 does *NOT* contain the feature this PR is testing, therefore it cannot be backported there.

v1v:
@mergifyio backport 9.3

PR #11163, opened 2025-11-13
[beatreceivers] Add support for kerberos

khushijain21:
## What does this PR do?
This PR ensures kerberos configuration is set on beats authextension.

## Why is it important?
To support Kerberos on beatreceivers

## Checklist

## Disruptive User Impact

None

## Related issues

- Relates https://github.com/elastic/beats/issues/46749


internal/pkg/otel/translate/otelconfig.go
```diff
@@ -639,20 +643,27 @@ func getBeatsAuthExtensionConfig(outputCfg *config.C) (map[string]any, error) {
...
+		err = newConfig.SetString("proxy_url", -1, defaultAuthSettings.Transport.Proxy.URL.String())
 		if err != nil {
 			return nil, fmt.Errorf("error settingg proxy url:%w ", err)
 		}
 	}
 
+	if defaultAuthSettings.Kerberos != nil {
+		err = newConfig.SetString("kerberos.auth_type", -1, defaultAuthSettings.Kerberos.AuthType.String())
+		if err != nil {
+			return nil, fmt.Errorf("error settingg kerberos auth type url:%w ", err)
```
pchila:
```suggestion
			return nil, fmt.Errorf("error setting kerberos auth type url:%w ", err)
```

internal/pkg/otel/translate/otelconfig_test.go
```diff
@@ -258,7 +258,6 @@ func TestGetOtelConfig(t *testing.T) {
 				"supported_protocols":     []interface{}{},
 				"verification_mode":       uint64(0),
 			},
-			"timeout": "1m30s",
```
pchila, line 261: Why did the timeout disappear from the expectedExtensionConfig ?

khushijain21: that's a miss, added it back

internal/pkg/otel/translate/otelconfig_test.go
```diff
@@ -1334,11 +1334,40 @@ func TestGetBeatsAuthExtensionConfig(t *testing.T) {
...
+					"password":    "testing",
+					"realm":       "elastic",
+				},
+			},
+			expected: map[string]any{
+				"continue_on_error":       true,
+				"idle_connection_timeout": "3s",
+				"timeout":                 "1m30s",
+				"kerberos": map[string]any{
+					"enabled":          nil,
```
swiatekm: Why is this nil?

internal/pkg/otel/translate/otelconfig.go
```diff
@@ -618,15 +619,18 @@ func BeatDataPath(componentId string) string {
 // getBeatsAuthExtensionConfig sets http transport settings on beatsauth
 // currently this is only supported for elasticsearch output
 func getBeatsAuthExtensionConfig(outputCfg *config.C) (map[string]any, error) {
-	defaultTransportSettings := elasticsearch.ESDefaultTransportSettings()
+
+	defaultAuthSettings := beatsauthextension.BeatsAuthConfig{
```
swiatekm: Considering we modify this value later, it should just be called `authSettings`.

internal/pkg/otel/translate/otelconfig_test.go
khushijain21: Because we don't explicity set enabled to true/false. And kerberos.enabled receives a pointer to bool so 
https://github.com/khushijain21/beats/blob/main/libbeat/common/transport/kerberos/config.go#L50

internal/pkg/otel/translate/output_elasticsearch.go
```diff
@@ -269,6 +268,12 @@ func cfgDecodeHookFunc() mapstructure.DecodeHookFunc {
 				return nil, fmt.Errorf("failed parsing proxy_url: %w", err)
 			}
 			return proxyURL, nil
+		case t == reflect.TypeOf(kerberos.AuthType(0)):
+			authType := kerberos.AuthType(0)
+			if err := authType.Unpack(data.(string)); err != nil {
```
ycombinator, line 277: Should we check first that the `data.(string)` type assertion succeeds before passing its value to `authType.Unpack`?

internal/pkg/otel/translate/output_elasticsearch.go
```diff
@@ -269,6 +268,12 @@ func cfgDecodeHookFunc() mapstructure.DecodeHookFunc {
 				return nil, fmt.Errorf("failed parsing proxy_url: %w", err)
 			}
 			return proxyURL, nil
+		case t == reflect.TypeOf(kerberos.AuthType(0)):
+			authType := kerberos.AuthType(0)
```
ycombinator: Curious why this can't be simply:


```suggestion
			var authType kerberos.AuthType
```

internal/pkg/otel/translate/output_elasticsearch.go
khushijain21: no we don't need to because we check the source type before we proceed here https://github.com/khushijain21/elastic-agent/blob/loadbalancefalse/internal/pkg/otel/translate/output_elasticsearch.go#L247

internal/pkg/otel/translate/otelconfig_test.go
swiatekm: I mean, if kerberos is enabled, shouldn't it be true instead?

khushijain21: Fixed it in recent commit

PR #11164, opened 2025-11-13
Exclude otel components from dependabot

swiatekm:
## What does this PR do?

Excludes elastic's otel components from dependabot updates.

## Why is it important?

These should be updated in lock-step with the otel core upgrades for now.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,8 @@ updates:
       - dependency-name: "helm.sh/helm/*"
     ignore:
       - dependency-name: "github.com/elastic/beats/*"
+      # Otel dependency updates are done in lock-step with Otel core framework updates
+      - dependency-name: "github.com/elastic/opentelemetry-collector-components/*"
     open-pull-requests-limit: 10
 
   - package-ecosystem: "github-actions"


.github/dependabot.yml
```diff
@@ -12,7 +12,6 @@ updates:
       - backport-active-all
     allow:
       - dependency-name: "github.com/elastic/*"
-      - dependency-name: "github.com/elastic/opentelemetry-collector-components/*"
```
pchila, line 15: Shouldn't this be moved under the `ignore`  list ?

blakerouse: Why doesn't `go.opentelemetry.io/collector`, `github.com/open-telemetry/opentelemetry-collector-contrib`, and `github.com/elastic/opentelemetry-lib` need to be ignored?

swiatekm: The way dependabot works is that if you specify `allow`, only dependencies explicitly matched by one of the rules get updated. Then `ignore` can remove from that list. Paolo is right in that we allow `github.com/elastic/*`, so we need to explicitly ignore `github.com/elastic/opentelemetry-collector-components/*`.

swiatekm: Done: https://github.com/elastic/elastic-agent/pull/11164/commits/267994be839df69eb9f91bdd2fd7cac99836404d

v1v (approved):

PR #11186, opened 2025-11-14
Allow setting the otel runtime per input type

swiatekm:
## What does this PR do?

Allows the runtime in which an input runs to be set per input type in a new `agent.internal.runtime` configuration section. There's three levels at which the runtime can be set, in order of increasing priority:

* The default for all input types.
* The default for all input types running in a specific beat.
* The default for a single input type.

Here's how the configuration might look like with all options set:

```yaml
agent:
  internal:
    runtime:
      default: otel
      filebeat:
        default: otel
        filestream: process
      metricbeat:
        default: process
        system/metrics: otel
```

The implementation is almost completely confined to the component package.

## Why is it important?

We intend to progressively make the otel runtime the default in future major releases, but we also want to let users switch back to the process runtime if they encounter bugs. These settings are intended to be transient - eventually, they'll be removed,.

## Checklist

Build agent locally and try playing around with the settings in the following config file:

```yaml
agent:
  logging:
    to_stderr: true
  internal:
    runtime:
      default: otel
      filebeat.filestream: process
inputs:
- id: filestream-filebeat
  type: filestream
  paths:
  - /tmp/nonexistent1
  prospector.scanner.fingerprint.enabled: false
  file_identity.native: ~
  use_output: default
- data_stream:
    namespace: default
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  type: system/metrics
  use_output: default
outputs:
  default:
    username: elastic
    password:
    hosts:
    - 127.0.0.1:9200
    type: elasticsearch
```

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/5567

### DIFF
--- a/elastic-agent.reference.yml
+++ b/elastic-agent.reference.yml
@@ -371,6 +371,7 @@ agent.logging.to_stderr: true
 # Advanced internal settings. No stability guarantees are provided. The options in this section may be removed at any
 # point without warning.
 # agent.internal:
+
 # Providers
 
 # Providers supply the key/values pairs that are used for variable substitution


swiatekm:
>     1. Let's document `agent.internal` in elastic-agent.reference.yml but not any of it's parameters. Just have the commented out section that says something like "Advanced internal and development settings. No guarantee of stability is provided".

Done.

>     2. Is there a test the ensures that the `agent.internal.runtime` settings don't override the `agent.monitoring._runtime_experimental` settings explicitly? Or is this guaranteed by the precedence order of the two? This was one of things I manually tested and it does what we want `agent.internal.runtime.default: process` does not force the agent to also use the process runtime for monitoring, that's controlled independently.

There's no explicit test for this, but there is a unit test verifying that `_runtime_experimental` set directly on an input overrides the settings in `agent.internal.runtime`, and that's what the monitoring section does. I'd like to refactor this in a follow-up to be less convoluted, but the current logic does what we want.

_meta/config/common.reference.p2.yml.tmpl
```diff
@@ -362,4 +362,7 @@ agent.logging.to_stderr: true
   # file. Defaults to false.
   # rotateonstartup: false
 
+# Advanced internal settings. No stability guarantees are provided. The options in this section may be removed at any
+# point without warning.
+# agent.internal:
```
cmacknz, line 367:
```suggestion
# agent.internal:

```

Should be a space here so it doesn't run into the providers section.

Issue #11169, opened 2025-11-13
[beats receivers]  host parsing failed for prometheus-collector: error parsing URL: parse "http://localhost:EDOT_COLLECTOR_METRICS_PORT": invalid port ":EDOT_COLLECTOR_METRICS_PORT" after host


cmacknz:
- Version: 9.2.1
- Operating System: Windows Server 2019 Standard

This was supposed to be fixed in https://github.com/elastic/elastic-agent/pull/10819 but we have a bug report that this is still happening in 9.2.1 now that the otel runtime is the default for self-monitoring in 9.2.1

Relevant section of state.yaml in the provided diagnostics:

```yaml
fleet_message: Connected
fleet_state: 2
log_level: info
message: 1 or more components/units in a failed state
state: 3
components:
    # ...
    - id: prometheus/metrics-monitoring
      state:
        component:
            apmconfig: null
            limits:
                gomaxprocs: 0
                source:
                    fields:
                        go_max_procs:
                            kind:
                                numbervalue: 0
        component_idx: 2
        features_idx: 2
        message: 'Healthy: communicating with pid ''10484'''
        pid: 0
        state: 2
        units:
            input-prometheus/metrics-monitoring-metrics-monitoring-collector:
                message: 'host parsing failed for prometheus-collector: error parsing URL: parse "http://localhost:EDOT_COLLECTOR_METRICS_PORT": invalid port ":EDOT_COLLECTOR_METRICS_PORT" after host'
                payload:
                    streams:
                        metrics-monitoring-collector:
                            error: ""
                            status: STARTING
                state: 4
            output-prometheus/metrics-monitoring:
                message: Healthy
                state: 2
        version_info:
            build_hash: 6485e11edf8854d7792dfc2999bf19db37315ea4
            meta:
                build_time: 2025-11-06 08:49:58 +0000 UTC
                commit: 6485e11edf8854d7792dfc2999bf19db37315ea4
            name: beat-v2-client
```

Interestingly the `prometheus/metrics-monitoring` input here is running as a sub-process. I think this is likely because we fell back to using the process runtime  because the user is using the Kafka output:

```
agent:
    download:
        sourceURI: https://artifactory.skoda.vwgroup.com/artifactory/chck-elastic-external-generic-local/agents/
    features: null
    monitoring:
        enabled: true
        logs: true
        metrics: true
        traces: false
        use_output: 60ada9f0-7907-11ee-bc53-e16d9f2c55ad
# ...
outputs:
[...]

Issue #10220, opened 2025-09-30
[beats receivers] Replace OTel collector internal telemetry monitoring with a scalable approach


faec:
The [initial implementation](https://github.com/elastic/elastic-agent/pull/9928) for ingesting OTel collector internal telemetry has some major limitations:

- Conversion to Beats-compatible fields for the Agent dashboards is done with a hard-coded Javascript processor, which is not very maintainable or testable.
- Basing the telemetry scrape on Metricbeat's prometheus input means that we can't assume all relevant metrics will be visible in the same event, since this input partitions fields by their label set (and the labels for the relevant fields vary significantly and are not guaranteed to be stable). This rules out support for some important metrics like `output.events.active` that require aggregating data from multiple collector telemetry variables. It also means we have questionable mitigations like mangling the metricbeat id so Elasticsearch doesn't reject events with disjoint variable sets as duplicates just because they have the same timestamp and source metadata.
- Some aspects of the config generation and field conversion only work for the monitoring case. Extending them to support the general case would significantly increase the complexity, and might be entirely infeasible with this approach.
- Less severe but still undesirable: this approach requires fetching the data over an open TCP port, even when running in the same process as the collector.

A sustainable solution would have the following attributes:
- Any necessary field conversions can access and logically depend on the full set of Collector telemetry fields (this requires at minimum a custom scraper for the data in Agent and/or Beats instead of adapting the existing Prometheus input).
- Non-trivial field conversions should be written in unit-tested Go code and updated in sync with the Collector version.
- If/when possible, scrape the data directly through in-process mechanisms, or through a private socket rather than generic TCP port.
    * This requires upstream changes, either to prometheus (https://github.com/prometheus/prometheus/issues/12024) or the collector.

The concrete design should also allow for a viable migration path to pure OTLP metrics, as definitions and dependencies stabilize enough to reliably build dashboards / integration support on that basis.

cmacknz:
Potentially related upstream issue https://github.com/open-telemetry/opentelemetry-collector/issues/4970

Also that monitoring requires configuration changes when dynamic providers is used is also a problem, it would be nice if we could stop requiring this but if not handling that will likely be handled as part of https://github.com/elastic/elastic-agent/issues/10316

PR #11204, opened 2025-11-15
Ensure monitoring the Otel collector never runs in a beat process

swiatekm:
## What does this PR do?

It ensures that the prometheus metrics input we use to monitor the otel collector always runs in said otel collector. The reason this has to be the case is that the input relies on an environment variable we only inject into the otel collector process, and not into the metricbeat process.

The fix is quite hacky, but I think this is acceptable for two reasons:
- We're redoing how this self-monitoring works in 9.3, and the problem will disappear.
- This is a symptom of a broader issue, where the self-monitoring config generation doesn't know which runtime a component will actually run in - just what we want it to run in. The fix for this is more involved and probably shouldn't go into a patch release.

## Why is it important?

If a user has a monitoring output which isn't supported by the otel runtime, their agent will become unhealthy for no good reason.

## Checklist

Build agent locally and run it using the following configuration, which falls back to the process runtime, as the `indices` setting on the elasticsearch output isn't supported for otel:

```yaml
agent:
  logging:
    to_stderr: true
  monitoring:
    enabled: true
inputs:
- data_stream:
    namespace: default
  streams:
  - data_stream:
      dataset: system.cpu
    metricsets:
    - cpu
  type: system/metrics
  use_output: default
outputs:
  default:
    username: elastic
    password: xxx
    hosts:
    - 127.0.0.1:9200
    type: elasticsearch
    indices: []
```

You should see a log line about the `prometheus/metrics-monitoring` output being skipped, and it shouldn't show up in status.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/11169

### DIFF
--- a/testing/integration/ess/beat_receivers_test.go
+++ b/testing/integration/ess/beat_receivers_test.go
@@ -817,6 +817,12 @@ agent.monitoring.enabled: false
 	}
 }
 
+// Log lines TestBeatsReceiverProcessRuntimeFallback checks for
+const (
+	otelRuntimeUnsupportedLogLineStart = "otel runtime is not supported"
+	prometheusInputSkippedLogLine      = "The Otel prometheus metrics monitoring input can't run in a beats process, skipping"
+)
+
 // TestBeatsReceiverProcessRuntimeFallback verifies that we fall back to the process runtime if the otel runtime
 // does not support the requested configuration.
 func TestBeatsReceiverProcessRuntimeFallback(t *testing.T) {
@@ -890,10 +896,10 @@ outputs:
 		}
 
 		if message, ok := logRecord["message"].(string); ok {
-			if strings.HasPrefix(message, "otel runtime is not supported") {
+			if strings.HasPrefix(message, otelRuntimeUnsupportedLogLineStart) {
 				unsupportedLogRecords = append(unsupportedLogRecords, logRecord)
 			}
-			if strings.HasPrefix(message, "The Otel prometheus metrics monitoring input can't run in a beats process, skipping") {
+			if strings.HasPrefix(message, prometheusInputSkippedLogLine) {
 				prometheusUnsupportedLogRecord = logRecord
 			}
 		}


testing/integration/ess/beat_receivers_test.go
```diff
@@ -902,11 +906,8 @@ agent.monitoring.enabled: false
 		}
 	})
 
-	require.NotNil(t, unsupportedLogRecord, "unsupported log message should be present")
-	message, ok := unsupportedLogRecord["message"].(string)
-	require.True(t, ok, "log message field should be a string")
-	expectedMessage := "otel runtime is not supported for component system/metrics-default, switching to process runtime, reason: unsupported configuration for system/metrics-default: error translating config for output: default, unit: system/metrics-default, error: indices is currently not supported: unsupported operation"
-	assert.Equal(t, expectedMessage, message)
+	assert.Len(t, unsupportedLogRecords, 5, "one log line for each component we try to run")
+	assert.NotEmpty(t, prometheusUnsupportedLogRecord, "should get a log line about Otel prometheus metrics input being skipped")
```
cmacknz, line 916: Before this fix the agent in this test should have been unhealthy correct?

That's the thing we are trying to fix, are we able to observe that directly in the test here instead of inferring it via a log line?

internal/pkg/agent/application/coordinator/coordinator.go
```diff
@@ -1817,6 +1817,12 @@ func (c *Coordinator) splitModelBetweenManagers(model *component.Model) (runtime
 		case component.OtelRuntimeManager:
 			otelComponents = append(otelComponents, comp)
 		case component.ProcessRuntimeManager:
+			// Hack to fix https://github.com/elastic/elastic-agent/issues/11169
+			// TODO: Remove this after https://github.com/elastic/elastic-agent/issues/10220 is resolved
+			if comp.ID == "prometheus/metrics-monitoring" {
```
cmacknz, line 1822: Create an issue to remove these lines and make it a sub-issue of https://github.com/elastic/elastic-agent/issues/10220, that will make sure we don't forget to come back here and address this.

swiatekm: https://github.com/elastic/elastic-agent/issues/11230

testing/integration/ess/beat_receivers_test.go
swiatekm: No, self-monitoring was disabled in this test, and now I enabled it. I believe this was done to make the test more narrowly scoped, but it ended up obscuring an actual issue.

What do you want to observe more directly? We're checking that the components are healthy and that they're running as beats processes in spite of the configuration. The log line checks are more to verify we're actually emitting the log lines.

testing/integration/ess/beat_receivers_test.go
```diff
@@ -872,13 +871,14 @@ agent.monitoring.enabled: false
 		status, statusErr := fixture.ExecStatus(ctx)
 		assert.NoError(collect, statusErr)
 		// we should be running beats processes even though the otel runtime was requested
-		assertBeatsHealthy(collect, &status, component.ProcessRuntimeManager, 1)
+		assertBeatsHealthy(collect, &status, component.ProcessRuntimeManager, 4)
```
pchila, line 880: the 4 is the expected number of beats that are running? The change is due to enabling self-monitoring ?

testing/integration/ess/beat_receivers_test.go
```diff
@@ -889,9 +889,13 @@ agent.monitoring.enabled: false
...
 		}
 
-		if message, ok := logRecord["message"].(string); ok && strings.HasPrefix(message, "otel runtime is not supported") {
-			unsupportedLogRecord = logRecord
-			break
+		if message, ok := logRecord["message"].(string); ok {
+			if strings.HasPrefix(message, "otel runtime is not supported") {
+				unsupportedLogRecords = append(unsupportedLogRecords, logRecord)
+			}
+			if strings.HasPrefix(message, "The Otel prometheus metrics monitoring input can't run in a beats process, skipping") {
```
pchila: maybe put the expected log message in a constant ?

pchila (commented):
Couple of questions, nothing major

testing/integration/ess/beat_receivers_test.go
swiatekm: Yes. One normal component and three from self-monitoring. The only thing this PR changes is how self-monitoring runs. It'd be five here without the change.

testing/integration/ess/beat_receivers_test.go
swiatekm: You mean a constant in one of the agent packages and then import it here?

pchila: no just a constant in this file, in order to find it and update it easily should the log change.
There are already enough dependencies from the integration tests to the main agent module, we don't need to add another one

testing/integration/ess/beat_receivers_test.go
cmacknz: I was wondering why the test did not detect this problem originally, which sounds like it was because monitoring was disabled, not that we weren't verifying the agent was unhealthy.

testing/integration/ess/beat_receivers_test.go
swiatekm: Sure, can do.

swiatekm: Done: https://github.com/elastic/elastic-agent/pull/11204/commits/f9294cb4ea6d9c255da98af83ff484b4b78da3c3

Issue #10157, opened 2025-09-25
[Hybrid Agent] Collector status reported to Fleet does not specify a unit type


cmacknz:
- Relates https://github.com/elastic/kibana/pull/236387#issuecomment-3333313047
- Relates https://github.com/elastic/kibana/pull/236387

When we report collector component status from a hybrid agent to Fleet, we do not specify a unit type. Partially this is because the collector has more component types than we currently model, but we could temporarily set receivers as input type units and exporters as output type units to help light up the Fleet UI.

We possibly could set the other component types like processors and connectors as input type units as well, extensions are a bit less clear on what we'd mark them as so possibly we just leave all these as empty.

```yaml
    {
      "id": "pipeline:metrics/otelcol-check-ce8269b6-d720-4b66-9fe7-a3fd05c9683a-otelcol-httpcheck-check",
      "message": "Healthy",
      "status": "HEALTHY",
      "type": "otel",
      "units": [
        {
          "id": "processor:transform/otelcol-check-ce8269b6-d720-4b66-9fe7-a3fd05c9683a-otelcol-httpcheck-check-routing",
          "message": "Healthy",
          "status": "HEALTHY",
          "type": ""
        },
        {
          "id": "receiver:httpcheck/otelcol-check-ce8269b6-d720-4b66-9fe7-a3fd05c9683a-otelcol-httpcheck-check",
          "message": "Healthy",
          "status": "HEALTHY",
          "type": ""
        },
        {
          "id": "connector:forward",
          "message": "Healthy",
          "status": "HEALTHY",
          "type": ""
        }
      ]
    },
    {
      "id": "pipeline:metrics",
      "message": "Healthy",
      "status": "HEALTHY",
      "type": "otel",
      "units": [
        {
          "id": "exporter:elasticsearch/fleet-elasticsearch-output",
          "message": "Healthy",
          "status": "HEALTHY",
          "type": ""
        },
        {
          "id": "connector:forward",
          "message": "Healthy",
          "status": "HEALTHY",
          "type": ""
        }
      ]
    }
```

The translation from component status to Fleet status happens below and one piece of complexity is the componentstatus API used does not expose the component kind.

https://github.com/elastic/elastic-agent/blob/cb9c826c8e66e75abdf50cbee635c61538b79f23/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L317-L326

ebeahan:
[...]

PR #11233, opened 2025-11-18
Set correct types for otel components in Fleet status

swiatekm:
## What does this PR do?

It ensures otel collector component status reported to Fleet has the right component type sets for receivers and exporters. Respectively, we have `exporter -> output` and `receiver -> input`.

It also does some internal refactoring. The Fleet status translation code was using its own `otelhelpers` package. I moved the contents of that package to `otel/translate` and made all the status translation code use it, unifying the logic between Fleet status and local status. This improves some of the error messages for local status as a side effect.

I've also added unit tests for the Fleet status translation.

## Why is it important?

Fleet knows about `input` and `output` component types, so it can now display status for native otel integrations.

## Checklist

This is slightly involved, as we don't have any native Otel integrations with actual otel collector configurations.

1. Build agent locally from this PR.
2. Build one of the proposed native otel integration locally. I checked out https://github.com/elastic/integrations/pull/15739 and ran `elastic-package build --zip` to generate the integration archive.
3. Upload the integration archive to Kibana.
4. Create an agent policy using the integration.
5. Start an agent locally from the built binary and enroll it into the policy.

You should see the integration show up in the UI:
Screenshot_20251117_173604

You can also look at the agent document to see the types set in the status.

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/10157

### DIFF
--- a/internal/pkg/otel/translate/status_test.go
+++ b/internal/pkg/otel/translate/status_test.go
@@ -183,7 +183,7 @@ func TestGetAllComponentState(t *testing.T) {
 						Units: map[runtime.ComponentUnitKey]runtime.ComponentUnitState{
 							runtime.ComponentUnitKey{UnitID: "filestream-unit", UnitType: client.UnitTypeInput}: {
 								State:   client.UnitStateStarting,
-								Message: "STARTING",
+								Message: "Starting",
 								Payload: map[string]any{
 									"streams": map[string]map[string]string{
 										"test-1": {
@@ -199,7 +199,7 @@ func TestGetAllComponentState(t *testing.T) {
 							},
 							runtime.ComponentUnitKey{UnitID: "filestream-default", UnitType: client.UnitTypeOutput}: {
 								State:   client.UnitStateStarting,
-								Message: "STARTING",
+								Message: "Starting",
 							},
 						},
 						VersionInfo: runtime.ComponentVersionInfo{


internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -318,8 +334,10 @@ func (f *FleetGateway) convertToCheckinComponents(components []runtime.Component
 	// OTel status is placed as a component for each top-level component in OTel
 	// and each subcomponent is a unit.
 	if collector != nil {
-		for id, item := range collector.ComponentStatusMap {
-			state, msg := otelhelpers.StateWithMessage(item)
+		sortedIds := slices.Sorted(maps.Keys(collector.ComponentStatusMap))
```
pchila: why do we need to sort the keys of the `collector.ComponentStatusMap` ? Especially since we then re-access the map by key ?
Same questions for the units below)

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
```diff
@@ -281,6 +283,20 @@ func (f *FleetGateway) convertToCheckinComponents(components []runtime.Component
 		return ""
 	}
 
+	otelComponentTypeString := func(componentStatusId string) string {
+		kind, _, err := translate.ParseEntityStatusId(componentStatusId)
+		if err != nil {
+			return ""
```
pchila, line 290: Can we get at least a log when we have issues parsing the kind of the otel component?
The output of this function is set directly in  `CheckinUnit.Type` a few lines lower in the file

swiatekm: Sure. This won't happen unless there's a bug in the otel healthcheck extension, so it shouldn't cause unnecessary log spam.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
swiatekm: The function takes a map as an input and returns a slice. Having the output order be consistent is a good practice in general, but here in particular I did it to make testing easier. testify has better diffs for ordered slices.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
swiatekm: Done: https://github.com/elastic/elastic-agent/pull/11233/commits/f32de378f7136816ad46e33ccffadd80d58ca7a1

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
pchila: If it's only for test assertions there's https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch that would still work without requiring sorting.
Having the output order consistent may not necessarily be a good practice, if it's not guaranteed by contract but consumer come to rely on the order (moreover if the output get marshaled/unmarshaled as an object order is not guaranteed once more).

swiatekm: `ElementsMatch` produces very unhelpful diffs. I originally had it there and got rid of it because I couldn't figure out what the problem was from the error message. I can revert it if you feel strongly about it, it's not that important.

cmacknz:
Thanks, I created https://github.com/elastic/kibana/issues/243720 to follow up on trying to report the receiver name instead of just `otelcol` which is what is defined in the package. This might require another tweak on the agent side but we'll deal with that once we agree what the best way to get more specific component names is.

internal/pkg/agent/application/gateway/fleet/fleet_gateway.go
pchila: I would like to avoid sorting components at each checkin in order to have nicer assertions in test.
If we need to sort to have nice diffs in test, let's sort in test and write a custom assertion there.

swiatekm: Allright, done: https://github.com/elastic/elastic-agent/pull/11233/commits/d1d65952ea142d2ecade6e7cb588789f599e1278.

cmacknz:
FYI the Fleet side has iterated on how this gets displayed to make the affected receiver more obvious: https://github.com/elastic/kibana/pull/244013

Issue #11171, opened 2025-11-13
[beats receivers] OTel manager failed: supervised collector exited with error: Error response from daemon: Sign in to continue using Docker Desktop


cmacknz:
- Version: 9.2.1
- Operating System: MacOS(15.7.2)

The supervised collector fails to start on MacOS (or presumably any system relying on Docker Desktop) if you are not signed in to Docker Desktop and signing in is mandatory (as it is at Elastic).

Relevant parts of state.yaml, where only `system/metrics-monitoring` appears even though we are attempting to start all of the monitoring beats. 

```yaml
fleet_message: Connected
fleet_state: 2
log_level: info
message: 'OTel manager failed: supervised collector (pid: 8109) exited with error: exit status 1'
state: 4
# ...
components:
    # This is the only monitoring beat with state reported in state.yaml
    - id: system/metrics-monitoring
      state:
        component:
            apmconfig: null
            limits:
                gomaxprocs: 0
                source:
                    fields:
                        go_max_procs:
                            kind:
                                numbervalue: 0
        component_idx: 0
        features_idx: 0
        message: 'Starting: spawned pid ''8140'''
        pid: 0
        state: 0
        units:
            input-system/metrics-monitoring-metrics-monitoring-endpoint_security:
                message: 'Starting: spawned pid ''8140'''
                state: 0
            output-system/metrics-monitoring:
                message: 'Starting: spawned pid ''8140'''
                state: 0
        version_info:
            build_hash: ""
            name: ""
```

Relevant logs where the primary error is `Exiting: error initializing processors: failed to start watcher: Error response from daemon: Sign in to continue using Docker Desktop. Membership in the [elastic] organization is required. Sign in enforced by your administrators (via registry.json)`

```json
{"log.level":"info","@timestamp":"2025-11-13T10:28:04.813Z","message":"Output reload is enabled, the beat will restart as needed on change of output config","component":{"binary":"metricbeat","dataset":"elastic_agent.metricbeat","id":"system/metrics-default","type":"system/metrics"},"log":{"source":"system/metrics-default"},"log.logger":"centralmgmt.V2-manager.centralmgmt","log.origin":{"file.line":203,"file.name":"management/managerV2.go","function":"github.com/elastic/beats/v7/x-pack/libbeat/management.NewV2AgentManagerWithClient"},"service.name":"metricbeat","ecs.version":"1.6.0","ecs.version":"1.6.0"}
[...]

PR #11234, opened 2025-11-18
[otelmanager] emit starting state in the beginning 

VihasMakwana:
## What does this PR do?

This PR introduces support for emitting a STARTING state when the collector is expect the collector to start. 

## Why is it important?

Right now, we default to a `STOPPED` state whenever an error occurs while accessing the healthcheck port. As a result, the `elastic-agent status` output does not show any monitoring components. If this error occurs consistently every time the collector starts, those components will never appear in `elastic-agent status`.

## Checklist

For mac,
1. Build agent from this branch
2. Sign out of docker desktop.
3. Install and enroll the agent into fleet.
4. Notice the status. You'll first see `Starting` state, then `Failed` state with message `OTel manager failed ... process exited with status 1`.
 
## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/11171

### DIFF
--- a/internal/pkg/otel/translate/status_test.go
+++ b/internal/pkg/otel/translate/status_test.go
@@ -167,6 +167,53 @@ func TestGetAllComponentState(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:       "component state starting",
+			components: []component.Component{fileStreamOtelComponent},
+			otelStatus: &status.AggregateStatus{
+				Event:              componentstatus.NewEvent(componentstatus.StatusStarting),
+				ComponentStatusMap: map[string]*status.AggregateStatus{},
+			},
+			expected: []runtime.ComponentComponentState{
+				{
+					Component: fileStreamOtelComponent,
+					State: runtime.ComponentState{
+						State:   client.UnitStateStarting,
+						Message: "STARTING",
+						Units: map[runtime.ComponentUnitKey]runtime.ComponentUnitState{
+							runtime.ComponentUnitKey{UnitID: "filestream-unit", UnitType: client.UnitTypeInput}: {
+								State:   client.UnitStateStarting,
+								Message: "STARTING",
+								Payload: map[string]any{
+									"streams": map[string]map[string]string{
+										"test-1": {
+											"error":  "",
+											"status": client.UnitStateStarting.String(),
+										},
+										"test-2": {
+											"error":  "",
+											"status": client.UnitStateStarting.String(),
+										},
+									},
+								},
+							},
+							runtime.ComponentUnitKey{UnitID: "filestream-default", UnitType: client.UnitTypeOutput}: {
+								State:   client.UnitStateStarting,
+								Message: "STARTING",
+							},
+						},
+						VersionInfo: runtime.ComponentVersionInfo{
+							Name:      OtelComponentName,
+							BuildHash: version.Commit(),
+							Meta: map[string]string{
+								"build_time": version.BuildTime().String(),
+								"commit":     version.Commit(),
+							},
+						},
+					},
+				},
+			},
+		},
 	}
 
 	for _, test := range tests {


internal/pkg/otel/translate/status.go
```diff
@@ -39,6 +41,15 @@ func GetAllComponentStates(otelStatus *status.AggregateStatus, components []comp
 				if compState, statusErr = getComponentState(pipelineStatus, comp); statusErr != nil {
 					return nil, statusErr
 				}
+			} else if isHealthcheckUnavailable(otelStatus) {
+				// healthcheck extension is not yet started. We return a STARTING state.
+				compState = runtime.ComponentComponentState{
+					Component: comp,
+					State: runtime.ComponentState{
+						State:   client.UnitStateStarting,
+						Message: "Starting",
```
VihasMakwana: We can emit the actual error as the `Message`. Right now, we just log it at debug level. wdyt?

VihasMakwana:
CI failing due to known issue. It will be fixed via https://github.com/elastic/elastic-agent/pull/11238

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -171,6 +171,13 @@ func (r *subprocessExecution) startCollector(ctx context.Context, baseLogger *lo
 					// after the collector exits, we need to report a nil status
 					r.reportSubprocessCollectorStatus(ctx, statusCh, nil)
 					return
+				default:
+					// if we face any other error (most likely, connection refused), emit a status with the error.
+					logger.Debugf("Received an unexpected error: %v. Emitting empty status", err)
+					status := &status.AggregateStatus{
+						Event: componentstatus.NewEvent(componentstatus.StatusNone, componentstatus.WithError(err)),
```
swiatekm: I'm somewhat suspicious of this change. This can potentially happen at any point, not just after starting the collector. Will we emit the right status in that case as well? Intuitively, I'd expect that if we can't get the status, and the collector is still running, we should return an error rather than emitting a fake status.

If we just want to have a `STARTING` status while the collector starts, let's emit that directly.

swiatekm (commented):
I'm not sure I like this approach. 

If we need to always emit a STARTING state, then we should do so explicitly, without waiting for the health check first. And if there's an error getting the status, we should set it as an error on the collector instead of doing this. Finally, if some errors from the collector need to be turned into a specific status for all the components, then that should happen separately somewhere in the otel manager.

Does that make sense?

internal/pkg/otel/manager/execution_subprocess.go
VihasMakwana: I understand what you mean.
What if we move this piece of code at the beginning of `startCollector`?
i.e. we emit starting state once and keep emiting statuses based on healthcheck's respnse.

Correct?

swiatekm: If the expectation is that we always should have a STARTING status even if the collector doesn't actually start, then yes. @cmacknz does that sound right to you?

cmacknz: STARTING can be used to represent a component that is expected to exist but doesn't yet. The error case should be FAILED.

We potentially should have introduced an UNKNOWN to represent "should exist and can't read it's status" but it's not worth adding a new state at this time IMO.

VihasMakwana: Makes sense. For now, I'll keep it `STARTING` because we expect it to start.

internal/pkg/otel/translate/status.go
VihasMakwana: Not applicable anymore.

internal/pkg/otel/manager/execution_subprocess.go
```diff
@@ -171,6 +171,9 @@ func (r *subprocessExecution) startCollector(ctx context.Context, baseLogger *lo
 					// after the collector exits, we need to report a nil status
 					r.reportSubprocessCollectorStatus(ctx, statusCh, nil)
 					return
+				default:
+					// if we face any other error (most likely, connection refused), log the error.
+					logger.Debugf("Received an unexpected error: %v", err)
```
VihasMakwana: I'm keeping this log as is because it might be helpful. I'm keeping this at debug level because the health check can be very noisy (we poll every `1s`)

internal/pkg/otel/manager/execution_subprocess.go
cmacknz: I think in the error case, we would reflect the collector error into each affected component, I added some examples in https://github.com/elastic/elastic-agent/issues/11171#issuecomment-3554264867

VihasMakwana:
The CI has been failing but the test passes on my Mac consistently. Weird. 
I've given it a retry.

VihasMakwana:
We already emit a `STARTING` state here: https://github.com/elastic/elastic-agent/blob/2b2dff737441ba5c931b91ebe91f0b0586767433/internal/pkg/otel/manager/execution_subprocess.go#L154-L155

We just need to handle that case while agent translation.

internal/pkg/otel/translate/status.go
```diff
@@ -39,6 +39,8 @@ func GetAllComponentStates(otelStatus *status.AggregateStatus, components []comp
 				if compState, statusErr = getComponentState(pipelineStatus, comp); statusErr != nil {
 					return nil, statusErr
 				}
+			} else if otelStatus != nil && otelStatus.Event != nil && otelStatus.Status() == componentstatus.StatusStarting {
+				compState = getComponentStartingState(comp)
```
swiatekm, line 43: We should have a unit test for this behavior in the translate package as well imo.

VihasMakwana: Done

swiatekm (approved):

PR #11236, opened 2025-11-18
Fix cleanup manual rollback

pchila:
## What does this PR do?
This PR fixes the relative path calculation for agent versioned homes to preserve introduced with https://github.com/elastic/elastic-agent/pull/9643.
Removes ambiguity in what a path to a versioned home to preserve should look like: it should always be `data/elastic-agent-<variable unique version part here>`.

Adds a few checks and resolves the relative path when calling rollback from command line.

## Why is it important?
To preserve elastic-agent versioned home directories correctly during cleanup.

## Checklist

## Related issues

- Relates #9643

## Questions to ask yourself

- ...

### DIFF
--- a/internal/pkg/agent/application/upgrade/rollback.go
+++ b/internal/pkg/agent/application/upgrade/rollback.go
@@ -177,14 +177,15 @@ func cleanup(log *logger.Logger, topDirPath string, removeMarker, keepLogs bool,
 	log.Infof("versioned homes to keep: %v", versionedHomesToKeep)
 
 	var cumulativeError error
-	relativeHomePaths := make([]string, len(versionedHomesToKeep))
-	for i, h := range versionedHomesToKeep {
+	relativeHomePaths := make([]string, 0, len(versionedHomesToKeep))
+	for _, h := range versionedHomesToKeep {
 		relHomePath, err := filepath.Rel(dataDirPath, filepath.Join(topDirPath, h))
 		if err != nil {
 			cumulativeError = goerrors.Join(cumulativeError, fmt.Errorf("extracting elastic-agent path relative to data directory from %s: %w", h, err))
-			continue
+			// best effort: try to use the entry as-is, without calculating the path relative to `data`
+			relHomePath = h
 		}
-		relativeHomePaths[i] = relHomePath
+		relativeHomePaths = append(relativeHomePaths, relHomePath)
 	}
 
 	log.Infof("Starting cleanup of versioned homes. Keeping: %v", relativeHomePaths)


internal/pkg/agent/application/upgrade/rollback.go
```diff
@@ -174,16 +174,21 @@ func cleanup(log *logger.Logger, topDirPath string, removeMarker, keepLogs bool,
...
+
+	var cumulativeError error
 	relativeHomePaths := make([]string, len(versionedHomesToKeep))
 	for i, h := range versionedHomesToKeep {
-		relHomePath, err := filepath.Rel("data", h)
+		relHomePath, err := filepath.Rel(dataDirPath, filepath.Join(topDirPath, h))
 		if err != nil {
-			return fmt.Errorf("extracting elastic-agent path relative to data directory from %s: %w", h, err)
+			cumulativeError = goerrors.Join(cumulativeError, fmt.Errorf("extracting elastic-agent path relative to data directory from %s: %w", h, err))
+			continue
```
ycombinator: Looks like we were failing fast in the cleanup before but we are no longer doing that.  Why is that?

internal/pkg/agent/cmd/watch.go
```diff
@@ -320,6 +333,13 @@ func rollback(log *logp.Logger, topDir string, client client.Client, installModi
 	// FIXME get the hash from the list of installs or the manifest or the versioned home
 	// This is only a placeholder in case there is no versionedHome defined (which we always have)
 	hash := ""
+	if filepath.IsAbs(versionedHome) {
+		// if the versioned home is an absolute path we need
```
ycombinator: Looks like an incomplete comment?

internal/pkg/agent/application/upgrade/rollback.go
pchila: This error can only happen if the watcher is passed some wrong paths in the list of versionedHomes to preserve during cleanup phase. 
If such an error occurs:
- (failing fast) -> Cleanup stops entirely, watcher exits, leaving obsolete installs on disk
- (accumulating the error) -> Cleanup keeps going and the entry causing the error is ignored, cleaning up everything that does not match entries in `relativeHomePaths`.

I was thinking if this should be a show stopper or if the watcher cleanup should keep going in best effort mode, I switched to accumulating the error and keep going but the previous behaviour is not wrong, I guess it depends on expectations.

What do you think should be the correct approach here ?

internal/pkg/agent/cmd/watch.go
pchila: yes, yes it is 😅 
Fixed in 18c8b7951788e66add007908970d8d7b9d16a96f

internal/pkg/agent/application/upgrade/rollback.go
ycombinator: Thanks for the explanation. I agree it's better to accumulate the errors and do a best effort cleanup.  

One minor change you might want to make with this approach is to not initialize the size of `relativeHomePaths` to `len(versionedHomesToKeep)` as its possible due to the `continue` here that the `relativeHomePaths` slice may end up containing fewer elements than its initial size. Not a showstopper though as we're probably looking at single-digit sizes most of the time.

pchila: The versionedHomes slice containes the versioned homes we want to keep relative to `.../data` directory.
This is then used as a [filter on line 193](https://github.com/elastic/elastic-agent/pull/11236/files/ca32577c0cfdda7fccaf4f48c3d8bb5183e8eddb#diff-3a6dbae8cacbb8a618f7ac9a68c7b7a12e9074ddf5e6e337b24836cb3e5714cbR193) to determine if a given versioned home should be preserved during cleanup.

Leaving some indices as `""` empty strings in case of errors will not impact the ability to preserve the other paths: for example,
if we have `.../data/elastic-agent-1` `.../someotherdir/elastic-agent-2` as `versionedHomesToKeep` we will end up with `[]string{"elastic-agent-1", ""}` slice. It's true that we have a small penalty for that element with a default value when calling `slices.Contains()`. We could work around that using `len(versionedHomesToKeep)` as capacity and appending instead of setting values in given indices.

pchila: Prepared a new version of the cleanup loop here 41d1b6a86d6fc97d16f06f58fd93d703f47c3689

